PR #85, opened 2018-12-31
copy stat to wasm and support windows

xmclark:
The function that copies stat into wasm memory does not easily account for different sized stat structs. This is the case with windows. This PR refactors this function to support windows.

### DIFF
--- a/src/apis/emscripten/utils.rs
+++ b/src/apis/emscripten/utils.rs
@@ -102,9 +102,7 @@ pub struct GuestStat {
     st_gid: u32,
     st_rdev: u64,
     st_size: u32,
-    #[cfg(not(target_os = "windows"))]
     st_blksize: u32,
-    #[cfg(not(target_os = "windows"))]
     st_blocks: u32,
     st_atime: u64,
     st_mtime: u64,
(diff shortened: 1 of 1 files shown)


src/apis/emscripten/utils.rs
```diff
@@ -92,29 +92,43 @@ pub unsafe fn copy_terminated_array_of_cstrs(
...
+pub struct GuestStat {
+    st_dev: u64,
+    st_ino: u32,
+    st_mode: u32,
+    st_nlink: u32,
+    st_uid: u32,
+    st_gid: u32,
+    st_rdev: u64,
+    st_size: u32,
+    #[cfg(not(target_os = "windows"))]
```
lachlansneff: The wasm representation doesn't change if the host is running on linux, darwin, or windows. It's invalid to change the behavior depending on the host environment.

xmclark: good feedback, I will default them! Thanks guys!

lachlansneff:
I don't think this is still setting all the items that it did before. Even if an item is zero, this should still set it.

xmclark:
Thanks @lachlansneff for the attention to detail and pointing out that the `struct stat` used in emscripten has extra fields that do not exist in the man page. I will be updating the PR shortly.

xmclark:
I updated the size of the first field. There was also a discrepancy for the last field (st_ino), but I think the correct size is u64 based on the reference struct, so I left it alone.

I used this struct as a reference:
https://github.com/kripken/emscripten/blob/6e4b98636618989bcd99308391e51aa1b81f4c61/system/lib/libc/musl/arch/emscripten/bits/stat.h#L4

PR #86, opened 2018-12-31
[WIP] Isolate WebAssembly program state from native code.

losfair:
TODO:

- [ ] Run WebAssembly code on its own stack
- [ ] Implement new memory layout for WebAssembly program: [linear memory] : [guard area] : [stack]

### DIFF
--- a/src/isolation/mod.rs
+++ b/src/isolation/mod.rs
@@ -1,0 +1,1 @@
+mod stack;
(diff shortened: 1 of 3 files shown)


lachlansneff:
This is a good idea. I believe fastly does something similar with the memory, guard, stack layout. Do keep in mind that it will soon be valid to have multiple instances that all use the same memory however.

PR #87, opened 2018-12-31
use libc clock constants and add macos build

xmclark:
This PR fixes a bug introduced that would cause nginx to fail at runtime. 

Support for the monotonic clock id is added because the compiled nginx binary that is checked in requires it. Nginx uses a [compile-time flag](https://github.com/nginx/nginx/blob/4bf4650f2f10f7bbacfe7a33da744f18951d416d/src/core/ngx_times.c#L196). Windows does not always have these clock id constants, so they are polyfilled. 

The constants in `libc` are used because the platforms do not all use the same conventions. We can't assume the constants values.

A super simple integration test is added that runs nginx. This has room for improvement.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,9 @@ jobs:
           keys:
             - v4-test-cargo-cache-linux-{{ arch }}-{{ checksum "Cargo.lock" }}
       - run: sudo apt-get install -y cmake
+      - run:
+          name: Run nginx
+          command:  cargo run -- run examples/nginx/nginx.wasm -- -v
       - run: make test
       - run: rustup component add rustfmt
       - run: make lint
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,39 @@ jobs:
             - target/debug/deps
           key: v4-test-cargo-cache-linux-{{ arch }}-{{ checksum "Cargo.lock" }}
 
+  test-macos:
+    macos:
+      xcode: "9.0"
+
+    steps:
+      - checkout
+      - restore_cache:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -11,6 +11,9 @@ jobs:
           keys:
             - v4-test-cargo-cache-linux-{{ arch }}-{{ checksum "Cargo.lock" }}
       - run: sudo apt-get install -y cmake
+      - run:
+          name: Run nginx
+          command:  cargo run -- run examples/nginx/nginx.wasm -- -v
```
syrusakbary: This will ouptut the nginx version: `nginx version: nginx/1.15.3`
Is this command using the clock contants under the hood? I tested in current `master` and it's running properly. However running `cargo run -- run examples/nginx/nginx.wasm -- -p examples/nginx -c nginx.conf` fails.

I guess we can either remove this integration test for now or make it in a way that really spins up the nginx server.

syrusakbary:
I have a comment about the integration test, but other than that it looks great!

.circleci/config.yml
xmclark: yeah I would like it to actually run the server! Do you have ideas on how to run the server, then exit gracefully? I think a shell script may work, but I'm open to other ideas.

xmclark: I am going to add integration tests in a different PR...

syrusakbary: Sounds good!

syrusakbary:
Before merging, there are some issues for building in mac. But once they are fixed (tests passing) it should be good to merge

PR #88, opened 2019-01-06
Generic implementation of emscripten data

bjfish:
This refactor attempts to make the implementation of `EmscriptenData` more generic in that a trait `ModuleEnvironment` is provided to allow an environment (imports + backing data) to be implemented in a more generic way. Also, environments are stored in the instance polymorphically as an enum `Vec<InstanceEnvironment>`.

Another goal here was to allow the possibility of de/activating multiple environments in the future. For example, running with emscripten, spectest, and debug imports at the same time.

One tradeoff was made with regards to the indexing of instance environments data as shown here: `instance.environments[EMSCRIPTEN_INSTANCE_INDEX]` which assumes this only this type of data to be at this index `0`.  This allows for faster access versus looking up via a hashmap but something more desirable should be investigated in the future. Any thoughts to make this better? Or should this be acceptable for now?


lachlansneff:
We could also make it a statically-sized array instead of a vec in order to make it possible for the compiler to avoid emitting bound-checks when indexing it.

lachlansneff:
What are the advantages of using an array of here instead of individual fields?

syrusakbary:
Feedback: I don't like too much the idea of multiple environments per instance.
I'm not sure what case we are trying to fulfil since -in general- the guest data is always going to be strictly tied to the ABI used by the module (emscripten, go?, ...).
And the modules only will have one ABI... therefore only one kind of guest data.

Perhaps I might be missing something there?

bjfish:
@syrusakbary What about adding additional imports and guest data? For example, if someone wants to run emscripten plus additional imports with guest data. Maybe Environment is the wrong naming for this but the concept is a way to package imports and guest data.

bjfish:
@syrusakbary I’ll update this to hold just one environment for now. If there is a need for multiple environments, that can be reviewed in the future.

bjfish:
@syrusakbary I've updated this to a single environment

PR #90, opened 2019-01-11
Moved emscripten into it's own lib

syrusakbary:
Moved emscripten into it's own lib


lib/emscripten/build/mod.rs
```diff
@@ -0,0 +1,11 @@
+use std::env;
+
+mod emtests;
+
+static SPECTESTS_ENV_VAR: &str = "WASM_EMSCRIPTEN_GENERATE_EMTESTS";
```
bjfish: typo `EMTESTS_ENV_VAR`

syrusakbary: Good catch. Just pushed a fix

PR #91, opened 2019-01-11
Improvement/nginx integraton test

xmclark:
Add an integration test for running nginx with wasmer.

This is a brute-force, straight-forward solution. A shell script that starts the nginx server, and then curls it.


tests/integration_tests.rs
```diff
@@ -0,0 +1,45 @@
...
+use std::fmt;
+use std::env;
+use std::fs::remove_dir_all;
+use std::path::PathBuf;
+use std::time::Duration;
+use std::{thread, time};
+
+
+#[test]
+fn test_nginx() {
```
syrusakbary: I think for the integrations simpler might be easier to do it directly with a shell script that can run in CircleCI.
(rather than a rust test itself, that we need to compile first to execute later).

Thoughts?

Makefile
```diff
@@ -28,7 +28,7 @@ precommit: lint test
 
 test:
 	# We use one thread so the emscripten stdouts doesn't collide
-	cargo test -- --test-threads=1 $(runargs) 
+	cargo test test_nginx # -- --test-threads=1 $(runargs)
```
appcypher: I think you committed this one by mistake.

tests/integration_tests.rs
xmclark: I went ahead and switched it to a shell script.

Makefile
xmclark: I went ahead and switched it to a shell script. This should be reverted now :)

PR #92, opened 2019-01-11
[WIP] Fix Emscripten Integration in New Runtime

appcypher:
- [x] Fix broken emscripten imports
- [ ] Fix other emscripten binding issues
- [ ] Run nginx and lua on new runtime successfully
- [ ] Fix emtests


lib/emscripten/src/lib.rs
```diff
@@ -104,1116 +112,1235 @@ macro_rules! mock_external {
     }};
 }
 
+struct EmscriptenGlobals {
+    pub data: Vec<(String, LocalGlobal, GlobalDesc)>,
+}
+
+impl EmscriptenGlobals {
```
syrusakbary: I don't think this abstraction is really necessary.

lib/emscripten/src/lib.rs
```diff
@@ -104,1116 +112,1235 @@ macro_rules! mock_external {
...
+
+    // Add globals.
+    for (name, ref global, desc) in em_globals.data {
+        let export = Export::Global {
+            local: global as *mut _,
+            global: desc,
+        };
+
+        imports.register_export("env", name, export);
+    }
```
syrusakbary: We can add globals directly, as we are doing with the other imported functions.
No need to depend in a EmscriptenGlobals abstraction (neither to walk an array at runtime).

appcypher: This impl is not complete, but this is needed because unlike previous Import impl, `Export::Global` hold pointer to the global to instead of moving the global value.

This means we can't create the globals before `generate_emscripten_env` and pass them in. In this case I'm using EmdcriptenGlobals

lib/emscripten/src/lib.rs
appcypher: Responded below

lib/emscripten/src/lib.rs
appcypher: Fixed the issue with previous impl.

.gitignore
```diff
@@ -2,3 +2,4 @@
 **/*.rs.bk
 /artifacts
 .DS_Store
+.idea
```
xmclark: 🙏 😂

lib/emscripten/src/lib.rs
syrusakbary: Can we hold the pointers to static variables instead?

lib/runtime/src/module.rs
```diff
@@ -34,7 +34,7 @@ pub struct ModuleInner {
     pub sig_registry: SigRegistry,
 }
 
-pub struct Module(Rc<ModuleInner>);
+pub struct Module(pub Rc<ModuleInner>);
```
lachlansneff: This shouldn't be public. This whole `Module`/`ModuleInner` stuff is a little weird, but I don't think it should be necessary to expose the `ModuleInner` from a `Module`.

lib/runtime/src/instance.rs
```diff
@@ -242,7 +242,24 @@ impl<'a> Iterator for ExportIter<'a> {
 
 // TODO Remove this later, only needed for compilation till emscripten is updated
 impl Instance {
-    pub fn memory_offset_addr(&self, _index: usize, _offset: usize) -> *const usize {
-        unimplemented!("TODO replace this emscripten stub")
+    pub fn memory_offset_addr(&self, index: usize, offset: usize) -> *const u8 {
```
lachlansneff: Just as a heads up, I'll get rid of this function at some point. Right now, the best replacement is to take `*mut vm::Ctx` as the last argument of the function, get its `local_backing`, and call `.memory(...)` or `.table(...)` on it.

src/bin/wasmer.rs
```diff
@@ -80,12 +81,16 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
...
+
     let import_object = if abi == webassembly::InstanceABI::Emscripten {
-        wasmer_emscripten::generate_emscripten_env()
+        wasmer_emscripten::generate_emscripten_env(&emscripten_globals)
     } else {
-        wasmer_runtime::Imports::new()
+        wasmer_runtime::import::Imports::new()
     };
 
+    let import_object = Rc::new(import_object);
```
lachlansneff: Imports don't use `Rc` anymore, but this is an easy fix once we get the kinks shaken out

lib/runtime/src/instance.rs
appcypher: Alright. Noted

lib/runtime/src/module.rs
appcypher: Yeah. Had to do that because of `memory_offset_addr`, but since you said its going to be scrapped then I can change that back

bjfish:
Merged to continue work in `vm_refactor` branch.

PR #95, opened 2019-01-17
Add data and data finalizer to Ctx

bjfish:


lib/runtime/src/vm.rs
```diff
@@ -28,12 +29,51 @@ pub struct Ctx {
...
+    fn drop(&mut self) {
+        (self.data_finalizer)(self.data);
+    }
 }
 
 impl Ctx {
     pub unsafe fn new(
+        local_backing: &mut LocalBacking,
+        import_backing: &mut ImportBacking
+    ) -> Self {
```
syrusakbary: It might be a good idea to reuse `new_with_data`.

```rust
Self::new_with_data(local_backing, import_backing, ptr::null_mut(), empty_finalizer)
```

Thoughts?

bjfish: @syrusakbary I actually intended this method to be temporary and I forgot to mention that (just trying to prove this concept). The new and new_with_data could be consolidated into new or we could keep new_with_data if you’d like.

syrusakbary: I think we can keep both, but one reusing the other so code is not dup.
However I don't have a strong preference, so... your call! :)

PR #101, opened 2019-01-21
Update README

Terkwood:
Updates the install instruction to work with recent versions of `cargo`.


syrusakbary:
Thanks for the fix @Terkwood 👍

PR #104, opened 2019-01-22
Library restructure

bjfish:
- runtime renamed to runtime-core
- new runtime lib added, re-exporting runtime-core, optionally depends on clif-backend
  - has `compile(wasm: &[u8])` method with default compiler


lib/runtime/src/lib.rs
```diff
@@ -1,51 +1,17 @@
...
-/// Compile a webassembly module using the provided compiler.
-pub fn compile(wasm: &[u8], compiler: &dyn backend::Compiler) -> CompileResult<module::Module> {
-    let token = backend::Token::generate();
-    compiler
-        .compile(wasm, token)
-        .map(|inner| module::Module::new(Rc::new(inner)))
+/// Compiles a WebAssembly module
+pub fn compile(wasm: &[u8]) -> error::CompileResult<module::Module> {
+    #[cfg(not(feature = "wasmer-clif-backend"))]
+    panic!("compile/1 is not available when default compiler is disabled");
```
lachlansneff: This should use `compile_error!("compile/1 is not available when default compiler is disabled")` instead of panic.

lachlansneff: Actually, I'm not sure about that. This would prevent it from even compiling without that feature enabled. Perhaps we should only define the compile function when there's a default compiler?

bjfish: I was thinking they’d use compile/2 and provide compiler parameter if default compiler is disabled.  I’m fine with either error at runtime or conditional inclusion of the method. I thought error at runtime may be more convenient.

bjfish: I was thinking they’d use compile/2 and provide compiler parameter if default compiler is disabled.  I’m fine with either error at runtime or conditional inclusion of the method. I thought error at runtime may be more convenient.

lachlansneff: I changed the name of `compile` in `runtime-core` to `compile_with` and reexported it in `runtime`.

PR #114, opened 2019-01-23
Emscripten update to new runtime

bjfish:
TODO 
- [x] Resolve all imports to run tests (See test failures: `cargo test -p wasmer-emscripten -- --test-threads=1`)
- [x] Initialize EmscriptenData and store in `vmctx.data`
- [x] Update functions using EmscriptenData
- [x] Handle program arguments
- [x] Review all commented out, `unimplemented()`, `TODO`
- [ ] Update branch with master (using new imports syntax)


src/webassembly/mod.rs
```diff
@@ -85,17 +87,128 @@ pub fn run_instance(
...
+    let space_offset = instance
+        .call("_malloc", &[Value::I32((string_len as i32) + 1)])
+        .unwrap();
+
+    let space_offset = match space_offset.as_slice() {
+        &[Value::I32(res)] => Some(res as u32),
+        _ => None,
+    }.unwrap();
+
+    let raw_memory = instance.inner.vmctx.memory(0)[space_offset as usize] as *mut u8;
```
syrusakbary: Rather than doing this, I think would be better to have a `vmctx` function in the instance itself (that points to the inner vmctx).

This way, you can do `instance.vmctx().memory(0)` ...

bjfish: cc: @lachlansneff ?

bjfish: I added a `ctx()` function in the PR.

src/bin/wasmer.rs
```diff
@@ -61,24 +61,29 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
...
     }
 
     let module = webassembly::compile(&wasm_binary[..])
         .map_err(|e| format!("Can't compile module: {:?}", e))?;
 
-    let (_abi, import_object) = if wasmer_emscripten::is_emscripten_module(&module) {
-        let emscripten_globals = wasmer_emscripten::EmscriptenGlobals::new();
+    let (_abi, import_object, em_globals) = if wasmer_emscripten::is_emscripten_module(&module) {
+        let (table_min, table_max) = wasmer_emscripten::get_emscripten_table_size(&module);
+        let (memory_min, memory_max) = wasmer_emscripten::get_emscripten_memory_size(&module);
```
syrusakbary: I think we should refactor this before merging.

lachlansneff:
It looks like this is passing ci. Ignoring the emtests, we should merge it.

Issue #51, opened 2018-12-12
windows support


xmclark:
There seem to be a few barriers to windows support. There is an outstanding bug on [`wabt-rs`](https://github.com/pepyakin/wabt-rs/issues/20) that prevents building on windows. There is a hacky work around, but non-ideal.

The codebase also seems to import many linux-only libraries, so those would have to refactored or conditionally compiled. 

Windows support would be a big win for portability, which I imagine is a priority for a tool that aims to be a cross-platform wasm runtime.

pepyakin:
It is a shame that wabt-rs doesn't work for Windows. Would love to help, but I don't have much experience with Windows nor have I a windows machine.

syrusakbary:
@pepyakin no worries, we will help you on that creating a PR on `wabt-rs` when the time comes :)

@xmclark we just added `Virtual memory allocation on Windows` with #54, so we are one step closer to be on windows :)

xmclark:
@syrusakbary that’s all great news! 

I’ve already started the ball rolling with @pepyakin on [windows support](https://github.com/pepyakin/wabt-rs/pull/31).

I’m also investigating signal support in libc for windows. Seems doable.

xmclark:
I think `libc` should have basic signals support on windows as of [this commit](https://github.com/rust-lang/libc/commit/027d4834615cf8bce3faa9c4a1c2706413a7b275). It may require more tinkering. 

I have spent the last few hours getting familiar with the details of stdio capture on windows. I think this is possible with just `libc` but it may be dirty. Time will tell! I am interested in any ideas for capture on windows.

xmclark:
@syrusakbary I think I have a solution for capturing the stdio, but I can't help feel that capturing stdio in-process seems a little round-about. A problem is windows pipes don't seem to be buffered and that has made it a bit more compilcated. The WinAPI has buffered pipes, but it would be a completely different implementation from the `libc pipe` implementation.

How about storing the tests in a test executable crate and spawning it from integration tests? The standard output can be captured with normal safe rust. There is no platform specific code or unsafe code. I can't imagine the performance is any worse. Thoughts?

syrusakbary:
We added experimental Windows, so it can run WebAssembly files.

There is still some work to do regarding Emscripten (#176) but other than that everything is tested and green 🎉.

We are preparing for Wasmer 0.2.0 and will close this issue once we publish the wasmer.exe file :)

CryZe:
[...]

PR #122, opened 2019-01-25
Update memory mod.rs to include missing mod statement on Windows

JohnDoneth:
Closes https://github.com/wasmerio/wasmer-rust-example/issues/3

Fixes the missing mod statement that exists in the Unix variant of this file. Currently it's blocking any building on Windows.

### DIFF
--- a/lib/runtime-core/src/sys/windows/mod.rs
+++ b/lib/runtime-core/src/sys/windows/mod.rs
@@ -1,1 +1,3 @@
+mod memory;
+
 pub use self::memory::{Memory, Protect};


CryZe:
This doesn't fix building on Windows though. The memory module itself has a ton of errors and once that builds there's even more in the clif-backend.

JohnDoneth:
@CryZe Ah, my apologies, I should have checked.

CryZe:
It seems a lot of new people that came here from the blog post ran into this though. So maybe we could all band together with the guys in #51 to make Windows support happen.

xmclark:
Windows support is crucial for adoption! Expect to see more progress on #51 soon!

lachlansneff:
To expand on @xmclark comment, Windows support is very important to us! There's quite a bit of work to do before it can work, however. I'll lay it out here:

1. Get `sys::Memory` working in Windows.
2. Add signal catching support for Windows. Currently, we use `sigaction`, `setjmp`, and `longjmp` for this.
3. Fix any build/linking issues. We recently made a big leap with that by removing the `libffi` dependency, but there may be additional work to do here.

lachlansneff:
We've just gotten windows support working, so this has been solved. It's currently in a pr, but we expect window support to be available in the 0.2.0 release.

CryZe:
Nice, I'll give it a try tomorrow :D

PR #129, opened 2019-01-29
Consolidate to single debug macro, activate using feature

bjfish:
E.g. activate using `wasmer-runtime-core = { path = "lib/runtime-core", features = ["debug"] }` in `cargo.toml`.


lib/runtime-core/src/macros.rs
```diff
@@ -1,6 +1,15 @@
+#[macro_export]
+#[cfg(feature = "debug")]
+macro_rules! debug {
+    ($fmt:expr) => (println!(concat!("wasmer-runtime(:{})::", $fmt), line!()));
+    ($fmt:expr, $($arg:tt)*) => (println!(concat!("wasmer-runtime(:{})::", $fmt, "\n"), line!(), $($arg)*));
+}
+
+#[macro_export]
+#[cfg(not(feature = "debug"))]
 macro_rules! debug {
```
lachlansneff: Does this create unused variable warnings when compiling without the debug feature?

bjfish: No it does not.

PR #130, opened 2019-01-29
Improve importing things

lachlansneff:
Now, you can import memories, tables, and globals like such:

```rust
let memory = Memory::new(MemoryDesc {
    min: 1,
    max: Some(1),
    shared: false,
})?;

let table = Table::new(TableDesc {
    ty: ElementType::Anyfunc,
    min: 10,
    max: None,
})?;

let global = Global::new(Value::I32(42));

    "env" => {
        "print_i32" => func!(print_num, [i32] -> [i32]),
        "memory" => memory,
        "global" => global,
        "table" => table,
    },
};

```


lib/runtime-core/src/vmcalls.rs
```diff
@@ -11,89 +13,106 @@ use crate::{
...
-        .grow_dynamic(by_pages)
-    {
-        // Store the new size back into the vmctx.
-        (*(*ctx).memories.add(memory_index.index())).size =
-            (old as usize + by_pages as usize) * LinearMemory::PAGE_SIZE as usize;
-        old
+    let local_memory = *ctx.memories.add(memory_index.index());
+    let memory = (*local_memory).memory as *mut DynamicMemory;
+
+    if let Some(old) = (*memory).grow(delta, &mut *local_memory) {
```
bjfish: Is there a way to make this less verbose? The js API for this is `memory.grow(1)`

lib/runtime-core/src/vm.rs
```diff
@@ -80,13 +80,13 @@ impl Ctx {
...
-            imported_funcs: import_backing.functions.as_mut_ptr(),
+            imported_memories: import_backing.vm_memories.as_mut_ptr(),
+            imported_tables: import_backing.vm_tables.as_mut_ptr(),
+            imported_globals: import_backing.vm_globals.as_mut_ptr(),
+            imported_funcs: import_backing.vm_functions.as_mut_ptr(),
 
-            local_backing,
-            import_backing,
+            // local_backing,
+            // import_backing,
```
bjfish: Remove commented out code

syrusakbary:
Love this!

lib/clif-backend/src/module_env.rs
```diff
@@ -20,6 +22,8 @@ pub struct ModuleEnv<'module, 'isa> {
     pub signatures: Map<SigIndex, ir::Signature>,
     globals: Map<GlobalIndex, cranelift_wasm::Global>,
     func_bodies: Map<LocalFuncIndex, ir::Function>,
+    pub deduplicated: PrimaryMap<cranelift_wasm::SignatureIndex, SigIndex>,
+    duplicated: HashMap<SigIndex, cranelift_wasm::SignatureIndex>,
```
bjfish: Maybe the naming could be more specific `deduplicated_signatures` `duplicated_signatures` or `signatures_all`/`signatures_unique`

lib/clif-backend/src/module_env.rs
```diff
@@ -48,23 +54,28 @@ impl<'module, 'isa, 'data> ModuleEnvironment<'data> for ModuleEnv<'module, 'isa>
 
     /// Declares a function signature to the environment.
     fn declare_signature(&mut self, sig: &ir::Signature) {
-        self.signatures.push(sig.clone());
-        self.module.sig_registry.register(Converter(sig).into());
+        let clif_sig_index = self.signatures.push(sig.clone());
+        let func_sig: FuncSig = Converter(sig).into();
+        let sig_index = self.module.sig_registry.lookup_sigindex(func_sig);
```
bjfish: `lookup_sig_index`

lib/clif-backend/src/relocation.rs
```diff
@@ -8,6 +8,18 @@ use cranelift_codegen::ir::{self, ExternalName, LibCall, SourceLoc, TrapCode};
...
+    pub const LOCAL_NAMESPACE: u32 = 1;
+    pub const IMPORT_NAMESPACE: u32 = 2;
+
+    pub const STATIC_MEM_GROW: u32 = 0;
+    pub const STATIC_MEM_SIZE: u32 = 1;
+    pub const SHARED_STATIC_MEM_GROW: u32 = 2;
+    pub const SHARED_STATIC_MEM_SIZE: u32 = 3;
+    pub const DYNAMIC_MEM_GROW: u32 = 4;
+    pub const DYNAMIC_MEM_SIZE: u32 = 5;
+}
```
bjfish: Should these be `enums`?

lib/runtime-core/src/backing.rs
```diff
@@ -163,141 +165,134 @@ impl LocalBacking {
...
-                                elements[init_base + i] = vm::Anyfunc { func_data, sig_id };
-                            }
-                        }
+                    let table = &tables[local_table_index];
+
+                    if (table.current_size() as usize) < init_base + init.elements.len() {
+                        let delta =
+                            (init_base + init.elements.len()) - table.current_size() as usize;
+                        // Grow the table if it's too small.
+                        table.grow(delta as u32).expect("couldn't grow table");
```
bjfish: Maybe a `RangeError` in the future?

lib/runtime-core/src/global.rs
```diff
@@ -0,0 +1,106 @@
...
+impl Global {
+    pub fn new(value: Value) -> Self {
+        Self::_new(value, false)
+    }
+
+    pub fn new_mutable(value: Value) -> Self {
+        Self::_new(value, true)
+    }
+
+    fn _new(value: Value, mutable: bool) -> Self {
```
bjfish: I always think of `_` as a prefix for unused names. Is there a convention for this? If not, maybe `new_internal`

lib/runtime-core/src/vmcalls.rs
lachlansneff: The user can just call `memory.grow(delta)`. This cruft in this function is because it's specifically accessing a particular type of memory. The user-exposed api doesn't expose that.

lib/runtime-core/src/vm.rs
lachlansneff: Oh, sorry, will do.

lib/clif-backend/src/module_env.rs
lachlansneff: This could probably be designed better tbh. I was just trying to get it to work as quickly as possible.

lib/clif-backend/src/relocation.rs
lachlansneff: These are required to be numbers, so constants are better in this scenario.

lib/runtime-core/src/backing.rs
lachlansneff: Yeah, definitely errors need to better. I left them out in a lot of cases.

lib/runtime-core/src/memory/dynamic.rs
```diff
@@ -0,0 +1,104 @@
+use crate::{
+    memory::{WASM_MAX_PAGES, WASM_PAGE_SIZE},
+    sys,
+    types::MemoryDesc,
+    vm,
+};
+
+pub const DYNAMIC_GUARD_SIZE: usize = 4096;
+
```
bjfish: Could we add some documentation describing `DynamicMemory`, `StaticMemory`, `SharedStaticMemory`, their differences and usages.

lib/runtime-core/src/memory/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+            MemoryType::SharedStatic => unimplemented!("shared memories are not yet implemented"),
+        };
+
+        Some(Memory {
+            desc,
+            storage: Rc::new(RefCell::new((memory_storage, vm_local_memory))),
+        })
+    }
+
+    pub fn description(&self) -> MemoryDesc {
```
bjfish: Should this be `descriptor` E.g. `tableDescriptor`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Memory#Syntax

lib/runtime-core/src/error.rs
```diff
@@ -49,14 +50,14 @@ pub enum LinkError {
     IncorrectMemoryDescription {
         namespace: String,
         name: String,
-        expected: Memory,
-        found: Memory,
+        expected: MemoryDesc,
+        found: MemoryDesc,
     },
     IncorrectTableDescription {
```
bjfish: Should this be `TableDescriptor`?

lib/runtime-core/src/memory/static_/unshared.rs
```diff
@@ -0,0 +1,99 @@
...
+        WASM_MAX_PAGES, WASM_PAGE_SIZE,
+    },
+    sys,
+    types::MemoryDesc,
+    vm,
+};
+
+pub struct StaticMemory {
+    memory: sys::Memory,
+    current: u32,
```
bjfish: Maybe name this and `current()`  `current_size` `actual_size`, or just `size`

lib/runtime-core/src/vm.rs
```diff
@@ -295,9 +261,11 @@ pub struct LocalMemory {
     /// Pointer to the bottom of this linear memory.
     pub base: *mut u8,
     /// Current size of this linear memory in bytes.
-    pub size: usize,
-    /// The local memory index.
-    pub index: LocalMemoryIndex,
+    pub bound: usize,
```
bjfish: Document `bound`

lib/runtime-core/src/memory/mod.rs
lachlansneff: Hmm, do you have a preference to use the `descriptor` term rather than `description`?

lib/runtime-core/src/memory/static_/unshared.rs
lachlansneff: Yeah, I wasn't sure what to name the various size functions.

lib/runtime-core/src/memory/mod.rs
bjfish: Yes, I think `descriptor` is the common terminology that I've seen in the `js-api`

lib/runtime-core/examples/simple/main.rs
```diff
@@ -1,16 +1,46 @@
...
+        min: 1,
+        max: Some(1),
+        shared: false,
+    })
+    .unwrap();
+
+    let global = Global::new(Value::I32(42));
+
+    let table = Table::new(TableDesc {
+        ty: ElementType::Anyfunc,
```
bjfish: Could `ty` be `element`? ref: https://github.com/WebAssembly/spec/blob/cf67a477d2d802329640ef4e6461e4c180d8f910/test/js-api/table/constructor.any.js#L57

lachlansneff: Okay, I agree with this.

lib/runtime-core/src/memory/static_/unshared.rs
lachlansneff: The current is the number of pages, not the actual size.

bjfish: Maybe `current_pages` or add some docs

lachlansneff: The js api just exposes the actual length of the memory, not the number of pages. Should we follow suite?

lib/runtime-core/src/error.rs
lachlansneff: Yep, I've changed these all.

lib/runtime-core/src/vm.rs
lachlansneff: It's already documented.

lib/runtime-core/src/memory/static_/unshared.rs
bjfish: Not exactly sure on this, this API appears to have pages as well: https://github.com/rossberg/wasm-c-api/blob/master/include/wasm.h#L450

lib/runtime-core/src/backing.rs
lachlansneff: I've added new errors.

lib/emscripten/src/lib.rs
```diff
@@ -1,15 +1,17 @@
 #[macro_use]
 extern crate wasmer_runtime_core;
+extern crate wasmer_runtime;
```
bjfish: @lachlansneff Should this depend on `wasmer_runtime`?

bjfish: during `cargo build --all` I see error[E0463]: can't find crate for `wasmer_runtime`
```
 --> lib/emscripten/src/lib.rs:3:1
  |
3 | extern crate wasmer_runtime;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate
```

bjfish: I'll just use `cargo build --all --exclude wasmer-emscripten` for now.

PR #135, opened 2019-02-04
C API

bjfish:
Opening this PR to start collecting feedback.

See `lib/runtime-c-api/README.md` about working with this library.

To Do:
- [x] Add table functions
- [x] Add global functions
- [x] Add memory access
- [x] Add exports
- [x] Update imports
- [x] Add error messaging
- [x] Add documentation
- [x] Add function signature
- [x] Consolidate results into one enum
- [x] Add `ifndef`via generator header
- [x] Cleanup


lib/runtime-c-api/tests/test-import-function.c
```diff
@@ -0,0 +1,48 @@
...
+    printf("In print_str\n");
+    print_str_called = true;
+}
+
+int main()
+{
+    wasmer_import_object_t *import_object = wasmer_import_object_new();
+    wasmer_value_tag params_sig[] = {WASM_I32, WASM_I32};
+    wasmer_value_tag returns_sig[] = {};
+    wasmer_imports_set_import_func(import_object, "env", "print_str", print_str, params_sig, 2, returns_sig, 0);
```
syrusakbary: Could be possible to have a macro for this? (regarding params and returns signature)

lachlansneff: I don't think a macro is necessary for this. Using simple arrays here is pretty simple.

lib/runtime-c-api/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+pub extern "C" fn wasmer_import_object_destroy(import_object: *mut wasmer_import_object_t) {
+    if !import_object.is_null() {
+        drop(unsafe { Box::from_raw(import_object as *mut ImportObject) });
+    }
+}
+
+#[allow(clippy::cast_ptr_alignment)]
+#[no_mangle]
+pub unsafe extern "C" fn wasmer_instantiate(
+    mut instance: *mut *mut wasmer_instance_t,
```
lachlansneff: Why double indirection here? If the user is going to have to keep passing around pointers to pointers, should `wasmer_instance_t` be a pointer type?

lib/runtime-c-api/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+    let returns: &[wasmer_value_tag] = slice::from_raw_parts(returns, returns_len as usize);
+    let returns: Vec<Type> = returns.iter().cloned().map(|x| x.into()).collect();
+
+    let export = Export::Function {
+        func: unsafe { FuncPointer::new(func as _) },
+        ctx: Context::Internal,
+        signature: Arc::new(FuncSig::new(params, returns)),
+    };
+
+    // TODO handle existing namespace
```
lachlansneff: Yeah, unfortunately, the way namespaces currently work doesn't allow this. We could either change the implementation, or the user could make a `namespace_object`, fill that with imports, and then import that into the `import_object`?

lib/runtime-c-api/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+
+#[repr(C)]
+#[derive(Clone)]
+pub struct wasmer_memory_t {
+    pub ptr: *mut uint8_t,
+    pub len: uint32_t,
+}
+
+#[no_mangle]
+pub extern "C" fn wasmer_import_object_new() -> *mut wasmer_import_object_t {
```
lachlansneff: It seems odd that this function returns the value, yet `wasmer_instantiate` takes it by reference?

lib/runtime-c-api/src/lib.rs
bjfish: This is the only signature with pointer to pointer because we want to initialize a pointer to instance at the given address. E.g. similar to this: https://libgit2.org/libgit2/#HEAD/group/repository/git_repository_init

lib/runtime-c-api/tests/test-import-function.c
xmclark: I also had some ideas about making the API more user friendly. Effectively sugar coating 😛 

I think keeping the C API simple is a good start, and letting other libraries wrap and make it easier. A higher level API for C++ may be nice, but it depends on how much we want to support.

It's also probably more idiomatic C to not use macros in a library like this.

lib/runtime-c-api/src/lib.rs
bjfish: I think it’s because instantiate has a lot of scenarios with possible error results, but new import object doesn’t have any likely failure scenarios.

lib/runtime-c-api/src/lib.rs
```diff
@@ -0,0 +1,282 @@
...
+                } => Value::I64(I64),
+                wasmer_value_t {
+                    tag: WASM_F32,
+                    value: wasmer_value { F32 },
+                } => Value::F32(F32),
+                wasmer_value_t {
+                    tag: WASM_F64,
+                    value: wasmer_value { F64 },
+                } => Value::F64(F64),
+                _ => panic!("not implemented"),
```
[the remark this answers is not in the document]
bjfish: Looks the same: https://doc.rust-lang.org/src/core/macros.rs.html#551-554

lib/runtime-c-api/src/lib.rs
bjfish: I think I'll update this to add a `namespace_object` to be inline with the current rust API.

lib/runtime-c-api/tests/test-memory.c
```diff
@@ -0,0 +1,59 @@
...
+//    bad_descriptor.max = 10;
+//    wasmer_result_t bad_memory_result = wasmer_memory_new(&bad_memory, bad_descriptor);
+//    printf("Bad memory result:  %d\n", bad_memory_result);
+//    assert(memory_result == WASMER_MEMORY_ERROR);
+//
+//    int error_len = wasmer_last_error_length();
+//    char *error_str = malloc(error_len);
+//    wasmer_last_error_message(error_str, error_len);
+//    assert(0 == strcmp(error_str, "Creation error"));
+//    free(error_str);
```
xmclark: what's up with this commented block?

lib/runtime-c-api/tests/test-tables.c
```diff
@@ -0,0 +1,44 @@
...
+    // printf("Table length:  %d\n", len_grow2);
+    // assert(len_grow2 == 15);
+
+//    wasmer_table_t *table_bad = NULL;
+//    wasmer_limits_t bad_descriptor;
+//    bad_descriptor.min = 15;
+//    bad_descriptor.max = 10;
+//    wasmer_result_t table_bad_result = wasmer_table_new(&table_bad, bad_descriptor);
+//    printf("Table result:  %d\n", table_bad_result);
+//    assert(table_result == WASMER_ERROR);
```
xmclark: same here, commented code?

lib/runtime-c-api/wasmer.h
```diff
@@ -0,0 +1,249 @@
...
+  uint32_t max;
+} wasmer_limits_t;
+
+typedef struct {
+
+} wasmer_table_t;
+
+/**
+ * Frees memory for the given Global
+ */
```
xmclark: nicely done!

lib/runtime-c-api/wasmer.h
```diff
@@ -0,0 +1,249 @@
+#ifndef WASMER_H
+#define WASMER_H
+
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
```
syrusakbary: Maybe we should add a:

```cpp
#ifdef __cplusplus
extern "C" {
#endif
```

And a 
```cpp
#ifdef __cplusplus
}  // extern "C"
#endif
```
at the end

bjfish: @syrusakbary There's another file `wasmer.hh` that is generated for `cpp`.

syrusakbary: Thanks for the clarification! Resolving my comment :)

lib/runtime-c-api/src/lib.rs
bjfish: I've updated instantiate to take an array of imports so this issue does not happen since I don't need to mutate these.

PR #137, opened 2019-02-05
make spectests compile in emscripten update PR

xmclark:
Maybe I am not understanding something...or maybe somebody else has a bigger picture. I'm making this PR so that I can get some feedback or explaination, instead of just pushing to the branch and maybe undoing some intentional work.

This gets the spectests at least compiling again.

I put my comments on the diff.

### DIFF
--- a/lib/spectests/build/spectests.rs
+++ b/lib/spectests/build/spectests.rs
@@ -421,7 +421,7 @@ fn {}_assert_invalid() {{
                 let func_name = format!("{}_assert_return_arithmetic_nan", self.command_name());
                 self.buffer.push_str(
                     format!(
-                        "fn {func_name}(vmctx: &mut Ctx) {{
+                        "fn {func_name}(instance: &mut Instance) {{
     println!(\"Executing function {{}}\", \"{func_name}\");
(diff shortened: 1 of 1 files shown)


lib/spectests/build/spectests.rs
```diff
@@ -370,23 +370,6 @@ fn test_module_{}() {{
...
-    //instance.start();
-}}\n",
-                start_module_call
-            )
-            .as_str(),
-        );
-        self.module_calls
-            .entry(self.last_module)
-            .or_insert(Vec::new())
-            .push(start_module_call);
```
xmclark: this seems unnecessary now...right? `instantiate` call the start function?

lib/spectests/build/spectests.rs
```diff
@@ -612,7 +595,7 @@ fn {}_assert_malformed() {{
                 let func_name = format!("{}_action_invoke", self.command_name());
                 self.buffer.push_str(
                     format!(
-                        "fn {func_name}(vmctx: &mut Ctx) -> Result<()> {{
+                        "fn {func_name}(instance: &mut Instance) -> Result<()> {{
```
xmclark: not sure why these were changed...the body of the function expects an instance

PR #138, opened 2019-02-05
memory fixes for windows

xmclark:
This PR fixes a few small issues with building memory code for windows. We use the winapi `VirtualFree` instead of `munmap`. A few crates were added that had been missing from the wasmer runtime crate.

### DIFF
--- a/lib/runtime-core/src/sys/windows/memory.rs
+++ b/lib/runtime-core/src/sys/windows/memory.rs
@@ -1,7 +1,9 @@
 use winapi::um::memoryapi::{
-    VirtualAlloc,
-    MEM_RESERVE, MEM_COMMIT,
-    PAGE_NOACCESS, PAGE_EXECUTE_READ, PAGE_READWRITE, PAGE_READONLY,
+    VirtualAlloc, VirtualFree
+};
+use winapi::um::winnt::{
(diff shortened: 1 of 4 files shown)


lib/runtime-core/src/sys/windows/memory.rs
```diff
@@ -103,7 +95,7 @@ impl Memory {
 impl Drop for Memory {
     fn drop(&mut self) {
         if !self.ptr.is_null() {
-            let success = unsafe { libc::munmap(self.ptr as _, self.size) };
+            let success = unsafe { VirtualFree(self.ptr as _, self.size, MEM_DECOMMIT) };
```
xmclark: the winapi equivalent of `munmap`

lib/runtime-core/src/sys/windows/memory.rs
```diff
@@ -27,17 +27,10 @@ impl Memory {
...
-                size,
-                MEM_RESERVE,
-                PAGE_NOACCESS,
-            )
-        };
+        let ptr = unsafe { VirtualAlloc(ptr::null_mut(), size, MEM_RESERVE, PAGE_NOACCESS) };
 
         if ptr.is_null() {
-            Err("unable to allocate memory")
+            Err(String::from("unable to allocate memory"))
```
lachlansneff: I'd prefer `.to_string` here, but it doesn't matter much.

PR #140, opened 2019-02-06
fix putenv and setenv for windows

xmclark:
This PR separates the different platform implementations for the emscripten env functions. 

This PR attempts to fix `putenv` and `setenv`. No binding is written in libc for `putenv` so I link it. The foreign function name is `_putenv`. 

There is no such function as `setenv` in libc on windows, so we fake it by constructing a new string in the form "name=value" from the two string pointers.

There is now some duplicated code because we have the `env` module copied for windows. Look for my comments in the diff to see what actually changed.


lib/emscripten/src/env/windows/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+}
+
+/// emscripten: _putenv // (name: *const char);
+pub fn _putenv(name: c_int, ctx: &mut Ctx) -> c_int {
+    debug!("emscripten::_putenv");
+
+    let name_addr = emscripten_memory_pointer!(ctx.memory(0), name) as *const c_char;
+
+    debug!("=> name({:?})", unsafe { CStr::from_ptr(name_addr) });
+    unsafe { putenv(name_addr) }
```
xmclark: this now uses the `putenv` that is declared above

lib/emscripten/src/env/windows/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+        let name = read_string_from_wasm(name, ctx.memory(0));
+        let value = read_string_from_wasm(value, ctx.memory(0));
+        let putenv_string = format!("{}={}", name, value);
+        let putenv_cstring = CString::from::<String>(putenv_string);
+        let putenv_raw_ptr = putenv_cstring.as_ptr();
+        debug!("=> name({:?})", CStr::from_ptr(name_addr));
+        debug!("=> value({:?})", CStr::from_ptr(value_addr));
+        putenv(putenv_raw_ptr)
+    }
+}
```
xmclark: this was re-written and uses a helper function to pull out the strings and concatenate them

lib/emscripten/src/env/windows/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+
+    unsafe { copy_cstr_into_wasm(ctx, c_str) }
+}
+
+unsafe fn read_string_from_wasm(pointer: c_int, memory: &Memory) -> String {
+    let string_cell = memory.view::<u8>()[(pointer as usize)..].as_ptr();
+    let string_ptr = string_cell.as_ptr();
+    let cstr_ptr = CStr::from_ptr(string_ptr);
+    String::From(cstr_ptr)
+}
```
xmclark: maybe this goes in utils?

lib/emscripten/src/env/windows/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+use crate::utils::{allocate_on_stack, copy_cstr_into_wasm, copy_terminated_array_of_cstrs};
+use crate::EmscriptenData;
+use wasmer_runtime_core::memory::Memory;
+use wasmer_runtime_core::vm::Ctx;
+
+#[link(name = "c")]
+extern "C" {
+    #[link_name = "_putenv"]
+    pub fn putenv(s: *const c_char) -> c_int;
+}
```
xmclark: not actually declared in libc, so we manually declare it with an alias

lib/emscripten/src/env/windows/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+
+    let c_str: *mut c_char = unsafe { getenv(name_addr as _) } as _;
+    if c_str.is_null() {
+        return 0;
+    }
+
+    unsafe { copy_cstr_into_wasm(ctx, c_str) }
+}
+
+unsafe fn read_string_from_wasm(pointer: c_int, memory: &Memory) -> String {
```
lachlansneff: I'd like to do this in a safer way actually.

```rust
fn read_string_from_wasm(memory: &Memory, pointer: u32) -> String {
    let v: Vec<_> = memory.view::<u8>()[pointer as usize]
        .iter()
        .take_while(|cell| cell.get() != 0)
        .map(|cell| cell.get())
        .collect();

    String::from_ut8(v.as_slice().unwrap()
}
```

lib/emscripten/src/env/windows/mod.rs
lachlansneff: Try to reduce the unsafe block to just the last line.

PR #144, opened 2019-02-06
create the windows io mod and link printf

xmclark:
This PR creates a new io mod for windows and links the printf function which is not defined in libc for windows.


lib/emscripten/src/io/windows.rs
```diff
@@ -0,0 +1,24 @@
+use libc::{c_char, c_int};
+use wasmer_runtime_core::vm::Ctx;
+
+// this may be problematic for msvc which uses macros for the printf family
+// this acts as a placeholder for now
```
lachlansneff: Can you specify that this will compile, but fail at link time?

xmclark: Of course! I will add a TODO comment and clarify. I will also drop in the `#[cfg_attr(all(windows, target_env="msvc"), link(name="legacy_stdio_definitions", kind="static-nobundle"))]`.

xmclark: ✔️

PR #145, opened 2019-02-06
let the compiler deduce types in syscalls

xmclark:
This PR is the first pass over `syscalls.rs`. In many cases, the signatures of libc functions differ slightly between the platforms. In many cases, different numeric type are used. This PR removes some explicit annotations and allows numeric types to be cast to the expected types. This moves us closer to windows support.


lib/emscripten/src/syscalls.rs
```diff
@@ -342,7 +342,7 @@ pub fn ___syscall102(which: c_int, mut varargs: VarArgs, ctx: &mut Ctx) -> c_int
             };
 
             type T = u32;
-            let payload = 1 as *const T as *const c_void;
```
lachlansneff: I still have no idea what this is, but it doesn't really pertain to this pr.

PR #150, opened 2019-02-07
fill unsetenv for windows

xmclark:
This PR fills `_unsetenv` for windows and re-uses the helper function `read_string_from_wasm` and the `setenv` function. This is required because `unsetenv` does not exist on windows.

### DIFF
--- a/lib/emscripten/src/utils.rs
+++ b/lib/emscripten/src/utils.rs
@@ -14,6 +14,7 @@ use wasmer_runtime_core::{
     units::Pages,
     vm::Ctx,
 };
+use std::ffi::CString;
 
 /// We check if a provided module is an Emscripten generated one
 pub fn is_emscripten_module(module: &Module) -> bool {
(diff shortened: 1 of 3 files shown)


lib/emscripten/src/utils.rs
```diff
@@ -158,11 +159,12 @@ pub unsafe fn copy_stat_into_wasm(ctx: &mut Ctx, buf: u32, stat: &stat) {
...
-    memory.view::<u8>()[(offset as usize)..]
+    let v: Vec<u8> = memory.view()[(offset as usize)..]
         .iter()
-        .take_while(|cell| cell.get() != 0)
-        .map(|cell| cell.get() as char)
-        .collect()
+        .map(|cell| cell.get())
+        .take_while(|&byte| byte != 0)
+        .collect();
+    String::from_utf8_lossy(&v).to_owned().to_string()
```
xmclark: this is modified to correctly handle utf8 and invalid utf8 strings

PR #164, opened 2019-02-08
[WIP] Single-pass native code generation.

losfair:


lib/dynasm-backend/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+
+impl DynasmCompiler {
+    pub fn new() -> DynasmCompiler {
+        DynasmCompiler {}
+    }
+}
+
+impl Compiler for DynasmCompiler {
+    fn compile(&self, wasm: &[u8], _: Token) -> CompileResult<ModuleInner> {
+        validate(wasm)?;
```
lachlansneff: Instead of validating and then parse the module, can you do both at the same time? We want to be able to support streaming compilation in the future.

syrusakbary:
Closing this PR for now.

PR #165, opened 2019-02-08
Fix emscripten setup, nginx works

lachlansneff:


lib/emscripten/src/storage.rs
```diff
@@ -4,14 +4,9 @@ pub fn align_memory(ptr: u32) -> u32 {
...
-//         "not enough memory for static allocation - increase total_memory!"
-//     );
-//     old_static_top
-// }
+pub fn static_alloc(static_top: &mut u32, size: u32) -> u32 {
+    let old_static_top = *static_top;
+    // NOTE: The `4294967280` is a u32 conversion of -16 as gotten from emscripten.
+    *static_top = (*static_top + size + 15) & 4294967280;
+    old_static_top
+}
```
xmclark: yay! get this commented code out.

xmclark:
Good job!

lib/emscripten/src/lib.rs
```diff
@@ -265,7 +222,7 @@ pub struct EmscriptenGlobals {
 }
 
 impl EmscriptenGlobals {
-    pub fn new(module: &Module) -> Self {
+    pub fn new(module: &Module /*, static_bump: u32 */) -> Self {
```
bjfish: Is this needed `/*, static_bump: u32 */` ?

lib/emscripten/src/lib.rs
```diff
@@ -284,24 +241,64 @@ impl EmscriptenGlobals {
...
+                memory_base,
+                table_base,
+                temp_double_ptr,
+
+                infinity: std::f64::INFINITY,
+                nan: std::f64::NAN,
+            }
         };
 
+        // let memory_base = STATIC_BASE;
```
bjfish: Can the commented out code be removed?

lib/emscripten/src/lib.rs
lachlansneff: the `static_bump` is emscripten module independent, but we've hardcoded the one for nginx here. The user has to pull the constant out of the generated js file, since there is no provided metadata in emscripten modules. We should make it variable at some point, and this is a reminder.

lib/emscripten/src/lib.rs
lachlansneff: Yeah, will do.

PR #167, opened 2019-02-08
Naive short circuiting implementation for user panics and results.

lachlansneff:
In addition to still supporting regular functions imported into wasm, this pr allows functions that return a `Result<T, E: Debug>`. If they return an `Err(...)`, the runtime will unwind back to where the wasm was called, and return `RuntimeError::User { ... }`. The same behavior also results from an imported function panicking.


lib/clif-backend/src/signal/mod.rs
```diff
@@ -118,6 +131,10 @@ impl ProtectedCaller for Caller {
             })
             .collect())
     }
+
+    fn get_early_aborter(&self) -> Box<dyn EarlyAborter> {
+        Box::new(Aborter)
+    }
```
xmclark: why do we need to put a zero size thing in a box? Any reason `do_early_abort` method can't be a standalone function?

lachlansneff: Since early abort will function differently between each backend, we need tie the implementation we use to the backend that compiled the wasm. Here, it doesn't actually end up being zero-sized. While the type itself has no size, there's still a v-table that is packaged along with it.

PR #168, opened 2019-02-09
add windows exception handling

xmclark:
This PR adds exception handling as an alternative to signal handling, on windows. 

The implementation is similar to that of signals, and we still use the `jmp` strategy for calling back with our exception data.

### DIFF
--- a/lib/clif-backend/src/resolver.rs
+++ b/lib/clif-backend/src/resolver.rs
@@ -223,9 +223,9 @@ impl FuncResolverBuilder {
                         LibCall::TruncF64 => libcalls::truncf64 as isize,
                         LibCall::NearestF64 => libcalls::nearbyintf64 as isize,
                         #[cfg(all(target_pointer_width = "64", target_os = "windows"))]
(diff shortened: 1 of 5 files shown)


lib/clif-backend/src/resolver.rs
```diff
@@ -223,9 +223,9 @@ impl FuncResolverBuilder {
                         LibCall::TruncF64 => libcalls::truncf64 as isize,
                         LibCall::NearestF64 => libcalls::nearbyintf64 as isize,
                         #[cfg(all(target_pointer_width = "64", target_os = "windows"))]
-                        Probestack => __chkstk as isize,
+                        LibCall::Probestack => __chkstk as isize,
                         #[cfg(not(target_os = "windows"))]
-                        Probestack => __rust_probestack as isize,
+                        LibCall::Probestack => __rust_probestack as isize,
```
xmclark: this is unrelated, but it was missed in a previous PR

xmclark:
closing this PR because the implementation will need to be different for experimental windows exception handling to work.

Issue #172, opened 2019-02-10
`abortOnCannotGrowMemory` has the wrong type signiture, `_emscripten_get_heap_size` and `_emscripten_resize_heap` missing


davidMcneil:
I am trying to compile and run the emscripten tests (`make emtests`). I try and run an output wasm file for example `clock_gettime.wasm` and get the following error:

```
"Can\'t instantiate module: LinkError([ImportNotFound { namespace: \"env\", name: \"_emscripten_get_heap_size\" }, ImportNotFound { namespace: \"env\", name: \"_emscripten_resize_heap\" }, IncorrectImportSignature { namespace: \"env\", name: \"abortOnCannotGrowMemory\", expected: FuncSig { params: [I32], returns: [I32] }, found: FuncSig { params: [], returns: [I32] } }])"
```

Comparing the resulting wasm files it looks like the type for `abortOnCannotGrowMemory` has changed from:
`(type $t2 (func (result i32)))`
to
`(type $t1 (func (param i32) (result i32)))`

It also looks like `_emscripten_get_heap_size` and `_emscripten_resize_heap` are missing. These are not even imported in the original wasm.

It seems like this may be an issue with emscripten and not wasmer. I am using emscripten version 1.38.26.

syrusakbary:
Taking a look now.
It seems memory management changed a bit: https://github.com/emscripten-core/emscripten/commit/1655d72383eaeb79c5bdd764c5502833bb9c457a

I think is feasible to support both cases (old-emscripten files, and new ones)

davidMcneil:
Compiling with emscripten version 1.38.21 fixes the issue.

davidMcneil:
This is by no means a complete solution, but it gets the tests to pass https://github.com/davidMcneil/wasmer/tree/172. I would be happy to put this into a more formal pull request. What would be the best way to determine the type signature of `abortOnCannotGrowMemory` to use?

syrusakbary:
Thanks for the changes @davidMcneil.

I've created the PR #173 to add a comment to have compatibility for both functions:
https://github.com/wasmerio/wasmer/pull/173

(any change in your branch will be reflected in the PR)

syrusakbary:
Closing the issue, as is fixed with #173

PR #173, opened 2019-02-11
Fix emscripten environment (#172)

syrusakbary:
Fix emscripten environment


lib/emscripten/src/memory.rs
```diff
@@ -33,8 +47,8 @@ pub fn enlarge_memory(_ctx: &mut Ctx) -> u32 {
 }
 
 /// emscripten: abortOnCannotGrowMemory
-pub fn abort_on_cannot_grow_memory(ctx: &mut Ctx) -> u32 {
-    debug!("emscripten::abort_on_cannot_grow_memory");
+pub fn abort_on_cannot_grow_memory(requested_size: u32, ctx: &mut Ctx) -> u32 {
```
syrusakbary: Can we have a `abort_on_cannot_grow_memory` and `abort_on_cannot_grow_memory_old`? (with the previous signature)
We will use one or the other depending on the expected signature from the function in the module.

That way we will have backwards compatibility with old-generated emscripten files.

## How to do it?

Creating a flag in the EmscriptenGlobals storing if the signature is one or the other:
https://github.com/wasmerio/wasmer/blob/c3da14d95896c27ed492c7461bb0ab856d9778e4/lib/emscripten/src/lib.rs#L242

And then use one function or the other depending on that flag.

davidMcneil: Thanks, an issue I think we are going to run into is that the types of functions are not going to be the same and these functions get converted into the `Func` type which will be generic over different argument list types.

davidMcneil: It looks like I can potential use the `new_from_ptr` call. Is this the best way to go?

davidMcneil: I think I have a solution using the `IsExport` trait object.

davidMcneil:
This should be ready to be merged. I tested locally with emcc version 1.38.21 and 1.38.26. The wasm files are built with version 1.38.26. Does it make sense to update these files or should I have kept the older version?

syrusakbary:
Looks great! Merging

PR #175, opened 2019-02-14
add windows exception handling in C

xmclark:
This PR adds experiemental support for windows. After much trial-and-error, we came up with a pure C solution for doing the exception handling (signal handling) and stack `jmp`ing. This is not idea, but it is a stop-gap solution and enables us to move forward on windows support (#51).


syrusakbary:
Love this! ❤️

CryZe:
Is C used here because setjmp is problematic in Rust or what's the reason?

syrusakbary:
I reviewed a bit more. I think naming the package `clif-backend-windows` it's a bit misleading, since only handles exception handling in windows and have no dependency over `clif-backend`.

What you think on renaming the package to `win-exception-handler` ?

lachlansneff:
Look at all of those passing tests 😎

xmclark:
Hey @CryZe yes `setjmp`/`longjmp` are a necessary evil here. We actually use them for unix too! In this case, we had trouble getting the `jmp`s to work across the FFI so this is a stop-gap solution written in pure C (as seen in this PR). We never `jmp` between Rust and C.

xmclark:
@syrusakbary I think a name change is warranted! I will change it to `win-exception-handler`. I am open to other suggestions.

xmclark:
I want 👀 from @bjfish or @lachlansneff before we merge.

lib/win-exception-handler/exception_handling/exception_handling.c
```diff
@@ -0,0 +1,71 @@
...
+exceptionHandler(struct _EXCEPTION_POINTERS *ExceptionInfo) {
+    EXCEPTION_RECORD* pExceptionRecord = ExceptionInfo->ExceptionRecord;
+    PCONTEXT pCONTEXT = ExceptionInfo->ContextRecord;
+    caughtExceptionAddress = pExceptionRecord->ExceptionAddress;
+    caughtInstructionPointer = pCONTEXT->Rip;
+    if (alreadyHandlingException == TRUE) {
+        return EXCEPTION_CONTINUE_SEARCH;
+    }
+    alreadyHandlingException = TRUE;
+    pCONTEXT->Rip = (uintptr_t)(&longjmpOutOfHere);
```
lachlansneff: Add some comments here.

> Basically, here, we coerce the os to resume us into a context that calls `longjmp` instead of just continuing. Presumably, we cannot `longjmp` out of the signal/exception context, like we can on unix.

PR #178, opened 2019-02-14
Fix/windows installer

xmclark:
This adds an installer that should auto publish to github when we merge to master and have a tag

### DIFF
--- a/src/bin/wasmer.rs
+++ b/src/bin/wasmer.rs
@@ -21,6 +21,7 @@ enum CLIOptions {
     Run(Run),
 
     /// Update wasmer to the latest version
+    #[cfg(not(target_os = "windows"))]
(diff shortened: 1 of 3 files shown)


src/bin/wasmer.rs
```diff
@@ -113,6 +114,7 @@ fn main() {
     let options = CLIOptions::from_args();
     match options {
         CLIOptions::Run(options) => run(options),
+        #[cfg(not(target_os = "windows"))]
         CLIOptions::SelfUpdate => update::self_update(),
```
syrusakbary: Rather than removing this option from the CLI I would prefer to raise an Error on the `self_update` function itself.

The API that we deliver in our CLI needs to be consistent across platforms.

PR #185, opened 2019-02-15
Grow as a trait

xmclark:
This PR refactors how `Memory` and `Table` grow their memory. Instead of concrete `impl`, we have a `Grow` trait. This trait is generic to the type that the memory grows. The `grow` method returns a  `Result<T, GrowError>`.

### DIFF
--- a/lib/runtime-core/src/backing.rs
+++ b/lib/runtime-core/src/backing.rs
@@ -1,3 +1,4 @@
+use crate::grow::Grow;
 use crate::{
     error::{LinkError, LinkResult},
     export::{Context, Export},
@@ -15,7 +16,6 @@ use crate::{
     vm,
 };
 use std::{slice, sync::Arc};
-use crate::grow::Grow;
 
 #[derive(Debug)]
 pub struct LocalBacking {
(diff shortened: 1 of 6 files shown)


lachlansneff:
What's the use case for this? Also, can you make the the type parameter an associated type instead?

xmclark:
Both `Table` and `Memory` were implementing a `grow` method. The behavior is similar enough that I think it warrants a trait. 

I made this PR because we wanted to change the contract for how we `grow`. We wanted to replace returned `Option` with a `Result`. The trait helps us realize this.  

Is an associated type better, or more conventional?

bjfish:
Are traits are a little excessive here? For example, the definition I see for traits here is:
`A trait is a collection of methods defined for an unknown type: Self. They can access other methods declared in the same trait.`: https://doc.rust-lang.org/rust-by-example/trait.html

However, in this case we should know the types already.

xmclark:
From the offline discussions: we were unsure if a new trait is warranted. Due to feedback, this PR will be closed, and I will implement with conrete methods, as it was before. We may revisit a new trait if it makes more sense. 🌈

PR #186, opened 2019-02-15
Validate descriptor max on creating new table or memory

xmclark:


lib/runtime-core/src/table/mod.rs
```diff
@@ -50,6 +50,12 @@ impl Table {
     /// # }
     /// ```
     pub fn new(desc: TableDescriptor) -> Result<Self, CreationError> {
+        if let Some(max) = desc.maximum {
+            if max < desc.minimum {
+                return Err(CreationError::UnableToCreateTable);
```
bjfish: Could we please have specific error messages for these which I think would be more user friendly? Also, they would help us debug issues.

E.g.  "maximum table size cannot be less than minimum"

xmclark: See the new commit. I added a new error.

PR #191, opened 2019-02-18
Create a grow error and refactor grow impl to return result

xmclark:
This PR adds some new error enum values for `GrowError` and a few nested error types that `Into` a `GrowError`. This will help us better understand memory related errors.


bjfish:
I think it would be nice if we could handle the different memory failures with separate error types and/or messages. For example, I see growing memory can fail in a variety of ways but currently the same error type would be returned without a message. Thoughts?

xmclark:
@bjfish I was hesitant at first...but after a full day of coding, a few cups of joe, and some carrots with hummus, I decided that you have a good point. I started working on implementing a more rich set of error types that we can use for growing memory. I will unmark this WIP tomorrow sometime for review again.

lib/runtime-c-api/src/lib.rs
```diff
@@ -208,14 +208,15 @@ pub extern "C" fn wasmer_memory_grow(
...
-            msg: "unable to grow memory".to_string(),
-        });
-        wasmer_result_t::WASMER_ERROR
+    let delta_result = memory.grow(Pages(delta));
+    match delta_result {
+        Ok(_) => wasmer_result_t::WASMER_OK,
+        Err(_) => {
+            update_last_error(CApiError {
+                msg: "unable to grow memory".to_string(),
+            });
```
bjfish: If the error implements "Error",  you can remove this CApiError and just pass the `Err(error)` to `update_last_error(error)`

xmclark: Right! Done. ✔️

lib/runtime-c-api/tests/test-tables.c
```diff
@@ -21,26 +21,29 @@ int main()
...
+    wasmer_table_t *table_bad = NULL;
+    wasmer_limits_t bad_descriptor;
+    bad_descriptor.min = 15;
+    wasmer_limit_option_t max2;
+    max2.has_some = true;
+    max2.some = 10;
+    bad_descriptor.max = max2;
+    wasmer_result_t table_bad_result = wasmer_table_new(&table_bad, bad_descriptor);
+    printf("Table result:  %d\n", table_bad_result);
+    assert(table_bad_result == WASMER_ERROR);
```
xmclark: @bjfish thanks for fixing this.

PR #193, opened 2019-02-20
stub wasm32-unknown-emscripten target imports

xmclark:
This PR stubs implements and hooks up the required imports for the `wasm32-unknown-emscripten` target. 

This rust target is not exactly the preferred way for building Wasm apps, but it is a good way for right now. It will enable developers to build standalone Wasm apps written in Rust that run on Wasmer.


lib/emscripten/src/lib.rs
```diff
@@ -303,206 +332,244 @@ pub fn generate_emscripten_env(globals: &mut EmscriptenGlobals) -> ImportObject
...
-          "Infinity" => Global::new(Value::F64(f64::INFINITY)),
-        },
-        "global.Math" => {
-            "pow" => func!(crate::math::pow),
-        },
-        "asm2wasm" => {
-            "f64-rem" => func!(crate::math::f64_rem),
-        },
-    }
+            "env" => {
```
lachlansneff: I think this got indented too much.

xmclark: fixed ✔️

PR #196, opened 2019-02-21
Use minimum limit as initial table size

bjfish:


lib/runtime-core/src/table/anyfunc.rs
```diff
@@ -53,10 +53,7 @@ impl AnyfuncTable {
         desc: TableDescriptor,
         local: &mut vm::LocalTable,
     ) -> Result<Box<Self>, CreationError> {
-        let initial_table_backing_len = match desc.maximum {
-            Some(max) => max,
-            None => desc.minimum,
-        } as usize;
+        let initial_table_backing_len = desc.minimum as usize;
```
lachlansneff: There are some additional optimizations we can do in the future, like reserving to the max if there is one, but it doesn't matter at this point.

PR #199, opened 2019-02-21
more emscripten stubs and urandom hack

xmclark:
This PR creates some technical debt. Hooray!

This adds some more `invoke_*` functions for emscripten wasm projects. It also adds a hack for missing a virtual file system. On windows, we create a temp file with some random data to fill the role of `/dev/urandom`. This was needed for some reason.


lachlansneff:
My guess for why the cli app needed some random numbers is as follows:

`structopt` is essentially just a wrapper around `clap`. `clap` uses HashMaps internally in a couple of places. `std::collections::HashMap` (and `std::collections::HashSet`) requires entropy when initializing because its hashing function, which attempts to make ddosing through specially crafted input impossible, requires entropy.

lib/emscripten/src/emscripten_target.rs
```diff
@@ -173,3 +173,393 @@ pub fn ___gxx_personality_v0(ctx: &mut Ctx, a: i32, b: i32, c: i32, d: i32, e: i
...
+pub fn nullFunc_viji(ctx: &mut Ctx, index: i32) {
+    debug!("emscripten::nullFunc_viji");
+}
+pub fn nullFunc_vijiii(ctx: &mut Ctx, index: i32) {
+    debug!("emscripten::nullFunc_vijiii");
+}
+pub fn nullFunc_vijj(ctx: &mut Ctx, index: i32) {
+    debug!("emscripten::nullFunc_vijj");
+}
+pub fn invoke_dii(ctx: &mut Ctx, index: i32, a1: i32, a2: i32) -> f64 {
```
lachlansneff: The `invoke_*` functions are intended to be used as a `try`, `catch` analog. We don't support that kind of thing right now, but we could think about registering another trap handler here? We should look into registering multiple trap handlers and checking each one.

PR #200, opened 2019-02-22
Added cache into standalone app

syrusakbary:
Added cache into standalone app.
This PR also removes `wasm_hash` from the ModuleInfo to use it directly with the cache api (load/store).


lachlansneff:
> Should there be an option to disable cache? I think it would be nice if integration tests ran for uncached, then cached.

@bjfish Good idea, cli flag?

syrusakbary:
@bjfish @lachlansneff I've just added an option `--disable-cache` in the CLI for `wasmer run`.
So we can do `wasmer run examples/lua.wasm --disable-cache -- -V`

PR #203, opened 2019-02-23
add execvp for emscripten

xmclark:
This PR adds support for `execvp` function. This can be used in emscripten wasmer to execute programs outside of the sandbox.


xmclark:
closing this because we will consolidate several changes together for the bundler app

PR #204, opened 2019-02-23
fix stack overflow in Error's Display impl

canndrew:
The `Display` impl for `Error` calls itself in an infinite loop and overflows the stack. This PR fixes the issue.


syrusakbary:
Everything looks great!
Once the lint issue is fixed we should be good to merge 👍

PR #206, opened 2019-02-23
Add C API module export descriptors

bjfish:
This is similar to the JS API `Module.exports()`


lib/runtime-c-api/wasmer.hh
```diff
@@ -110,6 +118,27 @@ wasmer_result_t wasmer_compile(wasmer_module_t **module,
                                uint8_t *wasm_bytes,
                                uint32_t wasm_bytes_len);
 
+/// Gets export descriptor kind
+wasmer_import_export_kind wasmer_export_descriptor_kind(wasmer_export_descriptor_t *export_);
```
xmclark: trailing `_` in `*export_`?

xmclark: nitpick

PR #210, opened 2019-02-25
no caching on windows

xmclark:
Due to issues with reading virtual memory on windows, we will disable caching on windows for 0.2.0 release.

### DIFF
--- a/src/bin/wasmer.rs
+++ b/src/bin/wasmer.rs
@@ -1,7 +1,6 @@
 extern crate structopt;
 
 use std::env;
-use std::fs;
 use std::fs::File;
 use std::io;
 use std::io::Read;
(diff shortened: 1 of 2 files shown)


src/bin/wasmer.rs
```diff
@@ -188,8 +192,10 @@ fn main() {
         CLIOptions::SelfUpdate => {
             println!("Self update is not supported on Windows. Use install instructions on the Wasmer homepage: https://wasmer.io");
         }
+        #[cfg(not(target_os = "windows"))]
         CLIOptions::Cache(cache) => match cache {
             Cache::Clean => {
+                use std::fs;
```
syrusakbary: Can we move this import to the head?

xmclark: @lachlansneff @bjfish do we mind this coding style? It is our preference. I moved it into this block because it is unused on the windows target.

PR #215, opened 2019-02-26
Add nginx LICENSE

markandrus:
Since [examples/nginx/nginx.wasm](https://github.com/wasmerio/wasmer/blob/master/examples/nginx/nginx.wasm) is nginx compiled with Emscripten, I believe the nginx license needs to be included with it. I got the nginx license from [here](https://github.com/nginx/nginx/blob/master/docs/text/LICENSE).


syrusakbary:
Good call! Merging

Issue #169, opened 2019-02-09
[runtime] Setup context data before instantiation or support closures


evgenykuzyakov:
Currently, instance calls `start_func` automatically when created here https://github.com/wasmerio/wasmer/blob/master/lib/runtime-core/src/instance.rs#L69

Before calling `start_func`, I want to be able to modify `Ctx` by modifying `.data` first. Would be nice to move it out of Instance::new(). 

Context:
We are using gas meter injector that adds some runtime callback after every operation in wasm. I assume it also modifies start_func and inserts some gas meter calls there as well. Since ctx.data is null at this moment it fails with the following error: `CallError(Runtime(Unknown { msg: "trap at 0x40 - segmentation violation" }))`

lachlansneff:
Hmm. I believe we'd be breaking the WebAssembly specification if we didn't automatically run the `$start` func. I'll think more about the upsides and downsides. We may be able to supply a way of setting the data before instantiation, or allow users to import closures that contain references to the outside state.

Would it be useful for your usecase if the runtime had explicit support for metering? It would likely have much less performance-impact than what you're doing now.

evgenykuzyakov:
> We may be able to supply a way of setting the data before instantiation, or allow users to import closures that contain references to the outside state.

Either way works for us. Supporting closures is cleaner, but as long as it doesn't affect performance. 

> Would it be useful for your usecase if the runtime had explicit support for metering? It would likely have much less performance-impact than what you're doing now.

That would be great. I think compute metering is the biggest overhead, because it requires context switching. Having it done by the runtime with some configuration that we can use would be huge perf boost. We can even receive it after entire execution if we could setup the limit beforehand.

syrusakbary:
We are triaging stale issues now.

It seems, as commented in #219 we closed the PR in favor of the Import State changes from #299: 7b0992e.

So that should be the path forward. Let us know if you have other ideas.

Closing issue for now :)

Issue #188, opened 2019-02-16
Incorrect Calling Conventions on Windows


CryZe:
Unfortunately #183 isn't actually quite resolved yet as the same problem exists for the wasm code calling into the host with an f64 argument. I believe we are dealing with S2 then or something (because of the additional context argument)? I haven't yet investigated this, so I don't quite know enough details yet.

CryZe:
It seems like the problem there is that ctx is actually passed after the f64 or something. f64 is in the XMM0 register even though the ABI requires it to be in the second.

CryZe:
![https://i.imgur.com/8Jgwgh9.png](https://i.imgur.com/8Jgwgh9.png)

Defining the wrap function to always be of "sysv64" calling convention fixes the bug, but that seems more like a hack rather than a fix, as cranelift should probably just emit the correct calling convention in the first place. I don't know where this is done though.

lachlansneff:
@CryZe Thanks for the deep dive! I "own" most of the runtime code, so I'll take a look at this on Monday. It's probably the calling convention getting set up wrong somewhere.

lachlansneff:
@CryZe Sorry this took so long, but I've just had time to investigate and indeed yes, cranelift implements the x64 calling convention incorrectly. I honestly don't blame them, it seems to waste registers.

Here's the spec: https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=vs-2017#example-of-argument-passing-3---mixed-ints-and-floats

syrusakbary:
We created an issue in Cranelift: https://github.com/CraneStation/cranelift/issues/691

syrusakbary:
This issue was fixed in: https://github.com/CraneStation/cranelift/pull/701 (related issue in cranelift: https://github.com/CraneStation/cranelift/issues/691)

Once we update the Cranelift version and this issue should be fixed.

Keeping the issue open until then!

CryZe:
The implementation was unfortunately incorrect. So they still need to fix that.

hrydgard:
I had similar problems with floating point arguments on Windows, but indeed appears to work correctly on master with the the updated cranelift.

Are you guys planning to tag a new release in the near future? Would be convenient :)

CryZe:
This seems to be resolved :)

PR #219, opened 2019-02-26
Remove ctx argument on imported functions and allow closures.

lachlansneff:
This allows you to use fully-fledged closures for imported functions.

```rust
let a_str = "this is a str";

    "env" => {
        "foo_bar" => Func::new(|other_thing| {
            println!("{}: {}", a_str, other_thing);
        },
    },
};
```

@evgenykuzyakov Does this fix the data initialization issue you were having?


CryZe:
This might accidentally also work around #188

lachlansneff:
@CryZe Could you elaborate? I didn't really change how imported function calling worked on windows.

CryZe:
Oh yeah, I thought ctx wouldn't be passed anymore with this, but looks like it still does. Nvm then

CryZe:
Oh yeah, if ctx / some env internally is still used as the first argument, then this still happens.

lachlansneff:
Oh, that's a shame. I haven't had time to look into fixing the issues you found; been working on some stuff that we're not ready to show yet.

evgenykuzyakov:
Yes. This should fix #169

bjfish:
Adding closures would be nice, I think whatever is done should be compatible with our current ctx usage in emscripten/C API/etc.  

It's not a big deal but this PR would be a little easier to read if the renaming of `ctx` to `env` was split into a new PR. Plus we should update the naming in the example repos/blog posts/documentation to `env` if this is changed.

Hywan:
Why do you want to remove the context argument?

lachlansneff:
This doesn't remove *per-say*, but really just repurposes it to hold some environmental data, whether that's the vm ctx or not. Then it adds some convenient sugar around that to make it easy to use closures as imported functions.

syrusakbary:
Closing this PR in favor of the Import State changes from #299: https://github.com/wasmerio/wasmer/commit/7b0992e44f21828a53c8ca8c06b234d24641c09f.

This new API let the user establish the context data very easily.
See example in WASI API: https://github.com/wasmerio/wasmer/blob/2aefa731a66616781f0632d756f1f68c9df831e5/lib/wasi/src/lib.rs#L33-L53

lib/win-exception-handler/src/exception_handling.rs
```diff
@@ -16,7 +16,7 @@ extern "C" {
     #[link_name = "callProtected"]
     pub fn __call_protected(
         trampoline: Trampoline,
-        ctx: *mut Ctx,
+        env: Option<NonNull<Env>>,
```
Hywan: The C code doesn't handle the `Option` correctly I guess (it expects a pointer).

PR #223, opened 2019-02-28
naively implement pipe

xmclark:


xmclark:
closing this because we will consolidate several changes together for the bundler app

PR #224, opened 2019-02-28
fix appveyor installer and build

xmclark:
This PR introduces a few changes. I reset and squashed some commits to make the intentions more clear. Here are some of the high level changes:

- Reverted changes to how we allocate virtual memory, this may have been related to the issues we were seeing in appveyor builds.
- We now unregister exception handlers have functions have been executed. This is naive, but it fixes issues with assertions and panics triggering the windows exception handlers when they shouldn't.
- There were a few changes made to the appveyor file including some comments and changes to how artifacts are published, and tests executed

I have run these tests several times now, and feel confident that we should not have many more fluke failures. There is still more investigation needed...for example, spectests will fail if run singled threaded, but will pass if running multithreaded.

### DIFF
--- a/lib/runtime-core/src/sys/windows/memory.rs
+++ b/lib/runtime-core/src/sys/windows/memory.rs
@@ -33,7 +33,7 @@ impl Memory {
 
         let protect = protection.to_protect_const();
 
-        let ptr = unsafe { VirtualAlloc(ptr::null_mut(), size, MEM_RESERVE | MEM_COMMIT, protect) };
+        let ptr = unsafe { VirtualAlloc(ptr::null_mut(), size, MEM_RESERVE, protect) };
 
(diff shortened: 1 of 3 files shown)


lib/runtime-core/src/sys/windows/memory.rs
```diff
@@ -33,7 +33,7 @@ impl Memory {
 
         let protect = protection.to_protect_const();
 
-        let ptr = unsafe { VirtualAlloc(ptr::null_mut(), size, MEM_RESERVE | MEM_COMMIT, protect) };
```
lachlansneff: This still works, even though we're not committing the memory?

xmclark: yes

PR #226, opened 2019-03-01
Remove debug option from run command.

calavera:
I was looking into the code and I noticed that this option is not used.
The `debug!` macro is used across the codebase, which looks more ideal.


syrusakbary:
The appveyor builds are flaky right now and we have a fix for it that will land today.
Once it's fixed we will review better this PR.

PR #232, opened 2019-03-04
fix(runtime-c-api) Fix several warnings

Hywan:
This patch fixes several warnings: Some generated by `rustc`, some other generated by Clippy.

I tried to generate 3 “clean” commits:

  1. To remove unsafe blocks inside unsafe functions,
  2. To remove unmutated mutable bindings,
  3. To resolve Clippy warnings.

I hope those changes are OK for you. I will comment the patch a little bit to provide further help.

### DIFF
--- a/lib/runtime-c-api/src/lib.rs
+++ b/lib/runtime-c-api/src/lib.rs
@@ -302,8 +302,7 @@ pub extern "C" fn wasmer_table_grow(
 #[no_mangle]
 pub extern "C" fn wasmer_table_length(table: *mut wasmer_table_t) -> uint32_t {
     let table = unsafe { &*(table as *mut Table) };
-    let len = table.size();
-    len
+    table.size()
 }
 
 /// Frees memory for the given Table
@@ -355,11 +354,10 @@ pub extern "C" fn wasmer_global_get_descriptor(
 ) -> wasmer_global_descriptor_t {
     let global = unsafe { &*(global as *mut Global) };
     let descriptor = global.descriptor();
-    let desc = wasmer_global_descriptor_t {
+    wasmer_global_descriptor_t {
         mutable: descriptor.mutable,
         kind: descriptor.ty.into(),
-    };
-    desc
+    }
 }
 
 /// Frees memory for the given Global
@@ -451,7 +449,7 @@ pub unsafe extern "C" fn wasmer_module_instantiate(
 
         let namespace = namespaces
             .entry(module_name)
-            .or_insert_with(|| Namespace::new());
+            .or_insert_with(Namespace::new);
 
         let export = match import.tag {
             wasmer_import_export_kind::WASM_MEMORY => {
@@ -544,9 +542,9 @@ pub unsafe extern "C" fn wasmer_export_descriptors_get(
     }
     let named_export_descriptors =
         &mut *(export_descriptors as *mut NamedExportDescriptors);
-    let ptr = &mut (*named_export_descriptors).0[idx as usize] as *mut NamedExportDescriptor
-        as *mut wasmer_export_descriptor_t;
-    ptr
+    &mut (*named_export_descriptors).0[idx as usize]
+        as *mut NamedExportDescriptor
+        as *mut wasmer_export_descriptor_t
 }
 
 /// Gets name for the export descriptor
@@ -717,9 +715,9 @@ pub unsafe extern "C" fn wasmer_import_descriptors_get(
     }
     let named_import_descriptors =
         &mut *(import_descriptors as *mut NamedImportDescriptors);
-    let ptr = &mut (*named_import_descriptors).0[idx as usize] as *mut NamedImportDescriptor
-        as *mut wasmer_import_descriptor_t;
-    ptr
+    &mut (*named_import_descriptors).0[idx as usize]
+        as *mut NamedImportDescriptor
+        as *mut wasmer_import_descriptor_t
 }
 
 /// Gets name for the import descriptor
@@ -810,7 +808,7 @@ pub unsafe extern "C" fn wasmer_instantiate(
 
         let namespace = namespaces
             .entry(module_name)
-            .or_insert_with(|| Namespace::new());
+            .or_insert_with(Namespace::new);
 
         let export = match import.tag {
             wasmer_import_export_kind::WASM_MEMORY => {
@@ -900,7 +898,7 @@ pub unsafe extern "C" fn wasmer_instance_call(
 
     match result {
         Ok(results_vec) => {
-            if results_vec.len() > 0 {
+            if !results_vec.is_empty() {
                 let ret = match results_vec[0] {
                     Value::I32(x) => wasmer_value_t {
                         tag: wasmer_value_tag::WASM_I32,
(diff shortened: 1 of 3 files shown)


lib/runtime-c-api/src/lib.rs
```diff
@@ -17,23 +17,23 @@ use wasmer_runtime_core::module::{ExportIndex, ImportName};
 use wasmer_runtime_core::types::{ElementType, FuncSig, MemoryDescriptor, TableDescriptor, Type};
 use wasmer_runtime_core::units::{Bytes, Pages};
 
-#[allow(non_camel_case_types)]
-pub struct wasmer_module_t();
+#[repr(C)]
```
Hywan: `repr(C)` implies `no_mangle` and `allow(non_camel_case_types)` as my tests shown. Not sure, but it has removed several warnings.

lib/runtime-c-api/src/lib.rs
```diff
@@ -17,23 +17,23 @@ use wasmer_runtime_core::module::{ExportIndex, ImportName};
...
+pub struct wasmer_module_t;
```
Hywan: I've removed `()` at the end of the structure declaration. I don't think they are required.

lib/runtime-c-api/src/lib.rs
```diff
@@ -164,8 +165,8 @@ pub unsafe extern "C" fn wasmer_validate(
     if wasm_bytes.is_null() {
         return false;
     }
-    let bytes: &[u8] =
-        unsafe { ::std::slice::from_raw_parts_mut(wasm_bytes, wasm_bytes_len as usize) };
+    let bytes: &[u8] = ::std::slice::from_raw_parts_mut(wasm_bytes, wasm_bytes_len as usize);
```
Hywan: A typical use of an `unsafe` block inside an `unsafe` function?

lib/runtime-c-api/src/lib.rs
```diff
@@ -180,7 +181,7 @@ pub unsafe extern "C" fn wasmer_validate(
 /// and `wasmer_last_error_message` to get an error message.
 #[no_mangle]
 pub unsafe extern "C" fn wasmer_memory_new(
-    mut memory: *mut *mut wasmer_memory_t,
+    memory: *mut *mut wasmer_memory_t,
```
Hywan: `memory` contains a mutable pointer, but the binding itself is not mutable. `rustc` generates a warning for that.

lib/runtime-c-api/src/lib.rs
```diff
@@ -301,8 +302,7 @@ pub extern "C" fn wasmer_table_grow(
 #[no_mangle]
 pub extern "C" fn wasmer_table_length(table: *mut wasmer_table_t) -> uint32_t {
     let table = unsafe { &*(table as *mut Table) };
-    let len = table.size();
-    len
+    table.size()
```
Hywan: Clippy suggests to return the value directly instead of the binding.

lib/runtime-c-api/src/lib.rs
```diff
@@ -451,21 +449,21 @@ pub unsafe extern "C" fn wasmer_module_instantiate(
 
         let namespace = namespaces
             .entry(module_name)
-            .or_insert_with(|| Namespace::new());
+            .or_insert_with(Namespace::new);
```
Hywan: Passing the function name directly instead of a closure saves work for the optimizer.

lib/runtime-c-api/src/lib.rs
```diff
@@ -451,21 +449,21 @@ pub unsafe extern "C" fn wasmer_module_instantiate(
...
         };
-        namespace.insert(import_name, unsafe { (&*export).clone() });
+        namespace.insert(import_name, (&*export).clone());
     }
     for (module_name, namespace) in namespaces.into_iter() {
         import_object.register(module_name, namespace);
     }
 
-    let module = unsafe { &*(module as *mut Module) };
+    let module = &*(module as *const Module);
```
Hywan: I _think_ it's not necessary to get a mutable module here. Correct me if I misunderstood the code!

lib/runtime-c-api/src/lib.rs
```diff
@@ -1436,14 +1421,12 @@ impl From<Type> for wasmer_value_tag {
 
 impl From<wasmer_value_tag> for Type {
     fn from(v: wasmer_value_tag) -> Self {
-        unsafe {
```
Hywan: I don't understand why the `match` statement is in an `unsafe` block. Did I miss something?

lib/runtime-c-api/src/lib.rs
```diff
@@ -1436,14 +1421,12 @@ impl From<Type> for wasmer_value_tag {
...
-            match v {
-                wasmer_value_tag::WASM_I32 => Type::I32,
-                wasmer_value_tag::WASM_I64 => Type::I64,
-                wasmer_value_tag::WASM_F32 => Type::F32,
-                wasmer_value_tag::WASM_F64 => Type::F64,
-                _ => panic!("not implemented"),
```
Hywan: If I get it correctly, the `_` arm is unreachable, but we don't trust the `v` input, is that right? In other words, we don't assume that `v` is correctly built.

lib/runtime-c-api/src/lib.rs
bjfish: The C API is fairly new and many things like this needed clean up.

lib/runtime-c-api/src/lib.rs
bjfish: 👍

Hywan:
Rebased on top of `master`.

lib/runtime-c-api/src/lib.rs
bjfish: Yes, these needed clean up after unsafe was added to the function declaration.

lib/runtime-c-api/src/lib.rs
bjfish: 👍

lib/runtime-c-api/src/lib.rs
bjfish: 👍, also when changes are made that may affect the header files you can run `make capi` from the root project. That and running tests is described in this projects `README`. Please run `make capi` for this PR if you  haven’t already.

Hywan: Done. I think we have an issue :-).

lib/runtime-c-api/wasmer.h
```diff
@@ -27,11 +27,9 @@ enum wasmer_value_tag {
 };
 typedef uint32_t wasmer_value_tag;
 
-typedef struct wasmer_instance_context_t wasmer_instance_context_t;
```
Hywan: This is smelly.

lib/runtime-c-api/wasmer.hh
```diff
@@ -24,11 +24,9 @@ enum class wasmer_value_tag : uint32_t {
   WASM_F64,
 };
 
-struct wasmer_instance_context_t;
```
Hywan: This is also smelly.

syrusakbary:
Once the conflicts are resolved (I think they were caused by merging #227), and we check cbindgen we should be good to merge.

lib/runtime-c-api/wasmer.h
Hywan: Oh, it's still here but below.

lib/runtime-c-api/wasmer.hh
Hywan: Still here, but below.

Hywan:
Everything is alright for me :+1:.

Hywan:
I think we need to update the C source file, because some types have changed (like `const *` for instance). It only generates warnings, but we need to fix them.

PR #233, opened 2019-03-04
implement more emscripten functions

xmclark:
This PR adds a few more functions for emscripten. It is the accumulation of some of the work I did last week that I think is ready for addition to Wasmer. This puts us a couple steps closer to running more "native-like" apps in Wasmer with Emscripten target.


lib/emscripten/src/exec.rs
```diff
@@ -0,0 +1,40 @@
+use libc::execvp;
+use std::cell::Cell;
+use std::ffi::CString;
+use wasmer_runtime_core::vm::Ctx;
+
+pub fn _execvp(ctx: &mut Ctx, command_name_offset: u32, argv_offset: u32) -> i32 {
```
Hywan: Why only one underscore, whilst the other ones have two? Shouldn't it be `__excevp`?

xmclark: I think I was trying to mirror the exact function name on msdn which uses a single underscore. Later I read the unix docs and there is *no* underscore. 

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/execvp-wexecvp?view=vs-2017
https://linux.die.net/man/3/execvp

Do you have a preference? Should we remove the underscore to match the unix docs?

Hywan: I don't see a valid reason to keep underscores for all functions actually. Maybe we can drop them.

PR #238, opened 2019-03-05
feat(runtime-c-api) Generate the C/C++ header files in `OUT_DIR` (+ copy in `CARGO_MANIFEST_DIR`).

Hywan:
This patch changes the directory where the C/C++ header files are
generated, from `CARGO_MANIFEST_DIR` to `OUT_DIR`. The goal is: When
`wasm-runtime-c-api` is used as a dependency of another Rust project,
then the C header files are accessible in the `target/` directory
(i.e. the `OUT_DIR`).

A copy is kept in the `CARGO_MANIFEST_DIR` file, so that they can be
versionnized.

Also, since `rustc` has a `--out-dir` directory, it increases the
flexibility of this approach: The user can generate the C header files
where she wants.

The `WASM_EMSCRIPTEN_GENERATE_C_API_HEADERS` environement variable has
been removed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,11 +115,13 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "cbindgen"
-version = "0.7.1"
+version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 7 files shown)


Hywan:
I agree to copy the C header files in the source directory.

I agree to remove the need for `WASM_EMSCRIPTEN_GENERATE_C_API_HEADERS`.

Hywan:
(Rebased on top of `master`).

PR #241, opened 2019-03-06
test(runtime-c-api) Deny all warnings

Hywan:
Waiting on #238. Will be rebased later.

This patch transforms all warnings into errors, so that a good code quality is ensured. Warnings in C or C++ can lead to catastrophic situations sometimes. Be strict regarding this.


lib/runtime-c-api/src/lib.rs
```diff
@@ -1241,7 +1241,7 @@ pub unsafe extern "C" fn wasmer_import_func_returns_arity(
 /// Frees memory for the given Func
 #[allow(clippy::cast_ptr_alignment)]
 #[no_mangle]
-pub extern "C" fn wasmer_import_func_destroy(func: *mut wasmer_import_func_t) {
```
bjfish: Should this one be `mut` since it is cast to `mut` and `drop` takes `&mut` ?

Hywan: Yeah, that's the only one I wasn't sure about. Thanks for the notice!

PR #243, opened 2019-03-06
chore(github) Update issue templates

Hywan:
This patch provides issue templates to get a more welcoming environment for our users. It's a mix between Github defaults, and https://github.com/rustwasm/wasm-bindgen templates.

Thoughts?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,35 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,35 @@
+---
```
Hywan: I don't know why the Github UI has created this duplicated file… 🤔

syrusakbary:
> I think we should list the commands needed to collect the wasmer version, rust version, target triple, etc.

I will do this in a PR apart :)

PR #245, opened 2019-03-07
[WIP] Get table from Ctx

tprk77:
Based on some [discussion on Spectrum](https://spectrum.chat/wasmer/general/is-it-possible-to-call-a-wasm-callback-from-rust~eee307f9-fdf9-4182-876c-59cf3395794c), this change adds a `table` function to `Ctx`.

Feedback appreciated!

### DIFF
--- a/lib/runtime-core/src/table/anyfunc.rs
+++ b/lib/runtime-core/src/table/anyfunc.rs
@@ -128,4 +128,8 @@ impl AnyfuncTable {
             Err(())
         }
     }
+
+    pub fn get(&mut self, index: u32) -> Option<vm::Anyfunc> {
+        self.backing.get(index as usize).map(|slot| *slot)
+    }
 }
--- a/lib/runtime-core/src/table/anyfunc.rs
+++ b/lib/runtime-core/src/table/anyfunc.rs
@@ -115,7 +115,7 @@ impl AnyfuncTable {
 
                     vm::Anyfunc {
                         func: func.raw(),
-                        ctx: func.instance_inner.vmctx,
+                        ctx: func.vmctx,
                         sig_id,
                     }
                 }
--- a/lib/runtime-core/src/instance.rs
+++ b/lib/runtime-core/src/instance.rs
@@ -206,7 +206,7 @@ impl Instance {
                 module: &self.module,
                 vmctx,
                 import_backing: &self.inner.import_backing,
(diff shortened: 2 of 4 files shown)


lachlansneff:
Looks great! Does it work for the specific issue you had?

tprk77:
I think there is a bit more to do here, but I'm still figuring it out. For example, `Table` has a `set` function, but I think now I need a `get` function. (Same thing for the `AnyfuncTable`.)

lachlansneff:
Yeah, I threw together the `Table` type pretty quickly without thinking about it too much. If you see a better way of doing it, we can definitely change it! :)

lib/runtime-core/src/table/anyfunc.rs
```diff
@@ -128,4 +128,12 @@ impl AnyfuncTable {
             Err(())
         }
     }
+
+    pub fn get(&mut self, index: u32) -> Result<vm::Anyfunc, ()> {
+        if let Some(slot) = self.backing.get(index as usize) {
+            Ok(*slot)
+        } else {
+            Err(())
```
syrusakbary: I think it might be a good idea to return an Option rather than a Result (since there is only one kind of error: if the function doesn't exist).

tprk77:
@lachlansneff I've had some time to look at this again, and I've been able to create and call a `DynFunc` from a `vm::Ctx`. But there's a problem. It appears that `ProtectedCaller` is not designed to work with non-export functions. I found this comment a little too late:

```rust
/// The functionality exposed by this trait is expected to be used
/// for calling functions exported by a webassembly module from
/// host code only.
pub trait ProtectedCaller: Send + Sync {
```

When I try to use `DynFunc` with a non-exported function, I get, predictably:

```
thread 'main' panicked at 'assertion failed: self.func_export_set.contains(&func_index)',
/.../wasmer/lib/clif-backend/src/signal/mod.rs:88:9
```

So I don't think this `DynFunc` approach is going to work, or anything else using `ProtectedCaller::call`. It seems like I actually need something like `ProtectedCaller` but a little more raw. I'm picturing something that takes a function pointer and a context directly. (Then does the trampoline, etc.)

Do you think that's a reasonable idea? Is that something I should try implementing?

tprk77:
On second thought, this is all very out of date, and I don't think any of it is salvageable.

PR #248, opened 2019-03-07
Add Compile command.

calavera:
This command allows people to compile a wasm module to machine code ahead
of time. It dumps the data in the cache, so future runs can use it.

It also changes the Cache trait to use Strings as keys, so it's more flexible than using the WasmHash object directly.

I've added a --from-cache flag to the Run command to allow loading modules directly from the cache without having access to the wasm or wat files. I'm not super sold on this flag.

I'm super open to suggestions and other options.

Enjoy this doggo taking a bath:

### DIFF
--- a/src/bin/wasmer.rs
+++ b/src/bin/wasmer.rs
@@ -37,9 +37,13 @@ enum CLIOptions {
 
 #[derive(Debug, StructOpt)]
 struct Run {
-    // Disable the cache
+    /// Disable the cache
     #[structopt(long = "disable-cache")]
     disable_cache: bool,
+
+    /// Load module from the cache
+    #[structopt(long = "from-cache")]
+    from_cache: Option<PathBuf>,
 
     /// Input file
     #[structopt(parse(from_os_str))]
@@ -113,15 +117,12 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
     #[cfg(not(target_os = "windows"))]
     let disable_cache = options.disable_cache;
 
-    let wasm_binary = load_wasm_binary(&options.path)?;
-
-    let module = if !disable_cache {
+    let module = if disable_cache {
+            let wasm_binary = load_wasm_binary(&options.path)?;
+            webassembly::compile(&wasm_binary[..])
+                .map_err(|e| format!("Can't compile module: {:?}", e))?
+    } else {
         // If we have cache enabled
-
-        // We generate a hash for the given binary, so we can use it as key
-        // for the Filesystem cache
-        let hash = WasmHash::generate(&wasm_binary);
-
         let wasmer_cache_dir = get_cache_dir();
 
         // We create a new cache instance.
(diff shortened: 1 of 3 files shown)


Hywan:
Thanks for the great PR!

I think the ideal workflows should be the following:

  1. `wasmer run <file.wasm>` to run a Wasm binary file. A cache of the compiled version is created on-the-fly, and re-used for the next runs,
  2. `wasmer compile <file.wasm>` to compile a Wasm file in the cache ahead of time, then simply run `wasm run <file.wasm>` would fetch it from the cache, as if it was a “next run” for the previous item.

Based on these assumptions, the `--from-cache` option is not necessary.

Thoughts?

calavera:
@Hywan that means that you need the wasm file where you compile and where you execute the module, that's not always the same servers.

calavera:
> Ignoring the required option seems confusing. Maybe a dedicated subcommand for running from a cache?

I thought about that too, and it would make sense because this feature is dependent on the cache. Do you have any suggestions for the subcomand name @xmclark ?

What about this?
```
wasmer cache exec WASM_HASH
```

lachlansneff:
My opinion is that we provide two cache subcommands:

`wasmer cache generate <wasm file>`
- takes an optional output directory argument
- outputs the path to the cache file on stdout.

and

`wasmer cache run foo/bar/hash`
- Takes a path to the cache as the argument.

calavera:
I'm also realizing that this might need the wasm path, or at least a name for the module to be executed. I'm thinking about what's the best way to solve that.

calavera:
> My opinion is that we provide two cache subcommands

I like this suggestion

calavera:
@lachlansneff, @xmclark I've implemented those two subcommands, tell me if this makes sense:

```
- `wasmer cache generate` to generate a machine object.
- `wasmer cache run` to run a cached object.

The run command requires two arguments. The first one is the name of the
program to run, to pass it as args[0] to the module, and the key for the
object in the cache. It also allows to send extra arguments to pass them
to the module. Example:

$ wasmer cache run nginx VERY_LONG_HASH -- -c /opt/nginx/config
```

I can write documentation for those subcommands if this is a good way to implement this.

src/bin/wasmer.rs
```diff
@@ -129,48 +194,81 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
...
+        FileSystemCache::new(wasmer_cache_dir).map_err(|e| format!("Cache error: {:?}", e))?
+    };
+
+    let key = String::from(options.module_key.to_str().unwrap());
+    let module = cache
+        .load(key)
+        .map_err(|e| format!("Can't execute module from cache: {:?}", e))?;
+
+    let name = String::from(options.module_name.to_str().unwrap());
+    run_wasm_module(&module, name, &options.args)
```
xmclark: A small nitpick: the `module_name` is unecessary on the CLI. Can you pass the hash as a string to `run_wasm_module` in place of the name? This will allow us to remove the `module_name` field from `CacheRun` too. 

This way all we have to type is:
`$ wasmer cache run VERY_LONG_HASH -- -c /opt/nginx/config`

calavera: I can do that, I was not sure if it'd be ok :+1:

lachlansneff:
bors try

calavera:
I've removed that extra argument in the `wasmer cache run` command. Do you have any conventions to write documentation? I don't see any docs in this repo besides the readme.

calavera:
@xmclark @lachlansneff it looks like Circle failed to download a dependency, and I cannot trigger a new job.

lachlansneff:
@calavera Seems like a Cargo.lock issue? Try resetting it?

calavera:
it looks like rebasing master did the trick @lachlansneff :confetti_ball:

xmclark:
bors try

calavera:
bors try

bjfish:
bors try

syrusakbary:
A small update on this PR: we are gathering more feedback (and learnings) about the compilation/caching runtime needs, to make sure we expose the right API for it.

We will comment here as soon as we are 100% clear on it (hopefully soon!)

syrusakbary:
We are on the final steps of rewriting Wasmer from scratch and this issue is now solved in the refactor (we have a `compile` subcommand).

syrusakbary:
Hi @calavera!

We finally added support for precompiling wasm files to native object files (or JIT).
The usage is quite easy:

Compile:

```
wasmer compile python.wasm -o python.so --native
```

Run:

```
wasmer run python.so
```

Note: if you use the native engine, the startup time to load them is just in the order of nanoseconds even for big wasm files.

PR #257, opened 2019-03-08
feat(runtime) Get ready for tiering

lachlansneff:
This PR adds various fixes and runtime changes in order to be ready to support compiler tiering.


lib/runtime-core/src/backing.rs
```diff
@@ -62,9 +64,23 @@ impl LocalBacking {
             vm_globals,
 
             dynamic_sigindices,
+            local_functions,
         }
     }
 
+    fn generate_local_functions(module: &ModuleInner) -> BoxedMap<LocalFuncIndex, *const vm::Func> {
+        (0..module.info.func_assoc.len() - module.info.imported_functions.len())
```
xmclark: is this constructing a range? do we have confidence this will always result in a positive set of numbers?

lachlansneff: Yeah, I do. The total number of functions will always be greater than or equal to the number of imported functions.

PR #261, opened 2019-03-11
feat(runtime-c-api) Add an API to update `vm::Ctx.data`

Hywan:
This patch adds 2 functions for the runtime C API, respectively
`wasmer_instance_context_data_set` and
`wasmer_instance_context_data_get`. The goal is to modify the
`vm::Ctx.data` field in the `runtime-core` library.

Since all imported functions receive this context, that's the
simplest way to pass dynamic data to them.


bjfish:
This looks good. Could you please add calling these functions to one of the C tests and assert the get matches what is set?

Hywan:
bors retry

Hywan:
bors retry

PR #262, opened 2019-03-11
create an llvm flag

xmclark:
Some users may not care about building for the LLVM backend, so it may be useful to isolate it with a feature flag. Building should now only build the cranelift backend and does not require any LLVM setup.


lachlansneff:
bors try

lib/emscripten/src/syscalls/windows.rs
```diff
@@ -40,8 +40,8 @@ pub fn ___syscall5(ctx: &mut Ctx, which: c_int, mut varargs: VarArgs) -> c_int {
                 emscripten_memory_pointer!(ctx.memory(0), urandom_file_offset) as *const i8;
             let fd = unsafe { open(raw_pointer_to_urandom_file, flags, mode) };
             debug!(
-                "=> pathname: {}, flags: {}, mode: {} = fd: {}\npath: {}",
```
lachlansneff: Did this stuff just get caught up in this pr?

lachlansneff:
Okay, let's actually merge it
bors r+

lib/emscripten/src/syscalls/windows.rs
xmclark: yeah I have no idea how the build was passing before...makes me suspicious.

PR #264, opened 2019-03-12
fix(runtime-core) `SharedMemory.desc` is never used, remove it

Hywan:
This patch removes the `desc` field of `SharedMemory`. This field is
never used. Maybe it will be used in the future when the
implementations will be written, but so far, it only generates
warnings.


bjfish:
Could we add a tag or something to allow it to be unused for now?

Hywan:
Yup :-).

xmclark:
bors try

PR #269, opened 2019-03-13
docs(runtime) Add some better runtime docs

lachlansneff:


lib/runtime-core/src/backing.rs
```diff
@@ -17,30 +17,31 @@ use crate::{
 };
 use std::slice;
 
+/// The `LocalBacking` "owns" the memory used by all the local resources.
+/// That is, local memories, tables, and globals (as well as some additional
+/// data for the virtual call machinery).
```
bjfish: One suggestion
> The `LocalBacking` "owns" the state used by

Could you clarify if LocalBacking belongs to an instance or module?

lib/runtime-core/src/backing.rs
```diff
@@ -102,6 +103,9 @@ impl LocalBacking {
         memories.into_boxed_map()
     }
 
+    /// Initialize each local memory.
```
bjfish: > Initializes each

Does `local` mean not internal or something else? Could we clarify what local is local to?

lib/runtime-core/src/backing.rs
```diff
@@ -174,6 +178,9 @@ impl LocalBacking {
         tables.into_boxed_map()
     }
 
+    /// This initializes all of the local tables, e.g.
```
bjfish: > Initializes all of the local tables
Consistency with above

lib/runtime-core/src/vm.rs
```diff
@@ -9,7 +9,16 @@ use std::{ffi::c_void, mem, ptr};
 
 /// The context of the currently running WebAssembly instance.
 ///
+/// This is implicitly passed to every webassembly function.
```
bjfish: WebAssembly

lib/runtime-core/src/vm.rs
```diff
@@ -9,7 +9,16 @@ use std::{ffi::c_void, mem, ptr};
...
+/// Since this is per-module, each field has a statically
```
bjfish: Should this be `per-instance`?

Hywan:
Thanks for the PR!

PR #270, opened 2019-03-14
feat(runtime-c-api) `wasmer_validate` expects a `*const uint8_t`

Hywan:
This patch updates the first argument of `wasmer_validate` from `*mut
uint8_t` to `*const uint8_t`. Indeed, the
`wasmer-runtime-core::validate` function doesn't expect a mutable
slice, so it's not required to expect a mutable array from C.

Also, it's likely for the Wasm bytes to be stored in the
`wasmer_byte_array` structure. The first field `bytes` is defined as
`*const uint8_t`. So this patch avoids a cast when writing a C++
program.


Hywan:
bors try

Hywan:
bors try

PR #271, opened 2019-03-15
feat: Implement `wasmer_module_serialize` and `wasmer_module_deserialize` functions in `runtime-c-api`

Hywan:
This PR implements 5 new functions in the `runtime-c-api` crate:

  1. `wasmer_module_serialize` to serialize a module into a `wasmer_serialized_module_t` type,
  2. `wasmer_module_deserialize` to deserialize a serialized module,
  3. `wasmer_serialized_module_bytes` to read the bytes in the `wasmer_serialized_module_t` type into a `wasmer_byte_array`,
  4. `wasmer_serialized_module_from_bytes` to transform a `wasmer_byte_array` into a `wasmer_serialized_module_t`,
  4. `wasmer_serialized_module_destroy` to destroy a `wasmer_serialized_module_t`.

Documentation and a test suite have been added.

We need to change the visibility of the `wasmer_runtime::default_compiler` function to public, since it is used in `runtime-c-api`.

The new test suite `test-module-serialize` does a full roundtrip: A module is compiled, then serialized, then deserialized, to finally be instantiated and a function is called on it.

Thoughts?


lib/runtime-c-api/src/lib.rs
```diff
@@ -569,6 +574,99 @@ pub unsafe extern "C" fn wasmer_export_descriptor_kind(
...
+        return wasmer_result_t::WASMER_ERROR;
+    }
+
+    let serialized_module: &[u8] = slice::from_raw_parts(
+        serialized_module_bytes,
+        serialized_module_bytes_len as usize,
+    );
+
+    match Artifact::deserialize(serialized_module) {
+        Ok(artifact) => match load_cache_with(artifact, default_compiler()) {
```
Hywan: If we do not want to make `default_compiler` public, then I can create a new function in `wasmer_runtime_core`:

```rust
pub unsafe fn load_cache(cache: Artifact) -> std::result::Result<module::Module, CacheError> {
    load_cache_with(cache, default_compiler())
}
```

Thoughts?

lib/runtime-c-api/src/lib.rs
```diff
@@ -569,6 +574,99 @@ pub unsafe extern "C" fn wasmer_export_descriptor_kind(
...
+        }
+    }
+}
+
+/// Deserialize the given bytes into a Module.
+///
+/// Returns `wasmer_result_t::WASMER_OK` upon success.
+///
+/// Returns `wasmer_result_t::WASMER_ERROR` upon failure. Use `wasmer_last_error_length`
+/// and `wasmer_last_error_message` to get an error message.
```
bjfish: I like to mention the _destroy functions in the documentation for functions that produce things that should be cleaned up later.

lib/runtime-c-api/src/lib.rs
bjfish: I’d be fine either way. Seems like a good idea to make default compiler public.

xmclark: a public `default_compiler` is cool with me

lib/runtime-c-api/src/lib.rs
```diff
@@ -569,6 +574,99 @@ pub unsafe extern "C" fn wasmer_export_descriptor_kind(
...
+
+    match module.cache() {
+        Ok(artifact) => match artifact.serialize() {
+            Ok(serialized_artifact) => {
+                *serialized_module = Box::into_raw(Box::new(wasmer_byte_array {
+                    bytes: serialized_artifact.as_ptr(),
+                    bytes_len: serialized_artifact.len() as u32,
+                })) as *mut wasmer_byte_array;
+
+                mem::forget(serialized_artifact);
```
lachlansneff: Does this leak memory?

lib/runtime/src/lib.rs
```diff
@@ -152,7 +152,8 @@ pub fn instantiate(wasm: &[u8], import_object: &ImportObject) -> error::Result<I
     module.instantiate(import_object)
 }
 
-fn default_compiler() -> &'static dyn Compiler {
+/// Get a single instance of the default compiler to use.
+pub fn default_compiler() -> &'static dyn Compiler {
```
lachlansneff: Ideally, this shouldn't need to use the default compiler, but it that's easiest, it's okay.

lib/runtime-c-api/src/lib.rs
Hywan: I will add a `wasmer_serialized_module_destroy` function.

lib/runtime-c-api/src/lib.rs
Hywan: Yes. If we don't, then the `wasmer_byte_array.bytes` field will contain a null pointer. It's up to the caller to free the memory.

lib/runtime-c-api/src/lib.rs
```diff
@@ -569,6 +574,106 @@ pub unsafe extern "C" fn wasmer_export_descriptor_kind(
...
+            });
+            wasmer_result_t::WASMER_ERROR
+        }
+    }
+}
+
+/// Frees memory for the given serialized Module.
+#[allow(clippy::cast_ptr_alignment)]
+#[no_mangle]
+pub extern "C" fn wasmer_module_serialization_destroy(serialized_module: *mut wasmer_byte_array) {
```
Hywan: @bjfish What do you think of that?

bjfish: Looks good but would `wasmer_serialized_module_destroy` be a better name?

lib/runtime-c-api/src/lib.rs
bjfish: @Hywan how about a `wasmer_serialized_module_t *` that points to the `serialized_artifact` on the rust side and a function to access the bytes `wasmer_byte_array wasmer_serialized_module_bytes(wasmer_module_t *serialized_module)`?  This would be similar to this: https://github.com/Hywan/wasmer/blob/f049e9b0b378ae52b6cef467364f25fcff5ec2d8/lib/runtime-c-api/src/lib.rs#L557

Hywan: That's a good idea!

lib/runtime-c-api/src/lib.rs
Hywan: Done.

lib/runtime-c-api/src/lib.rs
Hywan: Done.

Hywan:
I've added the `wasmer_serialized_module_from_bytes` function. The PR description has been updated accordingly. This function is required to transform a `wasmer_byte_array` into a `wasmer_serialized_module_t`. Indeed, when storing the serialized module bytes into a file for instance, one get the same bytes when reading the file and has to transform them into a `wasmer_serialized_module_t` so that it can be passed to the `wasmer_module_deserialize` function.

lib/runtime-c-api/src/lib.rs
```diff
@@ -569,12 +574,149 @@ pub unsafe extern "C" fn wasmer_export_descriptor_kind(
...
+///
+/// Returns `wasmer_result_t::WASMER_OK` upon success.
+///
+/// Returns `wasmer_result_t::WASMER_ERROR` upon failure. Use `wasmer_last_error_length`
+/// and `wasmer_last_error_message` to get an error message.
+#[allow(clippy::cast_ptr_alignment)]
+#[no_mangle]
+pub unsafe extern "C" fn wasmer_serialized_module_from_bytes(
+    serialized_module: *mut *mut wasmer_serialized_module_t,
+    serialized_module_bytes: *const wasmer_byte_array,
```
bjfish: Alternatively you could consider using the following format for bytes that I use elsewhere, for example:
```
pub unsafe extern "C" fn wasmer_compile(
    module: *mut *mut wasmer_module_t,
    wasm_bytes: *mut uint8_t,
    wasm_bytes_len: uint32_t,
```

lib/runtime-c-api/wasmer.h
```diff
@@ -504,6 +517,36 @@ wasmer_result_t wasmer_module_instantiate(const wasmer_module_t *module,
...
+
+/**
+ * Transform a sequence of bytes into a serialized module.
+ * The caller owns the object and should call `wasmer_serialized_module_destroy` to free it.
+ * Returns `wasmer_result_t::WASMER_OK` upon success.
+ * Returns `wasmer_result_t::WASMER_ERROR` upon failure. Use `wasmer_last_error_length`
+ * and `wasmer_last_error_message` to get an error message.
+ */
+wasmer_result_t wasmer_serialized_module_from_bytes(wasmer_serialized_module_t **serialized_module,
+                                                    const wasmer_byte_array *serialized_module_bytes);
```
bjfish: It may also be convenient to go directly to `wasmer_module_t` from bytes.

Hywan: As discussed in private, let's see if we need it in the future.

lib/runtime-c-api/src/lib.rs
Hywan: Done.

PR #274, opened 2019-03-15
feat(runtime-core) Nicer panic output

lachlansneff:
This adds the human-panic crate to print out a nicer panic message (as well as log the backtrace and whatnot).

Example output:

```
Well, this is embarrassing.

The Wasmer Runtime had a problem and crashed. To help us diagnose the problem you can send us a crash report.

We have generated a report file at "/var/folders/24/mfj_dsx13ln7jxrnrbb2vq0h0000gn/T/report-5012e1cd-3834-4ff2-8e3f-5b42c5735ca3.toml". Submit an issue or email with the subject of "The Wasmer Runtime Crash Report" and include the report as an attachment.

- Homepage: https://github.com/wasmerio/wasmer
- Authors: Wasmer Engineering <engineering@wasmer.io>

We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.

Thank you kindly!
```

### DIFF
--- a/lib/runtime-core/Cargo.toml
+++ b/lib/runtime-core/Cargo.toml
@@ -7,6 +7,9 @@ authors = ["The Wasmer Engineering Team <engineering@wasmer.io>"]
 repository = "https://github.com/wasmerio/wasmer"
 edition = "2018"
 
+[lib]
+crate-type = ["dylib"]
+
 [dependencies]
 nix = "0.12.0"
 page_size = "0.4.1"
@@ -17,6 +20,8 @@ indexmap = "1.0.2"
 errno = "0.2.4"
 libc = "0.2.49"
 hex = "0.3.2"
+human-panic = "1.0"
+ctor = "0.1.7"
 
 # Dependencies for caching.
 [dependencies.serde]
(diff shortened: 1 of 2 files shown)


lachlansneff:
bors try

syrusakbary:
Actually, I created an account on Sentry. It might be a good idea to send the error there as well? (and remove the "automated error collection" phrase :P )

lachlansneff:
I don't think we should automatically log the errors. It seems too much like telemetry. We can probably add stuff about submitting a report on sentry though.

lachlansneff:
bors try

CryZe:
Shouldn't this be behind a feature gate? This is great for wasmer, but not so great for other applications using wasmer as a dependency.

lachlansneff:
@CryZe Yeah, it seems like having this installed by default is a bad idea. Also, it doesn't always get installed (depends on the type of crate that depends on the runtime). I'll install this in the wasmer cli app, and let users do install it from the runtime if they want.

syrusakbary:
Closing the PR, as is stale.
We will re-do it when becomes a priority again

PR #276, opened 2019-03-17
Single-pass native code generation for x86-64 using dynasm.

losfair:


lib/dynasm-backend/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+    vm,
+};
+
+struct Placeholder;
+impl CacheGen for Placeholder {
+    fn generate_cache(
+        &self,
+        _module: &ModuleInner,
+    ) -> Result<(Box<ModuleInfo>, Box<[u8]>, Memory), CacheError> {
+        // unimplemented!()
```
bjfish: Please remove commented out code.

src/webassembly.rs
```diff
@@ -86,7 +87,11 @@ pub fn run_instance(
     if is_emscripten_module(module) {
         run_emscripten_instance(module, instance, path, args)?;
     } else {
-        instance.call("main", &[])?;
+        let args: Vec<Value> = args
+            .into_iter()
+            .map(|x| Value::I32(x.parse().unwrap()))
+            .collect();
+        println!("{:?}", instance.call("main", &args)?);
```
bjfish: I think this `println!` could be removed.

lib/runtime/src/lib.rs
```diff
@@ -159,7 +159,8 @@ fn default_compiler() -> &'static dyn Compiler {
     use wasmer_llvm_backend::LLVMCompiler as DefaultCompiler;
 
     #[cfg(not(feature = "llvm"))]
-    use wasmer_clif_backend::CraneliftCompiler as DefaultCompiler;
+    use wasmer_dynasm_backend::SinglePassCompiler as DefaultCompiler;
+    // use wasmer_clif_backend::CraneliftCompiler as DefaultCompiler; // TODO Fix default
```
bjfish: I think `Cranelift` should still be default for now?

bjfish:
In the "Files Changed" I saw " Submodule cranelift added at cb62a1" which should not be added.

bjfish:
bors try

bjfish:
@losfair I removed the cranelift submodule following these steps:
https://gist.github.com/myusuf3/7f645819ded92bda6677
You may have things to delete locally for this as well.

losfair:
bors try

losfair:
bors try

losfair:
@lachlansneff The LLVM backend did not pass some of the spectests.

lachlansneff:
I think some changes in the master got overwritten here. Can we rebase on top of master again or something?

lib/dynasm-backend/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+impl FuncResolver for Placeholder {
+    fn get(
+        &self,
+        _module: &ModuleInner,
+        _local_func_index: LocalFuncIndex,
+    ) -> Option<NonNull<vm::Func>> {
+        NonNull::new(0x3f3f3f3f3f3f3f3fusize as *mut vm::Func)
```
lachlansneff: This should get removed, no?

lib/dynasm-backend/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+pub struct SinglePassCompiler {}
+impl SinglePassCompiler {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl Compiler for SinglePassCompiler {
+    fn compile(&self, wasm: &[u8], _: Token) -> CompileResult<ModuleInner> {
+        let mut mcg = codegen_x64::X64ModuleCodeGenerator::new();
```
lachlansneff: Can we cfg this to error at compile-time if the target arch isn't supported?

lib/runtime-core/src/backing.rs
```diff
@@ -230,7 +230,12 @@ impl LocalBacking {
                                 }
                             };
 
-                            elements[init_base + i] = vm::Anyfunc { func, ctx, sig_id };
+                            elements[init_base + i] = vm::Anyfunc {
+                                func,
+                                ctx,
+                                sig_id,
+                                func_index: Some(func_index),
```
lachlansneff: Can you remove the func_index here? None of the other compilers expect an additional field.

lib/runtime-core/src/instance.rs
```diff
@@ -434,7 +434,7 @@ pub struct DynFunc<'a> {
     pub(crate) signature: Arc<FuncSig>,
     module: &'a ModuleInner,
     pub(crate) instance_inner: &'a InstanceInner,
-    func_index: FuncIndex,
+    pub func_index: FuncIndex,
```
lachlansneff: Why does this need to be public?

lib/runtime-core/src/table/anyfunc.rs
```diff
@@ -107,6 +107,7 @@ impl AnyfuncTable {
                         func: ptr,
                         ctx: ptr::null_mut(),
                         sig_id,
+                        func_index: None,
```
lachlansneff: Ditto

lib/runtime-core/src/table/anyfunc.rs
```diff
@@ -117,6 +118,7 @@ impl AnyfuncTable {
                         func: func.raw(),
                         ctx: func.instance_inner.vmctx,
                         sig_id,
+                        func_index: Some(func.func_index),
```
lachlansneff: Ditto

lib/runtime-core/src/vm.rs
```diff
@@ -14,25 +14,25 @@ use std::{ffi::c_void, mem, ptr};
 #[repr(C)]
 pub struct Ctx {
     /// A pointer to an array of locally-defined memories, indexed by `MemoryIndex`.
-    pub(crate) memories: *mut *mut LocalMemory,
+    pub memories: *mut *mut LocalMemory,
```
lachlansneff: Is it necessary for these to be public outside of the crate?

lib/dynasm-backend/src/lib.rs
losfair: It is already checked at L3-L7.

lib/runtime-core/src/vm.rs
losfair: These fields need to be public to be accessed from the codegen backend. Otherwise we have to access these fields by offsets which is not clean.

losfair:
There's no difference in `lib/llvm-backend` from master.

lib/runtime-core/src/vm.rs
lachlansneff: The ctx has been designed to be usable with falling back on rust. Can you use either offsets or implement it in assembly so that it doesn't need to call rust to do call_indirect?

lib/runtime-core/src/error.rs
```diff
@@ -20,6 +25,14 @@ pub enum CompileError {
     InternalError { msg: String },
 }
 
+impl From<BinaryReaderError> for CompileError {
```
lachlansneff: Please remove this. The runtime-core should not depend on wasmparser.

lib/runtime-core/src/types.rs
```diff
@@ -14,6 +18,21 @@ pub enum Type {
     F64,
 }
 
+impl Type {
+    pub fn from_wasmparser_type(other: ::wasmparser::Type) -> CompileResult<Type> {
```
lachlansneff: And this as well.

lib/runtime-core/src/types.rs
```diff
@@ -235,8 +254,8 @@ impl MemoryDescriptor {
     pub fn memory_type(self) -> MemoryType {
         match (self.maximum.is_some(), self.shared) {
             (true, true) => MemoryType::SharedStatic,
-            (true, false) => MemoryType::Static,
-            (false, false) => MemoryType::Dynamic,
+            (true, false) | (false, false) => MemoryType::Static,
```
lachlansneff: Also, please set this back to its original state.

lib/runtime-core/src/vm.rs
```diff
@@ -3,7 +3,7 @@ use crate::{
     memory::Memory,
     module::ModuleInner,
     structures::TypedIndex,
-    types::{LocalOrImport, MemoryIndex},
+    types::{FuncIndex, LocalOrImport, MemoryIndex},
```
lachlansneff: I don't think `FuncIndex` needs to be imported here.

src/bin/wasmer.rs
```diff
@@ -128,8 +128,9 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
                 let module = webassembly::compile(&wasm_binary[..])
                     .map_err(|e| format!("Can't compile module: {:?}", e))?;
 
-                // We save the module into a cache file
-                cache.store(hash, module.clone()).unwrap();
+                // We try to save the module into a cache file
+                cache.store(hash, module.clone()).unwrap_or_default();
```
lachlansneff: What does this do?

lib/dynasm-backend/src/lib.rs
lachlansneff: Okay.

lachlansneff:
> There's no difference in `lib/llvm-backend` from master.

Well, something is going wrong. It works in master, but not in this pr.

lib/runtime-core/src/vm.rs
losfair: Dynasm provides a quite good feature that computes offset of a struct field automatically by name. Even only for this reason, it's good to keep these fields public.

Also, I prefer to implement cold or large code paths in Rust to improve cleanness and make it less error-prone. Currently the only two functions implemented in assembly are `CONSTRUCT_STACK_AND_CALL_NATIVE` and `CALL_WASM` which both does some low-level stuff that converts calling conventions.

src/bin/wasmer.rs
losfair: The Dynasm backend does not support caching (and there's no point to cache code generated by a baseline backend).

losfair:
bors try

losfair:
bors try

losfair:
`cargo clippy --all` uses stable rust and is failing when checking `dynasm-backend`. However, clippy is not included in nightly rust and I didn't find any options to exclude crates from a clippy run.

Any ideas on fixing that?

bjfish:
@losfair I think there is a command to install on nightly , maybe `rustup component add clippy —toolchain=nightly`

src/bin/wasmer.rs
lachlansneff: Why `unwrap_or_default()` though?

syrusakbary: I did this change, so commenting here.

I used `unwrap_or_default` because the result needed to be unwrapped (otherwise a warning is generated), but I didn't want to fail if there was a problem (I wanted it to fail silently to be able to plug `dynasm` into the `wasmer` executable runtime).

losfair:
bors try

Cargo.toml
```diff
@@ -25,12 +25,11 @@ wasmer-clif-backend = { path = "lib/clif-backend" }
 wasmer-runtime = { path = "lib/runtime" }
 wasmer-runtime-core = { path = "lib/runtime-core" }
 wasmer-emscripten = { path = "lib/emscripten" }
-
-[target.'cfg(not(windows))'.dependencies]
```
bjfish: @xmclark 
Does this `not(windows)` basically mean that llvm can't be enabled via a feature here?

With llvm being a workspace member, are you able to run `cargo build --all` successfully on Windows?

Cargo.toml
```diff
@@ -25,12 +25,11 @@ wasmer-clif-backend = { path = "lib/clif-backend" }
...
 wasmer-llvm-backend = { path = "lib/llvm-backend", optional = true }
+wasmer-dynasm-backend = { path = "lib/dynasm-backend", optional = true }
 
 [workspace]
-members = ["lib/clif-backend", "lib/runtime", "lib/runtime-core", "lib/emscripten", "lib/spectests", "lib/win-exception-handler", "lib/runtime-c-api", "lib/llvm-backend"]
+members = ["lib/clif-backend", "lib/dynasm-backend", "lib/runtime", "lib/runtime-core", "lib/emscripten", "lib/spectests", "lib/win-exception-handler", "lib/runtime-c-api", "lib/llvm-backend"]
```
bjfish: I think having dynasm-backend as a member here will not compile on stable for `cargo build --all`. Maybe remove as a member for now?

Cargo.toml
losfair: Why would we want to run `cargo build --all`? It doesn't make sense to build non-binary crates individually.

bjfish: It’s useful in development to ensure changes in upstream upstream projects compile against all the other projects. For example, a change in runtime-core should be check against all other projects. It is also useful to run other commands that affect all projects in a workspace like `cargo clippy —all` or `cargo fmt —all`

xmclark: @bjfish correct. It would not work at all on windows if the attribute was still there.

lib/runtime-core/src/vm.rs
lachlansneff: The Context is exposed to the user of the runtime, so it's not ideal for these internal fields to be public. I'd be okay with making an `InternalVmCtx` struct that has these fields public, and then is part of the `Vmctx` struct.

lib/runtime-core/src/vm.rs
```diff
@@ -13,6 +13,25 @@ use std::{ffi::c_void, mem, ptr};
 #[derive(Debug)]
 #[repr(C)]
 pub struct Ctx {
+    // `internal` must be the first field of `Ctx`.
+    pub internal: InternalCtx,
```
lachlansneff: Can you make this private? You can safely cast from `Ctx` to `InternalCtx` even with this being private.

lib/runtime-core/src/vm.rs
```diff
@@ -13,6 +13,25 @@ use std::{ffi::c_void, mem, ptr};
...
+    pub data: *mut c_void,
+    pub data_finalizer: Option<extern "C" fn(data: *mut c_void)>,
+}
+
+/// The internal context of the currently running WebAssembly instance.
+///
+///
+#[derive(Debug)]
+#[repr(C)]
+pub struct InternalCtx {
```
lachlansneff: And make this `#[doc(hidden)]`.

lib/dynasm-backend/src/codegen_x64.rs
```diff
@@ -406,10 +406,8 @@ impl ProtectedCaller for X64ExecutionContext {
                 });
             }
             unsafe {
-                (
-                    (**(*_vmctx).internal.memories).base,
-                    (**(*_vmctx).internal.memories).bound,
-                )
+                let vmctx = _vmctx as *mut vm::InternalCtx;
```
lachlansneff: Just make the ctx parameter of this function `&mut vm::InternalCtx`.

losfair: Then the definition of the `ProtectedCaller` trait needs to be changed. I doubt whether this is really needed.

lachlansneff: Oh, my mistake. I thought this was part of a different function. Why does the protected caller need to figure out where the memory is?

losfair: Because `CALL_WASM` needs to setup a register containing memory information (base, bound) which is passed in as a paremeter.

lachlansneff: Okay. Let's merge this and then discuss implementation later.

PR #278, opened 2019-03-19
Add versioning to cache

### LINKED ISSUE #272: Cache files should be version-dependent
Keys used for cache files only depend on the original wasm file data.

When wasmer is updated or compiled differently, the old cache files keep being silently used forever.

Ideally, a UID generated at compile-time should be used as the BLAKE2B key or salt in order to have a cache directory seamlessly shared by multiple wasmer versions.

MarkMcCaskey:
resolves #272


MarkMcCaskey:
There are some issues that I noticed while doing this, the help text on `wasmer cache --help` seems to be missing (it's a default value which doesn't make much sense) and running `wasmer cache clean` before running anything with wasmer will cause a panic.  These should both be easy to fix

lib/runtime-core/src/cache.rs
```diff
@@ -207,3 +207,14 @@ pub trait Cache {
     fn load(&self, key: WasmHash) -> Result<Module, Self::LoadError>;
     fn store(&mut self, key: WasmHash, module: Module) -> Result<(), Self::StoreError>;
 }
+
+const WASMER_VERSION: &'static str = env!("CARGO_PKG_VERSION");
+
+/// Returns a unique ID generated from the version of Wasmer for use with cache versioning
+pub fn cache_versioned_sub_directory() -> String {
```
lachlansneff: I'm not sure if it's really necessary, but do you think it'd be possible to hash the version at compile time? `const` functions aren't advanced enough, but the build.rs would be able to.

lib/runtime-core/build.rs
```diff
@@ -0,0 +1,26 @@
...
+fn main() {
+    let mut state = blake2bp::State::new();
+    state.update(WASMER_VERSION.as_bytes());
+
+    let hasher = state.finalize();
+    let hash_string = hasher.to_hex().as_str().to_owned();
+
+    let crate_dir = env!("CARGO_MANIFEST_DIR");
+    let wasmer_version_hash_file = {
+        let mut path = PathBuf::from(&crate_dir);
```
lachlansneff: Try putting it in `env!(OUT_DIR)`. It's more idiomatic.

PR #281, opened 2019-03-19
feat(runtime) Cast and check CLI arguments to pass to the instance

Hywan:
When the Wasm binary isn't an emscripten module, this patch does 5 things:

  1. Check that the `main` function exists,
  2. Check the signature (arity and types),
  3. Cast the arguments into the correct Wasm value,
  4. Pass the arguments to the instance,
  5. Display the results.

I'm not sure about the usage of `eprintln!`. It's probably something
to update later, like the ability to pass a description message to
`ExportWrongType`.

Thoughts?

Before:

```
$ wasmer run nbody.wasm -- 7                                                                                                         
"Resolve(Signature { expected: FuncSig { params: [I32], returns: [F64] }, found: [] })"
```

After:

```sh
$ wasmer run nbody.wasm -- 7                                                                                                         
result_0 = f64 : -0.16907367446575297
```

Test file: [rust_nbody.wasm.zip](https://github.com/wasmerio/wasmer/files/2984035/rust_nbody.wasm.zip)


src/webassembly.rs
```diff
@@ -85,9 +89,108 @@ pub fn run_instance(
...
+                })
+            })
+            .and_then(|signature| {
+                let signature = signature.clone();
+                let parameter_types = signature.params();
+
+                if args.len() != parameter_types.len() {
+                    Err(CallError::Resolve(ResolveError::Signature {
+                        expected: (*signature).clone(),
+                        found: args.iter().map(|_| Type::I32).collect(),
```
Hywan: We don't know the type of the arguments yet, we need to parse them. For the error, let's assume it's `i32`.

lachlansneff:
Can you specify what type the arguments you pass in are?

syrusakbary:
> Can you specify what type the arguments you pass in are?

I think we can just try to cast it to what's expected and fail in case it's not possible. That way the way we do `wasmer run ...` becomes much simpler.
Thoughts?

Hywan:
All arguments are `&str` because they come from the CLI. The signature of the `main` function is fetched, and for each parameter, we try to parse the corresponding argument to the type of the parameter. So if `main: i32 x i32 x f64`, then for `wasmer run foo.wasm 1 2`, we will try to parse `1` and `2` as `i32` (and the returned value is an `f64`). If the parsing fails (i.e. if the “casting” fails), then an error is raised.

I think it's the best way to handle casting from `&str`.

Hywan:
As discussed in private, I will change this PR to be `wasmer call file.wasm function_name arg1 arg2`.

syrusakbary:
This changes have been merged in #1056. Closing the PR

PR #283, opened 2019-03-19
Add extend to imports

### LINKED ISSUE #258: Ability to combine two import objects
### Motivation

I should be able to combine two import objects from different origins. For example, I should be able to mix my own functions into the Emscripten import object.

### Proposed solution

See also PR #190.

MarkMcCaskey:
resolves #258 (early draft)

### DIFF
--- a/lib/runtime-core/src/import.rs
+++ b/lib/runtime-core/src/import.rs
@@ -41,7 +41,13 @@ impl IsExport for Export {
 /// }
 /// ```
 pub struct ImportObject {
-    map: Rc<RefCell<HashMap<String, Box<dyn LikeNamespace>>>>,
+    map: Rc<RefCell<HashMap<(String, String), Box<dyn IsExport>>>>,
+}
+
+pub struct ImportObjectEntry {
+    namespace: String,
+    identifier: String,
+    export: Export,
 }
 
 impl ImportObject {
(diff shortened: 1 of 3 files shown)


lib/runtime-core/src/import.rs
```diff
@@ -41,7 +41,13 @@ impl IsExport for Export {
 /// }
 /// ```
 pub struct ImportObject {
-    map: Rc<RefCell<HashMap<String, Box<dyn LikeNamespace>>>>,
+    map: Rc<RefCell<HashMap<(String, String), Box<dyn IsExport>>>>,
```
lachlansneff: Main issue with this is that we need to keep imported instances alive. I'm not sure the best way to do this, however.

MarkMcCaskey: Hmm, I'm not familiar enough with the code base yet to understand how this change affects the lifetimes of the Exports.  It seems like it should be the same, unless you mean that we were dropping entries by namespace before 🤔

lachlansneff: Yeah, the `Box<dyn LikeNamespace>` could contain an `Instance`. When getting a function pointer from an `ImportObject`, if the Instance isn't maintained in the `ImportObject`, the function pointer will dangle.

MarkMcCaskey: Hmm, it's possible to use two maps and check both to emulate a global name lookup.

Though, I'm still a bit unclear on the ownership... `ImportObject`s own `Instance`s and `Instance`s own `ImportObject`s?

syrusakbary:
Closing this in favor of #286

PR #286, opened 2019-03-23
Add extend to imports 

### LINKED ISSUE #258: Ability to combine two import objects

MarkMcCaskey:
a rewrite of #283 with a focus on simplicity

resolves #258

### DIFF
--- a/lib/runtime-core/src/import.rs
+++ b/lib/runtime-core/src/import.rs
@@ -1,5 +1,6 @@
 use crate::export::Export;
 use hashbrown::{hash_map::Entry, HashMap};
+use std::collections::VecDeque;
 use std::{
     cell::{Ref, RefCell},
     rc::Rc,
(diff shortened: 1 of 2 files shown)


lib/runtime-core/src/import.rs
```diff
@@ -97,6 +100,56 @@ impl ImportObject {
...
+        }
+    }
+}
+
+impl Extend<(String, String, Export)> for ImportObject {
+    fn extend<T: IntoIterator<Item = (String, String, Export)>>(&mut self, iter: T) {
+        let mut map = self.map.borrow_mut();
+        for (ns, id, exp) in iter.into_iter() {
+            if let Some(like_ns) = map.get_mut(&ns) {
+                like_ns.maybe_insert(&id, exp).expect(&format!(
```
syrusakbary: I don't think we should panic if the imports are overwritten. The same way extend doesn't panic on extending a `HashMap`.

MarkMcCaskey:
bors try

PR #294, opened 2019-03-26
add support for symbol maps

### LINKED ISSUE #293: Add symbol map support
`emcc` can generate symbol maps (`0:func_name\n1:func_name2...`), wasmer should have a debug option to use these in the debug logs

part of #225

MarkMcCaskey:
resolves #293

### DIFF
--- a/lib/clif-backend/src/resolver.rs
+++ b/lib/clif-backend/src/resolver.rs
@@ -374,26 +374,26 @@ fn round_up(n: usize, multiple: usize) -> usize {
 }
 
 extern "C" fn i32_print(_ctx: &mut vm::Ctx, n: i32) {
-    print!(" i32: {},", n);
+    eprint!(" i32: {},", n);
 }
 extern "C" fn i64_print(_ctx: &mut vm::Ctx, n: i64) {
-    print!(" i64: {},", n);
+    eprint!(" i64: {},", n);
 }
 extern "C" fn f32_print(_ctx: &mut vm::Ctx, n: f32) {
-    print!(" f32: {},", n);
+    eprint!(" f32: {},", n);
 }
 extern "C" fn f64_print(_ctx: &mut vm::Ctx, n: f64) {
-    print!(" f64: {},", n);
+    eprint!(" f64: {},", n);
 }
 extern "C" fn start_debug(ctx: &mut vm::Ctx, func_index: u32) {
-    if let Some(symbol_map) = &ctx.maybe_symbol_map {
+    if let Some(symbol_map) = unsafe { ctx.borrow_symbol_map() } {
         if let Some(fn_name) = symbol_map.get(&func_index) {
-            print!("func ({} ({})), args: [", fn_name, func_index);
+            eprint!("func ({} ({})), args: [", fn_name, func_index);
             return;
         }
     }
-    print!("func ({}), args: [", func_index);
+    eprint!("func ({}), args: [", func_index);
 }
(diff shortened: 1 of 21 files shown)


lib/clif-backend/src/resolver.rs
```diff
@@ -385,7 +385,13 @@ extern "C" fn f32_print(_ctx: &mut vm::Ctx, n: f32) {
 extern "C" fn f64_print(_ctx: &mut vm::Ctx, n: f64) {
     print!(" f64: {},", n);
 }
-extern "C" fn start_debug(_ctx: &mut vm::Ctx, func_index: u32) {
+extern "C" fn start_debug(ctx: &mut vm::Ctx, func_index: u32) {
+    if let Some(symbol_map) = &ctx.maybe_symbol_map {
+        if let Some(fn_name) = symbol_map.get(&func_index) {
+            print!("func ({} ({})), args: [", fn_name, func_index);
```
bjfish: Consider using `eprint` for debug statements.

lib/runtime-core/src/vm.rs
```diff
@@ -24,6 +25,8 @@ pub struct Ctx {
 
     pub data: *mut c_void,
     pub data_finalizer: Option<extern "C" fn(data: *mut c_void)>,
+
+    pub maybe_symbol_map: Option<HashMap<u32, String>>,
```
bjfish: I think `symbol_map` would also be an okay name for this, I do like `maybe_symbol_map` as a variable name but it looks a little weird to me as a field name since I don't think we do this anywhere else.

lib/runtime-core/src/module.rs
```diff
@@ -100,8 +100,12 @@ impl Module {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn instantiate(&self, import_object: &ImportObject) -> error::Result<Instance> {
-        Instance::new(Arc::clone(&self.inner), import_object)
```
bjfish: We should probably keep this public signature around for backwards compatability: `pub fn instantiate(&self, import_object: &ImportObject) -> error::Result<Instance> {`
I'm guessing we may have more configuration later on, so maybe we could have a `Config` struct to hold the `symbol_map`:
```
  pub fn instantiate_with_config(
        &self,
        import_object: &ImportObject,
        config: &Config,
    )
```

lib/runtime-core/src/instance.rs
```diff
@@ -63,7 +68,12 @@ impl Instance {
         // Initialize the vm::Ctx in-place after the backing
         // has been boxed.
         unsafe {
-            *inner.vmctx = vm::Ctx::new(&mut inner.backing, &mut inner.import_backing, &module)
+            *inner.vmctx = vm::Ctx::new(
+                &mut inner.backing,
+                &mut inner.import_backing,
+                &module,
+                maybe_symbol_map,
+            )
```
bjfish: @lachlansneff `Ctx` is per instance right?  In that case, then this data should be stored in the module level?

MarkMcCaskey:
> Can you post an image of some typical output with the symbols?

Sure!

Screen Shot 2019-03-27 at 15 35 25

MarkMcCaskey:
Also, as a heads up, this will be a breaking change to the wasmer cache.  The next time we have a release this will be fixed because of #272.

MarkMcCaskey:
bors retry

MarkMcCaskey:
bors retry

PR #295, opened 2019-03-27
generate null funcs

xmclark:
This PR automates adding the `nullfunc` imports. A new macro is added to help working with namespaces. All functions in the env namespace that start with `nullFunc_` will be added to the import object. This is easy because *all* null funcs have the same expected signature.


xmclark:
bors try

lib/runtime-core/src/macros.rs
```diff
@@ -75,3 +75,15 @@ macro_rules! __imports_internal {
         $ns
     };
 }
+
+#[macro_export]
+#[doc(hidden)]
+macro_rules! ecapseman {
```
bjfish: What does `ecapseman` mean?

lib/emscripten/src/lib.rs
```diff
@@ -770,7 +730,15 @@ pub fn generate_emscripten_env(globals: &mut EmscriptenGlobals) -> ImportObject
         "asm2wasm" => {
             "f64-rem" => func!(crate::math::f64_rem),
         },
-    }
+    };
+
+    import_object
+}
+
+pub fn nullfunc_x(ctx: &mut Ctx, _x: u32) {
```
syrusakbary: Should be just `nullfunc`? (without an x)

```
pub fn nullfunc_x(ctx: &mut Ctx) {
...
}
```

xmclark: sure, sound good

PR #298, opened 2019-03-27
Chimera: Tiering runtime

lachlansneff:
Chimera is wasmer's tiering runtime that combines all three backends to prioritize both startup latency and runtime performance.

This is very much a WIP.


Hywan:
\o/

bjfish:
Closing this as I believe this tiering PR has become stale now with some changes that have happened since: common parsing pipeline and the addition of OSR to singlepass. I'll create links to this as a reference for the future implementation.

Issue #297, opened 2019-03-27
Add WASI Interface to Wasmer


syrusakbary:
WASI is a WebAssembly System Interface.

Here's a detailed blogpost from Mozilla about it: https://hacks.mozilla.org/2019/03/standardizing-wasi-a-webassembly-system-interface/

Wasmer should support WASI, in a similar way we support emscripten.
So binaries/libraries can be executed with wasmer easily.

## Related issues

Rust is also trying to target `wasi` from rust libraries: https://github.com/rust-lang/rust/pull/59464

## Related docs

WASI API: https://github.com/CraneStation/wasmtime-wasi/blob/wasi/docs/WASI-api.md

syrusakbary:
We just merged #299 which adds initial (and experimental) support for WASI 🎉

We will keep iterating on the WASI integration, so keeping this issue open until is feature complete

jedisct1:
Wasmer's support for WASI is pretty decent already!

You should ask to be listed here https://wasi.dev

syrusakbary:
Thanks for the ping!

We will need to submit PRs to the following repos:
* https://github.com/cranestation/wasi
* https://github.com/CraneStation/wasi.dev

@Hywan do you want to handle this? :)

syrusakbary:
I've created the issue in the wasi.dev repo adding Wasmer: https://github.com/CraneStation/wasi.dev/pull/2

Closing the issue here since Wasmer WASI support is stable and mature now.

Thanks everyone for your support!

PR #299, opened 2019-03-28
Support for the WASI ABI.

lachlansneff:
part of #297


lib/runtime-core/src/instance.rs
```diff
@@ -63,7 +63,16 @@ impl Instance {
         // Initialize the vm::Ctx in-place after the backing
         // has been boxed.
         unsafe {
-            *inner.vmctx = vm::Ctx::new(&mut inner.backing, &mut inner.import_backing, &module)
+            *inner.vmctx = match imports.call_state_creator() {
+                Some((data, dtor)) => vm::Ctx::new_with_data(
```
bjfish: `dtor`? maybe `destructor` or `finalizer`?

lachlansneff: Yeah, either of those is fine. `dtor` was just short and it's common in the c++ world.

lib/wasi/src/syscalls/mod.rs
```diff
@@ -0,0 +1,218 @@
+use crate::state::WasiState;
+use wasmer_runtime_core::{memory::Memory, vm::Ctx};
+
+#[allow(clippy::mut_from_ref)]
+fn get_wasi_state(ctx: &Ctx) -> &mut WasiState {
+    unsafe { &mut *(ctx.data as *mut WasiState) }
```
bjfish: One idea: we could differentiate between, ModuleState, InstanceState, FunctionState (if we add that).

lib/wasi/src/state.rs
```diff
@@ -0,0 +1,271 @@
...
+    pub inode_counter: Cell<u64>,
+}
+
+impl WasiFs {
+    pub fn new() -> Result<Self, String> {
+        zbox_init_env();
+        Ok(Self {
+            repo: RepoOpener::new()
+                .create(true)
+                .open("mem://foo", "")
```
xmclark: let's call it something else

lib/runtime-core/src/macros.rs
```diff
@@ -57,6 +64,21 @@ macro_rules! imports {
...
+        };
+
+        let mut import_object = ImportObject::new_with_data($state_gen);
+
+        $({
+            let ns = $crate::__imports_internal!($ns);
+
+            import_object.register($ns_name, ns);
+        })*
+
```
syrusakbary: What this does? What is `__imports_internal` ?

lib/wasi/src/syscalls/unix/macos.rs
```diff
@@ -0,0 +1,12 @@
+use crate::syscalls::types::*;
+use std::cell::Cell;
+
+pub fn platform_fd_pread(
```
xmclark: Mark you know what to do

lib/wasi/src/syscalls/unix/linux.rs
```diff
@@ -0,0 +1,36 @@
...
+                fd as i32,
+                iovec.as_mut_ptr(),
+                iovs_len as i32,
+                offset as i64,
+            ),
+            iovec,
+        )
+    };
+    nread.set(result as u32);
+    /*for (i, arr_cell) in iovs.iter().enumerate() {
```
xmclark: 🗑

PR #300, opened 2019-03-28
hook up wasi to wasmer

MarkMcCaskey:


src/bin/wasmer.rs
```diff
@@ -215,6 +218,19 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
         )
     };
 
+    #[cfg(features = "wasi")]
+    let (_abi, import_object) = if wasmer_wasi::is_wasi_module(&module) {
+        (
+            InstanceABI::WASI,
+            wasmer_emscripten::generate_import_object(),
```
lachlansneff: emscripten?

src/bin/wasmer.rs
```diff
@@ -215,6 +218,16 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
         )
     };
 
+    #[cfg(features = "wasi")]
+    let (_abi, import_object) = if wasmer_wasi::is_wasi_module(&module) {
+        (InstanceABI::WASI, wasmer_wasi::generate_import_object())
```
lachlansneff: We need to pass in the args and envs. They are both `Vec<Vec<u8>>`.

MarkMcCaskey: Cool, was just going off the function signature at the time

PR #268, opened 2019-03-12
add experimental virtual filesystem

xmclark:
This PR introduces some changes to the runtime and adds a new feature: virtual file system.

In the future, we may want to bundle data with our wasm apps. That data can be made available to apps in a virtual filesystem by the wasmer runtime. To do this:

- wasmer must understand custom section data
- wasmer must re-wire emscripten calls to access a virtual file system
- these features are experimental and should be guarded by a feature flag (`--features vfs`)

~This PR adds support for the `read` syscall. ~
Just kidding, this PR does a lot more. Enough syscalls were re-implemented that a code reorg was warranted. 

Additions below ⬇️ 

Several new modules live under and around the `syscalls` module. These features are only supported on linux, but can be expanded to windows.

There are effectively two implementations: one for VFS and one for the host FS. In the future, we will have more tight control over access to the host, but this will do for now. The `select` function was also extracted into a new module as the complexity increased significantly. 

`select`should be refactored in the future to use a runtime like tokio, but that will need to wait.

The introduction of the libsodium dependency introduced some stumbling blocks, so I forked the zbox project and added a new build step. See it [here](https://github.com/wasmerio/zbox/blob/bundle-libsodium/build.rs). This will build libsodium at `cargo build` time and will automatically configure the environment. This will help with adoption and it is something we may be able to give back to zbox.

Additional changes:

After review, this PR is just a bit too large. We will implement the emscripten changes in another PR. This will help accelerate the work on WASI.

PR #302, opened 2019-03-28
add emscripten virtual filesystem

xmclark:
This PR is an extension to #268. It brings the virtual file system to the emscripten module. Running with the feature flag `--vfs` will instruct wasmer to look in the in-memory virtual filesystem when doing file system operations.

The emscripten syscalls module is reorganized to sepearate the host file system calls from the virtual file system calls. 

A custom implementation of select is made and move into a separate file.


xmclark:
This isn't going anywhere for awhile, so closing this PR.

PR #303, opened 2019-03-28
Add all wasi types

lachlansneff:
This adds all wasi types.


lachlansneff:
bors try

bjfish:
Maybe somebody could publish a set of common types as a library.

lachlansneff:
I'd like to write some safer wrappers around it as well.

PR #309, opened 2019-03-29
fix(llvm-backend) Remove an `example.rs` file.

Hywan:
I think this file is useless?


xmclark:
bors try

lachlansneff:
Yep, this file shouldn't be there!
bors r+

Issue #319, opened 2019-04-02
Accessing global arrays in the guest from host, when embedded?


hrydgard:
### Summary

When embedded in a host Rust application, it does not seem to be possible to access the contents of guest global arrays from the host. And it's not even possible to list the exports without linking to wasmer_runtime_core.

### Additional details

For example, if I add the following line to the example in https://github.com/wasmerio/wasmer-rust-example : 

```
#[no_mangle]
pub static mut SCRATCH_I : [u8; 1024] = [0; 1024];
```

And then loop through instance.exports() in main.rs:

```
    // (at the top of file: ) use wasmer_runtime_core::{export};
    for (name, export) in instance.exports() {
        match export {
            export::Export::Function{ func, ctx, signature } => {println!("  func: {}", name);},
            export::Export::Global( global ) => {println!("  global: {}", name);},
            _ => {},
        }
    }
```

I do get the following:
```
  func: hello_string_from_rust
  global: __data_end
  func: hello_wasm
  global: __heap_base
  global: SCRATCH_I
```
So SCRATCH_I is there.

But how can I access the memory? [Global](https://docs.rs/wasmer-runtime-core/0.2.1/wasmer_runtime_core/global/struct.Global.html) does not seem to have a public way to get the address - if I had that, I could stash it away and then go directly to memory.

Or am I missing something?

bjfish:
@lachlansneff I think we are missing `Export` from the public runtime API, can you confirm?

@hrydgard I modified your example to inspect the global value:
```
Export::Global(global) => {
   println!("  global: {}, {:?}", name, global);
}
```
which resulting in the following:
```
global: SCRATCH_I, Global { desc: GlobalDescriptor { mutable: false, ty: I32 }, value: I32(1053432) }
```
I suspect this value `1053432` to be an index into the memory, maybe some space setup for statics like this. So, I setup the array to contain some values to confirm this suspicion:
```
#[no_mangle]
pub static mut SCRATCH_I : [u8; 5] = [1, 2, 3, 4, 5];
```
and then print this part of the memory:
```
let i = 1053432;
let memory = instance.context().memory(0);
for b in memory.view::<u8>()[i as usize..(i + 5) as usize].iter() {
    println!("{:?}", b);
}
```
Results in:
```
Cell { value: 1 }
Cell { value: 2 }
Cell { value: 3 }
Cell { value: 4 }
Cell { value: 5 }
```
[...]

PR #320, opened 2019-04-03
Add Export and GlobalDescriptor to Runtime API

bjfish:
This was mentioned in the issue: https://github.com/wasmerio/wasmer/issues/319

It would be nice to be able to iterate exports without requiring `runtime-core`.


bjfish:
bors try

PR #322, opened 2019-04-03
compile invoke functions into module

### LINKED ISSUE #240: Dynamic invoke/dynCall
Currently wasmer requires a bespoke definition for each invoke/dynCall. This is somewhat non-ideal since the number of `invoke_[vid]+` functions is countably infinite. I took a whack at a DynFunc version of invoke, but `fn import_functions` returns `ImportedFunc`, which won't comfortably hold a DynFunc. It'd be nice to see a clean way to do this!

xmclark:
This PR fixes #240 by compiling invoke functions into a module at wasm-compile-time.

The solution generates some wat/wast and then uses wabt to convert it to wasm. The wasm is compiled into a module, and then instantiated. The exports are iterated and inserted into an 'env' namespace. The namespace is put into the imports object which is used to instantiate the main wasm module. 

Generating wat is a little slower, but the code was easier for me to reason about.


lib/emscripten/src/invoke.rs
```diff
@@ -0,0 +1,217 @@
...
+
+struct Sig {
+    pub args: Vec<InvokeArgType>,
+    pub rets: Vec<InvokeArgType>,
+}
+
+pub fn create_invoke_module(invoke_function_names: &Vec<String>) -> Module {
+    let wat = create_wat(invoke_function_names);
+    let wasm_binary = wabt::wat2wasm(wat.as_bytes()).unwrap();
+    let compiler = wasmer_clif_backend::CraneliftCompiler::new();
```
MarkMcCaskey: What does this mean for the multiple backends? 🤔

lib/emscripten/src/invoke.rs
```diff
@@ -0,0 +1,217 @@
...
+        Some(ret) => format!(" (result {})", ret.to_type_str()),
+        None => "".to_string(),
+    };
+    format!(
+        "(type $emscripten_{} (func{}{}))",
+        func_name, params_string, result_string
+    )
+}
+
+fn sig_to_func_wat(func_name: &str, sig: &Sig) -> String {
```
MarkMcCaskey: doc comments on some of the larger functions would be nice; brief explanation of what the `func wat` will look like would be good I think

syrusakbary:
I think rather than doing `signature checking -> wast -> convert to wasm -> compile to a Module using Cranelift -> get it's functions`, it might be more optimal to do `signature checking -> generate the functions using Cranelift directly`. Thoughts @xmclark ?

syrusakbary:
Here's the example on how trampolines are implemented in Cranelift: https://github.com/wasmerio/wasmer/blob/master/lib/clif-backend/src/trampoline.rs

syrusakbary:
If we want to merge this, I think we will need to research two things:
1. What's the cost of generating this functions at runtime in a real-world scenario (meaning, what's the cost end to end on generating the functions)
2. Add all the different emscripten signatures (since it's almost for free)

bjfish:
I think I also would prefer the approach `signature checking -> generate the functions using Cranelift directly` mentioned here: https://github.com/wasmerio/wasmer/pull/322#issuecomment-479607675
This `wat` approach seems a little heavy to me for dynamically generated code.

PR #323, opened 2019-04-03
add validate subcommand

MarkMcCaskey:


src/bin/wasmer.rs
```diff
@@ -240,6 +251,49 @@ fn run(options: Run) {
...
+    })?;
+
+    if !utils::is_wasm_binary(&wasm_binary) {
+        return Err(format!(
+            "Cannot recognize \"{}\" as a WASM binary",
+            wasm_path_as_str,
+        ));
+    }
+
+    wabt::Module::read_binary(wasm_binary, &Default::default())
```
bjfish: Runtime has a validate method:
https://github.com/wasmerio/wasmer/blob/bf823d2826aa2d17288103ba78bfc05a65e1f18c/lib/runtime/src/lib.rs#L87
It may be lighterweight solution than wabt.

MarkMcCaskey: Hmm, I think we'd want better error messages than `yes` or `no`, I'll look in to the functions that that `validate` fn calls

MarkMcCaskey: Thanks for pointing this out!

PR #338, opened 2019-04-09
Actually catch traps/panics/etc when using a typed func.

lachlansneff:
Some niche code in some of the other libraries will need to be added.

### DIFF
--- a/lib/clif-backend/src/signal/mod.rs
+++ b/lib/clif-backend/src/signal/mod.rs
@@ -2,12 +2,13 @@ use crate::relocation::{TrapData, TrapSink};
 use crate::trampoline::Trampolines;
 use hashbrown::HashSet;
 use libc::c_void;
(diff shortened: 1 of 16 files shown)


lachlansneff:
Okay, let's give this a try.
bors try

lachlansneff:
once more

bors try

lachlansneff:
trying to fix windows build
bors try

lachlansneff:
bors try

lachlansneff:
pls
bors try

bjfish:
It would be nice if there was more comments/documentation. For example, the `get_wasm_trampoline` could describe what is trampoline, where are they used, why are they're need. Or each of the traits/structs could describe their purpose.

lachlansneff:
@bjfish Good idea, I'll do that before merging.

lachlansneff:
*prays to the git gods*
bors r+

PR #342, opened 2019-04-10
feat(runtime-core) Make `UnsharedMemory` thread-safe

Hywan:
Use `Arc` instead of `Rc`, basically :-]. I <3 Rust.


Hywan:
No longer required.

PR #343, opened 2019-04-11
add preopened fd and fix/improve fs syscalls; fix wasi memory access

### LINKED ISSUE #356: Rust program can not open a file
Hello,

It is probably highly related to #297 currently but I prefer posting this issue to track the progression of the "fix".

I am currently testing wasmer, it is an awesome project, it work great with Rust, especially when compiling for the `wasm32-unknown-wasi` target.

However I tried a little program just to display the content of a file and hit a little problem: It is not possible to open a file. Here is the minimum program to reproduce that:

```rust
use std::{fs, io};

fn main() -> io::Result<()> {
    let file = fs::File::open("README.md")?;
    Ok(())
}
```

The error returned [is thrown by the Rust std library](https://github.com/rust-lang/rust/pull/59727/files#diff-4407938b27c53216e4b5afb5fc21ef36R663):

```
Custom { kind: Other, error: StringError("failed to find a preopened file descriptor through which \"xxx\" could be opened") }
```

Thank you for helping :)

MarkMcCaskey:
resolves #356


src/bin/wasmer.rs
```diff
@@ -66,9 +66,13 @@ struct Run {
     #[structopt(name = "--", raw(multiple = "true"))]
     args: Vec<String>,
 
-    /// Emscripten symbol map
-    #[structopt(long = "em-symbol-map", parse(from_os_str))]
+    /// Emscripten symbol ma>p
```
Kerollmops: Whoops :)

lib/wasi/src/state.rs
```diff
@@ -272,6 +409,19 @@ impl WasiFs {
         );
         Ok(idx)
     }
+
```
MarkMcCaskey: will rm

lib/wasi/src/syscalls/mod.rs
```diff
@@ -673,11 +685,12 @@ pub fn fd_read(
             let bytes = iov_inner.buf.deref(memory, 0, iov_inner.buf_len)?;
             let mut raw_bytes: &mut [u8] =
                 unsafe { &mut *(bytes as *const [_] as *mut [_] as *mut [u8]) };
-            bytes_read += reader.read(raw_bytes).map_err(|_| __WASI_EIO)? as u32;
+            bytes_read += reader.read(raw_bytes).map_err(|e| __WASI_EIO)? as u32;
         }
         Ok(bytes_read)
     }
 
+    debug!("think-fish");
```
lachlansneff: Don't forget to remove statements like this!

lib/wasi/src/syscalls/mod.rs
```diff
@@ -809,6 +825,8 @@ pub fn fd_seek(
         return __WASI_EACCES;
     }
 
+    debug!("Hmml");
```
lachlansneff: ditto

lib/wasi/src/ptr.rs
```diff
@@ -36,10 +36,11 @@ impl<T: Copy + ValueType> WasmPtr<T, Item> {
             return Err(__WASI_EFAULT);
         }
         unsafe {
-            let cell_ptr = memory
-                .view::<T>()
-                .get_unchecked((self.offset() as usize) / mem::size_of::<T>())
-                as *const _;
+            let aligner = |ptr: usize, align: usize| ptr & !(align - 1);
```
lachlansneff: Can you add a comment here saying what this does?

PR #367, opened 2019-04-19
Add caching support to llvm backend

lachlansneff:
This changes the `CacheGen` trait a little, since it was weird.


lachlansneff:
bors try

PR #370, opened 2019-04-20
Fix WASI by using u32 everywhere instead of u16

syrusakbary:
Long story short: we were seeing some undeterministic behavior when running wasmer with WASI on Linux.

It turned out we are doing some bad conversions from u16 to the inner type in the trampolines (or func type?), and this only showed up in certain systems.

This PR is a quick fix to get things running quickly, but we will proceed with a proper fix soon and revert this PR.

Note: the discovery of the bug was done by @bjfish, and this PR is just a very simple way to address it.

cc @lachlansneff

### DIFF
--- a/lib/wasi/src/syscalls/types.rs
+++ b/lib/wasi/src/syscalls/types.rs
@@ -42,84 +42,84 @@ pub struct __wasi_dirent_t {
     pub d_type: __wasi_filetype_t,
 }
 
-pub type __wasi_errno_t = u16;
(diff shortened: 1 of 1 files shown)


lachlansneff:
I don't think this is valid to do. It may work in this case, but the error type is defined to be a u16.

MarkMcCaskey:
In the shorter term, this will likely cause bad memory accesses with structs that use types that are `u16`s

Issue #361, opened 2019-04-16
WASI proc_exit throws exception instead of returning error code


kanaka:
### Describe the bug

The WASI `proc_exit` function causes an exception to be raised and does not set the requested error code.

### Steps to reproduce

The following wat code:

```
(module
  (import "wasi_unstable" "proc_exit"      (func $proc_exit (param i32)))

  (memory 10)
  (export "memory" (memory 0))

  (func $main
    (call $proc_exit (i32.const 7))
  )

  (start $main)
)
```

Results in the following behavior with wasmer 0.3.0:

```
$ ~/.wasmer/bin/wasmer run wasitests/exit.wasm
"Can\'t instantiate module: CallError(RuntimeError: Instance exited)"
$ echo $?
1
```

### Expected behavior

There should be no exception and the result error code should be set to the argument value. In the example above the exit code value should be 7.

### Actual behavior

Exception with return value of 1.

kanaka:
I checked again and the error message is different but the behavior is still the same:

```
$ wasmer run wasitests/exit.wasm 
"RuntimeError: WebAssembly trap occured during runtime: unknown"
$ echo $?
1
```

kanaka:
I checked wasmer 0.4.0 and this appears to be resolved. Thanks!

PR #373, opened 2019-04-20
Call process::exit in WASI proc_exit syscall.

kanaka:
Currently calling `proc_exit` causes an exception and does not set the exit code. This updates the syscall to use `std::process::exit` instead of calling `Err`.

Fixes: https://github.com/wasmerio/wasmer/issues/361

### DIFF
--- a/lib/wasi/src/syscalls/mod.rs
+++ b/lib/wasi/src/syscalls/mod.rs
@@ -13,6 +13,7 @@ use crate::{
 use rand::{thread_rng, Rng};
 use std::cell::Cell;
 use std::io::{self, Read, Seek, Write};
+use std::process;
 use wasmer_runtime_core::{debug, memory::Memory, vm::Ctx};
 
 #[cfg(any(target_os = "linux", target_os = "macos"))]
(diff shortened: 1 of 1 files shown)


lachlansneff:
Wasmer is designed to run multiple tenants and as a result, the WASI integration is also designed to be multi-tenant. Calling the exit syscall will just kill the process, even if there are other instances running.

There's some functionality missing in wasmer that needs to be implemented that will allow the runtime to retrieve the exit code from the thrown exception, which will fix this issue.

kanaka:
@lachlansneff ah, so I assume that the if there is only a single tenant then when that tenant exits, it's exit code (if there is one) becomes exit code of the wasmer process. If you have multiple tenants, I assume you would delay exit until all tenants have exited? Out of curiosity do you just pick the exit code of the last tenant "thread" to be the wasmer exit code?

I'll close this PR and just wait (impatiently) for the original issue to be addressed the right way for wasmer's architecture :-)

lachlansneff:
This issue will be fixed today, promise!

The wasmer cli program is one-tenant only right now, so it'll just return the exit code of that tenant.

PR #375, opened 2019-04-21
Introduce the MemoryViewMut type

Kerollmops:
Introduce the `MemoryViewMut` type which will allow us to manage the instance's memory safely and with ease because `DerefMut<Target=[T]>` is implemented on it and allow us to use [`copy_from_slice`](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice) to modify it.

I did not worked on the documentation yet, I am waiting for your approval about the correctness of this pull request first.

### DIFF
--- a/lib/runtime-core/src/memory/mod.rs
+++ b/lib/runtime-core/src/memory/mod.rs
@@ -17,7 +17,7 @@ use std::{
 pub use self::atomic::Atomic;
 pub use self::dynamic::DynamicMemory;
 pub use self::static_::{SharedStaticMemory, StaticMemory};
-pub use self::view::{Atomically, MemoryView};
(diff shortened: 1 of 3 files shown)


lachlansneff:
Hey, thanks for the PR!

Unfortunately, this is unsound. The reason View derefs to `[Cell<T>]` rather than just `[T]` is because that would create a situation where a immutably borrowed slice could be modified.

`[Cell<T>]` is mutable, you just have to explicitly use a for-loop to set each byte in it. (don't worry, it optimizes to a memcpy, I've checked)

Kerollmops:
Ho yeah I understand, even if you can prove to the language that you get a unique access to the slice, someone else could own the same slice in the form of an `[Cell<T>]` and modify it.

PR #376, opened 2019-04-21
Make the slice::from_raw_parts call safe

Kerollmops:
Casting a slice into another type (i.e. from `T` to `Atomic<T>`) can be made safe for Rust by using the [`#[repr(transparent)]`](https://github.com/rust-lang/rfcs/blob/master/text/1758-repr-transparent.md) on the `Atomic<T>` type.

This is now safe:
https://github.com/wasmerio/wasmer/blob/6478500b8fc52d4ecf118b376f4fe1182305338b/lib/runtime-core/src/memory/view.rs#L51


lachlansneff:
Awesome! This is great! Actually, while you're at it, could you update the `Atomic<T>` type to use the recently stabilized integer atomics instead of emulating them with AtomicUsize?

Kerollmops:
Do you want me to replace the inner `UnsafeCell<Wrapping<T>>` of the `Atomic` type and replace it with an std `AtomicX` type? It will make the `Atomic<X>` become `Atomic<AtomicX>` because each std `AtomicX` type is unique.

Or would you prefer me to create an `AtomicX` each time an action is performed? This would be similar to what is currently happening.

bjfish:
Hi @Kerollmops , I'm going to try to merge the original PR and have created a separate issue to track the refactor from AtomicUsize to integer atomics: https://github.com/wasmerio/wasmer/issues/410
Thanks.

PR #381, opened 2019-04-22
Hook up error propagation

lachlansneff:
You can now retrieve the error from a panicking or error-returning imported function.

Example:

```rust
#[derive(Debug)]
struct ExitCode {
    code: i32,
}

fn do_panic(_ctx: &mut Ctx) -> Result<i32, ExitCode> {
    Err(ExitCode { code: 42 })
}

// This wasm functions calls `do_panic`.
let foo: Func<(), i32> = instance.func(...)?;

let result = foo.call();

println!("result: {:?}", result);

if let Err(RuntimeError::Error { data }) = result {
    if let Ok(exit_code) = data.downcast::<ExitCode>() {
        println!("exit code: {:?}", exit_code);
    }
}
```

outputs:

```
result: Err(unknown error)
exit code: ExitCode { code: 42 }
```

### DIFF
--- a/lib/clif-backend/src/signal/unix.rs
+++ b/lib/clif-backend/src/signal/unix.rs
@@ -10,7 +10,7 @@
 //! unless you have memory unsafety elsewhere in your code.
 //!
 use crate::relocation::{TrapCode, TrapData};
-use crate::signal::HandlerData;
+use crate::signal::{HandlerData, RunErr};
 use libc::{c_int, c_void, siginfo_t};
 use nix::sys::signal::{
     sigaction, SaFlags, SigAction, SigHandler, SigSet, Signal, SIGBUS, SIGFPE, SIGILL, SIGSEGV,
@@ -18,7 +18,7 @@ use nix::sys::signal::{
 use std::cell::{Cell, UnsafeCell};
 use std::ptr;
 use std::sync::Once;
-use wasmer_runtime_core::error::{RuntimeError, RuntimeResult};
+use wasmer_runtime_core::typed_func::WasmTrapInfo;
 
 extern "C" fn signal_trap_handler(
     signum: ::nix::libc::c_int,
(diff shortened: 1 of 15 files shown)


lachlansneff:
@kanaka This pr will solve #373 once it's hooked up to the wasmer cli.

lib/clif-backend/src/signal/unix.rs
```diff
@@ -76,7 +76,7 @@ pub fn call_protected<T>(handler_data: &HandlerData, f: impl FnOnce() -> T) -> R
             *jmp_buf = prev_jmp_buf;
 
             if let Some(data) = super::TRAP_EARLY_DATA.with(|cell| cell.replace(None)) {
-                Err(RuntimeError::Error { data })
+                Err(RunErr::Error(data))
```
syrusakbary: I kind of like more `RuntimeError`? But I'm good with either. Maybe @MarkMcCaskey  or @xmclark can provide more input here.

lib/runtime/examples/call.rs
```diff
@@ -63,5 +70,11 @@ fn main() -> Result<(), error::Error> {
 
     println!("result: {:?}", result);
 
+    if let Err(RuntimeError::Error { data }) = result {
+        if let Ok(exit_code) = data.downcast::<ExitCode>() {
```
syrusakbary: Love this.

lib/clif-backend/src/signal/unix.rs
lachlansneff: It's not the same enum actually. `RuntimeError` is from the `runtime_core`, but `RunErr` is a local enum that's just used in the interaction between `call_protected` and the invoke function.

xmclark: `CallProtectedError`

lachlansneff: Ok, will rename.

lachlansneff:
bors try

PR #382, opened 2019-04-22
reduce suggested backends to the ones compiled into wasmer

### LINKED ISSUE #380: Backends that haven't been compiled in shouldn't be listed
`Backend::variants()` currently returns a static set of backends.

Which is quite confusing. The `wasmer run` help command always prints:

```text
OPTIONS:
        --backend <backend>                   [default: cranelift]  [possible values: singlepass, cranelift, llvm]
```

Even if neither `singlepass` nor `llvm` have been compiled in.

MarkMcCaskey:
resolves #380

### DIFF
--- a/src/bin/wasmer.rs
+++ b/src/bin/wasmer.rs
@@ -105,22 +105,13 @@ enum Backend {
 
 impl Backend {
     pub fn variants() -> &'static [&'static str] {
-        #[cfg(all(feature = "backend:singlepass", feature = "backend:llvm"))]
-        {
-            &["cranelift", "singlepass", "llvm"]
-        }
-        #[cfg(all(not(feature = "backend:singlepass"), feature = "backend:llvm"))]
-        {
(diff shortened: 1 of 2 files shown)


src/bin/wasmer.rs
```diff
@@ -105,7 +105,22 @@ enum Backend {
 
 impl Backend {
     pub fn variants() -> &'static [&'static str] {
-        &["singlepass", "cranelift", "llvm"]
+        #[cfg(all(feature = "backend:singlepass", feature = "backend:llvm"))]
```
lachlansneff: What if this was 
```rust
&[
    #[cfg(feature = "backend:singlepass")]
    "singlepass",
    #[cfg(feature = "backend:cranelift")]
    "cranelift",
    ...
]
```

instead?

MarkMcCaskey: I assumed that that wouldn't work due to the commas, but I can try it

Issue #362, opened 2019-04-16
Lack of _start export causes exception


kanaka:
### Describe the bug

wasmer 0.3.0 appears to assume that a wasm module exports a "_start" function. This should either be optional or dropped since (start ...) is already supported and is called before the exported "_start" routine.

### Steps to reproduce

The following wat file:

```
(module
  (memory 256)
  (export "memory" (memory 0))

  (import "wasi_unstable" "fd_write" (func $fd_write (param i32 i32 i32 i32) (result i32)))

  ;; __wasi_ciovec_t struct
  (data (i32.const 0) "\08\00\00\00") ;; buf: pointer to hello string
  (data (i32.const 4) "\0c\00\00\00") ;; buf_len: 12 characters
  ;; string
  (data (i32.const 8) "hello world\n")

  (func $main
    (drop (call $fd_write (i32.const 1) (i32.const 0) (i32.const 1) (i32.const 64)))
  )

  (start $main)
  ;;(export "_start" $main)
)
```

Will result in the following:

```
$ ~/.wasmer/bin/wasmer run wasitests/start.wasm
hello world
"ResolveError: ExportNotFound { name: \"_start\" }"
```

Note that the function specified by (start) is being called and then the exception is thrown. If both (start) and exported "_start" are used (uncomment export above) then the main function is executed twice as follows:

```
$ ~/.wasmer/bin/wasmer run wasitests/start.wasm
hello world
hello world
```

### Expected behavior

wasmer should not require "_start" to be exported since this functionality is already supported by the standard (start) mechanism.

syrusakbary:
Good catch! Thanks for the report

kanaka:
Related issue in lucet: https://github.com/fastly/lucet/issues/129#issuecomment-483915763 Based on discussion there seems like "_start" or something like it is needed (although I've suggested a different less confusing name) but the problem is that wasmer and lucet and throwing an exception when "_start" isn't exported (or is swallowing the exception).

bjfish:
This is also being discussed here: https://github.com/WebAssembly/WASI/issues/19

syrusakbary:
Closing the ticket here in favor of: https://github.com/WebAssembly/WASI/issues/19
:)

PR #383, opened 2019-04-22
Hook up wasi exit code to wasmer cli

lachlansneff:
Test by running:

```
> target/release/wasmer run examples/exit.wat
> echo $?
```

should echo "7".

Fixes (at least the exit code part) of #361.


lachlansneff:
bors try

src/bin/wasmer.rs
```diff
@@ -295,60 +297,81 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
...
+            );
 
-    let mut instance = module
-        .instantiate(&import_object)
-        .map_err(|e| format!("Can't instantiate module: {:?}", e))?;
+            let instance = module
+                .instantiate(&import_object)
+                .map_err(|e| format!("Can't instantiate module: {:?}", e))?;
 
-    webassembly::run_instance(
```
bjfish: Maybe this method could be deprecated if it will no longer be called.

lachlansneff: Yeah, this is a good idea.

src/bin/wasmer.rs
```diff
@@ -295,60 +297,81 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
...
-        &module,
-        &mut instance,
-        abi,
-        if let Some(cn) = &options.command_name {
-            cn
+            let start: Func<(), ()> = instance.func("_start").map_err(|e| format!("{:?}", e))?;
```
bjfish: There doesn't seem to be agreement on `_start` function yet (see linked wasi issue):
https://github.com/wasmerio/wasmer/issues/362

This is a little out of scope of this PR but maybe we could only call `_start` if it exists.

src/bin/wasmer.rs
```diff
@@ -295,60 +297,81 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
...
+                .args
+                .iter()
+                .map(|arg| arg.as_str())
+                .map(|x| Value::I32(x.parse().unwrap()))
+                .collect();
+            instance
+                .dyn_func("main")
+                .map_err(|e| format!("{:?}", e))?
+                .call(&args)
+                .map_err(|e| format!("{:?}", e))?;
```
bjfish: Also sort of unrelated to this PR, I hope this becomes a command:
https://github.com/wasmerio/wasmer/pull/281#issuecomment-475689836
`wasmer call file.wasm function_name arg1 arg2`

lachlansneff: This is a good idea, we should make an issue for that.

src/bin/wasmer.rs
lachlansneff: Yeah, I wasn't sure what to do for that. Maybe we should keep as is and make a different pr for that?

bjfish: Yeah, I think a separate PR.

PR #390, opened 2019-04-23
pin wapm version

xmclark:
This PR pins the version of wapm-cli that gets distributed during releases. We can re-create the release, and will always install the same version of wapm-cli (until we update on a new wapm-cli release).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -103,7 +103,7 @@ jobs:
       - run:
           name: Install Rust
           command: |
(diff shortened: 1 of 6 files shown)


syrusakbary:
It failed for some random reason (SSL error).
Triggering the rebuild now:
bors r+

MarkMcCaskey:
bors retry

MarkMcCaskey:
bors retry

PR #398, opened 2019-04-27
Middleware system for code instrumentation.

losfair:


bjfish:
bors try

lib/middleware-common/src/call_trace.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn feed_event<'a, 'b: 'a>(
+        &mut self,
+        op: Event<'a, 'b>,
+        module_info: &ModuleInfo,
+        sink: &mut EventSink<'a, 'b>,
+    ) -> Result<(), Self::Error> {
+        match op {
+            Event::Internal(InternalEvent::FunctionBegin(id)) => {
+                sink.push(Event::Internal(InternalEvent::Bkpt(Box::new(move |_| {
+                    println!("func ({})", id);
```
bjfish: Could we also add args and print to stderr? Similar to: https://github.com/wasmerio/wasmer/blob/master/lib/clif-backend/src/resolver.rs#L389

bjfish: It is also behind the "debug" feature flag: https://github.com/wasmerio/wasmer/blob/master/lib/clif-backend/src/module_env.rs#L389

lib/middleware-common/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+pub mod call_trace;
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
```
bjfish: This test can be removed. We should try to add some tests.

lib/runtime-core/src/codegen.rs
```diff
@@ -0,0 +1,256 @@
...
+
+pub enum Event<'a, 'b> {
+    Internal(InternalEvent),
+    Wasm(&'b Operator<'a>),
+}
+
+pub enum InternalEvent {
+    FunctionBegin(u32),
+    FunctionEnd,
+    Bkpt(Box<Fn(BkptInfo) + Send + Sync + 'static>),
```
bjfish: We can convert `Bkpt` -> `Breakpoint` to be more readable.

lib/runtime-core/src/codegen.rs
```diff
@@ -0,0 +1,256 @@
...
+        &self,
+        wasm: &[u8],
+        compiler_config: CompilerConfig,
+        _: Token,
+    ) -> CompileResult<ModuleInner> {
+        struct Placeholder;
+        impl CacheGen for Placeholder {
+            fn generate_cache(&self) -> Result<(Box<[u8]>, Memory), CacheError> {
+                Err(CacheError::Unknown(
+                    "the singlepass backend doesn't support caching yet".to_string(),
```
bjfish: `singlepass` -> `placeholder` ?

lib/middleware-common/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "wasmer-middleware-common"
+version = "0.3.0"
+repository = "https://github.com/wasmerio/wasmer"
+description = "Wasmer runtime common middlewares"
+license = "MIT"
+authors = ["The Wasmer Engineering Team <engineering@wasmer.io>"]
+edition = "2018"
+
+[dependencies]
+wasmer-runtime-core = { path = "../runtime-core", version = "0.3.0" }
```
bjfish: @losfair Should we include this in runtime core since it is small and with few dependencies?

losfair: There will be more "utility"-like middlewares in this crate (e.g. metering). I think it would be better to keep it separate.

lib/middleware-common/src/call_trace.rs
losfair: Changed to print to stderr. It needs some machine-specific tricks to print args and I think we can defer it a bit.

losfair:
bors try

losfair:
bors try

lib/middleware-common/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "wasmer-middleware-common"
+version = "0.3.0"
```
MarkMcCaskey: We should be on version `0.4.0` now

syrusakbary: Let's do 
```suggestion
version = "0.4.0"
```

lib/middleware-common/Cargo.toml
syrusakbary:
```suggestion
wasmer-runtime-core = { path = "../runtime-core", version = "0.4.0" }
```

PR #399, opened 2019-04-27
Add extending WASI plugin example

MarkMcCaskey:


examples/plugin-for-example/README.md
```diff
@@ -0,0 +1,33 @@
...
+cd ..
+# Run the example
+cargo run --example plugin
+```
+
+## Explanation
+
+In this example, we instantiate a system with an extended (WASI)[wasi] ABI, allowing our program to rely on Wasmer's implementation of the syscalls defined by WASI as well as our own that we made.  This allows us to use the full power of an existing ABI, like WASI, and give it super-powers for our specific use case.
+
+Because the Rust WASI doesn't support the crate type of `cdylib`, we have to include a main function which we don't use.
```
xmclark: I also discovered this detail over the weekend when I was experimenting with WASI. Thank you for adding the clarification here.

examples/plugin-for-example/README.md
```diff
@@ -0,0 +1,33 @@
...
+See the `wasmer/examples/plugin.rs` file for the source code of the host system.
+
+## Compiling
+
+```
+# Install an up to date version of Rust nightly
+# Add the target
+rustup target add wasm32-unknown-wasi
+# build it
+cargo build --release --target=wasm32-unknown-wasi
```
xmclark: This requires `cargo +nightly`

examples/plugin-for-example/README.md
```diff
@@ -0,0 +1,33 @@
...
+# copy it to examples folder
+cp ../../target/wasm32-unknown-wasi/release/plugin-for-example.wasm ../
```
xmclark: It may also be worth noting that WASI is only supported on windows with the gnu toolchain. If you can't find a way to work this detail into the example, that's ok, but it could be another stumbling block for some.

examples/plugin.rs
```diff
@@ -0,0 +1,28 @@
+use wasmer_runtime::{func, imports, instantiate};
+use wasmer_runtime_core::vm::Ctx;
+use wasmer_wasi::generate_import_object;
+
+static PLUGIN_WASM: &'static [u8] = include_bytes!("plugin-for-example.wasm");
```
xmclark: how about instead of statically importing the bytes, we load them at runtime? We already do this in our rust example project. I think loading the wasm at runtime more accurately represents the behavior of a "plugin" as plugins are dynamically loaded (to me).

examples/plugin-for-example/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
```
xmclark: something new: can we also publish the generated wat? It may be helpful for users to observe the generated wat, and they can also run it from wasmer. 

`wapm install wabt && wapm run wasm2wat ./target/wasm32-unknown-wasi/release/plugin-for-example.wasm > ./examples/plugin-for-example.wat` should do the trick! 🎩 🌠

examples/plugin.rs
```diff
@@ -0,0 +1,28 @@
...
+fn main() {
+    // WASI imports
+    let mut base_imports = generate_import_object(vec![], vec![], vec![]);
+    // env is the default namespace for extern functions
+    let custom_imports = imports! {
+        "env" => {
+            "it_works" => func!(it_works),
+        },
+    };
+    base_imports.extend(custom_imports);
```
xmclark: Oh this is great! I forgot we implemented this. This deserves a nice comment because of how powerful it is. Maybe something like this?:
```suggestion
    // The WASI imports object contains all required import functions for a WASI module to run.
    // Extend this imports with our custom imports containing "it_works" function so that our custom wasm code may run.
    base_imports.extend(custom_imports);
```

examples/plugin.rs
```diff
@@ -0,0 +1,28 @@
...
+    let instance =
+        instantiate(PLUGIN_WASM, &base_imports).expect("failed to instantiate wasm module");
+
+    let entry_point = instance.func::<(i32), i32>("plugin_entrypoint").unwrap();
```
xmclark:
```suggestion
    // get a reference to the function "plugin_entrypoint" which takes an i32 and returns an i32
    let entry_point = instance.func::<(i32), i32>("plugin_entrypoint").unwrap();
```

examples/plugin.rs
```diff
@@ -0,0 +1,28 @@
...
+    let result = entry_point.call(2).expect("failed to execute plugin");
```
xmclark:
```suggestion
    // call the "entry_point" function in WebAssembly with the number "2" as the i32 argument
    let result = entry_point.call(2).expect("failed to execute plugin");
```

examples/plugin-for-example/README.md
MarkMcCaskey: Okay, I'm not sure exactly what to add to say that.  What I mean is, does that show up in the `target` dir?

examples/plugin-for-example/Cargo.toml
MarkMcCaskey: As discussed, I'm not going to include it because it's too long and hard to read (It's almost entirely WASI code);

I'll add a line to the example readme about it though

examples/plugin-for-example/README.md
```diff
@@ -0,0 +1,33 @@
+# WASI plugin example
+
+In this example we extend the imports of Wasmer's WASI ABI to demonstrate how custom plugins work.
+
+See the `wasmer/examples/plugin.rs` file for the source code of the host system.
+
+## Compiling
+
```
xmclark:
```suggestion
_Attention Windows users: WASI target only works with the `nightly-x86_64-pc-windows-gnu` toolchain._ 
```

PR #404, opened 2019-04-30
Update to newer common wasmparser version

bjfish:
- versions should match to have matching types in a common parser
- the `0.29.2` version matches the current parser version of cranelift-wasm


bjfish:
bors try

PR #414, opened 2019-05-02
Exception logging (win32): Handle error codes correctly, add some more strings

hrydgard:
Ran into a situation with an unknown exception from Cranelift (will probably report that one separately). Turns out the signum was "1" though which does not seem to correspond to any of the Windows error codes, except possibly STATUS_GUARD_PAGE which is 0x80000001, but only if we lost the top bit somewhere.

On Windows, exceptions seemed to be trapped by callProtected, which is implemented here: https://github.com/wasmerio/wasmer/blob/cade9a666f5dfedfc9352718988aa26f21b028f4/lib/win-exception-handler/exception_handling/exception_handling.c . It did not seem to correctly store and retrieve the exception code, instead always returning 1: ```longjmp(jmpBuf, 1);```

So I fixed it. And now the log output looks like this:

```
unhandled trap at 1560d5e7bab - code #c0000005: segmentation violation
```

### DIFF
--- a/lib/clif-backend/src/signal/windows.rs
+++ b/lib/clif-backend/src/signal/windows.rs
@@ -4,7 +4,6 @@ use crate::trampoline::Trampoline;
 use std::cell::Cell;
 use std::ffi::c_void;
 use std::ptr::{self, NonNull};
-use wasmer_runtime_core::error::{RuntimeError, RuntimeResult};
 use wasmer_runtime_core::typed_func::WasmTrapInfo;
 use wasmer_runtime_core::vm::Ctx;
 use wasmer_runtime_core::vm::Func;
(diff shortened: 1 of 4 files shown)


syrusakbary:
bors try

lib/win-exception-handler/exception_handling/exception_handling.c
```diff
@@ -74,7 +77,7 @@ uint8_t callProtected(trampoline_t trampoline,
         return TRUE;
     }
 
-    out_result->code = (uint64_t)signum;
+    out_result->code = (uint64_t)caughtExceptionCode;
```
xmclark: Great catch! Thanks for fixing this.

lib/clif-backend/src/signal/windows.rs
```diff
@@ -80,10 +83,26 @@ pub fn call_protected(
...
+            EXCEPTION_INT_DIVIDE_BY_ZERO => "int div by zero",
+            EXCEPTION_INT_OVERFLOW => "int overflow",
+            EXCEPTION_PRIV_INSTRUCTION => "priv instruction",
+            EXCEPTION_IN_PAGE_ERROR => "in page error",
+            EXCEPTION_NONCONTINUABLE_EXCEPTION => "non continuable exception",
+            EXCEPTION_STACK_OVERFLOW => "stack overflow",
+            EXCEPTION_GUARD_PAGE => "guard page",
+            EXCEPTION_INVALID_HANDLE => "invalid handle",
+            EXCEPTION_POSSIBLE_DEADLOCK => "possible deadlock",
+            _ => "unknown exception code",
```
xmclark: Having these extra cases will help with diagnosing errors. Thanks for adding them.

PR #416, opened 2019-05-02
Remote code loading framework.

losfair:


bjfish:
It would be nice to have a README.md in `lib/kwasm` to explain the library and how to build and run the examples.

lib/runtime-core/src/backend.rs
```diff
@@ -80,6 +77,9 @@ pub trait RunnableModule: Send + Sync {
     fn get_trampoline(&self, info: &ModuleInfo, sig_index: SigIndex) -> Option<Wasm>;
 
     unsafe fn do_early_trap(&self, data: Box<dyn Any>) -> !;
+
+    fn get_code(&self) -> Option<&[u8]> { None }
+    fn get_offsets(&self) -> Option<Vec<usize>> { None }
 }
```
bjfish: Please add doc comments to describe what these are.

lib/runtime-core/src/backing.rs
```diff
@@ -449,10 +449,17 @@ fn import_functions(
...
             }
             None => {
-                link_errors.push(LinkError::ImportNotFound {
-                    namespace: namespace.to_string(),
-                    name: name.to_string(),
-                });
+                if imports.allow_missing_functions {
+                    functions.push(vm::ImportedFunc {
+                        func: ::std::ptr::null(),
+                        vmctx: ::std::ptr::null_mut(),
```
bjfish: It would be nice if we could mock missing imports again (which is a separate feature though). We used to mock imports and just print a message like "mock import `func_test` called".

lib/runtime-core/src/codegen.rs
```diff
@@ -43,6 +43,8 @@ pub trait ModuleCodeGenerator<FCG: FunctionCodeGenerator<E>, RM: RunnableModule,
 
     /// Adds an import function.
     fn feed_import_function(&mut self) -> Result<(), E>;
+
+    fn feed_compiler_config(&mut self, config: &CompilerConfig) -> Result<(), E> { Ok(()) }
```
bjfish: I think it would be nice if the trait methods were in the order that the parser calls them. So this one would be higher.

lib/runtime-core/src/backing.rs
losfair: With the currently implementation the program receives a SIGSEGV when a null imported function is called, which I think is fine for now.

The import mocking feature can be added in a separate PR.

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

bjfish:
Please rename the `kwasm-*` crates to add the `wasmer-` prefix for consistency.

syrusakbary:
bors try

losfair:
bors try

losfair:
bors try

syrusakbary:
bors try

losfair:
bors try

PR #429, opened 2019-05-09
implement some FS operations for rsign2

MarkMcCaskey:
Get https://github.com/jedisct1/rsign2 working with WASI on Wasmer (for end-to-end tests of wapm)

Current implementation is very hacky, but works!  It should have a couple passes of improvement before shipping


syrusakbary:
This is awesome! Reviewing now

PR #432, opened 2019-05-10
feat(runtime-c-api) Check buffer size before creating the slice, and fix `wasmer_last_error_message` returned value

Hywan:
It's safer to check the buffer size is large enough to hold the error
message before creating the slice from raw parts.

Also, this patch remove the need for `last_error`, simplifying the
code a little bit. The `length` variable is casted to `usize` once.

This patch also updates the returned value of `wasmer_last_error_message` by adding 1, so that it mimics the `wasmer_last_error_length` function that counts the trailing null byte.

### DIFF
--- a/lib/runtime-c-api/src/error.rs
+++ b/lib/runtime-c-api/src/error.rs
@@ -61,19 +61,19 @@ pub unsafe extern "C" fn wasmer_last_error_message(buffer: *mut c_char, length: 
         return -1;
     }
 
-    let last_error = match take_last_error() {
-        Some(err) => err,
+    let error_message = match take_last_error() {
+        Some(err) => err.to_string(),
         None => return 0,
     };
 
-    let error_message = last_error.to_string();
+    let length = length as usize;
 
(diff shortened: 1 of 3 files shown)


Hywan:
bors try

Hywan:
bors try

Hywan:
I'm not sure about the last patch. I want to write a small test suite before merging.

lib/runtime-c-api/src/error.rs
```diff
@@ -61,20 +61,20 @@ pub unsafe extern "C" fn wasmer_last_error_message(buffer: *mut c_char, length:
...
 
-    let error_message = last_error.to_string();
+    let length = length as usize;
 
-    let buffer = slice::from_raw_parts_mut(buffer as *mut u8, length as usize);
-
-    if error_message.len() >= buffer.len() {
-        // buffer to small for err  message
+    if error_message.len() >= length {
+        // buffer to small to hold the error message
```
syrusakbary:
```suggestion
        // buffer too small to hold the error message
```

PR #379, opened 2019-04-22
Fix small return types.

lachlansneff:
This fixes returning integers smaller than 32bits from imported functions. Also some work on fixing returning errors from functions.

PR #435, opened 2019-05-12
Fix f32/f64 conversion, add tests, and rename conversion method

### LINKED ISSUE #415: Floating point arguments broken in func references (win32, regression in 0.4.0)
### Describe the bug

Calling functions with floating point parameters directly with ```instance.call("function_name",...);```  for example now works great, but calling it via ```Func.call``` doesn't, since 0.4.0. It worked in 0.3.0.

### Steps to reproduce

This can be reproduced easily in https://github.com/wasmerio/wasmer-rust-example/ :

Change Cargo.toml to use wasmer 0.4.0.

Then modify main.rs as follows:
```
instance.call("hello_wasm", &[wasmer_runtime::Value::F32(16.0)])?;
let f : wasmer_runtime::Func<(f32)> = instance.func("hello_wasm").unwrap();
f.call(16.0);
```
Also make the following edit to wasm-sample-app:
```
pub extern fn hello_wasm(f : f32) {
    // Call the function we just imported and pass in
    // the offset of our string and its length as parameters.
    unsafe {
      //print_str(HELLO.as_ptr(), HELLO.len());
      let s = format!("{}", f);
      print_str(s.as_ptr(), s.len());
    }
}
```
Compile both and see the output:
```
16
0.000000000000000000000000000000000000000000022
```
instead of the expected:
```
16
16
```

bjfish:
Fixes #415

I believe removing this conversion here caused the regression: https://github.com/wasmerio/wasmer/pull/379/files#diff-5e7e23717d05d1d18be5b2e3b75f454dL158

I've added conversion tests and also renamed `NativeWasmType::{to_bits, from_bits}` to `to_binary/from_binary` because it seemed akward to overload the existing `f32::to_bits` method.

### DIFF
--- a/lib/runtime-core/src/typed_func.rs
+++ b/lib/runtime-core/src/typed_func.rs
@@ -247,7 +247,7 @@ impl<A: WasmExternType> WasmTypeList for (A,) {
     type CStruct = S1<A>;
     type RetArray = [u64; 1];
     fn from_ret_array(array: Self::RetArray) -> Self {
-        (WasmExternType::from_native(NativeWasmType::from_bits(
+        (WasmExternType::from_native(NativeWasmType::from_binary(
             array[0],
         )),)
     }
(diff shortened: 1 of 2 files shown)


bjfish:
bors try

syrusakbary:
Can you check that WASI integrations (`cowsay`, `lolcat`, ...) example works with this fix?
If so, it should be good to merge!

bjfish:
@syrusakbary cowsay works for me and I think we have an integration test for that, also I don't see any lolcat error. But I have not seen any colors being output to my shell on this or previous versions for lolcat.

PR #440, opened 2019-05-14
fix(runtime-c-api) `wasmer_instance_call` types must matche `wasmer_export_func_*_arity`

Hywan:
The `wasmer_export_func_params_arity` and
`wasmer_export_func_returns_arity` functions store the arity in a
`uint32_t`. The `wasmer_instance_call` expects `c_int`. There is a
type mismatch here. It's not annoying in C or C++, but in some other
languages that have bindings to C/C++, it can imply useless casting.

This patch changes `wasmer_instance_call` to expect `uint32_t` for
`params_len` and `results_len` to match the
`wasmer_export_func_*_arity` functions.

Other patches do the same for `wasmer_export_func_params` and
`wasmer_export_func_returns`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Blocks of changes will separated by version increments.
 
 ## **[Unreleased]**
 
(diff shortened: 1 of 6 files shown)


Hywan:
bors try

Hywan:
bors try

PR #443, opened 2019-05-14
Return impl Compiler from default_compiler to fix compilation features

bjfish:
runtime compilation was failing (llvm/singlepass features) for the new compiler implementations for the `default_compiler` function due to changing types and state being added to the compilers.

This PR:
- adds runtime library tests for these features to catch the regression earlier
- refactors `default_compiler` to return `impl Compiler` and no longer a singleton to fix the compilation issue


bjfish:
bors try

PR #444, opened 2019-05-15
chore(runtime-c-api) Update `Cargo.toml`

Hywan:
Just in case. `cbindgen` has a new version that escape reserved keywords in C and C++. Hopefully, it changes nothing in our code, but it's better to get the updates before hitting the problems.


syrusakbary:
The PR has some conflicts. It should be good to merge once we solve them 👍

syrusakbary:
We should do this. But the PR is now stale, so it will be easier to create a new one from `master` than to resolve the conflicts.

Closing this PR :)

PR #445, opened 2019-05-15
Better README.md link target for "WASI"

Uzume:
Just updating the link target for "WASI" in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


Uzume:
Of course this could also link to https://github.com/CraneStation/wasmtime/blob/master/docs/WASI-api.md and this link be moved to there which might be a better solution.

Hywan:
Indeed, thanks for the PR!

bors r+

PR #448, opened 2019-05-15
update WASI fs abstraction to open files on each rw

### LINKED ISSUE #438: Likely unnecessary write access to a file
Hi!

While playing with wasmer as an embeddable wasm execution engine I discovered that wasmer-wasi requests more wide file-access than required in the code during opening a file.

For example:

```rust
let file = File::open("/proc/uptime")?;
```

fails without root permissions, while it shouldn't.

Stracing resulted in:

```
[pid 18721] openat(AT_FDCWD, "./uptime", O_RDWR|O_CLOEXEC = -1 EACCES (Permission denied)
```

In pure Rust, this opens `/proc/uptime` in O_RDONLY mode. In the case of wasmer-wasi it also requests write-access (O_RDWR). Here is the suspected code:
https://github.com/wasmerio/wasmer/blob/9aac20e2159274f2d41195f3510eb6995f8595d0/lib/wasi/src/syscalls/mod.rs#L1464

Is this a bug or I'm doing something wrong?

PS. I preopened `/proc` and even changed the working directory to it. However, it is quite strange for me why the full open path is stripped to just `./uptime`.

OS: linux 5.0.3, x86_64 x86_64 GNU/Linux
Version = 0.4.1

MarkMcCaskey:
resolves #438 

It's unclear whether this change is a good idea or not.

I'll have to look in to it some more.  Additionally, this PR could break existing code, we should test it

### DIFF
--- a/lib/wasi/src/state.rs
+++ b/lib/wasi/src/state.rs
@@ -18,31 +18,53 @@ pub const MAX_SYMLINKS: usize = 100;
 
 #[derive(Debug)]
 pub enum WasiFile {
-    HostFile(fs::File),
+    HostFile { path: PathBuf, offset: u64 },
+}
+
+impl WasiFile {
+    pub fn new_host_file(path: PathBuf) -> WasiFile {
+        WasiFile::HostFile { path, offset: 0 }
+    }
 }
 
 impl Write for WasiFile {
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
         match self {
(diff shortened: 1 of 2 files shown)


MarkMcCaskey:
We can always add an optimization here, having an `Option<File>` in `WasiFile::HostFile` to avoid forcing disk synchronization/opening and closing the same files over and over.

My intuition is that the solution _as is_ has an unacceptable performance impact

MarkMcCaskey:
closing in favor of #449

PR #450, opened 2019-05-16
Metering.

losfair:
TODO
- [x] Add test to verify points used after function call
- [x] Add test to verify function call will trap
- [x] Add documentation


losfair:
bors try

losfair:
bors try

lib/runtime-core/src/vm.rs
```diff
@@ -92,6 +92,8 @@ pub struct InternalCtx {
 
     pub memory_base: *mut u8,
     pub memory_bound: usize,
+
+    pub internals: *mut [u64; INTERNALS_SIZE], // TODO: Make this dynamic?
```
bjfish: Should we consider embedding array?

lib/middleware-common/src/metering.rs
```diff
@@ -0,0 +1,90 @@
...
+                    | Operator::If { .. }
+                    | Operator::Else
+                    | Operator::Unreachable
+                    | Operator::Br { .. }
+                    | Operator::BrTable { .. }
+                    | Operator::BrIf { .. }
+                    | Operator::Call { .. }
+                    | Operator::CallIndirect { .. }
+                    | Operator::Return => {
+                        sink.push(Event::Internal(InternalEvent::GetInternal(0)));
```
bjfish: Maybe we could name the internals indexes using const or enum?
```
enum Internals {
  Points = 0
}
```

losfair: Changed to dynamically allocating indexes.

lib/runtime-core/src/vm.rs
losfair: Internals are intended to be statically allocated to middleware modules. I think a fixed-size array would be suitable for this case.

losfair:
bors try

lib/llvm-backend/src/code.rs
```diff
@@ -2659,7 +2716,11 @@ impl ModuleCodeGenerator<LLVMFunctionCodeGenerator, LLVMBackend, CodegenError>
         Ok(())
     }
 
-    unsafe fn from_cache(artifact: Artifact, _: Token) -> Result<ModuleInner, CacheError> {
+    unsafe fn from_cache(_artifact: Artifact, _: Token) -> Result<ModuleInner, CacheError> {
+        Err(CacheError::Unknown(
+            "caching is broken for LLVM backend".into(),
```
bjfish: Is there a specific case where this is broken?

losfair: Absolute address is used to reference the breakpoint table. Therefore caching won't work. To fix this, we need to dynamically relocate the breakpoint table.

bjfish: Okay, I think we should return an error here then only when metering is enabled.

src/bin/wasmer.rs
```diff
@@ -118,6 +120,9 @@ struct Run {
     /// Application arguments
     #[structopt(name = "--", raw(multiple = "true"))]
     args: Vec<String>,
+
+    #[structopt(long = "inst-limit")]
+    instruction_limit: Option<u64>,
```
bjfish: I don't believe this CLI option will be needed and the common use case would be using it as library.

lib/middleware-common/src/metering.rs
```diff
@@ -0,0 +1,276 @@
...
+                        sink.push(Event::WasmOwned(Operator::I64Const {
+                            value: self.limit as i64,
+                        }));
+                        sink.push(Event::WasmOwned(Operator::I64GeU));
+                        sink.push(Event::WasmOwned(Operator::If {
+                            ty: WpType::EmptyBlockType,
+                        }));
+                        sink.push(Event::Internal(InternalEvent::Breakpoint(Box::new(
+                            move |ctx| {
+                                eprintln!("execution limit reached");
```
bjfish: We should remove debugging printlns before merging.

bjfish:
bors try

losfair:
bors try

bjfish:
bors try

bjfish:
bors try

bjfish:
bors try

PR #451, opened 2019-05-17
implement map-dir for WASI; fix bug in path_open

### LINKED ISSUE #458: --mapdir flag
Great stuff!  I've been implementing my own Wasi library in [Assemblyscript](https://github.com/WebAssemblyOS/wasmos) and have had some initial successes using wasmer!

### Motivation

It would be nice for end users if their initial directory path was root, "/" instead of either a fullpath or ".".

Wasmtime has this option so that you can map an external pathname to an internal one.

e.g. `--mapdir=/tmp:/var/tmp`

### Proposed solution

Add command line option and open directories on the host with the path left of the ":" and name it with the left half in the runtime.

[Currently this is done here.](https://github.com/wasmerio/wasmer/blob/88124d926eaa45e2dfd6424f2b86cd3e728fa6ff/lib/wasi/src/state.rs#L193)

Wasmtime does it [here](https://github.com/CraneStation/wasmtime/blob/3ae7c60b1395fe25971f683828241e7fac8cb40b/src/wasmtime.rs#L135)

### Alternatives

One alternative is to just interpret each path as a mapped directory if there is a ":" in the name.  This seems like the faster way, but it would be better if the API reflected the difference.

MarkMcCaskey:
Colon separated mapping, implemented to be compatible with wasmtime

resolves #458 

example:
```Rust
use std::fs;

fn main() {
    let read_dir = fs::read_dir(".").unwrap();
    for entry in read_dir {
        println!("{:?}", entry.unwrap().path());
    }
}
```
```shell
$ cargo run --release --bin wasmer -- run --mapdir=.:src list-files.wasm
"./bin"
"./installer"
"./lib"
"./update"
"./utils"
"./webassembly"
```


MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

PR #461, opened 2019-05-22
fix(runtime-c-api) Prevent negative lengths in various places

Hywan:
This PR prevents receiving negative lengths in various places in the code.


Hywan:
bors try

PR #467, opened 2019-05-24
feat(runtime-c-api) `wasmer_instantiate` raises the source error

Hywan:
Instead of returning only “error instanting”, `wasmer_instantiate` can
return the real error message.


Hywan:
bors try

PR #473, opened 2019-05-29
add experimental rendering with opengl and sdl

MarkMcCaskey:
very much a work in progress

we'll need to handle webgl properly


MarkMcCaskey:
Closing this for now because it's out of date, there's been more discussion on the WASI side about using WebGPU and that seems like the way to go for now.

I can imagine a future where we support versions of OpenGL and other graphics libraries for extra compatibility, but I don't think this is something we'll be doing soon.

PR #476, opened 2019-05-30
fix bug in wasi::environ_get, fix off by one error in env_size_get

### LINKED ISSUE #475: Ability to set environment variables
I couldn't find a way to set an environment variable neither in the help nor in the Rust code.

### Motivation

Because it's an easy way to configure a program's behavior.

### Proposed solution

Add a `--env key=value` cli option.

MarkMcCaskey:
adds `--env` flag (resolves #475)
fixes other wasi env bugs
adds wasi envvar test


syrusakbary:
That's a great finding.

I think it will be good to:
* Add integration tests
* Only allow part of the environment variables to leak via a new CLI option

lib/wasi/wasitests/envvar.rs
```diff
@@ -0,0 +1,21 @@
...
+    let mut env_vars = env::vars()
+        .map(|(key, value)| format!("{}={}", key, value))
+        .collect::<Vec<String>>();
+
+    env_vars.sort();
+
+    println!("Env vars:");
+    for e in env_vars {
+        println!("{}", e);
+    }
```
syrusakbary: Can we add a case to retrieve a single env var when:
* env var is present (such as `DOG`)
* env var is not present (such as `UNKNOWN`)

MarkMcCaskey:
bors try

PR #479, opened 2019-06-02
Added missing symbols for `ffmpeg`

piranna:


hrydgard:
replacing _llvm_exp2_f32 with _llvm_exp2_f64 that way looks a bit dubious

piranna:
> replacing _llvm_exp2_f32 with _llvm_exp2_f64 that way looks a bit dubious

I know, but for some reason I got signature errors on this point. I'm open to discussion about how to fix it.

lib/emscripten/src/emscripten_target.rs
```diff
@@ -95,6 +95,10 @@ pub fn _pthread_attr_setstacksize(_ctx: &mut Ctx, _a: i32, _b: i32) -> i32 {
     debug!("emscripten::_pthread_attr_setstacksize");
     0
 }
+pub fn _pthread_cancel(_ctx: &mut Ctx, _a: i32) -> i32 {
+    debug!("emscripten::_pthread_cancel");
```
MarkMcCaskey: One small note:  I changed this in the last PR I did: I made pthread functions `trace` instead of `debug`.  We have to compile with `--features=trace` instead.

I think it's a good change though because they can be really verbose and they don't do anything.  I found that they were hiding the important information from Python so I changed them.

I'll post about it on Slack, too

syrusakbary:
Closing this PR, as is not relevant any longer

PR #481, opened 2019-06-03
Added context trampoline into runtime

syrusakbary:
Based on #450

### DIFF
--- a/lib/runtime-c-api/src/trampoline.rs
+++ b/lib/runtime-c-api/src/trampoline.rs
@@ -19,13 +19,25 @@ pub extern "C" fn wasmer_trampoline_buffer_builder_new() -> *mut wasmer_trampoli
 
 #[no_mangle]
 #[allow(clippy::cast_ptr_alignment)]
-pub unsafe extern "C" fn wasmer_trampoline_buffer_builder_add_function(
+pub unsafe extern "C" fn wasmer_trampoline_buffer_builder_add_context_trampoline(
     b: *mut wasmer_trampoline_buffer_builder_t,
     f: *const wasmer_trampoline_callable_t,
     ctx: *const c_void,
 ) -> usize {
     let b = &mut *(b as *mut TrampolineBufferBuilder);
-    b.add_function(f as *const CallTarget, ctx as *const CallContext)
+    b.add_context_trampoline(f as *const CallTarget, ctx as *const CallContext)
+}
+
+#[no_mangle]
+#[allow(clippy::cast_ptr_alignment)]
+pub unsafe extern "C" fn wasmer_trampoline_buffer_builder_add_callinfo_trampoline(
+    b: *mut wasmer_trampoline_buffer_builder_t,
+    f: *const wasmer_trampoline_callable_t,
+    ctx: *const c_void,
+    num_params: u32,
+) -> usize {
+    let b = &mut *(b as *mut TrampolineBufferBuilder);
+    b.add_callinfo_trampoline(::std::mem::transmute(f), ctx as *const CallContext, num_params)
 }
 
 #[no_mangle]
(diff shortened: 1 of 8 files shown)


syrusakbary:
bors try

lib/runtime-c-api/tests/test-import-function.c
```diff
@@ -14,8 +14,15 @@ typedef struct {
   int value;
 } context_data;
 
+struct print_str_context {
+    int call_count;
+};
+
 void print_str(wasmer_instance_context_t *ctx, int32_t ptr, int32_t len)
```
Hywan: This PR is an excellent start, thank you, but it doesn't solve the issue we have with dynamic languages like PHP, Python and so on.

In C, Go, Rust… `print_str` is declared with static arguments (here `ptr` and `len`). In dynamic languages, the function isn't statically declared. Generally, a “function” is a combination of one or more structures (in PHP for instance, this is two, in Python, this is one) and the VM has some mechanisms to call the function. We __cannot__ declare a list of arguments. Thus, an API like https://github.com/WebAssembly/wasm-c-api/blob/040b983b517ff8ef431f9396e8b98b096fa37d28/example/hello.c#L11-L13 where arguments are stored in an array (so are dynamic, opposed to statically declared) is ideal and the only way to make it work.

I don't know if I'm clear. Am I?

losfair: Fixed in latest commit.

lib/runtime-c-api/src/trampoline.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[no_mangle]
+#[allow(clippy::cast_ptr_alignment)]
+pub unsafe extern "C" fn wasmer_trampoline_buffer_builder_add_function(
+    b: *mut wasmer_trampoline_buffer_builder_t,
+    f: *const wasmer_trampoline_callable_t,
+    ctx: *const c_void,
+) -> usize {
+    let b = &mut *(b as *mut TrampolineBufferBuilder);
+    b.add_function(f as *const CallTarget, ctx as *const CallContext)
```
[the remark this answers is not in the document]
losfair: Do we really need to check null pointers explicitly that way?

It's easy to debug SIGSEGV's caused by null pointers.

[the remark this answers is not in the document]
Hywan: This is what we did in the other function, it's debatable :-).

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -0,0 +1,245 @@
...
+        idx
+    }
+
+    /// Adds a callinfo trampoline.
+    ///
+    /// This generates a trampoline function that collects `num_params` parameters into an array
+    /// and passes the array into `target` as the second argument when called. The first argument
+    /// of `target` is the `context` specified here.
+    ///
+    /// Note that non-integer parameters/variadic functions are not supported.
```
Hywan: Does it support floats (I think it does not matter since the bits can be interpreted as a float, but I prefer to ask)?

lib/runtime-c-api/tests/test-import-function-callinfo.c
```diff
@@ -0,0 +1,132 @@
...
+        .call_count = 0
+    };
+
+    printf("Creating trampoline buffer\n");
+    wasmer_trampoline_buffer_builder_t *tbb = wasmer_trampoline_buffer_builder_new();
+    unsigned long print_str_idx = wasmer_trampoline_buffer_builder_add_callinfo_trampoline(
+        tbb,
+        (wasmer_trampoline_callable_t *) print_str,
+        (void *) &local_context,
+        3
```
Hywan: 3 for the instance context + `print_str` arguments, is that correct?

I wonder if we should keep the instance context 🤔… it's for another PR anyway.

lib/runtime-core/src/trampoline_x64.rs
losfair: Floats are passed in different registers and it would be very complex to handle this case. I think it's better to just support integer parameters.

Hywan: So trampoline doesn't work with floats. It's important to know!

PR #486, opened 2019-06-08
fix stack offset check

laizy:
note: I have not read the whole code yet, so I might be wrong.


syrusakbary:
Thanks for contributing @laizy.

I think it will very useful to add a test case verifying the fix, so we assure it doesn't repeat in the future.
If you can add it to this PR it would be great!

losfair:
Thanks!

I've added a WASM test case for this and fixed a typo. The new PR is #487 .

PR #489, opened 2019-06-09
Su Engine: Decoding and encoding runtime state.

losfair:
This PR implements a managed runtime that is able to suspend a running WebAssembly program at arbitrary point in time, decode its machine state (registers and stack) into the corresponding WebAssembly abstract state (call frames, locals and the value stack), and encode the abstract state back into machine state for resuming execution later.

Features enabled by this PR include:

- Tier (compiler backend) switching at runtime.
- Debugging with backtraces and local variables.
- Suspending to disk.
- Live migration.

The name "Su" corresponds to 「溯」 in Chinese, originating from 「溯洄从之」 in The Book of Songs.


losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

lib/runtime-core/src/codegen.rs
```diff
@@ -9,13 +9,17 @@ use crate::{
...
+use std::collections::HashMap;
 use std::fmt;
 use std::fmt::Debug;
 use std::marker::PhantomData;
 use std::sync::{Arc, RwLock};
 use wasmparser::{self, WasmDecoder};
 use wasmparser::{Operator, Type as WpType};
 
+pub type BkptHandler = Box<Fn(BkptInfo) -> Result<(), Box<dyn Any>> + Send + Sync + 'static>;
+pub type BkptMap = Arc<HashMap<usize, BkptHandler>>;
```
bjfish: Style suggestion: I think we could use full names like "BreakpointMap".

lib/runtime-core/src/state.rs
```diff
@@ -0,0 +1,843 @@
...
+    pub globals: Vec<u64>,
+    pub execution_state: ExecutionStateImage,
+}
+
+impl ModuleStateMap {
+    fn lookup_call_ip(&self, ip: usize, base: usize) -> Option<(&FunctionStateMap, MachineState)> {
+        if ip < base || ip - base >= self.total_size {
+            None
+        } else {
+            //println!("lookup ip: {} in {:?}", ip - base, self.local_functions);
```
bjfish: Remove commented out code or make it `debug`.

lib/runtime-core/Cargo.toml
```diff
@@ -18,6 +18,8 @@ errno = "0.2.4"
 libc = "0.2.49"
 hex = "0.3.2"
 smallvec = "0.6.9"
+bincode = "1.1"
+colored = "1.8"
```
bjfish: Should `colored` be moved to a higher level project like `wasmer` 🤔

lib/runtime-core/image-loading-linux-x86-64.s
```diff
@@ -0,0 +1,69 @@
+# NOTE: Keep this consistent with `alternative_stack.rs`.
```
bjfish: Conceptually what is the "alternative stack" and how is it used?

lib/runtime-core/src/alternative_stack.rs
```diff
@@ -0,0 +1,462 @@
...
+    signum: ::nix::libc::c_int,
+    siginfo: *mut siginfo_t,
+    ucontext: *mut c_void,
+) {
+    unsafe {
+        let fault = get_fault_info(siginfo as _, ucontext);
+
+        let mut unwind_result: Box<dyn Any> = Box::new(());
+
+        let should_unwind = allocate_and_run(1048576, || {
```
bjfish: Could `1048576` be a constant?

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -1114,6 +1233,29 @@ impl X64FunctionCode {
                     call_movs.push((*param, x));
                 }
                 Location::Memory(_, _) => {
+                    match *param {
+                        Location::GPR(x) => {
+                            let content = m.state.register_values[X64Register::GPR(x).to_index().0];
+                            //assert!(content != MachineValue::Undefined);
```
bjfish: Assertions needed or should they be removed?

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -1466,6 +1702,18 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
     }
 
     fn feed_event(&mut self, ev: Event, module_info: &ModuleInfo) -> Result<(), CodegenError> {
+        let a = self.assembler.as_mut().unwrap();
```
bjfish: Style suggestion: consider naming this `assembler` instead of just `a`.

lib/runtime-core/src/alternative_stack.rs
```diff
@@ -0,0 +1,462 @@
...
+                unwind_result = Box::new(image);
+            } else {
+                use colored::*;
+                if es_image.frames.len() > 0 {
+                    eprintln!(
+                        "\n{}",
+                        "Wasmer encountered an error while running your WebAssembly program."
+                            .bold()
+                            .red()
+                    );
```
bjfish: Should we avoid using stderr and prefer providing users with an `Err` value to have them decide if stderr should be used?

lib/singlepass-backend/src/codegen_x64.rs
losfair: Since the assembler is used a lot in `feed_event`, I think renaming it to `assembler` will make the code too verbose.

lib/runtime-core/image-loading-linux-x86-64.s
losfair: `alternative_stack.rs` is mainly those code that works with the OS. Might be a good idea to rename it to something else.

lib/runtime-core/src/alternative_stack.rs
losfair: This is designed to be similar to how the Rust panic backtrace works. Some more discussion and thinking on this is needed though and we can change this in the future.

lib/runtime-core/src/state.rs
MarkMcCaskey: or make it `trace!` if it's very verbose 😸

lib/singlepass-backend/src/codegen_x64.rs
MarkMcCaskey: `debug_assert!` is also a good choice if you think you don't need it but future changes could cause it to be triggered

losfair: Added some comments to explain why this assertion should remain commented.

lib/runtime-core/Cargo.toml
losfair: How can we use dependencies in the top level workspace in `wasmer-runtime-core`?

bjfish: I think some user, like runtime and runtime-core users may not have a need for colored output since the usage is not in a terminal, for these cases could this be moved to a higher level project?

losfair: I think we can change this later when there's a need for high modularity. Currently I think it's okay to simulate Rust libstd's behavior on panic (output to stderr).

bjfish:
bors try

PR #494, opened 2019-06-12
fix(runtime-c-api) Replace deprecated type aliases from libc

Hywan:
Since https://github.com/rust-lang/libc/pull/1379, fixed width integer
type aliases are deprecated. Thus, this patch uses Rust types instead
of libc aliases.


Hywan:
bors try

PR #505, opened 2019-06-19
C API Variadic Import Func

bjfish:
The test here specifies a new API `wasmer_import_func_variadic_new` which would allow defining a new imported function which receives arguments as an array. For example, `void print_str(const wasmer_instance_context_t *ctx, const wasmer_value_t params[], wasmer_value_t results[])` in contrast to the current requirement of defining specific function types for an imported function: `void print_str(wasmer_instance_context_t *ctx, int32_t ptr, int32_t len)`.

Use case:
For dynamic language implementations (like Ruby), the desired import functions may not be defined until runtime. This makes it difficult to define the specific C ABI functions which would be required to satisfy function definitions.

Using the variadic style interface here, a C function can be defined in advance to handle variable function signatures defined at runtime.


Hywan:
The API is excellent :+1:

bjfish:
@Hywan Does the rust version of the API look good? https://github.com/wasmerio/wasmer/pull/505/commits/112d05a8bced1f48cd2af198eb121655f631c3f8

bjfish:
@Hywan has this API been created elsewhere already?

syrusakbary:
This should be now completely solved by the refactor that has just landed in `master` and the adoption of `wasm_c_api`.

Closing the PR :)

PR #513, opened 2019-06-27
Fix lseek in emscripten

syrusakbary:
Fix lseek in emscripten

### DIFF
--- a/lib/emscripten/src/syscalls/mod.rs
+++ b/lib/emscripten/src/syscalls/mod.rs
@@ -446,7 +446,7 @@ pub fn ___syscall140(ctx: &mut Ctx, _which: i32, mut varargs: VarArgs) -> i32 {
     let ret = unsafe { lseek(fd, offset, whence) };
     #[allow(clippy::cast_ptr_alignment)]
     let result_ptr_0 = emscripten_memory_pointer!(ctx.memory(0), result_ptr_value) as *mut i32;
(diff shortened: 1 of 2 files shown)


lib/emscripten/src/syscalls/mod.rs
```diff
@@ -437,23 +437,26 @@ pub fn ___syscall140(ctx: &mut Ctx, _which: i32, mut varargs: VarArgs) -> i32 {
     // -> c_int
     debug!("emscripten::___syscall140 (lseek) {}", _which);
     let fd: i32 = varargs.get(ctx);
-    let _ = varargs.get::<i32>(ctx); // ignore high offset
+    let _offset_high: i32 = varargs.get(ctx); // We don't use the offset high as emscripten skips it
     let offset_low: i32 = varargs.get(ctx);
-    let result_ptr_value = varargs.get::<i32>(ctx);
+    let result_ptr_value: i32 = varargs.get(ctx);
```
MarkMcCaskey:
```
let memory = ctx.memory(0);
let result_ptr_value: WasmPtr<i64> = varargs.get(ctx).unwrap();

let result_ptr =result_ptr_value.deref(memory);
result_ptr.set(ret);
```

Will do what you want I think!

MarkMcCaskey: and it doesn't need that `assert` or any of the `emscripten_memory_pointer` stuff

PR #515, opened 2019-06-30
Improved emscripten dyncalls

syrusakbary:
Improved emscripten dyncalls

### DIFF
--- a/lib/emscripten/src/emscripten_target.rs
+++ b/lib/emscripten/src/emscripten_target.rs
@@ -858,6 +858,10 @@ pub fn invoke_vijj(ctx: &mut Ctx, index: i32, a1: i32, a2: i32, a3: i32, a4: i32
         panic!("dyn_call_vijj is set to None");
     }
 }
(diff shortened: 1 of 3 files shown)


lib/emscripten/src/emscripten_target.rs
```diff
@@ -634,6 +634,11 @@ pub fn invoke_iij(ctx: &mut Ctx, index: i32, a1: i32, a2: i32, a3: i32) -> i32 {
     invoke!(ctx, dyn_call_iij, index, a1, a2, a3)
 }
 
+pub fn invoke_iji(ctx: &mut Ctx, index: i32, a1: i32, a2: i32, a3: i32) -> i32 {
```
piranna: Any reference to the definition of this syscall? I've been looking for it on internet and found no clue... :-/

syrusakbary: Is not a syscall. The dyncalls are the way emscripten wraps functions to be able to try/catch them natively (either via js, or Rust in our case).

PR #532, opened 2019-07-06
`dash` and `xz` support

piranna:
This pull request implement changes needed by both `dash` shell and `xz`. `xz` is fully working, but `dash` needs still some changes, though.


lib/emscripten/src/syscalls/unix.rs
```diff
@@ -418,30 +420,21 @@ pub fn ___syscall330(ctx: &mut Ctx, _which: c_int, mut varargs: VarArgs) -> pid_
...
-            debug!("ret(FIONBIO): {}", ret);
-            ret
-            // 0
-        }
-        21523 => {
-            // TIOCGWINSZ
-            let argp: u32 = varargs.get(ctx);
-            let argp_ptr = emscripten_memory_pointer!(ctx.memory(0), argp) as *mut c_void;
-            let ret = unsafe { ioctl(fd, TIOCGWINSZ, argp_ptr) };
+        TIOCGWINSZ => {
```
syrusakbary: The value of `TIOCGWINSZ` is different per OS in libc:
https://github.com/rust-lang/libc/search?q=TIOCGWINSZ&unscoped_q=TIOCGWINSZ

Since the WebAssembly is compiled with a fixed value, the value here should be fixed too.

```suggestion
        21523 => {
```

lib/emscripten/src/syscalls/unix.rs
```diff
@@ -452,13 +445,15 @@ pub fn ___syscall54(ctx: &mut Ctx, _which: c_int, mut varargs: VarArgs) -> c_int
                 ret
             }
         }
-        _ => {
-            debug!(
-                "emscripten::___syscall54 -> non implemented case {}",
-                request
-            );
+        TIOCSPGRP => {
```
syrusakbary: Change to the fixed value

lib/emscripten/src/syscalls/unix.rs
```diff
@@ -452,13 +445,15 @@ pub fn ___syscall54(ctx: &mut Ctx, _which: c_int, mut varargs: VarArgs) -> c_int
...
+            debug!("ret(TIOCSPGRP): (noop)");
             0
         }
+        _ => {
+            let ret = unsafe { ioctl(fd, request as _, argp_ptr) };
```
syrusakbary: We can't do this (as exposed in the previous comment).

syrusakbary: If you want to do it in a generic way, we need a proper mapping of fixed values to the `libc` values.

lib/emscripten/src/syscalls/unix.rs
piranna: Good catch, really, didn't thought about that point and instead focused on readibility. Ok, I'll remove the value fetch from `libc` and set it as a constact at file top.

lib/emscripten/src/syscalls/unix.rs
piranna: What targets are we focused or are we going to be focused in the short term at this moment beyond x86_64? AArch64? Maybe ARM in 32 bits? What flavour (v6 for Rapsberry Pi 1? v7? Cortex_A42?)

syrusakbary: x86_64 and AArch64 are the main ones... why is the target relevant?

piranna: To know to what platforms should I do the mapping.

syrusakbary: I think you don't really need to know the platform, you can use libc values for the real calls.

Here's an example on how:

```rust
use libc::TIOCSPGRP;

WASM_TIOCSPGRP = 21523;

// This is not syntactically valid:
let MAPPING = {
   WASM_TIOCSPGRP: TIOCSPGRP
}

// ...

let ret = unsafe { ioctl(fd, MAPPING[request], argp_ptr) };

```

piranna: Touchée, I got distracted with the previous comment about the values can be different for each platform. Yes, `libc` can do the mapping by ourselves.

I need to left work for today and go to the LGTB Proud parade to distract myself a bit :sweat_smile: :unicorn: :rainbow_flag: :-)

piranna: Done :-)

lib/emscripten/src/syscalls/unix.rs
```diff
@@ -77,6 +78,23 @@ use libc::{
...
+
+// Based on @syrusakbary sugerence at
+// https://github.com/wasmerio/wasmer/pull/532#discussion_r300837800
+fn translate_ioctl(wasm_ioctl: u32) -> c_ulong {
+    match wasm_ioctl {
+        WASM_FIOCLEX => FIOCLEX,
+        WASM_TIOCGWINSZ => TIOCGWINSZ,
+        WASM_TIOCSPGRP => TIOCSPGRP,
+
+        otherwise => otherwise as c_ulong,
```
syrusakbary: I think doing an `unimplemented!()` might be better, just in case.
Thoughts?

syrusakbary:
I added all this comments into #533, as I wanted to use the refactor of `ioctl`

syrusakbary:
Closing this as #533 is already merged and have the changes in this PR

PR #533, opened 2019-07-07
Make wasmer compatible with latest Emscripten

syrusakbary:
The new Emscripten upstream, which [uses the LLVM backend by default](https://v8.dev/blog/emscripten-llvm-wasm) has some internal changes. This PR adds support for them:

* [x] Automatically expose the methods named `_xxx` also as `xxx`
* [x] Refactored pthreads out of `emscripten_target`
* [x] Make `malloc` and other methods optional
* [x] Add "virtual" support for Shared memories (even if they are not supported yet in the global scope)
* [x] Added `contains_key` method to Namespace
* [x] Added `ctime` and `ctime_r` methods to Emscripten (so pypy can run)


lib/emscripten/src/env/mod.rs
```diff
@@ -22,11 +22,23 @@ use wasmer_runtime_core::{
...
 pub fn call_malloc_with_cast<T: Copy, Ty>(ctx: &mut Ctx, size: u32) -> WasmPtr<T, Ty> {
-    WasmPtr::new(get_emscripten_data(ctx).malloc.call(size).unwrap())
+    WasmPtr::new(
+        get_emscripten_data(ctx)
+            .malloc
+            .as_ref()
+            .unwrap()
+            .call(size)
+            .unwrap(),
+    )
```
piranna: Since this code is the same as `call_malloc()`, does it makes sense to call it to remove duplicated code?

lib/emscripten/src/memory.rs
```diff
@@ -65,6 +66,57 @@ pub fn _emscripten_resize_heap(ctx: &mut Ctx, requested_size: u32) -> u32 {
...
+//  }
+//  if ((newDynamicTop | 0) > (totalMemory | 0)) {
+//   if (_emscripten_resize_heap(newDynamicTop | 0) | 0) {} else {
+//    ___setErrNo(12);
+//    return -1;
+//   }
+//  }
+//  HEAP32[DYNAMICTOP_PTR >> 2] = newDynamicTop | 0;
+//  return oldDynamicTop | 0;
+// }
```
piranna: Why is this commented code being added? Will it be used in the future?

syrusakbary:
bors try

lib/emscripten/src/syscalls/unix.rs
```diff
@@ -72,8 +76,38 @@ use libc::{
     F_SETFD,
     SOL_SOCKET,
     TIOCGWINSZ,
+    TIOCSPGRP,
+    // TCGETS,
+    // TCSETSW,
 };
 
+const TCGETS: u64 = 0x5401;
+const TCSETSW: u64 = 0x5403;
```
piranna: These are not needed since they are the platform dependent translated values, not the wasm ones. Remove and use the ones from libc.

lib/emscripten/src/syscalls/unix.rs
```diff
@@ -72,8 +76,38 @@ use libc::{
...
+fn translate_ioctl(wasm_ioctl: u32) -> c_ulong {
+    match wasm_ioctl {
+        WASM_FIOCLEX => FIOCLEX,
+        WASM_TIOCGWINSZ => TIOCGWINSZ,
+        WASM_TIOCSPGRP => TIOCSPGRP,
+        WASM_FIONBIO => FIONBIO,
+        WASM_TCGETS => TCGETS,
+        WASM_TCSETSW => TCSETSW,
+        _otherwise => {
+            unimplemented!("The ioctl {} is not yet implemented", wasm_ioctl);
```
piranna: Definitely better, yes. Good point.

lib/emscripten/src/syscalls/unix.rs
syrusakbary: They are not exposed in macOS, that's why I had to do it this way.

piranna: Ok :+1: Maybe add a comment showing this corner case?

PR #536, opened 2019-07-08
Update cache key generation to use backend compiler name too

MarkMcCaskey:


MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/runtime-core/src/cache.rs
```diff
@@ -41,12 +42,13 @@ impl WasmHash {
     /// # Note:
     /// This does no verification that the supplied data
     /// is, in fact, a wasm module.
-    pub fn generate(wasm: &[u8]) -> Self {
+    pub fn generate(wasm: &[u8], backend: Backend) -> Self {
```
syrusakbary: Can we have this as a non-breaking change. I think near was relying on `generate` and I would like to keep the function signature the same.

Perhaps something like?
```
    pub fn generate_for_backend(wasm: &[u8], backend: Option<Backend>) -> Self {
    }
    pub fn generate(wasm: &[u8]) -> Self {
       Self::generate_for_backend(wasm, Some(Cranelift);
    }
```

MarkMcCaskey: Sure, the main reason I didn't do that is that it has the potential for bad things to happen, but I'll call it out in a doc comment and call it good enough!

MarkMcCaskey:
bors try

MarkMcCaskey:
bors retry

MarkMcCaskey:
bors retry

PR #542, opened 2019-07-09
Initial commit of SIMD support in wasmer.

nlewycky:
Add SIMD support to wasmer, mostly unimplemented. There's an implementation in the LLVM backend, but the other backends are left unimplemented, and the wasmer C API also leaves V128 unimplemented.

Updates to LLVM 8.0.


Cargo.toml
```diff
@@ -22,7 +22,7 @@ include = [
 byteorder = "1.3.1"
 errno = "0.2.4"
 structopt = "0.2.11"
-wabt = "0.7.2"
+wabt = { git = "https://github.com/pepyakin/wabt-rs", rev = "60c196ce2b946e271e66da80f413d84049ba0a76" }
```
bjfish: Is there a publishability concern here if this is got based?

nlewycky: Yes. I asked @MarkMcCaskey and he said it'd be fine to break wasmer-llvm-backend's pushability since it's never been pushed, but we shouldn't break the rest of wasmer right now.

lib/llvm-backend/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2018"
 
 [dependencies]
 wasmer-runtime-core = { path = "../runtime-core", version = "0.5.4" }
-inkwell = { git = "https://github.com/wasmerio/inkwell", branch = "llvm7-0" }
+inkwell = { git = "https://github.com/lumen/inkwell", rev = "64921e418c3e8f4b32f9beed56ebd5aac4e677c7", features = ["llvm8-0"] }
```
syrusakbary: Can we publish this branch in our own inkwell repo?

Cargo.toml
MarkMcCaskey: Yeah, we can deal this when we want to deal with publishing our LLVM work as a whole.  I figure that it won't be for quite a while, so it doesn't really matter, many of these changes will probably be upstreamed!

Good catch though, Brandon!

lib/llvm-backend/Cargo.toml
nlewycky: Done. https://github.com/wasmerio/inkwell/tree/llvm8-0

lib/runtime-core/src/instance.rs
```diff
@@ -595,13 +618,29 @@ fn call_func_with_index(
...
+            run_wasm(result.as_mut_ptr())?;
+
+            let mut bytes = [0u8; 16];
+            let lo = result[0].to_le_bytes();
+            let hi = result[1].to_le_bytes();
+            for i in 0..8 {
+                bytes[i] = lo[i];
+                bytes[i + 8] = hi[i];
+            }
+            rets.push(Value::V128(u128::from_le_bytes(bytes)));
```
MarkMcCaskey: you can use clone_from_slice like you do in other places; might be worth having a helper function for this

nlewycky: I don't see how to use clone_from_slice here? `clone_from_slice` will panic if the sizes are different, which they will be because `lo` is 8 bytes and `bytes` is 16 bytes?

lib/runtime-core/src/instance.rs
```diff
@@ -595,13 +618,29 @@ fn call_func_with_index(
         Type::I64 => Value::I64(raw as i64),
         Type::F32 => Value::F32(f32::from_bits(raw as u32)),
         Type::F64 => Value::F64(f64::from_bits(raw)),
+        _ => unreachable!(),
```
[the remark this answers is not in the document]
nlewycky: Done.

lib/runtime-c-api/src/value.rs
```diff
@@ -113,6 +114,7 @@ impl From<&wasmer_runtime::wasm::Type> for wasmer_value_tag {
             Type::I64 => wasmer_value_tag::WASM_I64,
             Type::F32 => wasmer_value_tag::WASM_F32,
             Type::F64 => wasmer_value_tag::WASM_F64,
+            _ => panic!("not implemented"),
```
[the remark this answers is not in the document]
nlewycky: Done.

lib/runtime-c-api/src/value.rs
```diff
@@ -76,6 +76,7 @@ impl From<Value> for wasmer_value_t {
                 tag: wasmer_value_tag::WASM_F64,
                 value: wasmer_value { F64: x },
             },
+            _ => panic!("not implemented"),
```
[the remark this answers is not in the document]
nlewycky: Done, updated a few other pre-existing instances in this file too. Please take a look.

lib/runtime-c-api/src/instance.rs
```diff
@@ -173,6 +173,7 @@ pub unsafe extern "C" fn wasmer_instance_call(
                         tag: wasmer_value_tag::WASM_F64,
                         value: wasmer_value { F64: x },
                     },
+                    _ => panic!("not implemented"),
```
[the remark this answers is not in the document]
nlewycky: Done.

lib/runtime-c-api/src/export.rs
```diff
@@ -402,6 +402,7 @@ pub unsafe extern "C" fn wasmer_export_func_call(
                         tag: wasmer_value_tag::WASM_F64,
                         value: wasmer_value { F64: x },
                     },
+                    _ => panic!("not implemented"),
```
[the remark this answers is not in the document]
nlewycky: Done.

lib/llvm-backend/src/code.rs
```diff
@@ -2268,6 +3884,413 @@ impl FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator {
...
+            Operator::I8x16AllTrue
+            | Operator::I16x8AllTrue
+            | Operator::I32x4AllTrue
+            | Operator::I64x2AllTrue => {
+                let vec_ty = match *op {
+                    Operator::I8x16AllTrue => intrinsics.i8x16_ty,
+                    Operator::I16x8AllTrue => intrinsics.i16x8_ty,
+                    Operator::I32x4AllTrue => intrinsics.i32x4_ty,
+                    Operator::I64x2AllTrue => intrinsics.i64x2_ty,
+                    _ => panic!(),
```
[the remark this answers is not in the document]
nlewycky: Used `unreachable!`.

lib/llvm-backend/src/code.rs
```diff
@@ -1071,18 +1315,262 @@ impl FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator {
...
+            Operator::I8x16AddSaturateS => {
+                let (v1, v2) = state.pop2()?;
+                let (v1, v2) = (v1.into_int_value(), v2.into_int_value());
+                let v1 = builder.build_bitcast(v1, intrinsics.i8x16_ty, "");
+                let v2 = builder.build_bitcast(v2, intrinsics.i8x16_ty, "");
+                let res = builder
+                    .build_call(intrinsics.sadd_sat_i8x16, &[v1, v2], "")
+                    .try_as_basic_value()
+                    .left()
+                    .unwrap();
```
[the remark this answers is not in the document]
nlewycky: There are zero calls to `expect()` in this file, but also this unwrap is guaranteed to succeed on the grounds that `intrinsics.sadd_sat_i8x16` is a function that doesn't return void. A call to a void returning function is the only reason we wouldn't get a basic_value back from a call.

lib/llvm-backend/src/code.rs
```diff
@@ -57,9 +60,165 @@ fn type_to_llvm(intrinsics: &Intrinsics, ty: Type) -> BasicTypeEnum {
...
+                .const_int(upper_bound, is_signed),
+            fvec_ty.get_element_type().into_float_type(),
+            "",
+        )
+    } else {
+        builder.build_unsigned_int_to_float(
+            ivec_ty
+                .get_element_type()
+                .into_int_type()
+                .const_int(upper_bound, is_signed),
```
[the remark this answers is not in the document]

lib/llvm-backend/src/code.rs
```diff
@@ -57,9 +60,165 @@ fn type_to_llvm(intrinsics: &Intrinsics, ty: Type) -> BasicTypeEnum {
...
+                .const_int(lower_bound, is_signed),
+            fvec_ty.get_element_type().into_float_type(),
+            "",
+        )
+    } else {
+        builder.build_unsigned_int_to_float(
+            ivec_ty
+                .get_element_type()
+                .into_int_type()
+                .const_int(lower_bound, is_signed),
```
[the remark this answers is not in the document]

lib/clif-backend/src/module.rs
```diff
@@ -107,6 +107,7 @@ impl From<Converter<ir::Type>> for Type {
             ir::types::I64 => Type::I64,
             ir::types::F32 => Type::F32,
             ir::types::F64 => Type::F64,
+            ir::types::I32X4 => Type::V128,
             _ => panic!("unsupported wasm type"),
```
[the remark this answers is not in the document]
nlewycky: Done.

Cargo.toml
```diff
@@ -61,7 +61,7 @@ members = [
 ]
 
 [build-dependencies]
-wabt = "0.7.2"
+wabt = { git = "https://github.com/wasmerio/wabt-rs.git", rev = "17188863021f0ed7721a1969da510efbc17615f2" }
```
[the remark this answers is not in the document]
nlewycky: https://github.com/pepyakin/wabt-rs/pull/47

syrusakbary:
bors try

syrusakbary:
bors try

syrusakbary:
bors try

lib/llvm-backend/src/code.rs
```diff
@@ -1071,18 +1300,262 @@ impl FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator {
...
+                    .try_as_basic_value()
+                    .left()
+                    .unwrap();
+                let res = builder.build_bitcast(res, intrinsics.i128_ty, "");
+                state.push1(res);
+            }
+            Operator::I16x8AddSaturateS => {
+                let (v1, v2) = state.pop2()?;
+                let (v1, v2) = (v1.into_int_value(), v2.into_int_value());
+                let v1 = builder.build_bitcast(v1, intrinsics.i16x8_ty, "");
```
bjfish: Should we name all the `""` name parameters all over? I thought they may be outputted to ll?

nlewycky: Depends on LLVM. A debug LLVM preserves instruction names and a release LLVM discards them (the symbol table for instruction names can be rather slow!). Each opcode implementation should copy the `state.var_name()` to one of the LLVM instructions, and that's usually the same point where we go from naming the rust variable an input value to a result. I think I put names on any instruction that wasn't obvious.

nlewycky: @bjfish I noticed that there were some opcodes which weren't using the `state.var_name()` at all. That should be fixed, please point it out if it isn't!

nlewycky:
bors try

lib/clif-backend/Cargo.toml
```diff
@@ -9,14 +9,14 @@ edition = "2018"
 
 [dependencies]
 wasmer-runtime-core = { path = "../runtime-core", version = "0.5.7" }
-cranelift-native = { version = "0.31" }
-cranelift-codegen = { version = "0.31" }
-cranelift-entity = { version = "0.31" }
-cranelift-frontend = { package = "wasmer-clif-fork-frontend", version = "0.31" }
-cranelift-wasm = { package = "wasmer-clif-fork-wasm", version = "0.31" }
+cranelift-native = { git = "https://github.com/wasmerio/cranelift.git", rev = "060c5a8e3553540144ca08ca028972c57693fa6e" }
```
syrusakbary: Can we update the versions of cranelift? (meaning: publishing our forks and use them here).

src/bin/wasmer.rs
```diff
@@ -316,7 +316,9 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
     };
 
     if !utils::is_wasm_binary(&wasm_binary) {
-        wasm_binary = wabt::wat2wasm(wasm_binary)
+        let mut features = wabt::Features::new();
+        features.enable_simd();
```
syrusakbary: I think it will better to only enable SIM when the indicated backend is LLVM
Thoughts?

syrusakbary:
```suggestion
        if options.backend == Backend::LLVM {
            features.enable_simd();
        }
```

nlewycky: Or follow the trend set by other implementations and require "--enable-simd" because it's a pre-standard feature?

MarkMcCaskey: It would be nice to standardize on an env var for this 🤔 

Maybe worth posting this to the https://github.com/webassembly/design repo

syrusakbary: > Or follow the trend set by other implementations and require "--enable-simd" because it's a pre-standard feature?

Having it behind a flag in Google Chrome makes sense because the implications of SIMD and because of their immense usage (so they need to be 1000% sure of everything working before enabling it).

However in our case I don't see any reason for not enabling it by default for LLVM (since our usage is still small and we can afford to iterate on it while shipping something that works for most of the cases).

nlewycky: I missed this conversation but posted it as a question on: https://github.com/WebAssembly/simd/issues/87

nlewycky:
bors try

lib/clif-backend/Cargo.toml
nlewycky: Done.

PR #543, opened 2019-07-10
update version numbers to 0.5.5

MarkMcCaskey:


MarkMcCaskey:
bors try

hrydgard:
Hi,

For the last few versions, the wasmer-runtime crate seems to have been left behind at 0.4.2 on crates.io (https://crates.io/crates/wasmer-runtime)  while the wasmer-runtime-core crate has been correctly updated (https://crates.io/crates/wasmer-runtime-core). Is there any reason for this?

Thanks!
Henrik

MarkMcCaskey:
> Hi,
> 
> For the last few versions, the wasmer-runtime crate seems to have been left behind at 0.4.2 on crates.io (https://crates.io/crates/wasmer-runtime) while the wasmer-runtime-core crate has been correctly updated (https://crates.io/crates/wasmer-runtime-core). Is there any reason for this?
> 
> Thanks!
> Henrik

Hi Henrik!

Thanks for the feedback and sorry about the inconvenience!  Actually the purpose of this PR is to finally ship a fix.  Due to depending on a git repository for some minor changes to Cranelift, we weren't able to publish some crates to crates.io.  We thought this would be a short-term thing but it hasn't turned out to be, so I'm in the process of setting things up so that we can publish our forks and then actually publish our crates.

Sorry about the lack of communication on this issue.  We're still figuring out our release process right now and will try to ensure something like this doesn't happen again!

MarkMcCaskey:
bors try

PR #544, opened 2019-07-10
Use bitcast instead of alloca+store+ptrcast+load sequence.

nlewycky:


nlewycky:
bors try

lib/llvm-backend/src/code.rs
```diff
@@ -78,22 +78,16 @@ fn trap_if_not_representatable_as_int(
...
 
     let (exponent, invalid_exponent) = {
-        let float_bits = {
-            let space = builder.build_alloca(int_ty, "space");
-            let float_ptr = builder.build_pointer_cast(space, float_ptr_ty, "float_ptr");
-            builder.build_store(float_ptr, value);
-            builder.build_load(space, "float_bits").into_int_value()
-        };
-
+        let float_bits = builder.build_bitcast(value, int_ty, "").into_int_value();
```
bjfish: Could name be something more descriptive than “” ? Might help with debugging later?

PR #555, opened 2019-07-12
wasi filesystem rewrite + implementation of many syscalls

MarkMcCaskey:
This lets the go compiler print its help text...

The rest of symlink code to follow...

This PR also does a major refactor of the file system code.  So we'll have to add more tests to be sure we didn't break anything


MarkMcCaskey:
bors try

lib/wasi-tests/wasitests/readlink.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    let mut some_contents = std::fs::File::open(sym_link_path).unwrap();
+
+    let mut buffer = [0; 128];
+
+    assert_eq!(some_contents.read(&mut buffer).unwrap(), 128);
+    let str_val = std::str::from_utf8(&buffer[..]).unwrap();
+    println!("{}", str_val);
+
+    // return to the current directory
```
syrusakbary: Curious why this is needed.

lib/wasi/src/state.rs
```diff
@@ -3,24 +3,30 @@
...
     fs,
     io::{self, Read, Seek, Write},
-    path::PathBuf,
+    path::{Path, PathBuf},
     time::SystemTime,
 };
 use wasmer_runtime_core::debug;
 
-pub const MAX_SYMLINKS: usize = 100;
+pub const MAX_SYMLINKS: u32 = 128;
```
syrusakbary: Is this number arbitrary? (or based in other impl?)

MarkMcCaskey: it's completely arbitrary, which I can explain in a comment.

It's just there to prevent infinite loops.  It seemed like a "big enough" number

lib/wasi-tests/wasitests/readlink.rs
syrusakbary: Same ^

MarkMcCaskey: I added it without checking if it was actually needed, it probably isn't

just the compulsion to close my open braces, I'll see if I can remove it

PR #557, opened 2019-07-13
Update loader-kernel feature name, fix compilation track state

bjfish:
- feature name `loader:kernel` -> `loader-kernel` (no colons)
- add track state `false` to loader kernel default, @losfair let me know if this default is incorrect
- cargo fmt


bjfish:
bors try

bjfish:
bors try

losfair:
Yeah setting track state to false for kernel loader is okay. The kernel loader does not use the state information.

Issue #558, opened 2019-07-14
'wasmer run' crashes when the .wasm file has no functions


nlewycky:
Thanks for the bug report!

### Describe the bug

CraneliftModuleCodeGenerator::signatures is an Option<> type. This type is only filled in when feed_signatures is called. If the module has no functions, feed_signatures is never called. Then, in finalize we have this code:
```rust
        let (func_resolver, backend_cache) = func_resolver_builder.finalize(
            &self.signatures.as_ref().unwrap(),
            Arc::clone(&trampolines),
            handler_data.clone(),
        )?;
```
which is attempting to unwrap a None.

### Steps to reproduce

```bash
echo '(module)' > empty.wat
wat2wasm empty.wat
wasmer run empty.wasm
```

The error is 
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:373:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```
with a stack trace of:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:373:21
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:47
   3: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:36
   4: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:200
   5: std::panicking::default_hook
             at src/libstd/panicking.rs:214
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:477
   7: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:384
   8: rust_begin_unwind
             at src/libstd/panicking.rs:311
   9: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  10: core::panicking::panic
             at src/libcore/panicking.rs:49
  11: core::option::Option<T>::unwrap
             at /rustc/0b680cfce544ff9a59d720020e397c4bf3346650/src/libcore/macros.rs:12
[...]

PR #559, opened 2019-07-14
Initial commit of a fuzzer. Run with "cargo fuzz run simple_instantiate".

nlewycky:
Used to discover issue #558.

We'll probably want to reconsider the default .gitignore of the artifacts and corpus directories. The fuzzer wastes a lot of time not having even a single exampel of a valid .wasm file to start with.


nlewycky:
> It would be nice if here was a `fuzz/README.md` that showed the command used to run the fuzzer.

Great point. Done!

PR #560, opened 2019-07-15
Ctx call finalizer

YaronWittenstein:
It seems that the `data_finalizer` of the`Ctx` is never being called upon instance destruction.

It seems a bit tricky to assert that in though

@syrusakbary @Hywan


YaronWittenstein:
I'll close this PR and re-open it in a more tidy manner

PR #561, opened 2019-07-15
* vm.rs - calling `Ctx` `data_finalizer` upon destruction

YaronWittenstein:
It seems that the `data_finalizer` field of the `Ctx` is never being called upon instance destruction.

It seems a bit tricky to assert that in though

@syrusakbary @Hywan


syrusakbary:
Ummm... tests are broken on windows, so we have to take a deeper look.
Will trigger bors to see the results on macos/linux.

syrusakbary:
bors try

syrusakbary:
(still investigating)

syrusakbary:
bors try

MarkMcCaskey:
Hey @YaronWittenstein I found the issue on our side!  We are breaking the contract on `std::mem::uninitialized()` and calling Drop on garbage memory interpreted as a Ctx.

(The reason is that we do something like
```
let ctx = Box::new(std::mem::uninit());
something_else.val = Box::into_raw(ctx);
something_else.val = real_ctx; // this is invalid because it causes the uninit Ctx to Drop
```

And the solution is to change the last line to

```
let mut real_ctx = real_ctx;
std::mem::swap(&mut real_ctx, &mut something_else.val);
std::mem::forget(real_ctx); // tell Rust to not Drop our garbage Ctx
```

I have a fix locally that I need to clean up -- and actually I need to port it to the new MaybeUninit style for Rust 1.38 -- but your PR was passing all the tests I tried on my computer last night.

Thanks!  I'll ping you again when its merged into master

MarkMcCaskey:
bors try

PR #563, opened 2019-07-15
improve wasitest infra and fix fseek test

MarkMcCaskey:
This PR:
- fixes arg parsing from comments
- fixes the mapdir test to have the native code doing the same thing
- makes `wasitests-generate` output stdout/stderr by default
- adds function to print stdout and stderr for a command if it fails
- compiles wasm with size optimizations
- strips generated wasm with `wasm-strip`

This makes wasm module sizes go down from 1.7, 1.9Mb to ~60kb, 80kb.  Preventing git bloat and reducing likely bottleneck of reading wasm modules from disk


MarkMcCaskey:
bors try

syrusakbary:
Merging manually, as tests passed :)

PR #567, opened 2019-07-17
State preserving for LLVM, and Singlepass -> LLVM OSR-based tier switching.

losfair:


lib/llvm-backend/src/code.rs
```diff
@@ -299,11 +301,67 @@ fn resolve_memory_ptr(
...
+    builder: &Builder,
+    local_function_id: usize,
+    target: &mut StackmapRegistry,
+    kind: StackmapEntryKind,
+    locals: &[PointerValue],
+    state: &State,
+) {
+    let stackmap_id = target.entries.len();
+
+    let mut params = vec![];
```
bjfish: Is capacity know here? `Vec::with_capacity`

lib/runtime-core/src/fault.rs
```diff
@@ -40,6 +40,7 @@ struct UnwindInfo {
 
 thread_local! {
     static UNWIND: UnsafeCell<Option<UnwindInfo>> = UnsafeCell::new(None);
+    static CURRENT_CTX: UnsafeCell<*mut vm::Ctx> = UnsafeCell::new(::std::ptr::null_mut());
```
nlewycky: I don't presently understand what it means for there to ever be more than one context?

lib/runtime-core/src/state.rs
```diff
@@ -634,6 +694,7 @@ pub mod x64 {
 
         catch_unsafe_unwind(
             || {
+                ::std::intrinsics::breakpoint();
```
nlewycky: Wait-- is this leftovers from earlier debugging or is this intended to `longjmp` to some other code? If so, where?

lib/runtime-core/src/state.rs
```diff
@@ -854,5 +946,36 @@ pub mod x64 {
...
+                7 => X64Register::GPR(GPR::RSP),
+                8 => X64Register::GPR(GPR::R8),
+                9 => X64Register::GPR(GPR::R9),
+                10 => X64Register::GPR(GPR::R10),
+                11 => X64Register::GPR(GPR::R11),
+                12 => X64Register::GPR(GPR::R12),
+                13 => X64Register::GPR(GPR::R13),
+                14 => X64Register::GPR(GPR::R14),
+                15 => X64Register::GPR(GPR::R15),
+
```
nlewycky: `16 => { unimplemented!("EIP"); }`? Or `unreachable`?

lib/runtime-core/src/state.rs
```diff
@@ -854,5 +946,36 @@ pub mod x64 {
...
+                17 => X64Register::XMM(XMM::XMM0),
+                18 => X64Register::XMM(XMM::XMM1),
+                19 => X64Register::XMM(XMM::XMM2),
+                20 => X64Register::XMM(XMM::XMM3),
+                21 => X64Register::XMM(XMM::XMM4),
+                22 => X64Register::XMM(XMM::XMM5),
+                23 => X64Register::XMM(XMM::XMM6),
+                24 => X64Register::XMM(XMM::XMM7),
+                _ => return None,
```
nlewycky: This applies to LLVM generated code too, right? Any reason to believe LLVM will never use XMM8--XMM31? YMM? ZMM/K (AVX-512)? I'm willing to say that we will never use x87 floating point stack or MMX registers.

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -1335,6 +1338,7 @@ impl X64FunctionCode {
             fsm.call_offsets.insert(
                 offset,
                 OffsetInfo {
+                    end_offset: offset + 1,
```
nlewycky: Is this a bug-fix? If so, could you pull it out into its own PR for easier review both of its own change and the OSR patch?

lib/runtime-core/src/state.rs
```diff
@@ -524,6 +583,7 @@ pub mod x64 {
 
             stack_offset -= 1;
             stack[stack_offset] = (code_base + activate_offset) as u64; // return address
+            println!("activating at {:?}", (code_base + activate_offset) as *const u8);
```
nlewycky: Debugging leftovers?

lib/llvm-backend/src/code.rs
```diff
@@ -17,14 +19,15 @@ use wasmer_runtime_core::{
     module::{ModuleInfo, ModuleInner},
     structures::{Map, TypedIndex},
     types::{
-        FuncIndex, FuncSig, GlobalIndex, LocalOrImport, MemoryIndex, SigIndex, TableIndex, Type,
+        FuncIndex, FuncSig, GlobalIndex, LocalOrImport, MemoryIndex, SigIndex, TableIndex, Type, ImportedFuncIndex,
```
nlewycky: Keep it alphabetical please.

losfair: Yeah will let rustfmt handle that.

lib/runtime-core/src/state.rs
losfair: Good point. Will add XMM8-XMM31 in. Since all values tracked by the stack map are not larger than 64 bits (SIMD is an exception, but can fix it later), saving just XMM8-31 should be enough?

lib/runtime-core/src/state.rs
losfair: This PR is not finished yet :) yeah this is just for debugging.

lib/runtime-core/src/fault.rs
losfair: In case of nested WASM invocations:

Native -> WASM -> (import function) Native -> WASM

lib/llvm-backend/src/backend.rs
```diff
@@ -285,13 +298,109 @@ impl LLVMBackend {
...
+            // All local functions.
+            for index in module_info.imported_functions.len()..module_info.func_assoc.len() {
+                let name = if cfg!(target_os = "macos") {
+                    format!("_fn{}", index)
+                } else {
+                    format!("fn{}", index)
+                };
+
+                let c_str = CString::new(name).unwrap();
+                let ptr = unsafe { get_func_symbol(module, c_str.as_ptr()) };
```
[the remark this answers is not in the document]
losfair: There are other functions (trampolines, etc.) than the ones purely translated from WASM functions. To avoid depending too much on code generation internals I think using the names is a better approach for now.

lib/singlepass-backend/src/codegen_x64.rs
losfair: This is a required change for OSR to work for LLVM, since we cannot determine the exact address of the trapping instruction (IR is more high-level).

lib/runtime-core/src/state.rs
losfair: Covered by the `return None` case. Or do you think a separate assertion is better?

syrusakbary:
@losfair @nlewycky I believe this might be useful when reviewing this PR:
https://github.com/wasmerio/wasmer/pull/298/files#diff-5ab40264a7ce86c159bf06530f13c3e1R9

lib/llvm-backend/src/backend.rs
```diff
@@ -231,10 +238,17 @@ pub struct LLVMBackend {
     module: *mut LLVMModule,
     #[allow(dead_code)]
     buffer: Arc<Buffer>,
+    msm: Option<ModuleStateMap>,
```
bjfish: Style suggestion: I think a more descriptive name should be used for fields: `module_state_map`.

lib/llvm-backend/src/backend.rs
```diff
@@ -273,25 +287,136 @@ impl LLVMBackend {
...
         };
 
+        /*
+
         static SIGNAL_HANDLER_INSTALLED: Once = Once::new();
 
         SIGNAL_HANDLER_INSTALLED.call_once(|| unsafe {
             crate::platform::install_signal_handler();
-        });
+        });*/
```
bjfish: Clean up commented out code.

lib/llvm-backend/src/backend.rs
```diff
@@ -273,25 +287,136 @@ impl LLVMBackend {
...
+                    llvm_backend_get_code_ptr(module),
+                    llvm_backend_get_code_size(module),
+                )
+            };
+            let mut msm = ModuleStateMap {
+                local_functions: Default::default(),
+                total_size: code_size,
+            };
+
+            let mut local_func_id_to_addr: Vec<usize> = Vec::new();
```
bjfish: Consider `Vec::with_capicity`

lib/llvm-backend/src/backend.rs
```diff
@@ -273,25 +287,136 @@ impl LLVMBackend {
...
+                        module_info,
+                        code_ptr as usize,
+                        &map,
+                        size_record,
+                        map_record,
+                        end_record.map(|x| (end_entry, *x)),
+                        &mut msm,
+                    );
+                } else {
+                    // TODO: optimized out?
```
bjfish: Please handle this branch.

losfair: If the map record is optimized out then doing nothing here is the correct behavior.

lib/llvm-backend/src/backend.rs
losfair: I think using `MSM = ModuleStateMap` and `FSM = FunctionStateMap` is fine if we keep it consistent everywhere, since it's quite common and `module_state_map` is too long to type...

lib/runtime-core/src/state.rs
```diff
@@ -682,31 +758,73 @@ pub mod x64 {
...
         mut stack: *const u64,
         initially_known_registers: [Option<u64>; 24],
         mut initial_address: Option<u64>,
     ) -> ExecutionStateImage {
         let mut known_registers: [Option<u64>; 24] = initially_known_registers;
         let mut results: Vec<WasmFunctionStateDump> = vec![];
+        let mut was_baseline = true;
 
-        for _ in 0.. {
+        for i in 0.. {
```
bjfish: `warning: unused variable: `i``

bjfish:
Please remove the `lib/runtime-c-api/tests/a.out` file.

bjfish:
bors try

lib/llvm-backend/src/backend.rs
```diff
@@ -204,25 +218,140 @@ impl LLVMBackend {
...
+                Self {
+                    module,
+                    buffer: Arc::clone(&buffer),
+                    msm: Some(msm),
+                    local_func_id_to_offset,
+                },
+                LLVMCache { buffer },
+            )
+        } else {
+            eprintln!("WARNING: No stack map");
```
bjfish: I  believe this warning could be removed. This prints multiple times in the spectests currently.

lib/runtime-core/src/state.rs
nlewycky: If it works then it's fine for now, but I can't promise that LLVM won't pack four 64-bit values together in a YMM register for example. It could if it wanted to, it just doesn't seem to want to.

What goes wrong if LLVM uses registers that aren't listed here? a) Will it cause problems at runtime or do we just realize we can't tier-switch at this point and ignore it? b) Is there a way I can test whether other registers are used, starting with a given Wasm file?

lib/runtime-core/src/tiering.rs
```diff
@@ -0,0 +1,236 @@
...
+    ctx: &Mutex<CtxWrapper>,
+    state: &OptimizationState,
+) {
+    let module = match compile_with_config(
+        &binary[..],
+        &*compiler,
+        CompilerConfig {
+            symbol_map: None,
+            track_state: true,
+            ..Default::default()
```
nlewycky: This doesn't work with the `--enable-simd` and `--enable-threads` flags, then? Maybe the `compiler` variable has a field with its CompilerConfig and you can use the fields from that?

lib/runtime-core/src/tiering.rs
```diff
@@ -0,0 +1,236 @@
...
+    module_info: &ModuleInfo,
+    wasm_binary: &[u8],
+    mut resume_image: Option<InstanceImage>,
+    import_object: &ImportObject,
+    start_raw: extern "C" fn(&mut Ctx),
+    baseline: &mut Instance,
+    optimized_backends: Vec<Box<dyn Fn() -> Box<dyn Compiler> + Send>>,
+    interactive_shell: F,
+) -> Result<(), String> {
+    unsafe {
```
nlewycky: Maybe `pub unsafe fn` instead?

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -418,7 +461,10 @@ impl ModuleCodeGenerator<X64FunctionCode, X64ExecutionContext, CodegenError>
         };
 
         let total_size = assembler.get_offset().0;
-        let output = assembler.finalize().unwrap();
+        let _output = assembler.finalize().unwrap();
```
nlewycky: The `_` prefix usually means unused, but this is used? Could you rename it? Actually, I expected that `output` would work here? You can `let` the same name twice in rust, using the first one in the initializer for the second.

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -1630,6 +1687,14 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
 
     fn begin_body(&mut self, _module_info: &ModuleInfo) -> Result<(), CodegenError> {
         let a = self.assembler.as_mut().unwrap();
+        let start_label = a.get_label();
+        // skip the patchpoint during normal execution
+        a.emit_jmp(Condition::None, start_label);
+        // patchpoint of 32 bytes
```
nlewycky: Proposed text: " // Patchpoint of 32 1-byte nops."

Nops come in different lengths, fewer longer nops is faster in-place, but when replacing you have to replace whole instruction sequences at a time.

lib/runtime-core/src/state.rs
```diff
@@ -28,22 +30,26 @@ pub struct MachineStateDiff {
...
 pub enum MachineValue {
     Undefined,
     Vmctx,
+    VmctxDeref(Vec<usize>),
     PreserveRegister(RegisterIndex),
     CopyStackBPRelative(i32), // relative to Base Pointer, in byte offset
     ExplicitShadow,           // indicates that all values above this are above the shadow region
     WasmStack(usize),
     WasmLocal(usize),
+    TwoHalves(Box<(MachineValue, MachineValue)>), // 32-bit values. TODO: optimize: add another type for inner "half" value to avoid boxing?
```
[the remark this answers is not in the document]
losfair: All top-level `MachineValue`'s are 64-bit. `TwoHalves` divides the value into two 32-bit values.

If in the future we find 16-bit and 8-bit values in the record, then we can just represent it with recursive `TwoHalves`.

lib/llvm-backend/src/backend.rs
```diff
@@ -204,25 +218,140 @@ impl LLVMBackend {
...
+
+            for r in &map.stk_map_records {
+                map_records.insert(r.patchpoint_id as usize, r);
+            }
+
+            for ((start_id, start_entry), (end_id, end_entry)) in stackmaps
+                .entries
+                .iter()
+                .enumerate()
+                .step_by(2)
```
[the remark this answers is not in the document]
losfair: Just to pack the entries by 2 each: (0, 1), (2, 3) ...

Is there any better ways for that?

lib/llvm-backend/src/backend.rs
```diff
@@ -204,25 +218,140 @@ impl LLVMBackend {
...
+                .zip(stackmaps.entries.iter().enumerate().skip(1).step_by(2))
```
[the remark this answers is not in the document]
losfair: See above

lib/singlepass-backend/src/codegen_x64.rs
losfair: `CodeMemory` isn't supposed to receive a type from dynasm. So there needs to be another statement to copy data from `_output` to `output`.

lib/runtime-core/src/state.rs
losfair: If the LLVM values "semantically" are at most 64 bits, then the equivalent runtime locations are unlikely to be larger than 64 bits - I don't think there's a way to represent in stackmap a value with different sizes between LLVM value and runtime location. Also, with the presence of stackmap some LLVM optimizations won't apply.

losfair: The compiler will panic if this function returns None, indicating a bug.

lib/runtime-core/src/tiering.rs
losfair: This is already checked in `wasmer.rs`.

lib/llvm-backend/src/backend.rs
```diff
@@ -300,9 +423,30 @@ impl RunnableModule for LLVMBackend {
...
+        })
+    }
+
+    fn get_local_function_offsets(&self) -> Option<Vec<usize>> {
+        Some(self.local_func_id_to_offset.clone())
+    }
+
+    fn get_module_state_map(&self) -> Option<ModuleStateMap> {
+        self.msm.clone()
+    }
```
MarkMcCaskey: I'd expect these to return references based on the name 🤔 though it's probably fine

lib/runtime-core/src/tiering.rs
```diff
@@ -0,0 +1,233 @@
...
+    });
+
+    {
+        let wasm_binary = wasm_binary.to_vec();
+        let ctx_box = ctx_box.clone();
+        let opt_state = opt_state.clone();
+        ::std::thread::spawn(move || {
+            for backend in optimized_backends {
+                if !ctx_box.lock().unwrap().0.is_null() {
+                    do_optimize(&wasm_binary, backend(), &ctx_box, &opt_state);
```
MarkMcCaskey: a ReadWriteLock might have better performance here than a Mutex

nlewycky:
AppVeyor on Windows:
```
error: unused import: `collections::BTreeMap`
  --> lib\llvm-backend\src\backend.rs:13:5
   |
13 |     collections::BTreeMap,
   |     ^^^^^^^^^^^^^^^^^^^^^
   |
note: lint level defined here
  --> lib\llvm-backend\src\lib.rs:4:5
   |
4  |     unused_imports,
   |     ^^^^^^^^^^^^^^
error: unused imports: `BTreeMap`, `HashMap`
 --> lib\llvm-backend\src\stackmap.rs:4:24
  |
4 | use std::collections::{BTreeMap, HashMap};
  |                        ^^^^^^^^  ^^^^^^^
error: unused import: `vm`
  --> lib\llvm-backend\src\stackmap.rs:11:5
   |
11 |     vm,
   |     ^^
```

Please just make the relevant imports conditional with the same conditions that use them. I know.

I'm not sure about circleci-lint? @MarkMcCaskey maybe?

syrusakbary:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
The failing `Test linux` test passes for me locally. Manually merging.

PR #576, opened 2019-07-24
fix Drop of uninit Ctx; use MaybeUninit

MarkMcCaskey:
I'm somewhat concerned that we need to call
`intrinsics::panic_if_uninhabited::<T>();` or something to tell the compiler that we know it's safe.

We just reinterpret the pointer later as initialized which is safe in theory due to `repr(transparent)`

this should unblock #561


MarkMcCaskey:
bors retry

MarkMcCaskey:
bors retry

PR #490, opened 2019-06-10
Add MiddlewareChain and StreamingCompiler to runtime

bjfish:
It would be nice to add `MiddlewareChain` and `StreamingCompiler` to runtime to be able to use these without having to include the `runtime-core` as a dependency. For example:
```rust
use wasmer_runtime::{MiddlewareChain, StreamingCompiler};
use wasmer_singlepass_backend::ModuleCodeGenerator as SinglePassMCG;
use wasmer_middleware_common::metering::Metering;
let limit = 1_000_000;
let c: StreamingCompiler<SinglePassMCG, _, _, _, _> = StreamingCompiler::new(move || {
    let mut chain = MiddlewareChain::new();
    chain.push(Metering::new(limit));
    chain
});            
wasmer_runtime::compile_with(&prepared_code, &c)
    .map_err(|e| Error::Wasmer(format!("{}", e)))
            chain
        });            
wasmer_runtime::compile_with(&prepared_code, &c)
    .map_err(|e| Error::Wasmer(format!("{}", e)))
```

PR #578, opened 2019-07-24
llvm-backend: export FunctionCodeGenerator and ModuleCodeGenerator

vavrusa:
This is in line with the singlepass-backend to be able to use LLVM MCG with
a StreamingCompiler (as in https://github.com/wasmerio/wasmer/pull/490).


syrusakbary:
Thanks for contributing! Merging :)

PR #580, opened 2019-07-25
Various improvements

syrusakbary:
* [x] Moved scripts to `scripts` folder
* [x] Improved Install documentation for different OS/platforms
* [x] Improved Wasmer Backends documentation


README.md
```diff
@@ -216,9 +229,8 @@ Below are some of the goals of this project (in order of priority):
 - [x] It should be 100% compatible with the [WebAssembly spec tests](https://github.com/wasmerio/wasmer/tree/master/lib/spectests/spectests)
 - [x] It should be fast _(partially achieved)_
 - [x] Support WASI - released in [0.3.0](https://github.com/wasmerio/wasmer/releases/tag/0.3.0)
-- [ ] Support Emscripten calls _(in the works)_
-- [ ] Support Rust ABI calls
```
piranna: What happened with Rust ABI calls?

MarkMcCaskey: I think Rust proudly does not have a stable ABI and won't have a stable ABI in the near-future.  This is a good thing for optimization and design without restriction.

I'm not sure what the reasoning behind it being there in the first place was though

piranna: Ok, I'm curious about why Rust would have an ABI at all being low level like C, so no runtime there should exist...

PR #584, opened 2019-07-26
Add --enable-simd flag to wasmer run and wasmer validate.

nlewycky:


lib/runtime-core/src/lib.rs
```diff
@@ -117,8 +117,26 @@ pub fn validate(wasm: &[u8]) -> bool {
...
     use wasmparser::WasmDecoder;
-    let mut parser = wasmparser::ValidatingParser::new(wasm, None);
+    let config = wasmparser::ValidatingParserConfig {
+        operator_config: wasmparser::OperatorValidatorConfig {
+            enable_simd: features.simd,
+            enable_bulk_memory: false,
+            enable_multi_value: false,
+            enable_reference_types: false,
+            enable_threads: false,
+        },
```
MarkMcCaskey: I like this being explicit, so don't change it, but here's a neat Rust trick you can do:
```
wasmparser::OperatorValidatorConfig {
  enable_simd: features.simd,
  ..wasmparser::OperatorValidatorConfig::default()
}
```

This is useful especially when there are a lot of fields that you don't care about (this code assumes OperatorValidatorConfig is Default and gives `false` for those values (which is what `derive(Default)` will do))

lib/runtime-core/src/backend.rs
```diff
@@ -110,6 +110,11 @@ impl Default for MemoryBoundCheckMode {
     }
 }
 
+#[derive(Default)]
```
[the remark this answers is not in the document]
nlewycky: Done.

lib/runtime-core/src/lib.rs
nlewycky: As it is the build will break when wasmparser changes their feature set. Do we want that?

MarkMcCaskey: Hmm, I mean we'll be updating wasmparser manually, so I think it's fine either way.

If you think that's something you might want to see, then keep it, otherwise you can use the syntax above

nlewycky: > error[E0599]: no function or associated item named `default` found for type `wasmparser::operators_validator::OperatorValidatorConfig` in the current scope

PR #585, opened 2019-07-26
add rust example to CI

MarkMcCaskey:


MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

.circleci/config.yml
```diff
@@ -129,6 +129,26 @@ jobs:
...
+          name: "Check Wasmer Rust example"
+          command: |
+            git clone https://github.com/wasmerio/wasmer-rust-example
+            rustup default stable
+            rustup target add wasm32-unknown-unknown
+            cd wasmer-rust-example
+            cd wasm-sample-app
+            cargo build --release
+            cd ..
+            sed -i 's/wasmer-runtime.*/wasmer-runtime = \{ path  = "..\/lib\/runtime" \}/g' Cargo.toml
```
syrusakbary: Since this is a bug with published crates, do you think this will trigger it? (using path instead of a pinned crates.io version)

MarkMcCaskey: Hmm, so this change will protect us from merging into master changes that break in the example.  In the publishing case: this will not catch it before we publish, but it will cause all builds to fail after

PR #586, opened 2019-07-26
Fixes case with empty module, handle case with no signatures

bjfish:
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:373:21


bjfish:
bors try

PR #587, opened 2019-07-27
Use --enable-simd to control whether SIMD is enabled in the wasmparser.

nlewycky:
Before this change, 'wasmer run --backend=llvm some-simd.wasm' would run without complaint.

Also, note that the flag is not part of the cache key, so after any successful run, we can run it again without passing the flag.

### DIFF
--- a/lib/runtime-core/src/codegen.rs
+++ b/lib/runtime-core/src/codegen.rs
@@ -1,6 +1,6 @@
 use crate::{
     backend::RunnableModule,
-    backend::{Backend, CacheGen, Compiler, CompilerConfig, Token},
+    backend::{Backend, CacheGen, Compiler, CompilerConfig, Features, Token},
     cache::{Artifact, Error as CacheError},
     error::{CompileError, CompileResult},
     module::{ModuleInfo, ModuleInner},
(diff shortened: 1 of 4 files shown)


lib/runtime-core/src/codegen.rs
```diff
@@ -137,22 +137,22 @@ impl<
...
+            enable_simd: features.simd,
             enable_bulk_memory: false,
             enable_multi_value: false,
         },
         mutable_global_imports: false,
     }
 }
 
-fn validate(bytes: &[u8]) -> CompileResult<()> {
+fn validate(bytes: &[u8], features: &Features) -> CompileResult<()> {
```
syrusakbary: Add a new method, rather than `validate` use `validate_with_features`, this way we will not break compatibility with the API.

nlewycky: I think this one isn't part of the exported API, the public one is `pub fn validate` in `lib.rs`? Should I avoid breaking compatibility here anyways?

syrusakbary: I think you are right on the API exposure (since validate was not public).
However I still see some utility into having two functions `validate` and `validate_with_feature`. Thoughts?

nlewycky: Renamed to `validate_with_features` and removed dead function `validate`.

PR #589, opened 2019-07-29
wasmer-c-api-changes: making `wasmer_byte_array` fields `public`

YaronWittenstein:
I've added this PR since I've tests that need access to `wasmer_byte_array` field.
Here is an example:
https://github.com/spacemeshos/svm/blob/52f04b52cc9a5227030f3dd3499e04c6cbe143d9/crates/svm-wasmer-c-api/tests/c_api.rs#L96

```rust
unsafe fn cast_wasmer_byte_array_to_string(wasmer_bytes: &wasmer_byte_array) -> String {
    let slice: &[u8] =
        std::slice::from_raw_parts(wasmer_bytes.bytes, wasmer_bytes.bytes_len as usize);

    if let Ok(s) = std::str::from_utf8(slice) {
        s.to_string()
    } else {
        panic!("error converting `wasmer_byte_array` to string")
    }
}
```

@syrusakbary @bjfish @Hywan


bjfish:
This looks good to me.
bors r+

PR #590, opened 2019-07-29
wasmer-c-api: error visibility changes 

YaronWittenstein:
I've changed the visibility to `public` in order to support code like here:
https://github.com/spacemeshos/svm/blob/52f04b52cc/crates/svm-wasmer-c-api/src/macros.rs#L150

@syrusakbary @bjfish @Hywan


bjfish:
LGTM, thanks!
bors r+

PR #593, opened 2019-07-29
Add a Docker sandbox for building Wasmer

penberg:
Building Wasmer is bit of a pain, because it requires Rust nightly, and
bunch of OS specific packages.

To make building easier, this adds a "build" script at the top-level
directory, which can be used to build Wasmer within a Docker sandbox
that has all the necessary dependencies installed. The build environment
is based on latest Ubuntu 19.04.

You first need to build a Docker image of the sandbox:

    docker build --file Dockerfile.build --tag wasmer-build .

Then, to build Wasmer, run:

    ./build make

To test Wasmer, run:

    ./build make test

and so on.

You can also drop into a shell within the Docker with:

    ./build

The "build" script bind mounts current directory as "/wasmer" in the
Docker container, which allows inspecting the build contents like you
had built them on your local machine.

For future improvements, we should consider:

  - Consolidation with existing Dockerfile (that is used for Circle CI)

  - Publishing the build sandbox image on Docker Hub so that people
    don't have to build it themselves

  - Moving dependency installation to separate script, which can be
    reused outside of the Docker sandbox.

The work has been inspired by "devtool" in the Firecracker project:

  https://github.com/firecracker-microvm/firecracker/blob/master/tools/devtool

and "dbuild" in the Scylla project:

  https://github.com/scylladb/scylla/blob/master/tools/toolchain/dbuild

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,0 +1,13 @@
+FROM ubuntu:19.04
+
+ARG RUST_TOOLCHAIN="nightly"
+
+ENV CARGO_HOME=/usr/local/rust
+ENV RUSTUP_HOME=/usr/local/rust
+ENV PATH="$PATH:$CARGO_HOME/bin"
+
+RUN apt-get update \
+    && apt-get -y install sudo strace curl cmake pkg-config python libssl-dev llvm-dev libz-dev gnuplot-nox \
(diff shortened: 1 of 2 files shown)


syrusakbary:
This is awesome! Merging 🎉

PR #594, opened 2019-07-30
Improvement cache in Windows

syrusakbary:
Caching was disabled on Windows, but can be re-enabled easily by improving the folder cache naming.

Reason why caching was disabled on Windows: We use a very long string (64 chars) for the wasmer version (hash). But we can use the version directly (no need to hashing)

### DIFF
--- a/lib/runtime-core/src/cache.rs
+++ b/lib/runtime-core/src/cache.rs
@@ -241,5 +241,3 @@ pub trait Cache {
 /// A unique ID generated from the version of Wasmer for use with cache versioning
 pub const WASMER_VERSION_HASH: &'static str =
     include_str!(concat!(env!("OUT_DIR"), "/wasmer_version_hash.txt"));
-
-pub const WASMER_VERSION: &'static str = env!("CARGO_PKG_VERSION");
--- a/lib/runtime-core/src/cache.rs
+++ b/lib/runtime-core/src/cache.rs
@@ -241,3 +241,5 @@ pub trait Cache {
 /// A unique ID generated from the version of Wasmer for use with cache versioning
 pub const WASMER_VERSION_HASH: &'static str =
     include_str!(concat!(env!("OUT_DIR"), "/wasmer_version_hash.txt"));
+
+pub const WASMER_VERSION: &'static str = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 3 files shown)


lib/runtime-core/src/cache.rs
```diff
@@ -241,3 +241,5 @@ pub trait Cache {
 /// A unique ID generated from the version of Wasmer for use with cache versioning
 pub const WASMER_VERSION_HASH: &'static str =
     include_str!(concat!(env!("OUT_DIR"), "/wasmer_version_hash.txt"));
+
+pub const WASMER_VERSION: &'static str = env!("CARGO_PKG_VERSION");
```
MarkMcCaskey: I think we're already reading this somewhere else in runtime-core `VERSION` in lib.rs;

We should reuse the constant, or define it in a shared `constants.rs` file and use it for both

lib/runtime/src/cache.rs
```diff
@@ -9,7 +9,7 @@ use std::{
 use wasmer_runtime_core::cache::Error as CacheError;
 pub use wasmer_runtime_core::{
     backend::Backend,
-    cache::{Artifact, Cache, WasmHash, WASMER_VERSION_HASH},
+    cache::{Artifact, Cache, WasmHash, WASMER_VERSION},
```
MarkMcCaskey: If it's just for use here, `runtime-core` already exports this as `VERSION`

lib/runtime/src/cache.rs
```diff
@@ -49,12 +49,7 @@ impl FileSystemCache {
     /// This method is unsafe because there's no way to ensure the artifacts
     /// stored in this cache haven't been corrupted or tampered with.
     pub unsafe fn new<P: Into<PathBuf>>(path: P) -> io::Result<Self> {
-        let path: PathBuf = {
-            let mut path = path.into();
-            path.push(WASMER_VERSION_HASH);
-            path
-        };
-
+        let path: PathBuf = path.into();
```
MarkMcCaskey: Did you mean to push `WASMER_VERSION` here?

MarkMcCaskey: If not, the import above isn't used

MarkMcCaskey:
Also the reasoning for caching being disabled on Windows seems strange:

I added the code to hash things under the guidance of Lachlan, but I recall caching on Windows being disabled before I joined and before we started versioning the cache.

Issue #583, opened 2019-07-25
Embedding API for wasmer-wasi


kitlith:
### Motivation

Currently, wasmer-wasi exposes an opaque interface to the host program. You can pass in program arguments, environment variables, and some preopened directories, and that's about it. However, host programs may want features such as the ability to redirect stdout/stderr to log files, fake a file in-memory, or the ability to hand file-descriptors directly to the program/plugin running under wasmer.

### Proposed solution

Add some APIs that host programs can use: (these are just some that I've thought of)
 - Added in #595:
   - [x] Add the ability to redirect stdin/stdout (perhaps to types implementing Read/Write?)
   - [x] Add the ability to open a file and get a file descriptor which can be passed to something running under wasmer.
     - ~~This may currently be possible thorugh the WasiFs Type, except it's private, etc.~~
   - [x] Perhaps instead of just files, types implementing Read/Write/Seek?
 - Others:
   - [ ] Custom directories?

### Alternatives

 - Implement WASI myself. I'd prefer to avoid this possibility, as there's already a perfectly good implementation here.
 - Fork wasmer-wasi. I'd rather discuss/get something in upstream if possible.

syrusakbary:
@kitlith I love this idea!

I would like to work together on this and help you get something upstream. Could be possible to get a bit more insight on how you are thinking on using the APIs (via some pseudocode)

MarkMcCaskey:
Thanks for the feedback!  This sounds like a great idea!

I'll post here soon with some more ideas;
as for anything implementing Read/Write/Seek: WASI has the notion of the Buffer file type which seems like a good fit.  I currently haven't implemented a lot of the logic for Buffer because I wasn't sure what exactly it would be, but this seems like a good place to start.

The current API also wasn't really designed with outside use of WASI in mind, either, I'd like to change some of the function signatures, but I'll have to wait until the next major release to do so!

kitlith:
How much would you have to change? Only two functions and one type are currently pub, so as long as those stay the same and only new stuff is added, it could be a minor release, right?

One thing I realized is that FDs are not limited to things that implement Read or Write or Seek -- because directories can do none of these things. I dunno how relevant this is to WASI, but it could be possible that someone wants to create in-memory directories. Maybe we have a trait WasiFdBacking that has default implementations for std::fs::File, some provided type representing a directory, etc. and can be easily implemented in terms of read/write/seek (automatically?) if present.

[...]

PR #595, opened 2019-07-30
Add useful functions for external use of WASI filesystem

MarkMcCaskey:
part of #583


lib/wasi/src/state.rs
```diff
@@ -20,75 +20,297 @@ use wasmer_runtime_core::debug;
...
+            _ => WasiFsError::UnknownError(err),
+        }
+    }
 }
 
-impl WasiFile {
-    pub fn close(self) {}
+/// This trait relies on your file closing when it goes out of scope via `Drop`
+/// it does not require `Drop` because `std::fs::File` does not implement it
+/// (TODO: figure out how that makes sense, because it clearly does)
```
kitlith: std::fs::File does not implement Drop, because it contains a std::sys::fs::File, which does not implement Drop as it contains a std::sys::fd::FileDesc, which implements Drop.

aka, only the innermost item needs to implement Drop, rust figures out the rest on it's own?

MarkMcCaskey: Thanks!  I was reading through the code, but didn't get that far -- that's good to know

lib/wasi/src/state.rs
```diff
@@ -20,75 +20,297 @@ use wasmer_runtime_core::debug;
...
     }
 
+    fn size(&self) -> u64 {
+        self.metadata().unwrap().len()
+    }
+}
+
+#[derive(Debug)]
+pub struct Stdout(std::io::Stdout);
+impl Read for Stdout {
```
kitlith: I feel like a 'nicer' way of doing this is to essentially encapsulate Read/Write/Seek in our trait, and provide default generic implementations if the type implements Read/Write/Seek.

(this is probably *not* nicer in many other ways, so meh.)

lib/wasi/src/state.rs
kitlith: so, yeah, apparently we should (almost?) never require an impl of Drop, because of the above.

kitlith:
I will note that currently this PR *by itself* does not expose any extra functionality, as WasiFs/WasiState and such is still not exported at the crate top-level.

lib/wasi/src/state.rs
MarkMcCaskey: Hmm, yeah, I see what you mean.  I think this is fine because while it is garbage code, we only need to write it once -- and I'll refactor the `state.rs` file to hide things like this because we mostly never need to see it -- and it lets us use all code working with the standard traits including third party libraries which could save us a lot of time in the future.

kitlith: I mean, so does this approach. The issue (either way, really) is that if a third party library is missing any one of these three, you have to wrap it in a newtype and create dummy implementations. At least with the 'wrapper trait' approach, you aren't implementing, say, Seek on something that can't actually support Seek. *shrug*

only solution I can think of that doesn't require boilerplate is probably 'wrapper trait' + specialization, and even then a) i don't think we want to do that and b) it doesn't appear to work like it does in my head so...

MarkMcCaskey:
@kitlith 
Let me know if it looks good to you

specifically,
```Rust
/// Get WasiState from a Ctx
pub unsafe fn get_wasi_state(ctx: &mut Ctx) -> &mut WasiState {
    &mut *(ctx.data as *mut WasiState)
}
```

Presumably in the future we can also provide higher level APIs that don't require using the raw WASI types which I also exported in this PR.  I wrapped the error type for the public facing APIs and maybe we can do the same with permissions and capabilities as well.  The fd capabilities should map very closely to what `std::file::OpenOptions` is doing, so a similar builder pattern should be simple to implement

MarkMcCaskey:
I updated the plugin example to use the new logging feature.  It relies on the guest calling into the host.  I think we will probably definitely need the builder pattern, too.  Because relying on the guest to do that is less than ideal.  There's also the possibility of Wasm transformations (the host performing arbitrary modifications to the Guest's Wasm before running) but I don't think that's something we support very well right now

kitlith:
I don't really see why the guest needs to call into the host, when there's [wasmer_runtime::Instance::context_mut](https://docs.rs/wasmer-runtime/0.5.7/wasmer_runtime/struct.Instance.html#method.context_mut)? I may be missing something.

And I don't think get_wasi_state should have been changed into a safe function, as it is only sound if the Ctx came from your WASI implementation, right? We (currently) must rely on the user to know this...

MarkMcCaskey:
@kitlith Wow, I did not know about that at all, but it makes sense that that would exist.  That would actually be useful in a personal project I'm making with Wasmer.  Thanks!  I'll update the example.


Hmm, that's a fair point.

So the reason why I made it safe was because in _our_ code we're casting a `&Ctx` to a `&mut Ctx`, which is definitely not safe... here I made it require a `&mut Ctx` so it's safe in that regard, but you're correct, calling this on a non-wasi context will do bad things and have undefined behavior.  I'll update the example and make it unsafe in a PR before we release tomorrow

PR #597, opened 2019-07-30
fix metering benchmark

### LINKED ISSUE #596: "make bench" fails because metering_benchmark has no compiler specified
When I run:

```
./build make bench
```

I get the following error:

```
     Running target/release/deps/metering_benchmark-771e924d5f51f188
Benchmarking Meter/No Metering: Warming up for 3.0000 sthread 'main' panicked at 'compiler not specified, activate a compiler via features', lib/middleware-common/benches/metering_benchmark.rs:168:5
```

Looking at the code, looks like none of the features were enabled:

```rust
#[cfg(not(any(feature = "llvm", feature = "clif", feature = "singlepass")))]
fn get_compiler(_limit: u64, metering: bool) -> impl Compiler {
    panic!("compiler not specified, activate a compiler via features");
    use wasmer_clif_backend::CraneliftCompiler;
    CraneliftCompiler::new()
}
```

If I try to enable the "backend-llvm" feature explicitly:

```
./build cargo bench --all --features backend-llvm
```

the benchmark fails in the exact same way.

I also tried to enable "llvm" feature:

```
./build cargo bench --all --features llvm
```

but it also fails:

```
error: Package `wasmer v0.5.7 (/wasmer)` does not have these features: `llvm`
```

Is the code broken or did I miss some magic to get `make bench` running?

MarkMcCaskey:
resolves #596

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,6 +115,8 @@ jobs:
           name: Check
           command: |
             make check
+            make compile-bench-singlepass
+            # TODO: add compile-bench-llvm and compile-bench-clif when they work
       - run:
           name: Release
           command: make release-fast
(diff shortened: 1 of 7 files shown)


penberg:
@MarkMcCaskey Can you also update `README.md` to describe the new targets?

PR #259, opened 2019-03-11
Re-enable llvm build in appveyor

bjfish:

PR #598, opened 2019-07-30
LLVM backend enabled for Windows

syrusakbary:
This PR:
* Enables LLVM for Windows
* Fixes Appveyor, using a customized build of LLVM 8
* Refactors the memory usage of LLVM

Note: this PR is based on the previous effort done by: #259


syrusakbary:
bors try

lib/llvm-backend/cpp/object_loader.cpp
```diff
@@ -75,16 +75,16 @@ struct MemoryManager : llvm::RuntimeDyld::MemoryManager {
     }
 
     virtual void registerEHFrames(uint8_t* addr, uint64_t LoadAddr, size_t size) override {
-        eh_frame_ptr = addr;
```
nlewycky: Don't commit commented out code.

Does removing this work on linux and mac?

lib/llvm-backend/src/platform/common.rs
```diff
@@ -0,0 +1,3 @@
+pub fn round_up_to_page_size(size: usize) -> usize {
```
nlewycky: Page size is not always 4kib on any given system. Windows and Linux both support different sized pages in the same running system.

On unix, it's `libc::sysconf(_SC_PAGESIZE)` but not on Windows, which suggests maybe this isn't really platform common.

lib/llvm-backend/src/platform/unix.rs
```diff
@@ -68,3 +74,60 @@ extern "C" fn signal_trap_handler(
...
     }
 }
+
+pub unsafe fn alloc_memory(
+    size: usize,
+    protect: MemProtect,
+    ptr_out: &mut *mut u8,
+    size_out: &mut usize,
+) -> LLVMResult {
+    let size = round_up_to_page_size(size);
```
nlewycky: Assuming the rust mmap follows POSIX mmap, neither the `addr` nor `length` arguments need to be page size aligned.

lib/llvm-backend/src/platform/unix.rs
```diff
@@ -68,3 +74,60 @@ extern "C" fn signal_trap_handler(
...
+
+    if res == 0 {
+        LLVMResult::OK
+    } else {
+        LLVMResult::PROTECT_FAILURE
+    }
+}
+
+pub unsafe fn dealloc_memory(ptr: *mut u8, size: usize) -> LLVMResult {
+    let res = munmap(ptr as _, round_up_to_page_size(size));
```
nlewycky: You shouldn't need round_up_to_page_size here either (though it helps to keep the sizes to mmap and munmap the same).

lib/llvm-backend/src/platform/win.rs
```diff
@@ -0,0 +1,80 @@
...
+use std::ptr;
+
+use winapi::um::memoryapi::{VirtualAlloc, VirtualFree};
+use winapi::um::winnt::{
+    MEM_COMMIT, MEM_DECOMMIT, MEM_RESERVE, PAGE_EXECUTE_READ, PAGE_NOACCESS, PAGE_READONLY,
+    PAGE_READWRITE,
+};
+
+pub unsafe fn visit_fde(_addr: *mut u8, _size: usize, _visitor: extern "C" fn(*mut u8)) {
+    // Do nothing on Windows
```
nlewycky: Question, is this ever called on Windows?

syrusakbary: Yup, from the `object_loader.cpp`

lib/llvm-backend/src/platform/common.rs
syrusakbary: This was same code as before, just moved into a new file.
Perhaps we can address this in a separate PR?

lib/llvm-backend/src/platform/unix.rs
syrusakbary: This was same code as before, just moved into a new file.
Perhaps we can address this in a separate PR?

lib/llvm-backend/src/platform/unix.rs
syrusakbary: This was same code as before, just moved into a new file.
Perhaps we can address this in a separate PR?

syrusakbary:
bors try

PR #606, opened 2019-08-01
Cleanup Wasmer

syrusakbary:
This PR moves a few things around trying to make Wasmer more compelling to use:
* Two less directories on the root (`fuzz` is moved to the `lib/` folder, `media` is moved to `src/installer` folder)
* Improved README
* Moved `ARCHITECTURE.md` into the `docs` folder.


bjfish:
Does fuzz and the instructions in fuzz readme.md still work?

syrusakbary:
> Does fuzz and the instructions in fuzz readme.md still work?
Just tried. It didn't. I reverted `fuzz` back to the root folder

> Looks good as far as I can tell, do we test the installer in CI? This change is probably fine, but having it tested would make me more confident
The installer it's only executed when creating tags. However I tried it locally and worked perfectly

syrusakbary:
bors cancel

PR #609, opened 2019-08-01
Update dependencies

syrusakbary:
This PR does:
* Update dependencies (to explicitly use only one version of a package, in all the places we use it)
* Remove explicit dependency on `hashbrown` since it's already adopted as the default algorithm for `HashMap`
* Small fixes on code to use latest versions
* Fixed order of `Instance` exports by using a explicit `IndexMap` instead of an arbitrary ordered `HashMap`


syrusakbary:
bors try

syrusakbary:
bors try

lib/runtime-abi/Cargo.toml
```diff
@@ -8,12 +8,11 @@ repository = "https://github.com/wasmerio/wasmer"
...
 
 [dependencies]
-libc = "0.2.50"
+libc = "0.2.60"
 wasmer-runtime-core = { path = "../runtime-core" }
-hashbrown = "0.1"
 failure = "0.1"
 tar = "0.4"
-wasmparser = "0.34.0"
+wasmparser = "0.35.1"
```
MarkMcCaskey: good catch, not sure how I missed this one

lib/runtime-c-api/tests/runtime_c_api_tests.rs
```diff
@@ -6,7 +6,7 @@ fn test_c_api() {
 
     run_command("cmake", project_tests_dir, Some("."));
     run_command("make", project_tests_dir, Some("-Wdev -Werror=dev"));
-    run_command("make", project_tests_dir, Some("test"));
+    run_command("make", project_tests_dir, Some("test VERBOSE=1"));
```
MarkMcCaskey: Is this an intentional change or debug code?

syrusakbary: It's intentional :)

lib/runtime-core/Cargo.toml
```diff
@@ -8,46 +8,46 @@ repository = "https://github.com/wasmerio/wasmer"
...
 [dependencies.serde-bench]
 version = "0.0.7"
 [dependencies.blake2b_simd]
-version = "0.4.1"
+version = "0.5.5"
 [dependencies.digest]
-version = "0.8.0"
-[dependencies.hashbrown]
-version = "0.1"
-features = ["serde"]
```
bjfish: This hashbrown `serde` feature was why I think we should re-test caching after removing `hashbrown`.

PR #615, opened 2019-08-01
clif-backend: Eliminate FunctionEnvironment construction in feed_event()

penberg:
The feed_event() function is called for every wasm binary instruction.
Let's optimize it by storing FunctionEnvironment object in
CraneliftFunctionCodeGenerator, rather than constructing it for every
feed_event() invocation.

This change reduces the time to run "ngix compile" benchmark by 68%:

Before:

```
  nginx compile           time:   [1.4152 s 1.4186 s 1.4222 s]
  Found 1 outliers among 10 measurements (10.00%)
    1 (10.00%) high mild
```

After:

```
  nginx compile           time:   [447.76 ms 448.32 ms 448.80 ms]
                          change: [-68.542% -68.440% -68.352%] (p = 0.00 < 0.05)
                          Performance has improved.
```

I assume some of the clone() calls are very expensive (Vec::clone(),
likely). I did see libc malloc()/free() high up in "perf top" profiles,
which are eliminted by this change.

I also looked into eliminating FunctionBuilder construction from
feed_event(). That turns out to be painful on lifetime rules because it
borrows bunch of other objects, so I am leaving it for someone who knows
the code better than I do.


bjfish:
bors try

syrusakbary:
I'm following up with some of the changes that were done in this commit to move `FunctionBuilder` out of the `feed_event`: https://github.com/wasmerio/wasmer/commit/d440776bc080cd31086ef4779af84b1b0e0081e2

syrusakbary:
After talking offline we decided to merge this PR and work on refactoring the `clif` backend so we can also move `FunctionBuilder` construction out of the `feed_event`

PR #616, opened 2019-08-01
runtime-c-api: create the import object separately from instance instantiation 

YaronWittenstein:
Creating an `import object` separately from the `wasmer` instance, let's us build the import object incrementally and support the use-case of having import functions (a.k.a `vmcalls`) implemented in 2 programming languages (Rust and Golang for example) and consumed by the same `wasmer` instance.


bjfish:
bors try

bjfish:
LGTM
bors r+

PR #621, opened 2019-08-02
Add deny dead_code lint and fix one error

bjfish:


bjfish:
bors try

bjfish:
bors try

syrusakbary:
Merging directly since there are no conflicts and tests passed.

lib/emscripten/src/env/mod.rs
```diff
@@ -30,6 +30,7 @@ pub fn call_malloc(ctx: &mut Ctx, size: u32) -> u32 {
         .unwrap()
 }
 
+#[warn(dead_code)]
```
MarkMcCaskey: is this different from `#[must_use]`?

MarkMcCaskey: Oh, it's normally deny but we're making it warn.  I understand

PR #622, opened 2019-08-02
runtime-c-api: `wasmer_instantiate` refactoring

YaronWittenstein:
instance.rs - `wasmer_instantiate` calls `wasmer_import_object_extend`


syrusakbary:
This should not be needed in the current codebase in `master`.

PR #623, opened 2019-08-02
Add support for (register) command in spec tests and enable importing mutable globals.

nlewycky:
Enable the part of the simd spec tests that now pass.


nlewycky:
bors try

PR #624, opened 2019-08-03
runtime-c-api: adding .gitignore with `wasmer.h` and `wasmer.hh` 

YaronWittenstein:
these files are auto-generated upon build


bjfish:
Hi @YaronWittenstein, we purposefully keep the header files in the repository to serve as quick access to documentation of the API and to be able to detect when the generated headers change.

The OUT dir headers are there to be used as part of the distribution.

PR #636, opened 2019-08-06
Fix traps with LLVM backend in Windows.

losfair:
Exception handling for LLVM backend is not working properly, possibly related to [this LLVM issue](https://bugs.llvm.org/show_bug.cgi?id=24233).

This PR tries to work around it by avoiding using C++ exceptions across stack frames owned by JIT-generated code.


lib/llvm-backend/cpp/object_loader.cpp
```diff
@@ -1,10 +1,60 @@
...
+  if(setjmp(stack_out)) {
+
+  } else {
+    func(userdata);
+  }
+}
+
+[[noreturn]]
+void unwinding_longjmp(jmp_buf stack_in) {
+  longjmp(stack_in, 42);
```
nlewycky: Just `1` here.

lib/llvm-backend/cpp/object_loader.cpp
```diff
@@ -1,10 +1,60 @@
 #include "object_loader.hh"
 #include <iostream>
 #include <memory>
+#include <setjmp.h>
 
 extern "C" void __register_frame(uint8_t *);
 extern "C" void __deregister_frame(uint8_t *);
 
+void unwinding_setjmp(jmp_buf stack_out, void (*func)(void *), void *userdata) {
```
nlewycky:
```
#ifdef __GNUC__
__attribute__((returns_twice))
#endif
void unwinding_setjmp(jmp_buf stack_out, void (*func)(void *), void *userdata) {
```

syrusakbary:
I was unable to have spectests passing on my local Windows:

```
C:\Users\syrusakbary\wasmer2>cargo test --manifest-path lib/spectests/Cargo.toml --release --features llvm -- --test-threads=1
```

Am I doing something wrong when testing?

syrusakbary:
I've just added the test in AppVeyor to check that Spectests are passing in Windows with LLVM

syrusakbary:
Weirdly enough, I tested without `--release` mode and it works perfectly.

```bash
# Works on Windows
cargo test --manifest-path lib/spectests/Cargo.toml --features llvm -- --test-threads=1

# Doesn't work on Windows
cargo test --manifest-path lib/spectests/Cargo.toml --release --features llvm -- --test-threads=1
```

lib/llvm-backend/cpp/object_loader.cpp
losfair: What `unwinding_setjmp` does is to call the callback `func` and catch `longjmp`'s backward.  It only returns once and I think `returns_twice` is not needed.

losfair:
I can only reproduce the test failure in release mode when not specifying `--test-threads=1`, and the error messages are like `IMAGE_REL_AMD64_ADDR32NB relocation requires anordered section layout.`. Might be related to calling LLVM from multiple threads.

With `--test-threads=1` both debug and release modes work for me. @syrusakbary are you using msvc or gnu abi?

syrusakbary:
@losfair emtests seem to be failing for llvm now (in linux)

lib/llvm-backend/cpp/object_loader.cpp
nlewycky: Oh yes, you're correct.

syrusakbary:
This PR should not be needed any longer thanks to our work in the refactor.
Closing it

PR #641, opened 2019-08-07
The feature is named "backend-cranelift" in Cargo.toml.

nlewycky:
Not a big deal right now because cranelift does not implement metering yet.


MarkMcCaskey:
Can we double check that this is correct?  There are multiple features across multiple sub-crates, grepping for `backend-cranelift` shows the feature in toml files and Rust files

syrusakbary:
Merging manually as there are no code changes nor changes in used behavior

PR #642, opened 2019-08-08
Unexclude some now passing assert_unlinkable

bjfish:


syrusakbary:
It seems tests passed. Merging manually (not sure why bors cancelled)

PR #648, opened 2019-08-09
Remove structopt dependency from LLVM

syrusakbary:
* [x] Remove structopt dependency from LLVM
* [x] Fixed `dyn` warning in the wasmer cli
* [x] Make `nix` an optional dependency for LLVM backend

### DIFF
--- a/lib/llvm-backend/src/lib.rs
+++ b/lib/llvm-backend/src/lib.rs
@@ -17,7 +17,6 @@ mod structs;
 mod trampolines;
 
 use std::path::PathBuf;
-use structopt::StructOpt;
 
 pub use code::LLVMFunctionCodeGenerator as FunctionCodeGenerator;
 pub use code::LLVMModuleCodeGenerator as ModuleCodeGenerator;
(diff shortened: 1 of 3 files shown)


lib/llvm-backend/src/lib.rs
```diff
@@ -31,23 +30,20 @@ pub type LLVMCompiler = SimpleStreamingCompilerGen<
     code::CodegenError,
 >;
 
-#[derive(Debug, StructOpt, Clone)]
+#[derive(Debug, Clone)]
 /// LLVM backend flags.
-pub struct CLIOptions {
+pub struct LLVMOptions {
```
MarkMcCaskey: I'm confused why this struct exists.  Is it publicly exposed to users?  It seems identical to the one with structopt defined on it

src/bin/wasmer.rs
```diff
@@ -363,8 +380,13 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
     #[cfg(feature = "backend-llvm")]
     {
         if options.backend == Backend::LLVM {
+            let options = options.backend_llvm_options.clone();
             unsafe {
```
MarkMcCaskey: If we're going to use mutable global variables, I'm a fan of what I did in Wapm which is wrap it in a write-once abstraction that is thread safe by way of Mutex or RW Lock.  Basically just having the compiler enforce that you can't use it wrong.


I'm generally a fan of no top-level unsafe.  `wasmer.rs` should be business logic and unsafe means you have to care about how things work which slows down reading and modifying the code.  Quickly implementing business logic changes is important from a business and product perspective, so in my opinion they don't mix well.

If this is a temporary fix then it's fine, but I think no unsafe in business logic is something we should stick to.
this looks like this unsafe was already here though -- I wasn't aware we were using this.

nlewycky: Ideally these shouldn't be global, they should be tied to each LLVMCompiler object in some fashion.

PR #649, opened 2019-08-09
Fix some spectest imports and namespace_table issue

bjfish:


lib/runtime-core/src/parse.rs
```diff
@@ -369,7 +368,16 @@ pub fn read_module<
 
                 info.write().unwrap().globals.push(global_init);
             }
-
```
bjfish: @losfair I think the issue here was modules without function bodies, so the following wasn't called. Please review.

lib/spectests/tests/excludes.txt
```diff
@@ -22,164 +22,48 @@ clif:skip:names.wast:*         # Names file has parsing error?
...
-clif:fail:data.wast:211 # AssertUnlinkable - caught panic Any
-clif:fail:data.wast:227 # AssertUnlinkable - caught panic Any
-clif:fail:data.wast:258 # AssertUnlinkable - caught panic Any
-clif:fail:data.wast:273 # AssertUnlinkable - caught panic Any
-clif:fail:elem.wast:101 # Module - caught panic Any
 clif:fail:elem.wast:143 # AssertUnlinkable - instantiate successful, expected unlinkable
 clif:fail:elem.wast:152 # AssertUnlinkable - caught panic Any
 clif:fail:elem.wast:161 # AssertUnlinkable - instantiate successful, expected unlinkable
 clif:fail:elem.wast:170 # AssertUnlinkable - instantiate successful, expected unlinkable
 clif:fail:elem.wast:178 # AssertUnlinkable - instantiate successful, expected unlinkable
```
bjfish: I reviewed some of these regressions and they seem unrelated. Most likely were false positives previously.

PR #650, opened 2019-08-09
implement wasi::path_rename

MarkMcCaskey:
Implement and test this new call, improving things at the same time (and introducing a few new bugs while I'm at it, called out in comments in the code, they're dependent on abstracting paths which is the second or third thing on my todo list right now)


MarkMcCaskey:
bors try

syrusakbary:
Since tests passed I think we can merge manually

PR #651, opened 2019-08-09
Fix LLVM speed regression

syrusakbary:
When canonicalizing all the NaNs with the LLVM backend, the performance was slowed down 2x.

This PR fixes that by reverting the canonicalization

@nlewycky

### DIFF
--- a/lib/llvm-backend/src/code.rs
+++ b/lib/llvm-backend/src/code.rs
@@ -360,45 +360,6 @@ fn trap_if_zero(
     builder.position_at_end(&shouldnt_trap_block);
 }
 
-// Replaces any NaN with the canonical QNaN, otherwise leaves the value alone.
-fn canonicalize_nans(
-    builder: &Builder,
-    intrinsics: &Intrinsics,
-    value: BasicValueEnum,
-) -> BasicValueEnum {
-    let f_ty = value.get_type();
-    let canonicalized = if f_ty.is_vector_type() {
-        let value = value.into_vector_value();
-        let f_ty = f_ty.into_vector_type();
-        let zero = f_ty.const_zero();
-        let nan_cmp = builder.build_float_compare(FloatPredicate::UNO, value, zero, "nan");
-        let canonical_qnan = f_ty
-            .get_element_type()
-            .into_float_type()
-            .const_float(std::f64::NAN);
-        let canonical_qnan = splat_vector(
-            builder,
-            intrinsics,
-            canonical_qnan.as_basic_value_enum(),
-            f_ty,
-            "",
-        );
-        builder
-            .build_select(nan_cmp, canonical_qnan, value, "")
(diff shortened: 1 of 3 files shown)


syrusakbary:
bors try

syrusakbary:
Previous timing:

```
wasmer run --backend=llvm --disable-cache --enable-simd   18.95s user 0.08s system 99% cpu 19.114 total
```

With the changes from this PR:
```
/Users/syrusakbary/Development/wasmer/target/release/wasmer run --backend=llv  14.21s user 0.03s system 99% cpu 14.263 total
```

nlewycky:
bors try

lib/llvm-backend/src/code.rs
```diff
@@ -399,6 +408,31 @@ fn canonicalize_nans(
     canonicalized
 }
 
+// Replaces any NaN with the canonical QNaN, otherwise leaves the value alone.
```
bjfish: Style idea: maybe consider the opposite name:  `make_canonical_nan`.

penzn:
Hi, I have reported the regression originally in wasmerio/c-wasm-simd128-example#1. If I reset `master` to commit right before 11f66d287b7e9ee81f49315c44e45f0089acf26a then I see the timings from the [article][a], but timings from the branch in this PR is still similar to `master`. Is that me or are you seeing that too?

[a]: https://medium.com/wasmer/webassembly-and-simd-13badb9bf1a8

Hywan:
@syrusakbary What's the status of this PR?

nlewycky:
This PR is obsoleted by PR #883 and PR #934.

nlewycky:
@penzn Yes, we noticed the same thing. It's a bit of a pain because LLVM doesn't say a lot about what happens with NaN values on operations, even as simple as an `fadd` instruction. It happened to work before, because it lowers to an addition that happens to preserve NaNs. Unfortunately, while the work I did to try to bring performance back did help on many examples, it hasn't helped to improve the c-wasm-simd128-example when built with SIMD. The primary difference there remains whether the LLVM autovectorizer manages to vectorize the `fdiv` instruction. That can be affected by many things, we've also changed the LLVM pass list to improve optimizations which should improve the chance that we get the vectorized fdiv in the simd example, but unfortunately we still don't. I'd still like to figure out how to get guaranteed correctness while preserving the performance, especially since it should be possible given that the x86-64 instructions give us what we need.

penzn:
@nlewycky I just rebuilt the benchmark with `-munimplemented-simd128`, which leads to vector `fdiv` in the wasm module, and performance is the same as with scalar `fdiv`. It also looks relatively close to native (given that I keep caching on). Can you elaborate a bit on what you are seeing, I'd be happy to help.

Maybe we should open an issue for this.

CC @topperc

Issue #644, opened 2019-08-08
[wasmparser library] Assertion failed in wasmparser::operators_validator::FuncState::push_block


pventuzelo:
This bug is related to the 3rd party library [wasmparser](https://github.com/yurydelendik/wasmparser.rs) used in wasmer

### Describe the bug

Assertion failed in `wasmparser::operators_validator::FuncState::push_block (operators_validator.rs:74)` i.e. [here](https://github.com/yurydelendik/wasmparser.rs/blob/master/src/operators_validator.rs#L74).
This issue is making wasmer to panicked.

### Status of my environment

```bash
echo "`wasmer -V` | `rustc -V` | `uname -m`"
wasmer 0.6.0 | rustc 1.38.0-nightly (0b680cfce 2019-07-09) | x86_64
```

### Steps to reproduce

Download [panic_assert_wasmparser_operators_validators.zip](https://github.com/wasmerio/wasmer/files/3481990/panic_assert_wasmparser_operators_validators.zip)
```
$ unzip panic_assert_wasmparser_operators_validators.zip
$ wasmer run panic_assert_wasmparser_operators_validators.wasm
thread 'main' panicked at 'assertion failed: stack_starts_at + index < self.stack_types.len()', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/wasmparser-0.34.0/src/operators_validator.rs:74:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

### Expected behavior

`wasmparser` should detect and handle the error properly when calling `assert_stack_type_at`.
`wasmer` should maybe catch the panics of `wasmparser` functions and quit properly.

### Actual behavior

```
$ RUST_BACKTRACE=1 wasmer run panic_assert_wasmparser_operators_validators.wasm
thread 'main' panicked at 'assertion failed: stack_starts_at + index < self.stack_types.len()', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/wasmparser-0.34.0/src/operators_validator.rs:74:9
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
[...]

PR #652, opened 2019-08-09
Update wasmparser (fix issue 644)

pventuzelo:
Update wasmparser since issue 644 have been fixed upstream.
https://github.com/wasmerio/wasmer/issues/644

Not sure if i'm supposed to update Cargo.lock.
Otherwise, maybe we should update .gitignore.


pventuzelo:
Not needed anymore, fixed in master

PR #655, opened 2019-08-10
Fix more bare dyn traits in runtime-c-api

bjfish:


bjfish:
There appears to be a transient error relating to llvm middleware.
bors r+

PR #657, opened 2019-08-10
Spectest Fixes

bjfish:
- Invalid table size for elems should produce link error
- Remove filter which prevents initializer offset check

### DIFF
--- a/lib/runtime-core/src/backing.rs
+++ b/lib/runtime-core/src/backing.rs
@@ -134,7 +134,6 @@ impl LocalBacking {
             .info
             .data_initializers
             .iter()
-            .filter(|init| init.data.len() > 0)
         {
             let init_base = match init.base {
                 Initializer::Const(Value::I32(offset)) => offset as u32,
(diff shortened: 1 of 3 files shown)


lib/runtime-core/src/backing.rs
```diff
@@ -237,9 +237,9 @@ impl LocalBacking {
                     let table = &tables[local_table_index];
 
                     if (table.size() as usize) < init_base + init.elements.len() {
-                        let delta = (init_base + init.elements.len()) - table.size() as usize;
-                        // Grow the table if it's too small.
-                        table.grow(delta as u32).expect("couldn't grow table");
```
syrusakbary: Just double-checking. We don't want to grow the table in this case, right?

bjfish: Yes, not to grow appears to be the correct implementation. I checked with one other implementation, the minimum table size should be enough to fit the elems intializer.

syrusakbary:
Two test failed on Windows:
```
clif:fail:data.wast:186 # AssertUnlinkable - caught panic Any
clif:fail:data.wast:194 # AssertUnlinkable - caught panic Any
```

PR #658, opened 2019-08-10
Re-use registered module instances in spectests

bjfish:
It appears spectests expects a register module instances to be same instance as named instances (can't instantiate a new instance):
https://github.com/wasmerio/wasmer/blob/1f8f8102229084a7f111f8239a9494c92184cb01/lib/spectests/spectests/elem.wast#L366


bjfish:
bors try

PR #659, opened 2019-08-10
Validate initializers before writing to memory

bjfish:
Fixes `linking.wast:342`

### DIFF
--- a/lib/runtime-core/src/backing.rs
+++ b/lib/runtime-core/src/backing.rs
@@ -159,14 +159,12 @@ impl LocalBacking {
                 LocalOrImport::Import(imported_memory_index) => {
                     // Write the initialization data to the memory that
                     // we think the imported memory is.
-                    unsafe {
-                        let local_memory = &*imports.vm_memories[imported_memory_index];
-                        let data_top = init_base + init.data.len();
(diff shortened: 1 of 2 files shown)


bjfish:
bors try

lib/runtime-core/src/backing.rs
```diff
@@ -143,17 +145,49 @@ impl LocalBacking {
...
                     if memory_desc.minimum.bytes().0 < data_top || data_top < init_base {
                         return Err(vec![LinkError::Generic {
                             message: "data segment does not fit".to_string(),
                         }]);
                     }
+                }
+                LocalOrImport::Import(imported_memory_index) => {
+                    // Write the initialization data to the memory that
+                    // we think the imported memory is.
+                    unsafe {
```
syrusakbary: Can we move the unsafe only for the local_memory retrieval?

```rust
let local_memory = unsafe { &*imports.vm_memories[imported_memory_index] };
 ```

bjfish: Done, I also minimized the other `unsafe` block at 38a8a0eb01d2802f7a5899e7276653d075f21a2d

PR #669, opened 2019-08-13
Finalize signatures & names when no function bodies

bjfish:


lib/runtime-core/src/parse.rs
```diff
@@ -369,12 +369,17 @@ pub fn read_module<
                 info.write().unwrap().globals.push(global_init);
             }
             ParserState::EndWasm => {
-                if namespace_builder.is_some() {
+                // TODO Consolidate with BeginFunction body if possible
+                if func_count == ::std::usize::MAX {
```
syrusakbary: Should be `>=` ?

bjfish: @syrusakbary `func_count` is initialized as follows:
```
    let mut func_count: usize = ::std::usize::MAX;
```
So, this branch would run if we had the initial size or some how had ` ::std::usize::MAX` functions which I could update this code to handle that scenario (not sure if thats possible).

syrusakbary: Umm... should we start with `func_count = 0`? Is there any reason on why is the max by default?

bjfish:
bors try

lib/runtime-core/src/parse.rs
bjfish: I've updated to start at `func_count = 0`. please re-review.

PR #674, opened 2019-08-14
Return compile error (llvm) for setting immutable global

bjfish:


lib/llvm-backend/src/code.rs
```diff
@@ -1151,7 +1151,9 @@ impl FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator {
                         builder.build_store(ptr_to_value, value);
                     }
                     GlobalCache::Const { value: _ } => {
-                        unreachable!("cannot set non-mutable globals")
+                        return Err(CodegenError {
+                            message: "global is immutable".to_string(),
```
MarkMcCaskey: It'd be nice to include more information here if it's available, like what global and what value maybe.  To make debugging easier

bjfish: I like having more context, I'll start thinking about this as I see more messages.

PR #677, opened 2019-08-14
Mark memory.size VM intrinsics as not modifying state, and throw.trap/breakpoint as functions which do not return.

nlewycky:


nlewycky:
Other potentially interesting attributes from llvm's list of function attributes:
  http://llvm.org/docs/LangRef.html#fnattrs
* `cold` (calls to this are guaranteed rare, implied by noreturn)
* `nounwind` (callee does not throw an exception)
* `readnone` (callee decides what to do (return or throw exception) based on its arguments and not global state)
* `willreturn`  (callee eventually returns or throws, but never exits the program)
* `argmemonly` (callee does not access global variables by name (it may access them if given a pointer to one as its argument))
* `inaccessiblememonly` and `inaccessiblemem_or_argmemonly` (similar to `readnone` but used only for JIT environments, indicates that the memory it accesses is private to the callee and not otherwise visible to the code represented in LLVM IR in this LLVM Module)

Some of these may also be used as attributes on pointer-typed arguments, such as indicating that a given pointer is `readonly` (no stores through this pointer), `readnone` (pointer is only used for its value, the pointed-to memory is never used), `nocapture` (the pointer value itself does not matter, only the contents of the pointed-to memory, used for copy elimination), `nonnull` which indicates that the pointer is literally not null (but nothing else), `dereferenceable(<n>)` and `dereferenceable_or_null(<n>)` which indicate that the pointer points to memory which is safe to deference the first _\<n\>_ bytes of, and `returned` which indicates that this argument is the value returned by the function.

PR #682, opened 2019-08-16
Specify wasmer as default-run binary

bjfish:


Hywan:
```
Caused by:
  the `default-run` manifest key is unstable

Caused by:
  feature `default-run` is required

consider adding `cargo-features = ["default-run"]` to the manifest
```

MarkMcCaskey:
> ```
> Caused by:
>   the `default-run` manifest key is unstable
> 
> Caused by:
>   feature `default-run` is required
> 
> consider adding `cargo-features = ["default-run"]` to the manifest
> ```

Hmmm,

This should be stable as of 1.37 maybe CI is out of date.  Side note: this'll make us dependent on 1.37 which maybe we don't want to do yet, we get a surprising amount of bug reports from people who think that they need nightly Rust to compile Wasmer even though we should be okay on 1.36 right now.  I think this is due to the error messages Rustc gives -- we can probably fix this with something like

```
#[cfg(rustc_version > 1.36)]
compile_error!("Please update to Rust 1.37 or greater");
```

I'll look into it and make a PR

MarkMcCaskey:
This requires using build.rs right now -- seems a bit hacky, there's a Rust issue to add version checking as a cfg or cfg_attr option; for now just updating the README in #712 ; if we notice that people are still confused, then we can go with compile time version checking

bjfish:
Closing this as low priotiry

MarkMcCaskey:
New Rustc comes out tomorrow I think -- maybe that's a good time to bump our required version and use this feature

PR #684, opened 2019-08-16
Add contributing.md

bjfish:


torch2424:
Sorry it took me so long to look at this @bjfish 😢 

One thing I think that would be good, and I find helpful, is a quick project layout explainer. Like what each major directory contains / is laid out. I feel like it really helps newcomers traverse the project 👍

PR #699, opened 2019-08-19
Abstract file access to use WasiPath trait

MarkMcCaskey:
Improving the embedded use case by removing code that assumes files are host files and replaces it with abstract code.

There's a lot of potential design work here, and I've chose to keep it simple to start.  We can iterate on it later

### DIFF
--- a/lib/wasi/src/state/mod.rs
+++ b/lib/wasi/src/state/mod.rs
@@ -69,7 +69,7 @@ pub enum Kind {
         handle: Option<Box<dyn WasiFile>>,
         /// The path on the host system where the file is located
         /// This is deprecated and will be removed in 0.7.0 or a shortly thereafter
-        path: PathBuf,
+        path: Box<dyn WasiPath>,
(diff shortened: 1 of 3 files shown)


MarkMcCaskey:
These changes need more design / design review

MarkMcCaskey:
Closing this PR because it needs more design:

It's not clear that it makes sense for the lines of abstraction to be drawn at the directory level like this, and if it does, it seems like it could be made simpler than having so many methods on the trait.  The reason is that as-is, the abstraction is powerful enough to do bad things but not powerful enough to do other useful things.  Maybe we want to allow key-value style filesystems, which would be hard here.

I think it's possible to create a reasonable abstraction here, but it will take a pretty good amount of focused effort.  Right now, I think the costs of implementing this are much higher than our gains.  We can introduce unsafe hacky APIs, like the create virtual dir on a not-yet-published branch from a few weeks ago, as as stop-gap.

PR #703, opened 2019-08-20
Fix compilation on AArch64 Linux.

losfair:
This only fixes compilation. Cranelift still reports many `not yet implemented` on AArch64.


lib/emscripten/src/syscalls/mod.rs
```diff
@@ -168,13 +168,13 @@ pub fn ___syscall38(ctx: &mut Ctx, _which: c_int, mut varargs: VarArgs) -> i32 {
     debug!("emscripten::___syscall38 (rename)");
     let old_path = varargs.get_str(ctx);
     let new_path = varargs.get_str(ctx);
-    let real_old_path_owned = get_cstr_path(ctx, old_path);
+    let real_old_path_owned = get_cstr_path(ctx, old_path as *const _);
```
MarkMcCaskey: I'm curious why this is an architecture specific issue :thinking:

losfair: Rust libstd for ARM (both 32 and 64) uses `u8` instead of `i8` in many places. I don't understand why though.

PR #710, opened 2019-08-22
fixing breaking change of `wasmparser`

YaronWittenstein:
Edit by @Hywan: Fix #707


Hywan:
cc @xmclark @nlewycky, can you get a second review please? Do we want to update wasmparser now?

Hywan:
bors try

nlewycky:
I'm surprised it passed. The version of `wasmparser` used by wasmer and used by the cranelift crate used in wasmer must match or there should be a build failure.

If it builds, that's great though?

YaronWittenstein:
> I'm surprised it passed. The version of `wasmparser` used by wasmer and used by the cranelift crate used in wasmer must match or there should be a build failure.
> 
> If it builds, that's great though?

_Bors_ failed above because of the duplicate `wasmerparser` versions

PR #713, opened 2019-08-22
Add AArch64 support for singlepass.

losfair:
This PR includes:

- Support for AArch64 (ARM64) in Singlepass backend. Implemented with a combination of x86_64 instruction translation and native code generation.
- State tracing and backtraces on AArch64/Singlepass. (Tiering is not implemented for this pair because there's no other backend supporting AArch64 yet)
- Debugging tools: `BlockTrace` middleware, and support for reading states of previous WebAssembly stack frames from middlewares.


Cargo.toml
```diff
@@ -58,7 +58,8 @@ members = [
     "lib/wasi-tests",
     "lib/emscripten-tests",
     "lib/middleware-common-tests",
-    "examples/plugin-for-example"
+    "examples/plugin-for-example",
+    "examples/hello_world"
```
nlewycky: Every other directory uses hyphens to separate words. Please follow suit.

lib/runtime-core/src/backend.rs
```diff
@@ -70,6 +70,82 @@ impl std::str::FromStr for Backend {
     }
 }
 
+#[derive(Copy, Clone, Debug)]
+pub enum Architecture {
```
nlewycky: At the risk of sounding silly, why do we need this? We ordinarily use cargo CFG to detect what this wasmer build is targeting, so why a runtime value to track it?

lib/runtime-core/src/backend.rs
```diff
@@ -70,6 +70,82 @@ impl std::str::FromStr for Backend {
...
+    X64,
+    Aarch64,
+}
+
+#[repr(u8)]
+#[derive(Copy, Clone, Debug)]
+pub enum InlineBreakpointType {
+    Trace,
```
nlewycky: What are these? Could you add doc comments?

lib/runtime-core/src/backend.rs
```diff
@@ -70,6 +70,82 @@ impl std::str::FromStr for Backend {
...
+                    None
+                }
+            }
+            _ => None,
+        },
+        Architecture::Aarch64 => match backend {
+            Backend::Singlepass => {
+                if code.len() < 12 {
+                    None
+                } else if &code[..8] == &[0, 0, 0, 0, 0xff, 0xff, 0xff, 0xff] {
```
nlewycky: This is an invalid instruction encoding. Why is that OK for Aarch64 but not X64?

lib/runtime-core/src/backend.rs
```diff
@@ -70,6 +70,82 @@ impl std::str::FromStr for Backend {
...
+    backend: Backend,
+    code: &[u8],
+) -> Option<InlineBreakpoint> {
+    match arch {
+        Architecture::X64 => match backend {
+            Backend::Singlepass => {
+                if code.len() < 7 {
+                    None
+                } else if &code[..6] == &[0x0f, 0x0b, 0x0f, 0xb9, 0xcd, 0xff] {
+                    // ud2 ud (int 0xff) code
```
nlewycky: What is int $0xff anyways?

lib/runtime-core/src/fault.rs
```diff
@@ -25,13 +34,19 @@ use std::process;
...
+#[cfg(not(target_arch = "x86_64"))]
+pub(crate) unsafe fn run_on_alternative_stack(stack_end: *mut u64, stack_begin: *mut u64) -> u64 {
+    unimplemented!("run_on_alternative_stack");
+}
+
 const TRAP_STACK_SIZE: usize = 1048576; // 1MB
 
-const SETJMP_BUFFER_LEN: usize = 27;
+const SETJMP_BUFFER_LEN: usize = 128;
 type SetJmpBuffer = [i32; SETJMP_BUFFER_LEN];
```
nlewycky: Could I request an `#repr(align(`_x_`))` here? It might already be aligned enough, but I would really like it to be explicit so that it's clear that it's been checked, and in the future needs to be updated.

lib/runtime-core/src/fault.rs
```diff
@@ -400,24 +552,29 @@ pub unsafe fn get_fault_info(siginfo: *const c_void, ucontext: *const c_void) ->
...
 
-    known_registers[X64Register::XMM(XMM::XMM0).to_index().0] = Some(read_xmm(&fpregs._xmm[0]));
-    known_registers[X64Register::XMM(XMM::XMM1).to_index().0] = Some(read_xmm(&fpregs._xmm[1]));
-    known_registers[X64Register::XMM(XMM::XMM2).to_index().0] = Some(read_xmm(&fpregs._xmm[2]));
-    known_registers[X64Register::XMM(XMM::XMM3).to_index().0] = Some(read_xmm(&fpregs._xmm[3]));
-    known_registers[X64Register::XMM(XMM::XMM4).to_index().0] = Some(read_xmm(&fpregs._xmm[4]));
-    known_registers[X64Register::XMM(XMM::XMM5).to_index().0] = Some(read_xmm(&fpregs._xmm[5]));
-    known_registers[X64Register::XMM(XMM::XMM6).to_index().0] = Some(read_xmm(&fpregs._xmm[6]));
-    known_registers[X64Register::XMM(XMM::XMM7).to_index().0] = Some(read_xmm(&fpregs._xmm[7]));
+    // WSL bug
```
nlewycky: ???

lib/singlepass-backend/src/emitter_x64.rs
```diff
@@ -168,6 +170,16 @@ pub trait Emitter {
     fn emit_call_location(&mut self, loc: Location);
 
     fn emit_bkpt(&mut self);
+
+    fn emit_homomorphic_host_redirection(&mut self, target: GPR);
```
nlewycky: I don't understand what this function is for from the name. Could you add a comment?

src/bin/wasmer.rs
```diff
@@ -397,7 +403,7 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
             .map_err(|e| format!("Can't convert from wast to wasm: {:?}", e))?;
     }
 
-    let compiler: Box<dyn Compiler> = match get_compiler_by_backend(options.backend) {
+    let compiler: Box<dyn Compiler> = match get_compiler_by_backend(options.backend, options) {
```
nlewycky: Maybe only pass `options` and let `get_compiler_by_backend` pull `backend` out of `options` itself?

lib/runtime-core/src/backend.rs
losfair: In case in the future we compile on one architecture and run on another architecture.

lib/runtime-core/src/backend.rs
losfair: I'm thinking it's better to use documented UDs for each architecture rather than undocumented ones?

lib/runtime-core/src/backend.rs
losfair: As a "special" pattern to prevent conflict with normal code generated from valid WebAssembly.

lib/runtime-core/src/fault.rs
losfair: On Windows Subsystem for Linux (WSL), the fpregs field is always null and no XMMs can be read. This bug is discussed in [a WSL issue](https://github.com/microsoft/WSL/issues/2555). Here we just skip fpregs if it's null to work around the bug.

lib/runtime-core/src/backend.rs
nlewycky: Right, so 0x00 0x00 0x00 0x00 is a valid undefined instruction `udf #0`. 0xff 0xff 0xff 0xff is just an invalid encoding though? If you want a different `udf` you could use `udf #65535` as 0xff 0xff 0x00 0x00, or switch it up and use `brk #0` at 0x00 0x00 0x20 0xd4.

nlewycky:
> @nlewycky Yeah currently lowering for different architectures are implemented at a very low level, and the x86 instruction set serves as a low-level "IR" before the actual assembler. I would like to gradually make changes to the low-level IR so that it is platform-neutral, and convert the `Emitter` trait to be more like a macro-assembler.
> 
> The benefits of the current design include bi-directional mappable machine-level layout (registers, stack, etc.) which makes OSR work without major rewrite for each new architecture, and a unique `codegen.rs` which enables the addition of many new WebAssembly opcodes without writing implementation for each architecture.

OK, I think we need to talk about this and agree on an approach. X86 does not serve as a low-level IR and building a low-level IR out of the wasm would make the system not be single pass. If we're going to have a macro assembler, we should use an existing compiler to minimize the amount of work we need to do (one that already supports platforms we want, etc.) and then run that macro assembler to produce the assembly snippets in a .rs file that then are built into wasmer. The code we need to run at runtime is stitching those snippets together.

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try-

losfair:
bors try

losfair:
bors try

losfair:
bors try

losfair:
bors try

lib/runtime-core/src/fault.rs
```diff
@@ -367,19 +449,109 @@ unsafe fn install_sighandler() {
...
 }
 
-/// Gets fault info for the given siginfo and context pointers.
+impl FaultInfo {
+    /// Parses the stack and builds an execution state image.
+    pub unsafe fn read_stack(&self, max_depth: Option<usize>) -> Option<ExecutionStateImage> {
+        let rsp = match self.known_registers[X64Register::GPR(GPR::RSP).to_index().0] {
+            Some(x) => x,
+            None => return None,
+        };
```
MarkMcCaskey: ⛳️ 
`let rsp = self.known_registers[X64Register::GPR(GPR::RSP).to_index().0]?;`

lib/runtime-core/src/fault.rs
```diff
@@ -425,33 +596,43 @@ pub unsafe fn get_fault_info(siginfo: *const c_void, ucontext: *const c_void) ->
...
+        known_registers[X64Register::XMM(XMM::XMM14).to_index().0] =
+            Some(read_xmm(&fpregs._xmm[14]));
+        known_registers[X64Register::XMM(XMM::XMM15).to_index().0] =
+            Some(read_xmm(&fpregs._xmm[15]));
+    }
 
     FaultInfo {
         faulting_addr: si_addr as usize as _,
-        ip: gregs[REG_RIP as usize] as _,
+        ip: std::mem::transmute::<&mut i64, &'static Cell<usize>>(&mut gregs[REG_RIP as usize]),
```
MarkMcCaskey: I think this can probably cause undefined behavior but it's pretty subtle.  The fact that a normal type in Rust exists that is being accessed with `Cell` or `UnsafeCell` is unsound.  `Cell` by way of `UnsafeCell` is the only way to have aliased mutability in Rust without undefined behavior.  This code allows aliased mutability where one side may be a Rust type without `Cell` or `UnsafeCell`.  I believe that is undefined behavior.

lib/runtime-core/src/state.rs
```diff
@@ -918,6 +1063,7 @@ pub mod x64 {
             let mut is_baseline: Option<bool> = None;
 
             for version in versions() {
+                //println!("Lookup IP: {:x}", ret_addr);
```
MarkMcCaskey: Debug code? probably should remove it

lib/runtime-core/src/state.rs
```diff
@@ -1105,144 +1251,10 @@ pub mod x64 {
                 stack: wasm_stack,
                 locals: wasm_locals,
             };
+            //println!("WFS = {:?}", wfs);
```
MarkMcCaskey: same here

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -4214,54 +4699,79 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                        Size::S64,
+                        loc,
+                        Location::XMM(tmp_in),
+                    );
+                    Self::emit_f64_int_conv_check(
+                        a,
+                        &mut self.machine,
+                        tmp_in,
+                        -1.0,
+                        18446744073709551616.0,
```
MarkMcCaskey: ⛳️ 
`u64::max_value() as f64` (if this is the same, then I think it's more readable)

src/bin/wasmer.rs
MarkMcCaskey: repinging on this

losfair: When tiering is enabled, we can be using a different `backend` than the one specified in `options`.

PR #736, opened 2019-08-30
Add metered compile and check functions to c api

ethanfrey:
Created `metering.rs` and exposed:

* wasmer_compile_with_limit - hardcoded to singlepass backend now, as only stable one
* wasmer_instance_get_points_used
* wasmer_instance_set_points_used

I also added a `metering` feature flag. And the functions default to `wasmer_compile` and "noops" if the feature is disabled (but expose the same API).

I will work on testing an example in the go-ext-wasm repo, then submit this for review, but I am happy for any early feedback as well.


lib/runtime/src/cache.rs
```diff
@@ -109,6 +109,7 @@ impl Cache for FileSystemCache {
 
     fn store(&mut self, key: WasmHash, module: Module) -> Result<(), CacheError> {
         let filename = key.encode();
+        // TODO: is the backend enough, or do we need the metering as well?
```
Hywan: I don't know. @bjfish any idea?

lib/runtime-c-api/src/lib.rs
```diff
@@ -96,6 +96,7 @@ pub mod global;
 pub mod import;
 pub mod instance;
 pub mod memory;
+pub mod metering;
```
Hywan: You can add the attribute `#[cfg(feature = "metering")]` here, and remove all the same attributes in `metering.rs`. I believe it will simplify the code :-).

lib/runtime-c-api/src/metering.rs
```diff
@@ -0,0 +1,111 @@
...
+    new_gas: u64,
+) {
+    use wasmer_middleware_common::metering;
+    let instance = &mut *(instance as *mut wasmer_runtime::Instance);
+    metering::set_points_used(instance, new_gas)
+}
+
+/*** placeholder implementation if metering feature off ***/
+
+// Without metering, wasmer_compile_with_limit is a copy of wasmer_compile
```
Hywan: I'm not sure we need fallbacks. If metering feature is on, then `wasmer_compile_with_limit` will exist, otherwise it will be absent.

lib/runtime-c-api/src/lib.rs
ethanfrey: The problem is that it would expose a different C ABI then, and the go wrapper would fail to compile. Since go doesn't support conditional compilation feature tags as rust does, I think it would break the go bindings to have different ABIs (okay, there is some stuff for architecture, but nothing like rust's feature switches).

I can change that out if this is not a problem, but wanted the least breaking change possible (especially since the abi is committed to the go-ext-wasmer repo)

lib/runtime-c-api/src/metering.rs
ethanfrey: See above point about the bindings - the ABI is copied into the other repo. I feel this option is more a way to merge for now without it being fully tested, then metering can be always on (remove conditions) when it is ready to release.

If this is fully supported in master of `go-ext-wasmer`, then it should be also in the default binary builds of the .so files

lib/runtime-c-api/Cargo.toml
```diff
@@ -24,12 +24,24 @@ default-features = false
...
+optional = true
+
+
 [features]
 default = ["cranelift-backend"]
 debug = ["wasmer-runtime/debug"]
 cranelift-backend = ["wasmer-runtime/cranelift", "wasmer-runtime/default-backend-cranelift"]
 llvm-backend = ["wasmer-runtime/llvm", "wasmer-runtime/default-backend-llvm"]
 singlepass-backend = ["wasmer-runtime/singlepass", "wasmer-runtime/default-backend-singlepass"]
+metering = ["wasmer-runtime/singlepass", "wasmer-middleware-common", "wasmer-singlepass-backend"]
```
Hywan: Since `singlepass-backend` is already a feature, you can simplify the “feature dependency” by writing:

```toml
metering = ["singlepass-backend", "wasmer-middleware-common"]
```

lib/runtime-c-api/src/metering.rs
```diff
@@ -0,0 +1,111 @@
...
+    c
+}
+
+// returns gas used
+#[allow(clippy::cast_ptr_alignment)]
+#[no_mangle]
+#[cfg(feature = "metering")]
+pub unsafe extern "C" fn wasmer_instance_get_points_used(instance: *mut wasmer_instance_t) -> u64 {
+    use wasmer_middleware_common::metering;
+    let instance = &*(instance as *const wasmer_runtime::Instance);
```
Hywan: Check that `instance` isn't a null pointer.

lib/runtime-c-api/src/metering.rs
```diff
@@ -0,0 +1,111 @@
...
+#[allow(clippy::cast_ptr_alignment)]
+#[cfg(feature = "metering")]
+#[no_mangle]
+pub unsafe extern "C" fn wasmer_compile_with_limit(
+    module: *mut *mut wasmer_module_t,
+    wasm_bytes: *mut u8,
+    wasm_bytes_len: u32,
+    gas_limit: u64,
+) -> wasmer_result_t {
+    let bytes: &[u8] = slice::from_raw_parts_mut(wasm_bytes, wasm_bytes_len as usize);
```
Hywan: Check that `module` and `wasm_bytes` aren't null pointers.

lib/runtime-c-api/src/metering.rs
```diff
@@ -0,0 +1,111 @@
...
+// sets gas used
+#[allow(clippy::cast_ptr_alignment)]
+#[no_mangle]
+#[cfg(feature = "metering")]
+pub unsafe extern "C" fn wasmer_instance_set_points_used(
+    instance: *mut wasmer_instance_t,
+    new_gas: u64,
+) {
+    use wasmer_middleware_common::metering;
+    let instance = &mut *(instance as *mut wasmer_runtime::Instance);
```
Hywan: Check that `instance` isn't a null pointer.

lib/runtime-c-api/src/metering.rs
```diff
@@ -0,0 +1,111 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn wasmer_compile_with_limit(
+    module: *mut *mut wasmer_module_t,
+    wasm_bytes: *mut u8,
+    wasm_bytes_len: u32,
+    _: u64,
+) -> wasmer_result_t {
+    let bytes: &[u8] = slice::from_raw_parts_mut(wasm_bytes, wasm_bytes_len as usize);
+    // TODO: this implicitly uses default_compiler() is that proper? maybe we override default_compiler
+    let result = wasmer_runtime::compile(bytes);
```
Hywan: You should coerce to compile with the singlepass backend here for the sake of consistent behaviors.

lib/runtime-c-api/src/metering.rs
```diff
@@ -0,0 +1,111 @@
...
+/// Creates a new Module with gas limit from the given wasm bytes.
+///
+/// Returns `wasmer_result_t::WASMER_OK` upon success.
+///
+/// Returns `wasmer_result_t::WASMER_ERROR` upon failure. Use `wasmer_last_error_length`
+/// and `wasmer_last_error_message` to get an error message.
+#[allow(clippy::cast_ptr_alignment)]
+#[cfg(feature = "metering")]
+#[no_mangle]
+pub unsafe extern "C" fn wasmer_compile_with_limit(
```
Hywan: I would rename the function `wasmer_compile_with_metering` or `_with_gas_metering` for being more explicit.

Hywan:
The fact that only the Singlepass backend support gas metering annoys me. The `go-ext-wasm` language integration works with the Cranelift backend. To support the Singlepass backend, it means that the shared library we share with our CI must embed two backends. Or we should distribute 2 shared libraries: One with the Singlepass backend, and one with the Cranelift backend.

Thoughts @syrusakbary?

lib/runtime-c-api/Cargo.toml
ethanfrey: I did that. It then sets `wasmer-runtime/default-backend-singlepass`.
and the default features set `wasmer-runtime/default-backend-cranelift`.

Setting two default backends caused a panic or compile issue. (I can look up the line if you want). I had to set this to avoid crashes. Ideal would be to disable the cranelift-backend completely for the metered build I guess. But I don't know how to disable default features.

lib/runtime-c-api/src/metering.rs
ethanfrey: Fair enough.
Or should we force the default compiler to single-pass when metering is on, with modified feature flags? Then we can just use default_compiler() consistently everywhere, which probably makes the cleanest code.

When cranelift supports metering, then it swaps out to that as default.

This approach would mean that builds with metering feature will be slower than the other ones, so it should remain a default off feature until cranelift support. But, it would allow me to complete the go bindings and integrations already.

Given the runtime performance of the wasm is not a major issue in most expected smart contracts until we do signature verification or heavy hashing in wasm, singlepass should work great for v1 of the cosmos-sdk integration. And if cranelift+metering work is on your roadmap for the next, say, 6 months, this should be plenty fine for the downstream updates (which would be seemless).

lib/runtime-c-api/src/metering.rs
ethanfrey: Yes, better name

lib/runtime-c-api/src/metering.rs
ethanfrey: To elaborate, cgo will panic on link error (no build) if the linked dll doesn't have the referenced methods. This is why I need stubs (go doesn't have smart feature flags)

ethanfrey:
Also, any idea how to get the C tests to run: eg. https://github.com/wasmerio/wasmer/pull/736/files#diff-105bd7892c64c7216319c0f4f58e0d18 ?

Hywan:
You should update CMake, https://github.com/wasmerio/wasmer/blob/master/lib/runtime-c-api/tests/CMakeLists.txt.

ethanfrey:
789f2ce fixes the features so metering will disable cranelift and enable singlepass as default compiler for a smoother experience. This requires running tests with the `--no-default--features` flag, like: `cargo test --features metering --no-default-features`

This fails on a serialization issue, but notably the same issue exists without metering: `cargo test --features singlepass-backend --no-default-features`. I will check master, but I think singlepass-backend is not properly handled in the c api (although worked fine in pure rust I believe). Guess non-cranelift path is relatively untested here.

lib/runtime-c-api/Cargo.toml
ethanfrey: Also note that singlepass-backend wouldn't even compile without an extra dependency added. Or maybe that is the problem. Hmmm...

lib/runtime/src/cache.rs
ethanfrey: Seems good as is.

lib/runtime-c-api/src/metering.rs
ethanfrey: We always use default compiler now when metering is enabled.
This is hardcoded as llvm now, but could be changed.

ethanfrey:
I addressed all issues and added a section to the README on feature flags.
The c test for testing metering and serialization of metered code is also enabled.

Note that this currently only passes all c tests for:

`cargo test --no-default-features --features llvm-backend,metering`

`test-module-metering-serialize` requires metering to be enabled (and working) and serialization to be working. serialization is known to be broken on singlepass backend (see #811) and gas metering / middleware is known to be broken on the cranelift backend (see #819)

AdamSLevy:
I rebased this PR on the latest master: https://github.com/AdamSLevy/wasmer/tree/metering

details here:
https://github.com/confio/wasmer/pull/1

lib/runtime-c-api/src/lib.rs
AdamSLevy: Golang build tags can be used to select the proper C ABI object to link against.

lib/runtime-c-api/src/metering.rs
AdamSLevy: I don't think that ` wasmer_compile_with_limit` should return without error if the feature is disabled. At least the function should inform the user that it can't enforce a limit.

Hywan:
Hello,

Now that Wasmer 1.0.0-beta2 is released, which contains a new middleware API, do you still need the features from the PR? If yes, can you rebase it, or do you want us to do that?

Note that we now align our C API onto the official standard Wasm C API. Middleware and metering are not part of this standard. That's why we have a vendor-specific API and I think such features will fit there.

ethanfrey:
Oh, we can close this. I stopped using the c api a year ago to write my logic in rust and use custom high level ffi bindings to go. 

I like the new Middleware design btw

PR #737, opened 2019-08-30
Run cargo fmt --all to make the ci linter happy

ethanfrey:
I noticed a CI failure in #736 due to linter. Some errors were my file, but many of these linter errors exist on master.

I just let `cargo fmt --all` fix them up, so the linter is happy on the CI and it is easier to check if a new PR broke something.

### DIFF
--- a/lib/middleware-common-tests/src/lib.rs
+++ b/lib/middleware-common-tests/src/lib.rs
@@ -148,5 +148,4 @@ mod tests {
         // verify it used the correct number of points
         assert_eq!(get_points_used(&instance), 109); // Used points will be slightly more than `limit` because of the way we do gas checking.
     }
-
 }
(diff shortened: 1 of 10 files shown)


MarkMcCaskey:
Hmm, this might be an issue of nightly vs stable linter.  I haven't noticed issues with this on PRs I shipped recently 🤔

ethanfrey:
I am running nightly (as required by SinglePasdBackend). I will try with stable.
Fair enough, the linter complained about this PR as well - mainly line-length.

Maybe there is a line-length setting in a global config?

ethanfrey:
Good call, I went to my other PR, and ran `cargo +stable fmt --all -- --check` and it was only a few points from my code. Not the other packages, good to know.

PR #756, opened 2019-09-05
feat(runtime-c-api) Allow to call `wasmer_export_func_call` with null inputs and a 0 arity

### LINKED ISSUE #753: wasmer_export_func_call 0 arg error
wasmer_export_func_call in wasmer.hh reports "params ptr is null", even if params_len is 0. Seems like it should just ignore the params pointer if params_len is 0.

Currently I'm working around the issue like this:
```
std::vector<wasmer_value_t> params;
// Marshall params...

if (params.size() == 0) {
  // Wasmer requires that our params ptr is valid, even if params_len is 0.
  params.push_back(ToValueI32(0));
}

wasmer_value_t ret;
ThrowIfFailed(wasmer_export_func_call(
      fn->fn, &params[0], fn->wasm_args.size(), &ret, fn->has_ret ? 1 : 0)));
```

Hywan:
Fix #753 

This PR allows to call `wasmer_export_func_call` with inputs set to NULL, and an arity of 0. It avoids to create an empty inputs array. So basically:

```c
wasmer_export_func_call(exported_function, NULL, 0, …, …);
```

instead of:

```c
wasmer_value_t inputs = {};
wasmer_export_func_call(exported_function, inputs, 0, …, …);
```

This PR also updates the test suite for exports. Instead of just the famous `sum` function, we now test more functions, so that all Wasm types are covered, including void functions.

Finally, this PR updates the `wasmer_export_func_call` signature to forbid negative arities.


Hywan:
bors try

lib/runtime-c-api/tests/test-exports.c
```diff
@@ -5,81 +5,483 @@
...
+        wasmer_byte_array name_bytes = wasmer_export_name(export);
+
+        assert(name_bytes.bytes_len == 3);
+
+        char expected[] = {'s', 'u', 'm'};
+
+        printf("Read export name:\n");
+
+        for (uint32_t idx = 0; idx < name_bytes.bytes_len; idx++) {
+            printf("%c\n", name_bytes.bytes[idx]);
```
nlewycky: Since we assert on it anyways, do we need this?

lib/runtime-c-api/tests/test-exports.c
```diff
@@ -5,81 +5,483 @@
...
+
+        uint32_t outputs_arity;
+        wasmer_export_func_returns_arity(exported_function, &outputs_arity);
+
+        assert(inputs_arity == 2);
+        assert(outputs_arity == 1);
+
+        printf("Check signature\n");
+
+        wasmer_value_tag *input_types = (wasmer_value_tag *) malloc(sizeof(wasmer_value_tag) * inputs_arity);
```
nlewycky: Prefer `calloc(y, sizeof(x))` to `malloc(sizeof(x) * y)`. It handles multiplication overflow better.

lib/runtime-c-api/tests/test-exports.c
```diff
@@ -5,81 +5,483 @@
...
+        wasmer_export_func_params(exported_function, input_types , inputs_arity);
```
nlewycky: Extra space after `input_types` before the comma.

lib/runtime-c-api/tests/test-exports.c
```diff
@@ -5,81 +5,483 @@
...
+        wasmer_import_export_kind export_kind = wasmer_export_kind(export);
+
+        assert(export_kind == WASM_FUNCTION);
+
+        const wasmer_export_func_t *exported_function = wasmer_export_to_func(export);
+        wasmer_byte_array name_bytes = wasmer_export_name(export);
+
+        assert(name_bytes.bytes_len == 7);
+
+        char expected[] = {'a', 'r', 'i', 't', 'y', '_', '0'};
```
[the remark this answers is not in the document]
Hywan: `sizeof("arity_0") - 1` actually (`wasmer_byte_array` contains a string that is not null-terminated). But thanks for the tips!

lib/runtime-c-api/tests/test-exports.c
Hywan: We don't. The first example is just here to print some stuff, but that's pretty useless (it was taken fromt the previous version).

lib/runtime-c-api/tests/test-exports.c
Hywan: Correct!

lib/runtime-c-api/tests/test-exports.c
Hywan: Good eyes, thanks!

Hywan:
:+1:

PR #783, opened 2019-09-12
Remove memory_base and memory_bound from InternalCtx.

nlewycky:
# Description
Remove the `memory_base` and `memory_bound` members on `InternalCtx`. These are redundant with the context's list of memories which also include a base and a bound. Use those instead.

# Review

### DIFF
--- a/lib/singlepass-backend/src/codegen_x64.rs
+++ b/lib/singlepass-backend/src/codegen_x64.rs
@@ -33,6 +33,7 @@ use wasmer_runtime_core::{
         TableIndex, Type,
     },
     vm::{self, LocalGlobal, LocalTable, INTERNALS_SIZE},
+    vmcalls,
 };
 use wasmparser::{Operator, Type as WpType, TypeOrFuncType as WpTypeOrFuncType};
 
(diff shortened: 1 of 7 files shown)


nlewycky:
bors try

nlewycky:
> Er, I can't see any information about bors' test failure?

@MarkMcCaskey shows that I have click the ✖ on the line with "Try #783:".

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -4075,26 +4113,36 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
-                    Location::Memory(
-                        Machine::get_vmctx_reg(),
-                        vm::Ctx::offset_intrinsics() as i32,
-                    ),
-                    Location::GPR(GPR::RAX),
-                );
-                a.emit_mov(
-                    Size::S64,
-                    Location::Memory(GPR::RAX, vm::Intrinsics::offset_memory_grow() as i32),
+                    Location::Imm64(func_value as _),
```
losfair: The change to use hardcoded function pointers for `memory_grow` would break `kernel-wasm` / in general any future features that require the JIT-generated code to be independent of its current runtime environment.

Maybe we can add a new compiler option to specify whether the code needs to be both position independent and "environment independent"?

nlewycky: > In general, I think we should remove the `imported_memories` and `imported_globals` fields in `Ctx` and merge them into `memories`/`globals`. That helps if we need to use the code in a position-independent way.

Yes, agreed. The spec even has text like `Globals are referenced through global indices, starting with the smallest index not referencing a global import.` which would suggest that we hold them in a list with imported globals first, followed by local globals. That seems like it could go into a future PR.

nlewycky: > The change to use hardcoded function pointers for `memory_grow` would break `kernel-wasm` / in general any future features that require the JIT-generated code to be independent of its current runtime environment.
> 
> Maybe we can add a new compiler option to specify whether the code needs to be both position independent and "environment independent"?

How does kernel-wasm work? Does it provide its own context?

For what purpose do we need position independent code?

In what circumstances is generated code expected to work with a different Ctx? The correctness of the code (and its security properties) are guaranteed in part by the static analysis done by the verification step. For instance, you can't generate code that accesses memory in the wasm module, then run it on a different context that has no memory. What if we start up in a new Ctx which matches the function, but the `memory.grow` function hasn't been called to grow the memory yet?

I think we can take a generated function and serialize it to deserialize it with the same version of `wasmer`. What about a newer version of wasmer? What kinds of changes are guaranteed to work vs. which will break? And do we have tests for this (for each release of wasmer, generate a bunch of OSR-save states to disk, commit them to tree, try loading them in future releases)?

My thinking is that, for now, a function is completely tied to the Ctx that was used to generate it, and that the Ctx is `Pin` in memory. We don't presently use the Pin type, but we do pass the pointer all around, so I wouldn't expect moving the Ctx to work.

losfair: `kernel-wasm` works by copying the generated code and the related state/specs (memory, globals, tables, etc. ) into the Linux kernel. The kernel then re-generates a `Ctx` using the provided state and specs, and calls WASM functions with the new `Ctx` as the first argument.

Besides `kernel-wasm` I can also imagine other situations where similar things can happen. Basically if we ever want to run the compiler and runtime in different process contexts, the generated code needs to be position independent and also independent of the value of `Ctx`.

The correctness of the structure of `Ctx` should be ensured by its creator.

I think that, we can generate position independent code without any assumptions about `Ctx` details initially in the *compiler*, and afterwards the *runtime* where the code actually runs in can perform optimizations like embedding the value of Ctx directly into the code.

The runtime state currently saved by the managed/"OSR" engine is very trivial - just a list of stack frames including all stack values and locals, plus global states. Given the serialization encoding (currently `bincode`) doesn't change, it is unlikely that saved images are incompatible across versions. But in case of breaking changes (like we added something to the serialized image), we can also prepend a version number to the header of the image. There's no test yet for this part, but it's good to have some.

nlewycky: Thanks! I've tried writing one long reply and it got out of hand, so I'm going to answer in pieces.

I don't think it makes sense to maintain a design separation between the compiler and the runtime. We can make all sorts of changes to `Ctx`, we might change whether memories are listed as vector(pair(base, bound)) to pair(vector(base), vector(bound)), or we might use variable bit rate encoding to shrink the number of bytes it takes to encode the bounds, or we might encode the bounds in the top bits of the base pointer, a common trick in pooled malloc implementations. Simply asking the runtime "what is the offset to the memory bounds" isn't fully flexible. Similarly we might change how AnyFunc works, or any number of things.

We could simply decide not to make those kinds of changes. The question then is, what's the benefit of making sure we have an ABI layer between the compiler and runtime?

nlewycky: As for embedding the value of `Ctx` in the function after the fact, I think if we wanted to do that, the best way would be to emit the true value of `Ctx` directly into the instructions, but also maintain a side-table, the relocation table. That way we emit the code in a single pass for singlepass, and we have the information we need to relocate the code, and we pay the cost of applying fixups at the time we make the copy.

nlewycky: The only time I can see that we need position-independent code is if we don't know the final address the code will live in after it's generated. Even across multiple processes, is there any reason we can't design the API such that that the process that will receive our code must reserve memory first and tell us the address to use?

I'm reasonably confident I can make this work for kernel-wasm too, though I understand if you want me to demonstrate that in the form of a PR. Let me know.

syrusakbary:
I believe this is no longer needed in the refactor, so we should be good to close the PR for now.
Let me know (or reopen the PR) if that's not the case @nlewycky

PR #785, opened 2019-09-13
Copy the spectest repository license from the official WebAssembly testsuite. These tests are under that license.

nlewycky:
Remove our own license statements on a few files since they're now redundant. They were all referencing this license.


Hywan:
Can you create the `CHANGELOG` entry please?

PR #787, opened 2019-09-13
feat(interface-types) Implement Interface Types (WIT)

Hywan:
So far, this PR implements a new crate named `wasmer-interface-types`. The goal is to implement the WebAssembly Interface Types (WIT) proposal, https://github.com/WebAssembly/interface-types. Keep in mind the proposal is highly unstable and experimental.

The `wasmer-interface-types` is designed to be _runtime agnostic_, which means that there is no connection between this crate and `wasmer-runtime-core` or `wasmer-runtime` (or any backend crates). The hope is that other runtimes could use `wasmer-interface-types` according to their needs.

The `wasmer-interface-types` is composed of 4 parts:
* AST: [The WIT language, represented as an AST](https://github.com/Hywan/wasmer/blob/feat-interface-types/lib/interface-types/src/ast.rs) (it's not really abstract though),
* Decoders: So far, it contains only [one parser that reads the WIT binary language](https://github.com/Hywan/wasmer/blob/feat-interface-types/lib/interface-types/src/decoders/binary.rs) and produces an AST,
* Encoders: So far, it contains only [one compiler that transforms the AST into "WIT.wat”](https://github.com/Hywan/wasmer/blob/feat-interface-types/lib/interface-types/src/encoders/wat.rs), i.e. the textual representation of an Interface Types AST
* Instructions: WIT defines a new concept called Adapters. An adapter contains a set of instructions. Those instructions are executed when data are bound. So, in more details, this module contains:
  * [`interpreter.rs`](https://github.com/Hywan/wasmer/blob/feat-interface-types/lib/interface-types/src/interpreter/mod.rs), a stack-based interpreter, defined by:
    * A compiler that transforms a set of instructions into a set of _executable_ instructions,
    * A stack,
    * A runtime that holds the “invocation inputs” (arguments of the interpreter), the stack, and the Wasm instance (which holds the exports, the imports, the memories, the tables etc.),
  * [`stack.rs`](https://github.com/Hywan/wasmer/blob/feat-interface-types/lib/interface-types/src/interpreter/stack.rs), a very light and generic stack implementation, exposing only the API required by the interpreter,
  * [`wasm/*.rs`](https://github.com/Hywan/wasmer/blob/feat-interface-types/lib/interface-types/src/interpreter/wasm/structures.rs), a set of enums, types, and traits to represent a Wasm runtime —basically this is the part a runtime should take a look to support the `wasmer-interface-types` crate—.

Progression:

* [x] WIT binary parser,
* [x] WIT.wat compiler,
* [ ] Interpreter:
  * [ ] Wasm runtime traits to represent:
    * [x] Instance,
    * [x] Export,
    * [x] Memory,
    * [x] Import,
    * [ ] Table,
  * [x] Generic stack implementation,
  * [ ] Implemented instructions:
    * [x] `arg.get`,
    * [x] `call`,
    * [x] `call-export`,
    * [x] `read-utf8`,
    * [x] `write-utf8`,
    * [ ] `as-wasm`,
    * [ ] `as-interface`
    * [ ] `table-ref-add`
    * [ ] `table-ref-get`
    * [ ] `call-method`
    * [ ] `make-record`
    * [ ] `get-field`
    * [ ] `const`
    * [ ] `fold-seq`
    * [ ] `add`
    * [ ] `mem-to-seq`
    * [ ] `load`
    * [ ] `seq.new`
    * [ ] `list.push`
    * [ ] `repeat-until`
    * [ ] `seq-to-mem`
    * [ ] `store`
* [ ] Plug `wasmer-interface-types` into `wasmer-runtime-core`


lib/interface-types/src/decoders/binary.rs
```diff
@@ -0,0 +1,734 @@
...
+    }
+
+    let length = input[0] as usize;
+    let mut input = &input[1..];
+
+    if input.len() < length {
+        return Err(Err::Error(make_error(input, ErrorKind::Eof)));
+    }
+
+    let mut items = vec![];
```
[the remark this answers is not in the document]
Hywan: Yup, thanks!

lib/interface-types/src/instructions/interpreter.rs
```diff
@@ -135,7 +141,7 @@ where
                                                 }
                                             }
                                             None => Err(format!(
-                                                "`{}` cannot call the exported function `{}` because there is no enought data in the stack for the arguments (needs {}).",
+                                                "`{}` cannot call the exported function `{}` because there is no enough data on the stack for the arguments (needs {}).",
```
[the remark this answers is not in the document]
Hywan: Haha :-p, good eyes!

lib/interface-types/src/lib.rs
```diff
@@ -0,0 +1,162 @@
...
+    use std::fs;
+    use wasmer_clif_backend::CraneliftCompiler;
+    use wasmer_runtime_core as runtime;
+
+    fn get_module() -> runtime::Module {
+        runtime::compile_with(
+            fs::read("tests/assets/hello_world.wasm")
+                .expect("Failed to read `tests/assets/hello_world.wasm`.")
+                .as_slice(),
+            &CraneliftCompiler::new(),
```
syrusakbary: Can we use `compile` instead of `compile_with` ?

```rust
        runtime::compile(
            fs::read("tests/assets/hello_world.wasm")
                .expect("Failed to read `tests/assets/hello_world.wasm`.")
                .as_slice()
        )
```

Hywan: I'm not sure those tests will remain here. I'll keep your comment open for the moment.

lib/interface-types/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+authors = ["The Wasmer Engineering Team <engineering@wasmer.io>"]
+repository = "https://github.com/wasmerio/wasmer"
+edition = "2018"
+
+[dependencies]
+nom = "5.0"
+
+[dev-dependencies]
+wasmer-runtime-core = { path = "../runtime-core", version = "0.7.0", features = ["backend-cranelift"] }
+wasmer-clif-backend = { path = "../clif-backend", version = "0.7.0" }
```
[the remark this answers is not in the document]
Hywan: I'm not sure I'll keep these dependencies for the tests. It will probably move directly in `wasmer-runtime-core`. Let's see. For the moment, I keep your comment open!

lib/interface-types/src/decoders/binary.rs
```diff
@@ -0,0 +1,940 @@
...
+    }
+
+    Ok((input, adapters))
+}
+
+/// Parse a list of forwarded exports.
+fn forwards<'input, E: ParseError<&'input [u8]>>(
+    input: &'input [u8],
+) -> IResult<&'input [u8], Vec<Forward>, E> {
+    let mut input = input;
```
MarkMcCaskey: ⛳️ I believe you can just put the `mut` in the argument list, like `mut input: &'input [u8],`

lib/interface-types/src/encoders/wat.rs
```diff
@@ -0,0 +1,716 @@
...
+//! # }
+//! ```
+
+use crate::{
+    ast::{Adapter, Export, Forward, Import, InterfaceType, Interfaces, Type},
+    interpreter::Instruction,
+};
+
+/// Encode an `InterfaceType` into a string.
+impl From<&InterfaceType> for String {
```
MarkMcCaskey: Consider using the [ToString trait](https://doc.rust-lang.org/std/string/trait.ToString.html) here (both is fine, but this seems to be exactly what `ToString` was created for)

lib/interface-types/src/encoders/wat.rs
```diff
@@ -0,0 +1,716 @@
...
+}
+
+/// Encode a list of `InterfaceType` representing inputs into a
+/// string.
+fn input_types_to_param(input_types: &[InterfaceType]) -> String {
+    if input_types.is_empty() {
+        "".into()
+    } else {
+        format!(
+            "\n  (param{})",
```
MarkMcCaskey: the fixed new line and indentation seems a bit inflexible but perhaps it's sufficient here!

lib/interface-types/src/encoders/wat.rs
```diff
@@ -0,0 +1,716 @@
...
+            Instruction::SeqNew(interface_type) => {
+                format!("seq.new {}", String::from(interface_type))
+            }
+            Instruction::ListPush => "list.push".into(),
+            Instruction::RepeatUntil(condition_index, step_index) => {
+                format!("repeat-until {} {}", condition_index, step_index)
+            }
+        }
+    }
+}
```
MarkMcCaskey: If the speed is ever a concern here we can implement these `X -> String` functions of a `&mut T` where `T: Write` to avoid the extra allocations.  Very low priority right now (and until the speed becomes a major concern), just writing this for future reference.

lib/interface-types/src/encoders/wat.rs
```diff
@@ -0,0 +1,605 @@
...
+            name = export.name,
+            inputs = input_types_to_param(&export.input_types),
+            outputs = output_types_to_result(&export.output_types),
+        )
+    }
+}
+
+impl<'input> From<&Type<'input>> for String {
+    fn from(_ty: &Type) -> Self {
+        unimplemented!()
```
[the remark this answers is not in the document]
MarkMcCaskey: still unimplemented 🤔

lib/interface-types/src/encoders/wat.rs
```diff
@@ -0,0 +1,716 @@
...
+                    instructions
+                        .iter()
+                        .fold(String::new(), |mut accumulator, instruction| {
+                            accumulator.push_str("\n  ");
+                            accumulator.push_str(&String::from(instruction));
+                            accumulator
+                        }),
+            ),
+
+            _ => unimplemented!(),
```
MarkMcCaskey: I'm a big advocate of always putting a string in `unimplemented`/`todo` so that it's easy to find and prints a message that a user can consume

lib/interface-types/src/encoders/wat.rs
```diff
@@ -0,0 +1,716 @@
...
+                        } => accumulator.push_str(&format!(
+                            "\n\n;; Interface, Adapter {}.{}\n",
+                            namespace, name
+                        )),
+
+                        Adapter::Export { name, .. } => {
+                            accumulator.push_str(&format!("\n\n;; Interface, Adapter {}\n", name))
+                        }
+
+                        _ => unimplemented!(),
```
MarkMcCaskey: Here too, any string is fine really, but ideally it should be identifiable and as helpful as it can be

lib/runtime-core/Cargo.toml
```diff
@@ -52,3 +52,4 @@ cc = "1.0"
 [features]
 managed = []
 deterministic-execution = ["wasmparser/deterministic"]
+interface-types = []
```
MarkMcCaskey: This seems to be unused 🤔

lib/interface-types/src/interpreter/instructions/call.rs
```diff
@@ -0,0 +1,187 @@
...
+    test_executable_instruction!(
+        test_call__invalid_local_import_index =
+            instructions: [
+                Instruction::Call { function_index: 42 },
+            ],
+            invocation_inputs: [
+                InterfaceValue::I32(3),
+                InterfaceValue::I32(4),
+            ],
+            instance: Instance { ..Default::default() },
```
MarkMcCaskey: ⛳️ `instance: Default::default()`

lib/interface-types/src/interpreter/wasm/structures.rs
```diff
@@ -0,0 +1,157 @@
+use super::values::{InterfaceType, InterfaceValue};
+use std::{cell::Cell, ops::Deref};
+
+pub trait TypedIndex: Copy + Clone {
+    fn new(index: usize) -> Self;
+    fn index(&self) -> usize;
+}
```
MarkMcCaskey: It would be good to document the reason why we can't use the runtime-core version of this trait!

lib/interface-types/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#![deny(
+    dead_code,
+    nonstandard_style,
+    unused_imports,
+    unused_mut,
+    unused_variables,
+    unused_unsafe,
+    unreachable_patterns
+)]
```
MarkMcCaskey: Could add deny missing docs to keep things documented!  Looks like we're close (we don't have to do this now)

We probably want to remove the 1 use of `unsafe` in this crate (I don't think we have any guarantees that we actually have utf8 bytes there) and once we do that, we can add `#![forbid(unsafe_code)]` as good practice (motivated by https://github.com/anderejd/cargo-geiger )

lib/interface-types/src/ast.rs
```diff
@@ -0,0 +1,173 @@
...
+#[derive(PartialEq, Debug)]
+pub struct Type<'input> {
+    /// The type name.
+    pub name: &'input str,
+
+    /// The field names.
+    pub fields: Vec<&'input str>,
+
+    /// The field types.
+    pub types: Vec<InterfaceType>,
```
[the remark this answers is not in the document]
Hywan: That's is very true. I'm updating that!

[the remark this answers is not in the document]
Hywan: Addressed in https://github.com/wasmerio/wasmer/pull/787/commits/b7941f401957a831a7e1d15003888d8290bd6e23.

lib/interface-types/src/ast.rs
```diff
@@ -0,0 +1,173 @@
...
+        namespace: &'input str,
+
+        /// The function name.
+        name: &'input str,
+
+        /// The function input types.
+        input_types: Vec<InterfaceType>,
+
+        /// The function output types.
+        output_types: Vec<InterfaceType>,
```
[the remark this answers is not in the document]
Hywan: What do you mean by:

> It's worth considering using the structs with the same fields inside this enum if you haven't already.

?

lib/interface-types/src/ast.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        /// The helper output types.
+        output_types: Vec<InterfaceType>,
+
+        /// The instructions of the adapter.
+        instructions: Vec<Instruction<'input>>,
+    },
+}
+
+/// Represented a forwarded export.
```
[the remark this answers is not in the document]
Hywan: It's not super defined in the spec yet :-D. It should be used to “rename” an export, like `main` to `_start`. I'm not sure it will be kept in the future.

lib/interface-types/src/decoders/binary.rs
```diff
@@ -0,0 +1,940 @@
...
+/// Parse an unsigned LEB with value no larger than a 64-bits number.
+fn leb<'input, E: ParseError<&'input [u8]>>(input: &'input [u8]) -> IResult<&'input [u8], u64, E> {
+    if input.is_empty() {
+        return Err(Err::Error(make_error(input, ErrorKind::Eof)));
+    }
+
+    let (output, bytes) = match input.iter().position(|&byte| byte & 0x80 == 0) {
+        Some(position) => (&input[position + 1..], &input[..=position]),
+        None => (&[] as &[u8], input),
+    };
```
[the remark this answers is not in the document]
Hywan: Yeah, it's like `LEB128` but for 64-bits numbers. It's a Little Endian Based unsigned integer. I'll update the documentation.

Check https://en.wikipedia.org/wiki/Variable-length_quantity and https://en.wikipedia.org/wiki/LEB128 for more context.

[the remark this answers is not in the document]
Hywan: Addressed by https://github.com/wasmerio/wasmer/pull/787/commits/1ad42d81cb7b39240adb65247bea942603e8cb5f.

lib/interface-types/src/decoders/binary.rs
```diff
@@ -0,0 +1,940 @@
...
+
+    Ok((
+        output,
+        bytes
+            .iter()
+            .rev()
+            .fold(0, |acc, byte| (acc << 7) | u64::from(byte & 0x7f)),
```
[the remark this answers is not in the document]
Hywan: That is true that we don't detect overflow here. Let's fix that.

Side note: There is a crate to handle such encoding, https://github.com/gimli-rs/leb128/.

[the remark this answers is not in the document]
Hywan: Addressed in https://github.com/wasmerio/wasmer/pull/787/commits/370fd6d86675d3255dbbed3f012fecb39a28d34a.

lib/interface-types/src/decoders/binary.rs
```diff
@@ -0,0 +1,940 @@
...
+
+/// Parse a UTF-8 string.
+fn string<'input, E: ParseError<&'input [u8]>>(
+    input: &'input [u8],
+) -> IResult<&'input [u8], &'input str, E> {
+    if input.is_empty() {
+        return Err(Err::Error(make_error(input, ErrorKind::Eof)));
+    }
+
+    let length = input[0] as usize;
```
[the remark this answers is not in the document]
Hywan: Good idea, but tests are failing. Since it's minor, I don't investigate more than that.

lib/interface-types/src/decoders/binary.rs
```diff
@@ -0,0 +1,734 @@
...
+    let input = &input[1..];
+
+    if input.len() < length {
+        return Err(Err::Error(make_error(input, ErrorKind::Eof)));
+    }
+
+    Ok((&input[length..], unsafe {
+        str::from_utf8_unchecked(&input[..length])
```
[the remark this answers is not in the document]
Hywan: The safe crate still needs the nightly version of rustc. I'm adding a comment.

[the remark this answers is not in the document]
Hywan: Actually, I've removed the `unsafe` block. Addressed in https://github.com/wasmerio/wasmer/pull/787/commits/3f0c32bf44bfb735749ed20a17eeac82884d8792.

lib/interface-types/src/decoders/binary.rs
Hywan: True! Addressed in https://github.com/wasmerio/wasmer/pull/787/commits/d8f923091f89686faa99fc7a8af378873f86c514.

lib/interface-types/src/encoders/wat.rs
Hywan: TIL about `ToString`, thanks!

Addressed in https://github.com/wasmerio/wasmer/pull/787/commits/13cee90d17620c76512f315a061d6d5442f11bbd.

lib/interface-types/src/encoders/wat.rs
Hywan: True. Just wanted a very simple version for debugging purposes first.

lib/interface-types/src/encoders/wat.rs
Hywan: It is!

lib/interface-types/src/encoders/wat.rs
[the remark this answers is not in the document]
Hywan: Because `Type` isn't implemented for the moment. Let's address in the future. Changing for `todo!()` with a note.

lib/interface-types/src/encoders/wat.rs
Hywan: Done!

lib/interface-types/src/encoders/wat.rs
Hywan: Done!

lib/interface-types/src/interpreter/instructions/call.rs
Hywan: Hehe, yup!

lib/interface-types/src/interpreter/wasm/structures.rs
Hywan: Because `wasmer-interface-types` is designed to be runtime agnostic, so with no connection with Wasmer (except in the name), so that it can be reused by other runtimes (I'm probably dreaming here 😇).

lib/interface-types/src/lib.rs
Hywan: `deny(missing_docs)` added!

Hywan: `forbid(unsafe_code)` added!

lib/runtime-core/Cargo.toml
Hywan: Oops, that's an artifact. Removing.

lib/interface-types/src/ast.rs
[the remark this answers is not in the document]
MarkMcCaskey: Most of the enum's fields are identical to the structs with the same/similar names.  It'd be possible to remove most of the fields in the enum and replace it with 1 instance of the correct struct.  It's a minor thing and I think there are convincing arguments for both doing it and not doing it.

PR #790, opened 2019-09-13
Switch llvm-backend from PIC+Default (small) to Static+Large.

nlewycky:
# Description

Should fix flaky test failure:
```
spectest-a277b7d051ac0dde: /home/nick/llvm-8.0.1.src/lib/ExecutionEngine/RuntimeDyld/RuntimeDyldELF.cpp:304: void llvm::RuntimeDyldELF::resolveX86_64Relocation(const llvm::SectionEntry&, uint64_t, uint64_t, uint32_t, int64_t, uint64_t): Assertion `isInt<32>(RealOffset)' failed.
```
(with assertions-enabled LLVM). The relocation in question is a `R_X86_64_PC32` with small offset against `.rodata`.

The switch from Default (small, JITDefault is a different default that we weren't using which is medium) to Large should fix the bug. Once we're using large code model, I think we can switch from PIC to static as either an optimization or a no-op.

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Blocks of changes will separated by version increments.
 
 ## **[Unreleased]**
 
+- [#790](https://github.com/wasmerio/wasmer/pull/790) Fix flaky test failure with LLVM, switch to large code model.
 - [#788](https://github.com/wasmerio/wasmer/pull/788) Use union merge on the changelog file.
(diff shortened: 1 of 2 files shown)


nlewycky:
bors try

nlewycky:
bors try

nlewycky:
On Slack, @losfair suggested that maybe the functions in `intrinsics.rs` are assumed by LLVM to be in the same code region and it uses PC-relative addressing for them. An alternative would be to add pointers to the VM functions to the context, then perform indirection through them.

PR #792, opened 2019-09-15
Remove colored CLI output from wasmer-runtime-core lib

repi:
When the colored output was originally added in https://github.com/wasmerio/wasmer/pull/489 and there was a discussion then about that it should ideally be in a higher-level crate rather than in the runtime-core library crate.

I agree with that, users of the library shouldn't be required to bring in the `colored` crate dependency and ideally also not have stdout/stderr output either, that should be controlled by the application that uses wasmer-runtime-core, not the library.

Disabling stdout/stderr output would be more intrusive but I wanted to at least not have colored output and another crate dependency so this change removes the colored output and the `colored` crate.

`colored` also had quite a few dependencies and, while well used, is not super actively maintained. So this change also removes 6 transitive dependencies of the `colored` crate which is great.

This could potentially be a feature flag instead also, but would be a bit messier.


syrusakbary:
@repi if you can update the cargo.lock it would be awesome, so we can merge!

repi:
Great, there now `Cargo.lock` config has been resolved

Issue #670, opened 2019-08-13
Callbacks between Wasm and Native code


tprk77:
### Motivation

I would like to use callbacks between Wasm and native code. I would like to implement a callback on the Wasm-side, pass it as an argument, and later invoke it from the native-side.

I basically want to do what's in this [StackOverflow question](https://stackoverflow.com/q/45387728), but with Wasmer and Rust.

### Proposed solution

On the native-side, which is embedding the VM, I want to receive a pointer to a function implemented on the Wasm-side, and somehow lookup and dispatch that function. I would like to be able to do this without any shims or exported functions.

It might look something like this:

```
// set_callback: (callback: extern fn(x: u32) -> ()) -> ()
pub fn set_callback(ctx: &mut Ctx, callback_ptr: u32) {
  // Somehow lookup callback_ptr in a table?
  ctx.get_func(callback_ptr).call(123);
}
```

### Additional context

I've attempted #245, but I haven't made any progress there. (Help please!)

I've implemented callbacks in my own [demo project](https://github.com/tprk77/experiments_in_ros2_rust_and_wasm), but it required some pretty undesirable work-arounds, like a export function shim.

There was also some [discussion on Spectrum](https://spectrum.chat/wasmer/general/is-it-possible-to-call-a-wasm-callback-from-rust~eee307f9-fdf9-4182-876c-59cf3395794c).

MarkMcCaskey:
I believe this is the same as or very similar to #638 ;

Thanks for the report -- I agree that this is an important thing to be able to do!

MarkMcCaskey:
I got this working in #803 with an example!!

MarkMcCaskey:
Closing because #803 merged in.  If you have any issues, don't hesitate to reopen this issue or file a new one!

tprk77:
This weekend I'll have time to integrate this change into my project. I'll let you know how it goes!

Issue #638, opened 2019-08-07
Ctx needs ability to invoke functions


autodidaddict:
### Motivation

My project has a wasmer runtime embedded in it, and my use case involves the host runtime invoking a guest module function. In response, the guest module can invoke multiple host functions. Within those host functions, I might need to make further guest function calls to interrogate data.

tl;dr host functions that are free-floating (not bound to a struct instance) have no way to invoke functions in the module, though they can read/write the module's memory block.

### Proposed solution

Add the `func` call to the `Ctx` struct, or add a reference to the `instance` that invoked the host function to the signature of the function pointed at by the `func!` macro.

### Alternatives

The only alternatives we have are trying to store global references to the `Instance`, but since the `Instance` triggers all kinds of thread safety errors as soon as we try and do that, it's not a viable option. Right now it appears that host functions _cannot_ call guest functions, which severely inhibits some RPC-style call patterns.

### Additional context

I'm looking at this functionality as a way to simplify the `wascap` interface standard to be more fluid and to avoid explicit cross-wasm boundary calls to allocate memory (decoupling allocation from the RPC interface). I can't do this if the host functions can't make guest function calls.

MarkMcCaskey:
I've created an example that does this, but unless I'm misunderstanding the spec, then there's an issue with the Wasm Rustc is generating https://github.com/rust-lang/rust/issues/64521 .

The only change needed is for `runtime-core::instance::call_func_with_index` to be public.

I've confirmed that Wasm and host fn callbacks now work with all 3 backends.

The issue is that given what Rustc currently does, it's not possible to tell the difference between a host fn and a wasm fn.  And not only that, but if the Rust code tries to do anything with a host-fn as a callback, it clobbers all the other indices making them invalid.

The way I'm doing it here is closer to #670, passing the functions directly.  I agree that we should have other ways to do this like this issue mentions!

MarkMcCaskey:
Okay!  Update on this front:

We have a PR out for making Instance Send (#807) which may resolve some of your issues, but it sounds like those might be a work around anyways.  #803 merged which allows calling functions via the table.  In order to get the table mapping, you'd have to get data from the Wasm side.  Is this something that seems feasible for your use case?  I think the biggest missing piece now that I think about it is that the change I made requires you to pass in arguments and doesn't let the caller query the signature of the function.  If that's an issue, I can prioritize looking into that, too.

thedavidmeister:
this seems really fundamental to me

[...]

PR #803, opened 2019-09-16
Add method to call function at index on Ctx

MarkMcCaskey:
For #638 and #670

```Rust
fn call_guest_fn(ctx: &mut Ctx, guest_fn: u32) -> u32 {
    println!("{}", guest_fn);

    let guest_fn_typed = unsafe { std::mem::transmute(guest_fn) };

    let result = ctx.call_with_table_index(guest_fn_typed, &[]).unwrap();
    println!("  -> {:?}", result);

    0
}
```
is what this looks like from the Host side

See `examples/callback.rs` for an example that doesn't require `transmute`

# Review


tprk77:
This looks awesome! I'm looking forward to using this.

PR #805, opened 2019-09-17
Replace panic! & unimplemented! in runtime-code and llvm-backend

pventuzelo:
# Description

Replace `unimplemented!` by already used `CodegenError` in `lib/llvm-backend/src/code.rs`
Replace `unimplemented!` by `Err` in `lib/llvm-backend/src/trampolines.rs`
Replace `panic!` by already used `BinaryReaderError` in `lib/runtime-core/src/parse.rs`

# Review

- [ ] Create a short description of the the change in the CHANGELOG.md file

### DIFF
--- a/lib/llvm-backend/src/code.rs
+++ b/lib/llvm-backend/src/code.rs
@@ -1176,7 +1176,11 @@ impl FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator {
                             BasicTypeEnum::FloatType(float_ty) => {
                                 float_ty.const_float(0.0).as_basic_value_enum()
                             }
-                            _ => unimplemented!(),
+                            _ => {
+                                return Err(CodegenError {
+                                    message: "Operator::End phi type unimplemented".to_string(),
+                                });
+                            }
                         };
                         state.push1(placeholder_value);
                         phi.as_instruction().erase_from_basic_block();
@@ -1741,7 +1745,12 @@ impl FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator {
                             _ => value,
                         });
                     }
-                    _ => unimplemented!("multi-value returns"),
+                    _ => {
+                        return Err(CodegenError {
+                            message: "Operator::CallIndirect multi-value returns unimplemented"
+                                .to_string(),
+                        });
+                    }
                 }
             }
 
@@ -6853,7 +6862,9 @@ impl FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator {
                 state.push1(result.try_as_basic_value().left().unwrap());
             }
             _ => {
-                unimplemented!("{:?}", op);
+                return Err(CodegenError {
+                    message: format!("Operator {:?} unimplemented", op),
+                });
             }
         }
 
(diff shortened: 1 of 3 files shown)


lib/llvm-backend/src/code.rs
```diff
@@ -1176,7 +1176,11 @@ impl FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator {
                             BasicTypeEnum::FloatType(float_ty) => {
                                 float_ty.const_float(0.0).as_basic_value_enum()
                             }
-                            _ => unimplemented!(),
```
nlewycky: `unreachable!()` or `unreachable!("Push PHI node of type {:?}", basic_ty)`

lib/llvm-backend/src/code.rs
```diff
@@ -1741,7 +1745,12 @@ impl FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator {
                             _ => value,
                         });
                     }
-                    _ => unimplemented!("multi-value returns"),
```
nlewycky: Do you really need this change? AFAICT this is a correct use of unimplemented. The WASM feature "multi-value returns" really is unimplemented in our VM (indeed, not just here, but throughout).

(Applies to the other mentions of multi-value return in this change.)

lib/llvm-backend/src/code.rs
```diff
@@ -6853,7 +6862,9 @@ impl FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator {
                 state.push1(result.try_as_basic_value().left().unwrap());
             }
             _ => {
-                unimplemented!("{:?}", op);
```
nlewycky: `unreachable!("Operator {:?} not handled", op);`

lib/llvm-backend/src/code.rs
pventuzelo: You totally right, it's a legitimate use of `unimplemented!`. 

I decided to replaced them because the function already return `Result` with `CodegenError`.
The cost for us to replace them by a returning error is nothing compared to having wasmer panicking if we use `unimplemented!` or even `unreachable!`.

nlewycky: I know these are unreachable, the parser is not supposed to accept multi-value return yet because it is an extension and we have not enabled it. It's a serious bug if this is reached in normal usage.

The benefit to keeping them is that when someone comes along to implement multi-value return (or another unimplemented feature) these are easily marked as such.

Can I ask what the downside is of letting this panic? It's bad for embedders if a library it's using panics, but as I stated that really shouldn't happen (or, if it does, I think this is worth panicking over!). Is it causing problems for the fuzzing effort?

lib/llvm-backend/src/code.rs
pventuzelo: This one should return an Error i think since we are not supporting all new opcode in future proposal.

lib/llvm-backend/src/code.rs
pventuzelo: Ok make sense, in that case i will convert `unimplemented!` to `unreachable` only for the ones related to multi-value return.

Regarding fuzzing, a panic generate an exit status `101` and an Error will be an exit status `1`.
This difference impact the bug classification of some fuzzers and make bugs triaging more complicated. Also, having too much panic triggered reduce the fuzzing speed since a panic will force the fuzzer to kill the current fuzzing thread and create a new one.

lib/llvm-backend/src/code.rs
nlewycky: OK.

PR #807, opened 2019-09-17
Implement Send for Instance

### LINKED ISSUE #748: Instance should implement Sync and Send
To be able to run Wasmer from different threads Instance will need to implement `Sync` and `Send`.

Related issue: https://github.com/amethyst/amethyst/issues/1729

MarkMcCaskey:
# Review

Resolves #748 

WIP

### Commit 1
- `Global`s use Arc instead of RC
- Export `Context` and `FuncPointer` manually implement Send
- `ImportObject` uses `Arc<Mutex<HashMap<...>>>` Instead of `Rc<RefCell<HashMap<...>>>`; removed `get_namespace` method in favor of continuation style to deal with locking the Mutex
- `Func` manually implements `Send` (TODO: this change needs to be checked in depth)
### Commit 2
- `unsafe impl Send for Export {}` (temporary to allow Memory to be not Send)
- RefCell -> Mutex in Global and Table
- Rc -> Arc in Table
- Namespace's `IsExport`s must be `Send` (Done to avoid touching much more of the code (i.e. `trait IsExport: Send` requires a lot -- maybe this is what we should do though)
- Make `Host` and `Wasm` `Func`s Send (manual implementation)
- Manual implementation for `LocalTable` and `AnyFunc`
### Commit 3
- rm placeholder `unsafe impl Send for Export {}`
- Manual implementation for `LocalBacking` and `ImportBacking` (both seemed to be not Send due to direct ownership of mutable pointers in their containers)
- ImportObject's state creator Fn trait object is now ` + Send + Sync + 'static` (required because it's in an Arc)
- Manually implement Send for `InstanceInner` because it holds a raw pointer, `LocalBacking` and `ImportBacking` are marked Send separately 
- Memory: All Rc -> Arc (including unshared memory); All RefCell -> Mutex (including unshared memory)
- Manual implementation of Send and Sync on `UnsharedMemoryInternal`
- Manual implementation of Send and Sync on `SharedMemoryInternal`
- Change `runtime-core::unix::memory::Memory.fd` from `Option<Rc<Rawfd>>` to `Option<Arc<Rawfd>>` (not strictly required for this change because Memory has manual implementations of Send and Sync, but Arc seems more correct here and there's no comment justifying the use of Rc)
- Manual implementation of Send for `ImportedFunc`


lib/runtime-core/src/import.rs
```diff
@@ -105,27 +106,60 @@ impl ImportObject {
         }
     }
 
-    pub fn get_namespace(&self, namespace: &str) -> Option<Ref<dyn LikeNamespace + 'static>> {
-        let map_ref = self.map.borrow();
-
+    /*pub fn get_namespace(
```
nlewycky: Commented out code -- delete?

lib/runtime-core/src/export.rs
```diff
@@ -11,6 +11,8 @@ pub enum Context {
     Internal,
 }
 
+unsafe impl Send for Context {}
```
nlewycky: Would it be possible to write a comment explaining why this is safe? Or why it can't be inferred?

lib/runtime-core/src/export.rs
```diff
@@ -26,6 +28,8 @@ pub enum Export {
 #[derive(Debug, Clone)]
 pub struct FuncPointer(*const vm::Func);
 
+unsafe impl Send for FuncPointer {}
```
nlewycky: For this in particular, I think it's relevant that we don't have any thread-local variables.

MarkMcCaskey: Hmm, can you explain what you mean?  I agree that thread-local variables can't be Send, is that something we've done or talked about in the past with FuncPointer?

bjfish:
It would be nice to add a test like this for module/instance:

https://github.com/wasmerio/wasmer/blob/801339acee38d9e51e912d5cf08d128501193b8a/lib/runtime-core/src/instance.rs#L736-L745

MarkMcCaskey:
bors try

Issue #811, opened 2019-09-18
runtime-c-api tests fail with singlepass-backend


ethanfrey:
### Describe the bug

 

```bash
echo "`wasmer -V` | `rustc -V` | `uname -m`"
```

```
wasmer: command not found
 | rustc 1.38.0-nightly (c43d03a19 2019-08-14) | x86_64
```

`git log -1` provides:

```
commit 6d351b514ccc15fa2f344de4ed68506a6d3baa23 (HEAD -> master)
Merge: 569d0386 91b199fe
Author: bors[bot] <26634292+bors[bot]@users.noreply.github.com>
Date:   Thu Aug 29 20:48:53 2019 +0000
```

----------------------------

I realized this was an old commit. 

I then updated to current master, fixed a typo, and while `cargo test` passes, 
`cargo test --features singlepass-backend --no-default-features` returns a linking error:

```
   Compiling wasmer-runtime-c-api v0.7.0 (/home/ethan/IdeaProjects/wasmer/lib/runtime-c-api)
error: linking with `cc` failed: exit code: 1
  |
...
note: /usr/bin/ld: /home/ethan/IdeaProjects/wasmer/target/debug/deps/libwasmer_runtime_core-dcfea6f3cbfc5444.rlib(image-loading-linux-x86-64.o): relocation R_X86_64_PC32 against symbol `get_boundary_register_preservation' can not be used when making a shared object; recompile with -fPIC
          /usr/bin/ld: final link failed: Bad value
          collect2: error: ld returned 1 exit status
```

from:
```
commit 0ea5b1b2bd87a704e2963bb92b9c0d96e2bf66ba (HEAD -> master, origin/staging, origin/master)
Merge: 174cc9f0 71b865aa
Author: bors[bot] <26634292+bors[bot]@users.noreply.github.com>
Date:   Wed Sep 18 16:59:37 2019 +0000
```

### Steps to reproduce
 

```shell
cd lib/c-runtime-api
cargo test --features singlepass-backend --no-default-features
```
fails on current `master`

Note that I had to use `--no-default-features` flags so this compiles at all.

### Expected behavior

Tests pass, as with cranelift or llvm

### Actual behavior

Result:
```
test 11
      Start 11: test-module-serialize

11: Test command: /home/ethan/IdeaProjects/wasmer/lib/runtime-c-api/tests/test-module-serialize
11: Test timeout computed to be: 9.99988e+06
[...]

PR #814, opened 2019-09-18
Fix issues with testing/linking singlepass in runtime-c-api

ethanfrey:
# Description

This resolves the regression from #811 (singlepass tests in runtime-c-api no longer even linking under linux). (The failing serialize test is still problematic)

Thanks to @nlewycky for debugging this with me and finding the solution.

# Review

- [ ] Create a short description of the the change in the CHANGELOG.md file


syrusakbary:
bors try

PR #817, opened 2019-09-19
Add feature matrix document

MarkMcCaskey:
[rendered](https://github.com/wasmerio/wasmer/blob/docs/add-feature-matrix/docs/feature_matrix.md)

# Review


torch2424:
@syrusakbary 

> This is a good first PR. Perhaps we can use emojis (✅| ✔️, ❌) instead of (Yes, No)

Hmmm. Do you happen to know how these are read for accessibility? 🤔 

But if they are read in a way that makes sense, I think it is a great idea! 😄

syrusakbary:
> Hmmm. Do you happen to know how these are read for accessibility? 🤔

No idea

MarkMcCaskey:
I checked and the osx screen reader can handle emoji reasonably well

torch2424:
@MarkMcCaskey Thanks for checking Mark! 😄

PR #821, opened 2019-09-20
Remove patch version in deps when not necessary

MarkMcCaskey:
This allows wasmer library users to have more control over the exact versions of deps that Wasmer uses.

# Review


Hywan:
:+1:

PR #828, opened 2019-09-23
No need to emit add of constant zero.

nlewycky:
# Description
While debugging singlepass generated code, I noticed unnecessary adds of zero. Removing them made the resulting assembly shorter and easier to analyze.


lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -1512,22 +1512,24 @@ impl X64FunctionCode {
...
-                        Size::S64,
-                        Location::Imm32(offset as u32),
-                        Location::GPR(tmp_addr),
-                    );
-                    a.emit_add(
-                        Size::S64,
-                        Location::Imm32(value_size as u32),
-                        Location::GPR(tmp_addr),
-                    );
+            if offset != 0 && value_size != 0 {
```
losfair: Should this be `if offset != 0 || value_size != 0` ?

nlewycky: Good catch!

Looking at this again, I don't think the case where we can avoid emitting an add here happens: `value_size` is never 0 in practice, so either the sum doesn't overflow or it does and we must emit both adds. There's no point placing another if-statement in the `None` branch since they both must be non-zero in order for an overflow to have occurred.

I've added a trivial "if the sum is zero, don't emit an add" which is obviously correct this time, but I don't think it'll ever fire.

nlewycky:
bors try

PR #831, opened 2019-09-24
Add support for atomic operations, excluding wait and notify, to singlepass.

nlewycky:
# Description
Adds support for atomic operations, excluding wait and notify, to singlepass. Enable with `--enable-threads`.

# Review


nlewycky:
bors try

lib/singlepass-backend/src/emitter_x64.rs
```diff
@@ -562,15 +566,15 @@ impl Emitter for Assembler {
             (Size::S64, Location::Memory(src, disp), Location::GPR(dst)) => {
                 dynasm!(self ; lea Rq(dst as u8), [Rq(src as u8) + disp]);
             }
-            _ => unreachable!(),
+            _ => panic!("LEA {:?} {:?} {:?}", sz, src, dst),
```
syrusakbary: Can you make the panics a bit more descriptive?

Just appending `singlepass reached `:

```suggestion
            _ => panic!("singlepass reached LEA {:?} {:?} {:?}", sz, src, dst),
```

nlewycky: Done.

lib/singlepass-backend/src/machine.rs
```diff
@@ -83,7 +83,14 @@ impl Machine {
 
     /// Releases a temporary GPR.
     pub fn release_temp_gpr(&mut self, gpr: GPR) {
-        assert_eq!(self.used_gprs.remove(&gpr), true);
+        assert!(self.used_gprs.remove(&gpr));
+    }
+
+    /// Specify that a given register is in use.
+    pub fn reserve_temp_gpr(&mut self, gpr: GPR) -> GPR {
```
losfair: Could you make the name a bit more distinguishable to `acquire_temp_gpr`, like `reserve_unused_gpr` ?

nlewycky: Done. I wanted to keep `temp` part of the name so that it refers to the RAX, RCX, RDX list of registers, not the RSI ... R11 list.

losfair: Okay, I thought this was used to reserve the RSI..R11 registers :)

MarkMcCaskey: Might be worth adding a comment explaining this if the new name doesn't make this explicit.  As someone without context on this, I'm sure it'd take be a bit to figure this out

nlewycky: The new name is `reserve_unused_temp_gpr`. It also has a comment on the function, and an assertion in the function. I'm in the position where I know what it does, so I can't see what about it might be confusing. Let me know if I can improve it?

nlewycky:
bors try

nlewycky:
Can I get an approval on this? Are @syrusakbary or @MarkMcCaskey OK with doing it or are we waiting for @losfair ?

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -1536,27 +1538,107 @@ impl X64FunctionCode {
...
+        if check_alignment && align != 1 {
+            let tmp_aligncheck = m.acquire_temp_gpr().unwrap();
+            //let tmp_mask = m.acquire_temp_gpr().unwrap();
+            a.emit_mov(
+                Size::S32,
+                Location::GPR(tmp_addr),
+                Location::GPR(tmp_aligncheck),
+            );
+            //a.emit_mov(Size::S64, Location::Imm64(align - 1), Location::GPR(tmp_mask));
+            //a.emit_and(Size::S64, Location::GPR(tmp_mask), Location::GPR(tmp_aligncheck));
```
MarkMcCaskey: Is this debug code?  Probably best to remove it or explain in a comment why it's there

nlewycky: Deleted.

It's the old code. We're under heavy register pressure here, and I rewrote it to use one less register, but kept the original around in case my changes didn't work properly. Thanks!

PR #850, opened 2019-09-30
Add builder API for WasiState

MarkMcCaskey:
Nicer to use and it checks for errors!

# Review


lib/wasi/src/state/builder.rs
```diff
@@ -0,0 +1,293 @@
...
+///            hm.insert("COLOR_OUTPUT", "TRUE");
+///            hm.insert("PATH", "/usr/bin");
+///            hm
+///          })
+///    .args(&["--verbose", "list"])
+///    .preopen_dir("src")
+///    .map_dir("dot", ".")
+///    .build()
+///    .unwrap();
+/// ```
```
syrusakbary: I was thinking that perhaps this API might be simpler? (removing `crate_wasi_state` and renaming `WasiStateBuilder` to `WASICommand` )

```
use wasmer_wasi::WASICommand;

WASICommand::new("")
   .env(b"HOME", "/home/home".to_string())
   .arg("--help")
   .envs({ let mut hm = std::collections::HashMap::new();
           hm.insert("COLOR_OUTPUT", "TRUE");
           hm.insert("PATH", "/usr/bin");
           hm
         })
   .args(&["--verbose", "list"])
   .preopen_dir("src")
   .map_dir("dot", ".")
   .build()
   .unwrap();
```

MarkMcCaskey:
I'm not implementing the command interface now.  I do like it though and it's something I want to revisit.  But I think it may not fit in this PR.  We could have a general `Wasm` command like structure that applies beyond WASI for constructing the Instance.  Something to think about, but I do think having an API path that's extremely simple is what we want.

MarkMcCaskey:
bors retry

PR #856, opened 2019-10-01
Add WASI support to C API

MarkMcCaskey:
This is an additive change (with one unrelated clean up of `*mut ptr -> *const ptr`).  It exposes the functions to get a WASI import object.  It also implements a function on import object to get an `import` with namespace and name.

These changes should be okay to ship now, we can follow up to finish adding methods to ImportObject on the C API side and start migrating away from  `*imports`.

# Review


lib/runtime-c-api/src/import.rs
```diff
@@ -51,6 +51,110 @@ pub unsafe extern "C" fn wasmer_import_object_new() -> *mut wasmer_import_object
...
+    pub unsafe extern "C" fn wasmer_wasi_generate_default_import_object(
+    ) -> *mut wasmer_import_object_t {
+        let import_object = Box::new(wasmer_wasi::generate_import_object(
+            vec![],
+            vec![],
+            vec![],
+            vec![],
+        ));
+
+        Box::into_raw(import_object) as *mut wasmer_import_object_t
```
Hywan: If you forget about a pointer, consider adding the “destroy” function in Rust to drop the pointer.

MarkMcCaskey: In the C example I wrote, I used `    wasmer_import_object_destroy(import_object);` which should take care of it, I believe!

MarkMcCaskey: (`wasmer_import_object_destroy` was an existing function)

lib/wasi-tests/build/wasitests.rs
```diff
@@ -145,7 +145,7 @@ pub fn compile(file: &str, ignores: &HashSet<String>) -> Option<String> {
         out_str.push_str("vec![");
 
         for entry in args.po_dirs {
-            out_str.push_str(&format!("\"{}\".to_string(),", entry));
+            out_str.push_str(&format!("std::path::PathBuf::from(\"{}\"),", entry));
```
MarkMcCaskey: oops, this made it onto the wrong PR

Hywan:
What is left to do?

MarkMcCaskey:
@Hywan I think it's done, we just wanted to get it tested (I made an example in the C API example repo and I'm working on implementing it in Go).  I think I know enough to say with reasonable confidence that we can ship this before finishing the implementation in Go because it just returns an import object but 🤷‍♂ .

One thing this code doesn't do is expose the errors properly.  I'd like to hear your thoughts on that.

lib/runtime-c-api/src/import.rs
```diff
@@ -51,12 +51,324 @@ pub unsafe extern "C" fn wasmer_import_object_new() -> *mut wasmer_import_object
...
+        name
+    } else {
+        update_last_error(CApiError {
+            msg: "error converting name to UTF-8 string".to_string(),
+        });
+        return wasmer_result_t::WASMER_ERROR;
+    };
+    if import.is_null() || import_export_value.is_null() {
+        update_last_error(CApiError {
+            msg: "pointer to import and import_export_value must not be null".to_string(),
```
Hywan:
```diff
- pointer
+ pointers
```

lib/runtime-c-api/src/import.rs
```diff
@@ -51,12 +51,324 @@ pub unsafe extern "C" fn wasmer_import_object_new() -> *mut wasmer_import_object
...
+    return i as i32;
+}
+
+#[no_mangle]
+/// Frees the memory acquired in `wasmer_import_object_get_functions`
+pub unsafe extern "C" fn wasmer_import_object_imports_destroy(
+    imports: *mut wasmer_import_t,
+    imports_len: u32,
+) {
+    // what's our null check policy here?
```
Hywan: If the pointer is NULL, we should raise an error. Since it's about destroying a value here, if it's null, there is nothing to destroy, so we usually go silent, and return directly.

lib/runtime-c-api/src/export.rs
```diff
@@ -85,12 +85,39 @@ pub union wasmer_import_export_value {
...
+}
+
+impl std::convert::TryFrom<u32> for wasmer_import_export_kind {
+    type Error = ();
+
+    fn try_from(value: u32) -> Result<Self, Self::Error> {
+        if value > 3 {
+            Err(())
+        } else {
+            Ok(unsafe { std::mem::transmute(value) })
```
Hywan: Replacing this function body by a `match` would help to remove the unsafe `mem::transmute` part.

Hywan:
@MarkMcCaskey I've created https://github.com/wasmerio/wasmer/pull/873 to fix some of my feedbacks in this PR. Feel free to include them if you think they are relevant.

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/runtime-c-api/src/import/mod.rs
```diff
@@ -51,12 +51,252 @@ pub unsafe extern "C" fn wasmer_import_object_new() -> *mut wasmer_import_object
...
+        wasmer_result_t::WASMER_OK
+    } else {
+        update_last_error(CApiError {
+            msg: format!("Export {} {} not found", namespace_str, name_str),
+        });
+        wasmer_result_t::WASMER_ERROR
+    }
+}
+
+#[no_mangle]
```
Hywan: Could you move the attribute after the comment please, so that it is consistent with the rest of the code?

lib/runtime-c-api/src/import/mod.rs
```diff
@@ -51,12 +51,252 @@ pub unsafe extern "C" fn wasmer_import_object_new() -> *mut wasmer_import_object
...
+            if let Export::Function { .. } = e {
+                true
+            } else {
+                false
+            }
+        })
+        .count() as i32
+}
+
+#[no_mangle]
```
Hywan: Same here as above.

lib/runtime-c-api/src/import/mod.rs
```diff
@@ -51,12 +51,252 @@ pub unsafe extern "C" fn wasmer_import_object_new() -> *mut wasmer_import_object
...
+/// Call `wasmer_import_object_imports_destroy` to free the memory allocated by this function.
+/// This function return -1 on error.
+pub unsafe extern "C" fn wasmer_import_object_get_functions(
+    import_object: *const wasmer_import_object_t,
+    imports: *mut wasmer_import_t,
+    imports_len: u32,
```
Hywan: What's the point of limiting the maximum number of imports to read?

lib/runtime-c-api/src/import/mod.rs
```diff
@@ -51,12 +51,252 @@ pub unsafe extern "C" fn wasmer_import_object_new() -> *mut wasmer_import_object
...
+                i += 1;
+            }
+            _ => (),
+        }
+    }
+
+    return i as i32;
+}
+
+#[no_mangle]
```
Hywan: Same note as above about attribute position.

Hywan:
Minor comments. The PR looks good. Nice work! Thanks.

lib/runtime-c-api/src/import/mod.rs
MarkMcCaskey: I did that so that the caller could control their own allocations

MarkMcCaskey: It is pretty weird since there's no way to say how much you want, but I figured it was necessary for it to be caller-allocated

Hywan: :+1:

Issue #679, opened 2019-08-14
[wasmer] multiple panics caused by unimplemented! macro


pventuzelo:
### Describe the bug

The macro `unimplemented!` is defined as following in: https://doc.rust-lang.org/1.6.0/std/macro.unimplemented!.html
> A standardized placeholder for marking unfinished code. It panics with the message "not yet implemented" when executed.

This macro is used multiple time in wasmer and make it panic:
```
thread 'main' panicked at 'not yet implemented', lib/clif-backend/src/resolver.rs:261:67
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

### Vulnerable code

Based on github search in this repository, 31 files are using unimplemented! macro (https://github.com/wasmerio/wasmer/search?q=unimplemented%21&type=Code)

``` bash
$ rg unimplemented!

src/webassembly.rs
56:    unimplemented!();

wapm-cli/src/config.rs
163:            unimplemented!()

src/bin/wasmer.rs
55:        unimplemented!()

lib/clif-backend/src/module.rs
111:            _ => unimplemented!("unsupported wasm type"),

lib/clif-backend/src/resolver.rs
260:                            VmCallKind::SharedStaticMemoryGrow => unimplemented!(),
261:                            VmCallKind::SharedStaticMemorySize => unimplemented!(),
278:                            VmCallKind::SharedStaticMemoryGrow => unimplemented!(),
279:                            VmCallKind::SharedStaticMemorySize => unimplemented!(),

lib/clif-backend/src/relocation.rs
108:        unimplemented!();
121:            _ => unimplemented!("unimplented reloc type: {}", reloc),
149:                        _ => unimplemented!(),
160:                        _ => unimplemented!(),
163:                    _ => unimplemented!(),
193:                    _ => unimplemented!("unimplemented libcall: {}", libcall),
211:        unimplemented!();

lib/clif-backend/src/code.rs
282:                    _ => unimplemented!(),

lib/runtime-core/src/vmcalls.rs
153:    unimplemented!()
159:    unimplemented!()

lib/runtime-core/src/vm.rs
211:                    MemoryType::SharedStatic => unimplemented!(),
219:                    MemoryType::SharedStatic => unimplemented!(),
868:                unimplemented!()
871:                unimplemented!()
876:                unimplemented!()

lib/runtime-core/src/instance.rs
495:        unimplemented!("Use the exports method instead");
512:        unimplemented!("Use the exports method instead");

[...]

PR #861, opened 2019-10-03
[fix issue #679] add details when calling unimplemented!

pventuzelo:
# Description

As discuss in issue #679, `unimplemented!` should provide a string to explain at least where the issue occurs. This pull request only add messages on existing `unimplemented!` and don't change any previous logics/behaviors in the code.

# Review


pventuzelo:
i just fixed an error happening during `build docs`.
Build need to run again.

PR #865, opened 2019-10-07
adding tests for `state_creator` of `import_object`

YaronWittenstein:
Part of the PR #807 changes was adding support for shared import objects between threads.

https://github.com/wasmerio/wasmer/pull/807/files#diff-d20cb4c5a883566b85be4cc046f45aa9R49

I've added tests/examples on how to create an `import object` with a state_creator
(function or closure)


lib/runtime-core/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![deny(
+#![allow(
```
MarkMcCaskey: This should be changed back!

lib/runtime-core/src/import.rs
```diff
@@ -254,6 +254,56 @@ mod test {
...
+
+        let (state, _dtor) = imports.call_state_creator().unwrap();
+        let data: &mut Data = unsafe { &mut *(state as *mut Data) };
+
+        assert_eq!(ptr, data.inner);
+    }
+
+    #[test]
+    fn state_creator_closure() {
+        let ptr = std::ptr::null();
```
MarkMcCaskey: Perhaps we should use a sentinel value, `0` is a common bit pattern in memory, something like:

`let ptr = unsafe { std::mem::transmute(0xB7B7A0A01234) };`

would make the test more robust I think

lib/runtime-core/src/lib.rs
YaronWittenstein: @MarkMcCaskey  may bad

YaronWittenstein:
> bors r+

@MarkMcCaskey can you please re-trigger `bors` ?

MarkMcCaskey:
Thanks!  Sorry about not shipping this before -- I only very recently learned/realized that not everyone can use bors on Wasmer! I assumed given an approval, people could ship their own PRs, but this seems to not be the case!  There are a few other PRs I didn't ship due to this misunderstanding which I'll be following up on

syrusakbary:
This is no longer needed on the refactor. Adding the label so we can close quickly the PR after publishing the refactor.

syrusakbary:
State creator is no longer supported in the import object.
We moved towards a cleaner API in the refactor, removing the need for state creation.

PR #867, opened 2019-10-07
Initial implementation of `assert_uninstantiable`.

nlewycky:
# Description
Add a basic implementation of `assert_uninstantiable` in the spectest runner. Fixes one line of spectests in all three backends.


MarkMcCaskey:
Actually, if it's sufficiently obvious given the rest of this file, then don't worry about it

nlewycky:
Assert that the module described is not instantiable, where instantiate is the process defined in the webassembly spec, for wasmer it means producing a `Instance` from the `Module`. I don't feel like it needs a comment, let me know if you want one.

nlewycky:
Thanks @syrusakbary !

PR #870, opened 2019-10-08
Fix unused value warning due to inkwell API change. NFC.

nlewycky:
# Description
Fix unused value warning due to inkwell API change. No functionality change.


nlewycky:
> I think it'd be useful to add a comment about why this `unwrap` is safe. If it's not always safe, then we should return an error

Done.

PR #875, opened 2019-10-13
doc: add link to SO in C api example

alocquet:
# Description

When I try to compile C API example, I didn't found easily the library needed to compile. I had to recompile wasmer before I found the library can be downlaoded in github release page. 

I make this PR just to update the doc so as other deveoppers don't take time to find SO file.

I have created an issue here : https://github.com/wasmerio/wasmer-c-api/issues/8 but I think the link as a better place in this README page.

# Review


syrusakbary:
Looks great! Merging

PR #877, opened 2019-10-14
Reimplement F32Min, F32Max, F64Min and F64Max.

nlewycky:
# Description
Reimplement F32Min, F32Max, F64Min and F64Max.

Adds XMM8--15 registers. Adds VMOVA, VBLEND and VXORP, and the VCMPUNORD and VCMPORD comparisons.

Fixes 419 test failures.


nlewycky:
bors try

nlewycky:
bors try

lib/runtime-core/src/state.rs
```diff
@@ -1023,6 +1023,14 @@ pub mod x64 {
...
         XMM6,
         XMM7,
+        XMM8,
+        XMM9,
+        XMM10,
+        XMM11,
+        XMM12,
+        XMM13,
+        XMM14,
+        XMM15,
```
losfair: Could you also add these registers to `invoke_call_return_on_stack` and `from_dwarf_regnum`? Seems that we are not using it for persistent values now but just in case.

nlewycky: Done.

BTW, I got the idea of using them for temporary values from the [x86-64 platform ABI](https://github.com/hjl-tools/x86-psABI/wiki/x86-64-psABI-1.0.pdf):
> | Register | Usage | callee saved |
> | -------- | ----- | ------------ |
> | `%xmm0–%xmm1` | used to pass and return floating point arguments | No |
>|  `%xmm2–%xmm7` | used to pass floating point arguments | No |
> | `%xmm8–%xmm15` | temporary registers | No |

losfair: The `run_on_alternative_stack` function in `runtime-core/image-loading-macos-x86-64.s` and `runtime-core/image-loading-linux-x86-64.s` also needs to be kept consistent with `invoke_call_return_on_stack` here, because the XMM values pushed onto the stack need to be actually loaded into the registers.

nlewycky: Great catch, thank you! Done.

PR #881, opened 2019-10-16
feat(runtime-core) Add documentation and make macros more readable.

Hywan:
It's a small PR.

After a long time, the macros are easy to read, but not at first
glance. I hope this PR will improve the situation: Same syntax used
everywhere, more spaces…


Hywan:
bors try

PR #883, opened 2019-10-17
For floating point operations, allow inputs to be arbitrary, including SNaNs.

nlewycky:
# Description
For floating point operations, allow inputs to be arbitrary, including SNaNs.

Instead of ensuring inputs are canonical NaNs on every operation, we tag outputs as pending such a canonicalization check, so that a sequence of computations can have a single canonicalization step at the end.

There's an extra wriggle for SIMD. The Wasm type system only indicates them as V128, so it's possible that we might do computations as F32x4Add, I8x16Add, F64x2Add in a row with no other computations in between. Since a canonicalization may change the bit patterns in a way that transforms one non-NaN to another non-NaN in the next subsequent instructions interpretation, most SIMD functions apply pending canonicalizations to their inputs, even the integer SIMD operations.

# Review


nlewycky:
@syrusakbary I'd appreciate if you could benchmark with this PR, please!

nlewycky:
bors try

nlewycky:
I ran some benchmarks myself with particle-repel:

> master 
> simd: 36.56s 36.69s 36.43s
> non-simd: 100.65s 100.66s, 100.62s
>
> feature/llvm-nan-but-fast
> simd: 48.69s 48.70s 48.74s
> non-simd: 62.48s 62.63s 62.40s

The speedup with non-SIMD is what I expected. I know we have to handle SIMD differently, and that it can put canonicalizations where we didn't use to have them sometimes, but I was not expecting a *slowdown* compared to our current approach.

It looks like wasi-sdk-5 doesn't emit a SIMD FDiv operation, but instead emits individual FDivs. At master, we got lucky with the LLVM autovectorizer reassembling it into a SIMD operation, but with this branch, we got unlucky and one of the fdivs moved into a different basic block. Overall we emit fewer instructions, even for the SIMD case, with this patch.

syrusakbary:
Good to merge once we have the changes integrated in the Changelog

PR #885, opened 2019-10-17
fix(cranelift-backend) Remove broken (and useless?) debug code

Hywan:
This debug code is broken (it doesn't compile). It can be assumed
nobody uses it, and can be considered as dead code. As such, this PR
removes it.

The `debug` feature is kept in `Cargo.toml` to activate `wasmer-runtime-core/debug`.


Hywan:
bors try

PR #886, opened 2019-10-17
fix(win-exception-handler) Remove `bindgen` and `regex` from cargo build deps

Hywan:
Those crates are not used.


Hywan:
bors try

PR #887, opened 2019-10-17
Get aggressive about running cargo check over all packages with all features.

nlewycky:
Fixes the one issue uncovered. The capstone disassembling support in the LLVM backend was broken. Fixed by removing it. Instead, use the `--llvm-object-file` flag to get a finished object file to disassemble with any disassembler.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,7 +153,37 @@ check-bench-llvm:
 check-bench: check-bench-singlepass check-bench-llvm
 
 check: check-bench
-	cargo check --release --features backend-singlepass,backend-llvm,loader-kernel,debug
+	# TODO: We wanted `--workspace --exclude wasmer-runtime`, but for
+	# https://github.com/rust-lang/cargo/issues/6745 .
+	cargo check -p wasmer-clif-backend -p wasmer-singlepass-backend -p wasmer-middleware-common -p wasmer-runtime-core -p wasmer-emscripten -p wasmer-llvm-backend -p wasmer-wasi -p wasmer-kernel-loader -p wasmer-dev-utils -p wasmer-wasi-tests -p wasmer-middleware-common-tests -p wasmer-emscripten-tests
+	cargo check --release -p wasmer-clif-backend -p wasmer-singlepass-backend -p wasmer-middleware-common -p wasmer-runtime-core -p wasmer-emscripten -p wasmer-llvm-backend -p wasmer-wasi -p wasmer-kernel-loader -p wasmer-dev-utils -p wasmer-wasi-tests -p wasmer-middleware-common-tests -p wasmer-emscripten-tests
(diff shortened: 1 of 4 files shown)


nlewycky:
bors try

Makefile
```diff
@@ -153,7 +153,37 @@ check-bench-llvm:
 check-bench: check-bench-singlepass check-bench-llvm
 
 check: check-bench
-	cargo check --release --features backend-singlepass,backend-llvm,loader-kernel,debug
+	# TODO: We wanted `--workspace --exclude wasmer-runtime`, but for
```
Hywan: “but for”? Is the phrasing correct here?

Hywan:
Looks good to me

Makefile
```diff
@@ -153,7 +153,37 @@ check-bench-llvm:
...
+	# https://github.com/rust-lang/cargo/issues/6745 .
+	cargo check -p wasmer-clif-backend -p wasmer-singlepass-backend -p wasmer-middleware-common -p wasmer-runtime-core -p wasmer-emscripten -p wasmer-llvm-backend -p wasmer-wasi -p wasmer-kernel-loader -p wasmer-dev-utils -p wasmer-wasi-tests -p wasmer-middleware-common-tests -p wasmer-emscripten-tests
+	cargo check --release -p wasmer-clif-backend -p wasmer-singlepass-backend -p wasmer-middleware-common -p wasmer-runtime-core -p wasmer-emscripten -p wasmer-llvm-backend -p wasmer-wasi -p wasmer-kernel-loader -p wasmer-dev-utils -p wasmer-wasi-tests -p wasmer-middleware-common-tests -p wasmer-emscripten-tests
+	cargo check -p wasmer-clif-backend -p wasmer-singlepass-backend -p wasmer-middleware-common -p wasmer-runtime-core -p wasmer-emscripten -p wasmer-llvm-backend -p wasmer-wasi -p wasmer-kernel-loader -p wasmer-dev-utils -p wasmer-wasi-tests -p wasmer-middleware-common-tests -p wasmer-emscripten-tests --all-features
```
syrusakbary: We can directly check with all features (and skip the previous checks). I believe it will alleviate time to check

Makefile
nlewycky: Doh, almost. "We would have used `--workspace --exclude wasmer-runtime`, but for <bug>" would have been correct. Fixed!

Makefile
nlewycky: Is that testing the same thing? Or is that a way to test 99% for less time?

nlewycky: Asked @syrusakbary about it in person, he wasn't sure what case I was worried about (case that `--enable-all` doesn't catch) and suggested there might be a smarter (less `make check` time) way to go about it. The problem I'm worried about is some feature A is on by default and relies on feature B which is off by default. With `--enable-all` both A and B will be on so it works. Without it, A is on, relies on B that isn't there, and fails.

@MarkMcCaskey could you take a look?

PR #889, opened 2019-10-18
fix(singlepass-backend) Use wasmparser from `runtime-core`

Hywan:
The `wasmer-runtime-core` crate re-exports the `wasmparser`
crate. This patch updates the `singlepass-backend` crate to use
`wasmparser` through the `wasmer-runtime-core` crate, which removes a
direct dependency for this crate.


Hywan:
bors try

nlewycky:
I don't understand this part of rust. Why should runtime-core re-export wasmparser at all? Why is it better to use the re-exported version?

Hywan:
@nlewycky Dependencies don't have to deal or to stick to a specific parser version. It's just… comfortable to work like that. `wasmer-runtime-core` re-exports the parser so that we are sure all the dependencies use the same version of the parser.

PR #910, opened 2019-10-28
Use getrandom instead of rand

### LINKED ISSUE #909: Use getrandom crate directly instead of rand?
Right now [`random_get`](https://github.com/wasmerio/wasmer/blob/master/lib/wasi/src/syscalls/mod.rs#L2454) uses `ThreadRng`. I think we should use `getrandom` crate directly instead, i.e. we should simply proxy host OS entropy, without an additional user-space layer. It will simplify Wasmer dependency tree (since `rand` is not used anywhere else) and potential performance impact is not important, since `random_get` will be probably used mostly only for seeding of user-space PRNGs on the WASI side.

Same can be done with this [code](https://github.com/wasmerio/wasmer/blob/master/lib/emscripten/src/syscalls/windows.rs#L42) as well, 32 byte token can be generated directly by `getrandom`.

newpavlov:
Closes #909

Note: I am reusing `__WASI_EIO` as an error code to denote a potential OS RNG failure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -451,7 +451,7 @@ dependencies = [
 
 [[package]]
 name = "getrandom"
-version = "0.1.12"
+version = "0.1.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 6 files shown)


MarkMcCaskey:
Looks like there's a small error on the windows side: we need to add a `&mut`!

newpavlov:
Done!

MarkMcCaskey:
Awesome, thanks!

PR #916, opened 2019-10-30
feat(runtime-core-tests) Introduce the new `wasmer-runtime-core-tests` crate

Hywan:
Extracted from https://github.com/wasmerio/wasmer/pull/882.

This non-publishable new crate contains a test suite for the
`wasmer-runtime-core` crate. So far, the test suite is rather small,
but it aims to be extended in a close future.

PR #882, opened 2019-10-17
feat: Support closures for imported functions

Hywan:
Reboot of https://github.com/wasmerio/wasmer/pull/219:

* [x] `runtime-core`: Support closures
   The trampoline and the invoke functions take a new pointer to a function
   environment (`*mut vm::FuncEnv`).
   The function environment `vm::FuncEnv` is an opaque pointer that can
   be of 2 kinds: Either a closure environment pointer, or `vm::Ctx`.
* [x] `clif-backend`: Support first point,
* [x] `llvm-backend`: Support first point,
* [x] `singlepass-backend`: Support first point,
* [x] `win-exception-handler`: The C part must be updated.

PR #917, opened 2019-10-30
feat(runtime-core) Host function without a `vm::Ctx` argument

Hywan:
Extracted from https://github.com/wasmerio/wasmer/pull/882.
~Includes https://github.com/wasmerio/wasmer/pull/916. Must not be merged until #916 lands in `master`.~
~If you want to compare only the different commits, see https://github.com/Hywan/wasmer/compare/feat-runtime-core-tests...Hywan:feat-runtime-core-host-function-without-vmctx?expand=1.~

This patch updates the `ExternalFunction` implementation to support host functions (aka imported functions) without having an explicit `vm::Ctx` argument.

It is thus possible to write:

```rust
fn add_one(x: i32) -> i32 {
    x + 1
}

    "env" => {
        "add_one" => func!(add_one);
    },
};
```

The previous form is still working though:

```rust
fn add_one(_: &mut vm::Ctx, x: i32) -> i32 {
    x + 1
}
```

The backends aren't modified. It means that the pointer to `vm::Ctx` is still inserted in the stack, but it is not used when the function doesn't need it.

We thought this is an acceptable trade-off.

About the C API. It has not check to ensure the type signature. Since the pointer to `vm::Ctx` is always inserted, the C API can digest host functions with or without a `vm::Ctx` argument. The C API uses [the `Export` + `FuncPointer` API](https://github.com/wasmerio/wasmer/blob/cf5b3cc09eff149ce415bd81846d84b2d2cdf3a3/lib/runtime-c-api/src/import/mod.rs#L630-L653) instead of going through the `Func` API (which is modified by this API), which is only possible since the C API only receives an opaque function pointer. Conclusion is: We can't ensure anything on the C side, but it will work with or without the `vm::Ctx` argument in theory.

### DIFF
--- a/lib/runtime-core/src/typed_func.rs
+++ b/lib/runtime-core/src/typed_func.rs
@@ -3,7 +3,7 @@ use crate::{
     export::{Context, Export, FuncPointer},
     import::IsExport,
     types::{FuncSig, NativeWasmType, Type, WasmExternType},
-    vm::{self, Ctx},
+    vm,
 };
 use std::{
     any::Any,
@@ -52,10 +52,10 @@ impl fmt::Display for WasmTrapInfo {
 /// of the `Func` struct.
 pub trait Kind {}
 
-pub type Trampoline = unsafe extern "C" fn(*mut Ctx, NonNull<vm::Func>, *const u64, *mut u64);
+pub type Trampoline = unsafe extern "C" fn(*mut vm::Ctx, NonNull<vm::Func>, *const u64, *mut u64);
 pub type Invoke = unsafe extern "C" fn(
     Trampoline,
-    *mut Ctx,
+    *mut vm::Ctx,
     NonNull<vm::Func>,
     *const u64,
     *mut u64,
@@ -173,7 +173,7 @@ where
 pub struct Func<'a, Args = (), Rets = (), Inner: Kind = Wasm> {
     inner: Inner,
     f: NonNull<vm::Func>,
-    ctx: *mut Ctx,
+    ctx: *mut vm::Ctx,
     _phantom: PhantomData<(&'a (), Args, Rets)>,
 }
 
@@ -188,7 +188,7 @@ where
     pub(crate) unsafe fn from_raw_parts(
         inner: Wasm,
         f: NonNull<vm::Func>,
-        ctx: *mut Ctx,
+        ctx: *mut vm::Ctx,
     ) -> Func<'a, Args, Rets, Wasm> {
         Func {
             inner,
@@ -267,7 +267,7 @@ impl WasmTypeList for Infallible {
         self,
         _: NonNull<vm::Func>,
         _: Wasm,
-        _: *mut Ctx,
+        _: *mut vm::Ctx,
     ) -> Result<Rets, RuntimeError>
     where
         Rets: WasmTypeList,
@@ -313,7 +313,7 @@ where
         self,
         f: NonNull<vm::Func>,
         wasm: Wasm,
-        ctx: *mut Ctx,
+        ctx: *mut vm::Ctx,
     ) -> Result<Rets, RuntimeError>
     where
         Rets: WasmTypeList,
@@ -405,7 +405,7 @@ macro_rules! impl_traits {
                 self,
                 f: NonNull<vm::Func>,
                 wasm: Wasm,
-                ctx: *mut Ctx,
+                ctx: *mut vm::Ctx,
             ) -> Result<Rets, RuntimeError>
             where
                 Rets: WasmTypeList
@@ -443,7 +443,7 @@ macro_rules! impl_traits {
             $( $x: WasmExternType, )*
             Rets: WasmTypeList,
             Trap: TrapEarly<Rets>,
-            FN: Fn(&mut Ctx $( , $x )*) -> Trap,
+            FN: Fn(&mut vm::Ctx $( , $x )*) -> Trap,
         {
             #[allow(non_snake_case)]
             fn to_raw(&self) -> NonNull<vm::Func> {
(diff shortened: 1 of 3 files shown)


Hywan:
bors try

lib/runtime-core/src/typed_func.rs
```diff
@@ -438,33 +472,91 @@ macro_rules! impl_traits {
             }
         }
 
-        impl< $( $x, )* Rets, Trap, FN > ExternalFunction<( $( $x ),* ), Rets> for FN
+        impl< $( $x, )* Rets, Trap, FN > ExternalFunction<ExplicitVmCtx, ( $( $x ),* ), Rets> for FN
```
Hywan: Implementation of `ExternalFunction<ExplicitVmCtx, …>`, where `FN` must be `Fn(&mut vm::Ctx, …)`. The `wrap` function has a `&mut vm::Ctx` argument.

Note for the ones who aren't familiar with this detail: The `wrap` function will be dereferenced to `*mut vm::Func` (opaque function pointer), and will be used as `FuncPointer` by the `Export` API. So `wrap` acts as a “wrapper” around the user-given function.

lib/runtime-core/src/typed_func.rs
```diff
@@ -438,33 +472,91 @@ macro_rules! impl_traits {
...
+                    );
+
+                    NonNull::new(unsafe {
+                        mem::transmute_copy::<_, *mut vm::Func>(self)
+                    }).unwrap()
+                }
+            }
+        }
+
+        impl< $( $x, )* Rets, Trap, FN > ExternalFunction<ImplicitVmCtx, ( $( $x ),* ), Rets> for FN
```
Hywan: Implementation of `ExternalFunction<ImplicitVmCtx, …>`, where `FN` must be `Fn(…)` (without `vm::Ctx`). The `wrap` function _still_ has a `&mut vm::Ctx` argument! Why? Because the backend always inserts a pointer to `vm::Ctx` in the stack, so `wrap` can read it.

Why grabbing `vm::Ctx` then? It is mandatory to trap error with `do_early_trap`.

lib/runtime-core/src/typed_func.rs
```diff
@@ -124,16 +124,49 @@ pub trait WasmTypeList {
...
 }
 
+/// Empty trait to specify the kind of `ExternalFunction`: With or
+/// without a `vm::Ctx` argument. See the `ExplicitVmCtx` and the
+/// `ImplicitVmCtx` structures.
+///
+/// This type is never aimed to be used by a user. It is used by the
+/// trait system to automatically generate an appropriate `wrap`
+/// function.
+pub trait ExternalFunctionKind {}
```
MarkMcCaskey: Can it be made `pub (crate)`?

lib/runtime-core/src/typed_func.rs
```diff
@@ -438,33 +472,91 @@ macro_rules! impl_traits {
...
+                    extern fn wrap<$( $x, )* Rets, Trap, FN>(
+                        vmctx: &mut vm::Ctx $( , $x: <$x as WasmExternType>::Native )*
+                    ) -> Rets::CStruct
+                    where
+                        $( $x: WasmExternType, )*
+                        Rets: WasmTypeList,
+                        Trap: TrapEarly<Rets>,
+                        FN: Fn(&mut vm::Ctx, $( $x, )*) -> Trap,
+                    {
+                        let f: FN = unsafe { mem::transmute_copy(&()) };
```
MarkMcCaskey: This line is terrifying and deserves some comments.  Also is `()` ABI safe? 🤔

lib/runtime-core/src/typed_func.rs
```diff
@@ -438,33 +472,91 @@ macro_rules! impl_traits {
...
+        impl< $( $x, )* Rets, Trap, FN > ExternalFunction<ExplicitVmCtx, ( $( $x ),* ), Rets> for FN
+        where
+            $( $x: WasmExternType, )*
+            Rets: WasmTypeList,
+            Trap: TrapEarly<Rets>,
+            FN: Fn(&mut vm::Ctx $( , $x )*) -> Trap,
+        {
+            #[allow(non_snake_case)]
+            fn to_raw(&self) -> NonNull<vm::Func> {
+                if mem::size_of::<Self>() == 0 {
```
MarkMcCaskey: Presumably Rust wasn't happy with you using the `Sized` bound?

lib/runtime-core/src/typed_func.rs
Hywan: Nah unfortunately, since it's used by `ExternalFunction` which is public.

lib/runtime-core/src/typed_func.rs
Hywan: `Self` refers to `FN` so to the `Fn` trait. Writing `mem::size_of::<Self>() == 0` is a way to tell: `FN` is a function pointer, or a closure with no captured environment.

This code will change in the next PR to support closures as host functions. So far, this PR only copy-paste the existing code by defining`vm::Ctx` as an optional argument.

lib/runtime-core/src/typed_func.rs
Hywan: This is the original code. I'm happy to discuss about this details but I prefer to address it in another PR. I'm more inclined to split the modifications of this part of the code in small PR. Thoughts?

Hywan: Additional note: The `wrap` function is almost a duplication: Only the `FN` definition is different, along with the way `f` is called (with or without `vmctx`).

MarkMcCaskey: If the code already existed, we can leave it for now.  I have no idea how it works though!

Hywan: :+1:

PR #920, opened 2019-10-30
Export WasmPtr from wasmer_runtime

MarkMcCaskey:


Hywan:
What motivates this decision?

MarkMcCaskey:
@Hywan It's a very useful thing for embedders to use and in some of our examples, we have to import `wasmer_runtime_core` just to use it.  It's the safest way that users can access memory, so I think it should be the default!

PR #921, opened 2019-10-30
Apply TBAA metadata in the LLVM backend.

nlewycky:
# Description
Inform LLVM that the pointers to memory, globals, locals, etc., are distinct.

# Review

### DIFF
--- a/lib/llvm-backend/src/intrinsics.rs
+++ b/lib/llvm-backend/src/intrinsics.rs
@@ -1183,7 +1183,8 @@ pub fn tbaa_label(
     //
     // LLVM TBAA supports many features, but we use it in a simple way, with
     // only scalar types that are children of the root node. Every TBAA type we
-    // declare is `noalias` with the others:
+    // declare is NoAlias with the others. See NoAlias, PartialAlias,
+    // MayAlias and MustAlias in the LLVM documentation:
     //   https://llvm.org/docs/AliasAnalysis.html#must-may-and-no-alias-responses
 
     let module = module.borrow_mut();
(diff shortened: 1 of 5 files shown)


lib/llvm-backend/src/intrinsics.rs
```diff
@@ -1027,3 +1164,43 @@ impl<'a> CtxType<'a> {
         }
     }
 }
+
+pub fn tbaa_label(
```
Hywan: Can you document this part specifically please? Ideally, it should explain what TBAA means (is it Type Based Alias Analysis?), what it is used for, and some examples. Thanks!

Hywan:
Is there a way to test this new feature?

lib/llvm-backend/src/intrinsics.rs
nlewycky: I added comments both to the function declaration, what it does for the caller, and in its implementation, how it does what it achieves, with the goal that a future developer could follow along with the newly commended code in one hand and the LLVM TBAA documentation in the other.

I didn't add examples. Let me know if you'd like me to extend the comments further.

nlewycky:
> Is there a way to test this new feature?

Don't I wish!!

The difference is only visible to the user in two ways, different generated object code (and with it different size and performance), and different LLVM IR (which in turn generates the different object code). We have some debugging flags --llvm-object-file and --llvm-post-opt-ir and --llvm-pre-opt-ir, but we don't have any tests for those either (and indeed they were broken for a while due to their lack of testing). I've asked @MarkMcCaskey how we should test those flags, but we didn't have any good ideas so I left it. I'd like to start by testing those flags, then we use them to test changes to the LLVM backend.

If we make a change like adding TBAA, that leads to a pre-opt-ir change, changing the pass pipeline to take advantage of TBAA leads to a post-opt-ir change, and object file change. Changing whether we emit PIC code only causes an object file change. Unfortunately, if we simply had golden file tests, making any of these three changes would impact every single test.

PR #923, opened 2019-11-02
Add probable fix of memory leak in trampoline code

### LINKED ISSUE #810: Memory leak in wasmer.hh trampoline API
I'm trying to import a function into a wasm instance using the trampoline API from wasmer.hh (`wasmer_trampoline_buffer_builder_new()` etc). I've got it working, but when I run it under ASAN it detects 48 bytes of leaks due to the trampoline buffer. The snippet that creates the trampoline looks like:

```
    wasmer_trampoline_buffer_builder_t* builder =
        wasmer_trampoline_buffer_builder_new();
    uintptr_t trampoline_id =
        wasmer_trampoline_buffer_builder_add_callinfo_trampoline(
            builder,
            reinterpret_cast<wasmer_trampoline_callable_t*>(Trampoline),
            reinterpret_cast<void*>(fn_imp), num_trampoline_args);
    fn_imp->buffer = wasmer_trampoline_buffer_builder_build(builder);

    const wasmer_trampoline_callable_t* trampoline =
        wasmer_trampoline_buffer_get_trampoline(fn_imp->buffer, trampoline_id);
    fn_imp->wasm_fn = wasmer_import_func_new(
        reinterpret_cast<void (*)(void*)>(
            const_cast<wasmer_trampoline_callable_t*>(trampoline)),
        args, num_args, rets, num_rets);
```

Later I destroy `buffer` and `wasm_fn` (in `fn_imp`'s destructor) like this:

```
  ~WasmFunctionImport() {
    wasmer_trampoline_buffer_destroy(buffer);
    wasmer_import_func_destroy(wasm_fn);
  }
```

I've confirmed that the leak is due to the buffer, and that these destroy functions are being called. If I comment out `wasmer_trampoline_buffer_destroy(buffer);`, the amount of memory leaked doesn't change, so it seems like that function isn't hooked up correctly and isn't really freeing the buffer properly.

The only other possibility I can think of is that making a trampoline buffer doesn't actually allocate any memory after it's built, and the 48 bytes is from some other coincidental allocation. It's difficult to investigate fully with the allocation hidden inside an opaque library.

Have you folks tried running the trampoline API tests under ASAN?

In case it matters, I'm using wasmer-runtime-c-api 0.6.0.

MarkMcCaskey:
might be what's needed for #810 ; but despite my best efforts I could not get asan working on osx, so I did not test it.

By my count this accounts for 40 bytes, so there may be another issue

edit: unless Rust is optimizing out turning a zero-sized type into a Box, in which case, this should account for 48 bytes

### DIFF
--- a/lib/runtime-c-api/src/trampoline.rs
+++ b/lib/runtime-c-api/src/trampoline.rs
@@ -61,7 +61,7 @@ pub unsafe extern "C" fn wasmer_trampoline_buffer_builder_build(
 #[allow(clippy::cast_ptr_alignment)]
 pub unsafe extern "C" fn wasmer_trampoline_buffer_destroy(buffer: *mut wasmer_trampoline_buffer_t) {
     if !buffer.is_null() {
-        Box::from_raw(buffer);
+        Box::from_raw(buffer as *mut TrampolineBuffer);
     }
 }
 


lib/runtime-c-api/src/trampoline.rs
```diff
@@ -61,7 +61,7 @@ pub unsafe extern "C" fn wasmer_trampoline_buffer_builder_build(
 #[allow(clippy::cast_ptr_alignment)]
 pub unsafe extern "C" fn wasmer_trampoline_buffer_destroy(buffer: *mut wasmer_trampoline_buffer_t) {
     if !buffer.is_null() {
-        Box::from_raw(buffer);
+        Box::from_raw(buffer as *mut TrampolineBuffer);
```
Hywan: We usually use the following syntax:

```rs
let _: Box<TrampolineBuffer> = Box::from_raw(buffer);
```

Technically, it's the same though.

However, what surprises me is: What type `Box::from_raw` returns in this case? `Box<wasmer_trampoline_buffer>`? In this case, it's size is 1, and so, yes, it leaks data.

PR #915, opened 2019-10-30
fix(runtime-core) Share the definition of `Trampoline` across all the backends

Hywan:
Extracted from https://github.com/wasmerio/wasmer/pull/882.

This patch updates all the backends to use the definition of
`Trampoline` as defined in the `wasmer_runtime_core::typed_func`
module. That way, there is no copy of that type, and as such, it is
easier to avoid regression (a simple `cargo check` does the job).

This patch also formats the `use` statements in the updated files.

PR #925, opened 2019-11-04
feat(runtime-core) Support closures with a captured environment as host functions

Hywan:
Reboot of #882 and #219.

For the moment, host functions (aka imported functions) can be regular function pointers, or (as a side-effect) closures without a captured environment. This PR extends the support of host functions to closures with a captured environment. This is required for many other features (incl. the Python integration, the Ruby integration, WebAssembly Interface Types [see #787], and so on).

This PR is the culmination of previous ones, notably #915, #916 and #917. 

### General idea

The user-defined host function is wrapped inside a `wrap` function. This wrapper function initially receives a `vm::Ctx` as its first argument, which is passed to the host function when necessary. The patch keeps this behavior but it comes from `vm::FuncCtx`, which is a new structure. A `vm::FuncCtx` is held by `vm::ImportedFunc` such as:

```rust
#[repr(C)]
pub struct ImportedFunc {
    pub(crate) func: *const Func,
    pub(crate) func_ctx: NonNull<FuncCtx>,
}
```

where `vm::FuncCtx` is:

```rust
#[repr(C)]
pub struct FuncCtx {
    pub(crate) vmctx: NonNull<Ctx>,
    pub(crate) func_env: Option<NonNull<FuncEnv>>,
}
```

where `vm::FuncEnv` is:

```rust
#[repr(transparent)]
pub struct FuncEnv(pub(self) *mut c_void);
```

i.e. a raw opaque pointer.

So the wrapper function of a host function receives a `vm::Ctx`, which is used to find out the associated `FuncCtx` (by using the import backing), which holds `vm::FuncEnv`. It holds a pointer to the closure captured environment.

#### How to get a pointer to a closure captured environment

A closure with a captured environment has a memory size greater than zero. This is how we detect it:

```rust
if mem::size_of::<Self>() != 0 { … }
```

To get a pointer to its captured environment, we use this statement:

```rust
NonNull::new(Box::into_raw(Box::new(self))).map(NonNull::cast)
```

(in `typed_func.rs`, in the `wrap` functions).

To reconstruct the closure based on the pointer, we use this statement:

```rust
let func: &FN = {
    let func: NonNull<FN> = func_env.cast();
    &*func.as_ptr()
};
```

That's basically how it works. And that's the core idea of this patch.

As a side effect, we have removed an undefined behavior (UB) in 2 places: The `mem::transmute(&())` has been removed (it was used to get the function pointer of `FN`). The transmute is replaced by `FuncEnv`, which provides a unified API, erasing the difference between host functions as closures with a captured environment, and host functions as function pointer. For a reason I ignore, the UB wasn't showing himself until this PR and a modification in the Singlepass backend. But now it's fixed.

#### Impact on `Backing`

After the modification on the `typed_func` and the `vm` modules, this is the other core idea of this patch: Updating the `backing` module so that `vm::ImportedFunc` replaces `vm::Ctx` by `vm::FuncCtx`.

When creating `vm::ImportedFunc`, a new `vm::FuncCtx` is created and its pointer is used. We are purposely leaking `vm::FuncCtx` so that the pointer is always valid. Hence the specific `Drop` implementation on `ImportBacking` to dereference the pointer, and to drop it properly.

#### Impact on the backends

Since the structure of `vm::ImportedFunc` has changed, backends must be updated. We must deref `FuncCtx` to reach its `vmctx` field.

#### Impact on `Instance`

Because `vm::ImportedFunc` has changed, it has a minor impact on the `instance` module, nothing crazy though.


lib/runtime-core/src/backing.rs
```diff
@@ -605,8 +626,8 @@ fn import_functions(
             None => {
                 if imports.allow_missing_functions {
                     functions.push(vm::ImportedFunc {
-                        func: ::std::ptr::null(),
-                        vmctx: ::std::ptr::null_mut(),
+                        func: ptr::null(),
```
syrusakbary: Do we still need `func` here? (since we have `func_ctx` that perhaps already have it?)

lib/runtime-core-tests/tests/imports.rs
```diff
@@ -1,117 +1,201 @@
...
+                    result,
+                    expected_value
+                )
+            ),
+        }
+    };
+}
+
+const SHIFT: i32 = 10;
+const shift: i32 = 100;
```
syrusakbary: Can add some comments on the source explaining what these variables are for? (just a quick comment should be fine)

syrusakbary:
Loving this PR ❤️

lachlansneff:
Well done, Ivan 👍

lib/runtime-core/src/backing.rs
syrusakbary: No worries, thanks for the explanation! :)

Hywan:
bors try

Hywan:
This PR also removes the `mem::transmute(&())` in `typed_func.rs` >`wrap`, which reveals to be an undefined behavior.

lib/runtime-core/src/instance.rs
```diff
@@ -449,7 +457,7 @@ impl InstanceInner {
                 let imported_func = &self.import_backing.vm_functions[imported_func_index];
                 (
                     imported_func.func as *const _,
-                    Context::External(imported_func.vmctx),
+                    Context::External(unsafe { imported_func.func_ctx.as_ref() }.vmctx.as_ptr()),
```
nlewycky: Why does this need `unsafe`?

lib/runtime-core/src/vm.rs
```diff
@@ -498,22 +498,49 @@ impl Ctx {
...
-/// Used to provide type safety (ish) for passing around function pointers.
-/// The typesystem ensures this cannot be dereferenced since an
-/// empty enum cannot actually exist.
+/// Represents a function pointer. It is mostly used in the
+/// `typed_func` module within the `wrap` functions, to wrap imported
+/// functions.
 #[repr(C)]
-pub struct Func(InnerFunc);
+pub struct Func {
+    _private: [u8; 0],
```
nlewycky: This looks subtle. Are you relying on `#[repr(C)]` to produce a zero length array in the C API?

In C89 such things are UB (but a GCC extension), in C99 the same feature is available with different syntax, as flexible array members. Details: https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html

lib/runtime-core/src/instance.rs
Hywan: [`NonNull::as_ref`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.as_ref) is marked unsafe.

lib/runtime-core/src/vm.rs
Hywan: The idea is to store a pointer that can't be deref (hence the `_private`). Do you have an other idea?

Hywan:
> There's a bunch of odd formatting changes, some lines are rewrapped, comments are made very short. Is there some difference in your rustfmt configuration?

Nop, it's just the regular `rustfmt` without any changes. I use emacs though, maybe there is other rules.

Hywan:
bors try

lib/runtime-core/src/vm.rs
lachlansneff: I think the current solution for this kind of problem is to have an unsized `extern type`.

MarkMcCaskey: @lachlansneff `std::mem::size_of::<[u8; 0]>()` and a struct containing it are 0.  That seems like that couldn't work unless there's some rule for keeping zero-sized types around at runtime that I don't know about.

It seems like `#[repr(transparent)]` with a private inner field should work here.

Let me know if you know of a reason why that can't work; you have a lot of low-level Rust expertise!

Hywan: Updated to `pub struct Func(pub(self) *mut c_void);`

Hywan:
bors try

Hywan:
bors try

Hywan:
Description updated!

Ready to review :-). cc @MarkMcCaskey @nlewycky

Hywan:
bors try

Hywan:
bors try-

Hywan:
bors try

Hywan:
I hope all bugs with `field-offset` are now fixed.

bors r+

PR #934, opened 2019-11-07
In LLVM backend, track which floats are guaranteed to be arithmetic, which makes the canonicalization a no-op.

nlewycky:
# Description
This is a reimplementation of the patch in PR #651.

Extend state.rs ExtraInfo to track more information about floats. In addition to tracking whether the value has a pending canonicalization of NaNs, also track whether the value is known to be arithmetic (which includes infinities, regular values, and non-signalling NaNs (aka. "arithmetic NaNs" in the webassembly spec)). When the value is arithmetic, the correct sequence of operations to canonicalize the value is a no-op. Therefore, we create a lattice where pending+arithmetic=arithmetic.

Also, this extends the tracking to track all values, including non-SIMD integers. That's why there are more places where pending canonicalizations are applied.

Looking at c-wasm-simd128-example, this provides no performance change to the non-SIMD case (takes 58s on my noisy dev machine). The SIMD case drops from 46s to 29s.

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Special thanks to [@newpavlov](https://github.com/newpavlov) and [@Maxgy](https:
 - [#939](https://github.com/wasmerio/wasmer/pull/939) Fix bug causing attempts to append to files with WASI to delete the contents of the file
 - [#940](https://github.com/wasmerio/wasmer/pull/940) Update supported Rust version to 1.38+
 - [#923](https://github.com/wasmerio/wasmer/pull/923) Fix memory leak in the C API caused by an incorrect cast in `wasmer_trampoline_buffer_destroy`
+- [#934](https://github.com/wasmerio/wasmer/pull/934) Simplify float expressions in the LLVM backend.
 - [#921](https://github.com/wasmerio/wasmer/pull/921) In LLVM backend, annotate all memory accesses with TBAA metadata.
 - [#883](https://github.com/wasmerio/wasmer/pull/883) Allow floating point operations to have arbitrary inputs, even including SNaNs.
 - [#856](https://github.com/wasmerio/wasmer/pull/856) Expose methods in the runtime C API to get a WASI import object
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - [#990](https://github.com/wasmerio/wasmer/pull/990) Default wasmer CLI to `run`.  Wasmer will now attempt to parse unrecognized command line options as if they were applied to the run command: `wasmer mywasm.wasm --dir=.` now works!
 - [#987](https://github.com/wasmerio/wasmer/pull/987) Fix `runtime-c-api` header files when compiled by gnuc.
 - [#957](https://github.com/wasmerio/wasmer/pull/957) Change the meaning of `wasmer_wasi::is_wasi_module` to detect any type of WASI module, add support for new wasi snapshot_preview1
+- [#934](https://github.com/wasmerio/wasmer/pull/934) Simplify float expressions in the LLVM backend.
 
 ## 0.10.2 - 2019-11-18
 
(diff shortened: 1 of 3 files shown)


lib/llvm-backend/src/state.rs
```diff
@@ -68,22 +69,122 @@ impl ControlFrame {
...
 
 #[derive(Debug, Eq, PartialEq, Copy, Clone, Hash)]
-pub enum ExtraInfo {
-    None,
-
-    // This values is required to be arithmetic 32-bit NaN (or 32x4) by the WAsm
+pub struct ExtraInfo {
+    state: u8,
+}
+impl ExtraInfo {
```
losfair: Why do we need to distinguish between `f32` and `f64` NaNs in `ExtraInfo`, since the opcodes and values are already typed?

nlewycky: Because of the SIMD opcodes which aren't typed. They're all just v128, so we might go from a F64x2Add to a F32x4Mul with no cast in between.

`v128_to_f32x4` checks for a pending f64 canonicalization and applies it, while not applying a pending f32 canonicalization. `v128_to_f64x2` does the opposite.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Special thanks to [@newpavlov](https://github.com/newpavlov) and [@Maxgy](https:
 - [#939](https://github.com/wasmerio/wasmer/pull/939) Fix bug causing attempts to append to files with WASI to delete the contents of the file
 - [#940](https://github.com/wasmerio/wasmer/pull/940) Update supported Rust version to 1.38+
 - [#923](https://github.com/wasmerio/wasmer/pull/923) Fix memory leak in the C API caused by an incorrect cast in `wasmer_trampoline_buffer_destroy`
+- [#934](https://github.com/wasmerio/wasmer/pull/934) Simplify float expressions in the LLVM backend.
```
MarkMcCaskey: Changelog entry should be moved

lib/llvm-backend/src/state.rs
```diff
@@ -68,22 +69,122 @@ impl ControlFrame {
...
+    // This value is required to be arithmetic 32-bit NaN (or 32x4) by the WAsm
     // machine, but which might not be in the LLVM value. The conversion to
     // arithmetic NaN is pending. It is required for correctness.
-    PendingF32NaN,
+    pub fn pending_f32_nan() -> ExtraInfo {
```
MarkMcCaskey: These functions can be `const`

lib/llvm-backend/src/state.rs
```diff
@@ -68,22 +69,122 @@ impl ControlFrame {
...
+        ExtraInfo { state: 4 }
+    }
+
+    // This value either does not contain a 64-bit NaN, or it contains an
+    // arithmetic NaN. In SIMD, applies to both lanes.
+    pub fn arithmetic_f64() -> ExtraInfo {
+        ExtraInfo { state: 8 }
+    }
+
+    pub fn has_pending_f32_nan(&self) -> bool {
```
MarkMcCaskey: These functions may be able to be `const`

lib/llvm-backend/src/state.rs
```diff
@@ -68,22 +69,122 @@ impl ControlFrame {
...
+                & !(ExtraInfo::pending_f32_nan().state | ExtraInfo::pending_f64_nan().state),
+        }
+    }
 }
 impl Default for ExtraInfo {
     fn default() -> Self {
-        ExtraInfo::None
+        ExtraInfo { state: 0 }
+    }
+}
```
MarkMcCaskey: This is the same as `derive(Default)` on `ExtraInfo`

lib/llvm-backend/src/state.rs
```diff
@@ -68,22 +69,122 @@ impl ControlFrame {
...
+// Union two ExtraInfos.
+impl BitOr for ExtraInfo {
+    type Output = Self;
+
+    fn bitor(self, other: Self) -> Self {
+        assert!(!(self.has_pending_f32_nan() && other.has_pending_f64_nan()));
+        assert!(!(self.has_pending_f64_nan() && other.has_pending_f32_nan()));
```
MarkMcCaskey: If you only want these in debug mode, use `debug_assert!`, these `asserts` will run on every use

lib/llvm-backend/src/code.rs
```diff
@@ -6054,7 +6395,7 @@ impl FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator {
                     Some(0),
                 );
                 let old = builder.build_int_z_extend(old, intrinsics.i64_ty, &state.var_name());
-                state.push1(old);
+                state.push1_extra(old, ExtraInfo::arithmetic_f64());
```
MarkMcCaskey: `Operator::I32AtomicRmwAdd` is different than the sub variant here, the sub has `arithmetic_f32`

CHANGELOG.md
nlewycky: Done.

lib/llvm-backend/src/state.rs
nlewycky: Done.

lib/llvm-backend/src/state.rs
nlewycky: They are. Done.

lib/llvm-backend/src/state.rs
nlewycky: So it is! Done.

lib/llvm-backend/src/state.rs
nlewycky: Good idea, done.

PR #938, opened 2019-11-08
Fuzz all the backends.

nlewycky:


pventuzelo:
Nice, i have already start calling all the backends in one fuzzers and check the result to do differential testings. I'm gonna use our code for compiler, your fonction get_llvm_compiler seems to be faster than mine.

PR #939, opened 2019-11-08
Fix WASI append bug, add test

### LINKED ISSUE #936: Appending to file does not work
I am trying to run a simple wasm program, see code [here](https://github.com/WilliamMartini/WASM/blob/master/rust/main.rs)

I am running: ```wasmer run --dir=. demo.wasm '{"a":"txt", "b":1}'```  
which should append ```txt 1``` to a ```log.txt``` file.  

However, the data is not appending but instead writing over. The same program compiled with native Rust (non-wasm) appends properly.

MarkMcCaskey:
resolves #936


lib/wasi-tests/build/wasitests.rs
```diff
@@ -83,7 +83,7 @@ pub fn compile(file: &str, ignores: &HashSet<String>) -> Option<String> {
         .arg("+nightly")
         .arg("--target=wasm32-wasi")
         .arg("-C")
-        .arg("opt-level=s")
+        .arg("opt-level=z")
```
nlewycky: This seems like it's part of a different change?

syrusakbary:
Thank you for the tests 🙏

lib/wasi-tests/build/wasitests.rs
MarkMcCaskey: It is, but using a new version of nightly locally means all the generated wasm modules will change anyways, so I figured I might as well make this change that I've been meaning to make for a while anyways

MarkMcCaskey:
bors retry

PR #942, opened 2019-11-10
Deny missing docs in runtime core and add missing docs

bjfish:
# Description
Deny missing docs in runtime core and add missing docs

# Review


lib/runtime-core/src/error.rs
```diff
@@ -1,13 +1,22 @@
+//! The error module contains the data structures and helper functions used to implement errors that
+//! are produced and returned from the wasmer runtime core.
 use crate::types::{FuncSig, GlobalDescriptor, MemoryDescriptor, TableDescriptor, Type};
 use core::borrow::Borrow;
 use std::any::Any;
 
+/// Aliases the standard `Result` type as `Result` within this module.
```
nlewycky: This comment is really a restatement of the code, when it should be a description of what this type is used for. Maybe something like "Result of an attempt to compile the provided WebAssembly module into a `Module`" to copy from the definition of `CompileError`.

lib/runtime-core/src/state.rs
```diff
@@ -1022,62 +1111,101 @@ pub mod x64 {
         unreachable!();
     }
 
+    /// A kind of GPR register
     #[repr(u8)]
     #[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
     pub enum GPR {
+        /// RAX Register
```
nlewycky: Is there a way to suppress the extra comments here? These aren't adding value, and they're doubling the number of lines.

lib/runtime-core/src/sys/unix/memory.rs
```diff
@@ -238,13 +249,19 @@ impl Clone for Memory {
...
 
+/// Kinds of memory protection.
 #[derive(Serialize, Deserialize, Debug, Copy, Clone, PartialEq, Eq)]
 #[allow(dead_code)]
 pub enum Protect {
+    /// No protection.
     None,
+    /// Read protection.
     Read,
+    /// Read/write protection.
```
nlewycky: I think this description is ambiguous (does "protection" refer to what is forbidden, so this is execute-only memory?). What it means is "read and write are permitted" with the implication that execution is not. It should say that. Suggestion: "Read and write only"

lib/runtime-core/src/vm.rs
```diff
@@ -547,15 +577,18 @@ pub struct LocalTable {
 unsafe impl Send for LocalTable {}
 
 impl LocalTable {
+    /// Offset base.
```
nlewycky: As written this sounds like the base of an offset. What this is, is the offset into the structure, at which the field "base" lives. This is used in singlepass-generated code to look up where the "base" pointer is, given a pointer to a LocalTable. Suggestion, "offset to base"?

lib/runtime-core/src/state.rs
bjfish: @nlewycky They can be suppressed individually using `#[doc(hidden)]` but that would be the same number of lines. I think they would all be suppressed with `#[doc(hidden)]` on the enum but then we wouldn't get a doc for that.

lib/runtime-core/src/sys/unix/memory.rs
bjfish: Updated these with your suggestion.

lib/runtime-core/src/error.rs
bjfish: I've updated these aliases using your suggestion.

lib/runtime-core/src/vm.rs
bjfish: I've updated this with your suggestion `Offset to base`.

bjfish:
bors try

bjfish:
bors try

Hywan:
👏

PR #948, opened 2019-11-11
Patch page_size to point to a fork that upgrades its dependency to winapi crate 0.3.8

hrydgard:
Until Elzair merges our PR, https://github.com/Elzair/page_size_rs/pull/1. Does not seem active.

For build perf, old winapi (<= 2.8) is super slow to build since it by defaults includes everything. Cargo.lock shrinks a little too.

Saves about 30 seconds of CPU during build (though with enough cores there's little difference in total built time, but if wasmer is built as part of a larger project, the difference appears).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -582,15 +582,6 @@ name = "kernel-net"
 version = "0.1.0"
 
 [[package]]
-name = "kernel32-sys"
-version = "0.2.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "winapi 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
- "winapi-build 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
 name = "lazy_static"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


syrusakbary:
@hrydgard I think if we merge this change, we will not be able to publish the crate to crates.io.

We'd be happy depending on a published fork on crates.io though, if you want to do a fork, publish it to crates.io and update the PR we will merge it promptly :)

hrydgard:
@syrusakbary In my opinion it's kind of a odd little crate. Anytime you care about what the page size is, you're likely also using other platform-specific functionality anyway, so the few extra lines to determine the page size using platform-specific code should be no big deal and not enough to motivate adding an external dependency.

So instead of us taking over maintenance of such a strange little crate, it might a better idea for you guys to get rid of the dependency on this crate, and simply take the contents and merge it into your code directly at the call sites, or just grab the lib.rs and rename it page_size.rs and just put it next to your other files?

hrydgard:
https://github.com/Elzair/page_size_rs has now been bumped to 0.4.2 which fixes this issue. Since you guys specify the dependency as "0.4", you can probably just close this after the next time you run cargo update.

syrusakbary:
We just merged #1093 ...things should be good now in master :)

PR #952, opened 2019-11-11
Generate C preprocessor code to hide things not on Windows

### LINKED ISSUE #804: C API doesn't mention platform restrictions regarding trampoline
### Describe the bug
The wasmer.h and wasmer.hh headers list the structs and functions associated with the trampoline features, however this feature is only enabled on unix x64 platforms.
The tests runtime-c-api/test-import-function and runtime-c-api/test-import-function-callinfo do not link on windows.

### Expected behavior
The header files should state that these aren't available or even better, remove them with macros.

MarkMcCaskey:
resolves #804 

# Review

### DIFF
--- a/lib/runtime-c-api/wasmer.hh
+++ b/lib/runtime-c-api/wasmer.hh
@@ -1,3 +1,20 @@
+
+#ifndef WASMER_H_MACROS
+#define WASMER_H_MACROS
+#if MSVC
+#ifdef _M_AMD64
+#define ARCH_X86_64
+#endif
+#endif
+
+#if GCC
+#ifdef __x86_64__
+#define ARCH_X86_64
+#endif
+#endif
+#endif // WASMER_H_MACROS
+
+
 #ifndef WASMER_H
 #define WASMER_H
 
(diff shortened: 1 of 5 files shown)


lib/runtime-c-api/wasmer.hh
```diff
@@ -612,32 +635,46 @@ uint32_t wasmer_table_length(wasmer_table_t *table);
 /// and `wasmer_last_error_message` to get an error message.
 wasmer_result_t wasmer_table_new(wasmer_table_t **table, wasmer_limits_t limits);
 
+#if (!defined(_WIN32) && defined(ARCH_X86_64))
```
nlewycky: Optional: these are clearly a group, just put a single #if around them to disable all these related trampoline functions for windows?

MarkMcCaskey: Yeah, it's unfortunate that it did that; if we manually edit the header file though then we have to do it after every time it's generated

MarkMcCaskey: I put it at the top level in Rust

PR #957, opened 2019-11-12
Add support for new WASI snapshot, backwards compat too

MarkMcCaskey:
# Review


Cargo.toml
```diff
@@ -86,13 +86,6 @@ backend-cranelift = [
     "wasmer-middleware-common-tests/clif",
     "wasmer-wasi-tests/clif"
 ]
-backend-llvm = [
-    "wasmer-llvm-backend",
-    "wasmer-runtime-core/backend-llvm",
-    "wasmer-runtime/llvm",
-    "wasmer-middleware-common-tests/llvm",
-    "wasmer-wasi-tests/llvm"
-]
```
syrusakbary: ?

lib/runtime-c-api/Cargo.toml
```diff
@@ -34,7 +34,6 @@ optional = true
 default = ["cranelift-backend", "wasi"]
 debug = ["wasmer-runtime/debug"]
 cranelift-backend = ["wasmer-runtime/cranelift", "wasmer-runtime/default-backend-cranelift"]
-llvm-backend = ["wasmer-runtime/llvm", "wasmer-runtime/default-backend-llvm"]
```
syrusakbary: ?

lib/wasi/src/utils.rs
```diff
@@ -1,18 +1,37 @@
...
+/// The version of WASI.  This is determined by the namespace string
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum WasiVersion {
+    /// "wasi_unstable"
+    Snapshot0,
+    /// "wasi_snapshot_preview1"
+    Snapshot1,
+}
+
+pub fn get_wasi_version(module: &Module) -> Option<WasiVersion> {
```
bjfish: Please rustdoc the public functions/types or suppress with doc hidden.

Cargo.lock
```diff
@@ -1504,7 +1504,6 @@ dependencies = [
  "tempfile 3.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "wabt 0.9.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "wasmer-clif-backend 0.10.1",
- "wasmer-llvm-backend 0.10.1",
```
syrusakbary: Probably you don't want to commit this either

lib/wasi/src/utils.rs
```diff
@@ -1,18 +1,37 @@
...
+    let mut import_iter = module
+        .info()
+        .imported_functions
+        .iter()
+        .map(|(_, import_name)| import_name.namespace_index);
+
+    // returns None if empty
+    let first = import_iter.next()?;
+    if import_iter.all(|idx| idx == first) {
```
syrusakbary: What is this `.all(|idx| idx == first)` doing? 

Because Emscripten implements part of `wasi` (setting some of the imported functions  as `wasi_unstable`), we need to check that all the imports have this namespace (not just one). That way we can differentiate Emscripten and WASI modules

lib/wasi/src/utils.rs
```diff
@@ -1,18 +1,37 @@
...
-        return false;
-    }
-    for (_, import_name) in &module.info().imported_functions {
-        let namespace = module
-            .info()
-            .namespace_table
-            .get(import_name.namespace_index);
-        if namespace != "wasi_unstable" {
-            return false;
+    get_wasi_version(module) == Some(WasiVersion::Snapshot1)
```
syrusakbary: I would do it just `.is_some()` ? (both snapshot0 and snapshot1 are wasi modules)

MarkMcCaskey: I was thinking that I'd attempt to preserve more of the semantics of the function call; either way this is a breaking change -- I think you're right that it doesn't make sense to keep the function name as what it is and not have that behavior.  I'll update it

lib/wasi/src/utils.rs
MarkMcCaskey: It checks that all namespaces have the same index, which I understood to be a faster way to check if they all share a namespace than looking up every one and doing string comparisons.

After it guarantees that they're all the same, it does logic on it.  I can add comments to clarify this logic

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/wasi/src/utils.rs
Hywan: Actually, it creates a bug. If the import namespaces are the following: `wasi_unstable` and `env`, we have 2 different namespaces, and `get_wasi_version` will return `None`. Working on a fix.

MarkMcCaskey: @Hywan as mentioned in #1028 , I think this is not a bug and that the behavior in #1028 is much closer to a bug because we'll start calling things which are not WASI (won't run with WASI imports) WASI and have no way to detect real WASI anymore.

Hywan: Is it specified in the specification that, in case of WASI, all imports must land in the same namespace? I think it is very restrictive, and would be break some design (like blockchain or stuff like that that have special namespaces too).

PR #960, opened 2019-11-13
feat(runtime-c-api) Add support for clang in `WASMER_H_MACROS`

Hywan:
In #952, the `WASMER_H_MACROS` constant has been defined. The `ARCH_X86_64` constant is defined under 2 conditions: If the compiler is MSVC + `_M_AMD64` is defined, or if the compiler is GCC + `__x86_64__` is defined.

Clang is missing. And it breaks some projects (like https://github.com/wasmerio/php-ext-wasm or https://github.com/wasmerio/go-ext-wasm for instance).

This patch supports Clang.

### DIFF
--- a/lib/runtime-c-api/build.rs
+++ b/lib/runtime-c-api/build.rs
@@ -21,7 +21,7 @@ fn main() {
     #endif
   #endif
 
-  #if defined(GCC)
+  #if defined(GCC) || defined(__clang__)
     #if defined(__x86_64__)
       #define ARCH_X86_64
     #endif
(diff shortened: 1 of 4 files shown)


lib/runtime-c-api/build.rs
```diff
@@ -13,20 +13,20 @@ fn main() {
...
+    #if defined(_M_AMD64)
+      #define ARCH_X86_64
+    #endif
+  #endif
 
-#if GCC
-#ifdef __x86_64__
-#define ARCH_X86_64
-#endif
+  #if defined(GCC) || defined(__clang__)
```
Hywan: Bug fix is here: `defined(__clang__)`.

Hywan:
bors try

PR #961, opened 2019-11-13
feat(runtime-core) `func!` supports closures

Hywan:
This patch updates the `func!` macro so that it can consume closures.

This patch also improves the documentation of the `func!` by explaining how to use it with examples (with function pointers or closures).


Hywan:
bors try

PR #968, opened 2019-11-14
Added invoke option to the command

syrusakbary:
# Description 

This PR adds an `--invoke` to the Wasmer CLI

# Review

### DIFF
--- a/src/bin/wasmer.rs
+++ b/src/bin/wasmer.rs
@@ -687,7 +687,10 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
                 args.push(Value::I32(x));
             }
 
-            let invoke_fn = options.invoke.clone().unwrap_or("main".to_owned());
+            let invoke_fn = match options.invoke.as_ref() {
+                Some(fun) => fun,
+                _ => "main",
+            };
             instance
                 .dyn_func(&invoke_fn)
                 .map_err(|e| format!("{:?}", e))?
--- a/src/bin/wasmer.rs
+++ b/src/bin/wasmer.rs
@@ -132,6 +132,10 @@ struct Run {
     )]
     backend: Backend,
 
+    /// Invoke a specified function
+    #[structopt(long = "invoke", short = "i")]
+    invoke: Option<String>,
+
     /// Emscripten symbol map
     #[structopt(long = "em-symbol-map", parse(from_os_str), group = "emscripten")]
     em_symbol_map: Option<PathBuf>,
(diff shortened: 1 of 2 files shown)


nlewycky:
Why invoke and not `-e` for entry?

src/bin/wasmer.rs
```diff
@@ -132,6 +132,10 @@ struct Run {
     )]
     backend: Backend,
 
+    /// Invoke a specified function
```
nlewycky: Maybe "specify the name of the start function" or "of the main function" of something like that?

bjfish:
It would be nice if this could take integer parameters like --invoke fun_name 2,3,4 ... this was also previously suggested to be a separate command `wasmer invoke fun_name 2,3,4`, and to print the result.

syrusakbary:
@nlewycky 
> Why invoke and not -e for entry?

It's not an entry function. But a function that you might want to call it alone. In general I'm following the naming of other implementations 

@bjfish 

> It would be nice if this could take integer parameters like --invoke fun_name 2,3,4 ... this was also previously suggested to be a separate command wasmer invoke fun_name 2,3,4, and to print the result.

Right now it will take the arguments provided in the CLI (transformed to the necessary i32, i64, ... values). Such as:

```sh
wasmer run myfile.wasm --invoke fun 1 2 3
```

src/bin/wasmer.rs
syrusakbary: Is not a main function, but a raw function that you might want to invoke.

src/bin/wasmer.rs
```diff
@@ -132,6 +132,10 @@ struct Run {
...
+    #[structopt(long = "invoke", short = "i")]
+    invoke: Option<String>,
```
MarkMcCaskey: ⛳️

Consider: 

```Rust
#[structopt(long = "invoke", short = "i", default = "main")]
invoke: String
``` 

This is less flexible, but for the code as it is now, this is a cleaner solution

src/bin/wasmer.rs
```diff
@@ -683,8 +687,12 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
                 args.push(Value::I32(x));
             }
 
+            let invoke_fn = match options.invoke.as_ref() {
+                Some(fun) => fun,
+                _ => "main",
+            };
```
MarkMcCaskey: If you don't take the above advice:

⛳️ 
```Rust
let invoke_fn = options.invoke.as_ref().unwrap_or("main");
```

src/bin/wasmer.rs
syrusakbary: I don't like that much, because depending on the integration you use (emscripten, wasi, ...) the default might vary. Will go through the other approach

src/bin/wasmer.rs
syrusakbary:
```suggestion
            let invoke_fn = options.invoke.as_ref().unwrap_or("main");
```

MarkMcCaskey: The problem with suggestions is that they only apply to 1 line and standard Rust fmt encourages splitting things across many lines.  It'd be extremely useful if they could apply to multiple lines and be batch applied; like a merge request of a PR but done inline

PR #973, opened 2019-11-16
Add sign extension spec tests; add sign extension to singlepass

MarkMcCaskey:
Adds missing spectests from official repo, implements instructions for singlepass

# Review


syrusakbary:
Nice!

nlewycky:
Sign extension ops are pre-standard still? Update the PrestandardFeatures struct in src/bin/wasmer.rs and propagate the flag through from the runner to the backends. Should be easy to search "simd" and follow the same pattern.

MarkMcCaskey:
> Sign extension ops are pre-standard still? Update the PrestandardFeatures struct in src/bin/wasmer.rs and propagate the flag through from the runner to the backends. Should be easy to search "simd" and follow the same pattern.

They are, they're already implemented in LLVM and Clif though and sounds good!

MarkMcCaskey:
bors try

src/bin/wasmer.rs
```diff
@@ -91,11 +91,41 @@ struct PrestandardFeatures {
     #[structopt(long = "enable-threads")]
     threads: bool,
 
+    /// Enable support for the threads proposal.
```
losfair: threads -> sign extension?

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -2948,6 +2949,125 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                let ret = self.machine.acquire_locations(
+                    a,
+                    &[(WpType::I32, MachineValue::WasmStack(self.value_stack.len()))],
+                    false,
+                )[0];
+                self.value_stack.push(ret);
+
+                let tmpg1 = self.machine.acquire_temp_gpr().unwrap();
+                a.emit_mov(Size::S32, loc, Location::GPR(tmpg1));
+                a.emit_and(Size::S32, Location::Imm32(0xFF), Location::GPR(tmpg1));
```
losfair: Do we really need to `and` the value with a corresponding bitmask first, while `mov{zx/sx}` is aware of the width?

MarkMcCaskey: Good call!  It seems that we don't!

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

PR #974, opened 2019-11-16
Update to Rust 1.39

losfair:
This PR updates the default Rust version to 1.39.

Necessary for code formatted with newer versions of `rustfmt` to pass the CI lint check.


Hywan:
bors try

MarkMcCaskey:
This breaks our Rustc stability guarantee and we should revert it unless we have a very good reason not to

PR #978, opened 2019-11-18
test(runtime-core) Test closures in `Func::new`

Hywan:
This patch contains 2 commits:

1. Test closures in `Func::new`
2. Fix a warning about `unused_parens` (since Rust 1.39, the behavior of `unused_parens` has changed a little bit).


lib/runtime-core/src/typed_func.rs
```diff
@@ -819,8 +818,14 @@ mod tests {
...
-                let _func = Func::new(with_vmctx);
-                let _func = Func::new(without_vmctx);
+                let _ = Func::new(with_vmctx);
+                let _ = Func::new(without_vmctx);
+                let _ = Func::new(|_: &mut vm::Ctx, $($x: i32),*| -> i32 {
+                    vec![$($x),*].iter().sum()
+                });
+                let _ = Func::new(|$($x: i32),*| -> i32 {
+                    vec![$($x),*].iter().sum()
+                });
```
bjfish: Do we call these and assert the results in some tests somewhere?

Hywan: Assertions here are the type-system. If it compiles, it means `Func` supports the described functions.

Issue #984, opened 2019-11-20
How to set proper architecture for C tests


AdamSLevy:
### Summary
 
I am attempting to run the runtime-c-api tests but I'm running into this compilation issue. I'm on an up to date Arch Linux system. I've run `rustup default nightly`.

```
$ cargo test -- --nocapture
...
[ 23%] Built target test-module-imports
Scanning dependencies of target test-import-function
[ 26%] Building C object CMakeFiles/test-import-function.dir/test-import-function.c.o

stderr:
/home/aslevy/repos/wasmer/lib/runtime-c-api/tests/test-import-function.c: In function ‘print_str’:
/home/aslevy/repos/wasmer/lib/runtime-c-api/tests/test-import-function.c:23:47: error: implicit declaration of function ‘wasmer_trampoline_get_context’ [-Werror=implicit-function-declaration]
   23 |     struct print_str_context *local_context = wasmer_trampoline_get_context();
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/aslevy/repos/wasmer/lib/runtime-c-api/tests/test-import-function.c:23:47: error: initialization of ‘struct print_str_context *’ from ‘int’ makes pointer from integer without a cast [-Werror=int-conversion]
/home/aslevy/repos/wasmer/lib/runtime-c-api/tests/test-import-function.c: In function ‘main’:
/home/aslevy/repos/wasmer/lib/runtime-c-api/tests/test-import-function.c:51:5: error: unknown type name ‘wasmer_trampoline_buffer_builder_t’
   51 |     wasmer_trampoline_buffer_builder_t *tbb = wasmer_trampoline_buffer_builder_new();
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/aslevy/repos/wasmer/lib/runtime-c-api/tests/test-import-function.c:51:47: error: implicit declaration of function ‘wasmer_trampoline_buffer_builder_new’ [-Werror=implicit-function-declaration]
   51 |     wasmer_trampoline_buffer_builder_t *tbb = wasmer_trampoline_buffer_builder_new();
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/aslevy/repos/wasmer/lib/runtime-c-api/tests/test-import-function.c:51:47: error: initialization of ‘int *’ from ‘int’ makes pointer from integer without a cast [-Werror=int-conversion]
[...]

PR #987, opened 2019-11-20
Fix runtime c api header preprocessor gcc

Hywan:
`ARCH_X86_64` is correctly defined for GCC or clang, but gnuc was
missing. This patch fixes that.

Address https://github.com/wasmerio/php-ext-wasm/issues/93
Fix https://github.com/wasmerio/wasmer/issues/984


Hywan:
bors try

Issue #993, opened 2019-11-21
Allow different Instances to have distinct limits from the same compiled module


AdamSLevy:
### Motivation

As I understand the current metering code, a fixed limit is hard coded at compile time. This means that if different limits are desired for different instantiations of the same module, it must be recompiled each time with the distinct limit. 

Allowing a dynamic limit would allow caching the serialized module for many applications that use a dynamic limit per instance like smart contracts on a blockchain.

### Proposed solution

Use an additional `wasmer_runtime_core::vm::InternalField` to store a Limit that may be set at instantiation through an additional API: 
```rust
pub fn reset_limit(instance: &mut Instance, limit u64) {
    instance.set_internal(&INTERNAL_FIELD_LIMIT, limit)
}
```

Define an additional static InternalField is declared in the module so that we change this:
https://github.com/wasmerio/wasmer/blob/9e144c623b113e4dc83e048502dd4d43ad61217b/lib/middleware-common/src/metering.rs#L9

to this

```rust
static INTERNAL_FIELD_LIMIT: InternalField = InternalField::allocate();
static INTERNAL_FIELD_USED: InternalField = InternalField::allocate();
```

And then change this:

https://github.com/wasmerio/wasmer/blob/9e144c623b113e4dc83e048502dd4d43ad61217b/lib/middleware-common/src/metering.rs#L40-L110

to this:

```rust
impl FunctionMiddleware for Metering {
    type Error = String;
    fn feed_event<'a, 'b: 'a>(
        &mut self,
        op: Event<'a, 'b>,
        _module_info: &ModuleInfo,
        sink: &mut EventSink<'a, 'b>,
    ) -> Result<(), Self::Error> {
        match op {
            Event::Internal(InternalEvent::FunctionBegin(_)) => {
                self.current_block = 0;
            }
            Event::Wasm(&ref op) | Event::WasmOwned(ref op) => {
                self.current_block += 1;
                match *op {
                    Operator::Loop { .. }
                    | Operator::Block { .. }
                    | Operator::End
                    | Operator::If { .. }
                    | Operator::Else
                    | Operator::Unreachable
                    | Operator::Br { .. }
                    | Operator::BrTable { .. }
                    | Operator::BrIf { .. }
                    | Operator::Call { .. }
                    | Operator::CallIndirect { .. }
                    | Operator::Return => {
                        sink.push(Event::Internal(InternalEvent::GetInternal(
                            INTERNAL_FIELD_USED.index() as _,
                        )));
[...]

PR #996, opened 2019-11-21
feat(middleware_common::metering): Dynamic limit

AdamSLevy:
Add a dynamic limit to the metering::Metering middleware.

# Description

Previously a limit was hardcoded at compile time. This commit adds an
additional InternalField to track a dynamic limit unique to each
instance. This commit adds fn metering::set_limit(&mut Instance,  u64)
to allow limits to be changed.

This allows a module compiled with metering to be cached and re-used to
instantiate Instances with different limits.

implements #993

# Review

### DIFF
--- a/lib/middleware-common-tests/benches/metering_benchmark.rs
+++ b/lib/middleware-common-tests/benches/metering_benchmark.rs
@@ -134,13 +134,13 @@ static WAT_GAS: &'static str = r#"
         "#;
 
 #[cfg(feature = "llvm")]
-fn get_compiler(limit: u64, metering: bool) -> impl Compiler {
+fn get_compiler(metering: bool) -> impl Compiler {
     use wasmer_llvm_backend::ModuleCodeGenerator;
     use wasmer_runtime_core::codegen::{MiddlewareChain, StreamingCompiler};
     let c: StreamingCompiler<ModuleCodeGenerator, _, _, _, _> = StreamingCompiler::new(move || {
         let mut chain = MiddlewareChain::new();
         if metering {
-            chain.push(Metering::new(limit));
+            chain.push(Metering::new());
         }
         chain
     });
@@ -149,13 +149,13 @@ fn get_compiler(limit: u64, metering: bool) -> impl Compiler {
 }
 
 #[cfg(feature = "singlepass")]
-fn get_compiler(limit: u64, metering: bool) -> impl Compiler {
+fn get_compiler(metering: bool) -> impl Compiler {
     use wasmer_runtime_core::codegen::{MiddlewareChain, StreamingCompiler};
     use wasmer_singlepass_backend::ModuleCodeGenerator as SinglePassMCG;
     let c: StreamingCompiler<SinglePassMCG, _, _, _, _> = StreamingCompiler::new(move || {
         let mut chain = MiddlewareChain::new();
         if metering {
-            chain.push(Metering::new(limit));
+            chain.push(Metering::new());
         }
         chain
     });
@@ -166,7 +166,7 @@ fn get_compiler(limit: u64, metering: bool) -> impl Compiler {
 compile_error!("compiler not specified, activate a compiler via features");
 
 #[cfg(feature = "clif")]
-fn get_compiler(_limit: u64, metering: bool) -> impl Compiler {
+fn get_compiler(metering: bool) -> impl Compiler {
     compile_error!("cranelift does not implement metering");
     use wasmer_clif_backend::CraneliftCompiler;
     CraneliftCompiler::new()
@@ -185,7 +185,7 @@ fn bench_metering(c: &mut Criterion) {
     c.bench(
         "Meter",
         Benchmark::new("No Metering", |b| {
-            let compiler = get_compiler(0, false);
+            let compiler = get_compiler(false);
             let wasm_binary = wat2wasm(WAT).unwrap();
             let module = compile_with(&wasm_binary, &compiler).unwrap();
             let import_object = imports! {};
@@ -194,7 +194,7 @@ fn bench_metering(c: &mut Criterion) {
             b.iter(|| black_box(add_to.call(100, 4)))
         })
         .with_function("Gas Metering", |b| {
-            let compiler = get_compiler(0, false);
+            let compiler = get_compiler(false);
             let gas_wasm_binary = wat2wasm(WAT_GAS).unwrap();
             let gas_module = compile_with(&gas_wasm_binary, &compiler).unwrap();
             let gas_import_object = imports! {
@@ -207,13 +207,14 @@ fn bench_metering(c: &mut Criterion) {
             b.iter(|| black_box(gas_add_to.call(100, 4)))
         })
         .with_function("Built-in Metering", |b| {
-            let metering_compiler = get_compiler(std::u64::MAX, true);
+            let metering_compiler = get_compiler(true);
             let wasm_binary = wat2wasm(WAT).unwrap();
             let metering_module = compile_with(&wasm_binary, &metering_compiler).unwrap();
             let metering_import_object = imports! {};
             let mut metering_instance = metering_module
                 .instantiate(&metering_import_object)
                 .unwrap();
+            metering::set_limit(&mut metering_instance, std::u64::MAX);
             metering::set_points_used(&mut metering_instance, 0u64);
             let metering_add_to: Func<(i32, i32), i32> = metering_instance.func("add_to").unwrap();
             b.iter(|| black_box(metering_add_to.call(100, 4)))
--- a/lib/middleware-common-tests/src/lib.rs
+++ b/lib/middleware-common-tests/src/lib.rs
@@ -12,22 +12,22 @@ mod tests {
     };
 
     #[cfg(feature = "llvm")]
-    fn get_compiler(limit: u64) -> (impl Compiler, Backend) {
+    fn get_compiler() -> (impl Compiler, Backend) {
         use wasmer_llvm_backend::ModuleCodeGenerator as LLVMMCG;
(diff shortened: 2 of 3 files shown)


AdamSLevy:
I have been experimenting with this patch and while it passes all tests I have figured out that this API will not allow a limit to be set for the "_start" function of a module, if defined.

This is because the limit may only be set via the `wasmer_runtime_core::Instance` or `wasmer_runtime_core::Ctx`, which isn't available to the user until they have an initialized instance, which I believe invokes the start function.

Thoughts?

syrusakbary:
I think it will be better to chime in more people that are using it (or might be using it soon).

@ethanfrey @mikhailOK @nhynes @YaronWittenstein @camilbancioiu @michaelvoronov what are your thoughts?

AdamSLevy:
Regarding the limitation I just mentioned, there must be some way to set the limit prior to start is called during Instance initialization. Haven't looked into it yet though.

I'd like to hear what others think.

ethanfrey:
Comments from reading test:

That's a very interesting idea, and nice to see someone else working on these issues :smile:

I wasn't sure the original design goal, but I just worked around it, given that `limit - used` is remaining limit. What I did was compile everything with a huge gas limit (1_000_000_000). Then before I execute a contract, determine what the limit should be (eg. 950_000), and `set_points_used` to the difference (eg. 999_050_000).

This seems a bit backwards, but works pretty good (as long as I never try to set a limit over the hardcoded compile limit). I am also happy for a more ergonomic approach. Note that I often cache instances to use over multiple function calls, and set a gas_limit before each call, rather than once per instance.

From the code:

I don't understand the metering framework, but this does seem to enable what I want. My big question is what happens if you execute a contract without calling `set_limit` first. Is default 0? I think it would be good to have a default very high - so not setting a limit is just allowing all contracts. But maybe it is better to default to 0. In any case, this should be defined and demonstrated with a test.

For my use case I could do something like this, right?

```rust
let module = compile_with(&wasm_binary, &get_compiler()).unwrap(); 
let mut instance = module.instantiate(&import_object).unwrap();

set_limit(instance, REAL_LIMIT_1);
set_points_used(instance, 0u64);
call_f(instance, args);

set_limit(instance, REAL_LIMIT_2);
set_points_used(instance, 0u64);
call_f(instance, args2);
```

Maybe something like `reset_limit` that sets the limit and sets points used to 0 at the same time would be helpful. I mean, if you are changing the limit, it is very likely a new environment and it doesn't make much sense to track old points used. Such an approach would definitely simplify my calling code and remove artificial hardcoded limit.

nhynes:
> what happens if you execute a contract without calling `set_limit` first

[the default is soft infinity](https://github.com/wasmerio/wasmer/blob/72864934f922851fb26d68c67e882733dbcfda94/lib/runtime-core/src/vm.rs#L154), which is good.

This PR would be fine for my workflow. I might only suggest a convenience method, `pub fn with_limit(points: usize) -> Self`, in support of metering the sub-instances of dynamically linked modules.

Also, I'm curious why the methods are in the module rather than instance. `Metering` isn't `Deref` anything, so there shouldn't be any name collision.

AdamSLevy:
@ethanfrey @nhynes Thanks for taking the time to review and chime in. 

> Note that I often cache instances to use over multiple function calls, and set a gas_limit before each call, rather than once per instance.

This is also something I want to be able to do. This seems to work as I expect in my initial testing, which was just me hacking around with the existing tests so far really.

> My big question is what happens if you execute a contract without calling set_limit first. Is default 0? I think it would be good to have a default very high - so not setting a limit is just allowing all contracts. But maybe it is better to default to 0. In any case, this should be defined and demonstrated with a test.

> the default is soft infinity, which is good.

@nhynes Thanks for pointing that out. I actually didn't yet know what that was initialized to.

AdamSLevy:
I also just want to bring up issue #999 that I just opened yesterday. The current metering approach is not perfect and has some flaws that I think should be addressed. Anyone using it for smart contracts may expect more exact metering akin to eWasm's approach. In my issue I propose a solution and would love to hear feedback from folks on this as well before I get too far into implementing a new PR for this.

AdamSLevy:
> Maybe something like reset_limit that sets the limit and sets points used to 0 at the same time would be helpful. I mean, if you are changing the limit, it is very likely a new environment and it doesn't make much sense to track old points used. Such an approach would definitely simplify my calling code and remove artificial hardcoded limit.

@ethanfrey I agree this is a good idea. I'm going to add this.

AdamSLevy:
rebased on the latest release

AdamSLevy:
> the default is soft infinity, which is good.

@nhynes the default is actually not set by the line you linked. That line is initializing what will be a unique index for the InternalField to a max invalid value, not the InternalFields value itself is.

There is still the question of the "start" function. Could a dev chime in about that? @syrusakbary How would a dynamically set limit like this affect "start" functions in modules?

AdamSLevy:
This has been rebased on master. I also added APIs so that the limit can be accessed and set from a Ctx as well as an Instance.

ethanfrey:
> This has been rebased on master. I also added APIs so that the limit can be accessed and set from a Ctx as well as an Instance.

This PR looks very good and as we are enhancing our gas metering for callbacks, this would be a very useful (not just interesting) feature to have - to control the limit and the gas_used separately. 

@syrusakbary I don't see any real arguments against this, just a lack of review and merging. I would be interested if this is something you see out-of-line with the wasmer design, or something that just needs a bit of polish to be merged. If you are willing to merge this, I am willing to update the code to current master unless @AdamSLevy feels like doing it again. It would be good to get a clear statement from the core wasmer team on this PR in any case.

syrusakbary:
@ethanfrey we have taken a look and it seems we would be good to merge once conflicts are resolved.
However let me chime more people, and I'll get back to you in a day :)

ethanfrey:
Thank you @syrusakbary 

I'm also happy to take another approach if there are some concerns. So the questions are:

* Is the feature desired by wasmer?
* Is this approach desired by wasmer?
* If both yes, what needs to be improved to merge it?
* If the second is no, what approach do you recommend?

AdamSLevy:
Very cool to see this move forward. @ethanfrey I don't have the bandwidth at this moment to update this PR so please feel free.

BTW to my knowledge the bug in the actual implementation of the Metering that I raised in issue #999 is still present. I have a fix for this in #1041 but there are some issues with merging it in as implemented due to some lifetime issues that I had to work around. That PR is more nuanced and I would be more willing and happy to do my part to help clarify and support getting that improved metering algorithm merged. My improved algorithm also will stop executing early in some circumstances where going over the limit is analytically guaranteed.

syrusakbary:
@ethanfrey  after discussing internally we should be good to merge:

> Is the feature desired by wasmer?

Yes

> Is this approach desired by wasmer?

Yes

> If both yes, what needs to be improved to merge it?

Resolve merge conflicts

ethanfrey:
@syrusakbary Awesome!

I'm tied up this week, but will take a swing at this early next week. Thank you for the quick response.

webmaster128:
This is getting important for us I as I believe [points used can overflow](https://github.com/CosmWasm/cosmwasm/issues/464#issuecomment-655593518). I rather see this merged than trying to debug the overflow problem very large limits and very large usage values.

Since everyone seems to be happy with this patch, I can update it next week. In a new PR? Or maybe @AdamSLevy wants to earn all the glory alone and does it?

AdamSLevy:
I certainly don't need all the glory.

I have a full time job now :) that has taken me away from using Wasmer and Rust :( 

I doubt I'll have time this week to address conflicts and update this PR. Please don't let me stand in the way of progress though.

I'll once again take this opportunity to mention the metering issue #999 which still allows for certain functions to arbitrarily exceed the limit. I am happy to answer any questions about this or my other PR if needed. I can be found on the Wasmer Slack as well.

Hywan:
Please @nlewycky, can you check whether this PR is still relevant since we have reshaped the middle API?

nlewycky:
I think this is still relevant. In master the limit is specified when the Metering middleware is created and that one constant value is applied to all functions.

We have a few choices:
1. Add the ability to change the initial_limit in Metering.
2. Make the limit a lambda so that we ask the user what the limit should be for a given function.
3. Expose the underlying global variable that's holding the limit.
4. Make the underlying global variable that's holding the limit an imported global variable instead of local to the module.
5. Make the global variable non-constant and put the metering middleware injects methods into the module that we can call to get/set the limit.

ethanfrey:
Looking forward to trying out the new api.

I have a work-around (set it to a huge value and preconsume gas before starting), but it would be nice to be able to set the gas limit before each invocation of an Instance. Or at least when creating an Instance from a Module. 

We try to compile one time and reuse the pre-compiled module multiple times (large speed-up), so only being able to set it compile time is a large limitation to the API.

webmaster128:
@nlewycky are you aware of `get_remaining_points`/`set_remaining_points` added in https://github.com/wasmerio/wasmer/pull/1867/files? I think your point 3. is basically implemented, with external wrappers in `Metering`.

 https://github.com/CosmWasm/cosmwasm/pull/668 shows this in action (with a custom re-implementation of `get_remaining_points`/`set_remaining_points` in order to allow using it after ahead of time compilation).

As far as I understand, the core of this PR has been superseded in current master. The only thing that is missing from my point of view is https://github.com/wasmerio/wasmer/issues/1931.

nlewycky:
> @nlewycky are you aware of `get_remaining_points`/`set_remaining_points` added in https://github.com/wasmerio/wasmer/pull/1867/files? I think your point 3. is basically implemented, with external wrappers in `Metering`.

Oh! I did miss that that PR was finished and merged, thanks!

> As far as I understand, the core of this PR has been superseded in current master. The only thing that is missing from my point of view is #1931.

Makes sense to me.

PR #1003, opened 2019-11-22
Use target from env in build.rs

ethanfrey:
# Description

Cross compile is broken for wasmer-runtime-core. I tried compiling from linux -> macos (with lots of setup of dependencies first) and it compiled, but errored on the `build.rs`, which was using linux assembly instead of macos assembly.

Long explanation here to capture (and share) my learnings on this:

Digging into this, it turns out the reason was the use of `if cfg!(all(target_os = "linux", target_arch = "x86_64")) {` to check for the target in `build.rs`. This consistently detected the os as linux, even when I did `cargo build --target x86_64-apple-darwin`. However, the `TARGET` env variable was showing macos.

I thought this was a cargo bug, but as @nlewycky explained, build.rs is compiled for the host system (the one doing compilation), and correctly sets the target_os to linux. While the rest of the source code will have `cfg(target_os)` set to `macos`. The solution here is to rely on the `CARGO_CFG_TARGET_{OS,ARCH}` environmental variables in `build.rs` if you want to detect the system for which the library is being compiled for. (Normally they are the same, this bug only shows up during cross-compilation).

# Review

I didn't add unit tests, here but am testing it when imported in another project where I have a cross-compile build system set up. I can link the example code here if you want to run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - [#990](https://github.com/wasmerio/wasmer/pull/990) Default wasmer CLI to `run`.  Wasmer will now attempt to parse unrecognized command line options as if they were applied to the run command: `wasmer mywasm.wasm --dir=.` now works!
 - [#987](https://github.com/wasmerio/wasmer/pull/987) Fix `runtime-c-api` header files when compiled by gnuc.
 - [#957](https://github.com/wasmerio/wasmer/pull/957) Change the meaning of `wasmer_wasi::is_wasi_module` to detect any type of WASI module, add support for new wasi snapshot_preview1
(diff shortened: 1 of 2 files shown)


ethanfrey:
To demo this working, you can check out the code from https://github.com/confio/go-cosmwasm/pull/14 

and run `make docker-image` to build the cross-compile environment
and `make docker-build` to build for linux and osx.

Output should be `api/libgo_cosmwasm.{so,dylib}`. This compiling shows that the build.rs file now builds a valid output. If you have a recent version of go installed, you can try `make test` to ensure that the cross-compiled dylib actually behaves correctly. (A friend has manually verified this).

ethanfrey:
This was fixed by a different PR and included in the `0.11.0` release. Thanks for the update and now cross-compiling works much better.

PR #1004, opened 2019-11-22
Adapt backend usage depending on wasm file executed

### LINKED ISSUE #998: Adapt backend usage depending on wasm file executed

### Motivation

When we do `wasmer run xyz.wasm`, we want to execute this module automatically with the singlepass backend if the file is larger than 10Mb.

### Proposed solution

Wasmer right now have `cranelift` as the default backend in the StructOpt. We should make another option to be `autodetect` (to be the default) and choose the backend depending on what's available and the file size.

d0iasm:
Adapt backend usage depending on wasm file executed in issue #998.
Close #998 

# Description
Add `auto` backend into a runtime-core and use it as a default backend.
The `auto` backend is equivalent to: 
* singlepass if singlepass is enabled and the wasm file size is larger than 10MiB, or singlepass is enable and the target architecture is aarch64.
* cranelift otherwise.

### DIFF
--- a/src/bin/wasmer.rs
+++ b/src/bin/wasmer.rs
@@ -10,7 +10,7 @@
 extern crate structopt;
 
 use std::env;
-use std::fs::{read_to_string, File};
+use std::fs::{metadata, read_to_string, File};
 use std::io;
 use std::io::Read;
 use std::path::PathBuf;
@@ -130,7 +130,7 @@ struct Run {
     #[cfg(target_arch = "x86_64")]
     #[structopt(
         long = "backend",
-        default_value = "cranelift",
+        default_value = "auto",
         case_insensitive = true,
         possible_values = Backend::variants(),
     )]
@@ -855,8 +855,30 @@ fn interactive_shell(mut ctx: InteractiveShellContext) -> ShellExitOperation {
     }
 }
 
-fn run(options: Run) {
-    match execute_wasm(&options) {
+fn update_backend(options: &mut Run) {
+    let binary_size = match metadata(&options.path) {
+        Ok(wasm_binary) => wasm_binary.len(),
+        Err(_e) => 0,
+    };
+
+    // Update backend when a backend flag is `auto`.
+    // Use the Singlepass backend if it's enabled and the file provided is larger
+    // than 10MiB (10485760 bytes), or it's enabled and the target architecture
+    // is AArch64. Otherwise, use the Cranelift backend.
+    if options.backend == Backend::Auto {
+        if Backend::variants().contains(&Backend::Singlepass.to_string())
+            && (binary_size > 10485760 || cfg!(target_arch = "aarch64"))
+        {
+            options.backend = Backend::Singlepass;
+        } else {
+            options.backend = Backend::Cranelift;
+        }
+    }
+}
+
+fn run(options: &mut Run) {
+    update_backend(options);
+    match execute_wasm(options) {
         Ok(()) => {}
         Err(message) => {
             eprintln!("Error: {}", message);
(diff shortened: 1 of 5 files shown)


src/bin/wasmer.rs
```diff
@@ -492,6 +492,20 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
         None
     };
 
+    // Update backend when a backend flag is `auto` or a default value.
```
Hywan: > or a default value

Is it still true? By reading the code, I assume it's when the `--backend` option is set to `auto` only?

src/bin/wasmer.rs
```diff
@@ -492,6 +492,20 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
...
+    // Use a singlepass if it's enable and the file provided is larger
```
Hywan:
```diff
- a singlepass
+ the Singlepass backend
```

Hywan:
```diff
- it's enable
+ it's enabled
```

src/bin/wasmer.rs
```diff
@@ -492,6 +492,20 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
...
+    // than 10MiB (10485760 bytes), or it's enable and the target architecture
```
Hywan:
```diff
- it's enable
+ it's enabled
```

src/bin/wasmer.rs
```diff
@@ -492,6 +492,20 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
...
+    // is aarch64. Otherwise, use a cranelift as a backend.
```
Hywan:
```diff
- use a cranelift as a backend
+ use the Cranelift backend
```

src/bin/wasmer.rs
```diff
@@ -492,6 +492,20 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
...
+    if options.backend == Backend::Auto {
+        if Backend::variants().contains(&Backend::Singlepass.to_string())
+            && (wasm_binary.len() > 10485760 || cfg!(target_arch = "aarch64"))
```
Hywan: Based on my comprehension of #998, when the `--backend` option is set to `auto`, we want to use Cranelift by default, or Singlepass if the module is larger than 10MiB; which translates as:

```rust
if Backend::variants().containts(&Backend::Singlepass.to_string()) // Singlepass is available
    && wasm_binary.len() > 10485760 // Large module, be a good candidate for Singlepass
{
    options.backend = Backend::Singlepass;
} else {
    options.backend = Backend::Cranelift;
}
```

Hywan:
@d0iasm Don't forget to update the `CHANGELOG.md` file too :-).

src/bin/wasmer.rs
d0iasm: I overlooked the Backend::default(). I updated the return value from `Cranelift` to `Auto` in Backend::default(), so this comment should be true.

src/bin/wasmer.rs
d0iasm: The issue #998 only mentions the file size, but I heard from @syrusakbary on Slack that the  Singlepass backend is used when the target arch is AArch64. Is it better to add about it to the issue #998, or to remove the condition relating to AArch64 for now?

d0iasm:
@MarkMcCaskey 
Thank you for your suggestions! I chose option 1 because option 2 would affect many places since `options` is used in other functions such as `execute_wasi`. I added a new function `update_backend` called from `run`. I also reverted `options: &mut Run` to `options: &Run` in `execute_wasm` argument.

d0iasm:
The test `cache::tests::test_file_system_cache_run` fails because the default compiler is auto but there is no "real" compiler of auto. So, I reverted `Backend::default()` to return `Backend::Cranelift`.

d0iasm:
bors try

Issue #1005, opened 2019-11-22
[embedded wasmer llvm] subtraction overflow panic in State::peekn_extra


pventuzelo:
# Describe the bug

This bug happen when you are calling `wasmer::compile_with` from another rust binary compiled in debug.

``` sh
$ ./target/debug/compile_debug panic_substract_overflow.wasm
Start compile_debug
File path: "panic_substract_overflow.wasm"
thread 'main' panicked at 'attempt to subtract with overflow', XXX/wasmer/lib/llvm-backend/src/state.rs:244:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

Issue come from line 244 of lib/llvm-backend/src/state.rs:
https://github.com/wasmerio/wasmer/blob/7bb570a7f93eab4e5e0894c713d859ef4129a019/lib/llvm-backend/src/state.rs#L239-L249

This issue is not happening with `wasmer` binary since an error will be through:
``` sh
$ wasmer run --backend llvm panic_substract_overflow.wasm
Error: Can't compile module: InternalError { msg: "Codegen(\"CodegenError { message: \\\"BinaryReaderError { message: \\\\\\\"invalid value stack\\\\\\\", offset: 18446744073709551615 }\\\" }\")" }
```

# Status of my environment

```sh
echo "`wasmer -V` | `rustc -V` | `uname -m`"
wasmer 0.10.2 | rustc 1.38.0-nightly (2d1a551e1 2019-08-08) | x86_64
```
version: 7bb570a7f93eab4e5e0894c713d859ef4129a019

### Steps to reproduce

Download [panic_substract_overflow.zip](https://github.com/wasmerio/wasmer/files/3880317/panic_substract_overflow.zip)

Testing source code:
``` rust
use std::env;
use std::fs::{File};
use std::io;
use std::io::Read;
use std::path::PathBuf;


extern crate wasmer_llvm_backend;
extern crate wasmer_runtime_core;
extern crate wasmer_runtime;

use wasmer_runtime::{
    compile_with,
};
use wasmer_runtime_core::backend::Compiler;

/// Read the contents of a file
fn read_contents(path: &PathBuf) -> Result<Vec<u8>, io::Error> {
    let mut buffer: Vec<u8> = Vec::new();
    let mut file = File::open(path)?;
    file.read_to_end(&mut buffer)?;
    // We force to close the file
    drop(file);
    Ok(buffer)
}

fn get_llvm_compiler() -> impl Compiler {
[...]

PR #1006, opened 2019-11-22
fix 1005 panic sub overflow

pventuzelo:
# Description

Fix issue https://github.com/wasmerio/wasmer/issues/1005

# Review

### DIFF
--- a/lib/llvm-backend/src/state.rs
+++ b/lib/llvm-backend/src/state.rs
@@ -240,14 +240,11 @@ impl State {
         &self,
         n: usize,
     ) -> Result<&[(BasicValueEnum, ExtraInfo)], BinaryReaderError> {
-        if self.stack.len() < n {
-            return Err(BinaryReaderError {
-                message: "invalid value stack",
-                offset: -1isize as usize,
-            });
-        }
+        let new_len = self.stack.len().checked_sub(n).ok_or(BinaryReaderError {
+            message: "invalid value stack",
+            offset: -1isize as usize,
+        })?;
 
-        let new_len = self.stack.len() - n;
         Ok(&self.stack[new_len..])
     }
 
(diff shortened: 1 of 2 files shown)


lib/llvm-backend/src/state.rs
```diff
@@ -240,12 +240,15 @@ impl State {
...
+        if self.stack.len() < n {
+            return Err(BinaryReaderError {
                 message: "invalid value stack",
                 offset: -1isize as usize,
-            })
+            });
+        }
+
+        let new_len = self.stack.len() - n;
+        Ok(&self.stack[new_len..])
```
MarkMcCaskey: Minor style feedback:

⛳️ 
```Rust
let new_len = self.stack.len().checked_sub(n).ok_or( /* error code goes here */ );
// do other stuff here
```

I think this is a bit more explicit and concise!

This can be combined into a singe statement, but either style is fine.

pventuzelo: I agree with you, it would has been more clear in your way but the code will more complicated at the end because either:
* `wasmparser::primitives:BinaryReaderError` doesn't implement `convert::From`
* `warning: unreachable expression`
* unused varaible will be detected by the compiler somehow so we will need to rename `new_len` to `_new_len`.

MarkMcCaskey: Hmm, sorry I'm not following, I'll write out fully what I meant:

I think the entire function can just be:

```Rust
pub fn peekn_extra(&self, n: usize) -> Result<&[(BasicValueEnum, ExtraInfo)], BinaryReaderError> {
   let new_len = self.stack.len().checked_sub(n)
            .ok_or(BinaryReaderError {
                message: "invalid value stack",
                offset: -1isize as usize,
            });
   Ok(&self.stack[new_len..])
}
```

Let me know if that doesn't make sense

pventuzelo: I've written the same and got some error but my bad i forgot to propagate the Err ;)

PR #1018, opened 2019-11-25
feat(runtime-core) Support polymorphic host functions

Hywan:
To be defined.

### DIFF
--- a/lib/runtime-core/src/typed_func.rs
+++ b/lib/runtime-core/src/typed_func.rs
@@ -349,7 +349,7 @@ where
     pub fn new_variadic<F, Trap>(func: F, arity: i8, signature: Arc<FuncSig>) -> Self
     where
         Trap: TrapEarly<Rets>,
-        F: Fn(&[Value]) -> Trap + 'static,
+        F: Fn(&mut vm::Ctx, &[Value]) -> Trap + 'static,
     {
         match arity {
             0 => Self::new_variadic_resolved::<Zero, _, _>(func, signature),
@@ -720,11 +720,11 @@ macro_rules! impl_traits {
             }
         }
 
-        impl<Rets, Trap, FN> VariadicHostFunction<ImplicitVmCtx, $arity, Rets> for FN
+        impl<Rets, Trap, FN> VariadicHostFunction<ExplicitVmCtx, $arity, Rets> for FN
         where
             Rets: WasmTypeList,
             Trap: TrapEarly<Rets>,
-            FN: Fn(&[Value]) -> Trap + 'static,
+            FN: Fn(&mut vm::Ctx, &[Value]) -> Trap + 'static,
         {
             #[allow(non_snake_case)]
             fn to_raw(self) -> (NonNull<vm::Func>, Option<NonNull<vm::FuncEnv>>) {
@@ -743,7 +743,7 @@ macro_rules! impl_traits {
                 where
                     Rets: WasmTypeList,
                     Trap: TrapEarly<Rets>,
-                    FN: Fn(&[Value]) -> Trap,
+                    FN: Fn(&mut vm::Ctx, &[Value]) -> Trap,
                 {
                     // Get the pointer to this `wrap` function.
                     let self_pointer = wrap::<Rets, Trap, FN> as *const vm::Func;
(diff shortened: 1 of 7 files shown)


lib/runtime-core/src/typed_func.rs
```diff
@@ -232,6 +281,7 @@ pub struct Func<'a, Args = (), Rets = (), Inner: Kind = Wasm> {
     inner: Inner,
     func: NonNull<vm::Func>,
     func_env: Option<NonNull<vm::FuncEnv>>,
+    signature: Option<Arc<FuncSig>>,
```
MarkMcCaskey: Is it not possible to use the phantom data to get this information?  If so it would be good to leave a comment about why!  Seems like we were able to get by with never storing this information at runtime before

lib/runtime-core/src/typed_func.rs
```diff
@@ -271,17 +323,66 @@ where
...
+            signature: None,
+            vmctx: ptr::null_mut(),
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<'a> Func<'a, (), (), Host> {
+    /// Creates a new `Func` with a specific signature.
+    pub fn new_variadic<FN, Error>(func: FN, arity: i8, signature: Arc<FuncSig>) -> Self
```
MarkMcCaskey: Why `i8` and not `u8`? 🤔

lib/runtime-core/src/typed_func.rs
```diff
@@ -570,84 +707,143 @@ macro_rules! impl_traits {
...
+                        Ok(Ok(returns)) => return {
+                            let mut bytes = [0u8; count_idents!( $( $x ),* ) * mem::size_of::<u64>()];
+                            let mut index = 0;
+
+                            for value in returns.iter() {
+                                match value {
+                                    Value::I32(value) => {
+                                        let next_index = index + mem::size_of::<i32>();
+
+                                        NativeEndian::write_i32(&mut bytes[index..next_index], *value);
```
MarkMcCaskey: This appears to be tightly packing, so `let mut bytes = [0u8; count_idents!( $( $x ),* ) * mem::size_of::<u64>()];` will have extra space at the end if you have an 32bit values; is that correct for alignment?

One other thing that could be important when doing unsafe stuff: `[u8; N]` has alignment 1.  Therefore it's undefined behavior in general to use that array of bytes as a type of higher alignment in Rust.  If you want a higher alignment, we need to construct the array as a type of something like `[u64; N / 8]`

lib/runtime-core/src/typed_func.rs
```diff
@@ -709,7 +905,10 @@ where
             func_env @ Some(_) => Context::ExternalWithEnv(self.vmctx, func_env),
             None => Context::Internal,
         };
-        let signature = Arc::new(FuncSig::new(Args::types(), Rets::types()));
+        let signature = match self.signature {
+            Some(ref signature) => signature.clone(),
+            None => Arc::new(FuncSig::new(Args::types(), Rets::types())),
+        };
```
MarkMcCaskey: Is there a reason why we need the `Some` branch here? 🤔 it seems like we already have this information without having to store it at runtime

lib/runtime-core/src/typed_func.rs
Hywan: Good catch :-)!

lib/runtime-core/src/typed_func.rs
Hywan: >  will have extra space at the end if you have an 32bit values

It is correct yes. Since we already know the size of the values at runtime, I decided to go that way: Allocating enough space for the largest values. The resulting value needs to be allocated on the stack.

> is that correct for alignment?

I believe it is yes, since the `S0`, `S1`, `S2` & co. structs have a C representation, I think… well, no, C hasn't a packing alignement, it has a padding alignment. I should address that!

lib/runtime-core/src/typed_func.rs
Hywan: `Args` and `Rets` represent the signature that is infered at compile-time for uniadic host functions. For variadic host functions, the signature is known at runtime, so we need to it in `Func`. I will explain that with a comment.

syrusakbary:
#1217 just got merged, so I think we can close this PR for now.

:)

PR #1019, opened 2019-11-26
Update llvm backend to work with latest inkwell.

nlewycky:
cargo update to pick up latest inkwell branch commit.

Add lifetime annotations to Module which now takes a lifetime.

Add <'ctx> across CtxType and Intrinsics. Remove it from reference bindings.

Use ManuallyDrop to ensure that context's members are dropped before the Context.

Co-authored-by: Mark McCaskey <mark@wasmer.io>


lib/llvm-backend/Cargo.toml
```diff
@@ -18,8 +18,8 @@ libc = "0.2.60"
 byteorder = "1"
 
 [dependencies.inkwell]
-git = "https://github.com/wasmerio/inkwell"
-branch = "llvm8-0"
+git = "https://github.com/TheDan64/inkwell"
+rev = "781620e9fa30e51a6e03bd0d49b5f5bb7a782520"
```
syrusakbary: Can we update our local inkwell repo and use it instead?

nlewycky: Of course we can, but I thought the idea was to move to master as that will match what gets uploaded to crates.io?

syrusakbary: Yeah, that would be even better!

PR #1022, opened 2019-11-27
Add caching support for Singlepass backend.

losfair:
This PR adds caching support for the Singlepass backend.

- [x] Implementation
- [x] AArch64 test


losfair:
bors try

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -239,6 +239,25 @@ pub struct X64ExecutionContext {
     msm: ModuleStateMap,
 }
 
+/// On-disk cache format.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct CacheImage {
+    /// Code for the whole module.
+    code: Vec<u8>,
+
+    /// Offsets to the beginnings of each function. (including trampoline, if any)
```
nlewycky:
```suggestion
    /// Offsets to the beginnings of each function (including trampoline, if any).
```

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -239,6 +239,25 @@ pub struct X64ExecutionContext {
...
+    function_pointers: Vec<usize>,
+
+    /// Offsets to the beginnings of each function after trampoline.
```
nlewycky: Is this the absolute offset in the file of the function, or is this the number of bytes between the function and the trampoline? Also does this mean that the order in the on-disk file is <function1, trampoline1>, <function2, trampoline2>, ...?

losfair:
bors try-

losfair:
bors try

lib/singlepass-backend/src/codegen_x64.rs
losfair: The absolute offset.

It is <trampoline1, function1>, <trampoline2, function2>, ... , though the trampolines only exist on AArch64.

losfair:
@nlewycky Didn't notice that we now have references as absolute addresses to outside data in the generated code - this itself is not expected should be fixed.

nlewycky:
> @nlewycky Didn't notice that we now have references as absolute addresses to outside data in the generated code - this itself is not expected and should be fixed.

OK. Unfortunately x86-64 doesn't have much of a way to put a constant into an xmm register except by going through memory. Maybe emit a jmp forward, write the constant bytes into the instruction stream, the load them as constant offset before %rip?

bjfish:
This could also be updated: https://github.com/wasmerio/wasmer/blob/master/docs/feature_matrix.md

losfair:
Seems that we are already using a temporary GPR to keep the address of the constant. Since the values are at most 64 bits wide, why not do `mov imm{32,64}, gpr` and then `mov gpr, xmm` directly?

losfair:
bors try

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -3416,10 +3501,10 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
                     // load float canonical nan
                     a.emit_mov(
                         Size::S64,
-                        Location::Imm64((&CANONICAL_NAN as *const u128) as u64),
+                        Location::Imm32(CANONICAL_NAN),
```
nlewycky: Optional: feel free to inline the constants if you like.

lib/singlepass-backend/src/codegen_x64.rs
nlewycky: Thanks! Could you clarify this in the comments? Maybe at the top with:
```
/// On-disk cache format.
/// Offsets are relative to the start of the file.
/// Trampolines are only present on AArch64.
```

ethanfrey:
Awesome! Looking forward to trying this out.

PR #1028, opened 2019-12-02
feat(wasi) Introduce strict/non-strict modes for `get_wasi_version`

Hywan:
Sequel of https://github.com/wasmerio/wasmer/pull/957.

If a module has multiple import namespaces, `get_wasi_version` is
broken because it assumes a module must only have a single namespace.

This patch updates `get_wasi_version` to introduce a `strict` flag:

* In strict mode, the previous behavior applies; only one namespace is expected to be found, and this namespace must be a WASI namespace to detect the version,
* In non-strict mode, at least one WASI namespace must be declared, and the first one is used to detect the version if any.

The non-strict mode is slower because it compares namespace strings.

2 new private constants have been declared: `SNAPSHOT0_NAMESPACE` and `SNAPSHOT1_NAMESPACE` to avoid repetition and bugs.


Hywan:
bors try

MarkMcCaskey:
I think the best path forward is to expose functions for partial and complete detection, or have a boolean for strict vs not-strict mode.  We can then put the solution we come up with later underneath that API

Hywan:
I think a strict mode is a good idea. Gonna update the PR.

Hywan:
@MarkMcCaskey Is it good to merge?

PR #1029, opened 2019-12-02
feat(wasi) Add the “floating” `WasiVersion::Latest` version.

Hywan:
In addition to `Snapshot0` and `Snapshot1`, I believe it is an
interesting API to provide the `Latest` version, so that the user can
write:

```rust
generate_import_object_for_version(WasiVersion::Latest, …);
```

This is a way to ensure that modules will run only if they come with
the latest WASI version (in case of security issues for instance), by
just updating the runtime.

Note that it can be dangerous if not used carefully, but we assume the
user knows what it does by sticking on a specific “floating” version.

Also note that the `Latest` version is never returned by any API. It
is provided only by the user.


Hywan:
bors try

Hywan:
> it'd be nice if we could assert that it's never returned too, like an attribute #[user_supplied] -- I doubt it's possible or worth the effort if we have to make it now, but it'd be nice

It would be nice yeah, but I didn't find an easy to do that… Let skip that part for the moment.

PR #1034, opened 2019-12-02
When modifying the value, also update its ExtraInfo.

nlewycky:
Fixes a debug_assert! on python3.7 and rustpython in wapm.


nlewycky:
> Looks fine to me -- do think there's anywhere else this same fix might apply?

There shouldn't be. When developing this code I've been indecisive and changed whether ExtraInfo gets propagated through functions like apply_pending_canonicalization along with the value, and that led to me auditing and re-auditing how every operator handles its ExtraInfo. I think this is a special case that I missed in my last read through the code due to Operator::Select being different from any other instructions.

PR #1038, opened 2019-12-04
Added support for indexed framebuffers in framebuffer files

torch2424:
As the title states!

We spoke offline about this over slack, this simply renames the file from `wasmerfb` -> `wasmerfb0` 😄


syrusakbary:
Merging manually

PR #1040, opened 2019-12-04
Add the ability to pass backend specific options through CompilerConfig.

nlewycky:
# Description
Add the ability to pass backend specific options through CompilerConfig.

Use this to replace wasmer_llvm_backend::GLOBAL_OPTIONS.

This adds a new item to CompilerConfig which can be any struct defined by a backend for backend specific configuration options. So far, we only implement the options that were previously in GLOBAL_OPTIONS, these move to wasmer.rs. The plan is to use these same callbacks for testing in llvm-backend-tests.

# Review

### DIFF
--- a/lib/llvm-backend/src/backend.rs
+++ b/lib/llvm-backend/src/backend.rs
@@ -2,6 +2,7 @@ use super::stackmap::StackmapRegistry;
 use crate::{
     intrinsics::Intrinsics,
     structs::{Callbacks, LLVMModule, LLVMResult, MemProtect},
+    LLVMCallbacks,
 };
 use inkwell::{
     memory_buffer::MemoryBuffer,
@@ -13,8 +14,6 @@ use std::{
     any::Any,
     cell::RefCell,
     ffi::{c_void, CString},
-    fs::File,
-    io::Write,
     mem,
     ops::Deref,
     ptr::{self, NonNull},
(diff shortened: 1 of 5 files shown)


lib/llvm-backend/src/backend.rs
```diff
@@ -176,23 +175,22 @@ impl LLVMBackend {
         _stackmaps: &StackmapRegistry,
         _module_info: &ModuleInfo,
         target_machine: &TargetMachine,
+        llvm_callbacks: &Option<Rc<RefCell<dyn LLVMCallbacks>>>,
```
MarkMcCaskey: Might make sense to make this take a &mut directly, the fact that it's an `Rc<RefCell<>>` doesn't seem relevant to the function

lib/runtime-core/src/backend.rs
```diff
@@ -200,6 +200,15 @@ pub struct Features {
     pub threads: bool,
 }
 
+#[derive(Debug)]
+pub struct BackendCompilerConfig(pub Box<dyn Any + 'static>);
```
MarkMcCaskey: Some doc comments and maybe some implementation-specific comments would be good to have here -- this code is a bit abstract and can be scary

src/bin/wasmer.rs
```diff
@@ -558,25 +595,22 @@ fn execute_wasm(options: &Run) -> Result<(), String> {
...
+            backend_specific_config = Some(BackendCompilerConfig(Box::new(LLVMBackendConfig {
+                callbacks: Some(Rc::new(RefCell::new(options.backend_llvm_options.clone()))),
+            })))
         }
     }
 
+    let compiler: Box<dyn Compiler> = match get_compiler_by_backend(options.backend, options) {
+        Some(x) => x,
+        None => return Err("the requested backend is not enabled".into()),
+    };
```
MarkMcCaskey: ⛳️ 
```
let compiler: Box<dyn Compiler> =
     get_compiler_by_backend(options.backend, options).ok_or_else(|| "the requested backend is not enabled".into())?;
```

lib/runtime-core/src/backend.rs
nlewycky: I added some, let me know if you think I got it or if there's more to add / something I forgot.

lib/llvm-backend/src/backend.rs
MarkMcCaskey: You should be able to have an `Option<&mut dyn LLVMCallbacks>` I believe

MarkMcCaskey: Something like `val.map(|inner| inner.borrow_mut())` might do the trick

nlewycky: Just marking down for review that Mark and I paired on this for a while and weren't able to make it work.

MarkMcCaskey: Or that the cure seemed worse than the disease, so we chose to not bother

Issue #999, opened 2019-11-22
Metering does not catch all functions that exceed limit


AdamSLevy:
### Describe the bug

A function may exceed a limit set by the `wasmer_middleware_common::metering::Metering` by an arbitrary amount so long as it avoids the following instructions.
```
Operator::Br
Operator::BrTable
Operator::BrIf
Operator::Call
Operator::CallIndirect
```

Of course the points used may be checked after the fact, but this is not obvious to users of the library. 

Moreover, the point of metering in many cases is to *prevent* execution beyond a certain point. The current metering implementation only checks if the limit was exceeded after a code segment has completed, but not even at the end of a function.

There are two issues that may be addressed separately but are related:
- Functions that have exceeded the limit may `end` without a limit check occurring. Causing implementations to not necessarily notice that a limit was exceeded.
- Functions may exceed limits to any amount so long as they avoid the above listed instructions that cause a check. A function could be crafted to cause a DOS attack against a metered runtime.

The first issue can probably simply be resolved by adding any operators that may end a function like `Operator::End` and `Operator::Return` to the list of operators that cause a limit check.

The second issue is more complex because it requires that the cost of a code segment be statically known so that a check can be inserted ahead of the code segment, but since we are streaming the code the current implementation doesn't know that cost at that time.

I think this can be addressed by allowing the `feed_event` function to cache `Events` in the `Metering` struct until a complete code segment is parsed and the cost is known. Then the cost can be added and the limit checked before pushing the cached operations and starting to parse the next code segment. When `Event::InternalEvent::FunctionEnd` is received, the last code segment can be flushed and pushed onto the `EventSink`.

### Steps to reproduce

I believe this behavior is actually known to the devs because the `middleware-core-tests` tests actually account for it.

In this test, the limit is set at 100, and the points used is 109. In this case the function did hit an operator that caused a limit check. Nine extra operations isn't so bad, but a function could make many more without a limit check occurring

https://github.com/wasmerio/wasmer/blob/d5f25aa14e3408efd3d09e3dac8c6bd602348c1b/lib/middleware-common-tests/src/lib.rs#L124-L150

A better example is this test case. This is expected to consume 74 points, so if we set the limit at 73 we should expect it to exceed the limit check and return an error during execution, but you will see that the test still passes.

https://github.com/wasmerio/wasmer/blob/d5f25aa14e3408efd3d09e3dac8c6bd602348c1b/lib/middleware-common-tests/src/lib.rs#L101-L121

[...]

PR #1021, opened 2019-11-27
Add WIP middleware with iterators experiment

MarkMcCaskey:
WIP ; publishing so we can discuss it

PR #1041, opened 2019-12-05
Improved Metering

AdamSLevy:
# Description

This PR refactors the Metering middleware to address the issues described in #999. In short, the current metering does not prevent execution of code segments that exceed the limit, but instead only checks if limits are exceeded after execution of the code segment. Additionally the current metering does not catch some situations where the limit exceeded in the last few instructions of a function.

This metering injects limit checks prior to each code segment, ensure that execution does not proceed if there is not enough gas to reach the next branching instruction. Although after an ExecutionLimitExceededError the reported gas used will be greater than the limit, the execution will not have actually proceeded through that code segment. 

Additionally, this metering will check further ahead than eWasm metering in certain circumstances.

For example if we have multiple nested loops in a block without any branching instructions,
```
METER
Block
    ... // no branching ops
    Loop
        METER
        ... // no branching ops
        Loop
            METER
            ...
            IF // first branching op since first METER
                ...
            END
        End
    End
End
```

Each METER represents injected metering code that adds the cost of the next segment to the used points and checks it against the limit, trapping if exceeded.

Each METER that I listed will actually ensure that there is enough gas to reach the first branching op, which in my above example is IF. 

I'm happy to answer more questions about the Metering design. I believe it would benefit from additional tests that use more complex wasm code.

## Implementation details

In order to implement this metering as such, a refactor of `parse::read_module()` and `codegen::MiddlewareChain.run()` and `codegen::EventSink` was required. 

This metering algorithm requires the ability to modify previously seen Events that have already been pushed onto the EventSink inside of feed_event. This means that this middleware chain needs to see all of the Events within a function prior to any subsequent chains seeing any of the Events. Additionally this means that all ParserState(Operator) events for a given function need to all be loaded into memory at the same time. 

In parse::read_module, I clone each Operator and store them in an events Vec<Event>. I load a complete function and then I pass the whole events Ven to MiddlewareChain.run(), which passes all events through each middleware chain. 

Unfortunately wasmparser::Operator does not implement clone, so I had to fork it and add the trait. I have opened an [issue on the latest wasmparser](https://github.com/bytecodealliance/wasmparser/issues/149) but for compatability I applied my patch to the [version that is currently](https://github.com/AdamSLevy/wasmparser/tree/OperatorClonev0.39.3) being used. 

## Issues

Currently my refactor of `parse::read_module` and `codegen::MiddlewareChain.run` breaks the singlepass backend. This is not due to the metering code but due to my refactor of the runtime-core, as I have isolated that patch in testing. Alternatively you can just comment out the line that adds the metering middleware entirely from the middleware-common-tests to show that the issue lies deeper. 

When running the middleware-common-tests compilation fails with singlepass I get the following failure:
```
Errors were encountered when committing before finalization: UnknownLabel(Dynamic(DynamicLabel(3)))
```
https://gist.github.com/AdamSLevy/455d2a50ec1211d20efaebcd9c06eb1c#file-gistfile1-txt-L42-L102

I have compared my code very closely with the master branch and I am 99% certain that all mcg and fcg functions are being called in the proper order in my refactor. The only difference I can see is that the Events are now all WasmOwned. 

However my metering code works perfectly with the LLVM backend. I have been working with @MarkMcCaskey to come up with this refactor and I have run into this issue and can't work past it as I believe it would require me to learn the singlepass backend and the dynasmrt Assembler. 

# Review

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,3 +110,6 @@ crate-type = ["bin"]
 [[example]]
 name = "callback"
 crate-type = ["bin"]
+
+[patch.crates-io]
+wasmparser =  { git = "https://github.com/AdamSLevy/wasmparser", branch = "OperatorClonev0.39.3" }
--- a/lib/middleware-common/src/metering.rs
+++ b/lib/middleware-common/src/metering.rs
@@ -6,7 +6,8 @@ use wasmer_runtime_core::{
     Instance,
 };
 
-static INTERNAL_FIELD: InternalField = InternalField::allocate();
+static INTERNAL_FIELD_USED: InternalField = InternalField::allocate();
+static INTERNAL_FIELD_LIMIT: InternalField = InternalField::allocate();
 
 /// Metering is a compiler middleware that calculates the cost of WebAssembly instructions at compile
 /// time and will count the cost of executed instructions at runtime. Within the Metering functionality,
@@ -21,16 +22,12 @@ static INTERNAL_FIELD: InternalField = InternalField::allocate();
 /// the same function calls so we can say that the metering is deterministic.
 ///
 pub struct Metering {
-    limit: u64,
     current_block: u64,
 }
 
 impl Metering {
-    pub fn new(limit: u64) -> Metering {
-        Metering {
-            limit,
-            current_block: 0,
-        }
+    pub fn new() -> Metering {
+        Metering { current_block: 0 }
     }
 }
 
@@ -65,14 +62,14 @@ impl FunctionMiddleware for Metering {
                     | Operator::CallIndirect { .. }
                     | Operator::Return => {
                         sink.push(Event::Internal(InternalEvent::GetInternal(
-                            INTERNAL_FIELD.index() as _,
+                            INTERNAL_FIELD_USED.index() as _,
                         )));
                         sink.push(Event::WasmOwned(Operator::I64Const {
                             value: self.current_block as i64,
                         }));
                         sink.push(Event::WasmOwned(Operator::I64Add));
                         sink.push(Event::Internal(InternalEvent::SetInternal(
-                            INTERNAL_FIELD.index() as _,
+                            INTERNAL_FIELD_USED.index() as _,
                         )));
                         self.current_block = 0;
                     }
@@ -85,11 +82,11 @@ impl FunctionMiddleware for Metering {
                     | Operator::Call { .. }
                     | Operator::CallIndirect { .. } => {
                         sink.push(Event::Internal(InternalEvent::GetInternal(
-                            INTERNAL_FIELD.index() as _,
+                            INTERNAL_FIELD_USED.index() as _,
                         )));
-                        sink.push(Event::WasmOwned(Operator::I64Const {
-                            value: self.limit as i64,
-                        }));
+                        sink.push(Event::Internal(InternalEvent::GetInternal(
+                            INTERNAL_FIELD_LIMIT.index() as _,
+                        )));
                         sink.push(Event::WasmOwned(Operator::I64GeU));
                         sink.push(Event::WasmOwned(Operator::If {
                             ty: WpTypeOrFuncType::Type(WpType::EmptyBlockType),
@@ -111,20 +108,36 @@ impl FunctionMiddleware for Metering {
 
 /// Returns the number of points used by an Instance.
 pub fn get_points_used(instance: &Instance) -> u64 {
-    instance.get_internal(&INTERNAL_FIELD)
+    instance.get_internal(&INTERNAL_FIELD_USED)
 }
 
 /// Sets the number of points used by an Instance.
 pub fn set_points_used(instance: &mut Instance, value: u64) {
-    instance.set_internal(&INTERNAL_FIELD, value);
+    instance.set_internal(&INTERNAL_FIELD_USED, value);
 }
 
 /// Returns the number of points used in a Ctx.
 pub fn get_points_used_ctx(ctx: &Ctx) -> u64 {
-    ctx.get_internal(&INTERNAL_FIELD)
+    ctx.get_internal(&INTERNAL_FIELD_USED)
 }
 
 /// Sets the number of points used in a Ctx.
 pub fn set_points_used_ctx(ctx: &mut Ctx, value: u64) {
-    ctx.set_internal(&INTERNAL_FIELD, value);
+    ctx.set_internal(&INTERNAL_FIELD_USED, value);
+}
+
+pub fn set_execution_limit(instance: &mut Instance, limit: u64) {
+    instance.set_internal(&INTERNAL_FIELD_LIMIT, limit);
+}
+
+pub fn set_execution_limit_ctx(ctx: &mut Ctx, limit: u64) {
+    ctx.set_internal(&INTERNAL_FIELD_LIMIT, limit);
+}
+
+pub fn get_execution_limit(instance: &Instance) -> u64 {
+    instance.get_internal(&INTERNAL_FIELD_LIMIT)
+}
+
+pub fn get_execution_limit_ctx(ctx: &Ctx) -> u64 {
+    ctx.get_internal(&INTERNAL_FIELD_LIMIT)
 }
--- a/lib/middleware-common/src/metering.rs
+++ b/lib/middleware-common/src/metering.rs
@@ -1,5 +1,5 @@
 use wasmer_runtime_core::{
-    codegen::{Event, EventSink, FunctionMiddleware, InternalEvent},
+    codegen::{Event, EventSink, FunctionMiddleware, InternalEvent::*},
     module::ModuleInfo,
     vm::{Ctx, InternalField},
     wasmparser::{Operator, Type as WpType, TypeOrFuncType as WpTypeOrFuncType},
@@ -9,25 +9,109 @@ use wasmer_runtime_core::{
 static INTERNAL_FIELD_USED: InternalField = InternalField::allocate();
 static INTERNAL_FIELD_LIMIT: InternalField = InternalField::allocate();
 
-/// Metering is a compiler middleware that calculates the cost of WebAssembly instructions at compile
-/// time and will count the cost of executed instructions at runtime. Within the Metering functionality,
-/// this instruction cost is called `points`.
+/// Metering is a compiler middleware that calculates the cost of WebAssembly instructions at
+/// compile time and will count the cost of executed instructions at runtime. Within the Metering
+/// functionality, this instruction cost is called `points`.
 ///
-/// The Metering struct takes a `limit` parameter which is the maximum number of points which can be
-/// used by an instance during a function call. If this limit is exceeded, the function call will
-/// trap. Each instance has a `points_used` field which can be used to track points used during
-/// a function call and should be set back to zero after a function call.
+/// Each instance has an `exec_limit` which is the maximum number of points which can be used by
+/// the instance during a function call. If this limit is exceeded, the function call will trap.
+/// Each instance has a `points_used` field which can be used to track points used during a
+/// function call and should be set back to zero after a function call.
 ///
 /// Each compiler backend with Metering enabled should produce the same cost used at runtime for
 /// the same function calls so we can say that the metering is deterministic.
-///
 pub struct Metering {
-    current_block: u64,
+    injections: Vec<Injection>,
+    current_block_injections: Vec<Injection>,
+    current_block_cost: u64,
 }
 
 impl Metering {
     pub fn new() -> Metering {
-        Metering { current_block: 0 }
+        Metering {
+            injections: Vec::new(),
+            current_block_injections: Vec::new(),
+            current_block_cost: 0,
+        }
+    }
+
+    fn set_costs<'a, 'b: 'a>(&mut self) {
+        for inj in &mut self.current_block_injections {
+            inj.check += self.current_block_cost;
+        }
+        // Set add of the last injection
+        if self.current_block_injections.len() > 0 {
+            let last_idx = self.current_block_injections.len() - 1;
+            self.current_block_injections[last_idx] = Injection {
+                add: self.current_block_cost,
+                check: 0,
+            };
+        }
+        self.current_block_cost = 0;
+    }
+
+    fn begin<'a, 'b: 'a>(&mut self) {
+        self.set_costs();
+        self.current_block_injections
+            .push(Injection { add: 0, check: 0 });
+    }
+    fn end<'a, 'b: 'a>(&mut self) {
+        self.set_costs();
+        self.injections.append(&mut self.current_block_injections);
+    }
+
+    fn inject_metering<'a, 'b: 'a>(&self, sink: &mut EventSink<'a, 'b>) {
+        let prev: Vec<Event> = sink.buffer.drain(..).collect();
+        let mut inj_idx: usize = 1;
+        for ev in prev {
+            match ev {
+                Event::Internal(FunctionBegin(_)) => {
+                    sink.push(ev);
+                    self.injections[0].inject(sink);
+                }
+                Event::Wasm(&ref op) | Event::WasmOwned(ref op) => match *op {
+                    Operator::End
+                    | Operator::If { .. }
+                    | Operator::Else
+                    | Operator::BrIf { .. }
+                    | Operator::Loop { .. }
+                    | Operator::Call { .. }
+                    | Operator::CallIndirect { .. } => {
+                        sink.push(ev);
+                        self.injections[inj_idx].inject(sink);
+                        inj_idx += 1;
+                    }
+                    _ => {
+                        sink.push(ev);
+                    }
+                },
+                _ => {
+                    sink.push(ev);
+                }
+            }
+        }
+    }
+
+    /// increment_cost adds 1 to the current_block_cost.
+    ///
+    /// Later this may be replaced with a cost map for assigning custom unique cost values to
+    /// specific Operators.
+    fn increment_cost<'a, 'b: 'a>(&mut self, ev: &Event<'a, 'b>) {
+        match ev {
+            Event::Internal(ref iev) => match iev {
+                FunctionBegin(_) | FunctionEnd | Breakpoint(_) => {
+                    return;
+                }
+                _ => {}
+            },
+            Event::Wasm(&ref op) | Event::WasmOwned(ref op) => match *op {
+                Operator::Unreachable | Operator::End | Operator::Else => {
+                    return;
+                }
+                _ => {}
+            },
+        }
+        self.current_block_cost += 1;
     }
 }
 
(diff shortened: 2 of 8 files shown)


AdamSLevy:
@ethanfrey this might interest you if you are using LLVM

lib/runtime-core/src/codegen.rs
```diff
@@ -298,18 +298,14 @@ impl MiddlewareChain {
...
+        mut sink: EventSink<'a, 'b>,
         module_info: &ModuleInfo,
     ) -> Result<(), String> {
-        let mut sink = EventSink {
-            buffer: SmallVec::new(),
-        };
-        sink.push(ev);
         for m in &mut self.chain {
-            let prev: SmallVec<[Event; 2]> = sink.buffer.drain().collect();
+            let prev: Vec<Event> = sink.buffer.drain(..).collect();
```
losfair: Why is `SmallVec` replaced by `Vec` here? In most cases where no middleware made changes to the opcode stream, there should only be one Event here and an heap allocation would be saved by SmallVec.

lib/runtime-core/src/parse.rs
```diff
@@ -266,28 +266,22 @@ pub fn read_module<
...
+                        ParserState::EndFunctionBody => {
+                            events.push(Event::Internal(InternalEvent::FunctionEnd));
+                            break;
                         }
-                        ParserState::EndFunctionBody => break,
                         _ => unreachable!(),
                     }
                 }
+
+                let events = EventSink { buffer: events };
```
losfair: Collecting all events into one big sink here would increase memory usage, especially for large functions. Is this really needed?

AdamSLevy:
@losfair I saw some comments you made on the code show up in my inbox but I don't see them here for some reason.

Thank you for taking the time for the review. 

Regarding use of Vec instead of SmallVec, this could be in error. I simply am new to Rust, and I knew that I'd be loading more than 2 Events into the Vec so that motivated the change.

Regarding the memory burden of loading all events of a function, I agree that this is not ideal. I think this could be reduced. 

The main goal here of this refactor is to allow middleware chains to be able to view some arbitrary number of opcodes to perform some static analysis, and then inject code at some other point. In the case of my Middleware, this means injecting metering code (gas used increase & limit check) with a placeholder value for the static cost, and then later replacing the placeholder value with the actual static cost once the end of the given code segment is reached. This is done by saving the index of the `const.i64` instruction that saves the cost, and then referring back to it to update it on some later call of `feed_event()`.

My current design relies on the caller of `feed_event` (the `codegen::MiddlewareChain.run` function), to allow EventSink to serve as a place that the chain can store Events without them being flushed. For simplicity of implementation I simply loaded all Events of a function.

I believe there is a better implementation which allows the chain to control how many Operators get loaded into memory, which would be less than an entire function in most cases. Instead of loading the entire function, a middleware can have its own Vec<Events> that it can clone Events into to cache them. It can then simply not push them onto the given EventSink until enough Events have been observed for the purposes of the middleware. In the case of this metering middleware, it would only need to cache as many Events as the longest non-branching code-segment. With this, my changes to runtime-core::codegen and runtime-core::parse could be reverted to the current implementation.

I actually originally attempted something like this but I ran into lifetime issues, and this was before I patched wasmparser to add the Clone trait to Operator.

MarkMcCaskey:
@losfair I think if we have the Opcodes as Clone, we can do the iterator style I prototyped before.  That should avoid having to store them all and shouldn't be any slower in the case of no middleware; I ran into unresolvable issues due to the lifetimes too when trying this out but it should work if they're Clone

AdamSLevy:
> Instead of loading the entire function, a middleware can have its own Vec that it can clone Events into to cache them. It can then simply not push them onto the given EventSink until enough Events have been observed for the purposes of the middleware.

I attempted to implement this but I could not get around lifetime issues.

Events can't seem to be stored in a Vec<Event<'a,'b>> that is a member of the Metering struct. Also putting them back on the EventSink creates lifetime conflicts in the same way. This is a frustrating limitation of storing Events in a Vec that is a member of the Metering struct. The Metering struct demands a lifetime that is unrelated to that of the incoming Events. This is why allowing them all to be put in the EventSink that MiddlewareChain.run provides gets around lifetime constraints of the Events.

MarkMcCaskey:
@AdamSLevy if Opcode from wasmparser is Clone then we can remove one of the lifetimes on `Event`.  We can only use the owned one because it's just way simpler.  I believe that should fix things!

AdamSLevy:
I'm not sure that will resolve all of the lifetime issues, but that would simplify things.

AdamSLevy:
If Event itself could also be cloned, that would make things easy.

MarkMcCaskey:
I agree, we should make Event Clone!  I can't remember a reason besides Opcode not being Clone that it wasn't

AdamSLevy:
I tried it and ran into an issue with InternalFunction::Breakpoint.

MarkMcCaskey:
@AdamSLevy we can fix that by changing 
```Rust
pub type BreakpointHandler =
    Box<dyn Fn(BreakpointInfo) -> Result<(), Box<dyn Any>> + Send + Sync + 'static>;
```

Into 
```Rust
use std::rc::Rc;
pub type BreakpointHandler =
    Rc<dyn Fn(BreakpointInfo) -> Result<(), Box<dyn Any>> + Send + Sync + 'static>;
```

I believe.  Box is a heap allocation so copying it doesn't make much sense, Rc is like a shared heap allocation!

I'm not sure what the impacts of a change like are.  If we need mutable access to the thing then we'll need `Rc<RefCell<...>>` and if we need to share it between threads then we'll need `Arc<Mutex<...>>`

ethanfrey:
I'm not sure why it doesn't work on singlepass and I definitely need to dig in there. I think the dynasm is the same reason it doesn't serialize.

I would be happy to test this out with llvm support, but given the inability to import llvm backend from crates.io (it cannot be published), I won't likely be using llvm in an app until that is resolved. But I can try to make a test case of some malicious wasm if that would help

AdamSLevy:
@ethanfrey Any contribution to figuring out why singlepass is not working or testing the robustness of this metering algorithm would be greatly appreciated. I have wanted to add more tests that establish that metering is actually preventing execution past certain points. 

For example, in the following pseudo-code, the metering injection algorithm should prevent execution past the first METER if there is not enough gas to reach the first branching statement (an `if` in the example), which may be nested within one or more loop.  
```
const.i32 0
block
    METER // ensures that there is enough gas to reach the next branching op
    ... // non-branching ops, possibly other loops and METERs that all ensure enough gas to reach the next branching op
    loop
        METER // ensures enough gas to reach the if
        ... // non-branching ops
        if // first branch
...
```
A good test could increment a variable between these METER statements and check that it was not incremented past a certain point based on various limits.

What is this issue regarding llvm? I am planning to use my forked version of wasmer that includes exposing this backend with metering through to Golang for my application. I haven't tested this on Windows yet, which I'd like to support. Will what you mentioned cause me any issues?

AdamSLevy:
@MarkMcCaskey If you publish a branch that adds the Clone trait to Event based on the latest `master`, I can attempt to refactor this metering to store Events internally in the Metering struct. This would reduce memory usage to whatever is required to store the Events of contiguously non-branching code sequences, as opposed to an entire function and it may happen to resolve the singlepass backend issue.

losfair:
@AdamSLevy I deleted my previous comment because I didn't fully read your changes by then. Sorry if there is any misunderstanding on that.

I think if `Event` is `Clone`, we won't really need to change the current `master`'s middleware system itself. The Metering middleware can simply cache everything within a local buffer until a terminator is found.

For the issue with singlepass failing to run cached code with metering injected, I think it is because the way singlepass handles `InternalEvent::Breakpoint` depends on runtime memory. This needs a fix and before that a workaround is to ensure `singlepass & (metering ^ caching)`.

AdamSLevy:
Aside from the issues discussed above, I have discovered that my metering algorithm has a bug that causes it to add too much gas to the points used. I'm working on a refactor for that issue.

AdamSLevy:
The latest force push rebases this PR on the latest master and also fixes the bugs in my metering algorithm that I mentioned in my last comment.

Cargo.toml
```diff
@@ -110,3 +110,6 @@ crate-type = ["bin"]
 [[example]]
 name = "callback"
 crate-type = ["bin"]
+
+[patch.crates-io]
+wasmparser =  { git = "https://github.com/AdamSLevy/wasmparser", branch = "OperatorClonev0.39.3" }
```
syrusakbary: Can you create a PR into `wasmparser` so we don't need to patch it here?

https://github.com/bytecodealliance/wasmparser/compare/master...AdamSLevy:OperatorClonev0.39.3

AdamSLevy: https://github.com/bytecodealliance/wasmparser/pull/157

AdamSLevy:
The latest force push simply rebased on master and added a small fix that properly re-initializes the metering state at the start of each function.

Cargo.toml
AdamSLevy: The issue you may find with my PR is that it is based on the latest wasmparser version. This version needs to match the version given by some external dependencies as well. That locks us to 0.39.3 atm, hence my branch based on that version specifically.

AdamSLevy: This is now merged upstream.

lib/middleware-common/src/metering.rs
```diff
@@ -1,37 +1,118 @@
...
-/// a function call and should be set back to zero after a function call.
+/// Each instance has an `exec_limit` which is the maximum number of points which can be used by
+/// the instance during a function call. If this limit is exceeded, the function call will trap.
+/// Each instance has a `points_used` field which can be used to track points used during a
+/// function call and should be set back to zero after a function call.
 ///
 /// Each compiler backend with Metering enabled should produce the same cost used at runtime for
 /// the same function calls so we can say that the metering is deterministic.
-///
 pub struct Metering {
```
MarkMcCaskey: Please add `#[derive(Debug)]` (lets us print it out in debug contexts), generally we want this on all of our types because it applies recursively!

lib/middleware-common/src/metering.rs
```diff
@@ -1,37 +1,118 @@
...
-    current_block: u64,
+    injections: Vec<Injection>,
+    current_block_injections: Vec<Injection>,
+    current_block_cost: u64,
 }
 
 impl Metering {
-    pub fn new(limit: u64) -> Metering {
+    pub fn new() -> Metering {
         Metering {
```
MarkMcCaskey: Small improvement, but you can use the keyword `Self` in `impl X` contexts to refer to the type being created, so this can be 

```Rust
pub fn new() -> Self {
    Self { ... }
}
```

One more improvement, is that we can add `#[derive(Default)]` to the struct (should look like `#[derive(Debug, Default)]`) and then do something like:

```Rust
pub fn new() -> Self {
   Self::default()
}
```

These are very minor points, but it makes editing the code a bit quicker by reducing redundancy; and if in the future we want a default struct but with some number of fields not default, we can use the syntax:

```Rust
Self {
   override_field1: value1,
   override_field2: value2,
   ..Self::default()
}
```

lib/middleware-common/src/metering.rs
```diff
@@ -1,37 +1,118 @@
...
-            limit,
-            current_block: 0,
+            injections: Vec::new(),
+            current_block_injections: Vec::new(),
+            current_block_cost: 0,
+        }
+    }
+
+    fn set_costs<'a, 'b: 'a>(&mut self) {
```
MarkMcCaskey: I don't believe these lifetime annotations, `<'a, 'b: 'a>` are used in this function, so we can remove them.

Because this is its own `impl` and not a trait method, we aren't required to make the type signature conform to anything extra

lib/middleware-common/src/metering.rs
```diff
@@ -1,37 +1,118 @@
...
+            inj.check += self.current_block_cost;
+        }
+        // Set add of the last injection
+        if self.current_block_injections.len() > 0 {
+            let last_idx = self.current_block_injections.len() - 1;
+            self.current_block_injections[last_idx] = Injection {
+                add: self.current_block_cost,
+                check: 0,
+            };
+        }
```
MarkMcCaskey: ⛳️ 
```Rust
if let Some(last) = self.current_block_injections.last_mut() {
    *last = Injection {
           add: self.current_black_cost,
           check: 0,
     };
}
```

lib/middleware-common/src/metering.rs
```diff
@@ -1,37 +1,118 @@
...
+        self.current_block_cost = 0;
+    }
+
+    fn begin<'a, 'b: 'a>(&mut self) {
```
MarkMcCaskey: Same here, I think we can remove these lifetime annotations

lib/middleware-common/src/metering.rs
```diff
@@ -1,37 +1,118 @@
...
+        self.set_costs();
+        self.current_block_injections
+            .push(Injection { add: 0, check: 0 });
+    }
+    fn end<'a, 'b: 'a>(&mut self) {
```
MarkMcCaskey: Same here, I think we can remove these lifetime annotations

lib/middleware-common/src/metering.rs
```diff
@@ -1,37 +1,118 @@
...
+    fn inject_metering<'a, 'b: 'a>(&self, sink: &mut EventSink<'a, 'b>) {
+        let prev: Vec<Event> = sink.buffer.drain(..).collect();
+        let mut inj_idx: usize = 1;
+        for ev in prev {
+            match ev {
+                Event::Internal(FunctionBegin(_)) => {
+                    sink.push(ev);
+                    self.injections[0].inject(sink);
+                }
+                Event::Wasm(&ref op) | Event::WasmOwned(ref op) => match *op {
```
MarkMcCaskey: ⛳️ 
I believe `&ref op` is exactly the same as `op` I think, `ref` makes it a reference and `&` pattern matches the reference away

lib/middleware-common/src/metering.rs
```diff
@@ -1,37 +1,118 @@
...
+    fn end<'a, 'b: 'a>(&mut self) {
+        self.set_costs();
+        self.injections.append(&mut self.current_block_injections);
+    }
+
+    fn inject_metering<'a, 'b: 'a>(&self, sink: &mut EventSink<'a, 'b>) {
+        let prev: Vec<Event> = sink.buffer.drain(..).collect();
+        let mut inj_idx: usize = 1;
+        for ev in prev {
+            match ev {
```
MarkMcCaskey: It looks like we could put `sink.push(ev);` here and remove it everywhere else.  Feel free to not do that if you think that there will be branches in the future where that's not the case

lib/runtime-core/src/codegen.rs
```diff
@@ -274,7 +273,8 @@ fn requires_pre_validation(backend: Backend) -> bool {
 
 /// A sink for parse events.
 pub struct EventSink<'a, 'b> {
-    buffer: SmallVec<[Event<'a, 'b>; 2]>,
+    /// document this
+    pub buffer: Vec<Event<'a, 'b>>,
```
MarkMcCaskey: We can talk about this later, but it may make sense to keep using a SmallVec, but maybe a larger SmallVec depending on how everything works; I'm mostly reviewing for Rust style right now so I didn't catch the event sink behavior

lib/runtime-core/src/parse.rs
```diff
@@ -266,28 +266,22 @@ pub fn read_module<
...
+                                events
+                                    .push(Event::Internal(InternalEvent::FunctionBegin(id as u32)));
                             }
-                            middlewares
-                                .run(Some(fcg), Event::Wasm(op), &info.read().unwrap())
-                                .map_err(|x| LoadError::Codegen(x))?;
+                            events.push(Event::WasmOwned(op.clone()));
+                        }
+                        ParserState::EndFunctionBody => {
+                            events.push(Event::Internal(InternalEvent::FunctionEnd));
```
MarkMcCaskey: I think we can keep the old style where we call middlewares at each step to prevent filling a buffer with the wasm ops.  That way middleware can do its own lookahead.  The one structural change we'll need to make is ensuring that middleware and our code can handle being given nothing or given more at once.

I'm still fond of using iterators, something like https://github.com/wasmerio/wasmer/pull/1021 , but I ran into issues with all the traits the last time I touched that and due to wasm op not being clone, it wasn't possible

MarkMcCaskey:
> @MarkMcCaskey If you publish a branch that adds the Clone trait to Event based on the latest `master`, I can attempt to refactor this metering to store Events internally in the Metering struct. This would reduce memory usage to whatever is required to store the Events of contiguously non-branching code sequences, as opposed to an entire function and it may happen to resolve the singlepass backend issue.

Sorry, I completely missed this comment!  I'll look into it

MarkMcCaskey:
bors try

MarkMcCaskey:
> > @MarkMcCaskey If you publish a branch that adds the Clone trait to Event based on the latest `master`, I can attempt to refactor this metering to store Events internally in the Metering struct. This would reduce memory usage to whatever is required to store the Events of contiguously non-branching code sequences, as opposed to an entire function and it may happen to resolve the singlepass backend issue.
> 
> Sorry, I completely missed this comment! I'll look into it

Following up on this, I made #1080 ; once that ships we can do the rest of the changes on master

lib/middleware-common/src/metering.rs
AdamSLevy: I thought so too but I rant into issues when I used any different combination here. I think the compiler requires the `&ref op` on `Event::Wasm` to match `Event::WasmOwned(ref op)` but I can't precisely articulate why.

This exact match pattern was used in the original metering code.

In this particular branch I don't believe I will ever see any `Event::Wasm` as everything is `Wasm::Owned` once it is passed to the middleware in my current refactor of `codegen`, but I didn't want to assume that would always be the case in the middleware.

lib/middleware-common/src/metering.rs
AdamSLevy: Thanks! I really appreciate the suggestions so I can improve my Rust skills.

lib/middleware-common/src/metering.rs
MarkMcCaskey: Okay, that makes sense -- I find it easy to believe that when matching on multiple things to the same body that the types need to line up exactly and somehow that this makes it work 🤷‍♂ .  Thanks for looking into it!

lib/middleware-common/src/metering.rs
AdamSLevy: In some cases the event still needs to be known after it is pushed, which would move the event and make it unusable. So I opted for being explicit and putting `sink.push(ev)` every place it is needed over trying to find a more clever way to reduce code.

lib/runtime-core/src/codegen.rs
AdamSLevy: Sure. I am ambivalent about what vector is used as I am still mostly ignorant to the subtle differences.

lib/runtime-core/src/parse.rs
AdamSLevy: >  That way middleware can do its own lookahead. 

I am still unsure how to achieve this with the current streaming design of WasmParser. 

> I'm still fond of using iterators, something like #1021 

How would this allow for a middleware to inject code? How would that affect the iterator? Perhaps this is a rudimentary Rust question, in which case just let me know and I'll do the appropriate Googling.

AdamSLevy:
> > > @MarkMcCaskey If you publish a branch that adds the Clone trait to Event based on the latest `master`, I can attempt to refactor this metering to store Events internally in the Metering struct. This would reduce memory usage to whatever is required to store the Events of contiguously non-branching code sequences, as opposed to an entire function and it may happen to resolve the singlepass backend issue.
> > 
> > 
> > Sorry, I completely missed this comment! I'll look into it

Actually we later discussed this on Discord. I added the Clone trait to Event using your above suggested modification of BreakpointHandler. However I was not able to get the Metering struct to own events with lifetimes defined by the caller of feed_event across calls due to lifetime issues.

> Following up on this, I made #1080 ; once that ships we can do the rest of the changes on master

This is good as it gets us the Clone trait on wasmparser::Operator, but I still do not know how the parse and codegen modules could be refactored to allow middleware to own/cache these Events between calls to feed_event.

Thanks for the thorough review. I'll get the above suggested changes pushed and rebased on master soon.

lib/middleware-common/src/metering.rs
MarkMcCaskey: Oh that's fair -- if you need it in the branches, then you can probably do it at the end of the loop 🤔 ; yeah, I was just thinking that if we want to push it every time, it's more "obviously correct" if the push happens unconditionally somewhere.  As it is now, the reader has to read all the branches to find out what the logic is in regards to the event and when editing has to make sure to do it in every new branch.  If it makes sense for there to be branches where it will not be pushed, then I think it's okay, but otherwise I think having it in one place is easier to understand.

lib/runtime-core/src/codegen.rs
MarkMcCaskey: In this case it's a very small thing, SmallVec is just a combination of static array and vector, so that you can avoid heap allocations in most cases while still supporting outliers that need an arbitrarily large vector.

It's especially useful if you're writing a generic data type which under normal circumstances only needs to hold a little bit of data, but it's possible that it could sometimes hold a lot or that some other code could use it in ways you didn't anticipate.  I'm not sure useful it is here 🤷‍♂

lib/runtime-core/src/parse.rs
MarkMcCaskey: The way middleware would work is by allowing any link in the chain to hold on to things.

So for example given the chain:

```
Parser -> A -> B -> C -> CodeGen
```

Any step can hold on to as many events as it likes.  So if `A` needs to have full lookahead then it would read from Parser until parser stopped returning anything and then start returning to `B`.

If `A` needs to look at functions and `B` needs to look at loops, then `A` can track the state of the streaming opcodes, when it sees a `function start` event it caches events until it sees `function end` and then dumps everything (including any modifications it made into B).  B does the same with loops and so on.

The Rust Iterator trait defines a function `next(&mut self) -> Option<Self::Item>` with the logic being that it returns `Some(T)` while it has items to return and returns `None` when there's nothing left to return and the iterator is done.

Using iterators, we'd have the `CodeGen` step get an opaque iterator and call `next` on it doing whatever it wants with it to generate code.

That iterator is step `C`, what `C` does is use its internal iterator to get events and then do logic on them, so when `C`.next() is called, `B`.next() is called which calls `A`.next() which calls `Parser`.next().  At any step in this chain, an iterator can call `next` on its internal iterator as many times as it wants allowing it to lookahead as much as it wants.  To be more specific about how it'll commonly work: once it's read all that it wants to into its internal buffer, the struct implementing it sets a flag or otherwise tracks that it's ready to return and returns the first item.  When that bit is set, it doesn't call `next()`, it just dumps its internal buffer/outputs its own edits and additions to the stream.

I like this API design a lot because it's very elegant, none of the links in the chain need to know about anything other than how to get more input.

MarkMcCaskey: And because Iterator is well supported in Rust, if you wanted to just get everything all at once, that's very easy to do, you can just do `internal_iter.collect::<Vec<_>>()` and you have a vector of everything.  In the future when generators and `yield` are in stable Rust, this type of API can be even nicer.  Instead of making an iterator explicitly, you just write a normal function but use the keyword `yield` instead of `return` for each element and magic happens that tracks the state of execution in your function for you.

lib/middleware-common/src/metering.rs
AdamSLevy: I'll take another look and see if I can't find a more elegant solution.

MarkMcCaskey: I just took another look and I see what you're saying now, if it doesn't make sense to do this, then that's fine too!

lib/runtime-core/src/parse.rs
AdamSLevy: Sounds like this is how it should work. I am not sure how to implement it but if you can then I will modify the middleware to use that approach.

webmaster128:
I guess this PR is outedated now there is a new metering middleware in master. But is the essence of this work included in the new metering schema?

syrusakbary:
Yeah, this PR is outdated now as is based on the Wasmer 0.x codebase which doesn't include the new structure brought by the refactor for 1.0.

I believe the essence is in included in the new metering schema.
Here's the new code in case is useful for review: https://github.com/wasmerio/wasmer/blob/master/lib/middlewares/src/metering.rs#L182-L216

New PRs would be more than welcome for improving the current metering middleware.
Closing this PR now :)

lib/middleware-common/src/metering.rs
```diff
@@ -1,37 +1,118 @@
...
+                Event::Internal(FunctionBegin(_)) => {
+                    sink.push(ev);
+                    self.injections[0].inject(sink);
+                }
+                Event::Wasm(&ref op) | Event::WasmOwned(ref op) => match *op {
+                    Operator::End
+                    | Operator::If { .. }
```
AdamSLevy, line 74: @syrusakbary 

It's been awhile since I had all the mental models loaded in my head, and I don't have the context on the new metering code you linked. But I believe that If needs to be included in the following list or else I think it may suffer a similar issue that this code was attempting to address, or at minimum calculate the code cost incorrectly. Of course the implications of adding that If operator would need to be considered in the context of the rest of the metering code. 

The reason I think it needs to be there is that an If technically starts a new section/branch of code. If the condition for the if is false, it will result in a jump. 

https://github.com/wasmerio/wasmer/blob/91ffef92efaa3803e541268c33a7b2af07d77eab/lib/middlewares/src/metering.rs#L182-L216

PR #1042, opened 2019-12-06
Make regression test work in release builds too.

nlewycky:
Fix this regression test to detect the bug it was looking for in release builds too.

This bug triggered an assertion failure in debug, and by examining the pre-opt IR, we can check for the bug in release mode too.

### DIFF
--- a/lib/llvm-backend-tests/tests/compile.rs
+++ b/lib/llvm-backend-tests/tests/compile.rs
@@ -33,10 +33,6 @@ impl LLVMCallbacks for RecordPreOptIR {
     fn preopt_ir_callback(&mut self, module: &InkwellModule) {
         self.preopt_ir = module.print_to_string().to_string();
     }
-
-    fn postopt_ir_callback(&mut self, _: &InkwellModule) {}
-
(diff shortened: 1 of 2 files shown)


lib/llvm-backend-tests/tests/compile.rs
```diff
@@ -1,40 +1,72 @@
...
+}
+
+impl LLVMCallbacks for RecordPreOptIR {
+    fn preopt_ir_callback(&mut self, module: &InkwellModule) {
+        self.preopt_ir = module.print_to_string().to_string();
+    }
+
+    fn postopt_ir_callback(&mut self, _: &InkwellModule) {}
+
+    fn obj_memory_buffer_callback(&mut self, _: &InkwellMemoryBuffer) {}
```
MarkMcCaskey: You can provide (empty) defaults in the trait itself.  It may also be good to use `unimplemented!()` here to catch mistakes.

nlewycky: Don't want to use `unimplemented!()` since it's totally reasonable to implement only one or two of them. Done!

PR #1043, opened 2019-12-06
[llvm-backend] remove most panic, assert, expect from llvm-backend code.

pventuzelo:
# Description

There is a bunch of calls to panicking functions/macros inside wasmer llvm-backend.
`panic!`, `assert` and `expect` will lead wasmer to panic if reached during execution.

# Review


syrusakbary:
Hi Patrick! After talking with the team it seems that we actually want to have most of this panics, since those panics are triggered in situations that should never happen, and it's a good thing for us to panic instead of making this runtime errors so in that case we realize of the error and fix the underlying issue.

However, if any of this panics causes the fuzzer to break, please let us know!

pventuzelo:
Hi @syrusakbary, Ok no problem. 

I will let you know if fuzzers triggered something.

syrusakbary:
Closing this PR for now. We will reopen if/when the fuzzer trigger the issues.

PR #1044, opened 2019-12-06
Initial commit for support of AArch64 in the llvm backend.

nlewycky:
# Description
Enables AArch64 in the llvm backend, and inkwell when the target_arch is aarch64. Adds relevant failing spectests to the excludes.txt.

# Review

### DIFF
--- a/.azure/install-llvm.yml
+++ b/.azure/install-llvm.yml
@@ -31,11 +31,14 @@ steps:
 
   - bash: |
       set -ex
+      mkdir Win64_Release
+      cd Win64_Release
       curl -OL https://github.com/wasmerio/llvm-build/releases/download/8.x/Win64_Release.zip
       7z x Win64_Release.zip
-      llvm=`pwd`/Win64_Release
+      llvm=`pwd`
       echo "##vso[task.prependpath]$llvm/bin"
       echo "##vso[task.setvariable variable=LLVM_SYS_80_PREFIX;]$llvm"
+      cd ..
     displayName: "Install LLVM (Windows)"
     condition: eq(variables['Agent.OS'], 'Windows_NT')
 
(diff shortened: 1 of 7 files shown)


nlewycky:
Given a Cargo.toml with:

```
[target.'cfg(target_arch = "x86_64")'.dependencies]
git = "https://github.com/TheDan64/inkwell"
rev = "781620e9fa30e51a6e03bd0d49b5f5bb7a782520"
default-features = false
features = ["llvm8-0", "target-x86"]

[target.'cfg(target_arch = "aarch64")'.dependencies]
git = "https://github.com/TheDan64/inkwell"
rev = "781620e9fa30e51a6e03bd0d49b5f5bb7a782520"
default-features = false
features = ["llvm8-0", "target-aarch64"]
```

when inkwell is built, both target-aarch64 and target-x86 are defined. This is due to https://github.com/rust-lang/cargo/issues/4866.

MarkMcCaskey:
@nlewycky I believe we can fix this with the `build.rs` tricks we did before as long as the `target-*` inkwell features don't pull in any dependencies in the Cargo.toml, it will work

nlewycky:
> @nlewycky I believe we can fix this with the `build.rs` tricks we did before as long as the `target-*` inkwell features don't pull in any dependencies in the Cargo.toml, it will work

I tried implementing this a couple ways, either with build.rs printing `cargo:rustc-cfg=feature="target-x86_64"` and having `target-x86_64 = ["inkwell/target-x86"]` in the Cargo.toml, or with build.rs printing `cargo:rustc-cfg=feature="inkwell/target-x86"` directly. Both times, inkwell is being built without the target being enabled, leading to an attempt to call `inkwell::Target::initialize_x86` which isn't declared in inkwell.

PR #1053, opened 2019-12-10
For RuntimeError and breakpoints, use Box<Any + Send> instead of Box<Any>.

### LINKED ISSUE #1049: Is it possible to resume after a panic?
I know the wasmer runtime will detect when a host function (i.e. something passed in using `imports!()`) panics and translate that to a `RuntimeError` (presumably around [here](https://github.com/wasmerio/wasmer/blob/71b342e5b1348b310d96266dfdff7a7bcd76c159/lib/runtime-core/src/typed_func.rs#L496-L513)). Is it possible for the top-level caller to use `std::panic::resume_unwind()` to continue panicking?

The `data` in `RuntimeError::Error` is `Box<Any>`, whereas `std::panic::resume_unwind()` expects `Box<dyn Any + Send>`.

I guess one possible solution is to match on the result of `call()` and then explicitly `panic!()` if a runtime error was detected. That would let us continue crashing, but would also mean we lose the original backtrace.

This may be related to #986 or #827.

nlewycky:
# Description
For RuntimeError and breakpoints, use Box<Any + Send> instead of Box\<Any\>.

# Review

Fixes #1049

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## **[Unreleased]**
 
+- [#1053](https://github.com/wasmerio/wasmer/pull/1053) For RuntimeError and breakpoints, use Box<Any + Send> instead of Box<Any>.
 - [#1050](https://github.com/wasmerio/wasmer/pull/1050) Attach C & C++ headers to releases.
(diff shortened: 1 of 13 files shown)


nlewycky:
bors try

nlewycky:
bors try

PR #1058, opened 2019-12-11
Fix issue 1057 + improve llvm/state.rs code

pventuzelo:
# Description

This pull request is doing:
- Fix issue https://github.com/wasmerio/wasmer/issues/1057 (subtraction overflow panic in State::peek1_extra) by using `checked_sub`
- replace other part of `state.rs` subject to potential substration overflow with `checked_sub`
- add more detail on the Errors messages in `state.rs`
- rename some variable for better understanding of the code.

# Review

### DIFF
--- a/lib/llvm-backend/src/state.rs
+++ b/lib/llvm-backend/src/state.rs
@@ -231,33 +231,41 @@ impl<'ctx> State<'ctx> {
 
     pub fn outermost_frame(&self) -> Result<&ControlFrame<'ctx>, BinaryReaderError> {
         self.control_stack.get(0).ok_or(BinaryReaderError {
-            message: "invalid control stack depth",
+            message: "outermost_frame: invalid control stack depth",
             offset: -1isize as usize,
         })
     }
 
     pub fn frame_at_depth(&self, depth: u32) -> Result<&ControlFrame<'ctx>, BinaryReaderError> {
-        let index = self.control_stack.len() - 1 - (depth as usize);
-        self.control_stack.get(index).ok_or(BinaryReaderError {
-            message: "invalid control stack depth",
-            offset: -1isize as usize,
-        })
+        let index = self
+            .control_stack
+            .len()
+            .checked_sub(1 + (depth as usize))
+            .ok_or(BinaryReaderError {
+                message: "frame_at_depth: invalid control stack depth",
+                offset: -1isize as usize,
+            })?;
+        Ok(&self.control_stack[index])
     }
 
     pub fn frame_at_depth_mut(
         &mut self,
         depth: u32,
     ) -> Result<&mut ControlFrame<'ctx>, BinaryReaderError> {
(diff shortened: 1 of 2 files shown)


lib/llvm-backend/src/state.rs
```diff
@@ -231,33 +231,41 @@ impl<'ctx> State<'ctx> {
...
     ) -> Result<&mut ControlFrame<'ctx>, BinaryReaderError> {
-        let index = self.control_stack.len() - 1 - (depth as usize);
-        self.control_stack.get_mut(index).ok_or(BinaryReaderError {
-            message: "invalid control stack depth",
-            offset: -1isize as usize,
-        })
+        let index = self
+            .control_stack
+            .len()
+            .checked_sub(1 + (depth as usize))
```
nlewycky: Here too.

lib/llvm-backend/src/state.rs
```diff
@@ -231,33 +231,41 @@ impl<'ctx> State<'ctx> {
...
     pub fn frame_at_depth(&self, depth: u32) -> Result<&ControlFrame<'ctx>, BinaryReaderError> {
-        let index = self.control_stack.len() - 1 - (depth as usize);
-        self.control_stack.get(index).ok_or(BinaryReaderError {
-            message: "invalid control stack depth",
-            offset: -1isize as usize,
-        })
+        let index = self
+            .control_stack
+            .len()
+            .checked_sub(1 + (depth as usize))
```
nlewycky:
```
.checked_sub(1)
.checked_sub(depth as usize)
```
?

pventuzelo: Not possible because `checked_sub` return an `Option`.

Do you prefer this way?
``` rust
        let tmp_index = self
            .control_stack
            .len()
            .checked_sub(1).ok_or(BinaryReaderError {
                message: "frame_at_depth: invalid control stack depth",
                offset: -1isize as usize,
            })?;
        let index = tmp_index.checked_sub(depth as usize)
            .ok_or(BinaryReaderError {
                message: "frame_at_depth: invalid control stack depth",
                offset: -1isize as usize,
            })?;
        Ok(&self.control_stack[index])
```

nlewycky: Doh! No, I prefer the code as it is, thanks.

MarkMcCaskey: It should be possible to use `.and_then` here.  I noticed the potential for overflow there too, but I figured it's very unlikely.  I think it should be fine as-is!

For future readers who may end up here who didn't understand what I meant:

```Rust
self
            .control_stack
            .len()
            .checked_sub(1)
            .and_then(|n| n.checked_sub(depth as usize))
            .ok_or(BinaryReaderError {
                message: "frame_at_depth: invalid control stack depth",
                offset: -1isize as usize,
            })?;
```

`and_then` acts like Monadic bind (`>>=`) in Haskell.

PR #1059, opened 2019-12-12
Add doc comments for WASI FD associated constants

MarkMcCaskey:
I want to try out something I'm going to call "daily docs" where I write documentation for one thing every day

### DIFF
--- a/lib/wasi/src/state/mod.rs
+++ b/lib/wasi/src/state/mod.rs
@@ -122,16 +122,30 @@ pub struct Fd {
     pub rights_inheriting: __wasi_rights_t,
     pub flags: __wasi_fdflags_t,
     pub offset: u64,
-    /// Used when reopening the file on the host system
+    /// Flags that determine how the [`Fd`] can be used.
+    ///
+    /// Used when reopening a [`HostFile`] during [`WasiState`] deserialization.
     pub open_flags: u16,
     pub inode: Inode,
 }
 
 impl Fd {
+    /// This [`Fd`] can be used with read system calls.
     pub const READ: u16 = 1;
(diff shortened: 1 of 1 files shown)


lib/wasi/src/state/mod.rs
```diff
@@ -128,10 +128,18 @@ pub struct Fd {
 }
 
 impl Fd {
+    /// This FD can be used with read system calls.
```
nlewycky: FD is unexplained, you could either write "file descriptor" here, or "descriptor" or document "pub struct Fd" above?

lib/wasi/src/state/mod.rs
```diff
@@ -128,10 +128,18 @@ pub struct Fd {
...
     pub const READ: u16 = 1;
+    /// This FD can be used with write system calls.
     pub const WRITE: u16 = 2;
+    /// This FD can append in write system calls. Note that append
```
nlewycky: "Note that append permissions require the write permission."

What happens if I open for read and append but not write? Does the open fail? Does it automatically set write for me?

lib/wasi/src/state/mod.rs
```diff
@@ -128,10 +128,18 @@ pub struct Fd {
 }
 
 impl Fd {
```
nlewycky: Are there any invariants across the whole group? Like, all file descriptors must be open for at least reading or writing?

lib/wasi/src/state/mod.rs
MarkMcCaskey: Yeah that should be ```[`Fd`]```, good call

PR #1062, opened 2019-12-13
Add emscripten functions and types to the C API

### LINKED ISSUE #574: Export Emscripten imports so can be used from C-API

### Motivation

Command `wasmer run` can load `wasm` compiled by Emscripten but `c-api` interface can't.  
`Wasmer` has already supported the Emscripten/WASI.  

### Proposed solution
1. Support importing both custom functions and Emscripten/WASI functions.
2. Keep features like`callinfo`, `import custom function`remain by c-api.

MarkMcCaskey:
resolves #574 

Left to do:
- [x] Add tests
- [x] Add memory set up, etc (need to break up monolithic ("run" function in wasmer_emscripten into pieces)
- [x] Add passing args, etc

# Review


MeirShpilraien:
@MarkMcCaskey thank you very much on the PR.

Just to make sure how to use it, After compiling my module I will have to call `wasmer_emscripten_get_emscripten_globals` then I will create `wasmer_import_object_t` from those globals using `wasmer_emscripten_generate_import_object` and then I will extend the created `wasmer_import_object_t` with my own modules and call `wasmer_module_import_instantiate`

Is that right?

MeirShpilraien:
@MarkMcCaskey I have tried this branch and now I am not getting the `Import not found` error and I can create an instance, but now it failed on wasmer_instance_call with this error `unknown trap at 0xb08 - segmentation violation`. As said before, when running with wasmer binary it runs just fine. Any idea what it might be?

MeirShpilraien:
Got it, there was an uninitialized variable that cause the segfault. I can confirm that its working.

MarkMcCaskey:
bors try

PR #1063, opened 2019-12-14
Add more doc comments to our WASI implementation

MarkMcCaskey:


lib/wasi/src/state/mod.rs
```diff
@@ -376,6 +376,7 @@ impl WasiFs {
         }
     }
 
+    /// Returns the next available inode index for creating a new inode.
```
nlewycky: As written, someone might think this can return the same value twice. This returns a new inode number, or allocates an inode, or something like that?

lib/wasi/src/state/builder.rs
```diff
@@ -1,4 +1,4 @@
-//! Builder code for [`WasiState`]
+//! Builder system for flexibly building [`WasiState`].
```
nlewycky: Optional: Expounds upon what "flexibly" means here?

lib/wasi/src/state/mod.rs
```diff
@@ -672,6 +682,14 @@ impl WasiFs {
         Ok(cur_inode)
     }
 
+    /// Splits a path into the first preopened directory that is a parent of it
```
nlewycky: Optional: is there guaranteed to be one? (Is it root? Maybe that doesn't belong documented here but somewhere else.)

lib/wasi/src/state/mod.rs
```diff
@@ -491,6 +492,15 @@ impl WasiFs {
         }
     }
 
+    /// Internal part of the core path resolution function which implements path
+    /// traversal logic such as going resolving relative path segments (such as
+    /// `.` and `..`) and resolving symlinks (while preventing infinite
+    /// loops/stack overflows).
+    ///
```
nlewycky: I think something this complicated should probably have a "given X produces Y" description, like, given a path and an open file descriptor of the base directory, returns the inode of the file that it refers to, or an error if it doesn't exist or an error occurred (such as infinite recursion resolving symlinks).

lib/wasi/src/state/mod.rs
MarkMcCaskey: I thought that too, but it takes a `&mut self` which I thought makes it sufficiently clear that it's not a pure function

PR #1068, opened 2019-12-16
Various cleanups for the singlepass backend on AArch64.

losfair:
ref: https://github.com/wasmerio/wasmer/pull/713


losfair:
bors try

PR #1071, opened 2019-12-16
Add support for non-trapping float-to-int conversions.

nlewycky:
# Description
Rewrites LLVM implementation of non-trapping float-to-int conversions. LLVM's conversion operations produce unspecified output when the input floats are out of range for the conversion.

Add implementation to singlepass for both x86-64 and AArch64.

# Review


syrusakbary:
bors try

lib/llvm-backend/src/code.rs
```diff
@@ -4472,10 +4564,36 @@ impl<'ctx> FunctionCodeGenerator<CodegenError> for LLVMFunctionCodeGenerator<'ct
...
+                let (v, i) = state.pop1_extra()?;
+                let v = apply_pending_canonicalization(builder, intrinsics, v, i);
+                let v = v.into_float_value();
+                let res = trunc_sat_scalar(
+                    builder,
+                    intrinsics.i32_ty,
+                    std::i32::MIN as u64,
+                    2147483520u64, // bits as f32: 0x4effffff
+                    std::i32::MIN as u32 as u64,
+                    std::i32::MAX as u32 as u64,
```
MarkMcCaskey: ⛳️ `i32::max_value()` doesn't need `std::` 🤔 either way is fine though

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -4626,6 +4835,63 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                    Size::S32,
+                    loc,
+                    Location::XMM(tmp_in),
+                );
+                Self::emit_f32_int_conv_check_sat(
+                    a,
+                    &mut self.machine,
+                    tmp_in,
+                    -2147483904.0,
+                    2147483648.0,
```
MarkMcCaskey: Names and/or explanations for numbers like these are always appreciated

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -4686,6 +4952,64 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                    Size::S32,
+                    loc,
+                    Location::XMM(tmp_in),
+                );
+                Self::emit_f32_int_conv_check_sat(
+                    a,
+                    &mut self.machine,
+                    tmp_in,
+                    -9223373136366403584.0,
+                    9223372036854775808.0,
```
MarkMcCaskey: Same here

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -4769,6 +5093,82 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                    Size::S32,
+                    loc,
+                    Location::XMM(tmp_in),
+                );
+                Self::emit_f32_int_conv_check_sat(
+                    a,
+                    &mut self.machine,
+                    tmp_in,
+                    -1.0,
+                    18446744073709551616.0,
```
MarkMcCaskey: here too, if these numbers are shared between the backends it might even make sense to put them in a shared location

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -4769,6 +5093,82 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                                Location::Imm32(1593835520u32),
+                                Location::GPR(tmp),
+                            ); //float 9.22337203E+18
+                            a.emit_mov(Size::S32, Location::GPR(tmp), Location::XMM(tmp_x1));
+                            a.emit_mov(Size::S32, Location::XMM(tmp_in), Location::XMM(tmp_x2));
+                            a.emit_vsubss(tmp_in, XMMOrMemory::XMM(tmp_x1), tmp_in);
+                            a.emit_cvttss2si_64(XMMOrMemory::XMM(tmp_in), tmp_out);
+                            a.emit_mov(
+                                Size::S64,
+                                Location::Imm64(0x8000000000000000u64),
```
MarkMcCaskey: Could this be `i64::max_value() as u64` or `1 << 63`?

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -4814,7 +5214,13 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
                         loc,
                         Location::XMM(tmp_in),
                     );
-                    Self::emit_f64_int_conv_check(a, &mut self.machine, tmp_in, -1.0, 4294967296.0);
+                    Self::emit_f64_int_conv_check_trap(
+                        a,
+                        &mut self.machine,
+                        tmp_in,
+                        -1.0,
+                        4294967296.0,
```
MarkMcCaskey: This number is appearing often

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -4824,6 +5230,58 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                    Size::S64,
+                    loc,
+                    Location::XMM(tmp_in),
+                );
+                Self::emit_f64_int_conv_check_sat(
+                    a,
+                    &mut self.machine,
+                    tmp_in,
+                    -1.0,
+                    4294967296.0,
```
MarkMcCaskey: perhaps a named constant is in order

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -4889,6 +5347,69 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                        tmp_in
+                    }
+                };
+
+                Self::emit_f64_int_conv_check_sat(
+                    a,
+                    &mut self.machine,
+                    real_in,
+                    -2147483649.0,
+                    2147483648.0,
```
MarkMcCaskey: these too

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -4949,6 +5470,64 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                    Size::S64,
+                    loc,
+                    Location::XMM(tmp_in),
+                );
+                Self::emit_f64_int_conv_check_sat(
+                    a,
+                    &mut self.machine,
+                    tmp_in,
+                    -9223372036854777856.0,
+                    9223372036854775808.0,
```
MarkMcCaskey: The bottom number is the same but the top number is slightly different:

from
```
                    -9223373136366403584.0,
                    9223372036854775808.0,
```

names would make that more obvious

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -4769,6 +5093,82 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                            Location::Imm64(std::u64::MAX),
+                            Location::GPR(tmp_out),
+                        );
+                    },
+                    None::<fn(_a: &mut Assembler, _m: &mut Machine)>,
+                    |a, m| {
+                        if a.arch_has_itruncf() {
+                            a.arch_emit_i64_trunc_uf32(tmp_in, tmp_out);
+                        } else {
+                            let tmp = m.acquire_temp_gpr().unwrap(); // r15
```
losfair: Seems that these specific register indexes (r15, xmm1, xmm3) are not depended on from anywhere else?

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -5033,6 +5612,83 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                    get_location_released(a, &mut self.machine, self.value_stack.pop().unwrap());
+                let ret = self.machine.acquire_locations(
+                    a,
+                    &[(WpType::I64, MachineValue::WasmStack(self.value_stack.len()))],
+                    false,
+                )[0];
+                self.value_stack.push(ret);
+
+                let tmp_out = self.machine.acquire_temp_gpr().unwrap();
+                let tmp_in = self.machine.acquire_temp_xmm().unwrap(); // xmm2
```
losfair: Same as above

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -5033,6 +5612,83 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
+                            Location::Imm64(std::u64::MAX),
+                            Location::GPR(tmp_out),
+                        );
+                    },
+                    None::<fn(_a: &mut Assembler, _m: &mut Machine)>,
+                    |a, m| {
+                        if a.arch_has_itruncf() {
+                            a.arch_emit_i64_trunc_uf64(tmp_in, tmp_out);
+                        } else {
+                            let tmp = m.acquire_temp_gpr().unwrap(); // r15
```
losfair: Here too

lib/llvm-backend/src/code.rs
```diff
@@ -227,6 +232,93 @@ fn trunc_sat<'ctx>(
...
 
+// Convert floating point vector to integer and saturate when out of range.
+// https://github.com/WebAssembly/nontrapping-float-to-int-conversions/blob/master/proposals/nontrapping-float-to-int-conversion/Overview.md
+fn trunc_sat_scalar<'ctx>(
+    builder: &Builder<'ctx>,
+    int_ty: IntType<'ctx>,
+    lower_bound: u64, // Exclusive (lowest representable value)
+    upper_bound: u64, // Exclusive (greatest representable value)
+    int_min_value: u64,
+    int_max_value: u64,
```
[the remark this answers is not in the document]
MarkMcCaskey: After reviewing the PR I think it's not as bad as I first thought.  I think it's probably possible to clean it up a bit though.  Low priority all things considered though

lib/singlepass-backend/src/codegen_x64.rs
nlewycky: These aren't quite shared between backends, though they absolutely could be. Note that these are f32/f64 while the llvm-backend uses u64, also llvm-backend uses values that are inside the valid range while singlepass uses values outside the valid range. I don't think there's any underlying need for these differences, but it might be that these are expressed in the way most convenient for lowering code in each backend. I'm going to punt that investigation and possible cleanup to a future PR.

lib/singlepass-backend/src/codegen_x64.rs
nlewycky: Fixed.

lib/singlepass-backend/src/codegen_x64.rs
nlewycky: Yup.

lib/singlepass-backend/src/codegen_x64.rs
nlewycky: Mhmm.

lib/singlepass-backend/src/codegen_x64.rs
nlewycky: `GEF64_LT_I64_MIN` and `LEF64_GT_I64_MAX`.

lib/singlepass-backend/src/codegen_x64.rs
nlewycky: Done.

lib/singlepass-backend/src/codegen_x64.rs
nlewycky: Done.

lib/singlepass-backend/src/codegen_x64.rs
nlewycky: Added a section of constants to the bottom of the file. (Same in the LLVM backend too.) Done.

lib/singlepass-backend/src/codegen_x64.rs
nlewycky: `GEF32_LT_I64_MIN` and `LEF32_GT_I64_MAX`.

lib/llvm-backend/src/code.rs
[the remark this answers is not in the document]
MarkMcCaskey: Just being more semantic where possible.  It's a minor thing but a number is a very open ended thing.  If it was a bunch of 0-sized structs with a method that returned an upper and lower bound, then mix ups would be less likely.

I don't think it's worth the effort here though -- I thought this was more complex than it actually is

lib/llvm-backend/src/code.rs
MarkMcCaskey: Yeah, I'm fine with either -- the only reason I mention it is that I haven't seen the MIN and MAX constants in a while.  Now that I see them again I do remember having seen them in the past.

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -8986,3 +9642,42 @@ fn sort_call_movs(movs: &mut [(Location, GPR)]) {
...
+/// Least Exact Float (64 bits) greater-than i64::MAX when rounding towards zero.
+const LEF64_GT_I64_MAX: f64 = 9223372036854775808.0;
+/// Greatest Exact Float (64 bits) less-than u32::MIN when rounding towards zero.
+const GEF64_LT_U32_MIN: f64 = -1.0;
+/// Least Exact Float (64 bits) greater-than u32::MAX when rounding towards zero.
+const LEF64_GT_U32_MAX: f64 = 4294967296.0;
+/// Greatest Exact Float (64 bits) less-than u64::MIN when rounding towards zero.
+const GEF64_LT_U64_MIN: f64 = -1.0;
+/// Least Exact Float (64 bits) greater-than u64::MAX when rounding towards zero.
+const LEF64_GT_U64_MAX: f64 = 18446744073709551616.0;
```
MarkMcCaskey: I'd prefer if these weren't duplicated; if they don't seem to fit into `runtime-core`, we could add `backend-common` crate for this kind of thing.  I'd like to encourage code reuse among the backends in general

PR #1072, opened 2019-12-16
Add minor changes to runtime/vm doc comments + make fns const fns

MarkMcCaskey:
Slightly improves readability of docs and makes more functions `const fn`


Hywan:
:+1:

PR #1073, opened 2019-12-17
Add llvm tests to AArch64 CI.

nlewycky:


syrusakbary:
This requires LLVM to be installed in the ARM machine. We have tested on our ARM machine, and everything works well.

However, for CI we rely on the prebuilt option of LLVM (https://github.com/wasmerio/llvm-build) and since this prebuilt is generated with Azure Pipelines (which doesn't have ARM option) we don't have yet an ARM prebuilt binary, and therefore we can't install easily LLVM on the travis ARM machine.

Closing the PR for now.

MarkMcCaskey:
@syrusakbary if this is something we really want, we can probably just have CI use a cached compiled LLVM from S3 or something.  We'd just have to download and unzip it and then create a script for building and uploading it that we can manually run for now (because we don't update LLVM right now anyways).

Doesn't seem super urgent to me, but I think we can quickly get a solution working when we want to!

PR #1075, opened 2019-12-17
Add include guards for WASI APIs in runtime-c-api

MarkMcCaskey:
Improves correctness, but the readability does suffer.

In the future we should probably have separate header files or find a way to merge the conditional includes


Hywan:
:+1:

PR #1078, opened 2019-12-18
Add more parameters to Func

Jasper-Bekkers:
# Description
We ran into the issue that we needed to pass more then the available number of parameters to wasm, so I've extended the parameter count for `Func` to have max 26 parameters now.

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## **[Unreleased]**
 
+- [#1078](https://github.com/wasmerio/wasmer/pull/1078) Add more parameters to Func
 - [#1062](https://github.com/wasmerio/wasmer/pull/1062) Expose some opt-in Emscripten functions to the C API
 - [#1032](https://github.com/wasmerio/wasmer/pull/1032) Change the signature of the Emscripten `abort` function to work with Emscripten 1.38.30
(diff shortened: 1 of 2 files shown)


Hywan:
I was wondering when someone would need more than 12 parameters. Here we are ;-).

Jasper-Bekkers:
> I was wondering when someone would need more than 12 parameters. Here we are ;-).

I'm not exactly happy about having to need them, but that's life.

The linter seems to complain about formatting, I'd personally prefer the single-line macro invocations but I can understand the project not wanting exceptions for it so I'll leave that up to you guys to decide :-)

Hywan:
The linter is rustfmt. Please follow its advice. We don't have any particular opinions on this topic, we let the linter decides for us ;-).

Hywan:
Or, add `#[rustfmt::skip]`, that's fine too! It's up to you :-).

Jasper-Bekkers:
I'll add the `#[rustfmt::skip]` I think it's visually a bit nicer and more inline with the formatting for the rest of the file.

Jasper-Bekkers:
It looks like TravisCI is failing because of some network issue

Hywan:
Yup, I agree. Let's see in a moment.

Hywan:
Thanks!

PR #1092, opened 2019-12-20
Add function to get nul-terminated strings from memory

### LINKED ISSUE #1086: Add method for reading C-style strings from WasmPtr
Suggested [on spectrum](https://spectrum.chat/wasmer/general/read-string-from-wasm-emscriptens-internal-utils~923a9bb9-4b33-4a23-b9ae-daa5b038f312) .

This is a very good idea, we should support getting a slice from a null-terminated array and probably have a function that just returns a `&str` too.  Edge cases to watch out for are null bytes in UTF-8 may be valid if in code point.  If that is the case, then we'll want to have a helper method specifically for reading UTF-8 from a WasmPtr with a null terminator.  ASCII can probably go through CStr

srenatus:
## Description

This is meant to fix #1086.

❓ I'm a bit new to this -- is this how you'd do it? Happy to take directions! 😃 

I've added a note regarding the special case in the comment.

Are there existing tests to expand?

## Review

### DIFF
--- a/lib/runtime-core/src/memory/ptr.rs
+++ b/lib/runtime-core/src/memory/ptr.rs
@@ -137,6 +137,17 @@ impl<T: Copy + ValueType> WasmPtr<T, Array> {
         let slice: &[u8] = unsafe { std::slice::from_raw_parts(ptr, str_len as usize) };
         std::str::from_utf8(slice).ok()
     }
+
+    /// Get a UTF-8 string representation of this `WasmPtr`, where the string is nul-terminated.
+    /// Note that this does not account for UTF-8 strings that _contain_ nul themselves,
+    /// [`get_utf8_string`] has to be used for those.
+    pub fn get_utf8_string_with_nul<'a>(self, memory: &'a Memory) -> Option<&'a str> {
+        memory.view::<u8>()[(self.offset as usize)..]
+            .iter()
+            .map(|cell| cell.get())
+            .position(|byte| byte == 0)
+            .and_then(|length| self.get_utf8_string(memory, length as u32))
+    }
 }
 
 unsafe impl<T: Copy, Ty> WasmExternType for WasmPtr<T, Ty> {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## **[Unreleased]**
 
+- [#1092](https://github.com/wasmerio/wasmer/pull/1092) Add `get_utf8_string_with_nul` to `WasmPtr` to read nul-terminated strings from memory.
 
 ## 0.12.0 - 2019-12-18
 


srenatus:
🤔 Not sure what happens on travis: https://travis-ci.org/wasmerio/wasmer/builds/627783425#L333 -- when I flip this locally, I get the opposite error, expected `i8`, found `u8`.

MarkMcCaskey:
Oh, one note about our linting/formatting: we stay 1 stable release behind the latest stable, so you'll have to format it with Rust 1.39 (`rustup install 1.39.0` to install it and then run `rustup override set 1.39.0` in the `wasmer` directory so the tools know to always use that version in this folder)

srenatus:
@MarkMcCaskey thanks for the insightful response! I'll try to get to it shortly. 😃

srenatus:
@MarkMcCaskey Can you have another look? I've landed on a slight alteration of your proposal, and renamed the null to nul. Thanks again for taking the time to lay out how `unsafe` is to be handled in Rust, I've learnt something today! 💡 

Let's see what travis thinks of this. (The error befor wasn't linting related, was it? But it seems like I'm unable to run `make spectests-singlepass` locally 🤔)

MarkMcCaskey:
@srenatus Singlepass requires nightly!  everything else works with stable 1.39!  Sounds good, I'll review it!

lib/runtime-core/src/memory/ptr.rs
```diff
@@ -137,6 +137,17 @@ impl<T: Copy + ValueType> WasmPtr<T, Array> {
         let slice: &[u8] = unsafe { std::slice::from_raw_parts(ptr, str_len as usize) };
         std::str::from_utf8(slice).ok()
     }
+
+    /// Get a UTF-8 string representation of this `WasmPtr`, where the string is nul-terminated.
+    /// Note that this does not account for UTF-8 strings that _contain_ nul themselves,
+    /// [`get_utf8_string`] has to be used for those.
+    pub fn get_utf8_string_nul_terminated<'a>(self, memory: &'a Memory) -> Option<&'a str> {
```
MarkMcCaskey: Can we change the name of this to `get_utf8_string_with_nul` to follow the naming conventions of the methods on CStr?

lib/runtime-core/src/memory/ptr.rs
```diff
@@ -137,6 +137,17 @@ impl<T: Copy + ValueType> WasmPtr<T, Array> {
...
+        memory.view::<u8>()[(self.offset as usize)..]
+            .iter()
+            .map(|cell| cell.get())
+            .position(|byte| byte == 0)
+            .and_then(|length| self.get_utf8_string(memory, length.try_into().unwrap()))
```
MarkMcCaskey: ⛳️ 
`length.try_into().unwrap()` -> `length as u32`

`try_into` is better for being super generic but Wasmer is written in a way that there's no chance that it will work on systems with pointers smaller than 32 bits, so it's safe to just do the cast directly!  If we want to run on 16bit machines in the future, we'll have to do a ton of refactoring anyways!

srenatus:
Done! 😃

PR #1099, opened 2019-12-21
Remove backend::Backend from runtime-core

syrusakbary:
# Description

This PR removes the dependency of a Backend in runtime-core. So it's agnostic and more backends can be plugged in easily.

Why this is important?
* By removing backends from wasmer-runtime-core we can make the runtime agnostic, so anyone can plug their own backend into Wasmer without needing to touch the main source code (V8, JavascriptCore, wasm3, ...).
* It simplifies the codebase and avoids code leaks from the backend to the runtime API.

# Review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,38 +222,9 @@ check: check-bench
 	# as default, and test a minimal set of features with only one backend
 	# at a time.
 	cargo check --manifest-path lib/runtime/Cargo.toml
-	# Check some of the cases where deterministic execution could matter
-	cargo check --manifest-path lib/runtime/Cargo.toml --features "deterministic-execution"
-	cargo check --manifest-path lib/runtime/Cargo.toml --no-default-features \
-		--features=default-backend-singlepass,deterministic-execution
-	cargo check --manifest-path lib/runtime/Cargo.toml --no-default-features \
-		--features=default-backend-llvm,deterministic-execution
-	cargo check --release --manifest-path lib/runtime/Cargo.toml
 
 	$(RUNTIME_CHECK) \
-		--features=cranelift,cache,debug,llvm,singlepass,default-backend-singlepass
-	$(RUNTIME_CHECK) --release \
-		--features=cranelift,cache,llvm,singlepass,default-backend-singlepass
-	$(RUNTIME_CHECK) \
-		--features=cranelift,cache,debug,llvm,singlepass,default-backend-cranelift
-	$(RUNTIME_CHECK) --release \
-		--features=cranelift,cache,llvm,singlepass,default-backend-cranelift
-	$(RUNTIME_CHECK) \
-		--features=cranelift,cache,debug,llvm,singlepass,default-backend-llvm
-	$(RUNTIME_CHECK) --release \
-		--features=cranelift,cache,llvm,singlepass,default-backend-llvm
-	$(RUNTIME_CHECK) \
-		--features=singlepass,default-backend-singlepass,debug
-	$(RUNTIME_CHECK) --release \
-		--features=singlepass,default-backend-singlepass
-	$(RUNTIME_CHECK) \
-		--features=cranelift,default-backend-cranelift,debug
-	$(RUNTIME_CHECK) --release \
-		--features=cranelift,default-backend-cranelift
-	$(RUNTIME_CHECK) \
-		--features=llvm,default-backend-llvm,debug
-	$(RUNTIME_CHECK) --release \
-		--features=llvm,default-backend-llvm
+		--features=singlepass,cranelift,llvm,cache,debug,deterministic-execution
 
 # Release
 release:
--- a/Makefile
+++ b/Makefile
@@ -224,7 +224,7 @@ check: check-bench
 	cargo check --manifest-path lib/runtime/Cargo.toml
 
 	$(RUNTIME_CHECK) \
-		--features=singlepass,cranelift,llvm,cache,debug,deterministic-execution
+		--features=default-backend-singlepass,singlepass,cranelift,llvm,cache,debug,deterministic-execution
 
 # Release
 release:
--- a/Makefile
+++ b/Makefile
@@ -222,8 +222,38 @@ check: check-bench
 	# as default, and test a minimal set of features with only one backend
 	# at a time.
 	cargo check --manifest-path lib/runtime/Cargo.toml
+	# Check some of the cases where deterministic execution could matter
+	cargo check --manifest-path lib/runtime/Cargo.toml --features "deterministic-execution"
+	cargo check --manifest-path lib/runtime/Cargo.toml --no-default-features \
+		--features=default-backend-singlepass,deterministic-execution
+	cargo check --manifest-path lib/runtime/Cargo.toml --no-default-features \
+		--features=default-backend-llvm,deterministic-execution
+	cargo check --release --manifest-path lib/runtime/Cargo.toml
 
 	$(RUNTIME_CHECK) \
+		--features=cranelift,cache,debug,llvm,singlepass,default-backend-singlepass
+	$(RUNTIME_CHECK) --release \
+		--features=cranelift,cache,llvm,singlepass,default-backend-singlepass
+	$(RUNTIME_CHECK) \
+		--features=cranelift,cache,debug,llvm,singlepass,default-backend-cranelift
+	$(RUNTIME_CHECK) --release \
+		--features=cranelift,cache,llvm,singlepass,default-backend-cranelift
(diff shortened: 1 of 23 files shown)


syrusakbary:
bors try

syrusakbary:
bors try

syrusakbary:
bors try-

syrusakbary:
bors try

Makefile
```diff
@@ -222,38 +222,9 @@ check: check-bench
 	# as default, and test a minimal set of features with only one backend
 	# at a time.
 	cargo check --manifest-path lib/runtime/Cargo.toml
-	# Check some of the cases where deterministic execution could matter
-	cargo check --manifest-path lib/runtime/Cargo.toml --features "deterministic-execution"
-	cargo check --manifest-path lib/runtime/Cargo.toml --no-default-features \
-		--features=default-backend-singlepass,deterministic-execution
-	cargo check --manifest-path lib/runtime/Cargo.toml --no-default-features \
-		--features=default-backend-llvm,deterministic-execution
-	cargo check --release --manifest-path lib/runtime/Cargo.toml
```
syrusakbary: Note: After a deep analysis I disabled all the unuseful checks, since only the default backends are mutually exclusive. For the rest we can check all the same features at once.

MarkMcCaskey: It's still useful to check things like `debug`  in both states with everything else enabled because that has a very high chance of breaking due to our lints, which may be an argument for disabling the lints.

The other features can interact, too, it's just less likely.  My opinion is that the cost of a breakage in the build is very expensive in terms of developer time and resyncing on master and the cost to check it in CI is very relatively cheap so I don't like this trade off.  My opinion is that we should test all valid feature sets in CI.  What `default-backend-singlepass,singlepass,cranelift,llvm,cache,debug,deterministic-execution` says to me is that using any default backend but singlepass with other compilers enabled isn't supported, you must use the debug flag and deterministic execution, and caching because we don't test any other uses, therefore we won't even know if they don't work

lib/clif-backend/src/code.rs
```diff
@@ -58,8 +58,8 @@ impl ModuleCodeGenerator<CraneliftFunctionCodeGenerator, Caller, CodegenError>
         unimplemented!("cross compilation is not available for clif backend")
     }
 
-    fn backend_id() -> Backend {
-        Backend::Cranelift
+    fn backend_id() -> String {
+        "cranelift".to_string()
```
losfair: I think it would be better to have typed backend ids. An approach similar to `InternalField` for middlewares might work: https://github.com/wasmerio/wasmer/blob/f2f1a08e07988947042252122048e3155b96bbe6/lib/runtime-core/src/vm.rs#L139

lib/runtime-core/src/vm.rs
```diff
@@ -1118,7 +1118,7 @@ mod vm_ctx_tests {
 
                 func_assoc: Map::new(),
                 signatures: Map::new(),
-                backend: Backend::Cranelift,
+                backend: "".to_string(),
```
MarkMcCaskey: ⛳️ 
`Default::default()`

src/bin/wasmer.rs
```diff
@@ -919,29 +926,42 @@ fn interactive_shell(mut ctx: InteractiveShellContext) -> ShellExitOperation {
...
-            && (binary_size > 10485760 || cfg!(target_arch = "aarch64"))
-        {
-            options.backend = Backend::Singlepass;
-        } else {
-            options.backend = Backend::Cranelift;
+    match backend {
+        Backend::Auto => {
+            #[cfg(feature = "backend-singlepass")]
+            {
+                let binary_size = match &_path.metadata() {
```
MarkMcCaskey: :golf: `_path` -> `path`

lib/runtime-core/src/backend.rs
```diff
@@ -104,22 +50,6 @@ pub struct InlineBreakpoint {
...
-    fn str_repr_matches() {
-        // if this test breaks, think hard about why it's breaking
-        // can we avoid having these be different?
-
-        for &backend in &[Backend::Cranelift, Backend::LLVM, Backend::Singlepass] {
-            assert_eq!(backend, Backend::from_str(backend.to_string()).unwrap());
-        }
-    }
-}
-
```
MarkMcCaskey: Looks like these tests didn't survive the migration -- looks like we should keep it around as long as we have an enum which can get out of date

lib/runtime-core/src/vm.rs
MarkMcCaskey: Also, this line is a bit strange, why would we have an invalid backend here?

syrusakbary:
> * the current changes to the Makefile in the check section remove most of the checking we do -- we can't ship that.
I just reverted that

> * We can improve type safety, in addition or instead of using a static string (described below), by providing a wrapper type -- this is more for our sake than the public API's sake, but I think it will improve readability in some places.
> * I think &'static str might be a better choice than String. String is a dynamic, runtime value, &'static str is a compile-time string. It seems that being forced to use values you declared at compile-time is less error prone than a dynamic string here. I personally can't think of any case where the string would need to be dynamic right now.

Agreed. I talked with @losfair and he thinks we can use a similar strategy as we are doing with Internal Fields (here's a POC: #1106 ).

I think we can merge this, and then merge the other PR apart, or we can do the static str changes meanwhile. Let me know!

lib/runtime-core/src/backend.rs
syrusakbary: This is now solved :)

MarkMcCaskey:
> I think we can merge this, and then merge the other PR apart, or we can do the static str changes meanwhile. Let me know!

Sure, I'm fine merging in in-progress work in general as long as we have plans to follow up and we don't release until the public APIs are in the shape we want them to be.

lib/clif-backend/src/code.rs
MarkMcCaskey: We can do this with a wrapper struct and (static) strings too.

I think we'll probably want to have a way to print out the name of the backend in either case, so we'll either need to have the string be part of the `id` or have a mapping from id to string.

lib/runtime-core/Cargo.toml
```diff
@@ -52,7 +52,7 @@ cc = "1.0"
 [features]
 debug = []
 trace = ["debug"]
-# backend flags used in conditional compilation of Backend::variants
+# backend flags no longer used
```
MarkMcCaskey: I'd prefer just deleting them if we're not using them!

lib/runtime/Cargo.toml
```diff
@@ -24,6 +24,14 @@ path = "../clif-backend"
...
 optional = true
 
+# Dependencies for caching.
+[dependencies.serde]
+version = "1.0"
+# This feature is required for serde to support serializing/deserializing reference counted pointers (e.g. Rc and Arc).
+features = ["rc"]
+[dependencies.serde_derive]
+version = "1.0"
+
```
MarkMcCaskey: It looks like this chunk of the Cargo.toml came from `runtime-core`, do you know if we can remove the feature from `runtime-core`?

lib/runtime/src/lib.rs
```diff
@@ -268,11 +345,31 @@ pub fn compiler_for_backend(backend: Backend) -> Option<Box<dyn Compiler>> {
...
+        // if this test breaks, think hard about why it's breaking
+        // can we avoid having these be different?
+
+        for &backend in &[
+            #[cfg(feature = "cranelift")]
+            Backend::Cranelift,
+            #[cfg(feature = "llvm")]
+            Backend::LLVM,
+            #[cfg(feature = "singlepass")]
+            Backend::Singlepass,
```
MarkMcCaskey: We should add `Auto` here as well

lib/runtime/Cargo.toml
syrusakbary: Unfortunately both need serde_derive as they use `#[derive(Serialize, Deserialize, ...)]`.

lib/runtime/src/lib.rs
syrusakbary: I skipped `Auto` here in propose since auto is not a real backend (but more like a feature that a frontend will use/define)

lib/runtime/Cargo.toml
MarkMcCaskey: Oops, the git diff shows 4 lines, I intended to point to just the top one with `features = ["rc"]`!

MarkMcCaskey: It's a very minor thing, just thought it'd be better to remove it if it's not needed

PR #1106, opened 2019-12-23
Generic unique identifiers.

losfair:
This PR adds a generic unique identifier implementation that can be used for e.g. middleware `InternalField`s and backend IDs.

TODO:

- [ ] Make it work on stable (macros?)
- [ ] Tests

### DIFF
--- a/lib/runtime-core/src/unique_id.rs
+++ b/lib/runtime-core/src/unique_id.rs
@@ -1,0 +1,78 @@
+//! Unique identifiers.
+
+use std::cell::UnsafeCell;
+use std::marker::PhantomData;
+use std::sync::{
+    atomic::{AtomicUsize, Ordering},
+    Once,
+};
+
+/// The backing structure providing configuration and storage for a UniqueId.
+pub trait UniqueIdBacking {
+    /// Returns the maximum amount of unique ids.
+    fn get_size() -> usize;
+
+    /// Returns the backing counter for the unique id.
+    fn get_counter() -> &'static AtomicUsize;
+}
+
+/// Defines a unique ID category.
+#[macro_export]
+macro_rules! define_unique_id_category {
+    ($name:ident, $max:expr) => {
+        /// Automatically generated unique ID category.
+        pub struct $name;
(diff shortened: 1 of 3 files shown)


lib/runtime-core/src/unique_id.rs
```diff
@@ -0,0 +1,78 @@
...
+    };
+}
+
+/// A unique identifier.
+pub struct UniqueId<B: UniqueIdBacking> {
+    /// Init once field.
+    init: Once,
+    /// Inner field.
+    inner: UnsafeCell<usize>,
+    _phantom: PhantomData<B>,
```
MarkMcCaskey: And if you have context on this part of the code, I'd appreciate a description of the phantom data and the entire thing struct in general -- if you don't understand the code then don't worry about it

lib/runtime-core/src/unique_id.rs
```diff
@@ -0,0 +1,78 @@
...
+        /// Automatically generated unique ID category.
+        pub struct $name;
+        impl $crate::unique_id::UniqueIdBacking for $name {
+            fn get_size() -> usize {
+                $max
+            }
+
+            fn get_counter() -> &'static ::std::sync::atomic::AtomicUsize {
+                static COUNTER: ::std::sync::atomic::AtomicUsize =
+                    ::std::sync::atomic::AtomicUsize::new(0);
```
MarkMcCaskey: We can remove the leading `::` now and consider importing `std::sync::atomic` or at least `sync` so this can be `sync::atomic::AtomicUsize` or `atomic::AtomicUsize`

lib/runtime-core/src/unique_id.rs
```diff
@@ -0,0 +1,78 @@
+//! Unique identifiers.
```
MarkMcCaskey: I'd appreciate more of a description here too, about exactly what it is and why we'd use it.  But this is the lowest priority of the doc comments I'm asking for

syrusakbary:
Thanks to our work in the refactor and new internal structures, this PR is no longer needed.
Closing it

PR #1120, opened 2020-01-07
Port to FreeBSD

syrusakbary:
# Description

I disovered this great patch today: https://github.com/MikaelUrankar/webasm/blob/master/www/wasmer/files/patch-freebsd

And then realized that @MikaelUrankar forked wasmer to add support for FreeBSD.

I'm creating this PR to follow up on that fork and merge the changes upstream.

@MikaelUrankar let us know if everything is good to go with this PR!

# Review

### DIFF
--- a/lib/emscripten/src/syscalls/unix.rs
+++ b/lib/emscripten/src/syscalls/unix.rs
@@ -130,7 +130,19 @@ extern "C" {
     pub fn lstat64(path: *const libc::c_char, buf: *mut c_void) -> c_int;
 }
 
-#[cfg(not(target_os = "macos"))]
+// Linking to functions that are not provided by rust libc
+#[cfg(target_os = "freebsd")]
+#[link(name = "c")]
+extern "C" {
+    pub fn wait4(pid: pid_t, status: *mut c_int, options: c_int, rusage: *mut rusage) -> pid_t;
+    pub fn fdatasync(fd: c_int) -> c_int;
+    pub fn ftruncate(fd: c_int, length: i64) -> c_int;
+    pub fn lstat(path: *const libc::c_char, buf: *mut stat) -> c_int;
+}
+
+#[cfg(target_os = "freebsd")]
+use libc::madvise;
+#[cfg(not(any(target_os = "freebsd", target_os = "macos")))]
 use libc::{fallocate, fdatasync, ftruncate64, lstat, madvise, wait4};
 
 // Another conditional constant for name resolution: Macos et iOS use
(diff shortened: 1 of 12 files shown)


lib/emscripten/src/syscalls/unix.rs
```diff
@@ -130,7 +130,18 @@ extern "C" {
     pub fn lstat64(path: *const libc::c_char, buf: *mut c_void) -> c_int;
 }
 
-#[cfg(not(target_os = "macos"))]
+// Linking to functions that are not provided by rust libc
```
nlewycky: Is this permanent? If not, I'd really appreciate it if this could include a link to an issue or PR.

lib/runtime-core/image-loading-freebsd-x86-64.s
```diff
@@ -0,0 +1,127 @@
+# NOTE: Keep this consistent with `fault.rs`.
```
nlewycky: I checked it visually, but it looks like it's a copy of the linux one. In your opinion should we just rename it to something like sysv-elf and use the same for both, or do you think there's something here that FreeBSD might change and need a different image loader for?

lib/emscripten/src/syscalls/unix.rs
```diff
@@ -99,7 +99,7 @@ const WASM_TCSETSW: u32 = 0x5403;
 fn translate_ioctl(wasm_ioctl: u32) -> c_ulong {
     match wasm_ioctl {
         WASM_FIOCLEX => FIOCLEX,
-        WASM_TIOCGWINSZ => TIOCGWINSZ,
+        WASM_TIOCGWINSZ => TIOCGWINSZ as u64,
```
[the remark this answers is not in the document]
syrusakbary:
```suggestion
        WASM_TIOCGWINSZ => TIOCGWINSZ as _,
```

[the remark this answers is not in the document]
MikaelUrankar: done

lib/runtime-core/image-loading-freebsd-x86-64.s
MikaelUrankar: Yes it's the same as the linux one. I think we can use the same for Linux/FreeBSD.

lib/emscripten/src/syscalls/unix.rs
MikaelUrankar: I can try to upstream the fn definition.
There is a PR for the ucontext/mcontext part https://github.com/rust-lang/libc/pull/1630

MikaelUrankar:
Hi,
Sorry for the late reply.
The tests pass on amd64, there are some tests that fail on aarch64, I don't know if it's specific to FreeBSD/aarch64 or if it fails on Linux too.

syrusakbary:
bors try

MarkMcCaskey:
Looks like the failing lint here should be fixed by merging / rebasing with wasmer master

syrusakbary:
aarch64 tests seem to be passing :)

MarkMcCaskey:
bors try

PR #1123, opened 2020-01-08
feat(runtime-c-api) Define the `DEPRECATED` C macro.

Hywan:
This PR defines a cross-compiler `DEPRECATED(message)` macro. It must
be used as follows in Rust:

```rust
/// This is a documentation.
/// cbindgen:prefix=DEPRECATED("This is a deprecation message.")
pub extern "C" fn wasmer_foo() -> c_uint {
    42
}
```

It will generate the following C header:

```c
/**
 * This is a documentation.
 */
DEPRECATED("This is a deprecation message.")
unsigned int wasmer_foo();
```

And once this code is used by a C compiler, it will print something
like this (example from Clang):

```
…/test.c:…:…: error: 'wasmer_foo' is deprecated: This is a deprecation message. [-Werror,-Wdeprecated-declarations]
    unsigned int x = wasmer_foo();
                     ^
…/wasmer.h:…:…: note: 'wasmer_foo' has been explicitly marked deprecated here
DEPRECATED("This is a deprecation message.")
^
…/wasmer.h:…:…: note: expanded from macro 'DEPRECATED'
```

This is required for further deprecations.

(cf https://github.com/eqrion/cbindgen/issues/408)

### DIFF
--- a/lib/runtime-c-api/build.rs
+++ b/lib/runtime-c-api/build.rs
@@ -24,10 +24,20 @@ fn main() {
 #  define ARCH_X86_64
 #endif
 
+// Compatibility with non-Clang compilers.
+#if !defined(__has_attribute)
+#  define __has_attribute(x) 0
+#endif
+
+// Compatibility with non-Clang compilers.
+#if !defined(__has_declspec_attribute)
+#  define __has_declspec_attribute(x) 0
+#endif
+
 // Define the `DEPRECATED` macro.
-#if defined(GCC) || defined(__GNUC__) || defined(__clang__)
+#if defined(GCC) || defined(__GNUC__) || __has_attribute(deprecated)
 #  define DEPRECATED(message) __attribute__((deprecated(message)))
-#elif defined(MSVC)
+#elif defined(MSVC) || __has_declspec_attribute(deprecated)
 #  define DEPRECATED(message) __declspec(deprecated(message))
 #endif
 
--- a/lib/runtime-c-api/build.rs
+++ b/lib/runtime-c-api/build.rs
@@ -14,18 +14,21 @@ fn main() {
 
     let mut pre_header = r#"
 #if !defined(WASMER_H_MACROS)
+
 #define WASMER_H_MACROS
 
-#if defined(MSVC)
-#if defined(_M_AMD64)
-#define ARCH_X86_64
-#endif
+// Define the `ARCH_X86_X64` constant.
+#if defined(MSVC) && defined(_M_AMD64)
+#  define ARCH_X86_64
+#elif (defined(GCC) || defined(__GNUC__) || defined(__clang__)) && defined(__x86_64__)
+#  define ARCH_X86_64
 #endif
 
+// Define the `DEPRECATED` macro.
 #if defined(GCC) || defined(__GNUC__) || defined(__clang__)
-#if defined(__x86_64__)
-#define ARCH_X86_64
-#endif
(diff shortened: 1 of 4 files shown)


Hywan:
Ideally, we should write a procedule macro like `#[deprecated-c(message)]` to add the `#[doc="cbindgen:prefix=DEPRECATED(message)]` on the sibling item, but eh, let's see if we really need it!

syrusakbary:
I think @nlewycky will be the best one to review it! Otherwise it looks great :)

lib/runtime-c-api/build.rs
```diff
@@ -14,18 +14,21 @@ fn main() {
...
+
 #define WASMER_H_MACROS
 
-#if defined(MSVC)
-#if defined(_M_AMD64)
-#define ARCH_X86_64
-#endif
+// Define the `ARCH_X86_X64` constant.
+#if defined(MSVC) && defined(_M_AMD64)
+#  define ARCH_X86_64
```
MarkMcCaskey: I thought the `#` had to be touching the `define` 🤔

MarkMcCaskey: To clarify, the `#` before and after the `#` are Rust syntax for raw strings (we can use any number of `#` if the string `"#` appears inside the string).  The rest of the `#` are part of the C header file

lib/runtime-c-api/build.rs
```diff
@@ -14,18 +14,21 @@ fn main() {
...
+#elif (defined(GCC) || defined(__GNUC__) || defined(__clang__)) && defined(__x86_64__)
+#  define ARCH_X86_64
 #endif
 
+// Define the `DEPRECATED` macro.
 #if defined(GCC) || defined(__GNUC__) || defined(__clang__)
```
nlewycky: While I think this should work as written, I don't think `__clang__` should be listed here. Clang can run in two modes, either emulating GCC which means both defining the `__GNUC__` macro and implementing gcc's `__attribute__` extensions, or emulating MSVC which means both defining `MSVC` and implementing `__declspec`. It turns out that this will work as is because even if clang is running as MSVC-compatible, it still understands `__attribute__` from GCC. However, I think it'd be better to just leave it out.

Or we can add specific clang support like we have specific GCC and MSVC support:
```
#ifndef __has_attribute
  #define __has_attribute(x) 0  // Compatibility with non-clang compilers.
#endif
#ifndef __has_declspec_attribute
  #define __has_declspec_attribute(x) 0  // Compatibility with non-clang compilers.
#endif
```
Those `__has_attribute` and `__has_declspec_attribute` are clang inventions, but designed with the idea that other compilers could implement the same thing. With the code above you could then write `#if defined(GCC) || defined(__GNUC__) || __has_attribute(deprecated)` here and `#elif defined(MSVC) || __has_declspec_attribute(deprecated)` below.

lib/runtime-c-api/build.rs
MarkMcCaskey: It's fine from a rust perspective, I just thought it was strange C -- thanks for the clarification!

Hywan: Putting spaces before `define` helps to provide some visual structure to the block.

Marking the conversation as resolved :-)!

lib/runtime-c-api/build.rs
Hywan: Thanks, it makes a lot of sense! I learned something today!

PR #1121, opened 2020-01-07
Add test for ImportObject's `allow_missing_functions`

MarkMcCaskey:
work in progress, testing to see if this test fails CI

resolves #1118 

# Review

- [ ] Add a short description of the the change to the CHANGELOG.md file

PR #1128, opened 2020-01-09
fix(runtime-core) Avoid crashing when missing host functions are allowed

### LINKED ISSUE #1118: allow_missing_functions doesn't work in release mode
In release mode, the `allow_missing_functions` option does not work as expected. Instantiation fails with a lot of `ImportNotFound` errors even with the field set to true.

Code to repro is [here](github.com/slipher/wasmer-bug). Steps:

```
git clone https://github.com/slipher/wasmer-bug
cd wasmer-bug
cargo build --release
target/release/sandbox cgame.wasm
```

In debug mode (`cargo build && target/debug/sandbox cgame.wasm`) it works as expected. I observed the same behavior of working only in debug mode on Windows and Linux machines.

Hywan:
Fix #1118.
Address #1121.

This PR fixes 2 things:

* When droping the import backing, check that `vm::FuncCtx` isn't null before dropping it,
* Use an `always_trap` as a placeholder host function when a host function is missing.

### DIFF
--- a/lib/runtime-core/src/typed_func.rs
+++ b/lib/runtime-core/src/typed_func.rs
@@ -733,6 +733,12 @@ where
     }
 }
 
+/// Function that always fails. It can be used as a placeholder when a
+/// host function is missing for instance.
+pub(crate) fn always_trap() -> Result<(), &'static str> {
+    Err("not implemented")
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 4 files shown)


lib/runtime-core/src/typed_func.rs
```diff
@@ -258,11 +258,6 @@ where
             _phantom: PhantomData,
         }
     }
-
-    /// Get the underlying func pointer.
-    pub fn get_vm_func(&self) -> NonNull<vm::Func> {
-        self.func
-    }
```
MarkMcCaskey: Is this a breaking change?

Hywan: Nop. It was implemented for `Wasm` only, now it's implemented for `Wasm` and `Host`.

PR #1129, opened 2020-01-09
Standard exception types for singlepass backend.

losfair:


losfair:
bors try

lib/runtime-core/src/error.rs
```diff
@@ -208,6 +209,8 @@ impl std::fmt::Display for RuntimeError {
                     write!(f, "\"{}\"", s)
                 } else if let Some(s) = data.downcast_ref::<&str>() {
                     write!(f, "\"{}\"", s)
+                } else if let Some(exc_code) = data.downcast_ref::<ExceptionCode>() {
+                    write!(f, "\"{:?}\"", exc_code)
```
Hywan: Maybe we should write it is an exception code, so that the user has more insight than the variant name. My suggestion:

```rust
write!(f, "exception code \"{:?}\"", exc_code)
```

losfair: Thanks! Fixed and now the error message looks like `Error: RuntimeError: Caught exception of type "Arithmetic".`.

losfair:
@nlewycky I just updated `excludes.txt` and removed most singlepass excludes., but there are still a few failures related to `linking.wast`. I'm not really sure what's going on there.

By the way, I think it would be good to refactor `RuntimeError` and remove the `Any` box (and make `ExceptionCode` a variant + remove plain strings) - maybe in a future PR.

fckt:
Great job! I have a small question: why these traps are `RuntimeError::Error`? Shouldn't it be a `RuntimeError::Trap`?

PR #1133, opened 2020-01-10
feat(runtime-c-api) Implement `wasmer_trap`

Hywan:
Idea is to provide an API to get fallible host function by calling `wasmer_trap` to run the Wasmer trapping API.

This is probably the easiest solution to not break the existing API, and not add a lot of complexity in the code.


lib/runtime-c-api/src/import/mod.rs
```diff
@@ -661,6 +668,24 @@ pub unsafe extern "C" fn wasmer_import_func_new(
...
 
+/// Hello
+#[no_mangle]
+#[allow(clippy::cast_ptr_alignment)]
+pub unsafe extern "C" fn wasmer_import_trap(
+    ctx: *const wasmer_instance_context_t,
+    error_message: *const c_char,
+) -> c_uint {
+    let ctx = &*(ctx as *const Ctx);
+    let error_message = CStr::from_ptr(error_message).to_str().unwrap();
```
MarkMcCaskey: `CStr::from_ptr` calls `slice::from_raw_parts` without checking if the pointer is null, [from raw parts](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html) says that the pointer must not be null; we should probably check it on our side then

lib/runtime-c-api/src/import/mod.rs
```diff
@@ -661,6 +668,24 @@ pub unsafe extern "C" fn wasmer_import_func_new(
...
+
+    (&*ctx.module)
```
MarkMcCaskey: technically this can trigger UB too, if the pointer is aligned to Rust's liking or if it's null/invalid (but this may be a larger issue in the C API); not sure there's as much we can do here, though checking for null is something

lib/runtime-c-api/src/import/mod.rs
Hywan: Thanks. It's a WIP ;-).

Hywan:
Ready for a review!

lib/runtime-c-api/src/import/mod.rs
```diff
@@ -661,6 +682,53 @@ pub unsafe extern "C" fn wasmer_import_func_new(
...
+
+    let ctx = &*(ctx as *const Ctx);
+    let error_message = CStr::from_ptr(error_message).to_str().unwrap();
+
+    (&*ctx.module)
+        .runnable_module
+        .do_early_trap(Box::new(error_message)); // never returns
+
+    #[allow(unreachable_code)]
+    wasmer_result_t::WASMER_OK
```
MarkMcCaskey: Perhaps we should panic here?  Seems like a small thing, not sure what best practice is here

lib/runtime-core/src/backend.rs
```diff
@@ -219,10 +219,11 @@ pub trait RunnableModule: Send + Sync {
     }
 
     /// A wasm trampoline contains the necessary data to dynamically call an exported wasm function.
-    /// Given a particular signature index, we are returned a trampoline that is matched with that
+    /// Given a particular signature index, we are returning a trampoline that is matched with that
```
MarkMcCaskey: I prefer `returned` to `returning` here;

`, we return a trampoline...` is probably best though.

lib/runtime-c-api/src/import/mod.rs
Hywan: cbindgen doesn't generate a C bindgen for a Rust function with `!` as returned type.

Since I need to error before calling `do_early_trap`, I've decided to return a `wasmer_result_t`. And in the best scenario, I return `WASMER_OK`, even if it's unreachable. I'll add more documentation to explain.

lib/runtime-core/src/backend.rs
Hywan: I thought it was an English typo.

Hywan:
Go for `wasmer_trap` :-).

lib/runtime-core/src/backend.rs
MarkMcCaskey: I believe it's grammatically correct and called the "passive voice", it's often discouraged in English though because it takes more words and ends up being less specific. For example, I generally prefer direct writing like, "This function takes a function signature and returns a corresponding trampoline".

PR #1140, opened 2020-01-13
Use blake3 instead of blake2_simd

syrusakbary:
# Description

Blake3 has a much faster hashing rate than Blake2/Blake2b_simd. This PR updates wasmer to use it as the default hashing algorithm.

More info here: https://www.infoq.com/news/2020/01/blake3-fast-crypto-hash/

# Review


syrusakbary:
bors try

lib/runtime-core/build.rs
```diff
@@ -1,13 +1,13 @@
-use blake2b_simd::blake2bp;
+use blake3;
```
CryZe: Unless I'm mistaken about how it works in build.rs, `use item;` never does anything meaningful anymore. Due to uniform paths, if something is reachable in a certain way via use, it is also reachable that way in other paths throughout the module. So since `use blake3;` works, `blake3::Hasher` must therefore already work in the module without having to specifically import it via use. rustc should honestly emit a warning there, but it doesn't yet.

lib/runtime-core/src/cache.rs
```diff
@@ -3,7 +3,7 @@
 //! and loaded to allow skipping compilation and fast startup.
 
 use crate::{module::ModuleInfo, sys::Memory};
-use blake2b_simd::blake2bp;
+use blake3;
```
CryZe: Same here.

syrusakbary:
bors try

lib/runtime-core/src/cache.rs
```diff
@@ -59,16 +58,16 @@ impl WasmHash {
     /// is, in fact, a wasm module.
     pub fn generate(wasm: &[u8]) -> Self {
         let mut first_part = [0u8; 32];
-        let mut second_part = [0u8; 32];
+        // We don't use the second part for now, since the blake3 hash is 32 byte array
+        let second_part = [0u8; 32];
```
MarkMcCaskey: This should be deleted -- it's too bad it produces a smaller digest, but it's probably fine

PR #1142, opened 2020-01-13
doc(runtime-c-api) Setup Doxygen

Hywan:
With @MarkMcCaskey, we discussed generating the documentation of `wasmer-runtime-c-api` with [Doxygen](http://www.doxygen.nl/).

This patch does the following:

1. Set up a `doxyfile`, which is the configuration file for Doxygen.
2. Set up a `doc/header.html` and `doc/footer.html` files, because Doxygen HTML output is… well, not a master piece 😉.
3. Set up a `doc/css/wasmer.css` stylesheet. I tried to mimic Wasmer color scheme. Usually, I'm not a bad designer, but Doxygen HTML output is so “alembic-ed” that it makes customizing the CSS a real challenge for the very short period of time I've.

Anyway, I believe it's an acceptable first step.

To generate the documentation:

```sh
$ doxygen doxyfile
```

And then open `doc/html/index.html`.

It looks like this:

Screen Shot 2020-01-13 at 15 20 09
Screen Shot 2020-01-13 at 15 20 25
Screen Shot 2020-01-13 at 15 20 35
Screen Shot 2020-01-13 at 15 20 57

The next step is to publish the documentation on Github Pages, thoughts @syrusakbary?

### DIFF
--- a/lib/runtime-c-api/.gitignore
+++ b/lib/runtime-c-api/.gitignore
@@ -1,0 +1,1 @@
+doc/html/
--- a/lib/runtime-c-api/doc/index.md
+++ b/lib/runtime-c-api/doc/index.md
@@ -1,0 +1,125 @@
+# Wasmer Runtime C API
+
+[Wasmer] is a standalone [WebAssembly] runtime, aiming to be fully
+compatible with WASI, Emscripten, Rust and Go. [Learn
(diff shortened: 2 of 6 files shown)


syrusakbary:
That's great! Thanks @Hywan :)

PR #1143, opened 2020-01-13
Set backend_id to static str

syrusakbary:
# Description

As per feedback in #1099, set backend_id to static str.

# Review

### DIFF
--- a/lib/runtime-core/src/parse.rs
+++ b/lib/runtime-core/src/parse.rs
@@ -82,7 +82,7 @@ pub fn read_module<
 
         func_assoc: Map::new(),
         signatures: Map::new(),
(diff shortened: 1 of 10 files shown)


lib/runtime-core/src/parse.rs
```diff
@@ -82,7 +82,7 @@ pub fn read_module<
 
         func_assoc: Map::new(),
         signatures: Map::new(),
-        backend: MCG::backend_id(),
+        backend: MCG::backend_id().to_string(),
```
MarkMcCaskey: If we know that it has to be `&'static str`, then we can change it to be a `&'static str` in the structs too and remove all uses of `to_string()`!

PR #1159, opened 2020-01-17
Document runtime-core::vmcalls functions

MarkMcCaskey:


lib/runtime-core/src/vmcalls.rs
```diff
@@ -1,3 +1,5 @@
+//! Unchecked functions for querying the size of memories and growing them.
```
nlewycky: Yes, but I think it makes more sense to focus on the fact that these are "vmcalls", that is, these are the implementatons of functions called directly by our generated code. That includes the code for the `memory.grow` opcode, but could include any VM calls in the future. (While I suppose these can also be called from Rust, I don't know whether they are, and they probably shouldn't be.)

lib/runtime-core/src/vmcalls.rs
```diff
@@ -31,16 +43,32 @@ pub unsafe extern "C" fn local_static_memory_grow(
     ret
 }
 
+/// Get the size in [`Pages`] of a local [`StaticMemory`].
```
nlewycky: Optional: I'd rather that were
```suggestion
/// Get the size of a local [`StaticMemory`] in [`Pages`].
```

PR #1168, opened 2020-01-20
doc(runtime-c-api) Improve documentation of the `instance`, `error` and `export` modules

Hywan:
Daily doc', that turned to be an obsession. The entire `instance`, `error` modules are fully documented. The `exports` module's documentation have been improved too.

This PR also contains some “bug fixes” when `instance` pointers are null for instance (in this case, we do nothing or we return `void`).

### DIFF
--- a/lib/runtime-c-api/src/instance.rs
+++ b/lib/runtime-c-api/src/instance.rs
@@ -418,7 +418,7 @@ pub unsafe extern "C" fn wasmer_instance_exports(
 ///
 /// // You can read your data.
 /// {
-///     my_data *data = wasmer_instance_context_data_get(wasmer_instance_context_get(instance));
+///     my_data *data = (my_data*) wasmer_instance_context_data_get(wasmer_instance_context_get(instance));
 ///     // …
 /// }
 /// ```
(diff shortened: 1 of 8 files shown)


lib/runtime-c-api/src/instance.rs
```diff
@@ -171,52 +178,84 @@ pub unsafe extern "C" fn wasmer_instantiate(
...
 
-/// Given:
-/// * A prepared `wasmer` import-object
-/// * A compiled wasmer module
+/// Returns the instance context. Learn more by looking at the
+/// `wasmer_instance_context_t` struct.
 ///
-/// Instantiates a wasmer instance
-#[no_mangle]
-pub unsafe extern "C" fn wasmer_module_import_instantiate(
```
Hywan: This function has been moved to the `module` module.

MarkMcCaskey:
Not directly related to this PR, but on the topic of documentation of `unsafe` functions:  I'd like to eventually turn on the clippy lint that denies `unsafe` functions without a `# Safety` section in its doc comments.  My reasoning is that the entire point of unsafe is that it means that the caller has to uphold the invariants manually, thus if we aren't documenting what those invariants are, then we have absolutely no chance of using our unsafe functions correctly over time.  The C API contains a lot of unsafe functions but it's a bit of a different case -- what are your thoughts on this lint in the context of the C API?  It's trivial to disable it, but I think it might still be valuable to make sure we're explicit on every unsafe function about what exactly we're assuming.

Hywan:
I agree we must share with the user what “unsafe” means in this situation. If Clippy has such lints, we should turn it on.

PR #1174, opened 2020-01-24
Clean up misc. bits of runtime-core

MarkMcCaskey:
Also includes the lock file which has been showing up in my diffs for a while, so I guess I should commit it


lib/runtime-core/src/fault.rs
```diff
@@ -259,7 +259,7 @@ pub fn allocate_and_run<R, F: FnOnce() -> R>(size: usize, f: F) -> R {
         const NUM_SAVED_REGISTERS: usize = 31;
         let stack_begin = stack
             .as_mut_ptr()
-            .offset((end_offset - 4 - NUM_SAVED_REGISTERS) as isize);
+            .add(end_offset - 4 - NUM_SAVED_REGISTERS);
         let stack_end = stack.as_mut_ptr().offset(end_offset as isize);
```
nlewycky: What about this one?

Issue #1178, opened 2020-01-27
Make wabt an optional dependency


losfair:
Currently we depend on `wabt` for tests and various type definitions. However `wabt` requires a C++ compiler to build, which is not included in the [musl](https://www.musl-libc.org/) toolchain.

This issue is blocking https://github.com/wasmerio/wasmer/issues/1173 , and needs to be fixed.

nlewycky:
Why doesn't the musl toolchain include a C++ compiler? More broadly, is there documentation covering what it means to be a rust toolchain and why it includes a C++ compiler or how we should depend on C++ code from rust if it's not part of the rust toolchain package, etc?

losfair:
Related:

- https://github.com/rust-lang/cargo/issues/3359
- https://github.com/emk/rust-musl-builder/issues/53

I tried symlinking `g++` to `musl-g++` but it doesn't work:

```
error: could not find native static library `stdc++`, perhaps an -L flag is missing?

error: aborting due to previous error

error: could not compile `wabt-sys`
```

musl is a libc implementation and does not include C++ standard libraries. To depend on C++ code w/o dynamic linking I think we need to modify the build system to use a native C++ compiler, but there can be ABI incompability issues.

Issue #1173, opened 2020-01-24
Cannot compile with target `x86_64-unknown-linux-musl`


ethanfrey:
### Describe the bug

I am embedding wasmer in a go application and want to make a proper static binary. I also had issues with the `cdylib` I compiled under recent Ubuntu not working on old versions of linux (like CentOS7, which uses glibc 2.12 I believe). 

So, after installing the toolchains, I ended up with something like: 

`RUSTFLAGS="-C target-feature=-crt-static" rustup run nightly cargo build --release  --target x86_64-unknown-linux-musl`

This managed to compile and link most of the deps, but it dies on `wasmer-runtime-core`

This is build on Ubuntu 18.04 with a recent rust nightly (12-26-2019) and wasmer 0.12.

I ran the following beforehand:

```sh
	sudo apt install musl-tools
	rustup target add x86_64-unknown-linux-musl
	rustup target add x86_64-unknown-linux-musl --toolchain=nightly
```

### Expected behavior

I thought this would compile.

### Actual behavior

```
error[E0432]: unresolved imports `libc::_libc_xmmreg`, `libc::REG_R10`, `libc::REG_R11`, `libc::REG_R12`, `libc::REG_R13`, `libc::REG_R14`, `libc::REG_R15`, `libc::REG_R8`, `libc::REG_R9`, `libc::REG_RAX`, `libc::REG_RBP`, `libc::REG_RBX`, `libc::REG_RCX`, `libc::REG_RDI`, `libc::REG_RDX`, `libc::REG_RIP`, `libc::REG_RSI`, `libc::REG_RSP`
   --> /home/ethan/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-runtime-core-0.12.0/src/fault.rs:552:9
    |
552 |         _libc_xmmreg, ucontext_t, REG_R10, REG_R11, REG_R12, REG_R13, REG_R14, REG_R15, REG_R8,
    |         ^^^^^^^^^^^^              ^^^^^^^  ^^^^^^^  ^^^^^^^  ^^^^^^^  ^^^^^^^  ^^^^^^^  ^^^^^^ no `REG_R8` in the root
    |         |                         |        |        |        |        |        |
    |         |                         |        |        |        |        |        no `REG_R15` in the root
    |         |                         |        |        |        |        no `REG_R14` in the root
    |         |                         |        |        |        no `REG_R13` in the root
    |         |                         |        |        no `REG_R12` in the root
[...]

PR #1180, opened 2020-01-28
Fix compilation for target `x86_64-unknown-linux-musl`.

losfair:
Fixes compilation failure when built for `x86_64-unknown-linux-musl`.

- Allows disabling the `wabt` feature to avoid depending on a C++ compiler.
- Disables XMM register dump on fault as a workaround for missing musl libc types (https://github.com/rust-lang/libc/pull/1646).

Fixes https://github.com/wasmerio/wasmer/issues/1178 and https://github.com/wasmerio/wasmer/issues/1173 .

### DIFF
--- a/lib/runtime-core/src/fault.rs
+++ b/lib/runtime-core/src/fault.rs
@@ -570,13 +570,13 @@ pub unsafe fn get_fault_info(siginfo: *const c_void, ucontext: *mut c_void) -> F
 #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
 /// Get fault info from siginfo and ucontext.
 pub unsafe fn get_fault_info(siginfo: *const c_void, ucontext: *mut c_void) -> FaultInfo {
-    use crate::state::x64::XMM;
     use libc::{
-        _libc_xmmreg, ucontext_t, REG_R10, REG_R11, REG_R12, REG_R13, REG_R14, REG_R15, REG_R8,
+        ucontext_t, REG_R10, REG_R11, REG_R12, REG_R13, REG_R14, REG_R15, REG_R8,
         REG_R9, REG_RAX, REG_RBP, REG_RBX, REG_RCX, REG_RDI, REG_RDX, REG_RIP, REG_RSI, REG_RSP,
     };
 
-    fn read_xmm(reg: &_libc_xmmreg) -> u64 {
+    #[cfg(not(target_env = "musl"))]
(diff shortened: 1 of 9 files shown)


losfair:
bors try

losfair:
bors try-

losfair:
bors try

lib/runtime-core/src/fault.rs
```diff
@@ -615,30 +615,46 @@ pub unsafe fn get_fault_info(siginfo: *const c_void, ucontext: *mut c_void) -> F
...
-            Some(read_xmm(&fpregs._xmm[12]));
-        known_registers[X64Register::XMM(XMM::XMM13).to_index().0] =
-            Some(read_xmm(&fpregs._xmm[13]));
-        known_registers[X64Register::XMM(XMM::XMM14).to_index().0] =
-            Some(read_xmm(&fpregs._xmm[14]));
-        known_registers[X64Register::XMM(XMM::XMM15).to_index().0] =
-            Some(read_xmm(&fpregs._xmm[15]));
+    // Skip reading floating point registers when building with musl libc.
+    // FIXME: Depends on https://github.com/rust-lang/libc/pull/1646
+    #[cfg(not(target_env = "musl"))]
```
MarkMcCaskey: Do we need to handle the other case too?  Like

```
#[cfg(target_env = "musl")]
unimplemented!("blocked on https://github.com/rust-lang/libc/pull/1646");
```

Or will it work fine as-is?

losfair: This is on the always-taken path for fault handling so panicking here will break things that depend on fault handling, like middlewares. Now all XMM registers will be defaulted to zero if not dumped; do you think it is fine?

MarkMcCaskey: Hmm, I'm not familiar enough to know what the impacts of that are: if you think this leaves the system in a sensible state where execution makes sense then I'm happy with it!  Otherwise, I don't think compiling and failing silently is necessarily a step up from panicking or not compiling.

losfair: The only case where doing so is possibly incorrect is when we take a managed snapshot and then unwind out from WebAssembly code (where floating point state will be lost). This feature is not widely used (yet) and I think it is "cold" enough to ignore for now and defer to after a future upstream fix in musl-libc.

MarkMcCaskey: Okay, that sounds reasonable to me!  We just need to remember to not forget to update this!

PR #1181, opened 2020-01-29
doc(runtime-c-api) Improve documentation of the `value` module

Hywan:
Hopla. Daily doc'!


lib/runtime-c-api/doxyfile
```diff
@@ -118,7 +118,6 @@ WARN_LOGFILE           =
 #---------------------------------------------------------------------------
 INPUT                  = doc/index.md
 INPUT                 += wasmer.h
-INPUT                 += wasmer.hh
```
Hywan: The idea here is to avoid variants/members duplications in Doxygen for structs and enums.

PR #1184, opened 2020-01-31
Add minor documentation updates

MarkMcCaskey:


lib/llvm-backend/src/intrinsics.rs
```diff
@@ -1,3 +1,8 @@
+//! Code for dealing with LLVM and VM intrinsics.
+//!
+//! Intrinsics are built-in operations. LLVM intrinsics are used to precisely
```
nlewycky: > LLVM intrinsics are used to precisely compute values

Optional: I'm not sure I'd emphasize precision of computation here, makes it sound like we're using them because they give more exact bits in a floating point calculation. There's some things in LLVM which are instructions and others relegated to intrinsics, and that once meant something, but now it really doesn't, the difference is just whether all passes need to know about it (like how match statements in rust cover everything) or if they can just treat it like a call to an opaque function. I wouldn't try to describe them at all, but https://llvm.org/docs/LangRef.html#intrinsic-functions gives LLVM's wording.

PR #1192, opened 2020-02-06
Use `ExceptionCode` for error representation.

losfair:
Extends https://github.com/wasmerio/wasmer/pull/1129 to all backends.


losfair:
bors try

PR #1195, opened 2020-02-07
Add a new test that two instances from the same import object can have their tables mutated independently.

nlewycky:


lib/runtime-core-tests/tests/imports.rs
```diff
@@ -292,3 +292,51 @@ test!(
...
+)
+"#;
+    let wasm_binary2 = wat2wasm(MODULE2.as_bytes()).expect("WASM not valid or malformed");
+    let module2 = compile_with(&wasm_binary2, &get_compiler()).unwrap();
+    let import_object2 = imports! {
+        "env" => {
+            "table" => instance1b.get_export("table").unwrap(),
+        },
+    };
+    let instance2 = module2.instantiate(&import_object2).unwrap();
```
MarkMcCaskey: Does instantiating both modules with this explicit import object change anything? `f2a` and `f2b` here.  I'm surprised this test passes 🤔

Hywan:
Still relevant. Must be ported to the new API.

PR #1206, opened 2020-02-11
Implement DerefMut for non-atomic `MemoryView`s

MarkMcCaskey:
This is probably more correct for cases where we need a mutable slice to memory...

Unfortunately it's still trivial to use this API in an unsound way (this change does not affect that).

The return type of `deref_mut` must be the same as the return type of `deref`.

Looking for feedback and ideas about memory generally here.

# Review


MarkMcCaskey:
@Hywan 

thinking about this change again I realized that this allows aliased `&mut` to the same data very easily which is definitely bad -- probably worse than what we currently have.  Because we can get a `MemoryView` with a `&Memory`, we can get multiple of them and then get multiple `&mut` to the slice through this.  So I think we can't ship this PR as-is

MarkMcCaskey:
Closing for now as aliasing `&mut T` is not the API we want here

PR #1209, opened 2020-02-11
Maybe improve safety of MemoryView Deref

MarkMcCaskey:
Possibly slightly better... (effectively does exactly the same thing the same way):

Notably this code still violates the safety rules of its unsafe functions:

It violates rule 2 from https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html :

> The memory referenced by the returned slice must not be accessed through any other pointer (not derived from the return value) for the duration of lifetime ’a. Both read and write accesses are forbidden.

Some comments in the stdlib indicate that casting between `Cell<T>` and `T` isn't something that's safe to do outside of the stdlib 🤷‍♂ .

Fundamentally, `MemoryView` is unsound.  Our docs say that it does not have undefined behavior, but we don't properly justify that.   I believe it relies very heavily on undefined behavior.

Looking for feedback here about what we can do.  It seems like we'll have to break the public API to make a sound API... if we can decide on a sufficiently good one, we can deprecate the old one and roll out the new one to prevent user breakage.

We may need to use something like https://docs.rs/crossbeam/0.7.3/crossbeam/atomic/struct.AtomicCell.html or implement something like it.

Where I'm at in regards to memory access from the host is that it's not possible to make any kind of efficient API that is not entirely `unsafe`.  Even given that, we need to figure out the bounds with which the `unsafe` functions can be used relatively safely.  I think even if we marked our existing functions as `unsafe`, it's too complex to figure out whether or not you're using it correctly...

Bonus points if your solution works across multiple threads properly...

# Review


Hywan:
I mean, the PR updates the code in a better way, so I'm tempted to approve it. Can we open an issue to not forget to fix the safety issue of `MemoryView`? There is another important undefined behavior with it: When the memory is reallocated, `MemoryView` isn't invalidated, and thus can be used but it points to the old memory space, which is dangerous.

losfair:
Can you provide a link to the standard library source that indicates unsafety for casting between `&T` and `&Cell<T>`? For trivial types like the builtin ones (`u8`, `u32`, etc.) I believe it is safe - if there are any special cases, we can add an unsafe marker trait for that.

MarkMcCaskey:
> Can you provide a link to the standard library source that indicates unsafety for casting between `&T` and `&Cell<T>`? For trivial types like the builtin ones (`u8`, `u32`, etc.) I believe it is safe - if there are any special cases, we can add an unsafe marker trait for that.

It's not documented in any obvious place, I found it in a non-doc comment in the source code of UnsafeCell: https://doc.rust-lang.org/src/core/cell.rs.html#1558-1561

Quoting:

>  // We can just cast the pointer from `UnsafeCell<T>` to `T` because of
>  // #[repr(transparent)]. This exploits libstd's special status, there is
>  // no guarantee for user code that this will work in future versions of the compiler!

So it sounds like it works but that there's no guarantee that this will continue to work.  Although I'm not strictly sure, if you dig around the source code of the Rust stdlib, you'll find lots of special markers which let it do things that normal Rust code cannot.  For example, the implementation of box uses the old `box` keyword and a magic attribute...  This doesn't look as magic, but I'm not really sure how to tell what is and isn't magic

MarkMcCaskey:
> I mean, the PR updates the code in a better way, so I'm tempted to approve it. Can we open an issue to not forget to fix the safety issue of `MemoryView`? There is another important undefined behavior with it: When the memory is reallocated, `MemoryView` isn't invalidated, and thus can be used but it points to the old memory space, which is dangerous.



> I mean, the PR updates the code in a better way, so I'm tempted to approve it. Can we open an issue to not forget to fix the safety issue of `MemoryView`? There is another important undefined behavior with it: When the memory is reallocated, `MemoryView` isn't invalidated, and thus can be used but it points to the old memory space, which is dangerous.

I agree that a tracking issue would be good!  I'm going to wait a bit longer to merge this so we can discuss it as much as people care to!

So `MemoryView` itself has a semi-public `new` method, as long as the lifetime of `MemoryView` is tied to the lifetime of `Memory`, ~then it should be okay~.... actually nevermind!  You're correct.  Because `Memory::grow` takes a `&self` and not a `&mut self`, we can in fact grow it while we have a memory view... if it took a `&mut self`, this would be a non-issue.

I'm not sure why all of `Memory`'s methods take a `&self` instead of a `&mut self`: it has internal locks but the problem is that `MemoryView` does not hold on to that lock.  Perhaps that's an easy fix to make that the case though.  Though that just means that we've created an API that will deadlock...  the lock probably has a `try_lock` method or something and we could do some kind of `async` style thing to access memory but ideally we could prevent this at compile time.  Though I suppose introducing multiple threads really does complicate things

losfair:
@MarkMcCaskey 

> We can just cast the pointer from `UnsafeCell<T>` to `T` because of #[repr(transparent)].

My understanding is that as long as `#[repr(transparent)]` isn't removed from libstd definition of `UnsafeCell`, casting should always be safe. Although the repr attribute is theoretically an implementation detail, it is unlikely to be removed any time soon.

The [Rust nomicon book](https://doc.rust-lang.org/nomicon/other-reprs.html) also states that:

> (repr(transparent)) The goal is to make it possible to transmute between the single field and the struct. An example of that is UnsafeCell, which can be transmuted into the type it wraps.

MarkMcCaskey:
@losfair that makes sense to me -- thanks for the nomicon link!

I think this is fine then, though I still have a preference for this code as it gets a mutable slice which I think is more correct given that we're giving mutable access to the data through `Cell`s.  We can probably also do the cast or transmute with a mutable slice too though.

MarkMcCaskey:
@losfair good call, I just merged master and compiled both with `codegen-units = 1`.

The two binaries are identical in release mode!

PR #1211, opened 2020-02-12
fix(runtime-c-api) Avoid casting a non-mut to mut pointer.

Hywan:
Depends on #1206.

Use `slice::as_mut_ptr` to get a mutable pointer. Then casting `*mut
Cell<u8>` to `mut u8` is safe because identical bitwise.

This patch avoids casting a non-mutable pointer `slice::as_ptr` to a
mutable pointer `*mut Cell<u8>` to `*mut u8`.


MarkMcCaskey:
Closing this because #1206 shouldn't be shipped as-is (aliasing `&mut` is not the direction our API should go!)

PR #1212, opened 2020-02-13
Add support for GDB JIT debugging

MarkMcCaskey:
This PR adds support for JIT debugging to Wasmer with the Cranelift backend using a fork of `wasmtime-debug`.

The motivation for this change is partially inspired by the feature in Wasmtime and the implementation is largely derived from Wasmtime's `wasmtime-debug` crate and not included in this PR.  This implementation is currently Cranelift-only (if LLVM has value tracking we can add this there too without too much effort; we'd have to do the value tracking ourselves in Singlepass and I don't have enough context to know how hard that would be) and is based on a generic fork of the `wasmtime-debug` -- which will be published and uploaded in another repo.

This PR started out implementing the [Wasm-DWARF](https://yurydelendik.github.io/webassembly-dwarf/) reading and writing with gimli but after working on it for a few days, reading a chunk of the DWARF spec, seeing that Wasmtime had solved this well, and realizing how long this would likely take, I decided that it didn't make sense to spend the engineering effort there so I made a copy of `wasmtime-debug` and removed some of the less portable Cranelift pieces (very minor changes) and all code relying on data structures from wasmtime.  The resulting crate is completely generic and would work fine with Wasmtime or any other Wasm runtime at the cost of requiring some `transmute`s or a linear pass over the debug data to reconstruct it in terms of the new types exposed by the fork.  Perhaps there's a cleaner way to handle that that I haven't considered.

The integration with the GDB JIT interface is from the LLVM examples (I don't remember if I properly attributed everything in this PR/version of the code -- I still have the other branches locally though which I'll review before shipping this) and some of the code in this PR is from Wasmtime/Cranelift source code such as the sorting of the `ebb`s in `clif_backend::resolver`.  I spent a long time debugging some subtle bugs and ended up using a few things from Wasmtime's integration with `wasmtime-debug` and some bits from `cranelift-wasm` and `cranelift-codegen`.

If there's interest from other people in working on the generic `wasmtime-debug` fork, I'm happy to get other maintainers involved and/or move it to a shared organization.

Special thanks to [Yury Delendik](https://github.com/yurydelendik) and the other `wasmtime-debug` authors for their work on Wasm debugging.  Also shout out to Cranelift for the nice API for tracking variables/data.

### TODO:
- [x] Update attributions file for LLVM, [wasm-dwarf](https://github.com/yurydelendik/wasm-dwarf), and Wasmtime/Cranelift and do another pass over code to make sure we're in compliance with the licenses from the relevant projects and have properly attributed the code used from other projects.
- [x] Adjust API of wasm-debug based on feedback
- [x] Discuss with Nick integration with LLVM
- [x] Discuss with Heyang integration with Singlepass
- [x] Adjust implementation based on feedback from team (traits modified, etc.)
- [x] Clean up some pointer wrangling code
- [x] Add opt-in feature to wasmer-runtime-core to enale wasm-debug so library users who won't use debug info are not affected

# Review


lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+    first_entry: *mut JITCodeEntry,
+}
+
+/// Global variable that the GDB JIT interface will read the data from.
+/// The data is in the form of a doubly linked list.  This global variable acts
+/// as a head node with extra information about the operation that we want the
+/// debugger to perform.
+#[no_mangle]
+#[allow(non_upper_case_globals)]
+pub(crate) static mut __jit_debug_descriptor: JitDebugDescriptor = JitDebugDescriptor {
```
nlewycky: Is the repeated type mandatory here?

lib/runtime-core/src/structures/map.rs
```diff
@@ -78,6 +78,11 @@ where
     pub fn into_vec(self) -> Vec<V> {
         self.elems
     }
+
+    /// Iterate over the values of the map in order
```
nlewycky:
```suggestion
    /// Iterate over the values of the map in order.
```

lib/runtime-core/src/vm.rs
```diff
@@ -134,6 +134,9 @@ pub struct InternalCtx {
 
     /// Interrupt signal mem.
     pub interrupt_signal_mem: *mut u8,
+
+    /// hmm
```
nlewycky: hmm?

lib/runtime-core/src/vm.rs
```diff
@@ -336,6 +340,8 @@ impl Ctx {
                 };
                 ((*mem).base, (*mem).bound)
             };
+
+        dbg!(mem_bound);
```
nlewycky: dbg code

lib/runtime-core/src/instance.rs
```diff
@@ -79,12 +79,13 @@ impl Instance {
...
             let backing = &mut *(&mut inner.backing as *mut _);
             let import_backing = &mut *(&mut inner.import_backing as *mut _);
-            let real_ctx = match imports.call_state_creator() {
+            let mut real_ctx = match imports.call_state_creator() {
                 Some((data, dtor)) => {
                     vm::Ctx::new_with_data(backing, import_backing, &module, data, dtor)
                 }
                 None => vm::Ctx::new(backing, import_backing, &module),
             };
+            real_ctx.internal.ctx = vmctx.as_mut_ptr();
```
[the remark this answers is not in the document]
MarkMcCaskey: Oops, that's debug code, should be removed!

lib/runtime-core/src/jit_debug.rs
MarkMcCaskey: Yeah, statics have to have labeled types

MarkMcCaskey: Or at least global statics do

lib/runtime-core/src/vm.rs
MarkMcCaskey: Oh that should be removed!

lib/runtime-core/src/codegen.rs
```diff
@@ -112,14 +112,45 @@ pub trait ModuleCodeGenerator<FCG: FunctionCodeGenerator<E>, RM: RunnableModule,
...
+        &mut self,
+        module_info: Arc<RwLock<ModuleInfo>>,
+        loc: (u32, u32),
+    ) -> Result<&mut FCG, E>;
     /// Finalizes this module.
-    fn finalize(self, module_info: &ModuleInfo) -> Result<(RM, Box<dyn CacheGen>), E>;
+    fn finalize(
+        self,
+        module_info: &ModuleInfo,
+    ) -> Result<((RM, Option<DebugMetadata>), Box<dyn CacheGen>), E>;
```
syrusakbary: Why did you do that?

Shouldn't be simpler a single tuple?

```suggestion
    ) -> Result<(RM, Option<DebugMetadata>, Box<dyn CacheGen>), E>;
```

lib/runtime-core/src/module.rs
```diff
@@ -77,6 +78,14 @@ pub struct ModuleInfo {
...
     /// Custom sections.
     pub custom_sections: HashMap<String, Vec<u8>>,
+
+    /// Flag controlling whether or not debug information for use in a debugger
+    /// will be generated.
+    pub generate_debug_info: bool,
+
+    #[serde(skip)]
+    /// Resource manager of debug information being used by a debugger.
+    pub debug_info_manager: jit_debug::JITCodeDebugInfoManager,
```
syrusakbary: Shouldn't this fields be optional depending on compilation features?

lib/runtime-core/src/codegen.rs
```diff
@@ -296,6 +364,7 @@ impl MiddlewareChain {
         fcg: Option<&mut FCG>,
         ev: Event,
         module_info: &ModuleInfo,
+        loc: u32,
```
syrusakbary: Are we changing the public API? If so, is there any way to either:
* Have this added into the CHANGELOG as [BREAKING CHANGE]
* Minimize this? (not sure how, we will have to research)

lib/runtime-core/src/module.rs
MarkMcCaskey: Fair enough, I can make that happen and also exclude the jit_debug module from runtime-core, too.

Will need to double check that this shows up on `docs.rs` though because it's not a default feature and ideally it'd still be documented

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+    unsafe {
+        std::ptr::read_volatile(&x);
+    }
+}
+
+/// The operation that the debugger should perform with the entry that we gave it.
+#[allow(non_camel_case_types)]
+#[derive(Debug)]
+#[repr(u32)]
+pub(crate) enum JITAction {
```
Hywan:
```suggestion
pub(crate) enum JitAction {
```

and thus, you can remove the `#[allow(non_camel_case_types)]` directive.

Or… it should be `#[no_mangle]`d?

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// Unregister the given code.
+    JIT_UNREGISTER_FN = 2,
+}
+
+/// Node of the doubly linked list that the GDB JIT interface reads from.
+#[no_mangle]
+#[repr(C)]
+pub(crate) struct JITCodeEntry {
+    /// Next entry in the linked list.
+    next: *mut JITCodeEntry,
```
Hywan:
```suggestion
    next: *mut Self,
```

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// Previous entry in the linked list.
+    prev: *mut JITCodeEntry,
```
Hywan:
```suggestion
    prev: *mut Self,
```

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// Pointer to the data we want the debugger to read.
+    symfile_addr: *mut u8,
```
Hywan:
```suggestion
    symfile_addr: *const c_char,
```

`c_char` is from `std::os::raw::c_char`. I know it's is equivalent to `u8` but I think it's best to use the C equivalent definitions since it targets or mimics C.

`*const` because the C signature is `const char* symfile_addr` (as defined in [JIT Declarations](https://sourceware.org/gdb/current/onlinedocs/gdb.html#Declarations)).

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+            symfile_size: 0,
+        }
+    }
+}
+
+/// Head node of the doubly linked list that the GDB JIT interface expects.
+#[no_mangle]
+#[repr(C)]
+pub(crate) struct JitDebugDescriptor {
+    /// The version of the JIT interface to use (presumably, TODO: double check this)
```
Hywan: I double-checked, and it looks good to me.

```suggestion
    /// The version of the JIT interface to use.
```

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+    version: u32,
+    /// Which action to perform, see [`JITAction`].
+    action_flag: u32,
```
Hywan: Since `JITAction` has a `u32` representation, I think it could be safer (from a type point of view) to write:

```suggestion
    action_flag: JITAction,
```

and it would still be valid for C I guess.

Thoughts?

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// The entry in the list that the `action_flag` applies to.
+    relevant_entry: *mut JITCodeEntry,
+    /// The first entry in the doubly linked list.
+    first_entry: *mut JITCodeEntry,
+}
+
+/// Global variable that the GDB JIT interface will read the data from.
+/// The data is in the form of a doubly linked list.  This global variable acts
```
Hywan:
```suggestion
/// The data is in the form of a doubly linked list. This global variable acts
```

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+/// as a head node with extra information about the operation that we want the
+/// debugger to perform.
+#[no_mangle]
+#[allow(non_upper_case_globals)]
+pub(crate) static mut __jit_debug_descriptor: JitDebugDescriptor = JitDebugDescriptor {
+    version: 1,
+    action_flag: JITAction::JIT_NOACTION as _,
```
Hywan: Even better than the official documentation, good job 😉!

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+    relevant_entry: ptr::null_mut(),
+    first_entry: ptr::null_mut(),
+};
+
+/// Prepend an item to the front of the `__jit_debug_descriptor` entry list
+///
+/// # Safety
+/// - Access to underlying global variable is unsynchronized.
```
Hywan: It means the JIT engine must not be multi-threaded, is that correct? Do we have a way to detect that (for what I looked for, no)?

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+/// - Pointer to [`JITCodeEntry`] should point to a valid entry.
+unsafe fn push_front(jce: *mut JITCodeEntry) {
+    if __jit_debug_descriptor.first_entry.is_null() {
+        __jit_debug_descriptor.first_entry = jce;
+    } else {
+        let old_first = __jit_debug_descriptor.first_entry;
+        debug_assert!((*old_first).prev.is_null());
```
Hywan: I would turn `debug_assert!` into `assert!` here considering how important it is. Thoughts?

```suggestion
        assert!((*old_first).prev.is_null());
```

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+        __jit_debug_descriptor.first_entry = jce;
+    }
+}
+
+/// Removes an entry from the doubly linked list, updating both nodes that it's
+/// connected to.
+///
+/// # Safety
+/// - Access to underlying global variable is unsynchronized.
+/// - Pointer must point to a valid `JitCodeEntry`.
```
Hywan:
```suggestion
/// - Pointer must point to a valid `JITCodeEntry`.
```

or

```suggestion
/// - Pointer to [`JITCodeEntry`] must point to a valid entry.
```

(to be consistent with `push_front`).

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+    action_flag: JITAction::JIT_NOACTION as _,
+    relevant_entry: ptr::null_mut(),
+    first_entry: ptr::null_mut(),
+};
+
+/// Prepend an item to the front of the `__jit_debug_descriptor` entry list
+///
+/// # Safety
+/// - Access to underlying global variable is unsynchronized.
+/// - Pointer to [`JITCodeEntry`] should point to a valid entry.
```
Hywan:
```suggestion
/// - Pointer to [`JITCodeEntry`] must point to a valid entry.
```

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+
+/// Removes an entry from the doubly linked list, updating both nodes that it's
+/// connected to.
+///
+/// # Safety
+/// - Access to underlying global variable is unsynchronized.
+/// - Pointer must point to a valid `JitCodeEntry`.
+unsafe fn remove_node(jce: *mut JITCodeEntry) {
+    if __jit_debug_descriptor.first_entry == jce {
+        debug_assert!((*jce).prev.is_null());
```
Hywan: It's up to you to decide 🙂.

```suggestion
        assert!((*jce).prev.is_null());
```

lib/runtime-core/src/jit_debug.rs
Hywan: Are we sure we want `JITAction` to be `pub(crate)` too?

```suggestion
enum JITAction {
```

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// Register the given code.
+    JIT_REGISTER_FN = 1,
+    /// Unregister the given code.
+    JIT_UNREGISTER_FN = 2,
+}
+
+/// Node of the doubly linked list that the GDB JIT interface reads from.
+#[no_mangle]
+#[repr(C)]
+pub(crate) struct JITCodeEntry {
```
Hywan: Are we sure we want `JITCodeEntry` to be public in the crate? I don't see any reason. Am I missing something?

```suggestion
struct JITCodeEntry {
```

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+            symfile_addr: ptr::null_mut(),
+            symfile_size: 0,
+        }
+    }
+}
+
+/// Head node of the doubly linked list that the GDB JIT interface expects.
+#[no_mangle]
+#[repr(C)]
+pub(crate) struct JitDebugDescriptor {
```
Hywan: Are we sure we want `JitDebugDescriptor` to be public in the crate? I don't see any reason. Am I missing something?

```suggestion
struct JitDebugDescriptor {
```

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+    first_entry: *mut JITCodeEntry,
+}
+
+/// Global variable that the GDB JIT interface will read the data from.
+/// The data is in the form of a doubly linked list.  This global variable acts
+/// as a head node with extra information about the operation that we want the
+/// debugger to perform.
+#[no_mangle]
+#[allow(non_upper_case_globals)]
+pub(crate) static mut __jit_debug_descriptor: JitDebugDescriptor = JitDebugDescriptor {
```
Hywan: Are we sure we want `__jit_debug_descriptor` to be public in the crate? I don't see any reason. Am I missing something?

```suggestion
static mut __jit_debug_descriptor: JitDebugDescriptor = JitDebugDescriptor {
```

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,195 @@
...
+                entry.symfile_size as _,
+                entry.symfile_size as _,
+            );
+        }
+    }
+}
+
+/// Manager of debug info registered with the debugger.
+#[derive(Debug, Clone, Default)]
+pub struct JITCodeDebugInfoManager {
```
Hywan:
```suggestion
pub(crate) struct JITCodeDebugInfoManager {
```

since there is only one field, which is private, and one method, which is visible to the crate only.

lib/runtime-core/src/module.rs
```diff
@@ -77,6 +79,15 @@ pub struct ModuleInfo {
...
     pub custom_sections: HashMap<String, Vec<u8>>,
+
+    /// Flag controlling whether or not debug information for use in a debugger
+    /// will be generated.
+    pub generate_debug_info: bool,
+
+    #[cfg(feature = "generate-debug-information")]
+    #[serde(skip)]
+    /// Resource manager of debug information being used by a debugger.
+    pub debug_info_manager: jit_debug::JITCodeDebugInfoManager,
```
Hywan:
```suggestion
    pub(crate) debug_info_manager: jit_debug::JITCodeDebugInfoManager,
```

related to my comment above.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - [#1233](https://github.com/wasmerio/wasmer/pull/1233) Improved Wasmer C API release artifacts.
 - [#1216](https://github.com/wasmerio/wasmer/pull/1216) `wasmer-interface-types` receives a binary encoder.
+- [#1212](https://github.com/wasmerio/wasmer/pull/1212) Add `--generate-debug-info` and `-g` flags to `wasmer run` to generate debug information during compilation that is passed via the GDB JIT interface to a debugger to allow source-level debugging of Wasm files.  Currently only available on clif-backend, see PR for more information on its implementation.
```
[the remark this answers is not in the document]
MarkMcCaskey: I'm a big fan of two spaces after periods.  I try to suppress it when writing in a shared space because inconsistent style is worse, but I find it much more readable, it chunks the sentences in my brain better and helps me read more quickly.

From Googling it, using two spaces after a period apparently means I'm over 40 years old and stuck to my ways using the typewriter 😆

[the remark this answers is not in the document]
MarkMcCaskey: there's quite a few instances of `\.  ` in CHANGELOG.md haha, probably all from me

lib/clif-backend/src/resolver.rs
```diff
@@ -96,59 +96,169 @@ impl FuncResolverBuilder {
...
-                    },
-                )
-                .collect();
+        let generate_debug_info = info.generate_debug_info;
+        let fb = function_bodies.iter().collect::<Vec<(_, _)>>();
+
+        /// Data about the the compiled machine code.
+        type CompileMetadata = (
+            LocalFuncIndex,
+            Option<(CompiledFunctionData, ValueLabelsRanges, Vec<Option<i32>>)>,
```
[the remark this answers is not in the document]
MarkMcCaskey: Uh, it came from the stackslot type that `wasm-debug` used to depend on.  The offset it returns is a number.  I believe they're usually negative numbers actually because they're in terms of RBP and not RSI 🤔

lib/runtime-core/src/jit_debug.rs
MarkMcCaskey: Yeah, that's fair, I believe I changed it from `const` to `mut` to indicate that the Rust code was allowed to retake ownership of it and destroy it and so that Rust knew that the pointer could be mutated.  I believe casting `*mut T` to a `*const T` (and then back to `*mut T` when we destroy it) is acceptable by Rust and won't trigger UB, so I can do that

lib/runtime-core/src/jit_debug.rs
MarkMcCaskey: Yeah, the one concern I have is that, as a global variable exported to FFI, if it gets mutated then that triggers undefined behavior.  I think in this case it's probably fine though as it shouldn't be mutated by the thing we're talking to with FFI

lib/runtime-core/src/jit_debug.rs
MarkMcCaskey: So right now, we only submit after code generation has finished so it's a non-issue.  It does raise the issue that there's a single global variable being shared between multiple instances potentially, but the use case for debug information is for debugging and I think things like breakpoints are per-thread.

Having a global mutex here is the right thing to do though.

lib/runtime-core/src/jit_debug.rs
MarkMcCaskey: Sure, I'm fine with that!

lib/runtime-core/src/jit_debug.rs
MarkMcCaskey: It's part of module which has all public fields.  I wasn't sure if I should make the first private field in the struct or just make it public.  Making it public seemed like the safer choice

lib/runtime-core/src/module.rs
MarkMcCaskey: I was a bit worried about the implications of making the first non-public field, but I suppose it'll be fine

MarkMcCaskey:
bors try

lib/runtime-core/src/parse.rs
```diff
@@ -206,88 +210,107 @@ pub fn read_module<
...
+                    for param in sig.params() {
+                        fcg.feed_param(type_to_wp_type(*param))
+                            .map_err(|x| LoadError::Codegen(format!("{:?}", x)))?;
+                    }
                 }
-                for param in sig.params() {
-                    fcg.feed_param(type_to_wp_type(*param))
-                        .map_err(|x| LoadError::Codegen(format!("{:?}", x)))?;
+
+                let mut operator_parser = parser.create_binary_reader();
```
losfair: `create_binary_reader` returns a raw reader without validation: https://docs.rs/wasmparser/0.51.2/src/wasmparser/validator.rs.html#721-729 , but backends expect a pre-validated stream of events.

MarkMcCaskey: Good catch, the reason is that we need the local info too and can't create a validating operator parser if we're not on `FunctionBegin` -- although, it seems like tests should definitely be failing if we disabled validation on the insides of function bodies...

If we don't have tests for that yet, we absolutely should

lib/runtime-core/src/jit_debug.rs
```diff
@@ -0,0 +1,214 @@
...
+/// control to read the data we prepared.
+// Implementation of this function is derived from wasmtime and is licensed under
+// the Apache 2.0 license.  See ATTRIBUTIONS.md for full license and more
+// information.
+#[no_mangle]
+#[inline(never)]
+extern "C" fn __jit_debug_register_code() {
+    // This code exists to prevent optimization of this function so that the
+    // GDB JIT interface behaves as expected
+    let x = 42;
```
syrusakbary: What is this number?

lib/clif-backend/src/code.rs
```diff
@@ -1091,7 +1108,12 @@ impl FunctionCodeGenerator<CodegenError> for CraneliftFunctionCodeGenerator {
         Ok(())
     }
 
-    fn feed_event(&mut self, event: Event, _module_info: &ModuleInfo) -> Result<(), CodegenError> {
+    fn feed_event(
+        &mut self,
+        event: Event,
+        _module_info: &ModuleInfo,
+        loc: u32,
```
syrusakbary: Can we rename this to `source_loc`, I know it's more verbose but it's also clearer:

```suggestion
        source_loc: u32,
```

lib/runtime-core/src/jit_debug.rs
nlewycky: https://en.wikipedia.org/wiki/42_(number)#The_Hitchhiker's_Guide_to_the_Galaxy

MarkMcCaskey: placeholder value, doesn't do anything.  We need something to force a volatile read on

lib/runtime-core/src/parse.rs
MarkMcCaskey: I filed an issue on WasmParser to unblock this!

MarkMcCaskey: Okay, this should be resolved now!

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## **[Unreleased]**
 
+- [#1212](https://github.com/wasmerio/wasmer/pull/1212) [BREAKING CHANGE] Add `--generate-debug-info` and `-g` flags to `wasmer run` to generate debug information during compilation that is passed via the GDB JIT interface to a debugger to allow source-level debugging of Wasm files. Currently only available on clif-backend. As part of this change, the public middleware API is broken, there is now a `source_loc` parameter that should be passed through if applicable. Additionally there are breaking changes on the compiler trait functions such as `feed_local`, `feed_event` as well as `ModuleCodeGenerator::finalize`.
```
syrusakbary:
```suggestion
- [#1212](https://github.com/wasmerio/wasmer/pull/1212) [BREAKING CHANGE on internal APIs] Add `--generate-debug-info` and `-g` flags to `wasmer run` to generate debug information during compilation that is passed via the GDB JIT interface to a debugger to allow source-level debugging of Wasm files. Currently only available on clif-backend. As part of this change, the public middleware API has a breaking change, there is now a `source_loc` parameter that should be passed through if applicable. Additionally there are breaking changes on the compiler trait functions such as `feed_local`, `feed_event` as well as `ModuleCodeGenerator::finalize`.
```

MarkMcCaskey: > BREAKING CHANGE on internal APIs

Do you mean external APIs?

syrusakbary: I wanted to remove some alarmism from it.
We can do `BREAKING CHANGE on Middleware and internal APIs`

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## **[Unreleased]**
 
+- [#1212](https://github.com/wasmerio/wasmer/pull/1212) [BREAKING CHANGE to public APIs] Add support for GDB JIT debugging:
```
syrusakbary:
```suggestion
- [#1212](https://github.com/wasmerio/wasmer/pull/1212) Add support for GDB JIT debugging:
```

webmaster128:
How should I interprete the new `source_loc`? Is it the 0-based index of the wasm blob in bytes?

We're implementing a

```rust
impl FunctionMiddleware for DeterministicMiddleware {
    type Error = CompileError;

    fn feed_event<'a, 'b: 'a>(
        &mut self,
        event: Event<'a, 'b>,
        _module_info: &ModuleInfo,
        sink: &mut EventSink<'a, 'b>,
        source_loc: u32,
    ) -> Result<(), Self::Error> {
        match event {
            Event::Wasm(op) => parse_wasm_opcode(op)?,
            Event::WasmOwned(ref op) => parse_wasm_opcode(op)?,
            _ => (),
        };

        sink.push(event);
        Ok(())
    }
}
```

and could pass the new value into the error message if that makes sense.

MarkMcCaskey:
@webmaster128 `source_loc`s should be forwarded without modification for most middleware.  Yeah, `source_loc`s are offsets into the Wasm binary from the start of the code section if I remember correctly -- though I don't know if that's subject to change in the future.  From the perspective of most of the code touching them, they can just be treated as a black-box.  You could use them in error messages but the semantics of the number may break in the future if the Wasm DWARF standard changes, so I would recommend not relying on it for anything critical if you can avoid it!

webmaster128:
Thanks a lot, @MarkMcCaskey, that's very helpful. I'll ignore the value then. I've no place to forward it to.

MarkMcCaskey:
Sure!  We'll update the CHANGELOG.md with any breaking changes, so if the Wasm DWARF standard does change, we'd document it there.  So if you check that when updating Wasmer, you shouldn't have any bad surprises like the meaning of `source_loc` changing (if you do rely on it)!

ouiliame:
Hi, was this feature removed? I can't seem to trace why it was deprecated, was it due to lack of maintenance? Seems like a pretty important feature to leave out.

ptitSeb:
It was removed, but that looks like a nice feature, so I'll check if it can be added back.

PR #1217, opened 2020-02-14
Polymorphic host functions based on dynamic trampoline generation.

losfair:
This PR implements polymorphic host functions by dynamically generating the "glue" code that translates platform arguments to an array in `runtime-core`.

TODO:

- ~~Multiple return values.~~ Deferring to a future multivalue PR.
- [x] Dynamic signatures for polymorphic functions.
- [x] Use a proper executable memory allocator.


losfair:
bors try

lib/runtime-core-tests/tests/imports.rs
```diff
@@ -142,6 +154,16 @@ fn imported_functions_forms(test: &dyn Fn(&Instance)) {
                 Ok(n + 1)
             }),
 
+            "callback_closure_polymorphic" => Func::<i32, i32, _>::new_polymorphic(
+                Arc::new(FuncSig::new(vec![Type::I32], vec![Type::I32])),
+                |_, params| -> Vec<Value> {
```
Hywan: Is `_` the `vm::Ctx`?

lib/runtime-core-tests/tests/imports.rs
```diff
@@ -142,6 +154,16 @@ fn imported_functions_forms(test: &dyn Fn(&Instance)) {
                 Ok(n + 1)
             }),
 
+            "callback_closure_polymorphic" => Func::<i32, i32, _>::new_polymorphic(
```
Hywan: The fact that we need to statically write the types here (`Func::<i32, i32, _>`) makes it unsable in a dynamically typed language such as Python, Ruby or PHP. In #1018, I found a workaround for that, maybe we can duplicate it here. I'll try to port it here and see how it fits.

lib/runtime-core-tests/tests/imports.rs
losfair: Yep

Hywan:
Please, check if https://github.com/wasmerio/wasmer/pull/1225 can be included in this PR.

lib/runtime-core-tests/tests/imports.rs
Hywan: Done in #1225.

losfair:
bors try

losfair:
bors try

lib/runtime-core/src/typed_func.rs
```diff
@@ -246,6 +297,97 @@ where
     }
 }
 
+impl<'a> ErasedFunc<'a> {
+    /// Creates a polymorphic function.
+    #[allow(unused_variables)]
+    #[cfg(all(unix, target_arch = "x86_64"))]
```
syrusakbary: Can we have another implementation of `new_polymorphic` that is for not `x86_64`?

`#[cfg(not(all(unix, target_arch = "x86_64")))]` that raises a `compile_error!("Polymorphic functions are not implemented for this architecture")`

lib/runtime-core/src/typed_func.rs
```diff
@@ -246,6 +297,97 @@ where
...
+    pub fn new_polymorphic<F>(signature: Arc<FuncSig>, func: F) -> Self
+    where
+        F: Fn(&mut vm::Ctx, &[crate::types::Value]) -> Vec<crate::types::Value> + 'static,
+    {
+        use crate::trampoline_x64::*;
```
syrusakbary: Rather than star imports, can we be explicit on imports (I was having some trouble trying to see where `CallContext` was coming from)

lib/runtime-core/src/typed_func.rs
losfair: Making it a compile error will prevent `runtime_core` to build on non-`x86_64` architectures, while making it a runtime error doesn't make a lot of sense since external code should be aware of the unsupported status at compile time. Therefore I think leaving it "missing" for now would be fine.

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -29,6 +32,95 @@ lazy_static! {
...
+
+/// The allocation state of a `TrampBuffer`.
+struct AllocState {
+    /// Records all allocated blocks in `buffer`.
+    blocks: BTreeMap<usize, usize>,
+}
+
+impl TrampBuffer {
+    /// Creates a trampoline buffer with a given (virtual) size.
+    pub(self) fn new(size: usize) -> TrampBuffer {
```
Hywan:
```suggestion
    fn new(size: usize) -> TrampBuffer {
```

This is strictly the same, and avoids confusion.

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -29,6 +32,95 @@ lazy_static! {
...
+        TrampBuffer {
+            buffer: mem,
+            alloc: Mutex::new(AllocState {
+                blocks: BTreeMap::new(),
+            }),
+        }
+    }
+
+    /// Removes a previously-`insert`ed trampoline.
+    pub(self) unsafe fn remove(&self, start: NonNull<u8>) {
```
Hywan: Can you add a `# Safety` section to explain the `unsafe` keyword please?

Also, as above:

```suggestion
    unsafe fn remove(&self, start: NonNull<u8>) {
```

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -292,4 +399,85 @@ mod tests {
...
+                            >(ptr);
+                            assert_eq!(f(1, 2, 3, 4, 5, 6, 7), 28);
+                        }
+                        _ => unreachable!(),
+                    }
+                    buffer.remove(ptr);
+                }
+            }
+
+            let expected = (0..=n as u64).fold(0u64, |a, b| a + b);
```
Hywan:
```suggestion
            let expected = (0..=n as u64).sum();
```

lib/runtime-core/src/typed_func.rs
```diff
@@ -190,18 +190,69 @@ where
     }
 }
 
+/// Represents a type-erased function provided by either the host or the WebAssembly program.
+#[allow(dead_code)]
```
Hywan: Why `dead_code` is allowed?

lib/runtime-core/src/typed_func.rs
```diff
@@ -246,6 +297,97 @@ where
...
+                .enumerate()
+                .map(|(i, t)| {
+                    let i = i + 1; // skip vmctx
+                    match *t {
+                        Type::I32 => Value::I32(*args.offset(i as _) as i32),
+                        Type::I64 => Value::I64(*args.offset(i as _) as i64),
+                        Type::F32 => Value::F32(f32::from_bits(*args.offset(i as _) as u32)),
+                        Type::F64 => Value::F64(f64::from_bits(*args.offset(i as _) as u64)),
+                        Type::V128 => {
+                            panic!("enter_host_polymorphic: 128-bit types are not supported")
```
Hywan:
```suggestion
                            todo!("enter_host_polymorphic: 128-bit types are not supported")
```

maybe?

lib/runtime-core/src/trampoline_x64.rs
losfair: Thanks and fixed. I thought `pub(self)` is needed to use these functions in tests.

losfair:
bors try

lib/runtime-core-tests/tests/imports.rs
```diff
@@ -142,6 +154,16 @@ fn imported_functions_forms(test: &dyn Fn(&Instance)) {
                 Ok(n + 1)
             }),
 
+            "callback_closure_polymorphic" => ErasedFunc::new_polymorphic(
```
syrusakbary: After offline chatting I think other name like `DynamicFunc::new` would make the API a bit clearer for the user.

This will imply:
* Renaming `ErasedFunc` to `DynamicFunc`
* Renaming the `new_polymorphic` method to just `new`

Thoughts @losfair ?

lib/runtime-core/src/typed_func.rs
```diff
@@ -246,6 +296,97 @@ where
...
+    #[allow(unused_variables)]
+    #[cfg(all(unix, target_arch = "x86_64"))]
+    pub fn new_polymorphic<F>(signature: Arc<FuncSig>, func: F) -> Self
+    where
+        F: Fn(&mut vm::Ctx, &[crate::types::Value]) -> Vec<crate::types::Value> + 'static,
+    {
+        use crate::trampoline_x64::{CallContext, TrampolineBufferBuilder};
+        use crate::types::Value;
+        use std::convert::TryFrom;
+
```
syrusakbary: Can we also enforce that `F` is a function pointer or a closure _without_ a captured environment? (meaning: not allowing closure with environments for now)

Something similar to what we are doing in the other `to_raw` functions:

```
                    // `FN` is a function pointer, or a closure
                    // _without_ a captured environment.
                    if mem::size_of::<Self>() == 0 {
                        NonNull::new(&self as *const _ as *mut vm::FuncEnv)
                    }
                    else {
                        unimplemented!("DynamicFunc with captured closures are not implemented");
                    }
```

lib/runtime-core-tests/tests/imports.rs
syrusakbary:
```suggestion
            "callback_closure_polymorphic" => DynamicFunc::new(
```

lib/runtime-core/src/typed_func.rs
losfair: It is not needed for dynamic/polymorphic functions since proper fat pointer `Box<Fn>` calls and drops is implemented: https://github.com/wasmerio/wasmer/blob/31a72e59fbd96263277a5827b3576351b6698ed0/lib/runtime-core/src/typed_func.rs#L313

losfair: After some internal discussion we decided to temporarily disallow closures with captured environment so that more future changes are possible without breaking the public API.

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -29,6 +32,97 @@ lazy_static! {
...
+/// The allocation state of a `TrampBuffer`.
+struct AllocState {
+    /// Records all allocated blocks in `buffer`.
+    blocks: BTreeMap<usize, usize>,
+}
+
+impl TrampBuffer {
+    /// Creates a trampoline buffer with a given (virtual) size.
+    fn new(size: usize) -> TrampBuffer {
+        // Pre-allocate 64 MiB of virtual memory for code.
```
nlewycky: 64 MiB? Isn't this value controlled by the `size` parameter?

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -29,6 +32,97 @@ lazy_static! {
...
+    ///
+    /// Returns `None` if no memory is available.
+    fn insert(&self, buf: &[u8]) -> Option<NonNull<u8>> {
+        // First, assume an available start position...
+        let mut assumed_start: usize = 0;
+
+        let mut alloc = self.alloc.lock().unwrap();
+        let mut found = false;
+
+        // Then, try invalidating that assumption...
```
nlewycky: How common is freeing blocks? And inserting new ones?

Should we instead store a single allocated region, plus a list of "free" regions, keeping consecutive free regions coalesced? This loop would then try to allocate out of the free list. We could even keep the free list sorted by size instead of start location, so that we can quickly find the smallest block that will fit?

Should we even bother reusing freed memory here at all? If I understand right, we produce at most one trampoline per imported function, which is a finite number for a given wasm instance? We could have one of these per instance or module, then just allocate off the end only and delete the whole thing when the instance is deleted?

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -29,6 +32,97 @@ lazy_static! {
...
+                // This free block can be used.
+                found = true;
+                break;
+            }
+        }
+
+        if !found {
+            // No previous free blocks were found. Try allocating at the end.
+            if self.buffer.len() - assumed_start < buf.len() {
+                // No more free space. Cannot allocate.
```
nlewycky: `mremap`? Or is the problem that we have live pointers into this space and we can't deal with the memory moving.

We could store a list of buffers?

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -29,6 +32,97 @@ lazy_static! {
...
+                return None;
+            } else {
```
nlewycky: The empty else is unusual. I'd suggest just moving the comment after the `!found` and saying "We have enough space to extend towards the end." or something like that?

lib/runtime-core/src/typed_func.rs
```diff
@@ -190,18 +190,68 @@ where
...
 pub struct Func<'a, Args = (), Rets = (), Inner: Kind = Wasm> {
     inner: Inner,
+
+    /// The function pointer.
     func: NonNull<vm::Func>,
+
+    /// The function environment.
     func_env: Option<NonNull<vm::FuncEnv>>,
+
+    /// The famous `vm::Ctx`.
```
nlewycky: :smile:

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## 0.14.1 - 2020-02-24
 
+- [#1217](https://github.com/wasmerio/wasmer/pull/1217) Polymorphic host functions based on dynamic trampoline generation.
```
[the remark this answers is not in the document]
syrusakbary: Good catch!

lib/runtime-core/src/trampoline_x64.rs
losfair: Yeah the free region method would be the ideal way to track dynamic allocation, as in common memory allocators. Here I just kept track of used blocks for simplicity, maybe it's good to change this.

The lifetime of `DynamicFunc` is unbounded and we cannot associate it to an instance safely at creation time. Or do you think the public API should be refactored and `DynamicFunc::new` should be made a method on `Instance`?

lib/runtime-core/src/trampoline_x64.rs
losfair: Yeah the memory cannot be moved because it will invalidate pointers.

Storing a list of buffers would work but why not just allocate a large amount of virtual memory and let the operating system manage their physical allocations? (hence `MAP_NORESERVE` is used) Some `madvise`s are missing though.

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -29,6 +32,97 @@ lazy_static! {
...
+    /// A fixed-(virtual)-size executable+writable buffer for storing trampolines.
+    buffer: CodeMemory,
+
+    /// Allocation state.
+    alloc: Mutex<AllocState>,
+}
+
+/// The allocation state of a `TrampBuffer`.
+struct AllocState {
+    /// Records all allocated blocks in `buffer`.
```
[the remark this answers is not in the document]
losfair: The start and end of all used blocks.

lib/runtime-core/src/loader.rs
```diff
@@ -169,7 +171,7 @@ impl CodeMemory {
                 std::ptr::null_mut(),
                 size,
                 PROT_READ | PROT_WRITE,
-                MAP_PRIVATE | MAP_ANON,
+                MAP_PRIVATE | MAP_ANON | MAP_NORESERVE,
```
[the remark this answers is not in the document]
losfair: Prevent the OS from pre-allocating swap space and failing the `mmap` if neither swap nor real memory is available. Useful if we want to do something like allocating a large amount of virtual memory without physical backing (which is the trampoline case).

lib/runtime-core/src/trampoline_x64.rs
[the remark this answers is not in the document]
nlewycky: Thanks! Could you add that to the comment?

lib/runtime-core/src/trampoline_x64.rs
nlewycky: I can't imagine that we'll run out of 64MiB for trampolines, but if we do decide it's a problem we can extend it to store a list of slabs then.

lib/runtime-core/src/trampoline_x64.rs
nlewycky: I was just thinking it'd be faster to run until the end before trying to use up spaces in the middle. The normal case is that this search will return zero free spaces and we don't need to do a scan until we're actually out of memory.

Talked about `DynamicFunc::new` with Syrus and Mark, we agreed that putting them on `Instance` is the wrong way to go. Functions can exist outside of instances anyways, and we're currently failing a spectest over that.

lib/runtime-core/src/trampoline_x64.rs
nlewycky: Optional: I still feel weird about this else. Both because it's an empty-else (*every* if without an else has an empty-else!) and because it's an else after a return (you never need an else after a return statement!). The "extend towards the end" comment describes what we do after this `if !found { }` block anyways.

syrusakbary: Yup, agreed. Would be great to remove the empty else 👍

lib/runtime-core/src/trampoline_x64.rs
losfair: I just implemented it in a simple way so that we would not increase our code's size too much with "yet another allocator". For performance optimization, can we port over an external allocator like `wee_alloc`?

losfair: Merging this first and optimizations on the executable memory allocator can be made in another PR.

tessi:
thanks @losfair for this feature. This was the missing piece for me to implement support for imported functions in my elixir-wrapper. I was about to ask you for this feature, when I spotted it in the changelogs. thanks ❤️

PR #1223, opened 2020-02-17
feat(runtime-core) Implement `TryFrom<native_type>` for `Value`

Hywan:
Extracted from #1018.

This PR implements `TryFrom` for `Value`. It is required to support dynamically-typed values for polymorphic host functions.

This PR also refactors the code by using a macro (`value_conversions!`) to implement `From` and `TryFrom` in one shot.


Hywan:
Precisely, it can fail. For instance, let `v` be a `Value`. You want a `f32` but it's a `I32`. You will use `try_into()` and it might fail.

MarkMcCaskey:
> Precisely, it can fail. For instance, let `v` be a `Value`. You want a `f32` but it's a `I32`. You will use `try_into()` and it might fail.

Oh didn't realize it was bidirectional conversion, I thought it was the same conversion implemented with both `From` and `TryFrom`!

Looks good to me!

PR #1225, opened 2020-02-17
feat(runtime-core) Allow dynamic signature for polymorphic host functions

Hywan:
Extracted from #1018.

⚠️  This PR targets #1217, not `master`! ⚠️ 

---

This patch adds a new field in `Func`: `signature`. It contains the
signature of the host function.

For non-polymorphic host functions, the signature is computed from the
`Args` and `Rets` implementation parameters at compile-time.

For polymorphic host functions though, to be fully dynamic, the
signature given to `new_polymorphic` is used in `Func` as the correct
signature.

The improvement is the following:

```diff
- Func::<i32, i32, _>::new_polymorphic(…)
+ Func::new_polymorphic(…)
```

with this approach, we are more dynamic, and thus more easily embeddable in dynamic languages such as Python, Ruby, PHP and so on.

### DIFF
--- a/lib/runtime-core-tests/tests/imports.rs
+++ b/lib/runtime-core-tests/tests/imports.rs
@@ -154,7 +154,7 @@ fn imported_functions_forms(test: &dyn Fn(&Instance)) {
                 Ok(n + 1)
             }),
 
-            "callback_closure_polymorphic" => Func::<i32, i32, _>::new_polymorphic(
+            "callback_closure_polymorphic" => Func::new_polymorphic(
                 Arc::new(FuncSig::new(vec![Type::I32], vec![Type::I32])),
                 |_, params| -> Vec<Value> {
                     match params[0] {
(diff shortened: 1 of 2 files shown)


Hywan:
I'm afraid it will cause a regression. Let me think about it.

Hywan:
Yeah, in our doc, we have code like `let add_one: Func<u32, u32> = instance.func("add_one")?;`. This code will be broken.

losfair:
Yeah this is a breaking change.

Can we add something like `ErasedFunc` that provides a type-erased `Func` then (and maybe implement `From<Func<...>>` for `ErasedFunc` to keep the current API unchanged)? Having type parameters that do not always match the actual function type sounds confusing to me.

Hywan:
It does also sound confusing to me.
The `ErasedFunc` idea is good. I'll update the patch.

losfair:
Merging this first, will continue in https://github.com/wasmerio/wasmer/pull/1217 .

PR #1231, opened 2020-02-18
feature: calling an instance function by its export index

YaronWittenstein:
# Description
Adding `Instance#call_func_with_index` and `Instance#dyn_func_with_index` public methods.
This is useful for code that references a function by its export index. 

For example: in the blockchain world, every byte that can be saved from the chain is a blessing.
So instead of storing the function name inside the Smart-Contract transaction, we could store
its index.

# Review

- [x] Calling an instance function by its export index.
- [x] Retrieving a callable function by its export index.

### DIFF
--- a/lib/runtime-core/src/instance.rs
+++ b/lib/runtime-core/src/instance.rs
@@ -284,37 +284,45 @@ impl Instance {
     /// # }
     /// ```
     pub fn dyn_func(&self, name: &str) -> ResolveResult<DynFunc> {
-        let export_index =
-            self.module
-                .info
-                .exports
-                .get(name)
-                .ok_or_else(|| ResolveError::ExportNotFound {
-                    name: name.to_string(),
-                })?;
+        let func_index = self.resolve_func(name)?;
 
-        if let ExportIndex::Func(func_index) = export_index {
-            let sig_index = *self
-                .module
-                .info
-                .func_assoc
-                .get(*func_index)
-                .expect("broken invariant, incorrect func index");
-            let signature =
-                SigRegistry.lookup_signature_ref(&self.module.info.signatures[sig_index]);
+        self.dyn_func_with_index(func_index)
+    }
 
-            Ok(DynFunc {
-                signature,
-                module: &self.module,
-                instance_inner: &self.inner,
-                func_index: *func_index,
-            })
-        } else {
-            Err(ResolveError::ExportWrongType {
-                name: name.to_string(),
-            }
-            .into())
-        }
+    /// This returns the representation of a function that can be called
+    /// safely.
+    ///
+    /// # Usage:
+    /// ```
+    /// # use wasmer_runtime_core::Instance;
+    /// # use wasmer_runtime_core::error::CallResult;
+    /// # fn call_foo(instance: &mut Instance) -> CallResult<()> {
+    ///
+    /// let index = 0;
+    /// instance
+    ///     .dyn_func_with_index(index)?
+    ///     .call(&[])?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn dyn_func_with_index(&self, func_index: usize) -> ResolveResult<DynFunc> {
+        let func_index = FuncIndex::new(func_index);
+
+        let sig_index = *self
+            .module
+            .info
+            .func_assoc
+            .get(func_index)
+            .expect("broken invariant, incorrect func index");
+
+        let signature = SigRegistry.lookup_signature_ref(&self.module.info.signatures[sig_index]);
+
+        Ok(DynFunc {
+            signature,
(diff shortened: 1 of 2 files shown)


Hywan:
I understand the purpose of this, but using function index is… I don't find a good term except “unstable”. The function index can change from one module compilation to another.

I'm willing to accept the PR because I understand the use-case and it makes a lot of sense to me, but I would like insights from @syrusakbary for instance. Is it the kind of API we want to offer to the users?

YaronWittenstein:
> I understand the purpose of this, but using function index is… I don't find a good term except “unstable”. The function index can change from one module compilation to another.
> 
> I'm willing to accept the PR because I understand the use-case and it makes a lot of sense to me, but I would like insights from @syrusakbary for instance. Is it the kind of API we want to offer to the users?

I've assumed that the function is deterministic between different compilations.
so that the first export function has relative index 0, the second index 1 and so on.

To my understanding, the function index of the API takes into account the number of imported functions too. So if we import 5 functions in a module, then the absolute index of the first exported function will be
5 and so on. @Hywan @syrusakbary

Hywan:
What I'm saying is that if the program changes, the import indexes can change too, but that's OK. Blockchains have different needs, and I'm OK to merge this PR as is. Just wanted to double-check with @syrusakbary.

lib/runtime-core/src/instance.rs
```diff
@@ -284,37 +303,45 @@ impl Instance {
...
+    /// ```
+    pub fn dyn_func_with_index(&self, func_index: usize) -> ResolveResult<DynFunc> {
+        let func_index = FuncIndex::new(func_index);
+
+        let sig_index = *self
+            .module
+            .info
+            .func_assoc
+            .get(func_index)
+            .expect("broken invariant, incorrect func index");
```
MarkMcCaskey: > broken invariant, incorrect func index

It's not really a broken invariant if it's a user argument.  When taking user input, we'll want to handle the out of bounds case without panicking.  The panic is there because before it was theoretically impossible (due to invariants) for this check to fail -- that is no longer the case.

lib/runtime-core/src/instance.rs
```diff
@@ -164,88 +164,107 @@ impl Instance {
...
-                .get(*func_index)
-                .expect("broken invariant, incorrect func index");
-            let signature =
-                SigRegistry.lookup_signature_ref(&self.module.info.signatures[sig_index]);
+        let sig_index = *self
+            .module
+            .info
+            .func_assoc
+            .get(func_index)
+            .expect("broken invariant, incorrect func index");
```
MarkMcCaskey: Same here, we should handle this case when taking user input

lib/runtime-core/src/instance.rs
```diff
@@ -164,88 +164,107 @@ impl Instance {
...
     {
-        let export_index =
-            self.module
-                .info
-                .exports
-                .get(name)
-                .ok_or_else(|| ResolveError::ExportNotFound {
-                    name: name.to_string(),
-                })?;
+        let func_index = FuncIndex::new(func_index);
```
MarkMcCaskey: I'm undecided on whether or not this function should take a `FuncIndex` or a `usize`.  The reason we use typed indices is so that numbers don't get mixed up.  Perhaps the context here is strong enough that we don't need the extra type safety. 

Hmmm, `FuncIndex::new` is at least undocumented and perhaps not public.  Then I suppose it's a matter of if this function should be called with arbitrary user-supplied numbers or if the `FuncIndex` must come from somewhere.  If the `FuncIndex` comes from a validated source then my comments about the `expect` regarding the `invariant` are incorrect because we'd be validating when handing out the `FuncIndex`s (though I'm not sure that's possible to do in general).

I'd like to hear some more opinions/arguments on this before shipping!

syrusakbary:
(following up with @YaronWittenstein in Slack to evaluate better the use case)

MarkMcCaskey:
For future reference, a solution to this problem is to build your own cache of function index to function name using [Instance::exports](https://docs.rs/wasmer-runtime-core/0.16.2/wasmer_runtime_core/instance/struct.Instance.html#method.exports) and [Instance::resolve_func](https://docs.rs/wasmer-runtime-core/0.16.2/wasmer_runtime_core/instance/struct.Instance.html#method.resolve_func)

That might look something like:

```Rust
instance.exports()
        .filter(|(_, ex)| matches!(ex, Export::Function { .. }))
        .map(|(name, _)| (instance.resolve_func(&name).unwrap(), name.clone())
        .collect::<HashMap<usize, String>>();
```

Note: I didn't test this code and this relies on the new (as of 1.42.0) [`matches!` macro](https://doc.rust-lang.org/std/macro.matches.html)

PR #1233, opened 2020-02-20
Improved Wasmer C API release artifacts

syrusakbary:
# Description

This PR updates the artifacts generated for Wasmer C API, in a way that is much more consumable, including a quick README and License.

So, after running `make capi && make build-capi` We will have a `wasmer-c-api.tar.gz` file with the following structure
```
/
  lib/
    libwasmer.a
    libwasmer.so
  include/
    wasmer.h
    wasmer.hh
  README.md
  LICENSE
```

See example generated artifact here: 
[wasmer-c-api.tar.gz](https://github.com/wasmerio/wasmer/files/4228560/wasmer-c-api.tar.gz)

# Review


lib/runtime/Cargo.toml
```diff
@@ -2,7 +2,6 @@
 name = "wasmer-runtime"
 version = "0.13.1"
 description = "Wasmer runtime library"
-documentation = "https://wasmerio.github.io/wasmer/c/runtime-c-api/"
```
syrusakbary: This was a wrong doc link, intended to be the `runtime-c-api` not for the `runtime`.

We are fixing it in this PR :)

syrusakbary:
bors try

syrusakbary:
bors try

syrusakbary:
bors try

syrusakbary:
bors try-

syrusakbary:
bors try

syrusakbary:
bors try

Makefile
```diff
@@ -294,6 +294,33 @@ build-install:
...
+	cp target/release/libwasmer_runtime_c_api.dylib ./capi/lib/libwasmer.dylib
+	cp target/release/libwasmer_runtime_c_api.dylib ./capi/lib/libwasmer.a
+	# Fix the rpath for the dylib
+	install_name_tool -id "@rpath/libwasmer.dylib" ./capi/lib/libwasmer.dylib
+else
+	cp target/release/libwasmer_runtime_c_api.so ./capi/lib/libwasmer.so
+	cp target/release/libwasmer_runtime_c_api.a ./capi/lib/libwasmer.a
+endif
+endif
+
```
Hywan: Please remove this empty line.

Makefile
```diff
@@ -294,6 +294,33 @@ build-install:
...
+	find target/release/build -name 'wasmer.h*' -exec cp {} ./capi/include ';'
+	cp LICENSE ./capi/LICENSE
+	cp lib/runtime-c-api/distribution/* ./capi/
```
Hywan: The `distribution` directory is added in this PR.
What do you think adding the `doc/` directory instead? It already contains [an `index.md` file](https://github.com/wasmerio/wasmer/blob/master/lib/runtime-c-api/doc/index.md) that can be used as the `README.md` for this kind of archive (modulo some reformulations). Thoughts?

The idea behind my proposal is to limit the number of `README.md` and documentation to maintain.

lib/runtime-c-api/distribution/README.md
```diff
@@ -0,0 +1,32 @@
...
+clang YOUR_FILE -Iinclude -lwasmer -Llib
+```
+
+> Note: append ` -rpath lib` if you are in macOS.
+
+## Examples
+
+You can check examples of how to use the Wasmer C API here:
+
+https://docs.wasmer.io/integrations/c/examples
```
Hywan: This link is broken.

syrusakbary: The link would work on we release the new docs :)

https://new-docs.wasmer.io/integrations/c/examples

Makefile
syrusakbary: That's a good point. I just fixed it to reuse the same `doc/index.md` file!

syrusakbary:
> Should we include the lib/runtime-c-api/doc/ directory? I think it should be great to offer offline documentation?

I tried, but is not trivial to do it (since it requires doxygen installed in all the distributions OR manage more artifacts between pipelines).
I think we should do it in a PR apart

> Should we rename libwasmer_runtime_c_api to libwasmer? I understand it is simpler and shorter, but it's less self-describing. Could it “block” us for future projects or distributed/shared libraries?

I think we should keep it as simple as possible for now. `libwasmer` is a bit more aligned, since our header is named `wasmer.h`.

Happy to revisit this in the future though :)

Makefile
```diff
@@ -294,6 +294,31 @@ build-install:
 	cp ./target/release/wasmer ./install/bin/
 	tar -C ./install -zcvf wasmer.tar.gz bin/wapm bin/wasmer
 
+UNAME_S := $(shell uname -s)
+
+build-capi:
```
MarkMcCaskey: It seems that `build-capi` doesn't actually build the capi, it assumes it's there and copies it.  I think we should rename it!

Makefile
```diff
@@ -306,9 +331,11 @@ publish-release:
 dep-graph:
 	cargo deps --optional-deps --filter wasmer-wasi wasmer-wasi-tests wasmer-kernel-loader wasmer-dev-utils wasmer-llvm-backend wasmer-emscripten wasmer-emscripten-tests wasmer-runtime-core wasmer-runtime wasmer-middleware-common wasmer-middleware-common-tests wasmer-singlepass-backend wasmer-clif-backend wasmer --manifest-path Cargo.toml | dot -Tpng > wasmer_depgraph.png
 
-docs:
-	cargo doc --features=backend-singlepass,backend-cranelift,backend-llvm,docs,wasi,managed --workspace --document-private-items --no-deps
+docs-capi:
 	cd lib/runtime-c-api/ && doxygen doxyfile && cd ..
```
MarkMcCaskey: You didn't write this code in this PR, but it looks wrong, we cd 2 directories up and 1 directory back.

`cd -` pops back to where you `cd`'d from no matter where it was!

Hywan:
Good job :-).

PR #1234, opened 2020-02-20
Check for unused excluded spectest failures.

losfair:
Previously if a spectest failure was (maybe unintentionally) fixed then the corresponding line in `excludes.txt` will be silently ignored. This PR adds a check that each excluded spectest failure actually occurred.


losfair:
bors try

losfair:
bors try

losfair:
bors try

syrusakbary:
(merged without bors as `bors try` succeeded)

PR #1238, opened 2020-02-21
doc(readme) Fix logo

Hywan:
Addresses regression introduced in #1236.

This patch updates paths to the logo, that has moved into the `assets/` directory.


syrusakbary:
Good catch!

PR #1259, opened 2020-03-02
expose the readwrite section

Geal:
# Description

The readwrite section is part of the code generated by LLVM, and should be accessible if we want to move things around when compiling position independent code

### DIFF
--- a/lib/runtime-core/src/backend.rs
+++ b/lib/runtime-core/src/backend.rs
@@ -247,6 +247,10 @@ pub trait RunnableModule: Send + Sync {
         None
     }
 
+    fn get_read_section(&self) -> Option<&[u8]> {
+        None
+    }
+
     /// Returns the beginning offsets of all functions, including import trampolines.
     fn get_offsets(&self) -> Option<Vec<usize>> {
         None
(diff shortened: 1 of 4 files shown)


Geal:
while I'm at it, it would be just as useful to open access to the read section, should I update the PR with that?

lib/runtime-core/src/backend.rs
```diff
@@ -243,6 +243,14 @@ pub trait RunnableModule: Send + Sync {
         None
     }
 
+    fn get_readwrite_section(&self) -> Option<&[u8]> {
```
Hywan: Please add documentation here :-).

lib/runtime-core/src/backend.rs
```diff
@@ -243,6 +243,14 @@ pub trait RunnableModule: Send + Sync {
...
+        None
+    }
+
+    fn get_read_section(&self) -> Option<&[u8]> {
```
Hywan: That part needs documentation too.

syrusakbary:
Closing PR since it was targeting the old Wasmer codebase and the codebase has changed widely

PR #1263, opened 2020-03-03
Fix WASI directory bugs, improve WASI preopen APIs

MarkMcCaskey:
Fixes the issue with the `/`, adds support for preopened directories with `/` in it and `./` in it:
- `./` forces it to open in "the current directory" (`.` (technically `/.` due to the implementation))
- `/` opens it in the root
- no prefix now behaves the same as `./`
- multiple `/` in a preopen alias will traverse the filesystem and open directories as needed

New logic likely not fully applied, `WasiFs::new()` does _not_ support the new logic (this is still used in `generate_import_object_snapshot0`....)

I just realized I didn't retest the `--dir=a/b/c` case at all, that logic has not been updated here

# Review


Hywan:
Probably still relevant.

PR #1264, opened 2020-03-03
feat(interface-types) Implement lifting and lowering instructions

Hywan:
For the moment, WIT defines [a bunch of lifting and lowering instructions](https://github.com/WebAssembly/interface-types/blob/master/proposals/interface-types/working-notes/instr.md). The encoders, the decoders and the AST support them, but we didn't have any instruction implementations for them. This PR implements the instructions for real.

To handle casting failure, I decided to use the `TryFrom` trait, instead of `as`, such as:

```rust
let x: i32 = 128;
let y: i8 = x.try_into().unwrap(); // fails
```

while:

```rust
let x: i32 = 128;
let y: i8 = x as _; // just overflows silently
```

Not all `*x` instructions aren't implement _yet_.

### DIFF
--- a/lib/interface-types/src/interpreter/instructions/lowering_lifting.rs
+++ b/lib/interface-types/src/interpreter/instructions/lowering_lifting.rs
@@ -1,0 +1,342 @@
+use crate::interpreter::wasm::values::InterfaceValue;
+use std::convert::TryInto;
+
+macro_rules! lowering_lifting {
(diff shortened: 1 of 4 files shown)


Hywan:
> I'm not exactly sure why lowering_lifting has a function that returns a closure instead of just having the function take runtime itself 🤔

That's how the instruction implementations are designed for the moment :-).

PR #1265, opened 2020-03-04
Improve `WasmPtr` documentation

MarkMcCaskey:
This came up multiple times when responding to people on our issue tracker.

Adds an example and a lot more information about `WasmPtr` and `ValueType`

### DIFF
--- a/lib/runtime-core/src/memory/ptr.rs
+++ b/lib/runtime-core/src/memory/ptr.rs
@@ -73,7 +73,7 @@ impl<T: Copy + ValueType> WasmPtr<T, Item> {
     /// Dereference the `WasmPtr` getting access to a `&Cell<T>` allowing for
     /// reading and mutating of the inner value.
     ///
-    /// This method is unsound to use with unsynchronized shared memory.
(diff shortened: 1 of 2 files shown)


lib/runtime-core/src/memory/ptr.rs
```diff
@@ -48,8 +66,16 @@ fn align_pointer(ptr: usize, align: usize) -> usize {
...
 
+/// Methods for `WasmPtr`s to data that can be dereferenced, namely to types
+/// that implement [`ValueType`], meaning that they're valid for all possible
+/// bit patterns.
 impl<T: Copy + ValueType> WasmPtr<T, Item> {
-    /// Dereference this `WasmPtr`.
+    /// Dereference the `WasmPtr` getting access to a `&Cell<T>` allowing for
+    /// reading and mutating of the inner value.
+    ///
+    /// This method is unsound to use with unsynchronized shared memory.
```
syrusakbary:
```suggestion
    /// This method is unsound if used with unsynchronized shared memory.
```

PR #1271, opened 2020-03-05
Fix floating point calling convention for `DynamicFunc` trampolines.

losfair:
- [x] Allocate GPR/XMM registers according to on argument types in runtime-core/trampoline_x64.
- [ ] Previously singlepass used GPR for all arguments. Needs to be fixed.


syrusakbary:
I believe this is fixed in wasmer `master` thanks to reborn. I'll close the PR for now.
Could you confirm that's the case @losfair ?

Thanks!

PR #1272, opened 2020-03-05
Fix WasmPtr to work with accesses accessing the final valid byte

### LINKED ISSUE #1258: WASI fd_write and fd_read fails on memory boundary
### Describe the bug

WASI `fd_write` and `fd_read` fails if the iovec describes a buffer such that its right boundary coincides with the memory right boundary.

Example:

There's 1 page of memory (0x10000 bytes).

Callng `fd_read` or `fd_write` with  `iovec[ 65532, 4 ]`:

```
    FFFC FFFD FFFE FFFF
───┬────┬────┬────┬────╖
...│    │    │    │    ║
───┴────┴────┴────┴────╜
   │←───── len: 4 ────→│
   ↑ offset: 0xFFFC (65532)
```

The wasm memory size is just enough to fit this buffer. But the call errs with code 21. 

If len is changed to 3, it works nicely.

If memory size is changed to 2 (2 * 0x10000 bytes), it works nicely.

**wasmer version** 0.14.1
**OS** x86_64 GNU/Linux (Ubuntu 18.04)
Rust compiler is not used.

### Steps to reproduce

Run this .wat file with `wasmer run`

https://gist.github.com/subzey/7dd2cc27274b59322d78878f73a0944d

### Expected behavior

It should output `TEST` into stdout and exit with `0` exit code.

### Actual behavior

It outputs nothing and exists with `21`.

### Additional context

This file executed with `wasmtime` works as expected.

MarkMcCaskey:
Resolves #1258 

The fix was to change `>=` into `>`.  Doing this made us vulnerable to accessing memory to create an empty slice just out of bounds so we have to add some additional checks to make sure that that can't happen.

This PR also prevents accessing arrays of length 0 (the length bound is non-inclusive, so length 0 is never valid) and prevents access of zero-sized types.

The zero-sized type checks will probably be inlined (or will be in the future as `const fn` gets more mature) so provide no additional overhead.  The checking of if length == 0 does add some overhead, but on modern CPUs it shouldn't be an issue as it's a branch that should be always false in normal use.

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## **[Unreleased]**
 
+- [#1272](https://github.com/wasmerio/wasmer/pull/1272) Fix off-by-one error bug when accessing memory with a `WasmPtr` that contains the last valid byte of memory. Also changes the behavior of `WasmPtr<T, Array>` with a length of 0 and `WasmPtr<T>` where `std::mem::size_of::<T>()` is 0 to always return `None`
+
 ## 0.15.0 - 2020-03-04
 
 - [#1263](https://github.com/wasmerio/wasmer/pull/1263) Changed the behavior of some WASI syscalls to now handle preopened directories more properly. Changed default `--debug` logging to only show Wasmer-related messages.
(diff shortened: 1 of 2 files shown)


Hywan:
:+1:

PR #1273, opened 2020-03-06
compile in PIC mode

Geal:
# Description
this enables compiling position independent code. Linking this to #1259: if the code is moved, the read and readwrite section mist be moved as well, and stay at the same offsets relatively.

### DIFF
--- a/lib/llvm-backend/src/code.rs
+++ b/lib/llvm-backend/src/code.rs
@@ -8690,7 +8690,7 @@ impl<'ctx> ModuleCodeGenerator<LLVMFunctionCodeGenerator<'ctx>, LLVMBackend, Cod
                 &cpu_name.unwrap_or(TargetMachine::get_host_cpu_name().to_string()),
                 &cpu_features.unwrap_or(TargetMachine::get_host_cpu_features().to_string()),
                 OptimizationLevel::Aggressive,
-                RelocMode::Static,
+                RelocMode::PIC,
                 CodeModel::Large,
             )
             .unwrap();


Hywan:
bors try

nlewycky:
We can't take this PR right now. If you apply it in and run `make spectests-llvm` in a loop, building against an assertions-enabled build of llvm 8.0, you will eventually hit this assertion:

```
spectest-6f3eded9b60cbf91: /home/nicholas/llvm-8.0.0.src/lib/ExecutionEngine/RuntimeDyld/RuntimeDyldELF.cpp:304: void llvm::RuntimeDyldELF::resolveX86_64Relocation(const llvm::SectionEntry&, uint64_t, uint64_t, uint32_t, int64_t, uint64_t): Assertion `isInt<32>(RealOffset)' failed.
```

This happens about 1 in 250 runs in my testing, and is probably relying on ASLR to randomly generate the conditions where it occurs. I think the proximate cause is that distance between some instruction and the data it needs to refer to is too large to apply the PIC relocation. As to why that is, whether LLVM should have picked a different instruction sequence (we are building in large code model after all) or if we need to keep some things within a maximum distance apart in order to use PIC, I'm not sure.

This was originally changed in 59597b9e8302b756445ef1794616e07be654071a.

Hywan:
This PR is now outdated, and I'm not sure it's still required (by that I mean your initial issue might have changed). Can you confirm @Geal? Until then, I'm closing this issue, and I'm fully dispoed to reopen it if required.

Geal:
Not required anymore, thanks!

PR #1280, opened 2020-03-09
fix(runtime-core) Protecting a memory of size zero raises an error on Windows

Hywan:
The other methods on `Memory` are gated against a memory of size 0, except the `protect` method.

It raised bugs in https://github.com/wasmerio/python-ext-wasm/pull/100.


Hywan:
This patch is tested in https://github.com/wasmerio/python-ext-wasm/pull/142, and the bug is solved. This PR is thus valid.

lib/runtime-core/src/sys/windows/memory.rs
```diff
@@ -110,6 +110,13 @@ impl Memory {
         let size = round_up_to_page_size(range_end - range_start, page_size);
         assert!(size <= self.size);
 
+        // Allocating a memory of size 0 raises an error on Windows.
+        if size == 0 {
+            self.protection = Protect::None;
```
Hywan: Maybe we should just `self.protection = protect` here, and return `()`. I'm not sure what is best. The function receives a `protect` argument, and the user expects the memory to get this protection variant, not `None`.

syrusakbary: I think we should do `self.protection = protect`. This will shield better for the future.

```suggestion
            self.protection = protect;
```

syrusakbary: I did the change and tests started to fail?

Hywan: So maybe we shouldn't change for `protect` :-p.

syrusakbary:
```suggestion
            self.protection = Protect::None;
```

lib/runtime-core/src/sys/windows/memory.rs
```diff
@@ -110,6 +110,13 @@ impl Memory {
         let size = round_up_to_page_size(range_end - range_start, page_size);
         assert!(size <= self.size);
 
+        // Allocating a memory of size 0 raises an error on Windows.
+        if size == 0 {
+            self.protection = protect;
```
syrusakbary:
```suggestion
            self.protection = Protect::None;
```

syrusakbary:
Still fails @Hywan

Hywan:
Rollback to `self.protection = Protect::None`?

syrusakbary:
Both failed

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

lib/runtime-core/src/sys/windows/memory.rs
```diff
@@ -45,7 +46,10 @@ impl Memory {
         let ptr = unsafe { VirtualAlloc(ptr::null_mut(), size, flags, protect) };
 
         if ptr.is_null() {
-            Err("unable to allocate memory".to_string())
+            Err(format!(
+                "unable to allocate memory (error code: %d)",
```
MarkMcCaskey, line 50: What's up with the `%d` error formatting here?

does this compile?

Hywan: Oopsy :-p

syrusakbary:
This should be solved in the refactor.

Hywan:
I guess yes, we need to test though.

syrusakbary:

Hywan:
Confirmed it is!

PR #1283, opened 2020-03-09
Workaround for floating point arguments and return values in `DynamicFunc`s.

losfair:
This PR makes floating point arguments and return values for `DynamicFunc`s work correctly in all three backends.

Previously Singlepass used integer registers for all arguments. This PR adds another thin trampoline layer just before control is transferred to the import function, so that arguments will be rearranged strictly according to the System V ABI.

The full fix would require singlepass to implement the SysV calling convention internally too: https://github.com/wasmerio/wasmer/pull/1271 . This is just a workaround.


losfair:
bors try

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -246,44 +248,49 @@ impl TrampolineBufferBuilder {
         &mut self,
         target: unsafe extern "C" fn(*const CallContext, *const u64) -> u64,
         context: *const CallContext,
-        num_params: u32,
+        params: &[Type],
```
nlewycky: Any reason this shouldn't just take the full function signature? As it happens it only used to need the number of parameters, now it needs the whole list of the types, but isn't that an implementation detail?

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -407,6 +418,43 @@ mod tests {
         assert_eq!(ret, 136);
     }
 
+    #[test]
+    fn test_trampolines_with_floating_point() {
+        unsafe extern "C" fn inner(n: *const CallContext, args: *const u64) -> u64 {
+            let n = n as usize;
+            let mut result: u64 = 0;
+            for i in 0..n {
```
nlewycky: Sorry I'm confused – `n` was a pointer to a CallContext, so if it's at address 0x00005620d48999e8 then this loop will run 94699004729832 times?

lib/runtime-core/src/typed_func.rs
```diff
@@ -360,11 +386,27 @@ impl<'a> DynamicFunc<'a> {
             func: Box::new(func),
         });
         let ctx = Box::into_raw(ctx);
-        builder.add_callinfo_trampoline(
-            enter_host_polymorphic,
-            ctx as *const _,
-            (signature.params().len() + 1) as u32, // +vmctx
-        );
+
+        let mut native_param_types = vec![Type::I32]; // vm::Ctx is the first parameter.
```
nlewycky: Given that pointers are 64-bits, why wouldn't this be a Type::I64?

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -880,6 +893,92 @@ impl ModuleCodeGenerator<X64FunctionCode, X64ExecutionContext, CodegenError>
...
+            let mut param_locations: Vec<Location> = vec![];
+
+            // Allocate stack space for arguments.
+            let stack_offset: i32 = if sig.params().len() > 5 {
+                5 * 8
+            } else {
+                (sig.params().len() as i32) * 8
+            };
+            if stack_offset > 0 {
+                a.emit_sub(
```
nlewycky: Note that x86-64 SysV ABI requires a 16-byte stack alignment, not 8-byte. While a fault due to misaligment is rare, it could happen if the callee uses aligned movs to access the stack. (I'm not sure if this is a bug because I don't know whether something else is guaranteed to realign the stack between here and the point where we next enter non-singlepass-generated code. Doesn't have to be an import, could also be a vmcalls function.)

lib/runtime-core/src/parse.rs
```diff
@@ -110,11 +110,46 @@ pub fn read_module<
...
+                    let info_read = info.read().unwrap();
+                    mcg.feed_signatures(info_read.signatures.clone())
+                        .map_err(|x| LoadError::Codegen(format!("{:?}", x)))?;
+                }
+            }
+            _ => {}
+        }
+        match *state {
+            ParserState::BeginFunctionBody { .. } | ParserState::EndWasm => {
+                if !mcg_info_fed {
```
MarkMcCaskey: If we use separate `match` statements, moving the `if !mcg_*_fed {` checks outside the `match *state` will probably have better performance due to better branch prediction.

lib/runtime-core/src/parse.rs
```diff
@@ -110,11 +110,46 @@ pub fn read_module<
...
     let mut name_builder = Some(StringTableBuilder::new());
     let mut func_count: usize = 0;
+    let mut mcg_signatures_fed = false;
     let mut mcg_info_fed = false;
 
     loop {
         use wasmparser::ParserState;
         let state = parser.read();
+
+        // Feed signature and namespace information as early as possible.
```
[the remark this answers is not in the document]
MarkMcCaskey: Maybe it makes sense to use the Wasmparser types that let us be more specific about what exactly we want: https://docs.rs/wasmparser/0.51.4/wasmparser/struct.CodeSectionReader.html .  They have a bunch of section-specific parsers, there aren't validating versions of these parsers though, so we'd have to do a validation pass first

lib/spectests/tests/spectest.rs
```diff
@@ -342,7 +352,7 @@ mod tests {
                                     file: filename.to_string(),
                                     line: line,
                                     kind: format!("{}", "Module"),
-                                    message: format!("caught panic {:?}", e),
+                                    message: format!("caught panic {:?}", format_panic(&e)),
```
MarkMcCaskey: `{:?}` => `{}`, debug printing a string will do extra quotes to it.  If this is your intention, then I prefer being explicit with `\"` as debug printing output can change.

lib/runtime-core-tests/tests/imports.rs
```diff
@@ -154,13 +185,77 @@ fn imported_functions_forms(test: &dyn Fn(&Instance)) {
...
+
+                    let memory = vmctx.memory(0);
+                    let shift_ = shift + memory.view::<i32>()[0].get();
+                    let i: i32 = (&inputs[0]).try_into().unwrap();
+                    let j: i64 = (&inputs[1]).try_into().unwrap();
+                    let k: f32 = (&inputs[2]).try_into().unwrap();
+
+                    dbg!(i);
+                    dbg!(j);
+                    dbg!(k);
```
MarkMcCaskey: debug code, should be removed before shipping!

lib/runtime-core/src/trampoline_x64.rs
```diff
@@ -395,8 +402,12 @@ mod tests {
         }
         let mut builder = TrampolineBufferBuilder::new();
         let ctx = TestContext { value: 100 };
-        let idx =
-            builder.add_callinfo_trampoline(do_add, &ctx as *const TestContext as *const _, 8);
+        let param_types: Vec<Type> = (0..8).map(|_| Type::I32).collect();
```
MarkMcCaskey: ⛳️  `let param_types = vec![Type::I32; 8];`

lib/runtime-core/src/trampoline_x64.rs
losfair: The `*const CallContext` is casted from the length of the parameter list as an integer in the following call to `add_callinfo_trampoline`, so it's not really a pointer.

lib/singlepass-backend/src/codegen_x64.rs
losfair: The call to the actual import is a tail call and the stack space allocated here will be released before jumping to the target function.

lib/runtime-core/src/trampoline_x64.rs
losfair: Will add a comment.

lib/runtime-core/src/parse.rs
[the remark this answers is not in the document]
losfair: Since this block borrows `mcg`, there would be lifetime issues if this is made into a closure. And yeah this is about code reuse. The same code was already duplicated twice for `BeginFunctionBody` and `EndWasm`, and this change will add one more duplication if done "in place".

lib/runtime-core/src/trampoline_x64.rs
losfair: This function does not prepend the WebAssembly-specific `vm::Ctx` argument and is just a general-purpose trampoline generator, therefore should not take a WebAssembly `FuncSig`. I added a `returns` argument just in case of future compatibility.

lib/runtime-core/src/parse.rs
losfair: I think this is a kind of simple optimization that can be left to the compiler?

losfair:
bors try

lib/runtime-core/src/parse.rs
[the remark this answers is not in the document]
MarkMcCaskey: `mcg` could be passed as an argument here to make closures work

lib/runtime-core/src/parse.rs
nlewycky: Depends whether the compiler can prove `*match` is side-effect free. I'm sorta torn on this one, I think we should write whatever is clearest to the reader, but I'm not sure that pulling this part of the match out in advance is actually clearer. I'm happy to let you and Mark decide.

lib/runtime-core/src/trampoline_x64.rs
nlewycky: OK, but you can't just go around passing integers as pointers. The compiler makes the assumption that the pointer is sufficiently aligned to point to a `CallContext`, which could lead it to conclude that the bottom bits are zero. All you do here is use it as a `usize` which will cast it, but if you were to pass, say, a bitfield through, the compiler would actually delete the tests of the bottom bits.

I don't know about Rust, but in C and C++ the rule is that all pointers must actually point to a valid object when created (whenever an expression of pointer type is evaluated). You're guaranteed that casting a pointer to an appropriately sized int and back to the same pointer type gets you the original pointer back, but that's a different property from being able to take an int, cast it to a pointer, and put it back to an int. For a crazy example that no compiler I know of does, if there's a value in the range [0, 4) the compiler could pack it in the same storage as the int.

If this is a temporary part of the workaround that will go away with the permanent fix, let me know. Casting this way doesn't seem to go wrong in practice, much.

losfair: This is part of the test code and does not go into the compiled crate, so the worst case is one more failing test and we know something is wrong here.

lib/runtime-core/src/parse.rs
losfair: Will make it into a closure.

lib/runtime-core/src/trampoline_x64.rs
nlewycky: Oh no, that's never the worst case. :-) The worst case is something like "not only does this test pass when it should actually fail, it causes the other tests running in other threads to incorrectly pass when they should be failing". Anyway, I don't think we need to change this right now.

PR #1284, opened 2020-03-10
feat(interface-types) Implement string and memory instructions

Hywan:
Previously, we had `read-utf8` and `write-utf8` instructions. The latest version of the standard proposes respectively `memory-to-string` and `string-to-memory`. The semantics is close though.


Hywan:
bors try

PR #1285, opened 2020-03-10
feat(interface-types) Greatly improve errors

Hywan:
~This PR is build on top of #1284. It must be merged first. [View the exact diff](https://github.com/Hywan/wasmer/compare/feat-interface-types-instructions-string-and-memory...Hywan:feat-interface-types-better-error?expand=1).~

Errors in `wasmer-interface-types` were just `&'static str`, which isn't quite great 😉. This PR introduces the `errors` module, with structures that represent errors. A `Display` implementation maps the errors to strings.


Hywan:
bors try

lib/interface-types/src/interpreter/instructions/memory_to_string.rs
```diff
@@ -1,53 +1,57 @@
...
-                    None => Err(format!(
-                        "`{}` failed because there is no memory to read.",
-                        instruction_name
-                    ))
-                }
-                None => Err(format!(
-                    "`{}` failed because there is not enough data on the stack (needs 2).",
-                    instruction_name,
-                ))
+            if memory_view.len() < pointer + length {
```
MarkMcCaskey: I think this may be affected by the same bug I recently fixed in WasmPtr,

Given a String, "hi" at address 0xFFFE the memory looks like

0xFFFE: 'h'
0xFFFF: 'i'

ptr (0xFFFE) + len (2) = 0x10000, failing this bounds check.  The fix, in general, is annoying when you handle length = 0 and zero sized types.

See https://github.com/wasmerio/wasmer/pull/1272 for my fix.

Better to have false positives than false negatives in bounds-checking code, but leaving a comment here as a heads up.  It might make sense to do a full pass for these types of bugs and fix them in a separate PR if you think it might clutter this one.

MarkMcCaskey: To clarify more, we want to make sure out of bounds pointers are never derefed, so ptr = 0x10000 and length = 0 will not be caught by the naive solution of changing `<` to `<=` here.

lib/interface-types/src/interpreter/mod.rs
```diff
@@ -165,7 +168,7 @@ where
         for executable_instruction in self.iter() {
             match executable_instruction(&mut runtime) {
                 Ok(_) => continue,
-                Err(message) => return Err(message),
+                Err(error) => return Err(error),
```
MarkMcCaskey: ⛳️ 
```Rust
for executable_instruction in self.iter() { 
   executable_instruction(&mut runtime)?;
}
```

lib/interface-types/src/interpreter/instructions/memory_to_string.rs
Hywan: I'm not sure I understand. I've pushed a new commit to use inclusive an range, and consequently a special condition to handle empty string, but I'm not sure there is a bug right now. The `test_memory_to_string` already reads at bounds.

lib/interface-types/src/errors.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// The targeted type.
+    ///
+    /// `InterfaceType` is used to represent the native type by
+    /// associativity.
+    pub to: InterfaceType,
+}
+
+/// Structure to represent the errors for instructions.
+#[derive(Debug)]
+pub struct InstructionError {
```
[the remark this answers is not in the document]
Hywan: Done!

lib/interface-types/src/interpreter/mod.rs
Hywan: Of course 🤦‍♂…

PR #1288, opened 2020-03-11
Update some dependencies manually, removing duplicate deps

MarkMcCaskey:
Removes all duplicate deps (as seen by `cargo +stable tree -d`).  Primarily by updating `wasm-debug` but also by through updating some other dependencies.


MarkMcCaskey:
bors try

Cargo.lock
```diff
@@ -1802,20 +1810,74 @@ version = "0.9.0+wasi-snapshot-preview1"
...
 
+[[package]]
+name = "wasm-bindgen"
+version = "0.2.59"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3557c397ab5a8e347d434782bcd31fc1483d927a6826804cec05cc792ee2519d"
+dependencies = [
+ "cfg-if",
+ "wasm-bindgen-macro",
+]
```
syrusakbary: Again wasm-bindgen here??

MarkMcCaskey: This is from criterion via plotters via js-sys, I believe it should be a dev-dependency

MarkMcCaskey: Which, from their Cargo.toml, seems to be only for `[target.'cfg(target_arch = "wasm32")'.dependencies]`

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

MarkMcCaskey:
bors merge

PR #1289, opened 2020-03-11
Enable `DynamicFunc` for closures with captured environment.

losfair:
Previously we disabled `DynamicFunc` for any non-zero-sized closures to leave space for future changes. However this feature is critical for applications that needs to bring context with host functions, like integrations with dynamic lauguages. So it might be good to enable it.

A question left is: should we put this behind a feature flag or enable it by default?

@Hywan @syrusakbary


syrusakbary:
Let's put it behind a feature flag for now @losfair

Hywan:
I think a flag is OK for the moment.

PR #1290, opened 2020-03-11
Fix `wax` command packaging for install script

MarkMcCaskey:
Fixes bug causing `wax` to not actually be installed (or be executable) by the install script


lmorchard:
Just to circle back - this works for me now! :+1:

PR #1300, opened 2020-03-13
Update `wasi-tests` to test multiple versions of WASI

MarkMcCaskey:
This change also cleans up the `wasi-test` generation a bit.  It's
actually still really, really messy, but at least now it's split up
into easier to understand chunks.

There's still a lot of low-hanging fruit in terms of improving the
readability and maintainability of the code.

Improves test coverage over what's on master.  Things in `ignores.txt` weren't being tested on master either; we should do a follow up to remove them from there (will require finding a good solution for raw WASI calls)

# Review


lib/wasi-tests/build/wasitests.rs
```diff
@@ -170,39 +219,89 @@ fn test_{rs_module_name}() {{
...
+        out.push(test_mod);
+    }
+
+    out
+}
+
+fn run_prelude(should_gen_all: bool) -> &'static [WasiVersion] {
+    if should_gen_all {
+        println!(
+            "Generating WASI tests for all version of WASI. Run with WASI_TEST_GENERATE_ALL=0 to only generate the latest tests."
```
syrusakbary:
```suggestion
            "Generating WASI tests for all versions of WASI. Run with WASI_TEST_GENERATE_ALL=0 to only generate the latest tests."
```

lib/wasi-tests/tests/wasitests/mod.rs
```diff
@@ -5,26 +5,49 @@
...
-mod mapdir;
-mod path_link;
-mod path_rename;
-mod path_symlink;
-mod poll_oneoff;
-mod quine;
-mod readlink;
-mod wasi_sees_virtual_root;
-mod writing;
+mod snapshot1_close_preopen_fd;
```
syrusakbary: Perhaps having them in different dirs might be good? (not sure)

MarkMcCaskey: I agree, I think that's a good change.  It's a bit non-trivial to update it though as it will require updating some of the structure of the code and the way the ignore list works -- I think that'd be a good follow up issue, but I can prioritize it now if you want!

PR #1303, opened 2020-03-15
NaN canonicalization for singlepass backend.

losfair:
- [x] Implementation
- [x] Test


lib/spectests/tests/excludes.txt
```diff
@@ -266,6 +266,30 @@ clif:fail:data.wast:266:windows # AssertUnlinkable - caught panic Any
 clif:fail:data.wast:186:windows # AssertUnlinkable - caught panic Any
 clif:fail:data.wast:194:windows # AssertUnlinkable - caught panic Any
 
+# NaN canonicalization
+clif:fail:wasmer.wast:105 # AssertReturn - result I32(2143289345) ("0x7fc00001") does not match expected I32(2143289344) ("0x7fc00000")
```
syrusakbary: Any idea of why this happens? I thought Cranelift supported NaN canonicalization

nlewycky: Looks like it defaults to off in cranelift and nothing turns it on?

https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/wasmerio/cranelift%24+enable_nan_canonicalization&patternType=literal
https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/wasmerio/wasmer%24+enable_nan_canonicalization&patternType=literal

syrusakbary: @losfair as part of this PR can you make Cranelift read the `nan_canonicalization` field from the Compiler `Config` and use it accordingly? Thanks!

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -6339,6 +6833,14 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
```
nlewycky: Don't you need to pass in canonicalized nans? If the value is cncl_f32, it needs to be canonicalized because the callee might observe it?

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -6559,19 +7074,52 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
...
-                    Self::emit_relaxed_binop(
-                        a,
-                        &mut self.machine,
-                        Assembler::emit_mov,
-                        Size::S64,
-                        loc,
-                        Location::GPR(GPR::RAX),
-                    );
+                    match frame.returns[0] {
+                        WpType::F32 | WpType::F64 => {
```
nlewycky: I think this is correct because you can apply canonicalization early, but I think this is unnecessary?

losfair: Code reference here is outdated so I'm not sure which part you are referring to; but if you mean the `Phi`-like branch like `Br`-ing out, early canonicalization is necessary because we don't have the static tracking information after.

losfair:
bors try

losfair:
bors try

lib/singlepass-backend/src/codegen_x64.rs
```diff
@@ -6390,6 +6956,27 @@ impl FunctionCodeGenerator<CodegenError> for X64FunctionCode {
                     .collect();
                 self.machine.release_locations_only_regs(&params);
 
+                while let Some(fp) = self.fp_stack.last() {
```
syrusakbary: Can you add a comment on what this while loop is doing?

losfair:
bors try

PR #1306, opened 2020-03-16
feat(runtime-core) Introduce `Instance::dyn_func_by_index`

Hywan:
This PR proposes a new API `Instance::dyn_func_by_index` to call an exported function by its index. It is very similar to `Instance::dync_func`, except that it takes an index (`usize`) instead of a name (`&str`) to identify the function.


Hywan:
Hold on. I think another API would be better, something like `dyn_func_by_index` (similarly to `dyn_func`). It will help to get the signature of the function, which is better.

MarkMcCaskey:
I don't know the context for this PR, but when this came up last time the suggestion I gave was to build a cache on the user side of function indices to name or function indices to `DynFunc` if possible.  It's possible to iterate through all functions and get their name and function indices, then you just collect it into a map from index to name.

Hywan:
For the context: WIT defines an instruction called `call-core <index>`, and another one `string-to-memory <index>` where `<index>`  is a function index (in the former case, a function, in the latter case, an allocator function). In `wasmer-interface-types`, we don't need this new API because we have access to everything inside `wasmer-runtime-core`. However, to support WIT in language integrations like Python (with polyfills), we need to expose this API publicaly, hence `dyn_func_by_index`.

MarkMcCaskey:
Hmm, I think there were good reasons for why we didn't want to ship this as an API the last time it came up.  The reasons I can remember are that indices may not be guaranteed to be stable and that the API is a bit too low level and easy to misuse.

It seems that the external caching option has better performance anyways (constant time function index to Dynfunc) at the cost of storing a mapping from index to name for each function and doing 1 linear scan.

PR #1310, opened 2020-03-16
upgrade the blake3 dependency to 0.2

oconnor663:
Version 0.2 makes assembly implementations available. They're off by default, and I haven't enabled them here, because they require the build machine to have a C toolchain installed. But if that's already a requirement for Wasmer, we could enable them with the `"c"` feature, for both better runtime performance and faster build times.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -86,9 +86,9 @@ checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
 
 [[package]]
 name = "blake3"
-version = "0.1.5"
+version = "0.2.2"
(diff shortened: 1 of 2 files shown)


syrusakbary:
This looks great!

Note: We are also using blake3 in our [wapm-cli](https://github.com/wasmerio/wapm-cli) project

syrusakbary:
@oconnor663 could you push the `Cargo.lock` file as well? Thanks!

oconnor663:
Added `Cargo.lock`.

> Note: We are also using blake3 in our wapm-cli project

Awesome, just created https://github.com/wasmerio/wapm-cli/pull/188 over there.

PR #1313, opened 2020-03-17
Add types and methods to provide updated API

MarkMcCaskey:
Improving the API in a number of ways.

<details><summary>Current status </summary>

- [x] Get exports from Instance
  - [x] Func
     - [x] Documented
     - [x] Tested
  - [x] DynFunc
    - [x] Documented
    - [x] Tested
  - [x] Memory
    - [x] Documented
    - [x] Tested
  - [x] Table
    - [x] Documented
    - [ ] Tested ; fully testing this now will be difficult, blocked on Table API being completed
  - [x] Global
    - [x] Documented
    - [x] Tested
  - [x] Field syntax (fairly non-trivial)
- [x] Get imports from Module
  - [ ] Maybe update this to be an iterator instead of a Vec (side note, we may want to have a way to access specific types of imports too)
  - [x] Documented
  - [x] Tested
- [x] Get exports from Module
  - [x] Documented
  - [x] Tested
- [x] Get custom section from Module
  - [x] Figure out correct solution
  - [x] Ship separate PR that updates custom section code
  - [x] Documented
  - [x] Tested
- [ ] Updated Memory API
  - [x] Added conversion methods to bytes/pages with From
  - [ ] Documented
  - [ ] Tested
- [ ] Table APIs ; blocked on `wrap` being not linear time update (this update should also make it possible to retrieve a `Func` from `vm::Anyfunc`)
  - [ ] Table set (implemented needs to be checked)
    - [ ] Documented with examples
    - [x] Implemented
  - [ ] Table grow (implemented needs to be checked)
    - [ ] Documented with examples
    - [x] Implemented
  - [ ] Table get
    - [x] Structure implemented
    - [ ] Documented
    - [ ] Tested
- [ ] Module APIs (probably separate)
- [x] Update import object macro to handle a lack of trailing commas as well
  - [x] Add line in changelog about it
  - [x] Tested

</details>

# Review


lib/runtime-core-tests/tests/imports.rs
```diff
@@ -2,15 +2,44 @@ use std::{convert::TryInto, sync::Arc};
...
+    let module = compile_with(&wasm_binary, &get_compiler()).unwrap();
+    let import_object = imports! {};
+    let instance = module.instantiate(&import_object).unwrap();
+
+    let my_global: Global = instance.exports_new().get("my_global").unwrap();
+    assert_eq!(my_global.get(), Value::I32(45));
+    let double: Func<i32, i32> = instance.exports_new().get("double").unwrap();
+    assert_eq!(double.call(5).unwrap(), 10);
+    let add_one: DynFunc = instance.exports_new().get("add_one").unwrap();
+    assert_eq!(add_one.call(&[Value::I32(5)]).unwrap(), &[Value::I32(6)]);
```
syrusakbary: I have no idea how the call works without a vmctx... but 🤷‍♂️

MarkMcCaskey: The VmCtx is in the instance which `DynFunc` holds a reference to!

syrusakbary: Oh! Somehow I missed that we were actually instantiating a Wasm module. I thought you created a DynFunc by hand... my brain is not working as well today haha

lib/runtime-core/src/instance.rs
```diff
@@ -814,6 +822,89 @@ impl<'a> DynFunc<'a> {
...
+/// Use [`Instance::exports_new`] to get an `Exports` from an [`Instance`].
+pub struct Exports<'a> {
+    instance: &'a Instance,
+}
+
+impl<'a> Exports<'a> {
+    /// Get an export.
+    ///
+    /// ```
+    /// # use wasmer_runtime_core::{DynFunc, Func, Instance};
```
syrusakbary: Can this be `wasmer_runtime` instead? I would prefer people depending on the the outer shell.

MarkMcCaskey: Yeah, it's possible to move the doc comments with examples to `wasmer_runtime` instead, though the `#` on that line means that it's invisible to users in the docs, so it doesn't matter here

lib/runtime-core/src/module.rs
```diff
@@ -173,6 +326,107 @@ impl Clone for Module {
...
+    // TODO: why does function have no data?
+    Function,
+    /// The import is a global variable.
+    Global {
+        /// Whether or not the variable can be mutated.
+        mutable: bool,
+        /// The Wasm type that the global variable holds.
+        // TODO: attempt to understand explanation about 128bit globals:
+        // https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#webassembly-module-instatiation
+        type_: Type,
```
syrusakbary: I just realized of this. I think using `ty` might be a better field name :)

MarkMcCaskey: We can also do `r#type` if we want a field named type, but I think it's better in general to not use that syntax because it might confuse people not familiar with it

MarkMcCaskey: (`r#` makes it literal and allows you to use any name, including reserved keywords.  It was added to allow adding new keywords into the language and having a clear migration path for old code to not break.  It can also be used to not have to name things weirdly, but it requires the syntax to use)

lib/api-tests/tests/high_level_api.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    assert_eq!(outside_global.get(), Value::I32(15));
+    update_outside_global.call().unwrap();
+    assert_eq!(outside_global.get(), Value::I32(15 * 2));
+    update_outside_global.call().unwrap();
+    assert_eq!(outside_global.get(), Value::I32(15 * 2 * 2));
+}
+
+// copied from Rust stdlib: https://doc.rust-lang.org/nightly/nightly-rustc/src/serialize/leb128.rs.html#4-14
+macro_rules! impl_write_unsigned_leb128 {
```
syrusakbary: Why this is needed?

MarkMcCaskey: It's not strictly needed, but in case we needed the other functions, I figured it might be good to have around.  I can macro-expand/inline the one function I need for the custom section writing if you'd prefer that!

lib/runtime-core/src/instance.rs
```diff
@@ -58,6 +61,11 @@ pub struct Instance {
 }
 
 impl Instance {
+    /// Access the new exports API.
+    /// TODO: rename etc.
+    pub fn exports_new(&self) -> Exports {
```
syrusakbary: Can we rename it to `exports_map` for now?

MarkMcCaskey: Sure!  I have no intention of shipping with the name `exports_name`, it's just a placeholder for now

syrusakbary:
Can you make `instance.func` to use the new externals function under the hood? (and also mark it as deprecated)

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/api-tests/tests/high_level_api.rs
```diff
@@ -0,0 +1,177 @@
...
+    assert_eq!(num_exports, 6);
+    let (
+        mut test_table_seen,
+        mut test_global_seen,
+        mut test_ret_2_seen,
+        mut test_ret_4_seen,
+        mut set_test_global_seen,
+        mut update_outside_global_seen,
+    ) = (false, false, false, false, false, false);
+    for export in module.exports() {
```
syrusakbary: Shouldn't be better to just to have the exports to derive from `Eq`, `PartialEq` and then do a comparison on the expected exports via equality?

```rust
assert_eq!(module.exports(), [export::Export{kind:ExportKind::Table}, ...]);
```

lib/runtime-core/src/instance.rs
```diff
@@ -283,38 +225,9 @@ impl Instance {
     /// # Ok(())
     /// # }
     /// ```
+    #[deprecated(since = "0.17.0", note = "Please use `instance.exports.get()` instead")]
```
syrusakbary: Awesome.

```suggestion
    #[deprecated(since = "0.17.0", note = "Please use `instance.exports.get(name)` instead")]
```

lib/runtime-core/src/instance.rs
```diff
@@ -170,82 +181,13 @@ impl Instance {
...
```
syrusakbary:
```suggestion
    #[deprecated(since = "0.17.0", note = "Please use `instance.exports.get(name)` instead")]
```

lib/runtime-core/src/instance.rs
```diff
@@ -411,6 +324,123 @@ impl Instance {
     }
 }
 
+/// Private function implementing the inner logic of `Instance::func`
+fn func<'a, Args, Rets>(
```
syrusakbary: When / for what this function is used?

lib/api-tests/tests/high_level_api.rs
MarkMcCaskey: Hmm, I believe that would require them to either be orderable or hashable so we can compare them in an order-agnostic way but that wouldn't give us the property that we see every export exactly once. This seemed like a relatively simple way to completely test it in a way that's order agnostic (I didn't want the test to break if we change an implementation detail and the order they're returned in changes)

syrusakbary: I think exports should always return things in the proper order (as they are defined in the wasm).

lib/api-tests/tests/high_level_api.rs
```diff
@@ -0,0 +1,177 @@
...
+    out.extend(&custom_section_length_as_leb128);
+    out.extend(&name_length_as_leb128);
+    out.extend(custom_section_name.as_bytes());
+    out.extend(custom_section_contents);
+
+    out
+}
+
+#[test]
+fn it_works() {
```
syrusakbary: Right now we are testing a lot of different things in one, which makes harder to spot multiple issues as one function will fail in cascade.

We are testing:
1. Custom sections
2. Module exports (should be ordered)
3. Instantiation
4. Variables working without instantiating them:
   * Outside global
4. Exports working
   * Functions: function retrieval and function calls
   * Tables: table retrieval
   * Globals: globals retrieval and globals getters

I think all those should be tested separately

lib/api/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// this implementation should be moved
+impl CompiledModule for Module {
+    fn new(bytes: impl AsRef<[u8]>) -> error::CompileResult<Module> {
+        let bytes = bytes.as_ref();
+        wasmer_runtime_core::compile_with(bytes, &default_compiler())
+    }
+
+    fn from_binary(_bytes: impl AsRef<[u8]>) -> error::CompileResult<Module> {
```
syrusakbary: I think that should be just a `Result` ?

lib/runtime-core/src/instance.rs
```diff
@@ -814,6 +844,134 @@ impl<'a> DynFunc<'a> {
...
+    fn get_self(exports: &'a Exports, name: &str) -> ResolveResult<Self> {
+        let (inst_inner, module) = exports.get_inner();
+        dyn_func(module, inst_inner, name)
+    }
+}
+
+impl<'a, Args: WasmTypeList, Rets: WasmTypeList> Exportable<'a> for Func<'a, Args, Rets, Wasm> {
+    fn get_self(exports: &'a Exports, name: &str) -> ResolveResult<Self> {
+        let (inst_inner, module) = exports.get_inner();
+        func(module, inst_inner, name)
```
syrusakbary: I think it will be cleared to move `func` contents into here? (unless func is used somewhere else)

lib/runtime-core/src/instance.rs
```diff
@@ -814,6 +844,134 @@ impl<'a> DynFunc<'a> {
...
+                name: name.to_string(),
+            })
+        }
+    }
+}
+
+impl<'a> Exportable<'a> for DynFunc<'a> {
+    fn get_self(exports: &'a Exports, name: &str) -> ResolveResult<Self> {
+        let (inst_inner, module) = exports.get_inner();
+        dyn_func(module, inst_inner, name)
```
syrusakbary: I think it will be cleared to move `dyn_func` contents into here? (unless dyn_func is used somewhere else)

lib/runtime-core/src/macros.rs
```diff
@@ -126,11 +126,87 @@ macro_rules! __imports_internal {
...
+                "func2" => func!(func)
+            }
+        };
+        let _ = imports! {
+            "env" => {
+                "func1" => func!(func),
+                "func2" => func!(func),
+            }
+        };
+        use std::{ffi, ptr};
```
syrusakbary: Can we move the `imports!` with state into it's own test?

lib/runtime-core/src/module.rs
```diff
@@ -164,6 +166,152 @@ impl Module {
...
+    ///           .collect::<Vec<String>>();
+    ///
+    /// // And here we count the number of global variables exported by this module.
+    /// let num_globals =
+    ///     module.exports()
+    ///           .filter(|ed| ed.kind == ExportKind::Global)
+    ///           .count();
+    /// # }
+    /// ```
+    pub fn exports(&self) -> impl Iterator<Item = ExportDescriptor> + '_ {
```
syrusakbary: I think this should return just the array, not an iterator?

Thoughts?

lib/runtime-core/src/module.rs
```diff
@@ -174,6 +322,107 @@ impl Clone for Module {
...
 
+/// The type of import. This indicates whether the import is a function, global, memory, or table.
+// TODO: discuss and research Kind vs Type;
+// `Kind` has meaning to me from Haskell as an incomplete type like
+// `List` which is of kind `* -> *`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ImportType {
+    /// The import is a function.
+    // TODO: why does function have no data?
+    Function,
```
syrusakbary: I think we should have things like:

```
Func(FuncType),
Global(GlobalType),
Table(TableType),
Memory(MemoryType),
```

That way, there is no duplication of the descriptor types.

lib/api-tests/tests/high_level_api.rs
MarkMcCaskey: Sure, I'll update the docs on `ModuleInfo.exports` to say that it's supposed to be ordered and that we shouldn't use the APIs from ordered map that make it not insertion ordered.

lib/runtime/README.md
```diff
@@ -83,7 +84,8 @@ fn main() -> error::Result<()> {
     let instance = instantiate(WASM, &import_object)?;
 
     let values = instance
-        .dyn_func("add_one")?
+        .exports
+        .get::<DynFunc>("add_one")?
```
syrusakbary: Somehow I feel that `DynFunc` and `Func` should be merged together, as they reference to the same thing, just with a different API for calling the function.

We can have something like:

```rust
    let add_one = instance
        .exports
        .get::<Func<i32, i32>>("add_one")?;
    // For dynfunc usage
    let values = add_one.call(&[Value::I32(42)])?;
    // For normal func usage
    let values = add_one.call_native(42)?;
```

Thoughts?

lib/api/src/lib.rs
MarkMcCaskey: What did you have in mind for the error type?

syrusakbary: We can use `anyhow::Error` :)

lib/api/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+
+pub mod error {
+    //! Various error types returned by Wasmer APIs.
+    pub use wasmer_runtime_core::error::*;
+}
+
+pub use prelude::*;
+
+/// Idea for generic trait; consider rename; it will need to be moved somewhere else
+pub trait CompiledModule {
```
syrusakbary: I think this trait should be named `Module`. We can put this contents on the `wasm-common` crate I was working on.

lib/api/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+pub mod units {
+    //! Various unit types.
+    pub use wasmer_runtime_core::units::{Bytes, Pages};
+}
+
+pub mod types {
+    //! Types used in the Wasm runtime and conversion functions.
+    pub use wasmer_runtime_core::types::*;
```
syrusakbary: Can we explicit on the types? It makes easier to reason about the exposed API

lib/api/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+}
+
+pub mod error {
+    //! Various error types returned by Wasmer APIs.
+    pub use wasmer_runtime_core::error::*;
```
syrusakbary: Can we be explicit on the errors exposed? It makes easier to reason about the API

lib/api/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+            .map_err(|_| todo!("Current error enum can't handle this case"))?;
+
+        Module::from_binary(bytes.as_slice())
+        */
+    }
+
+    fn validate(_bytes: impl AsRef<[u8]>) -> error::CompileResult<()> {
+        todo!("validate")
+    }
+}
```
syrusakbary: Note: Can we have a trait for `Instance` as well?

It should have the fields `module` (returns a reference to the Module), `exports` (the instantiated exports).

lib/runtime-core/src/module.rs
MarkMcCaskey: It's definitely a trade off worth considering.

As I see it, returning an array has:
- Pros
  - simpler in the simplest use case
- Cons
  - takes longer at runtime: if we didn't want all the exports or we wanted a different structure, we now have to pay the cost twice
  - forces extra memory use through heap allocation instead of letting the caller decide
  - is less flexible: it forces many uses of this to now be more verbose while allowing the `Vec` case to be handled with a `collect::Vec<_>()`.  It makes the API less composable and interoperable with generic Rust abstractions.

If the API is for users new to Rust then returning a `Vec` makes sense to me, otherwise I think it makes sense to use iterators pretty much everywhere.  The reason being that an iterator is strictly more expressive than a `Vec` (just call `collect` and you get a vec) at the cost of being a concept that people may not be familiar with.

lib/runtime/README.md
MarkMcCaskey: That's a pretty interesting idea -- the problem is that this seems to remove a lot of the usefulness of DynFunc -- a big benefit of it is that it does not have a type signature which means you can just have `Vec<DynFunc>` and that you can operate on it without knowing or caring what the type signature is.  In the above example, we'd still need to know the type signature statically to be able to get it.

This sounds like something that will take more design work!

MarkMcCaskey: To clarify, it may be possible to implement this in a way to get that effect easily but I don't see any obvious solutions right away

syrusakbary: Sounds good. Let's handle that apart.

lib/runtime-core/src/table/mod.rs
```diff
@@ -18,12 +19,76 @@ pub use self::anyfunc::Anyfunc;
...
+
 /// Kind of table element.
+// note to implementors: all types in `Element` should implement `StorableInTable`.
 pub enum Element<'a> {
     /// Anyfunc.
     Anyfunc(Anyfunc<'a>),
 }
 
+// delegation implementation for `Element`
+impl<'a> StorableInTable for Element<'a> {
```
syrusakbary: Looking at the reference WebAssembly, Element should tend to be just a `Value` type.

lib/api/src/lib.rs
MarkMcCaskey: Hmm, I'll have to research it more but my experience with third-party error crates is that they provide a nice way to bundle all your errors together but it comes at the cost of making the errors essentially impossible to handle in code.  Perhaps `anyhow` has a solution for that, but I'd be hesitant to go behind basic `enums` and maybe some `structs` if we expect users to able to handle the errors.

lib/api/src/lib.rs
MarkMcCaskey: Sure, I agree that it's not really in-scope for this.  Though we may run into issues with usability if it has the same name as a concrete type that implements it.

lib/api/src/lib.rs
MarkMcCaskey: > It should have the fields module

traits cant' have fields but I can make methods that do that.  Reiterating I'm not a fan of duplicating the names when we'll have to export both -- I think that might end up forcing users to write code like `<wasmer::Module as wasmer::traits::Module>::method`

lib/runtime-core/src/table/mod.rs
MarkMcCaskey: Hmm, I was under the impression that that wasn't in the current version of Wasm -- I'm not really following, can you link to the location in the spec you're talking about?

MarkMcCaskey:
bors try

lib/runtime-core/src/module.rs
MarkMcCaskey: We don't currently have a `FuncDescriptor` and the original design doc didn't have any data on `Func` either, I'd be interested in revisiting this together

MarkMcCaskey:
bors try

lib/api/src/lib.rs
MarkMcCaskey: From https://github.com/dtolnay/anyhow#comparison-to-thiserror :

> Use Anyhow if you don't care what error type your functions return, you just want it to be easy. This is common in application code. Use thiserror if you are a library that wants to design your own dedicated error type(s) so that on failures the caller gets exactly the information that you choose.

This confirms my assumption that the library is for when you don't want to handle the errors and shouldn't be used in our public API.  We can use `thiserror` (by the same author) if we want a library to help us with error types, but I think it might be simpler to just be explicit with an enum.

lib/runtime-core/src/table/mod.rs
syrusakbary: https://github.com/WebAssembly/wasm-c-api/blob/master/include/wasm.hh#L344

lib/runtime-core/src/module.rs
MarkMcCaskey: Follow up: I spent some time trying to make `imports` return an iterator too and I think it's not possible on stable Rust without a decent amount of work, so that's another point in favor of using `Vec`, it's potentially more stable to large internal breakage because it's less flexible.

lib/runtime-core/src/table/mod.rs
MarkMcCaskey: Do we want to expose functionality in our API that is not yet implemented?  Here's the JS API as a counter point: https://webassembly.github.io/spec/js-api/#enumdef-tablekind

lib/api/src/lib.rs
MarkMcCaskey: I think the traits are out of scope for the current PR, as they won't live in the `api` crate.  I thought I commented that "traits can't have fields" before but I don't see it, so I'll say it again:

Traits can't have fields but we can have trait methods which do those things.

lib/api/src/lib.rs
MarkMcCaskey: It's difficult to do even if we explicitly list them because we're implicitly reexporting methods on the types we export 🤔 ... it looks like Rust doesn't warn about that, you can have methods that return errors that are not public via `api` but are public via runtime core...

This will be non-trivial to test and I really don't like it in general 🤔

MarkMcCaskey:
bors try

lib/runtime-core/src/instance.rs
```diff
@@ -242,6 +242,10 @@ impl Instance {
     /// # Ok(())
     /// # }
     /// ```
+    #[deprecated(
+        since = "0.17.0",
+        note = "Please use `let f: DynFunc = instance.exports.get(name)?; f.call(params)?;` instead"
+    )]
     pub fn call(&self, name: &str, params: &[Value]) -> CallResult<Vec<Value>> {
         let export_index =
             self.module
```
syrusakbary: Can we simplify this implementation, using the inner implementation instead?

```rust
let func: DynFunc = self.exports.get(&name)?;
func.call(&params)?
```

lib/api/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    fn from_binary(_bytes: impl AsRef<[u8]>) -> error::CompileResult<Module> {
+        todo!("from_binary: how is this different from `new`?")
+    }
+    fn from_binary_unchecked(_bytes: impl AsRef<[u8]>) -> error::CompileResult<Module> {
+        todo!("from_binary_unchecked")
+    }
+    fn from_file(_file: impl AsRef<std::path::Path>) -> error::CompileResult<Module> {
+        todo!("from_file");
```
syrusakbary: Can we implement those?

lib/api/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+    fn new(bytes: impl AsRef<[u8]>) -> error::CompileResult<Module> {
+        let bytes = bytes.as_ref();
+        wasmer_runtime_core::compile_with(bytes, &default_compiler())
+    }
+
+    fn from_binary(_bytes: impl AsRef<[u8]>) -> error::CompileResult<Module> {
+        todo!("from_binary: how is this different from `new`?")
+    }
+    fn from_binary_unchecked(_bytes: impl AsRef<[u8]>) -> error::CompileResult<Module> {
+        todo!("from_binary_unchecked")
```
syrusakbary: Can we implement this?

lib/api/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+
+// this implementation should be moved
+impl CompiledModule for Module {
+    fn new(bytes: impl AsRef<[u8]>) -> error::CompileResult<Module> {
+        let bytes = bytes.as_ref();
+        wasmer_runtime_core::compile_with(bytes, &default_compiler())
+    }
+
+    fn from_binary(_bytes: impl AsRef<[u8]>) -> error::CompileResult<Module> {
+        todo!("from_binary: how is this different from `new`?")
```
syrusakbary: Implementation missing :)

Note: `new` calls from_binary. But perhaps we want to add automatic support for converting text files into wasm (in `new` not in `from_binary`).

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/runtime-core/src/vm.rs
```diff
@@ -716,10 +716,19 @@ impl LocalGlobal {
 }
 
 /// Identifier for a function signature.
+///
+/// A transparent `SigIndex`
 #[derive(Debug, Clone, Copy)]
 #[repr(transparent)]
 pub struct SigId(pub u32);
```
syrusakbary: Should we replace all uses of `SigId` with `SigIndex` ?

lib/runtime-core/src/module.rs
```diff
@@ -164,6 +170,154 @@ impl Module {
...
+    }
+}
+
+/// Type describing an export that the [`Module`] provides.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ExportDescriptor<'a> {
+    /// The name identifying the export.
+    pub name: &'a str,
+    /// The type of the export.
+    pub ty: ExportType,
```
syrusakbary: Can we have here ty: `ExternType` ?

MarkMcCaskey:
bors try

lib/runtime-core/src/vm.rs
MarkMcCaskey: The fact that it's `transparent` may be important to its use.  I just added the comment because I get confused every time I see it.

syrusakbary: Is there any way to verify that we are making use of `transparent` (kind of: tests without it will pass?)

If tests pass, then we can just remove this type and use SigIndex directly.

MarkMcCaskey: It's stored by pointer and might be accessed by the generator code, having it not be transparent could mean that depending on the optimization level, for example, it might just corrupt the data that generated code sees.  It is possible to have tests for this, but it's probably not the simplest thing to test.  I don't really know how it's used, so maybe it's easier to test than I think it is.

MarkMcCaskey:
bors try-

PR #1315, opened 2020-03-18
Fix windows cross compilation

Jake-Shadle:
# Description
When compiling to windows from any host other than windows, the `win-exception-handler` crate would not actually compile the C code since the compilation was guarded by `#[cfg(target_os = "windows")]`, which is evaluated when the build script itself is compiled which would mean it would != windows, which would eventually cause a link error due to the missing library.

This change gets the target os via the `CARGO_CFG_TARGET_OS` environment variable, which is set by cargo when invoking the build script, and will evaluate to `windows` when actually targetting windows, no matter the host platform.

It also changed from using cmake to use the cc crate to compile the code, as cmake seemed like massive overkill for a single C file, as well as complicating cross compilation again, though I can change it back to cmake if you don't want that change.

I also fixed a warning produced by clang by adding `#include <intrin.h>`, though there was another warning due to the trampoline function pointer taking a non-const `wasmer_instance_context_t*`, but wasmer actually sending in a const *.

# Review

### DIFF
--- a/lib/win-exception-handler/build.rs
+++ b/lib/win-exception-handler/build.rs
@@ -7,4 +7,8 @@ fn main() {
         println!("cargo:rustc-link-search=native={}", dst.display());
         println!("cargo:rustc-link-lib=static={}", project_name);
     }
+    cc::Build::new()
+        .include("exception_handling")
+        .file("exception_handling/exception_handling.c")
+        .compile("exception_handling");
 }
--- a/lib/win-exception-handler/build.rs
+++ b/lib/win-exception-handler/build.rs
@@ -1,12 +1,8 @@
 fn main() {
-    #[cfg(target_os = "windows")]
-    {
-        use cmake::Config;
-        let project_name = "exception_handling";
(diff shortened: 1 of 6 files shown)


lib/win-exception-handler/build.rs
```diff
@@ -1,10 +1,10 @@
 fn main() {
-    #[cfg(target_os = "windows")]
-    {
-        use cmake::Config;
-        let project_name = "exception_handling";
-        let dst = Config::new(project_name).build();
-        println!("cargo:rustc-link-search=native={}", dst.display());
-        println!("cargo:rustc-link-lib=static={}", project_name);
```
MarkMcCaskey: Why are these lines deleted?  Surely, we still need to link the compiled output, or does `cc` somehow do that as well?

Jake-Shadle: Sorry, yes, cc automatically

1. Determines the compiler to use based on the environment
2. Specifies additional linker search directories
3. Additional libraries to link to

PR #1321, opened 2020-03-20
Create allocas for all stack operations.

nlewycky:
Next up, we need to use the same allocas for the same depth of stack across two sides of an if-else.

After that, remove PHI node creation.


Hywan:
Probably still relevant. Might be related to https://github.com/wasmerio/wasmer/issues/2235 (really not sure, must be confirmed).

PR #1329, opened 2020-03-24
feat(interface-types) New numbers and strings instructions

Hywan:
This PR does 3 things:

* Transforms the `x-to-y` number lowerings and liftings instructions to the new set of instructions `y.from_x`. It's just the opposite. It's more logical, but the diff is dense due to that. The `lowering_lifting` module is renamed `numbers`.
* Renames `memory-to-string` to `string.lift_memory`, and `string-to-memory` to `string.lower_memory`. It's also more logical, but the diff is dense due to that too. The `memory_to_string` and `string_to_module` modules are merged into the `strings` module.
* Introduce the new `string.size` instruction (in the `strings` module).

I think commits are clean and atomic enough, so it might be better to review it like that.

For binary encoding/decoding, the binary representation is absolutely entirely arbitrary (since the working notes define nothing, obviously).

Cf https://github.com/WebAssembly/interface-types/blob/master/proposals/interface-types/working-notes/Instructions.md


lib/interface-types/src/interpreter/instructions/strings.rs
```diff
@@ -0,0 +1,386 @@
...
+                .wasm_instance
+                .memory(memory_index as usize)
+                .ok_or_else(|| {
+                    InstructionError::new(
+                        instruction,
+                        InstructionErrorKind::MemoryIsMissing { memory_index },
+                    )
+                })?;
+
+            let pointer = to_native::<i32>(&inputs[0], instruction)? as usize;
```
MarkMcCaskey: Does the fact this is an `i32` cause a problem?  It seems that we can't access the upper 2gb of memory by casting from `i32` to `usize` because `-1i32 as usize` is `18446744073709551615`; `-1i32 as u32 as usize` is `4294967295` which seems closer to what we'd want here.

If this is a mistake, perhaps we should add some tests using the upper 2gb of memory to make sure everything is working as expected as this issue will likely be in a lot of places!

lib/interface-types/src/interpreter/instructions/strings.rs
```diff
@@ -0,0 +1,386 @@
...
+                Instruction::StringLiftMemory,
+            ],
+            invocation_inputs: [
+                InterfaceValue::I32(0),
+                //              ^^^^^^ pointer
+                InterfaceValue::I32(13),
+                //              ^^^^^^^ length
+            ],
+            instance: Instance {
+                memory: Memory::new("Hello, World!".as_bytes().iter().map(|u| Cell::new(*u)).collect()),
```
MarkMcCaskey: I'm not familiar with `Memory::new` actually, but it seems that we don't need to pass in `Cell` unless this is creating a memory that's the size of the string and using that memory directly.

What I assume is happening is that we're copying these bytes into the real memory so the fact these are wrapped in `Cell` shouldn't matter here as I understand it.  `Cell` is just important from a data creation perspective if that data can be modified without `&mut` but if we're copying it, then that's not a concern.

Hywan: In this case, `Memory` is just a test object, not the real `Memory` struct from `runtime-core` :-).

lib/interface-types/src/interpreter/instructions/strings.rs
Hywan: Hmm. The pointer is of kind `i32` in the spec. It is regular a Wasm core type `i32`. Our API needs a `usize`, hence the casting.

You're right we need to check it is positive. I'm opening an issue to focus on that.

PR #1332, opened 2020-03-24
Add `CompilerConfig` opt to disable IR verification in debug mode

### LINKED ISSUE #1330: Allow disabling cranelift verifier explicitely
### Motivation

At [CosmWasm](https://github.com/CosmWasm/cosmwasm) we use debug builds of wasmer+cranelift for integration testing smart contracts that have been compiled to Wasm files before. Due to [a change in the cranelift backend with debug assertions](https://github.com/wasmerio/wasmer/commit/5e3147db0882e26db70c962d3e3cd0b62d8281cb#diff-02f9535bb166cb207b858de844a0c257R46), those tests now run significantly slower than before. Cranelift is our default backend for development purposes since it compiles with Rust stable. In order to increase the experience for smart contract developers, it would be great to be able to disable verifier independent of the build mode.

### Proposed solution

There are two solutions that come to my mind:
1. Reconsider the decision to enable the verifyier by default whenever Wasmer is build with debug assertions. Maybe it makes sense to allow enabling it explicitely.
2. Create a flag that allows disabling the verifier, which overrides the defaults

### Alternatives

It turns out that is it [possible to override](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides) `wasmer-clif-backend`'s build settings via:

```
[profile.dev.package.wasmer-clif-backend]
debug-assertions = false
opt-level = 1
```

However, this assumes the cranelift specific verifier is always bound to `cfg!(debug_assertions)`.

Also this override seems to be doable in the final project only, and the package `cosmwasm-vm` cannot set this for all of its users.

### Additional context

- Discussion in CosmWasm: https://github.com/CosmWasm/cosmwasm/issues/155
- Example integration test: [VM](https://github.com/CosmWasm/cosmwasm/blob/v0.7.2/packages/vm/Cargo.toml#L14-L39), [smart contract](https://github.com/CosmWasm/cosmwasm/blob/v0.7.2/contracts/hackatom/Cargo.toml#L23-L38), [integration test](https://github.com/CosmWasm/cosmwasm/blob/v0.7.2/contracts/hackatom/tests/integration.rs), [build and run command](https://github.com/CosmWasm/cosmwasm/blob/v0.7.2/contracts/hackatom/.cargo/config#L4)

MarkMcCaskey:
Resolves #1330 

# Review

### DIFF
--- a/lib/clif-backend/src/lib.rs
+++ b/lib/clif-backend/src/lib.rs
@@ -49,13 +49,15 @@ fn get_isa(config: Option<&CompilerConfig>) -> Box<dyn isa::TargetIsa> {
             if config.nan_canonicalization {
                 builder.set("enable_nan_canonicalization", "true").unwrap();
             }
-            enable_verifier = !config.disable_debug_mode_verification;
+            enable_verifier = config.enable_verification;
         } else {
             // Set defaults if no config found.
-            enable_verifier = true;
+            // NOTE: cfg(test) probably does nothing when not running `cargo test`
+            //       on this crate
+            enable_verifier = cfg!(test) || cfg!(debug_assertions);
         }
 
-        if (cfg!(test) || cfg!(debug_assertions)) && enable_verifier {
+        if enable_verifier {
             builder.set("enable_verifier", "true").unwrap();
         } else {
             builder.set("enable_verifier", "false").unwrap();
--- a/lib/clif-backend/src/lib.rs
+++ b/lib/clif-backend/src/lib.rs
@@ -57,11 +57,12 @@ fn get_isa(config: Option<&CompilerConfig>) -> Box<dyn isa::TargetIsa> {
             enable_verifier = cfg!(test) || cfg!(debug_assertions);
         }
 
-        if enable_verifier {
-            builder.set("enable_verifier", "true").unwrap();
-        } else {
-            builder.set("enable_verifier", "false").unwrap();
-        }
+        builder
+            .set(
+                "enable_verifier",
+                if enable_verifier { "true" } else { "false" },
+            )
+            .unwrap();
 
         let flags = settings::Flags::new(builder);
         debug_assert_eq!(flags.opt_level(), settings::OptLevel::SpeedAndSize);
--- a/lib/clif-backend/src/lib.rs
+++ b/lib/clif-backend/src/lib.rs
@@ -43,16 +43,22 @@ fn get_isa(config: Option<&CompilerConfig>) -> Box<dyn isa::TargetIsa> {
         builder.set("opt_level", "speed_and_size").unwrap();
         builder.set("enable_jump_tables", "false").unwrap();
 
-        if cfg!(test) || cfg!(debug_assertions) {
-            builder.set("enable_verifier", "true").unwrap();
-        } else {
-            builder.set("enable_verifier", "false").unwrap();
-        }
+        let enable_verifier: bool;
 
         if let Some(config) = config {
             if config.nan_canonicalization {
                 builder.set("enable_nan_canonicalization", "true").unwrap();
             }
+            enable_verifier = !config.disable_debug_mode_verification;
+        } else {
+            // Set defaults if no config found.
+            enable_verifier = true;
+        }
+
+        if (cfg!(test) || cfg!(debug_assertions)) && enable_verifier {
+            builder.set("enable_verifier", "true").unwrap();
+        } else {
+            builder.set("enable_verifier", "false").unwrap();
         }
 
         let flags = settings::Flags::new(builder);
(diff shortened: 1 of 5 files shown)


lib/clif-backend/src/lib.rs
```diff
@@ -43,16 +43,22 @@ fn get_isa(config: Option<&CompilerConfig>) -> Box<dyn isa::TargetIsa> {
...
+        let enable_verifier: bool;
 
         if let Some(config) = config {
             if config.nan_canonicalization {
                 builder.set("enable_nan_canonicalization", "true").unwrap();
             }
+            enable_verifier = !config.disable_debug_mode_verification;
+        } else {
+            // Set defaults if no config found.
+            enable_verifier = true;
```
nlewycky: How about `enable_verifier = cfg!(test) || cfg!(debug_assertions)`?

lib/runtime-core/src/backend.rs
```diff
@@ -136,6 +136,12 @@ pub struct CompilerConfig {
     /// Enabling this makes execution deterministic but increases runtime overhead.
     pub nan_canonicalization: bool,
 
+    /// Turns off verification that is done by default when `debug_assertions` are enabled
+    /// (for example in 'debug' builds). Enabling this flag will make compilation faster at the
+    /// cost of not detecting bugs in the compiler. The verification steps that this flag
+    /// disables are disabled by default in 'release' builds.
+    pub disable_debug_mode_verification: bool,
```
nlewycky: Is there anything stopping us from turning verification on in a release build? Say, for users who are prepared to pay for extra assurance? If we can do that, could we make this control the setting unilaterally, and also remove the negative?

syrusakbary: Yup, agreed with Nick. Since debug verification is probably always false, we can do something like:

```
    pub enable_ir_verification: bool,
```

(note: this can work for both Cranelift and LLVM, since they both have IR verifiers).

nlewycky: I'd like to rename it to something like "extra_safety_checks" or "backend_internal_verification" or something along those lines. There's nothing special about clif/llvm here, singlepass could have its own safety checks too, and it doesn't have an IR.

syrusakbary: What about `enable_internal_verification` or `enable_verification` (I kind of like keeping backend out, since this might be a setting on the backend config itself?)

lib/clif-backend/src/lib.rs
MarkMcCaskey: That would require duplicating that boolean expression because it needs to be checked on the other case too!  I considered it but decided against it

lib/runtime-core/src/backend.rs
MarkMcCaskey: It's phrased in the negative because we `derive(Default)` which means it defaults to false haha.  I can make a manual implementation of Default though

lib/clif-backend/src/lib.rs
```diff
@@ -43,16 +43,24 @@ fn get_isa(config: Option<&CompilerConfig>) -> Box<dyn isa::TargetIsa> {
...
             }
+            enable_verifier = config.enable_verification;
+        } else {
+            // Set defaults if no config found.
+            // NOTE: cfg(test) probably does nothing when not running `cargo test`
+            //       on this crate
+            enable_verifier = cfg!(test) || cfg!(debug_assertions);
+        }
+
+        if enable_verifier {
```
nlewycky: Optional.
:golf: `builder.set("enable_verifier", if enable_verifier { "true" } else { "false" }).unwrap();`

lib/runtime-core/src/backend.rs
```diff
@@ -148,6 +155,30 @@ pub struct CompilerConfig {
     pub generate_debug_info: bool,
 }
 
+impl Default for CompilerConfig {
+    fn default() -> Self {
+        Self {
+            symbol_map: Default::default(),
```
nlewycky: Couldn't you use `..Default::default()` for these?

MarkMcCaskey: That would infinite loop I believe!  My understanding of what `..Default::default()` does is that it calls default on the type you're constructing and then overrides the fields you specified.  Because we're defining default, I don't believe we can use that syntax here.

MarkMcCaskey: Yeah, just verified this:

```
warning: function cannot return without recursing
   --> lib/runtime-core/src/backend.rs:159:5
    |
159 |     fn default() -> Self {
    |     ^^^^^^^^^^^^^^^^^^^^ cannot return without recursing
...
178 |             ..Default::default()
    |               ------------------ recursive call site
    |
    = note: `#[warn(unconditional_recursion)]` on by default
    = help: a `loop` may express intention better if this is on purpose
```

MarkMcCaskey:
bors retry

PR #1335, opened 2020-03-26
 fix(runtime-c-api) Change mutability of `memory` to`const` in `wasmer_memory_data_length`

### LINKED ISSUE #1314: Inconsistent use of `const` in memory pointer in C API
Reported in https://github.com/wasmerio/docs.wasmer.io/issues/43.

```
  const wasmer_memory_t *memory = wasmer_instance_context_memory(ctx, 0);
  wasmer_memory_data_length(memory);
```

Causes a warning on Clang (OSX) and is possibly an error with GCC on Linux.  `wasmer_instance_context_memory` returns a `const wasmer_memory_t*` but `wasmer_memory_data_length` wants a `wasmer_memory_t*`.

This bug is affecting our documentation examples, ideally we should be testing these in CI on both sides.  See https://github.com/wasmerio/docs.wasmer.io/issues/46 for the issue on fixing CI on the docs repo side.

Hywan:
This PR changes mutability of `memory` for `const` in `wasmer_memory_data_length` to be more consistent regarding the returned value of `wasmer_instance_context_memory`.

Fixes #1314. 
Fixes https://github.com/wasmerio/docs.wasmer.io/issues/43.


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## **[Unreleased]**
 
+- [#1335](https://github.com/wasmerio/wasmer/pull/1335) Change mutability of `memory` for `const` in `wasmer_memory_data_length` in the C API
```
MarkMcCaskey: Minor nit, I think "`memory` for `const`` should be "`memory` to `const`"

Hywan: Thanks!

PR #1337, opened 2020-03-26
feat(interface-types) Better handling of i32 to usize casts

### LINKED ISSUE #1334: fix(interface-types) Casting `i32` to `usize` may cause problems
In WIT instructions, we cast `i32` to `usize`. It may cause problems, as noticed in https://github.com/wasmerio/wasmer/pull/1329/files#r397396580. We must handle those cases.

Hywan:
Fix #1334 

This PR introduces a new `NegativeValue` error.
This PR fixes some `usize` and `u32` types for indexes (small changes).
Finally, this PR casts `i32` to `usize` by using the `TryFrom` implementation. That way:

```rust
let pointer = to_native::<i32>(&inputs[0], instruction)? as usize;
```

becomes:

```rust
let pointer: usize = to_native::<i32>(&inputs[0], instruction)?.try_into().map_err(|_| {
    InstructionError::new(
        instruction,
        InstructionErrorKind::NegativeValue { subject: "pointer" },
    )
})?;
```

It's more verbose, but it handles the casting correctly.

Note: Maybe we should implement `From<TryFromIntError>` for `InstructionErrorKind`? What do you think?

Edit: With `From<TryFromIntError>`, it looks like this:

```rust
let pointer: usize = to_native::<i32>(&inputs[0], instruction)? // convert from WIT to native `i32`
    .try_into() // convert to `usize`
    .map_err(|e| (e, "pointer").into()) // convert the `TryFromIntError` to `InstructionErrorKind::NegativeValue`
    .map_err(|k| InstructionError::new(instruction, k))?; // convert to an `InstructionError`
```

It is indeed simpler. Keeping things like this.

### DIFF
--- a/lib/interface-types/src/decoders/binary.rs
+++ b/lib/interface-types/src/decoders/binary.rs
@@ -169,7 +169,7 @@ fn instruction<'input, E: ParseError<&'input [u8]>>(
             (
                 input,
                 Instruction::CallCore {
-                    function_index: argument_0 as usize,
+                    function_index: argument_0 as u32,
                 },
             )
         }
--- a/lib/interface-types/src/decoders/wat.rs
+++ b/lib/interface-types/src/decoders/wat.rs
@@ -146,7 +146,7 @@ impl<'a> Parse<'a> for Instruction {
             parser.parse::<keyword::call_core>()?;
 
             Ok(Instruction::CallCore {
-                function_index: parser.parse::<u64>()? as usize,
+                function_index: parser.parse::<u32>()?,
             })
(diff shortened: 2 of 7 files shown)


syrusakbary:
> Note: Maybe we should implement From<TryFromIntError> for InstructionErrorKind? What do you think?

Not sure, I really think we should do whatever is easier to implement/maintain for now :) (perhaps less code is better)

MarkMcCaskey:
bors merge

MarkMcCaskey:
bors retry

PR #1343, opened 2020-03-30
update blake3 to version 0.3.0

oconnor663:
Apologies for showing up again so soon :)

Version 0.3 switched to using assembly implementations by default, which
both perform better and build more quickly. This invokes the C compiler
from build.rs. If Wasmer would prefer to avoid that build dependency, we
could enable the new `pure` feature.

PR #1350, opened 2020-04-02
update blake3 to 0.3.1

oconnor663:
Version 0.3.0 caused problems because it required a C compiler with
AVX-512 support, which broke Android x86 cross-compilation. Version
0.3.1 automatically falls back to a pure Rust build when the C compiler
either doesn't exist or doesn't support the flags we need.


oconnor663:
(The previous attempt was https://github.com/wasmerio/wasmer/pull/1343.)

PR #1358, opened 2020-04-07
Update C API to use new API

MarkMcCaskey:
# Review


MarkMcCaskey:
bors try

lib/api/src/lib.rs
```diff
@@ -217,3 +222,23 @@ impl CompiledModule for Module {
...
     }
 }
+
+// Below this line is things copied from `wasmer-runtime` to make the C API work.
+// All these additions should be reviewed carefully before shipping.
+
+/// Compile WebAssembly binary code into a [`Module`].
+/// This function is useful if it is necessary to
+/// compile a module before it can be instantiated
+/// (otherwise, the [`instantiate`] function should be used).
```
syrusakbary, line 232:
```suggestion
```

lib/api/src/lib.rs
```diff
@@ -217,3 +222,23 @@ impl CompiledModule for Module {
...
+///
+/// [`Module`]: struct.Module.html
+/// [`instantiate`]: fn.instantiate.html
```
syrusakbary, line 235: I thought Rust docs did that automatically?

lib/runtime-c-api/src/import/mod.rs
```diff
@@ -17,12 +17,11 @@ use std::{
     ptr, slice,
     sync::Arc,
 };
-use wasmer_runtime::{Ctx, Global, Memory, Module, Table};
+use wasmer::import::{ImportObject, ImportObjectIterator};
+use wasmer::wasm::{Ctx, FuncSig, Global, Memory, Module, Table, Type};
 use wasmer_runtime_core::{
     export::{Context, Export, FuncPointer},
```
syrusakbary, line 23: Perhaps we can reuse the `ExportType` or `ExternType` for Export?

lib/api/src/lib.rs
MarkMcCaskey: I believe it's a nightly rustdoc thing; docs.rs uses it by default.  But it relies on it being in scope; either way that doc comment was just copied and pasted from runtime

Hywan: It's called intra-doc link resolution. It's a nightly only feature. Also, we may want to add:

```rust
![deny(missing_docs, intra_doc_link_resolution_failure)]
```

everywhere. It protects again broken links.

docs.rs uses nighty to generate the doc, by the way.

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/runtime-c-api/src/instance.rs
```diff
@@ -10,11 +10,9 @@ use crate::{
...
 use libc::{c_char, c_int, c_void};
 use std::{collections::HashMap, ffi::CStr, ptr, slice};
-use wasmer_runtime::{Ctx, Global, Instance, Memory, Table, Value};
-use wasmer_runtime_core::{
-    export::Export,
-    import::{ImportObject, Namespace},
-};
+use wasmer::import::{ImportObject, Namespace};
+use wasmer::wasm::{Ctx, Global, Instance, Memory, Table, Value};
+use wasmer_runtime_core::export::Export;
```
syrusakbary, line 15: Why export is still needed?

MarkMcCaskey: This is the Instance export, the one that is executable, which is different from a module export which is not:

I actually reexported `Export` in the tests PR that just landed so we don't have to depend on runtime-core for this anymore

MarkMcCaskey:
bors try

syrusakbary:
I commented on the Wasm-C-API repo what we were discussing to get more thoughts on the nomenclature:

https://github.com/WebAssembly/wasm-c-api/issues/135

syrusakbary:
Note for readers: in future PRs we will adopt the *Type nomenclature, moving away from *Descriptor.

PR #1362, opened 2020-04-07
feat(interface-types) Remove allocator index from `string.lower_memory`

Hywan:
This PR updates `string.lower_memory` to remove the allocator index. Indeed, the string pointer is assumed to be present on the stack.
Also, this PR updates `string.size` to pop, and not to peek, the string to compute the length from.

That way, it matches the WIT proposal.


Hywan:
Also, I noticed that the string to be lowered must be loaded twice on the stack. I've opened a discussion here, https://github.com/WebAssembly/interface-types/issues/107.

Hywan:
This PR respects the actual proposal, not the discussion. I prefer to match the current state of the proposal for the moment.

PR #1366, opened 2020-04-08
Reuse compile-with-config, add read_custom_sections config field

### LINKED ISSUE #1359: Why is there a custom section in compile_with but not in compile_with_config
### Summary

The documentation of `compile_with_config` says, _The same as `compile_with` but changes the compiler behavior with the values in the `CompilerConfig`_. However, the code seems to have a non-trivial difference in `import_custom_sections`:

```rust
/// Compile a [`Module`] using the provided compiler from
/// WebAssembly binary code. This function is useful if it
/// is necessary to a compile a module before it can be instantiated
/// and must be used if you wish to use a different backend from the default.
pub fn compile_with(
    wasm: &[u8],
    compiler: &dyn backend::Compiler,
) -> CompileResult<module::Module> {
    let token = backend::Token::generate();
    compiler
        .compile(wasm, Default::default(), token)
        .map(|mut inner| {
            let inner_info: &mut crate::module::ModuleInfo = &mut inner.info;
            inner_info.import_custom_sections(wasm).unwrap();
            module::Module::new(Arc::new(inner))
        })
}

/// The same as `compile_with` but changes the compiler behavior
/// with the values in the `CompilerConfig`
pub fn compile_with_config(
    wasm: &[u8],
    compiler: &dyn backend::Compiler,
    compiler_config: backend::CompilerConfig,
) -> CompileResult<module::Module> {
    let token = backend::Token::generate();
    compiler
        .compile(wasm, compiler_config, token)
        .map(|inner| module::Module::new(Arc::new(inner)))
}
```

Wouldn't it be possible to implement `compile_with` via `compile_with_config` to ensure they only differ in the one documented aspect?

### Additional details

Reading code because I try to configure the default memory `memory(0)`.

syrusakbary:
# Description

This PR fixes #1359 and also adds a `read_custom_sections` field to the BackendConfig (so we pay no default overhead when reading custom sections).

# Review

### DIFF
--- a/lib/runtime-core/src/backend.rs
+++ b/lib/runtime-core/src/backend.rs
@@ -143,6 +143,11 @@ pub struct CompilerConfig {
     /// These verifications are disabled by default in 'release' builds.
     pub enable_verification: bool,
 
+    /// Should we automatically attach the custom sections to the Module?
+    ///
+    /// It's off by default as it takes more cpu and memory to read and store the sections.
+    pub read_custom_sections: bool,
+
     pub features: Features,
 
     // Target info. Presently only supported by LLVM.
@@ -164,6 +169,7 @@ impl Default for CompilerConfig {
             track_state: Default::default(),
             full_preemption: Default::default(),
             nan_canonicalization: Default::default(),
+            read_custom_sections: Default::default(),
             features: Default::default(),
             triple: Default::default(),
             cpu_name: Default::default(),
--- a/lib/runtime-core/src/lib.rs
+++ b/lib/runtime-core/src/lib.rs
@@ -122,10 +122,11 @@ pub fn compile_with_config(
     compiler_config: backend::CompilerConfig,
 ) -> CompileResult<module::Module> {
     let token = backend::Token::generate();
+    let read_custom_sections = compiler_config.read_custom_sections;
     compiler
         .compile(wasm, compiler_config, token)
         .map(|mut inner| {
-            if compiler_config.read_custom_sections {
+            if read_custom_sections {
                 let inner_info: &mut crate::module::ModuleInfo = &mut inner.info;
                 inner_info.import_custom_sections(wasm).unwrap();
             }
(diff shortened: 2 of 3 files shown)


lib/runtime-core/src/lib.rs
```diff
@@ -111,14 +111,7 @@ pub fn compile_with(
...
 ) -> CompileResult<module::Module> {
-    let token = backend::Token::generate();
-    compiler
-        .compile(wasm, Default::default(), token)
-        .map(|mut inner| {
-            let inner_info: &mut crate::module::ModuleInfo = &mut inner.info;
-            inner_info.import_custom_sections(wasm).unwrap();
-            module::Module::new(Arc::new(inner))
-        })
+    compile_with_config(wasm, compiler, Default::default())
```
webmaster128, line 114: ❤️

lib/runtime-core/src/backend.rs
```diff
@@ -143,6 +143,11 @@ pub struct CompilerConfig {
     /// These verifications are disabled by default in 'release' builds.
     pub enable_verification: bool,
 
+    /// Should we automatically attach the custom sections to the Module?
+    ///
+    /// It's off by default as it takes more cpu and memory to read and store the sections.
```
Hywan, line 148: Well, I would argue that it should be `on` by default.
We should offer everything to the user by default, and then it's up to the user to disable things according to its context and constraints.

MarkMcCaskey: I agree, plus that way is a simpler change.  Syrus and I discussed this yesterday and it seems for now we're going to have that overhead but that an ideal design removes the overhead even when we parse custom sections.  All we really need to do if we keep the Wasm module in memory, is store offsets into it for the custom sections, requiring no copying or real overhead (not even a double parse, as we do now, if our planned improvement of our parsing logic goes through)

syrusakbary:
This API is deprecated in new Wasmer `master` thanks to latest refactor that just landed a few days ago. Closing PR!

PR #1370, opened 2020-04-09
feat(interface-types) Crate reorganization

Hywan:
This PR re-organizes the crate:

* `interpreter/wasm/values.rs` moves to `values.rs`,
* `interpreter/wasm/serde/*` moves to `serde/*`,
* `types.rs` is new, and includes `InterfaceType` and `RecordType`,
* Update the documentation.

Bonus: `InterfaceValue::Record`  now takes a `Vec1` rather than a `Vec` to match `InterfaceType::Record`.


lib/interface-types/src/serde/ser.rs
```diff
@@ -432,7 +441,9 @@ impl<'a> ser::SerializeStruct for &'a mut Serializer {
     }
 
     fn end(self) -> Result<Self::Ok, Self::Error> {
-        let record = InterfaceValue::Record(self.pop()?);
+        let record = InterfaceValue::Record(
+            Vec1::new(self.pop()?).map_err(|_| Self::Error::RecordNeedsAtLeastOneField)?,
```
MarkMcCaskey, line 445: Could also use something like `TryFrom` here which returns the correct error automatically

Hywan: Yes, that's a good idea! I'll do that in another PR, so that this one isn't blocked.

PR #1371, opened 2020-04-09
Update tests to use new API where easily possible

MarkMcCaskey:
The primary thing that wasn't updated yet is the compile and compile_with... functions.

### DIFF
--- a/lib/api/src/lib.rs
+++ b/lib/api/src/lib.rs
@@ -58,7 +58,7 @@ pub mod memory {
 }
 
 pub mod wasm {
-    //! Various types exposed by the Wasmer Runtime.
+    //! Various types exposed by the Wasmer Runtime relating to Wasm.
     //!
     //! TODO: Add index with links to sub sections
     //!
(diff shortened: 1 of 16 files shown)


MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/api/src/lib.rs
```diff
@@ -58,23 +62,214 @@ pub mod memory {
...
+
+            #[cfg(any(feature = "default-backend-cranelift", feature = "docs"))]
+            return Backend::Cranelift;
+
+            #[cfg(all(feature = "default-backend-llvm", not(feature = "docs")))]
+            return Backend::LLVM;
+        }
+    }
+
+    impl std::str::FromStr for Backend {
```
syrusakbary, line 152: Nice touch!

MarkMcCaskey: It's copied and pasted from `wasmer_runtime`, but I agree!  It's very nice!

PR #1382, opened 2020-04-15
Test Refactor

syrusakbary:
# Description

This PR is the continuation of #1380.
It refactors our testing infrastructure based on various points:
* There is no longer a "default compiler" on tests
* Tests are automatically generated for: emscripten and wasi
* It accelerates testing into multiple threads when possible (with the exception of the llvm backend)
* It automatically detects the backends available in the host

Summary of changes:
* [x] Removed all extra logic for test creation in emscripten test generator
* [x] Removed all extra logic for test creation in the wasi test generator
* [x] Divided wasmer wast test in different files (under `tests/custom/`)
* [x] Refactored trap asserts using the wast format
* [x] Improved the build script by adding emscripten and wasi test generators (separated from build)
* [x] Improved WASI errors to be Rust error types (by deriving from `thiserror::Error`)
* [x] Fixed toolchain error creation
* [x] Removed leaking of testing logic from generators into build
* [x] Refactored import tests to be compatible with multiple backends
* [x] Improved wasmer binary to work even when no backends are set (useful for testing without any backend available).
* [x] Removed assumption that Cranelift will be the default in the `wasmer` binary
* [x] Migrated middleware

# Review


lib/wasi/src/state/builder.rs
```diff
@@ -56,15 +57,23 @@ impl std::fmt::Debug for WasiStateBuilder {
 }
 
 /// Error type returned when bad data is given to [`WasiStateBuilder`].
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Error, Debug, PartialEq, Eq)]
```
syrusakbary, line 60: Note, this is needed because `anyhow` (used in the wasi tests) needs to receive an Rust error, and it throw me a compiler error when trying to map our errors to anyhow.

So it's actually something that makes our API better as well!
I prefered to do it in one stand, rather than in a PR apart because otherwise I will forget.

build.rs
```diff
@@ -41,71 +45,199 @@ fn wast_processor(out: &mut Testsuite, p: PathBuf) -> Option<Test> {
     })
 }
 
+/// Given a Testsuite and a path, process the path in case is a Emscripten
+/// wasm file.
+fn emscripten_processor(out: &mut Testsuite, p: PathBuf) -> Option<Test> {
```
syrusakbary, line 50: Note: I have to move this processors out of build

Cargo.toml
```diff
@@ -76,21 +80,25 @@ rustc_version = "0.2"
...
-default = ["fast-tests", "wasi", "backend-cranelift", "wabt"]
+# Don't add the backend features in default, please add them on the Makefile
+# since we might want to autoconfigure them depending on the availability on the host.
+default = ["wasi", "wabt"]
 "loader-kernel" = ["wasmer-kernel-loader"]
 debug = ["fern", "log/max_level_debug", "log/release_max_level_debug"]
 trace = ["fern", "log/max_level_trace", "log/release_max_level_trace"]
 docs = ["wasmer-runtime/docs"]
-# This feature will allow cargo test to run much faster
-fast-tests = []
```
syrusakbary, line 93: This feature was not used anywhere, hence the removal :)

lib/runtime/src/lib.rs
```diff
@@ -291,7 +298,7 @@ pub fn instantiate(wasm: &[u8], import_object: &ImportObject) -> error::Result<I
 /// The output of this function can be controlled by the mutually
 /// exclusive `default-backend-llvm`, `default-backend-singlepass`,
 /// and `default-backend-cranelift` feature flags.
-pub fn default_compiler() -> impl Compiler {
+pub fn default_compiler() -> Box<dyn Compiler> {
```
MarkMcCaskey, line 301: What's the reason this is now a dynamic trait object?

MarkMcCaskey: I thought it might be because the compiler can't be statically known anymore, but that doesn't seem to be the case

Cargo.toml
```diff
@@ -76,21 +80,25 @@ rustc_version = "0.2"
 [dev-dependencies]
 anyhow = "1.0.19"
 wasmer-wast = { path = "tests/wast" }
+lazy_static = "1.4.0"
```
MarkMcCaskey, line 83: If we avoid patch versions when it doesn't matter to us, we can reduce build times by avoiding duplicated dependencies

build.rs
```diff
@@ -41,71 +45,199 @@ fn wast_processor(out: &mut Testsuite, p: PathBuf) -> Option<Test> {
...
+    );
+
+    Some(Test {
+        name: testname.to_string(),
+        body: body.to_string(),
+    })
+}
+
+fn is_truthy_env(name: &str) -> bool {
+    env::var(name).unwrap_or("0".to_string()) == "1"
```
MarkMcCaskey, line 129: ⛳ 
`env::var(name).ok().as_deref().unwrap_or("0") == "1"` avoids the extra allocation so does `env::var(name).map(|n| n == "1").unwrap_or_default()`

Alternatively `env::var(name).unwrap_or_else(|| "0".to_string()) == "1"` avoids the extra allocation on the happy path but not the unhappy path.

Super minor detail that doesn't matter here -- this is more of a general style note.

tests/generate-wasi-tests/src/set_up_toolchain.rs
```diff
@@ -12,19 +12,22 @@ fn install_toolchain(toolchain_name: &str) {
...
+
+    println!("Installing rustup WASI target");
+    let rustup_out = Command::new("rustup")
+        .arg(format!("+{}", toolchain_name))
+        .arg("target")
+        .arg("add")
+        .arg("wasm32-wasi")
+        .output()
+        .expect("Failed to wasi target in Rust toolchain");
+    util::print_info_on_error(&rustup_out, "WASI TARGET IN TOOLCHAIN INSTAL FAILED");
```
nlewycky, line 24:
```suggestion
    util::print_info_on_error(&rustup_out, "WASI TARGET IN TOOLCHAIN INSTALL FAILED");
```

build.rs
```diff
@@ -41,71 +45,199 @@ fn wast_processor(out: &mut Testsuite, p: PathBuf) -> Option<Test> {
...
+        let mut out_ext = p.clone();
+        out_ext.set_extension("out");
+        // let mut txt_ext = p.clone();
+        // txt_ext.set_extension("txt");
+        if out_ext.exists() {
+            out_ext
+        }
+        // else if txt_ext.exists() {
+        //     txt_ext
+        // }
```
MarkMcCaskey, line 107: looks like some debug code got left over here

build.rs
MarkMcCaskey, line 50: I'm still not a fan of including specific logic in a workspace build.rs like this:

It's a relatively minor point and as our build.rs is currently only doing 1 thing the only cost is increasing the amount of visible complexity in the codebase.  And my real concern is about setting a bad example for future editors: ideally the correct thing (keeping test generation and other compile-time logic separate) should be easy, otherwise people won't do it.

So not the biggest deal, but I do think this does increase the chance of having a tangled mess that we'll have to untangle in the future (it's not at all right now, but after N more features get added it might be).

Makefile
```diff
@@ -1,5 +1,45 @@
...
+backends :=
+
+ifeq ($(ARCH), x86_64)
+  # In X64, Cranelift is enabled
+  backends += cranelift
+  # LLVM is enabled if not in Windows
+  ifneq ($(OS), Windows_NT)
+	LLVM_VERSION := $(shell llvm-config --version)
+	# If findstring is not empty, then it have found the value
+    ifneq (, $(findstring 9,$(LLVM_VERSION)))
```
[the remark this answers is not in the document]
syrusakbary: Are we using LLVM 10 in `master` ?

build.rs
```diff
@@ -41,71 +45,199 @@ fn wast_processor(out: &mut Testsuite, p: PathBuf) -> Option<Test> {
...
+    let ext = p.extension()?;
+    // Only look at wast files.
+    if ext != "wasm" {
+        return None;
+    }
+
+    let outfile = {
+        let mut out_ext = p.clone();
+        out_ext.set_extension("out");
+        // let mut txt_ext = p.clone();
```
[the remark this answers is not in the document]
syrusakbary:
```suggestion
```

lib/runtime/src/lib.rs
syrusakbary: The main reason for this is because I wanted `runtime` to be compilable even when no backends where set (via the `backend-` feature).

However, this caused a compilation error in Rust, mentioning that it didn't know the size of Compiler (since for Rust this looked like an empty function).
I tried to resolve it by doing `panic!` at the end of the function body, but the compilation error still happened, so I changed into a Box<dyn Compiler>` that Rust liked it.

build.rs
syrusakbary:
```suggestion
env::var(name).map(|n| n == "1").unwrap_or_default()
```

build.rs
syrusakbary: Yeah, I will move the processors back to the tests.
The general body of `main` in build.rs I think it's already minimal, but descriptive, so I think we can take out almost everything out of main.

Makefile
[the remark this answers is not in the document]
nlewycky: Oh! We are not. Still waiting on an inkwell crate.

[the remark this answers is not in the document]
syrusakbary: Is 9 the LLVM version we are using in master, or is 8?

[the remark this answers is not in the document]
nlewycky: It's still 8.

lib/runtime/src/lib.rs
```diff
@@ -320,7 +327,19 @@ pub fn default_compiler() -> impl Compiler {
...
 
-    DefaultCompiler::new()
+    #[cfg(any(
+        feature = "default-backend-singlepass",
+        feature = "default-backend-cranelift",
+        feature = "default-backend-llvm",
+    ))]
+    return Box::new(DefaultCompiler::new());
+
+    #[cfg(not(any(
```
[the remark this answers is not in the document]
syrusakbary: I tried your suggestion but it turns out `DefaultCompiler` will be not defined if the return statement if doesn't have the `#[cfg(...)]` the line before, and causes Rust compilation to fail.

build.rs
syrusakbary: Perhaps there are other ways of doing this, let me know if you have ideas @MarkMcCaskey!

lib/runtime/src/lib.rs
syrusakbary: Perhaps there are other ways of doing this, let me know if you have ideas @MarkMcCaskey!

build.rs
syrusakbary: Fixed!

syrusakbary:
bors try

lib/runtime/src/lib.rs
MarkMcCaskey: From an efficiency perspective, this doesn't matter at all, but there are some usability concerns.

If users can interact with compilers outside of the `Compiler` trait, then `impl Compiler` is strictly better because it's more direct:

```rust
let compiler: LLVMCompiler = default_compiler();
compiler.llvm_compiler_specific_method();
```

Whereas with `Box<dyn Compiler>` the user has to do some kind of downcasting (I'm not sure how to do this without looking at docs, but I believe it's possible and will require an `unwrap`).

Also to avoid API breaking it might be worth it to make it work...

One way to do this is to add a Fail compiler which panics on every method... though that really raises the question, in what scenario will a user be compiling with no backends and be calling `default_compiler`?  The only scenario I can think of right now is if the user is making a mistake: if that's the case, then perhaps we want to just remove the `default_compiler` function if no compiler backends are enabled 🤔

tests/test-generator/src/processors.rs
```diff
@@ -0,0 +1,101 @@
...
+/// Given a Testsuite and a path, process the path in case is a wast
+/// file.
+pub fn wast_processor(out: &mut Testsuite, p: PathBuf) -> Option<Test> {
+    let ext = p.extension()?;
+    // Only look at wast files.
+    if ext != "wast" {
+        return None;
+    }
+
+    // Ignore files starting with `.`, which could be editor temporary files
```
MarkMcCaskey, line 14: There are crates that do file globbing that respect `.gitignores` I think 🤔 , might be simpler

tests/test-generator/src/processors.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    let testname = extract_name(&p);
+    let compiler = out.path.get(0).unwrap();
+
+    // The implementation of `run_wast` lives in /tests/spectest.rs
+    let body = format!("crate::run_wast(r#\"{}\"#, \"{}\")", p.display(), compiler);
+
+    Some(Test {
+        name: testname.to_string(),
+        body: body.to_string(),
```
MarkMcCaskey, line 27:
```suggestion
        name: testname,
        body,
```

tests/test-generator/src/processors.rs
```diff
@@ -0,0 +1,101 @@
...
+    let body = format!(
+        "crate::run_wasi(r#\"{}\"#, r#\"{}\"#, \"{}\")",
+        p.display(),
+        outfile.display(),
+        compiler
+    );
+
+    Some(Test {
+        name: testname.to_string(),
+        body: body.to_string(),
```
MarkMcCaskey, line 99:
```suggestion
        name: testname,
        body,
```

tests/wasi_serialization.rs
```diff
@@ -1,65 +1,70 @@
...
+        let second_entry: Func<(), i32> = instance.exports.get("second_entry").unwrap();
+        let result = second_entry.call().unwrap();
+        assert_eq!(result, true as i32);
+    }
 
-#[allow(clippy::mut_from_ref)]
-pub(crate) fn get_wasi_state(ctx: &Ctx) -> &mut WasiState {
-    unsafe { state::get_wasi_state(&mut *(ctx as *const Ctx as *mut Ctx)) }
+    #[allow(clippy::mut_from_ref)]
+    pub(crate) fn get_wasi_state(ctx: &Ctx) -> &mut WasiState {
```
MarkMcCaskey, line 67: Note to self, I think this can be removed now and it definitely should be!

Or actually, it looks like we may be able to change `ctx: &Ctx` to `ctx: &mut Ctx` and remove the lint rule above

lib/runtime-c-api/src/module.rs
```diff
@@ -285,7 +285,7 @@ pub unsafe extern "C" fn wasmer_module_deserialize(
     let serialized_module: &[u8] = &*(serialized_module as *const &[u8]);
 
     match Artifact::deserialize(serialized_module) {
-        Ok(artifact) => match load_cache_with(artifact, &default_compiler()) {
+        Ok(artifact) => match load_cache_with(artifact, &*default_compiler()) {
```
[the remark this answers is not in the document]
syrusakbary: I tried with both (`default_compiler()` and `&default_compiler()`) and Rust complained.

src/commands/run.rs
```diff
@@ -474,7 +474,12 @@ impl LLVMCallbacks for LLVMCLIOptions {
 
 /// Execute a wasm/wat file
 fn execute_wasm(options: &Run) -> Result<(), String> {
-    if options.generate_debug_info && options.backend != Backend::Cranelift {
+    #[cfg(feature = "backend-cranelift")]
+    let in_cranelift = options.backend == Backend::Cranelift;
+    #[cfg(not(feature = "backend-cranelift"))]
+    let in_cranelift = false;
```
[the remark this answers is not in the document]
syrusakbary: I was wondering if there was a better way.. your suggestion works!

syrusakbary:
bors try

src/commands/run.rs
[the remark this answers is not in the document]
syrusakbary: I tried, and the code doesn't compile:

```rust
error[E0599]: no variant or associated item named `Cranelift` found for type `wasmer_runtime::Backend` in the current scope
   --> src/commands/run.rs:477:91
    |
477 |     let in_cranelift = cfg!(feature = "backend-cranelift") && options.backend == Backend::Cranelift;
    |                                                                                           ^^^^^^^^^ variant or associated item not found in `wasmer_runtime::Backend`
```

tests/generate-emscripten-tests/src/emtests.rs
```diff
@@ -60,111 +61,32 @@ pub fn compile(file: &str, ignores: &HashSet<String>) -> Option<String> {
...
+
         let _wasm_compilation = Command::new("emcc")
             .arg(file)
             .arg("-s")
             .arg("WASM=1")
             .arg("-o")
             .arg(output_str)
             .output()
-            .expect("failed to execute process");
+            .expect("failed to execute emcc process. Is `emcc` available in your system?");
```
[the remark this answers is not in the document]
syrusakbary: Good feedback, perhaps in other PR when (if) we prioritize back Emscripten

Makefile
```diff
@@ -177,9 +219,9 @@ integration-tests: release-clif examples
 	./tests/integration_tests/cowsay/test.sh
 
 examples:
-	cargo run --example plugin
-	cargo run --example callback
-
+	cargo build --release $(backend_features) --examples
+	test -f target/release/examples/callback && ./target/release/examples/callback || echo "skipping callback test"
+	test -f target/release/examples/plugin && ./target/release/examples/plugin || echo "skipping plugin test"
```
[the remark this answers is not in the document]
syrusakbary: The issue is that `make examples` didn't worked in non-x86 devices.

This PR uncovered that, so we needed a way to make it work.

tests/override.rs
```diff
@@ -1,64 +0,0 @@
-use wabt::wat2wasm;
```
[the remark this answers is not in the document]
syrusakbary: Yes, the "call-overwritten-element" test exists in the spectests in `tests/spectests/elem.wast`.

Even if it doesn't seems so, this file was autogenerated from the first versions of spectests (I wrote it the first version of the autogenerator, that's how I know :P )

tests/spectest_semantics.rs
```diff
@@ -1,39 +0,0 @@
...
-mod tests {
-    use wabt::wat2wasm;
-    use wasmer::error::{CallError, ExceptionCode, RuntimeError};
-    use wasmer::import::ImportObject;
-
-    // The semantics of stack overflow are documented at:
-    // https://webassembly.org/docs/semantics/#stack-overflow
-    #[test]
-    #[ignore]
-    fn test_stack_overflow() {
```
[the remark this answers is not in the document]
syrusakbary: Yup, I've created the `tests/custom/stack-overflow.wast` that tests it.

lib/runtime/src/lib.rs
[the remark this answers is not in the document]
syrusakbary: I'm going to mark this as resolved, since there seem to be no obvious way to do it properly.

Perhaps we can open another PR apart using the `cfg-if` crate :)

tests/wasi_serialization.rs
syrusakbary: Fixed!

tests/test-generator/src/processors.rs
syrusakbary: Just looked into it. The most obvious crate is:
https://crates.io/crates/ignore

But I think it's not trivial (as a few mins of work) to add it. If required we can add it in a PR apart. The simple solution works for now though :)

tests/generate-wasi-tests/src/set_up_toolchain.rs
```diff
@@ -12,19 +12,22 @@ fn install_toolchain(toolchain_name: &str) {
...
     util::print_info_on_error(&rustup_out, "TOOLCHAIN INSTALL FAILED");
+
+    println!("Installing rustup WASI target");
+    let rustup_out = Command::new("rustup")
+        .arg(format!("+{}", toolchain_name))
+        .arg("target")
+        .arg("add")
+        .arg("wasm32-wasi")
+        .output()
+        .expect("Failed to wasi target in Rust toolchain");
```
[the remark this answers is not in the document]
syrusakbary: Thanks!

tests/generate-wasi-tests/src/wasitests.rs
```diff
@@ -258,117 +172,74 @@ fn compile(
             .to_string()
     };
     let base_dir = Path::new(file).parent().unwrap();
-    generate_native_output(temp_dir, &file, &rs_mod_name).expect("Generate native output");
+    let output = generate_native_output(temp_dir, &file, &rs_mod_name, &options.args)
+        .expect("Generate native output");
+
+    let test = WasiTest { output, options };
+    let test_serialized = serde_json::to_string_pretty(&test).unwrap();
```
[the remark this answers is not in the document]
syrusakbary: This var is actually used later, when we write the wasi test content for each of the wasi versions.

PR #1399, opened 2020-04-21
Rename `*Descriptor` -> `*Type` to match Wasm spec

MarkMcCaskey:
Renaming things to be better aligned with the Wasm spec.

We'll be adding compatibility definitions with a deprecated attribute where possible.

# Review

### DIFF
--- a/lib/runtime-core/src/error.rs
+++ b/lib/runtime-core/src/error.rs
@@ -1,7 +1,7 @@
 //! The error module contains the data structures and helper functions used to implement errors that
 //! are produced and returned from the wasmer runtime core.
 use crate::backend::ExceptionCode;
-use crate::types::{FuncSig, GlobalDescriptor, MemoryDescriptor, TableDescriptor, Type};
+use crate::types::{FuncSig, GlobalType, MemoryType, TableType, Type};
 use core::borrow::Borrow;
 use std::any::Any;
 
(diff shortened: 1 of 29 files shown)


MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

tests/high_level_api.rs
```diff
@@ -87,36 +87,36 @@ wasmer_backends! {
...
+                    ty: export::ExternType::Table(TableType {
                         element: ElementType::Anyfunc,
                         minimum: 2,
                         maximum: None,
                     }),
                 },
-                export::ExportDescriptor {
+                export::ExportType {
                     name: "test-global",
-                    ty: export::ExternDescriptor::Global(GlobalDescriptor {
```
nlewycky, line 100: This field `ty` should be named `desc` for descriptor. The two places descriptors show up in the spec are as part of export types and import types (see 2.5.10 and 2.5.11). The export type is {name, desc} and the desc is {func | table | mem | global}.

MarkMcCaskey: Very good catch!  I just discussed the use of `desc` with Syrus now and we looked at the Wasm C API was doing as a reference and it seems that it just generally avoids indices in favor of other types, like `wasm_exporttype` here (https://github.com/WebAssembly/wasm-c-api/blob/master/include/wasm.h#L295).  So we're going to defer this for now, but this is something worth following up on.

MarkMcCaskey:
bors try

lib/runtime-core/src/error.rs
```diff
@@ -100,37 +100,37 @@ pub enum LinkError {
         name: String,
     },
     /// The memory descriptor provided does not match the expected descriptor.
-    IncorrectMemoryDescriptor {
+    IncorrectMemoryType {
```
MarkMcCaskey, line 103: Something I want reviewers to note about this PR is that it introduces breaking changes in `LinkError`, the error variants are renamed here:  I don't see a good way to make this change without breaking these.  Well, unless we're fine with the errors referring to them as `descriptors`

syrusakbary:
After talking offline, we will delay merging this PR to batch other breaking API changges

syrusakbary:

PR #1410, opened 2020-04-28
feat(runtime-core) Implement `ImportObjectIterator::size_hint`.

Hywan:
Because it is helpful in some circumstances.


Hywan:
bors try

MarkMcCaskey:
bors try-

syrusakbary:
Note for the future, there is a trait that we can use in Rust for that: `ExactSizeIterator`:

https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html

PR #1412, opened 2020-04-29
fix(singlepass-backend) Fix argument overwriting in a corner case when calling functions.

### LINKED ISSUE #1409: Args value is mangled when host function is called
When host function is called by Wasm program some arguments receive a different value than the one passed by the Wasm program.

This bug reproduces with the following Wasmer Rust crates:
```
wasmer-runtime = { version = "0.16.2", features = ["default-backend-singlepass"], default-features = false }
wasmer-runtime-core = { version = "0.16.2" }
```
and it does not reproduce with the following crates:
```
wasmer-runtime = { version = "0.13.1", features = ["default-backend-singlepass"], default-features = false }
wasmer-runtime-core = { version = "0.13.1" }
```

#### Version 0.16.2

In Terminal 1:
a) This needs to be run once:
```bash
git clone https://github.com/nearprotocol/nearcore/
cd nearcore
git checkout master_storage_write_log
cargo build --package neard --bin near
```
b) This can be run multiple times
```
rm -rf /tmp/testdir
./target/debug/near --home=/tmp/testdir init --chain-id= --test-seed=alice.near --account-id=test.near --fast
./target/debug/near --home=/tmp/testdir run
```

In Terminal 2:
```bash
git clone https://github.com/nearprotocol/near-evm/
cd near-evm
git checkout storage_write_log
./build.sh
cargo test --package near-evm --test cryptozombies_rpc test_all_in_one
```

Observe the console output in Terminal 1:
```
LOG storage_write 33 67119368 8 1047344 18446744073709551614
CALLED storage_write 33 67119368 8 18446744073709551614 18446744073709551614
```

The fist line is printed by the Wasm program using `log_utf8` host function before calling `storage_write` host function, see: https://github.com/near/near-sdk-rs/blob/storage_write_log/near-sdk/src/environment/env.rs#L641
The second line is printed by the `storage_write` host function right after it was called, see: https://github.com/nearprotocol/nearcore/blob/master_storage_write_log/runtime/near-vm-logic/src/logic.rs#L1879

The arguments do not match.

#### Version 0.13.1

Now, in Terminal 1, instead of `git checkout master_storage_write_log` do `git checkout beta_storage_write_log`
and run the rest of the commands.

Observe:
```
LOG storage_write 33 67119368 8 1047344 18446744073709551614
CALLED storage_write 33 67119368 8 1047344 18446744073709551614
````

What changed? -- We are now using Wasmer 0.13.1 instead of 0.16.2

### Expected behavior
Host receives exactly the same arguments that were passed by the guest program.

### Actual behavior
Host receives different arguments than those passed by the guest program.

### Versions
```
rustup 1.21.1 (7832b2ebe 2019-12-20)
cargo 1.44.0-nightly (7019b3ed3 2020-03-17)
rustc 1.44.0-nightly (f509b26a7 2020-03-18)
```

### PR DESCRIPTION
**Pull request by losfair**
Fixes #1409.

### DIFF
--- a/lib/singlepass-backend/src/codegen_x64.rs
+++ b/lib/singlepass-backend/src/codegen_x64.rs
@@ -1942,7 +1942,10 @@ impl X64FunctionCode {
 
     /// Emits a System V call sequence.
     ///
-    /// This function must not use RAX before `cb` is called.
+    /// This function will not use RAX before `cb` is called.
+    ///
+    /// The caller MUST NOT hold any temporary registers allocated by `acquire_temp_gpr` when calling
+    /// this function.
     fn emit_call_sysv<I: Iterator<Item = Location>, F: FnOnce(&mut Assembler)>(
         a: &mut Assembler,
         m: &mut Machine,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/singlepass-backend/src/codegen_x64.rs [hunk 1]
```diff
@@ -2063,14 +2066,17 @@ impl X64FunctionCode {
                             // Dummy value slot to be filled with `mov`.
                             a.emit_push(Size::S64, Location::GPR(GPR::RAX));
 
-                            // Use R10 as the temporary register here, since it is callee-saved and not
-                            // used by the callback `cb`.
-                            a.emit_mov(Size::S64, *param, Location::GPR(GPR::R10));
+                            // Use RCX as the temporary register here, since:
+                            // - It is a temporary register that is not used for any persistent value.
+                            // - This register as an argument location is only written to after `sort_call_movs`.'
```
nlewycky, line 2071:
```suggestion
                            // - This register as an argument location is only written to after `sort_call_movs`.
```

PR #1416, opened 2020-04-29
Fix `make check` so it runs in CI

MarkMcCaskey:
Before shipping we should also investigate:

- [ ] Why this wasn't causing CI failures even though it was returning a non-zero value
- [ ] That the adjusted benchmark checking code is checking what we want it to

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -255,13 +255,16 @@ install:
 
 # Checks
 check-bench-singlepass:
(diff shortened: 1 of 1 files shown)


MarkMcCaskey:
bors try

syrusakbary:
This should not be needed with the work from the refactor. Closing PR

PR #1417, opened 2020-04-30
Store signature indices for all functions, not just exports

Redrield:
# Description
In playing with wasmer for use in a wasm plugin system, I found that it was possible to create a situation where attempting to call an existing function with a valid table index would result in a panic.

As it is right now, Rust functions are only exported from the wasm binary if they are extern "C", and marked no_mangle. When trying to run the file with the cranelift backend, trampolines are only generated for signatures from exported functions. In short, if the host is given a table index of a function that isn't exported, and doesn't share a function signature with any exported function, the execution will panic. An example of this can be replicated with the code in [this repo](https://github.com/Redrield/wasm-plugin-system/)

This PR modifies trampoline generation in the cranelift backend to consider all signature types. This seems to be the way that the LLVM backend does it already, so no modification there is required.

# Review

### DIFF
--- a/lib/clif-backend/src/trampoline.rs
+++ b/lib/clif-backend/src/trampoline.rs
@@ -8,7 +8,7 @@ use cranelift_codegen::{
 use std::{collections::HashMap, iter, mem};
 use wasmer_runtime_core::{
     backend::sys::{Memory, Protect},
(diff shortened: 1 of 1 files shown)


nlewycky:
It makes sense to me. Is singlepass affected?

syrusakbary:

Issue #1429, opened 2020-05-05
Wasm execution result is different from native rust


evgenykuzyakov:
### Describe the bug

I'm using latest `uint` crate and found that results in multiplication overflow when within Wasm and wasmer (`wasmer-runtime = { version = "0.13.1", features = ["default-backend-singlepass"], default-features = false }`). Reproducible with `0.16.2` as well.

Here is the minimum reproduction: https://github.com/evgenykuzyakov/wasmer-rust-example/tree/math-bug

```
~/code/wasmer-rust-example/math-bug ~/code/wasmer-rust-example
Testing in rust
    Updating crates.io index
   Compiling byteorder v1.3.4
   Compiling static_assertions v1.1.0
   Compiling rustc-hex v2.1.0
   Compiling uint v0.8.3
   Compiling math-bug v0.1.0 (/Users/ekwork/code/wasmer-rust-example/math-bug)
    Finished test [unoptimized + debuginfo] target(s) in 6.04s
     Running target/debug/deps/math_bug-ca1b3b206f3bb89d

running 2 tests
test tests::test_multiply_4 ... ok
test tests::test_multiply_8 ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

   Doc-tests math-bug

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

Compiling wasm
   Compiling byteorder v1.3.4
   Compiling rustc-hex v2.1.0
   Compiling static_assertions v1.1.0
   Compiling uint v0.8.3
   Compiling math-bug v0.1.0 (/Users/ekwork/code/wasmer-rust-example/math-bug)
    Finished release [optimized] target(s) in 4.28s
~/code/wasmer-rust-example
   Compiling wasmer-run v0.1.0 (/Users/ekwork/code/wasmer-rust-example)
    Finished release [optimized] target(s) in 0.76s
     Running `target/release/wasmer-run`
Running multiply_8
multiply_8 is OK
Running multiply_4
Error: CallError(RuntimeError: unknown error)
```

### Steps to reproduce

1. Clone https://github.com/evgenykuzyakov/wasmer-rust-example/tree/math-bug
2. Need `nightly` toolchain for `wasmer`. And `stable` with `wasm32-unknown-unknown` component.
3. Run `./run.sh`
4. Observe `wasm` execution panics.

### Expected behavior

[...]

PR #1436, opened 2020-05-11
fix(singlepass-backend): I64ExtendI32U: Garbage in upper 32 bits shouldn't propagate to the result.

losfair:
Fixes https://github.com/wasmerio/wasmer/issues/1429 .


syrusakbary:
Waiting on a test before merge

tests/custom/int-extend-garbage.wast
```diff
@@ -0,0 +1,35 @@
+;; https://github.com/wasmerio/wasmer/pull/1436
```
nlewycky, line 1: Please link to the issue instead of to the PR?
```suggestion
;; https://github.com/wasmerio/wasmer/issues/1429
```

PR #1453, opened 2020-05-22
Fix trap reporting for unreachable errors in the Singlepass backend

MarkMcCaskey:
This PR updates to push code versions in `Instance::new` so that singlepass reports errors correctly.

This is required for the fault handler to read from the exception table and report the exception type correctly in Singlepass:

This fix does not ever pop it, it doesn't update the code in run.rs which both pushes and pops (so that may have broken too (needs investigation))

Other considerations, this logic seems to only apply to `unix`, even if tests don't fail, we should double check windows (and get the tests running there too if they're not)

# Review

### DIFF
--- a/lib/runtime-core/src/instance.rs
+++ b/lib/runtime-core/src/instance.rs
@@ -2,8 +2,6 @@
 //!
 //! An "instance", or "instantiated module", is a compiled WebAssembly [`Module`] with its
 //! corresponding imports (via [`ImportObject`]) that is ready to execute.
-#[cfg(unix)]
-use crate::fault::{pop_code_version, push_code_version};
 use crate::{
     backend::RunnableModule,
     backing::{ImportBacking, LocalBacking},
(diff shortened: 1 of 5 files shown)


MarkMcCaskey:
bors try

lib/runtime-core/src/instance.rs
```diff
@@ -106,6 +108,22 @@ impl Instance {
...
         };
 
+        // We need to push the code version so that the exception table can be read
+        // in the feault handler so that we can report traps correctly
+        #[cfg(unix)]
+        {
+            if let Some(msm) = instance.module.runnable_module.get_module_state_map() {
+                push_code_version(CodeVersion {
+                    baseline: true,
+                    msm: msm,
```
nlewycky, line 118:
```suggestion
                    msm,
```

MarkMcCaskey:
I don't think this is ready to ship yet by the way, I think we probably need feedback from Heyang.  I mentioned the known problems above!

lib/runtime-core/src/instance.rs
```diff
@@ -106,6 +108,22 @@ impl Instance {
             import_object: imports.clone(),
         };
 
+        // We need to push the code version so that the exception table can be read
+        // in the feault handler so that we can report traps correctly
+        #[cfg(unix)]
+        {
+            if let Some(msm) = instance.module.runnable_module.get_module_state_map() {
+                push_code_version(CodeVersion {
```
losfair, line 116: We need to pop the code version when an `Instance` is `drop`ped. The code version won't be released otherwise, and there will be a memory leak.

MarkMcCaskey: For sure -- thanks!  It seems the current system relies on `Instance`s being dropped in the opposite order that they were created, otherwise `pop_code_version` may pop the wrong one.  I think this is probably okay for now but it definitely seems like a bug

Issue #842, opened 2019-09-29
Signal handling improvements


michaelvoronov:
### Motivation

In Fluence we decided to move to Wasmer :). And since our node is written in Scala, we intend to use Wasmer as a JNI module. But almost all Java VMs actively use signals for various tasks (such as asymmetric Dekker synchronization). It turned out that it is incompatible with the signal handling scheme in the clif backend since it just calls abort() functions if a signal doesn't belong to the Wasm side.

Also, it could hurt other environments if you try to embed Wasmer.

### Proposed solution

The most straightforward improvement is just calling the previous handler instead of abort() like this [one](https://github.com/wasmerio/wasmer/compare/master...fluencelabs:clif_jni_hardering
) (I've already tested it in our network). Ideally, it needs a more complex algo taking into account layouts of Wasm heap allocation, stack and so on to determine the exact destination of a signal.

P.S. Really, we don't depend much on this feature to switch to Wasmer, because there is a nice `libjsig` mechanism in Hotspot. But it would be more convenient for us to rely only on proper signal handling scheme inside Wasmer.

syrusakbary:
Thanks for creating the issue!

We are reviewing it internally and will follow here once we have the strategy a bit more clear

michaelvoronov:
Hmm, It seems that the signal management related to managed singlepass context should be refactored to be embeddable... It doesn't even work with `libjsig`.

michaelvoronov:
And also, one wanted design point: in Fluence, we are dealing with multi modules backends. Each module has its memory, and in terms of Wasmer, they have different instances. And we will have some predefined trusted modules such as Redis, SQLite that each developer could easily integrate as a component in his backend. It will be known for sure that these modules don't have any compiler/jit bombs, and it is possible to use the fastest llvm backend for them safely (It could even be precompiled in llvm ir). But untrusted user's modules should still use the singlepass compiler.

At now, each backend compiler in Wasmer uses its signal scheme, and it seems that they aren't compatible with each other (to be honest, I haven't checked, but clif and singlepass have different trap handlers without the ability to call previously registered handlers). So, it would be nice to have such a possibility after signal handler refactoring.

zackangelo:
I'm trying to call Wasmer from a JNI module and I believe I'm bumping into this issue. The first JNI invocation succeeds, but subsequent ones result in the process crashing with error code 134. 

[...]

PR #1496, opened 2020-07-08
Interrupt signal propagation

hanjukim:
# Description
If we embed wasmer to a process, wasmer will override interrupt(SIGINT) signal handler. In result, previous signal handler will not be called when SIGINT is fired, and if it gets fired again, process will be stopped by process::abort() which leads to output a long stack trace.

Related Issue: https://github.com/wasmerio/wasmer/issues/842

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## **[Unreleased]**
+- [#1496](https://github.com/wasmerio/wasmer/pull/1496) Propagate interrupt signal(SIGINT) to previous signal handler
 
 ## 0.17.1 - 2020-06-24
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2048,9 +2048,9 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.53"
+version = "1.0.56"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "993948e75b189211a9b31a7528f950c6adc21f9720b6438ff80a7fa2f864cea2"
+checksum = "3433e879a558dde8b5e8feb2a04899cf34fdde1fafb894687e52105fc1162ac3"
 dependencies = [
  "itoa",
  "ryu",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -210,9 +210,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.54"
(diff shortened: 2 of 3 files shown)


hanjukim:
@Hywan Can you review this? This change is essential to shutdown process gracefully.

syrusakbary:
Hi @hanjukim we are in the mid of the process of upgrading Wasmer, that's why we are a bit slower replying here.

We will be super happy to invite you to it so you can try the new Wasmer, would you mind joining our slack channel so I can follow up there and make sure the issue is fixed? https://slack.wasmer.io/ Thanks! 🤗

Hywan:
@hanjukim Sorry for the late reply. As @syrusakbary told you, a new version of Wasmer is coming, and your issue is likely to be resolved already. If you accept the invitation, you can try to replicate your issue. Thoughts?

hanjukim:
> @hanjukim Sorry for the late reply. As @syrusakbary told you, a new version of Wasmer is coming, and your issue is likely to be resolved already. If you accept the invitation, you can try to replicate your issue. Thoughts?

@syrusakbary I justed joined to wasmer slack. Which branch should I use for testing new version candidate? Thanka for your help.

hanjukim:
Ok. I reviewd new Wasmer (wasmer-reborn) and it seems a lot of things are changing. We are using cosmwasm which is a library that provides building cosmos-compatible wasm smart contracts and it is dependent on wasmer. Is there any chance that this PR could be merged to next release of (old) wasmer? If not, we will have to decide to fork every dependent repositories for our software release.

syrusakbary:
> I reviewd new Wasmer (wasmer-reborn) and it seems a lot of things are changing

That's right. Thanks though for taking a sneak peek of the project!

> Is there any chance that this PR could be merged to next release of (old) wasmer?

Yeah, I think we can do that to facilitate your work!

nlewycky:
While reviewing this PR, I began to wonder why we were installing a hander for SIGINT at all, and discovered that it's tied to the `managed` cfg feature. I moved the relevant code to be conditional under that feature, so we no longer install any handler for SIGINT by default. https://github.com/wasmerio/wasmer/pull/1552/

If you'd like to update this PR so that you can use `managed` and install your own signal handler for sigint, please rebase on top of the `0.x` branch. Thanks!

syrusakbary:
This PR was done for the old codebase, which as commented by Nick it's also solved in `master`. Closing PR.

PS: thanks anyway for the contribution @hanjukim !

PR #1537, opened 2020-08-14
Prepare for 1.0.0-alpha01.0 release

MarkMcCaskey:
This is the first alpha release of `1.0.0`: expect breaking changes.

The version suffix `alpha01.0` was chosen so that we can ship updates
that automatically update with the latter number and can prevent
auto-updates by incrementing the former number (which is not actually
a number as far as semver is concerned)

I think our release process is likely broken and we'll have to do a lot of manual work.  I'm deferring updating the changelog until later as I intend to write a decent amount there -- will update it as a separate PR after the release.

# Review

### DIFF
--- a/lib/engine/Cargo.toml
+++ b/lib/engine/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmer-engine"
-version = "1.0.0-alpha.1"
+version = "1.0.0-alpha01.0"
 authors = ["Wasmer Engineering Team <engineering@wasmer.io>"]
 description = "Wasmer Engine abstraction"
 license = "(Apache-2.0 WITH LLVM-exception) or MIT"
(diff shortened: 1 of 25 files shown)


lib/engine/Cargo.toml
```diff
@@ -11,9 +11,9 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
-wasmer-types = { path = "../wasmer-types", version = "1.0.0-alpha.1" }
-wasmer-compiler = { path = "../compiler", version = "1.0.0-alpha.1", default-features = false }
```
nlewycky, line 15: These default features were just `serde` and `std`? Nothing about particular compilers or engines?

MarkMcCaskey: Yeah, that's correct

PR #1546, opened 2020-08-17
Fix up Cargo manifests for publishing Wasmer

### LINKED ISSUE #1547: Wasmer (API) depedency on engine-dummy
Seen in #1546 ; the `wasmer` (API) crate has a dev-dependency on `wasmer-engine-dummy`.

We should refactor to remove this dependency so that we don't need to publish the dummy crate (which is really needed by the wasmer CLI) to publish the API crate.

MarkMcCaskey:
I needed to fix these up to publish the wasmer crates

also resolves #1547 

# Review


tests/lib/engine-dummy/Cargo.toml
```diff
@@ -2,9 +2,9 @@
 name = "wasmer-engine-dummy"
 version = "1.0.0-alpha01.0"
 authors = ["Wasmer Engineering Team <engineering@wasmer.io>"]
-description = "wasmer testing utils"
+description = "Wasmer placeholder engine"
+license = "MIT"
 edition = "2018"
-publish = false
```
nlewycky, line 7: Why publish engine-dummy?

MarkMcCaskey: It's a dependency of `wasmer` (API) because it's used in Wasmer the CLI...

I agree that this was unfortunate, but I figured it was better to publish it now and refactor it later as I thought the refactor would be relatively large and that there'd be no way for me to publish it last week.


I agree that this is something we should file an issue for and investigate

MarkMcCaskey: So to clarify, it's an accidental hard-requirement that it be published.  It may be possible to rework it so that we don't need to publish it in the future, but I chose to just publish it rather than try to fix it on Friday

syrusakbary: @MarkMcCaskey CLI doesn't use engine-dummy as far as I can tell?

PR #1548, opened 2020-08-17
Add coverage checking to CI

MarkMcCaskey:
Get code coverage running in CI

- [ ] It looks like this may be blocked on setting an env var in CI with a value from https://codecov.io/gh/wasmerio/wasmer

# Review


.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,29 @@
+on: [push]
+
+name: build
+
+env:
+  RUST_BACKTRACE: 1
```
nlewycky, line 6: Is this intended to land or just for debugging this PR? Is this needed for coverage or is this for debugging when things do go wrong?

.github/workflows/coverage.yaml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: 1.45.0
+          override: true
+      - name: Generate Coverage Report
+        run: |
+          cargo install cargo-tarpaulin
+          cargo tarpaulin --forward --features "cranelift,singlepass,test-cranelift,test-no-traps" --release -t120 --out Xml -- --skip traps:: --skip spec::linking
+          ls
```
nlewycky, line 24: What's the `ls` for?

MarkMcCaskey: old debug code

.github/workflows/coverage.yaml
MarkMcCaskey: Both, it doesn't hurt to have it, it's not needed but it's better to have it enabled in CI

nlewycky: Sounds good!

tests/lib/wast/src/wast.rs
```diff
@@ -171,6 +173,8 @@ impl Wast {
                 let result = self.perform_invoke(call);
                 self.assert_trap(result, message)?;
             }
+            #[cfg(feature = "test-no-traps")]
```
nlewycky, line 176: Is there an issue filed either on us or on tarpaulin for this? Could you put a link to it in a comment here?

PR #1549, opened 2020-08-17
Add script to automate publishing crates to crates.io

MarkMcCaskey:
This script uses hardcoded dependencies but uses topological sort to find the order to publish the crates in and then publishes them.  The script handles failure by skipping crates that have already been published and includes a `--dry-run` flag for testing it.

It wouldn't be very hard to parse the toml files to get the dependencies, but I chose to keep things extremely simple to start, we can improve this script later as needed.

### DIFF
--- a/scripts/publish.py
+++ b/scripts/publish.py
@@ -1,7 +1,8 @@
 #! /usr/bin/env python3
 
 # This is a script for publishing the wasmer crates to crates.io.
-# It should be run in the root of wasmer like `python3 scripts/publish.py`.
+# It should be run in the root of wasmer like `python3 scripts/publish.py --no-dry-run`.
+# By default the script executes a test run and does not publish the crates to crates.io.
 
 # install dependencies:
 # pip3 install toposort
@@ -65,7 +66,7 @@ location = {
     "wasmer-wasi-experimental-io-devices": "wasi-experimental-io-devices",
 }
 
-dry_run = False
+no_dry_run = False
 
 def get_latest_version_for_crate(crate_name: str) -> Optional[str]:
     output = subprocess.run(["cargo", "search", crate_name], capture_output=True)
@@ -88,22 +89,22 @@ def publish_crate(crate: str):
     starting_dir = os.getcwd()
     os.chdir("lib/{}".format(location[crate]))
 
-    global dry_run
-    if dry_run:
-        print("In dry-run: not publishing crate `{}`".format(crate))
-    else:
+    global no_dry_run
+    if no_dry_run:
         output = subprocess.run(["cargo", "publish"])
+    else:
+        print("In dry-run: not publishing crate `{}`".format(crate))
 
     os.chdir(starting_dir)
 
 def main():
     parser = argparse.ArgumentParser(description='Publish the Wasmer crates to crates.io')
-    parser.add_argument('--dry-run', default=False, action='store_true',
+    parser.add_argument('--no-dry-run', default=False, action='store_true',
                         help='Run the script without actually publishing anything to crates.io')
     args = vars(parser.parse_args())
 
-    global dry_run
-    dry_run = args['dry_run']
+    global no_dry_run
+    no_dry_run = args['no_dry_run']
 
     # get the order to publish the crates in
     order = list(toposort_flatten(dep_graph, sort=True))
--- a/scripts/publish.py
+++ b/scripts/publish.py
@@ -1,0 +1,123 @@
+#! /bin/python3
+
+# This is a script for publishing the wasmer crates to crates.io.
+# It should be run in the root of wasmer like `python3 scripts/publish.py`.
+
+# install dependencies:
+# pip3 install toposort
+
+import argparse
+import os
+import re
+import subprocess
+import time
+
+from typing import Optional
+from toposort import toposort_flatten
+
+
+# TODO find this automatically
+target_version = "1.0.0-alpha01.1"
+
+# TODO generate this by parsing toml files
+dep_graph = {
+    "wasmer-types": set([]),
+    "wasmer-vm": set(["wasmer-types"]),
+    "wasmer-compiler": set(["wasmer-vm", "wasmer-types"]),
+    "wasmer-object": set(["wasmer-types", "wasmer-compiler"]),
+    "wasmer-engine": set(["wasmer-types", "wasmer-vm", "wasmer-compiler"]),
+    "wasmer-compiler-singlepass": set(["wasmer-types", "wasmer-vm", "wasmer-compiler"]),
+    "wasmer-compiler-cranelift": set(["wasmer-types", "wasmer-vm", "wasmer-compiler"]),
+    "wasmer-compiler-llvm": set(["wasmer-types", "wasmer-vm", "wasmer-compiler"]),
+    "wasmer-engine-jit": set(["wasmer-types", "wasmer-vm", "wasmer-compiler", "wasmer-engine"]),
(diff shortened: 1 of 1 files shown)


scripts/publish.py
```diff
@@ -0,0 +1,123 @@
+#! /bin/python3
```
nlewycky, line 1: Unusual shebang? Either `#! /usr/bin/python3` or `#! /usr/bin/env python3`?

scripts/publish.py
```diff
@@ -0,0 +1,123 @@
...
+# pip3 install toposort
+
+import argparse
+import os
+import re
+import subprocess
+import time
+
+from typing import Optional
+from toposort import toposort_flatten
```
nlewycky, line 16: Something like this should work:
```
try:
    from toposort import toposort_flatten
except ImportError:
    print("Please install toposort, `pip3 install toposort`")
```

scripts/publish.py
```diff
@@ -0,0 +1,123 @@
...
+                   "wasmer-compiler-llvm", "wasmer-compiler", "wasmer-engine", "wasmer-engine-jit",
+                   "wasmer-engine-native", "wasmer-types"]),
+    "wasmer-cache": set(["wasmer"]),
+    "wasmer-wasi": set(["wasmer"]),
+    "wasmer-wasi-experimental-io-devices": set(["wasmer-wasi"]),
+    "wasmer-emscripten": set(["wasmer"]),
+}
+
+# where each crate is located in `lib`
+# this can also be generated from the toml files
```
nlewycky, line 45:
```suggestion
# TODO: this could also be generated from the toml files
```

scripts/publish.py
```diff
@@ -0,0 +1,123 @@
...
+import time
+
+from typing import Optional
+from toposort import toposort_flatten
+
+
+# TODO find this automatically
+target_version = "1.0.0-alpha01.1"
+
+# TODO generate this by parsing toml files
```
nlewycky, line 22:
```suggestion
# TODO: generate this by parsing toml files
```

scripts/publish.py
```diff
@@ -0,0 +1,123 @@
...
+import os
+import re
+import subprocess
+import time
+
+from typing import Optional
+from toposort import toposort_flatten
+
+
+# TODO find this automatically
```
nlewycky, line 19:
```suggestion
# TODO: find this automatically
```

scripts/publish.py
```diff
@@ -0,0 +1,123 @@
...
+    "wasmer-engine-jit": "engine-jit",
+    "wasmer-engine-native": "engine-native",
+    "wasmer-cache": "cache",
+    "wasmer": "api",
+    "wasmer-wasi": "wasi",
+    "wasmer-emscripten": "emscripten",
+    "wasmer-wasi-experimental-io-devices": "wasi-experimental-io-devices",
+}
+
+dry_run = False
```
nlewycky, line 65: Default to True, rename `dry_run` arg to `nodry_run`?

scripts/publish.py
```diff
@@ -0,0 +1,123 @@
...
+def publish_crate(crate: str):
+    os.chdir("lib/{}".format(location[crate]))
+
+    global dry_run
+    if dry_run:
+        print("In dry-run: not publishing crate `{}`".format(crate))
+    else:
+        output = subprocess.run(["cargo", "publish"])
+
+    os.chdir("../..")
```
nlewycky, line 94: Note that this may have surprising behaviour if lib/location_crate goes through a symlink. Obviously we don't expect that so it's up to you whether to fix it.

scripts/publish.py
```diff
@@ -0,0 +1,123 @@
...
+        result = prog.match(line)
+        if result:
+            return result.group(1)
+
+def is_crate_already_published(crate_name: str) -> bool:
+    found_string = get_latest_version_for_crate(crate_name)
+    if found_string is None:
+        return False
+
+    return target_version == found_string
```
nlewycky, line 82: Strictly speaking this only implements "is crate already published" as long as there is no newer published version of the crate than we're trying to publish. Which shouldn't happen if all releases use this script and set the version correctly, but it would be nice if it could detect or recover from manual crate releases. (Again, this is not a situation we intend to put ourselves in, but unlike the symlink case, I can see someone tripping over this for real, probably during a fire drill related to why we had a manual release of one crate to begin with.)

scripts/publish.py
MarkMcCaskey: So I based the command on `cargo publish` which also has a `--dry-run` flag and defaults to publishing, which is why that made sense to me

scripts/publish.py
MarkMcCaskey: Good call, I updated it to store the starting dir and return to it; this also solves the problem with `deprecated/runtime-*` which aren't part of the script yet

scripts/publish.py
nlewycky: I would still prefer to default to a dry run here. Some tools like python linters work by running the code (hence the `if name == "__main__":` at the end) or someone may be editing the code or refactoring it into a larger library and we want to avoid accidents. I agree that having a `dry_run` is the normal approach and our reasons don't apply to `cargo publish` itself, but I'd still rather this program and tools like it operate in "safe by default" mode with an explicit decision to turn on the business end. To me it's less about this script and more about a general rule. I could live with us keeping `--dry_run` here if you feel strongly about it.

scripts/publish.py
MarkMcCaskey: Yeah, I'm not too worried about it  -- I think it's good enough for now; if this fails and we republish something, then it will just cause `cargo publish` to fail, which will either cause python to throw an exception or continue executing as normal, either outcome seems fine

MarkMcCaskey: Also having to understand versions seems a bit too advanced for this script -- the script expects all the Cargo.toml files to have the correct version to publish already... the hardcoded version should really just come from the `toml` file itself to allow the script to work correctly with transient network failures or whatever

scripts/publish.py
MarkMcCaskey: Hmm, interesting point, the main reason why I think not defaulting `dry-run` makes sense is that `dry-run` is only useful when editing the script which I perceive to be a less common operation than running it.

I'm seeing this script more as a quick hack like the `scripts/update-version.sh` I made than as a robust tool: it's a simple piece of code that does the right thing enough that it's a useful way to save some time.

I don't feel that strongly though and we can make it opt-out of test mode

scripts/publish.py
nlewycky: It's just the mismatch between what it does and the name. If it were "is_version_already_published" I'd be fine with it since it does exactly what it says on the tin, but this kind of minor semantic mismatch is exactly the kind of thing I'm looking for in a code review.

scripts/publish.py
nlewycky: > I'm seeing this script more as a quick hack like the scripts/update-version.sh I made than as a robust tool: it's a simple piece of code that does the right thing enough that it's a useful way to save some time.

This is only temporary unless it works. :)

Up to you whether to change it.

scripts/publish.py
MarkMcCaskey: oh, you are right that if a later version is published, that won't show up with `cargo search`, but I don't think there's any way to check if the crate is already published given a later version without attempting to download it, curling crates.io, or attempting to publish it.  I think it's good enough for now as it will serve its purpose of speeding up retries in case of failure and should continue working even if this check fails.

PR #1551, opened 2020-08-18
Reintegrate with bors

MarkMcCaskey:
Attempting to get bors working again to speed up CI

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1,25 +1,3 @@
-on: [push]
-
-name: build
-
-env:
-  RUST_BACKTRACE: 1
-
-jobs:
-  lint:
-    name: Code lint
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - name: Install Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: 1.45.0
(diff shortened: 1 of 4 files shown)


MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

.github/workflows/main.yaml
```diff
@@ -1,25 +1,3 @@
-on: [push]
-
-name: build
-
-env:
```
nlewycky, line 5: Do you want to keep RUST_BACKTRACE=1?

MarkMcCaskey: Good catch, yeah I'll add it back

MarkMcCaskey:
This PR is more or less ready to ship but we'll need to adjust some admin settings and maybe tweak bors (need to investigate that timeout (I suspect it's the first line in bors.toml)).

I'm going to delay shipping this PR until next week

bors.toml
```diff
@@ -1,5 +1,8 @@
 status = [
-    "wasmerio.wasmer"
+    "wasmerio.wasmer",
```
syrusakbary, line 2: I think we don't need this status check. This was the status name of tests in Azure

PR #1554, opened 2020-08-18
Bump the minimum rust version to 1.45.2.

nlewycky:
# Description
Bump the minimum required rustc version to 1.45.2, marking the first time that singlepass will be able to build without a nightly compiler.

This is a bit ahead of schedule (we normally support one behind the most recent stable) but we expect the next stable to be released in about a week, which should happen before this commit is released.

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## **[Unreleased]**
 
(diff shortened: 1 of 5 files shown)


nlewycky:
> Looks good to me, but I think we also need to update the singlepass dyn asm dep (and Cargo.lock) for this to work on stable

Done!

PR #1557, opened 2020-08-20
Fix support of dynamic memories in compiler-llvm.

nlewycky:
Fix implementation of dynamic memory styles with compiler-llvm.

Make VMMemoryDefinition::current_length a u32, from a usize. The usize is host-specific not target-specific, and trying to use u64 causes problems with the heap in compiler-cranelift.

### DIFF
--- a/lib/api/src/externals/memory.rs
+++ b/lib/api/src/externals/memory.rs
@@ -2,6 +2,7 @@ use crate::exports::{ExportError, Exportable};
 use crate::externals::Extern;
 use crate::store::Store;
 use crate::{MemoryType, MemoryView};
+use std::convert::TryInto;
 use std::slice;
 use std::sync::Arc;
 use wasmer_types::{Pages, ValueType};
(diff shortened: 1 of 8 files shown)


lib/api/src/externals/memory.rs
```diff
@@ -84,7 +85,7 @@ impl Memory {
     }
 
     /// Returns the size (in bytes) of the `Memory`.
-    pub fn data_size(&self) -> usize {
+    pub fn data_size(&self) -> u32 {
```
MarkMcCaskey, line 88: This change seems fine on `Memory` a concrete type, but for something like a Memory trait, we probably want `u64`.  Wasm64 will eventually ship with more than 4gb of memory

lib/vm/src/instance.rs
```diff
@@ -1145,7 +1145,7 @@ unsafe fn get_memory_slice<'instance>(
         let import = instance.imported_memory(init.location.memory_index);
         import.definition.as_ref().clone()
     };
-    slice::from_raw_parts_mut(memory.base, memory.current_length)
+    slice::from_raw_parts_mut(memory.base, memory.current_length.try_into().unwrap())
```
MarkMcCaskey, line 1148: Converting u32 to usize in a failable way is probably overkill -- I don't think Wasmer is likely to run on any platform where a u32 can't always convert into a usize.  This is consistent with some other code I've seen in the wasmer code base though so you don't have to change it, it just seems a bit odd to me

lib/api/src/externals/memory.rs
nlewycky: Done.

lib/vm/src/instance.rs
nlewycky: It's what the compiler error recommended. I have a slight preference to avoid `as` casting (very convenient, but also seems like a quick way to get corner-case bugs).

PR #1558, opened 2020-08-20
feat(wasmer-types) Implement `NativeWasmType` for `u128`

Hywan:
I wonder why it wasn't implemented before. Am I missing something? cc @MarkMcCaskey @nlewycky

### DIFF
--- a/lib/wasmer-types/src/native.rs
+++ b/lib/wasmer-types/src/native.rs
@@ -71,6 +71,7 @@ impl NativeWasmType for i32 {
         bits as _
     }
 }
+
 impl NativeWasmType for i64 {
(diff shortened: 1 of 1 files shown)


nlewycky:
This adds NativeWasmType for i128 but doesn't include an associated use of it anywhere, not even a test. Is there a straight-forward way for this PR to include a usage of the type, even if it's just a test. Maybe the `mod test_native_type` at the bottom of the same file?

Hywan:
Tests added.

I don't need this feature in Wasmer itself, but rather in `python-ext-wasm`.

syrusakbary:
I'll force merge since somehow this PR is not tested

PR #1559, opened 2020-08-20
Fix build of windows dlls from linux with engine-native + compiler-llvm.

nlewycky:
# Description
The experimental interface between engine-native and compiler-llvm allows llvm to emit an object file and build a shared object out of it that engine-native will dlopen, or LoadLibraryW on windows. This PR fixes the problems we run into producing a dll file with cross-compilation toolchain.

There are other bugs in compiler-llvm which mean that with this PR applied we produce DLLs that don't work on windows, but those are cross-platform bugs fixed in #1557 .

# Review

### DIFF
--- a/lib/compiler-llvm/src/compiler.rs
+++ b/lib/compiler-llvm/src/compiler.rs
@@ -4,7 +4,7 @@ use crate::translator::FuncTranslator;
 use crate::CompiledFunctionKind;
 use inkwell::context::Context;
 use inkwell::memory_buffer::MemoryBuffer;
-use inkwell::module::Module;
+use inkwell::module::{Linkage, Module};
 use inkwell::targets::FileType;
 use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
 use wasmer_compiler::{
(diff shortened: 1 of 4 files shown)


lib/compiler-llvm/src/compiler.rs
```diff
@@ -194,6 +198,12 @@ impl LLVMCompiler {
         let memory_buffer = target_machine
             .write_to_memory_buffer(&merged_module, FileType::Object)
             .unwrap();
+        if let Some(ref callbacks) = self.config.callbacks {
+            callbacks.obj_memory_buffer(
+                &CompiledFunctionKind::Local(function_body_inputs.iter().next().unwrap().0),
```
MarkMcCaskey, line 203: As a reader of this code with no context, it's not obvious to me why this is correct, a comment about why the `unwrap` won't fail would be useful here (why do we know that there is at least one element?)

nlewycky: Excellent catch! Really, why am I passing in one random function here. No, we should have an enum entry that correctly describes what we're actually calling the callback for.

Done.

PR #1566, opened 2020-08-26
Add support for WASI opening Unix special files

### LINKED ISSUE #1565: Opening /dev/null leads to panic: not yet implemented
### Describe the bug

If a program attempts to open `/dev/null` or any other device file, Wasmer panics with the error saying "not yet implemented".

```sh
zsh: command not found: rustc
wasmer 0.17.1 |  | x86_64
```

### Steps to reproduce

1. Put the following code in `opendevnull.c`:

```c
#include <stdio.h>
int main() {
  fputs("open: /dev/null\n", stderr);
  if (fopen("/dev/null", "r")==NULL) perror("open");
  return 0;
}
```

2. Compile using `wasienv`:
```sh
$ wasicc opendevnull.c -o opendevnull.wasm
```

3. Run
```sh
$ wasmer run --dir=/ opendevnull.wasm
open: /dev/null
thread 'main' panicked at 'not yet implemented: state::get_inode_at_path unknown file type: not file, directory, or symlink', lib/wasi/src/state/mod.rs:818:33
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
Error: error: User(Any)
```

### Expected behavior

Opening `/dev/null` should've succeeded.

### Actual behavior

Wasmer panics with the error saying "not yet implemented.

### Additional context

I want to use the existing test suite to ensure that the program ported to Wasmer works as expected. I'd like to have as few modifications to the test suite as possible.

The test suite depends heavily on `/dev/null`. I could make it work for the output going to `/dev/null` easily with `--mapdir` feature (so that it actually goes to a file named "null" in a temporary dir). Unfortunately, `/dev/null` is also often used for input and I don't have a workaround.

MarkMcCaskey:
Resolves #1565 

# Review


tests/wasi-wast/wasi/snapshot1/unix_open_special_files.wast
```diff
@@ -0,0 +1,5 @@
+(wasi_test "unix_open_special_files.wasm"
+  (map_dirs "/dev:/dev")
+  (assert_return (i64.const 0))
+  (assert_stdout "13\n")
+)
```
nlewycky, line 5: newline please!

tests/wasi-wast/wasi/unstable/unix_open_special_files.wast
```diff
@@ -0,0 +1,5 @@
+(wasi_test "unix_open_special_files.wasm"
+  (map_dirs "/dev:/dev")
+  (assert_return (i64.const 0))
+  (assert_stdout "13\n")
+)
```
nlewycky, line 5: newline please

lib/wasi/src/state/mod.rs
```diff
@@ -1196,7 +1239,18 @@ impl WasiFs {
         is_preopened: bool,
         name: String,
     ) -> Inode {
-        let mut stat = __wasi_filestat_t::default();
+        let stat = __wasi_filestat_t::default();
+        self.create_inode_with_stat(kind, is_preopened, name, stat)
+    }
+
+    /// creates an inode with the given filestat and insert it.
```
nlewycky, line 1246:
```suggestion
    /// Creates an inode with the given filestat and insert it.
```
Rule of thumb is that either it's a sentence which starts with a capital and ends with a period (or question or exclamation mark) or it's a sentence fragment and has neither.

lib/wasi/src/state/mod.rs
```diff
@@ -819,6 +819,56 @@ impl WasiFs {
...
+                                    if let Kind::Dir {
+                                        ref mut entries, ..
+                                    } = &mut self.inodes[cur_inode].kind
+                                    {
+                                        entries.insert(
+                                            component.as_os_str().to_string_lossy().to_string(),
+                                            new_inode,
+                                        );
+                                    } else {
+                                        unreachable!(
```
nlewycky, line 864: Just a comment, I'm not sure whether this is reachable or not? If the place in the wasi-fs where the host path is being injected is controlled by the user then it seems like it should be a full Err? Otherwise, if this is part of mirroring the host paths then this is fine as is.

tests/wasi-wast/wasi/unstable/unix_open_special_files.wast
MarkMcCaskey: these are generated files, I can regenerate them all but it'd probably be best to do that separately

lib/wasi/src/state/mod.rs
MarkMcCaskey: It shouldn't be reachable without unsafe I believe: it's not possible to put files in the virtual root right now, they have to come from a preopened directory and that directory will be mounted at `/`.

Issue #1568, opened 2020-08-29
Dangling pointer in JIT Engine


sebpuetz:
### Describe the bug

Initially reported by @hywan at https://github.com/PyO3/pyo3/issues/1120

-----

This test causes abort on linux:

~~~Rust
    #[test]
    fn test2() {
        let store = wasmer::Store::default();
        let module = wasmer::Module::new(&store, "(module)").unwrap();
        mem::drop(store);
        mem::drop(module); // abort
    }
~~~

Stepping through the code  shows that the abort happens while dropping `wasmer_engine_jit::unwind::systemv::UnwindRegistry`. More specifically, the last call before abort is `__deregister_frame` in  `UnwindRegistry::drop`.

`Artifact` holds a pointer to a `Vec` owned by `Store` that gets passed to `__deregister_frame`. If the `Store` goes out first, the pointer in `Artifact`'s `UnwindRegistry` becomes invalid but gets used anyways.

Allocation happens here: https://github.com/wasmerio/wasmer/blob/80290e48055655a63f05f2c329f2ace4adbc712f/lib/engine-jit/src/artifact.rs#L165-L166

Subverting lifetimes happens here: https://github.com/wasmerio/wasmer/blob/80290e48055655a63f05f2c329f2ace4adbc712f/lib/engine-jit/src/unwind/systemv.rs#L86

Assuming dead-pointers are still alive happens here: https://github.com/wasmerio/wasmer/blob/80290e48055655a63f05f2c329f2ace4adbc712f/lib/engine-jit/src/unwind/systemv.rs#L104-L106

A fix is switching the order of the struct members in `wasmer::Module` since dropping happens in reverse order. It'd probably be cleaner to keep the `Vec` alive independently of member order.

~~~Rust
pub struct Module {
    artifact: Arc<dyn Artifact>,
    store: Store,
}
~~~

### Steps to reproduce

Create project with test and run cargo test.

### Expected behavior
Exit succesfully

### Actual behavior
(signal: 6, SIGABRT: process abort signal)

sebpuetz:
Adding 
~~~Rust
impl Drop for CodeMemory {
    fn drop(&mut self) {
        for section in self.read_sections.iter() {
            println!("read section: {:p}", section.as_ptr());
        }
    }
}
~~~

and this print demonstrates that the Vec is dropped before the pointer to it gets passed to `__deregister_frame`:
~~~Rust
impl Drop for UnwindRegistry {
[...]

PR #1580, opened 2020-08-31
test: Test the JIT engine being dropped twice through the Store

Hywan:
I think we have a bug in the `engine-jit` where the `UnwindRegistry` on Linux (Ubuntu) fails when dropped twice.

This PR adds a test to check that.

### DIFF
--- a/lib/api/tests/store.rs
+++ b/lib/api/tests/store.rs
@@ -1,0 +1,19 @@
+use anyhow::Result;
+use wasmer::*;
+
+#[test]
+fn store_dropped_twice() -> Result<()> {
+    let wat = r#"(module)"#;
+
+    {
+        let store = Store::default();
+        let _module = Module::new(&store, wat)?;
+    }
+
+    {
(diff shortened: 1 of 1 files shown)


Hywan:
bors try

Hywan:
bors try

Hywan:
bors cancel

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

syrusakbary:
bors try

sebpuetz:
Just wondering if you saw #1568: The `UnwindRegistry` is **not** dropped twice. A `Vec<u8>` owned by `CodeMemory` inside the `Store` containing the empty frame gets dropped before deregistering the frames. This `Vec`'s buffer then gets overwritten and the orginally 0-length entry gets changed to some nonsense which causes the abort.

Hywan:
Yup, this PR was just a test!

PR #1582, opened 2020-08-31
fix(api) Drop `Store` before `Artifact` when dropping `Module`.

Hywan:
When `Module` is dropped, its fields are dropped in reverse order. The
store must be dropped first, as the engine might holds pointers to the
artifact, with specific `Drop` implementations. In order to avoid
having dangling pointers, we must drop `store` first.

Fixed found by @sebpuetz in https://github.com/wasmerio/wasmer/issues/1568. Thank you!

### DIFF
--- a/lib/api/src/module.rs
+++ b/lib/api/src/module.rs
@@ -30,10 +30,15 @@ pub enum IoCompileError {
 ///
 /// Cloning a module is cheap: it does a shallow copy of the compiled
 /// contents rather than a deep copy.
+// Note on the field order. When `Module` is dropped, its fields are
+// dropped in reverse order. The store must be dropped first, as the
+// engine might holds pointers to the artifact, with specific `Drop`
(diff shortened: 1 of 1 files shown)


Hywan:
We should add tests for that.

syrusakbary (approved):

syrusakbary:
Should we add this test? (reported in #1568)

```rust
    #[test]
    fn test2() {
        let store = wasmer::Store::default();
        let module = wasmer::Module::new(&store, "(module)").unwrap();
        mem::drop(store);
        mem::drop(module); // abort
    }
```

Hywan:
It's not sufficient. I'm not able to reproduce in Rust only. I'm able to reproduce it with Python though. Maybe we can just assert that the order of dropping is correct. I've an idea for that.

Hywan:
Outdated by https://github.com/wasmerio/wasmer/pull/1628.

PR #1585, opened 2020-08-31
Add "object file" engine

MarkMcCaskey:
Engine name is a work in progress

# Review


lib/c-api/src/wasm_c_api/mod.rs
```diff
@@ -120,6 +156,12 @@ pub unsafe extern "C" fn wasm_instance_new(
 #[no_mangle]
 pub unsafe extern "C" fn wasm_instance_delete(_instance: Option<Box<wasm_instance_t>>) {}
 
+// TODO: NOT part of the standard Wasm C API
+#[no_mangle]
+pub unsafe extern "C" fn wasm_instance_get_vmctx_ptr(instance: &wasm_instance_t) -> *mut c_void {
```
syrusakbary, line 161: Where / How this is used?

lib/cli/src/commands/compile.rs
```diff
@@ -83,6 +94,19 @@ impl Compile {
...
         let module = Module::from_file(&store, &self.path)?;
         let _ = module.serialize_to_file(&self.output)?;
+        #[cfg(feature = "object-file")]
+        if let Some(header_file_src) = module.artifact().create_header_file() {
+            // for C code
+            let mut header = std::fs::OpenOptions::new()
+                .create(true)
+                .truncate(true)
+                .write(true)
+                .open("test.h")?;
```
syrusakbary, line 104: I think it would be good to receive a filename from the CLI args rather than always using `test.h`

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -111,7 +111,7 @@ impl FuncTranslator {
         // TODO: mark vmctx nofree
         func.add_attribute(AttributeLoc::Function, intrinsics.stack_probe);
         func.set_personality_function(intrinsics.personality);
-        func.as_global_value().set_section(FUNCTION_SECTION);
+        //func.as_global_value().set_section(FUNCTION_SECTION);
```
syrusakbary, line 114: cc @nlewycky

lib/compiler-llvm/src/trampoline/wasm.rs
```diff
@@ -63,9 +63,9 @@ impl FuncTrampoline {
         );
 
         let trampoline_func = module.add_function(name, trampoline_ty, Some(Linkage::External));
-        trampoline_func
-            .as_global_value()
-            .set_section(FUNCTION_SECTION);
+        /*trampoline_func
+        .as_global_value()
+        .set_section(FUNCTION_SECTION);*/
```
syrusakbary, line 68: Is this intended?

lib/compiler/src/compiler.rs
```diff
@@ -99,6 +99,24 @@ pub trait Compiler {
     ) -> Option<Result<Vec<u8>, CompileError>> {
         None
     }
+
+    /// Compiles a module into an object file that can be built with libwasmer
+    /// into a stand-alone executable.
+    ///
+    /// It returns the bytes as a `&[u8]` or a [`CompileError`].
+    fn experimental_object_file_compile_module<'data, 'module>(
```
syrusakbary, line 107: I believe we don't need this. Probably we can replicate what the native engine is doing for generating the object file. Thoughts @MarkMcCaskey ?

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,647 @@
...
+        });
+        c_statements.push(CStatement::Declaration {
+            name: "WASMER_METADATA".to_string(),
+            array: true,
+            is_extern: true,
+            is_const: true,
+            ctype: CType::U8,
+            definition: None,
+        });
+        for (function_local_index, _function_len) in self.metadata.function_body_lengths.iter() {
```
syrusakbary, line 262: Walking through body lengths to use the function local index only seems a bit of misuse.

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,647 @@
...
+        };
+
+        //let host_target = Triple::host();
+        //let is_cross_compiling = target_triple != &host_target;
+
+        Self::from_parts_crosscompiled(&mut *engine_inner, metadata, obj_bytes, metadata_length)
+    }
+
+    /// Generate the header file that goes with the generated object file.
+    pub fn generate_header_file(&self) -> String {
```
syrusakbary, line 241: I believe we can put this as a utility function outside of the Artifact, this will help on having testing, for example. Or using that header-function in an isolated way.

A function that given a `ModuleInfo` returns the String.
I think we could generate everything needed from that `ModuleInfo` (rather than taking &self).

lib/c-api/src/wasm_c_api/mod.rs
MarkMcCaskey: It's needed to call the trampolines directly -- currently I'm using it in a test C file I made that I didn't commit

lib/engine-object-file/src/artifact.rs
MarkMcCaskey: I agree, I copied this code from native engine though haha

lib/compiler-llvm/src/trampoline/wasm.rs
MarkMcCaskey: Yeah, I needed this to get the symbols to show up with objdump... it looks like it still works if I uncomment it though

lib/compiler/src/compiler.rs
MarkMcCaskey: Yeah, I agree, right now the code paths are the same, I thought I was going to have to do different things is why I split it out

lib/engine-object-file/src/artifact.rs
MarkMcCaskey: Hmm, well we need all the symbol names to put into the header file which is currently part of the engine/artifact, maybe there's a way to make that work though

lib/engine/src/artifact.rs
```diff
@@ -71,6 +71,11 @@ pub trait Artifact: Send + Sync {
         Ok(())
     }
 
+    /// Create a header file for use with a native object file;
+    fn create_header_file(&self) -> Option<String> {
```
syrusakbary, line 75: I'm not sure that we should modify the engine. It seems this case is only needed for this specific engine?

MarkMcCaskey: Yeah the issue is that it needs access to the internals.

The way I see it:
- We can add trait methods that allow us to downcast and get specific implementations out of `dyn Artifact`
- We can have a method like `create_header_file`
- We can attempt to change the calling code so we don't deal with abstract types, we deal with `ObjectFileEngine` directly
- We can restructure the way symbol naming in object files work and separate the logic of header file generation from this engine
- We can expose an API for symbol naming via `Artifact` so the header file generation code can get the information it needs

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,647 @@
...
+            is_extern: true,
+            is_const: true,
+            ctype: CType::U8,
+            definition: None,
+        });
+        for (function_local_index, _function_len) in self.metadata.function_body_lengths.iter() {
+            let function_name = self
+                .metadata
+                .symbol_to_name(Symbol::LocalFunction(function_local_index));
+            // TODO: figure out the signtaure here too
```
nlewycky, line 266:
```suggestion
            // TODO: figure out the signature here too
```

syrusakbary:
I believe we can get things simpler if we make `generate_header_file` non depending on self (just on two params that doesn't depend on the artifact). This will remove the need for `downcast_ref` and also simplify other things.

But, how we can achieve it?
It should be simple if we make `generate_header_file` static and receiving only the arguments: `ModuleInfo` and `SymbolRegistry`.

```rust
pub fn generate_header_file(module_info: &ModuleInfo, registry: &dyn SymbolRegistry) -> String {
    ...
}
```

lib/engine-object-file/src/artifact.rs
syrusakbary: Just added a comment in the PR that guides on how to solve that (hopefully).

https://github.com/wasmerio/wasmer/pull/1585#issuecomment-686587245

lib/engine-object-file/README.md
```diff
@@ -0,0 +1,186 @@
...
+        }
+        wasi_env_set_instance(wasi_env, instance);
+        
+        // WASI is now set up.
+
+        void* vmctx = wasm_instance_get_vmctx_ptr(instance);
+        wasm_val_t* inout[2] = { NULL, NULL };
+
+        // We're able to call our compiled functions directly through their trampolines.
+        wasmer_trampoline_function_call__1(vmctx, wasmer_function__1, &inout);
```
syrusakbary, line 163: We should probably just use the normal wasm c api for calling functions.
Probably this code is not needed?

lib/c-api/src/wasm_c_api/mod.rs
syrusakbary: Why someone would call the trampolines directly? I think it's better to enforce users to use the normal wasm-c-api to call wasm functions.
Thoughts @MarkMcCaskey

lib/engine-object-file/README.md
```diff
@@ -0,0 +1,186 @@
...
+#include "my_wasm.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+
+void wasmer_function__1(void);
+void wasmer_trampoline_function_call__1(void*, void*, void*);
+
+// todo: add to wasmer_wasm.h or something
+void* wasm_instance_get_vmctx_ptr(wasm_instance_t*);
```
syrusakbary, line 39: In general, I'm a bit wary of exposing a function like this. I don't think we should expose the vmctx to the user, since it leaks internal structures

MarkMcCaskey: I see what you're saying, but it's required to be able to call things directly and the vmctx is completely opaque from the perspective of the user. I'll get back to this after investigating the tradeoff of calling through the Wasm C API

lib/engine-jit/src/artifact.rs
```diff
@@ -303,4 +303,8 @@ impl Artifact for JITArtifact {
         serialized.extend(bytes);
         Ok(serialized)
     }
+
+    fn symbol_registry(&self) -> &dyn SymbolRegistry {
+        unimplemented!("TODO: figure out why engine JIT doesn't have a SymbolRegistry")
```
syrusakbary, line 308: `SymbolRegistry` only makes sense for object files. JIT would never need it.

lib/engine-object-file/README.md
syrusakbary: Let's add a comment in the code about this, ship that comment with the code as TODO and then add tests and merge the PR (once reviewed).

tests/integration/cli/src/compile.rs
```diff
@@ -0,0 +1,162 @@
...
+        match self {
+            Engine::Jit => "--jit",
+            Engine::Native => "--native",
+            Engine::ObjectFile => "--object-file",
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+pub enum Backend {
```
syrusakbary, line 27: Can we name this Compiler rather than backend? (I know, it's also hard for me, I was so used to backend).

tests/integration/cli/src/object_file_engine_test_c_source.c
```diff
@@ -0,0 +1,143 @@
...
+        }
+        wasi_env_set_instance(wasi_env, instance);
+        
+        // WASI is now set up.
+
+        void* vmctx = wasm_instance_get_vmctx_ptr(instance);
+        wasm_val_t* inout[2] = { NULL, NULL };
+
+        // We're able to call our compiled functions directly through their trampolines.
+        wasmer_trampoline_function_call__1(vmctx, wasmer_function__1, &inout);
```
syrusakbary, line 136: Should we check the results to make sure we got the right one? (or that the stdout is what we expect)

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/engine-object-file/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "wasmer-engine-object-file"
+version = "1.0.0-alpha3"
+authors = ["Wasmer Engineering Team <engineering@wasmer.io>"]
+description = "Wasmer Object File Engine"
+categories = ["wasm"]
+keywords = ["webassembly", "wasm"]
+repository = "https://github.com/wasmerio/wasmer"
+# TODO: make a readme explaining what this crate is
```
Hywan, line 9: This `TODO` can be removed I guess.

lib/engine-object-file/README.md
```diff
@@ -0,0 +1,186 @@
+# Wasmer Engine Object File
+
+This is an [engine](https://crates.io/crates/wasmer-engine) for the [wasmer](https://crates.io/crates/wasmer/1.0.0-alpha01.0) WebAssembly VM.
```
Hywan, line 3: alpha01?

syrusakbary:
```suggestion
This is an [engine](https://crates.io/crates/wasmer-engine) for the [wasmer](https://crates.io/crates/wasmer/1.0.0-alpha3) WebAssembly VM.
```

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors  try-

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

MarkMcCaskey:
@hywan has asked to not merge changes to the C API into master until #1641 merges; commenting here so we don't merge this for now

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/engine-jit/src/artifact.rs
syrusakbary: Ping: please fix this! :)

lib/engine/src/artifact.rs
```diff
@@ -147,4 +147,10 @@ pub trait Artifact: Send + Sync {
             .finish_instantiation(&data_initializers)
             .map_err(|trap| InstantiationError::Start(RuntimeError::from_trap(trap)))
     }
+
+    /// Get the `SymbolRegistry` used to generate the names used in the Artifact.
+    fn symbol_registry(&self) -> &dyn SymbolRegistry;
+
+    /// The length in bytes of the metadata in the serialized output.
+    fn metadata_length(&self) -> usize;
```
syrusakbary, line 155: It seems a bit wrong that we have this on the artifact

lib/cli/src/commands/compile.rs
```diff
@@ -81,12 +96,49 @@ impl Compile {
...
         let module = Module::from_file(&store, &self.path)?;
         let _ = module.serialize_to_file(&self.output)?;
         eprintln!(
             "✔ File compiled successfully to `{}`.",
             self.output.display(),
         );
+
+        #[cfg(feature = "object-file")]
+        if engine_type == EngineType::ObjectFile {
+            let symbol_registry = module.artifact().symbol_registry();
```
syrusakbary, line 109: Can we cast the artifact to `NativeArtifact` and access the methods there? I don't think all Artifacts should have `symbol_registry` or `metadata_length`

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,485 @@
...
+            let mut obj = get_object_for_target(&target_triple).map_err(to_compile_error)?;
+            emit_data(&mut obj, WASMER_METADATA_SYMBOL, &metadata_binary)
+                .map_err(to_compile_error)?;
+            emit_compilation(&mut obj, compilation, &metadata, &target_triple)
+                .map_err(to_compile_error)?;
+            obj.write().map_err(to_compile_error)?
+        };
+
+        //let host_target = Triple::host();
+        //let is_cross_compiling = target_triple != &host_target;
```
Hywan, line 235:
```suggestion
```

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,485 @@
...
+    pub fn from_parts_crosscompiled(
+        engine_inner: &mut ObjectFileEngineInner,
+        metadata: ModuleMetadata,
+        module_bytes: Vec<u8>,
+        metadata_length: usize,
+    ) -> Result<Self, CompileError> {
+        let finished_functions: PrimaryMap<LocalFunctionIndex, FunctionBodyPtr> = PrimaryMap::new();
+        let finished_dynamic_function_trampolines: PrimaryMap<FunctionIndex, FunctionBodyPtr> =
+            PrimaryMap::new();
+        //let signatures: PrimaryMap<SignatureIndex, VMSharedSignatureIndex> = PrimaryMap::new();
```
Hywan, line 258:
```suggestion
```

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,485 @@
...
+                byte_buffer[j] = bytes[cur_offset + j];
+            }
+            sp.ptr = usize::from_ne_bytes(byte_buffer);
+            cur_offset += WORD_SIZE;
+            // REVIEW: we can also serialize and read back lengths, do we want to do this?
+            /*for j in 0..WORD_SIZE {
+                byte_buffer[j] = bytes[cur_offset + j];
+            }
+                sp.len = usize::from_ne_bytes(byte_buffer);
+                cur_offset += WORD_SIZE;*/
```
Hywan, line 341:
```suggestion
```

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,485 @@
...
+            let vm_shared_idx = signature_registry.register(&func_type);
+            sig_map.insert(sig_idx, vm_shared_idx);
+
+            let mut sp = SlicePtr { ptr: 0, len: 0 };
+            for j in 0..WORD_SIZE {
+                byte_buffer[j] = bytes[cur_offset + j];
+            }
+            sp.ptr = usize::from_ne_bytes(byte_buffer);
+            cur_offset += WORD_SIZE;
+            // REVIEW: we can also serialize and read back lengths, do we want to do this?
```
Hywan, line 336: What do you mean?

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,485 @@
...
+
+            let func_type = &metadata.compile_info.module.signatures[SignatureIndex::new(i)];
+
+            engine_inner.add_trampoline(func_type, trampoline);
+            // REVIEW: we can also serialize and read back lengths, do we want to do this?
+            /*for j in 0..WORD_SIZE {
+                byte_buffer[j] = bytes[cur_offset + j];
+            }
+                sp.len = usize::from_ne_bytes(byte_buffer);
+                cur_offset += WORD_SIZE;*/
```
Hywan, line 378:
```suggestion
```

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,485 @@
...
+                byte_buffer[j] = bytes[cur_offset + j];
+            }
+            cur_offset += WORD_SIZE;
+            let trampoline_ptr_bytes = usize::from_ne_bytes(byte_buffer);
+            let trampoline = mem::transmute::<usize, VMTrampoline>(trampoline_ptr_bytes);
+
+            let func_type = &metadata.compile_info.module.signatures[SignatureIndex::new(i)];
+
+            engine_inner.add_trampoline(func_type, trampoline);
+            // REVIEW: we can also serialize and read back lengths, do we want to do this?
```
Hywan, line 373: Same question as above.

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,485 @@
...
+            }
+            sp.ptr = usize::from_ne_bytes(byte_buffer);
+            cur_offset += WORD_SIZE;
+
+            // REVIEW: we can also serialize and read back lengths, do we want to do this?
+            /*for j in 0..WORD_SIZE {
+                byte_buffer[j] = bytes[cur_offset + j];
+            }
+                sp.len = usize::from_ne_bytes(byte_buffer);
+                cur_offset += WORD_SIZE;*/
```
Hywan, line 401:
```suggestion
```

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,485 @@
...
+        let num_dynamic_trampoline_functions = usize::from_ne_bytes(byte_buffer);
+        for _i in 0..num_dynamic_trampoline_functions {
+            let mut sp = SlicePtr { ptr: 0, len: 0 };
+            for j in 0..WORD_SIZE {
+                byte_buffer[j] = bytes[cur_offset + j];
+            }
+            sp.ptr = usize::from_ne_bytes(byte_buffer);
+            cur_offset += WORD_SIZE;
+
+            // REVIEW: we can also serialize and read back lengths, do we want to do this?
```
Hywan, line 396: Same question as above.

lib/engine-object-file/src/artifact.rs
```diff
@@ -0,0 +1,485 @@
...
+
+            let fp = FunctionBodyPtr(mem::transmute(sp));
+
+            finished_dynamic_function_trampolines.push(fp);
+        }
+
+        Ok(Self {
+            metadata,
+            // TODO: review
+            module_bytes: vec![],
```
Hywan, line 411: I would prefer `None` rather than an empty vec. Thoughts?

lib/engine-object-file/src/engine.rs
```diff
@@ -0,0 +1,252 @@
...
+    pub fn trampoline(&self, sig: VMSharedSignatureIndex) -> Option<VMTrampoline> {
+        self.trampolines.get(&sig).cloned()
+    }
+
+    pub(crate) fn add_trampoline(&mut self, func_type: &FunctionType, trampoline: VMTrampoline) {
+        let index = self.signatures.register(&func_type);
+        // We always use (for now) the latest trampoline compiled
+        // TODO: we need to deallocate trampolines as the compiled modules
+        // where they belong become unallocated.
+        self.trampolines.insert(index, trampoline);
```
Hywan, line 250: Still todo?

Hywan (commented):
Excellent work. I'm impressed by everything. Congrats!

I've 41 comments, but nothing super important, it's just minor stuff, except for the C header generation, where I think introducing a “global suffix” for all functions and types could help to “brand” code from our users, and clarify things.

lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -1,44 +1,118 @@
...
-pub extern "C" fn wasm_config_new() -> *mut wasm_config_t {
-    todo!("wasm_config_new")
-    //ptr::null_mut()
+pub extern "C" fn wasm_config_new() -> Box<wasm_config_t> {
+    Box::new(wasm_config_t::default())
 }
 
-#[allow(non_camel_case_types)]
+#[no_mangle]
+pub extern "C" fn wasm_config_set_compiler(
```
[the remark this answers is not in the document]
MarkMcCaskey: The reason I used `wasm_config` is to keep the pattern of it operating on the `wasm_config_t` type.  We could do `wasmer_wasm_config_set_compiler` if you prefer for all custom methods, but it's not a `wasmer_config` in my opinion!

lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -1,44 +1,118 @@
...
+    config: &mut wasm_config_t,
+    compiler: wasmer_compiler_t,
+) {
+    config.compiler = compiler;
+}
+
+#[no_mangle]
+pub extern "C" fn wasm_config_set_engine(config: &mut wasm_config_t, engine: wasmer_engine_t) {
```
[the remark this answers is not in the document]
MarkMcCaskey: These methods are for use with `wasm_engine_new_with_config`, without this method there's no way to select an engine!

See https://github.com/wasmerio/wasmer/pull/1585#discussion_r500495760 for my thoughts on `wasmer_config`

lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -1,44 +1,118 @@
...
+    config.engine = engine;
+}
+
+/// cbindgen:ignore
+#[repr(C)]
```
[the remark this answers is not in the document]
MarkMcCaskey: Sure, it feels a bit weird to not being using `repr(C)` types when dealing with pointers and C FFI but yeah if it's opaque it's probably fine

lib/c-api/src/wasm_c_api/wasmer.rs
```diff
@@ -0,0 +1,9 @@
+//! Wasmer-specific extensions to the Wasm C API.
```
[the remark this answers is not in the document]
MarkMcCaskey: > Should we rename those functions wasmer_…?

possibly, I don't know how much it matters: it probably plays better with others in regards to linking with conditional compilation and all that... but if users are using multiple producers of the Wasm C API in the same binary, then don't they already have that problem?

It would be good to call out wasmer specific functions I suppose.. I'm undecided on this!

> Should we include engine and config functions, and compiler/engine enums too?

I thought about this and decided against it.  The decision was fairly arbitrary but I thought that encapsulating/keeping code near where it's used made more sense than putting it all in one file. I'm open to changing my mind on this or considering alternatives though if you feel strongly about one specific way

lib/c-api/src/wasm_c_api/engine.rs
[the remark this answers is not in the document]
syrusakbary: I prefer `wasmer_config_set_compiler` over `wasmer_wasm_config_set_compiler` (I'm more agnostic on `wasm_` vs `wasmer_`).
We can assume wasmer will touch wasm stuff :)

lib/cli/src/c_gen/mod.rs
```diff
@@ -0,0 +1,537 @@
...
+        target_type: CType,
+        /// The thing being cast.
+        expression: Box<CStatement>,
+    },
+    /// Typedef one type to another.
+    TypeDef {
+        /// The type of the thing being typedef'd.
+        source_type: CType,
+        /// The new name by which this type may be called.
+        new_name: String,
```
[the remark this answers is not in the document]
MarkMcCaskey: That's a good idea -- didn't consider it

lib/c-api/build.rs
```diff
@@ -374,4 +374,9 @@ fn exclude_items_from_wasm_c_api(builder: Builder) -> Builder {
         .exclude_item("wasi_get_start_function")
         .exclude_item("wasi_get_wasi_version")
         .exclude_item("wasi_version_t")
+        .exclude_item("wasm_instance_get_vmctx_ptr")
```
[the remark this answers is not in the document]
MarkMcCaskey: Perhaps I misunderstood how the header file generation works, but my understanding is that we wanted it to be not ignored so that it's part of `wasmer_wasm.h`.

Or are you saying that we can specify `wasmer_c_api::wasmer` in `exclude_item` and everything just works?

lib/c-api/src/wasm_c_api/mod.rs
```diff
@@ -29,3 +28,5 @@ pub mod value;
 
 #[cfg(feature = "wasi")]
 pub mod wasi;
+
+pub mod wasmer;
```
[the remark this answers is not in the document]
MarkMcCaskey: I thought we couldn't do that because we want these things in `wasmer_wasm.h` 🤔

lib/engine-object-file/README.md
```diff
@@ -0,0 +1,183 @@
...
+```C
+#include "wasmer_wasm.h"
+#include "wasm.h"
+#include "my_wasm.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+
+void wasmer_function__1(void);
+void wasmer_trampoline_function_call__1(void*, void*, void*);
```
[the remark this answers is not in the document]
MarkMcCaskey: Add comments to the generated header file? Probably a good idea

lib/engine-object-file/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+// a bit of a hack; TODO: clean this up
+typedef struct my_byte_vec_t {
+        size_t size;
+        char* data;
+} my_byte_vec_t;
```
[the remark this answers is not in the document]
MarkMcCaskey: Yeah that works!  I was under the impression that it didn't work for some reason... I've cleaned it up!

lib/engine-object-file/README.md
MarkMcCaskey: There's a series of type requirements that we can't meet:

Quoting from when I wrote about it on Slack previously:

----

`wasm_func_callback_t` = `struct wasm_trap_t *(*)(const struct wasm_val_t *, struct wasm_val_t *)`.
`wasm_func_new` takes `wasm_func_callback_t` + `wasm_functype_t`  and returns `wasm_func_t`
`wasm_func_call` takes `wasm_func_t` + args + rets returns `wasm_trap_t*`

```
#[repr(C)]
pub struct wasm_func_t {
    inner: Function,
    // this is how we ensure the instance stays alive
    instance: Option<Arc<Instance>>,
}
```
Is our current definition of `wasm_func_t`

We don’t have anything of the correct type to use our functions as `wasm_func_t` right now
it may be possible to change `wasm_func_t` and/or make a `wasmer_func_new` or something to make things work
but it’s invalid to cast any of our pointers to a `wasm_func_callback_t` or `wasm_func_t` right now

MarkMcCaskey: Following up on my previous comment, now that we have traps implemented... it's probably possible to generate wrapper functions so that we can have a bunch of `wasm_func_t`s... we'd need a helper function or two 🤔 ; fundamentally what we're doing is making a trampoline so that we can make another trampoline for it though... so perhaps having a wasmer-specific lower level method to get a `wasm_func_t` directly is the best option here

lib/c-api/src/wasm_c_api/engine.rs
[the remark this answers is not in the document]
Hywan: Actually, I prefer `wasmer_wasm_config_set_compiler` too. I don't have an irrevocable opinion though.

lib/engine-object-file/src/artifact.rs
MarkMcCaskey: This has to do with the FunctionBodyLengths... currently they're all 0 and everything works, I wasn't sure if we needed it

lib/engine-object-file/src/engine.rs
MarkMcCaskey: Yeah looks like it... this code is duplicated from `engine-native` and seems to be just an oversight on both.  I don't really have context on it right now

lib/c-api/src/wasm_c_api/wasmer.rs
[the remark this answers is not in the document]
Hywan: I don't have a strong opinion on this one. I trust you to take the best decision!

lib/c-api/build.rs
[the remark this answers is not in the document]
Hywan: This `wasm_instance_get_vmctx_ptr` is already marked as `cbindgen:ignore` right? So you don't need to exclude this item for the `wasmer.h*` files.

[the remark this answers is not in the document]
Hywan: Your next comment makes me realize that `wasm_instance_get_vmctx_ptr` is a public function.

Forget my comment!

lib/c-api/src/wasm_c_api/mod.rs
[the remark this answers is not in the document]
Hywan: Resolved by https://github.com/wasmerio/wasmer/pull/1585#discussion_r500545167.

lib/engine-object-file/README.md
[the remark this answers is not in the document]
Hywan: Yup. And inside the `README.md` too, please.

lib/engine-object-file/src/artifact.rs
MarkMcCaskey: Yeah, or perhaps we should just store the bytes we deserialized with. That seems like a less leaky abstraction, in general I'd expect a Serialized + Deserialized thing to be identical and for a deserialized thing to be able to be serialized again

lib/engine-object-file/README.md
```diff
@@ -35,26 +39,26 @@ Now lets create a program to link with this object file.
...
     wasmer_last_error_message(error_str, error_len);
     printf("Error str: `%s`\n", error_str);
 }
 
 
 int main() {
         printf("Initializing...\n");
-        wasm_engine_t* engine = wasm_engine_new();
+        wasm_config_t* config = wasm_config_new();
+        wasm_config_set_compiler(config, CRANELIFT);
```
syrusakbary, line 59: This seem a bit wrong. We should create a file object **headless** engine.

tests/integration/cli/src/object_file_engine_test_c_source.c
```diff
@@ -0,0 +1,145 @@
...
+    wasmer_last_error_message(error_str, error_len);
+    printf("Error str: `%s`\n", error_str);
+}
+
+
+int main() {
+        printf("Initializing...\n");
+        wasm_config_t* config = wasm_config_new();
+        wasm_config_set_compiler(config, CRANELIFT);
+        wasm_config_set_engine(config, OBJECT_FILE);
```
syrusakbary, line 33: This seem a bit wrong. We should create a file object **headless** engine.

syrusakbary: Meaning: it should not require any compiler attached to Wasmer

tests/integration/cli/src/object_file_engine_test_c_source.c
```diff
@@ -0,0 +1,145 @@
...
+        current_offset += sizeof(size_t);
+
+        memcpy(memory_buffer + current_offset, (void*)&function_trampolines[0], sizeof(function_trampolines));
+        current_offset += sizeof(function_trampolines);
+
+        memcpy(memory_buffer + current_offset, (void*)&num_dynamic_function_trampoline_pointers, sizeof(size_t));
+        current_offset += sizeof(size_t);
+
+        memcpy(memory_buffer + current_offset, (void*)&dynamic_function_trampoline_pointers[0], sizeof(dynamic_function_trampoline_pointers));
+        current_offset += sizeof(dynamic_function_trampoline_pointers);
```
syrusakbary, line 80: Why do we need this?

lib/engine-object-file/README.md
```diff
@@ -0,0 +1,189 @@
...
+        current_offset += sizeof(size_t);
+
+        memcpy(memory_buffer + current_offset, (void*)&function_trampolines[0], sizeof(function_trampolines));
+        current_offset += sizeof(function_trampolines);
+
+        memcpy(memory_buffer + current_offset, (void*)&num_dynamic_function_trampoline_pointers, sizeof(size_t));
+        current_offset += sizeof(size_t);
+
+        memcpy(memory_buffer + current_offset, (void*)&dynamic_function_trampoline_pointers[0], sizeof(dynamic_function_trampoline_pointers));
+        current_offset += sizeof(dynamic_function_trampoline_pointers);
```
syrusakbary, line 107: I'm very confused by this. Why do we need it?

tests/integration/cli/src/object_file_engine_test_c_source.c
MarkMcCaskey: Yeah, currently whether or not an engine is headless is determined purely by whether the `libwasmer_c_api` was compiled with the `compiler` feature or not.

I agree that it'd be good to allow users to select it too.

syrusakbary: I mean that the example should not even mention a compiler here.

Can we just remove the line?
```
        wasm_config_set_compiler(config, CRANELIFT);
```

tests/integration/cli/src/object_file_engine_test_c_source.c
MarkMcCaskey: >         // We need to pass all the bytes as one big buffer so we have to do all this logic to memcpy
>        // the various pieces together from the generated header file.

Deserializing requires one contiguous array of bytes.  We have to pass this info to the ObjectFileEngine for it to deserialize.  We could make a `deserialize_vectored` function which doesn't require memcpying, it allows the bytes to come from multiple sources.

tests/integration/cli/src/object_file_engine_test_c_source.c
syrusakbary: In this kind of engine, the wasm is already precompiled.
So is not that users should be able to select it... is more that it doesn't make sense to select any compiler at all for this engine example.

MarkMcCaskey: Oh sure, we can remove it but it's functionally the same: it will default to using `Cranelift`, I was just being explicit but yeah we can remove it.

MarkMcCaskey: We can make it so that absence of a compiler means headless mode, but perhaps it's better to be explicit about it because most users will want to use a compiler most of the time

syrusakbary: Yeah, but not in the case of the object-file engine. In this case compilers are not needed nor desired

lib/engine-object-file/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+        char* byte_ptr = (char*)&WASMER_METADATA[0];
+
+        // We need to pass all the bytes as one big buffer so we have to do all this logic to memcpy
+        // the various pieces together from the generated header file.
+        //
+        // We should provide a `deseralize_vectored` function to avoid requiring this extra work.
+
+        size_t num_function_pointers
+                = sizeof(function_pointers) / sizeof(void*);
```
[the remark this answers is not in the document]
MarkMcCaskey: Yeah, I think this is a good idea to be able to use multiple header files at the same time: Syrus suggested keying the setup logic split into a helper function by some user defined key, like the name of the wasm file. I've added that but currently it's a no-op, more generally we will have duplicate symbols and I'm not sure how we'll want to handle that.  I think you're right that should probably be a core argument to `wasmer compile` for this use case.

lib/engine-object-file/README.md
```diff
@@ -0,0 +1,183 @@
...
+        size_t num_function_trampolines
+                = sizeof(function_trampolines) / sizeof(void*);
+        size_t num_dynamic_function_trampoline_pointers
+                = sizeof(dynamic_function_trampoline_pointers) / sizeof(void*);
```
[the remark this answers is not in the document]
MarkMcCaskey: I split this logic into helper functions that are now in the generated header! So now that logic belongs to us and we can update it as needed.

lib/engine-object-file/src/artifact.rs
MarkMcCaskey: It's apparently useful for debug output.  I'm going to not fix this in this PR (I spent some time trying to do so today and wasn't successful), but we can add it later -- I have comments in all the places that would need to be updated.

syrusakbary (approved):

PR #1590, opened 2020-09-03
Fix soundness issue in vm::Global

MarkMcCaskey:
The `unsafe impl` for `Send` and `Sync` relies on the global being
locked when it's being accessed; the get and get_mut functions did
not (and could not) do this, so we replace them with `get` and `set`
methods which operate on `Value`s directly.

Additionally the `get_mut` function took a `&self` and returned a
`&mut` to the underlying data which allows for aliased mutable
references to the same data which is another soundness issue.

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## **[Unreleased]**
+- [#1590](https://github.com/wasmerio/wasmer/pull/1590) Fix soundness issue in API on vm::Global
 - [#1566](https://github.com/wasmerio/wasmer/pull/1566) Add support for opening special Unix files to the WASI FS
 
 ## TODO: 1.0.0-alpha1.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## **[Unreleased]**
-- [#1590](https://github.com/wasmerio/wasmer/pull/1590) Fix soundness issue in API on vm::Global
+- [#1590](https://github.com/wasmerio/wasmer/pull/1590) Fix soundness issue in API of vm::Global
 - [#1566](https://github.com/wasmerio/wasmer/pull/1566) Add support for opening special Unix files to the WASI FS
 
 ## TODO: 1.0.0-alpha1.0
--- a/lib/api/src/externals/global.rs
+++ b/lib/api/src/externals/global.rs
@@ -43,7 +43,7 @@ impl Global {
         });
         unsafe {
             global
-                .set(val.clone())
+                .set_unchecked(val.clone())
                 .map_err(|e| RuntimeError::new(format!("create global for {:?}: {}", val, e)))?;
         };
 
@@ -76,25 +76,13 @@ impl Global {
     /// * The global is not mutable
     /// * The type of the `Val` doesn't matches the Global type.
     pub fn set(&self, val: Val) -> Result<(), RuntimeError> {
-        if self.ty().mutability != Mutability::Var {
-            return Err(RuntimeError::new(
-                "immutable global cannot be set".to_string(),
-            ));
-        }
-        if val.ty() != self.ty().ty {
-            return Err(RuntimeError::new(format!(
-                "global of type {:?} cannot be set to {:?}",
-                self.ty().ty,
-                val.ty()
-            )));
-        }
         if !val.comes_from_same_store(&self.store) {
             return Err(RuntimeError::new("cross-`Store` values are not supported"));
         }
         unsafe {
             self.global
                 .set(val)
-                .map_err(|e| RuntimeError::new(format!("Failed to set global: {}", e)))?;
+                .map_err(|e| RuntimeError::new(format!("{}", e)))?;
         }
         Ok(())
     }
--- a/lib/api/src/externals/global.rs
+++ b/lib/api/src/externals/global.rs
@@ -1,7 +1,7 @@
 use crate::exports::{ExportError, Exportable};
 use crate::externals::Extern;
 use crate::store::{Store, StoreObject};
-use crate::types::{Val, ValType};
+use crate::types::Val;
 use crate::GlobalType;
 use crate::Mutability;
 use crate::RuntimeError;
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## **[Unreleased]**
+- [#1590](https://github.com/wasmerio/wasmer/pull/1590) Fix soundness issue in API on vm::Global
```
nlewycky, line 4:
```suggestion
- [#1590](https://github.com/wasmerio/wasmer/pull/1590) Fix soundness issue in API of vm::Global
```

lib/api/src/externals/global.rs
```diff
@@ -42,14 +42,9 @@ impl Global {
...
-            match val {
-                Val::I32(x) => *global.get_mut().as_i32_mut() = x,
-                Val::I64(x) => *global.get_mut().as_i64_mut() = x,
-                Val::F32(x) => *global.get_mut().as_f32_mut() = x,
-                Val::F64(x) => *global.get_mut().as_f64_mut() = x,
-                Val::V128(x) => *global.get_mut().as_u128_bits_mut() = x.to_ne_bytes(),
-                _ => return Err(RuntimeError::new(format!("create_global for {:?}", val))),
-            }
+            global
+                .set(val.clone())
```
nlewycky, line 46: I wish we didn't need to clone here. I guess the only reason is for the error message?

lib/vm/src/global.rs
```diff
@@ -63,13 +63,57 @@ impl Global {
...
+    pub fn get<T>(&self) -> Value<T> {
+        let _global_guard = self.lock.lock().unwrap();
+        unsafe {
+            let definition = &*self.vm_global_definition.get();
+            match self.ty().ty {
+                Type::I32 => Value::from(*definition.as_i32()),
+                Type::I64 => Value::from(*definition.as_i64()),
+                Type::F32 => Value::F32(*definition.as_f32()),
+                Type::F64 => Value::F64(*definition.as_f64()),
+                Type::V128 => Value::V128(*definition.as_u128()),
```
nlewycky, line 76: The fact two of these use From and the other two use enumerator constructors bothers me a little. Any reason for the inconsistency?

lib/vm/src/global.rs
```diff
@@ -63,13 +63,57 @@ impl Global {
...
+                _ => unimplemented!("Global::get for {:?}", self.ty),
+            }
+        }
+    }
+
+    /// Set a value for the global.
+    ///
+    /// # Safety
+    /// The caller should check that the `val` comes from the same store as this global.
+    pub unsafe fn set<T>(&self, val: Value<T>) -> Result<(), GlobalError> {
```
nlewycky, line 86: This is still a setter on a non-mutable self. Is that really what we want?

lib/vm/src/global.rs
```diff
@@ -63,13 +63,57 @@ impl Global {
...
-    pub fn get_mut(&self) -> &mut VMGlobalDefinition {
-        unsafe { &mut *self.vm_global_definition.get() }
+    /// Set a value from the global (unchecked)
+    ///
+    /// # Safety
+    /// The caller should check that the `val` comes from the same store as this global.
+    /// The caller should also ensure that this global is synchronized. Otherwise, use
+    /// `set` instead.
+    pub unsafe fn set_unchecked<T>(&self, val: Value<T>) -> Result<(), GlobalError> {
+        // ideally we'd use atomics here
```
nlewycky, line 107: Is this a TODO? Do you mean that you wish the assignment into `*definition` was atomic? The comment mentions that this should be synchronized, do we still need atomics here even if it's synchronized, or are those separate?

lib/api/src/externals/global.rs
MarkMcCaskey: It's a relatively small type that doesn't allocate on Clone so I'm not too worried, but yeah, the only reason for it is the error message. for what it's worth, I'd expect Rust to optimize this decently and to produce roughly the same code as was being generated before

lib/vm/src/global.rs
MarkMcCaskey: didn't notice that! Good catch

lib/vm/src/global.rs
MarkMcCaskey: Yeah, it doesn't matter if it's mutable if it's using the `Mutex` and `UnsafeCell` internally; additionally the caller can't get a `&mut` to it without doing bad things

lib/vm/src/global.rs
MarkMcCaskey: Yeah, the idea is that we'd use atomics instead of a mutex on platforms with atomics that go up to 16 bytes... if there are any platforms that have that

MarkMcCaskey: I don't think it's particularly actionable, but I also don't think leaving that comment there is hurting anything, it might inspire a future person to try something!

nlewycky: I just think the comment is terse, it doesn't describe why an atomic is better than not atomic. Like, generally we don't use atomics and using atomics is slower on CPU, and makes for uglier code. So why the suggestion to use one here? What would it help? I'm trying to improve the comment "// TODO: if we used atomics here, we could ..." but I don't know enough to write that comment myself.

lib/vm/src/global.rs
nlewycky: If using `from` always works, we could move that call outside the match.

MarkMcCaskey: Good thought but each branch has to return the same value for that to work! So I don't think it'd be possible to structure this so that that worked

MarkMcCaskey: I mean you could probably do `&dyn Into<Value>` as the type that they all share, but it's a little tricky to make that work

nlewycky (approved):

syrusakbary:
Note that might be useful for the future:

Prior to this PR, the logic that was handling how to get/set values in Globals was done in the end layer (the wasmer API). We did that because once we add support for reference types, the final implementation of it will also be living on the wasmer API (not on the runtime). Moving it to the runtime might cause a leak of logic down the stream.

Moving the logic to the runtime (as this PR is doing) might cause difficulties on the long term when approaching reference types. But I guess we can solve for that when the time comes.

PR #1591, opened 2020-09-04
Add release workflow

MarkMcCaskey:
Setting things up so we can release; this will probably take a lot of iteration to get working properly so WIP

# Review


.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }} # This pulls from the CREATE RELEASE step above, referencing it's ID to get its outputs object, which include a `upload_url`. See this blog post for more info: https://jasonet.co/posts/new-features-of-github-actions/#passing-data-to-future-steps 
+          asset_path: wasmer-windows/wasmer-windows.exe
+          asset_name: wasmer-windows.exe
+          asset_content_type: application/vnd.microsoft.portable-executable
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: wasmer-ubuntu/wasmer.tar.gz
+          asset_name: wasmer-linux-amd64.tar.gz
+          asset_content_type: application/zip
```
syrusakbary, line 51: This seem wrong for tar.gz

.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: wasmer-macos/wasmer.tar.gz
+          asset_name: wasmer-darwin-amd64.tar.gz
+          asset_content_type: application/zip
```
syrusakbary, line 56: This seem wrong for tar.gz

.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+    - '[0-9]+\.[0-9]+\.[0-9]+.*' # Push events to matching v*, i.e. v1.0, v20.15.10
```
syrusakbary, line 5: Should we start forcing `vX.Y.Z` tags? (this is forcing `X.Y.Z`)

.github/workflows/main.yaml
```diff
@@ -129,9 +129,19 @@ jobs:
       - name: Build Wasmer binary
         run: |
           make build-wasmer
-          # TODO: build wapm
-          # make build-wapm
+      - name: Build Wapm binary
+        run: |
+          make build-wapm
```
syrusakbary, line 134: Note: build-wapm is currently empty I think. Might be worth to add a comment for now

.github/workflows/release.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+jobs:
+  build:
+    name: Upload Release Asset
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Download Artifacts
+        uses: actions/download-artifact@v2
```
syrusakbary, line 17: If it depends on downloading artifacts, we probably need to make this job dependent of the other one?

.github/workflows/release.yaml
MarkMcCaskey: Oh yeah, that comment is incorrect it's from the template I copied it from but I edited the regex, we can use a prefix of `v` if you want -- I think we may need to update some code in wapm if we do though

.github/workflows/release.yaml
MarkMcCaskey: Hmm, I searched for it before doing it but checking on my own I get:

```
file --mime-type ~/Downloads/sqlite.tar.gz
/Users/mark/Downloads/sqlite.tar.gz: application/gzip
```

Not sure if `gzip` vs `zip` is important

.github/workflows/release.yaml
MarkMcCaskey: Yeah, I was thinking the release workflow would be triggered independently -- perhaps it's possible to also make it dependent, though I'm not sure if that's possible.

There's a lot of ways to make a release process with github actions and I haven't fully decided which one I even really like yet.  We should probably discuss this for a few minutes, as you built the previous release pipeline

.github/workflows/main.yaml
MarkMcCaskey: It seems to be building wapm in my tests on my fork

.github/workflows/main.yaml
```diff
@@ -23,24 +9,45 @@ on:
       - 'master'
       - 'staging'
       - 'trying'
+    tags:
+      - '[0-9]+.[0-9]+.[0-9]+*'
```
nlewycky, line 13: 1. I'm concerned that this ends in `+*` which is uncommon in regex syntax (I would expect it to mean ((one or more of the preceding) zero or more of the preceding).

2. Optional: consider namespacing the release tag names, like `release/1.2.3`. 

Is this a full match or a partial match?

.github/workflows/main.yaml
```diff
@@ -129,9 +136,69 @@ jobs:
...
+        with:
+          name: wasmer-${{ matrix.os }}
+          path: dist
+
+  release:
+    needs: [setup, test]
+    runs-on: ubuntu-latest
+    if: needs.setup.outputs.DOING_RELEASE == '1'
+    steps:
+      - name: Download The Artifacts
```
nlewycky, line 158:
```suggestion
      - name: Download the Artifacts
```

.github/workflows/main.yaml
```diff
@@ -23,24 +9,45 @@ on:
...
+      VERSION: ${{ steps.setup.outputs.VERSION }}
+      DOING_RELEASE: ${{ steps.setup.outputs.DOING_RELEASE }}
+    steps:
+      - name: Set up env vars
+        id: setup
+        shell: bash
+        run: |
+          VERSION=${GITHUB_REF/refs\/tags\//}
+          echo ::set-output name=VERSION::${VERSION}
+          DOING_RELEASE=$(echo $VERSION | grep -q '[0-9]\+\.[0-9]\+\.[0-9]\+\(-.*\)\?' && echo -n '1' || echo -n '0')
```
nlewycky, line 28: Maybe
```suggestion
          DOING_RELEASE=$(echo $VERSION | grep -cq '[0-9]\+\.[0-9]\+\.[0-9]\+\(-.*\)\?')
```
?

MarkMcCaskey: We need the result as a string I think, I'm not 100% sure.

This line was a nightmare to get working correctly

MarkMcCaskey: Yeah, the env var isn't set if I do `-cq` with no echo

.github/workflows/main.yaml
MarkMcCaskey: I was thinking of this as a title and generally `the` isn't capitalized in titles

https://grammar.yourdictionary.com/capitalization/rules-for-capitalization-in-titles.html

MarkMcCaskey: Oh, that's what you're suggesting, I misread this as the opposite. 👍 for the fix

.github/workflows/main.yaml
MarkMcCaskey: Yeah, I agree this is bad, perhaps it needs a comment, but this isn't regex... it's a thing that looks like regex. `*` means  `[^/]+` (or maybe `[^/]*`) here I think 🤦

MarkMcCaskey: I actually can't express what I actually want (which I express in the grep command for the VERSION var) because it's not expressive enough, so I had to just replace the entire last part with `*`

MarkMcCaskey: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

nlewycky (approved):

PR #1600, opened 2020-09-08
wasmer_instance_call: allow null params

yjh0502:
allow null params if params_len is zero

# Description
wasmer_export_func_call allows null params when params_len is zero, but wasmer_instance_call is not. The PR makes wasmer_instance_call to accept null params too.

# Review


MarkMcCaskey (approved):
Looks good to me

PR #1602, opened 2020-09-08
Fix bug relating to type conversion in function calls

### LINKED ISSUE #1596: Regression in master: panic triggered by passing negative values to host functions
Passing `-1` as rights to `__wasi_path_open` triggers panic in the current master, works just fine in the latest release. 

```sh
wasmer 0.17.1 | rustc 1.46.0 (04488afe3 2020-08-24) | x86_64
```

```
$ (cd ~/wasmer.git && git describe --tags)
0.17.1-2043-gd7893fe5b
```

### Steps to reproduce

 1. `test.c`:

```c
#include <wasi/core.h>
#include <wasi/libc-find-relpath.h>
#include <string.h>
#include <stdio.h>

int main() {
  const char *p = "/tmp";
  int fd = __wasilibc_find_relpath(p, 0, 0, &p);
  __wasi_fd_t res;
  __wasi_errno_t rc;
  rc = __wasi_path_open(fd, 0, p, strlen(p),
                        __WASI_O_DIRECTORY, -1, -1, 0, &res);
  if (rc == 0) printf("fd %d\n", res);
  else printf("error %d\n", rc);
  return 0;
}
```

  2. `wasicc test.c -o test.wasm`

  3. run using Wasmer binary from master: `~/wasmer.git/target/release/wasmer run --dir=/ test.wasm`

### Expected behavior

A file descriptor is opened and `fd N` line is printed. This is what I'm getting with the latest release:

```
$ wasmer run --dir=/ test.wasm
fd 5
```

### Actual behavior

Wasmer binary from master panics:

```
thread 'main' panicked at 'out of range type conversion attempt (tried to convert `i64` to `u64`): TryFromIntError(())', /home/nickz/wasmer.git/lib/api/src/externals/function.rs:600:43
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/libunwind.rs:86
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.46/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:78
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:59
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1076
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1537
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:62
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:49
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:198
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:217
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:526
  11: rust_begin_unwind
             at src/libstd/panicking.rs:437
  12: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  13: core::option::expect_none_failed
             at src/libcore/option.rs:1269
  14: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  15: <Func as wasmer::externals::function::inner::HostFunction<(A1,A2,A3,A4,A5,A6,A7,A8,A9),Rets,wasmer::externals::function::inner::WithEnv,Env>>::function_body_ptr::func_wrapper
  16: <unknown>
  17: <unknown>
  18: <unknown>
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

MarkMcCaskey:
resolves #1596 

I attempted to use constant expressions so we didn't have to duplicate the macro, but `std::mem::transmute` is a compile error if you attempt to transmute from A to B where size_of(A) != size_of(B), even if it's in a branch like `if false`

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## **[Unreleased]**
+- [#1602](https://github.com/wasmerio/wasmer/pull/1602) Fix panic when calling host functions with negative numbers in certain situations
 - [#1590](https://github.com/wasmerio/wasmer/pull/1590) Fix soundness issue in API of vm::Global
 - [#1566](https://github.com/wasmerio/wasmer/pull/1566) Add support for opening special Unix files to the WASI FS
 
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -630,16 +630,12 @@ mod inner {
 
                     #[inline]
                     fn from_native(native: Self::Native) -> Self {
-                        unsafe {
-                            std::mem::transmute::<$native_type, $type>(native)
-                        }
+                        Self::from_ne_bytes(Self::Native::to_ne_bytes(native))
                     }
 
                     #[inline]
                     fn to_native(self) -> Self::Native {
-                        unsafe {
-                            std::mem::transmute::<$type, $native_type>(self)
-                        }
+                        Self::Native::from_ne_bytes(Self::to_ne_bytes(self))
                     }
                 }
             )*
@@ -669,16 +665,7 @@ mod inner {
         #[test]
         fn test_to_native() {
             assert_eq!(7i8.to_native(), 7i32);
-        }
-
-        #[test]
-        #[should_panic(
-            expected = "out of range type conversion attempt (tried to convert `u32` to `i32`)"
-        )]
-        fn test_to_native_panics() {
-            use std::{i32, u32};
-
-            assert_eq!(u32::MAX.to_native(), i32::MAX);
+            assert_eq!(u32::MAX.to_native(), -1);
         }
     }
 
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -597,24 +597,12 @@ mod inner {
 
                     #[inline]
                     fn from_native(native: Self::Native) -> Self {
-                        native.try_into().expect(concat!(
-                            "out of range type conversion attempt (tried to convert `",
-                            stringify!($native_type),
-                            "` to `",
-                            stringify!($type),
-                            "`)",
-                        ))
+                        native as Self
                     }
 
                     #[inline]
                     fn to_native(self) -> Self::Native {
-                        self.try_into().expect(concat!(
-                            "out of range type conversion attempt (tried to convert `",
-                            stringify!($type),
-                            "` to `",
-                            stringify!($native_type),
-                            "`)",
-                        ))
+                        self as Self::Native
                     }
                 }
             )*
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -621,11 +621,39 @@ mod inner {
         };
     }
 
+    macro_rules! from_to_native_wasm_type_same_size {
+        ( $( $type:ty => $native_type:ty ),* ) => {
+            $(
+                #[allow(clippy::use_self)]
+                unsafe impl FromToNativeWasmType for $type {
+                    type Native = $native_type;
+
+                    #[inline]
+                    fn from_native(native: Self::Native) -> Self {
+                        unsafe {
+                            std::mem::transmute::<$native_type, $type>(native)
+                        }
+                    }
+
+                    #[inline]
+                    fn to_native(self) -> Self::Native {
+                        unsafe {
+                            std::mem::transmute::<$type, $native_type>(self)
+                        }
+                    }
+                }
+            )*
+        };
+    }
+
     from_to_native_wasm_type!(
         i8 => i32,
         u8 => i32,
         i16 => i32,
-        u16 => i32,
+        u16 => i32
+    );
+
+    from_to_native_wasm_type_same_size!(
         i32 => i32,
         u32 => i32,
         i64 => i64,
--- a/lib/api/tests/module.rs
+++ b/lib/api/tests/module.rs
@@ -156,3 +156,57 @@ fn exports() -> Result<()> {
     );
     Ok(())
 }
+
+#[test]
+fn calling_host_functions_with_negative_values_works() -> Result<()> {
+    let store = Store::default();
+    let wat = r#"(module
+    (import "host" "host_func1" (func (param i64)))
+    (import "host" "host_func2" (func (param i32)))
+    (import "host" "host_func3" (func (param i64)))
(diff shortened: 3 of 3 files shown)


lib/api/src/externals/function.rs
```diff
@@ -621,11 +621,39 @@ mod inner {
...
+
     from_to_native_wasm_type!(
         i8 => i32,
         u8 => i32,
         i16 => i32,
-        u16 => i32,
+        u16 => i32
+    );
+
+    from_to_native_wasm_type_same_size!(
```
nlewycky, line 656: I really don't understand the naming here, `from_to_X_Y`? I read it expecting "from X to Y"? Maybe "define casts" instead?

lib/api/src/externals/function.rs
```diff
@@ -621,11 +621,39 @@ mod inner {
...
+        ( $( $type:ty => $native_type:ty ),* ) => {
+            $(
+                #[allow(clippy::use_self)]
+                unsafe impl FromToNativeWasmType for $type {
+                    type Native = $native_type;
+
+                    #[inline]
+                    fn from_native(native: Self::Native) -> Self {
+                        unsafe {
+                            std::mem::transmute::<$native_type, $type>(native)
```
nlewycky, line 634: Does this have to be an unsafe transmute? What does it mean if the endianness of the system is different, do we want to produce a different result? Could this be written with a from_ne_bytes / to_ne_bytes pair?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## **[Unreleased]**
+- [#1602](https://github.com/wasmerio/wasmer/pull/1602) Fix panic when calling host functions with negative numbers in certain situations
```
nlewycky, line 4: Could we get a test for that?

lib/api/src/externals/function.rs
MarkMcCaskey: I thought this wasn't possible, but it seems it is possible to do that since Rust 1.40.0 ; yeah we can do this without unsafe, I don't know if there's any overhead for doing so... I can check

MarkMcCaskey: Interestingly, it looks like the `X::from_ne_bytes(Y::to_ne_bytes(value))` is the _most_ efficient way to express this, it gets compiled out entirely in 1.45.2 and 1.46.0, whereas the transmute doesn't in the test I just made

CHANGELOG.md
MarkMcCaskey: Sure, any recommendations on where a test this basic should go? 😆

lib/api/src/externals/function.rs
nlewycky: I think it'd be reasonable to file a bug on rustc that transmute doesn't get compiled out in this case. Regardless we should prefer code that is safe and expressive, and if that just happens to get us better code generation, all the better!

lib/api/src/externals/function.rs
MarkMcCaskey: 🤷 That's what the thing is called right now, it's a trait for bidirectional conversion

nlewycky: OK. I still dislike the naming but it doesn't need to rolled into this PR.

lib/api/src/externals/function.rs
MarkMcCaskey:
```Rust
use std::convert::TryInto;

#[inline(never)]
pub fn convert_try(i: u32) -> i32 {
    i.try_into().unwrap()
}

#[inline(never)]
pub fn convert_trans(i: u32) -> i32 {
    unsafe { std::mem::transmute(i) }
}

#[inline(never)]
pub fn convert_to_from(i: u32) -> i32 {
   i32::from_ne_bytes(u32::to_ne_bytes(i))
}
```

is the thing I tested, it's likely that the others would get compiled out if they're actually used, but only the bottom one was completely missing from the generated code.

with this being the generated code

```asm
example::convert_try:
        push    rax
        test    edi, edi
        js      .LBB1_1
        mov     eax, edi
        pop     rcx
        ret
.LBB1_1:
        lea     rdi, [rip + .L__unnamed_1]
        lea     rcx, [rip + .L__unnamed_2]
        lea     r8, [rip + .L__unnamed_3]
        mov     rdx, rsp
        mov     esi, 43
        call    qword ptr [rip + core::result::unwrap_failed@GOTPCREL]
        ud2

example::convert_trans:
        mov     eax, edi
        ret
```


Normally, I'd agree in regards to safe/expressive but this is potentially very, very hot code in Wasmer as it happens on every parameter every time the host calls into the guest, so I think it's worth us going out of our way to make sure things here are fast.

lib/api/src/externals/function.rs
MarkMcCaskey: Perhaps file an issue for it, I remember Ivan talking about it at the time. I thought the name was a bit strange too, but it seemed like a good option given the other names he considered

CHANGELOG.md
nlewycky: Hrm, I don't really see a good place for it. My first choice is to create a new place, `tests/api` which tests lib/api or `tests/regression` which contains tiny tests to ensure that we don't have regressions on fixed PRs. I somewhat prefer the latter because I think we'll want a grab-bag directory for just "PRs we've fixed in the past" and tests can have names like `pr1602.rs`.

MarkMcCaskey: Oh didn't see this comment until now... hmm that's a good idea, I put in the api/tests for now because it is testing the API.

It would be a good idea to have a place to track all the tests we added this way as well as having more explicit regression tests

MarkMcCaskey: Oh, so the `tests/api` are the unit tests on the api crate so it's at `lib/api/tests`

syrusakbary (approved):

MarkMcCaskey:
bors try

PR #1603, opened 2020-09-08
Update codecov.yaml

MarkMcCaskey:
Adds a more lenient threshold for the patch diff (perhaps it should be even higher than 5%) and moves the codecov.yml file into the `.github` folder, which the codecov docs say is also okay!

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -4,3 +4,8 @@ coverage:
       default:
         target: auto
         threshold: 1%
+    patch:
+      default:
+        target: auto
+        threshold: 5%
+        base: auto


nlewycky:
What does a 5% threshold mean?

MarkMcCaskey:
> What does a 5% threshold mean?

It means that if the patch diff was at 30% and you drop it to 25% then it's still okay.

I'm in favor of large (possibly larger) thresholds for patch diffs given our current amount of test coverage because cargo tarpaulin with our current settings has a ton of false positives / negatives and sometimes editing unrelated code will adversely affect you when it shouldn't I think

nlewycky (approved):

PR #1607, opened 2020-09-10
Feature/llvm pic mcsmall

nlewycky:
# Description

# Review


lib/engine-jit/src/link.rs
```diff
@@ -37,22 +37,18 @@ fn apply_relocation(
     };
 
     match r.kind {
-        #[cfg(target_pointer_width = "64")]
```
syrusakbary, line 40: This was intentional. JIT can only support the target pointer widths for the target architecture.

lib/engine-jit/src/engine.rs
```diff
@@ -157,7 +157,7 @@ pub struct JITEngineInner {
     features: Features,
     /// The code memory is responsible of publishing the compiled
     /// functions to memory.
-    code_memory: CodeMemory,
+    code_memory: Vec<CodeMemory>,
```
syrusakbary, line 160: I'm trying to understand the needs/implications of this. I guess now each artifact will have its own `CodeMemory`?
Is that assumption accurate?

lib/engine-jit/src/link.rs
nlewycky: The relocation specifies how many bits it modifies, regardless of the pointer width on the machine. For example AArch64 has LD_PREL_LO19 which edits exactly 19 bits of the value.

lib/engine-jit/src/engine.rs
nlewycky: Yes, and each `CodeMemory` has one `Mmap`. To get small code model to work, everything a compiled module needs is within a single 32-bit region.

lib/engine-jit/src/link.rs
syrusakbary: Yeah, I understand that.

What I mean is, before  if you were trying to run code compiled for a 64 bit machine, in a 32 bit machine, it would have fall back to a `panic!` -> https://github.com/wasmerio/wasmer/blob/742d73d0c3136393a0f578b2d8bca9f92ce184b8/lib/engine-jit/src/link.rs#L57-L60 

By removing the `cfg` macro, it will silently fail until you actually try to run it. Which I think is a worse developer experience.

PR #1619, opened 2020-09-14
feat(native-engine) Panic if `gcc`/`clang-10` is missing

Hywan:
When `gcc` or `clang-10` is absent, we get a non-obvious error:

> Compilation error: No such file or directory (os error 2)

It comes from https://github.com/wasmerio/wasmer/blob/2cd12213fb256f475c4e1dc87aefd422115665f7/lib/engine-native/src/artifact.rs#L288-L298.

This PR improves that by panicking when `gcc` or `clang` isn't present.


lib/engine-native/src/engine.rs
```diff
@@ -27,6 +29,10 @@ impl NativeEngine {
     /// Create a new `NativeEngine` with the given config
     #[cfg(feature = "compiler")]
     pub fn new(compiler: Box<dyn Compiler + Send>, target: Target, features: Features) -> Self {
+        if let Err(_) = which("gcc").or_else(|_| which("clang-10")) {
```
syrusakbary, line 32: We should try (by this order):
* `clang-10` (cross compile)
* `clang` (cross compiled)
* `gcc`

lib/engine-native/src/engine.rs
```diff
@@ -27,6 +29,10 @@ impl NativeEngine {
...
+            panic!("gcc/clang is not found, it is required for using the `NativeEngine`");
+        }
```
syrusakbary, line 34: Once we know which compiler to use, we should save it in the struct, along with if supports cross_compilation or not.

To then, use that command in the artifact generation.

syrusakbary (approved):

PR #1621, opened 2020-09-14
Prepare for 1.0.0-alpha02.1 release

MarkMcCaskey:
Now we're publishing wapm

# Review


nlewycky (commented):
I'm not sure we want to make an alpha-point release that includes an additional change (https://github.com/wasmerio/wasmer/pull/1619) vs. moving to alpha03, but otherwise looks good.

PR #1625, opened 2020-09-15
Synchronize mutable env in functions

### LINKED ISSUE #1612: Calling a host function with environment in parallel results in data race
### Describe the bug

Causing a host function to be called in parallel leads to aliased mutable references to the function's `Env`. This is a data race and is undefined behavior.

Here's a complete example:

```rust
use wasmer::{Cranelift, Function, Instance, Module, Store, Value, JIT};

fn main() {
    let mut compiler_config = Cranelift::new();
    let engine = JIT::new(&mut compiler_config).engine();
    let store = Store::new(&engine);

    let wasm_source = "
    (module
        (type (;0;) (func (param i32) (result i32)))
        (import \"env\" \"host_function\" (func $host_function (type 0)))
        (func $guest_function (type 0) (param i32) (result i32)
          local.get 0
          call $host_function)
        (table (;0;) 1 1 funcref)
        (memory (;0;) 16)
        (global (;0;) (mut i32) (i32.const 1048576))
        (global (;1;) i32 (i32.const 1048576))
        (global (;2;) i32 (i32.const 1048576))
        (export \"memory\" (memory 0))
        (export \"guest_function\" (func $guest_function))
        (export \"__data_end\" (global 1))
        (export \"__heap_base\" (global 2)))
    ";
    let module = Module::new(&store, wasm_source).unwrap();
    let import_object = wasmer::imports! {
        "env" => {
            // host_function has a `u32` environment
            "host_function" => Function::new_native_with_env(&store, 10u32, host_function),
        }
    };

    // spawn 8 threads to call guest_function in parallel
    let mut threads = Vec::new();
    for _ in 0..8 {
            for _ in 0..1_000_000 {
                let guest_function = instance.exports.get_function("guest_function").unwrap();
                let _result = guest_function.call(&[Value::I32(1000)]).unwrap();
            }
        }));
    }

    for thread in threads {
        thread.join().unwrap();
    }
}

fn host_function(env: &mut u32, x: u32) -> u32 {
    *env += 1;
    if *env % 1_000_000 == 0 {
        println!("{}", env);
    }
    x
}

```

Running with `wasmer-1.0.0-alpha02.0` and `cargo run --release` gives me this output on Linux:

```
1000079
1000131
1000262
1001312
2000104
2000220
3000017
3000307
````

There's a pretty clear data race here—the correct output would be 

```
1000000
2000000
3000000
4000000
5000000
6000000
7000000
8000000
```

### Steps to reproduce
Run the above example with `cargo run --release`

### Expected behavior
This looks like the API for host function environments needs to be changed so the function doesn't receive a mutable reference to its `Env`.

### PR DESCRIPTION
**Pull request by MarkMcCaskey**
Resolves #1612

This PR wraps all `Env`s in `std::sync::Mutex` ensuring that mutable access cannot happen concurrently.  This PR also fixes some undefined behavior / broken code that was accidentally working with `VMDynamicFunctionWithEnv`.

As a next step, we should design new APIs to give users more control over `Env`, by for example having an `env` and `env_mut` variation where the `env` variation passes a `&Env` allowing users to synchronize themselves if they want.  Additionally, it may be valuable to add unsafe versions of these functions for power users of the API.  Finally, we should investigate if it's possible to allow direct `&mut` access in a single threaded context with a sound API (by using trait constraints).

# Review

### DIFF
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -6,9 +6,10 @@ use crate::FunctionType;
 use crate::NativeFunc;
 use crate::RuntimeError;
 pub use inner::{FromToNativeWasmType, HostFunction, WasmTypeList, WithEnv, WithoutEnv};
-use std::cell::RefCell;
+use std::borrow::BorrowMut;
 use std::cmp::max;
 use std::fmt;
+use std::sync::Mutex;
 use wasmer_vm::{
     raise_user_trap, resume_panic, wasmer_call_trampoline, Export, ExportFunction,
     VMCallerCheckedAnyfunc, VMContext, VMDynamicFunctionContext, VMFunctionBody, VMFunctionKind,
@@ -126,7 +127,7 @@ impl Function {
         Env: Sized + 'static,
     {
         let dynamic_ctx = VMDynamicFunctionContext::from_context(VMDynamicFunctionWithEnv {
-            env: RefCell::new(env),
+            env: Mutex::new(Box::new(env)),
             func: Box::new(func),
             function_type: ty.clone(),
         });
@@ -226,7 +227,7 @@ impl Function {
         // Wasm-defined functions have a `VMContext`.
         // In the case of Host-defined functions `VMContext` is whatever environment
         // the user want to attach to the function.
-        let box_env = Box::new(env);
+        let box_env = Box::new(Mutex::new(env));
         let vmctx = Box::into_raw(box_env) as *mut _ as *mut VMContext;
         let signature = function.ty();
 
@@ -409,7 +410,6 @@ impl Function {
                 )));
             }
         }
-
         Ok(NativeFunc::new(
             self.store.clone(),
             self.exported.address,
@@ -463,6 +463,7 @@ impl VMDynamicFunction for VMDynamicFunctionWithoutEnv {
     }
 }
 
+#[repr(C)]
 pub(crate) struct VMDynamicFunctionWithEnv<Env>
 where
     Env: Sized + 'static,
@@ -470,7 +471,9 @@ where
     function_type: FunctionType,
     #[allow(clippy::type_complexity)]
     func: Box<dyn Fn(&mut Env, &[Val]) -> Result<Vec<Val>, RuntimeError> + 'static>,
-    env: RefCell<Env>,
+    // We have to box this because of the way we call the `call` function, we don't know the
+    // size of the env when calling from NativeFunc.
+    env: Mutex<Box<Env>>,
 }
 
 impl<Env> VMDynamicFunction for VMDynamicFunctionWithEnv<Env>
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by MarkMcCaskey** [comment]
bors try

File: lib/api/src/externals/function.rs [hunk 1]
```diff
@@ -463,24 +463,27 @@ impl VMDynamicFunction for VMDynamicFunctionWithoutEnv {
...
 
 impl<Env> VMDynamicFunction for VMDynamicFunctionWithEnv<Env>
 where
     Env: Sized + 'static,
 {
     fn call(&self, args: &[Val]) -> Result<Vec<Val>, RuntimeError> {
-        // TODO: the `&mut *self.env.as_ptr()` is likely invoking some "mild"
-        //      undefined behavior due to how it's used in the static fn call
-        unsafe { (*self.func)(&mut *self.env.as_ptr(), &args) }
+        let mut env_guard = self.env.lock().unwrap();
```
syrusakbary, line 484: I'm not really a fan of this. This is a clear performance penalty that we are now incurring for all function calls.

lib/api/src/externals/function.rs
```diff
@@ -1022,18 +1025,22 @@ mod inner {
...
                         $( $x: FromToNativeWasmType, )*
                         Rets: WasmTypeList,
                         RetsAsResult: IntoResult<Rets>,
                         Env: Sized,
                         Func: Fn(&mut Env, $( $x ),* ) -> RetsAsResult + 'static
                     {
+                        use std::borrow::BorrowMut;
+
                         let func: &Func = unsafe { &*(&() as *const () as *const Func) };
+                        let mut env_guard = env.lock().unwrap();
```
syrusakbary, line 1039: I'm not really a fan of this. This is a clear performance penalty that we are now incurring for all function calls.

Hywan: We should measure this penalty, because I don't see a lot more options to solve this sync problem.

MarkMcCaskey: I'm completely in favor of measurement, but I'm also sympathetic to wanting a fast path for singlethreaded code which will likely be the majority of uses (at least in the short-term).

Also, I agree that yeah, this change is nice because it just works without much changes.  I wrote up a design doc about solutions to this problem and I'll ping people about it next week: I think it's probable that we can make a more complex solution to this problem, but it will require breaking changes to our API or duplicating our API (singlethreaded version vs general version).  I also talked about some more exotic options with Nick in regards to things like updating vtables at runtime to dynamically synchronize code when it becomes multithreaded but I don't think we're realistically going to be trying anything fancy any time soon and also I haven't thought of any fancy solutions that aren't a bad trade off in terms of complexity.

MarkMcCaskey:
This PR is not being actively developed, we've decided to pursue #1663 instead

PR #1628, opened 2020-09-16
Change engine-jit to keep all data in an artifact within one mmap.

nlewycky:
# Description
This change keeps the parts of one artifact together, which lessens the restrictions on code emitted by the compilers. For example, they can use 32-bit addressing for their own functions and data on a 64-bit system, though access to the wasmer VM and through the VMContext still has to use the full host width.

In passing, implements a few cleanups. A new SectionBodyPtr type is introduced for type safety. Each UnwindRegistry is now singularly owned by a CodeMemory.

# Review


lib/engine-jit/src/code_memory.rs
```diff
@@ -17,193 +13,144 @@ use wasmer_vm::{FunctionBodyPtr, Mmap, VMFunctionBody};
...
+            let (func_buf, next_buf) = buf.split_at_mut(len);
+            buf = next_buf;
+            bytes += len;
 
-    /// Make all allocated memory executable.
-    pub fn publish(&mut self) {
-        self.push_current(0)
-            .expect("failed to push current memory map");
+            let vmfunc = Self::copy_function(&mut self.unwind_registry, func, func_buf);
+            assert!(vmfunc as *mut _ as *mut u8 as usize % ARCH_FUNCTION_ALIGNMENT == 0);
```
MarkMcCaskey, line 98: You can probably just cast this directly to a usize once it's a pointer, I don't think it matters that it be a `*mut u8` first

lib/engine-jit/src/code_memory.rs
```diff
@@ -17,193 +13,144 @@ use wasmer_vm::{FunctionBodyPtr, Mmap, VMFunctionBody};
...
+            function_result.push(vmfunc);
+        }
+        for section in executable_sections {
+            let section = &section.bytes;
+            assert!(buf.as_mut_ptr() as *mut _ as *mut u8 as usize % ARCH_FUNCTION_ALIGNMENT == 0);
```
MarkMcCaskey, line 103: same here

lib/engine-jit/src/code_memory.rs
MarkMcCaskey: Also consider using `assert_eq!` which will print out what the value is if it's not equal, whereas normal assert will only say that the given boolean expression failed

lib/engine-jit/src/code_memory.rs
```diff
@@ -17,193 +13,144 @@ use wasmer_vm::{FunctionBodyPtr, Mmap, VMFunctionBody};
...
+        if !data_sections.is_empty() {
+            // Data sections have different page permissions from the executable
+            // code that came before it, so they need to be on different pages.
+            let padding = round_up(bytes, page_size) - bytes;
+            buf = buf.split_at_mut(padding).1;
+
+            for section in data_sections {
+                let section = &section.bytes;
+                assert!(
+                    buf.as_mut_ptr() as *mut _ as *mut u8 as usize % DATA_SECTION_ALIGNMENT == 0
```
MarkMcCaskey, line 123: here too

lib/engine-jit/src/code_memory.rs
```diff
@@ -224,41 +171,35 @@ impl CodeMemory {
...
     fn copy_function<'a>(
         registry: &mut UnwindRegistry,
-        base_address: usize,
         func: &FunctionBody,
-        func_start: u32,
         buf: &'a mut [u8],
-    ) -> (u32, &'a mut [u8], &'a mut [VMFunctionBody]) {
-        assert!((func_start as usize) % ARCH_FUNCTION_ALIGNMENT == 0);
+    ) -> &'a mut [VMFunctionBody] {
+        assert!((buf.as_ptr() as usize) % ARCH_FUNCTION_ALIGNMENT == 0);
```
MarkMcCaskey, line 177: ⛳  `assert_eq!`

lib/engine-jit/src/engine.rs
```diff
@@ -193,130 +192,117 @@ impl JITEngineInner {
...
-            //     // it's signature is already allocated.
-            //     continue;
-            // }
-            let ptr = self
-                .code_memory
-                .allocate_for_function(registry, &compiled_function)
+        let (mut allocated_functions, allocated_executable_sections, allocated_data_sections) =
+            self.code_memory
+                .last_mut()
+                .unwrap()
```
MarkMcCaskey, line 226: Why is this correct?

lib/engine-jit/src/engine.rs
```diff
@@ -193,130 +192,117 @@ impl JITEngineInner {
...
+
+        let allocated_functions_result = allocated_functions
+            .drain(0..functions.len())
+            .map(|slice| FunctionBodyPtr(slice as *mut [_]))
+            .collect::<PrimaryMap<LocalFunctionIndex, _>>();
+
+        let mut allocated_function_call_trampolines: PrimaryMap<SignatureIndex, FunctionBodyPtr> =
+            PrimaryMap::new();
+        for (sig_index, _) in function_call_trampolines.iter() {
+            let func_type = module.signatures.get(sig_index).unwrap();
```
MarkMcCaskey, line 247: probably the same as `module.signatures[sig_index]` which is generally equivalent to `get(idx).unwrap()`

lib/engine-jit/src/engine.rs
```diff
@@ -193,130 +192,117 @@ impl JITEngineInner {
...
+            let index = self.signatures.register(&func_type);
+            let ptr = allocated_functions
+                .drain(0..1)
+                .map(|slice| FunctionBodyPtr(slice as *mut [_]))
+                .collect::<Vec<_>>()[0];
```
MarkMcCaskey, line 252: ⛳ consider using `.next()` instead of allocating on the heap

MarkMcCaskey: Also depending on how we access `allocated_functions`, if this is an operation that happens a lot, we may want to use a data structure which has efficient popping off the front like a `std::collections::VecDeque`

lib/engine-jit/src/engine.rs
```diff
@@ -193,130 +192,117 @@ impl JITEngineInner {
...
     }
 
-    /// Publish the unwind registry into code memory.
-    pub(crate) fn publish_unwind_registry(&mut self, unwind_registry: Arc<UnwindRegistry>) {
-        self.code_memory.publish_unwind_registry(unwind_registry);
+    /// Register DWARF-type exception handling information associated with the code.
+    pub(crate) fn publish_eh_frame(&mut self, eh_frame: Option<&[u8]>) -> Result<(), CompileError> {
+        self.code_memory
+            .last_mut()
+            .unwrap()
```
MarkMcCaskey, line 299: Is this guaranteed?

MarkMcCaskey (commented):
Looks pretty good to me as far as I could see from my initial review.  I didn't fully think through some of the code for correctness though

lib/engine-jit/src/engine.rs
nlewycky: Why is the unwrap correct? The immediately previous line of code is:
```
self.code_memory.push(CodeMemory::new());
```

lib/engine-jit/src/engine.rs
nlewycky: Roughly these types have a) init empty b) fill with data c) publish data. If you skip fill then the publish will fail and that's fine to panic! because it's definitely an internal error inside engine-jit. The only thing I'd change here is to add an assert! that self.code_memory isn't empty, and that isn't really a change to the behaviour.

The only other issue is why the last one, as in, can you do init+fill+init+fill+publish+publish? In a word, no, we always do these actions together while setting up a new artifact. Artifact::from_parts is what drives this.

lib/engine-jit/src/engine.rs
nlewycky: Used zip + 1 call to drain(0..function_call_trampolines.len()).

lib/engine-jit/src/engine.rs
MarkMcCaskey: Yeah that's fine!  Only point I'd mention is that it'd be good to have the preconditions listed in the doc comments but that sounds reasonable!

MarkMcCaskey (approved):
Looks good to me

nlewycky:
bors try

PR #1629, opened 2020-09-17
fix(c api) Remove the `trampoline` module since it's not used

Hywan:
Need to be double-checked by @MarkMcCaskey. I'm not sure we are going to port this module since we have an entire new `Function` API that covers the features provided by the old trampoline API.


MarkMcCaskey (approved):
I think this is a good change -- sorry about the delay I didn't notice this until now!

The only reason I didn't delete it is so we can think about it more and make sure it's really what we want

Issue #1632, opened 2020-09-17
InstanceHandler::memory_grow performs unsynchronized write to shared memory


bjorn3:
https://github.com/wasmerio/wasmer/blob/a2e744aba6c4898841f8a67f4d56a24f204ee439/lib/vm/src/instance.rs#L436

This happens through `set_memory`

https://github.com/wasmerio/wasmer/blob/a2e744aba6c4898841f8a67f4d56a24f204ee439/lib/vm/src/instance.rs#L224

Because `InstanceHandler` is `Send` and `Clone` shares the same underlying `Instance` this can cause two threads to write to the same place at the same time in a non-atomic way. This is a data-race. Because `VMMemoryDefenition` is bigger than a pointer this will cause a mangled value being stored in some cases even if LLVM doesn't take advantage of the UB.

syrusakbary:
We believe this will be solved by our refactor in Wasmer 3.0

Issue #1630, opened 2020-09-17
Instance host_state not required to be Sync, yet can be accessed from multiple threads at the same time.


bjorn3:
https://github.com/wasmerio/wasmer/blob/a2e744aba6c4898841f8a67f4d56a24f204ee439/lib/vm/src/instance.rs#L98

1. Create new `InstanceHandle`
2. Clone `InstanceHandle`
3. Send clone to another thread
4. Call `handle.host_state()` on both handles
5. You now have a reference to a non-sync type in two threads. This is UB and may for example be exploited by using `RefCell` as `host_state` to get a race-condition.

syrusakbary:
Thanks for opening the issue @bjorn3!

Here's the related bug reported in wasmtime: https://github.com/bytecodealliance/wasmtime/issues/793 for future readers.

## Possible solution

Since we are not longer using `host_state` perhaps we can get rid of it directly

syrusakbary:
Thanks for creating the issue. We are working on refactoring our API for Wasmer 3.0

syrusakbary:
This issue will be fixed in wasmer3 (already fixed in the `wasmer3` branch)

PR #1633, opened 2020-09-18
Make InstanceHandle not clonable to avoid thread data races

syrusakbary:
This PR is trying to fix the underlying issues that caused #1632 and #1630 by making InstanceHandle not clonable.

The PR needs extra review.

### DIFF
--- a/lib/api/src/instance.rs
+++ b/lib/api/src/instance.rs
@@ -4,6 +4,7 @@ use crate::module::Module;
 use crate::store::Store;
 use crate::InstantiationError;
 use std::fmt;
+use std::sync::Arc;
 use wasmer_engine::Resolver;
 use wasmer_vm::InstanceHandle;
 
@@ -17,7 +18,7 @@ use wasmer_vm::InstanceHandle;
 /// Spec: https://webassembly.github.io/spec/core/exec/runtime.html#module-instances
 #[derive(Clone)]
 pub struct Instance {
-    handle: InstanceHandle,
+    handle: Arc<InstanceHandle>,
     module: Module,
     /// The exports for an instance.
     pub exports: Exports,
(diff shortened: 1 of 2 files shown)


lib/api/src/instance.rs
```diff
@@ -17,7 +18,7 @@ use wasmer_vm::InstanceHandle;
 /// Spec: https://webassembly.github.io/spec/core/exec/runtime.html#module-instances
 #[derive(Clone)]
 pub struct Instance {
-    handle: InstanceHandle,
+    handle: Arc<InstanceHandle>,
```
bjorn3, line 21: This fixes the unsafety because `InstanceHandle` is not `Sync` and thus `Arc<InstanceHandle>` is not `Send`. This does also mean that `Instance` is now also not `Send`.

lib/vm/src/instance.rs
```diff
@@ -767,7 +767,14 @@ pub struct InstanceHandle {
 }
 
 /// # Safety
-/// This is safe because there is no thread-specific logic in `InstanceHandle`.
+/// 
+/// This is safe because there is no thread-specific logic in `InstanceHandle`,
+/// and because `InstanceHandle` is not clonable.
```
bjorn3, line 772:
```suggestion
/// and because `InstanceHandle` is neither clonable nor sync.
```

lib/vm/src/instance.rs
```diff
@@ -767,7 +767,14 @@ pub struct InstanceHandle {
...
+/// 
+/// If `InstanceHandle` could be cloned, then other issues appear as reported by:
```
bjorn3, line 774:
```suggestion
/// If `InstanceHandle` could be cloned and/or would be sync, then other issues appear as reported by:
```

syrusakbary:
This should be now solved in master. Closing the PR

PR #1635, opened 2020-09-18
feat(c-api) Add `wat2wasm` to the `wasm-c-api` module

Hywan:
This PR adds the `wat2wasm` function to the `wasm-c-api` module.

Caveat of this patch: The `wasmer_wasm.h` seems to be written by hand. Consequently, if the `lib/c-api/` crate is compiled without the `wat` feature (which is the default), the `wat2wasm` function will still be defined in `wasmer_wasm.h`.
Proposal: We must generate the `wasmer_wasm.h` header file with the same approach than we use for `wasm.h`. It can be done in a next PR.

Thoughts @MarkMcCaskey

### DIFF
--- a/lib/c-api/src/wasm_c_api/wat.rs
+++ b/lib/c-api/src/wasm_c_api/wat.rs
@@ -21,10 +21,8 @@ pub unsafe extern "C" fn wat2wasm(wat: &wasm_byte_vec_t) -> Option<Box<wasm_byte
     let mut result: Vec<u8> = result.into_owned();
     result.shrink_to_fit();
 
-    let wasm = wasm_byte_vec_t {
+    Some(Box::new(wasm_byte_vec_t {
         size: result.len(),
         data: result.as_mut_ptr(),
-    };
-
-    Some(Box::new(wasm))
+    }))
 }
--- a/lib/c-api/src/wasm_c_api/wat.rs
+++ b/lib/c-api/src/wasm_c_api/wat.rs
@@ -1,0 +1,30 @@
+use super::types::wasm_byte_vec_t;
+
+/// Parses in-memory bytes as either the WAT format, or a binary Wasm
+/// module. This is wasmer-specific.
+///
+/// In case of failure, `wat2wasm` returns `NULL`.
+#[cfg(feature = "wat")]
+#[no_mangle]
+pub unsafe extern "C" fn wat2wasm(wat: &wasm_byte_vec_t) -> Option<Box<wasm_byte_vec_t>> {
+    let wat: &[u8] = wat.into_slice()?;
+
+    let result = match wasmer::wat2wasm(wat) {
+        Ok(result) => result,
+        Err(error) => {
+            crate::error::update_last_error(error);
+
+            return None;
+        }
+    };
+
+    let mut result: Vec<u8> = result.into_owned();
+    result.shrink_to_fit();
+
+    let wasm = wasm_byte_vec_t {
+        size: result.len(),
+        data: result.as_mut_ptr(),
+    };
+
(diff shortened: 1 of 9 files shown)


MarkMcCaskey (commented):
We discussed organization of the C APIs privately, and I think with some of the changes we talked about we can probably generate that file, but for now it's non-trivial to do so.

I'm unsure if we want to make this transformation explicit or not.  In Wasmer we allow taking a `&[u8]` and we'll handle both wat and wasm.  Perhaps we just want to do that for the C API too.

If we want to make it explicit, then this change seems fine.  How would users free this data though?  And one more point, when turning things like `Vec` or `String` into a pointer + length, it's important to `shrink_to_fit` first so we don't leak memory.  `Vec` has 3 numbers: pointer, length, and capacity.  If capacity != length (which is very likely), then we can't reconstruct the `Vec` correctly and will probably do a partial free

Hywan:
I've rebased this PR on top of the `master` branch.
I've also added `shrint_to_fit`.
Finally, I've changed the `wat2wasm` function type signature:

```diff
- unsafe extern "C" fn wat2wasm(wat: &wasm_byte_vec_t, wasm: &mut wasm_byte_vec_t) -> Option<()>
+ unsafe extern "C" fn wat2wasm(wat: &wasm_byte_vec_t) -> Option<Box<wasm_byte_vec_t>>
```

I believe this API is better. If it returns `NULL`, it's clear there is an error; Also it helps to use `?` in the implementation. It also makes explicit that the callee is responsible of freeing the data.

Thoughts?

syrusakbary (approved):

lib/c-api/src/wasm_c_api/wat.rs
```diff
@@ -0,0 +1,28 @@
...
+    let wat: &[u8] = wat.into_slice()?;
+
+    let result = match wasmer::wat2wasm(wat) {
+        Ok(result) => result,
+        Err(error) => {
+            crate::error::update_last_error(error);
+
+            return None;
+        }
+    };
```
MarkMcCaskey, line 19: Is this different than what `c_try!` does?

lib/c-api/src/wasm_c_api/wat.rs
```diff
@@ -0,0 +1,28 @@
...
+
+    let mut result: Vec<u8> = result.into_owned();
+    result.shrink_to_fit();
+
+    Some(Box::new(wasm_byte_vec_t {
+        size: result.len(),
+        data: result.as_mut_ptr(),
+    }))
```
MarkMcCaskey, line 27: We recently landed a change that should let you just do `let result: wasm_byte_vec_t = result.into_owned().into();`!

It encapsulates all the logic and ensures it works. The current code, for example uses the memory after free! `Vec<u8>` will be dropped at the end of this scope and the returned pointer will be invalid. The conversion methods will take care of all these details for us.

MarkMcCaskey (approved):
Looks good!  Other than some small bugs it looks good to go!  One other thing, perhaps we want to add an example/test for this function (it may have caught the use after free bug, though it's highly likely it would not have)

lib/c-api/src/wasm_c_api/wat.rs
Hywan: It was, but now that the return type is an `Option`, it's not!

lib/c-api/src/wasm_c_api/wat.rs
Hywan: Ah true, I forgot about that. Thanks!

MarkMcCaskey (approved):
Looks great!

PR #1636, opened 2020-09-18
feat(c-api) Implement `wasm_module_validate`

Hywan:
This PR implements `wasm_module_validate`.

It returns a bool, but it populates the error handler too. I took the liberty to update the `c_try!` macro to return any kind of value, so that it is generic over the returned type of the function using `c_try!`.

### DIFF
--- a/lib/c-api/src/wasm_c_api/module.rs
+++ b/lib/c-api/src/wasm_c_api/module.rs
@@ -3,6 +3,7 @@ use super::types::{
     wasm_byte_vec_t, wasm_exporttype_t, wasm_exporttype_vec_t, wasm_importtype_t,
     wasm_importtype_vec_t,
 };
+use crate::error::update_last_error;
 use std::ptr::NonNull;
 use std::slice;
 use std::sync::Arc;
@@ -31,6 +32,30 @@ pub unsafe extern "C" fn wasm_module_new(
 
 #[no_mangle]
 pub unsafe extern "C" fn wasm_module_delete(_module: Option<Box<wasm_module_t>>) {}
+
+#[no_mangle]
+pub unsafe extern "C" fn wasm_module_validate(
+    store_ptr: Option<NonNull<wasm_store_t>>,
+    bytes: &wasm_byte_vec_t,
+) -> bool {
+    // TODO: review lifetime of byte slice.
+    let wasm_byte_slice: &[u8] = slice::from_raw_parts(bytes.data, bytes.size);
+
+    if store_ptr.is_none() {
+        return false;
+    }
+
+    let store_ptr = store_ptr.unwrap().cast::<Store>();
+    let store = store_ptr.as_ref();
+
+    if let Err(error) = Module::validate(store, wasm_byte_slice) {
+        update_last_error(error);
+
+        false
+    } else {
+        true
+    }
+}
 
 #[no_mangle]
 pub unsafe extern "C" fn wasm_module_exports(
--- a/lib/c-api/src/wasm_c_api/module.rs
+++ b/lib/c-api/src/wasm_c_api/module.rs
@@ -7,7 +7,7 @@ use crate::error::update_last_error;
 use std::ptr::NonNull;
 use std::slice;
 use std::sync::Arc;
-use wasmer::{Module, Store};
+use wasmer::Module;
 
 #[allow(non_camel_case_types)]
 pub struct wasm_module_t {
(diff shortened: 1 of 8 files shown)


lib/c-api/src/wasm_c_api/mod.rs
```diff
@@ -208,6 +208,26 @@ pub unsafe extern "C" fn wasm_module_new(
...
+    if store_ptr.is_none() {
+        return false;
+    }
+
+    let store_ptr = store_ptr.unwrap().cast::<Store>();
+    let store = store_ptr.as_ref();
+
+    c_try!(Module::validate(store, wasm_byte_slice); otherwise return false);
+
+    true
```
MarkMcCaskey, line 228: I think this code should be able to just be

```Rust
     c_try!(Module::validate(store, wasm_byte_slice); otherwise return false)
```

lib/c-api/src/wasm_c_api/mod.rs
```diff
@@ -208,6 +208,26 @@ pub unsafe extern "C" fn wasm_module_new(
     }))
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn wasm_module_validate(
+    store_ptr: Option<NonNull<wasm_store_t>>,
```
MarkMcCaskey, line 213: This is something I've thought about for a while, we probably can just do `&wasm_store_t` if we need immutable access to it.  The thing you did was good though as it's following the lead of a lot of the code in this file.  Feel free to keep it as is and we can update a lot of the uses of store in a separate PR later.

lib/c-api/src/wasm_c_api/mod.rs
```diff
@@ -208,6 +208,26 @@ pub unsafe extern "C" fn wasm_module_new(
...
+    bytes: &wasm_byte_vec_t,
+) -> bool {
+    // TODO: review lifetime of byte slice
+    let wasm_byte_slice: &[u8] = slice::from_raw_parts_mut(bytes.data, bytes.size);
+
+    if store_ptr.is_none() {
+        return false;
+    }
+
+    let store_ptr = store_ptr.unwrap().cast::<Store>();
```
MarkMcCaskey, line 223: You could use `c_try!` here if you wanted

MarkMcCaskey (approved):
Looks good to me!

I was initially a bit uncertain about having anything but the bare minimum syntax in a macro, but I think it's probably fine!

lib/c-api/src/wasm_c_api/mod.rs
Hywan: Nah, because `Module::validate` returns a `Result`.

MarkMcCaskey: I thought that part of `c_try` was for things that returned `Result`

Hywan:
@MarkMcCaskey I've rebased the PR on top of the `master` branch. I took this as an opportunity to not change the `c_try!` macro. The code is simpler.

I don't use `c_try!` at all because the function returns a `bool`, not a `Result`.

Finally, I prefer to use `Option<NonNull<wasm_store_t>>` rather than `*const wasm_store_t`, it provides a better Rust idiomatic API from my point of view.

Thoughts?

MarkMcCaskey:
> Finally, I prefer to use `Option<NonNull<wasm_store_t>>` rather than `*const wasm_store_t`, it provides a better Rust idiomatic API from my point of view.
> 
> Thoughts?

My biggest issue with `NonNull` is that it implies ownership... It's too bad Rust doesn't have an immutable `NonNull` pointer... cbindgen will generate code that implies ownership with NonNull I believe...

I think `&wasm_store_t` is probably fine. As long as the C API doesn't say we need to handle null pointers for the store, then I think it's fine to go into UB world if the C caller messes up.

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -32,6 +33,30 @@ pub unsafe extern "C" fn wasm_module_new(
 #[no_mangle]
 pub unsafe extern "C" fn wasm_module_delete(_module: Option<Box<wasm_module_t>>) {}
 
+#[no_mangle]
+pub unsafe extern "C" fn wasm_module_validate(
+    store_ptr: Option<NonNull<wasm_store_t>>,
```
MarkMcCaskey, line 38: I commented elsewhere on this PR about the implied semantics of `NonNull`, `store: &wasm_store_t` is probably better, especially if we're calling `unwrap` on it

MarkMcCaskey (approved):
Other than that, looks good to me!

Hywan:
I've updated the definition of `wasm_store_t` from being an opaque Rust type to a real Rust struct such as:

```rust
struct wasm_store {
    inner: Store
}
```

The rest of the patch above updates the code accordingly.

Thoughts?

MarkMcCaskey (approved):
Looks great to me!

PR #1639, opened 2020-09-21
feat(c-api) Implement `wasm_importtype_delete`

Hywan:
This PR implements `wasm_importtype_delete`. Thoughts @MarkMcCaskey?

### DIFF
--- a/lib/c-api/src/wasm_c_api/mod.rs
+++ b/lib/c-api/src/wasm_c_api/mod.rs
@@ -1869,6 +1869,9 @@ pub extern "C" fn wasm_importtype_type(
     unsafe { et.extern_type.as_ref() }
 }
 
+#[no_mangle]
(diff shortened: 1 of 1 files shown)


MarkMcCaskey (approved):
Looks good to me!  This function probably does not need to be `unsafe` but it doesn't really mater.

The reason it doesn't need to be `unsafe` is that given the type signature, there's no way to call this function incorrectly.  `unsafe` means that Rust code calling this function needs to abide by certain invariants, but there's no way to misuse this function from safe/sound Rust.  The fact that the C API can call it incorrectly doesn't really matter because `unsafe` only applies to Rust code compiled with this code.

Hywan:
I generally mark functions targeting the C world as unsafe, because I don't trust anything from C, but it's probably a bad practise. Thanks for the comment!

PR #1641, opened 2020-09-22
feat(c-api) Move the WasmER C API as deprecated, and clean up the Wasm C API

Hywan:
This PR moves the WasmER C API inside a `lib/c-api/src/deprecated/` directory. It also moves the documentation inside a `lib/c-api/doc/deprecated/` directory. It updates doxygen accordingly.

This PR also generates the `wasmer_wasm.h` file automatically. It's not correct yet though. 2 PR are pending on cbindgen's side to fix that. Edit: We are now using our own fork of cbindgen (https://github.com/eqrion/cbindgen/compare/master...Hywan:incoming?expand=1, it contains the PR we need), the `wasmer_wasm.h` file is much better now, I need to test it though.

This PR also splits the `wasm_c_api` into separate modules. It's now cleaner.


lib/c-api/README.md
```diff
@@ -1,136 +1,2 @@
...
-
-```bash
-wasmer config --libdir
-```
-
-> Note: You can also download the libraries or header files directly
-from [Wasmer release page].
-
-The full C API documentation can be found here:
-https://wasmerio.github.io/wasmer/c-api/
```
syrusakbary, line 27: We should keep this. It's going to be useful regardless

Hywan: I've removed it because it might outdated. It's a reminder for me to not forget to update that part. The PR is still a work in progress :-).

lib/c-api/build.rs
```diff
@@ -40,67 +25,356 @@ fn main() {
...
+                "feature",
+                "include-deprecated",
+                INCLUDE_DEPRECATED_FEATURE_AS_C_DEFINE,
+            )
+            // Do not include `wasm.h` as it doesn't contain any
+            // documentation.
+            // We export the same functions, with documentations and
+            // examples, which is better for the user.
+            //
+            //.with_include("wasm.h")
```
MarkMcCaskey, line 135: I think we should generally use the standard `wasm.h` in all our tests to make sure we're conforming to the signatures in it.  Alternatively we can have some kind of verification step where we diff our header file against it.  I'm worried that distributing a custom version of wasm.h is a good way to accidentally not support the real wasm C API

lib/c-api/src/deprecated/mod.rs
```diff
@@ -0,0 +1,180 @@
+//! # Wasmer Runtime C API
```
MarkMcCaskey, line 1: for future reference I think `git mv` can avoid the delete/readd problem and will preserve history (I often forget to use it too)

lib/c-api/src/wasm_c_api/types/frame.rs
```diff
@@ -0,0 +1,5 @@
+#[derive(Debug)]
```
MarkMcCaskey, line 1: For types like frame and byte we should reexport them all at `types/mod.rs` or put them all in the same file I think

lib/c-api/wasmer_wasm.h
```diff
@@ -1,105 +1,543 @@
...
-//
-// The start function takes 0 arguments and returns 0 values.
-own wasm_func_t* wasi_get_start_function(wasm_instance_t*);
+typedef struct Option_Box_wasm_extern_t Option_Box_wasm_extern_t;
+
+typedef struct Option_Box_wasm_externtype_t Option_Box_wasm_externtype_t;
+
+typedef struct Option_Box_wasm_func_t Option_Box_wasm_func_t;
+
+typedef struct Option_Box_wasm_functype_t Option_Box_wasm_functype_t;
```
MarkMcCaskey, line 98: Hmm

MarkMcCaskey (commented):
Looks pretty good so far to me!

lib/c-api/build.rs
Hywan: I agree with that.

lib/c-api/src/deprecated/mod.rs
Hywan: Same here :-p.

lib/c-api/src/wasm_c_api/types/frame.rs
Hywan: Good to me.

lib/c-api/wasmer_wasm.h
Hywan: Yup, that's wrong for the moment!

Hywan: Most of them have been fixed now.

lib/c-api/build.rs
Hywan: `wasm.h` is now included, and functions and types are not exported from our code any longer by cbindgen.

Hywan:
I'm updating the test suites. Don't merge it for the moment.

Hywan:
OK, everything works now!

The source _and_ the tests are divided into 2 test suites: `deprecated` and `wasm_c_api`.
The `wasmer_wasm.h` header file is automatically generated, along with the `wasmer.h` one. The former represents the `wasm_c_api` API including our extra API, and `wasmer.h` represents the deprecated API.

`cargo test --release` to test everything.
`cargo check` or similar to build all the header files.

Let's merge with `master` and we are good to go.

lib/c-api/src/wasm_c_api/store.rs
```diff
@@ -0,0 +1,31 @@
+use super::engine::wasm_engine_t;
+use std::ptr::NonNull;
+use wasmer::Store;
+
+/// Opaque wrapper around `Store`
+/// cbindgen:ignore
+#[allow(non_camel_case_types)]
+pub struct wasm_store_t {}
+
+/// cbindgen:ignore
```
MarkMcCaskey, line 10: Is there a reason we're putting `/// cibndgen:ignore` on everything inside a module instead of at the module level?

MarkMcCaskey (approved):
Ideally we'd reduce use of `/// cbindgen:ignore` on individual items and do it at the module level.

Generally looks good to me though, unfortunately we're losing a lot of git history which may have been useful but probably not much we can reasonably do about that

Hywan:
bors try

lib/c-api/Cargo.toml
```diff
@@ -70,14 +71,18 @@ llvm = [
...
 
 # This is for compatibility for old usage
 singlepass-backend = ["singlepass"]
 cranelift-backend = ["cranelift"]
 llvm-backend = ["llvm"]
 
 [build-dependencies]
-cbindgen = { version = "0.14.3" }
+#cbindgen = "0.14.6"
+cbindgen = { git = "https://github.com/eqrion/cbindgen", branch = "master" }
```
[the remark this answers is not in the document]
Hywan: What if we target a specific commit?

lib/c-api/build.rs
```diff
@@ -40,67 +25,368 @@ fn main() {
...
+            .with_language(Language::C)
+            .with_crate(crate_dir)
+            .with_include_guard("WASMER_WASM_H")
+            .with_header(&pre_header)
+            .with_define("target_family", "windows", "_WIN32")
+            .with_define("target_arch", "x86_64", "ARCH_X86_64")
+            .with_define("feature", "jit", JIT_FEATURE_AS_C_DEFINE)
+            .with_define("feature", "compiler", COMPILER_FEATURE_AS_C_DEFINE)
+            .with_define("feature", "wasi", WASI_FEATURE_AS_C_DEFINE)
+            .with_define("feature", "emscripten", EMSCRIPTEN_FEATURE_AS_C_DEFINE)
```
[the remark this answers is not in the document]
Hywan: Good idea.

Hywan:
> Ideally we'd reduce use of /// cbindgen:ignore on individual items and do it at the module level.

If we `cbindgen:ignore` the entire `wasm_c_api` module, then we won't be able to include the `wasm_c_api::wasi` module. But we can ignore sibling modules. I'm updating the code accordingly.

> Generally looks good to me though, unfortunately we're losing a lot of git history which may have been useful but probably not much we can reasonably do about that

I just backport 2 commits manually. All the other have been ported. `git rebase` or `git merge` were in trouble with this PR :-/.

lib/c-api/build.rs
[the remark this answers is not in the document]
Hywan: Done.

lib/c-api/wasmer_wasm.h
```diff
@@ -1,105 +1,140 @@
...
 
-// Have the WASI program print directly to stderr
-void wasi_config_inherit_stderr(wasi_config_t*);
+#if defined(WASMER_WASI_ENABLED)
+void wasi_config_inherit_stderr(wasi_config_t *config);
+#endif
 
-// Have the WASI program read directly to stdin
-//void wasi_config_inherit_stdin(wasi_config_t*);
+#if defined(WASMER_WASI_ENABLED)
```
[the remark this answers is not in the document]
Hywan: Yeah, I know, I would like to work on that, but in another PR.

lib/c-api/src/wasm_c_api/store.rs
Hywan: Nop. Updated :-).

lib/c-api/src/deprecated/mod.rs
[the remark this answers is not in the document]
Hywan: Done.

lib/c-api/src/lib.rs
```diff
@@ -2,89 +2,6 @@
 #![doc(html_logo_url = "https://avatars3.githubusercontent.com/u/44205449?s=200&v=4")]
 // temporary while in transition
 #![allow(unused_variables)]
-
-//! # Wasmer Runtime C API
```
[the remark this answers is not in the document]
Hywan: Done.

lib/c-api/tests/wasm_c_api/test-early-exit.c
```diff
@@ -3,8 +3,9 @@
 #include <stdlib.h>
 #include <string.h>
 
-#include "wasm.h"
-#include "wasmer_wasm.h"
+#include "../../wasmer_wasm.h"
```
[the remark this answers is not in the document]
syrusakbary: Agreed, `"wasmer_wasm.h"` looks cleaner (without the `../`)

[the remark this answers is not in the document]
Hywan: Of course, that was temporary, just forgot to clean that up!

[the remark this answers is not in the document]
Hywan: Done.

Hywan:
bors try

Hywan:
\o/

lib/c-api/wasmer_wasm.h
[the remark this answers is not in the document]
MarkMcCaskey: Yeah, I agree! Another PR is the way to do that

lib/c-api/Cargo.toml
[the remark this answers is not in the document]
MarkMcCaskey: That won't help with publishing to crates.io, but I think we shouldn't block merging this on that just because of how much trouble it would be to delay it with other changes to the C API...

If it's on master then we can ask for another release and hopefully get one soon!

PR #1642, opened 2020-09-22
Fix up aarch64 release logic

MarkMcCaskey:
Fixed in the nightly release repo; `make package` names the file `arm64` so we rename it before uploading


nlewycky (approved):
I wish I understood why `make packages` named it with arm64?

PR #1647, opened 2020-09-24
Make CodeMemory owned by JITArtifact

nlewycky:
# Description
CodeMemory lives longer than it needs to because it's owned by the JITEngineInner. The JITArtifact would be a better place for it, once the artifact is gone, you can't call any functions on it.

In order to do this, we need to better keep track of the JITArtifact so that we don't drop it too early. Change InstanceHandle to hold an Arc<InstanceHandleInner> where the InstanceHandleInner is what used to be InstanceHandle.

In Function, when it's a WasmFunctionDefinition hold an InstanceHandle to keep the instance alive. In NativeFunc, hold an Option<InstanceHandle> to keep the instance alive when the target is a wasm function, or None for a host function.

# Review


nlewycky:
We've decided not to do this, at least not for now. We don't reference count the items inside a single store yet, and it would take a pervasive change to change that.

PR #1659, opened 2020-09-29
Remove `extern "C"` from a non-pub function that returns non-FFI'able type.

nlewycky:
Remove `extern "C"` from a non-pub function that returns non-FFI'able type.

Add `extern "C"` to two `#[no_mangle] pub` functions.


MarkMcCaskey (approved):
thanks! looks good

PR #1660, opened 2020-09-30
Allow mapdir aliases with starting `/`

### LINKED ISSUE #1503: The mapdir option unexpectedly maps to the current working directory
### Describe the bug

Take the following rename(2) test from the [wasi-test](https://github.com/caspervonb/wasi-test) project.

```c
// { "preopens": { "/scratch": "scratch" } }

#include <assert.h>
#include <stdio.h>

int main(void) {
	char oldname[] = "/scratch/oldfile";
	char newname[] = "/scratch/newfile";

	FILE *oldfile = fopen(oldname, "w");
	assert(oldfile != NULL);
	assert(fclose(oldfile) == 0);

	assert(rename(oldname, newname) == 0);

	FILE *newfile = fopen(newname, "r");
	assert(newfile != NULL);
	assert(fclose(newfile) == 0);

	return 0;
}
```

Which gets invoked as the following command

```
 wasmer run --mapdir=/scratch:scratch integration/rename.wasm 
```

```sh
wasmer 0.16.2 | rustc 1.43.1 (8d69840ab 2020-05-04) | x86_64
```

### Steps to reproduce

1. git clone https://github.com/caspervonb/wasi-test
2. python3 build.py
2. wasmer run --mapdir=/scratch:scratch integration/rename.wasm
4. mkdir scratch
5. ls scratch

### Expected behavior

/scratch to be mapped to ./scratch and it to contain a single file

### Actual behavior

/scratch is mapped to . which contains the created file

MarkMcCaskey:
Stripping leading `/` is valid because we mount mapped dirs at `/` regardless.

Resolves #1503 

This solution is a bit hacky, there may be a more elegant solution.  I'll look for one tomorrow morning when I'm more mentally fresh.


syrusakbary:
Can we add tests to the PR?

syrusakbary (approved):

lib/wasi/src/state/builder.rs
```diff
@@ -438,7 +438,17 @@ impl PreopenDirBuilder {
...
+        let alias = if let Some(first_byte) = alias.as_bytes().first() {
+            if *first_byte == b'/' {
+                &alias[1..]
+            } else {
+                alias
+            }
+        } else {
+            alias
+        }
+        .to_string();
```
syrusakbary, line 450: We should use `String::trim_start_matches`.

https://doc.rust-lang.org/std/string/struct.String.html#method.trim_start_matches

PR #1663, opened 2020-10-01
Make Function Env immutable

### LINKED ISSUE #1612: Calling a host function with environment in parallel results in data race

MarkMcCaskey:
Alternative / related to #1625

Resolves #1612 

# Review


MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

Hywan:
I agree `Env` must be immutable, and let the user use interior mutability (with `Cell`-like types) if necessary. Thanks for addressing that!

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

MarkMcCaskey:
bors try

lib/api/Cargo.toml
```diff
@@ -67,6 +67,8 @@ llvm = [
     "wasmer-compiler-llvm",
     "compiler",
 ]
+# enables internal features used by the deprecated API.
+deprecated = []
```
nlewycky, line 71: Optional: really wish we didn't have to do this. Deprecated ought to be built on top of the API. Or put another way, anything deprecated does should be supported by our real API as something that other library consumer may also want to do.

Not that I'm offering an alternative. IIUC we considered wrapping it in a Mutex and rejected that?

lib/api/src/externals/function.rs
```diff
@@ -235,7 +238,50 @@ impl Function {
...
+                vmctx,
+                signature,
+                call_trampoline: None,
+            },
+        }
+    }
+
+    /// Function used by the deprecated API to call a function with a `&mut` Env.
+    ///
+    /// This is  not a stable API and may be broken at any time.
```
nlewycky, line 261:
```suggestion
    /// This is not a stable API and may be broken at any time.
```

lib/api/src/externals/function.rs
```diff
@@ -989,6 +1043,18 @@ mod inner {
 
     impl HostFunctionKind for WithEnv {}
 
+    /// An empty struct to help Rust typing to determine
+    /// when a `HostFunction` does have an environment.
```
nlewycky, line 1047:
```suggestion
    /// when a `HostFunction` has an environment.
```

lib/api/src/externals/function.rs
```diff
@@ -989,6 +1043,18 @@ mod inner {
...
+    ///
+    /// This environment is passed by `&mut` and exists soley for the deprecated
```
nlewycky, line 1049:
```suggestion
    /// This environment is passed by `&mut` and exists solely for the deprecated
```

lib/deprecated/runtime-core/src/vm.rs
```diff
@@ -38,6 +40,9 @@ pub struct Ctx {
     pub data_finalizer: Option<fn(data: *mut c_void)>,
 }
 
+/// We mark `Ctx` as a legacy env that can be passed by `&mut`.
+unsafe impl LegacyEnv for Ctx {}
```
nlewycky, line 44: Optional: what about `UnsafeMutableEnv`?

lib/vm/src/instance.rs
```diff
@@ -278,8 +278,8 @@ impl Instance {
     }
 
     /// Return a raw pointer to the vmctx used by compiled wasm code.
-    pub fn vmctx_ptr(&self) -> *const VMContext {
-        self.vmctx() as *const VMContext
+    pub fn vmctx_ptr(&self) -> *mut VMContext {
```
nlewycky: Where's the code that needs this to be mutable?

lib/api/Cargo.toml
MarkMcCaskey: > Not that I'm offering an alternative. IIUC we considered wrapping it in a Mutex and rejected that?

That was a different PR, if it were as simple as just conditionally putting something behind a mutex, then I agree.  I tried 2-3 times to make it work without editing the API and this is the only way I've been able to make it work.

I agree that it kind of sucks, but I figure the deprecated API won't be around forever so 🤷

lib/vm/src/instance.rs
MarkMcCaskey: It's like that on master

https://github.com/wasmerio/wasmer/blob/master/lib/vm/src/instance.rs#L281-L283

I probably looked at master and got confused or something 🤷

MarkMcCaskey: Actually I don't even know how you commented on that, I don't see that in the diff 😆

lib/deprecated/runtime-core/src/vm.rs
MarkMcCaskey: Yeah, I'm open to a new name, that does sound more descriptive

lib/c-api/wasmer_wasm.h
```diff
@@ -29,9 +29,6 @@
 #  define DEPRECATED(message) __declspec(deprecated(message))
 #endif
 
-// The `jit` feature has been enabled for this build.
-#define WASMER_JIT_ENABLED
-
```
nlewycky, line 34: ??

nlewycky (approved):
IMO given how all the parts deprecated needs are now `unsafe` and documented as non-stable, you could remove the `deprecated` feature. A pro of this is that we don't have two possible builds of lib/api, with deprecated enabled and disabled. Maybe that doesn't matter.

lib/c-api/wasmer_wasm.h
MarkMcCaskey: Oops, not sure how that ended up in there

syrusakbary (approved):
Looks great to merge, once the "deprecated" comment from Nick is addressed :)

PR #1684, opened 2020-10-06
feat(c-api) Reduce the size of `libwasmer_c_api` by 23% + include `wasmer_wasm.h` inside the package

Hywan:
This PR does 2 things. Yes, that's not ideal, but it's 2 small things.

1. Reduce the size of the `libwsmer_c_api`. On my machine, it drops from 4.8.Mib to 3.7Mib.
2. Include the `wasmer_wasm.h` file inside the Wasmer packages.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,6 +124,10 @@ test-llvm = [
 # context where signal handling is a problem, such as tarpaulin for code coverage.
 test-no-traps = ["wasmer-wast/test-no-traps"]
 
+[profile.release.package.wasmer-c-api]
+opt-level = "z"
+codegen-units = 1
+
 [[bench]]
 name = "static_and_dynamic_functions"
 harness = false
--- a/Makefile
+++ b/Makefile
@@ -177,6 +177,7 @@ package-capi:
 	mkdir -p "package/include"
 	mkdir -p "package/lib"
 	cp lib/c-api/wasmer.h* package/include
+	cp lib/c-api/wasmer_wasm.h* package/include
 	cp lib/c-api/doc/index.md package/include/README.md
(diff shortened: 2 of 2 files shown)


syrusakbary (approved):

syrusakbary:
Looks good to merge!

Cargo.toml
```diff
@@ -124,6 +124,10 @@ test-llvm = [
 # context where signal handling is a problem, such as tarpaulin for code coverage.
 test-no-traps = ["wasmer-wast/test-no-traps"]
 
+[profile.release.package.wasmer-c-api]
+opt-level = "z"
```
MarkMcCaskey, line 128: `z` will prioritize code size over execution speed, something to think about.  Cargo recently introduced custom release profiles, I believe they're still nightly only though.  so in the future it'll be trivial to have settings for "smallest binary" vs "fastest binary"

Makefile
```diff
@@ -77,14 +77,28 @@ build-capi: build-capi-cranelift
...
 	cargo build --manifest-path lib/c-api/Cargo.toml --release \
 		--no-default-features --features jit,cranelift,wasi
+	${MAKE} strip-capi
 
 build-capi-llvm:
 	cargo build --manifest-path lib/c-api/Cargo.toml --release \
 		--no-default-features --features jit,llvm,wasi
+	${MAKE} strip-capi
+
+strip-capi:
```
MarkMcCaskey, line 92: stripping may make debugging problems harder, something to consider

Cargo.toml
Hywan: I know. Do we favor speed over size?

PR #1685, opened 2020-10-06
feat(c-api) Implement `wasm_exporttype_delete`

Hywan:
This PR implements the destructor for `wasm_exporttype_t`.


MarkMcCaskey (approved):
Looks great to me!  The one thing I'd want to confirm before shipping this is the ownership of `wasm_exporttype_t`, this looks like it has a legacy implementation so I'm a bit skeptical...

Yeah, exporttype_t owns both the things passed to it in new, so currently this implementation will leak memory because they're just stored as `NonNull` and never freed.

Hywan:
`wasm_exporttype_t` is built with `name` and `extern_type`, both are `NonNull`. When `wasm_exporttype_t` is dropped, none of its fields are dropped, because `NonNull` does not free the pointer when it's dropped.

I believe it is up to the user to free the `name` and the `extern_type` in this case. `wasm_exporttype_t` just owns “pointers”, not the values.

Thoughts?

Hywan:
Note: That's a little more trickier actually. It's likely that the `wasm_exporttype_t` will be created by `wasm_module_exports`. And thus, `name` and `extern_type` are created by our `From<&ExportType> for wasm_exporttype_t` implementation. Everything is put on on the stack, and no one owns it.

The solution would be to add an `owns_fields: bool` field to `wasm_exporttype_t`. If `true`, then we can `Box::from_raw` to construct a `Box` and then properly free the data. Otherwise, it's the responsability of the user to free it.

See my patch above.

PR #1687, opened 2020-10-07
Add Table example; fix local Table and Memory metadata ownership

MarkMcCaskey:
Adds an example using the Table API.

Work in progress, debugging an issue with growing using the Table API...

Things I'd like to have before shipping:
- [x] Code / comments clear enough that what's happening is fully understandable with no extra context
- [ ] Test for dynamic memory (I don't know how to test this actually works with dynamic memory without editing code, this bug manifests most strongly with dynamic memory -- it may be possible to test memory without it though) - deferred, see #1769 

# Review


syrusakbary (approved):

Hywan:
Thanks for working on this.

Hywan (commented):
Please don't forget to update `examples/README.md`.

MarkMcCaskey:
bors try

lib/engine/src/tunables.rs
```diff
@@ -19,14 +21,28 @@ pub trait Tunables {
...
     /// Create a memory given a memory type
+    ///
+    /// `vm_definition_location` should either point to a valid memory location
+    /// in VM memory or be `None` to indicate that the metadata for the memory
+    /// should be owned by the host.
     fn create_memory(
         &self,
         ty: &MemoryType,
         style: &MemoryStyle,
+        vm_definition_location: Option<NonNull<VMMemoryDefinition>>,
```
syrusakbary, line 32: If `VMMemoryDefinition` is now owned by the create_memory, how it will be Null?

lib/vm/src/memory.rs
```diff
@@ -155,7 +177,11 @@ struct WasmMmap {
 
 impl LinearMemory {
     /// Create a new linear memory instance with specified minimum and maximum number of wasm pages.
-    pub fn new(memory: &MemoryType, style: &MemoryStyle) -> Result<Self, MemoryError> {
+    pub fn new(
+        memory: &MemoryType,
+        style: &MemoryStyle,
+        vm_memory_location: Option<NonNull<VMMemoryDefinition>>,
+    ) -> Result<Self, MemoryError> {
```
syrusakbary, line 184: I don't love this API. I think we should divide into two functions.

```rust
    pub fn new(memory: &MemoryType, style: &MemoryStyle) -> Result<Self, MemoryError> {
```

And:

```rust
    pub fn from_definition(memory: &MemoryType, style: &MemoryStyle, definition: `VMMemoryDefinition`) -> Result<Self, MemoryError> {
```

(same for tables)

lib/vm/src/memory.rs
```diff
@@ -135,13 +135,35 @@ pub struct LinearMemory {
...
+#[derive(Debug)]
+enum VMMemoryDefinitionOwnership {
+    /// The `VMMemoryDefinition` is owned by the `Instance` and we should use
+    /// its memory. This is how a local memory that's exported should be stored.
+    VMOwned(NonNull<VMMemoryDefinition>),
+    /// The `VMMemoryDefinition` is owned by the host and we should manage its
+    /// memory. This is how an imported memory that doesn't come from another
+    /// Wasm module should be stored.
+    HostOwned(Box<UnsafeCell<VMMemoryDefinition>>),
+}
```
syrusakbary, line 156: I'm not super fan of this division. A memory is a memory, regardless of who owns it.

lib/engine/src/tunables.rs
MarkMcCaskey: I tried to address that in the comment above:

```Rust
    /// `vm_definition_location` should either point to a valid memory location
    /// in VM memory or be `None` to indicate that the metadata for the memory
    /// should be owned by the host.
```

The idea is that `None` means that it should be owned by the thing implementing the `Memory` trait itself.  Let me know if you think there's a clearer way to word that

MarkMcCaskey: I tried to address that in the comment above:

```Rust
    /// `vm_definition_location` should either point to a valid memory location
    /// in VM memory or be `None` to indicate that the metadata for the memory
    /// should be owned by the host.
```

The idea is that `None` means that it should be owned by the thing implementing the `Memory` trait itself.  Let me know if you think there's a clearer way to word that

lib/vm/src/memory.rs
MarkMcCaskey: Hmm, yeah that makes sense -- I'm open to that too

lib/vm/src/memory.rs
MarkMcCaskey: Yeah, this is encapsulated internally and the reason we store it this way is so that freeing memory works correctly:

`VMMemoryDefinitionOwnership::VMOwned` won't free the memory when it gets dropped but `VMMemoryDefinitionOwnership::HostOwned` will.

I agree that it might make sense to define methods on this type to avoid having to `match` in the implementation code.

examples/README.md
```diff
@@ -182,6 +182,35 @@ example.
 
    </details>
 
+### Basic Usage
```
Hywan, line 185: Maybe we should merge the `exported-function` item with this tables and memories items, into a single section named Externals or Externs, I don't know, something like that. Also don't forget to add links to the example files :-).

examples/memory.rs
```diff
@@ -0,0 +1,89 @@
+use wasmer::{
+    imports, wat2wasm, Extern, Function, Instance, Memory, MemoryType, Module, NativeFunc, Pages,
+    Store, Table, TableType, Type, Value,
+};
+use wasmer_compiler_cranelift::Cranelift;
+use wasmer_engine_jit::JIT;
+
+// this example is a work in progress:
+// TODO: clean it up and comment it
```
Hywan, line 9: Still actual ;-).

examples/memory.rs
```diff
@@ -0,0 +1,89 @@
...
+fn main() -> anyhow::Result<()> {
+    let wasm_bytes = wat2wasm(
+        r#"
+(module
+  (type $mem_size_t (func (result i32)))
+  (type $get_at_t (func (param i32) (result i32)))
+  (type $set_at_t (func (param i32) (param i32)))
+
+  (memory $mem 1)
+  ;;(import "env" "memory" (memory $mem 1))
```
Hywan, line 20: ?

Hywan:
I did a very quick review, but discussion happens on Slack, easier for the moment.

examples/memory.rs
MarkMcCaskey: Yeah I was intending to ship it like that 😆 that's a `TODO` for another time.  I can do it before shipping but I didn't want to block on it

examples/memory.rs
MarkMcCaskey: Oh this was testing code for testing the control flow of imported vs local/exported memory

lib/vm/src/memory.rs
nlewycky: Hypothetically if we don't fix it this way, then here's another thing we could do, not that it wouldn't be a ton of work. Make sure that VM* objects exist only inside the VMContext/Instance. wasmer_api:: would need two objects for memory, one that would be a handle to any kind of memory and the other which would be a host memory (detached from instance/vmcontext) and would contain the backing store. The VMMemoryImport would no longer point to a VMMemoryDefinition because a host Memory wouldn't have one, it would have to point to all the relevant parts (the backing, the enough to call grow and size, etc.) directly. The user has to keep the host memory and backing store alive, probably by Rc or Arc, and we could somewhat fold it into the handle (api::Memory) with an enum that indicates whether it's holding a whole backing store, or pointing to a VMMemoryDefinition.

lib/vm/src/table.rs
```diff
@@ -112,7 +112,20 @@ pub struct LinearTable {
     table: TableType,
     /// Our chosen implementation style.
     style: TableStyle,
-    vm_table_definition: Box<UnsafeCell<VMTableDefinition>>,
+    vm_table_definition: VMTableDefinitionOwnership,
+}
+
+/// A type to help manage who is responsible for the backing table of them
```
nlewycky, line 118:
```suggestion
/// A type to help manage who is responsible for the backing table of the
```

lib/api/src/tunables.rs
```diff
@@ -84,17 +88,55 @@ impl BaseTunables for Tunables {
...
     ) -> Result<Arc<dyn Memory>, MemoryError> {
         Ok(Arc::new(LinearMemory::new(&ty, &style)?))
     }
 
-    /// Create a table given a [`TableType`] and a [`TableStyle`].
-    fn create_table(&self, ty: &TableType, style: &TableStyle) -> Result<Arc<dyn Table>, String> {
+    /// Create a memory owned by the VM given a [`MemoryType`] and a [`MemoryStyle`].
+    ///
+    /// # Safety
+    /// - `vm_definition_location` must point to a valid location in VM memory.
```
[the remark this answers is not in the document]
MarkMcCaskey: Yeah, I just meant that it should be owned by the `VM`, I don't think it matters from this functions perspective how that is done.  I can make it more precise though

lib/engine/src/tunables.rs
```diff
@@ -46,28 +75,36 @@ pub trait Tunables {
             let mi = MemoryIndex::new(index);
             let ty = &module.memories[mi];
             let style = &memory_styles[mi];
+            // TODO: error handling
```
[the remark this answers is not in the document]
MarkMcCaskey: Thanks for the ping, this would probably be best to clean up in this PR

lib/vm/src/instance.rs
```diff
@@ -772,6 +764,74 @@ pub struct InstanceHandle {
 unsafe impl Send for InstanceHandle {}
 
 impl InstanceHandle {
+    /// Allocates an instance for use with `InstanceHandle::new`.
+    ///
+    /// Returns the instance pointer and the [`VMOffsets`] that describe the
+    /// memory.
+    pub fn allocate_instance(module: &ModuleInfo) -> (NonNull<u8>, VMOffsets) {
```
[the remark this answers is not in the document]
MarkMcCaskey: I believe `VMOffsets` should be computable from just a Module -- the main thing this type does is tell us how many local functions, globals, etc. there are and does some arithmetic based on that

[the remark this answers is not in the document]
MarkMcCaskey: `VmContext` is probably a good change 👍 ; will need to look into if there are any edge cases to watch out for if we deref a pointer to a zero-sized type

lib/vm/src/memory.rs
MarkMcCaskey: The idea being that `VMMemoryImport` would never own its metadata?

I'm not sure I fully understood what you're suggesting but to make sure we're on the same page:
- The generated code always needs a `VMMemoryDefinition`.
   - If it's local, it lives in VM-owned memory
   - if it's imported it lives somewhere else (either another VM's memory or the host)
- In the current code, due to `Arc`, an Instance using an imported host type will ensure that it's never freed while the Instance is using it.
- We need an extra layer of indirection for imported VM Objects to make sure that modifications are reflected in the generated code.  It's possible to remove this layer of indirection but it means we'd need update logic.

nlewycky: Right, the idea being that VMMemoryImport would never own its metadata. In this model the generated code would need to change to work with a VMMemoryImport.

The compiler-generated code needs to know the base pointer and the size. It also has to be able to grow and size, both of which go through libcalls (even in the local owned memory case). More subtly we need invalidation rules for the base pointer and the size, for instance can we assume they never change at any time the function is executing?

One possible solution. You could have the VMMemoryImport have ptr_to_ versions of the fields the VMMemoryDefinition has. Instead of ptr-to-vmmemorydefinition and arc<dyn Memory> it would have a ptr_to_base and ptr_to_current_length, then the Arc<dyn Memory> (or whatever we call the object that keeps the pointed-to-memory alive, if anything) goes in the Instance::memories, not the VMContext objects.

The important part of my suggestion is to start with a rule that VM* objects are only ever part of a VMContext and then follow that through the rest of the design.

lib/vm/src/instance.rs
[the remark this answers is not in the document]
MarkMcCaskey: It's actually not a pointer to `VMContext` and I think changing it to that would be a bit confusing, it really is the pointer to `vm::Instance` (which includes the `VMContext`) but we don't expose that type so we can't make the pointer accurate, I think the best thing to do is to leave it as a pointer to u8 for now.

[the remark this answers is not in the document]
nlewycky: Why not pointer to `vm::Instance`? I really did think it was a pointer to VMContext, this would solve that?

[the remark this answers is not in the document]
MarkMcCaskey: `vm::Instance` is not public and can't be part of public type signatures, we don't expose that type

lib/vm/src/instance.rs
```diff
@@ -843,19 +892,15 @@ impl InstanceHandle {
...
-                alloc::handle_alloc_error(layout);
-            }
             ptr::write(instance_ptr, instance);
             Self {
                 instance: instance_ptr,
             }
         };
         let instance = handle.instance();
 
+        // split this into two steps
```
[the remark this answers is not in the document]
MarkMcCaskey: Good catch, I think it's an old implementation note for myself: that's where allocation used to happen, but I moved it out (into two steps), I'll delete it

lib/vm/src/memory.rs
MarkMcCaskey: > The important part of my suggestion is to start with a rule that VM* objects are only ever part of a VMContext and then follow that through the rest of the design.

I think this makes ownership harder in the case of a VM object imported into multiple instances, who is responsible for freeing it? What about memories created when there are no Instances created?

I'm not necessarily opposed to the changes but I think we should land this PR first, this change is self-contained and fixes user-facing bugs; the implementation can easily be changed later in my opinion.

syrusakbary (approved):

PR #1689, opened 2020-10-08
feat(c-api) Use `Option<NonNull<wasm_valtype_t>>` in `wasm_valtype_kind`

Hywan:
A more Rust-FFI idiomatic way to handle null pointer.

Note: In `wasm_valtype_kind`, it's tricky to handle the error because
we _must_ return a `wasm_valtype_kind` value. For the moment, it
continues to panic, which is probably the best tradeoff.

### DIFF
--- a/lib/c-api/src/wasm_c_api/types/value.rs
+++ b/lib/c-api/src/wasm_c_api/types/value.rs
@@ -1,5 +1,6 @@
 use super::super::value::wasm_valkind_t;
 use std::convert::TryInto;
+use std::ptr::NonNull;
 use wasmer::ValType;
 
 #[allow(non_camel_case_types)]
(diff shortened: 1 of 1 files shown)


lib/c-api/src/wasm_c_api/types/value.rs
```diff
@@ -83,10 +84,11 @@ pub extern "C" fn wasm_valtype_new(kind: wasm_valkind_t) -> Option<Box<wasm_valt
...
 
 #[no_mangle]
-pub unsafe extern "C" fn wasm_valtype_kind(valtype: *const wasm_valtype_t) -> wasm_valkind_t {
-    if valtype.is_null() {
-        // TODO: handle error
-        panic!("wasm_valtype_kind: argument is null pointer");
-    }
-    return (*valtype).valkind as wasm_valkind_t;
+pub unsafe extern "C" fn wasm_valtype_kind(
+    valtype: Option<NonNull<wasm_valtype_t>>,
```
MarkMcCaskey, line 88: Could we just use `Option<&wasm_valtype_t>` here?

Hywan: We can yes. To be honest, I'm still not sure what the policy regarding `null` value in our implementation or in the Wasm C API.

syrusakbary (approved):

jubianchi (approved):
Seems good to me 👍

PR #1690, opened 2020-10-08
fix(c-api) `wasm_limits_t` contains `Pages`, not `Bytes` + sentinel value

Hywan:
When building a `wasm_memorytype_t` with `wasm_memorytype_new`, we
pass a `wasm_limits_t`, where `min` and `max` represent `Pages`. This
semantics is set by `wasm_memorytype_new` itself where `min` and `max`
from `wasm_limits_t` are used to compute `Pages`, which are then passed
to `MemoryType`.

Then, in `wasm_memorytype_limits`, we expect to get the same
`wasm_limits_t` given to `wasm_memorytype_new`. But it's not!

The same `MemoryType` is read, good. The `minimum` and `maximum`
fields are `Pages`, good. Then, we compute the `min` and `max` values
for the resulting `wasm_limits_t`, which receive `Page.bytes().0`, not
good! We don't want the number of bytes, but the number of pages.

This patch fixes that.

Edit: This patch also fixes the max limit sentinel value. In `wasm.h`, it is defined by `wasm_limits_max_default` (`0xffffffff`). In our code, it is correctly used in `wasm_memorytype_new`, but not in `wasm_memorytype_limits`. A new constant has been introduced: `LIMITS_MAX_SENTINEL`. Not super happy with this name though.

### DIFF
--- a/lib/c-api/src/wasm_c_api/types/memory.rs
+++ b/lib/c-api/src/wasm_c_api/types/memory.rs
@@ -58,6 +58,6 @@ pub unsafe extern "C" fn wasm_memorytype_limits(mt: &wasm_memorytype_t) -> *cons
 
     Box::into_raw(Box::new(wasm_limits_t {
         min: md.minimum.0 as _,
-        max: md.maximum.map(|max| max.0 as _).unwrap_or(0),
+        max: md.maximum.map(|max| max.0 as _).unwrap_or(u32::max_value()),
     }))
 }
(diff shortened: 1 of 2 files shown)


syrusakbary (approved):

lib/c-api/src/wasm_c_api/types/memory.rs
```diff
@@ -54,8 +57,12 @@ pub unsafe extern "C" fn wasm_memorytype_delete(_memorytype: Option<Box<wasm_mem
...
     let md = mt.as_memorytype();
+
     Box::into_raw(Box::new(wasm_limits_t {
-        min: md.minimum.bytes().0 as _,
-        max: md.maximum.map(|max| max.bytes().0 as _).unwrap_or(0),
+        min: md.minimum.0 as _,
+        max: md
+            .maximum
+            .map(|max| max.0 as _)
+            .unwrap_or(LIMITS_MAX_SENTINEL),
```
MarkMcCaskey, line 66: Good catch, I didn't see that

PR #1693, opened 2020-10-08
Add `wasmer create-exe`

MarkMcCaskey:
This adds the `wasmer create-exe` subcommand.  This subcommand is a combination of `wasmer compile --object-file` and linking that compiled Wasm with a main function and libwasmer.  Put more plainly: it lets us transform Wasm modules into native executables in one step.

In order for this to work we need:
- [x] Ship wasm.h with Wasmer or use different mechanism to find it
- [x] Ship wasmer_wasm.h with Wasmer
- [x] Requires up to date libwasmer... had to build one and copy it over, may fail in CI because of this... will be fixed with next release though
- [x] More gracefully handle wasmer installed without WASMER_DIR, etc (even if just error messages, should be tested)
- [x] Add tests

# Review


lib/cli/src/commands/create_exe.rs
```diff
@@ -0,0 +1,273 @@
...
+use std::fs;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+use structopt::StructOpt;
+use wasmer::*;
+
+const WASMER_MAIN_C_SOURCE: &[u8] = include_bytes!("wasmer_create_exe_main.c");
+
+// TODO: to get this working locally I had to add wasmer_wasm.h and wasm.h to `.wasmer/include` manually.
+// this needs to be fixed before we can ship this.
```
syrusakbary, line 15: Make sure we ship `wasmer_wasm.h` and `wasm.h` before merging

lib/cli/src/commands/wasmer_create_exe_main.c
```diff
@@ -0,0 +1,89 @@
...
+        }
+        
+        // We have now finished the memory buffer book keeping and we have a valid Module.
+
+        // In this example we're passing some JavaScript source code as a command line argument
+        // to a WASI module that can evaluate JavaScript.
+        wasi_config_t* wasi_config = wasi_config_new("constant_value_here");
+        const char* js_string = "function greet(name) { return JSON.stringify('Hello, ' + name); }; print(greet('World'));";
+        wasi_config_arg(wasi_config, "--eval");
+        wasi_config_arg(wasi_config, js_string);
```
syrusakbary, line 46: We probably don't want this.

We would probably want to replace `constant_value_here` with the first argument to main.
We also would like to relay the rest of the arguments also to WASI.

It would be a good idea to get some other thoughts on mounted dirs and environments. Perhaps those can be provided as configuration to the `create-exe` subcommand.

lib/cli/src/commands/create_exe.rs
MarkMcCaskey: It's on the todo list at the top of the issue (you may need to refresh the page to get Github to show you that though)

lib/cli/src/commands/create_exe.rs
```diff
@@ -0,0 +1,273 @@
...
+    let mut path = PathBuf::from(env::var("WASMER_DIR")?);
+    path.push("lib");
+
+    #[cfg(not(windows))]
+    path.push("libwasmer.a");
+    #[cfg(windows)]
+    path.push("libwasmer.lib");
+
+    Ok(path)
+}
```
syrusakbary, line 168: It would be a good idea to abstract this out, and also reuse them on the `lib/cli/src/commands/config.rs` file

lib/cli/src/store.rs
```diff
@@ -36,100 +55,16 @@ pub struct StoreOptions {
...
-    #[structopt(long, conflicts_with_all = &["native", "object_file"])]
-    jit: bool,
-
-    /// Use Native Engine.
-    #[structopt(long, conflicts_with_all = &["jit", "object_file"])]
-    native: bool,
-
-    /// Use ObjectFile Engine.
-    #[structopt(long, conflicts_with_all = &["jit", "native"])]
-    object_file: bool,
```
syrusakbary, line 49: Great call on moving this out

Hywan:
I'm not sure what `create-exe` does. Can you elaborate please?

lib/cli/src/commands/create_exe.rs
MarkMcCaskey: Yeah it looks like the config logic isn't tested right now... changes Ivan and I have been making to the C API are going to break it I think...

The config code implies we can just do `-lwasmer` which is interesting that would make the code have fewer `cfg`s in it.

MarkMcCaskey: The code in `wasmer config` for linking against wasmer seems to default to the dynamic libwasmer 🤔

lib/cli/src/commands/wasmer_create_exe_main.c
```diff
@@ -0,0 +1,172 @@
...
+}
+#endif
+
+int main(int argc, char* argv[]) {
+  wasm_config_t* config = wasm_config_new();
+  wasm_config_set_engine(config, OBJECT_FILE);
+  wasm_engine_t* engine = wasm_engine_new_with_config(config);
+  wasm_store_t* store = wasm_store_new(engine);
+  
+  wasm_module_t* module = wasmer_object_file_engine_new(store, "qjs.wasm");
```
syrusakbary, line 114: I think this is a typo. Can we use the first argument in `argv` (or "wasmer-exe" in case argc == 0)

lib/cli/src/commands/create_exe.rs
```diff
@@ -0,0 +1,300 @@
...
+                    .canonicalize()
+                    .context("Failed to find libwasmer")?,
+            );
+        let command = if let Some(target) = &self.target {
+            command.arg("-target").arg(format!("{}", target))
+        } else {
+            command
+        };
+        #[cfg(windows)]
+        let command = command.arg("-luserenv").arg("-lWs2_32").arg("-ladvapi32");
```
syrusakbary, line 279: Can we comment what this libraries are necessary for? (a comment in the code would be good enough)

lib/cli/src/commands/create_exe.rs
```diff
@@ -0,0 +1,300 @@
...
+    let mut path = get_wasmer_dir()?;
+    path.push("include");
+    Ok(path)
+}
+
+/// path to the static libwasmer
+fn get_libwasmer_path() -> anyhow::Result<PathBuf> {
+    let mut path = get_wasmer_dir()?;
+    path.push("lib");
+
```
syrusakbary, line 170: Add a comment:

```rust
/// TODO: use headless wasmer
```

lib/cli/src/commands/create_exe.rs
```diff
@@ -0,0 +1,300 @@
...
+        println!("Engine: {}", engine_type.to_string());
+        println!("Compiler: {}", compiler_type.to_string());
+        println!("Target: {}", target.triple());
+
+        let working_dir = tempfile::tempdir()?;
+        let starting_cd = env::current_dir()?;
+        let output_path = starting_cd.join(&self.output);
+        env::set_current_dir(&working_dir)?;
+
+        // TODO: encapsulate compile code
```
syrusakbary, line 76: What this means?

lib/cli/src/commands/create_exe.rs
```diff
@@ -0,0 +1,300 @@
...
+        #[cfg(windows)]
+        let c_src_obj = PathBuf::from("wasmer_main.obj");
+
+        {
+            let mut c_src_file = fs::OpenOptions::new()
+                .create_new(true)
+                .write(true)
+                .open(&c_src_path)
+                .context("Failed to open C source code file")?;
+            // TODO:
```
syrusakbary, line 130: TODO?

tests/integration/cli/src/link_code.rs
```diff
@@ -0,0 +1,70 @@
...
+                self.object_paths
+                    .iter()
+                    .map(|path| path.canonicalize().unwrap()),
+            )
+            .arg(&self.libwasmer_path.canonicalize()?);
+        #[cfg(windows)]
+        let command = command.arg("-luserenv").arg("-lWs2_32").arg("-ladvapi32");
+        #[cfg(not(windows))]
+        // TODO: remove `-lffi` before shipping
+        let command = command.arg("-ldl").arg("-lm").arg("-pthread").arg("-lffi");
```
syrusakbary, line 56: TODO: Remove this :)

tests/integration/cli/src/link_code.rs
```diff
@@ -0,0 +1,70 @@
...
+            .current_dir(&self.current_dir)
+            .arg(&self.optimization_flag)
+            .args(
+                self.object_paths
+                    .iter()
+                    .map(|path| path.canonicalize().unwrap()),
+            )
+            .arg(&self.libwasmer_path.canonicalize()?);
+        #[cfg(windows)]
+        let command = command.arg("-luserenv").arg("-lWs2_32").arg("-ladvapi32");
```
syrusakbary, line 53: It seems this logic is duplicated?

lib/cli/src/commands/wasmer_create_exe_main.c
MarkMcCaskey: argc should never be 0 I think, there's always supposed to be 1 arg which is the name of the program... We could handle it but I'm not aware of a situation in which that can happen

tests/integration/cli/src/link_code.rs
MarkMcCaskey: Yeah, this is test code.  I'm not sure where a good place to share this code would be.

We could maybe make it public in the `wasmer-cli` crate and then depend on it, but I think it's probably fine to duplicate

Those flags are required to link with `wasmer compile`, but since we do that in create-exe we also do that

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/cli/src/commands/create_exe.rs
```diff
@@ -0,0 +1,304 @@
...
+/// Compile the C code.
+fn run_c_compile(
+    path_to_c_src: &Path,
+    output_name: &Path,
+    target: Option<Triple>,
+) -> anyhow::Result<()> {
+    #[cfg(not(windows))]
+    let c_compiler = "cc";
+    #[cfg(windows)]
+    let c_compiler = "clang++";
```
syrusakbary, line 191: Just caught this. Why clang++ in Windows? Can we add a comment?

syrusakbary (approved):

jaykrell:
Sounds great, not very turnkey or well documented from what I see so far..

Hywan:
It is true. We are going to document that more. Thanks!

PR #1694, opened 2020-10-08
Update to cbindgen 0.15.0

MarkMcCaskey:
Removes a git dependency which would prevent us from publishing the C API as a crate to crates.io.

It seems that cbindgen 0.15.0 stops sorting output by default, so we now manually specify that we want that.

### DIFF
--- a/lib/c-api/Cargo.toml
+++ b/lib/c-api/Cargo.toml
@@ -88,5 +88,4 @@ cranelift-backend = ["cranelift"]
 llvm-backend = ["llvm"]
 
 [build-dependencies]
-#cbindgen = "0.14.6"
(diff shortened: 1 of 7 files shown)


syrusakbary (approved):

lib/c-api/Cargo.toml
```diff
@@ -88,5 +88,4 @@ cranelift-backend = ["cranelift"]
 llvm-backend = ["llvm"]
 
 [build-dependencies]
-#cbindgen = "0.14.6"
-cbindgen = { git = "https://github.com/eqrion/cbindgen", branch = "master" }
\ No newline at end of file
+cbindgen = "0.15"
```
nlewycky, line 91: newline please?

nlewycky (approved):

PR #1699, opened 2020-10-09
feat(c-api) Update `wasm-c-api` repository

Hywan:
This change is important because `wasm.h` contains new functions, like
`wasm_name_new_from_string_nt`, which are useful for the Go
implementation.


lib/c-api/tests/wasm_c_api/wasm-c-api/README.md
```diff
@@ -94,6 +94,7 @@ Currently, known implementations of this API are included in
 * V8 natively (both C and C++)
 * Wabt (only C?)
 * Wasmtime (only C?)
+* [Wasmer](https://github.com/wasmerio/wasmer/tree/master/lib/c-api) (only C, C++ coming soon)
```
MarkMcCaskey, line 97: Unrelated to this PR but I want to put this idea out into the universe before I forget:

When doing the release we have `c-api/docs/deprecated` and we currently package that as `README.md` in the C API that we distribute...

We should probably make a note that it's the deprecated API and file an issue to start building docs for Wasmer with the Wasm C API!

MarkMcCaskey (approved):
Looks good to me...

we may need to update some of our own tests as well if any behavior has changed

lib/c-api/tests/wasm_c_api/wasm-c-api/README.md
syrusakbary: Completely!

Hywan (commented):
In addition to the updates required by the changes in  `wasm.h`, I've updated `wasi/mod.rs` (the `wasm_get_imports` function) to use a `wasm_extern_vec_t`, so that it matches the definition of `wasm_instance_new`, which is simpler and more consistent.

Hywan:
@MarkMcCaskey We must be very careful with the last update of wasm.h. The following code is now a common pattern:

```
wasm_val_vec_t args = WASM_EMPTY_VEC;
```

where `WASM_EMPTY_VEC` is a new macro such as:

```
#define WASM_EMPTY_VEC {0, NULL}
```

So basically, it creates a `wasm_val_vec_t { .size = 0, .data = NULL }`.

Why is it dangerous for us? Because our `into_slice` implementation for `wasm_$name_vec_t` returns `None` if `data` is `null`. Yup. So an empty array passed to `into_slice` will return `None`.

I don't know if we should change the behavior of `into_slice`. Maybe we should. At least to return an empty static slice (`[]`)? Thoughts?

Hywan:
All `wasm-c-api` examples/tests are green. Let's migrate our own tests.

Hywan:
All good. All `wasmer_c_api` tests are green. Let's see how it impacts all the project.

Hywan:
bors try

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -86,30 +94,36 @@ pub unsafe extern "C" fn wasm_func_new_with_env(
...
                     kind: wasm_valkind_enum::WASM_I64 as _,
                     of: wasm_val_inner { int64_t: 0 },
                 };
                 num_rets
-            ];
+            ]
+            .into();
 
-            let _traps = callback(*env, processed_args.as_ptr(), results.as_mut_ptr());
+            let _traps = callback(*env, &processed_args, &mut results);
```
MarkMcCaskey, line 113: This might not be the same, we should be careful of the difference between fat pointers and thin pointers.

`as_ptr` and `as_mut_ptr` generally pass word sized pointers, references to slices are double words and references to vecs are different data entirely

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -124,21 +138,29 @@ pub unsafe extern "C" fn wasm_func_delete(_func: Option<Box<wasm_func_t>>) {}
...
+    let params = args
+        .into_slice()
+        .map(|slice| {
+            slice
+                .into_iter()
+                .map(TryInto::try_into)
+                .collect::<Result<Vec<Val>, _>>()
+                .expect("Argument conversion failed")
+        })
+        .unwrap_or_else(|| Vec::new());
```
MarkMcCaskey, line 153: `unwrap_or_default` also works here

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -124,21 +138,29 @@ pub unsafe extern "C" fn wasm_func_delete(_func: Option<Box<wasm_func_t>>) {}
...
-        .collect::<Result<_, _>>()
-        .ok()?;
+    let params = args
+        .into_slice()
+        .map(|slice| {
+            slice
+                .into_iter()
+                .map(TryInto::try_into)
+                .collect::<Result<Vec<Val>, _>>()
+                .expect("Argument conversion failed")
```
MarkMcCaskey, line 151: We used to report these failures as errors to the user: why is it `TryInto` instead of `Into`? Is this failure something users could encounter?

lib/c-api/src/wasm_c_api/instance.rs
```diff
@@ -12,56 +12,21 @@ pub struct wasm_instance_t {
...
     _traps: *mut *mut wasm_trap_t,
 ) -> Option<Box<wasm_instance_t>> {
     let wasm_module = &module.inner;
     let module_imports = wasm_module.imports();
     let module_import_count = module_imports.len();
-    let imports = argument_import_iter(imports);
     let resolver: OrderedResolver = imports
+        .into_slice()
+        .map(|imports| imports.iter())
+        .unwrap_or_else(|| [].iter())
```
MarkMcCaskey, line 29: `unwrap_or_default` _might_ work here

lib/c-api/src/wasm_c_api/externals/function.rs
Hywan: Good eye, indeed!

lib/c-api/src/wasm_c_api/externals/function.rs
Hywan: Yes, because `wasm_val_t` contains an enum that we must build from `u8`. We can out-of-bounds values. A `TryInto` is appropriate here.

lib/c-api/src/wasm_c_api/instance.rs
Hywan: Nah, it doesn't, good try :-p.

lib/c-api/src/wasm_c_api/externals/function.rs
Hywan: `processed_args` and `results` are now of kind `wasm_val_vec_t`. I believe it is safe to use `&` in this case.

Hywan:
@MarkMcCaskey Please, can you review the following changes:

* https://github.com/wasmerio/wasmer/pull/1699/files#diff-b7936550e9c4136a8e2504b59f17b1dbdc8fd2df8aed8a92299dd012b013054d
* https://github.com/wasmerio/wasmer/pull/1699/files#diff-520b2626a7b19427742c3ec992a4346fe6a9334e7b591b2e5f68644889d25372

Thanks!

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

Hywan:
bors cancel

Hywan:
bors try-

Hywan:
bors try

MarkMcCaskey:
bors try

PR #1700, opened 2020-10-09
feat(c-api) Implement `wasm_externtype_copy`

Hywan:
This patch implements the missing `wasm_externtype_copy` function.

### DIFF
--- a/lib/c-api/src/wasm_c_api/types/extern_.rs
+++ b/lib/c-api/src/wasm_c_api/types/extern_.rs
@@ -21,6 +21,15 @@ pub unsafe extern "C" fn wasm_extern_type(e: &wasm_extern_t) -> Box<wasm_externt
 #[no_mangle]
 pub unsafe extern "C" fn wasm_externtype_delete(_et: Option<Box<wasm_externtype_t>>) {}
 
+#[no_mangle]
+pub extern "C" fn wasm_externtype_copy(
+    wasm_externtype: &wasm_externtype_t,
+) -> Box<wasm_externtype_t> {
+    Box::new(wasm_externtype_t {
+        inner: wasm_externtype.inner.clone(),
+    })
+}
+
 impl From<ExternType> for wasm_externtype_t {
     fn from(other: ExternType) -> Self {
         Self { inner: other }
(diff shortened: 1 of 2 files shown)


lib/c-api/src/wasm_c_api/types/extern_.rs
```diff
@@ -21,6 +21,16 @@ pub unsafe extern "C" fn wasm_extern_type(e: &wasm_extern_t) -> Box<wasm_externt
 #[no_mangle]
 pub unsafe extern "C" fn wasm_externtype_delete(_et: Option<Box<wasm_externtype_t>>) {}
 
+#[no_mangle]
+pub unsafe extern "C" fn wasm_externtype_copy(
```
MarkMcCaskey, line 25: This function does not need to be `unsafe` as it's valid within a Rust context.

All `unsafe` means is that the Rust type system is insufficient to cover the behavior of the function. But when we use higher level types, this is often no longer the case.  And `unsafe` has no effect in regards to FFI.

This is a super minor change and it's a mistake I make constantly too and we have a lot of instances of this in the wasm C API, so I don't think it's worth blocking this PR over until we start doing a good job at removing it 😆

lib/c-api/src/wasm_c_api/types/extern_.rs
```diff
@@ -21,6 +21,16 @@ pub unsafe extern "C" fn wasm_extern_type(e: &wasm_extern_t) -> Box<wasm_externt
...
+    wasm_externtype: &wasm_externtype_t,
+) -> Box<wasm_externtype_t> {
+    // do shallow copy
```
MarkMcCaskey, line 28: Is this really a shallow copy?  It makes a new box and clones the internal type

MarkMcCaskey (approved):

lib/c-api/src/wasm_c_api/types/extern_.rs
Hywan: Yeah, that was a reflex attached to this codebase :-D. Fixed.

lib/c-api/src/wasm_c_api/types/extern_.rs
Hywan: It's not. Fixed!

PR #1709, opened 2020-10-12
feat(c-api) Implement `wasm_module_name` and `wasm_module_set_name`

Hywan:
Until my official proposal made here, https://github.com/WebAssembly/wasm-c-api/pull/157 is accepted or rejected, I implement those functions as “wasmer-specific” API.


lib/c-api/src/wasm_c_api/wasmer.rs
```diff
@@ -1,9 +1,35 @@
...
+
+    *out = name.as_bytes().to_vec().into();
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn wasm_module_set_name(module: &wasm_module_t, name: &wasm_name_t) -> bool {
+    let name = match name.into_slice() {
+        Some(name) => match str::from_utf8(name) {
+            Ok(name) => name,
+            Err(_) => return false, // not ideal!
```
MarkMcCaskey, line 29: sometimes what I do for cases like this is to make a private inner function which just returns Option<T> so you can use c_try internally and then do all the logic there and then convert it to whatever return type you actually need...

You can even do it with an anonymous function too

MarkMcCaskey (approved):

lib/c-api/src/wasm_c_api/wasmer.rs
Hywan: I noticed, but I thought this use case was too small to deserve an inner function :-).

PR #1710, opened 2020-10-12
Move function call trampolines to Artifact instead of EngineInner.

nlewycky:
# Description
Move function call trampolines to Artifact instead of EngineInner.

# Review


lib/vm/src/export.rs
```diff
@@ -35,6 +35,8 @@ pub struct ExportFunction {
     pub signature: FunctionType,
     /// The function kind (it defines how it's the signature that provided `address` have)
     pub kind: VMFunctionKind,
+    /// Address of the function call trampoline owned by the same VMContext that owns the VMFunctionBody. May be None when FunctionType == Dynamic or vmctx == nullptr.
```
syrusakbary, line 38: Note: we should remove `or vmctx == nullptr`, it doesn't make sense for trampoline to be None if vmctx is null
```suggestion
    /// Address of the function call trampoline owned by the same VMContext that owns the VMFunctionBody.
    /// May be None when FunctionType == Dynamic.
```

nlewycky: If vmctx == nullptr then it must be a host function, and host functions never have trampolines. Is that not right?

syrusakbary: It got me a bit confused. Perhaps this description is actually better:
```suggestion
    /// Address of the function call trampoline owned by the same VMContext that owns the VMFunctionBody.
    /// May be None when the function is an host-function (FunctionType == Dynamic or vmctx == nullptr)
```

syrusakbary (changes requested):
We can also delete the engine `register_signature`.

And with a bit more effort, we could do the `lookup_signature` as well

Thoughts @nlewycky ?

nlewycky:
> We can also delete the engine `register_signature`.
> 
> And with a bit more effort, we could do the `lookup_signature` as well
> 
> Thoughts @nlewycky ?

I tried removing them while making this PR, then had to put them back. We still need a single signature index across all Instances in the same Store. I can investigate trying again once this PR has landed.

syrusakbary:
> We can also delete the engine register_signature.

I believe `register_signature` is no longer used anywhere?

nlewycky:
> > We can also delete the engine register_signature.
> 
> I believe `register_signature` is no longer used anywhere?

https://github.com/wasmerio/wasmer/blob/ff15be3a7528b16dba804ae7b4b6a73bb62d2886/lib/api/src/externals/function.rs#L379

```
    pub(crate) fn checked_anyfunc(&self) -> VMCallerCheckedAnyfunc {
        let vmsignature = self
            .store
            .engine()
            .register_signature(&self.exported.signature);
        VMCallerCheckedAnyfunc {
            func_ptr: self.exported.address,
            type_index: vmsignature,
            vmctx: self.exported.vmctx,
        }
    }
```

syrusakbary:
bors try

lib/vm/src/export.rs
```diff
@@ -35,6 +35,9 @@ pub struct ExportFunction {
     pub signature: FunctionType,
     /// The function kind (it defines how it's the signature that provided `address` have)
     pub kind: VMFunctionKind,
+    /// Address of the function call trampoline owned by the same VMContext that owns the VMFunctionBody.
+    /// May be None when the function is an host-function (FunctionType == Dynamic or vmctx == nullptr).
+    pub trampoline: Option<VMTrampoline>,
```
syrusakbary, line 40:
```suggestion
    pub call_trampoline: Option<VMTrampoline>,
```

syrusakbary (approved):

PR #1711, opened 2020-10-12
Export the VM functions when building on ARM.

nlewycky:
This is needed to be able to run engine-native built object files on aarch64 linux.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,9 +1,4 @@
-[target.x86_64-unknown-linux-gnu]
-rustflags = [
-  "-C", "link-arg=-Wl,-E"
-]
-
-[target.aarch64-unknown-linux-gnu]
+[target.'cfg(target_os = "linux")']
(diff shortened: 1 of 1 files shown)


.cargo/config.toml
```diff
@@ -2,3 +2,8 @@
 rustflags = [
   "-C", "link-arg=-Wl,-E"
 ]
+
+[target.aarch64-unknown-linux-gnu]
+rustflags = [
+  "-C", "link-arg=-Wl,-E"
```
syrusakbary, line 8: Should we add them as a flag always? (not just fofr x86_64 or aarch64, but for all architectures)

nlewycky: That would be great, but the flag only works on Linux. I don't think this file allows globs or regexes?

nlewycky: Done.

syrusakbary (approved):

PR #1715, opened 2020-10-13
fix(c-api) Register errors from `wasm_module_serialize`

Hywan:
This patch updates `wasm_module_serialize` to register errors with`update_last_error` if any.


MarkMcCaskey (approved):
Seems good, perhaps we want to write an "error" value out at out_ptr though, like all 0s.

Currently this can report errors but there's no easy way to check if this function errored because `update_last_error` is global and could have been set by something else

Hywan:
Yes I know. The strategy for the other similar API is to keep the `out` parameter as is. Usually, the user passes a `NULL` value, so if `NULL` is still there, it's a sign for an error. The `wasm.h` doesn't precise anything about such behavior :-/.

PR #1720, opened 2020-10-14
Autodetect llvm regardless of architecture.

nlewycky:


syrusakbary (approved):

nlewycky:
We can't land this just yet, it will cause `make test` to fail on arm as it runs `make test-llvm-jit` which doesn't work.

MarkMcCaskey (approved):

nlewycky:
This is now substantially different from when it was last reviewed, please take another look!

jubianchi (approved):

Issue #1360, opened 2020-04-07
Allow configuration of default memory


webmaster128:
### Motivation

I'd like to be able to limit the memory consumption of a Wasm sandbox. It seems like `MemoryDescriptor` can do that. However, I don't see a way to configure the default memory (`memory(0)`), which is created for me automatically.

### Proposed solution

After digging into the code, I found a way to patch the memory limit into Wasmer (0.16.2) like this:

```diff
diff --git a/lib/runtime-core/src/lib.rs b/lib/runtime-core/src/lib.rs
index 7b6168c5f..a9cd4189a 100644
--- a/lib/runtime-core/src/lib.rs
+++ b/lib/runtime-core/src/lib.rs
@@ -103,6 +103,9 @@ pub mod prelude {
     pub use crate::{func, imports};
 }
 
+use crate::types::{LocalMemoryIndex};
+use crate::units::{Pages};
+
 /// Compile a [`Module`] using the provided compiler from
 /// WebAssembly binary code. This function is useful if it
 /// is necessary to a compile a module before it can be instantiated
@@ -117,6 +120,11 @@ pub fn compile_with(
         .map(|mut inner| {
             let inner_info: &mut crate::module::ModuleInfo = &mut inner.info;
             inner_info.import_custom_sections(wasm).unwrap();
+            let keys: Vec<LocalMemoryIndex> = inner_info.memories.iter().map(|(k, _)| k).collect();
+            for key in keys {
+                inner_info.memories[key].maximum = Some(Pages(512)); // 32 MiB
+            }
+            // println!("Module memories: {:?}", inner_info.memories);
             module::Module::new(Arc::new(inner))
         })
 }
```

The loop iterates over one key only. For some visibility reasons I was not able to construct `LocalMemoryIndex(0)` directly.

This patch changes the default memory from

> Memory descriptor: MemoryDescriptor { minimum: 17 pages, maximum: None, shared: false, memory_type: Dynamic }; size: 18 pages (1179648 bytes)

to

> Memory descriptor: MemoryDescriptor { minimum: 17 pages, maximum: Some(512 pages), shared: false, memory_type: Dynamic }; size: 18 pages (1179648 bytes)

and in my tests this new maximum seems to be respected.

The problem with this is primarily, that `compiler.compile` is explicitely designed to be crate internal and I don't see an API that allows for setting memory limits.

### Alternatives

I don't know if this is the correct approach to set memory limits.

### Additional context

[...]

PR #1730, opened 2020-10-16
Create example for limiting memory with Tunables

### LINKED ISSUE #1588: How to use new Tunables API
We need an example how to use the new Tunables API

webmaster128:
Closes #1588

# Description

This is an attempt for solving #1360 with the new Tunables API. It could be developed further to address #1588 if you like.

I'd appreciate a code review to ensure I'm on the right track.

Open questions:
1. Is it expected that I had to add `wasmer-vm`?
2. Should I really create a new `Target` for this use case when engine already has one set?
3. I used `BaseTunables` for the base implementation and sticked with `Tunables` for the trait name, because it makes most sense to me. However, in `lib/api/src/tunables.rs` it is done the other way round. Any thoughts on the naming issue?
4. The import collision between `wamer::Memory` and `wasmer_vm::Memory` is inconvenient. Can it be avoided or do I do it correctly here?

# Review

### DIFF
--- a/examples/tunables_limit_memory.rs
+++ b/examples/tunables_limit_memory.rs
@@ -1,12 +1,15 @@
+use std::ptr::NonNull;
 use std::sync::Arc;
+
 use wasmer::{
-    imports, wat2wasm, Instance, Memory, MemoryError, MemoryType, Module, Pages, Store, TableType,
-    Target, Tunables as BaseTunables,
+    imports,
+    vm::{self, MemoryError, MemoryStyle, TableStyle, VMMemoryDefinition, VMTableDefinition},
+    wat2wasm, Instance, Memory, MemoryType, Module, Pages, Store, TableType, Target,
+    Tunables as BaseTunables,
 };
 use wasmer_compiler_cranelift::Cranelift;
 use wasmer_engine::Tunables;
 use wasmer_engine_jit::JIT;
-use wasmer_vm::{Memory as MemoryTrait, MemoryStyle, Table, TableStyle};
 
 /// A custom tunables that allows you to set a memory limit
 struct LimitingTunables {
@@ -25,6 +28,28 @@ impl LimitingTunables {
             base: BaseTunables::for_target(target),
         }
     }
+
+    /// Takes in input type as requested by the guest and returns a
+    /// limited memory type that is ready for processing.
+    fn adjust_memory(&self, requested: &MemoryType) -> Result<MemoryType, MemoryError> {
+        if requested.minimum > self.max_memory {
+            return Err(MemoryError::Generic(
+                "Minimum of requested memory exceeds the allowed memory limit".to_string(),
+            ));
+        }
+
+        if let Some(max) = requested.maximum {
+            if max > self.max_memory {
+                return Err(MemoryError::Generic(
+                    "Maximum of requested memory exceeds the allowed memory limit".to_string(),
+                ));
+            }
+        }
+
+        let mut adjusted = requested.clone();
+        adjusted.maximum = Some(self.max_memory);
+        Ok(adjusted)
+    }
 }
 
 impl Tunables for LimitingTunables {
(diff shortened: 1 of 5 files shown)


examples/tunables_limit_memory.rs
```diff
@@ -0,0 +1,131 @@
+use std::sync::Arc;
+use wasmer::{
+    imports, wat2wasm, Instance, Memory, MemoryError, MemoryType, Module, Pages, Store, TableType,
+    Tunables as BaseTunables,
+};
+use wasmer_compiler::Target;
+use wasmer_compiler_cranelift::Cranelift;
+use wasmer_engine::Tunables;
+use wasmer_engine_jit::JIT;
+use wasmer_vm::{Memory as MemoryTrait, MemoryStyle, Table, TableStyle};
```
syrusakbary, line 10: For the example, I think it would be better to get some of the structs directly from Wasmer (so people don't depend on other subpackages).

```suggestion
use wasmer::{
    imports, wat2wasm, Instance, Memory, MemoryError, MemoryType, Module, Pages, Store, TableType,
    Tunables, WasmerTunables, Target, Memory, RuntimeMemory, MemoryStyle, RuntimeTable, TableStyle
};
use wasmer_compiler_cranelift::Cranelift;
use wasmer_engine_jit::JIT;
```

This will imply exposing the following structs in the Wasmer API (`lib/api/src/lib.rs`):

```rust
pub use crate::tunables::Tunables as WasmerTunables;

pub use wasmer_engine::Tunables;
pub use wasmer_vm::{Memory as RunnableMemory, MemoryStyle, Table as RunnableTable, TableStyle};
```

-------

Note: other alternative would be exposing the subcrates as submodules off the API:

```suggestion
use wasmer::{
    imports, wat2wasm, Instance, Memory, MemoryError, MemoryType, Module, Pages, Store, TableType,
    Target, Tunables as BaseTunables,
    engine::Tunables,
    vm::{Memory as MemoryTrait, MemoryStyle, Table, TableStyle};
};
use wasmer_compiler_cranelift::Cranelift;
use wasmer_engine_jit::JIT;
```

```rust
pub mod engine {
    pub use wasmer_engine::Tunables;
}
pub mod vm {
    pub use wasmer_vm::{Memory, MemoryStyle, Table, TableStyle};
}
```

Thoughts @MarkMcCaskey ?

syrusakbary (commented):
Thanks a lot for this PR, it pushes things on the right direction.

I just ping'd @MarkMcCaskey so we can see how to handle properly the imports, but other than that, it looks great... good work!

examples/tunables_limit_memory.rs
MarkMcCaskey: Yeah I can see either way, for things that conflict in name like `Memory` namespacing can be a nice way to deal with it, I think expecting people to know about `use x as y` syntax in Rust is probably not a good assumption (though maybe the error message you get tells you how to fix it), so I'd be more in favor of the public submodule idea than exporting 2 things with the same name at the same scope level.

examples/tunables_limit_memory.rs
```diff
@@ -0,0 +1,131 @@
...
+            ));
+        }
+
+        if let Some(max) = requested.maximum {
+            if max > self.max_memory {
+                return Err(MemoryError::Generic(
+                    "Maximum of requested memory exceeds the allowed memory limit".to_string(),
+                ));
+            }
+        }
```
MarkMcCaskey, line 66: One thing, if we expect users to be able to handle these errors programmatically, we should use semantic error values rather than a String via `MemoryError::Generic` if this error is something fatal then a String is fine

MarkMcCaskey: Oh, I just realized that this is an example 😆 , this is neat!

The same still holds though, we may want to add error values for these if we want users to be able to handle them better.  We can adjust this this after shipping this example though, so let's ship it as-is for now and then follow up with possibly adding error values that are easier to match on.

webmaster128: I'm happy wither way. I could create the error types, but then again the whole concept of memory limits is a customization and the errors also belong in the customization. So maybe the question is whether or not the error type must be customizable in `Result<Arc<dyn MemoryTrait>, MemoryError>`.

examples/tunables_limit_memory.rs
webmaster128: > I think expecting people to know about use x as y syntax in Rust is probably not a good assumption (though maybe the error message you get tells you how to fix it)

Agreed. However, in my case the bigger problem was very very confusing error messages about types that are `Sized` or size not known at compile time when I had `Memory` (trait) importet and then used `Vec<Memory>` in the code, which referes to the exported `Memory` (struct). Took a while to realize there are more than one of those in Wasmer.

webmaster128:
I can update this PR at any time. Just waiting for some clear guidance on the Wasmer exports, especially how to deal with `vm` and the collisions. This I cannot decide really.

MarkMcCaskey:
@webmaster128 thanks for the ping!  I think for now let's go with exporting a sub-module like `vm` from `wasmer`, so you can use `use wasmer::vm::Memory as _;` to get the trait in scope or use it directly as `vm::Memory`.

I personally think it's probably also a good idea to rename the `Memory` trait in `vm` to something unambiguous as well, but we can do that later -- the wasmer sub-module should be a reasonable way to unblock this!

webmaster128:
Alright, thanks. I now created `wasmer::vm`, such that I could remove `wasmer-vm` from the root Cargo.toml again. This is also updated to the changes in the Tunables trait.

The example got a little fancier as the base Tunables is now generic, such that you can use `LimitingTunables` on top of any Tunables implementation.

lib/api/src/lib.rs
```diff
@@ -81,7 +81,17 @@ pub use wasmer_types::{
     Atomically, Bytes, GlobalInit, LocalFunctionIndex, MemoryView, Pages, ValueType,
     WASM_MAX_PAGES, WASM_MIN_PAGES, WASM_PAGE_SIZE,
 };
+
+// TODO: should those be moved into wasmer::vm as well?
```
MarkMcCaskey, line 85: Perhaps, but they should probably stay at the top level too!

MarkMcCaskey (approved):
Looks good to me! I'll ping Syrus too

webmaster128:
Hope you don't mind I linked this to close #1588. Let me know if I am missing something.

lib/api/src/lib.rs
webmaster128: okay, feel free to just kick out the comment when you are happy with the setup

syrusakbary (approved):

syrusakbary:
Looks great. Merging.

bors r+

PR #1731, opened 2020-10-16
For store instructions, always load first.

nlewycky:
Sometimes the store may be partially-successful before trapping, for instance if it is partway in valid memory and partway into the guard page.

Use a load instruction before the store to ensure that all the memory is addressable. The loaded value is discarded.

NB. We don't apply this to atomics. It's not clear whether atomic stores can be half-committed.

Fixes align.wast and memory_trap.wast on aarch64.


syrusakbary (commented):
Shall we re-enable this tests on aarch64?

nlewycky:
> Shall we re-enable this tests on aarch64?

They aren't disabled. They're just failing.

The cranelift ones are disabled, but we simply never turned on testing with llvm on arm yet.

syrusakbary (approved):

PR #1738, opened 2020-10-19
Begin work on fixing up vec_uninit code in wasm_c_api

MarkMcCaskey:
PR intentionally unfinished. Before continuing, I want to discuss the fix with @Hywan . The solution in this PR is more correct than what we had before but it has some major problems, namely that users from the C API can't safely free an uninitialized vec due to the way uninit is being used here.

I think the most correct solution that we can have here is to avoid using higher level Rust types and probably alloc and dealloc slabs of memory directly, not involving `Drop` at all... this may have other side effects though.  Some types in vecs now have `Arc` to keep `Instance` alive, avoiding `Drop` may cause problems if the vec itself owns these types.

# Review

### DIFF
--- a/lib/c-api/src/wasm_c_api/externals/mod.rs
+++ b/lib/c-api/src/wasm_c_api/externals/mod.rs
@@ -3,6 +3,7 @@ mod global;
 mod memory;
 mod table;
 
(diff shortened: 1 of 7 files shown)


MarkMcCaskey:
I think this is outdated now, going to close it

PR #1739, opened 2020-10-20
Add host env via trait with proc macro

MarkMcCaskey:
This PR includes a trait-based host-env init and deinit flow (including the plumbing to call these functions) as well as a procedural macro to generate this code.

For example,
```Rust
/// The environment provided to the WASI imports.
#[derive(Debug, Clone, WasmerEnv)]
pub struct WasiEnv {
    state: Arc<Mutex<WasiState>>,
    #[wasmer(export)]
    memory: LazyInit<Memory>,
    #[wasmer(export(name = "real_name"))]
    sum: LazyInit<NativeFunc<(i32, i32), i32>>,
}
```

will generate code roughly looking like:

```Rust
impl ::wasmer::WasmerEnv for WasiEnv {
    fn finish(&mut self, instance: &::wasmer::Instance) {
        let memory: Memory = instance.exports.get_with_generics("memory").unwrap();
        self.memory.initialize(memory);
        let sum: NativeFunc<(i32, i32), i32> = instance.exports.get_with_generics("real_name").unwrap();
        self.sum.initialize(sum);
    }
}
impl WasiEnv {
    pub fn memory_ref(&self) -> Option<&Memory> {
        self.memory.get()
    }
    pub fn sum_ref(&self) -> Option<&NativeFunc<(i32, i32), i32>> {
        self.sum.get()
    }
    pub unsafe fn memory_ref_unchecked(&self) -> &Memory {
        self.memory.get_unchecked()
    }
    pub unsafe fn sum_ref_unchecked(&self) -> &NativeFunc<(i32, i32), i32> {
        self.sum.get_unchecked()
    }
}
```

The exact details of the interface are subject to change and should be reviewed thoroughly, for example we will want to be able to handle errors properly instead of `unwrap`ing and assuming they don't happen.

note to self: The generated helper functions can be used in an unsound way if called before we instantiate, but it'd be nice to not have to use unsafe or deal with `Option` when being called from a host function...

# Review

- [x] Review lazy init abstraction
  - [x] Review its thread-safety
  - [x] Review its interface
- [x] Review the `WasmerEnv` trait
  - [x] Review the interface
  - [x] Review the plumbing (function pointer casting, data flow, requirement that all types implement `WasmerEnv`)
- [ ] Review the proc macro
  - [ ] Review the implementation, likely simpler ways to do certain things but proc macros are kind of confusing
  - [x] Review test coverage
  - [x] Review test coverage of unhappy paths / error messages, etc


MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

nlewycky (commented):
Could you break the vmctx -> union part to its own PR?

examples/wasi.rs
```diff
@@ -51,9 +51,6 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let import_object = wasi_env.import_object(&module)?;
     let instance = Instance::new(&module, &import_object)?;
 
-    // WASI requires to explicitly set the memory for the `WasiEnv`
-    wasi_env.set_memory(instance.exports.get_memory("memory")?.clone());
```
Hywan, line 55: I'm glad to see this going away.

lib/api/src/externals/function.rs
```diff
@@ -124,7 +129,7 @@ impl Function {
     pub fn new_with_env<F, Env>(store: &Store, ty: &FunctionType, env: Env, func: F) -> Self
     where
         F: Fn(&mut Env, &[Val]) -> Result<Vec<Val>, RuntimeError> + 'static,
-        Env: Sized + 'static,
+        Env: Sized + crate::WasmerEnv + 'static,
```
Hywan, line 132: Can't we `use wasmer::WasmerEnv` and replace the code as such:

```suggestion
        Env: Sized + WasmerEnv + 'static,
```

Thanks :-).
```

File: lib/api/src/lib.rs [hunk 3]
```diff
@@ -118,3 +120,118 @@ pub use wasmer_engine_native::{Native, NativeArtifact, NativeEngine};
 
 /// Version number of this crate.
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+// --------------------------------------------------------------
+// TODO: put this in a proper location, just prototyping for now:
+// TODO: rename everything, all names are throw-away names
```
Hywan, line 126: I'm glad to read this :-D.

lib/api/src/lib.rs
```diff
@@ -118,3 +120,118 @@ pub use wasmer_engine_native::{Native, NativeArtifact, NativeEngine};
...
+
+/// Prototype trait for finishing envs.
+pub trait WasmerEnv {
+    /// The function that Wasmer will call on your type to let it finish
+    /// instantiating.
```
Hywan, line 131: We must explain the workflow here. When the “instantation finishes”? Is it really after the Wasmer instantiation, or before something else? What actions the user can do? What can the user change? An example of a `finish` function would be helpful, at least to explain why it is required or necessary?

lib/api/src/lib.rs
```diff
@@ -118,3 +120,118 @@ pub use wasmer_engine_native::{Native, NativeArtifact, NativeEngine};
...
+    pub fn initialize(&mut self, value: T) -> bool {
+        if self.initialized {
+            return false;
+        }
+        unsafe {
+            self.data.as_mut_ptr().write(value);
+        }
+        self.initialized = true;
+        true
+    }
```
Hywan, line 191: Excellent.

lib/api/src/lib.rs
```diff
@@ -118,3 +120,118 @@ pub use wasmer_engine_native::{Native, NativeArtifact, NativeEngine};
...
+                initialized: true,
+            }
+        } else {
+            Self {
+                data: std::mem::MaybeUninit::uninit(),
+                initialized: false,
+            }
+        }
+    }
+}
```
Hywan, line 216: Why do we want `InitAfterInstance` to be clonable? Is it to preserve the behavior of `T` in this case (because `T` is clonable, then `InitAfterInstance<T>` is also clonable)?

lib/api/src/lib.rs
```diff
@@ -118,3 +120,118 @@ pub use wasmer_engine_native::{Native, NativeArtifact, NativeEngine};
...
+
+impl<T> Default for InitAfterInstance<T> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+unsafe impl<T: Send> Send for InitAfterInstance<T> {}
```
Hywan, line 235: Why `InitAfterInstance` must implement `Send`? Is it to preserve the behavior of `T` in this case?

lib/derive/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+fn impl_wasmer_env_for_struct(
+    name: &Ident,
+    data: &DataStruct,
+    _attrs: &[Attribute],
+) -> TokenStream {
+    let (trait_methods, helper_methods) = derive_struct_fields(data);
+    quote! {
+        impl ::wasmer::WasmerEnv for #name {
```
Hywan, line 27: How do manage lifetime for `struct #name`? Like `struct<'a> S<'a> { … }`? It is OK to ignore this for the moment, but it can become really complex.

lib/engine/src/artifact.rs
```diff
@@ -103,6 +103,20 @@ pub trait Artifact: Send + Sync + Upcastable {
...
+
+        let mut thunks = vec![];
+        // ------------
+        for (func_init, func) in imports
+            .host_function_env_initializers
+            .values()
+            .cloned()
+            .zip(imports.functions.values())
+        {
+            let host_env = func.vmctx as _;
```
Hywan, line 115: Hmm I predict complex bug with this line in the future 🤔.

lib/vm/src/export.rs
```diff
@@ -31,6 +31,8 @@ pub struct ExportFunction {
     pub address: *const VMFunctionBody,
     /// Pointer to the containing `VMContext`.
     pub vmctx: *mut VMContext,
+    /// temp code to set vmctx for host functions
+    pub function_ptr: Option<fn(*mut std::ffi::c_void, *const std::ffi::c_void)>,
```
Hywan, line 35: I don't understand what it does exactly, and why the name `function_ptr`? Can you explain please?

lib/vm/src/vmcontext.rs
```diff
@@ -22,7 +22,7 @@ pub struct VMFunctionImport {
     /// A pointer to the imported function body.
     pub body: *const VMFunctionBody,
 
-    /// A pointer to the `VMContext` that owns the function.
+    /// A pointer to the `VMContext` that owns the function or host data.
```
Hywan, line 25: Can't we use an `enum` to differentiate between the two? In the pre-1.0 era, we used `vmctx` as an opaque pointer to hold various kinds of values, and it was really complex to debug and clearly error-prone. I don't want to reproduce this here.

lib/api/src/lib.rs
MarkMcCaskey: Hmm, on second thought I'm not exactly sure why I felt the need to add this, I think it's just because `WasiEnv` was clone and it let me keep `derive(Clone)` on it

lib/api/src/lib.rs
MarkMcCaskey: This was also for `WasiEnv` but more largely because the way Wasmer works, we can pass functions to other threads and call it meaning that it's `Send`... perhaps it can also be used in a `Sync` way but I didn't think so...

Someone filed an issue for us to review this and I agree, that we should probably audit what we actually allow users to do

lib/derive/src/lib.rs
MarkMcCaskey: We don't at all, I'm hoping to avoid that.

I posted on Slack about wanting to remove the lifetime from `NativeFunc` and friends because I don't think we're actually using it in practice and it is somewhat of a blocker to making this work

lib/engine/src/artifact.rs
MarkMcCaskey: Yeah in #1753 I'm trying to make use of `vmctx` more explicit so that users of the code are more careful with it...

Any suggestions for this use ?

lib/vm/src/export.rs
MarkMcCaskey: This is `WasmerEnv::finish` for the env used by this function, due to our heirarchy of types, we can't be explicit about the type signature because this is in `vm` and it takes `&api::Instance` as an argument.

It's really not ideal but I haven't come up with a better way to get access to the initializer function yet. This is the thing I was hoping to get the most feedback on early

lib/vm/src/vmcontext.rs
MarkMcCaskey: To clarify, this PR does nothing new in regards to vmctx, it simply documents what we've been doing from the start.

I agree that I'd also like to avoid it.  #1753 at least makes the behavior explicit (it's really bad that it's not even obvious what it is....).

I don't really see a reasonable alternative given our code right now, this is low level code that our generated code interacts with.  But this double use of a pointer is a high risk thing to do from a security perspective and I don't feel like it's simple enough to audit our code even with #1753.  We'd have to change the way generated code works I think

syrusakbary (commented):
The PR I think it moves to the good direction.

Feedback:
1. Test the wasm-c-api `wasm_func_new_with_env` with a custom finalizer
2. Test the behavior of a creating a custom environment. That the finalizer and free methods get called.

lib/vm/src/export.rs
syrusakbary: I'm unclear on why we need this here. The env finisher should live in the InstanceHandle, right?

lib/vm/src/instance.rs
```diff
@@ -296,13 +315,18 @@ impl Instance {
...
             ExportIndex::Function(index) => {
                 let sig_index = &self.module.functions[*index];
-                let (address, vmctx) = if let Some(def_index) = self.module.local_func_index(*index)
-                {
-                    (self.functions[def_index].0 as *const _, self.vmctx_ptr())
-                } else {
-                    let import = self.imported_function(*index);
-                    (import.body, import.vmctx)
-                };
+                let (address, vmctx, function_ptr) =
```
syrusakbary, line 318: This needs to be revisited. I'm not sure why we need `function_ptr` in an exported function from the InstanceHandle

lib/api/src/lib.rs
syrusakbary: Probably we don't need WasiEnv to be clonable after this PR?

lib/api/src/exports.rs
```diff
@@ -150,6 +150,19 @@ impl Exports {
             .map_err(|_| ExportError::IncompatibleType)
     }
 
+    /// Hack to get this working with nativefunc too
+    pub fn get_with_generics<'a, T, Args, Rets>(&'a self, name: &str) -> Result<T, ExportError>
```
syrusakbary, line 154: `get_native_function_with_generics` ?

lib/vm/src/export.rs
MarkMcCaskey: Good question!

This is how the function pointers get into the `InstanceHandle`: we lose type information, but `wasmer::externals::function::Function` methods have access to the initializer, we have to store it somewhere for later and this is the best place I've found.

The little code window above that Ivan commented on is showing out of date code, but I left a comment there asking for feedback if anyone knows a better way to get this information to where we need it.

syrusakbary: Why don't store it in `InstanceHandle` ?

lib/api/src/exports.rs
MarkMcCaskey: This lets us use the `derive(WasmerEnv)` with `NativeFunc`

The issue is that `instance.exports.get()` does not work for `NativeFunc` because `NativeFunc` has generic arguments that need to be known.  This is an alternative `get` function that _does_ work with `NativeFunc` because it takes generics as arguments.

This way we have a single entrypoint that works for all the types we support: we're offloading the work of deciding what logic to run to the typesystem / traits.  Without this we can't handle `NativeFunc` without special casing it in the macro (that is, we have to have users explicitly call it out for us because we need to decide which type of code to generate. This is what I had before with the `#[wasmer(export(native_func("hello")))]` syntax.)

lib/vm/src/export.rs
MarkMcCaskey: We do, but we have to get that information to the `InstanceHandle`.

The data flow starts when a user makes a new `Function`:
1. [Function::new_with_env](https://github.com/wasmerio/wasmer/blob/feature/host-env-prototype/lib/api/src/externals/function.rs#L130-L161) : by way of the trait constraint, we can get the initializer and turn it into a function pointer.  We then put it in `ExportFunction` because it seemed like the most obvious place to put it.

2. We need this data when instantiating [Instance::new](https://github.com/wasmerio/wasmer/blob/feature/host-env-prototype/lib/api/src/instance.rs#L73-L76), [Module::instantiate](https://github.com/wasmerio/wasmer/blob/feature/host-env-prototype/lib/api/src/module.rs#L261-L279) , we only get a module and a resolver (the resolver is where all the functions are living)

3. We end up in [engine::Artifact::instantiate](https://github.com/wasmerio/wasmer/blob/feature/host-env-prototype/lib/engine/src/artifact.rs#L89-L158) here we extract these values from the `Imports` during instantiation and make it part of the `InstanceHandle`

4. Later we call a function on `InstanceHandle` to do something with this stored data

lib/api/src/exports.rs
syrusakbary: I meant that we should rename it to `get_native_function_with_generics` :P

MarkMcCaskey: Oh, we can but we're calling it for all types not just `NativeFunc`, it's `Exportable` / `Instance::get` + NativeFunc support, it works for `Memory`, `Table`, and everything `Exportable` too, which is why it's useful!

webmaster128:
Thank you for the great work here. If you want this to be tested, feel free to ping me. We use access both memory and exported functions via Env in import implementations. Right now we're doing it the manual way.

lib/api/tests/externals.rs
```diff
@@ -377,3 +377,33 @@ fn function_outlives_instance() -> Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn manually_call_function() -> Result<()> {
```
syrusakbary, line 382: Awesome! This is a great test, although the name of the test is not super clear.

Can we name it `manually_generate_wasmer_env` instead ?

lib/derive/src/lib.rs
MarkMcCaskey: I added lifetime support to the proc macro a bit ago, I think things should hopefully "just work"

lib/api/src/env.rs
```diff
@@ -0,0 +1,202 @@
...
+///     }
+/// }
+/// ```
+pub trait WasmerEnv {
+    /// The function that Wasmer will call on your type to let it finish
+    /// setting up the environment with data from the `Instance`.
+    ///
+    /// This function is called after `Instance` is created but before it is
+    /// returned to the user via `Instance::new`.
+    fn finish(&mut self, instance: &Instance) -> Result<(), HostEnvInitError>;
```
syrusakbary, line 63: I don't really like the name `finish`. `setup_instance`, `init_from_instance` or other names might be better. Other ideas?

syrusakbary: Also, why not have a default implementation that returns `Ok(())`?

That will simplify things I think.

MarkMcCaskey: > I don't really like the name finish. setup_instance, init_from_instance or other names might be better. Other ideas?

Yeah, I agree that it probably needs a better name.

Of those three, the one I like the most is `init_from_instance`.  I'll think about it for a bit 🤔

lib/api/src/externals/function.rs
```diff
@@ -126,7 +129,7 @@ impl Function {
     pub fn new_with_env<F, Env>(store: &Store, ty: &FunctionType, env: Env, func: F) -> Self
     where
         F: Fn(&mut Env, &[Val]) -> Result<Vec<Val>, RuntimeError> + 'static,
-        Env: Sized + 'static,
+        Env: Sized + WasmerEnv + 'static,
```
syrusakbary, line 132: Can we have here Env as `dyn Any`, and then a check on WasmerEnv at runtime?

MarkMcCaskey: I believe it's not possible to downcast from one trait object into another: it requires downcasting into a concrete type.  Also I think it's just a limitation of the trait system, but there aren't any workarounds for this either.  What I mean is that you can'1t implement a trait on all types but have specializations on certain types (this would let us implement `WasmerEnv` on everything, or another trait which would let us know if `WasmerEnv` was implemented).

syrusakbary:
What will happen if we provide a non `LazyInit` field?

```rust
/// The environment provided to the WASI imports.
#[derive(Debug, Clone, WasmerEnv)]
pub struct MyEnv {
    #[wasmer(export)]
    memory: Memory, // It should be LazyInit<Memory>
}
```

Will a proper error showed? (Kind of: `please wrap your type with `LazyInit`)

And in the case is not an export?

```rust
/// The environment provided to the WASI imports.
#[derive(Debug, Clone, WasmerEnv)]
pub struct MyEnv {
    #[wasmer(export)]
    memory: usize, // It should error saying that usize is not a possible Wasm Export type
}
```

MarkMcCaskey:
> What will happen if we provide a non `LazyInit` field?
> 
> ```rust
> /// The environment provided to the WASI imports.
> #[derive(Debug, Clone, WasmerEnv)]
> pub struct MyEnv {
>     #[wasmer(export)]
>     memory: Memory, // It should be LazyInit<Memory>
> }
> ```
> 
> Will a proper error showed? (Kind of: `please wrap your type with "LazyInit"`)
> 
> And in the case is not an export?
> 
> ```rust
> /// The environment provided to the WASI imports.
> #[derive(Debug, Clone, WasmerEnv)]
> pub struct MyEnv {
>     #[wasmer(export)]
>     memory: usize, // It should error saying that usize is not a possible Wasm Export type
> }
> ```

We have a test for that! https://github.com/wasmerio/wasmer/blob/7131caa8d9ee806dd2fd7df4ab5e77365c8d4ff0/lib/derive/tests/compile-fail/no-lazy-init.rs

The error message for that line is ```WasmerEnv derive expects all `exports` to be wrapped in `LazyInit````; we don't run these tests in CI because they're really, really slow for annoying reasons. It requires a full recompilation of Wasmer from scratch otherwise it fails to build 🤷

lib/vm/src/export.rs
```diff
@@ -31,6 +31,16 @@ pub struct ExportFunction {
...
+    /// Function pointer to `WasmerEnv::finish(&mut self, instance: &Instance)`.
+    ///
+    /// This function is called to finish setting up the environment after
+    /// we create the `api::Instance`.
+    // META: if you have a better idea of how to get this function to where it
+    // needs to be, please let me know.
+    pub function_ptr: Option<
+        // TODO: review the return value here...
+        fn(*mut std::ffi::c_void, *const std::ffi::c_void) -> Result<(), *mut std::ffi::c_void>,
+    >,
```
syrusakbary, line 43: I don't think `function_ptr` should live in the `ExportFunction`. Perhaps we need a better abstraction in the engine itself.
I don't think it should leak to the VM (other than when we init the instance)

MarkMcCaskey: Well, the problem is that we're getting it from a `Resolver` (specifically the import object);

Following the types we get:

```Rust
    pub fn ImportObject::get_export(&self, module: &str, name: &str) -> Option<Export>
    fn NameResolver::resolve_by_name(&self, module: &str, name: &str) -> Option<Export>
```

```Rust
/// The value of an export passed from one instance to another.
#[derive(Debug, Clone)]
pub enum Export {
    /// A function export value.
    Function(ExportFunction),

    /// A table export value.
    Table(ExportTable),

    /// A memory export value.
    Memory(ExportMemory),

    /// A global export value.
    Global(ExportGlobal),
}
```

So I'm not sure where else we could put it if this is how we're getting the functions

lib/api/src/instance.rs
```diff
@@ -85,11 +124,19 @@ impl Instance {
...
             module: module.clone(),
             exports,
-        })
+        };
+
+        unsafe {
+            instance
+                .handle
+                .initialize_host_envs::<HostEnvInitError>(&instance as *const _ as *const _)?;
+        }
```
syrusakbary, line 137: TODO: Syrus, check this deeply, research alternatives if any

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -92,16 +92,40 @@ pub unsafe extern "C" fn wasm_func_new_with_env(
...
+    }
+
+    impl wasmer::WasmerEnv for WrapperEnv {
+        fn init_with_instance(
+            &mut self,
+            _instance: &wasmer::Instance,
+        ) -> Result<(), wasmer::HostEnvInitError> {
+            Ok(())
+        }
+    }
```
syrusakbary, line 125: This should not be needed any longer.

lib/api/src/externals/function.rs
```diff
@@ -140,6 +143,12 @@ impl Function {
         let vmctx = VMFunctionEnvironment {
             host_env: Box::into_raw(Box::new(dynamic_ctx)) as *mut _,
         };
+        // TODO: look into removing transmute by changing API type signatures
```
syrusakbary, line 146: How we can change the API?

MarkMcCaskey: There's no good way to do this that I can think of yet. The only way to avoid the transmutes would be to to make the `WasmerEnv::init_with_instance` function take `void*` directly... but that seems like a bad trade off.

lib/vm/src/export.rs
syrusakbary: Yeah, perhaps we need two separate types. `Export` and `VMExport` (`VMExport` will be what now `Export` is)

syrusakbary: For example, `ExportFunction` now takes:
```
    /// The function kind (specifies the calling convention for the function).
    pub kind: VMFunctionKind,
    /// May be None when the function is a host-function (FunctionType == Dynamic or vmctx == nullptr).
```

But those fields are really only needed for the engine

lib/api/src/exports.rs
syrusakbary: In the `get` function we are using `Exportable<'a>`. However in this function decl we are using `Args` and `Rets` which kind of implies a function? (or that's why I assumed it was for a function).

Why would we want this function working for `Memory` and `Table`?

MarkMcCaskey: The reason is that we now have a single function by which can talk about anything that can come from an `Instance`, whether it has `Args` and `Rets` or not.

It was created to solve the problem of the procedural macro not knowing anything about what was tagged with `#[export]`.  Before that if we wanted to get `NativeFunc` with it's `Args` and `Rets` we had to call it out explicitly in the macro: `#[export(native_func)]`.  The new API lets us not care what it is because we have a single interface by which we can get anything.

syrusakbary: Ok, got it.

syrusakbary: Should we leave this API public only for the crate? (but not for outside)
This might prevent unintended usage for now

lib/api/src/externals/function.rs
```diff
@@ -283,6 +304,11 @@ impl Function {
             host_env: Box::into_raw(box_env) as *mut _,
         };
         let signature = function.ty();
+        // TODO: look into removing transmute by changing API type signatures
+        let function_ptr = Some(std::mem::transmute::<
+            fn(_, _) -> Result<(), _>,
+            fn(_, _) -> Result<(), _>,
+        >(Env::init_with_instance));
```
syrusakbary, line 311: I'm not super sure why this is needed. Can you provide some context @MarkMcCaskey ?

lib/api/src/externals/function.rs
```diff
@@ -291,6 +317,7 @@ impl Function {
                 address,
                 kind: VMFunctionKind::Static,
                 vmctx,
+                function_ptr,
```
syrusakbary, line 320: We should move this out of this struct.

lib/api/src/instance.rs
```diff
@@ -85,11 +124,21 @@ impl Instance {
...
 
-        Ok(Self {
+        let instance = Self {
             handle: Arc::new(Mutex::new(handle)),
             module: module.clone(),
             exports,
-        })
+        };
+
+        unsafe {
```
syrusakbary, line 133: I'm not super fan of this function.
I'm thinking on ways we can do it differently, but haven't got any so far.

Can you add a block comment before the unsafe function commenting what it does, why is unsafe and why is needed to do it after having the instance?

lib/api/src/exports.rs
MarkMcCaskey: Well, it needs to be public to be called by the macro-expanded code: this is the primary interface that the macro will use when it expands

lib/api/src/externals/function.rs
MarkMcCaskey: This is the same as https://github.com/wasmerio/wasmer/pull/1739#discussion_r526446162 ; it may be possible but I'm not exactly sure how yet

lib/api/src/native.rs
```diff
@@ -68,6 +68,8 @@ where
             address: other.address,
             vmctx: other.vmctx,
             signature,
+            // TODO:
+            function_ptr: None,
```
syrusakbary, line 72: I really think this doesn't belong here.

lib/api/src/externals/function.rs
MarkMcCaskey: Which part? The function_ptr at all or the transmute?

We need the function pointer to initialize the data, so far this is the most direct place to put it so we can access where we need it.

We have to transmute it because `WasmerEnv` isn't known to the `vm` so we can't use the real name of the type, I have some work in progress trying to make it so vm can see `WasmerEnv` but it's a bit stuck

lib/c-api/src/deprecated/import/mod.rs
```diff
@@ -634,26 +634,27 @@ pub unsafe extern "C" fn wasmer_import_func_params_arity(
...
-/// struct used to hold on to `UnsafeMutableEnv` pointer as well as the function.
-/// we need to do this to initialize the context ptr inside of `UnsafeMutableEnv` when
+/// struct used to hold on to `LegacyEnv` pointer as well as the function.
+/// we need to do this to initialize the context ptr inside of `LegacyEnv` when
 /// instantiating the module.
 #[derive(Debug)]
 pub(crate) struct FunctionWrapper {
     pub(crate) func: NonNull<Function>,
-    pub(crate) legacy_env: NonNull<UnsafeMutableEnv>,
+    pub(crate) legacy_env: NonNull<LegacyEnv>,
```
syrusakbary, line 657: I actually like more `UnsafeMutableLegacyEnv`. Verbose but it conveys the meaning

lib/c-api/src/deprecated/import/wasi.rs
```diff
@@ -213,11 +213,6 @@ pub unsafe extern "C" fn wasmer_wasi_generate_default_import_object() -> *mut wa
     let mut wasi_state_builder = wasi::WasiState::new("wasmer-wasi-default-program-name");
     let wasi_state = wasi_state_builder.build().unwrap();
     let mut wasi_env = wasi::WasiEnv::new(wasi_state);
-    // this API will now leak a `Memory`
-    let memory_type = MemoryType::new(0, None, false);
-    let memory = Memory::new(store, memory_type).expect("create memory");
-    wasi_env.set_memory(memory);
-    // TODO(mark): review lifetime of `Memory` here
```
syrusakbary, line 220: I have no idea how this worked before. We should pass the memory exported from the instance. The previous code was not doing that :O

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -174,21 +174,28 @@ pub extern "C" fn wasi_env_new(mut config: Box<wasi_config_t>) -> Option<Box<was
...
 
     true
 }
 
+// Dead code: deprecate or remove
 #[no_mangle]
-pub extern "C" fn wasi_env_set_memory(env: &mut wasi_env_t, memory: &wasm_memory_t) {
-    env.inner.set_memory(memory.inner.clone());
+pub extern "C" fn wasi_env_set_memory(_env: &mut wasi_env_t, _memory: &wasm_memory_t) {
+    //env.inner.set_memory(memory.inner.clone());
```
syrusakbary, line 198: Comment why this is no longer needed

lib/wasi/src/lib.rs
```diff
@@ -172,7 +82,7 @@ impl WasiEnv {
 
     /// Get a reference to the memory
     pub fn memory(&self) -> &Memory {
-        self.memory.get_memory().expect("The expected Memory is not attached to the `WasiEnv`. Did you forgot to call wasi_env.set_memory(...)?")
+        self.memory_ref().unwrap()
```
syrusakbary, line 85: Use expect rather than unwrap.

lib/api/src/instance.rs
```diff
@@ -85,11 +124,21 @@ impl Instance {
...
+            instance
+                .handle
+                .lock()
+                .unwrap()
```
syrusakbary, line 137: Use expect rather than unwrap

lib/emscripten/src/env/mod.rs
```diff
@@ -47,15 +47,14 @@ pub fn call_memalign(ctx: &EmEnv, alignment: u32, size: u32) -> u32 {
...
+        .memset_ref()
         .unwrap()
         .call(pointer, value, size)
         .unwrap()
 }
 
-pub(crate) fn get_emscripten_data<'a>(ctx: &'a EmEnv) -> &'a mut EmscriptenData<'static> {
-    unsafe { &mut **ctx.data }
+pub(crate) fn get_emscripten_data(ctx: &EmEnv) -> MutexGuard<EmscriptenData> {
+    ctx.data.lock().unwrap()
```
syrusakbary, line 57: `expect`

lib/c-api/src/deprecated/import/wasi.rs
MarkMcCaskey: I think it actually was, it's just that what this code was doing was being overwritten later.  It was really bad, glad it's being removed

lib/wasi/src/lib.rs
MarkMcCaskey: The idea was that It shouldn't be possible to use this incorrectly anymore, but if done manually then I suppose it could be used incorrectly

lib/api/src/instance.rs
MarkMcCaskey: `unwrap` on `Mutex::lock` isn't worth handling, there's nothing useful we can provide to users here

MarkMcCaskey: To clarify this is only possible if there was a panic on another thread while holding the mutex.  If that happens, we just want to terminate Wasmer as soon as possible

lib/emscripten/src/env/mod.rs
MarkMcCaskey: Same as https://github.com/wasmerio/wasmer/pull/1739#discussion_r529028014

lib/c-api/src/deprecated/import/mod.rs
MarkMcCaskey: So this change is actually fixing an issue from the find and replace done on in another PR.  It should really be split out and merged to master separately, but I accidentally renamed a struct with the same name in another place by doing bulk find and replace, this just corrects that mistake

lib/c-api/src/wasm_c_api/wasi/mod.rs
MarkMcCaskey: Yeah, I was putting this off for wanting to find a deprecation process but I'll add a user visible comment about it being deprecated for now and we can figure out what to do later

syrusakbary (approved):

Issue #1744, opened 2020-10-20
How do traps work in the C API?


liamappelbe:
I assume traps are for exception handling, but the [C API](https://github.com/wasmerio/wasmer/blob/master/lib/c-api/tests/wasm_c_api/wasm-c-api/include/wasm.h) isn't well documented, and I'm not sure how to use them.

I have a wasm module for testing that just imports a function and calls it. I'm filling the import with a Dart function (represented as a `wasm_func_callback_with_env_t`) and I'm able to call back from wasm into Dart successfully. Now I want to handle the case where the Dart function throws an exception.

My current approach is to convert the exception and stack trace into a string, and stick that string in a `wasm_trap_t` using `wasm_trap_new`. `wasm_func_callback_with_env_t` returns a `wasm_trap_t`, so naturally I just return the trap from the callback. The wasm function is being invoked with `wasm_func_call`, which also returns a `wasm_trap_t`, so I'm checking if it returns a non-null trap.

So I assumed the way all this would work is that if the callback returns a non-null trap, then the engine halts execution and immediately returns the trap from `wasm_func_call` (and doesn't execute the rest of the wasm function). I want to get the original Dart exception string from that trap, and throw it again as a Dart exception. But when I look at the return value of `wasm_func_call`, it's always null, regardless of whether `wasm_func_callback_with_env_t` returned a trap or not.

How should I be using traps? Also, why does `wasm_instance_new` take a `wasm_trap_t**`?

Hywan:
Hey :-),

For you, the sun is probably shining,
Which means for me, since hours, it's for sure sleeping.
Midnight is close.
My brain has turned off already.
However here is a short disclose,
As much as possible friendly:
https://github.com/wasmerio/wasmer/blob/master/lib/c-api/tests/wasm_c_api/wasm-c-api/example/trap.c.

I'll provide a longer answer later :-).

Hywan:
Hello @liamappelbe,

[...]

PR #1751, opened 2020-10-22
feat(c-api) Correctly implement “trap” in `wasm_func_new*`

Hywan:
The implementation of “trap” in `wasm_func_new` was “incorrect”. It's
more idiomatic to return a `RuntimeError` than raising it in this
case, so that we don't duplicate locations where runtime errors are
raised.

The implementation of “trap” in `wasm_func_new_with_env` was
missing. This patch implements a similar strategy than the sibling
function.

Fixes https://github.com/wasmerio/wasmer/issues/1744.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - [#1710](https://github.com/wasmerio/wasmer/pull/1710) Memory for function call trampolines is now owned by the Artifact.
 ### Added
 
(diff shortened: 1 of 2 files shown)


Hywan:
Fixes https://github.com/wasmerio/wasmer/issues/1744.

jubianchi (approved):

PR #1752, opened 2020-10-22
Add more examples for the Rust API

jubianchi:
# Description

The goal is to cover most of the Rust API with examples (and thus tests).

Some other PRs already exist for that purpose:
* #1730
* #1687 

# Review

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -181,6 +181,20 @@ example.
 
    </details>
 
+10. [**Exported memory**][exported-memory], explains how to read from 
+    and write to exported memory.
+   
+   _Keywords_: export, memory.
+
+   <details>
+   <summary><em>Execute the example</em></summary>
+
+   ```shell
+   $ cargo run --example exported-memory --release --features "cranelift"
+   ```
+
+   </details>
+
 ### Imports
 
 10. [**Imported global**][imported-global], explains how to work with
@@ -237,6 +251,7 @@ example.
 [cross-compilation]: ./engine_cross_compilation.rs
 [exported-global]: ./exports_global.rs
 [exported-function]: ./exports_function.rs
+[exported-memory]: ./exports_memory.rs
 [imported-global]: ./imports_global.rs
 [imported-function]: ./imports_function.rs
 [wasi]: ./wasi.rs
--- a/examples/README.md
+++ b/examples/README.md
@@ -150,8 +150,23 @@ example.
    </details>
 
 ### Exports
+
+8. [**Exported global**][exported-global], explains how to work with
+   exported globals: get/set their value, have information about their 
+   type.
    
-8. [**Exported function**][exported-function], explains how to get and
+   _Keywords_: export, global.
+
+   <details>
+   <summary><em>Execute the example</em></summary>
+
+   ```shell
+   $ cargo run --example exported-globals --release --features "cranelift"
+   ```
+
+   </details>
+   
+9. [**Exported function**][exported-function], explains how to get and
    how to call an exported function. They come in 2 flavors: dynamic,
    and “static”/native. The pros and cons are discussed briefly.
    
@@ -166,9 +181,25 @@ example.
 
    </details>
 
+### Imports
+
+10. [**Imported global**][imported-global], explains how to work with
+   imported globals: create globals, import them, get/set their value.
+   
+   _Keywords_: import, global.
+
+   <details>
+   <summary><em>Execute the example</em></summary>
+
+   ```shell
+   $ cargo run --example imported-globals --release --features "cranelift"
+   ```
+
+   </details>
+
 ### Integrations
 
-9. [**WASI**][wasi], explains how to use the [WebAssembly System
+11. [**WASI**][wasi], explains how to use the [WebAssembly System
    Interface][WASI] (WASI), i.e. the [`wasmer-wasi`] crate.
    
    _Keywords_: wasi, system, interface
@@ -189,7 +220,9 @@ example.
 [compiler-cranelift]: ./compiler_cranelift.rs
 [compiler-llvm]: ./compiler_llvm.rs
 [cross-compilation]: ./engine_cross_compilation.rs
+[exported-global]: ./exports_global.rs
 [exported-function]: ./exports_function.rs
+[imported-global]: imports_global.rs
 [wasi]: ./wasi.rs
 [`wasmer-compiler-singlepass`]: https://github.com/wasmerio/wasmer/tree/master/lib/compiler-singlepass
 [`wasmer-compiler-cranelift`]: https://github.com/wasmerio/wasmer/tree/master/lib/compiler-cranelift
--- a/examples/README.md
+++ b/examples/README.md
@@ -197,9 +197,24 @@ example.
 
    </details>
 
+11. [**Imported function**][imported-function], explains how to define 
+   an imported function. They come in 2 flavors: dynamic,
+   and “static”/native.
+   
+   _Keywords_: import, function, dynamic, static, native.
+
+   <details>
+   <summary><em>Execute the example</em></summary>
+
+   ```shell
+   $ cargo run --example imported-function --release --features "cranelift"
+   ```
+
+   </details>
+
 ### Integrations
 
-11. [**WASI**][wasi], explains how to use the [WebAssembly System
+12. [**WASI**][wasi], explains how to use the [WebAssembly System
    Interface][WASI] (WASI), i.e. the [`wasmer-wasi`] crate.
    
    _Keywords_: wasi, system, interface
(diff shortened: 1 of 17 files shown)


jubianchi:
bors try

examples/README.md
```diff
@@ -166,9 +181,54 @@ example.
...
+   <details>
+   <summary><em>Execute the example</em></summary>
+
+   ```shell
+   $ cargo run --example imported-globals --release --features "cranelift"
+   ```
+
+   </details>
+
+12. [**Imported function**][imported-function], explains how to define
```
MarkMcCaskey, line 214: I think GitHub markdown will number correctly for us if we just use `1.` everywhere; we can do that in a separate PR though, just putting the idea out there in case I forget

examples/README.md
```diff
@@ -189,7 +249,11 @@ example.
 [compiler-cranelift]: ./compiler_cranelift.rs
 [compiler-llvm]: ./compiler_llvm.rs
 [cross-compilation]: ./engine_cross_compilation.rs
+[exported-global]: exports_global.rs
```
MarkMcCaskey, line 252: The other files start with `./`, I have no idea if that's significant for GitHub markdown links, but might be good to make this consistent

examples/exports_memory.rs
```diff
@@ -0,0 +1,135 @@
...
+    println!("Instantiating module...");
+    // Let's instantiate the Wasm module.
+    let instance = Instance::new(&module, &import_object)?;
+
+    let store = instance
+        .exports
+        .get_native_function::<(i32, i32), ()>("store")?;
+    let load = instance
+        .exports
+        .get_native_function::<(), (i32, i32)>("load")?;
```
MarkMcCaskey, line 64: Consider changing the return type of this to `(WasmPtr<u8, Array>, i32)`

examples/exports_memory.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    // We now know where to fin our string, let's read it.
+    //
+    // We will get bytes out of the memory so we need to
+    // decode them into a string.
+    let iter = view[offset as usize..(offset + length) as usize]
+        .iter()
+        .map(Cell::get);
+    let bytes: Vec<u8> = iter.collect();
+    println!("Memory contents: {:?}", str::from_utf8(&bytes)?);
```
MarkMcCaskey, line 101: Then we can just do
```Rust
let (ptr, length) = load.call()?;
println!("String offset: {:?}", ptr.offset());
println!("String length: {:?}", length);

let wasm_str = ptr.get_utf8_string(memory, length).unwrap();
println!("Memory contents: {}", wasm_str); 
```

https://docs.rs/wasmer/1.0.0-alpha4/wasmer/struct.WasmPtr.html#method.get_utf8_string

examples/exports_memory.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    // What about changing part of the contents of the memory
+    // with a more appropriate string?
+    //
+    // The Wasm module exports a `store` function which lets us
+    // write to the memory. Let's us this function to change the
+    // string.
+    for (index, byte) in "Wasmer!".as_bytes().iter().enumerate() {
+        store.call(offset + 7 + index as i32, *byte as i32)?;
+    }
```
MarkMcCaskey, line 111: We also want to do bulk storage, we can get a pointer to memory on the host and just write directly to it.

examples/exports_memory.rs
```diff
@@ -0,0 +1,135 @@
...
+    // query the offset and length again.
+    let (offset, length) = load.call()?;
+    println!("String offset: {:?}", offset);
+    println!("String length: {:?}", length);
+
+    let iter = view[offset as usize..(offset + length) as usize]
+        .iter()
+        .map(Cell::get);
+    let bytes: Vec<u8> = iter.collect();
+    println!("New memory contents: {:?}", str::from_utf8(&bytes)?);
```
MarkMcCaskey, line 125: This code can also use `WasmPtr`

MarkMcCaskey (commented):
Looks good! I appreciate your attention to detail with documenting the errors and examples generally in the doc comments!

I just have a few suggestions to make some things a bit simpler!

examples/README.md
Hywan: Didn't know that. I'm trying:

```md
1. Hello
1. World

ozkedokze

3. Oh
1. Does it
1. Work?
```

1. Hello
1. World

ozkedokze

3. Oh
1. Does it
1. Work?

lib/api/src/lib.rs
```diff
@@ -1,4 +1,8 @@
 //! Wasmer API
+#![doc(
+    html_logo_url = "https://avatars3.githubusercontent.com/u/44205449?s=200&v=4",
```
Hywan, line 3:
```suggestion
    html_logo_url = "https://github.com/wasmerio.png?size=200",
```

That's a pro-tip, but I think it's better to use that URL to save us stupid maintenance in the future. Thoughts?

examples/exports_memory.rs
jubianchi: @Mark I did something here but your feedback would be much appreciated.

examples/exports_memory.rs
```diff
@@ -0,0 +1,123 @@
...
+    // To do that, we'll get a pointer to a `MemoryView`, set its
+    // offset to the same as the string and write to it.
+    let view = memory.view::<u8>();
+    let source = "Hello, Wasmer!";
+
+    unsafe {
+        let dest = view.as_ptr().add(ptr.offset() as usize) as *mut u8;
+
+        std::ptr::copy(source.as_ptr(), dest, source.len());
+    }
```
MarkMcCaskey, line 103: This is a reasonable implementation! And in fact I think it'd be a good idea for us to expose a `memcpy` like function(s) on `WasmPtr` for users, but I think ideally we'd want to avoid using `unsafe` when showing our API, so perhaps we can go with a safe solution and file an issue to make a `memcpy` helper and clean this up.

Our API is actually very subtle and a bit odd here but because we're returning things wrapped in `Cell`, we can just mutate it.

https://docs.rs/wasmer/1.0.0-alpha4/wasmer/struct.WasmPtr.html#method.deref-1

So we can probably do something like

```Rust
let values = ptr.deref(memory, 0, source.len()).unwrap();
for (i, c) in source.bytes().enumerate() {
    values[i].set(*c);
}
```

And this should not be significantly slower than  memcpy, `Cell` is a compile-time abstraction and just tells Rust that "interior mutability" is allowed.  The only way this can be slower than memcpy is if bounds checks can't be eliminated or it prevents some other optimization for some reason.

But in general I think this is probably the best way to do it without using `unsafe` right now

MarkMcCaskey (approved):
Looks good overall to me!  I think we should probably stick to safe Rust for our examples in most cases : needing to use `unsafe` for basic use cases is a good signal that our API is not sufficient! So I think we should use that to improve our API

examples/exports_memory.rs
```diff
@@ -0,0 +1,123 @@
...
+    let str = ptr.get_utf8_string(memory, length as u32).unwrap();
+    println!("Memory contents: {:?}", str);
+
+    // What about changing the contents of the memory with a more
+    // appropriate string?
+    //
+    // To do that, we'll get a pointer to a `MemoryView`, set its
+    // offset to the same as the string and write to it.
+    let view = memory.view::<u8>();
+    let source = "Hello, Wasmer!";
```
MarkMcCaskey, line 97: One other thing, this may be a bit too advanced to want to include here but if you're using a string as purely binary data, you can use the `b"this is a string"` syntax to get a raw byte string which is of type `&[u8]` instead of `&str` (which is a UTF-8 validated string), this can sometimes simplify code when you really want to just deal with some literal bytes. (for example, you wouldn't need to call `.bytes()` on the string when iterating it, you can just iterate it)

examples/exports_memory.rs
jubianchi: @mark thanks, I'll switch to this.

Actually, this was my first try but I did think it was not good enough. When I first read your (first) comment, I was thinking you wanted to totally get rid of the loop ;).

MarkMcCaskey (approved):
Looks good to me!

lib/cli/src/lib.rs
```diff
@@ -11,7 +11,7 @@
     unstable_features
 )]
 #![doc(html_favicon_url = "https://wasmer.io/static/icons/favicon.ico")]
-#![doc(html_logo_url = "https://avatars3.githubusercontent.com/u/44205449?s=200&v=4")]
+#![doc(html_logo_url = "https://github.com/wasmerio.png?size=200")]
```
syrusakbary, line 14: Nice trick!

PR #1753, opened 2020-10-22
Use a union for VMContext for functions

MarkMcCaskey:
This makes the code more self documenting and correctly uses `unsafe`
to communicate that it's the caller's responsibility to ensure that the
code paths that can lead to the access can only write the value they
expect to be there.

spun off of #1739

- [x] There's still some inconsistent use of `vmctx` and `extra_data` might be good to unify this more.
- [x] Improve docs on the new type

### DIFF
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -373,8 +373,7 @@ impl Function {
             Self {
                 store: store.clone(),
                 definition: FunctionDefinition::Host(HostFunctionDefinition {
-                    // TOOD: make safe function on this union to check for null
-                    has_env: !unsafe { wasmer_export.vmctx.host_env.is_null() },
+                    has_env: !wasmer_export.vmctx.is_null(),
                 }),
                 exported: wasmer_export,
             }
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -11,8 +11,8 @@ use std::cmp::max;
 use std::fmt;
 use wasmer_vm::{
     raise_user_trap, resume_panic, wasmer_call_trampoline, Export, ExportFunction,
-    VMCallerCheckedAnyfunc, VMContext, VMDynamicFunctionContext, VMFunctionBody, VMFunctionKind,
-    VMTrampoline,
+    VMCallerCheckedAnyfunc, VMDynamicFunctionContext, VMFunctionBody, VMFunctionExtraData,
+    VMFunctionKind, VMTrampoline,
 };
 
 /// A function defined in the Wasm module
@@ -85,7 +85,9 @@ impl Function {
         // The engine linker will replace the address with one pointing to a
         // generated dynamic trampoline.
         let address = std::ptr::null() as *const VMFunctionBody;
-        let vmctx = Box::into_raw(Box::new(dynamic_ctx)) as *mut VMContext;
+        let vmctx = VMFunctionExtraData {
+            host_env: Box::into_raw(Box::new(dynamic_ctx)) as *mut _,
+        };
 
         Self {
             store: store.clone(),
@@ -135,7 +137,9 @@ impl Function {
         // The engine linker will replace the address with one pointing to a
         // generated dynamic trampoline.
         let address = std::ptr::null() as *const VMFunctionBody;
-        let vmctx = Box::into_raw(Box::new(dynamic_ctx)) as *mut VMContext;
+        let vmctx = VMFunctionExtraData {
+            host_env: Box::into_raw(Box::new(dynamic_ctx)) as *mut _,
+        };
 
         Self {
             store: store.clone(),
@@ -176,7 +180,9 @@ impl Function {
     {
         let function = inner::Function::<Args, Rets>::new(func);
         let address = function.address() as *const VMFunctionBody;
-        let vmctx = std::ptr::null_mut() as *mut _ as *mut VMContext;
+        let vmctx = VMFunctionExtraData {
+            host_env: std::ptr::null_mut() as *mut _,
+        };
         let signature = function.ty();
 
         Self {
@@ -230,7 +236,9 @@ impl Function {
         // In the case of Host-defined functions `VMContext` is whatever environment
         // the user want to attach to the function.
         let box_env = Box::new(env);
-        let vmctx = Box::into_raw(box_env) as *mut _ as *mut VMContext;
+        let vmctx = VMFunctionExtraData {
+            host_env: Box::into_raw(box_env) as *mut _,
+        };
         let signature = function.ty();
 
         Self {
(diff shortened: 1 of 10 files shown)


lib/api/src/externals/function.rs
```diff
@@ -365,7 +373,8 @@ impl Function {
             Self {
                 store: store.clone(),
                 definition: FunctionDefinition::Host(HostFunctionDefinition {
-                    has_env: !wasmer_export.vmctx.is_null(),
+                    // TOOD: make safe function on this union to check for null
```
nlewycky, line 376:
```suggestion
                    // TODO: make safe function on this union to check for null
```

lib/vm/src/vmcontext.rs
```diff
@@ -15,15 +15,37 @@ use std::ptr::{self, NonNull};
 use std::sync::Arc;
 use std::u32;
 
+/// We stop lying about what this daat is
+/// TODO:
```
nlewycky, line 19: Did you mean to include these? A comment about "the first parameter may be a VMContext parameter for a wasm function call, or an arbitrary pointer produced by the host for host function calls" might make sense here?

lib/vm/src/vmcontext.rs
```diff
@@ -15,15 +15,37 @@ use std::ptr::{self, NonNull};
...
+#[derive(Copy, Clone)]
+pub union VMFunctionExtraData {
+    /// Wasm function, it has a real VMContext:
```
nlewycky, line 22:
```suggestion
    /// Wasm functions take a VMContext
```

lib/vm/src/vmcontext.rs
```diff
@@ -15,15 +15,37 @@ use std::ptr::{self, NonNull};
...
+    pub vmctx: *mut VMContext,
+    /// Host functions can have custom environments
+    pub host_env: *mut std::ffi::c_void,
+}
+
+impl std::fmt::Debug for VMFunctionExtraData {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "FunctionExtarData union")
```
nlewycky, line 30:
```suggestion
        write!(f, "FunctionExtraData union")
```

lib/vm/src/vmcontext.rs
```diff
@@ -15,15 +15,37 @@ use std::ptr::{self, NonNull};
...
+    }
+}
```
nlewycky, line 32: Do we need a custom Debug impl? At most we could see whether the pointer is null?

lib/vm/src/vmcontext.rs
```diff
@@ -15,15 +15,37 @@ use std::ptr::{self, NonNull};
 use std::sync::Arc;
 use std::u32;
 
+/// We stop lying about what this daat is
+/// TODO:
+#[derive(Copy, Clone)]
+pub union VMFunctionExtraData {
```
nlewycky, line 21: Optional: consider the name VMFunctionContextParameter or VMFunctionEnvParameter, or possibly CallbackParameter. Whether it's a wasm function or a host function this is giving that function its context/env.

lib/vm/src/vmcontext.rs
```diff
@@ -15,15 +15,37 @@ use std::ptr::{self, NonNull};
...
+        write!(f, "FunctionExtarData union")
+    }
+}
+
+impl std::cmp::PartialEq for VMFunctionExtraData {
+    fn eq(&self, rhs: &Self) -> bool {
+        unsafe { self.host_env as usize == rhs.host_env as usize }
+    }
+}
+
```
nlewycky, line 39: Do you want to impl VMFunctionExtraData { pub fn is_null() ... } ?

lib/vm/src/vmcontext.rs
MarkMcCaskey: Those were dev-comments, that I blindly ported over from the other PR, yeah will clean up

lib/vm/src/vmcontext.rs
MarkMcCaskey: You can't derive `Debug` on unions in Rust, but yeah we can definitely at least print some info here, and because it's really a union of exclusively pointers, we can honestly print the hex of the pointer too

lib/vm/src/vmcontext.rs
MarkMcCaskey: I'm down for renaming, but I'll wait for more feedback before renaming!  I like your suggestions though

nlewycky (approved):

lib/vm/src/vmcontext.rs
Hywan: I also like `VMFunctionContextParameter`. I would have use `VMFunctionEnvironment` first, but it creates a naming conflict with “host environment” or similar.

Hywan: Don't forget to rename the `extra_data` field everywhere accordingly too :-).

lib/vm/src/vmcontext.rs
```diff
@@ -15,15 +15,48 @@ use std::ptr::{self, NonNull};
...
+/// Union representing the first parameter passed when calling a function.
+///
+/// It may either be a pointer to the [`VMContext`] if it's a Wasm function
+/// or a pointer to arbitrary data controlled by the host if it's a host function.
+#[derive(Copy, Clone)]
+pub union VMFunctionExtraData {
+    /// Wasm functions take a pointer to [`VMContext`].
+    pub vmctx: *mut VMContext,
+    /// Host functions can have custom environments.
+    pub host_env: *mut std::ffi::c_void,
```
Hywan, line 27: As far as I understand this PR (which backport and enhance (!) a pre-1.0 feature), `host_env` can hold 2 kind of pointers: `VMContextWithoutEnv` and `VMContextWithEnv`. To clarify things, wouldn't it be possible to write this union as follows:

```rust
pub enum Either<L, R> {
    Left(L),
    Right(R),
}

pub union VMFunctionContextParameter {
    pub vmctx: *mut VMContext,
    pub host_env: Either<*mut VMContextWithEnv, *mut VMContextWithoutEnv>,
}
```

`mem::size_of::<VMFunctionContextParameter>` would be around 16 bytes, which is quite “heavy”. Without `Either`, it's 8 bytes.

If we don't really care about the size of this pointer, I would suggest to use `Either` or a similar enum, to be super explicity about what kind of pointer `host_env` can hold.

lib/vm/src/vmcontext.rs
syrusakbary: I like `VMFunctionEnvironment`

lib/vm/src/vmcontext.rs
MarkMcCaskey: My understanding is that there are 3 states that can exist on this single pointer:

1. The pointer is used in the context of a compiled Wasm function. The pointer is a `VMContext`.
2. The pointer is used in the context of a host function with an environment.  The pointer is a pointer to user data.
3. The pointer is used in the context of a host function without an environment.  The pointer is never used and we usually set it to null.

I don't think we need to represent the difference between 2 and 3 really, but I think the difference between 1 and 2 can be quite important.  Please let me know if my mental model of this is not the same as yours!

Hywan: Your mental model matches mine. I don't know if that's a good news though 🤔.

😄

PR #1756, opened 2020-10-23
Add a comment indicating what generates this file.

nlewycky:


MarkMcCaskey (approved):
Can you regenerate the header file so this comment gets included before shipping this? Otherwise all subsequent builds will have changes in this file

MarkMcCaskey (approved):

PR #1758, opened 2020-10-23
Globals are already stored inside the VMContext.

nlewycky:
I claim these two TODOs are both detritus. Change my mind :wink:

### DIFF
--- a/lib/vm/src/instance.rs
+++ b/lib/vm/src/instance.rs
@@ -255,7 +255,6 @@ impl Instance {
     /// Return the indexed `VMGlobalDefinition`.
     fn global_ptr(&self, index: LocalGlobalIndex) -> NonNull<VMGlobalDefinition> {
         let index = usize::try_from(index.as_u32()).unwrap();
-        // TODO:
         NonNull::new(unsafe { *self.globals_ptr().add(index) }).unwrap()
     }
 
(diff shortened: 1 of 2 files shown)


lib/vm/src/vmcontext.rs
```diff
@@ -909,8 +909,6 @@ impl VMBuiltinFunctionsArray {
 /// The struct here is empty, as the sizes of these fields are dynamic, and
 /// we can't describe them in Rust's type system. Sufficient memory is
 /// allocated at runtime.
-///
-/// TODO: We could move the globals into the `vmctx` allocation too.
```
MarkMcCaskey, line 913: Looking at `InstanceHandle`, it looks like globals have an extra layer of indirection on them

```Rust
            instance.globals_ptr() as *mut NonNull<VMGlobalDefinition>,
```

Whereas everything else only has 1, so it looks like this comment is still valid.

We could probably fix this in the same way that #1687 is fixing memory/table

lib/vm/src/instance.rs
```diff
@@ -255,7 +255,6 @@ impl Instance {
     /// Return the indexed `VMGlobalDefinition`.
     fn global_ptr(&self, index: LocalGlobalIndex) -> NonNull<VMGlobalDefinition> {
         let index = usize::try_from(index.as_u32()).unwrap();
-        // TODO:
```
MarkMcCaskey, line 258: Not sure what this `TODO` is referring to... maybe the double pointer thing?

syrusakbary:
Closing this PR as is stale

PR #1761, opened 2020-10-26
feat(c-api) Implement the `traps` argument of `wasm_instance_new`

Hywan:
# Description

When running `Instance::new`, it can error with an
`InstantiationError`. There is 2 scenarii:

1. Either it's a `InstantiationError::Link`. In this case, the
   `wasm_instance_new` function must return `NULL` and register the
   error in the Wasmer error registry.

2. Either it's a `InstantiationError::Start`. In this case, the
   `wasm_instance_new` function must return `NULL` and the error must be
   converted into a `wasm_trap_t*`, which is stored in the `wasm_trap_t**`.

# Review

Fixes https://github.com/wasmerio/wasmer/issues/1744.


Hywan:
I'm not sure why `wasm_instance_new` expects a `wasm_trap_t**`, i.e. an array of `wasm_trap_t*`. I don't see how can have more than one trap. Maybe @MarkMcCaskey or @syrusakbary can help me here.

lib/c-api/src/wasm_c_api/instance.rs
```diff
@@ -32,7 +31,35 @@ pub unsafe extern "C" fn wasm_instance_new(
...
+            return None;
+        }
+
+        Err(InstantiationError::Start(runtime_error)) => {
+            let pointer = {
+                let trap: Box<wasm_trap_t> = Box::new(runtime_error.into());
+                let mut traps: Vec<*mut wasm_trap_t> = Vec::with_capacity(1);
+                traps.push(Box::into_raw(trap));
+
+                debug_assert!(traps.len() == traps.capacity());
```
jubianchi, line 49: Do we really need this here?

lib/c-api/src/wasm_c_api/instance.rs
```diff
@@ -17,8 +17,7 @@ pub unsafe extern "C" fn wasm_instance_new(
     _store: &wasm_store_t,
     module: &wasm_module_t,
     imports: &wasm_extern_vec_t,
-    // own
```
MarkMcCaskey, line 20: Might be useful to put this comment back if it's correct, otherwise it's not clear what the ownership is of `*mut *mut`, `own` means we're responsible for freeing it I believe

lib/c-api/src/wasm_c_api/instance.rs
MarkMcCaskey: I think it's probably good to have, alternatively we can call `shrink_to_fit` on the vector and guarantee that it's always true.  Methods like `into_boxed_slice` will shrink the allocation too, because slices have no notion of capacity

lib/c-api/src/wasm_c_api/instance.rs
```diff
@@ -32,7 +31,35 @@ pub unsafe extern "C" fn wasm_instance_new(
...
+        Err(InstantiationError::Start(runtime_error)) => {
+            let pointer = {
+                let trap: Box<wasm_trap_t> = Box::new(runtime_error.into());
+                let mut traps: Vec<*mut wasm_trap_t> = Vec::with_capacity(1);
+                traps.push(Box::into_raw(trap));
+
+                debug_assert!(traps.len() == traps.capacity());
+
+                let pointer = traps.as_mut_ptr();
+                mem::forget(traps);
```
MarkMcCaskey, line 52: This code seems reasonable to me, ideally we could encapsulate more of it but I don't know if that's practical...

But without encapsulation, outside of the context of this PR diff, it's not obvious that this is a memory leak (which I believe it is)... we should leave a comment about it or try to fix it in this PR

MarkMcCaskey (commented):
Looks like a good change overall, but I think we need to iterate a bit more on the ownership of this

lib/c-api/src/wasm_c_api/instance.rs
syrusakbary: I believe it would be good to not have it here.

The debug asserts have bitten us before, and the logic is implied by `Vec::with_capacity(1)`

Hywan: Removed.

lib/c-api/src/wasm_c_api/instance.rs
Hywan: We can't own it. We can't free it. The `start.c` example explicitely drops the `trap` value:

https://github.com/wasmerio/wasmer/blob/86f3438acd23ccb3026143bcebd96b49ddf659f3/lib/c-api/tests/wasm_c_api/wasm-c-api/example/start.c#L58-L60

and later:

https://github.com/wasmerio/wasmer/blob/86f3438acd23ccb3026143bcebd96b49ddf659f3/lib/c-api/tests/wasm_c_api/wasm-c-api/example/start.c#L95

I believe that the value is owned by the caller, not by the callee.

lib/c-api/src/wasm_c_api/instance.rs
Hywan: We leak it because we don't own it. The caller owns it.

lib/c-api/src/wasm_c_api/instance.rs
MarkMcCaskey: Huh, `own wasm_trap_t**` is weird in the context of using arguments as outputs.

It seems like there's no need for the `Vec` at all then, this just seems to be a double pointer for the sake of setting a `trap*`

lib/c-api/src/wasm_c_api/instance.rs
```diff
@@ -32,7 +31,34 @@ pub unsafe extern "C" fn wasm_instance_new(
...
+                traps.push(Box::into_raw(trap));
+                traps.shrink_to_fit();
+
+                let pointer = traps.as_mut_ptr();
+                mem::forget(traps);
+
+                pointer
+            };
+
+            *traps = *pointer;
```
MarkMcCaskey, line 56:
```suggestion
            *traps = pointer;
```

lib/c-api/src/wasm_c_api/instance.rs
```diff
@@ -32,7 +31,34 @@ pub unsafe extern "C" fn wasm_instance_new(
...
+        Err(InstantiationError::Link(link_error)) => {
+            crate::error::update_last_error(link_error);
+
+            return None;
+        }
+
+        Err(InstantiationError::Start(runtime_error)) => {
+            let pointer = {
+                let trap: Box<wasm_trap_t> = Box::new(runtime_error.into());
+                let mut traps: Vec<*mut wasm_trap_t> = Vec::with_capacity(1);
```
MarkMcCaskey, line 46: There's no need for us to allocate 2 indirections, we only need 1

MarkMcCaskey (commented):
Based on the example in start.c, it seems that we just need one pointer here, not two.  The `Vec` is never freed nor used here!

lib/c-api/src/wasm_c_api/instance.rs
MarkMcCaskey: To be fair, the debug asserts did exactly what they were supposed to: they caught a memory leak 😆 the primary issue is that we don't test with debug_asserts enabled so we didn't catch the bug, a user did... perhaps they should just be normal `assert!` or we should test with `debug_assert!`...

MarkMcCaskey: That said, I strongly prefer writing code in such a way that this isn't necessary, we were able to abstract a lot of our Vec into pointer code away and now this logic can't be messed up.

lib/c-api/src/wasm_c_api/instance.rs
Hywan: So `wasm_trap_t**` must not be interpreted as an array of `wasm_trap_t*`, hmm. Let's fix that then.

jubianchi (approved):

PR #1762, opened 2020-10-26
fix(wasi) Allow the `=` sign in the environment variable value.

### LINKED ISSUE #1708: wasmer_wasi: Unable to set env values containing = (equal sign)
### Problem

I was trying to run CGI apps using `wasmer-wasi`. You need to pass the client headers to the CGI app using environment variables, but some client headers contain equal signs (e.g `sec-ch-ua: "Chromium";v="86", "\"Not\\A;Brand";v="99", "Google Chrome";v="86"`), which are rejected by `wasmer-wasi`, but work fine using `wasmtime_wasi`.

### Bug

Setting a environment variable to a value containing an equal sign (=) is rejected by `WasiStateBuilder::build`, even though only keys may not contain equal signs. The first equal sign seperates the key from the value. Any other equal signs should be treated as part of the value.

```yaml
[dependencies]
wasmer-wasi = "1.0.0-alpha3"
```

```rust
fn main() {
    wasmer_wasi::WasiState::new("hello").env("DUMMY", "X=1").finalize().expect("this should work");
}
```

```
thread 'main' panicked at 'this should work: EnvironmentVariableFormatError("found \'=\' in env var string \"DUMMY=X=1\" (key=value)")', src/main.rs:2:73
```

### Fix

Removing the incorrect check for multiple equal signs:

```rust
diff --git a/lib/wasi/src/state/builder.rs b/lib/wasi/src/state/builder.rs
index f6694bb02..6d5118e69 100644
--- a/lib/wasi/src/state/builder.rs
+++ b/lib/wasi/src/state/builder.rs
@@ -329,22 +329,9 @@ impl WasiStateBuilder {
             }
         }
         for env in self.envs.iter() {
-            let mut eq_seen = false;
             for b in env.iter() {
                 match *b {
-                    b'=' => {
-                        if eq_seen {
-                            return Err(WasiStateCreationError::EnvironmentVariableFormatError(
-                                format!(
-                                    "found '=' in env var string \"{}\" (key=value)",
-                                    std::str::from_utf8(env)
-                                        .unwrap_or("Inner error: env var is invalid_utf8!")
-                                ),
-                            ));
-                        }
-                        eq_seen = true;
-                    }
-                    0 => {
+                     0 => {
                         return Err(WasiStateCreationError::EnvironmentVariableFormatError(
                             format!(
                                 "found nul byte in env var string \"{}\" (key=value)",
```

It might make sense to add a check for equal signs in the key in the function `WasiStateBuilder::env`, so that `WasiStateBuilder::build` can return an error in such cases.

### Sample wasm app

```rust
fn main() {
    for (name,value) in std::env::vars() {
        println!("{} = {}", name, value);
    }
}
```

Output using wasmtime:
```sh
$ wasmtime run --env DUMMY=X=1 sample.wasm
DUMMY = X=1
```

Output using wasmer:
```sh
$ wasmer sample.wasm --env DUMMY=X=1
Error: Env vars must be of the form <var_name>=<value>. Found DUMMY=X=1
```

Hywan:
# Description

Having `a=b` means that the environment variable key is `a`, and its
value is `b`.

Having `a=b=c` means that the key is `a` and its value is `b=c`.

It's OK to get this, e.g. within a CGI context where values can hold
strings such as `sec-ch-ua: "Chromium;v=86"` etc. See
#1708 for a longer
explanation.

# Review

Fixes #1708.

### DIFF
--- a/lib/wasi/src/state/builder.rs
+++ b/lib/wasi/src/state/builder.rs
@@ -334,7 +334,7 @@ impl WasiStateBuilder {
                 return Err(WasiStateCreationError::EnvironmentVariableFormatError(
                     format!(
                         "found nul byte in env var string \"{}\" (key=value)",
-                        std::str::from_utf8(env).unwrap_or("Inner error: env var is invalid_utf8!")
+                        String::from_utf8_lossy(env)
                     ),
                 ));
             }
--- a/lib/wasi/src/state/builder.rs
+++ b/lib/wasi/src/state/builder.rs
@@ -93,8 +93,10 @@ fn validate_mapped_dir_alias(alias: &str) -> Result<(), WasiStateCreationError> 
 // return stdout somehow, it's unclear what that API should look like)
 impl WasiStateBuilder {
     /// Add an environment variable pair.
-    /// Environment variable keys must not contain the byte `=` (0x3d)
-    /// or nul (0x0).
+    ///
+    /// Both the key and value of an environment variable must not
+    /// contain a nul byte (`0x0`), and the key must not contain the
+    /// `=` byte (`0x3d`).
     pub fn env<Key, Value>(&mut self, key: Key, value: Value) -> &mut Self
     where
         Key: AsRef<[u8]>,
@@ -107,6 +109,7 @@ impl WasiStateBuilder {
     }
 
     /// Add an argument.
+    ///
     /// Arguments must not contain the nul (0x0) byte
     pub fn arg<Arg>(&mut self, arg: Arg) -> &mut Self
     where
@@ -118,7 +121,10 @@ impl WasiStateBuilder {
     }
 
     /// Add multiple environment variable pairs.
-    /// Keys must not contain the `=` (0x3d) or nul (0x0) byte.
+    ///
+    /// Both the key and value of the environment variables must not
+    /// contain a nul byte (`0x0`), and the key must not contain the
+    /// `=` byte (`0x3d`).
     pub fn envs<I, Key, Value>(&mut self, env_pairs: I) -> &mut Self
     where
         I: IntoIterator<Item = (Key, Value)>,
@@ -133,6 +139,7 @@ impl WasiStateBuilder {
     }
 
     /// Add multiple arguments.
+    ///
     /// Arguments must not contain the nul (0x0) byte
     pub fn args<I, Arg>(&mut self, args: I) -> &mut Self
     where
@@ -147,6 +154,7 @@ impl WasiStateBuilder {
     }
 
     /// Preopen a directory
+    ///
     /// This opens the given directory at the virtual root, `/`, and allows
     /// the WASI module to read and write to the given directory.
     pub fn preopen_dir<FilePath>(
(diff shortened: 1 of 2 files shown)


Hywan:
bors try

lib/wasi/src/state/builder.rs
```diff
@@ -328,33 +328,15 @@ impl WasiStateBuilder {
...
-                            ),
-                        ));
-                    }
-                    _ => (),
-                }
+            if env.iter().find(|&&ch| ch == 0).is_some() {
+                return Err(WasiStateCreationError::EnvironmentVariableFormatError(
+                    format!(
+                        "found nul byte in env var string \"{}\" (key=value)",
+                        std::str::from_utf8(env).unwrap_or("Inner error: env var is invalid_utf8!")
```
Hywan, line 337: We may want to use `str::from_utf8_lossy` here. Thoughts @MarkMcCaskey or @jubianchi?

ensch:
Should  `WasiState::new("hello").env("KEY=BAD", "Value").finalize()` now be allowed and set the key "KEY" to the value "BAD=Value"?

Hywan:
Yeah, the validation step happens in the `build()` method, not in the `env()` or `envs()` methods. I can change that a little bit. The modification implies to not flatten the env key + value in `env()` or `envs()`, but in `build()` after the validation. How does it sound @MarkMcCaskey?

ensch:
Maybe `env()` and `envs()` should set a flag if called with invalid input, so `build()` can return an error like before?

syrusakbary (approved):

lib/wasi/src/state/builder.rs
MarkMcCaskey: Yeah, I can believe that that'd be a good change here

lib/wasi/src/state/builder.rs
```diff
@@ -508,17 +490,21 @@ mod test {
 
     #[test]
     fn env_var_errors() {
+        // `a=b` means key is `a` and value is `b`, which is OK.
+        // `a=b=c` means key is `a` and value is `b=c`, which is OK too.
         let output = create_wasi_state("test_prog")
-            .env("HOM=E", "/home/home")
```
MarkMcCaskey, line 512: Do we want to put the test checking the failure of `=` in the key back?

MarkMcCaskey (approved):
Looks good! I think we can probably make a new test rather than removing one though

lib/wasi/src/state/builder.rs
Hywan: done.

lib/wasi/src/state/builder.rs
Hywan: Yes. I've updated the test cases to handle all the scenarios.

Hywan:
The last patch keeps the env's key and value separate until the `build()` method is called. It allows to apply different validations resp. for the key and the value: Both must not contain a nul byte, but only the key cannot contain an equal sign.

This patch also updates the test cases accordingly.

Finally this patch updates the `envs` and `args` methods to reuse respectively `env` and `arg` to avoid code repetition.

PR #1763, opened 2020-10-26
Test C API with the native engine as well.

### LINKED ISSUE #1726: test capi for each engine
Currently the makefile has separate rules for each compiler like `test-capi-singlepass` and `test-capi-cranelift`, but these build with the JIT. The C API works with multiple engines and you can specify which engine through the API, but the testing run by the Makefile tests with the JIT only. We should run the CAPI tests for all engines as well as all compilers.

nlewycky:
Fixes #1726 .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ build-docs-capi:
 	cd lib/c-api/ && doxygen doxyfile
 
 # We use cranelift as the default backend for the capi for now
-build-capi: build-capi-cranelift
+build-capi: build-capi-cranelift-jit
 
 build-capi-singlepass-jit:
 	cargo build --manifest-path lib/c-api/Cargo.toml --release \
--- a/Makefile
+++ b/Makefile
@@ -94,21 +94,29 @@ build-docs-capi:
 # We use cranelift as the default backend for the capi for now
 build-capi: build-capi-cranelift
 
-build-capi-singlepass:
+build-capi-singlepass-jit:
 	cargo build --manifest-path lib/c-api/Cargo.toml --release \
 		--no-default-features --features wat,jit,object-file,singlepass,wasi
 
-build-capi-cranelift:
+build-capi-cranelift-jit:
 	cargo build --manifest-path lib/c-api/Cargo.toml --release \
 		--no-default-features --features wat,jit,object-file,cranelift,wasi
 
+build-capi-cranelift-native:
+	cargo build --manifest-path lib/c-api/Cargo.toml --release \
+		--no-default-features --features wat,native,object-file,cranelift,wasi
+
 build-capi-cranelift-system-libffi:
 	cargo build --manifest-path lib/c-api/Cargo.toml --release \
-		--no-default-features --features wat,jit,object-file,cranelift,wasi,system-libffi
+		--no-default-features --features wat,jit,native,object-file,cranelift,wasi,system-libffi
 
-build-capi-llvm:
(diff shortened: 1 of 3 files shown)


lib/c-api/tests/deprecated/test-module-serialize.c
```diff
@@ -28,18 +28,6 @@ int main()
...
-    assert(serialized_module_bytes.bytes[1] == 'w');
-    assert(serialized_module_bytes.bytes[2] == 'a');
-    assert(serialized_module_bytes.bytes[3] == 's');
-    assert(serialized_module_bytes.bytes[4] == 'm');
-    assert(serialized_module_bytes.bytes[5] == 'e');
-    assert(serialized_module_bytes.bytes[6] == 'r');
-    assert(serialized_module_bytes.bytes[7] == '-');
-    assert(serialized_module_bytes.bytes[8] == 'j');
-    assert(serialized_module_bytes.bytes[9] == 'i');
-    assert(serialized_module_bytes.bytes[10] == 't');
```
MarkMcCaskey, line 42: Since this is the deprecated API it may not be worth testing, but it'd be good to have some logic here for verifying the binary blob we created isn't garbage.

I don't know if there's enough commonality to make that work... It might also make sense to have `CMake` able to pass defines to the C source so that the C code being compiled can be aware of this kind of thing and make adjustments as necessary.

Actually if we're testing that it deserializes, then it's probably sufficiently well tested.

Makefile
```diff
@@ -94,22 +94,30 @@ build-docs-capi:
 # We use cranelift as the default backend for the capi for now
 build-capi: build-capi-cranelift
```
MarkMcCaskey, line 95: This command looks like it will fail now

MarkMcCaskey (commented):
Not sure what we're shipping right now but it might be good to have an option to bundle in more of the features in the C API, I don't think we have to be selective about backends or engines, other than binary size, they should all work together...

I think moving away from more compile time conditional logic and towards more runtime conditional logic for stuff like this is a good idea, especially at the level that the C API currently sits at, there's very little compile-time code at all

Makefile
nlewycky: Thanks, fixed.

nlewycky:
> Not sure what we're shipping right now but it might be good to have an option to bundle in more of the features in the C API, I don't think we have to be selective about backends or engines, other than binary size, they should all work together...
> 
> I think moving away from more compile time conditional logic and towards more runtime conditional logic for stuff like this is a good idea, especially at the level that the C API currently sits at, there's very little compile-time code at all

The build does support having jit,native,object-file all specified at the same time and it picks one based on a hard-coded priority list in src/wasm_c_api/engine.rs. To run the tests with a particular engine we have to compile without the other engines.

I agree it'd be better to build the tests once and have them run internally with different engines. I'm not sure how we'd do that without any source code changes to wasm-c-api examples. Maybe an environment variable? I decided that the tests run quickly enough that adding more runs of them isn't a big problem and that I didn't want to design something more sophisticated. What do you think?

lib/c-api/tests/deprecated/test-module-serialize.c
nlewycky: > Actually if we're testing that it deserializes, then it's probably sufficiently well tested.

Yep, that's exactly the thought process I went through.

MarkMcCaskey (approved):

PR #1764, opened 2020-10-27
Fix issue with `path_rename` moving down 1 directory

### LINKED ISSUE #1759: Renaming a file unexpectedly places it outside of pre-opened directory
### Describe the bug
Whenever a file is renamed it's placed outside of pre-opened directory. It happens only if the file isn't in a subdirectory.

```sh
wasmer 1.0.0-alpha4 | rustc 1.47.0 (18bf6b4f0 2020-10-07) | x86_64
```

### Steps to reproduce
1. Download [uutils.wasm](https://registry-cdn.wapm.io/contents/_/coreutils/0.0.1/target/wasm32-wasi/release/uutils.wasm) from WAPM
2. Create directory dir with a sample file foo in it
3. Run `wasmer --dir=./dir uutils.wasm -- mv dir/foo dir/bar`

### Expected behavior
File ./dir/foo is renamed to ./dir/bar

### Actual behavior
File ./dir/foo is renamed to ./bar

### Additional context
`wasmer --dir=./dir uutils.wasm -- mv dir/dir2/foo dir/dir2/bar` works correctly

MarkMcCaskey:
Resolves #1759 

Due to file metadata caching in WasiFS this behavior isn't easily observable with tests because as far as the rest of the WASI calls are concerned the file is in the right place:

perhaps we should consider removing some of that caching, but it's a large change and may have other implications...
Ideally we do want to not be dependent on the file system when dealing with open files... the issue here is that when we rename it, we still consider it to be a file that WasiFS is actively managing. This maybe relatively easy to fix on its own

# Review


syrusakbary:
Add test?

MarkMcCaskey:
> Add test?

Yeah been working on it, but as I said, it's not easily observable from inside our WASI system. I think I've probably found a way to test it though

Hywan (approved):

jubianchi (approved):

PR #1765, opened 2020-10-27
doc(c-api) Update documentation.

Hywan:
# Description

This patch adds documentation to the  `engine.rs` module.

# Review


lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -58,21 +65,25 @@ impl Default for wasmer_engine_t {
     }
 }
 
+/// A configuration holds the compiler and the engine used by the store.
```
MarkMcCaskey, line 68: Nit: the config could be used for other configuration in the future and it's none of the user's business what it contains 😆 

I think explaining what it's used for is useful though

MarkMcCaskey (approved):

PR #1767, opened 2020-10-27
feat(makefile) Add more `build-capi-*` targets

Hywan:
# Description

The new targets are the following:

* `build-capi-<compiler-name>` to compile the C API with only one
  compiler, and 3 engines: JIT, native, and object-file.
* `build-capi-<compiler-name>-<engine-name>` to compile the C API with
  only one compiler and only one engine, defined by `engine-name`.

One question remains: What is `build-capi-cranelift-system-libffi`?

# Review

- [ ] ~Add a short description of the the change to the CHANGELOG.md file~ not necesary

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,15 +94,35 @@ build-docs-capi:
 # We use cranelift as the default backend for the capi for now
 build-capi: build-capi-cranelift-jit
 
+build-capi-singlepass:
+	cargo build --manifest-path lib/c-api/Cargo.toml --release \
+		--no-default-features --features wat,jit,native,object-file,singlepass,wasi
+
 build-capi-singlepass-jit:
 	cargo build --manifest-path lib/c-api/Cargo.toml --release \
-		--no-default-features --features wat,jit,object-file,singlepass,wasi
+		--no-default-features --features wat,jit,singlepass,wasi
+
+build-capi-singlepass-native:
+	cargo build --manifest-path lib/c-api/Cargo.toml --release \
+		--no-default-features --features wat,native,singlepass,wasi
+
+build-capi-singlepass-object-file:
+	cargo build --manifest-path lib/c-api/Cargo.toml --release \
+		--no-default-features --features wat,object-file,singlepass,wasi
+
+build-capi-cranelift:
+	cargo build --manifest-path lib/c-api/Cargo.toml --release \
+		--no-default-features --features wat,jit,native,object-file,cranelift,wasi
 
 build-capi-cranelift-jit:
 	cargo build --manifest-path lib/c-api/Cargo.toml --release \
-		--no-default-features --features wat,jit,object-file,cranelift,wasi
+		--no-default-features --features wat,jit,cranelift,wasi
 
 build-capi-cranelift-native:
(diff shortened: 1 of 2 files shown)


MarkMcCaskey (commented):
I mentioned this to Nick when he made a PR increasing the number of rules for the C API in the makefile too: I think we should move away from compile time options on the C API.

Almost none of the C API is feature gated and it's designed to work with all the features compiled.  The only thing that doesn't work with this are the official Wasm-c-api examples unmodified, we can't test those in the feature matrix, but perhaps we can use an environment variable or something instead.

I think we should use the C API like users will be interacting with it and that aside from specialized/minimal use cases, users won't be compiling the C API with custom features.  Perhaps that's what these makefile rules are for but maybe that should live in `lib/c-api` and not the top level Makefile / maybe users should use `cargo build` directly if they want full control.

As for the `system-ffi` option, that's a workaround to compiling the C API on my computer.  Mac OSX has a permissions system from hell that refuses to allow libffi to be compiled from source on an external harddrive (even when I explicitly gave permission to everything that I could see was involved and tried to disable all the protections on my mac 🤷 ).  It's a workaround to an annoying bug in OSX that affects me.

nlewycky:
If you write the equivalent test-capi-*-object-file rules, the tests will fail. I'm not sure it makes sense to add the build-* rules without the matching test-* rules, and then those tests should pass.

MarkMcCaskey (approved):
Approving so that I'm not blocking you if this is something you need soon.

I'd like to avoid compile time logic in the C API though, though a Makefile in `lib/c-api` isn't as bad in my opinion because it doesn't encourage the use of compile time flags as much by default.

My preference is for there to be 1 canonical version of libwasmer which most people use most of the time.  We should support specialized and light builds but shouldn't make it the default or encourage it; I think this is a simpler user experience and simpler from our side too.

Hywan:
The original purpose of this PR was to embed the 3 engines inside `libwasmer`. I then noticed that targets were missing. And here we are.

I propose to merge this PR as is. Then cleanup the possible targets. We need to find a solution to provide the 3 _compilers_ anyway, so it will reopen the discussion!

syrusakbary (approved):

jubianchi:
bors try

Makefile
```diff
@@ -92,31 +92,59 @@ build-docs-capi:
...
-build-capi: build-capi-cranelift-jit
+build-capi: build-capi-cranelift
+
+build-capi-singlepass:
+	cargo build --manifest-path lib/c-api/Cargo.toml --release \
+		--no-default-features --features wat,jit,native,object-file,singlepass,wasi
 
 build-capi-singlepass-jit:
 	cargo build --manifest-path lib/c-api/Cargo.toml --release \
-		--no-default-features --features wat,jit,object-file,singlepass,wasi
```
jubianchi, line 99: `object-file` was required for integration tests.

To fix that, I changed the target used by `build-capi`.

jubianchi (approved):

jubianchi:
bors try

jubianchi:
bors try

PR #1770, opened 2020-10-27
Enable these tests now that they pass.

nlewycky:
Also fix a syntax error, use assert_eq! instead of assert! to compare two values for equality.

Also in passing, convert a comment about something not being done into a TODO.


MarkMcCaskey (approved):
🎉

PR #1773, opened 2020-10-28
Test Tunables::memory_style

webmaster128:
# Description

Using a static memory bound larger than the requested maximum seems like a waste of memory. Is there a good reason to do that? Or does this PR fix a bug?

# Review

### DIFF
--- a/lib/api/src/tunables.rs
+++ b/lib/api/src/tunables.rs
@@ -71,7 +71,7 @@ impl BaseTunables for Tunables {
         let maximum = memory.maximum.unwrap_or_else(Pages::max_value);
         if maximum <= self.static_memory_bound {
             MemoryStyle::Static {
-                bound: self.static_memory_bound,
+                bound: maximum,
                 offset_guard_size: self.static_memory_offset_guard_size,
             }
         } else {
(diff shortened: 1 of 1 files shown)


MarkMcCaskey (commented):
I'm not sure if this code affects it but I know that we use a trick with virtual memory to elide bounds checks on static memory.  From memory, here's how it works:

if we have a static bound of something like 4GiB then we can say, if we always have 32bit Wasm pointers (which we always do for now), then they're always in bounds and we don't need to check them before dereferencing them.  If the pointer points past the max, then it will attempt to read protected virtual memory and trap.
This doesn't actually use 4GiB of RAM on most modern operating systems because all of the memory that's inaccessible in that 4GiB is purely virtual.  It does use up 4GiB of the virtual address space but the upper limit of that on 64bit Linux, for example, is extremely large.

I can look into this in more detail to see if this actually affects that or not.

webmaster128:
> if we have a static bound of something like 4GiB then we can say, if we always have 32bit Wasm pointers (which we always do for now), then they're always in bounds and we don't need to check them before dereferencing them.

Okay, so a static memory larger than `maximum` is a performance optimization? Seems like this is documented for the 64 bit case:

```rust
        let (mut static_memory_bound, mut static_memory_offset_guard_size): (Pages, u64) =
            match pointer_width {
                PointerWidth::U16 => (0x400.into(), 0x1000),
                PointerWidth::U32 => (0x4000.into(), 0x1_0000),
                // Static Memory Bound:
                //   Allocating 4 GiB of address space let us avoid the
                //   need for explicit bounds checks.
                // Static Memory Guard size:
                //   Allocating 2 GiB of address space lets us translate wasm
                //   offsets into x86 offsets as aggressively as we can.
                PointerWidth::U64 => (0x1_0000.into(), 0x8000_0000),
            };
```

Happy to adapt the code and tests if that is the case. With a small comment, this would be easier to review. The current code looks like a bug to me.

syrusakbary (approved):
> Okay, so a static memory larger than maximum is a performance optimization?

That's right, however we should be safe since we set the guard size separately with `static_memory_offset_guard_size`.

-----

I think the PR does indeed fixes the behavior. If all tests passes we should be good to go!

webmaster128:
Hmm, but then this PR removed the documented bounds optimization _Allocating 4 GiB of address space let us avoid the need for explicit bounds checks._ for 64 bit wasm in case maximum is lower than 4 GiB.

syrusakbary:
Yeah, chatting now with @MarkMcCaskey about this. Thanks following up @webmaster128

MarkMcCaskey:
Yeah, our tests would have caught this so it's fine!

```
failures:

---- wast::cranelift::spec::data stdout ----
Running wast `tests/wast/spec/data.wast` with the cranelift compiler
thread 'wast::cranelift::spec::data' panicked at 'assertion failed: `(left >= right)`
  left: `2 pages`,
 right: `65536 pages`', lib/engine/src/resolver.rs:194:29

---- wast::cranelift::spec::imports stdout ----
Running wast `tests/wast/spec/imports.wast` with the cranelift compiler
13: i32
14: i32
42: f32
13: i32
13: i32
13: f32
13: i32
25: f64
53: f64
24: f64
24: f64
24: f64
thread 'wast::cranelift::spec::imports' panicked at 'assertion failed: `(left >= right)`
  left: `2 pages`,
 right: `65536 pages`', lib/engine/src/resolver.rs:194:29

---- wast::cranelift::spec::linking stdout ----
Running wast `tests/wast/spec/linking.wast` with the cranelift compiler
thread 'wast::cranelift::spec::linking' panicked at 'assertion failed: `(left >= right)`
  left: `5 pages`,
 right: `65536 pages`', lib/engine/src/resolver.rs:194:29


failures:
    wast::cranelift::spec::data
    wast::cranelift::spec::imports
    wast::cranelift::spec::linking
```

The thing I noticed was the documentation on `MemoryStyle::Static`:

```Rust
    Static {
        /// The number of mapped and unmapped pages.
        bound: Pages,
```

The change didn't make sense to me given that comment: mapped and unmapped is everything.  The change was changing it to be only the pages that we'd ever be able to map.

webmaster128:
Alright, thank you. Reverted the code change, added a short comment and adapted the tests.

MarkMcCaskey (approved):
Thanks!  Let's try it

PR #1776, opened 2020-10-29
NFC. Make FunctionBodyPtr hold a const pointer.

nlewycky:


MarkMcCaskey (approved):
Seems fine to me, but I'd expect there to be more places in the code that need to be updated

nlewycky:
`make test` passed for me :shrug:

PR #1781, opened 2020-10-30
Cranelift upgrade

syrusakbary:
# Description

Upgrade Cranelift to `0.67`. This upgrade also enables all SIMD tests (with one small exception operator that is still not fixed in Cranelift).

### DIFF
--- a/lib/compiler-cranelift/src/func_environ.rs
+++ b/lib/compiler-cranelift/src/func_environ.rs
@@ -11,7 +11,9 @@ use cranelift_codegen::ir::immediates::{Offset32, Uimm64};
 use cranelift_codegen::ir::types::*;
 use cranelift_codegen::ir::{AbiParam, ArgumentPurpose, Function, InstBuilder, Signature};
 use cranelift_codegen::isa::TargetFrontendConfig;
+use cranelift_frontend::FunctionBuilder;
 use std::convert::TryFrom;
+use wasmer_compiler::wasmparser::Type;
 use wasmer_compiler::{WasmError, WasmResult};
 use wasmer_types::entity::EntityRef;
 use wasmer_types::entity::PrimaryMap;
@@ -538,10 +540,11 @@ impl<'module_environment> BaseFuncEnvironment for FuncEnvironment<'module_enviro
 
     fn translate_table_grow(
         &mut self,
-        _: cranelift_codegen::cursor::FuncCursor<'_>,
-        _: TableIndex,
-        _: ir::Value,
-        _: ir::Value,
+        mut _pos: cranelift_codegen::cursor::FuncCursor<'_>,
+        _table_index: TableIndex,
+        _table: ir::Table,
+        _delta: ir::Value,
+        _init_value: ir::Value,
     ) -> WasmResult<ir::Value> {
         Err(WasmError::Unsupported(
             "the `table.grow` instruction is not supported yet".into(),
@@ -550,9 +553,10 @@ impl<'module_environment> BaseFuncEnvironment for FuncEnvironment<'module_enviro
 
     fn translate_table_get(
         &mut self,
-        _: cranelift_codegen::cursor::FuncCursor<'_>,
-        _: TableIndex,
-        _: ir::Value,
+        _builder: &mut FunctionBuilder,
+        _table_index: TableIndex,
+        _table: ir::Table,
+        _index: ir::Value,
     ) -> WasmResult<ir::Value> {
         Err(WasmError::Unsupported(
             "the `table.get` instruction is not supported yet".into(),
@@ -561,10 +565,11 @@ impl<'module_environment> BaseFuncEnvironment for FuncEnvironment<'module_enviro
 
     fn translate_table_set(
         &mut self,
-        _: cranelift_codegen::cursor::FuncCursor<'_>,
-        _: TableIndex,
-        _: ir::Value,
-        _: ir::Value,
+        _builder: &mut FunctionBuilder,
+        _table_index: TableIndex,
+        _table: ir::Table,
+        _value: ir::Value,
+        _index: ir::Value,
     ) -> WasmResult<()> {
         Err(WasmError::Unsupported(
             "the `table.set` instruction is not supported yet".into(),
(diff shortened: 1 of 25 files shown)


lib/compiler-cranelift/src/func_environ.rs
```diff
@@ -573,21 +578,56 @@ impl<'module_environment> BaseFuncEnvironment for FuncEnvironment<'module_enviro
...
     }
 
+    fn translate_ref_null(
+        &mut self,
+        mut pos: cranelift_codegen::cursor::FuncCursor,
+        ty: Type,
+    ) -> WasmResult<ir::Value> {
+        Ok(match ty {
+            Type::FuncRef => pos.ins().iconst(self.pointer_type(), 0),
+            // Type::ExternRef => pos.ins().null(self.reference_type(ty)),
```
nlewycky, line 599: Optional: generally we don't check in commented out code.

lib/compiler-cranelift/src/translator/code_translator.rs
```diff
@@ -37,17 +37,19 @@ use cranelift_codegen::ir::condcodes::{FloatCC, IntCC};
...
+use smallvec::SmallVec;
 use std::vec::Vec;
+
 use wasmer_compiler::wasmparser::{MemoryImmediate, Operator};
 use wasmer_compiler::{to_wasm_error, WasmResult};
 use wasmer_compiler::{wasm_unsupported, ModuleTranslationState};
 use wasmer_types::{FunctionIndex, GlobalIndex, MemoryIndex, SignatureIndex, TableIndex};
 
-// Clippy warns about "flags: _" but its important to document that the flags field is ignored
+// Clippy warns about "align: _" but its important to document that the flags field is ignored
```
nlewycky, line 52: Should this be:
```suggestion
// Clippy warns about "align: _" but its important to document that the aign field is ignored
```

lib/compiler-cranelift/src/translator/code_translator.rs
```diff
@@ -1291,15 +1381,27 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             )?;
             let splatted = builder.ins().splat(type_of(op), state.pop1());
             state.push1(splatted)
+
+            // let opcode = ir::Opcode::LoadSplat;
```
nlewycky, line 1385: commented out code again

lib/compiler-cranelift/src/translator/code_translator.rs
```diff
@@ -1460,6 +1566,11 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             let bool_result = builder.ins().vall_true(a);
             state.push1(builder.ins().bint(I32, bool_result))
         }
+        Operator::I8x16Bitmask | Operator::I16x8Bitmask | Operator::I32x4Bitmask => {
+            unimplemented!("SIMD Operator {:?} not yet implemented", op);
+            // let a = pop1_with_bitcast(state, type_of(op), builder);
```
nlewycky, line 1571: commented-out code

lib/compiler-cranelift/src/translator/code_translator.rs
```diff
@@ -1535,6 +1646,21 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
             let (a, b) = pop2_with_bitcast(state, type_of(op), builder);
             state.push1(builder.ins().fmin(a, b))
         }
+        Operator::F32x4PMax | Operator::F64x2PMax => {
+            unimplemented!("SIMD Operator {:?} not yet implemented", op);
+            // let (a, b) = pop2_with_bitcast(state, type_of(op), builder);
```
nlewycky, line 1651: commented-out code here and a few lines later

lib/compiler-cranelift/src/translator/code_translator.rs
```diff
@@ -1894,7 +2062,7 @@ fn mem_op_size(opcode: ir::Opcode, ty: Type) -> u32 {
         ir::Opcode::Istore8 | ir::Opcode::Sload8 | ir::Opcode::Uload8 => 1,
         ir::Opcode::Istore16 | ir::Opcode::Sload16 | ir::Opcode::Uload16 => 2,
         ir::Opcode::Istore32 | ir::Opcode::Sload32 | ir::Opcode::Uload32 => 4,
-        ir::Opcode::Store | ir::Opcode::Load => ty.bytes(),
+        ir::Opcode::Store | ir::Opcode::Load /* | ir::Opcode::LoadSplat */=> ty.bytes(),
```
nlewycky, line 2065: I think this would be wrong for LoadSplat -- LoadSplat loads a small size (1 byte / 2 byte / 4 byte / 8 byte) and splats it to the full 16 byte. If you use the type of the value pushed onto the stack by the loadsplat instruction then you would always return 16 for the V128. The function is supposed to return how many bytes it loads/stores.

lib/compiler-cranelift/src/translator/code_translator.rs
```diff
@@ -2161,6 +2571,80 @@ pub fn optionally_bitcast_vector(
     }
 }
 
+#[inline(always)]
+fn is_non_canonical_v128(ty: ir::Type) -> bool {
+    match ty {
+        B8X16 | B16X8 | B32X4 | B64X2 | I64X2 | I32X4 | I16X8 | F32X4 | F64X2 => true,
+        _ => false,
+    }
```
nlewycky, line 2579: Use `matches!`.

nlewycky (approved):

lib/compiler-cranelift/src/translator/code_translator.rs
syrusakbary: It will be useful from the future (is directly ported from cranelift-wasm), that's why is useful keeping it!

lib/compiler-cranelift/src/translator/code_translator.rs
syrusakbary: It will be useful for the future (is directly ported from cranelift-wasm, but can't be used just yet since it requires for Cranelift to publish a new version), that's why is useful keeping it!

lib/compiler-cranelift/src/translator/code_translator.rs
syrusakbary: It will be useful for the future (is directly ported from cranelift-wasm, but can't be used just yet since it requires for Cranelift to publish a new version), that's why is useful keeping it!

lib/compiler-cranelift/src/func_environ.rs
syrusakbary: It will be useful when we add reference types. I think it will be useful to keep it!

lib/compiler-cranelift/src/translator/code_translator.rs
syrusakbary: It's directly ported from cranelift-wasm.

PR #1783, opened 2020-10-30
feat(c-api) Handle initialized but empty results in `wasm_func_call`

Hywan:
# Description

Our implementation of `wasm_func_call` was correct for C code as
follows:

```c
wasm_val_vec_t arguments = WASM_EMPTY_VEC;
wasm_val_vec_t results = WASM_EMPTY_VEC;
wasm_func_call(func, &arguments, &results);
```

However, for a C code such as:

```c
wasm_val_t vals[1];
wasm_val_vec_t arguments = WASM_EMPTY_VEC;
wasm_val_vec_t results = WASM_ARRAY_VEC(vals);
wasm_func_call(func, &arguments, &results);
```

the `vals` array were kept empty/unchanged. Why?

Because `wasm_func_call` was replacing the value of `results` by a new
`wasm_val_vec_t`. It is correct when `results` is an empty vector, but
it is incorrect when `results` is initialized with empty values.

This patch tries to detect this pattern: If `results.data` is `null`,
it means the vector is empty/uninitialized, and we can set a new
`wasm_val_vec_t`, otherwise it means the vector is initialized with
empty values, and we need to update each item individually.

This pattern can be found in the `global.c` example of the Wasm C API: https://github.com/wasmerio/wasmer/blob/591691bc17795e0f547a06bd8227ecc37f9aec39/lib/c-api/tests/wasm_c_api/wasm-c-api/example/global.c#L40-L47

Without this patch, this test fails.

# Review

### DIFF
--- a/lib/c-api/src/wasm_c_api/externals/function.rs
+++ b/lib/c-api/src/wasm_c_api/externals/function.rs
@@ -167,7 +167,7 @@ pub unsafe extern "C" fn wasm_func_call(
                 .expect("Results conversion failed");
 
             // `results` is an uninitialized vector. Set a new value.
-            if results.size == 0 || results.data.is_null() {
+            if results.is_uninitialized() {
                 *results = vals.into();
             }
             // `results` is an initialized but empty vector. Fill it
--- a/lib/c-api/src/wasm_c_api/externals/function.rs
+++ b/lib/c-api/src/wasm_c_api/externals/function.rs
@@ -154,18 +154,34 @@ pub unsafe extern "C" fn wasm_func_call(
                 .into_iter()
                 .map(TryInto::try_into)
                 .collect::<Result<Vec<Val>, _>>()
-                .expect("Argument conversion failed")
+                .expect("Arguments conversion failed")
         })
         .unwrap_or_default();
 
     match func.inner.call(&params) {
         Ok(wasm_results) => {
-            *results = wasm_results
+            let vals = wasm_results
                 .into_iter()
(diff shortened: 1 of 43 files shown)


Hywan:
Opened question: Do we need to error when the size of `results` is not equal to the number of results?

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -154,18 +154,34 @@ pub unsafe extern "C" fn wasm_func_call(
...
+            let vals = wasm_results
                 .into_iter()
                 .map(TryInto::try_into)
                 .collect::<Result<Vec<wasm_val_t>, _>>()
-                .expect("Argument conversion failed")
-                .into();
+                .expect("Results conversion failed");
+
+            // `results` is an uninitialized vector. Set a new value.
+            if results.size == 0 || results.data.is_null() {
```
Hywan: Should we implement a function like `is_uninitialized` function for `wasm_$name_vec_t`? I believe yes.

Hywan: Done.

Hywan:
bors try

lib/c-api/src/wasm_c_api/externals/function.rs
jubianchi: I think having such a function would make our intentions clear in the code.

jubianchi (approved):

PR #1787, opened 2020-10-30
We're happy with the name "Global".

nlewycky:


jubianchi (approved):

lib/vm/src/global.rs
```diff
@@ -6,10 +6,9 @@ use thiserror::Error;
 use wasmer_types::{GlobalType, Mutability, Type, Value};
 
 #[derive(Debug)]
-/// TODO: figure out a decent name for this thing
```
syrusakbary, line 9:
```suggestion
/// A Global instance
```

nlewycky: github doesn't let me commit suggestions that replace deleted lines. :(

PR #1789, opened 2020-10-30
Make storage of globals use a union and make getters for it safe.

nlewycky:


lib/vm/src/vmcontext.rs
```diff
@@ -421,14 +422,42 @@ mod test_vmtable_definition {
...
+#[derive(Clone, Copy)]
+#[repr(C, align(16))]
+pub union VMGlobalDefinitionStorage {
+    as_i32: i32,
+    as_u32: u32,
+    as_f32: f32,
+    as_i64: i64,
+    as_u64: u64,
+    as_f64: f64,
+    as_u128: u128,
```
syrusakbary, line 441: I'm very confident this not what we want.
This will duplicate the data in the structure.

If you change bytes, all the representations shall change.

lib/vm/src/vmcontext.rs
```diff
@@ -421,14 +422,42 @@ mod test_vmtable_definition {
...
+    bytes: [u8; 16],
+}
```
MarkMcCaskey, line 443: Seems like a reasonable change!  The one thing I'd mention is that I believe Wasm numbers are always Little Endian, I'd suspect this to do the wrong thing on a Big Endian machine

lib/vm/src/vmcontext.rs
syrusakbary: That means the `as_i32`, `as_...` should just be views over the provided storage

nlewycky: This is a union, not a struct. The in-memory representation hasn't changed. The generated code from the compilers still works fine.

syrusakbary: What kind of magic trick is this one?

lib/vm/src/vmcontext.rs
```diff
@@ -469,207 +498,89 @@ mod test_vmglobal_definition {
...
     /// Return a reference to the value as an f32.
-    ///
-    /// # Safety
-    ///
-    /// To be defined (TODO).
-    #[allow(clippy::cast_ptr_alignment)]
-    pub unsafe fn as_f32(&self) -> &f32 {
-        &*(self.storage.as_ref().as_ptr() as *const f32)
+    pub fn as_f32(&self) -> &f32 {
+        unsafe { &self.storage.as_f32 }
```
MarkMcCaskey, line 548: Is this really safe? Is it possible that the bit representation of the `f32` could be invalid (I know Wasm NaNs are a bit specific...) if another type was inside?

lib/vm/src/vmcontext.rs
```diff
@@ -469,207 +498,89 @@ mod test_vmglobal_definition {
...
 
     /// Return a reference to the value as an u128.
-    ///
-    /// # Safety
-    ///
-    /// To be defined (TODO).
-    #[allow(clippy::cast_ptr_alignment)]
-    pub unsafe fn as_u128(&self) -> &u128 {
-        &*(self.storage.as_ref().as_ptr() as *const u128)
+    pub fn as_u128(&self) -> &u128 {
```
MarkMcCaskey, line 567: These being safe means we could alias memory with references to the same data but different types 🤔 ; I have no idea if that's okay or not 😆 , something to look into before shipping

lib/vm/src/vmcontext.rs
nlewycky: A `union` is `mem::transmute` in disguise. The union has the size of the largest of its members (not the sum) and the alignment of the max of its members. So all of these are sharing the same [u8; 16] space (though some of them like i32 are smaller and only using part of the space, but that's fine). As long as it's safe to transmute -- which it is because these are all primitive data types with no possible invalid values and because we initialized the memory -- it's safe read or write through any of these at any time.

MarkMcCaskey (approved):
The one other thing I'd like to consider is how `VMGlobalDefinitionStorage` will be accessed:

is `&mut self` really a proper guarantee with this type?  I mean if it's not that's technically not your problem, the code below looks probably sound to me, but for practical purposes it'd be good to take a look at the caller's and verify that `&mut self` actually implies exclusivity.

Oh one other thing, holding a reference, mutable or not, to Global data could possibly cause problems if that is then mutated 🤔 by Wasm or possibly another global...

I suspect, we could probably do

```Rust
let global1 = instance.exports.get("my_global").unwrap();
let global2 = instance.exports.get("my_global").unwrap();

let num_ref = global1.something.as_u32();
*global2.something.as_i32_mut() = 0;
```

Which would be bad...

lib/vm/src/vmcontext.rs
nlewycky: Ah ok, that's no good then. I'll move the `unsafe`'s back to the function declarations.

lib/vm/src/vmcontext.rs
nlewycky: IEEE floats have no invalid bit patterns, just NaNs. The one I needed to check was `i32` and `i64`. In C89 signed integers are permitted to have one invalid representation which isn't used on 2's complement machines. I decided it must be OK because transmute is used in the standard library to implement from_ne_bytes.

lib/vm/src/vmcontext.rs
nlewycky: I think the intention is that the caller always use the correct accessor type, like as_i32() / as_i32_mut() if the wasm type of the global is I32. On the other hand, it's not rust-unsafe to switch up the types, it just may produce a surprising value.

lib/vm/src/vmcontext.rs
MarkMcCaskey: It can still be safe as long as it takes a `&mut self` that will guarantee exclusivity!

MarkMcCaskey: Assuming sound Rust

nlewycky: This is now `pub fn to_u128(&self) -> u128 {`.

nlewycky:
> Oh one other thing, holding a reference, mutable or not, to Global data could possibly cause problems if that is then mutated  by Wasm or possibly another global...

The `Safety` section now mentions the requirement that all accesses (not just writes, the reads too) have to go through the mutable borrow as long as it exists.

MarkMcCaskey (approved):

syrusakbary:
bors try

syrusakbary (commented):
Before merging I think it would be good for @Hywan to review the PR as well

lib/vm/src/vmcontext.rs
Hywan: It's not strictly unsafe, but yeah, as you said, it produces surprising values if the type isn't correctly selected.

lib/vm/src/instance.rs
```diff
@@ -1163,9 +1163,9 @@ fn get_memory_init_start(init: &DataInitializer<'_>, instance: &Instance) -> usi
     if let Some(base) = init.location.base {
         let val = unsafe {
             if let Some(def_index) = instance.module.local_global_index(base) {
-                *instance.global(def_index).as_u32()
+                instance.global(def_index).to_u32()
```
Hywan, line 1166: I'm surprised we use a `u32` here. I understand that we need a `u32`, but Wasm doesn't support `u32`, only `i32`. Did we “cheat” by storing random/non-standard bytes in a Global?

lib/vm/src/vmcontext.rs
```diff
@@ -421,14 +422,42 @@ mod test_vmtable_definition {
...
+
 /// The storage for a WebAssembly global defined within the instance.
 ///
 /// TODO: Pack the globals more densely, rather than using the same size
 /// for every type.
 #[derive(Debug, Clone)]
 #[repr(C, align(16))]
 pub struct VMGlobalDefinition {
-    storage: [u8; 16],
+    storage: VMGlobalDefinitionStorage,
```
Hywan, line 460: How hard would it be to store the `Global`'s type here? That way we could panic if the requested type isn't appropriate (e.g. casting a `f32` to `i32`).

Hywan (approved):
I like the `union` approach. It makes the intent clearer. Thanks for that!

Few notes:

* I'm surprised to see the `u32` and `u64` types: they aren't standard; are we cheating by storing non-standard bytes in a Global?
* Also, I'm wondering whether we should store the Global's type inside `VMGlobalDefinition`. That would make sense actually, it's part of its definition to me. It would also help to panic if the requested type isn't appropriate. Bonus: it would also help to build `Value` with a simple `From` implementation. Note: It seems that `Global` already fulfills this role by holding a `VMGlobalDefinition` + a `GlobalType`.

Anyway, the patch is good as is. My review opens new questions that can be addressed by further patches.

lib/vm/src/instance.rs
nlewycky: Wasm's `I32` is neither rust's `i32` nor rust's `u32`. It is 32 bits long, but it doesn't include whether the number is signed or unsigned in a division, for example, while rust's i32/u32 does.

lib/vm/src/vmcontext.rs
nlewycky: It really doesn't belong here. The VM* structs are an API to the compiler-generated code, and the type doesn't need to include the type since it's fixed when the wasm is being compiled.

jubianchi (approved):

lib/vm/src/instance.rs
Hywan: OK.

lib/vm/src/vmcontext.rs
Hywan: I'm good with that.

PR #1791, opened 2020-10-30
Document Send and Sync for ExportFunction.

nlewycky:
In passing, improve comments on VMFunctionKind.


MarkMcCaskey (approved):
Seems reasonable to me

PR #1795, opened 2020-11-02
test(c-api) Introduce inline-c-rs, and use it!

Hywan:
# Description

This patch uses the new [`inline-c-rs` crate](https://github.com/Hywan/inline-c-rs/). Since it is immature for the moment, we use the `master` branch of this crate for the moment. A release should follow quickly.

A new helper function has been added to the `wasmer_wasm.h` file: `wasm_byte_vec_new_from_string`, which is similar to `wasm_name_new_from_string`.

Unit testing the code has allowed to find bugs, which are corrected by this patch.

Unit tests for the `wat` and the `module` modules have been written.

Bonus: 2 memory leaks have been fixed inside this PR. It would have been easier in other PR retrospectively, sorry for the reviewers.

# Review


lib/c-api/Cargo.toml
```diff
@@ -39,6 +39,9 @@ paste = "0.1"
 # Commented out for now until we can find a solution to the exported function problem
 # wasmer-wasm-c-api = { version = "1.0.0-alpha4", path = "crates/wasm-c-api" }
 
+[dev-dependencies]
+inline-c = { git = "https://github.com/Hywan/inline-c-rs", branch = "master" }
```
jubianchi, line 43: Any plan on having this released?

Asking this because we may want to have guarantees on the API being stable over time.

jubianchi (commented):
I think this kind of tool will help us write C test in a more rust-fashion.

Having a single tool to run them (i.e `cargo test`) is a good thing I think. Also, being able to have the tests near the code they test feels more like the Rust style and helps reading/maintaining tests.

lib/c-api/Cargo.toml
Hywan: Sure! For the moment, `inline-c-rs` just assume `clang` is installed and uses only that. I'm planning to move to [`cc-rs`](https://github.com/alexcrichton/cc-rs/) if I can hack the lib to do what I want, i.e. to compile for all platforms. (ping @alexcrichton)

Hywan: Done. I'm releasing the first version of `inline-c-rs`.

Hywan:
bors try

Hywan:
I'm digging into more ownership and leaking issues. It will make this PR less easy to review. I propose to merge this PR if tests are OK, and then open a new PR to investigate more ownership and leaking issues. Thoughts @nlewycky and @jubianchi?

lib/c-api/src/wasm_c_api/types/function.rs
```diff
@@ -97,7 +96,6 @@ pub unsafe extern "C" fn wasm_functype_params(ft: &wasm_functype_t) -> *const wa
     out as *const _
 }
 
-// TODO: fix memory leak
```
Hywan, line 100: For those deletions, take a look at https://github.com/wasmerio/wasmer/pull/1795/commits/8ae283507e4912ef9d56a11f18e3edde867825b0.

lib/c-api/build.rs
```diff
@@ -389,3 +390,19 @@ fn exclude_items_from_wasm_c_api(builder: Builder) -> Builder {
         .exclude_item("wasmer_engine_t")
         .exclude_item("wat2wasm")
 }
+
+fn build_inline_c_env_vars() {
+    let mut shared_object_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+    shared_object_dir.pop(); // removes `c-api`
+    shared_object_dir.pop(); // removes `lib`
```
nlewycky, line 397: Optional:
```suggestion
    assert_eq!(shared_object_dir.pop(), "c-api".into());
    assert_eq!(shared_object_dir.pop(), "lib".into());
```
(note: code not actually tested) The idea is to check this so that it fails noisily instead of silently causing surprises down the line.

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -115,3 +115,364 @@ pub unsafe extern "C" fn wasm_module_serialize(
...
+
+            int main() {
+                wasm_engine_t* engine = wasm_engine_new();
+                wasm_store_t* store = wasm_store_new(engine);
+
+                wasm_byte_vec_t wat;
+                wasm_byte_vec_new_from_string(&wat, "(module)");
+                wasm_byte_vec_t* wasm = wat2wasm(&wat);
+
+                assert(wasm_module_validate(store, wasm));
```
nlewycky, line 136: Unlike `assert!` in rust, this does not run `wasm_module_validate` in a release build. The assert() macro is defined to nothing in release builds. Therefore this won't test module_validate.

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -115,3 +115,364 @@ pub unsafe extern "C" fn wasm_module_serialize(
...
+    }
+
+    #[test]
+    fn test_module_exports() {
+        (assert_c! {
+            #include <string.h>
+            #include "tests/wasmer_wasm.h"
+
+            void assert_exporttype_name(const wasm_exporttype_t* exporttype, const char* expected) {
+                assert(strncmp(wasm_exporttype_name(exporttype)->data, expected, strlen(expected)) == 0);
```
nlewycky, line 181: This is just strcmp. strncmp + strlen isn't any safer than strcmp. Unlike rust strings where strings store the length before the string data, C strings run until the first NUL byte. Strlen will just run off the end, same as strcmp would.

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -115,3 +115,364 @@ pub unsafe extern "C" fn wasm_module_serialize(
...
+                    wasm_functype_t* func_type = wasm_externtype_as_functype((wasm_externtype_t*) extern_type);
+
+                    const wasm_valtype_vec_t* func_params = wasm_functype_params(func_type);
+                    assert(func_params->size == 2);
+                    assert(wasm_valtype_kind(func_params->data[0]) == WASM_I32);
+                    assert(wasm_valtype_kind(func_params->data[1]) == WASM_I64);
+                    wasm_valtype_vec_delete((wasm_valtype_vec_t*) func_params);
+
+                    const wasm_valtype_vec_t* func_results = wasm_functype_results(func_type);
+                    assert(func_results->size == 0);
```
nlewycky, line 224: Note that this will generate an unused variable warning when built in release mode. Consider adding `(void)func_results;`.

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -115,3 +115,364 @@ pub unsafe extern "C" fn wasm_module_serialize(
...
+                assert(export_types.size == 4);
+
+                {
+                    wasm_exporttype_t* export_type = export_types.data[0];
+                    assert_exporttype_name(export_type, "function");
+
+                    const wasm_externtype_t* extern_type = wasm_exporttype_type(export_type);
+                    assert(wasm_externtype_kind(extern_type) == WASM_EXTERN_FUNC);
+
+                    wasm_functype_t* func_type = wasm_externtype_as_functype((wasm_externtype_t*) extern_type);
```
nlewycky, line 215:
```suggestion
                    const wasm_functype_t* func_type = wasm_externtype_as_functype_const(extern_type);
```

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -115,3 +115,364 @@ pub unsafe extern "C" fn wasm_module_serialize(
...
+
+                    const wasm_valtype_vec_t* func_params = wasm_functype_params(func_type);
+                    assert(func_params->size == 2);
+                    assert(wasm_valtype_kind(func_params->data[0]) == WASM_I32);
+                    assert(wasm_valtype_kind(func_params->data[1]) == WASM_I64);
+                    wasm_valtype_vec_delete((wasm_valtype_vec_t*) func_params);
+
+                    const wasm_valtype_vec_t* func_results = wasm_functype_results(func_type);
+                    assert(func_results->size == 0);
+                    wasm_valtype_vec_delete((wasm_valtype_vec_t*) func_results);
```
nlewycky, line 225: > ### wasm_functype_results()
> const wasm_valtype_vec_t* wasm_functype_results(const wasm_functype_t*)
> Returns the list of results of this function type.
> The returned memory is owned by the wasm_functype_t argument, the caller should not deallocate it.
 -- https://docs.wasmtime.dev/c-api/wasm_8h.html#a5ca8dc78814cb58f10be35b92e615974

Notably this is const-incorrectness, you ordinarily shouldn't free a const pointer. It would indicate either a bug in our code or in wasm.h.

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -115,3 +115,364 @@ pub unsafe extern "C" fn wasm_module_serialize(
...
+    }
+
+    #[test]
+    fn test_module_imports() {
+        (assert_c! {
+            #include <string.h>
+            #include "tests/wasmer_wasm.h"
+
+            void assert_importtype_name(const wasm_name_t* name, const char* expected) {
+                assert(strncmp(name->data, expected, strlen(expected)) == 0);
```
nlewycky, line 285:
```suggestion
                assert(strncmp(name->data, expected, name->size) == 0 && name->size == strlen(expected));
```
If `expected` is shorter than `name`, then the strncmp will stop at the NUL byte terminating `expected`, and the compare will fail, short-circuit evaluation skips the strlen part as the assert fires.
If `name` is shorter than `expected` and they match all the way through to name's end, the strlen will fail.

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -115,3 +115,364 @@ pub unsafe extern "C" fn wasm_module_serialize(
...
+                wasm_engine_delete(engine);
+            }
+        })
+        .success();
+    }
+
+    #[test]
+    fn test_module_exports() {
+        (assert_c! {
+            #include <string.h>
```
nlewycky, line 177: Consider moving this into tests/wasmer_wasm.h. That header file will include everything our tests need and keeps the test code clean.

lib/c-api/wasmer_wasm.h
```diff
@@ -29,6 +29,9 @@
 #  define DEPRECATED(message) __declspec(deprecated(message))
 #endif
 
+// The `jit` feature has been enabled for this build.
+#define WASMER_JIT_ENABLED
+
```
nlewycky, line 34: ?

lib/c-api/tests/wasmer_wasm.h
```diff
@@ -0,0 +1,13 @@
+// This header file is used only for test purposes! It is used by unit
+// test inside the `src/` directory for the moment.
+
+#include <assert.h>
+#include "../wasmer_wasm.h"
+
+// Wasmer-specific shortcut to quickly create a `wasm_byte_vec_t` from
+// a string.
+static inline void wasm_byte_vec_new_from_string(
```
nlewycky, line 9: Perhaps our own functions shouldn't use the standard `wasm_` prefix? Maybe `wasmer_`?

lib/c-api/src/wasm_c_api/wat.rs
```diff
@@ -12,3 +12,73 @@ pub unsafe extern "C" fn wat2wasm(wat: &wasm_byte_vec_t) -> Option<Box<wasm_byte
...
+            #include "tests/wasmer_wasm.h"
+
+            int main() {
+                wasm_engine_t* engine = wasm_engine_new();
+                wasm_store_t* store = wasm_store_new(engine);
+
+                wasm_byte_vec_t wat = {
+                    .data = "(module",
+                    .size = 7,
+                };
```
nlewycky, line 70:
```suggestion
                wasm_byte_vec_t wat;
                wasm_byte_vec_new_from_string(&wat, "(module");
```

lib/c-api/src/wasm_c_api/wat.rs
```diff
@@ -12,3 +12,73 @@ pub unsafe extern "C" fn wat2wasm(wat: &wasm_byte_vec_t) -> Option<Box<wasm_byte
...
+            #include "tests/wasmer_wasm.h"
+
+            int main() {
+                wasm_engine_t* engine = wasm_engine_new();
+                wasm_store_t* store = wasm_store_new(engine);
+
+                wasm_byte_vec_t wat = {
+                    .data = "(module)",
+                    .size = 8,
+                };
```
nlewycky, line 32:
```suggestion
                wasm_byte_vec_t wat;
                wasm_byte_vec_new_from_string(&wat, "(module)");
```

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -115,3 +115,364 @@ pub unsafe extern "C" fn wasm_module_serialize(
...
+        .success();
+    }
+
+    #[test]
+    fn test_module_imports() {
+        (assert_c! {
+            #include <string.h>
+            #include "tests/wasmer_wasm.h"
+
+            void assert_importtype_name(const wasm_name_t* name, const char* expected) {
```
nlewycky, line 284: `static inline`?

lib/c-api/build.rs
Hywan: That's an excellent idea!

lib/c-api/src/wasm_c_api/module.rs
Hywan: Hmm, so we'll have to write our own `assert` :-).

lib/c-api/src/wasm_c_api/module.rs
Hywan: Maybe we should use `wasm_exporttype_name(exporttype)->size` instead of `strlen`. Thoughts?

lib/c-api/src/wasm_c_api/module.rs
Hywan: Except if we write our own `assert` :-D!

lib/c-api/src/wasm_c_api/module.rs
Hywan: I wanted to report that to the WG, because `wasm_functype_results` actually returns a `const *` here.

lib/c-api/tests/wasmer_wasm.h
Hywan: Good idea.

lib/c-api/src/wasm_c_api/module.rs
Hywan: Good idea.

Hywan:
How are we supposed to do that? We can't track that… My understanding is that the caller is responsible to free the value generated by `wasm_functype_(params|results)`, it's the opposite.

nlewycky:
> How are we supposed to do that? We can't track that… My understanding is that the caller is responsible to free the value generated by `wasm_functype_(params|results)`, it's the opposite.

As far as I can tell the only way to implement is to make it a pure getter that returns a pointer to the params that already exists inside the functype. The params and results are filled in when the C-API functtype object is created, always.

lib/c-api/build.rs
Hywan: `pop` returns `bool`. The solution is to do this:

```
assert_eq!(shared_object_dir.file_name(), Some(OsStr::new("c-api")));
shared_object_dir.pop();
```

and so on.

lib/c-api/src/wasm_c_api/module.rs
Hywan: Agreed. Done!

lib/c-api/src/wasm_c_api/module.rs
Hywan: Our own `wasmer_assert` macro has been implemented. It mimics the standard `assert` macro, with a different name, to avoid confusion.

jubianchi (dismissed):
Two comments from @nlewycky left to address but otherwise, seems good to me.

lib/c-api/src/wasm_c_api/module.rs
Hywan: OK, the `wasm.h` declares the functions as is:

```c
WASM_API_EXTERN const wasm_valtype_vec_t* wasm_functype_params(const wasm_functype_t*);
WASM_API_EXTERN const wasm_valtype_vec_t* wasm_functype_results(const wasm_functype_t*);
```

The output result of both functions has no `own`, so the ownership is still to `wasm_functype_t`. We need to fix our code.

Hywan: Those memory leaks are fixed!

lib/c-api/src/wasm_c_api/module.rs
Hywan: Done, thanks.

lib/c-api/wasmer_wasm.h
Hywan: It was missing in the `wasmer_wasm.h` file. We should provide the `jit` feature.

lib/c-api/src/wasm_c_api/wat.rs
Hywan: Done.

Hywan:
bors try

Makefile
```diff
@@ -90,6 +90,8 @@ build-docs:
 
 build-docs-capi:
 	cd lib/c-api/ && doxygen doxyfile
+	RUSTDOCFLAGS="" \
```
nlewycky, line 93: Why does this need to be set?

Makefile
```diff
@@ -90,6 +90,8 @@ build-docs:
...
+		cargo doc --manifest-path lib/c-api/Cargo.toml --no-deps --features wat,jit,object-file,cranelift,wasi
```
nlewycky, line 94: Please include native here?

lib/c-api/src/wasm_c_api/module.rs
nlewycky: This suggestion should be possible now that we no longer delete the func_Type? (And the same thing for the rest of the blocks in this test.)

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -14,6 +14,48 @@ pub struct wasm_module_t {
...
+/// int main() {
+///     wasm_engine_t* engine = wasm_engine_new();
+///     wasm_store_t* store = wasm_store_new(engine);
+///    
+///     wasm_byte_vec_t wat;
+///     wasmer_byte_vec_new_from_string(&wat, "(module)");
+///     wasm_byte_vec_t* wasm = wat2wasm(&wat);
+///    
+///     wasm_module_t* module = wasm_module_new(store, wasm);
+///
```
nlewycky, line 45: Optional: this extra newline here looks odd to me.

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -115,3 +157,376 @@ pub unsafe extern "C" fn wasm_module_serialize(
...
+                    "(module\n"
+                    "  (func (export \"function\") (param i32 i64))\n"
+                    "  (global (export \"global\") i32 (i32.const 7))\n"
+                    "  (table (export \"table\") 0 funcref)\n"
+                    "  (memory (export \"memory\") 1))"
+                );
+                wasm_byte_vec_t* wasm = wat2wasm(&wat);
+
+                wasm_module_t* module = wasm_module_new(store, wasm);
+
```
nlewycky, line 239: Optional: here too.

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -115,3 +157,376 @@ pub unsafe extern "C" fn wasm_module_serialize(
...
+                    const wasm_name_t* import_name = wasm_importtype_name(import_type);
+                    wasmer_assert_name(import_name, "function");
+
+                    const wasm_externtype_t* extern_type = wasm_importtype_type(import_type);
+                    assert(wasm_externtype_kind(extern_type) == WASM_EXTERN_FUNC);
+
+                    wasm_functype_t* func_type = wasm_externtype_as_functype((wasm_externtype_t*) extern_type);
+
+                    const wasm_valtype_vec_t* func_params = wasm_functype_params(func_type);
+                    assert(func_params->size == 0);
```
nlewycky, line 374: Optional: consider if you want `assert(func_params && func_params->size == 0);`.

lib/c-api/src/wasm_c_api/module.rs
```diff
@@ -115,3 +157,376 @@ pub unsafe extern "C" fn wasm_module_serialize(
...
+    #[test]
+    fn test_module_serialize_and_deserialize() {
+        (assert_c! {
+            #include "tests/wasmer_wasm.h"
+
+            int main() {
+                wasm_engine_t* engine = wasm_engine_new();
+                wasm_store_t* store = wasm_store_new(engine);
+
+                wasm_byte_vec_t wat;
```
nlewycky, line 492: This isn't deleted.

I see that `lib/c-api/tests/wasm_c_api/test-wat2wasm.c` doesn't delete it either, but I can't see how that isn't just a leak? If `wat2wasm` takes ownership, then that fact needs to be added to the function comment in `wasmer_wasm.h`.

(A tip, any program you've built can be run with `valgrind program` to see leaks on exit.)

lib/c-api/src/wasm_c_api/types/extern_.rs
```diff
@@ -75,9 +36,81 @@ impl From<&ExternType> for wasm_externkind_enum {
...
+
 #[no_mangle]
-pub unsafe extern "C" fn wasm_externtype_kind(et: &wasm_externtype_t) -> wasm_externkind_t {
-    wasm_externkind_enum::from(&et.inner) as wasm_externkind_t
+pub unsafe extern "C" fn wasm_extern_kind(r#extern: &wasm_extern_t) -> wasm_externkind_t {
+    wasm_externkind_enum::from(r#extern.inner.ty()) as wasm_externkind_t
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn wasm_externtype_delete(_et: Option<Box<wasm_externtype_t>>) {}
```
nlewycky, line 97: Optional: FWIW I'd just call it `_` in this case.

lib/c-api/tests/wasmer_wasm.h
```diff
@@ -0,0 +1,26 @@
+// This header file is used only for test purposes! It is used by unit
+// test inside the `src/` directory for the moment.
+
+#if !defined(TEST_WASMER_WASM)
+
+#define TEST_WASMER_WASM
```
nlewycky, line 6:
```suggestion
#ifndef TEST_WASMER_WASM
#define TEST_WASMER_WASM
```

lib/c-api/tests/wasmer_wasm.h
```diff
@@ -0,0 +1,26 @@
...
+void wasmer_assert_name(const wasm_name_t* name, const char* expected) {
+  assert(name->size == strlen(expected) && strncmp(name->data, expected, name->size) == 0);
+}
+
+
+// Helper to quickly create a `wasm_byte_vec_t` from a string, à la
+// `wasm_name_new_from_string`.
+static inline void wasmer_byte_vec_new_from_string(
+  wasm_byte_vec_t* out, const char* s
+) {
```
nlewycky, line 22: Consider running `clang-format -i filename` over this file. (`-i` is in-place editing.)

nlewycky (approved):

Makefile
Hywan: Done.

Makefile
Hywan: Done.

lib/c-api/src/wasm_c_api/module.rs
Hywan: Done.

lib/c-api/src/wasm_c_api/module.rs
Hywan: Done.

lib/c-api/src/wasm_c_api/module.rs
Hywan: Correct!

lib/c-api/src/wasm_c_api/module.rs
Hywan: Done, thanks!

lib/c-api/src/wasm_c_api/module.rs
Hywan: `wat2wasm` must not take ownership of the bytes. I'm fixing that. Thanks!

lib/c-api/src/wasm_c_api/types/extern_.rs
Hywan: I like being super explicit. In this case, I would rename it to `_extern_type` actually :-D.

lib/c-api/tests/wasmer_wasm.h
Hywan: Done, thanks.

Hywan:
bors try

jubianchi (approved):

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

lib/c-api/Cargo.toml
```diff
@@ -92,4 +92,4 @@ cranelift-backend = ["cranelift"]
 llvm-backend = ["llvm"]
 
 [build-dependencies]
-cbindgen = "0.15"
+cbindgen = "0.15"
```
jubianchi, line 95: Newline?

Hywan: Yup.

nlewycky (changes requested):
(unapproving for `rpath`.)

jubianchi:
bors try

lib/c-api/build.rs
```diff
@@ -419,12 +419,15 @@ fn build_inline_c_env_vars() {
         L = shared_object_dir.clone(),
     );
 
-    // Add `-lwasmer_c_api` and, on Linux,
-    // `-rpath=<shared_object_dir>`, to the linker argument.
     println!(
-        "cargo:rustc-env=INLINE_C_RS_LDFLAGS=-lwasmer_c_api,{rpath}",
-        rpath = if cfg!(target_os = "linux") {
-            format!("-rpath={}", shared_object_dir)
+        "cargo:rustc-env=INLINE_C_RS_LDFLAGS={shared_object_dir}/libwasmer_c_api.{ext}",
```
syrusakbary, line 423: In the case of windows I believe is `wasmer_c_api.dll` (without the lib)

jubianchi:
bors try-

lib/c-api/build.rs
```diff
@@ -419,14 +419,20 @@ fn build_inline_c_env_vars() {
...
-        },
+            format!(
+                "libwasmer_c_api.{ext}",
+                ext = if cfg!(target_os = "macos") {
+                    "dylib".to_string()
+                } else {
+                    "so".to_string()
+                }
+            )
+        }
```
syrusakbary: This can be much simpler this way:

```suggestion
        lib = if cfg!(target_os = "windows") {
            "wasmer_c_api.dll".to_string()
        } else if cfg!(target_os = "macos") {
            "libwasmer_c_api.dylib".to_string()
        }
        else {
            "libwasmer_c_api.so".to_string()
        }
```

lib/c-api/build.rs
jubianchi: fixed!

— https://github.com/wasmerio/wasmer/pull/1795/commits/6c9a487f2214719dee050638fd96229d8f88c09b

jubianchi:
bors try

lib/c-api/Cargo.toml
```diff
@@ -39,6 +39,9 @@ paste = "0.1"
 # Commented out for now until we can find a solution to the exported function problem
 # wasmer-wasm-c-api = { version = "1.0.0-alpha5", path = "crates/wasm-c-api" }
 
+[dev-dependencies]
+inline-c = { git = "https://github.com/Hywan/inline-c-rs", branch = "fix-linux-shared-object" }
```
syrusakbary, line 43: Can we use a published version of the crate?

jubianchi: Yes we will. We are waiting for the last bits of https://github.com/Hywan/inline-c-rs/pull/4 before releasing the crate.

lib/c-api/build.rs
```diff
@@ -389,3 +390,44 @@ fn exclude_items_from_wasm_c_api(builder: Builder) -> Builder {
...
+    let shared_object_dir = shared_object_dir.as_path().to_string_lossy();
+    let include_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+
+    // The following options mean:
+    //
+    // * `-I`, add `include_dir` to include search path,
+    // * `-L`, add `shared_object_dir` to library search path,
+    // * `-D_DEBUG`, enable debug mode to enable `assert.h`.
+    println!(
+        "cargo:rustc-env=INLINE_C_RS_CFLAGS=-I{I} -L{L} -D_DEBUG",
```
nlewycky, line 417: (No action required.) I think I see why this didn't work: `-L` isn't really a compiler flag, it's a linker flag. It'll be ignored when not linking, and since it's not passed along to the linker (not in `INLINE_C_RS_LDFLAGS`) it would cause `-lwasmer_c_api` to fail because the linker wouldn't know to search our shared_object_dir.

nlewycky (approved):

PR #1796, opened 2020-11-02
Make FunctionBodyPtr a thin pointer.

nlewycky:
Also document Send+Sync for FunctionBodyPtr.


Hywan (approved):
I appreciate the simplification of the code. Thanks!

Hywan:
bors try

lib/engine-object-file/src/artifact.rs
```diff
@@ -319,10 +319,12 @@ impl ObjectFileArtifact {
 
             byte_buffer[0..WORD_SIZE]
                 .clone_from_slice(&bytes[cur_offset..(cur_offset + WORD_SIZE)]);
+            let fp = usize::from_ne_bytes(byte_buffer);
```
syrusakbary: Can you explain what is going on here?

nlewycky: `bytes` is a slice of bytes, at some offset (`cur_offset`) into that stream of bytes, there is a pointer. We grab the pointer out of that stream of bytes and ultimately cast it to FunctionBodyPtr.

I don't know why we're doing this, but this is what the code was already doing. I could look into why if you want?

lib/engine/src/trap/frame_info.rs
```diff
@@ -227,18 +227,15 @@ impl Drop for GlobalFrameInfoRegistration {
 /// dropped, will be used to unregister all name information from this map.
 pub fn register(
     module: Arc<ModuleInfo>,
-    finished_functions: &BoxedSlice<LocalFunctionIndex, FunctionBodyPtr>,
+    finished_functions: &BoxedSlice<LocalFunctionIndex, (FunctionBodyPtr, usize)>,
```
syrusakbary, line 230: Can we have a struct that owns both the pointer and the length? (rather than an unnamed tuple)

nlewycky: Done

lib/engine-jit/src/engine.rs
```diff
@@ -193,7 +193,7 @@ impl JITEngineInner {
         custom_sections: &PrimaryMap<SectionIndex, CustomSection>,
     ) -> Result<
         (
-            PrimaryMap<LocalFunctionIndex, FunctionBodyPtr>,
+            PrimaryMap<LocalFunctionIndex, (FunctionBodyPtr, usize)>,
```
syrusakbary, line 196: Update this to use `FunctionExtent` ?

lib/engine-jit/src/link.rs
```diff
@@ -8,20 +8,17 @@ use wasmer_compiler::{
...
 use wasmer_types::LocalFunctionIndex;
 use wasmer_vm::ModuleInfo;
-use wasmer_vm::{FunctionBodyPtr, SectionBodyPtr, VMFunctionBody};
+use wasmer_vm::{FunctionBodyPtr, SectionBodyPtr};
 
 fn apply_relocation(
     body: usize,
     r: &Relocation,
-    allocated_functions: &PrimaryMap<LocalFunctionIndex, FunctionBodyPtr>,
+    allocated_functions: &PrimaryMap<LocalFunctionIndex, (FunctionBodyPtr, usize)>,
```
syrusakbary, line 16: Update this to use `FunctionExtent` ?

lib/engine-jit/src/link.rs
```diff
@@ -69,7 +65,7 @@ fn apply_relocation(
 /// required relocations and jump tables.
 pub fn link_module(
     _module: &ModuleInfo,
-    allocated_functions: &PrimaryMap<LocalFunctionIndex, FunctionBodyPtr>,
+    allocated_functions: &PrimaryMap<LocalFunctionIndex, (FunctionBodyPtr, usize)>,
```
syrusakbary, line 68: Update me!

lib/engine-jit/src/engine.rs
nlewycky: Right, sorry. Done.

lib/engine/src/trap/frame_info.rs
```diff
@@ -219,6 +219,16 @@ impl Drop for GlobalFrameInfoRegistration {
     }
 }
 
+/// Represents a continuous region of executable memory starting with a function
+/// entry point.
+pub struct FunctionExtent {
```
MarkMcCaskey, line 224: `#[derive(Debug)]` please!

MarkMcCaskey: Does this need to be `repr(C)`? 🤔

lib/engine/src/trap/frame_info.rs
```diff
@@ -227,18 +237,22 @@ impl Drop for GlobalFrameInfoRegistration {
...
+    for (
+        i,
+        FunctionExtent {
+            ptr: start,
+            length: len,
+        },
+    ) in finished_functions.iter()
+    {
+        let start = **start as usize;
+        let end = start + len;
```
MarkMcCaskey, line 255: Nice improvement!

I think the layout of `*const [T]` isn't guaranteed to be `ptr, len` so this is definitely an improvement in correctness

lib/vm/src/lib.rs
```diff
@@ -58,24 +58,26 @@ pub use crate::vmoffsets::{TargetSharedSignatureIndex, VMOffsets};
...
 
 /// # Safety
-/// TODO:
+/// The VMFunctionBody that this points to is opaque, so there's no data to
+/// read or write through this pointer. This is essentially a usize.
 unsafe impl Send for FunctionBodyPtr {}
 /// # Safety
-/// TODO:
+/// The VMFunctionBody that this points to is opaque, so there's no data to
+/// read or write through this pointer. This is essentially a usize.
```
MarkMcCaskey, line 80: It's surprising that `Send` isn't automatically implemented for types that are both `Sync` and `Copy`

MarkMcCaskey (approved):
Looks good overall to me!

lib/engine-object-file/src/artifact.rs
MarkMcCaskey: This code is parsing in metadata stored in the object file, it's creating types like `usize` from the buffer of bytes

lib/engine/src/trap/frame_info.rs
nlewycky: I don't think it needs to be repr(C) right now, but I'd like to use FunctionBodyPtr in the VM types so I'll go ahead and add it now.

PR #1799, opened 2020-11-06
Update this script so that it works on Linux.

nlewycky:
Make `fd` the default name of the program, but allow for the user to specify another, such as `FD=fdfind scripts/update-version.sh`.

Fix the sed command to take the filename to modify in-place and specify that the regex string is the sed-expression to execute.

# Review

 - [ ] Please check that this still works on a Mac.

### DIFF
--- a/scripts/update-version.sh
+++ b/scripts/update-version.sh
@@ -1,5 +1,6 @@
 #! /bin/sh
 
+# How to install `fd`: https://github.com/sharkdp/fd#installation
 : "${FD:=fd}"
 
 # A script to update the version of all the crates at the same time
--- a/scripts/update-version.sh
+++ b/scripts/update-version.sh
@@ -1,12 +1,16 @@
+#! /bin/sh
+
+: "${FD:=fd}"
+
 # A script to update the version of all the crates at the same time
 PREVIOUS_VERSION='1.0.0-alpha3'
 NEXT_VERSION='1.0.0-alpha4'
 
 # quick hack
-fd Cargo.toml --exec sed -i '' "s/version = \"$PREVIOUS_VERSION\"/version = \"$NEXT_VERSION\"/"
+${FD} Cargo.toml --exec sed -i '{}' -e "s/version = \"$PREVIOUS_VERSION\"/version = \"$NEXT_VERSION\"/"
 echo "manually check changes to Cargo.toml"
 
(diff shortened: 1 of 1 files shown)


scripts/update-version.sh
```diff
@@ -1,12 +1,16 @@
+#! /bin/sh
+
+: "${FD:=fd}"
```
syrusakbary, line 3: Do we really need this? (can we add a comment on how to install fd (via cargo and apt-get)

jubianchi: On MacOS, `fd` can also be installed using homebrew.

nlewycky: I added this because when you install `fd` with apt-get, the command is `fdfind`. I assume we need to support both names, but I kept `fd` as the default on the assumption that it was Debian renaming things. If `fdfind` is the popular name, I can make that the default. Or, we can remove the default and use `fdfind` unconditionally. Thoughts?

jubianchi: `fd` seems to be the right name to me. `fd-find` being the name of the crate which provides the `fd` CLI.

I think being able to change the name of the CLI is great to have. We ensure the script will work everywhere, whatever the path/name of the `fd` CLI is.

jubianchi, line 3:
```suggestion
: # How to install `fd`: https://github.com/sharkdp/fd#installation
: "${FD:=fd}"
```

jubianchi (approved):
Works on MacOS after installing `fd` and `gnu-sed` with `brew`.

nlewycky:
> Works on MacOS after installing `fd` and `gnu-sed` with `brew`.

It shouldn't need gnu-sed, bsd sed should work too? Does it not?

jubianchi:
@nlewycky unfortunately `-i` as a totally different meaning on BSD `sed` (https://www.freebsd.org/cgi/man.cgi?query=sed&sektion=&n=1)

syrusakbary (approved):

MarkMcCaskey:
@jubianchi good call here, the script now doesn't work on mac 😆 

I wrote this script as a hack in like 30 seconds a long time ago, maybe it's time to solve this problem properly 🤔 

As for `fd`, I can just never remember all the flags required to make `find` work in a sensible way, so I went with `fd` 😆 , if `fd` is a problem `find` should work just fine too

PR #1812, opened 2020-11-11
fix(ci): Test and then rebuild the C API

jubianchi:
Before this patch we were building the C API with the `build-capi`
target and then using the `test-capi` target to run the tests.

The `test-capi` runs the test suite against several builds of the C API
(different feature set each time).

After the tests, we packaged the C API. Conclusion: packaging the API
depended on the `test-capi` results and not on the `build-capi` results
thus avoiding us to ensure the packaged version is built with a known
feature set (adding a new test build/feature set could change the build
result).

With this patch, we first run the test (using `test-capi`) and then
rebuild the C API using `build-capi`. We know ensure the packaged
version will have a know feature set.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -143,8 +143,8 @@ jobs:
       - run: make test
       - name: Build and Test C API
         run: |
-          make build-capi
           make test-capi
+          make build-capi
         if: matrix.os != 'windows-latest'
       - name: Build C API on Windows
         run: make build-capi


jubianchi:
bors try

nlewycky:
I don't like the way these rules are non-hermetic, I think `make package-capi`'s behaviour should never depend on what's run before it. What would you think of making `package-capi` depend on `build-capi`?

jubianchi:
@nlewycky totally agree, let me do the change you suggested.

syrusakbary:
I think we should keep the PR as is.

The separation between build and packaging is completely indented.
It helps for people that want to build wasmer in a custom way, but still use the package subcommand.

syrusakbary (approved):

PR #1819, opened 2020-11-17
fix(ci): Regex correctly matches tags when doing release

jubianchi:
Before this patch, a branch name like `1.0.0-testing-something` was
considered as being the source of a release.

With this new regex we ensure we never use a wrong tag/branch to create
a release.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -26,7 +26,7 @@ jobs:
         run: |
           VERSION=${GITHUB_REF/refs\/tags\//}
           echo ::set-output name=VERSION::${VERSION}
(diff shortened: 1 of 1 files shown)


Hywan (approved):
I so much hate the regex syntax of `grep`. Good job handling that, thanks!

syrusakbary:
I'm not super sure about this.

Semver doesn't specify that only `alpha`, `beta`, `rc` subtags are valid. Semver allows any text there, and I think is better to fit to that?

jubianchi:
@syrusakbary you are right, I'll make the regex less strict on the pre-release tag while still giving a high level of security.

jubianchi:
Fixed the limits of the previous regular expression. It now supports more things: https://regex101.com/r/8crJQF/1

syrusakbary (approved):

PR #1824, opened 2020-11-18
Capi improvements

syrusakbary:
# Description

This PR improves:
* [x] Adds a `deprecated` feature to the wasm-c-api to ship the deprecated API (having `libffi` as a optional dependency)
* [x] Adds commands to build the wasm-c-api in headless mode

# Review

### DIFF
--- a/lib/c-api/src/lib.rs
+++ b/lib/c-api/src/lib.rs
@@ -30,6 +30,7 @@
     unreachable_patterns
 )]
 
+#[cfg(feature = "deprecated")]
 pub mod deprecated;
 pub mod error;
 mod ordered_resolver;
--- a/lib/c-api/src/wasm_c_api/engine.rs
+++ b/lib/c-api/src/wasm_c_api/engine.rs
@@ -13,6 +13,7 @@ use wasmer_engine_object_file::ObjectFile;
 ///
 /// This is a Wasmer-specific type with Wasmer-specific functions for
 /// manipulating it.
+#[cfg(feature = "compiler")]
 #[derive(Debug, Copy, Clone)]
 #[repr(C)]
 pub enum wasmer_compiler_t {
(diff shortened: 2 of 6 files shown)


lib/c-api/wasmer_wasm.h
```diff
@@ -32,9 +32,6 @@
 // The `jit` feature has been enabled for this build.
 #define WASMER_JIT_ENABLED
 
-// The `compiler` feature has been enabled for this build.
-#define WASMER_COMPILER_ENABLED
-
```
MarkMcCaskey, line 37: We may want to commit the version of the header file with this enabled

lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -72,6 +74,7 @@ impl Default for wasmer_engine_t {
 #[derive(Debug, Default)]
 #[repr(C)]
 pub struct wasm_config_t {
+    #[cfg(feature = "compiler")]
```
MarkMcCaskey, line 77: Minor comment, but it's a bit unfortunate that if the user's libwasmer and wasmer_wasm.h don't agree on the `compiler` feature, this will likely silently do the wrong thing because it overlaps with `engine` and uses the same numbers.

We can probably ignore this for now, but it'd be nice if there was some kind of way to fail at compile time or link time if the header differs from the library being used...

Actually in the case of building something for shared wasmer, this could actually be a problem...

One hacky fix for now is to move `engine` above `compiler` in `wasm_config_t`... then at least it won't confilct

MarkMcCaskey (approved):
Looks good to me!

lib/c-api/src/lib.rs
```diff
@@ -30,6 +30,7 @@
     unreachable_patterns
 )]
 
+#[cfg(feature = "deprecated")]
```
Hywan, line 33: Reminder: `cbindgen` ignores this. Is it OK for you?

PR #1826, opened 2020-11-19
feat(api) Faster `Exports::from_iter`

Hywan:
# Description

Instead of creating a new `Exports` with `Exports::new` and inserting
all pair `(String, Extern)` one after the other with
`Exports::insert`, this patch updates the code to create an `Exports`
directly with its `map` field built with `IndexMap::from_iter`, so
that we avoid calling `Arc::get_mut(…).unwrap().insert(…)` for every
pair.

# Review

- ~[ ] Add a short description of the the change to the CHANGELOG.md file~ (not sure it's necessary)


Hywan:
bors try

jubianchi (approved):

MarkMcCaskey (approved):
Looks good to me!

PR #1827, opened 2020-11-19
feat(vm) Don't clone the export in `Instance::lookup`

Hywan:
# Description

This patch does 2 things:

1. It simplifies the code by using `?` on an `Option`,
2. It then removes the need to clone the `export`.

# Review

- ~[ ] Add a short description of the the change to the CHANGELOG.md file~ (not sure this is necessary)

### DIFF
--- a/lib/vm/src/instance.rs
+++ b/lib/vm/src/instance.rs
@@ -284,11 +284,8 @@ impl Instance {
 
     /// Lookup an export with the given name.
     pub fn lookup(&self, field: &str) -> Option<Export> {
-        let export = if let Some(export) = self.module.exports.get(field) {
-            export.clone()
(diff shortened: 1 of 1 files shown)


Hywan:
bors try

jubianchi (approved):

MarkMcCaskey (approved):
Nice improvement!  This seems like the kind of thing clippy could find, I wonder if we should start using clippy in CI in a more strict way to keep our code simple and normalized

Hywan:
> Nice improvement! This seems like the kind of thing clippy could find, I wonder if we should start using clippy in CI in a more strict way to keep our code simple and normalized

We may want to use https://github.com/actions-rs/clippy.

PR #1828, opened 2020-11-19
feat(vm) Simplify pointer arithmetic by using existing helper methods

Hywan:
# Description

This patch has started by using the `Instance::vmctx_plus_offset`
method to replace the `&self.vmctx as *const VMContext as *const
u8).add(usize::try_from(offset).unwrap)` pieces of code, as it is
designed to do that.

Then I noticed that `vmctx_plus_offset` was already used by
`Instance::tables_ptr` and `Instance::memories_ptr`. Exactly what were
needed by `Instance::memory_index` and `Instance::table_index`
respectively.

So finally, this patch just updates the `memory_index` and
`table_index` method to use `memories_ptr` and `tables_ptr`. The code
stays the same, it's just much simpler.

# Review

- ~[ ] Add a short description of the the change to the CHANGELOG.md file~ (not necessary I think)

### DIFF
--- a/lib/vm/src/instance.rs
+++ b/lib/vm/src/instance.rs
@@ -417,11 +417,7 @@ impl Instance {
 
     /// Return the table index for the given `VMTableDefinition`.
     pub(crate) fn table_index(&self, table: &VMTableDefinition) -> LocalTableIndex {
-        let offsets = &self.offsets;
-        let begin = unsafe {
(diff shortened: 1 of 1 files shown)


Hywan:
bors try

jubianchi (approved):

MarkMcCaskey (approved):
Nice!

PR #1829, opened 2020-11-19
chore(ci): Use the new llvm-custom-builds

jubianchi:
# Review

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -43,15 +43,14 @@ jobs:
           - build: linux
             os: ubuntu-18.04
             rust: 1.47.0
-            llvm_url: 'https://github.com/wasmerio/llvm-build/releases/download/10.x/Ubuntu1604_Release.tar.xz'
-            # llvm_url: 'https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/clang+llvm-10.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz'
+            llvm_url: 'https://github.com/wasmerio/llvm-custom-builds/releases/download/10.x/linux-amd64.tar.gz'
             artifact_name: 'wasmer-linux-amd64'
             cross_compilation_artifact_name: 'cross_compiled_from_linux'
             run_integration_tests: true
           - build: macos
             os: macos-latest
             rust: 1.47.0
-            llvm_url: 'https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/clang+llvm-10.0.0-x86_64-apple-darwin.tar.xz'
+            llvm_url: 'https://github.com/wasmerio/llvm-custom-builds/releases/download/10.x/darwin-amd64.tar.gz'
             artifact_name: 'wasmer-darwin-amd64'
             cross_compilation_artifact_name: 'cross_compiled_from_mac'
             run_integration_tests: true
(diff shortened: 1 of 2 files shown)


jubianchi:
bors try

.github/workflows/main.yaml
```diff
@@ -123,19 +123,17 @@ jobs:
...
-        #   7z x llvm-installer.exe -oC:/llvm-10
-        #   echo C:/llvm-10/bin >> $GITHUB_PATH
-        #   echo "LLVM_SYS_100_PREFIX=C:/llvm-10" >> $GITHUB_ENV
-        #   echo "LIBCLANG_PATH=C:/llvm-10/bin/libclang.dll" >> $GITHUB_ENV
-      - name: Install LLVM (Unix)
+          curl --proto '=https' --tlsv1.2 -sSf ${{ matrix.llvm_url }} -L -o llvm.tar.gz
+          7z x llvm.tar.gz -oC:/llvm-10
+          echo C:/llvm-10/bin >> $GITHUB_PATH
+          echo "LLVM_SYS_100_PREFIX=C:/llvm-10" >> $GITHUB_ENV
+          echo "LIBCLANG_PATH=C:/llvm-10/bin/libclang.dll" >> $GITHUB_ENV
```
Hywan, line 130: The only difference between Windows and Linux+macOS here is the location where LLVM is uncompressed. Note that `7z` is available on all platforms, and `tar xzvf … -C …` also works on all platforms.

Consequently, we may want to remove the `if:` step's condition, and merge the two steps into a single one.

Thoughts?

Hywan (commented):
Apart from my comment, I think it's good :+1:.

.github/workflows/main.yaml
jubianchi: Oh yes, if we can have the exact same step on every platform, it could be great, will have a look at that.

Thanks for the hint.

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try-

jubianchi:
bors try

MarkMcCaskey (commented):
Looks like aarch64 is still using the old LLVM url here; so `env.LLVM_DIR` is never set.

you can specify `shell: bash` to get a uniform environment across platforms if you run into issue with Windows, for example.

Also when you get this working, please update it on wasmer nightly as well!  Sorry about the lack of process here!  If you can think of a way to simplify updating things in 2 places, please let me know

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try-

jubianchi:
bors try

jubianchi:
bors try-

jubianchi:
bors try

.github/workflows/main.yaml
```diff
@@ -103,18 +103,13 @@ jobs:
...
-        # run: |
-        #   curl --proto '=https' --tlsv1.2 -sSf https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/LLVM-10.0.0-win64.exe -L -o llvm-installer.exe
-        #   7z x llvm-installer.exe -oC:/llvm-10
-        #   echo C:/llvm-10/bin >> $GITHUB_PATH
-        #   echo "LLVM_SYS_100_PREFIX=C:/llvm-10" >> $GITHUB_ENV
-        #   echo "LIBCLANG_PATH=C:/llvm-10/bin/libclang.dll" >> $GITHUB_ENV
-      - name: Install LLVM (Unix)
-        if: matrix.os != 'windows-latest' && matrix.target != 'aarch64-apple-darwin'
+      - name: Install LLVM
+        if: matrix.os != 'windows-latest' && matrix.os != 'macos-11.0'
```
MarkMcCaskey, line 107: `matrix.os != 'macos-11.0'` this will always be true, the name in `matrix.os` for macos is `macos-latest`

MarkMcCaskey (commented):
Looks good to me, want to try it on the nightly release first?

.github/workflows/main.yaml
jubianchi: What we want here is skip `macos-arm64` for which `matrix.os` is `macos-11.0`

MarkMcCaskey: Oh, didn't see that! it was hidden in the diff viewer, looks good to me then

Hywan (approved):
Looks good to me! Would be great to use our own Windows builds, but let's do that in another PR!

jubianchi:
@MarkMcCaskey yes, I was waiting for a nightly to pass. I've merged the PR there and will wait for it to pass a build before merging #1829

jubianchi:
Nightly passed with the new LLVM builds (https://github.com/wasmerio/wasmer-nightly/actions/runs/386156412).

bors r+

PR #1837, opened 2020-11-23
feat(vm) Ability to use exports after the instance has been freed

Hywan:
# Description

Sequel of https://github.com/wasmerio/wasmer/pull/1822. Before #1822, we were able to use an export from an instance even if the instance were freed. That was a nice feature, but since #1822 is merged, it's now an undefined behavior. Indeed, this feature was working because of a memory leak. This PR is an attempt to safely restore this behavior!

At the same time, this patch fixes the way the `Layout` of `Instance` is calculated (it was wrong).

There is now an `InstanceAllocator` that can be viewed as a `Arc<Instance>` basically. The `InstanceAllocator` type is responsible to calculate the layout of an `Instance`, to allocate it, and to deallocate it. The `InstanceHandle` allocates it along with constructing the `Instance`. A clone of the `InstanceAllocator` is passed into exports when they are created. Consequently, every export owns a clone of the `InstanceAllocator`.

The `InstanceAllocator` frees the `Instance` only when all clones of itself are dropped, just like an `Arc`. `InstanceAllocator` is the only type that “owns” the pointer to the `Instance` for more safety. That way, we stop having copies of the same `Instance` pointer everywhere.

So by (re-)implementing the “use-export-after-freed-instance” feature, we also add more soundness, fix silent bugs (like incorrect `Layout` calculation), and add more safety in the code by having less raw pointers.

# To do

- [x] Zero memory leak with `valgrind`
- [x] Zero errors for memory writes with `valgrind`
- [x] Our example works
- [x] Extensive documentation
- [x] Update `InstanceHandle::from_vmctx` (only used by the trap API) [note: the method has been removed]
- [x] All tests pass

# Example we want to pass

This example is now a test in the `lib/api/` crate.

<details>

<summary>The example we want to work</summary>

```rust
use wasmer::*;

fn main() {
    let store = Store::default();
    let module = Module::new(
        &store,
        "
    (module
      (type $sum_t (func (param i32 i32) (result i32)))
      (func $sum_f (type $sum_t) (param $x i32) (param $y i32) (result i32)
        local.get $x
        local.get $y
        i32.add)
      (export \"sum\" (func $sum_f)))
",
    )
    .unwrap();

    let import_object = ImportObject::new();
    let instance2 = instance.clone();
    let instance3 = instance.clone();

    let sum = instance.exports.get_function("sum").unwrap().clone();

    drop(instance);
    drop(instance2);
    drop(instance3);

    dbg!(sum.call(&[Value::I32(1), Value::I32(2)]).unwrap());
}
```

</details>

# Review


lib/vm/src/export.rs
```diff
@@ -38,6 +39,9 @@ pub struct ExportFunction {
     /// Address of the function call trampoline owned by the same VMContext that owns the VMFunctionBody.
     /// May be None when the function is a host-function (FunctionType == Dynamic or vmctx == nullptr).
     pub call_trampoline: Option<VMTrampoline>,
+    /// A reference to the instance, to ensure it's not deallocated
+    /// before the function.
+    pub instance_allocator: Option<Arc<InstanceAllocator>>,
```
MarkMcCaskey, line 44: You mentioned it on Slack, but it'd be good to explain the semantics of the `Option` in these doc comments

lib/vm/src/instance.rs
```diff
@@ -761,99 +680,98 @@ impl Instance {
...
-    /// Whether `Self` owns `self.instance`. It's not always the case;
-    /// e.g. when `Self` is built with `Self::from_vmctx`.
-    ///
-    /// This information is necessary to know whether `Self` should
-    /// deallocate `self.instance`.
-    owned_instance: bool,
+#[derive(Debug)]
+#[repr(C)]
+pub struct InstanceAllocator {
+    ptr_to_dealloc: *const Arc<Self>,
```
MarkMcCaskey, line 686: Confused about the `Arc<Self>` here

lib/vm/src/instance.rs
```diff
@@ -761,99 +680,98 @@ impl Instance {
...
+    ptr_layout: Layout,
+    instance: mem::ManuallyDrop<Instance>,
 }
 
-/// # Safety
-/// This is safe because there is no thread-specific logic in `InstanceHandle`.
-/// TODO: this needs extra review
-unsafe impl Send for InstanceHandle {}
+impl InstanceAllocator {
+    #[inline(always)]
```
MarkMcCaskey, line 692: minor nit but probably unnecessary: Rust is theoretically pretty good at inlining, this attribute therefore only says "if Rust ever decides it's a bad idea to inline this for some reason, do it anyways" which maybe we don't want.

Don't let this block anything though, this is the kind of thing we can figure out later and do a code-base wide review on our use of `#[inline]` and `#[inline(always)]`

lib/vm/src/instance.rs
```diff
@@ -761,99 +680,98 @@ impl Instance {
...
 
-impl InstanceHandle {
-    /// Allocates an instance for use with `InstanceHandle::new`.
-    ///
-    /// Returns the instance pointer and the [`VMOffsets`] that describe the
-    /// memory buffer pointed to by the instance pointer.
-    pub fn allocate_instance(module: &ModuleInfo) -> (NonNull<u8>, VMOffsets) {
-        let offsets = VMOffsets::new(mem::size_of::<*const u8>() as u8, module);
+    fn instance_layout(offsets: &VMOffsets) -> Layout {
+        let size = mem::size_of::<Arc<mem::ManuallyDrop<Self>>>()
```
MarkMcCaskey, line 702: This seems wrong to me 🤔

MarkMcCaskey: But perhaps I don't really understand -- as far as I can tell the `Arc<ManuallyDrop<Self>>` is a lie

lib/vm/src/instance.rs
```diff
@@ -761,99 +680,98 @@ impl Instance {
...
+            .checked_add(usize::try_from(offsets.size_of_vmctx()).unwrap())
+            .unwrap();
+        let align = mem::align_of::<Arc<mem::ManuallyDrop<Self>>>();
```
MarkMcCaskey, line 705: This also seems likely to be wrong, it could under-align fields that have higher alignment than this small type

lib/vm/src/instance.rs
```diff
@@ -761,99 +680,98 @@ impl Instance {
...
-        let num_memories = offsets.num_local_memories;
-        let num_memories = usize::try_from(num_memories).unwrap();
-        let mut out = Vec::with_capacity(num_memories);
-        // TODO: better encapsulate this logic, this shouldn't be duplicated
-        let base_ptr = instance_ptr.as_ptr().add(std::mem::size_of::<Instance>());
-        for i in 0..num_memories {
-            let mem_offset = offsets.vmctx_vmmemory_definition(LocalMemoryIndex::new(i));
-            let mem_offset = usize::try_from(mem_offset).unwrap();
+    unsafe fn deallocate_instance(&mut self) {
+        mem::ManuallyDrop::drop(&mut self.instance);
```
MarkMcCaskey, line 726: TODO: review this very closely, this looks like it's dropping the wrong thing (the pointer, not the thing it's pointing to)

MarkMcCaskey (commented):
Good work!  This is very complicated and I have low confidence in the entire implementation being correct but I think it's a good start.  The design idea seems like it should work as far as I can tell right now!

lib/vm/src/export.rs
Hywan: Correct, I'll!

lib/vm/src/instance.rs
Hywan: I'll add very verbose comment to explain everything here :-).

lib/vm/src/instance.rs
Hywan: It is. It's an error in my merge. Fixing it.

Hywan:
Finally:

```
==92344== Memcheck, a memory error detector
==92344== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==92344== Using Valgrind-3.17.0.GIT and LibVEX; rerun with -h for copyright info
==92344== Command: target/debug/bug
==92344==
Dropping `wasmer::Instance`
Dropping `wasmer::Instance`
Dropping `wasmer::Instance`
Trying to drop `wasmer_vm::InstanceAllocator`...
... not now
Dropping `wasmer::Function`
Trying to drop `wasmer_vm::InstanceAllocator`...
... not now
[src/main.rs:30] sum.call(&[Value::I32(1), Value::I32(2)]).unwrap() = [
    I32(3),
]
Dropping `wasmer::Function`
Trying to drop `wasmer_vm::InstanceAllocator`...
Yep, dropping it.
Dropping `wasmer_vm::Instance`
--92344:0:schedule VG_(sema_down): read returned -4
--92344:0:schedule VG_(sema_down): read returned -4
--92344:0:schedule VG_(sema_down): read returned -4
--92344:0:schedule VG_(sema_down): read returned -4
--92344:0:schedule VG_(sema_down): read returned -4
--92344:0:schedule VG_(sema_down): read returned -4
--92344:0:schedule VG_(sema_down): read returned -4
--92344:0:schedule VG_(sema_down): read returned -4
==92344==
==92344== HEAP SUMMARY:
==92344==     in use at exit: 84,594 bytes in 341 blocks
==92344==   total heap usage: 762 allocs, 421 frees, 145,029 bytes allocated
==92344==
==92344== LEAK SUMMARY:
==92344==    definitely lost: 0 bytes in 0 blocks
==92344==    indirectly lost: 0 bytes in 0 blocks
==92344==      possibly lost: 0 bytes in 0 blocks
==92344==    still reachable: 68,961 bytes in 178 blocks
==92344==         suppressed: 15,633 bytes in 163 blocks
==92344== Reachable blocks (those to which a pointer was found) are not shown.
==92344== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==92344==
==92344== For lists of detected and suppressed errors, rerun with: -s
==92344== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 5 from 5)
```

lib/vm/src/instance.rs
Hywan: Anyway, the refactoring removed this :-).

lib/vm/src/instance.rs
Hywan: The code has radically changed. This doesn't hold any more.

lib/vm/src/instance.rs
Hywan: This has radically changed. The commented code has been removed.

lib/vm/src/instance.rs
```diff
@@ -273,95 +259,15 @@ impl Instance {
...
-
-    /// Return an iterator over the exports of this instance.
-    ///
-    /// Specifically, it provides access to the key-value pairs, where the keys
-    /// are export names, and the values are export declarations which can be
-    /// resolved `lookup_by_declaration`.
-    pub fn exports(&self) -> indexmap::map::Iter<String, ExportIndex> {
-        self.module.exports.iter()
-    }
-
```
Hywan, line 364: Moved inside `InstanceHandle`, so that we can inject `InstanceAllocator` clones inside the exports.

lib/vm/src/instance.rs
```diff
@@ -535,14 +442,6 @@ impl Instance {
...
     }
 
-    fn alloc_layout(offsets: &VMOffsets) -> Layout {
-        let size = mem::size_of::<Self>()
-            .checked_add(usize::try_from(offsets.size_of_vmctx()).unwrap())
-            .unwrap();
-        let align = mem::align_of::<Self>();
-        Layout::from_size_align(size, align).unwrap()
-    }
-
```
Hywan, line 545: Moved inside `InstanceAllocator`, it's the new place to be.

lib/vm/src/instance.rs
```diff
@@ -761,110 +661,272 @@ impl Instance {
...
+    #[inline]
+    pub(crate) fn instance_ref<'a>(&'a self) -> &'a Instance {
+        // SAFETY: The pointer is properly aligned, it is
+        // “dereferencable”, it points to an initialized memory of
+        // `Instance`, and the reference has the lifetime `'a`.
+        unsafe { self.instance.as_ref() }
+    }
+}
+
+/// TODO: Review this super carefully.
```
Hywan, line 802: TODO

lib/vm/src/instance.rs
```diff
@@ -978,6 +1052,97 @@ impl InstanceHandle {
...
+        let instance = (&*vmctx).instance();
+
+        Self {
+            instance: Arc::from_raw(instance as *const InstanceAllocator),
+            // We consider `vmctx` owns the `Instance`, not `Self`.
+            //owned_instance: false,
+        }
+        */
+        todo!()
+    }
```
Hywan, line 1072: TODO

lib/vm/src/instance.rs
```diff
@@ -978,6 +1052,97 @@ impl InstanceHandle {
...
+        for i in 0..num_tables {
+            let table_offset = offsets.vmctx_vmtable_definition(LocalTableIndex::new(i));
+            let table_offset = usize::try_from(table_offset).unwrap();
+
+            let new_ptr = NonNull::new_unchecked(base_ptr.add(table_offset));
+
+            out.push(new_ptr.cast());
+        }
+        out
+    }
```
Hywan, line 1144: Moved from above, and changed a little bit (just code simplification, nothing fancy).

lib/vm/src/instance.rs
```diff
@@ -987,64 +1152,132 @@ impl InstanceHandle {
...
+                    address,
+                    // Any function received is already static at this point as:
+                    // 1. All locally defined functions in the Wasm have a static signature.
+                    // 2. All the imported functions are already static (because
+                    //    they point to the trampolines rather than the dynamic addresses).
+                    kind: VMFunctionKind::Static,
+                    signature,
+                    vmctx,
+                    call_trampoline,
+                    instance_allocator: Some(instance),
```
Hywan, line 1229: That's 🆕.

lib/vm/src/instance.rs
```diff
@@ -987,64 +1152,132 @@ impl InstanceHandle {
...
+                let from = if let Some(def_index) = instance_ref.module.local_table_index(*index) {
+                    instance_ref.tables[def_index].clone()
+                } else {
+                    let import = instance_ref.imported_table(*index);
+                    import.from.clone()
+                };
+
+                ExportTable {
+                    from,
+                    instance_allocator: Some(instance),
```
Hywan, line 1244: That's 🆕.

lib/vm/src/instance.rs
```diff
@@ -987,64 +1152,132 @@ impl InstanceHandle {
...
+                let from = if let Some(def_index) = instance_ref.module.local_memory_index(*index) {
+                    instance_ref.memories[def_index].clone()
+                } else {
+                    let import = instance_ref.imported_memory(*index);
+                    import.from.clone()
+                };
+
+                ExportMemory {
+                    from,
+                    instance_allocator: Some(instance),
```
Hywan, line 1259: That's 🆕.

lib/vm/src/instance.rs
```diff
@@ -987,64 +1152,132 @@ impl InstanceHandle {
...
+                        instance_ref.globals[def_index].clone()
+                    } else {
+                        let import = instance_ref.imported_global(*index);
+                        import.from.clone()
+                    }
+                };
+
+                ExportGlobal {
+                    from,
+                    instance_allocator: Some(instance),
```
Hywan, line 1276: That's 🆕.

jubianchi:
bors try

lib/vm/src/instance.rs
```diff
@@ -761,110 +661,272 @@ impl Instance {
...
+        //
+        // > It is important to enforce any possible access to the object in one
+        // > thread (through an existing reference) to *happen before* deleting
+        // > the object in a different thread. This is achieved by a "release"
+        // > operation after dropping a reference (any access to the object
+        // > through this reference must obviously happened before), and an
+        // > "acquire" operation before deleting the object.
+        //
+        // [1]: https://www.boost.org/doc/libs/1_55_0/doc/html/atomic/usage_examples.html
+        atomic::fence(atomic::Ordering::Acquire);
```
MarkMcCaskey, line 884: When we do this, don't we have to check that it wasn't incremented? 🤔

lib/vm/src/instance.rs
```diff
@@ -978,6 +1052,97 @@ impl InstanceHandle {
...
+    /// # Safety
+    /// - `instance_ptr` must point to enough memory that all of the offsets in
+    ///   `offsets` point to valid locations in memory.
+    pub unsafe fn memory_definition_locations(
+        instance_ptr: NonNull<u8>,
+        offsets: &VMOffsets,
+    ) -> Vec<NonNull<VMMemoryDefinition>> {
+        // `NonNull<u8>` here actually means `NonNull<Instance>`. See
+        // `Self::allocate_instance` to understand why.
+        let instance_ptr: NonNull<Instance> = instance_ptr.cast();
```
MarkMcCaskey, line 1093: Doesn't this cast make the `add` below incorrect?  The logic is written in terms of bytes and ptr::add multiplies by the size.

Thanks for calling out that you made changes here or I never would have caught this.  In the future it's probably best to avoid changing moved code in a PR this big and complex: github's diff viewer makes understanding these types of changes really difficult.

lib/vm/src/instance.rs
```diff
@@ -978,6 +1052,97 @@ impl InstanceHandle {
...
+    /// # Safety
+    /// - `instance_ptr` must point to enough memory that all of the offsets in
+    ///   `offsets` point to valid locations in memory.
+    pub unsafe fn table_definition_locations(
+        instance_ptr: NonNull<u8>,
+        offsets: &VMOffsets,
+    ) -> Vec<NonNull<VMTableDefinition>> {
+        // `NonNull<u8>` here actually means `NonNull<Instance>`. See
+        // `Self::allocate_instance` to understand why.
+        let instance_ptr: NonNull<Instance> = instance_ptr.cast();
```
MarkMcCaskey, line 1127: This also causes the pointer math below to be incorrect

MarkMcCaskey (commented):
Nice improvements, lots of really tricky code though.

I'm thinking it might make the code a bit easier to read to put `InstanceAllocator` in a new file or something.  Purely for the sake of encapsulating and hiding complex code that we shouldn't mess with most of the time.

syrusakbary:
Random question.
Let's say we have two instances:
1. Instance 1. Which exports a memory created in the wasm (memory1)
2. Instance 2, that takes a memory from it's imports, and reexports it at the same time (memory 2)

Let's say, we take the memory1 and instantiate the instance2 with this memory1, to then take the exported memory from instance 2 (memory2, which should be the same as memory1).

Which one will be the instance deallocator for memory2?

MarkMcCaskey:
> Random question.
> Let's say we have two instances:
> 
>     1. Instance 1. Which exports a memory created in the wasm (memory1)
> 
>     2. Instance 2, that takes a memory from it's imports, and reexports it at the same time (memory 2)
> 
> 
> Let's say, we take the memory1 and instantiate the instance2 with this memory1, to then take the exported memory from instance 2 (memory2, which should be the same as memory1).
> 
> Which one will be the instance deallocator for memory2?

Due to the memory layout of local vs non-local memories, that memory's data lives in instance 1 (where it was local), reexporting the memory from instance 2 should be fine.  The only issue is that we'd need to be sure instance 1 can't be freed while instance 2 is using a memory from it!

Hywan:
bors try

lib/vm/src/instance.rs
Hywan: Indeed, https://doc.rust-lang.org/std/primitive.pointer.html#method.add:

> `count` is in units of `T`; e.g., a `count` of 3 represents a pointer offset of `3 * size_of::<T>()` bytes.

That's probably why memories and tables are failing. I'm fixing that! Thanks for noticing!

Hywan:
bors try

lib/vm/src/instance.rs
```diff
@@ -761,110 +661,272 @@ impl Instance {
...
+
+/// TODO: Review this super carefully.
+unsafe impl Send for InstanceAllocator {}
+unsafe impl Sync for InstanceAllocator {}
+
+/// A soft limit on the amount of references that may be made to an `InstanceAllocator`.
+///
+/// Going above this limit will make the program to panic at exactly
+/// `MAX_REFCOUNT` references.
+const MAX_REFCOUNT: usize = std::usize::MAX - 1;
```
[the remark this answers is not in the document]
Hywan: Good idea. Fixed!

lib/vm/src/instance.rs
```diff
@@ -761,110 +661,272 @@ impl Instance {
...
-    /// This function lets us create `Table` objects on the host with backing
-    /// memory in the VM.
+    /// This creates another `InstanceAllocator` using the same
+    /// `instance` pointer, increasing the strong reference count.
+    #[inline]
+    fn clone(&self) -> Self {
+        let old_size = self.strong.fetch_add(1, atomic::Ordering::Relaxed);
+
+        if old_size > MAX_REFCOUNT {
+            panic!("Too much references of `InstanceAllocator`");
```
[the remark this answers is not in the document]
Hywan: Thanks! Fixed.

Hywan:
bors try

Hywan:
bors try

lib/vm/src/instance.rs
```diff
@@ -761,110 +661,272 @@ impl Instance {
...
 
-    /// Get the locations of where the `VMTableDefinition`s should be stored.
+    /// Get a reference to the `Instance`.
+    #[inline]
+    pub(crate) fn instance_ref<'a>(&'a self) -> &'a Instance {
+        // SAFETY: The pointer is properly aligned, it is
+        // “dereferencable”, it points to an initialized memory of
+        // `Instance`, and the reference has the lifetime `'a`.
+        unsafe { self.instance.as_ref() }
+    }
```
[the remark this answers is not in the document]
Hywan: `self.instance` is a correctly aligned and initialized pointer to `Instance`  in `new`, and in `instance_ref` (now renamed `as_ref`). I don't see why this method or `new` must be `unsafe`?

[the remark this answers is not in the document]
Hywan: Also, the pointer to `Instance` is only owned, and can only be freed by `InstanceAllocator`. Only `&Instance` can be used outside this struct. So except if someone deref the reference and do something bad with it, this code above is safe :-p.

Hywan:
bors try

lib/vm/src/instance.rs
```diff
@@ -761,110 +661,272 @@ impl Instance {
...
+impl Clone for InstanceAllocator {
+    /// Makes a clone of `InstanceAllocator`.
     ///
-    /// This function lets us create `Table` objects on the host with backing
-    /// memory in the VM.
+    /// This creates another `InstanceAllocator` using the same
+    /// `instance` pointer, increasing the strong reference count.
+    #[inline]
+    fn clone(&self) -> Self {
+        let old_size = self.strong.fetch_add(1, atomic::Ordering::Relaxed);
```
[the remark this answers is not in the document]
Hywan: Done!

lib/vm/src/instance.rs
Hywan: This code is from `Arc` itself. I trust them more than I trust myself.

lib/vm/src/instance.rs
Hywan: Yup, fixed!

lib/vm/src/instance.rs
```diff
@@ -987,64 +1151,132 @@ impl InstanceHandle {
...
+        let export = self.module().exports.get(field)?;
+
+        Some(self.lookup_by_declaration(&export))
     }
 
     /// Lookup an export with the given export declaration.
     pub fn lookup_by_declaration(&self, export: &ExportIndex) -> Export {
-        self.instance().lookup_by_declaration(export)
+        let instance = self.instance().clone();
+        let instance_ref = instance.as_ref();
```
Hywan, line 1198: That's 🆕. This is where the final magic happens!

lib/vm/src/instance.rs
```diff
@@ -761,110 +661,272 @@ impl Instance {
...
-        let offsets = VMOffsets::new(mem::size_of::<*const u8>() as u8, module);
+    /// Calculate the appropriate layout for `Instance`.
+    fn instance_layout(offsets: &VMOffsets) -> Layout {
+        let size = mem::size_of::<Instance>()
+            .checked_add(
+                usize::try_from(offsets.size_of_vmctx())
+                    .expect("Failed to convert the size of `vmctx` to a `usize`"),
+            )
+            .expect("Failed to compute the size of `Instance`");
+        let align = mem::align_of::<Instance>();
```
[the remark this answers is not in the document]
Hywan: Actually, it's from the old code base, but I believe it is wrong. It works only because the size of `Instance` + the size of `VMContext` still fits in the alignment of `Instance` only.

I can use a drafted code I developed when working on this patch to compute a more change-proof calculation for the alignment.

Hywan:
bors try

Hywan:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

lib/vm/src/instance.rs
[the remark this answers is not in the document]
Hywan: I think we should do something like this:

```rust
fn instance_Layout(offsets: &VMOffsets) -> Layout {
        let (layout, _offset) = Layout::new::<Instance>()
            .extend(
                Layout::array::<u8>(
                    usize::try_from(offsets.size_of_vmctx())
                        .expect("Failed to convert the size of `vmctx` to a `usize`"),
                )
                .expect("Failed to create a layout for `VMContext`"),
            )
            .expect("Failed to extend to `Instance` layout to include `VMContext`");

        dbg!(layout.pad_to_align())
}
```

On my machine, it gives:

```
Layout {
    size_: 400,
    align_: 16,
}
```

which is correct.

Thoughts?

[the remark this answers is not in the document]
Hywan: Hmm, actually the size of `VMContext` is given by `VMOffsets.size_of_vmctx` and I'm even not sure it computes the size correctly. Digging.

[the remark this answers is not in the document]
Hywan: I can't decide whether we should use `Layout::array::<u8>` or `Layout::array::<usize>`. For me, `VMContext` contains pointers, so it should be an array of `usize`, but I'm not sure. Using `u8` gives the same results as we have on `master`, but it may be wrong!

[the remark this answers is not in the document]
Hywan: Let's keep `u8` for the moment. As you said earlier “Somewhat unrelated”. https://github.com/wasmerio/wasmer/pull/1837/commits/0ec3b70b2267079e4f3ac84fd24b8cf465b0e6d8 already improves how the real `Instance` layout is calculated, which is good, let's keep things as they are.

Hywan:
bors try

Hywan:
bors try

Hywan:
Maybe we should add tests specifically for that :-p?

Hywan:
bors try

lib/vm/src/instance.rs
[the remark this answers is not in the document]
MarkMcCaskey: That seems reasonable to me, the one thing I'd add is that I think we need to make sure we have proper padding (to force proper alignment) between the known struct and the dynamic extension of it.

Just because the struct itself is aligned to 16 doesn't mean that the dynamic part will be...

lib/vm/src/instance.rs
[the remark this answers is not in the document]
MarkMcCaskey: > self.instance is a correctly aligned and initialized pointer to Instance in new, and in instance_ref (now renamed as_ref). I don't see why this method or new must be unsafe?

I don't believe we can guarantee that with safe Rust, therefore we can't make that assumption without `unsafe`. `unsafe` means "read the docs for how to use this function correctly", if there's a way to use the function that compiles without `unsafe` that will cause memory safety issues, it's unsound and the API should be marked `unsafe`.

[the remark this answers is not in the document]
MarkMcCaskey: Notably, anything that takes a `NonNull<T>` that assumes the pointer is anything but garbage must be `unsafe` because `NonNull::dangling` can be used in safe Rust to get a pointer to arbitrary memory.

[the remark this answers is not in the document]
MarkMcCaskey: Because this isn't a public method, I think this is a relatively small deal, but I think it's still best to be precise here!

lib/vm/src/instance.rs
MarkMcCaskey: One thing to keep in mind is that this code is not doing the same thing as `Arc`, I'm unclear on if it's different enough to matter though.

Notably the std::arc::Arc source code defers actually deallocating the underlying memory until all the `Weak` pointers are dropped.

And that code does similar checks:

```Rust
        if inner.weak.fetch_sub(1, Release) == 1 {
            acquire!(inner.weak);
            unsafe { Global.dealloc(self.ptr.cast(), Layout::for_value(self.ptr.as_ref())) }
        }
```

After reviewing `InstanceAllocator` again, I think it's probably similar enough that we can call it correct!

MarkMcCaskey (commented):
Looks good to me, biggest thing I'd want to see changed before shipping is the unsoundness in on our internal `InstanceAllocator` API.

lib/vm/src/instance.rs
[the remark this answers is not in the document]
Hywan: I guess `Layout` already handles that case. The dynamic part is handled by `Layout::array::<u8>` which is the extension of `Layout::new::<Instance>`.

In anyways, the actual code is _not_ worst than what we had before ;-).

lib/vm/src/instance.rs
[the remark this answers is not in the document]
Hywan: I see! If you consider `InstanceAllocator` alone, then yes, I fully agree. I was talking about `InstanceHandle` + `InstanceAllocator`, that's why I wasn't aligned with you! I'm updating the code now to make `InstanceAllocator::new` unsafe. If the constructor is safely called, then `as_ref` will be safe too.

Done!

lib/vm/src/instance.rs
Hywan: :+1:

Hywan:
Everything has been addressed. Thanks @MarkMcCaskey for the deep review. I'm fixing the conflicts with `master` and I'll merge the PR.

Hywan:
bors try

jubianchi (approved):

PR #1839, opened 2020-11-23
Metering middleware.

nlewycky:
Based on https://github.com/wasmerio/wasmer-reborn/pull/129 . Original description:

- [x] Necessary APIs for modifying module info.
- [x] Global transformation.
- [x] Per-function transformation.
- [x] Testing and examples.


Cargo.toml
```diff
@@ -23,6 +23,7 @@ wasmer-wasi = { version = "1.0.0-alpha5", path = "lib/wasi", optional = true }
 wasmer-wast = { version = "1.0.0-alpha5", path = "tests/lib/wast", optional = true }
 wasmer-cache = { version = "1.0.0-alpha5", path = "lib/cache", optional = true }
 wasmer-types = { version = "1.0.0-alpha5", path = "lib/wasmer-types" }
+wasmer-middlewares = { version = "1.0.0-alpha.5", path = "lib/middlewares", optional = true }
```
MarkMcCaskey, line 26: Slight typo in the version, should drop the `.`

webmaster128:
Thanks for bringing this over. Happy to test this work in an external project. Just ping me when you think this makes sense.

lib/compiler-singlepass/src/compiler.rs
```diff
@@ -47,16 +47,19 @@ impl Compiler for SinglepassCompiler {
...
     ) -> Result<Compilation, CompileError> {
         if compile_info.features.multi_value {
             return Err(CompileError::UnsupportedFeature("multivalue".to_string()));
         }
-        let vmoffsets = VMOffsets::new(8, &compile_info.module);
         let memory_styles = &compile_info.memory_styles;
         let table_styles = &compile_info.table_styles;
+        let mut module = (*compile_info.module).clone();
+        self.config.middlewares.apply_on_module_info(&mut module);
+        compile_info.module = Arc::new(module);
```
MarkMcCaskey, line 61: This will leave all the old `Arc<Module>`s unaffected, this should be looked at closely for bugs

lib/middlewares/src/metering.rs
```diff
@@ -0,0 +1,158 @@
...
+                    state.push_operator(Operator::I64LtU);
+                    state.push_operator(Operator::If { ty: WpTypeOrFuncType::Type(WpType::EmptyBlockType) });
+                    state.push_operator(Operator::Unreachable); // FIXME: Signal the error properly.
+                    state.push_operator(Operator::End);
+
+                    // globals[remaining_points_index] -= self.accumulated_cost;
+                    state.push_operator(Operator::GlobalGet { global_index: self.remaining_points_index.as_u32() });
+                    state.push_operator(Operator::I64Const { value: self.accumulated_cost as i64 });
+                    state.push_operator(Operator::I64Sub);
+                    state.push_operator(Operator::GlobalSet { global_index: self.remaining_points_index.as_u32() });
```
MarkMcCaskey, line 147: Might be ergonomic to implement https://doc.rust-lang.org/std/iter/trait.Extend.html for this, so you can put a bunch of operators in a collection like `&[]` and push them all in

MarkMcCaskey (commented):
Looks pretty reasonable to me!

I'd like to see an example using metering too as the place where we explain how to use it, but that can be done separately after this merges

lib/compiler-singlepass/src/compiler.rs
nlewycky: Indeed that was a source of the bug fixed in 363a28c of this PR. I'm trusting that the tests will catch any real problems here. (Also, there are no old Arc's? At least not in this function?)

lib/middlewares/src/metering.rs
nlewycky: Done. Two new `Extend`s.

Cargo.toml
nlewycky: Done, thanks.

lib/compiler-singlepass/src/compiler.rs
MarkMcCaskey: `compile_info.module` before it gets overwritten is pointing to the old Arc, anything that it's shared with won't get the update

MarkMcCaskey: Yeah I'm not sure about its correctness in context, but it's possible that this could cause problems later

lib/middlewares/src/metering.rs
```diff
@@ -0,0 +1,160 @@
...
+use wasmer_types::GlobalIndex;
+use wasmer_vm::ModuleInfo;
+
+/// The module-level metering middleware.
+///
+/// # Panic
+///
+/// An instance of `Metering` should not be shared among different modules, since it tracks
+/// module-specific information like the global index to store metering state. Attempts to use
+/// a `Metering` instance from multiple modules will result in a panic.
```
webmaster128, line 22: If an instance of `Metering` can be used for one module only, and it is installed via `CompilerConfig::push_middleware`, does this mean you need a new compiler (and engine and store) for each metered module?

syrusakbary: That's right. The current structure makes easier reusing the modules in headless engines.

syrusakbary (approved):

Issue #1840, opened 2020-11-24
Support closures as host functions tracking issue


MarkMcCaskey:
Wasmer does not currently support closures (functions with captured environments) as host functions, this is a tracking issue for progress on this feature.

The current plan is to reject closures in 1.0 and possibly support them in the future.  Please let us know if this feature is important to your use case.

aidanhs:
This seems like it'd help the ergonomics of environments a bunch.

Simple example:
```
Function::new_native_with_env(&store, script_len,
    move |&script_len_env: &u32| -> u32 { script_len_env })
```
vs
```
Function::new_native(&store,
    move || -> u32 { script_len })
```

----

Slightly more complex example:
```
struct MemEnv {
    #[wasmer(export)]
    memory: LazyInit<Memory>,
}
Function::new_native_with_env(&store, Default::default(),
    move |env: &MemEnv, data: WasmPtr<u8, Array>| -> () {    
          let memory = env.memory_ref().expect("memory not set up");    
          let data_cell = mytry!(data.deref(memory, 0, script_len).ok_or(()));    
          for (cell, &b) in data_cell.into_iter().zip(script.iter()) {    
              cell.set(b)    
          }    
      })
```
vs
```
struct ScriptEnv {
    script: Vec<u8>,
    script_len: u32,
    #[wasmer(export)]
    memory: LazyInit<Memory>,
}
let script_env = ScriptEnv { script, script_len, memory: Default::default() };
Function::new_native_with_env(&store, script_env,
    move |env: &ScriptEnv, data: WasmPtr<u8, Array>| -> () {    
          let memory = env.memory_ref().expect("memory not set up");    
          let data_cell = mytry!(data.deref(memory, 0, env.script_len).ok_or(()));    
          for (cell, &b) in data_cell.into_iter().zip(env.script.iter()) {    
              cell.set(b)    
          }    
      })
```

----

This latter example has a couple of points worth observing:
1. many functions may want a `Memory`, and so if I could capture environments I can use the single `MemoryEnv` struct between them and capture the rest. With no capturing, I have to define a fresh struct for every set of things I want to carry in the environment
[...]

PR #1841, opened 2020-11-24
Disable closures as host functions for now + docs + tests

### LINKED ISSUE #1811: Function::new_native support for closures with captured environments
### Describe the bug

`Function::new_native` seems to be playing fast and loose with closures.

`Cargo.toml`:
```toml
[dependencies]
anyhow = "1.0.34"
wasmer = "1.0.0-alpha5"
wasmer-compiler-cranelift = "1.0.0-alpha5"
wasmer-engine-jit = "1.0.0-alpha5"
```

For reference this (correctly) prints "42":
```rust
use anyhow::Result;
use wasmer::*;

fn main() -> Result<()> {
    let wat = r#"(module
        (import "console" "log" (func $log (param i32)))
        (func (export "main")
          i32.const 42
          call $log
        )
    )"#;
    let cranelift = Cranelift::default();
    let store = Store::new(&JIT::new(&cranelift).engine());
    let module = Module::new(&store, wat)?;
        "console" => {
            "log" => Function::new_native(&store, |value: i32| {
                println!("{}", value);
            })
        }
    };
    let main: NativeFunc<(), ()> = instance
        .exports
        .get_native_function("main")?;
    main.call()?;
    Ok(())
}
```

This (incorrectly) crashes the program with a stack overflow:
```rust
use anyhow::Result;
use wasmer::*;
use std::sync::Arc;

fn main() -> Result<()> {
    let wat = r#"(module
        (import "console" "log" (func $log (param i32)))
        (func (export "main")
          i32.const 42
          call $log
        )
    )"#;
    let cranelift = Cranelift::default();
    let store = Store::new(&JIT::new(&cranelift).engine());
    let module = Module::new(&store, wat)?;
    let message = "message";
    let message_arc = Arc::new(message);
        "console" => {
            "log" => Function::new_native(&store, move |value: i32| {
                println!("{}, {}", message_arc, value);
            })
        }
    };
    let main: NativeFunc<(), ()> = instance
        .exports
        .get_native_function("main")?;
    main.call()?;
    Ok(())
}
```

This (incorrectly) prints 0:
```rust
use anyhow::Result;
use wasmer::*;

fn main() -> Result<()> {
    let wat = r#"(module
        (import "console" "log" (func $log (param i32)))
        (func (export "main")
          i32.const 42
          call $log
        )
    )"#;
    let cranelift = Cranelift::default();
    let store = Store::new(&JIT::new(&cranelift).engine());
    let module = Module::new(&store, wat)?;
    let state = 1337;
        "console" => {
            "log" => Function::new_native(&store, move |_: i32| {
                println!("{}", state);
            })
        }
    };
    let main: NativeFunc<(), ()> = instance
        .exports
        .get_native_function("main")?;
    main.call()?;
    Ok(())
}
```

### Additional context

Globally-installed things:
```
wasmer 1.0.0-alpha4
rustc 1.47.0 (18bf6b4f0 2020-10-07)
Windows 10 10.0.19042 N/A Build 19042
```

MarkMcCaskey:
Resolves #1811 for now

Part of #1840 

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 
 ### Fixed
 
-- [#1841](https://github.com/wasmerio/wasmer/pull/1841) We will now panic when attempting to use a function with a captured env as a host function. Previously this would silently do the wrong thing. See [#1840](https://github.com/wasmerio/wasmer/pull/1840) for info about Wasmer's support of closures as host functions.
+- [#1841](https://github.com/wasmerio/wasmer/pull/1841) We will now panic when attempting to use a native function with a captured env as a host function. Previously this would silently do the wrong thing. See [#1840](https://github.com/wasmerio/wasmer/pull/1840) for info about Wasmer's support of closures as host functions.
 - [#1764](https://github.com/wasmerio/wasmer/pull/1764) Fix bug in WASI `path_rename` allowing renamed files to be 1 directory below a preopened directory.
 
 ## 1.0.0-alpha5 - 2020-11-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 
 ### Fixed
 
+- [#1841](https://github.com/wasmerio/wasmer/pull/1841) We will now panic when attempting to use a function with a captured env as a host function. Previously this would silently do the wrong thing. See [#1840](https://github.com/wasmerio/wasmer/pull/1840) for info about Wasmer's support of closures as host functions.
 - [#1764](https://github.com/wasmerio/wasmer/pull/1764) Fix bug in WASI `path_rename` allowing renamed files to be 1 directory below a preopened directory.
 
 ## 1.0.0-alpha5 - 2020-11-06
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -51,6 +51,11 @@ pub enum FunctionDefinition {
 /// during execution of the function.
 ///
 /// Spec: https://webassembly.github.io/spec/core/exec/runtime.html#function-instances
+///
+/// # Panics
+/// - Closures (functions with captured environments) are not currently supported.
+///   Attempting to create a `Function` with one will result in a panic.
+///   [Closures as host functions tracking issue](https://github.com/wasmerio/wasmer/issues/1840)
 #[derive(Clone, PartialEq)]
 pub struct Function {
     pub(crate) store: Store,
@@ -79,6 +84,9 @@ impl Function {
     where
         F: Fn(&[Val]) -> Result<Vec<Val>, RuntimeError> + 'static,
     {
+        if std::mem::size_of::<F>() != 0 {
+            Self::closures_unsupported_panic();
+        }
         let dynamic_ctx = VMDynamicFunctionContext::from_context(VMDynamicFunctionWithoutEnv {
             func: Box::new(func),
             function_type: ty.clone(),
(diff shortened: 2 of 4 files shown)


lib/api/src/externals/function.rs
```diff
@@ -130,6 +138,9 @@ impl Function {
         F: Fn(&Env, &[Val]) -> Result<Vec<Val>, RuntimeError> + 'static,
         Env: Sized + 'static,
     {
+        if std::mem::size_of::<F>() != 0 {
+            Self::closures_unsupported_panic();
+        }
```
syrusakbary, line 143: I believe normal functions (not native), should still work. We should only disable native ones

MarkMcCaskey: Oh, makes sense!  I've updated it to reflect that now

jubianchi (approved):

Hywan:
It's going to break `wasmer-python`, and probably `wasmer-c-api` (didn't check for that one). Is it temporary?

syrusakbary:
We don't support native closures with Wasmer... are they used in wasmer-python ?

webmaster128:
We use this feature for a year of so without any issues. In our case, the `api` is copied into the closure: https://github.com/CosmWasm/cosmwasm/blob/upgrade-to-wasmer-reborn/packages/vm/src/instance.rs#L124-L130. For some reason this still compiles using wamer at 82fc0080ba6380a13dc2bdaea907cdf250f4d762. What am I missing here? Maybe I don't know what a "closure" is? Maybe copies are okay? Is the `move` making a difference?

webmaster128:
In https://doc.rust-lang.org/book/ch13-01-closures.html#capturing-the-environment-with-closures there is a good explanation regarding naming and the different capture types. In Rust all anonymous functions are closures (no "lambda" as used in other languages). Then there are different types of environment capturing. And it turns out some of those captures (all captured data owned by the function) are still possible with this PR.

Hywan:
> We don't support native closures with Wasmer... are they used in wasmer-python ?

Yes they are: https://github.com/wasmerio/wasmer-python/blob/99bd88d7f8a7cc4083625767d0cac77052456cb1/packages/api/src/externals/function.rs#L145-L193

It captures the `result_types` variable. Not that we can still store them in the environment 🤔.

webmaster128:
I found why my code is not affected by this change. We use `Function::new_with_env`, but this PR is limited to `Function::new_native` and friends. Is this a missing check or should only "native" functions have the new restriction?

tests/compilers/native_functions.rs
```diff
@@ -78,6 +78,90 @@ fn native_function_works_for_wasm() -> Result<()> {
...
+    let store = get_store(false);
+    let state = 3;
+    let env = 4;
+    Function::new_native_with_env(&store, env, move |_env: &i32, _: i32| {
+        println!("{}", state);
+    });
+}
+
+#[test]
+fn lambdas_with_no_env_work() -> Result<()> {
```
webmaster128, line 107:
```suggestion
fn non_native_functions_and_closures_with_no_env_work() -> Result<()> {
```

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 
 ### Fixed
 
+- [#1841](https://github.com/wasmerio/wasmer/pull/1841) We will now panic when attempting to use a function with a captured env as a host function. Previously this would silently do the wrong thing. See [#1840](https://github.com/wasmerio/wasmer/pull/1840) for info about Wasmer's support of closures as host functions.
```
webmaster128, line 24:
```suggestion
- [#1841](https://github.com/wasmerio/wasmer/pull/1841) We will now panic when attempting to use a native function with a captured env as a host function. Previously this would silently do the wrong thing. See [#1840](https://github.com/wasmerio/wasmer/pull/1840) for info about Wasmer's support of closures as host functions.
```

syrusakbary:
> This means wasmerio/wasmer-python#420 was not needed.

Yeah, you are right @webmaster128

syrusakbary (approved):

PR #1843, opened 2020-11-25
Add "Hello World" example

### LINKED ISSUE #1656: "Hello world" example
We should have an example that aims show Wasmer working but is as simple as possible.  Many of the existing examples use more advanced features and it'd be good to have an explicit example dedicated to being as simple as possible.

Edit: after some more searching I may have overestimated how much simpler an example can be, but as long as it's even a little simpler than existing examples, it should be worthwhile.

MarkMcCaskey:
resolves #1656 

This is a bit redundant with the `instance` example (which I only noticed after finishing this example) but I thought it was good to explain the core concepts explicitly.  I ended up writing this as part of the intro docs for the Wasmer API crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -245,3 +245,8 @@ required-features = ["cranelift"]
 name = "imported-function-env"
 path = "examples/imports_function_env.rs"
 required-features = ["cranelift"]
+
+[[example]]
+name = "hello-world"
+path = "examples/hello_world.rs"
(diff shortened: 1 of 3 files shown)


jubianchi (approved):
I think it's always good to have examples even if they may seem redundant sometimes.

Better too much than not enough

syrusakbary (approved):

PR #1844, opened 2020-11-25
Implement and export some WasiFile's

TheLostLambda:
# Description
Hello again! I'm gradually trying to upstream some of the things that I found really helpful when writing a [WASM plugin system](https://github.com/mosaic-org/wasm-module-experiments) with Wasmer recently :slightly_smiling_face: 

It's super useful to be able to rebind the stdin and stdout of a WASI environment, but as it currently stands, nothing (public) actually implements `WasiFile`!

In this PR, I've:
- Made the existing `WasiFile` implementers public (if there is a reason to keep them private, that's okay too)
- Added a `Pipe` that serves as a two-way stdio pipe, sorta like [`std::process::Stdio`](https://doc.rust-lang.org/std/process/struct.Stdio.html). I use this to communicate bidirectionally with my WASI plugin!

I'm happy to add doc-comments if we're happy with this addition :)

# Review


lib/wasi/src/state/types.rs
```diff
@@ -970,6 +970,77 @@ impl WasiFile for Stdin {
...
+        Ok(())
+    }
+    fn bytes_available(&self) -> Result<usize, WasiFsError> {
+        Ok(self.buffer.len())
+    }
+}
+
+impl Read for Pipe {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        (&self.buffer[..]).read(buf)
```
MarkMcCaskey, line 1021: Does calling read multiple times do the correct thing or does it read the same data?

lib/wasi/src/state/types.rs
```diff
@@ -970,6 +970,77 @@ impl WasiFile for Stdin {
     }
 }
 
+/// For piping stdio. Stores all output / input in a byte-vector.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Pipe {
+    buffer: Vec<u8>,
```
MarkMcCaskey, line 976: It probably makes sense to use a [`std::collections::VecDeque`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html) here so we can pop off the front and append to the back in constant time

lib/wasi/src/lib.rs
```diff
@@ -22,8 +22,8 @@ mod utils;
 use crate::syscalls::*;
 
 pub use crate::state::{
-    Fd, WasiFile, WasiFs, WasiFsError, WasiState, WasiStateBuilder, WasiStateCreationError,
-    ALL_RIGHTS, VIRTUAL_ROOT_FD,
+    Fd, HostFile, Pipe, Stderr, Stdin, Stdout, WasiFile, WasiFs, WasiFsError, WasiState,
```
MarkMcCaskey, line 25: I think it's probably fine to make `Stdout`, `Stderr`, and `Stdin` public but I'm not sure about `HostFile`.

Is it useful to make `HostFile` public? My thoughts are that it's very likely that it's a type full of weird implementation details that we probably don't want to have to not break

MarkMcCaskey (commented):
Interesting idea! Thanks for the PR! I think this makes sense to add to our WASI implementation but it'd be nice to have an example or a test using this type.

lib/wasi/src/lib.rs
TheLostLambda: I can imagine some niche uses for it, like piping the output of some WASM to a file on the host machine, but I don't mind keeping that private for now. If we add the `Pipe`, then users should still be able to pipe a file themselves.

TheLostLambda:
Hey @MarkMcCaskey !

Good call about `Read` not behaving quite as it should! That should be fixed by these changes. The suggestion about `VecDeque` is also a good one, but I think I may have found an alternative that is nice as well? Let me know what you think though!

I found out that [`std::io::Cursor`](https://doc.rust-lang.org/std/io/struct.Cursor.html) is *very* similar to the `Pipe` I'm trying to implement. In fact, I'm infuriatingly close to just being able to write `impl WasiFile for Cursor<Vec<u8>>`, but alas, it doesn't implement `Serialize` or `Deserialize`, so I had to write some [wrapper code](https://serde.rs/remote-derive.html). I've put up an [issue here](https://github.com/serde-rs/serde/issues/1930) to upstream this, but `serde`'s code is a little intimidating, so I'm not holding my breath.

Overall though, this should at least guarantee some sensible `Read`, `Write`, and `Seek` implementations for `Pipe`. Let me know if you think this is a good solution!

Some tests / examples sound like a great idea! Do you have a particular place that you'd like me to write up a `Pipe` example in?

Thanks again for the review and the help!

MarkMcCaskey:
Thanks for iterating on this!

> Overall though, this should at least guarantee some sensible Read, Write, and Seek implementations for Pipe. Let me know if you think this is a good solution!

I think that makes sense, the one real problem I have with the `Cusor<Vec<u8>>` implementation is that as far as I can tell it will never free memory when reading, so using `Pipe` will build up memory pressure and eventually cause a problem.  Additionally, due to the way Vec resizing works, the `Pipe` will get slower over time due to copying the old vec on resize. With `VecDeque` we can free memory once it's been read and easily set up an upper bound on how much data a pipe can buffer (though I think it's fine to leave bounding pipes for the future).

> Some tests / examples sound like a great idea! Do you have a particular place that you'd like me to write up a Pipe example in?

If you'd like to add some unit tests (for things like the Read bug above where reading twice would give the same data), I think it's fine to put them in the same file that the type is defined in.

I think more useful than unit tests, would be an example in the top level `examples` directory showing how this type is useful. Alternatively, you can possibly show that in a doc comment on `Pipe` itself, though it may be trickier to actually demonstrate it there.


One more note, I think most pipes can't be `Seek`ed, so dropping the `Cursor` shouldn't be a huge deal.  I'm not familiar with seekable pipes but from a quick google search it seems like they do sometimes exist 🤷 .

I think switching to VecDeque would be an improvement, but we can always change it later provided we're clear about the interface (seekable pipes vs non-seekable pipes).

Thanks for thinking about this!

TheLostLambda:
Thanks again for all of the helpful feedback! I think you are right, we can take a simpler approach here. I think I'm fine with non-seeking pipes (I don't really see any case where I'd need that), and I've just moved to a `VecDeque` that clears data as it's read. Nothing gets read twice anymore and `.read_to_string(&mut buf)` works like a charm – replacing the `Display` implementation I was using to get a string of the contents.

When writing an example, I figured I'd copy the existing WASI example and demonstrate how the pipes work using a simple WASI program that reverses its input (a `String` from stdin).

Compiling my example into WASM using [this](https://github.com/wasmerio/wasi-tests) yields the following, rather unhelpful test, but maybe that's alright? Since this is part of an example?

```
(wasi_test "pipe_reverse.wasm"
  (assert_return (i64.const 0))
  (assert_stdout "\n")
)
```

Regardless, the new example works like a charm, printing the following:
```
Compiling module...
Creating `WasiEnv`...
Instantiating module with WASI imports...
Writing "racecar go zoom" to the WASI stdin...
Call WASI `_start` function...
Reading from the WASI stdout...
Read "mooz og racecar" from the WASI stdout!
```

Let me know what you think!

examples/wasi_pipes.rs
```diff
@@ -0,0 +1,86 @@
...
+        .finalize()?;
+
+    println!("Instantiating module with WASI imports...");
+    // Then, we get the import object related to our WASI
+    // and attach it to the Wasm instance.
+    let import_object = wasi_env.import_object(&module)?;
+    let instance = Instance::new(&module, &import_object)?;
+
+    // WASI requires to explicitly set the memory for the `WasiEnv`
+    wasi_env.set_memory(instance.exports.get_memory("memory")?.clone());
```
MarkMcCaskey, line 53: This should no longer be required on master!

MarkMcCaskey (approved):
This looks great, thanks!

There are a few things I need to do, like get it up to date with master and adjust a few things that have changed.

Aside from that we have a bit of a weird process with wasi tests where we generate things, so I'll do that as well by moving it over to https://github.com/wasmerio/wasi-tests and then pulling in the updated stuff into current Wasmer.

Thanks for doing this!

PR #1851, opened 2020-11-30
test+doc(c-api): Test and document the C A PI

Hywan:
# Description

This PR improves the test suites and the documentation of the C API.

# Review


lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -79,15 +79,74 @@ pub struct wasm_config_t {
...
+/// ```
 ///
 /// cbindgen:ignore
 #[no_mangle]
 pub extern "C" fn wasm_config_new() -> Box<wasm_config_t> {
     Box::new(wasm_config_t::default())
 }
 
-/// Configure the compiler to use.
+/// Updates the configuration to specify a particular compiler to use.
```
jubianchi, line 118: Should we add something that makes it clear for users that this is not part of the standard Wasm C API?

lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -97,7 +156,38 @@ pub extern "C" fn wasm_config_set_compiler(
     config.compiler = compiler;
 }
 
-/// Configure the engine to use.
+/// Updates the configuration to specify a particular engine to use.
```
jubianchi, line 159: Same question as `wasm_config_set_compiler`

lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -305,3 +459,25 @@ pub extern "C" fn wasm_engine_new_with_config(
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use inline_c::assert_c;
+
+    #[test]
+    fn test_engine_new() {
```
jubianchi, line 468: Is this test really useful? It seems redundant with the examples from the documentation.

lib/c-api/src/wasm_c_api/instance.rs
```diff
@@ -7,11 +7,28 @@ use std::mem;
...
+///
+/// ## Errors
+///
+/// The function can fail in 2 ways, as defined by the specification:
+///
+/// 1. Link errors that happen when plugging the imports into the
+///    instance,
+/// 2. Runtime errors that happen when running the module `start`
+///    function.
+///
```
jubianchi, line 27: I think we should be clear about the fact that errors will not result in panics but in the `traps` being set.

Also, you are refering to the "specification": it could be a good thing to add a link to it don't you think?

lib/c-api/src/wasm_c_api/engine.rs
Hywan: I initially wanted to improve this test, but our C API doesn't offer enough features. Let's keep this as is, and I'll improve it later.

lib/c-api/src/wasm_c_api/engine.rs
Hywan: Done.

lib/c-api/src/wasm_c_api/engine.rs
Hywan: done.

lib/c-api/src/wasm_c_api/instance.rs
Hywan: I don't find the paragraph in the spec anymore. I've removed that part of the sentence :-).

Hywan:
bors try

Hywan:
bors try

Hywan:
So. I pushed a commit. Git says everything is up-to-date. But the commit doesn't appear here. Hmmm.

Hywan:
bors try

jubianchi:
bors try

jubianchi (approved):

PR #1853, opened 2020-12-01
fix(c-api) Memory leak around `wasm_$name_vec_delete`

Hywan:
# Description

WIP

# Review

- [ ] Add a short description of the the change to the CHANGELOG.md file

PR #1855, opened 2020-12-01
Fix memory leak in wat2wasm function

MarkMcCaskey:
Alternative to #1853 ; 

Found with @nlewycky 

# Review

### DIFF
--- a/lib/c-api/src/wasm_c_api/wat.rs
+++ b/lib/c-api/src/wasm_c_api/wat.rs
@@ -6,11 +6,20 @@ use super::types::wasm_byte_vec_t;
 /// In case of failure, `wat2wasm` returns `NULL`.
 #[cfg(feature = "wat")]
 #[no_mangle]
-pub unsafe extern "C" fn wat2wasm(wat: &wasm_byte_vec_t) -> Option<Box<wasm_byte_vec_t>> {
-    let wat: &[u8] = wat.into_slice()?;
-    let result: wasm_byte_vec_t = c_try!(wasmer::wat2wasm(wat)).into_owned().into();
+pub unsafe extern "C" fn wat2wasm(wat: &wasm_byte_vec_t, out: &mut wasm_byte_vec_t) {
+    let wat: &[u8] = match wat.into_slice() {
+        Some(v) => v,
+        _ => return,
+    };
+    let result: wasm_byte_vec_t = match wasmer::wat2wasm(wat) {
+        Ok(val) => val.into_owned().into(),
+        Err(err) => {
+            crate::error::update_last_error(err);
+            return;
+        }
+    };
 
-    Some(Box::new(result))
+    *out = result;
 }
 
 #[cfg(test)]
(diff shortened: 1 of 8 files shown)


lib/c-api/src/wasm_c_api/wat.rs
```diff
@@ -6,11 +6,20 @@ use super::types::wasm_byte_vec_t;
...
+    let wat: &[u8] = match wat.into_slice() {
+        Some(v) => v,
+        _ => return,
+    };
+    let result: wasm_byte_vec_t = match wasmer::wat2wasm(wat) {
+        Ok(val) => val.into_owned().into(),
+        Err(err) => {
+            crate::error::update_last_error(err);
+            return;
+        }
```
nlewycky, line 19: How does the user know whether `wat2wasm` succeeded? Is there a way to initialize a `wasm_byte_vec_t` to "clear" which the user can then test for? Or does it always check for errors with a separate "did the last thing fail" call after every call to `wat2wasm`?

MarkMcCaskey: That's a fair point, in this case I'd assume the user would initialize to null prior and check for that, but perhaps it makes more sense for us to zero the fields for them on failure

nlewycky: You can also use the return value if you like.

lib/c-api/tests/wasm_c_api/test-wat2wasm.c
```diff
@@ -20,31 +20,32 @@ int main(int argc, const char* argv[]) {
     .data = "(module)",
     .size = 8,
   };
-  wasm_byte_vec_t *wasm = wat2wasm(&wat);
+  wasm_byte_vec_t wasm;
+  wat2wasm(&wat, &wasm);
 
-  if (!wasm) {
+  if (!wasm.data) {
     printf("> Error compiler WAT to Wasm!\n");
```
nlewycky, line 27:
```suggestion
    printf("> Error compiling WAT to Wasm!\n");
```

lib/c-api/wasmer_wasm.h
```diff
@@ -239,6 +239,6 @@ int wasmer_last_error_message(char *buffer, int length);
  *
  * In case of failure, `wat2wasm` returns `NULL`.
```
nlewycky, line 240:
```suggestion
 * In case of failure, `wat2wasm` fills the `out->data = NULL` and `out->size = 0`.
```

nlewycky (approved):

lib/c-api/wasmer_wasm.h
```diff
@@ -237,8 +237,8 @@ int wasmer_last_error_message(char *buffer, int length);
  * Parses in-memory bytes as either the WAT format, or a binary Wasm
  * module. This is wasmer-specific.
  *
- * In case of failure, `wat2wasm` returns `NULL`.
+ * In case of failure, `wat2wasm` fills the `out->data = NULL` and `out->size = 0`.
```
nlewycky: Grammar here is a little off, but meh.

MarkMcCaskey: I fixed it!

Hywan:
:+1:

PR #1863, opened 2020-12-02
Remove knowledge of middleware from the compilers

nlewycky:
# Description
This pushes knowledge of middlewares out of the compilers and into the engines.

### DIFF
--- a/lib/cli/Cargo.toml
+++ b/lib/cli/Cargo.toml
@@ -54,8 +54,8 @@ default = [
     "wat",
     "wast",
     "jit",
-    "native",
-    "object-file",
+#    "native",
+#    "object-file",
     "cache",
     "wasi",
     "emscripten",
@@ -81,8 +81,8 @@ wat = ["wasmer/wat"]
 compiler = [
     "wasmer-compiler/translator",
     "wasmer-engine-jit/compiler",
-    "wasmer-engine-native/compiler",
-    "wasmer-engine-object-file/compiler",
+#    "wasmer-engine-native/compiler",
+#    "wasmer-engine-object-file/compiler",
 ]
 experimental-io-devices = [
     "wasmer-wasi-experimental-io-devices",
--- a/lib/compiler/src/compiler.rs
+++ b/lib/compiler/src/compiler.rs
@@ -9,6 +9,7 @@ use crate::module::CompileModuleInfo;
 use crate::target::Target;
 use crate::translator::ModuleMiddleware;
 use crate::FunctionBodyData;
+use crate::MiddlewareBinaryReader;
 use crate::ModuleTranslationState;
 use crate::SectionIndex;
 use wasmer_types::entity::PrimaryMap;
@@ -87,7 +88,7 @@ pub trait Compiler {
         module: &'module mut CompileModuleInfo,
         module_translation: &ModuleTranslationState,
         // The list of function bodies
-        function_body_inputs: PrimaryMap<LocalFunctionIndex, FunctionBodyData<'data>>,
+        function_body_inputs: PrimaryMap<LocalFunctionIndex, MiddlewareBinaryReader>,
     ) -> Result<Compilation, CompileError>;
 
     /// Compiles a module into a native object file.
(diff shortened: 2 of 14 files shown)


lib/cli/Cargo.toml
```diff
@@ -54,8 +54,8 @@ default = [
     "wat",
     "wast",
     "jit",
-    "native",
-    "object-file",
+#    "native",
+#    "object-file",
```
syrusakbary, line 58: Uncomment me

lib/cli/Cargo.toml
```diff
@@ -81,8 +81,8 @@ wat = ["wasmer/wat"]
 compiler = [
     "wasmer-compiler/translator",
     "wasmer-engine-jit/compiler",
-    "wasmer-engine-native/compiler",
-    "wasmer-engine-object-file/compiler",
+#    "wasmer-engine-native/compiler",
+#    "wasmer-engine-object-file/compiler",
```
syrusakbary, line 85: Uncomment me

lib/cli/Cargo.toml
nlewycky: Done.

lib/cli/Cargo.toml
nlewycky: Done.

lib/compiler/src/compiler.rs
```diff
@@ -99,12 +99,14 @@ pub trait Compiler {
...
         _symbol_registry: &dyn SymbolRegistry,
         // The metadata to inject into the wasmer_metadata section of the object file.
         _wasmer_metadata: &[u8],
-    ) -> Option<Result<Vec<u8>, CompileError>> {
-        None
+    ) -> Result<Vec<u8>, CompileError> {
+        Err(CompileError::Codegen(
+            "This backend does not support `experimental_native_compile_module`.".into(),
+        ))
     }
```
syrusakbary, line 110: We return None in propose, so it can fallback to the non-experimental one

nlewycky: The API changed to take the function body data by move instead of by reference, so once moved you no longer hold the data to make the fallback call. In its place, a "supports_experimental_native_compile_module" method is provided so that the caller can decide which to use and then move the data once.

Why does it now take by move instead of by reference? Previously the PrimaryMap was from local function index to raw wasm data that was then interpreted, but now it's to a streaming interface. To use a streaming interface you call something like a `get_next()` method which returns different data each time. To do that, the streaming object must be mutating under the hood, so a non-mutating reference is no longer a choice.

It might be possible to make this work with a mutable reference, I didn't try that.

lib/engine-jit/src/artifact.rs
```diff
@@ -79,6 +84,23 @@ impl JITArtifact {
 
         let compiler = inner_jit.compiler()?;
 
+        let function_body_inputs = translation
+            .function_body_inputs
+            .iter()
```
Hywan, line 89: Can we use `par_iter` here?

lib/engine-jit/src/builder.rs
```diff
@@ -40,6 +44,21 @@ impl<'a> JIT<'a> {
...
+    }
+
+    /// Append a stack of middlewares
+    pub fn middlewares<I: Iterator<Item = Arc<dyn ModuleMiddleware>>>(
+        mut self,
+        middlewares: I,
+    ) -> Self {
+        self.middleware.extend(middlewares);
+        self
+    }
```
Hywan, line 60: :+1:

lib/compiler/src/compiler.rs
```diff
@@ -87,9 +82,14 @@ pub trait Compiler {
         module: &'module mut CompileModuleInfo,
         module_translation: &ModuleTranslationState,
         // The list of function bodies
-        function_body_inputs: PrimaryMap<LocalFunctionIndex, FunctionBodyData<'data>>,
+        function_body_inputs: PrimaryMap<LocalFunctionIndex, MiddlewareBinaryReader>,
     ) -> Result<Compilation, CompileError>;
 
+    /// Whether this compiler supports the experimental native object file method.
+    fn supports_experimental_native_compile_module(&self) -> bool {
```
Hywan, line 89:
```suggestion
    const fn supports_experimental_native_compile_module(&self) -> bool {
```

Hywan (approved):
I like this PR very much. I believe the resulting code is cleaner that way. The middleware logic is centralized inside engines, not dispersed between engines and compilers.

Few comments about small details.

Issue #1714, opened 2020-10-12
Memory leak when creating new Store, Module and Instance


evq:
### Describe the bug
Memory leak

### Steps to reproduce
Reproduces using the following variation of https://github.com/wasmerio/wasmer-python/blob/master/examples/appendices/simple.py

```
from wasmer import Store, Module, Instance
import os

__dir__ = os.path.dirname(os.path.realpath(__file__))

while True:
    module = Module(Store(), open(__dir__ + '/simple.wasm', 'rb').read())

    instance = Instance(module)
    result = instance.exports.sum(1, 2)

    print(result) # 3!
```

### Expected behavior

The program should not leak memory.

### Actual behavior

Memory is leaked eventually leading to:
```
Traceback (most recent call last):
  File "simple.py", line 9, in <module>
    instance = Instance(module)
RuntimeError: Insufficient resources: Failed to create memory: Error when allocating memory: Cannot allocate memory (os error 12)
```

MarkMcCaskey:
Related #1584 #1667 , I think all these are referring to the same bug

Issue #1667, opened 2020-10-02
Environment passed to Function::new_native_with_env is never dropped and leaks memory


bkolobara:
I'm running wasmer `1.0.0-alpha.3` and recently run into a problem that I'm not sure how to work around. For every instance I crate a new `ImportObject` capturing an environment with some functions.

The issue is that this environment is never dropped and I can't figure out a way of releasing the memory. My environment is a bit heavier and after running the app for some time I start running out of memory.

I created a minimal example just to illustrate the issue:
```rust
use wasmer::{Store, Instance, Module, Exports, ImportObject, Function};

struct Env {
    multiplier: i32,
}

impl Drop for Env {
    fn drop(&mut self) {
        println!("Dropping: {}", self.multiplier);
    }
}

fn main() {
    let store = Store::default();
    let module = Module::new(&store, "(module (import \"env\" \"sum_and_multiply\" (func (param i32 i32) (result i32))))").unwrap();

    let env_dropped = Env { multiplier: 1 };
    let env_never_dropped = Env { multiplier: 2 };
    let f = Function::new_native_with_env(&store, env_never_dropped, sum_and_multiply);

    let mut exports = Exports::new();
    exports.insert("sum_and_multiply", f);

    let mut import_object = ImportObject::new();
    import_object.register("env", exports);

    let instance = Instance::new(&module, &import_object).unwrap();
}

fn sum_and_multiply(env: &mut Env, a: i32, b: i32) -> i32 {
    (a + b) * env.multiplier
}
```

The output of this app is:
```
Dropping: 1
```

As you can see `env_dropped` is never moved and drops, but `env_never_dropped` doesn't get it's `drop` method called. Even after the instance, module and store are all gone.

Ideally I would expect that once there are no more `Instances` referencing the `ImportObject` that  the resources get released.

It can also be that I structured my app in an unconventional way. Is there maybe a better approach to this?

bkolobara:
I think this is related to #1584.

MarkMcCaskey:
Thanks for the issue, I think this is the same as #1584 too;

I believe once either #1663 or #1625 lands this will be easier to fix as the general ownership of Env is clarified a bit more in those PRs.

We can probably make a solution that's compatible with both of those PRs though

jubianchi:
Closing this as #1865 got merged and it fixes the issue. Here is the original snippet, adapted with new requirements:

```rust
[...]

PR #1865, opened 2020-12-03
Fix memory leak in host function envs

### LINKED ISSUE #1584: vmctx leaks?
It seems that `vmctx` is memory-leaked here:

1.  `Function::new_with_env` 
https://github.com/wasmerio/wasmer/blob/518e0f32226abffff78d01af195d6de0dbaea00b/lib/api/src/externals/function.rs#L137

2. `Function::new_native_with_env` https://github.com/wasmerio/wasmer/blob/518e0f32226abffff78d01af195d6de0dbaea00b/lib/api/src/externals/function.rs#L230

MarkMcCaskey:
TODO: link to issue

This PR contains a number of changes:

1. Make `WasmerEnv: Clone`
2. Store a pointer to the `clone` function when creating a host function (Notably this is a feature that wouldn't work even if we _could_ use a proper trait object because you can't have a `Sized` trait object and `Clone: Sized`).
3. Store a pointer to the `drop` function when creating a host function.
4. Clone the env via pointer every time an `Instance` is made. Therefore each `Instance` gets its own, unique `Env` per host function with `Env`.
5. Add reference counting and drop logic to a sub-field of `wasmer_export::ExportFunction` which frees the original version of the `Env` (the thing that gets cloned each time an `Instance` is made) with the `drop` function pointer.
6. Change some logic in `vm::Instance` from SoA (struct of arrays) to AoS (array of structs): this uses more memory but is a bit less error prone and can be easily changed later.
7. Add logic on this new struct (`vm::ImportEnv`) that contains the function pointers for each import in `Instance` to drop (with the `drop` fn pointer) when the `vm::Instance` is being dropped. This fixes the original memory leak.
8. Add wrapper functions inside the host function creation functions which makes the layout of the user supplied env-pointer the responsibility of each function.  Thus, rather than `drop` being `Env::drop`, it's a function which frees all wrapper types, traverses indirections and frees the internal `Env` with `Env::drop`.  This simplifies code at the cost of making the `host_env` pointer (`vmctx`) not consistent in terms of what it actually points to.  This change fixes another memory leak related to the creation of host functions.

tl;dr: we're leaning into manually doing virtual method dispatch on `WasmerEnv`s and it actually works great! The biggest issue I have with the PR as-is is that the code isn't as clean/readable/robust as I'd ideally like it to be.

Edit (by @Hywan): This PR fixes #1584, #1714, #1865, #1667.

# Review


lib/engine/src/export.rs
```diff
@@ -49,12 +52,14 @@ impl From<VMExport> for Export {
...
     /// The VM function, containing most of the data.
-    pub vm_function: VMExportFunction,
+    pub vm_function: Arc<VMExportFunction>,
     /// Function pointer to `WasmerEnv::init_with_instance(&mut self, instance: &Instance)`.
     ///
     /// This function is called to finish setting up the environment after
     /// we create the `api::Instance`.
     pub import_init_function_ptr: Option<ImportInitializerFuncPtr>,
+    /// The destructor to free the host environment.
+    pub host_env_drop_fn: Option<fn(*mut std::ffi::c_void)>,
```
syrusakbary, line 62: A function can be exported twice. How do you verify that `host_env_drop_fn` is only called once?

syrusakbary (commented):
A function can be exported twice. How do you verify that `host_env_drop_fn` is only called once?

lib/engine/src/export.rs
MarkMcCaskey: Do we have any tests for this? I believe this should behave the same as initialization, which should also only run once

MarkMcCaskey:
bors try

Cargo.toml
```diff
@@ -15,7 +15,7 @@ wasmer-compiler = { version = "1.0.0-beta1", path = "lib/compiler" }
 wasmer-compiler-cranelift = { version = "1.0.0-beta1", path = "lib/compiler-cranelift", optional = true }
 wasmer-compiler-singlepass = { version = "1.0.0-beta1", path = "lib/compiler-singlepass", optional = true }
 wasmer-compiler-llvm = { version = "1.0.0-beta1", path = "lib/compiler-llvm", optional = true }
-wasmer-emscripten = { version = "1.0.0-beta1", path = "lib/emscripten", optional = true }
+#wasmer-emscripten = { version = "1.0.0-beta1", path = "lib/emscripten", optional = true }
```
nlewycky, line 18: Not that I think you've forgotten, but since this isn't a draft PR, I'll note that this PR can't land with this change.

lib/api/src/env.rs
```diff
@@ -94,29 +95,37 @@ impl WasmerEnv for isize {}
...
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicI8 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicU8 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicI16 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicU16 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicI32 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicU32 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicI64 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicUsize {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicIsize {}
+//impl WasmerEnv for dyn ::std::any::Any + Clone {}
```
nlewycky, line 108: What do you think about this? Re-enable it? It's not obvious to me what the end-user API impact of having or removing this is.

lib/api/src/env.rs
```diff
@@ -94,29 +95,37 @@ impl WasmerEnv for isize {}
...
+        guard.init_with_instance(instance)
     }
 }
 
+/*impl<T: WasmerEnv> WasmerEnv for &'static T {
+    fn init_with_instance(&mut self, instance: &Instance) -> Result<(), HostEnvInitError> {
+        T::init_with_instance()
+        (*self).init_with_instance(instance)
+    }
+}*/
```
nlewycky, line 127: This should be safe to remove, we don't want to have this code back, right?

lib/api/src/externals/function.rs
```diff
@@ -87,23 +88,41 @@ impl Function {
     where
         F: Fn(&[Val]) -> Result<Vec<Val>, RuntimeError> + 'static,
     {
-        let dynamic_ctx = VMDynamicFunctionContext::from_context(VMDynamicFunctionWithoutEnv {
-            func: Box::new(func),
-            function_type: ty.clone(),
-        });
+        let dynamic_ctx: VMDynamicFunctionContext<VMDynamicFunctionWithoutEnv> =
```
nlewycky, line 91: Visually the type here is redundant with the same two types written on the next line. Does rustc need this?

lib/api/src/externals/function.rs
```diff
@@ -714,9 +794,10 @@ pub(crate) trait VMDynamicFunction {
     fn function_type(&self) -> &FunctionType;
 }
 
+#[derive(Clone)]
 pub(crate) struct VMDynamicFunctionWithoutEnv {
     #[allow(clippy::type_complexity)]
-    func: Box<dyn Fn(&[Val]) -> Result<Vec<Val>, RuntimeError> + 'static>,
+    func: Arc<dyn Fn(&[Val]) -> Result<Vec<Val>, RuntimeError> + 'static>,
```
nlewycky, line 800: I'm surprised to see `Arc` in a VM struct. Either this shouldn't be a VM struct because the compiler doesn't need to read its contents, or the Arc is actually an FFI safe type, or maybe this one field of this struct isn't used by the compiler and that's OK and we should put a comment on it indicating as much.

lib/engine/src/export.rs
```diff
@@ -44,17 +46,44 @@ impl From<VMExport> for Export {
     }
 }
 
+/// TODO: rename, etc.
+#[derive(Debug, PartialEq)]
+pub struct ExportFunctionMetadata {
+    /// duplicated here so we can free it....
```
nlewycky, line 52: This comment needs some more exposition. (Duplicated from where? So who can free it?)

lib/engine/src/export.rs
```diff
@@ -44,17 +46,44 @@ impl From<VMExport> for Export {
...
+    pub host_env_clone_fn: fn(*mut std::ffi::c_void) -> *mut std::ffi::c_void,
+    /// The destructor to free the host environment.
+    pub host_env_drop_fn: fn(*mut std::ffi::c_void),
+}
+
+// We have to free `host_env` here because we always clone it before using it
+// so all the `host_env`s freed at the `Instance` level won't touch the original.
+impl Drop for ExportFunctionMetadata {
+    fn drop(&mut self) {
+        dbg!("DROPPING ORIGINAL HOST ENV!");
```
nlewycky, line 72: dbg!

lib/api/src/env.rs
MarkMcCaskey: It's not possible to do this, because `Clone` requires `Sized` and you can't have trait objects that are sized. That code is not allowed by Rust

lib/api/src/externals/function.rs
MarkMcCaskey: This code is high risk for very bad bugs as it's doing a lot of nasty low level stuff.  I explicitly adding the type did actually fix a bug here before I believe, but I think it's best just to be extremely explicit to the compiler and the reader what the precise types of everything is here.

The bug before was that I was calling `clone` but that `clone` was actually going to `&T` and cloning the reference, not the `T` itself.

I agree that it's redundant, but I think it's probably fine to be excessively explicit when doing dangerous low-level things

lib/api/src/externals/function.rs
MarkMcCaskey: So this function has the name "VM" but it's actually in the API crate.

The only way this is interacted with at the compiler / vm levels is via `Clone` I believe (and perhaps passing a pointer to the `VMDynamicFunctionWithoutEnv` type itself around)

lib/engine/src/export.rs
MarkMcCaskey: Good call, yeah I'll improve the docs.

The field here is never used here outside of `Drop`, it's only there for `Drop`.

lib/api/src/externals/function.rs
MarkMcCaskey: @syrusakbary thoughts on the names of these types? I agree with Nick that having types that start with `VM` in API is a bit confusing to start. Also as Nick said, types that start with `VM` that aren't meant for low level access seems bad too.

syrusakbary: Agreed, perhaps the `VM` part of the name should be removed

lib/api/src/externals/function.rs
```diff
@@ -87,23 +88,41 @@ impl Function {
     where
         F: Fn(&[Val]) -> Result<Vec<Val>, RuntimeError> + 'static,
     {
-        let dynamic_ctx = VMDynamicFunctionContext::from_context(VMDynamicFunctionWithoutEnv {
-            func: Box::new(func),
-            function_type: ty.clone(),
-        });
+        let dynamic_ctx: VMDynamicFunctionContext<DynamicFunctionWithoutEnv> =
```
Hywan, line 91: I bet this type annotation isn't required (regardless, I prefer having type annotation in this kind of code, so it's perfectly fine to keep it, just wondered whether it changes something).

lib/api/src/externals/function.rs
```diff
@@ -727,19 +808,27 @@ impl VMDynamicFunction for VMDynamicFunctionWithoutEnv {
...
 }
 
-#[repr(C)]
-pub(crate) struct VMDynamicFunctionWithEnv<Env>
+pub(crate) struct DynamicFunctionWithEnv<Env>
 where
     Env: Sized + 'static,
 {
-    // This field _must_ come first in this struct.
-    env: Box<Env>,
```
Hywan, line 736: Why `env` is no longer required to be at first position?

lib/api/src/types.rs
```diff
@@ -76,7 +76,7 @@ impl ValFuncRef for Val {
         let export = wasmer_engine::ExportFunction {
             // TODO:
             // figure out if we ever need a value here: need testing with complicated import patterns
-            import_init_function_ptr: None,
```
Hywan, line 79: Is the `TODO` above still valid?

lib/engine/src/export.rs
```diff
@@ -44,17 +46,57 @@ impl From<VMExport> for Export {
     }
 }
 
+/// Extra metadata about `ExportFunction`s.
+///
+/// The metadata acts as a kind of manual virtual dispatch. We store the
+/// user-supplied `WasmerEnv` as a void pointer and have methods on it
+/// that have been adapted to accept a void pointer.
+#[derive(Debug, PartialEq)]
+pub struct ExportFunctionMetadata {
```
Hywan, line 55: I like that.

lib/engine/src/export.rs
```diff
@@ -44,17 +46,57 @@ impl From<VMExport> for Export {
...
+    /// the `Drop` implementation. This field is the "master copy" of the env,
+    /// that is, the original env passed in by the user. Every time we create
+    /// an `Instance` we clone this with the `host_env_clone_fn` field.
+    ///
+    /// Thus, we only bother to store the master copy at all here so that
+    /// we can free it.
+    ///
+    /// See `wasmer_vm::export::VMExportFunction::vmctx` for the version of
+    /// this pointer that is used by the VM when creating an `Instance`.
+    pub host_env: *mut std::ffi::c_void,
```
Hywan, line 68: Idea: Create a `new` constructor method rather than having all fields `pub`. I think it could be “dangerous”, it's better to hide everything here. Thoughts?

Hywan: Same for all the fields in this type.

lib/engine/src/export.rs
```diff
@@ -44,17 +46,57 @@ impl From<VMExport> for Export {
...
 #[derive(Debug, Clone, PartialEq)]
 pub struct ExportFunction {
     /// The VM function, containing most of the data.
     pub vm_function: VMExportFunction,
-    /// Function pointer to `WasmerEnv::init_with_instance(&mut self, instance: &Instance)`.
-    ///
-    /// This function is called to finish setting up the environment after
-    /// we create the `api::Instance`.
-    pub import_init_function_ptr: Option<ImportInitializerFuncPtr>,
+    /// TODO:
```
Hywan, line 98: What's left to do?

lib/vm/src/instance.rs
```diff
@@ -109,6 +105,45 @@ pub(crate) struct Instance {
     vmctx: VMContext,
 }
 
+/// TODO: figure out what to do with Clone here
+///       We can probably remove the Arc... just need to figure that out.
+/// TODO:
```
Hywan, line 110: 2 `TODO`, we are lucky! 😉

lib/vm/src/instance.rs
```diff
@@ -109,6 +105,45 @@ pub(crate) struct Instance {
...
+#[derive(Debug)]
+pub struct ImportEnv {
+    /// TODO:
+    pub env: *mut std::ffi::c_void,
+    /// TODO:
+    pub clone: Option<fn(*mut std::ffi::c_void) -> *mut std::ffi::c_void>,
```
Hywan, line 116: Why is `clone` optional?

Hywan (commented):
I like the PR. In some places it simplifies code, in other, it's a little bit harder to understand. I think another refactoring round would be required in a close future, but let's keep the priority as is: fixing the memory leak.

There is a lot of `TODO` in this PR. Are you going to fix them all? I believe a majority should be fixed before merging this PR.

Hywan:
This PR fixes #1584, #1714, #1865.

lib/api/src/externals/function.rs
MarkMcCaskey: Not strictly no, I thought it was helpful for understanding what's happening with the unsafe block here though.

Actually all these wrapper functions can be created by a helper function, so I should probably just do that

lib/api/src/externals/function.rs
MarkMcCaskey: The reason before was due to a bad hack where we'd double deref a pointer ignoring the fact that it was in a struct.

We now have each function that is called on the `env` traverse things correctly, so for example we create a wrapper function where we walk through the types and find what we need.  So if we need to operate on the `Box<env>` itself, we can find it in each function.  I think this is a simpler solution because then all the functions always just work on the `env`, we had an issue where the clone and drop function needed to be called on the top level the wrapper dynamic func context thing but the initialize function needed to be called on the inner env.  This fixes that by making the thing each function gets consistent.

lib/engine/src/export.rs
MarkMcCaskey: Oh documentation, we have some lints which prevent the code from compiling without doc comments everywhere so I just write `TODO` when I don't want to write doc-comments while still developing things.  I'll fill them out now though

lib/api/src/types.rs
MarkMcCaskey: Yeah, I'm not 100% confident in this code which is why that comment is still there, I can investigate it later... I'm not aware of a good way to test this yet, I can investigate that though

lib/vm/src/instance.rs
MarkMcCaskey: That's a good question! I believe I've clarified this now by making all 3 states explicit with an `enum`!

lib/api/src/env.rs
```diff
@@ -94,26 +95,26 @@ impl WasmerEnv for isize {}
...
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicBool {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicI8 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicU8 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicI16 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicU16 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicI32 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicU32 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicI64 {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicUsize {}
+impl<'a> WasmerEnv for &'a ::std::sync::atomic::AtomicIsize {}
```
syrusakbary, line 107: Why we need the `<'a>` ?

lib/vm/src/instance.rs
```diff
@@ -100,7 +95,8 @@ pub(crate) struct Instance {
     ///
     /// TODO: Be sure to test with serialize/deserialize and imported
     /// functions from other Wasm modules.
-    import_initializers: ImportInitializerThunks,
+    /// TODO: update this comment
```
syrusakbary, line 98: Fix me

lib/engine/src/artifact.rs
```diff
@@ -96,7 +96,7 @@ pub trait Artifact: Send + Sync + Upcastable {
 
         let module = self.module();
         let (instance_ptr, offsets) = InstanceHandle::allocate_instance(&module);
-        let (imports, import_initializers) = {
+        let (imports, import_envs) = {
```
syrusakbary, line 99: I think a var like `import_function_env_initializers` is a bit more clear. Thoughts?

lib/vm/src/instance.rs
```diff
@@ -109,6 +105,139 @@ pub(crate) struct Instance {
     vmctx: VMContext,
 }
 
+/// A collection of data about host envs.
+///
+/// Due to invariants in the `HostEnv` variant, this type should only
+/// be constructed by one of the given methods, not directly.
+#[derive(Debug)]
+pub enum ImportEnv {
```
syrusakbary, line 113: `VMImportFunctionEnv` might be a better name

lib/vm/src/instance.rs
```diff
@@ -100,7 +95,8 @@ pub(crate) struct Instance {
     ///
     /// TODO: Be sure to test with serialize/deserialize and imported
     /// functions from other Wasm modules.
-    import_initializers: ImportInitializerThunks,
+    /// TODO: update this comment
+    import_envs: BoxedSlice<FunctionIndex, ImportEnv>,
```
syrusakbary, line 99: `imported_function_envs`

lib/vm/src/imports.rs
```diff
@@ -68,25 +63,9 @@ impl Imports {
...
-                    (func_init, host_env)
-                })
-                .collect()
-        } else {
-            vec![]
-        };
-        // ensure we only call these functions once and free this now useless memory.
-        self.host_function_env_initializers = None;
-        result
+    pub fn get_import_initializers(&mut self) -> BoxedSlice<FunctionIndex, ImportEnv> {
```
syrusakbary, line 66: `get_imported_function_initializers`

lib/api/src/types.rs
syrusakbary: We should investigate. I bet is fine as None, but it will be good to verify

lib/engine/src/artifact.rs
MarkMcCaskey: hmm, I think `import_envs` makes more sense because a bunch of things called `ImportEnv`, it's more than just the initializers now

MarkMcCaskey: I'm open to renaming the type though if `ImportEnv` isn't descriptive enough

syrusakbary: But they are only relevant on the context of a function, right?

MarkMcCaskey: Yeah, I was more objecting to the `initializers` part, I think `function` is good 👍  saw your other comments, I'll update some names

lib/vm/src/instance.rs
MarkMcCaskey: I left `VM` off of the type because of something Nick said a few days or weeks ago but what his mental model around `VM` types are: that they should be `repr(C)` and are directly interacted with by the VM.  I think there's a decent argument that this isn't a VM type and it doesn't even need to be defined in `wasmer_vm` actually, we could put it in `wasmer_types` because it doesn't contain anything VM-specific.

lib/vm/src/instance/mod.rs
```diff
@@ -113,6 +107,139 @@ pub(crate) struct Instance {
...
 
+/// A collection of data about host envs.
+///
+/// Due to invariants in the `HostEnv` variant, this type should only
+/// be constructed by one of the given methods, not directly.
+#[derive(Debug)]
+pub enum ImportFunctionEnv {
+    /// The `vmctx` pointer does not refer to a host env, there is no
+    /// metadata about it.
+    NoEnv,
```
syrusakbary, line 118: I think is better to name this: `VMEnv`.
And have the `HostEnv` to be an enum with all the different options:
* NoEnv
* Env

Also, the VM should not be aware of dynamic functions or not. So I believe `DynamicHostEnvWithNoInnerEnv` should be done via `HostEnv`

lib/vm/src/imports.rs
```diff
@@ -17,9 +17,7 @@ pub struct Imports {
     /// space may affect Wasm runtime performance due to increased cache pressure.
     ///
     /// We make it optional so that we can free the data after use.
-    pub host_function_env_initializers: Option<
-        BoxedSlice<FunctionIndex, (Option<ImportInitializerFuncPtr>, *mut std::ffi::c_void)>,
-    >,
+    pub host_function_env_initializers: Option<BoxedSlice<FunctionIndex, ImportFunctionEnv>>,
```
syrusakbary, line 20: This should have the same type as `functions` field, this means that there should be no need to wrap things with an `Option`.

```suggestion
    pub host_function_env_initializers: BoxedSlice<FunctionIndex, ImportFunctionEnv>,
```

lib/vm/src/instance/mod.rs
MarkMcCaskey: Same thought on the `VM` prefix, I'm not sure it makes sense to use it on types that aren't related to the VM.

Yeah, the extra case here is just to make the multiple cases more explicit.  Otherwise there's data that's sometimes there and sometimes not and it's not exactly clear why. This enum is more descriptive than prescriptive: it's describing the state of the world as it is, not what it should be. Perhaps we should move this type out of `vm`, there's no reason it has to be defined in VM.

syrusakbary: For the VM, there is not such thing as DynamicHostEnv. This is something that we do in the engine level, not VM

syrusakbary: Meaning: the `vm` only handles the functions with their end ABI. Is the engine the one that creates the trampolines and so.

lib/vm/src/imports.rs
MarkMcCaskey: The use of option here is for correctness / hinting at access patterns and to reduce the amount of dead memory we have: once we get this thing once it's both incorrect to access it again (in regards to initializers) and wasteful to keep it around here.

Previously, the initializers would also be destroyed after initialization making it impossible to call it multiple times.  I believe this PR changes that in order to keep some of the code simpler, but it may be worth changing it back.

lib/vm/src/instance/mod.rs
syrusakbary: Note: there is some "dangling" code in the VM that is aware of dynamic functions. But thanks to the PR where we move most of the function stuff to the engine, it should be easy to move the rest as well.

lib/vm/src/imports.rs
MarkMcCaskey: When certain data has precise semantics I find the mental model of "consumable" data to be a good one because it's more explicit about what's happening and it also reduces peak memory usage of data that is no good to anyone anymore.  The data can only be misused by keeping it around, so maybe it's better to just get rid of it.

syrusakbary: I might have missed, where do we consume the `host_function_env_initializers` field?

syrusakbary: Oh, I gues the `take` in `get_imported_function_envs` consumes it?

syrusakbary: If so, can we comment that on the `host_function_env_initializers`. It was not trivial for me to realize that the field was consumable!

MarkMcCaskey: I agree, it may be better to create a wrapper type that's more explicit than `Option` in wasmer_types or something.

`UseOnce` or something, something that implies that it always exists at the start but once it's used, it's gone

syrusakbary: I think we can use `Option` for now (it will be simpler!), but just with a clear comment on why is that way :)

syrusakbary (approved):

examples/imports_function_env.rs
```diff
@@ -72,15 +71,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // We derive a default implementation of `WasmerEnv` here.
     #[derive(WasmerEnv, Clone)]
     struct Env {
-        counter: Arc<RefCell<i32>>,
+        counter: Arc<Mutex<i32>>,
```
syrusakbary, line 74: Why do we need to change the examples? The previous examples should work as is. Or am I missing something?

MarkMcCaskey: Yeah, the previous examples were incorrect, we just didn't enforce it. `WasmerEnv` needs to be `Send` and `Sync` because `Instance` is `Send` and `Sync` and we don't synchronize access to functions. So it's possible in safe Rust to get a `NativeFunc` on multiple threads that points to the same underlying function.  In order for the `WasmerEnv` to be able to handle being used this way, it must be both `Send` and `Sync`.  This shouldn't affect much if any real code. For example, WASI was not affected at all by this change.  Emscripten was only affected because it was doing some weird stuff with pointers.

webmaster128:
> Edit (by @Hywan): This PR fixes #1584, #1714, #1865, #1667.

Note that GitHub only closes one ticket per keyword. In order to connect all of them, you need something like: fixes #1584, fixes #1714, fixes #1865, fixes #1667.

PR #1867, opened 2020-12-03
feat(examples): Add metering example

jubianchi:
# To DO

- [x] How to use metering in a module
- [x] How to meter a function call (see cost of gas before and after)
- [x] Demonstrate that if the gas is over the given limit, it should fail.

# Review

### DIFF
--- a/examples/metering.rs
+++ b/examples/metering.rs
@@ -1,0 +1,165 @@
+//! Wasmer will let you easily run Wasm module in a Rust host.
+//!
+//! This example illustrates the basics of using Wasmer metering features:
+//!
+//!   1. How to enable metering in a module
+//!   2. How to meter a specific function call
+//!   3. How to make execution fails if cost exceeds a given limit
+//!
+//! You can run the example directly by executing in Wasmer root:
+//!
+//! ```shell
+//! cargo run --example metering --release --features "cranelift"
+//! ```
+//!
+//! Ready?
+
+use anyhow::bail;
+use std::sync::Arc;
+use wasmer::wasmparser::Operator;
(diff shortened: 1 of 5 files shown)


examples/metering.rs
```diff
@@ -0,0 +1,97 @@
...
+    // Create a Store.
+    // Note that we don't need to specify the engine/compiler if we want to use
+    // the default provided by Wasmer.
+    // You can use `Store::default()` for that.
+    let metering = Metering::new(5, |operator| -> u64 {
+        match operator {
+            Operator::LocalGet { .. }
+            | Operator::I32Add { .. }=> 2,
+            _ => {
+                dbg!(operator);
```
Hywan, line 51: Is it going to be removed?

jubianchi: yes it will, this patch is basically a draft/wip just so I don't loose the code in my stash or somewhere ;)

syrusakbary (commented):
The metering example should show:

1. How to use metering in a module (done)
2. How to meter a function call (see cost of gas before and after)
3. Demonstrate that if the gas is over the given limit, it should fail.

syrusakbary (approved):

examples/metering.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    println!(
+        "Remaining points: {:?}",
+        metering.get_remaining_points(&instance)
+    );
+    Ok(())
+}
+
+#[test]
+fn test_exported_function() -> Result<(), Box<dyn std::error::Error>> {
```
jubianchi, line 133: I'll change that

PR #1874, opened 2020-12-04
Set compiler config to be owned (following wasm-c-api)

syrusakbary:
# Description

Set compiler config to be owned (following wasm-c-api)

# Review

### DIFF
--- a/lib/compiler-llvm/src/config.rs
+++ b/lib/compiler-llvm/src/config.rs
@@ -207,8 +207,8 @@ impl CompilerConfig for LLVM {
     }
 
     /// Transform it into the compiler.
-    fn compiler(&self) -> Box<dyn Compiler + Send> {
-        Box::new(LLVMCompiler::new(&self))
+    fn compiler(self: Box<Self>) -> Box<dyn Compiler + Send> {
+        Box::new(LLVMCompiler::new(*self))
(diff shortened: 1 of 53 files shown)


syrusakbary:
bors try

lib/compiler-llvm/src/config.rs
```diff
@@ -207,8 +207,8 @@ impl CompilerConfig for LLVM {
     }
 
     /// Transform it into the compiler.
-    fn compiler(&self) -> Box<dyn Compiler + Send> {
-        Box::new(LLVMCompiler::new(&self))
+    fn compiler(self: Box<Self>) -> Box<dyn Compiler + Send> {
```
MarkMcCaskey, line 210: Consider using `From`/`Into` traits here; not sure how this is used but that could be a bit cleaner

MarkMcCaskey (approved):
Looks good to me

lib/compiler-llvm/src/config.rs
syrusakbary: I tried, couldn't make it work with `From`. We can do it in a PR apart if necessary

syrusakbary:
Merging directly to alleviate CI. Tests are passing locally

PR #1876, opened 2020-12-04
Add test showing issue with reusing import objects w/ WasmerEnv

MarkMcCaskey:
Run

```
cargo test --release --features "cranelift singlepass" --features "test-cranelift test-jit" -- multi_use_host_fn_manages_memory_correctly --nocapture
```

to reproduce the bug.

notice that the `dbg!` line is called twice, this means the env is being init'd twice, (it won't do anything the second time, so the second instance will see the wrong `Memory`)

This bug is being fixed in #1865

# Review


MarkMcCaskey:
This was fixed and this test case added in #1865

PR #1877, opened 2020-12-04
Fixed compilation issue

syrusakbary:
Fixed compilation issue in master

# Description

# Review

### DIFF
--- a/lib/engine-jit/src/builder.rs
+++ b/lib/engine-jit/src/builder.rs
@@ -4,7 +4,7 @@ use wasmer_compiler::{CompilerConfig, Features, Target};
 /// The JIT builder
 pub struct JIT {
     #[allow(dead_code)]
-    compiler_config: Option<Box<dyn CompilerConfig + Send + Sync>>,
+    compiler_config: Option<Box<dyn CompilerConfig>>,
     target: Option<Target>,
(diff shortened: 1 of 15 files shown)


lib/engine-jit/src/builder.rs
```diff
@@ -4,16 +4,19 @@ use wasmer_compiler::{CompilerConfig, Features, Target};
 /// The JIT builder
 pub struct JIT {
     #[allow(dead_code)]
-    compiler_config: Option<Box<dyn CompilerConfig>>,
+    compiler_config: Option<Box<dyn CompilerConfig + Send + Sync>>,
```
MarkMcCaskey, line 7: It's possible to make `Send` and `Sync` a requirement of the `CompilerConfig` itself so we don't have to write it everywhere.

Depends on if we ever think there will be a need for a `CompilerConfig` that can't be made thread safe

MarkMcCaskey (approved):

lib/engine-jit/src/builder.rs
syrusakbary: Done! Thanks

syrusakbary:
This time I'm merging manually and I did run `make test` successfully locally

PR #1879, opened 2020-12-05
Update lint to use the same Rust version as CI

MarkMcCaskey:
Forgot to this when bumping the version we use in CI, ran into an issue in a PR I'm working on, this will fix it


nlewycky (approved):

MarkMcCaskey:
bors  r+

PR #1883, opened 2020-12-06
Better in headers wasm-c-api

syrusakbary:
# Description

Previously the `wasmer.h` header corresponded to the (now) deprecated Wasmer API.
However, as we move forward and adopted `wasm-c-api`, we couldn't use `wasmer.h` for it because `wasmer.h` and `wasmer_wasm.h` had conflicting names.

Now, this PR solves that by having the old Wasmer API to use the WasmKind enum so people can only use one header always.

# Review


syrusakbary:
bors try

lib/c-api/build.rs
```diff
@@ -79,6 +79,10 @@ fn build_wasm_c_api_headers(crate_dir: &str, out_dir: &str) {
 #if !defined(WASMER_WASM_H_MACROS)
 
 #define WASMER_WASM_H_MACROS
+
+// Include the deprecated API for import compatibility reasons
+#include "wasmer_deprecated.h"
```
Hywan, line 84: I don't know what to think about this right now 🤔.

Hywan (commented):
So the idea is to use `wasmer.h` everytime, whatever the user needds the deprecated API or the standard++ API, is that correct?

syrusakbary:
> So the idea is to use wasmer.h everytime, whatever the user needds the deprecated API or the standard++ API, is that correct?

Yup, that's right!

lib/c-api/build.rs
```diff
@@ -124,19 +128,23 @@ fn build_wasm_c_api_headers(crate_dir: &str, out_dir: &str) {
...
     let mut out_header_file = PathBuf::from(out_dir);
-    out_header_file.push("wasmer");
+    out_header_file.push("wasmer_deprecated");
 
     let mut pre_header = format!(
         r#"// The Wasmer C/C++ header file.
 
 #if !defined(WASMER_H_MACROS)
 
+// Include the WASP API for using the wasm_valkind_enum type.
```
MarkMcCaskey, line 143: `WASP`: is this a typo?

lib/c-api/build.rs
nlewycky, line 84: I'd suggest doing this the other way. Make users of the deprecated API `#include "wasmer_deprecated.h"` which in turn includes (and re-exports) the full `#include "wasmer.h"`. That way the deprecated doesn't need to redeclare any duplicate stuff, and people can easily build to the new API without using old parts by accident, and the new API is available to anyone using the old API to facilitate updating to the new API.

lib/c-api/wasmer_deprecated.h
```diff
@@ -0,0 +1,1480 @@
...
+#if !defined(__has_declspec_attribute)
+#  define __has_declspec_attribute(x) 0
+#endif
+
+// Define the `DEPRECATED` macro.
+#if defined(GCC) || defined(__GNUC__) || __has_attribute(deprecated)
+#  define DEPRECATED(message) __attribute__((deprecated(message)))
+#elif defined(MSVC) || __has_declspec_attribute(deprecated)
+#  define DEPRECATED(message) __declspec(deprecated(message))
+#endif
```
nlewycky, line 33: Yes, perfect! It seems that we should be marking the entire deprecated API with `DEPRECATED("use X instead")` now?

lib/c-api/build.rs
syrusakbary: We want to do it this way because I don't want to have breaking changes for people that are using the `#include "wasmer.h"` header for using the API.

That's the main reason for this. We will probably remove this on 2.0 :)

MarkMcCaskey (approved):
Seems like a good change overall!  My biggest concern is implicitly encouraging the deprecated API by making it always imported / weirdness from accidentally mixing the APIs.

lib/c-api/build.rs
syrusakbary: Yup it is!

syrusakbary:
```suggestion
// Include the Wasm API for using the wasm_valkind_enum type.
```

lib/c-api/wasmer_deprecated.h
MarkMcCaskey: It'd be nice to have this macro available for our custom extensions to the Wasmer C API, we have 2 functions related to initializing WASI which are now deprecated and not marked like this, only marked by a comment

MarkMcCaskey:
bors try

Hywan:
> My biggest concern is implicitly encouraging the deprecated API by making it always imported / weirdness from accidentally mixing the APIs.

That's my biggest concern too. With an IDE, a user could just run autocompletion and use functions or types from the deprecated API by mistake :-/. That's really not ideal. And for that, I am likely to stand against this PR if we don't find a correct solution to avoid such scenario.

syrusakbary:
Closing this PR as it was already merged in master with #2375

Issue #1872, opened 2020-12-04
Naming conflict Tunables between wasmer/wasmer_engine


webmaster128:
### Describe the bug

According to Syrus' statement (https://github.com/CosmWasm/cosmwasm/commit/950f63183a266d9114f80085b66d679b7a6c33ca#r44799222), all symbols from the crate `wasmer-engine` should be available in `wasmer`. When I tried that, I ran into the naming conflict `Tunables`, which is the trait in `wasmer-engine` and the reference implementation in `wasmer`.

In my project, I resolved the conflict as follows:

```rust
use wasmer::{Bytes, Engine, Pages, Singlepass, Store, Target, Tunables as ReferenceTunables, JIT};
use wasmer_engine::Tunables;
```

because I don't think `Tunables` is not a great name for one specific `Tunables` implementation. Maybe we adopt this here? Or resolve the conflict differently.

### Steps to reproduce

NA

### Expected behavior

All symbols from `wasmer_engine` are re-exported in `wasmer`

### Actual behavior

`wasmer_engine::Tunables` and `wasmer::Tunables` are different things.

webmaster128:
I can work on this, but need a 👍/👎 if you would be happy with `Tunables` for the trait and `ReferenceTunables` (or `BaseTunables`) for the implementation.

Hywan:
I'm OK to try something. Ping me on your PR so that I can review it :-).

PR #1894, opened 2020-12-07
Expose wasmer::{CraneliftOptLevel, LLVMOptLevel}

webmaster128:
# Description

This is required to do something like

```rust
use wasmer::{Cranelift, CraneliftOptLevel};

let mut compiler = Cranelift::default();
compiler.opt_level(CraneliftOptLevel::None);
let engine = JIT::new(compiler).engine();
// ...
```

# Review

### DIFF
--- a/lib/compiler-cranelift/src/config.rs
+++ b/lib/compiler-cranelift/src/config.rs
@@ -11,7 +11,7 @@ use wasmer_compiler::{
 /// Possible optimization levels for the Cranelift codegen backend.
 #[non_exhaustive]
 #[derive(Clone, Debug)]
-pub enum OptLevel {
+pub enum CraneliftOptLevel {
     /// No optimizations performed, minimizes compilation time by disabling most
     /// optimizations.
     None,
@@ -33,7 +33,7 @@ pub struct Cranelift {
     enable_verifier: bool,
     enable_simd: bool,
     enable_pic: bool,
-    opt_level: OptLevel,
+    opt_level: CraneliftOptLevel,
     /// The middleware chain.
     pub(crate) middlewares: Vec<Arc<dyn ModuleMiddleware>>,
 }
@@ -45,7 +45,7 @@ impl Cranelift {
         Self {
             enable_nan_canonicalization: false,
             enable_verifier: false,
-            opt_level: OptLevel::Speed,
+            opt_level: CraneliftOptLevel::Speed,
             enable_pic: false,
             enable_simd: true,
             middlewares: vec![],
@@ -68,7 +68,7 @@ impl Cranelift {
     }
 
     /// The optimization levels when optimizing the IR.
-    pub fn opt_level(&mut self, opt_level: OptLevel) -> &mut Self {
+    pub fn opt_level(&mut self, opt_level: CraneliftOptLevel) -> &mut Self {
         self.opt_level = opt_level;
         self
     }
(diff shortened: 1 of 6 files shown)


lib/compiler-cranelift/src/config.rs
```diff
@@ -11,7 +11,7 @@ use wasmer_compiler::{
 /// Possible optimization levels for the Cranelift codegen backend.
 #[non_exhaustive]
 #[derive(Clone, Debug)]
-pub enum OptLevel {
+pub enum CraneliftOptLevel {
```
Hywan, line 14: Why do we need to rename this type?

Hywan:
Thanks for the PR!

lib/compiler-cranelift/src/config.rs
webmaster128: Not strictly necessary, but with this rename we avoid naming collisions in the `wasmer::` high level API namespace. In the not too far future there is probably `wasmer::OptimizationLevel` for the LLVM version, making it hard to reason about to which backends the types belong.

The other option would be use have `pub use wasmer_compiler_cranelift::{Cranelift, OptLevel as CraneliftOptLevel};`, which gives us `wasmer_compiler_cranelift::OptLevel` and `wasmer::CraneliftOptLevel`. But in my opinion this creates more confusion than it helps.

What do you think?

syrusakbary: Not sure about this change.
You think is likely companies will use two compilers within the same codebase? This PR moves into prefixing all the public-facing objects of compilers with `CompilernameXXX`

webmaster128: > You think is likely companies will use two compilers within the same codebase?

Well, we do. I don't know much about other use cases.
But even if they don't. The caller code `use wasmer::OptLevel;` should not refer to compiler X in one project and compiler Y in a different project, right?

> This PR moves into prefixing all the public-facing objects of compilers with `CompilernameXXX`

Right. It is based on the idea that you promoted recently: make all high level API available in the `wasmer::` namespace. This is a noble goal, but hard to get right. In https://github.com/wasmerio/wasmer/pull/1730 we were fighting imports, https://github.com/wasmerio/wasmer/issues/1872 documents a conflict now we have at least very similar names for multiple compilers.

webmaster128: > @webmaster128 Could you extend this renaming to other compilers in this PR, so that it's consistent across the entire codebase please?

So far this is the only new compiler-specific export in `wasmer::`:

```rust
#[cfg(feature = "singlepass")]
pub use wasmer_compiler_singlepass::Singlepass;

#[cfg(feature = "cranelift")]
pub use wasmer_compiler_cranelift::{Cranelift, CraneliftOptLevel};

#[cfg(feature = "llvm")]
pub use wasmer_compiler_llvm::LLVM;
```

However, I now also added `LLVMOptLevel` to show the direction in which this is heading.

Hywan (approved):
Thanks!

webmaster128:
Fixed the compile issue

Hywan:
bors try

webmaster128:
I think this is ready to be merged. Or should I update it again to master?

Hywan:
Please can you just update the `CHANGELOG.md`, since we have released the 1.0.0-beta2 it's outdated. And I'll merge it! Thanks.

webmaster128:
> Please can you just update the `CHANGELOG.md`, since we have released the 1.0.0-beta2 it's outdated.

Sure, done

Hywan:
Thanks!

bors r+

PR #1901, opened 2020-12-08
Fix use-after free with `NativeFunc`

MarkMcCaskey:
Prior to this change, `NativeFunc` didn't keep the `Instance` alive.

This PR does the simple thing of storing `ExportFunction` in `NativeFunc`: this stores data isn't strictly necessary for `NativeFunc` and can be optimized to only include a subset of the fields of `ExportFunction`, however we would need better test coverage to ensure the transformation into and out of `NativeFunc` still works as expected.  I think it's a reasonable trade off to just store `ExportFunction` directly as it solves a lot of problems and is a very simple solution.

Note: The new indirect field accesses may slow down `call` on `NativeFunc`, this is overhead that may be worth avoiding by duplicating these fields in `NativeFunc`. I'll investigate that and write about it here.

I've benchmarked it and it seems that there is no overhead from the indirect field accesses.

### DIFF
--- a/lib/vm/src/instance.rs
+++ b/lib/vm/src/instance.rs
@@ -933,6 +933,7 @@ impl Drop for InstanceAllocator {
         // > "acquire" operation before deleting the object.
         //
         // [1]: https://www.boost.org/doc/libs/1_55_0/doc/html/atomic/usage_examples.html
+        dbg!("Freeing instance allocator");
         atomic::fence(atomic::Ordering::Acquire);
 
         // Now we can deallocate the instance. Note that we don't
--- a/lib/vm/src/instance.rs
+++ b/lib/vm/src/instance.rs
@@ -933,7 +933,6 @@ impl Drop for InstanceAllocator {
         // > "acquire" operation before deleting the object.
         //
         // [1]: https://www.boost.org/doc/libs/1_55_0/doc/html/atomic/usage_examples.html
-        dbg!("Freeing instance allocator");
         atomic::fence(atomic::Ordering::Acquire);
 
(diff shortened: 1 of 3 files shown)


lib/vm/src/instance.rs
```diff
@@ -933,6 +933,7 @@ impl Drop for InstanceAllocator {
         // > "acquire" operation before deleting the object.
         //
         // [1]: https://www.boost.org/doc/libs/1_55_0/doc/html/atomic/usage_examples.html
+        dbg!("Freeing instance allocator");
```
syrusakbary, line 936: Remove me

syrusakbary (approved):
Other than the small nit (`dbg!`), this is ready to be shipped!

syrusakbary:
Also, do you think it's reasonable to have a test case for it?

syrusakbary:
(closed by mistake)

MarkMcCaskey:
> Also, do you think it's reasonable to have a test case for it?

We can certainly have a test case for it but it's not a test that will pass on this branch and fail on master.  Well, it _may_ fail on master but when I tested it previously, it appeared to work, it was only via that `dbg` statement that you could tell the function was being called after the `Instance` had been freed.

Maybe there's a way we can use custom memory allocators to test this kind of thing... 2 problems though:
- I believe we `mmap` function bodies
- custom allocators are nightly only...

Maybe we can have a feature that we can use during tests where when we unmmap something we modify it in such a way that any further accesses fail

syrusakbary:
Ok, let's not worry about testing then on this PR

Hywan:
:+1:

PR #1902, opened 2020-12-09
Fix memory leak in `InstanceAllocator` on failure

MarkMcCaskey:
This code needs to be cleaned up, just hacked it out really quickly.

The idea is to use a builder pattern to create a type-safe state machine for setting up things for the `InstanceAllocator`; this should allow us to remove `unsafe` in our "public" API because we can guarantee certain invariants because of the restricted nature of these types.

The names for the new types are long and don't really make sense but I didn't want to get blocked on that: suggestions welcome!

In combination with #1865 , this PR should fix the all the memory leaks seen by asan while running `make test-cranelift-jit` 🎉 🎉 🎉

### DIFF
--- a/lib/vm/src/export.rs
+++ b/lib/vm/src/export.rs
@@ -2,7 +2,7 @@
 // Attributions: https://github.com/wasmerio/wasmer/blob/master/ATTRIBUTIONS.md
 
 use crate::global::Global;
-use crate::instance::InstanceAllocator;
+use crate::instance::InstanceRef;
 use crate::memory::{Memory, MemoryStyle};
 use crate::table::{Table, TableStyle};
 use crate::vmcontext::{VMFunctionBody, VMFunctionEnvironment, VMFunctionKind, VMTrampoline};
@@ -49,8 +49,8 @@ pub struct VMExportFunction {
     pub call_trampoline: Option<VMTrampoline>,
 
     /// A “reference” to the instance through the
-    /// `InstanceAllocator`. `None` if it is a host function.
-    pub instance_allocator: Option<InstanceAllocator>,
+    /// `InstanceRef`. `None` if it is a host function.
+    pub instance_allocator: Option<InstanceRef>,
 }
 
 /// # Safety
@@ -74,8 +74,8 @@ pub struct VMExportTable {
     pub from: Arc<dyn Table>,
 
     /// A “reference” to the instance through the
-    /// `InstanceAllocator`. `None` if it is a host function.
-    pub instance_allocator: Option<InstanceAllocator>,
+    /// `InstanceRef`. `None` if it is a host function.
+    pub instance_allocator: Option<InstanceRef>,
 }
 
 /// # Safety
@@ -120,8 +120,8 @@ pub struct VMExportMemory {
     pub from: Arc<dyn Memory>,
 
     /// A “reference” to the instance through the
-    /// `InstanceAllocator`. `None` if it is a host function.
-    pub instance_allocator: Option<InstanceAllocator>,
+    /// `InstanceRef`. `None` if it is a host function.
+    pub instance_allocator: Option<InstanceRef>,
 }
 
 /// # Safety
@@ -166,8 +166,8 @@ pub struct VMExportGlobal {
     pub from: Arc<Global>,
 
     /// A “reference” to the instance through the
-    /// `InstanceAllocator`. `None` if it is a host function.
-    pub instance_allocator: Option<InstanceAllocator>,
+    /// `InstanceRef`. `None` if it is a host function.
+    pub instance_allocator: Option<InstanceRef>,
 }
 
 /// # Safety
--- a/lib/vm/src/instance.rs
+++ b/lib/vm/src/instance.rs
@@ -688,6 +688,248 @@ impl Instance {
     }
 }
 
+/// TODO: rename this
+pub struct UnpreparedInstanceAllocatorSetter {
+    instance_ptr: NonNull<Instance>,
+    instance_layout: Layout,
+    /// Only in an `Option` so we can `Option::take` it before it's
+    /// dropped.
+    offsets: Option<VMOffsets>,
+    /// Whether to drop the inner values or not.
+    consumed: bool,
+}
+
+impl Drop for UnpreparedInstanceAllocatorSetter {
+    fn drop(&mut self) {
+        if !self.consumed {
+            let instance_ptr = self.instance_ptr.as_ptr();
+            unsafe {
+                std::alloc::dealloc(instance_ptr as *mut u8, self.instance_layout);
+            }
+        }
+    }
+}
+
+impl UnpreparedInstanceAllocatorSetter {
+    /// Allocate `Instance` (it is an uninitialized pointer).
+    ///
+    /// `offsets` is used to compute the layout with `Self::instance_layout`.
+    pub fn new(offsets: VMOffsets) -> Self {
+        let instance_layout = Self::instance_layout(&offsets);
+
+        #[allow(clippy::cast_ptr_alignment)]
+        let instance_ptr = unsafe { alloc::alloc(instance_layout) as *mut Instance };
+
+        let instance_ptr = if let Some(ptr) = NonNull::new(instance_ptr) {
+            ptr
+        } else {
+            alloc::handle_alloc_error(instance_layout);
+        };
+
+        Self {
+            instance_ptr: instance_ptr.cast(),
+            instance_layout,
+            offsets: Some(offsets),
+            consumed: false,
+        }
+    }
+
+    /// Calculate the appropriate layout for `Instance`.
+    fn instance_layout(offsets: &VMOffsets) -> Layout {
+        let vmctx_size = usize::try_from(offsets.size_of_vmctx())
+            .expect("Failed to convert the size of `vmctx` to a `usize`");
+
+        let instance_vmctx_layout =
+            Layout::array::<u8>(vmctx_size).expect("Failed to create a layout for `VMContext`");
+
+        let (instance_layout, _offset) = Layout::new::<Instance>()
+            .extend(instance_vmctx_layout)
+            .expect("Failed to extend to `Instance` layout to include `VMContext`");
+
+        instance_layout.pad_to_align()
+    }
+
+    /// Prepare the instance allocator setter by getting pointers into key
+    /// locations which can be used to initialize data.
+    pub fn prepare(
+        mut self,
+    ) -> (
+        HalfPreparedInstanceAllocatorSetter,
+        Vec<NonNull<VMMemoryDefinition>>,
+        Vec<NonNull<VMTableDefinition>>,
+    ) {
+        let memories = unsafe { self.memory_definition_locations() };
+        let tables = unsafe { self.table_definition_locations() };
+        self.consumed = true;
+        let instance_ptr = self.instance_ptr;
+        let instance_layout = self.instance_layout;
+        let offsets = self.offsets.take();
+        let prepared = HalfPreparedInstanceAllocatorSetter {
+            instance_ptr,
+            instance_layout,
+            offsets,
+            consumed: false,
+        };
+
+        (prepared, memories, tables)
+    }
+
+    /// Get the locations of where the local `VMMemoryDefinition`s should be stored.
+    ///
+    /// This function lets us create `Memory` objects on the host with backing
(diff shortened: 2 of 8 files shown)


lib/vm/src/instance.rs
```diff
@@ -688,6 +688,218 @@ impl Instance {
     }
 }
 
+/// TODO: rename this
+/// You must not put any type in this struct with drop logic as
+/// `Drop` will not be called on this type if it transitions into the
+/// next state.
+pub struct UnpreparedInstanceAllocatorSetter {
```
syrusakbary, line 695: What you think about `UnpreparedInstance` / `InstanceLayout` ?

MarkMcCaskey: Well, it's certainly an improvement!

`UnpreparedInstance` isn't too bad, I'll review it with fresh eyes tomorrow

lib/vm/src/instance.rs
```diff
@@ -963,11 +1094,15 @@ impl InstanceHandle {
...
+        UnpreparedInstanceAllocatorSetter,
+        Vec<NonNull<VMMemoryDefinition>>,
+        Vec<NonNull<VMTableDefinition>>,
+    ) {
         let offsets = VMOffsets::new(mem::size_of::<*const u8>() as u8, module);
-        let (instance_ptr, _instance_layout) = InstanceAllocator::allocate_instance(&offsets);
-
-        (instance_ptr.cast(), offsets)
+        UnpreparedInstanceAllocatorSetter::new(offsets)
     }
```
syrusakbary, line 1106: Can we move this function to the `UnpreparedInstance` struct?

MarkMcCaskey: I thought it'd be more discoverable there but I think it's fine to put it on the type itself too

syrusakbary (approved):

lib/vm/src/instance.rs
```diff
@@ -688,6 +688,189 @@ impl Instance {
...
+}
+
+impl Drop for UnpreparedInstance {
+    fn drop(&mut self) {
+        if !self.consumed {
+            // If `consumed` has not been set, then we still have ownership
+            // over the buffer and must free it.
+            let instance_ptr = self.instance_ptr.as_ptr();
+            unsafe {
+                std::alloc::dealloc(instance_ptr as *mut u8, self.instance_layout);
```
Hywan, line 722: I believe we should also use `ptr::drop_in_place(instance_ptr);` in order to properly drop `Instance`. For example, if `Instance` has a `Drop` implementation, `drop_in_place` will call it, whilst `alloc::dealloc` won't.

Hywan: The actual code in this PR might also be unsafe. Nothing forbids `self.instance_ptr` to be dangling, and I don't know what happens when one tries to dealloc a dangling pointer 🤔.

lib/vm/src/instance.rs
```diff
@@ -688,6 +688,189 @@ impl Instance {
...
+    /// tables respectively. These pointers should be written to before
+    /// calling [`InstanceHandle::new`].
+    pub fn new(
+        module: &ModuleInfo,
+    ) -> (
+        UnpreparedInstance,
+        Vec<NonNull<VMMemoryDefinition>>,
+        Vec<NonNull<VMTableDefinition>>,
+    ) {
+        let offsets = VMOffsets::new(mem::size_of::<*const u8>() as u8, module);
```
Hywan, line 743: Side note: I've noticed this `mem::size_of::<*const u8>() as u8` and I wanted to simplify it to `mem::size_of::<usize>() as u8`. We compute the size of a pointer here, and `usize` is exactly that.

That's not a mandatory feedback though 😃.

lib/vm/src/instance.rs
```diff
@@ -688,6 +688,189 @@ impl Instance {
...
+
+        let unprepared = Self {
+            instance_ptr,
+            instance_layout,
+            offsets,
+            consumed: false,
+        };
+
+        let memories = unsafe { unprepared.memory_definition_locations() };
+        let tables = unsafe { unprepared.table_definition_locations() };
```
Hywan, line 763: Shouldn't we specify why it is OK to have `unsafe` blocks here?

lib/vm/src/instance.rs
```diff
@@ -688,6 +688,189 @@ impl Instance {
...
+/// metadata when creating an [`Instance`].
+///
+/// This type will free the allocated memory if it's dropped before
+/// being used.
+///
+/// The [`UnpreparedInstance::instance_layout`] computes the correct
+/// layout to represent the wanted [`Instance`].
+///
+/// Then we use this layout to allocate an empty `Instance` properly.
+pub struct UnpreparedInstance {
```
Hywan, line 701: I would suggest `InstanceBuilder` rather than `UnpreparedInstance`. Thoughts?

lib/vm/src/instance.rs
```diff
@@ -688,6 +688,189 @@ impl Instance {
...
+
+            let new_ptr = NonNull::new_unchecked(base_ptr.add(table_offset));
+
+            out.push(new_ptr.cast());
+        }
+        out
+    }
+
+    /// Finish preparing by writing the [`Instance`] into memory.
+    pub(crate) fn write_instance(mut self, instance: Instance) -> InstanceAllocator {
```
Hywan, line 849: I'm not sure it's needed to get a `pub(crate)` here. A private visbility should be enough.

lib/vm/src/instance.rs
```diff
@@ -688,6 +688,189 @@ impl Instance {
...
+        // transition into the new state.
+        self.consumed = true;
+
+        // This is correct because of the invariants of `Self` and
+        // because we write `Instance` to the pointer in this function.
+        unsafe { InstanceAllocator::new(instance, instance_layout) }
+    }
+
+    /// Get the [`VMOffsets`] for the allocated buffer.
+    pub(crate) fn offsets(&self) -> VMOffsets {
```
Hywan, line 869: Why `pub(crate)`?

lib/vm/src/instance.rs
```diff
@@ -688,6 +688,189 @@ impl Instance {
...
+        self.offsets.clone()
```
Hywan, line 870: Why cloning? Do we want a “read-only” API? In any case, I think returning `&self.offsets` should be OK. Thoughts?

lib/vm/src/instance.rs
```diff
@@ -688,6 +688,189 @@ impl Instance {
...
+    }
+}
+
 /// An `InstanceAllocator` is responsible to allocate, to deallocate,
 /// and to give access to an `Instance`, in such a way that `Instance`
 /// is unique, can be shared, safely, across threads, without
```
Hywan, line 876: This is outdated now. `InstanceBuilder` is responsible to allocate it, not `InstanceAllocator` (which sounds weird…).

lib/vm/src/instance.rs
```diff
@@ -769,48 +940,22 @@ impl InstanceAllocator {
...
     /// `instance` must a non-null, non-dangling, properly aligned,
     /// and correctly initialized pointer to `Instance`. See
-    /// `InstanceHandle::new` for an example of how to correctly use
+    /// [`UnpreparedInstance::new`] for an example of how to correctly use
     /// this API.
-    unsafe fn new(instance: NonNull<Instance>, instance_layout: Layout) -> Self {
-        Self {
+    /// TODO: update docs
+    pub(crate) unsafe fn new(instance: NonNull<Instance>, instance_layout: Layout) -> Self {
+        InstanceAllocator {
```
Hywan, line 947:
```suggestion
        Self {
```

lib/vm/src/instance.rs
```diff
@@ -769,48 +940,22 @@ impl InstanceAllocator {
     ///
     /// `instance` must a non-null, non-dangling, properly aligned,
     /// and correctly initialized pointer to `Instance`. See
-    /// `InstanceHandle::new` for an example of how to correctly use
+    /// [`UnpreparedInstance::new`] for an example of how to correctly use
     /// this API.
-    unsafe fn new(instance: NonNull<Instance>, instance_layout: Layout) -> Self {
-        Self {
+    /// TODO: update docs
+    pub(crate) unsafe fn new(instance: NonNull<Instance>, instance_layout: Layout) -> Self {
```
Hywan, line 946: Why `pub(crate)`?

Hywan (commented):
Thank you for tackling this. It was obvious for me that this code needed another rewrite; the first rewrite (introduced in #1837) was focusing on providing more safety and a new feature, but the fact the creation of `InstanceAllocator` were happening with multiple methods from different types was weird and error-prone. So, good job!

After reading this PR carefully, I wonder whether we could rename `UnpreparedInstance` to `InstanceAllocator` or `InstanceBuilder`, and rename `InstanceAllocator` to `InstanceRef`.

I'm also surprised to see 2 destructors for `Instance`: one in `UnpreparedInstance`, and one in `InstanceAllocator`. Should we merge them both? Or maybe `UnpreparedInstance` should call the one from `InstanceAllocator` based on `consumed`'s state?

Thank you!

lib/vm/src/instance.rs
MarkMcCaskey: > I believe we should also use ptr::drop_in_place(instance_ptr); in order to properly drop Instance. For example, if Instance has a Drop implementation, drop_in_place will call it, whilst alloc::dealloc won't.


The intent of this type was that it would never exist with the `Instance` written to it, it was purely managing the buffer prior to that step.  Due to feedback from Syrus, an intermediate type which made the distinction of having the `Instance` written to it was removed and now there are a few lines where, if dropped, this type would do the wrong thing (however `consumed` is also set there, so it wouldn't free anything anyways).

I think it probably does not make sense to add drop logic for `Instance` here unless we end up merging types more somehow.  I still like having an additional type / passing it to into the constructor of `InstanceHandle`, it would allow us to reduce unsafety and simplify the code but I got strong feedback against it.

--------

> The actual code in this PR might also be unsafe. Nothing forbids self.instance_ptr to be dangling, and I don't know what happens when one tries to dealloc a dangling pointer 🤔.

That's a good thought but as far as I can tell it is not possible.  The only way it could be possible is because this type is defined in a huge file where people could access its field directly.  I think the solution is to move it into another file where its private fields can remain private.  Because we railroad the user with (now an extremely simple) type-state machine, the user can't use the API incorrectly as there's no way to give a pointer that we didn't create ourselves.

lib/vm/src/instance.rs
MarkMcCaskey: I wanted to make a distinction between things that things that are not this type can call and what is supposed to be private.  It doesn't really make sense though, I should just move this type into its own module and make it more explicit

syrusakbary:
Following from what we talked in Slack, 

I think what is now `InstanceAllocator` should be `InstanceRef`  and what is `InstanceBuilder` should be `InstanceAllocator`.
Once that's done, we should be good to merge

lib/vm/src/instance.rs
Hywan: All good to me :-). Thanks for your reply.

lib/vm/src/instance/builder.rs
```diff
@@ -0,0 +1,197 @@
...
+        #[allow(clippy::cast_ptr_alignment)]
+        let instance_ptr = unsafe { alloc::alloc(instance_layout) as *mut Instance };
+
+        let instance_ptr = if let Some(ptr) = NonNull::new(instance_ptr) {
+            ptr
+        } else {
+            alloc::handle_alloc_error(instance_layout);
+        };
+
+        let unprepared = Self {
```
Hywan, line 75: `let builder`? :-)

syrusakbary: We are going with allocator. So perhaps `let allocator`?

syrusakbary (approved):

lib/vm/src/export.rs
```diff
@@ -74,8 +74,8 @@ pub struct VMExportTable {
     pub from: Arc<dyn Table>,
 
     /// A “reference” to the instance through the
-    /// `InstanceAllocator`. `None` if it is a host function.
-    pub instance_allocator: Option<InstanceAllocator>,
+    /// `InstanceRef`. `None` if it is a host function.
+    pub instance_allocator: Option<InstanceRef>,
```
Hywan, line 78: Ideally, we should rename this to `instance`. I'll make a PR to update them.

PR #1908, opened 2020-12-09
Implement TryFrom<Value<T>> for i32/u32/i64/u64/f32/f64

webmaster128:
# Description

After the migration from 0.17 to 1.0 users have to handle much more dynamic return values. This PR makes working with `Value<T>` more convenient by implementing `TryFrom<Value<T>>` for `i32`/`u32`/`i64`/`u64`/`f32`/`f64`.

This PR is tested and demonstrated at https://github.com/CosmWasm/cosmwasm/pull/658

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
  
 * [#1867](https://github.com/wasmerio/wasmer/pull/1867) Added `Metering::get_remaining_points` and `Metering::set_remaining_points` 
 * [#1881](https://github.com/wasmerio/wasmer/pull/1881) Added `UnsupportedTarget` error to `CompileError`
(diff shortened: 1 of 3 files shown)


MarkMcCaskey (approved):
Looks good to me!

One thing to note because I see some formatting changes, Wasmer is 1 behind the latest stable so you may need to reformat with Rust 1.47.0

syrusakbary:
It seems it failed for external reasons to this PR.
Trying again

bors r+

webmaster128:
> I see some formatting changes

The imports in `lib/wasmer-types/src/lib.rs`? This was caused by adding `convert` to the list. Then some of then ended up in the next line.

PR #1909, opened 2020-12-09
Allow constructing V128 from byte array

webmaster128:
# Description

V128 stores 16 bytes fixed length. Constructing it from `&[u8]` turns compile-time checks into a runtime assertion, which is not ideal I guess.

With this new constructor we save a vector allocation and imcrease readability.

# Review


MarkMcCaskey (approved):
Nice, thanks!

Hywan:
Neat, thanks!

PR #1911, opened 2020-12-10
Add support for const function signatures

webmaster128:
# Description

`FunctionType` contains variable length vectors and I'm sure there is a good reason for this design. However, it makes creating them a bit annoying since it connot be done in constants. With this PR I propose a representation that uses a tuple of arrays, which can be converted into `FunctionType` on demand. This allows developers to use constant signatures and push them to a better place in their code.

In https://github.com/CosmWasm/cosmwasm/pull/659/files you see a demonstration of how this deduplicates code and pushes the definitions out of my way.

Since the conversion `&FunctionType` to `FunctionType` is implemented here, this should no be breaking anyone's code. In cases where you can pass an owned `FunctionType` instead of `&FunctionType` you save one clone.

# Review

### DIFF
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -69,7 +69,7 @@ pub struct Function {
 impl Function {
     /// Creates a new host `Function` (dynamic) with the provided signature.
     ///
-    /// # Example
+    /// # Examples
     ///
     /// ```
     /// # use wasmer::{Function, FunctionType, Type, Store, Value};
@@ -82,11 +82,27 @@ impl Function {
     ///     Ok(vec![Value::I32(sum)])
     /// });
     /// ```
+    ///
+    /// With constant signature:
+    ///
+    /// ```
+    /// # use wasmer::{Function, FunctionType, Type, Store, Value};
+    /// # let store = Store::default();
+    /// #
+    /// const I32_I32_TO_I32: ([Type; 2], [Type; 1]) = ([Type::I32, Type::I32], [Type::I32]);
+    ///
+    /// let f = Function::new(&store, I32_I32_TO_I32, |args| {
+    ///     let sum = args[0].unwrap_i32() + args[1].unwrap_i32();
+    ///     Ok(vec![Value::I32(sum)])
+    /// });
+    /// ```
     #[allow(clippy::cast_ptr_alignment)]
-    pub fn new<F>(store: &Store, ty: &FunctionType, func: F) -> Self
+    pub fn new<FT, F>(store: &Store, ty: FT, func: F) -> Self
     where
+        FT: Into<FunctionType>,
         F: Fn(&[Val]) -> Result<Vec<Val>, RuntimeError> + 'static,
     {
+        let ty: FunctionType = ty.into();
         let dynamic_ctx = VMDynamicFunctionContext::from_context(VMDynamicFunctionWithoutEnv {
             func: Box::new(func),
             function_type: ty.clone(),
@@ -108,7 +124,7 @@ impl Function {
                     address,
                     kind: VMFunctionKind::Dynamic,
                     vmctx,
-                    signature: ty.clone(),
+                    signature: ty,
                     call_trampoline: None,
                     instance_allocator: None,
                 },
@@ -118,7 +134,7 @@ impl Function {
 
     /// Creates a new host `Function` (dynamic) with the provided signature and environment.
     ///
-    /// # Example
+    /// # Examples
     ///
     /// ```
     /// # use wasmer::{Function, FunctionType, Type, Store, Value, WasmerEnv};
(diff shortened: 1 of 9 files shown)


lib/api/src/externals/function.rs
```diff
@@ -82,11 +82,27 @@ impl Function {
     ///     Ok(vec![Value::I32(sum)])
     /// });
     /// ```
+    ///
+    /// With constant signature
```
Hywan, line 86:
```suggestion
    /// With constant signature:
```

lib/api/src/externals/function.rs
```diff
@@ -137,12 +153,33 @@ impl Function {
     ///     Ok(vec![Value::I32(result)])
     /// });
     /// ```
+    ///
+    /// With constant signature
```
Hywan, line 157:
```suggestion
    /// With constant signature:
```

lib/wasmer-types/src/types.rs
```diff
@@ -276,6 +276,39 @@ impl fmt::Display for FunctionType {
...
+    7,0 7,1 7,2 7,3 7,4 7,5 7,6 7,7 7,8 7,9
+    8,0 8,1 8,2 8,3 8,4 8,5 8,6 8,7 8,8 8,9
+    9,0 9,1 9,2 9,3 9,4 9,5 9,6 9,7 9,8 9,9
+}
+
+impl From<&FunctionType> for FunctionType {
+    fn from(as_ref: &FunctionType) -> Self {
+        as_ref.clone()
+    }
+}
```
Hywan, line 310: :+1:

Hywan (approved):
I like the idea. I don't see exactly how it improves Wasmer itself, but I can understand how it can be convenient for the user!

I'm approving this, but I would appreciate another feedback on this from @syrusakbary, @nlewycky, @MarkMcCaskey or @jubianchi before merging :-).

syrusakbary (approved):
Looks good to me!

webmaster128:
yo bors, I rebased this PR to lastest master and fixed the compile error

webmaster128:
Oh, the failing test now shows we are using some `&&wasmer::FunctionType`. I guess this is not intended and should be `&wasmer::FunctionType`, right?

```
error[E0277]: the trait bound `wasmer::FunctionType: std::convert::From<&&wasmer::FunctionType>` is not satisfied
   --> lib/c-api/src/wasm_c_api/externals/function.rs:81:20
    |
81  |     let function = Function::new(&store.inner, &func_sig, inner_callback);
    |                    ^^^^^^^^^^^^^ the trait `std::convert::From<&&wasmer::FunctionType>` is not implemented for `wasmer::FunctionType`
    | 
   ::: /home/runner/work/wasmer/wasmer/lib/api/src/externals/function.rs:102:13
    |
102 |         FT: Into<FunctionType>,
    |             ------------------ required by this bound in `wasmer::Function::new`
    |
    = help: the following implementations were found:
              <wasmer::FunctionType as std::convert::From<&wasmer::FunctionType>>
              <wasmer::FunctionType as std::convert::From<([wasmer::ValType; 0], [wasmer::ValType; 0])>>
              <wasmer::FunctionType as std::convert::From<([wasmer::ValType; 0], [wasmer::ValType; 1])>>
              <wasmer::FunctionType as std::convert::From<([wasmer::ValType; 0], [wasmer::ValType; 2])>>
            and 97 others
    = note: required because of the requirements on the impl of `std::convert::Into<wasmer::FunctionType>` for `&&wasmer::FunctionType`

error[E0277]: the trait bound `wasmer::FunctionType: std::convert::From<&&wasmer::FunctionType>` is not satisfied
   --> lib/c-api/src/wasm_c_api/externals/function.rs:155:20
    |
155 |     let function = Function::new_with_env(
    |                    ^^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::From<&&wasmer::FunctionType>` is not implemented for `wasmer::FunctionType`
    | 
   ::: /home/runner/work/wasmer/wasmer/lib/api/src/externals/function.rs:178:13
    |
178 |         FT: Into<FunctionType>,
    |             ------------------ required by this bound in `wasmer::Function::new_with_env`
    |
    = help: the following implementations were found:
              <wasmer::FunctionType as std::convert::From<&wasmer::FunctionType>>
              <wasmer::FunctionType as std::convert::From<([wasmer::ValType; 0], [wasmer::ValType; 0])>>
              <wasmer::FunctionType as std::convert::From<([wasmer::ValType; 0], [wasmer::ValType; 1])>>
              <wasmer::FunctionType as std::convert::From<([wasmer::ValType; 0], [wasmer::ValType; 2])>>
            and 97 others
    = note: required because of the requirements on the impl of `std::convert::Into<wasmer::FunctionType>` for `&&wasmer::FunctionType`

error: aborting due to 2 previous errors
```

webmaster128:
I'll take care of this now

webmaster128:
Rebased to latest master and 1 commit added (8a8e8dc981734430c1f8aedf509680878ac94578)

Hywan:
bors try

MarkMcCaskey (approved):

webmaster128:
Could you restart this job? https://github.com/wasmerio/wasmer/runs/1551129968 Seems like it never started

MarkMcCaskey:
Random follow up but I wanted to ping you, @webmaster128 , to make sure you know about [`NativeFunc`](https://docs.rs/wasmer/1.0.0-beta1/wasmer/struct.NativeFunc.html): if you know the signature of your host function at compile time, you can avoid specifying the type signature of your host function entirely by using `Function::new_native` and `Function::new_native_with_env` and it should have less runtime overhead.

We just had another issue where someone missed that, so I wanted to make sure you knew about it too in case the docs aren't clear here.

webmaster128:
Thank you @MarkMcCaskey, much appreciated. This was not clear to me indeed for a long time, especially when upgrading from [statically typed `Func` closures in 0.17](https://github.com/CosmWasm/cosmwasm/blob/pre-reborn/packages/vm/src/instance.rs#L98) to the new API. Now I already migrated all functions to native that do not capture variables. One function could be refactored to storing a variable in Env instead of capturing. And some are remaining to be refactored from capturing closures to native functions accessing data via Env.

webmaster128:
I'm very open to discuss if this PR is more destructive than helpful. Does it encourage users to follow the wrong path? Could all those testing dummies be replaced with statically typed closures? https://github.com/CosmWasm/cosmwasm/pull/663/commits/65a7d1f7ab7fe5f56b2eb10e6e5789094314300d

MarkMcCaskey:
@webmaster128 Yeah, it seems like those examples could be made into "native" functions if you wanted to!  Don't worry about it, I think this PR is a good change, it's nice to have the ability to avoid heap allocations.  I've updated the docs locally to make it more clear what the options are!

Just wanted to make sure you knew about it in case you didn't!

webmaster128:
Alright cool. One thought why this might be tricky for people who are migrating: it seems like the default chained from native to dynamic. To me `Function::new` feels like the direct replacement of `Func::new`. But it looks like you got a native function in 0.17 and get a dynamic function in 1.0.

PR #1912, opened 2020-12-10
fix: Fix test_set_host_global_immutable test

jubianchi:
wasm_globaltype_new takes ownership of its first argument, valtype. The test was freeing the valtype when it should not

### DIFF
--- a/lib/c-api/src/wasm_c_api/externals/global.rs
+++ b/lib/c-api/src/wasm_c_api/externals/global.rs
@@ -95,16 +95,17 @@ mod tests {
                 wasm_val_t forty_two = WASM_F32_VAL(42);
                 wasm_val_t forty_three = WASM_F32_VAL(43);
 
-                wasm_valtype_t* type = wasm_valtype_new_i32();
-                wasm_globaltype_t* global_type = wasm_globaltype_new(type, WASM_CONST);
+                wasm_valtype_t* valtype = wasm_valtype_new_i32();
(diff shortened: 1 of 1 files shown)


lib/c-api/src/wasm_c_api/externals/global.rs
```diff
@@ -95,16 +95,17 @@ mod tests {
                 wasm_val_t forty_two = WASM_F32_VAL(42);
                 wasm_val_t forty_three = WASM_F32_VAL(43);
 
-                wasm_valtype_t* type = wasm_valtype_new_i32();
-                wasm_globaltype_t* global_type = wasm_globaltype_new(type, WASM_CONST);
+                wasm_valtype_t* valtype = wasm_valtype_new_i32();
+                wasm_globaltype_t* global_type = wasm_globaltype_new(valtype, WASM_CONST);
```
Hywan, line 99: Ownership of `valtype` is transferred to `wasm_globaltype_new`, so no need to free it, that's correct.

lib/c-api/src/wasm_c_api/externals/global.rs
```diff
@@ -95,16 +95,17 @@ mod tests {
...
                 wasm_global_t* global = wasm_global_new(store, global_type, &forty_two);
 
+                wasm_globaltype_delete(global_type);
```
Hywan, line 102: Ownership of `global_type` is _not_ transferred to `wasm_global_new`, so it's correct to free it!

Hywan (approved):
Good job :-).

PR #1914, opened 2020-12-10
Handle overflow errors in Bytes -> Pages conversion

webmaster128:
# Description

Before the Bytes to Pages conversion silently returned wrong results when the number of pages was too large.

# Review

### DIFF
--- a/lib/wasmer-types/src/units.rs
+++ b/lib/wasmer-types/src/units.rs
@@ -1,8 +1,10 @@
+use crate::lib::std::convert::TryFrom;
 use crate::lib::std::fmt;
 use crate::lib::std::ops::{Add, Sub};
 #[cfg(feature = "enable-serde")]
 use serde::{Deserialize, Serialize};
 use std::convert::TryInto;
+use thiserror::Error;
 
 /// WebAssembly page sizes are fixed to be 64KiB.
 /// Note: large page support may be added in an opt-in manner in the [future].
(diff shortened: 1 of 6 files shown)


lib/wasmer-types/src/units.rs
```diff
@@ -108,9 +109,16 @@ where
     }
 }
 
-impl From<Bytes> for Pages {
-    fn from(bytes: Bytes) -> Self {
-        Self((bytes.0 / WASM_PAGE_SIZE) as u32)
+const PAGES_EXCEED_UINT32: &'static str = "Number of pages exceeds uint32 range";
```
Hywan, line 112: You don't need a constant here. The constant is used only once.

lib/wasmer-types/src/units.rs
```diff
@@ -108,9 +109,16 @@ where
...
+
+impl TryFrom<Bytes> for Pages {
+    type Error = &'static str;
+
+    fn try_from(bytes: Bytes) -> Result<Self, Self::Error> {
+        let pages: u32 = (bytes.0 / WASM_PAGE_SIZE)
+            .try_into()
+            .or(Err(PAGES_EXCEED_UINT32))?;
```
Hywan, line 120: I believe you can write:

```suggestion
            .try_into()
            .map_err(PAGES_EXCEED_UINT32)?;
```

Hywan (commented):
Thank you very much for the PR! Apart from the little comment, everything looks good to me!

lib/wasmer-types/src/units.rs
webmaster128: `map_err` takes a callback, which is what I wanted to avoid. It would have to be

```rust
.map_err(|_original_error| PAGES_EXCEED_UINT32)?;
```

However, we can use this as well. Which way do you like more?

lib/wasmer-types/src/units.rs
webmaster128: I can inline it, no problem. Do you prefer that?

webmaster128:
Thanks for the positive feedback @Hywan. After looking at the error again I thought we better do this properly and use an error type instead of a generic string. This shows the user's type system that there there is only one possible error reason that in many use cases can be handled by defauling to `u32::MAX`.

syrusakbary (approved):

PR #1924, opened 2020-12-12
Export wasmer::{BaseTunables, Tunables}

### LINKED ISSUE #1872: Naming conflict Tunables between wasmer/wasmer_engine

webmaster128:
Closes #1872

ping @Hywan 

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - [#1851](https://github.com/wasmerio/wasmer/pull/1851) Improve test suite and documentation of the Wasmer C API
 - [#1874](https://github.com/wasmerio/wasmer/pull/1874) Set `CompilerConfig` to be owned (following wasm-c-api)
 - [#1880](https://github.com/wasmerio/wasmer/pull/1880) Remove cmake dependency for tests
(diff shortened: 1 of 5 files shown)


Hywan (approved):
It looks good to me. Thanks!

Issue #1237, opened 2020-02-21
Compiler configuration with `-O` or `-C`/`--codegen` flag on `wasmer`


Hywan:
Considering (i) #1212, (ii) we have multiple backends addressing different needs, (iii) we may add more profiling information or execution context, I'm more and more thinking about adding the very classical compiler `-C`/`--codegen` option to the `wasmer` executable. Here is a list of options I've in mind (inspired from `rustc`):

* `-C opt-level=value` where `value` can be:
  * `0` to `3`, where `0` implies `--generate-debug-info` (introduced in #1212),
  * `s` and `z` to optimise for speed or small size,
  * the option controls the output of the compiler, but we may even use this option to run `wasm-opt` on the Wasm module before starting compilation, just an idea,
* `-C backend-arguments=value` where `value` is different according to the backend:
   *  arguments to the LLVM backend (@nick may confirm that it could be useful to enable/disable some passes, checks or assertions),
   * arguments to the Singlepass backend (@losfair may confirm whether it is useful or not),
   * arguments to the Cranelift backend,
* `-C overflow-checks=yes/no` to check integer overflows,
* `-C debuginfo=value` where `value` can be:
  * `0` for no debug info,
  * `1` for line tables only,
  * `2` for full debug information,
* `-C profile-generate=value` to enable PGO (I assume it can work only with the LLVM backend for the moment, we don't support PGO in Singlepass or Cranelift, cc @nick),
* `-C profile-use=value` to use a PGO profile file,
* `-C target-cpu` and `target-features` to cross-compile or to enable some specific features for a target,
* `-C simd=yes/no` to replace the `--enable-simd` option,
* `-C thread=yes/no` to replace the `--enable-thread` option,
* `-C wasm-features=values` to replace the `--enable-all` features,
* `-C abi-arguments=values` where `value` depends on the ABI: emscripten or WASI, even if the current option (`--em-entrypoint`, `--em-symbol-map`, `--dir` etc.) are for runtime only, not compile-time. We should find a way to simplify those options too, that's for another issue.

And thus, we can have the very classical alias option `-Ovalue` for `-C opt-level=value`.

I think it is important to get a standard way to configure the backends and the compilations before adding too much options.

Also, it would imply to rethinkg a little bit the `wasmer` command. For the moment, we have the following commands:

* `run` to compile and cache (if not cached), and run a module,
* `cache` to clean the caches,
* `validate` to validate a module.

Here is my proposal:

[...]

PR #1932, opened 2020-12-14
Put unsandboxed Emscripten exec behind a flag

MarkMcCaskey:
Our implementation of the emscripten API can be fairly called "unsandboxed", this PR makes use of our emscripten ABI opt-in for security reasons.

# Review


lib/cli/src/commands/run.rs
```diff
@@ -54,6 +54,12 @@ pub struct Run {
     #[structopt(flatten)]
     wasi: Wasi,
 
+    /// Allow the Wasm module to be executed in an unsandboxed way with
+    /// the `Emscripten` ABI.
+    #[cfg(feature = "emscripten")]
+    #[structopt(long = "allow-emscripten")]
+    allow_emscripten: bool,
```
syrusakbary, line 61: I think a flag like `--emscripten` might be better and simpler to use for the user?

MarkMcCaskey: I agree, the only reason I went with `--allow-emscripten` was that `--emscripten` sounds like it _has_ to be `emscripten` or it will fail, whereas `allow-emscripten` makes the distinction that nothing is being forced to be anything, just that something else is now allowed.

Hywan: I agree with @MarkMcCaskey here.
Also, I take this opportunity to remind that I'm a bit afraid of addings flags and options one by one until having a long list of non-sense options. Hence this proposal https://github.com/wasmerio/wasmer/issues/1237.

It's not a blocker for this PR though!

Hywan (approved):

Hywan:
bors try

Issue #1922, opened 2020-12-12
Is it possible to avoid syscalls during Function::call?


ul:
### Summary

This question is in the context of using WASM for realtime audio synthesis. I invoke `Function::call` within an audio thread 48000 times per second. In the profiler,  `sigprogmask` and `sigaltstack` get a noticeable share of CPU time (see attached screenshot). While overhead itself is more or less acceptable as it stays constant with the growth of the executed WASM code, the unpredictability of the latency incurred by syscalls is bothersome for my use-case, I get a high variation in the trace with some calls taking longer than my per-frame budget. So my question is as follows: is it possible to implement `Function::call` without the need for syscalls?

### Additional details

<img width="919" alt="Screen Shot 2020-12-12 at 15 42 15" src="https://user-images.githubusercontent.com/233227/101975769-2c898080-3c93-11eb-98cc-06658ca8d4c2.png">

syrusakbary:
As functions can fail, we need to wrap them so if they fail we catch the traps properly.
However, this "catch" mechanism is done each time per function call and can indeed be optimized.

A few months ago we were thinking on creating an `unsafe` function `call_unchecked` so the user can catch errors themselves manually (rather than having Wasmer do it for them).
In your use case that means that you will choose how to catch the error. And you could save the 918ms spent on the traphandlers code.

How this would be implemented?

We would need a function similar to:
```rust
// Similar to wasmer_call_trampoline, but without catch_traps
pub unsafe fn wasmer_call_trampoline_unchecked(
    vmctx: VMFunctionEnvironment,
    trampoline: VMTrampoline,
    callee: *const VMFunctionBody,
    values_vec: *mut u8,
) -> Result<(), Trap> {
    mem::transmute::<_, extern "C" fn(VMFunctionEnvironment, *const VMFunctionBody, *mut u8)>(
        trampoline,
    )(vmctx, callee, values_vec)
}
```

And then the user would need to do something similar to:

```rust
catch_traps(vmctx, || {
    for i in 1..48000 {
        unsafe {
            func.call_unchecked(...);
        }
    }
}
```

This way, catch_traps will only be called once rather than 48,000 times and our code will still be safe to execute (returning any error when it happens.

ul:
Thank you for the explanation @syrusakbary! Indeed it would be nice to have such flexible API for traps to allow for some optimisations in cases like mine.

syrusakbary:
Hi @ul,

[...]

PR #1935, opened 2020-12-15
Provide an API for unchecked wasm calls and setting trap handlers manually

ul:
# Description

This is an attempt to provide an API which allows reducing syscall overhead when calling a hot wasm function, as outlined in #1922  

# Review

- [ ] Expose `wasmer::vm::catch_traps` together with `{Function, NativeFunc}::{call_unchecked, get_vmctx}` to allow reducing syscall overhead when calling a hot wasm function.

### DIFF
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -15,9 +15,9 @@ use std::fmt;
 use std::sync::Arc;
 use wasmer_engine::{Export, ExportFunction, ExportFunctionMetadata};
 use wasmer_vm::{
-    raise_user_trap, resume_panic, wasmer_call_trampoline, VMCallerCheckedAnyfunc,
-    VMDynamicFunctionContext, VMExportFunction, VMFunctionBody, VMFunctionEnvironment,
-    VMFunctionKind, VMTrampoline,
+    raise_user_trap, resume_panic, wasmer_call_trampoline, wasmer_call_trampoline_unchecked,
+    VMCallerCheckedAnyfunc, VMDynamicFunctionContext, VMExportFunction, VMFunctionBody,
+    VMFunctionEnvironment, VMFunctionKind, VMTrampoline,
 };
 
 /// A function defined in the Wasm module
@@ -488,6 +488,7 @@ impl Function {
         func: &WasmFunctionDefinition,
         params: &[Val],
         results: &mut [Val],
+        trampoline_checked: bool,
     ) -> Result<(), RuntimeError> {
         let format_types_for_error_message = |items: &[Val]| {
             items
@@ -530,15 +531,26 @@ impl Function {
         }
 
         // Call the trampoline.
-        if let Err(error) = unsafe {
-            wasmer_call_trampoline(
-                self.exported.vm_function.vmctx,
-                func.trampoline,
-                self.exported.vm_function.address,
-                values_vec.as_mut_ptr() as *mut u8,
-            )
-        } {
-            return Err(RuntimeError::from_trap(error));
+        if trampoline_checked {
+            if let Err(error) = unsafe {
+                wasmer_call_trampoline(
+                    self.exported.vm_function.vmctx,
+                    func.trampoline,
+                    self.exported.vm_function.address,
+                    values_vec.as_mut_ptr() as *mut u8,
+                )
+            } {
+                return Err(RuntimeError::from_trap(error));
+            }
+        } else {
+            unsafe {
+                wasmer_call_trampoline_unchecked(
+                    self.exported.vm_function.vmctx,
+                    func.trampoline,
+                    self.exported.vm_function.address,
+                    values_vec.as_mut_ptr() as *mut u8,
+                )
+            }
         }
 
         // Load the return values out of `values_vec`.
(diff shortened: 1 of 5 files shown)


lib/api/src/externals/function.rs
```diff
@@ -466,15 +467,26 @@ impl Function {
...
-        if let Err(error) = unsafe {
-            wasmer_call_trampoline(
-                self.exported.vm_function.vmctx,
-                func.trampoline,
-                self.exported.vm_function.address,
-                values_vec.as_mut_ptr() as *mut u8,
-            )
-        } {
-            return Err(RuntimeError::from_trap(error));
+        if trampoline_checked {
```
MarkMcCaskey, line 470: Seems good for a start, we can easily implement this without having to do the check at runtime though! We can make a trait and use static dispatch to generate 2 versions of this function.

We can do that as a follow up to this PR though

lib/api/src/externals/function.rs
```diff
@@ -424,6 +424,7 @@ impl Function {
         func: &WasmFunctionDefinition,
         params: &[Val],
         results: &mut [Val],
+        trampoline_checked: bool,
```
MarkMcCaskey, line 427: `fn call_wasm` now needs to be unsafe if we have a way to call it without checks.  Alternatively make a separate function.

We'll need to document in the `# Safety` section the correct way to call this function.

lib/api/src/externals/function.rs
```diff
@@ -562,14 +574,70 @@ impl Function {
...
+    /// let vmctx = sum.get_vmctx();
+    /// unsafe {
+    ///     vm::catch_traps(vmctx, || {
+    ///         for i in 0..10 {
+    ///             assert_eq!(sum.call(&[Value::I32(1), Value::I32(2)]).unwrap().to_vec(), vec![Value::I32(3)]);
+    ///         }
+    ///     });
+    /// }
+    /// ```
+    pub fn call_unchecked(&self, params: &[Val]) -> Result<Box<[Val]>, RuntimeError> {
```
MarkMcCaskey, line 623: This function must be `unsafe` and it should have all prerequisites to call it correctly documented in a `# Safety` section.

lib/api/src/externals/function.rs
```diff
@@ -562,14 +574,70 @@ impl Function {
...
+                self.call_wasm(&wasm, params, &mut results, false)?;
+            }
+            _ => unimplemented!("The function definition isn't supported for the moment"),
+        }
+
+        Ok(results.into_boxed_slice())
+    }
+
+    /// Get a VM Context to manually handle traps when calling `call_unchecked`.
+    pub fn get_vmctx(&self) -> VMFunctionEnvironment {
```
MarkMcCaskey, line 637: This function needs additional documentation about how to use it correctly.

It may also need to be `unsafe` depending on which functions exist that can take `VMFunctionEnvironment`.

`VMFunctionEnvironment` can either be a real `vmctx` or a host env, for `catch_traps` we must have a real `vmctx`. Which means this method must be called on an exported Wasm function.

This method doesn't strictly have to be unsafe if all other functions take a `VMFunctionEnvironment` are `unsafe` (which they should be); though the distinction between the types of `VMFunctionEnvironment` would be a useful thing to document here.

lib/api/src/native.rs
```diff
@@ -112,8 +112,26 @@ macro_rules! impl_native_traits {
...
         {
+
             /// Call the typed func and return results.
             pub fn call(&self, $( $x: $x, )* ) -> Result<Rets, RuntimeError> {
+                self.call_(true, $( $x, )*)
+            }
+
+            /// Call the typed func and return results.
+            /// This would not catch any traps and must be called within a callback passed to `wasmer_vm::catch_traps`.
+            pub fn call_unchecked(&self, $( $x: $x, )* ) -> Result<Rets, RuntimeError> {
```
MarkMcCaskey, line 123: This function must be `unsafe` with prerequisites listed in a `# Safety` section of a doc comment.

lib/api/src/native.rs
```diff
@@ -112,8 +112,26 @@ macro_rules! impl_native_traits {
...
+                self.call_(false, $( $x, )*)
+            }
+
+            /// Get a VM Context to manually handle traps when calling `call_unchecked`.
+            pub fn get_vmctx(&self) -> VMFunctionEnvironment {
```
MarkMcCaskey, line 128: Similar thing to what I mentioned before

lib/api/src/native.rs
```diff
@@ -112,8 +112,26 @@ macro_rules! impl_native_traits {
...
+                self.exported.vm_function.vmctx
+            }
+
+            /// Call the typed func and return results.
+            /// When calling a wasm function, allows to do checked or unchecked call.
+            fn call_(&self, trampoline_checked: bool, $( $x: $x, )* ) -> Result<Rets, RuntimeError> {
```
MarkMcCaskey, line 134: This function must be `unsafe` and have the prerequisites for correct usage documented with a `# Safety` section.

Also, we can avoid the runtime check here as well.

lib/vm/src/trap/traphandlers.rs
```diff
@@ -459,6 +459,35 @@ pub unsafe fn wasmer_call_trampoline(
...
+///   callee vmctx, the caller vmctx, the `callee` argument below, and then the
+///   `values_vec` argument.
+/// * `callee` - the third argument to the `trampoline` function
+/// * `values_vec` - points to a buffer which holds the incoming arguments, and to
+///   which the outgoing return values will be written.
+///
+/// # Safety
+///
+/// Wildly unsafe because it calls raw function pointers and reads/writes raw
+/// function pointers.
```
MarkMcCaskey, line 479: `# Safety` section should be updated with information about how to call this correctly (that is, how do you do the "check" that's being omitted)

MarkMcCaskey (commented):
Looks like a good start -- thanks for trying to get this done!

I was looking into this as well and I discovered that it's actually pretty tricky.

There are 2 types of checks that we have: `catch_traps` (for Wasm functions) and `std::panic::catch_unwind(std::panic::AssertUnwindSafe(||` for host functions.  I don't think it's reasonable to expect users to remember/identify which check they need, additionally as of right now `catch_traps` needs a vmctx for the `Instance` being used which means the user can't call other Wasm functions that don't also share that `vmctx`.

Perhaps we can make a wrapper function for controlling the context in which this happens, something like:

```Rust
    /// TODO:
    /// Wrapper to safely call many functions with less overhead.
    ///
    /// # Safety
    /// - All Wasm functions must be from a single `Instance`.
    /// - A valid function exported from Wasm must be passed in
    ///   `wasm_funtion`: this function uses this Wasm function to find
    ///   the instance that all the Wasm functions will use.
    pub unsafe fn call_many<F, Ret>(f: F, wasm_function: &Function) -> Result<Ret, RuntimeError>
    where
        Ret: Sized,
        F: FnMut() -> Ret,
    {
        let mut result: Ret;
        let vmctx = match wasm_function {
            Function {
                definition: FunctionDefinition::Wasm(_),
                exported:
                    ExportFunction {
                        vm_function: VMExportFunction { vmctx, .. },
                        ..
                    },
                ..
            } => vmctx,
            // TODO: proper error handling?
            _ => panic!("`wasm_function` must be a Wasm function"),
        };
        // TODO: is this nested order correct?
        std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
            wasmer_vm::catch_traps(*vmctx, move || {
                result = f();
            })
        }))
        .map_err(|e| RuntimeError::new(format!("{:?}", e)))?;
        Ok(result)
    }
    ```
    
    which is some wip code I made when looking into this this morning.

**Comment by MarkMcCaskey** [comment]
@ul I just heard from @syrusakbary that he wants to ship a change changing `catch_traps` first, so heads up on that. If you want to reduce the number of times we iterate on this, you probably want to wait for those changes to land!

**Comment by syrusakbary** [comment]
Some more context: I want to make `catch_traps` vmctx agnostic. And to take a `&Store` instead. So once we follow up with that we should be good to merge this PR!

File: lib/api/src/native.rs (hunk 6 above)
Comment (syrusakbary) [reply lib/api/src/native.rs:128]: Note: we shouldn't need this once a future `catch_traps` refactor (using `&Store` instead of `vmctx`) lands

File: lib/api/src/externals/function.rs (hunk 4 above)
Comment (syrusakbary) [reply lib/api/src/externals/function.rs:637]: Note: we shouldn't need this function once a future `catch_traps` refactor (using `&Store` instead of `vmctx`) lands

File: lib/api/src/externals/function.rs (hunk 1 above)
Comment (syrusakbary) [reply lib/api/src/externals/function.rs:470]: Can we have the function as `#[inline]` so the compiler does that for us?

Comment (MarkMcCaskey) [reply lib/api/src/externals/function.rs:470]: Yeah! Although I believe you don't need to mark functions as `#[inline]` to have them be inlined! In fact I'm not sure if `#[inline]` does anything other than document that you want it be inlined. I know `#[inline(always)]` and `#[inline(never)]` definitely do things though.

That is one of the nice things about static dispatch in situations like this: we trade binary size (i.e. we generate more code) for specialized versions of the functions that are optimized to do 1 thing.

Comment (MarkMcCaskey) [reply lib/api/src/externals/function.rs:470]: To clarify what I mean by using static dispatch here we can have something like:

```Rust
trait Trait {
    unsafe fn call_trampoline(..) -> ...;
}

struct Checked;
impl Trait for Checked {
   unsafe fn call_trampoline(..) -> .. {
      call_checked(..)
    }
}

struct Unchecked;
impl Trait for Unchecked {
   unsafe fn call_trampoline(..) -> .. {
      call_unchecked(..)
    }
}

fn call<T: Trait>(...) -> ... {
   // then instead of an if statement we just call
   unsafe {
      T::call_trampoline(..)
    }
}
```

ul:
Thanks for a thorough review and supplied context! I'm for sure keen to wait for `catch_trap` changes.

syrusakbary:
Closing this PR in favor of #2319

PR #1936, opened 2020-12-15
feat: Rename the `wasmer-engine-native` to `wasmer-engine-shared-library`

Hywan:
# Description

This engine had a bad name, since “native” means everything and
nothing at the same time. This engine generates a shared object
library, so we decided to name it `wasmer-engine-shared-library`.

To do once this PR is merged:

- [ ] Update the documentation (docs.wasmer.io)
- [ ] Update the Github issue labels
- [ ] Update the language integrations
- [ ] Do something on crates.io to yank `wasmer-engine-native` probably?

# Review


Hywan:
bors try

lib/cli/src/store.rs
```diff
@@ -19,16 +19,16 @@ pub struct StoreOptions {
...
-    #[structopt(long, conflicts_with_all = &["native", "object_file"])]
+    /// Use the JIT Engine.
+    #[structopt(long, conflicts_with_all = &["shared_library", "object_file"])]
     jit: bool,
 
-    /// Use Native Engine.
+    /// Use the Shared Library Engine.
     #[structopt(long, conflicts_with_all = &["jit", "object_file"])]
-    native: bool,
+    shared_library: bool,
```
MarkMcCaskey, line 28: Perhaps add an alias of `native`? Unsure what we want to do in regards to CLI flag stability

MarkMcCaskey (approved):
We'll need to put a task somewhere to publish under the old name once more (in a non-breaking way) marking it as deprecated and linking to the new one in the readme.

I'm not 100% on the new name but I do think `native` is a confusing name and I've probably gotten used to it, so I'd say this name change is probably an upgrade 👍

lib/cli/src/store.rs
Hywan: The 1.0 is not released yet, not sure we want to have compatibility between alpha, beta, and stable version 🤔.

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

Hywan:
Closed in favor of #2356.

PR #1941, opened 2020-12-16
Allow using get_remaining_points/set_remaining_points without the Metering instance

webmaster128:
# Description

In order to be able to use `get_remaining_points`/`set_remaining_points` in an ahead of time compilation scenario, those should be acessible withouht the compile time tooling. It turns out this is easy since `&self` is never used anyways.

In this PR I turn them into free function. I tried static functions first (i.e. calls like `Metering::get_remaining_points(&instance)`), but this brings users in big trouble with respect to generic arguments of `struct Metering<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync>`. We could use a different struct for namespacing purposes (e.g. `MeteringManager::get_remaining_points(&instance)`), but this feels like Java style.

# Review

### DIFF
--- a/examples/metering.rs
+++ b/examples/metering.rs
@@ -21,7 +21,7 @@ use wasmer::CompilerConfig;
 use wasmer::{imports, wat2wasm, Instance, Module, Store};
 use wasmer_compiler_cranelift::Cranelift;
 use wasmer_engine_jit::JIT;
-use wasmer_middlewares::Metering;
+use wasmer_middlewares::metering::{get_remaining_points, set_remaining_points, Metering};
 
 fn main() -> anyhow::Result<()> {
     // Let's declare the Wasm module.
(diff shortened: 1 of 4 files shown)


examples/metering.rs
```diff
@@ -21,7 +21,7 @@ use wasmer::CompilerConfig;
 use wasmer::{imports, wat2wasm, Instance, Module, Store};
 use wasmer_compiler_cranelift::Cranelift;
 use wasmer_engine_jit::JIT;
-use wasmer_middlewares::Metering;
+use wasmer_middlewares::{get_remaining_points, set_remaining_points, Metering};
```
jubianchi, line 24: Can we have the `get_remaining_points`, `set_remaining_points` only in `metering`?

I'm afraid in the future, when we'll have more middlewares that we have either conflicts os `use` satatements looking like 
```
use wasmer_middlewares::{get_remaining_points, set_remaining_points, some_middleware_func, other-middleware_func, yet_another_middleware_func, Metering};
```

Having something like 

```
use wasmer_middlewares::{
    metering::{get_remaining_points, set_remaining_points}, 
    Metering,
};
```

I think showing this as an example seems cleaner (to me at least, but I may be wrong).

webmaster128: Nice idea! As you can see in the PR description I struggeled with namespacing of those. I did not realize the `metering` module is public already. Will update.

Otherwise good?

webmaster128: Updated in the latest commit

syrusakbary (approved):

PR #1942, opened 2020-12-16
Prepare for 1.0.0-beta2 release

MarkMcCaskey:
Let's get some of the super useful changes we've been making into the hands of our users!

# Review

### DIFF
--- a/scripts/publish.py
+++ b/scripts/publish.py
@@ -21,7 +21,7 @@ except ImportError:
 
 
 # TODO: find this automatically
-target_version = "1.0.0-beta1"
+target_version = "1.0.0-beta2"
 
 # TODO: generate this by parsing toml files
 dep_graph = {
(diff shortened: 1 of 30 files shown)


scripts/publish.py
```diff
@@ -49,6 +49,7 @@
     "wasmer-c-api": set(["wasmer", "wasmer-compiler", "wasmer-compiler-cranelift", "wasmer-compiler-singlepass",
                          "wasmer-compiler-llvm", "wasmer-emscripten", "wasmer-engine", "wasmer-engine-jit",
                          "wasmer-engine-native", "wasmer-engine-object-file", "wasmer-wasi", "wasmer-types"]),
+    "wasmer-middlewares": set(["wasmer", "wasmer-types", "wasmer-vm"]),
```
nlewycky, line 52: (No action required:) huh, so beta1 didn't have the middleware? Were the users using the revision from git then?

nlewycky (approved):

scripts/publish.py
jubianchi: I think yes.

jubianchi (approved):

scripts/publish.py
MarkMcCaskey: Yeah, it seems so!  We updated the version but did not publish it last time

PR #1944, opened 2020-12-16
Clean up host env code, add more `Send`, `Sync` bounds

MarkMcCaskey:
This PR contains:
- Code clean up, refactored the unsafe metadata creation into a function which can probably be safe
- Add `Send` + `Sync` bounds in some places where they should have been to begin with

# Review

### DIFF
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -67,6 +67,49 @@ pub struct Function {
     pub(crate) exported: ExportFunction,
 }
 
+fn build_export_function_metadata<Env>(
+    env: Env,
+    import_init_function_ptr: for<'a> fn(
+        &'a mut Env,
+        &'a crate::Instance,
+    ) -> Result<(), crate::HostEnvInitError>,
+) -> (*mut std::ffi::c_void, ExportFunctionMetadata)
+where
+    Env: Clone + Sized + 'static + Send + Sync,
+{
+    let import_init_function_ptr = Some(unsafe {
+        std::mem::transmute::<fn(_, _) -> Result<(), _>, fn(_, _) -> Result<(), _>>(
+            import_init_function_ptr,
+        )
+    });
+    let host_env_clone_fn: fn(*mut std::ffi::c_void) -> *mut std::ffi::c_void = |ptr| {
+        let duped_env: Env = unsafe {
+            let ptr: *mut Env = ptr as _;
+            let item: &Env = &*ptr;
+            item.clone()
+        };
+        Box::into_raw(Box::new(duped_env)) as _
+    };
+    let host_env_drop_fn: fn(*mut std::ffi::c_void) = |ptr: *mut std::ffi::c_void| {
+        unsafe { Box::from_raw(ptr as *mut Env) };
+    };
+    let env = Box::into_raw(Box::new(env)) as _;
+
+    // # Safety
+    // - All these functions work on all threads
+    // - The host env is `Send`.
+    let metadata = unsafe {
+        ExportFunctionMetadata::new(
+            env,
+            import_init_function_ptr,
+            host_env_clone_fn,
+            host_env_drop_fn,
+        )
+    };
+
+    (env, metadata)
+}
+
 impl Function {
     /// Creates a new host `Function` (dynamic) with the provided signature.
     ///
@@ -104,7 +147,7 @@ impl Function {
     pub fn new<FT, F>(store: &Store, ty: FT, func: F) -> Self
     where
         FT: Into<FunctionType>,
-        F: Fn(&[Val]) -> Result<Vec<Val>, RuntimeError> + 'static,
+        F: Fn(&[Val]) -> Result<Vec<Val>, RuntimeError> + 'static + Send + Sync,
     {
         let ty: FunctionType = ty.into();
         let dynamic_ctx: VMDynamicFunctionContext<DynamicFunctionWithoutEnv> =
@@ -209,7 +252,7 @@ impl Function {
     pub fn new_with_env<FT, F, Env>(store: &Store, ty: FT, env: Env, func: F) -> Self
     where
         FT: Into<FunctionType>,
-        F: Fn(&Env, &[Val]) -> Result<Vec<Val>, RuntimeError> + 'static,
+        F: Fn(&Env, &[Val]) -> Result<Vec<Val>, RuntimeError> + 'static + Send + Sync,
         Env: Sized + WasmerEnv + 'static,
     {
         let ty: FunctionType = ty.into();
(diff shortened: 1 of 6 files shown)


lib/api/src/externals/function.rs
```diff
@@ -67,6 +67,49 @@ pub struct Function {
     pub(crate) exported: ExportFunction,
 }
 
+fn build_export_function_metadata<Env>(
+    env: Env,
+    import_init_function_ptr: for<'a> fn(
+        &'a mut Env,
+        &'a crate::Instance,
+    ) -> Result<(), crate::HostEnvInitError>,
```
syrusakbary, line 75: What is the `for` syntax? First time I'm seeing this!

lib/api/src/externals/function.rs
```diff
@@ -104,7 +147,7 @@ impl Function {
     pub fn new<FT, F>(store: &Store, ty: FT, func: F) -> Self
     where
         FT: Into<FunctionType>,
-        F: Fn(&[Val]) -> Result<Vec<Val>, RuntimeError> + 'static,
+        F: Fn(&[Val]) -> Result<Vec<Val>, RuntimeError> + 'static + Send + Sync,
```
syrusakbary, line 150: What it means for require a function to be `Send + Sync` ?
Could it still be possible to be called in parallel through different threads?

lib/api/src/externals/function.rs
MarkMcCaskey: Haha yeah, it's not super common

it's used to bind a lifetime in a scope, it's a universal qualifier (like ∀), it's just saying "for all lifetimes `'a`, this thing, it's a way of constraining the type (and it's required here because it'll be inferred incorrectly)

lib/api/src/externals/function.rs
MarkMcCaskey: I believe `Sync` with `Fn` does mean that the function can be called from multiple threads at the same time. `Send` just means it can be executed on any thread.

For `Send`, this means functions can't use thread-local variables.

For `Sync`, this means functions should not use unsynchronized global variables.

I can't think of any other cases where this is really an issue right now, I'm sure there are a few other restrictions, but most functions that use `unsafe` correctly or don't use `unsafe` at all should be both `Send` and `Sync`

lib/api/src/externals/function.rs
```diff
@@ -67,6 +67,49 @@ pub struct Function {
...
+        let duped_env: Env = unsafe {
+            let ptr: *mut Env = ptr as _;
+            let item: &Env = &*ptr;
+            item.clone()
+        };
+        Box::into_raw(Box::new(duped_env)) as _
+    };
+    let host_env_drop_fn: fn(*mut std::ffi::c_void) = |ptr: *mut std::ffi::c_void| {
+        unsafe { Box::from_raw(ptr as *mut Env) };
+    };
```
Hywan, line 95: Just for the sake of consistency with my previous proposal, and for more simplicity:

```suggestion
    let host_env_drop_fn: fn(*mut std::ffi::c_void) = |ptr| {
        unsafe { Box::from_raw(ptr.cast::<Env>()) };
    };
```

lib/api/src/externals/function.rs
Hywan: I think it also means that the environment captured by the function must also implement `Send` + `Sync`.

Confirmed with:

```rust
use std::rc::Rc;

fn main() {
    fn check<F>(f: F) where F: Fn() + 'static + Send + Sync {}
    
    let x = Rc::new(42);
    check(|| { let _ = x; });
}
```

which raises the following issue:

```
error[E0277]: `Rc<i32>` cannot be shared between threads safely
 --> src/main.rs:7:5
  |
4 |     fn check<F>(f: F) where F: Fn() + 'static + Send + Sync {}
  |                                                 ---- required by this bound in `check`
...
7 |     check(|| { let _ = x; });
  |     ^^^^^ `Rc<i32>` cannot be shared between threads safely
  |
  = help: the trait `Sync` is not implemented for `Rc<i32>`
  = note: required because of the requirements on the impl of `Send` for `&Rc<i32>`
  = note: required because it appears within the type `[closure@src/main.rs:7:11: 7:28]`
```

([see the playground](https://play.rust-lang.org/))

lib/api/src/externals/function.rs
```diff
@@ -67,6 +67,49 @@ pub struct Function {
...
+    let metadata = unsafe {
+        ExportFunctionMetadata::new(
+            env,
+            import_init_function_ptr,
+            host_env_clone_fn,
+            host_env_drop_fn,
+        )
+    };
+
+    (env, metadata)
```
Hywan, line 110: Why not just returning `metadata` here, and let the rest of the code accessing `metadata.env` to get the environment? Is it by convenience?

lib/api/src/externals/function.rs
```diff
@@ -220,58 +263,27 @@ impl Function {
                 function_type: ty.clone(),
             });
 
+        let import_init_function_ptr: for<'a> fn(&'a mut _, &'a _) -> Result<(), _> =
+            |env: &mut VMDynamicFunctionContext<DynamicFunctionWithEnv<Env>>,
+             instance: &crate::Instance| {
+                Env::init_with_instance(&mut *env.ctx.env, instance)
+            };
```
Hywan, line 270: 👍

Hywan (approved):
I like this PR. It's a very welcomed clean up!

I'm approving it to not be a blocker, but I would be delighted to see my comments addressed 🙂.

lib/api/src/externals/function.rs
Hywan: Yes, obviously there is a difference. The difference with `'a` on the function itself is that it will be inferred from the calling context. That's not what we want! Forget it.

lib/api/src/externals/function.rs
```diff
@@ -67,6 +67,49 @@ pub struct Function {
...
+    ) -> Result<(), crate::HostEnvInitError>,
+) -> (*mut std::ffi::c_void, ExportFunctionMetadata)
+where
+    Env: Clone + Sized + 'static + Send + Sync,
+{
+    let import_init_function_ptr = Some(unsafe {
+        std::mem::transmute::<fn(_, _) -> Result<(), _>, fn(_, _) -> Result<(), _>>(
+            import_init_function_ptr,
+        )
+    });
```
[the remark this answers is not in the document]
MarkMcCaskey: Yeah, that's true, I thought listing both types was good documentation though: `std::mem::transmute::<fn(_, _) -> Result<(), _>, fn(_, _) -> Result<(), _>>(` communicates that the function arguments and error value are changing, but the number of arguments, and the fact that it's `Result<(),` are not.  I'll rereview the code to see if this extra info is useful, but I think it's good to be super explicit with things like `transmute` as so many things can go wrong

lib/api/src/externals/function.rs
```diff
@@ -67,6 +67,49 @@ pub struct Function {
...
+    let host_env_clone_fn: fn(*mut std::ffi::c_void) -> *mut std::ffi::c_void = |ptr| {
+        let duped_env: Env = unsafe {
+            let ptr: *mut Env = ptr as _;
+            let item: &Env = &*ptr;
+            item.clone()
+        };
+        Box::into_raw(Box::new(duped_env)) as _
+    };
```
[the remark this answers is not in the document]
MarkMcCaskey: Huh I had no idea `cast` was a method on plain pointers... Good to know!

lib/api/src/externals/function.rs
MarkMcCaskey: `metadata.env` is private! It can only be seen in `wasmer_engine` not in `wasmer_API`, this is for the best as the contents of the `metadata` struct can be dangerous

lib/api/src/externals/function.rs
MarkMcCaskey: @Hywan , I think having a `Send + Sync` environment is ultimately what we need for the Wasmer API... because `Instance` is `Send + Sync`, so a single function can be called from multiple threads concurrently and in parallel.  I don't know a good way around this yet but I think `Send + Sync` does fix everything for us

MarkMcCaskey: The alternative might be to have separate code and abstractions for non-thread safe things, but that seems pretty complicated 🤔 . I think encouraging code to work on multiple threads by default is a very Rusty thing anyways and I think it positions us and users of `wasmer` well to scale their applications

PR #1947, opened 2020-12-16
Include a NUL byte in the message returned by wasm_trap_message().

nlewycky:


jubianchi (approved):

MarkMcCaskey (approved):
Interesting, I assumed that strings wouldn't be nul-terminated because we pass back a pointer and a length; what was the motivating reason for this change?

nlewycky:
wasm.h has this to say:
```
typedef wasm_name_t wasm_message_t;  // null terminated
```
so I'd guess that wasm_message_t should be NUL terminated but wasm_name_t may not be.

In practice it came up in wasm-c-api/examples/start.cc which prints out a string by pointer:
```
  std::cout << "> " << trap->message().get() << std::endl;
```
Since it's using a pointer with no length, operator<< calls strlen() to find the end of the string which in turn reads off the end.

MarkMcCaskey:
That's good to know!  I'd expect this to be a more widespread issue, affecting everywhere `message` is used

Hywan:
:+1:

Hywan:
This PR was actually incorrect. `wasm_trap_t` works on `wasm_message_t` which is supposed to always contain a null-terminated string. But `wasm_trap_t` uses `RuntimeError` which is a valid Rust string that must not contain a null byte. In the current code at the time of writing, `RuntimeError` receives a null terminated string (which is arguably incorrect), a `wasm_trap_message` always add a new null byte at the end of the string. It results in a doubly-null-terminated string. This is incorrect :-).

PR #1949, opened 2020-12-17
test+doc(c-api) Fix vecs that must be boxed vecs + don't transmute uninitialized boxed vecs

Hywan:
# Description

This PR improve test coverage of the C API.

~It's a work in progress. I've paused this work because we need to focus on something else, but one of the two test cases that have been added reveals an important bug.~

The `wasm_frame_vec_t`, `wasm_functype_vec_t`, `wasm_globaltype_vec_t`, `wasm_memorytype_vec_t` and `wasm_tabletype_vec_t` are incorrectly defined. Indeed, in `wasm.h` they are all defined with `WASM_DECLARE_TYPE` (except for `wasm_frame_vec_t`, but it's exactly the same), which declares the vectors as `WAM_DECLARE_VEC(name, *)`. This `*` is important. It means that we _must not_ declare thoses types with our own `wasm_declare_vec!` macro, but rather with `wasm_declare_boxed_vec!`.

~Fixing this bug impact many places in the code. I didn't have time for the moment, but it's probably not a big deal!~ This PR fixes those bugs with https://github.com/wasmerio/wasmer/pull/1949/commits/e10ad512bcdcac420d0f1206b48e33aae0833c75, https://github.com/wasmerio/wasmer/pull/1949/commits/75ceb0d9e0b8bd97777a0b6443f44b089f0564c2, https://github.com/wasmerio/wasmer/pull/1949/commits/a3c7a2d75235ea1e606c2432ace6bfbd0041157c, https://github.com/wasmerio/wasmer/pull/1949/commits/a0c34fe850069db5fdc567190044df0f3ae4cbcf. Bonus for `wasm_frame_t` that was also incorrectly implemented, the fix lands in https://github.com/wasmerio/wasmer/pull/1949/commits/4abf6f81da5e6b053ffa132d0c58b4ed4a8612a5 and is possible thanks to https://github.com/wasmerio/wasmer/pull/1949/commits/43026e651ed0d22f45c228033a39ebec6db259b9, with a side effect of simplifying the code in https://github.com/wasmerio/wasmer/pull/1949/commits/940dea72e7e78ea8f4abf167e94ae6d5c3ac4c46.

While continuing to improve the test coverage, another bug has been found. In case of a boxed vector, `wasm_$name_vec_delete` was crashing if the given vector was uninitialized. Basically the following code was crashing:

```
wasm_exporttype_vec_t vector;
wasm_exporttype_vec_new_uninitialized(&vector, 3);
wasm_exporttype_vec_delete(&vector);
```

This is now fixed in https://github.com/wasmerio/wasmer/pull/1949/commits/8aa08225cd5575bb6a590bdf11dcfd0ceb772203, and tested in https://github.com/wasmerio/wasmer/pull/1949/commits/96169de8f0e6f4f693a8c46dfce493df10eca7ac.

The PR adds 34 test cases, and has found+fixes 6 bugs.

# Review

### DIFF
--- a/lib/c-api/src/wasm_c_api/macros.rs
+++ b/lib/c-api/src/wasm_c_api/macros.rs
@@ -345,13 +345,7 @@ See the [`wasm_" $name "_vec_t`] type to get an example."]
                 if let Some(vec) = ptr {
                     if !vec.data.is_null() {
                         let data: Vec<*mut [<wasm_ $name _t>]> = Vec::from_raw_parts(vec.data, vec.size, vec.size);
-
-                        // If the vector has been initialized (we check
-                        // only the first item), we can transmute items to
-                        // `Box`es.
-                        if vec.size > 0 && !data[0].is_null() {
-                            let _data: Vec<Box<[<wasm_ $name _t>]>> = ::std::mem::transmute(data);
-                        }
+                        let _data: Vec<Option<Box<[<wasm_ $name _t>]>>> = ::std::mem::transmute(data);
 
                         vec.data = ::std::ptr::null_mut();
                         vec.size = 0;
--- a/lib/c-api/src/wasm_c_api/macros.rs
+++ b/lib/c-api/src/wasm_c_api/macros.rs
@@ -274,7 +274,14 @@ Read the documentation of [`wasm_" $name "_t`] to see more concrete examples."]
                 let vec = &mut *ptr;
                 if !vec.data.is_null() {
                     let data: Vec<*mut [<wasm_ $name _t>]> = Vec::from_raw_parts(vec.data, vec.size, vec.size);
-                    let _data: Vec<Box<[<wasm_ $name _t>]>> = ::std::mem::transmute(data);
+
+                    // If the vector has been initialized (we check
+                    // only the first item), we can transmute items to
+                    // `Box`es.
+                    if vec.size > 0 && !data[0].is_null() {
+                        let _data: Vec<Box<[<wasm_ $name _t>]>> = ::std::mem::transmute(data);
+                    }
+
                     vec.data = ::std::ptr::null_mut();
                     vec.size = 0;
                 }
@@ -299,7 +306,6 @@ macro_rules! wasm_declare_ref_base {
             }
 
             // TODO: finish this...
-
         }
     };
 }
--- a/lib/c-api/src/wasm_c_api/macros.rs
+++ b/lib/c-api/src/wasm_c_api/macros.rs
@@ -326,7 +326,7 @@ int main() {
 ```"]
             #[no_mangle]
             pub unsafe extern "C" fn [<wasm_ $name _vec_new_uninitialized>](out: *mut [<wasm_ $name _vec_t>], length: usize) {
-                let mut bytes: Vec<*mut [<wasm_ $name _t>]> = Vec::with_capacity(length);
+                let mut bytes: Vec<*mut [<wasm_ $name _t>]> = vec![::std::ptr::null_mut(); length];
                 let pointer = bytes.as_mut_ptr();
 
                 (*out).data = pointer;
--- a/lib/c-api/src/wasm_c_api/macros.rs
+++ b/lib/c-api/src/wasm_c_api/macros.rs
@@ -341,21 +341,21 @@ int main() {
 
 See the [`wasm_" $name "_vec_t`] type to get an example."]
             #[no_mangle]
-            pub unsafe extern "C" fn [<wasm_ $name _vec_delete>](ptr: *mut [<wasm_ $name _vec_t>]) {
-                let vec = &mut *ptr;
+            pub unsafe extern "C" fn [<wasm_ $name _vec_delete>](ptr: Option<&mut [<wasm_ $name _vec_t>]>) {
+                if let Some(vec) = ptr {
+                    if !vec.data.is_null() {
+                        let data: Vec<*mut [<wasm_ $name _t>]> = Vec::from_raw_parts(vec.data, vec.size, vec.size);
 
-                if !vec.data.is_null() {
-                    let data: Vec<*mut [<wasm_ $name _t>]> = Vec::from_raw_parts(vec.data, vec.size, vec.size);
+                        // If the vector has been initialized (we check
+                        // only the first item), we can transmute items to
+                        // `Box`es.
+                        if vec.size > 0 && !data[0].is_null() {
+                            let _data: Vec<Box<[<wasm_ $name _t>]>> = ::std::mem::transmute(data);
+                        }
 
-                    // If the vector has been initialized (we check
-                    // only the first item), we can transmute items to
-                    // `Box`es.
-                    if vec.size > 0 && !data[0].is_null() {
-                        let _data: Vec<Box<[<wasm_ $name _t>]>> = ::std::mem::transmute(data);
+                        vec.data = ::std::ptr::null_mut();
+                        vec.size = 0;
                     }
-
-                    vec.data = ::std::ptr::null_mut();
-                    vec.size = 0;
                 }
             }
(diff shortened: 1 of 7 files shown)


lib/c-api/src/wasm_c_api/macros.rs
```diff
@@ -153,6 +246,27 @@ macro_rules! wasm_declare_boxed_vec {
                 }
             }
 
+            impl<'a, T: Into<[<wasm_ $name _t>]> + Clone> From<&'a [T]> for [<wasm_ $name _vec_t>] {
```
Hywan, line 249: Such implementation exists for regular vecs. Now it exists for boxed vecs too.

lib/c-api/src/wasm_c_api/macros.rs
```diff
@@ -171,33 +285,75 @@ macro_rules! wasm_declare_boxed_vec {
...
                 if !vec.data.is_null() {
                     let data: Vec<*mut [<wasm_ $name _t>]> = Vec::from_raw_parts(vec.data, vec.size, vec.size);
-                    let _data: Vec<Box<[<wasm_ $name _t>]>> = ::std::mem::transmute(data);
+
+                    // If the vector has been initialized (we check
+                    // only the first item), we can transmute items to
+                    // `Box`es.
+                    if vec.size > 0 && !data[0].is_null() {
+                        let _data: Vec<Box<[<wasm_ $name _t>]>> = ::std::mem::transmute(data);
+                    }
```
Hywan, line 355: This changes allows to delete an uninitialized boxed vec. Should we check for all items or checking the first is safe enough?

Hywan:
bors try

Hywan:
Too bad, it's working on `aarch64-apple-darwin`, but not on `x86_64`…

Hywan:
Following `Vec::with_capacity`, we are redirected to [`RawVec::with_capacity`](https://github.com/rust-lang/rust/blob/e261649593cf9c2707f7b30a61c46c4469c67ebb/library/alloc/src/raw_vec.rs#L92). If we continue the lead, we reach `RawVec::with_capacity_in`:

https://github.com/rust-lang/rust/blob/e261649593cf9c2707f7b30a61c46c4469c67ebb/library/alloc/src/raw_vec.rs#L128-L130

Notice the difference with `RawVec::with_capacity_zeroed_in`:

https://github.com/rust-lang/rust/blob/e261649593cf9c2707f7b30a61c46c4469c67ebb/library/alloc/src/raw_vec.rs#L135-L137

Ideally, the `wasm_$name_vec_new_uninitialized` should zeroed the items, for safety reasons. Otherwise, this change https://github.com/wasmerio/wasmer/pull/1949/commits/8aa08225cd5575bb6a590bdf11dcfd0ceb772203 will not work because we have no guarantee that the items are zeroed. This API is public for `RawVec`; it is only used by `slice::new_zeroed_slice`, which is unstable and behind a feature flag.

We could probably just used `vec![ptr::null(); len]`: yes the vector will allocate memory immediately and everytime, but they are all aimed to be filled one day.

lib/c-api/src/wasm_c_api/macros.rs
```diff
@@ -171,33 +285,75 @@ macro_rules! wasm_declare_boxed_vec {
...
             }
 
-            #[doc = "Deletes a vector of [`wasm_" $name "_t`]."]
+            #[doc = "Deletes a vector of [`wasm_" $name "_t`].
+
+# Example
+
+See the [`wasm_" $name "_vec_t`] type to get an example."]
             #[no_mangle]
             pub unsafe extern "C" fn [<wasm_ $name _vec_delete>](ptr: *mut [<wasm_ $name _vec_t>]) {
```
MarkMcCaskey, line 344: We can change this to `ptr: Option<&mut [<wasm_ $name _vec_t>]`  like the other delete function

lib/c-api/src/wasm_c_api/macros.rs
```diff
@@ -171,33 +285,75 @@ macro_rules! wasm_declare_boxed_vec {
...
                 let vec = &mut *ptr;
```
MarkMcCaskey, line 345: Generally `delete` functions should no-op on `null` I think, we do this pattern in a few places

lib/c-api/src/wasm_c_api/macros.rs
```diff
@@ -171,33 +285,75 @@ macro_rules! wasm_declare_boxed_vec {
...
+
                 if !vec.data.is_null() {
                     let data: Vec<*mut [<wasm_ $name _t>]> = Vec::from_raw_parts(vec.data, vec.size, vec.size);
-                    let _data: Vec<Box<[<wasm_ $name _t>]>> = ::std::mem::transmute(data);
+
+                    // If the vector has been initialized (we check
+                    // only the first item), we can transmute items to
+                    // `Box`es.
+                    if vec.size > 0 && !data[0].is_null() {
+                        let _data: Vec<Box<[<wasm_ $name _t>]>> = ::std::mem::transmute(data);
```
MarkMcCaskey, line 354: I don't feel great about transmuting a `repr(Rust)` type, but I suppose we were already doing that 🤔 .

lib/c-api/src/wasm_c_api/macros.rs
MarkMcCaskey: For handling uninitialized, I think we just have to make a representation of `uninit` that makes sense... I think the code as-is will break in real code (for example partially initialized vectors.  I'll get back to you with more details after finishing the review and looking at the code locally

MarkMcCaskey:
> Too bad, it's working on `aarch64-apple-darwin`, but not on `x86_64`…

How do I reproduce the failure? Locally `make test-capi-cranelift-jit` is passing on `x86_64` macos

lib/c-api/src/wasm_c_api/macros.rs
Hywan: Good idea!

lib/c-api/src/wasm_c_api/macros.rs
Hywan: Removed with `Option<&mut T>` in https://github.com/wasmerio/wasmer/pull/1949/commits/51fe219efbb260193753441a0a42c704ea9cfd2d.

lib/c-api/src/wasm_c_api/macros.rs
Hywan: It was indeed present in the code before. I've just added the `vec.size > 0 && …` condition.

Hywan:
To handle partially uninitialized boxed vector, I propose to transmute the `Vec<*mut T>` to `Vec<Option<Box<T>>>` so that it handles safely all items individually. The corresponding commit is https://github.com/wasmerio/wasmer/pull/1949/commits/a68a1e67f414bcbe9f8c4bb1187bef2dacd79764.

Hywan:
bors try

lib/c-api/src/wasm_c_api/types/function.rs
```diff
@@ -111,8 +73,11 @@ pub unsafe extern "C" fn wasm_functype_new(
         .map(|val| val.as_ref().into())
         .collect::<Vec<_>>();
 
-    wasm_valtype_vec_delete(Box::into_raw(params));
-    wasm_valtype_vec_delete(Box::into_raw(results));
+    wasm_valtype_vec_delete(Some(params.as_mut()));
+    wasm_valtype_vec_delete(Some(results.as_mut()));
+
+    mem::forget(params);
+    mem::forget(results);
```
MarkMcCaskey, line 80: Ideally, we'd have `Drop` implemented and it'd do this for us

MarkMcCaskey (approved):
I have some other small comments but it overlaps with #1944, and I can change them there!

Issue #1931, opened 2020-12-14
Allow users to handle metering errors


webmaster128:
### Motivation

It would be good to be able to differentiate _out of gas_ `Metering` errors from other runtime errors. Right now you get

```
RuntimeError { source: Trap(UnreachableCodeReached), wasm_trace: [FrameInfo { module_name: "<module>", func_index: 0, function_name: Some("add_one_f"), func_start: SourceLoc(37), instr: SourceLoc(44) }], native_trace:    0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: <unknown>
  12: <unknown>
 }
```

when adding this error debugger to the metering example:

```diff
diff --git a/examples/metering.rs b/examples/metering.rs
index ed588dda3..ea6a9e556 100644
--- a/examples/metering.rs
+++ b/examples/metering.rs
@@ -131,8 +131,11 @@ fn main() -> anyhow::Result<()> {
                 result
             );
         }
-        Err(_) => {
-            println!("Calling `add_one` failed: not enough gas points remaining.");
+        Err(err) => {
+            println!(
+                "Calling `add_one` failed: not enough gas points remaining: {:?}",
+                err
+            );
         }
     }
```

So it does not seems to be possible to differentiate point depletion from other traps and panics.

### Proposed solution

Unfortunately I don't know how to implement.

### Alternatives

?

### Additional context

I'm trying to re-enable metering after upgrading from Wasmer 0.17 to master.

nlewycky:
The middleware can only do what normal wasm can do, which includes the same kinds of trap that non-metering wasm could contain. Can you tell it's a out-of-gas by querying `Metering::get_remaining_points()`? Is that good enough?

webmaster128:
> Can you tell it's a out-of-gas by querying `Metering::get_remaining_points()`? Is that good enough?

Maybe, can try. For this to work all remaining points must be set to 0 in case of a gas depletion (which is probably desired anyways since you don't want guests to double spend their points). The [current behaviour is different](https://github.com/wasmerio/wasmer/blob/22c7bc8e19de54090f3f0a2197ad62c4a5b72c95/examples/metering.rs#L139-L142).

If runtime error + zero points left is interpreted as _out of gas_, you cannot differentiate traps that fully consume all available gas and not more.

[...]

PR #1951, opened 2020-12-17
feat: Add an error code on the metering middleware

jubianchi:
_This is a proposition for #1931._

~~**⚠️ Do not merge as-is. Wait for #1941 to get merged or to provide an alternative**~~

Basically, the metering middleware now injects a new global, `error_code` with its own helper `get_error_code` which will allow the host to correctly handle metering errors. Now, a `remaining_points == 0` and `error_code != NoError` means the execution of the guest module consumed all the gas points.

# Review

### DIFF
--- a/examples/metering.rs
+++ b/examples/metering.rs
@@ -21,7 +21,10 @@ use wasmer::CompilerConfig;
 use wasmer::{imports, wat2wasm, Instance, Module, Store};
 use wasmer_compiler_cranelift::Cranelift;
 use wasmer_engine_jit::JIT;
-use wasmer_middlewares::metering::{get_remaining_points, set_remaining_points, Metering};
+use wasmer_middlewares::{
+    metering::{get_remaining_points, set_remaining_points, MeteringPoints},
+    Metering,
+};
 
 fn main() -> anyhow::Result<()> {
     // Let's declare the Wasm module.
@@ -55,11 +58,10 @@ fn main() -> anyhow::Result<()> {
 
     // Now let's create our metering middleware.
     //
-    // `Metering` needs to be configured with a limit (the gas limit) and
-    // a cost function.
+    // `Metering` needs to be configured with a limit and a cost function.
     //
     // For each `Operator`, the metering middleware will call the cost
-    // function and subtract the cost from the gas.
+    // function and subtract the cost from the remaining points.
     let metering = Arc::new(Metering::new(10, cost_function));
     let mut compiler_config = Cranelift::default();
     compiler_config.push_middleware(metering);
@@ -93,14 +95,17 @@ fn main() -> anyhow::Result<()> {
     println!("Calling `add_one` function once...");
     add_one.call(1)?;
 
-    // As you can see here, after the first call we have 6 remaining gas points.
+    // As you can see here, after the first call we have 6 remaining points.
     //
     // This is correct, here are the details of how it has been computed:
     // * `local.get $value` is a `Operator::LocalGet` which costs 1 point;
     // * `i32.const` is a `Operator::I32Const` which costs 1 point;
     // * `i32.add` is a `Operator::I32Add` which costs 2 points.
     let remaining_points_after_first_call = get_remaining_points(&instance);
-    assert_eq!(remaining_points_after_first_call, 6);
+    assert_eq!(
+        remaining_points_after_first_call,
+        MeteringPoints::Remaining(6)
+    );
 
     println!(
         "Remaining points after the first call: {:?}",
@@ -110,18 +115,21 @@ fn main() -> anyhow::Result<()> {
     println!("Calling `add_one` function twice...");
     add_one.call(1)?;
 
-    // We spent 4 more gas points with the second call.
+    // We spent 4 more points with the second call.
     // We have 2 remaining points.
     let remaining_points_after_second_call = get_remaining_points(&instance);
-    assert_eq!(remaining_points_after_second_call, 2);
+    assert_eq!(
+        remaining_points_after_second_call,
+        MeteringPoints::Remaining(2)
+    );
 
     println!(
         "Remaining points after the second call: {:?}",
         remaining_points_after_second_call
     );
 
-    // Because calling our `add_one` function consumes 4 gas points,
-    // calling it a third time will fail: we already consume 8 gas
+    // Because calling our `add_one` function consumes 4 points,
+    // calling it a third time will fail: we already consume 8
     // points, there are only two remaining.
     println!("Calling `add_one` function a third time...");
     match add_one.call(1) {
@@ -132,27 +140,27 @@ fn main() -> anyhow::Result<()> {
             );
         }
         Err(_) => {
-            println!("Calling `add_one` failed: not enough gas points remaining.");
+            println!("Calling `add_one` failed.");
+
+            // Because the last needed more than the remaining points, we should have an error.
+            let remaining_points = get_remaining_points(&instance);
+
+            match remaining_points {
+                MeteringPoints::Remaining(..) => {
+                    bail!("No metering error: there are remaining points")
+                }
+                MeteringPoints::Exhausted => println!("Not enough points remaining"),
+            }
         }
     }
 
-    // Becasue the previous call failed, it did not consume any gas point.
-    // We still have 2 remaining points.
-    let remaining_points_after_third_call = get_remaining_points(&instance);
-    assert_eq!(remaining_points_after_third_call, 2);
-
-    println!(
-        "Remaining points after third call: {:?}",
-        remaining_points_after_third_call
-    );
-
     // Now let's see how we can set a new limit...
-    println!("Set new remaining points points to 10");
+    println!("Set new remaining points to 10");
     let new_limit = 10;
     set_remaining_points(&instance, new_limit);
 
     let remaining_points = get_remaining_points(&instance);
-    assert_eq!(remaining_points, new_limit);
+    assert_eq!(remaining_points, MeteringPoints::Remaining(new_limit));
 
     println!("Remaining points: {:?}", remaining_points);
 
--- a/lib/middlewares/src/metering.rs
+++ b/lib/middlewares/src/metering.rs
@@ -12,6 +12,34 @@ use wasmer::{
 use wasmer_types::GlobalIndex;
 use wasmer_vm::ModuleInfo;
 
+#[derive(Clone)]
+struct MeteringGlobalIndexes(GlobalIndex, GlobalIndex);
+
+impl MeteringGlobalIndexes {
+    /// The global index in the current module for remaining points.
+    fn remaining_points(&self) -> GlobalIndex {
+        self.0
+    }
+
+    /// The global index in the current module for a boolean indicating whether points are exhausted
+    /// or not.
+    /// This boolean is represented as a i32 global:
+    ///   * 0: there are remaining points
+    ///   * 1: points have been exhausted
+    fn points_exhausted(&self) -> GlobalIndex {
+        self.1
+    }
+}
+
+impl fmt::Debug for MeteringGlobalIndexes {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("MeteringGlobalIndexes")
+            .field("remaining_points", &self.remaining_points())
+            .field("points_exhausted", &self.points_exhausted())
+            .finish()
+    }
+}
+
 /// The module-level metering middleware.
 ///
 /// # Panic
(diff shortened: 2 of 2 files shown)


jubianchi:
ping @webmaster128

lib/middlewares/src/metering.rs
```diff
@@ -39,43 +43,51 @@ pub struct FunctionMetering<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync
...
+                    _ => Err(()),
+                }
+            },
+            Value::I64(value ) => {
+                match value  {
+                    value  if value  == MeteringError::NoError as i64 => Ok(MeteringError::NoError),
+                    value  if value  == MeteringError::OutOfGas as i64 => Ok(MeteringError::OutOfGas),
+                    _ => Err(()),
+                }
+            }
```
webmaster128, line 75: Why support two different Wasm `Value` types for this conversion? We know in which type the error code is stored, right?

lib/middlewares/src/metering.rs
```diff
@@ -186,3 +220,150 @@ impl<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync> FunctionMiddleware
...
+
+        // Second call
+        add_one.call(1).unwrap();
+        assert_eq!(get_remaining_points(&instance), 2);
+
+        // Third call fails due to limit
+        assert!(add_one.call(1).is_err());
+        // TODO: what do we expect now? 0 or 2? See https://github.com/wasmerio/wasmer/issues/1931
+        // assert_eq!(metering.get_remaining_points(&instance), 2);
+        // assert_eq!(metering.get_remaining_points(&instance), 0);
```
webmaster128, line 337: Could you enable this check, now that the behaviour is defined to set remaing gas to 0.

webmaster128 (commented):
Looks good, some first thoughts

lib/middlewares/src/metering.rs
jubianchi: indeed, TBH I don't know why I covered both i32 and i64 🤔

lib/middlewares/src/metering.rs
```diff
@@ -75,29 +111,45 @@ impl<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync + 'static> ModuleMiddl
...
             .push(GlobalInit::I64Const(self.initial_limit as i64));
 
         module_info.exports.insert(
             "remaining_points".to_string(),
-            ExportIndex::Global(global_index),
+            ExportIndex::Global(remaining_points_global_index),
+        );
+
+        module_info.exports.insert(
+            "error_code".to_string(),
```
MarkMcCaskey, line 151: Small nit, if these global export names can conflict with user global variables, lets use something less likely to conflict, like prefixing them with `wasmer_metering_`

lib/middlewares/src/metering.rs
```diff
@@ -183,6 +240,16 @@ pub fn get_remaining_points(instance: &Instance) -> u64 {
...
 }
 
+pub fn get_error(instance: &Instance) -> MeteringError {
+    instance
+        .exports
+        .get_global("error_code")
+        .expect("Can't get `error_code` from Instance")
+        .get()
+        .try_into()
+        .expect("`error_code` from Instance has wrong type")
```
MarkMcCaskey, line 250: Something like "(metering) `error_code` has an invalid value" is a bit more clear

lib/middlewares/src/metering.rs
```diff
@@ -183,6 +240,16 @@ pub fn get_remaining_points(instance: &Instance) -> u64 {
         .expect("`remaining_points` from Instance has wrong type")
 }
 
+pub fn get_error(instance: &Instance) -> MeteringError {
+    instance
+        .exports
+        .get_global("error_code")
+        .expect("Can't get `error_code` from Instance")
```
MarkMcCaskey, line 247: Same here, this should mention the context (metering). Not necessarily if the global variable name is changed to include the word "metering"

MarkMcCaskey (commented):
The structure and design looks reasonable to me! I have some feedback about some of the implementation details though!

lib/middlewares/src/metering.rs
```diff
@@ -36,53 +41,63 @@ pub struct FunctionMetering<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync
...
     remaining_points_index: GlobalIndex,
 
     /// Accumulated cost of the current basic block.
     accumulated_cost: u64,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum MeteringError {
+    NoError,
+    OutOfGas,
```
webmaster128, line 57: The term "gas" is not used elsewhere in the implementation as far as I know. I'd go with something like

```suggestion
    OutOfPoints,
```

lib/middlewares/src/metering.rs
```diff
@@ -36,53 +41,63 @@ pub struct FunctionMetering<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync
...
+}
+
+impl<T> TryFrom<Value<T>> for MeteringError {
+    type Error = ();
+
+    fn try_from(v: Value<T>) -> Result<Self, Self::Error> {
+        match v {
+            Value::I32(value) => value.try_into(),
+            _ => Err(()),
+        }
```
webmaster128, line 79: We could also convert to `u32` because `impl<T> TryFrom<Value<T>> for u32` already exists. Then we cam make the `MeteringError` enum `#[repr(u32)]` and avoid a lot of casting

lib/middlewares/src/metering.rs
```diff
@@ -36,53 +41,63 @@ pub struct FunctionMetering<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync
...
 
     /// Accumulated cost of the current basic block.
     accumulated_cost: u64,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum MeteringError {
+    NoError,
+    OutOfGas,
+}
```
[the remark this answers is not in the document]
jubianchi: Actually this is what I was asking to the team: I can't think of any other use case so there are two alternatives here:
* remove the `enum` and only use a boolean;
* leave the enum as is and apply your suggestion.

webmaster128:
This PR covers all functionality required for #1931 – very cool. Freel free to reject my suggestions and personal preferences if you don't like them.

One more thing that I can think of (but don't need on my own), is the ability to reset the error. Imagine a use case where Wasm runs for a while and hits a safety limit. Now the user has the choice to resume the execution. Then it would be nice if you could increase the limit and reset the error.

lib/middlewares/src/metering.rs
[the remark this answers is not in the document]
webmaster128: The third option is to keep the 2 value enum, which adds some nice naming to the two possible states. Even with no `#[non_exhaustive]`, the enum is more future proof as adding enum cases will be breaking but in a less disruptive way.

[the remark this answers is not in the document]
webmaster128: I don't have a strong preference here. All 3 seem reasonable to me.

jubianchi:
@webmaster128 @syrusakbary what do you think of reseting the error each time we call `set_remaining_points`?

webmaster128:
> @webmaster128 @syrusakbary what do you think of reseting the error each time we call `set_remaining_points`?

I think there is two layers of abstraction here:
1. getters/setters for the metering state variables
2. a higher level concept of metering points that can be empty or exceeded

In 1., the values are independent it it would be very confusing if they affect each other when setting them. However, we can think about exposing 2. to user, where
- `set_remaining_points` resets the error
- `get_remaining_points` returns an
   ```rust
   enum MeteringPoints {
       /// The given amount of points >= 0 is left 
       Remaining(u64),
       /// The limit was exceeded
       Depleted,
   }
   ```
This would remove the impossible state _error = 1 and points > 0_ from the user's handling code.

jubianchi:
@webmaster128 thanks for the suggestion.

I like the second proposition. @syrusakbary @nlewycky what do you think about that?

lib/middlewares/src/metering.rs
```diff
@@ -173,14 +211,28 @@ impl<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync> FunctionMiddleware
...
+    let error: u64 = instance
+        .exports
+        .get_global("wasmer_metering_missing_points")
+        .expect("Can't get `wasmer_metering_missing_points` from Instance")
+        .get()
+        .try_into()
+        .expect("`wasmer_metering_missing_points` from Instance has wrong type");
+
+    if error > 0 {
+        return MeteringPoints::Depleted(points, error);
```
webmaster128, line 232: With this new higher level interface I don't think we need the error code anymore.
I think the `points` are not interesting in the `Depleted` case as well.

lib/middlewares/src/metering.rs
```diff
@@ -173,14 +211,28 @@ impl<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync> FunctionMiddleware
 ///
 /// The instance Module must have been processed with the [`Metering`] middleware
 /// at compile time, otherwise this will panic.
-pub fn get_remaining_points(instance: &Instance) -> u64 {
-    instance
+pub fn get_remaining_points(instance: &Instance) -> MeteringPoints {
+    let points = instance
```
webmaster128, line 215: I'd only access that value when there is no depletion error

examples/metering.rs
```diff
@@ -132,27 +140,31 @@ fn main() -> anyhow::Result<()> {
...
+            let remaining_points = get_remaining_points(&instance);
+
+            match remaining_points {
+                MeteringPoints::Remaining(..) => {
+                    bail!("No metering error: there are remaining points")
+                }
+                MeteringPoints::Depleted(remaining, missing) => println!(
+                    "Not enough points remaining: {:?} points remaining but need {:?}",
+                    remaining,
+                    remaining + missing
```
webmaster128, line 155: I'd avoid exposing `missing` to the user, as this is not what is missing for the completion of the execution but just what is missing for the next step. I don't think it is very useful.

webmaster128 (commented):
Cool API

lib/middlewares/src/metering.rs
```diff
@@ -37,19 +40,29 @@ pub struct FunctionMetering<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync
...
     remaining_points_index: GlobalIndex,
 
     /// Accumulated cost of the current basic block.
     accumulated_cost: u64,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum MeteringPoints {
+    Remaining(u64),
+    Depleted(u64, u64),
```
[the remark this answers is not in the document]
jubianchi: Not required as this will disappear ;)

lib/middlewares/src/metering.rs
```diff
@@ -173,14 +183,20 @@ impl<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync> FunctionMiddleware
...
-        .get_global("remaining_points")
-        .expect("Can't get `remaining_points` from Instance")
+        .get_global("wasmer_metering_remaining_points")
+        .expect("Can't get `wasmer_metering_remaining_points` from Instance")
         .get()
         .try_into()
-        .expect("`remaining_points` from Instance has wrong type")
+        .expect("`wasmer_metering_remaining_points` from Instance has wrong type");
+
+    if points < 0 {
```
webmaster128, line 195: With this single i64/u64 value, any u64 value geater than i64::MAX is interpreted as a negative value and error. E.g.

```rust
set_remaining_points(&instance, u64::MAX);
get_remaining_points(&instance) // depletion
```

I'd rather stick to two variables than limiting the values the user can use

jubianchi:
bors try

lib/middlewares/src/metering.rs
```diff
@@ -173,14 +218,28 @@ impl<F: Fn(&Operator) -> u64 + Copy + Clone + Send + Sync> FunctionMiddleware
...
+        .exports
+        .get_global("wasmer_metering_points_exhausted")
+        .expect("Can't get `wasmer_metering_points_exhausted` from Instance")
+        .get()
+        .try_into()
+        .expect("`wasmer_metering_points_exhausted` from Instance has wrong type");
+
+    if exhausted > 0 {
+        return MeteringPoints::Exhausted;
+    }
```
webmaster128, line 232: Beautiful! Now the value of `wasmer_metering_remaining_points` is completely irrelevant to the user in the error case. This implementation detail is never exposed via the API.

lib/middlewares/src/metering.rs
```diff
@@ -195,10 +254,17 @@ pub fn get_remaining_points(instance: &Instance) -> u64 {
...
+        .expect("Can't get `wasmer_metering_remaining_points` from Instance")
         .set(points.into())
-        .expect("Can't set `remaining_points` in Instance");
+        .expect("Can't set `wasmer_metering_remaining_points` in Instance");
+
+    instance
+        .exports
+        .get_global("wasmer_metering_points_exhausted")
+        .expect("Can't get `wasmer_metering_points_exhausted` from Instance")
+        .set(0.into())
```
webmaster128, line 266: This seems to be the default type, but we could make it a little more review-frindly.

```suggestion
        .set(0i32.into())
```

webmaster128 (commented):
Great work. Some ideas, feel free to dismiss. The only thing we should clean up one way or another is the exports.

MarkMcCaskey (approved):

PR #1962, opened 2020-12-21
Reduce scope of wasmparser::BinaryReaderError in the codebase

### LINKED ISSUE #1950: Error type of FunctionMiddleware::feed is private
### Describe the bug

The error type of the `FunctionMiddleware::feed` is `BinaryReaderError`:

```rust
use wasmparser::{BinaryReader, Operator, Result as WpResult, Type};

// …

/// A function middleware specialized for a single function.
pub trait FunctionMiddleware: Debug {
    /// Processes the given operator.
    fn feed<'a>(
        &mut self,
        operator: Operator<'a>,
        state: &mut MiddlewareReaderState<'a>,
    ) -> WpResult<()> {
        state.push_operator(operator);
        Ok(())
    }
}
```

and

```rust
pub type Result<T, E = BinaryReaderError> = result::Result<T, E>;
```

However, `BinaryReaderError` is a type in the `wasmparser` crate with no public constructor. So if I understand correctly, a user of Wasmer cannot implement a `FunctionMiddleware` that errors.

### Steps to reproduce

–

### Expected behavior

The error type is either constructible or can be defined by the user ([as in Wasmer 0.17](https://github.com/wasmerio/wasmer/blob/0.17.1/lib/runtime-core/src/codegen.rs#L430-L431)). 

### Actual behavior

The error type is fixed to `wasmparser::BinaryReaderError`, which is not constructible.

### Additional context

I'm trying rewrite a the `Deterministic` middleware, orginally created here: https://github.com/spacemeshos/svm/blob/5df80288c8b9a5ab3665297251c283cb614ebb81/crates/svm-compiler/src/middleware/validation.rs.

webmaster128:
~Based on #1963~

Closes #1950

# Description

There is an error conversion pileline `BinaryReaderError` -> `WasmError`. In this PR, the same conversion happens earlier, such that `WasmError`/`WasmResult` can be used in middlewares.

Questions:
- [ ] Should middlewares be allowed to produce all `WasmError` cases? Or should we introduce a `WasmError::MiddlewareError(String)`?
- [x] ~Should `to_wasm_error` be removed as part of this PR?~ Extracted to https://github.com/wasmerio/wasmer/pull/1963

# Review

### DIFF
--- a/lib/compiler-singlepass/src/compiler.rs
+++ b/lib/compiler-singlepass/src/compiler.rs
@@ -9,7 +9,6 @@ use crate::codegen_x64::{
 use crate::config::Singlepass;
 use rayon::prelude::{IntoParallelIterator, IntoParallelRefIterator, ParallelIterator};
 use std::sync::Arc;
-use wasmer_compiler::wasmparser::BinaryReaderError;
 use wasmer_compiler::TrapInformation;
 use wasmer_compiler::{
     Architecture, CompileModuleInfo, CompilerConfig, MiddlewareBinaryReader, ModuleMiddlewareChain,
@@ -92,9 +91,9 @@ impl Compiler for SinglepassCompiler {
 
                 // This local list excludes arguments.
                 let mut locals = vec![];
-                let num_locals = reader.read_local_count().map_err(to_compile_error)?;
+                let num_locals = reader.read_local_count()?;
                 for _ in 0..num_locals {
-                    let (count, ty) = reader.read_local_decl().map_err(to_compile_error)?;
+                    let (count, ty) = reader.read_local_decl()?;
                     for _ in 0..count {
                         locals.push(ty);
                     }
@@ -113,7 +112,7 @@ impl Compiler for SinglepassCompiler {
 
                 while generator.has_control_frames() {
                     generator.set_srcloc(reader.original_position() as u32);
-                    let op = reader.read_operator().map_err(to_compile_error)?;
+                    let op = reader.read_operator()?;
                     generator.feed_operator(op).map_err(to_compile_error)?;
                 }
 
@@ -155,12 +154,6 @@ impl Compiler for SinglepassCompiler {
 
 trait ToCompileError {
     fn to_compile_error(self) -> CompileError;
-}
-
-impl ToCompileError for BinaryReaderError {
-    fn to_compile_error(self) -> CompileError {
-        CompileError::Codegen(self.message().into())
-    }
 }
 
 impl ToCompileError for CodegenError {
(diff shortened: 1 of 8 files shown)


syrusakbary (approved):

webmaster128:
Thank you for the feedback @syrusakbary. I'm having compile issues right now. I'll let you know when I have a demo available showing the integration of this PR in a project.

lib/compiler-singlepass/src/compiler.rs
```diff
@@ -113,7 +112,7 @@ impl Compiler for SinglepassCompiler {
 
                 while generator.has_control_frames() {
                     generator.set_srcloc(reader.original_position() as u32);
-                    let op = reader.read_operator().map_err(to_compile_error)?;
+                    let op = reader.read_operator()?;
```
webmaster128, line 115: Before this PR, `BinaryReaderError` was converted to `CompileError::Codegen` at this place. Not it is converted to `CompileError::Wasm`. Is this change correct?

webmaster128:
Alright, I'm happy now. Demo here: https://github.com/CosmWasm/cosmwasm/pull/675/files.
Just the one question: https://github.com/wasmerio/wasmer/pull/1962#discussion_r547531097

MarkMcCaskey (approved):
Looks good to me

lib/compiler-singlepass/src/compiler.rs
MarkMcCaskey: Hmm, it's a fuzzy line for sure.  Perhaps the idea for calling it "codegen" was that we only read more operators here based on state of the code generator... whereas calling it a "wasm" error suggests that if the invariant here isn't met, then it's because of a problem with the Wasm...

I would defer to the types of errors that this function (and functions like it) are returning... if they're mostly returning codegen errors, then I'd expect that to be what we want here, likewise with Wasm

webmaster128: Alright, `reader.read_operator()` now returns `WasmError`, so as long as you don't have a better idea, I leave it to the default type converter.

One thought I had about this code section: Maybe `to_compile_error` should be renamed to `to_codegen_error` if it should always produces `CompileError::Codegen`, independent of the source type.

MarkMcCaskey: Hmm, it looks like `to_compile_error` comes from the trait / method, so it probably makes sense to keep it is as used if that's a widely used abstraction here -- I'm not super familiar with this part of wasmer's code

webmaster128:
build fixed

PR #1979, opened 2021-01-04
Make `WasmPtr::get_utf8_string` return `String`

MarkMcCaskey:
The old functionality is readded as an `unsafe` method,
`WasmPtr::get_utf8_str`.

~It's important to note that `WasmPtr::get_utf8_string` isn't _entirely_ sound and technically should be marked `unsafe` as well, but this change is a massive improvement over what we had before.~

~For future reference, the reason `WasmPtr::get_utf8_string` still has some soundness issues is that we can't guarantee exclusive access to the memory while parsing in the string, we temporarily hold a `&[u8]` and hope it doesn't get mutated.~ 

~It's possible to implement this method in a more correct way by copying each byte as we read it into a `Vec` and converting that into a String.~

EDIT: 
fixed in https://github.com/wasmerio/wasmer/pull/1979/commits/b592ed83afb9742853cc6d9458fa28457c0290a4; `WasmPtr::get_utf8_string` is sound and uses no `unsafe` internally now.

# Review

### DIFF
--- a/lib/api/src/ptr.rs
+++ b/lib/api/src/ptr.rs
@@ -188,7 +188,15 @@ impl<T: Copy + ValueType> WasmPtr<T, Array> {
     /// Note that this method returns a reference to Wasm linear memory. The
     /// underlying data can be mutated if the Wasm is allowed to execute or
     /// an aliasing `WasmPtr` is used to mutate memory.
-    pub fn get_utf8_string(self, memory: &Memory, str_len: u32) -> Option<&str> {
+    ///
+    /// # Safety
+    /// `str` has invariants that must not be broken by mutating Wasm memory.
+    /// Thus the caller must ensure exclusive access to memory or otherwise ensure
+    /// that the backing memory is not modified while the reference is held.
+    ///
+    /// Additionally, if `memory` is dynamic, the caller must also ensure that `memory`
+    /// is not grown while the reference is held.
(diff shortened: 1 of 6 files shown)


syrusakbary (approved):

lib/api/src/ptr.rs
```diff
@@ -185,36 +185,81 @@ impl<T: Copy + ValueType> WasmPtr<T, Array> {
...
+    /// data can be mutated if the Wasm is allowed to execute or an aliasing
+    /// `WasmPtr` is used to mutate memory.
+    ///
+    /// `str` has invariants that must not be broken by mutating Wasm memory.
+    /// Thus the caller must ensure that the backing memory is not modified
+    /// while the reference is held.
+    ///
+    /// Additionally, if `memory` is dynamic, the caller must also ensure that `memory`
+    /// is not grown while the reference is held.
+    pub unsafe fn get_utf8_str<'a>(self, memory: &'a Memory, str_len: u32) -> Option<&'a str> {
```
syrusakbary, line 203: Can we mark this function as deprecated? @MarkMcCaskey

MarkMcCaskey: We could, but I don't know if we want to! It's dangerous to use, but it is more efficient as it doesn't do any memory allocations, so it's nice to have, it just needs to be used very carefully.

syrusakbary: Ok, let's merge then!

Issue #1943, opened 2020-12-16
Reusing an Artifact (JIT) in multiple instances leads to crashes


webmaster128:
### Summary

For caching purposes, I'd like to store and reuse artifact for multiple instances. An artifact from cache is combined with a new `Store` to make a `Module` which is then instantiated. This works fine if I cache serialized versions of the artifact. However, it would be very nice to avoid this serialization/deserialization for in-memory artifact caching.

With the help of [this API adjustment](https://github.com/wasmerio/wasmer/compare/master...webmaster128:module-from-artifact), I implemented a cache for `Arc<dyn Artifact>`: https://github.com/CosmWasm/cosmwasm/pull/666. This works fine until the test `use_multiple_cached_instances_of_same_contract` somes in, which creates and executes multiple instances of the same artifact. Then it crashes with a `(signal: 11, SIGSEGV: invalid memory reference)` on Linux (CI and locally) and errors with `RuntimeError: out of bounds memory access` on a the Mac.

Now the real question: is instantiating an Artifact multiple times supported? Is this a bug or am I missing something here?

### Additional details

–

syrusakbary:
> Now the real question: is instantiating an Artifact multiple times supported? Is this a bug or am I missing something here?

It should be theoretically possible to instantiate an artifact multiple times. However, the `Artifact` struct is only intended to be used by the engines.
Your API might have uncovered other bug, we'll take a look into it. Thanks for reporting it!

webmaster128:
> However, the `Artifact` struct is only intended to be used by the engines.

I only operated on `Arc<dyn Artifact>`s, i.e. the trait. The only two methods I used are `module.artifact()` and `Module::from_artifact`, so I did not try to do anything super-internal.

webmaster128:
I dug into this a little bit more. After refining the tests on my side, it turned out the issue also comes up with a single instance after caching. It happens **when a module is instantiated from an artifact after the original store dropped**. In https://github.com/wasmerio/wasmer/pull/1974 a minimal test case is shown to reproduce the issue. As far as I can tell the behaviour is the same for JIT and Native.

Hywan:
I've dug into this.

A `Module` compiles a `dyn Artifact`. The artifact concept belongs to the engine. For example, for the JIT engine, we have `JITArtifact`.

An artifact is responsible to construct an `Instance`. Thus, it is possible to build several `Instance`s from the same `dyn Artifact`.

[...]

PR #1983, opened 2021-01-05
feat(api) Remove `Module::artifact`

Hywan:
# Description

This patch removes the `Module::artifact` method since it is too
dangerous (see #1943 for more
details). This method has initially been introduced as a hack.

This patch replaces `Module::artifact` by `Module::query_artifact`
which is another hack, but less dangerous. The artifact can no longer
be extracted from the `Module`; it is only possible to query it to
fetch some information.

cc @MarkMcCaskey @webmaster128 

# Review

- [ ] ~Add a short description of the the change to the CHANGELOG.md file~ not necessary since this was a hidden feature

### DIFF
--- a/lib/api/src/module.rs
+++ b/lib/api/src/module.rs
@@ -251,6 +251,9 @@ impl Module {
         Ok(Self::from_artifact(store, artifact))
     }
 
+    /// Low-level constructor. It must be used to build a unique
+    /// `Module`. It must not be used to build back a `Module` from an
+    /// already compiled `Artifact`.
     fn from_artifact(store: &Store, artifact: Arc<dyn Artifact>) -> Self {
         Self {
             store: store.clone(),
@@ -411,14 +414,40 @@ impl Module {
         &self.artifact.module_ref()
     }
 
-    /// Gets the [`Artifact`] used internally by the Module.
+    /// Query the artifact to fetch information from it without
+    /// extracting the artifact from the module.
+    ///
+    /// The artifact defined in the `Module` is abstracted behind `dyn
+    /// Artifact`. When querying, downcasting to a concrete artifact
+    /// type is done by inferring the type from the closure used to
+    /// query. Something like this:
+    ///
+    /// ```rust,no_run
+    /// use wasmer_engine_object_file::ObjectFileArtifact;
+    ///
+    /// let symbol_registry = unsafe {
+    ///     module.query_artifact(|artifact: &ObjectFileArtifact| {
+    ///         Some(Arc::clone(artifact.symbol_registry()))
+    ///     })
+    /// };
+    /// ```
+    ///
+    /// If downcasting the artifact to `A` fails, the method will
+    /// return `None`.
     ///
     /// This API is hidden because it's not necessarily stable;
     /// this functionality is required for some core functionality though, like
-    /// the object file engine.
+    /// the [`wasmer-engine-object-file`].
     #[doc(hidden)]
-    pub fn artifact(&self) -> &Arc<dyn Artifact> {
-        &self.artifact
+    pub unsafe fn query_artifact<A, Q, R>(&self, query: Q) -> Option<Arc<R>>
+    where
+        A: Artifact + 'static,
+        Q: FnOnce(&A) -> Option<Arc<R>>,
+        R: ?Sized,
+    {
+        let artifact_ref: &A = self.artifact.downcast_ref()?;
+
+        query(artifact_ref)
     }
 }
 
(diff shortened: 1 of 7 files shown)


lib/api/src/module.rs
```diff
@@ -411,14 +414,40 @@ impl Module {
...
     /// this functionality is required for some core functionality though, like
-    /// the object file engine.
+    /// the [`wasmer-engine-object-file`].
     #[doc(hidden)]
-    pub fn artifact(&self) -> &Arc<dyn Artifact> {
-        &self.artifact
+    pub unsafe fn query_artifact<A, Q, R>(&self, query: Q) -> Option<Arc<R>>
+    where
+        A: Artifact + 'static,
+        Q: FnOnce(&A) -> Option<Arc<R>>,
```
MarkMcCaskey, line 445: Returning something always wrapped in an `Arc` seems perhaps overly restrictive... I think we could probably just make `R` also `Cloned`, or does that not work?

lib/api/src/module.rs
```diff
@@ -411,14 +414,40 @@ impl Module {
...
+    /// return `None`.
     ///
     /// This API is hidden because it's not necessarily stable;
     /// this functionality is required for some core functionality though, like
-    /// the object file engine.
+    /// the [`wasmer-engine-object-file`].
     #[doc(hidden)]
-    pub fn artifact(&self) -> &Arc<dyn Artifact> {
-        &self.artifact
+    pub unsafe fn query_artifact<A, Q, R>(&self, query: Q) -> Option<Arc<R>>
```
MarkMcCaskey, line 442: `unsafe` functions must have a `# Safety` section explaining how the function must be used to avoid UB, etc.

lib/cli/src/commands/compile.rs
```diff
@@ -105,15 +106,26 @@ impl Compile {
...
+            use wasmer_engine_object_file::ObjectFileArtifact;
+
+            let symbol_registry = unsafe {
+                module.query_artifact(|artifact: &ObjectFileArtifact| {
+                    Some(Arc::clone(artifact.symbol_registry()))
+                })
+            };
+            let metadata_length = unsafe {
+                module.query_artifact(|artifact: &ObjectFileArtifact| {
+                    Some(Arc::new(artifact.metadata_length()))
```
MarkMcCaskey, line 118: For example, the use of `Arc` here seems unnecessary

MarkMcCaskey: We may also be able to do this with 1 unsafe block, returning a tuple of both things we care about

lib/engine-jit/src/artifact.rs
```diff
@@ -48,14 +48,14 @@ impl JITArtifact {
     #[cfg(feature = "compiler")]
     pub fn new(
         jit: &JITEngine,
-        data: &[u8],
+        binary: &[u8],
```
MarkMcCaskey, line 51: It seems this argument is called `data` in all the other engines, we can change it but it should be consistent

`rg -F 'data: &[u8]'` shows the other uses

lib/api/src/module.rs
webmaster128, line 442: Any specific reason to add the `unsafe` here? To me the description is clear without it.

MarkMcCaskey (commented):
Seems like it could be a good improvement, needs a few changes though!

webmaster128 (approved):
LGTM. Let's kick my commit out of the history here as we clearly do not use anything from it.

lib/api/src/module.rs
Hywan: I wanted to go straight to the point. Setting it to `R` means we can't return a reference, that's annoying. We should not want to implement `Clone` for `SymbolRegistry`s for instance. Thoughts?

lib/cli/src/commands/compile.rs
Hywan: Yup, using `Arc` here is… stupid :-/.

lib/api/src/module.rs
MarkMcCaskey: Well, my idea is that `Arc` is clone, so we can still always return `Arc` if we want... maybe that makes the function harder to use if users have to specify what they mean directly as a type argument (though perhaps making this function harder to use is a feature 😉 )

lib/api/src/module.rs
Hywan: It's another trick I use to indicate it's unstable. I hope it catches the user's attention.

syrusakbary: I think we can remove this one altogether. No need to return the artifact at all.

lib/api/src/module.rs
syrusakbary: I think we can just remove this one.

lib/engine-jit/src/artifact.rs
Hywan: Reverted!

Hywan:
Ready to merge?

ioan555:
Yess

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try-

lib/api/src/module.rs
```diff
@@ -251,6 +251,9 @@ impl Module {
         Ok(Self::from_artifact(store, artifact))
     }
 
+    /// Low-level constructor. It must be used to build a unique
+    /// `Module`. It must not be used to build back a `Module` from an
+    /// already compiled `Artifact`.
```
MarkMcCaskey, line 256: Depending on how bad the consequences for misuse are here, we should consider marking this as `unsafe`.

lib/cli/src/commands/compile.rs
```diff
@@ -105,15 +106,26 @@ impl Compile {
...
+                module.query_artifact(|artifact: &ObjectFileArtifact| {
+                    Some(Arc::clone(artifact.symbol_registry()))
+                })
+            };
+            let metadata_length = unsafe {
+                module.query_artifact(|artifact: &ObjectFileArtifact| {
+                    Some(artifact.metadata_length())
+                })
+            };
+
```
MarkMcCaskey, line 121: This can be done in 1 unsafe block or even in 1 query; I have a slight preference for that because `unsafe` should be a marker for failure-prone code but this is not, so reducing the use of keyword `unsafe` by half is a good thing here.

lib/cli/src/commands/create_exe.rs
```diff
@@ -81,17 +83,22 @@ impl CreateExe {
...
-        let metadata_length = artifact.metadata_length();
+        let symbol_registry = unsafe {
+            module.query_artifact(|artifact: &ObjectFileArtifact| {
+                Some(Arc::clone(artifact.symbol_registry()))
+            })
+        };
+        let metadata_length = unsafe {
+            module.query_artifact(|artifact: &ObjectFileArtifact| Some(artifact.metadata_length()))
+        };
+
```
MarkMcCaskey, line 94: Same here

lib/engine-object-file/src/artifact.rs
```diff
@@ -384,12 +384,12 @@ impl ObjectFileArtifact {
...
             metadata_length: 0,
-            symbol_registry,
+            symbol_registry: Arc::new(symbol_registry),
         })
     }
 
-    /// Get the `SymbolRegistry` used to generate the names used in the Artifact.
-    pub fn symbol_registry(&self) -> &dyn SymbolRegistry {
+    /// Get the `SymbolRegistry` used to generate the names used in the artifact.
+    pub fn symbol_registry(&self) -> &Arc<ModuleMetadataSymbolRegistry> {
```
MarkMcCaskey, line 392: Technically this is a breaking change in the public API of `engine-object-file`.

Not sure what best practice is here for our own, essentially internal APIs. It is unfortunate that this is at least theoretically fully usable on crates.io and we'd be breaking it.

lib/engine-object-file/src/serialize.rs
```diff
@@ -13,8 +13,9 @@ pub struct ModuleMetadata {
     pub function_body_lengths: PrimaryMap<LocalFunctionIndex, u64>,
 }
 
+/// Symbol registry of the [`ModuleMetadata`].
 pub struct ModuleMetadataSymbolRegistry {
-    pub prefix: String,
+    pub(crate) prefix: String,
```
MarkMcCaskey, line 18: This as well, would be a breaking change

MarkMcCaskey (approved):

PR #1993, opened 2021-01-07
Handle error case of emscripten + wasi module

MarkMcCaskey:
Need to test this on an emscripten module compiled with WASI imports.

Motivation for this PR is that we get asked a lot about this: our error messages here aren't very clear.

Also included in this PR are numerous small tweaks to get wasmer to compile with non-default cargo features

# Review

### DIFF
--- a/lib/cli/src/commands/compile.rs
+++ b/lib/cli/src/commands/compile.rs
@@ -41,8 +41,8 @@ impl Compile {
     pub(crate) fn get_recommend_extension(
         engine_type: &EngineType,
         target_triple: &Triple,
-    ) -> &'static str {
-        match engine_type {
+    ) -> Result<&'static str> {
+        Ok(match engine_type {
             #[cfg(feature = "native")]
             EngineType::Native => {
                 wasmer_engine_native::NativeArtifact::get_default_extension(target_triple)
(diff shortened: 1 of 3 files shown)


lib/cli/src/commands/compile.rs
```diff
@@ -54,7 +54,7 @@ impl Compile {
                 wasmer_engine_object_file::ObjectFileArtifact::get_default_extension(target_triple)
             }
             #[cfg(not(all(feature = "native", feature = "jit", feature = "object-file")))]
-            _ => bail!("selected engine type is not compiled in"),
+            _ => panic!("selected engine type is not compiled in"),
```
syrusakbary, line 57: Since it’s done on the CLI level, and the error is kind of expected I believe bail is better suited.

MarkMcCaskey: For context, the function does not return `anyhow::Result` so we'd have to update that and all the call locations, which may be fine, I just did the smallest change to get it to compile.

MarkMcCaskey: I agree though that `bail` is better, all things being equal -- I'll see if making it `anyhow::Result` is simple

PR #2003, opened 2021-01-11
chore: Build Wasmer on musl

### LINKED ISSUE #1482: Compatibility with Alpine Linux

### Motivation

Many tiny Docker deployments use Alpine Linux as its base to reduce size of images and containers. Currently, wasmer is incompatible with such environments because it's compiled against glibc.

### Proposed solution

Provide a build of wasmer that can work with musl libc as well.

### Additional context

For background, initially I stumbled upon this due to non-sensical error after installation:

```bash
/ # ls /root/.wasmer/bin/wasmer
/root/.wasmer/bin/wasmer
/ # /root/.wasmer/bin/wasmer
sh: /root/.wasmer/bin/wasmer: not found
```

After searching around, I found explanation of a similar issue as well a workaround that creates a symbolic link for musl libc pretending to be full one: https://stackoverflow.com/questions/34729748/installed-go-binary-not-found-in-path-on-alpine-linux-docker

That helped a bit, but there are still missing functions due to differences in libc implementations:

```bash
/ # mkdir /lib64
/ # ln -s /lib/libc.musl-x86_64.so.1 /lib64/ld-linux-x86-64.so.2
/ # /root/.wasmer/bin/wasmer
Error relocating /root/.wasmer/bin/wasmer: __snprintf_chk: symbol not found
Error relocating /root/.wasmer/bin/wasmer: mallinfo: symbol not found
Error relocating /root/.wasmer/bin/wasmer: __fprintf_chk: symbol not found
Error relocating /root/.wasmer/bin/wasmer: __vsnprintf_chk: symbol not found
Error relocating /root/.wasmer/bin/wasmer: __longjmp_chk: symbol not found
```

Hence, I think it would be best if wasmer would be built against musl as well.

### LINKED ISSUE #1766: alpine build error
### Describe the bug

try build from alpine, get this error:

```sh

  --- stderr
  autoreconf: Entering directory `.'
  autoreconf: configure.ac: not using Gettext
  autoreconf: running: aclocal -I m4
  autoreconf: configure.ac: tracing
  autoreconf: running: libtoolize --copy
  autoreconf: running: /usr/bin/autoconf
  autoreconf: running: /usr/bin/autoheader
  autoreconf: running: automake --add-missing --copy --no-force
  configure.ac:31: installing './compile'
  configure.ac:19: installing './install-sh'
  configure.ac:19: installing './missing'
  Makefile.am: installing './depcomp'
  doc/Makefile.am:3: installing 'doc/mdate-sh'
  doc/Makefile.am:3: installing 'doc/texinfo.tex'
  autoreconf: Leaving directory `.'
  ../src/java_raw_api.c: In function 'ffi_java_raw_call':
  ../src/java_raw_api.c:317:3: warning: 'ffi_java_raw_to_ptrarray' is deprecated [-Wdeprecated-declarations]
    317 |   ffi_java_raw_to_ptrarray (cif, raw, avalue);
        |   ^~~~~~~~~~~~~~~~~~~~~~~~
  ../src/java_raw_api.c:76:1: note: declared here
     76 | ffi_java_raw_to_ptrarray (ffi_cif *cif, ffi_java_raw *raw, void **args)
        | ^~~~~~~~~~~~~~~~~~~~~~~~
  ../src/java_raw_api.c: In function 'ffi_java_translate_args':
  ../src/java_raw_api.c:328:3: warning: 'ffi_java_raw_size' is deprecated [-Wdeprecated-declarations]
    328 |   ffi_java_raw *raw = (ffi_java_raw*)alloca (ffi_java_raw_size (cif));
        |   ^~~~~~~~~~~~
  ../src/java_raw_api.c:45:1: note: declared here
     45 | ffi_java_raw_size (ffi_cif *cif)
        | ^~~~~~~~~~~~~~~~~
  ../src/java_raw_api.c:331:3: warning: 'ffi_java_ptrarray_to_raw' is deprecated [-Wdeprecated-declarations]
    331 |   ffi_java_ptrarray_to_raw (cif, avalue, raw);
        |   ^~~~~~~~~~~~~~~~~~~~~~~~
  ../src/java_raw_api.c:157:1: note: declared here
    157 | ffi_java_ptrarray_to_raw (ffi_cif *cif, void **args, ffi_java_raw *raw)
        | ^~~~~~~~~~~~~~~~~~~~~~~~
  ../src/java_raw_api.c: In function 'ffi_prep_java_raw_closure':
  ../src/java_raw_api.c:369:3: warning: 'ffi_prep_java_raw_closure_loc' is deprecated [-Wdeprecated-declarations]
    369 |   return ffi_prep_java_raw_closure_loc (cl, cif, fun, user_data, cl);
        |   ^~~~~~
  ../src/java_raw_api.c:337:1: note: declared here
    337 | ffi_prep_java_raw_closure_loc (ffi_java_raw_closure* cl,
        | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  ar: `u' modifier ignored since `D' is the default (see `U')
  ar: `u' modifier ignored since `D' is the default (see `U')
  thread 'main' panicked at 'Unable to find libclang: "the `libclang` shared library at /usr/lib/libclang.so.10 could not be opened: Dynamic loading not supported"', /opt/build/php/.cargo/registry/src/github.com-1ecc6299db9ec823/bindgen-0.52.0/src/lib.rs:1895:31
  stack backtrace:
     0: rust_begin_unwind
               at /rustc/ffa2e7ae8fbf9badc035740db949b9dae271c29f/library/std/src/panicking.rs:483:5
     1: core::panicking::panic_fmt
               at /rustc/ffa2e7ae8fbf9badc035740db949b9dae271c29f/library/core/src/panicking.rs:85:14
     2: core::option::expect_none_failed
               at /rustc/ffa2e7ae8fbf9badc035740db949b9dae271c29f/library/core/src/option.rs:1268:5
     3: core::result::Result<T,E>::expect
     4: core::ops::function::FnOnce::call_once
     5: lazy_static::lazy::Lazy<T>::get::{{closure}}
     6: std::sync::once::Once::call_once::{{closure}}
     7: std::sync::once::Once::call_inner
               at /rustc/ffa2e7ae8fbf9badc035740db949b9dae271c29f/library/std/src/sync/once.rs:419:21
     8: std::sync::once::Once::call_once
     9: <bindgen::ensure_libclang_is_loaded::LIBCLANG as core::ops::deref::Deref>::deref
    10: bindgen::ensure_libclang_is_loaded
    11: bindgen::Bindings::generate
    12: bindgen::Builder::generate
    13: build_script_build::generate_bindings
    14: build_script_build::main
    15: core::ops::function::FnOnce::call_once
  note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
warning: build failed, waiting for other jobs to finish...
warning: 2 warnings emitted

error: build failed
```

### Steps to reproduce

build from alpine/edge,  already install musl-dev clang-dev llvm-dev. 

```sh
ls -alFh /usr/lib/libclang.so.10
-rwxr-xr-x    1 root     root       26.0M Jul 30 14:48 /usr/lib/libclang.so.10*
 ldd /usr/lib/libclang.so.10
	/lib/ld-musl-x86_64.so.1 (0x7fcca8fcf000)
	libLLVM-10.so => /usr/lib/../lib/libLLVM-10.so (0x7fcca38de000)
	libstdc++.so.6 => /usr/lib/../lib/libstdc++.so.6 (0x7fcca373d000)
	libgcc_s.so.1 => /usr/lib/../lib/libgcc_s.so.1 (0x7fcca3723000)
	libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7fcca8fcf000)
	libffi.so.7 => /usr/lib/../lib/libffi.so.7 (0x7fcca3718000)
	libz.so.1 => /lib/libz.so.1 (0x7fcca36fe000)
	libxml2.so.2 => /usr/lib/../lib/libxml2.so.2 (0x7fcca35d4000)
	liblzma.so.5 => /usr/lib/../lib/liblzma.so.5 (0x7fcca35b1000)
```

### Expected behavior

expect build c-api static lib.

jubianchi:
This patch adds a specific build for musl. Currently, it will only support JIT engine.

I also changes the workflow definition a bit so we don't depend on the OS name but rather on the environment ID which is our convention.

Closes #1482
Closes #1766

# Description

# Review

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -38,7 +38,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        build: [linux-x64, macos-x64, macos-arm64, windows-x64, linux-aarch64]
         include:
           - build: linux-x64
             os: ubuntu-18.04
@@ -77,14 +76,15 @@ jobs:
       CARGO_SCCACHE_VERSION: 0.2.14-alpha.0-parity
       SCCACHE_AZURE_BLOB_CONTAINER: wasmerstoragesccacheblob
       SCCACHE_AZURE_CONNECTION_STRING: ${{ secrets.SCCACHE_AZURE_CONNECTION_STRING }}
+      TARGET: ${{ matrix.target }}
     steps:
       - uses: actions/checkout@v2
       - name: Set up libstdc++ on Linux
+        if: matrix.build == 'linux-x64'
         run: |
           sudo apt-get update -y
           sudo apt-get install -y --allow-downgrades libstdc++6=8.4.0-1ubuntu1~18.04
           sudo apt-get install --reinstall g++-8
-        if: matrix.os == 'ubuntu-18.04'
       - name: Install Rust ${{ matrix.rust }}
         uses: actions-rs/toolchain@v1
         with:
@@ -125,7 +125,6 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
           key: ${{ matrix.build }}-${{ matrix.target }}-cargo-${{ hashFiles('Cargo.lock') }}-v1
-      # Install sccache
       - uses: actions/cache@v2
         with:
           path: ${{ runner.tool_cache }}/cargo-sccache
@@ -173,7 +172,7 @@ jobs:
       - name: Test
         run: |
           make test
-        if: matrix.target != 'aarch64-apple-darwin'
+        if: matrix.build != 'macos-arm64'
       - name: Test C API
         run: |
           make test-capi
@@ -190,9 +189,9 @@ jobs:
         run: |
           make build-wasmer
       - name: Build Wapm binary
+        if: needs.setup.outputs.DOING_RELEASE == '1'
         run: |
           make build-wapm
-        if: needs.setup.outputs.DOING_RELEASE == '1'
       - name: Install Nightly Rust for Headless
         uses: actions-rs/toolchain@v1
         with:
@@ -202,6 +201,7 @@ jobs:
           components: "rust-src"
         if: needs.setup.outputs.DOING_RELEASE == '1'
       - name: Build Minimal Wasmer Headless
+        if: needs.setup.outputs.DOING_RELEASE == '1'
         run: |
           cargo install xargo
           echo "" >> Cargo.toml
@@ -216,17 +216,6 @@ jobs:
           echo "codegen-units = 1" >> Cargo.toml
           echo "rpath = false" >> Cargo.toml
           make build-wasmer-headless-minimal
-        if: needs.setup.outputs.DOING_RELEASE == '1'
-      - name: Copy target binaries
-        run: |
-          mkdir -p target/release
-          cp target/${{matrix.target}}/release/wasmer* target/release
-          cp target/${{matrix.target}}/release/libwasmer* target/release
-          if [ -d "wapm-cli" ]; then
-            mkdir -p wapm-cli/target/release 
-            cp wapm-cli/target/${{matrix.target}}/release/wapm* wapm-cli/target/release
-          fi
-        if: matrix.target
       - name: Dist
         run: |
           make distribution
@@ -243,7 +232,7 @@ jobs:
           make test-integration
         if: matrix.run_integration_tests && matrix.os != 'windows-latest'
       - name: Cross compile from Linux
-        if: matrix.os == 'ubuntu-18.04'
+        if: matrix.build == 'linux-x64'
         shell: bash
         run: |
           ls target/release
@@ -282,6 +271,64 @@ jobs:
         with:
           name: ${{ matrix.cross_compilation_artifact_name }}
           path: cross
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: ${{ matrix.artifact_name }}
+          path: dist
+          if-no-files-found: error
+          retention-days: 1
+
+  test-docker:
+    name: Test on ${{ matrix.build }}
+    runs-on: ubuntu-latest
+    needs: setup
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - build: linux-musl-x64
+            image: alpine:latest
+            rust: 1.49
+            artifact_name: 'wasmer-linux-musl-amd64'
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: addnab/docker-run-action@v1
+        with:
+          image: ${{ matrix.image }}
+          options: -v ${{ github.workspace }}:/work
+          run: |
+            set -e
+
+            # Set up tools
+            apk add musl-dev curl make libtool libffi-dev gcc automake autoconf git
+
+            # Install Rust ${{ matrix.rust }}
+            curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs > rustup.sh
+            sh ./rustup.sh -y
+            source ~/.cargo/env
+            rm -f ./rustup.sh
+
+            # Set up Rust
+            rustup toolchain install ${{ matrix.rust }}
+            rustup default ${{ matrix.rust }}
+
+            # Change working directory
+            cd /work
+
+            # Test
+            make test test-capi
+
+            # Build C API, Wasmer & Wapm
+            make build-capi build-wasmer build-wapm
+
+            # Dist
+            make distribution
+
+            # Run integration tests
+            export WASMER_DIR=`pwd`/package
+            make test-integration
       - name: Upload Artifacts
         uses: actions/upload-artifact@v2
         with:
@@ -340,7 +387,7 @@ jobs:
   #        wasmer run artifacts/cross/cross_compiled_from_mac/win_from_mac.wjit -- --eval "function greet(name) { return JSON.stringify('Hello, ' + name); }; print(greet('World'));"
 
   release:
-    needs: [setup, test, test-cross-compile-on-linux, test-cross-compile-on-mac] #, test-cross-compile-on-win]
+    needs: [setup, test, test-docker, test-cross-compile-on-linux, test-cross-compile-on-mac] #, test-cross-compile-on-win]
     runs-on: ubuntu-latest
     if: needs.setup.outputs.DOING_RELEASE == '1'
     steps:
@@ -348,7 +395,6 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           path: artifacts
-
       - name: Create Release
         id: create_release
         uses: actions/create-release@v1
@@ -359,7 +405,6 @@ jobs:
           release_name: Release ${{ needs.setup.outputs.VERSION }}
           draft: true
           prerelease: false
-
       - name: Upload Release Asset Windows Installer
         uses: actions/upload-release-asset@v1
         env:
@@ -369,7 +414,6 @@ jobs:
           asset_path: artifacts/wasmer-windows-amd64/WasmerInstaller.exe
           asset_name: wasmer-windows.exe
           asset_content_type: application/vnd.microsoft.portable-executable
-
       - name: Upload Release Asset Windows
         uses: actions/upload-release-asset@v1
         env:
@@ -379,7 +423,6 @@ jobs:
           asset_path: artifacts/wasmer-windows-amd64/wasmer.tar.gz
           asset_name: wasmer-windows-amd64.tar.gz
           asset_content_type: application/gzip
-
       - name: Upload Release Asset Linux amd64
         uses: actions/upload-release-asset@v1
         env:
@@ -389,7 +432,16 @@ jobs:
           asset_path: artifacts/wasmer-linux-amd64/wasmer.tar.gz
           asset_name: wasmer-linux-amd64.tar.gz
           asset_content_type: application/gzip
-
+      - name: Upload Release Asset Linux amd64 (musl)
+        id: upload-release-asset-linux-musl-amd64
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: artifacts/wasmer-linux-musl-amd64/wasmer.tar.gz
+          asset_name: wasmer-linux-musl-amd64.tar.gz
+          asset_content_type: application/gzip
       - name: Upload Release Asset Mac amd64
         uses: actions/upload-release-asset@v1
         env:
@@ -399,7 +451,6 @@ jobs:
           asset_path: artifacts/wasmer-darwin-amd64/wasmer.tar.gz
           asset_name: wasmer-darwin-amd64.tar.gz
           asset_content_type: application/gzip
-
       - name: Upload Release Asset Mac arm64
         uses: actions/upload-release-asset@v1
         env:
@@ -409,7 +460,6 @@ jobs:
           asset_path: artifacts/wasmer-darwin-arm64/wasmer.tar.gz
           asset_name: wasmer-darwin-arm64.tar.gz
           asset_content_type: application/gzip
-
       - name: Upload Release Asset Linux aarch64
         uses: actions/upload-release-asset@v1
         env:
--- a/Makefile
+++ b/Makefile
@@ -295,21 +295,34 @@ capi_compiler_features := --features $(subst $(space),$(comma),$(filter-out llvm
 ifneq (, $(filter 1, $(IS_DARWIN) $(IS_LINUX)))
 	bold := $(shell tput bold 2>/dev/null || echo -n '')
 	green := $(shell tput setaf 2 2>/dev/null || echo -n '')
+	yellow := $(shell tput setaf 3 2>/dev/null || echo -n '')
 	reset := $(shell tput sgr0 2>/dev/null || echo -n '')
 endif
 
 HOST_TARGET=$(shell rustup show | grep 'Default host: ' | cut -d':' -f2 | tr -d ' ')
+
+TARGET_DIR := target/release
+
+ifneq (, $(TARGET))
+	TARGET_DIR := target/$(TARGET)/release
+endif
 
 $(info -----------)
 $(info $(bold)$(green)INFORMATION$(reset))
 $(info -----------)
 $(info )
 $(info Host Target: `$(bold)$(green)$(HOST_TARGET)$(reset)`.)
+ifneq (, $(TARGET))
+	# We use spaces instead of tabs to indent `$(info)`
+	# otherwise it's considered as a command outside a
+	# target and it will fail.
+        $(info Build Target: $(bold)$(green)$(TARGET)$(reset) $(yellow)($(TARGET_DIR))$(reset))
+endif
 ifneq (, $(LIBC))
-		# We use spaces instead of tabs to indent `$(info)`
-		# otherwise it's considered as a command outside a
-		# target and it will fail.
-                $(info C standard library: $(bold)$(green)$(LIBC)$(reset))
+	# We use spaces instead of tabs to indent `$(info)`
+	# otherwise it's considered as a command outside a
+	# target and it will fail.
+        $(info C standard library: $(bold)$(green)$(LIBC)$(reset))
 endif
 $(info Enabled Compilers: $(bold)$(green)$(subst $(space),$(reset)$(comma)$(space)$(bold)$(green),$(compilers))$(reset).)
 $(info Compilers + engines pairs (for testing): $(bold)$(green)${compilers_engines}$(reset))
@@ -324,7 +337,6 @@ $(info $(bold)$(green)RULE EXECUTION$(reset))
 $(info --------------)
 $(info )
 $(info )
-
 
 ############
 # Building #
@@ -579,13 +591,13 @@ package-wapm:
 	mkdir -p "package/bin"
 ifneq (, $(filter 1, $(IS_DARWIN) $(IS_LINUX)))
 	if [ -d "wapm-cli" ]; then \
-		cp wapm-cli/target/release/wapm package/bin/ ;\
+		cp wapm-cli/$(TARGET_DIR)/wapm package/bin/ ;\
 		echo "#!/bin/bash\nwapm execute \"\$$@\"" > package/bin/wax ;\
 		chmod +x package/bin/wax ;\
 	fi
 else
 	if [ -d "wapm-cli" ]; then \
-		cp wapm-cli/target/release/wapm package/bin/ ;\
+		cp wapm-cli/$(TARGET_DIR)/wapm package/bin/ ;\
 	fi
 ifeq ($(IS_DARWIN), 1)
 	codesign -s - package/bin/wapm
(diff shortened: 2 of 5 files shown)


jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try-

jubianchi:
bors try

jubianchi:
bors try-

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

syrusakbary:
Why no support for the native engine?

Neves-Adriano (approved):
👍🏻

jubianchi:
bors try

jubianchi:
@syrusakbary there are problem with the native engine and linkage: currently, musl is statically linked and it crates problem with dynamic loading.

I chose to disable these things for now and then experiment more on using musl as a shared library. 

Reading the rust documentation it seems as easy as using a `target-feature=-crt-static` but it does not work (and I don't know why yet)

jubianchi:
bors try

jubianchi:
bors try-

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try-

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try-

Makefile
```diff
@@ -79,9 +88,9 @@ compilers := $(filter-out ,$(compilers))
 test_compilers_engines := $(filter-out ,$(test_compilers_engines))
 
 ifneq ($(OS), Windows_NT)
-	bold := $(shell tput bold)
-	green := $(shell tput setaf 2)
-	reset := $(shell tput sgr0)
+	bold := $(shell tput bold 2>/dev/null || echo -n '')
+	green := $(shell tput setaf 2 2>/dev/null || echo -n '')
+	reset := $(shell tput sgr0 2>/dev/null || echo -n '')
```
Hywan, line 93: :+1:

Makefile
```diff
@@ -376,7 +389,9 @@ ifeq ($(UNAME_S), Darwin)
 	# Fix the rpath for the dylib
 	install_name_tool -id "@rpath/libwasmer.dylib" package/lib/libwasmer.dylib
 else
-	cp target/release/libwasmer_c_api.so package/lib/libwasmer.so
+	if [ -f "target/$(HOST_TARGET)/release/wasmer-headless" ]; then \
```
Hywan, line 392: I presume you wanted to write:

```suggestion
	if [ -f "target/$(HOST_TARGET)/wasmer-headless" ]; then \
```

But why moving `libwasmer_c_api.so` to `libwasmer.so` if `wasmer-headless` exists?

examples/engine_cross_compilation.rs
```diff
@@ -97,13 +97,3 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-
-#[test]
-#[cfg(not(any(
-    windows,
-    // We don't support yet crosscompilation in macOS with Apple Silicon
-    all(target_os = "macos", target_arch = "aarch64")
-)))]
-fn test_cross_compilation() -> Result<(), Box<dyn std::error::Error>> {
-    main()
-}
```
Hywan, line 109: I assumed you wanted to remove the `#[cfg]` but keep the `fn test_cross_compilation`, is that correct?

.github/workflows/main.yaml
```diff
@@ -171,22 +181,31 @@ jobs:
       - name: Test
         run: |
           make test
-        if: matrix.target != 'aarch64-apple-darwin'
+        if: matrix.target != 'aarch64-apple-darwin' && matrix.build != 'linux-musl-x64'
+      - name: Test (musl)
+        if: matrix.build == 'linux-musl-x64'
+        run: |
+          LIBC=musl WASMER_CAPI_USE_SYSTEM_LIBFFI=1 make test
```
MarkMcCaskey, line 188: I wonder if it's possible to have this be one step by using the matrix to set the env vars ahead of time...

That would be optimal but I find it easy to believe that that's something that sounds way easier than it actually is to do.

.github/workflows/main.yaml
```diff
@@ -171,22 +181,31 @@ jobs:
...
       - name: Test C API
         run: |
           make test-capi
-        if: matrix.os != 'windows-latest' && matrix.target != 'aarch64-apple-darwin' # we can't test yet on Apple Silicon or Windows
+        if: matrix.os != 'windows-latest' && matrix.target != 'aarch64-apple-darwin' && matrix.build != 'linux-musl-x64' # we can't test yet on Apple Silicon or Windows
```
MarkMcCaskey, line 192: At some point we may just want to encode the logic in the matrix directly as a variable like `should-test-capi` and have `true`/`false` for each target.

I'll leave it up to your judgement what you think is better, just thinking out loud

.github/workflows/main.yaml
```diff
@@ -171,22 +181,31 @@ jobs:
...
-        if: matrix.target == 'aarch64-apple-darwin'
+        if: matrix.target == 'aarch64-apple-darwin' || matrix.build == 'linux-musl-x64'
       - name: Build Wasmer binary
+        if: matrix.build != 'linux-musl-x64'
         run: |
           make build-wasmer
+      - name: Build Wasmer binary (musl)
+        if: matrix.build == 'linux-musl-x64'
+        run: |
+          LIBC=musl WASMER_CAPI_USE_SYSTEM_LIBFFI=1 make build-wasmer
```
MarkMcCaskey, line 208: Yeah, if we could set env vars then we wouldn't need to duplicate this either

MarkMcCaskey: It'll also scale better as we add more things

Makefile
MarkMcCaskey: I'm also unclear on the logic here

MarkMcCaskey (commented):
Looks pretty good to me

One thing to consider before shipping: does `wapm` build properly on musl?

It might also be a good idea to get this into `wasmer-nightly`, some recent changes I mad ein `main.yaml` also need to be ported to wasmer-nightly I think... 🤔

.github/workflows/main.yaml
jubianchi: @MarkMcCaskey I totally agree.

I started my PR with much more things in the workflow/makefile (improvements, optimizations) but ran into problem and reverted everything to try to find the root cause.

I think we can explore this in a dedicated PR.

jubianchi:
bors try

.github/workflows/main.yaml
```diff
@@ -171,26 +181,35 @@ jobs:
       - name: Test
         run: |
           make test
-        if: matrix.target != 'aarch64-apple-darwin'
+        if: matrix.target != 'aarch64-apple-darwin' && matrix.build != 'linux-musl-x64'
```
Hywan, line 184: For the sake of consistency:

```suggestion
        if: matrix.build != 'macos-arm64' && matrix.build != 'linux-musl-x64'
```

.github/workflows/main.yaml
Hywan: I don't understand. We don't test the C API with musl, but we build it (just below this line)? Would be nice to test it, don't you think :-p?

jubianchi:
> I don't understand. We don't test the C API with musl, but we build it (just below this line)? Would be nice to test it, don't you think :-p?

Some tests are failing with musl but I wanted to have the library so I can use it to test without having to build it manually.

I will fix the workflow to not build the C API with musl: it does not really work so let's not ship it.

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try-

jubianchi:
bors trybors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try-

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try-

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

jubianchi:
bors try

.cargo/config.toml
```diff
@@ -1,4 +1,4 @@
 [target.'cfg(target_os = "linux")']
 rustflags = [
-  "-C", "link-arg=-Wl,-E"
+  "-C", "link-arg=-Wl,-E",
```
nlewycky, line 3: (No action required.) Could you explain?

.github/workflows/main.yaml
```diff
@@ -216,12 +216,13 @@ jobs:
           echo "codegen-units = 1" >> Cargo.toml
           echo "rpath = false" >> Cargo.toml
           make build-wasmer-headless-minimal
-        if: needs.setup.outputs.DOING_RELEASE == '1'
       - name: Copy target binaries
         run: |
           mkdir -p target/release
           cp target/${{matrix.target}}/release/wasmer* target/release
-          cp target/${{matrix.target}}/release/libwasmer* target/release
+          if find target/${{matrix.target}}/release -type f -name 'libwasmer*' -maxdepth 1 | grep -q "."; then
```
nlewycky, line 223: Why would there ever be a libwasmer without a `.` in the name? Maybe this could have a comment?

.github/workflows/main.yaml
```diff
@@ -290,6 +291,73 @@ jobs:
...
+            rustup default ${{ matrix.rust }}
+
+            # Change working directory
+            cd /work
+
+            # Test
+            make test
+
+            # Test C API
+            make test-capi
```
nlewycky, line 336: Any particular reason not to use `make test test-capi`?

Makefile
```diff
@@ -323,6 +325,9 @@ $(info --------------)
 $(info )
 $(info )
 
+ifneq (, $(TARGET))
+	TARGET := $(TARGET)/
+endif
```
nlewycky, line 330: How about TARGET_DIR := target/$(TARGET)/release when true or TARGET_DIR := target/release when TARGET is blank? Then we can simplify a lot of paths.

My underlying problem with this is that a single variable with one name switched its "logical type" from being a string containing just the target type, to being a directory name.

lib/c-api/tests/.gitignore
```diff
@@ -1,6 +1,11 @@
 # ignore wasm-c-api binaries
 wasm-c-api-*
 test-*
+wasm-c-api/example/*
```
nlewycky, line 4: Is this change related to musl?

.cargo/config.toml
jubianchi: Something I forgot to remove while testing different configurations but I like having trailing-comma when it's allowed to reduce diff when adding lines into lists.

.github/workflows/main.yaml
jubianchi: only reason is to keep the script as close as possible to the workflow steps we use in the previous job.

I'll make the tests steps a one-liner ;)

lib/c-api/tests/.gitignore
jubianchi: nope.

It can be extracted to a dedicated PR if needed.

.github/workflows/main.yaml
jubianchi: This step will be removed:  the `TARGET` variable in the Makefile will do the job :)

jubianchi:
bors try

.github/workflows/main.yaml
```diff
@@ -389,7 +432,16 @@ jobs:
...
+      - name: Upload Release Asset Linux amd64 (musl)
+        id: upload-release-asset-linux-musl-amd64
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: artifacts/wasmer-linux-musl-amd64/wasmer.tar.gz
+          asset_name: wasmer-linux-musl-amd64.tar.gz
+          asset_content_type: application/gzip
```
Hywan, line 444: It seems like this code is a duplication of the code below (see my comment).

.github/workflows/main.yaml
```diff
@@ -419,6 +469,15 @@ jobs:
...
           asset_content_type: application/gzip
+      - name: Upload Release Asset Linux (musl) amd64
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: artifacts/wasmer-linux-musl-amd64/wasmer.tar.gz
+          asset_name: wasmer-linux-musl-amd64.tar.gz
+          asset_content_type: application/gzip
```
Hywan, line 480: This is the duplicated code I'm mentionning in my comment above (line 435).

We should keep this block, and remove the previous one. The `id` key isn't necessary.

Makefile
```diff
@@ -616,25 +630,28 @@ package-capi:
...
+		cp $(TARGET_DIR)/libwasmer_c_api.dylib package/lib/libwasmer.dylib ;\
+		install_name_tool -id "@rpath/libwasmer.dylib" package/lib/libwasmer.dylib ;\
+	fi
+
+	if [ -f $(TARGET_DIR)/libwasmer_c_api.so ]; then \
+		cp $(TARGET_DIR)/libwasmer_c_api.so package/lib/libwasmer.so ;\
+	fi
+	if [ -f $(TARGET_DIR)/libwasmer_c_api.a ]; then \
+		cp $(TARGET_DIR)/libwasmer_c_api.a package/lib/libwasmer.a ;\
+	fi
```
Hywan, line 654: I like this new approach. It's more “error-proof”.

Hywan (approved):
The asset upload duplication must be fixed, but apart from that, I'm approving this PR. Good job!

.github/workflows/main.yaml
```diff
@@ -38,7 +38,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        build: [linux-x64, macos-x64, macos-arm64, windows-x64, linux-aarch64]
```
MarkMcCaskey, line 41: Huh, I had just assumed that line was always required

.github/workflows/main.yaml
```diff
@@ -173,7 +172,7 @@ jobs:
       - name: Test
         run: |
           make test
-        if: matrix.target != 'aarch64-apple-darwin'
+        if: matrix.build != 'macos-arm64'
```
MarkMcCaskey, line 175: Are target and build the same here? It seems like they could be different

.github/workflows/main.yaml
```diff
@@ -216,17 +216,6 @@ jobs:
...
-      - name: Copy target binaries
-        run: |
-          mkdir -p target/release
-          cp target/${{matrix.target}}/release/wasmer* target/release
-          cp target/${{matrix.target}}/release/libwasmer* target/release
-          if [ -d "wapm-cli" ]; then
-            mkdir -p wapm-cli/target/release 
-            cp wapm-cli/target/${{matrix.target}}/release/wapm* wapm-cli/target/release
-          fi
-        if: matrix.target
```
MarkMcCaskey, line 229: This looks like a work-around for aarch64 macos, do we not need it anymore?

.github/workflows/main.yaml
```diff
@@ -290,6 +279,64 @@ jobs:
...
+            image: alpine:latest
+            rust: 1.49
+            artifact_name: 'wasmer-linux-musl-amd64'
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: addnab/docker-run-action@v1
+        with:
+          image: ${{ matrix.image }}
+          options: -v ${{ github.workspace }}:/work
```
MarkMcCaskey, line 300: is the `:/` intentional or should this just be `/`?

Makefile
MarkMcCaskey: I was actually just about to comment that I think it's more error prone as we'll silently fail if we're missing any of these files and will just have to wait until a user reports or we notice that we didn't release one of the files we intended to.

Ideally our packaging logic would be tested and not live in our CI file. We've lost a ton of things done in CI over time from all the CI migrations we've done -- I wouldn't put anything too critical there

MarkMcCaskey: I do think the above code is better though, but we do need tests for our packaging logic and the change above makes the need for tests greater.

.github/workflows/main.yaml
jubianchi: It intentional, this part is a Docker volume mapping which syntax is `<path-on-the-host>:<path-in-the-container>`

.github/workflows/main.yaml
jubianchi: With the new `TARGET_DIR` variable in the `Makefile` this step is not required anymore.

.github/workflows/main.yaml
jubianchi: `matrix.target` and `matrix.os` are names defined by either Rust or GitHub. `matrix.build` is something we chosse on our side so I prefer to rely on our naming convention here. 

If GitHub forces us to change the `os` this will not impact the workflow. Same for `target` (even if it's unlikely they will change)

Makefile
jubianchi: @MarkMcCaskey the problem with our previous strategy here was we relied on the host OS to define the files to copy.

This is not always the right thing to do, for example when we cross-compile this is why I changed that. Also, our strategy was not adaptative: if a user decides to change his build configuration (for example move from `cdylib` to `staticlib`) he'll have to change the Makefile. With this change, everything will automatically work.

.github/workflows/main.yaml
MarkMcCaskey: Sure that makes sense, I just want to check that there aren't extra semantics here. If I remember correctly, arm64 Mac used `target` or some other key for a very specific purpose and I think the difference might have been important because of cross-compilation or something

Makefile
MarkMcCaskey: Yeah that makes sense, my one concern is just that it's less specific. Rather than encoding an expected state of the world "given condition X, we must have files Y, Z, ..." we just take whatever is there. So any bug that affects the files we produce will now go unnoticed.

I think the real fix is to just have actual testing for our release process but I think that's out of scope for this PR.

.github/workflows/main.yaml
jubianchi: @markandrus yeah, to build for M1 we cross compile so we define (in the matrix) the Rust target to use and we were also using this in our conditionals.

But now, we have a problem: when GitHub will support M1 directly, we won't cross-compile anymore (we won't force the target through the matrix). Without my change, this would also require to change all the conditionals. With my change, we only update the matrix and we are done.

Makefile
Hywan: Let's accept the following patch, and then immediately open a new PR to test the `package/` directory? We also have various scheduled tests that run based`wasmer-nightly` (like in `wasmer-go`, `wasmer-php` etc.).

PR #2011, opened 2021-01-12
Feature/headless wasmer revamp

syrusakbary:
# Description

This PR:
* [x] Adds a new wasmer-headless binary
* [x] Simplifies the Makefile process
* [x] Improves Makefile so the stripe binary doesn't strip out the wasmer symbols


Makefile
```diff
@@ -329,16 +332,9 @@ endif
 endif
 
 package-minimal-headless-wasmer:
-ifdef cross_compiling_to_mac_aarch64
-	if [ -f "target/aarch64-apple-darwin/release/wasmer" ]; then \
-		cp target/aarch64-apple-darwin/release/wasmer package/bin/wasmer-headless ;\
+	if [ -f "target/$(HOST_TARGET)/release/wasmer-headless" ]; then \
```
MarkMcCaskey, line 335: Don't we still need the special casing for `aarch64` macos here? if we're not building on actual aarch64 macos machines, I believe this will not work correctly

lib/cli/Cargo.toml
```diff
@@ -107,5 +107,4 @@ llvm = [
 ]
 debug = ["fern", "log", "wasmer-wasi/logging"]
 disable-all-logging = ["wasmer-wasi/disable-all-logging"]
-headless = []
-headless-minimal = ["headless", "disable-all-logging", "wasi", "native", "jit"]
+headless = ["disable-all-logging", "wasi", "native", "jit"]
```
MarkMcCaskey, line 110: I wanted to make a distinction between "headless" (should disable certain features) and "headless-minimal" (compile it to be as small as possible"), because it may be useful to build in `headless` mode without those features

lib/cli/src/bin/wasmer-headless.rs
```diff
@@ -0,0 +1,61 @@
+use anyhow::Result;
```
MarkMcCaskey, line 1: We don't have to duplicate this; I have a fix locally that doesn't duplicate it

Makefile
syrusakbary: Not sure if I follow why.
`HOST_TARGET` will be `aarch64-apple-darwin` when cross compiling to `macos-arm64` since we setup the target in cargo config:
https://github.com/wasmerio/wasmer/blob/master/.github/workflows/main.yaml#L138-L144

Am I missing something?

MarkMcCaskey: I would assume
```
 HOST_TARGET=$(shell rustup show | grep 'Default host: ' | cut -d':' -f2 | tr -d ' ') 
 ```
 
 Will not give the correct results there, but I'm just assuming

File: lib/cli/Cargo.toml (hunk 2 above)
Comment (syrusakbary) [reply lib/cli/Cargo.toml:110]: The thing we use now `wasmer-headless` binary for it (for creating the minimal headless).
So I thought we already have redundancy with the new binary name (since it can be only generated via xargo), and as such we don't need an extra flag to indicate it. Thoughts?

Comment (MarkMcCaskey) [reply lib/cli/Cargo.toml:110]: wasmer-headless can be generated without `xargo`, in fact, making `cli` have 2 `bin`s means `wasmer-headless` will be produced every time

Comment (syrusakbary) [reply lib/cli/Cargo.toml:110]: Via the `Makefile` only the `wasmer` bin would be generated.
I specifically added the `--bin` flags to the `make build-wasmer` commands

File: Makefile (hunk 1 above)
Comment (syrusakbary) [reply Makefile:335]: I just re-checked and you are right, however I just found out the previous logic was also wrong.

```cmake
cross_compiling_to_mac_aarch64 =
ifneq (,$(filter $(ARCH),aarch64 arm64))
	test_compilers_engines += cranelift-jit
	ifneq (, $(findstring llvm,$(compilers)))
		test_compilers_engines += llvm-native
	endif
	# if we are in macos arm64, we use the system libffi for the capi
	ifeq ($(UNAME_S), Darwin)
		cross_compiling_to_mac_aarch64 = yes
		use_system_ffi = yes
	endif
endif
```

This can never be true. `cross_compiling_to_mac_aarch64` will never be yes unless the host is a macOS M1 itself

syrusakbary: Re-checking this: 

```
HOST_TARGET=$(shell rustup show | grep 'Default host: ' | cut -d':' -f2 | tr -d ' ') 
```

Should return `aarch64-apple-darwin` when cross-compiling. Since it's the default (and only) one installed for that.
So we should be good!

lib/cli/Cargo.toml
MarkMcCaskey: Okay, I didn't realize that!

Another thing to consider is that since we're 1.0.0, we have to be careful about what the features mean: I don't think it makes sense for `headless` to imply `disable-all-logging`. If we want to only have 1 feature I think we should delete `headless-minimal` and keep `headless = []` as it was. Then we can manually specify the feature set for the minimal build in the Makefile.

I suppose breaking changes in the cli crate is less of an issue because we're not publishing it currently now that I think about it. But even for aesthetic reasons, I don't think we should tie disabling features to headless builds to enabling other features. I think headless wasmer with logging for example is a sensible thing to do.

syrusakbary: Yeah, I think that's a good tradeoff... can we do those changes on your PR directly?

lib/cli/src/bin/wasmer-headless.rs
syrusakbary: Cool, let's address the fix directly on your PR?

syrusakbary: Fixed

lib/cli/Cargo.toml
syrusakbary: Fixed

PR #2012, opened 2021-01-13
Refactor singlepass init stack assembly

slave5vw:
# Description

The old method generated 1 mov assembly instruction per stack slot.
This takes up 15 bytes * slots count size. So more local stack variables use, the larger the code size.
The suggested method is the same as c memset.
Initialize with a fixed instructions (24 bytes) using the rep stosq assembly instruction. It works on a range basis.
https://www.felixcloutier.com/x86/stos:stosb:stosw:stosd:stosq

However, in this way, the skip_stack_guard_page test fails.
https://github.com/wasmerio/wasmer/blob/1b49fe8475e335c6cdbb702a7100ba044201afd0/lib/vm/src/trap/traphandlers.rs#L124
The reason is that DF Flag is 0, so it is accessed in reverse order and initialized.
However, The vm trap handler checks the last address and determines it as a heap oob without range.
the stack oob is correct, and the test condition fails because it is a stack oob occurrence.

I think there are two ways. 
1. improve stack oob check in VM trap handler 
-> Abstract. There is no information on the range of access, so this may not be possible.
2. pushf/popf/std(set DF=1), and access to sequential order
-> Not recommended. Unnecessarily increases assembly size.

So for now, I added it to ignores.txt like cranelift/llvm.

# Review


losfair:
Thanks for the pull request!

I added stack probe for large allocations before `rep stosq` and it should fix the `skip_stack_guard_page` test.

losfair:
bors try

losfair:
bors try

syrusakbary:
bors try-

syrusakbary:
bors try

lib/compiler-singlepass/src/machine.rs
```diff
@@ -447,18 +450,48 @@ impl Machine {
...
 
+        // Stack probe.
+        //
+        // `rep stosq` writes data from low address to high address and may skip the stack guard page.
+        // so here we probe it explicitly when needed.
+        for i in (n_params..n).step_by(NATIVE_PAGE_SIZE / 8).skip(1) {
+            a.emit_mov(Size::S64, Location::Imm32(0), locations[i]);
+        }
+
+        //Initialize all normal locals to zero.
```
Hywan, line 468:
```suggestion
        // Initialize all normal locals to zero.
```

lib/compiler-singlepass/src/machine.rs
```diff
@@ -447,18 +450,48 @@ impl Machine {
...
+                    last_stack_loc = cmp::min(last_stack_loc, locations[i]);
+                }
+                Location::GPR(_) => {
+                    a.emit_mov(Size::S64, Location::Imm32(0), locations[i]);
+                }
+                _ => unreachable!(),
+            }
+        }
+        if init_stack_loc_cnt > 0 {
+            // Since this assemblies takes up 24 bytes, If initialize more than 2 slots, These assemblies are smallar.
```
Hywan, line 484:
```suggestion
            // Since these assemblies take up to 24 bytes, if more than 2 slots are initialized, then they are smaller.
```

Not sure here about the intended meaning? :-)

Hywan (commented):
Thanks for the PR!

lib/compiler-singlepass/src/machine.rs
brew0722: thanks for corrections!

slave5vw: thanks for correction! 
Suggestion is a feature I've never seen on github. I accepted it, and the commit message was not pretty, is it okay?

Hywan: That's OK yes :-)

Hywan:
@losfair Can you approve this PR if it's OK please? :-)

losfair (approved):

PR #2033, opened 2021-01-21
chore: Publish documentation from master branch 

jubianchi:
# Description

A documentation is published on https://wasmerio.github.io/wasmer/ from master but it is not up to date anymore.

This PR adds a `documentation` workflow to automatically publish the doc on each push on master.

# Review

### DIFF
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -1,0 +1,29 @@
+on:
+  push:
+    branches:
+      - 'master'
+
+name: Documentation
+
+jobs:
+  documentation:
+    name: Documentation
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up dependencies
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y libxkbcommon-dev doxygen
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: 1.48.0
+          override: true
+      - name: Build & package documentation
+        run: make package-docs
+      - name: Publish documentation
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./package/docs
--- a/Makefile
+++ b/Makefile
@@ -383,11 +383,11 @@ endif
 
 package-docs: build-docs build-docs-capi
 	mkdir -p "package/docs"
-	mkdir -p "package/docs/c"
+	mkdir -p "package/docs/c/runtime-c-api"
 	cp -R target/doc package/docs/crates
-	cp -R lib/c-api/doc/html package/docs/c-api
-	echo '<!-- Build $(SOURCE_VERSION) --><meta http-equiv="refresh" content="0; url=rust/wasmer_vm/index.html">' > package/docs/index.html
-	echo '<!-- Build $(SOURCE_VERSION) --><meta http-equiv="refresh" content="0; url=wasmer_vm/index.html">' > package/docs/crates/index.html
+	cp -R lib/c-api/doc/deprecated/html/ package/docs/c/runtime-c-api
+	echo '<!-- Build $(SOURCE_VERSION) --><meta http-equiv="refresh" content="0; url=crates/wasmer/index.html">' > package/docs/index.html
+	echo '<!-- Build $(SOURCE_VERSION) --><meta http-equiv="refresh" content="0; url=wasmer/index.html">' > package/docs/crates/index.html
 
 package: package-wapm package-wasmer package-minimal-headless-wasmer package-capi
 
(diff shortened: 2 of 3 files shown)


Hywan:
<3

Makefile
```diff
@@ -146,11 +146,11 @@ else
 endif
 
 build-docs:
-	cargo doc --release $(compiler_features) --document-private-items --no-deps --workspace
+	cargo doc --release --no-deps --manifest-path lib/api/Cargo.toml
```
MarkMcCaskey, line 149: Unsure about this change here, might want to talk with @nlewycky about it.

My understanding of our custom build of the docs was to be different from what docs.rs does, but these settings seem like it'd provide the same (or maybe a worse?) experience than docs.rs.

The `--no-deps` in particular, I think means a lot of types that could be linked to in the generated docs won't be linked to.

Makefile
```diff
@@ -146,11 +146,11 @@ else
...
 
 build-docs-capi:
 	cd lib/c-api/doc/deprecated/ && doxygen doxyfile
-	cargo doc --manifest-path lib/c-api/Cargo.toml --no-deps --features wat,jit,object-file,native,cranelift,wasi
+	cargo doc --release --no-deps --manifest-path lib/c-api/Cargo.toml
```
MarkMcCaskey, line 153: I think this probably isn't what we want, disabling those features means those features won't show up in the generated docs. I think we should at least have the same features enabled as what we release with

Makefile
```diff
@@ -383,11 +383,11 @@ endif
 
 package-docs: build-docs build-docs-capi
 	mkdir -p "package/docs"
-	mkdir -p "package/docs/c"
+	mkdir -p "package/docs/c/runtime-c-api"
```
MarkMcCaskey, line 386: It looks like we're removing the non-deprecated C docs from our releases here. We probably don't want to do that

MarkMcCaskey (commented):
Thanks for working on this!  I think we'll probably want to test this out a bit and iterate on it before shipping it though!

Makefile
jubianchi: I think we are good:

```
package/docs
├── c
│   └── runtime-c-api
└── crates
    ├── implementors
    ├── src
    ├── test_generator
    ├── wasmer
    ├── wasmer_c_api
    ├── wasmer_cache
    ├── wasmer_cli
    ├── wasmer_compiler
    ├── wasmer_compiler_cranelift
    ├── wasmer_compiler_llvm
    ├── wasmer_compiler_singlepass
    ├── wasmer_derive
    ├── wasmer_emscripten
    ├── wasmer_engine
    ├── wasmer_engine_dummy
    ├── wasmer_engine_jit
    ├── wasmer_engine_native
    ├── wasmer_engine_object_file
    ├── wasmer_integration_tests_cli
    ├── wasmer_middlewares
    ├── wasmer_object
    ├── wasmer_types
    ├── wasmer_vm
    ├── wasmer_wasi
    ├── wasmer_wasi_experimental_io_devices
    └── wasmer_wast

29 directories

```

MarkMcCaskey: What I meant is that the line

```
 	cp -R lib/c-api/doc/html package/docs/c-api 
 ```
 was changed to
 ```
  	cp -R lib/c-api/doc/deprecated/html/ package/docs/c/runtime-c-api 
  ```
  
  It looks like the crate-level C api is included in the docs, but I thought we had some other stuff too

Comment (MarkMcCaskey) [reply Makefile:386]: I think we should probably also rename `runtime-c-api` to `deprecated-api` or something, or otherwise make it more clear what it is!

Comment (jubianchi) [reply Makefile:386]: I kept the `runtime-c-api` name to not break existing links.

About the change on the `cp` command: the source directory changed, the doxygen part moved to another place so I had to change the command.

Comment (MarkMcCaskey) [reply Makefile:386]: Huh, it looks like we don't package docs into our releases anymore, I think we used to

Comment (MarkMcCaskey) [reply Makefile:386]: I guess it's fine then; I looked through the Makefile for the variable `SOURCE_VERSION` which is used in the package-docs command and as far as I can tell it doesn't exist.  I'm skeptical that it'll do the right thing -- it seems that code may now be out of date with respect to the rest of the Makefile, does the command do what you expect locally?

**Review by MarkMcCaskey — APPROVED** [review]

File: .github/workflows/documentation.yaml [hunk 4]
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          toolchain: 1.48.0
+          override: true
+      - name: Build & package documentation
+        run: make package-docs
+      - name: Publish documentation
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./package/docs
```
syrusakbary, line 25: I assume this will push to the `gh-pages` branch, right?

syrusakbary (approved):
LGTM

.github/workflows/documentation.yaml
jubianchi: yes right!

jubianchi:
bors retry

PR #2036, opened 2021-01-21
Use Box<[Type]>` inside of `FunctionType`

MarkMcCaskey:
This saves 16 bytes (on 64bit systems) on the stack for each `FunctionType` and
additional space on the heap by not allowing resizing (like Vec does)

There are ways to save even more memory here but they're all heavier weight than this simple change.

### DIFF
--- a/lib/wasmer-types/src/types.rs
+++ b/lib/wasmer-types/src/types.rs
@@ -229,9 +229,9 @@ impl ExternType {
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub struct FunctionType {
     /// The parameters of the function
-    params: Vec<Type>,
+    params: Box<[Type]>,
     /// The return values of the function
-    results: Vec<Type>,
+    results: Box<[Type]>,
 }
 
 impl FunctionType {
@@ -242,8 +242,8 @@ impl FunctionType {
         Returns: Into<Vec<Type>>,
     {
         Self {
-            params: params.into(),
-            results: returns.into(),
+            params: params.into().into_boxed_slice(),
+            results: returns.into().into_boxed_slice(),
         }
     }
 


syrusakbary (approved):

lib/wasmer-types/src/types.rs
```diff
@@ -242,8 +242,8 @@ impl FunctionType {
         Returns: Into<Vec<Type>>,
     {
         Self {
-            params: params.into(),
-            results: returns.into(),
+            params: params.into().into_boxed_slice(),
+            results: returns.into().into_boxed_slice(),
```
webmaster128, line 246: Rust can infer the target type of `returns.into()` to the temporary `Vec<Type>`? Impressive.

lib/wasmer-types/src/types.rs
```diff
@@ -229,9 +229,9 @@ impl ExternType {
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
 pub struct FunctionType {
     /// The parameters of the function
-    params: Vec<Type>,
+    params: Box<[Type]>,
```
webmaster128, line 232: Does this affect serde somehow?

webmaster128 (approved):

lib/wasmer-types/src/types.rs
MarkMcCaskey: I don't believe it should, but that's a good thing to check -- thanks!

lib/wasmer-types/src/types.rs
MarkMcCaskey: In this case, `params` and `returns` and generic and the _only_ thing we know about these types is that they implement `Into<Vec<Type>>`, but I agree that it is nice!

lib/wasmer-types/src/types.rs
MarkMcCaskey: Alright, checked the serde source code: they're both serialized as sequences, so there shouldn't be any effect on the serialization in any format

syrusakbary:
Can wait until we can use Github Actions in our own M1 machines

syrusakbary:
Merging manually since all tests pass except of the macos-11 due to issues in Github Actions (not in our codebase)

PR #2037, opened 2021-01-22
Improve parallelism of llvm builds with native or object file engine.

nlewycky:
On my 24 core system, a 21.5s build of qjs.so (llvm + native) is reduced to an 8.2s build.


syrusakbary (approved):
This looks great!

syrusakbary:
Merging manually, is failing because of macos 11.0 issues in Github actions

PR #2040, opened 2021-01-22
feat(vm) Add `Instance::offsets` and `InstanceHandle::vmoffsets`

### LINKED ISSUE #2032: Making VMOffsets accessible from the InstanceHandle
Hello! In my project, I have to rewrite parts of the VM context, and for that I use the `wasm_vm::VMOffsets` structure.
Right now, to access it, I write:

```rust
// with handle a &wasmer_vm::InstanceHandle
let module = handle.module_ref();
let vmoffsets = VMOffsets::new(std::mem::size_of::<*mut u8>() as u8, module);
```

Would it be possible to add a method to `InstanceHandle` so that it can directly return a `VMOffsets` that makes sense for it? Making sure that it has the same pointer size, etc.

Hywan:
# Description

This PR adds `Instance::offets`  and `InstanceHandle::vmoffsets`. It fixes #2032.
This PR also cleans up the `mod.rs` module by extracting the `InstanceRef` implementation into its own `ref.rs` module.

# Review

### DIFF
--- a/lib/vm/src/instance/mod.rs
+++ b/lib/vm/src/instance/mod.rs
@@ -8,8 +8,10 @@
 //! wrapper around an `InstanceRef`.
 
 mod allocator;
+mod r#ref;
 
 pub use allocator::InstanceAllocator;
+pub use r#ref::InstanceRef;
 
 use crate::export::VMExport;
 use crate::global::Global;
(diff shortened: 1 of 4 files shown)


Hywan:
bors try

jubianchi (approved):
Seems good, waiting for @Geal feedback on that.

lib/vm/src/instance/mod.rs
```diff
@@ -1141,6 +941,13 @@ impl InstanceHandle {
         self.instance().as_ref().vmctx_ptr()
     }
 
+    /// Return a reference to the `VMOffsets` to get offsets in the
+    /// `Self::vmctx_ptr` region. Be careful when doing pointer
+    /// arithmetic!
```
MarkMcCaskey, line 946: I don't know if we need to document anything here or if this changes anything in regards to what is public / what we're saying we won't break, but something to keep in mind here!

MarkMcCaskey (approved):

Issue #2028, opened 2021-01-17
CLI: --env should allow variables that contain `=`


artlogic:
### Describe the bug

When passing environment variables into `wasmer run` via the `--env` flag, the program will fail if the environment variable contains an `=`. It's sometimes useful or necessary to include an `=` in an environment variable. I believe this should be allowed.

Versions: `wasmer 1.0.0 |  | x86_64`

### Steps to reproduce

* Run: `wasmer run --env=MYVAR=1+2=3 test.wasm`
* Output: `error: Invalid value for '--env <KEY=VALUE>...': Env vars must be of the form <var_name>=<value>. Found MYVAR=1+2=3`

### Expected behavior

The value should be allowed and parsed correctly.

### Actual behavior

An error message is displayed and the program stops.

MarkMcCaskey:
Thanks for the report!  I thought we had this issue and the past and fixed it... I'll look into it later, just to clarify the expected behavior here is that the variable is parsed as:

name = `MYVAR`
value = `1+2=3`

is that correct?

artlogic:
That's perfect @MarkMcCaskey!

Hywan:
@MarkMcCaskey Are you refering to https://github.com/wasmerio/wasmer/pull/1762?

MarkMcCaskey:
> @MarkMcCaskey Are you refering to #1762?

Yeah, I think so. I knew someone worked on this in the past, unsure if it's since been broken (and we're not testing it?) or if something else happened.

Hywan:
This error comes from the `cli` crate itself. I'm fixing it.

PR #2042, opened 2021-01-22
feat(cli) Improve environment variables parsing

Hywan:
# Description

This PR fixes https://github.com/wasmerio/wasmer/issues/2028. In https://github.com/wasmerio/wasmer/pull/1762 we have already improved environment variables syntax, but the same work must be done in the CLI crate.

The following environment variables are considered incorrect:

* `A`, no equal sign,
* `A=`, value is missing,
* `=A`, key is missing.

The following environment variables are considered correct:

* `A=B` with `A` for the name and `B` for the value,
* `A=B=C` with `A` for the name and `B=C` for the value.

The environment variable is trimmed before being parsed with [`str::trim`](https://doc.rust-lang.org/std/primitive.str.html#method.trim).

This PR also adds tests for the `parse_envvar` function, which was missing.

# Review


lib/cli/src/utils.rs
```diff
@@ -40,14 +40,53 @@ pub fn parse_mapdir(entry: &str) -> Result<(String, PathBuf)> {
...
+            "Environment variable is not well formed, the `name` is missing in `<name>=<value>`; got `{}`",
+            &entry
+        ),
+
+        Some(position) if position == entry.len() - 1 => bail!(
+            "Environment variable is not well formed, the `value` is missing in `<name>=<value>`; got `{}`",
             &entry
+        ),
+
+        Some(position) => Ok((entry[..position].into(), entry[position + 1..].into())),
```
MarkMcCaskey, line 63: The actual splitting can just be done with just https://doc.rust-lang.org/std/primitive.str.html#method.split_at ; but we won't have as many error cases to report.

Just a heads up, I'm fine keeping it as-is

MarkMcCaskey (approved):
Looks good , thanks for the tests!

lib/cli/src/utils.rs
Hywan: We would need to deal with empty string for the left- or right-end side of the `=`. I think the code would not be simplified that much. Thanks for the heads-up!

PR #2044, opened 2021-01-22
feat(c-api) Do not build C headers if the env var `DOCS_RS` exists

### LINKED ISSUE #1954: `wasmer-c-api`'s documentation is broken on docs.rs
The `wasmer-c-api`'s documentation can not be generated on docs.rs because of this:

https://docs.rs/crate/wasmer-c-api/1.0.0-beta2/builds/325937

```
[INFO] [stderr] error: failed to run custom build command for `wasmer-c-api v1.0.0-beta2 (/opt/rustwide/workdir)`
[INFO] [stderr] 
[INFO] [stderr] Caused by:
[INFO] [stderr]   process didn't exit successfully: `/opt/rustwide/target/debug/build/wasmer-c-api-e213afef7dae2643/build-script-build` (exit code: 101)
[INFO] [stderr]   --- stderr
[INFO] [stderr]   thread 'main' panicked at 'Unable to copy the generated C bindings: Os { code: 30, kind: Other, message: "Read-only file system" }', build.rs:124:14
[INFO] [stderr]   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It would be great to see how we could solve this.

Hywan:
# Description

Fix #1954.

According to https://docs.rs/about/builds, the `DOCS_RS` environment variable is set when the documentation is built from docs.rs. In this case, we must not build the C headers because it generates such error (see #1954 to learn more):

> Unable to copy the generated C bindings: `Os { code: 30, kind: Other, message: "Read-only file system" }`

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - [#2041](https://github.com/wasmerio/wasmer/pull/2041) Documentation diagrams now have a solid white background rather than a transparent background.
 
 ### Fixed
+- [#2044](https://github.com/wasmerio/wasmer/pull/2044) Do not build C headers on docs.rs.
 
 ## 1.0.1 - 2021-01-12
 
--- a/lib/c-api/build.rs
+++ b/lib/c-api/build.rs
@@ -1,3 +1,8 @@
+//! This build script aims at:
+//!
+//! * generating the C header files for the C API,
+//! * setting `inline-c` up.
+
 use cbindgen::{Builder, Language};
 use std::{env, fs, path::PathBuf};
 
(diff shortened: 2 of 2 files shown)


MarkMcCaskey (approved):
This is an improvement but it's still broken when doing `cargo publish`; we'll have to either `cargo publish --no-verify` or `DOCS_RS=1 cargo publish`; currently it's the 1 crate not covered by publish.py: we have to publish it manually.

(An easy fix would be to extend publish.py to allow setting env vars)

Hywan:
Let's merge this PR for the moment, and let's add another to solve the `cargo publish` problem :-).

bors r+

PR #2045, opened 2021-01-22
fix(doc): Install LLVM before building the documentation

jubianchi:


MarkMcCaskey (approved):
Looks good to me!

I wish we were able to get more code reuse between our workflows

maybe we just need a lint system because my real concern is that duplication like this means it's likely it'll be updated in one place but not the other and things will break in the future. Just thinking out loud, I don't think there's anything actionable we can do in this PR for this, really

PR #2048, opened 2021-01-22
Llvm parallel improvement

syrusakbary:
# Description

This PR aims to greatly improve parallelization by removing the serialization (`write_bitcode_to_memory().as_slice().to_vec()`) and deserialization (`Module::parse_bitcode_from_buffer`) needed for each of the generated LLVM Modules.
This should improve great timings on LLVM compilation natively!

# Review


lib/compiler-llvm/src/compiler.rs
```diff
@@ -72,6 +71,25 @@ impl SymbolRegistry for ShortNames {
...
+    fn into_inner(self) -> Module<'a> {
+        Arc::try_unwrap(self.module).unwrap().into_inner().unwrap()
+    }
+    fn link_in_module(&self, linked: SendModule<'a>) {
+        self.module.lock().unwrap().link_in_module(linked.into_inner());
+    }
+}
+
+unsafe impl Send for SendModule<'_> {}
+
```
syrusakbary, line 92: I created this struct in order to be able to use LLVM modules across threads easily.

nlewycky:
Could you please try out a compile with `wasmer compile --llvm --native --enable-verifier`? I suspect this is linking modules that were created with different contexts which has undefined behaviour in LLVM that the verifier tries to catch.

lib/compiler-llvm/src/compiler.rs
```diff
@@ -72,6 +71,25 @@ impl SymbolRegistry for ShortNames {
...
+    }
+    fn into_inner(self) -> Module<'a> {
+        Arc::try_unwrap(self.module).unwrap().into_inner().unwrap()
+    }
+    fn link_in_module(&self, linked: SendModule<'a>) {
+        self.module.lock().unwrap().link_in_module(linked.into_inner());
+    }
+}
+
+unsafe impl Send for SendModule<'_> {}
```
MarkMcCaskey, line 91: I think this line is incorrect, looking at the docs of `std::sync::Mutex` it does:

```Rust
impl<T: ?Sized + Send> Send for Mutex<T>
impl<T: ?Sized + Send> Sync for Mutex<T>
```

A `std::sync::Mutex` can make a `Send` type both `Send` and `Sync` but it can't make a type that's not `Send` into either.

Similarly, `std::sync::Arc` has:

```Rust
impl<T> Send for Arc<T> where
    T: Send + Sync + ?Sized
impl<T> Sync for Arc<T> where
    T: Send + Sync + ?Sized
```
can make a `Send + Sync` type `Send + Sync` but it can't do either to types that aren't already `Send + Sync`.

So in this case we can use `Arc<Mutex<T>>` to take a `T` that's `Send` and make it `Clone`able, mutatable on any thread and `Send + Sync`, but we can't take a type that's not `Send` and do that.

Inkwell's `Module` is neither `Send` nor `Sync`, so we have a bit of a problem.

-----

`Sync` is pretty complicated and it makes sense that a lot of types can't reasonably be `Sync` (but we can make them `Sync` with `std::sync::Mutex` if they're `Send`).

`Send` on the other hand, is a lot easier to achieve, so if something isn't `Send`, there's usually a good reason. For example, if a type fundamentally relies on thread-local storage, it can't ever be `Send` because it can't be sent between threads without the behavior breaking. Because inkwell is a wrapper over C++, there may a bunch of thread-specific logic happening behind the scenes that isn't safe to do from another thread.

It may also be that it's an oversight and that `inkwell::Module` contains raw pointers, for example, but could otherwise be `Send`.

Looking into why that might be, we see:

```Rust
#[derive(Debug, PartialEq, Eq)]
pub struct Module<'ctx> {
    data_layout: RefCell<Option<DataLayout>>,
    pub(crate) module: Cell<LLVMModuleRef>,
    pub(crate) owned_by_ee: RefCell<Option<ExecutionEngine<'ctx>>>,
    _marker: PhantomData<&'ctx Context>,
}
```

Both `Cell` and `RefCell` are `Send` if the type they contain is `Send`. So our suspects are
- `DataLayout`
- `LLVMModuleRef`
- `ExecutionEngine`

----

`DataLayout` is neither `Send` nor `Sync` because it contains `LLVMStringOrRaw`

`LLVMModuleRef` is neither `Send` nor `Sync` because it's just a pointer: `type LLVMModuleRef = *mut LLVMModule;`. (Pointers default to be neither `Send` nor `Sync` as a safety precaution because raw pointers mean low level things are happening, but there's nothing inherently non-thread safe about pointers.)

`ExecutionEngine` is neither `Send` nor `Sync` because of various types that indirectly point to `Rc` (which is neither `Send` nor `Sync`) and other internal types which likely are not thread safe for the same reason as `LLVMStringOrRaw`


----

My take from looking over this is that it's extremely likely that inkwell was 1. not designed to be used in a threaded context 2. has not had effort put in to make it so.

You'd have to ask @nlewycky about how LLVM itself works, but from the Rust side, it shouldn't be too hard to make the inkwell wrapper thread safe, though it will add overhead in the form of using `Arc` instead of `Rc` and `Mutex` instead of `RefCell`, which will reduce performance in the singlethreaded use case but 🤷 , they're pretty fast and LLVM is inherently slow due to what it is and how much work it does, I think the performance lost at the Rust API layer is noise.

To be clear, the path forward here is either updating inkwell or using `unsafe` to work around inkwell and never updating inkwell because changes to inkwell might break our work around

lib/compiler-llvm/src/compiler.rs
nlewycky: This is a good idea, nearly every (are there _any_ exceptions?) LLVM type is Send, we should add them to inkwell.

syrusakbary:
Closing PR based on feedback

PR #2049, opened 2021-01-22
Remove sequencing point between the mapping steps.

nlewycky:
Slightly improves performance with llvm+native, for me with 24 cores, qjs.so builds go from 8.2s to 7.9s.


syrusakbary (approved):

Hywan:
:+1:

PR #2053, opened 2021-01-25
feat(c-api) Implement `wasi_get_unordered_imports`

Hywan:
Depend on https://github.com/wasmerio/wasmer/pull/2083.

# Description

The problem I've with the C API for WASI is the following:

```c
// At some point, we write this:
bool success = wasi_get_imports(store, module, wasi_env, &imports_to_collect);
```

This function iterates over the module imports, and for each entry, it looks inside the (inner generated) WASI import object if a matching entry exists. If it doesn't, it fails. See:

https://github.com/wasmerio/wasmer/blob/6b028410c23da088d62a6b6919e2c086931ad101/lib/c-api/src/wasm_c_api/wasi/mod.rs#L333-L356

So… if a module requires WASI and WASI only, it works like a charm of course.
But, if a module requires WASI + other imports, it will fail.

And I think it's a common pattern to use WASI + some other imports (like `math.random` or anything related to I/O etc etc.).

What we need is an API to collect all the WASI imports, in no particular order, by module/namespace and name, so that we can implement an API like `ImportObject` to re-order the imports when needed (i.e. when passing a vec of externs to `wasm_instance_new` for example).

# Review


lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -304,7 +305,124 @@ pub unsafe extern "C" fn wasi_get_wasi_version(module: &wasm_module_t) -> wasi_v
...
+/// Non-standard function to get the name of a `wasm_named_extern_t`.
+#[no_mangle]
+pub extern "C" fn wasm_named_extern_name(named_extern: &wasm_named_extern_t) -> &wasm_name_t {
+    named_extern.name.as_ref()
+}
+
+/// Non-standard function to get the wrapped extern of a
+/// `wasm_named_extern_t`.
+#[no_mangle]
+pub extern "C" fn wasm_named_extern_extern(named_extern: &wasm_named_extern_t) -> &wasm_extern_t {
```
MarkMcCaskey, line 338: Review use of `extern` here, it seems the use of `extern` in the Wasm C API is different than that of the Wasm spec / our API.

MarkMcCaskey: What I mean by that is, let's consider possibly renaming this. haha

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -304,7 +305,124 @@ pub unsafe extern "C" fn wasi_get_wasi_version(module: &wasm_module_t) -> wasi_v
...
+            let module = {
+                let mut module: Box<str> = module.into_boxed_str();
+                let module_ptr = module.as_mut_ptr();
+                let module_len = module.bytes().len();
+                let module_inner = wasm_name_t {
+                    size: module_len,
+                    data: module_ptr,
+                };
+
+                Box::leak(module);
```
MarkMcCaskey, line 386: It'd be nice if we had a conversion function for this, I think we have something like this for `Vec<T>` or something, it'd be fine to have a `From` / `Into` because this code needs to exist in multiple places and anything but a standard way to do it means we'll get weird inconsistencies or bugs

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -304,7 +305,124 @@ pub unsafe extern "C" fn wasi_get_wasi_version(module: &wasm_module_t) -> wasi_v
...
+            let name = {
+                let mut name: Box<str> = name.into_boxed_str();
+                let name_ptr = name.as_mut_ptr();
+                let name_len = name.bytes().len();
+                let name_inner = wasm_name_t {
+                    size: name_len,
+                    data: name_ptr,
+                };
+
+                Box::leak(name);
```
MarkMcCaskey, line 399: For example it could be used here

lib/c-api/src/wasm_c_api/wasi/mod.rs
MarkMcCaskey: Also as part of this, lets check the usage of `wasm_name_t`, where does it get freed? Are the strings usually host-owned? Need to make sure everything works together and we're meeting all the expectations of the standard types we're using

MarkMcCaskey (commented):
I like it! But I'm hesitant to add new APIs that users may use without a lot of thought as we'll be stuck with them for a while

lib/c-api/src/wasm_c_api/wasi/mod.rs
Hywan: Yup, totally agree.

My proposal:

1. `wasm_named_extern_inner`,
2. `wasm_named_extern_unwrap`.

What's your favorite?

lib/c-api/src/wasm_c_api/wasi/mod.rs
Hywan: It's unclear in this case. I think it's best to give the entire ownership to the caller in this case, so `imports: wasm_named_extern_vec_t` is owned by the caller.

We can free the entire vector with `wasm_named_extern_vec_delete`. However, in real code, we often need to keep a copy of the inner `wasm_extern_t` value. Problem is that `wasm_extern_copy` doesn't exist for the moment, I might implement it as part of this PR too.

lib/c-api/src/wasm_c_api/wasi/mod.rs
Hywan: (I've renamed the function `wasm_named_extern_unwrap` for the moment).

lib/c-api/src/wasm_c_api/wasi/mod.rs
Hywan: Done.

lib/c-api/src/wasm_c_api/wasi/mod.rs
Hywan: Done!

Hywan:
> I like it! But I'm hesitant to add new APIs that users may use without a lot of thought as we'll be stuck with them for a while

Well, in an ideal world, we would need an `ImportObject` API in C, but we are stuck with the Wasm C API, so I'm not sure we can do better for the moment. WASI has no standard C API for the moment neither.

lib/c-api/src/wasm_c_api/externals/mod.rs
```diff
@@ -20,6 +20,84 @@ pub struct wasm_extern_t {
...
+///     wasm_byte_vec_delete(&wat);
+///     wasm_store_delete(store);
+///     wasm_engine_delete(engine);
+///
+///     return 0;
+/// }
+/// #     })
+/// #     .success();
+/// # }
+/// ```
```
MarkMcCaskey, line 87: General feedback: this is a really long comment for a trivial function

lib/c-api/src/wasm_c_api/types/export.rs
```diff
@@ -42,19 +42,7 @@ impl From<ExportType> for wasm_exporttype_t {
...
-            let name_inner = wasm_name_t {
-                size: str_len,
-                data: char_ptr,
-            };
-            Box::leak(heap_str);
-
-            Box::new(name_inner)
-        };
-
+        let name = Box::new(other.name().to_string().into());
```
MarkMcCaskey, line 45: For readability it'd be nice to explicitly annotate the type here.  I know that rust analyzer and the IDE tools can show it when editing the code, but it's still nice to make the code clear as plain text. Otherwise, I like the code reuse!

lib/c-api/src/wasm_c_api/types/import.rs
```diff
@@ -50,37 +50,13 @@ impl From<ImportType> for wasm_importtype_t {
...
-                size: str_len,
-                data: char_ptr,
-            };
-            Box::leak(heap_str);
-
-            Box::new(name_inner)
-        };
-
+        let module = Box::new(other.module().to_string().into());
+        let name = Box::new(other.name().to_string().into());
```
MarkMcCaskey, line 54: Here too, type annotations would very slightly improve readability in my opinion

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -344,7 +345,98 @@ pub unsafe extern "C" fn wasi_get_wasi_version(module: &wasm_module_t) -> wasi_v
...
+/// Non-standard type wrapping `wasm_extern_t` with the addition of
+/// two `wasm_name_t` respectively for the module name and the name of
+/// the extern (very likely to be an import). This non-standard type
+/// is used by the non-standard `wasi_get_unordered_imports` function.
+#[allow(non_camel_case_types)]
+#[derive(Clone)]
+pub struct wasm_named_extern_t {
+    module: Box<wasm_name_t>,
+    name: Box<wasm_name_t>,
+    r#extern: Box<wasm_extern_t>,
```
MarkMcCaskey, line 357: Note to self (and to you!): we should review the ownership here and how this type is used: this says all `wasm_name_t` structs are owned by us and `Cloning` this type will deep copy data that we own meaning the names and extern will need to be freed separately

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -344,7 +345,98 @@ pub unsafe extern "C" fn wasi_get_wasi_version(module: &wasm_module_t) -> wasi_v
...
+            let name = Box::new(name.into());
+            let extern_inner = Extern::from_vm_export(store, export);
+
+            Box::new(wasm_named_extern_t {
+                module,
+                name,
+                r#extern: Box::new(wasm_extern_t {
+                    instance: None,
+                    inner: extern_inner,
+                }),
```
MarkMcCaskey, line 427: TODO: review `instance: None`, aren't `wasm_extern_t`s callable? is `instance: None` a problem for data being passed to the user?

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -409,12 +501,6 @@ pub unsafe extern "C" fn wasi_get_start_function(
     }))
 }
 
-/// Delete a `wasm_extern_t` allocated by the API.
-///
-/// cbindgen:ignore
-#[no_mangle]
-pub unsafe extern "C" fn wasm_extern_delete(_item: Option<Box<wasm_extern_t>>) {}
-
```
MarkMcCaskey, line 417: What happened to this function?

lib/c-api/wasmer.h
```diff
@@ -443,64 +434,6 @@ wasmer_result_t wasmer_compile(wasmer_module_t **module,
...
- * Execute global constructors (required if the module is compiled from C++)
- * and sets up the internal environment.
- *
- * This function sets the data pointer in the same way that
- * [`wasmer_instance_context_data_set`] does.
- */
-wasmer_result_t wasmer_emscripten_set_up(wasmer_instance_t *instance,
-                                         wasmer_emscripten_globals_t *globals);
-#endif
-
```
MarkMcCaskey, line 503: Do we want to commit these deletions?

lib/c-api/wasmer.hh
```diff
@@ -130,22 +126,6 @@ struct wasmer_byte_array {
   uint32_t bytes_len;
 };
 
-#if defined(WASMER_EMSCRIPTEN_ENABLED)
-/// Type used to construct an import_object_t with Emscripten imports.
-struct wasmer_emscripten_globals_t {
-
-};
-#endif
```
MarkMcCaskey, line 138: Here too

lib/c-api/wasmer_wasm.h
```diff
@@ -457,6 +497,88 @@ void wasm_module_name(const wasm_module_t *module, wasm_name_t *out);
  */
 bool wasm_module_set_name(wasm_module_t *module, const wasm_name_t *name);
 
+#if defined(WASMER_WASI_ENABLED)
+/**
+ * Non-standard function to get the module name of a
+ * `wasm_named_extern_t`.
+ */
+const wasm_name_t *wasm_named_extern_module(const wasm_named_extern_t *named_extern);
```
MarkMcCaskey, line 505: This may be a wider problem but we need to mark things as `own` in our extension to the Wasm C API, otherwise it's ~really hard~ impossible to know what needs to be freed by who

lib/c-api/wasmer_wasm.h
```diff
@@ -457,6 +497,88 @@ void wasm_module_name(const wasm_module_t *module, wasm_name_t *out);
...
+                               wasm_named_extern_t *const *init);
+#endif
+
+#if defined(WASMER_WASI_ENABLED)
+/**
+ *Creates an empty vector of [`wasm_named_extern_t`].
+ *
+ *# Example
+ *
+ *```rust # use inline_c::assert_c; # fn main() { #    (assert_c! { # #include "tests/wasmer_wasm.h" # int main() {     // Creates an empty vector of `wasm_named_extern_t`.     wasm_named_extern_vec_t vector;     wasm_named_extern_vec_new_empty(&vector);
```
MarkMcCaskey, line 557: Do we usually include stuff like this in the generated header file? Seems like noise

lib/c-api/wasmer_wasm.h
```diff
@@ -457,6 +497,88 @@ void wasm_module_name(const wasm_module_t *module, wasm_name_t *out);
...
+void wasm_named_extern_vec_new_empty(wasm_named_extern_vec_t *out);
+#endif
+
+#if defined(WASMER_WASI_ENABLED)
+/**
+ *Creates a new uninitialized vector of [`wasm_named_extern_t`].
+ *
+ *# Example
+ *
+ *```rust # use inline_c::assert_c; # fn main() { #    (assert_c! { # #include "tests/wasmer_wasm.h" # int main() {     // Creates an empty vector of `wasm_named_extern_t`.     wasm_named_extern_vec_t vector;     wasm_named_extern_vec_new_uninitialized(&vector, 3);
```
MarkMcCaskey, line 572: here too

MarkMcCaskey (commented):
There's a lot to unpack here, this is my initial review; I'll need to do another pass about ownership / see more examples of the ownership in action (maybe we need a test using these new functions to do something)

lib/c-api/src/wasm_c_api/externals/mod.rs
Hywan: Totally agree ^^. I'll change that to a test instead.

lib/c-api/src/wasm_c_api/types/export.rs
Hywan: :+1:

lib/c-api/src/wasm_c_api/types/import.rs
Hywan: :+1:

lib/c-api/src/wasm_c_api/wasi/mod.rs
Hywan: It's been moved into `externals/mod.rs`, with the other `wasm_extern_*` type and functions.

lib/c-api/wasmer.h
Hywan: To be honest, I didn't know why they were still here. As far as I know, our C API no longer ship with emscripten, is that right?

lib/c-api/wasmer_wasm.h
Hywan: Yup. I just don't know how to add attributes inside the C headers from Rust. I'm not even sure if it's doable.

lib/c-api/wasmer_wasm.h
Hywan: It is noise. It's from Rust comments, and automatically passed to C bindings with cbindgen. I'm not sure about how to solve that yet. We need those comments on the Rust side, but we don't want them inside the C headers.

Actually, we don't need any comments in the C headers at all. We have an online documentation, which is the best place where everything is documented. I'll try to see if we can remove comments in C headers with cbindgen.

Hywan: I've created a PR on cbindgen to skip documentation generation, https://github.com/eqrion/cbindgen/pull/652.

Hywan: PR to remove the documentation, https://github.com/wasmerio/wasmer/pull/2083

lib/c-api/src/wasm_c_api/externals/mod.rs
Hywan: Done.

lib/c-api/src/wasm_c_api/types/export.rs
Hywan: Done.

lib/c-api/src/wasm_c_api/types/import.rs
Hywan: Done.

lib/c-api/src/wasm_c_api/wasi/mod.rs
Hywan: Yes, the type owns everything (meant by the usage of `Box` here). I don't see anything incorrect here, should I?

lib/c-api/src/wasm_c_api/wasi/mod.rs
Hywan: We don't have an access to `Instance` here since we read the externs from `Module`.

`wasm_extern_t` is defined as:

```rust
pub struct wasm_extern_t {
    // this is how we ensure the instance stays alive
    pub(crate) instance: Option<Arc<Instance>>,
    pub(crate) inner: Extern,
}
```

We don't need to ensure that `Instance` stays alive, it doesn't exist at this step :-).

lib/c-api/wasmer_wasm.h
Hywan: I'll explain that in the documentation.

Hywan: (in this case, ownership isn't transferred by the way)

Hywan: While documenting onwership, I found a memory leak. We said in `wasi_get_imports` and in `wasi_get_unordered_imports` that the ownership of the `wasi_env_t` parameter is taken by the function, which wasn't the case. It was memory leak.

It's not fixed.

lib/c-api/build.rs
```diff
@@ -89,8 +87,41 @@ fn build_wasm_c_api_headers(crate_dir: &str, out_dir: &str) {
...
+//
+// This file is automatically generated by `lib/c-api/build.rs` of the
+// [`wasmer-c-api`] Rust crate.
+//
+// # Stability
+//
+// The [`wasm-c-api`] standard API is a _living_ standard. There is no
+// commitment for stability yet. We (Wasmer) will try our best to keep
+// backward compatibility as much as possible. Nonetheless, some
+// necessary API aren't yet standardized, and as such, we provide a
```
MarkMcCaskey, line 101: Small typo, should be `APIs`, I'm not going to review the grammar too closely, that kind of thing is trivial to fix later

lib/c-api/build.rs
```diff
@@ -89,8 +87,41 @@ fn build_wasm_c_api_headers(crate_dir: &str, out_dir: &str) {
...
+// custom API, e.g. `wasi_*` types and functions.
+//
+// The documentation makes it clear whether a function is unstable.
+// 
+// When a type or a function will be deprecated, it will be marked as
+// such with the appropriated compiler warning, and will be removed at
+// the next release round.
```
MarkMcCaskey, line 108: We need a tool for keeping track of what we said we'd remove. We have deprecated items in our Rust APIs (for example wasmer-wasi) saying that they should be removed before 1.0.0 launch but they're still there

lib/c-api/src/wasm_c_api/wasi/mod.rs
MarkMcCaskey: Yeah, looks fine to me on further review!

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -349,24 +345,137 @@ pub unsafe extern "C" fn wasi_get_wasi_version(module: &wasm_module_t) -> wasi_v
...
+pub unsafe extern "C" fn wasi_get_unordered_imports(
+    store: Option<&wasm_store_t>,
+    module: Option<&wasm_module_t>,
+    wasi_env: Option<Box<wasi_env_t>>,
+    imports: &mut wasm_named_extern_vec_t,
+) -> bool {
+    wasi_get_unordered_imports_inner(store, module, wasi_env, imports).is_some()
+}
+
+unsafe fn wasi_get_unordered_imports_inner(
```
MarkMcCaskey, line 412: Unclear if this function actually needs to be unsafe, same with any function not taking a pointer, really. We can always audit this later though

MarkMcCaskey (approved):
Alright, if there are any problems with this I'm not able to see them right now!

I did become aware again that we're using `unsafe` in a lot of places that we don't need to, but we can fix that separately

lib/c-api/build.rs
Hywan: We should just mark them as `#[deprecated]` I think, and remove all the types and functions before the release. Thoughts?

lib/c-api/src/wasm_c_api/wasi/mod.rs
Hywan: True! I'll remove the `unsafe` keyword.

lib/c-api/build.rs
MarkMcCaskey: Yeah, I'm saying that the function marked with that attribute wasn't removed before 1.0.0; I don't think we should add more manual, error-prone steps to the release to check all `deprecated` functions too, because in all likelihood that'll cause problems that need to be resolved and we'll just end up with a blocked release. We need like notifications a few days in advance of a release to figure those out I think. Not strictly related to this PR, but it's something we haven't figured out yet

Hywan:
bors try

PR #2054, opened 2021-01-25
Add `wasm_config_delete` to the Wasm C API

MarkMcCaskey:
Fixes 1 missing import reported in #2052 

This function is relatively new to the Wasm C API and its implementation is trivial

### DIFF
--- a/lib/c-api/src/wasm_c_api/engine.rs
+++ b/lib/c-api/src/wasm_c_api/engine.rs
@@ -131,6 +131,36 @@ pub extern "C" fn wasm_config_new() -> Box<wasm_config_t> {
     Box::new(wasm_config_t::default())
 }
 
+/// Delete a Wasmer config object.
+///
+/// This function does not need to be called if `wasm_engine_new_with_config` or
+/// another function that takes ownership of the `wasm_config_t` is called.
+///
+/// # Example
+///
+/// ```rust
+/// # use inline_c::assert_c;
+/// # fn main() {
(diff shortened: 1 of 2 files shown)


jubianchi (approved):

lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -131,6 +131,36 @@ pub extern "C" fn wasm_config_new() -> Box<wasm_config_t> {
...
+///     // Delete the configuration
+///     wasm_config_delete(config);
+///
+///     return 0;
+/// }
+/// #    })
+/// #    .success();
+/// # }
+/// ```
+/// cbindgen:ignore
```
Hywan, line 160: Why ignoring it in the C API?

Hywan (changes requested):

lib/c-api/src/wasm_c_api/engine.rs
Hywan: If ignored, it won't be present in the `.h` files, and I don't see how the doctest can work :-).

MarkMcCaskey: This is part of wasm.h , I made it ignored because `wasm_config_new` is also ignored and this is functionally the same

Hywan: Oh true!

Hywan (approved):

PR #2056, opened 2021-01-25
Depend on enumset instead of wasmer_enumset

coolreader18:
I'm not sure what ramifications this will have for semver, but since it's still relatively close to the release maybe 1.0.2 could be published and 1.0.{0,1} could be yanked? Not sure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### Changed
 - [#2042](https://github.com/wasmerio/wasmer/pull/2042) Parse more exotic environment variables in `wasmer run`.
 - [#2041](https://github.com/wasmerio/wasmer/pull/2041) Documentation diagrams now have a solid white background rather than a transparent background.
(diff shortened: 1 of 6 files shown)


syrusakbary (approved):

syrusakbary:
The audit failed because of external reasons to this PR. Merging

MarkMcCaskey:
Glad to hear it got fixed!

Yeah, so the transition to wasmer-enumset was technically a breaking change (though as far as we know, no one was affected), but because the version we had already shipped was effectively broken, we were fine shipping it without incrementing...

Going the other way though, I'm not sure what the best practice is. This is a pretty exceptional circumstance as far as I know.

Another option which doesn't break anything is to modify the `wasmer-enumset` crate to just export the real `enumset`... that could reduce _some_ breakage but it's effectively the same thing and probably not worth it.

I'll talk with the team about this today

Hywan:
> though as far as we know, no one was affected

Except `wasmer-python` :-). I can manage dealing with the breakage though.

PR #2058, opened 2021-01-26
fix(c-api) `wasi_version_t` wasn't bound correctly to C

Hywan:
# Description

Because the enum variants weren't in uppercase _and_ because one variant (`InvalidVersion`) has a non-literal value, `cbindgen` was ignoring the variants. This patch fixes that. It's not a breaking changes since the WASI version constants weren't present in the `wasmer_wasm.h` file before.

Note: `u32::max_value()` is soft-deprecated, we should use `u32::MAX` instead. Notice that none of them works with `cbindgen`, so I've hardcoded the value of `u32::MAX` here.

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - [#2056](https://github.com/wasmerio/wasmer/pull/2056) Change back to depend on the `enumset` crate instead of `wasmer_enumset`
 
 ### Fixed
+- [#2058](https://github.com/wasmerio/wasmer/pull/2058) `wasm_version_t` wasn't bound correctly to C.
 - [#2044](https://github.com/wasmerio/wasmer/pull/2044) Do not build C headers on docs.rs.
 
 ## 1.0.1 - 2021-01-12
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 - [#2056](https://github.com/wasmerio/wasmer/pull/2056) Change back to depend on the `enumset` crate instead of `wasmer_enumset`
 
 ### Fixed
-- [#2058](https://github.com/wasmerio/wasmer/pull/2058) `wasm_version_t` wasn't bound correctly to C.
+- [#2058](https://github.com/wasmerio/wasmer/pull/2058) Expose WASI versions to C correctly.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 - [#2056](https://github.com/wasmerio/wasmer/pull/2056) Change back to depend on the `enumset` crate instead of `wasmer_enumset`
 
 ### Fixed
+- [#2058](https://github.com/wasmerio/wasmer/pull/2058) `wasm_version_t` wasn't bound correctly to C.
```
MarkMcCaskey, line 19: This message in the changelog isn't meaningful without context, perhaps change it to something that users can understand as it affects them

MarkMcCaskey, line 19: This message in the changelog isn't meaningful without context, perhaps change it to something that users can understand as it affects them

MarkMcCaskey (approved):

jubianchi (approved):

syrusakbary (approved):

CHANGELOG.md
Hywan:
```suggestion
- [#2058](https://github.com/wasmerio/wasmer/pull/2058) Expose WASI versions to C correctly.
```

Hywan:
Error on Windows (for the record):

```
D:\a\wasmer\wasmer\tests\integration\cli/../../../lib/c-api\wasmer_wasm.h:105:18: error: typedef redefinition with different types ('uint32_t' (aka 'unsigned int') vs 'wasi_version_t')
    typedef uint32_t wasi_version_t;
                     ^
    D:\a\wasmer\wasmer\tests\integration\cli/../../../lib/c-api\wasmer_wasm.h:70:6: note: previous definition is here
    enum wasi_version_t {
```

Hywan:
https://github.com/wasmerio/wasmer/pull/2058/commits/7a435ef740b4c31e8662261b1ad67e43d2f8d0ad

> On Windows, using a `u32` representation for `wasi_version_t` fails if
a C++ compiler is used to treat a C program. So we change our strategy
here. We use a C representation to be FFI-safe, and the
`INVALID_VERSION` variant is now set to -1 instead of `u32::MAX`.
> 
> This patch also adds unit tests for `wasi_get_wasi_version` so that we
are sure of the behavior of this `type` on all platforms.

PR #2059, opened 2021-01-26
fix(wasi) Fix the logic behind inherited/captured `stdin`, `stdout` and `stderr`

Hywan:
# Description

This patch is both a fix and a feature!

First, let's no longer derive from `Default` for `wasi_config_t`. By
default, we want to inherit `stdin`, `stdout` and `stderr`. The
default for `bool` is `false`; we want `true` here.

Second, let's update `wasi_config_new` to correctly set `inherit_*`
fields to `true`.

Third, lets' create `wasi_config_capture_*` functions. By default,
`std*` are inherited, so we need functions to capture them. That's the
new feature this patch introduces. The `wasi_config_inherit_*`
functions are kept for the sake of backward compatibility. Ideally, we
would want an API like `wasi_config_capture_*(capture: bool)`, but it
would duplicate the API somehow.

Fourth, let's fix `wasi_env_new`. We want to capture `stdout` and
`stderr` if and only if the `inherit_*` fields are set to
`false`. There was bug here. That's why everything was working
correctly by the way: `bool::default()` is `false`, and we have this
inverted condition here, so everything was working as expected because
of a double error. The only bug was that it wasn't possible to capture
`std*` before.

# Review

### DIFF
--- a/lib/c-api/src/wasm_c_api/wasi/mod.rs
+++ b/lib/c-api/src/wasm_c_api/wasi/mod.rs
@@ -23,13 +23,12 @@ use wasmer_wasi::{
     WasiStateBuilder, WasiVersion,
 };
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 #[allow(non_camel_case_types)]
 pub struct wasi_config_t {
     inherit_stdout: bool,
     inherit_stderr: bool,
     inherit_stdin: bool,
-    /// cbindgen:ignore
     state_builder: WasiStateBuilder,
 }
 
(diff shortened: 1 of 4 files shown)


Hywan:
bors try

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -132,14 +133,31 @@ pub unsafe extern "C" fn wasi_config_mapdir(
...
 #[no_mangle]
 pub extern "C" fn wasi_config_inherit_stderr(config: &mut wasi_config_t) {
     config.inherit_stderr = true;
 }
+
+#[no_mangle]
+pub extern "C" fn wasi_config_capture_stdin(config: &mut wasi_config_t) {
+    config.inherit_stdin = false;
+}
+
```
MarkMcCaskey, line 160: What does it mean to capture stdin here?

I don't think we have any APIs to make use of this, perhaps keep it commented out so we don't export this symbol until we have a use for it.

MarkMcCaskey (approved):
Thanks! Looks good

lib/c-api/src/wasm_c_api/wasi/mod.rs
Hywan: Let's keep `wasi_config_capture_stdin` apart for the moment :-).

PR #2060, opened 2021-01-26
Add align, nofree and nonnull to vmctx pointer. Dereferenceable is still TODO.

nlewycky:
This gives LLVM more information about the vmctx pointer which it can use for better code generation and faster reasoning while compiling.

### DIFF
--- a/lib/compiler-llvm/src/abi/aarch64_systemv.rs
+++ b/lib/compiler-llvm/src/abi/aarch64_systemv.rs
@@ -11,6 +11,8 @@ use inkwell::{
 use wasmer_compiler::CompileError;
 use wasmer_types::{FunctionType as FuncSig, Type};
 
+use std::convert::TryInto;
+
 /// Implementation of the [`Abi`] trait for the Aarch64 ABI on Linux.
 pub struct Aarch64SystemV {}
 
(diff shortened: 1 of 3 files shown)


lib/compiler-llvm/src/abi/aarch64_systemv.rs
```diff
@@ -47,19 +49,36 @@ impl Abi for Aarch64SystemV {
...
         Ok(match sig.results() {
             [] => (
                 intrinsics
                     .void_ty
                     .fn_type(&param_types.collect::<Result<Vec<_>, _>>()?, false),
-                vec![],
+                vmctx_attributes
+                    .into_iter()
+                    .map(|x| (x, AttributeLoc::Param(0)))
+                    .collect(),
```
syrusakbary, line 71: This code seems repeated across the different match branches.
Why don't we move it up so we only write it once?

lib/compiler-llvm/src/abi/x86_64_systemv.rs
```diff
@@ -67,14 +81,20 @@ impl Abi for X86_64SystemV {
                 intrinsics
                     .void_ty
                     .fn_type(&param_types.collect::<Result<Vec<_>, _>>()?, false),
-                vec![],
+                vmctx_attributes
+                    .into_iter()
+                    .map(|x| (x, AttributeLoc::Param(0)))
+                    .collect(),
```
syrusakbary, line 87: This code seems repeated across the different match branches.
Why don't we move it up so we only write it once?

lib/compiler-llvm/src/abi/aarch64_systemv.rs
nlewycky: Done.

lib/compiler-llvm/src/abi/x86_64_systemv.rs
nlewycky: Done.

syrusakbary (approved):

MarkMcCaskey (approved):

PR #2063, opened 2021-01-27
Add `build-capi-all` command to Makefile for all compilers

MarkMcCaskey:
This is a useful command to have, maybe it should be the default?

There are still some issues with LLVM here (need to link zlib and curses) when linking against libwasmer.

Also LLVM is not detected on my computer by the Makefile even though it builds fine

# Review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,6 +155,10 @@ build-docs-capi:
 # We use cranelift as the default backend for the capi for now
 build-capi: build-capi-cranelift
 
+build-capi-all:
+	cargo build --manifest-path lib/c-api/Cargo.toml --release \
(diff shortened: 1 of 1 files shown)


syrusakbary:
Yeah, it might be a good idea to make it the default!

MarkMcCaskey:
> Yeah, it might be a good idea to make it the default!

I agree! but I think we need to fix the zlib and curses dependency issue before we can ship it to users. Otherwise, we'd be making it harder than it needs to be to build against wasmer

Hywan:
It doesn't build all the compilers, but the available compilers, is that right? We should add a line of documentation to specify that. Just a recommendation, nothing blocking.

PR #2064, opened 2021-01-27
chore: Do not try to copy libwasmer_c_api.so if it does not exist

jubianchi:
This will help resolving wasmerio/wasmer-nightly#9 and #2003

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,10 +2,12 @@
 ifneq ($(OS), Windows_NT)
 	ARCH := $(shell uname -m)
 	UNAME_S := $(shell uname -s)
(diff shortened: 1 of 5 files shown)


MarkMcCaskey (approved):
Looks good to me assuming you tested it locally!

Makefiles are weird about shell and it wouldn't surprise me if this change breaks something

syrusakbary:
Will the .so still be included in the distribution artifact for general Linux?

jubianchi:
@syrusakbary yes it will: if the `.so` is here, it will be packaged, if it's not, only the `.a` will be packaged.

jubianchi:
bors try

Hywan (approved):

PR #2068, opened 2021-01-28
Remove dependency on raw-cpuid in favor of is_x86_feature_detected

syrusakbary:
# Description

Remove dependency on raw-cpuid in favor of the macro `std:is_x86_feature_detected!`.

# Review

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1529,15 +1529,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "raw-cpuid"
-version = "9.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


syrusakbary:
bors try

Hywan (approved):
Perfect, thanks!

MarkMcCaskey (approved):
Looks good! 

The one thing I'd caution is the use of a string-based API with conditionals. It might be good to make sure we're actually executing those branches and that the strings are correct.

I don't know if we have a good way to test for things like this but something to watch out for with changes like this: it's very easy to introduce regressions here

syrusakbary:
Yeah, I agree with you Mark.
I’ve rechecked the strings twice.  There is no an ideal easy way to test it :(

I’ll merge now and we can think on extra ways of testing. If we see them we can open other PRs

PR #2069, opened 2021-01-28
doc(c-api) Use the last `README.md`, not the deprecated one

Hywan:
# Description

This patch updates the `README.md` used in the `include/` directory of the packaged C API. We must use the last `lib/c-api/README.md` rather than the `lib/c-api/doc/deprecated/index.md` file. That's a mistake.

Also, I wonder why `wasmer_c_api.{dll|lib)` on Windows isn't renamed to `wasmer.{dll|lib}` as we do for ther other platforms?

# Review


MarkMcCaskey (approved):
In regards to the naming on Windows: I remember asking that same question > 6-12 months ago and getting an answer from Syrus but I have no idea what the answer was. I think there's an actual reason though

Hywan:
cc @syrusakbary regarding my question in the first comment?

bors r+

PR #2071, opened 2021-01-28
feat(c-api) Rename the `wasmer` module to `unstable`

Hywan:
# Description

This patch is the first step to prepare more unstable API (like cross-compilation etc.). Nothing changes for the moment from the user perspective, it's not a breaking change. It's just a reorganisation of the code, and better documentation. Basically the `wasmer_c_api::wasm_c_api::wasmer` module has been renamed `unstable`, so that it's super clear :-).

# Review

### DIFF
--- a/lib/c-api/src/wasm_c_api/mod.rs
+++ b/lib/c-api/src/wasm_c_api/mod.rs
@@ -230,6 +230,12 @@ pub mod trap;
 /// cbindgen:ignore
 pub mod types;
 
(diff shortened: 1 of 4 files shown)


MarkMcCaskey (approved):
To be fair, I don't think `wasmer_wasm.h` is any less stable than `wasm.h`: it has breaking changes all the time.

So I'm not sure about how valuable marking this as "unstable" is, but it's fine

PR #2072, opened 2021-01-28
feat(c-api) Implement cross-compilation API in C

Hywan:
# Description

Depends on https://github.com/wasmerio/wasmer/pull/2071 (merged).

This patch implements `wasm_target_t`, `wasm_triple_t`, `wasm_cpu_features_t`.

This patch also implements `wasm_config_set_target` to set… a `wasm_target_t`. The `wasm_config_new_with_engine` function has been updated to handle the new `wasm_target_t`.

# Example

```c
int main() {
    // Create the configuration.
    wasm_config_t* config = wasm_config_new();

    // Set the target.
    {
      // Declare the target triple.
      wasm_triple_t* triple;
  
      {
          wasm_name_t triple_name;
          wasm_name_new_from_string(&triple_name, "x86_64-apple-darwin");
  
          triple = wasm_triple_new(&triple_name);
  
          wasm_name_delete(&triple_name);
      }
  
      assert(triple);
  
      // Declare the target CPU features.
      wasm_cpu_features_t* cpu_features = wasm_cpu_features_new();
  
      {
          wasm_name_t cpu_feature_name;
          wasm_name_new_from_string(&cpu_feature_name, "sse2");
  
          wasm_cpu_features_add(cpu_features, &cpu_feature_name);
  
          wasm_name_delete(&cpu_feature_name);
      }
  
      assert(cpu_features);
  
      // Create the target!
      wasm_target_t* target = wasm_target_new(triple, cpu_features);
      assert(target);
  
      // Set the target onto the configuration!
      wasm_config_set_target(config, target);
    }

    // Create the engine.
    wasm_engine_t* engine = wasm_engine_new_with_config(config);

    // Check we have an engine!
    assert(engine);

    // Free everything.
    wasm_engine_delete(engine);

    return 0;
}
```

# Review


lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -467,7 +484,13 @@ pub extern "C" fn wasm_engine_new_with_config(
                         // There are currently no uses of the object-file engine + compiler from the C API.
                         // So we run in headless mode.
                         if #[cfg(feature = "object-file")] {
-                            Arc::new(ObjectFile::headless().engine())
+                            let mut builder = ObjectFile::headless();
+
+                            if let Some(target) = config.target {
+                                builder = builder.target(target.inner);
```
MarkMcCaskey, line 490: Does it make sense to set the target for a headless-only engine?

It's fine for symmetry I suppose but I don't think this can ever do anything

lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -489,7 +518,13 @@ pub extern "C" fn wasm_engine_new_with_config(
...
                     cfg_if! {
                         if #[cfg(feature = "native")] {
-                            Arc::new(Native::headless().engine())
+                            let mut builder = Native::headless();
+
+                            if let Some(target) = config.target {
+                                builder = builder.target(target.inner);
+                            }
+
+                            Arc::new(builder.engine())
```
MarkMcCaskey, line 527: These ones too

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
```diff
@@ -0,0 +1,318 @@
...
+
+/// Creates a new [`wasm_target_t`].
+///
+/// It takes ownership of `triple` and `cpu_features`.
+///
+/// # Example
+///
+/// See the module's documentation.
+#[no_mangle]
+pub unsafe extern "C" fn wasm_target_new(
```
MarkMcCaskey, line 84: Probably doesn't need to be unsafe

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
```diff
@@ -0,0 +1,318 @@
...
+    }))
+}
+
+/// Delete a [`wasm_target_t`].
+///
+/// # Example
+///
+/// See the module's documentation.
+#[no_mangle]
+pub unsafe extern "C" fn wasm_target_delete(_target: Option<Box<wasm_target_t>>) {}
```
MarkMcCaskey, line 102: same, no need for unsafe

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
```diff
@@ -0,0 +1,318 @@
...
+///
+/// See [`wasm_triple_t`] or [`wasm_triple_new_from_host`].
+#[no_mangle]
+pub unsafe extern "C" fn wasm_triple_new(
+    triple: Option<&wasm_name_t>,
+) -> Option<Box<wasm_triple_t>> {
+    let triple = triple?;
+    let triple = c_try!(CStr::from_bytes_with_nul_unchecked(slice::from_raw_parts(
+        triple.data,
+        triple.size + 1
```
MarkMcCaskey, line 153: Why is this correct?

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
```diff
@@ -0,0 +1,318 @@
...
+///     return 0;
+/// }
+/// #    })
+/// #    .success();
+/// # }
+/// ```
+///
+/// See also [`wasm_triple_new`].
+#[no_mangle]
+pub unsafe extern "C" fn wasm_triple_new_from_host() -> Box<wasm_triple_t> {
```
MarkMcCaskey, line 187: Doesn't need to be unsafe

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
```diff
@@ -0,0 +1,318 @@
...
+    })
+}
+
+/// Delete a [`wasm_triple_t`].
+///
+/// # Example
+///
+/// See [`wasm_triple_t`].
+#[no_mangle]
+pub unsafe extern "C" fn wasm_triple_delete(_triple: Option<Box<wasm_triple_t>>) {}
```
MarkMcCaskey, line 199: same

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
```diff
@@ -0,0 +1,318 @@
...
+    inner: EnumSet<CpuFeature>,
+}
+
+/// Create a new [`wasm_cpu_features_t`].
+///
+/// # Example
+///
+/// See [`wasm_cpu_features_t`].
+#[no_mangle]
+pub unsafe extern "C" fn wasm_cpu_features_new() -> Box<wasm_cpu_features_t> {
```
MarkMcCaskey, line 269: same

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
```diff
@@ -0,0 +1,318 @@
...
+    })
+}
+
+/// Delete a [`wasm_cpu_features_t`].
+///
+/// # Example
+///
+/// See [`wasm_cpu_features_t`].
+#[no_mangle]
+pub unsafe extern "C" fn wasm_cpu_features_delete(_cpu_features: Option<Box<wasm_cpu_features_t>>) {
```
MarkMcCaskey, line 281: same

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
```diff
@@ -0,0 +1,318 @@
...
+/// Add a new CPU feature into the set represented by
+/// [`wasm_cpu_features_t`].
+///
+/// # Example
+///
+/// See [`wasm_cpu_features_t`].
+#[no_mangle]
+pub unsafe extern "C" fn wasm_cpu_features_add(
+    cpu_features: Option<&mut wasm_cpu_features_t>,
+    feature: Option<&wasm_name_t>,
```
MarkMcCaskey, line 293: I don't think we need `Option` around all user-supplied pointers, the standard wasm-c-api doesn't bother with that in a lot of cases, if the pointer is invalid it'll crash or do what C does

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
```diff
@@ -0,0 +1,318 @@
...
+        _ => return false,
+    };
+    let feature = match feature {
+        Some(feature) => feature,
+        _ => return false,
+    };
+    let feature = c_try!(
+        CStr::from_bytes_with_nul_unchecked(slice::from_raw_parts(
+            feature.data,
+            feature.size + 1,
```
MarkMcCaskey, line 306: I'm skeptical of this

MarkMcCaskey (approved):
Looks pretty good overall

lib/c-api/src/wasm_c_api/engine.rs
Hywan: Yeah I was wondering if that was making sense. Using the target makes sense _only_ when compiling, but we never know, maybe one day, someone, somewhere… :-p. Let's keep it for “the symmetry” as you said, how does it sound?

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
Hywan: Resolved. Thanks.

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
Hywan: Solved, thanks!

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
Hywan: 🤦 

`CStr` expects a C string, a string that is nul-terminated. It's not the case here. I've updated the code to use `str::from_utf8(slice::from_raw_parts(triple.data, triple.size))`, which is the correct way to use `wasm_name_t` :-).

Thanks!

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
Hywan: Yup, solved!

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
Hywan: Yeah but I prefer to prevent bad usage of our API. That's what we already do everywhere almost.

lib/c-api/src/wasm_c_api/unstable/target_lexicon.rs
Hywan: Fixed!

Hywan:
bors r-

waiting on https://github.com/wasmerio/wasmer/pull/2083 to finish

PR #2073, opened 2021-01-29
Initial commit of new MemoryUsage trait.

nlewycky:


syrusakbary:
Can we name it memoryprofiler or similar? It's a bit confusing to refer to it as wasmer-memory (since Wasm already has a memory concept)

nlewycky:
> Can we name it memoryprofiler or similar? It's a bit confusing to refer to it as wasmer-memory (since Wasm already has a memory concept)

Good point! Done.

lib/memoryprofiler/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "wasmer-memoryprofiler"
```
MarkMcCaskey, line 2: I think we should consider not including it in this repo and not naming it anything to do with Wasmer. It really is a standalone utility that could be generally useful and it shouldn't follow our versioning scheme for the rest of Wasmer.

1.0 implies stability and a new thing we just made is likely not 1.0-stable especially because we want to merge early and often into master and release whenever

lib/memoryprofiler/derive/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+) -> proc_macro::TokenStream {
+    let lifetimes_and_generics = &generics.params;
+    let where_clause = &generics.where_clause;
+    let sum = match &data.fields {
+        Fields::Named(ref fields) => fields
+            .named
+            .iter()
+            .map(|field| {
+                let id = field.ident.as_ref().unwrap();
+                quote! { MemoryUsage::size_of(&self.#id) }
```
MarkMcCaskey, line 37: You can use `quote_spanned` and get the span from the field so that way if there's an error message (for example one field does not implement `MemoryUsage`) the compiler will point to the line containing that field and it'll show that field in the error message

lib/memoryprofiler/derive/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+            })
+            .collect(),
+        Fields::Unit => vec![],
+        Fields::Unnamed(fields) => (0..(fields.unnamed.iter().count()))
+            .into_iter()
+            .map(|field| {
+                let id = syn::Index::from(field);
+                quote! { MemoryUsage::size_of(&self.#id) }
```
MarkMcCaskey, line 45: Maybe there's an equivalent of `quote_spanned` for this too

lib/memoryprofiler/derive/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+    .fold(quote! { 0 }, |x, y| quote! { #x + #y });
+
+    (quote! {
+        #[allow(dead_code)]
+        impl < #lifetimes_and_generics > MemoryUsage for #struct_name < #lifetimes_and_generics > #where_clause {
+            fn size_of(&self) -> usize {
+                #sum
+            }
+        }
+    })
```
MarkMcCaskey, line 60: There's another thing where you can provide a dummy quoted section of code for error messages (so code that looks like what you're expanding but with none of the specifics); might be useful to have; if this is its own repo, it'll be easier to test the proc-macro error messages. The error message testing I got working with `wasmer-derive` works great but requires `cargo clean` after every run, so we don't test it in CI and don't test it locally either because that's a big burden to do in this project.

lib/memoryprofiler/derive/tests/basic.rs
```diff
@@ -0,0 +1,23 @@
+use wasmer_memoryprofiler::r#trait::MemoryUsage;
+use wasmer_memoryprofiler_derive::MemoryUsage;
```
MarkMcCaskey, line 2: Mentioned it on slack but I'll say it again here: Rust doesn't care about name collisions with derive macros and traits. In fact it's actually a big feature if they're the same because then you can just `use` one thing and you get both in scope!

MarkMcCaskey: (Not strictly relevant for this file, but generally it's useful; i.e. the trait shouldn't be publicly exported under `r#trait`)

lib/memoryprofiler/src/trait.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+// pointer
+// Pointers aren't necessarily safe to dereference, even if they're nonnull.
+
+// references
+impl<T: MemoryUsage> MemoryUsage for &T {
+    fn size_of(&self) -> usize {
+        (**self).size_of()
```
MarkMcCaskey, line 83: I believe we also want to account for the memory usage of the `&` itself, which would be `usize`

lib/memoryprofiler/src/trait.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+impl<T: MemoryUsage> MemoryUsage for &mut T {
+    fn size_of(&self) -> usize {
+        (**self).size_of()
```
MarkMcCaskey, line 88: same here

lib/memoryprofiler/src/trait.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+// slices
+impl<T: MemoryUsage> MemoryUsage for [T] {
+    fn size_of(&self) -> usize {
+        self.iter().map(MemoryUsage::size_of).sum()
```
MarkMcCaskey, line 95: Plus 2 `usize` for the pointer and length of the slice

lib/memoryprofiler/src/trait.rs
```diff
@@ -0,0 +1,169 @@
...
+impl<T: MemoryUsage, const N: usize> MemoryUsage for [T; N] {
+    fn size_of(&self) -> usize {
+        self.iter().map(MemoryUsage::size_of).sum()
+    }
+}
+
+// strs
+impl MemoryUsage for str {
+    fn size_of(&self) -> usize {
+        self.as_bytes().size_of()
```
MarkMcCaskey, line 109: Hmm okay this is interesting, will need testing to make sure we don't over or under count but I think `str` is the same as `[u8]`; meaning I'm not sure what it's overhead is: `&str` and `&[u8]` should have an extra 2 `usize` on the stack though!

lib/memoryprofiler/src/trait.rs
```diff
@@ -0,0 +1,169 @@
...
+//impl<T: MemoryUsage> MemoryUsage for Pin<T> {
+//}
+
+// TODO: Mutex
+
+// TODO: NonNull might be possible when '*const T' is MemoryUsage.
+
+impl<T: MemoryUsage> MemoryUsage for Option<T> {
+    fn size_of(&self) -> usize {
+        std::mem::size_of::<Option<()>>() + self.iter().map(MemoryUsage::size_of).sum::<usize>()
```
MarkMcCaskey, line 134: I don't know if this is correct, Option does optimizations depending on what it's holding so for example

`NonNull<T>` and `Option<NonNull<T>>` are the same size

MarkMcCaskey: Alright lets break it down: we know in the `None` case it's always just the size of the stack space taken up so `std::mem::size_of::<Option<T>>()` is all we need.

In the `Some` case we need to know the memory usage of the inner type because of heap usage and then `std::mem::size_of::<Option<T>>()`: something I've been thinking about with this trait is having another method and trying to make a distinction between the static size allocated and the dynamic size.

We might actually need that to make that work. A neat thing we can do for the static one is basically just have it be `std::mem::size_of::<T>()` where it takes the argument by generic type.

I don't think we can do `MemoryUsage: Sized` and rely on `std::mem::size_of` for this though because that would prevent this trait from being derived by other traits that are then used as trait objects: the desired behavior we want for traits with no real stack size is to say that they don't have one (this applies for unsized types like `[u8]` too)


A bit stream of consciousness but ping me if you want to discuss anything about this today

lib/memoryprofiler/src/trait.rs
```diff
@@ -0,0 +1,169 @@
...
+// TODO: Ref, RefCell, RefMut
+
+//impl<T: MemoryUsage> MemoryUsage for Result<T> {
+//}
+
+// TODO: RwLock
+
+// string?
+
+// TODO: UnsafeCell
```
MarkMcCaskey, line 149: Yeah there's actually a ton of types but we can always add them incrementally!

For example https://doc.rust-lang.org/std/num/struct.NonZeroU8.html

and the rest of std::collections

MarkMcCaskey: Depending on our needs for Wasmer, too, we may also need to add features to this crate that conditionally implement this trait on third party types too, due to the orphan rule.

So for example if we use crossbeam::Mutex anywhere (which we might) then the only place this trait can be implemented is in this crate or in `crossbeam`. So the standard way to handle that is to have a feature called `crossbeam` that makes this crate depend on crossbeam and implement the trait for those types.

MarkMcCaskey (commented):
Looks very good! Looks like you have a solid start on it!

There's some improvements we can do for macro error reporting and some small details to work out with the trait itself but I think you're absolutely on the right path!

lib/memoryprofiler/Cargo.toml
nlewycky: I absolutely agree with you, but I don't know where else to put it. A new repo in github? Or as a directory in wasmer/ which isn't part of the wasmer workspace?

nlewycky:
FYI, related: https://github.com/dtolnay/quote/blob/master/src/lib.rs#L209-L229

lib/memoryprofiler/derive/src/lib.rs
nlewycky: Done.

Do you know any way to do this for `quote! ( x + y )` where I have one span for `x` and another for `y`?

lib/memoryprofiler/derive/src/lib.rs
nlewycky: I looked but couldn't find anything.

lib/memoryprofiler/derive/tests/basic.rs
nlewycky: I'm not sure what you're suggesting. Ideally I would like the trait to be `MemoryUsage` so that you can `impl MemoryUsage for Whatever` and also for the proc_macro to be used as `#[derive(MemoryUsage)]`. The trait needs to be public so that people can impl it for their own types. I'm not sure how to arrange it so that you can use one thing and get both types when they have to be in separate crates because proc macro crates much be separate from everything else.

lib/memoryprofiler/src/trait.rs
nlewycky: I've replaced this with:
```
    fn size_of(&self) -> usize {
        std::mem::size_of::<&T>()
    }
```
because if we visit the referred-to object then it may be double-counted with anyone else holding the immutable reference. For mutable references the same logic applies even though there can only be one mutable reference it still has to be referring to an object which is owned by somebody elsewhere.

lib/memoryprofiler/src/trait.rs
nlewycky: Done.

lib/memoryprofiler/src/trait.rs
nlewycky: I think `std::mem::size_of::<Option<T>>() + self.iter().map(MemoryUsage::size_of).sum::<usize>()` gets exactly what we want here? The size of the Option including NonNull optimization if applicable, then visit the member if it's holding one?

lib/memoryprofiler/src/trait.rs
nlewycky: I think it's clear that we should include an implementation for the `std::collections`. I initially though this crate would be separate from Wasmer, but once I thought about the orphan rule I realized that we might want it to be an implemention of the trait for all the types used in Wasmer which somewhat attaches it to wasmer.

Or we can use features as you suggest.

lib/memoryprofiler/Cargo.toml
Hywan: https://github.com/wasmerio/loupe

Hywan:
Closed since we have been doing all this in [the `loupe` crate](https://github.com/wasmerio/loupe/).

PR #2074, opened 2021-01-29
feat(package) Be consistent on Windows regarding `libwasmer` name.

Hywan:
# Description

Windows package libraries aren't named consistently regarding other platforms. This patch fixes that.

# Review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -378,8 +378,8 @@ package-capi:
 	cp lib/c-api/wasm.h* package/include
 	cp lib/c-api/README.md package/include/README.md
 ifeq ($(OS), Windows_NT)
-	cp target/release/wasmer_c_api.dll package/lib
-	cp target/release/wasmer_c_api.lib package/lib
+	cp target/release/wasmer_c_api.dll package/lib/wasmer.dll
+	cp target/release/wasmer_c_api.lib package/lib/wasmer.lib
 else
 ifeq ($(UNAME_S), Darwin)
 	# For some reason in macOS arm64 there are issues if we copy constantly in the install_name_tool util


Makefile
```diff
@@ -378,8 +378,8 @@ package-capi:
 	cp lib/c-api/wasm.h* package/include
 	cp lib/c-api/README.md package/include/README.md
 ifeq ($(OS), Windows_NT)
-	cp target/release/wasmer_c_api.dll package/lib
-	cp target/release/wasmer_c_api.lib package/lib
+	cp target/release/wasmer_c_api.dll package/lib/libwasmer.dll
```
jubianchi, line 381: AFAIR on Windows we should not add the `lib` prefix in file names.

Hywan: Correct! Fixed.

jubianchi (approved):

Makefile
```diff
@@ -378,8 +378,8 @@ package-capi:
 	cp lib/c-api/wasm.h* package/include
 	cp lib/c-api/README.md package/include/README.md
 ifeq ($(OS), Windows_NT)
-	cp target/release/wasmer_c_api.dll package/lib
-	cp target/release/wasmer_c_api.lib package/lib
+	cp target/release/wasmer_c_api.dll package/lib/wasmer.dll
+	cp target/release/wasmer_c_api.lib package/lib/wasmer.lib
```
MarkMcCaskey, line 382: It'd be good to check with Syrus on the name inconsistency, if I remember correctly they've always had a different name than other platforms and I believe I've verified that that was intentional in the past

syrusakbary: I just re-checked. This looks good!

syrusakbary (approved):

syrusakbary:
Merging although is going to cause `create-exe` fail temporarily. I'll fix that in master directly

PR #2082, opened 2021-02-01
Define shims for stack probes using naked functions with RFC 2873 style asm!

### LINKED ISSUE #1722: probestack broken with engine-native and compiler-llvm
# Describe the bug

When building with --llvm and --native, `wast::llvm::spec::skip_stack_guard_page` fails with "out of bounds memory access" instead of the expected "call stack exhausted".

Let's take a quick look at one example:
```
(lldb) disassemble -s 0x7ffff77a0061
wasmer_nativexpTE17.so`wasmer_function__1:
    0x7ffff77a0061 <+1>:  pushq  %r15
    0x7ffff77a0063 <+3>:  pushq  %r14
    0x7ffff77a0065 <+5>:  pushq  %r13
    0x7ffff77a0067 <+7>:  pushq  %r12
    0x7ffff77a0069 <+9>:  pushq  %rbx
    0x7ffff77a006a <+10>: movl   $0x2098, %eax             ; imm = 0x2098 
    0x7ffff77a006f <+15>: callq  0x7ffff77a0020            ; symbol stub for: wasmer_probestack
    0x7ffff77a0074 <+20>: subq   %rax, %rsp
    0x7ffff77a0077 <+23>: movq   0x8(%rdi), %rbx
    0x7ffff77a007b <+27>: callq  0x7ffff77a0030            ; symbol stub for: wasmer_function__1
```

That wasmer_probestack stub is a PLT entry which looks like this:
```
(lldb) disassemble -s 0x7ffff77a0020
wasmer_nativexpTE17.so`wasmer_probestack:
    0x7ffff77a0020 <+0>:  jmpq   *0x9ffa(%rip)             ;  + 32
    0x7ffff77a0026 <+6>:  pushq  $0x1
    0x7ffff77a002b <+11>: jmp    0x7ffff77a0000
```
(you can ignore everything but the first instruction, __rust_probestack ends in a retq which will tail call return past this frame back to 0x7ffff77a006f)

That jmpq resolves to `0x000055555bde4148` which looks like bad code:
```
* thread #2, name = 'wast::llvm::spe', stop reason = instruction step into
    frame #0: 0x000055555bde4148
->  0x55555bde4148: jp     0x55555bde40df
    0x55555bde414a: movl   %ebx, 0x55(%rbx)
    0x55555bde414d: pushq  %rbp
    0x55555bde414e: addb   %al, (%rax)
```
but note that it is indeed the address of wasmer_probestack!
```
(lldb) p &wasmer_probestack
(void (**)()) $38 = 0x000055555bde4148
(lldb) p wasmer_probestack
(void (*)()) $39 = 0x000055555b89957a (compilers-95b529d63c259da9`__rust_probestack)
```
Somehow we jumped to the address of a global variable pointing to __rust_probestack instead of calling directly into __rust_probestack.

nlewycky:
This should fix #1722.

It's not clear whether we can merge this, new features are enabled in the lib.rs


lib/vm/src/libcalls.rs
```diff
@@ -399,8 +398,53 @@ pub unsafe extern "C" fn wasmer_raise_trap(trap_code: TrapCode) -> ! {
...
+            asm!("jmp {}", sym __rust_probestack, options (noreturn));
+        }
+    } else {
+        /// Probestack check is not used on this platform.
+        #[no_mangle]
+        pub unsafe extern "C" fn wasmer_probestack() {
+            panic!();
+        }
+    }
+}
```
syrusakbary, line 447: It might be a good idea to move this to the probestack file, if possible

syrusakbary (commented):
It would be a good idea to re-enable the skipped test

Hywan:
:+1:

lib/vm/src/libcalls.rs
```diff
@@ -399,8 +398,53 @@ pub unsafe extern "C" fn wasmer_raise_trap(trap_code: TrapCode) -> ! {
...
 /// This function does not follow the standard function ABI, and is called as
 /// part of the function prologue.
-#[no_mangle]
-pub static wasmer_probestack: unsafe extern "C" fn() = PROBESTACK;
+cfg_if::cfg_if! {
+    if #[cfg(all(
+            target_os = "windows",
+            target_env = "msvc",
+            target_pointer_width = "64"
+            ))] {
```
MarkMcCaskey, line 406: And if we want to ship this for nigthly-only, we should feature gate this code based on nightly vs not nightly

syrusakbary (commented):
Since this is in the VM layer, and we already have C code in VM layer, perhaps we can bypass the "asm" feature if the user is not in nightly when compiling by using a asm function implemented in C.

nlewycky:
> Since this is in the VM layer, and we already have C code in VM layer, perhaps we can bypass the "asm" feature if the user is not in nightly when compiling by using a asm function implemented in C.

C proper, as in official standard C, doesn't permit assembly in it either. That said, gcc and clang support assembly in C files for linux and mac with the same syntax. We'd need to do something different for Windows though.

PR #2084, opened 2021-02-01
fix(c-api) Avoid more than one call to the function environment finalizer

Hywan:
# Description

A function environment `WrapperEnv` can be cloned. In case the
original and the cloned environments are being dropped, the
`self.finalizer` function —if any— is called twice. That's a bug. It
must be called only once.

This patch updates the code to apply the finalizer only once by taking
it —if any— the first time it's called.

This bug has been raised by @jubianchi when working with PHP and the Zend Engine GC.

# Review


MarkMcCaskey (approved):

jubianchi (approved):
Approved & tested agains `wasmer-php`

PR #2086, opened 2021-02-01
Make the memory.{grow,size} library calls libcalls instead of pointer-to-function fields on the Instance.

nlewycky:
For now this keeps the fields since they are still used by cranelift and singlepass, but it ports llvm over to calling the functions directly.

### DIFF
--- a/lib/vm/src/libcalls.rs
+++ b/lib/vm/src/libcalls.rs
@@ -137,6 +137,7 @@ pub extern "C" fn wasmer_f64_nearest(x: f64) -> f64 {
 /// # Safety
 ///
 /// `vmctx` must be valid and not null.
+#[no_mangle]
 pub unsafe extern "C" fn wasmer_memory32_grow(
     vmctx: *mut VMContext,
     delta: u32,
@@ -156,6 +157,7 @@ pub unsafe extern "C" fn wasmer_memory32_grow(
 /// # Safety
 ///
 /// `vmctx` must be valid and not null.
+#[no_mangle]
 pub unsafe extern "C" fn wasmer_imported_memory32_grow(
     vmctx: *mut VMContext,
     delta: u32,
@@ -175,6 +177,7 @@ pub unsafe extern "C" fn wasmer_imported_memory32_grow(
 /// # Safety
 ///
 /// `vmctx` must be valid and not null.
+#[no_mangle]
 pub unsafe extern "C" fn wasmer_memory32_size(vmctx: *mut VMContext, memory_index: u32) -> u32 {
     let instance = (&*vmctx).instance();
     let memory_index = LocalMemoryIndex::from_u32(memory_index);
@@ -187,6 +190,7 @@ pub unsafe extern "C" fn wasmer_memory32_size(vmctx: *mut VMContext, memory_inde
 /// # Safety
 ///
 /// `vmctx` must be valid and not null.
+#[no_mangle]
 pub unsafe extern "C" fn wasmer_imported_memory32_size(
     vmctx: *mut VMContext,
     memory_index: u32,
@@ -435,8 +439,20 @@ pub enum LibCall {
     /// trunc.f32
     TruncF32,
 
-    /// frunc.f64
+    /// trunc.f64
     TruncF64,
+
+    /// memory.grow for a locally defined memory
+    Memory32Grow,
+
+    /// memory.grow for an imported memory
+    ImportedMemory32Grow,
+
+    /// memory.size for a locally defined memory
+    Memory32Size,
+
+    /// memory.size for an imported memory
+    ImportedMemory32Size,
 }
 
 impl LibCall {
@@ -453,6 +469,10 @@ impl LibCall {
             Self::RaiseTrap => wasmer_raise_trap as usize,
             Self::TruncF32 => wasmer_f32_trunc as usize,
             Self::TruncF64 => wasmer_f64_trunc as usize,
+            Self::Memory32Grow => wasmer_memory32_grow as usize,
+            Self::ImportedMemory32Grow => wasmer_imported_memory32_grow as usize,
+            Self::Memory32Size => wasmer_memory32_size as usize,
+            Self::ImportedMemory32Size => wasmer_imported_memory32_size as usize,
         }
     }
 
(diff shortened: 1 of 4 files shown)


MarkMcCaskey (approved):
looks good to me

lib/vm/src/libcalls.rs
```diff
@@ -453,6 +469,10 @@ impl LibCall {
             Self::RaiseTrap => wasmer_raise_trap as usize,
             Self::TruncF32 => wasmer_f32_trunc as usize,
             Self::TruncF64 => wasmer_f64_trunc as usize,
+            Self::Memory32Grow => wasmer_memory32_grow as usize,
```
syrusakbary, line 472: This is a bit confusing.
Were this libcalls not available before? How cranelift was using them?

nlewycky: They were not libcalls before! This PR doesn't change how singlepass or cranelift use them, which is to use the "vm builtins" mechanism where every instance has pointers-to these functions as fields in the instance. See translate_memory_grow in codege-cranelift's func_environ.rs:
```
        let (vmctx, func_addr) = self.translate_load_builtin_function_address(&mut pos, func_idx);
        let call_inst = pos
            .ins()
            .call_indirect(func_sig, func_addr, &[vmctx, val, memory_index]);
```
`translate_load_builtin_function_address` will load the function pointer out of the instance itself (using VMBuiltinFunctionIndex on VMOffsets` and then return that function pointer.

IMO the direction should be that we remove the VMBuiltinFunction system entirely, unless you know of a good reason why we want to have function for different instances at runtime and don't know this at compile time.

syrusakbary:
I'm lacking context. To me it seems useful that this functions `memory_grow`, `memory_size`, etc. belong to `VMBuiltinFunctionIndex` so they can be "parched" at runtime when instantiating the module. At least theoretically.

What is the reasoning behind this change?

nlewycky:
> I'm lacking context. To me it seems useful that this functions `memory_grow`, `memory_size`, etc. belong to `VMBuiltinFunctionIndex` so they can be "parched" at runtime when instantiating the module. At least theoretically.
> 
> What is the reasoning behind this change?

It started with "what are these cached_memory_grow and cached_memory_size in LLVM". The're hashmaps that are indexed from memory index to address-of-memory.grow or memory.size function. Well that's odd, because there's only two memory.grow functions, one for locally defined and one for imported. So why a hash table? In fact, calculating a hash is probably slower than just checking whether a MemoryIndex is local or imported so this is just slowing down compilation.

After some digging I read through the "VM builtins" and I think we should remove it. VMBuiltinFunctionIndex is more complicated to understand how it works, uses up more runtime memory, and requires more memory accesses and slower code emitted by the compilers as compared to libcalls which can do the exact same things. But before I merge this PR, I do want to check my understanding with you first, we should make sure we agree that we both understand the code here.

In favour of "VM builtins", they are optimized for the case where the functions are different on each instance and the compiler can't possibly know what function will be called -- not even the function name. This could make sense if we want to switch to a different implementation of `memory.grow` after the module is not only compiled, but after it is instantiated. We could go to instance.builtin_functions_ptr[some_offset] = &different_function.

But I simply can't see why we would ever want to do that? This seems like unnecessary complexity.

With "libcalls", when you want to call the function, you call it. If we want different behaviour for different instances, the libcall implementation would have to look at the data in the instance (through the `* const vmctx` it receives as first parameter) to decide what to do.

To make this clear, I'll show the assembly. Our .wasm file:
```
(module
    (memory 0)
    (func (export "size") (result i32) (memory.size 0))
)
```
produces this code with VM builtins:
```
        pushq   %rax
        xorl    %esi, %esi
        callq   *48(%rdi)
        popq    %rcx
        retq
```
and this code with libcall:
```
        pushq   %rax
        xorl    %esi, %esi
        callq   wasmer_memory32_size
        popq    %rcx
        retq
```
The difference is that the former is using %rdi (the vmctx pointer) plus 0x48, load 64-bits from there, then call that address. The latter is calling the function named wasmer_memory32_size. For instance `dlopen` will push the address of `wasmer_memory32_size` directly into the instruction stream as a relocation. In both cases they get the vmctx pointer as first argument and 0 for memory index as second argument.

This is what we already do for "wasmer_raise_trap" for example.

Hywan (approved):
Looks good to me!

syrusakbary:
Feedback after having some thoughts and reading your analysis.

It's true that `VMBuiltinFunction` is only needed for functions who's address may change after compiling the module.

That is, for example:
```
module = Module(store, bytes)

instance1 = module.instantiate()
instance2 = module.instantiate()

# We want instance1 and instance2 to have different addresses for some functions (that are not imported nor wasm-defined, nor a math library).
# Or we want instance1 to have a different address for some functions once is already instantiated
```

I think the reason for `VMBuiltinFunction` existence is more philosophical: while truncate a f64 is a function that will need to stay constant to assure spec viability (that means, that it has only one valid implementation), `memory_grow` or others can have multiple implementations depending on the VM layer.

Having said this, in the practical level, how much we expect this to change:
1. Changing the address of VM functions once a module has been compiled
2. Changing the address of VM functions once a module has been instantiated

Either for 1. or 2. we need the concept of `VMBuiltinFunction`...so I think the main question relies on: is there any case where 1 or 2 is going to be needed?

nlewycky:
> Either for 1. or 2. we need the concept of VMBuiltinFunction

I don't think we do! If we ever want to add the situation where `memory.grow` works in two ways on two different instances, we could put that data in a field on the instance and read that in function that is called. Of course that's less direct, but just to answer the point about need, literally.

This philosophical rationale is exactly why I wanted to check with you before moving ahead. I'm certain the functionality can be the same either way, but I identified that there might be a philosophical reason why this is the most straight-forward way to do what it's doing.

Ultimately I think that `memory.grow` is the same as `f32.truncate`, and that's how we ought to be thinking about it when thinking about the VM and the compiled-code <-> VM interface. Even in the future when we want 64-bit memories that are different from our 32-bit memories this will have to be shown in the module, so we'll know when compiling, and can call a new function.

> The relocations happen at "Module" creation time, not instantiation.

Fair enough! What I meant to convey was "each instantiation has the same function pointer" because it's on the Artifact.

PR #2089, opened 2021-02-02
chore: Update LLVM in the documentation workflow

jubianchi:
The documentation workflow is broken since we updated to LLVM 11. This PR fixes that.

### DIFF
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -30,7 +30,7 @@ jobs:
           echo "LLVM_SYS_100_PREFIX=${{ env.LLVM_DIR }}" >> $GITHUB_ENV
         env:
           LLVM_DIR: ${{ github.workspace }}/llvm-10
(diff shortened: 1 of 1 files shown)


Hywan (approved):

.github/workflows/documentation.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
           echo "LLVM_SYS_100_PREFIX=${{ env.LLVM_DIR }}" >> $GITHUB_ENV
         env:
           LLVM_DIR: ${{ github.workspace }}/llvm-10
```
Hywan, line 32:
```suggestion
          LLVM_DIR: ${{ github.workspace }}/llvm-11
```

.github/workflows/documentation.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
           echo "LLVM_SYS_100_PREFIX=${{ env.LLVM_DIR }}" >> $GITHUB_ENV
```
Hywan, line 30: Should we use `LLVM_SYS_110_PREFIX` here?

PR #2090, opened 2021-02-02
fix(c-api) Call `wasi_env_delete` manually

Hywan:
# Description

`wasi_get_imports` isn't taking ownership of `wasi_env_t` (despites
what is written in the documentation). And it must not take ownership
of it, since one could use it with the `wasi_env_read_stdout` &
sibling functions after having called `wasi_get_imports`.

Consequently, this patch calls `wasi_env_delete` to avoid leaking
memory.

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - [#2056](https://github.com/wasmerio/wasmer/pull/2056) Change back to depend on the `enumset` crate instead of `wasmer_enumset`
 
 ### Fixed
+- [#2090](https://github.com/wasmerio/wasmer/pull/2090) `wasi_env_t` needs to be freed with `wasi_env_delete` in the C API.
 - [#2084](https://github.com/wasmerio/wasmer/pull/2084) Avoid calling the function environment finalizer more than once when the environment has been cloned in the C API.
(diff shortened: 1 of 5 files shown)


MarkMcCaskey (commented):
Posted on Slack, but I think this is not symmetric with what we do in Rust and that the `wasi_env_t` is invalid so it's a bit odd to require manually deleting it in my opinion

Hywan:
We don't take ownership of `wasi_env_t`. Just a reference:

https://github.com/wasmerio/wasmer/blob/3ed1e47add8ec3e477a0bde4dbdaf698322e32cf/lib/c-api/src/wasm_c_api/wasi/mod.rs#L536-L542

And we just clone the inner value:

https://github.com/wasmerio/wasmer/blob/3ed1e47add8ec3e477a0bde4dbdaf698322e32cf/lib/c-api/src/wasm_c_api/wasi/mod.rs#L564

So `WasiEnv` is still available and fully working :-).

MarkMcCaskey (approved):
Alright, if it works then maybe that's a feature. The docs on the `wasi_env_t` type itself might need to be updated, I suspect they probably say that it does not need to be deleted.

Hywan:
Documentation has always been updated for `wasi_get_imports` and `wasi_get_unordered_imports`.

Hywan:
I believe these is something bigger happening here. cc @MarkMcCaskey

PR #2092, opened 2021-02-03
Add reference types

MarkMcCaskey:
This PR adds support for reference types in the API and in our compilers.

Wasm C API support should probably be a separate PR after this one lands.

# Status
- [x] spectests passing
  - [x] compiler-cranelift
  - [x] compiler-singlepass
  - [x] compiler-llvm
- [x] Tables work with references
- [x] Globals work with references
- [x] FuncRefs for local functions
- [x] FuncRefs for imported functions
- [ ] ExternRefs
  - [x] Work as values in Wasm
  - [x] Tests showing it being passed to/from host functions
  - [x] Top level API for interacting with ExternRefs
  - [ ] Reference counting works as expected, fully tested
- [ ] ~~Wasm C API support (should be done on a PR branched off of this branch)~~ (not going to do this in this PR)
- [x] ~Host info (see Wasm C API and wasmer Rust API)~ (not going to do this in this PR. The value of this feature seems low and the cost seems high; unclear if we even want it at the Rust API level)

# Review

# misc


lib/api/src/types.rs
```diff
@@ -19,56 +19,60 @@ pub type Val = Value<Function>;
 impl StoreObject for Val {
     fn comes_from_same_store(&self, store: &Store) -> bool {
         match self {
-            Self::FuncRef(f) => Store::same(store, f.store()),
+            Self::FuncRef(None) => true,
+            Self::FuncRef(Some(f)) => Store::same(store, f.store()),
             Self::ExternRef(ExternRef::Ref(_)) | Self::ExternRef(ExternRef::Other(_)) => false,
-            Self::ExternRef(ExternRef::Null) => true,
+            Self::ExternRef(ExternRef::Null) => todo!("update this code"),
```
syrusakbary, line 25: I think if it's a null extern ref we can assume it comes from the same store (at least on the way comes from the same store is currently used).

lib/vm/src/vmcontext.rs
```diff
@@ -866,9 +885,25 @@ impl VMBuiltinFunctionIndex {
...
+        Self(16)
+    }
+    /// Returns an index for wasm's `table.grow` instruction for imported tables.
+    pub const fn get_imported_table_grow_index() -> Self {
+        Self(17)
+    }
     /// Returns the total number of builtin functions.
     pub const fn builtin_functions_total_number() -> u32 {
-        14
+        18
```
syrusakbary, line 906: I'm curious: perhaps there is a way to automate this in safe rust? (so we don't need to do it manually each time we add fields)

MarkMcCaskey: Yeah, I briefly thought about that too. We can definitely do it with a macro, in a number of ways. I think Nick may have removed this code though or maybe the PR is still open. I believe this is the thing he was considering dropping in favor of the other, very similar thing

lib/api/src/types.rs
MarkMcCaskey: Oh yeah, for sure!  I think the reason I did that is because I intend to delete `ExternRef::Null` and pretty much entirely redo how `ExternRef` works! I just haven't started that process yet.

The existing code used `Val::ExternRef(ExternRef::Null)` for null funcrefs too, I needed to update all uses of that because that's now outdated so I made code paths using that panic with `todo!`. This is very much in-development code, I just got went off on 15 tangents while implementing this and haven't come back there yet!

MarkMcCaskey: Following up on that, I don't think most of this PR is reviewable yet. There are a few things that I think are close enough to how I want them to look that they could be reviewed, but most of it is just figuring things out and I expect substantial changes to most of the code in this PR so far. 

All that is why I delayed making a PR for as long as a I have -- I only made a PR this morning to make it easier for Nick to give me feedback on some of the compiler-llvm code I was writing!

lib/vm/src/vmcontext.rs
```diff
@@ -762,6 +792,17 @@ pub struct VMCallerCheckedAnyfunc {
     // If more elements are added here, remember to add offset_of tests below!
 }
 
+impl VMCallerCheckedAnyfunc {
+    /// Check if this anyfunc is a null funcref.
+    // TODO: we probably want to clean this up before shipping, maybe actually add
+    // an extra layer of indirection
```
MarkMcCaskey, line 798: I did add that layer of indirection, this is a really out of date comment.

Note to self, clean this up

lib/vm/src/table.rs
```diff
@@ -102,11 +104,58 @@ pub trait Table: fmt::Debug + Send + Sync {
...
+    }
+}
+
+// TODO: review what repr(C) means on unions, make sure this just acts like a usize.
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub union TableElement {
+    pub(crate) extern_ref: VMExternRef,
+    pub(crate) func_ref: VMFuncRef,
+}
```
syrusakbary, line 132: I'm unable to see the difference between `TableReference` and `TableElement` (other than the internal representation).
We should probably simplify this.

lib/api/src/externals/function.rs
```diff
@@ -651,16 +673,14 @@ impl Function {
...
-    pub(crate) fn checked_anyfunc(&self) -> VMCallerCheckedAnyfunc {
-        let vmsignature = self
-            .store
-            .engine()
-            .register_signature(&self.exported.vm_function.signature);
-        VMCallerCheckedAnyfunc {
+    pub(crate) fn checked_anyfunc(&self) -> VMFuncRef {
+        let engine = self.store.engine();
+        let vmsignature = engine.register_signature(&self.exported.vm_function.signature);
+        engine.register_function_metadata(VMCallerCheckedAnyfunc {
```
syrusakbary, line 679: I'm not super sure about this model. (the engine keeping track of the metadata) as a parent -> children relation.
I need to think more about this

lib/api/src/externals/function.rs
```diff
@@ -890,8 +910,10 @@ impl<T: VMDynamicFunction> VMDynamicFunctionCall<T> for VMDynamicFunctionContext
         let result = panic::catch_unwind(AssertUnwindSafe(|| {
             let func_ty = self.ctx.function_type();
             let mut args = Vec::with_capacity(func_ty.params().len());
+            // TODO: we need a store here, where do we get it?
+            let hack = 3;
```
syrusakbary, line 914: ?

lib/vm/src/func_data_registry.rs
```diff
@@ -0,0 +1,102 @@
+//! TODO: document this
+//!
+//! We store metadata for functions here. This is what a funcref actually points to.
```
syrusakbary, line 3: This seems a bit convoluted. Why do we need a registry of function metadata? Why we can't have the metadata on the definition itself?

lib/api/src/externals/function.rs
MarkMcCaskey: Yeah, the specific issue that this solves is the problem of arbitrary functions inserted into tables. Without that issue there are simpler solutions.

Explicitly, the problem is that users at runtime can get access to an exported or imported table, and insert any arbitrary function into it. Including functions that have signatures that do not show up in either the exported or imported functions of that module (we also handle that case with `engine.register_signature`). Thus we need a `FuncRef` (which is a pointer to data somewhere else) for any possible function a user can create at runtime, additionally we have to store that data somewhere that the `FuncRef` points to.

lib/api/src/externals/function.rs
MarkMcCaskey: This is the `Value` change mentioned above:

>  Passing a store via &dyn Any to some methods and the garbage placeholder values we pass when we don't have a store.

Currently because of the way that Value works we need a `Store` when dealing with `FuncRef`. Additionally because `Value` doesn't know about `Store` due to the shape of our dependency tree, we have to pass it secretly using traits and `dyn Any`. Here we don't have access to a `Store` (but probably need one). This was a hack to get it to compile but I believe the fact that it hasn't failed any of our tests yet is a symptom of a lack of tests, specifically tests calling functions that pass and return `FuncRef`s as values.

lib/vm/src/func_data_registry.rs
MarkMcCaskey: Generally speaking it's because of the distinction between local and imported functions and the case I mentioned above with arbitrary functions being allowed to be stored in Tables that the Wasm can interact with.

Additionally, the current model for `FuncRef`s is that they live as long as the engine / store does, so we store the data here.

I don't disagree that the current solution is more complicated than we want, but other than a few small improvements to the current system, I don't have a plan to fix it yet.

lib/vm/src/table.rs
MarkMcCaskey: Well the intent is that their internal representation is different. `union` means the size of the entire thing is the size of biggest thing in it, so it's 1 word. Using an enum means you need extra space for the tag. We could use an `enum` and `usize` as a `void*` in some places but I don't know if I'd consider that a simplification.

What are you thinking here for simplifying it?

lib/vm/src/instance/mod.rs
```diff
@@ -513,71 +519,131 @@ impl Instance {
...
+            .unwrap_or_else(|| panic!("no table for index {}", table_index.index()))
+            .size()
+    }
+
+    /// Returns the number of elements in a given imported table.
+    ///
+    /// # Safety
+    /// TODO: review safety of this. (safety message copied from memory.size)
+    /// This and `imported_table_grow` are currently unsafe because they
+    /// dereference the table import's pointers.
```
syrusakbary, line 535: TODO: Review safety

lib/compiler-cranelift/src/translator/code_translator.rs
```diff
@@ -78,27 +78,35 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
             let ty = builder.func.dfg.value_type(val);
             let local_type = environ.get_local_type(*local_index).unwrap();
             if local_type == WasmerType::ExternRef {
-                environ.translate_externref_dec(builder.cursor(), val)?;
-            }
-            if ty.is_ref() {
+                debug_assert!(_metadata.ref_counted);
+                let existing_val = builder.use_var(Variable::with_u32(*local_index));
+                environ.translate_externref_dec(builder.cursor(), existing_val)?;
                 environ.translate_externref_inc(builder.cursor(), val)?;
```
nlewycky, line 107: The value already has one use on the local value stack, and then local.set drops it (dec-ref) and it lives on in the local value slot (inc-ref). That dec+inc is a no-op so we shouldn't be inc-ref'ing it here.

Also, always inc before dec please. That's not the bug here, but in general it prevents problems when val == existing_val.

lib/compiler-cranelift/src/translator/code_translator.rs
```diff
@@ -552,41 +593,84 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
+            // decrement the ref count of all ref-counted values as we pop them off the stack
+            for arg in args
+                .iter()
+                .zip(args_metadata.iter())
+                .filter(|(_, metadata)| metadata.ref_counted)
+                .map(|(arg, _)| arg)
+            {
+                // TODO: figure out if we can avoid this
+                environ.translate_externref_dec(builder.cursor(), *arg)?;
+            }
```
nlewycky, line 624: The callee has to be responsible for dec-ref'ing all its locals. Consider a) caller does an inc-ref as part of a call b) callee does a local.set on top of the value causing a dec-ref c) we have no idea whether to dec-ref again or not. So, the callee has to do it on exit, and we don't do this loop after calls.

lib/compiler-cranelift/src/translator/code_translator.rs
```diff
@@ -1317,25 +1401,29 @@ pub fn translate_operator<FE: FuncEnvironment + ?Sized>(
...
             let table_index = TableIndex::from_u32(*index);
             let table = state.get_or_create_table(builder.func, *index, environ)?;
-            let value = state.pop1();
-            let index = state.pop1();
+            let (value, value_metadata) = state.pop1();
+            let (index, _) = state.pop1();
             environ.translate_table_set(builder, table_index, table, value, index)?;
+            if value_metadata.ref_counted {
+                // TODO: look into using "move"-like semantics here to avoid doing this
+                environ.translate_externref_dec(builder.cursor(), value)?;
```
nlewycky, line 1415: Why? Does table_set do an inc-ref? If it didn't, we wouldn't need the dec-ref here.

(Oh I guess that's what the comment is about. OK.)

lib/compiler-cranelift/src/translator/func_translator.rs
```diff
@@ -248,6 +248,12 @@ fn parse_function_body<FE: FuncEnvironment + ?Sized>(
         environ.after_translate_operator(&op, builder, state)?;
     }
 
+    // When returning we drop all values in locals and on the stack.
+    // TODO:
+    // - [ ] figure out which values on the stack are being returned so we can filter them out from being dropped
+    // - [ ] drop all locals
```
nlewycky, line 254: It looks like this is done?

lib/compiler-cranelift/src/translator/func_translator.rs
```diff
@@ -256,6 +262,29 @@ fn parse_function_body<FE: FuncEnvironment + ?Sized>(
...
+                .iter()
+                .zip(state.metadata_stack.iter())
+                .take(num_elems_to_drop)
+                .filter(|(_, metadata)| metadata.ref_counted)
+                .map(|(val, _)| val)
+            {
+                environ.translate_externref_dec(builder.cursor(), *val)?;
+            }
+
+            // TODO: look into what `state.reachable` check above does as well as `!builder.is_unreachable`, do we need that too for ref counting?
```
nlewycky, line 286: This is a very good question. What happens to externref upon unreachable? Are they dec-ref'd?

lib/compiler-cranelift/src/translator/code_translator.rs
MarkMcCaskey: Yep, fixed locally! I agree

lib/compiler-cranelift/src/translator/code_translator.rs
MarkMcCaskey: Good call on the `inc` and ordering!

lib/compiler-cranelift/src/translator/code_translator.rs
MarkMcCaskey: Oh good point, I was confusing arguments with the stack. We already do inc/dec logic for locals on function enter/exit. The extra inc/dec in call is a no-op. I'll remove both 👍

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/wasmer-types/Cargo.toml
```diff
@@ -22,3 +22,6 @@ default = ["std", "enable-serde"]
 std = ["serde/std"]
 core = []
 enable-serde = ["serde", "cranelift-entity/enable-serde"]
+
+# experimental / in-development features
+experimental-reference-types-extern-ref = []
```
syrusakbary, line 27: I'd name it `reference-types` directly. It's experimental per se, but it's unlikely to have another implementation (all implementations wil be experimental), a feature can be also experimental per se.
Therefore simplifying the feature name seems like a good thing to do!

lib/c-api/Cargo.toml
```diff
@@ -53,7 +53,7 @@ wat = ["wasmer/wat"]
 wasi = ["wasmer-wasi", "typetag", "serde"]
 engine = []
 middlewares = ["wasmer-middlewares"]
-deprecated = ["libffi"]
+deprecated = ["libffi", "wasmer/experimental-reference-types-extern-ref"]
```
syrusakbary, line 56: Ummm I'm not sure if this was really intended?
We should have a new feature for reference types: `reference-types = ["wasmer/reference-types"]`

lib/vm/src/module.rs
```diff
@@ -70,10 +70,10 @@ pub struct ModuleInfo {
     pub table_initializers: Vec<TableInitializer>,
 
     /// WebAssembly passive elements.
-    pub passive_elements: PrimaryMap<ElemIndex, Box<[FunctionIndex]>>,
+    pub passive_elements: HashMap<ElemIndex, Box<[FunctionIndex]>>,
```
syrusakbary, line 73: We don't want this change in master, somehow is reverting the PR #2183?

lib/wasmer-types/Cargo.toml
MarkMcCaskey: That's good feedback but I think the name `reference-types` would be misleading because this feature only concerns the `ExternRef` API in Wasmer. Reference types are always enabled and can't easily be disabled, statically in the code I mean.

The runtime config that allows Wasm modules to be compiled with reference types has it disabled by default. So users must update that config option or pass the `--enable-reference-types` flag to enable it wit the Wasmer CLI currently.

Happy to change the name of the feature flag though

lib/c-api/Cargo.toml
MarkMcCaskey: We can make a separate feature for enabling it but it's not currently usable in the C API so it won't do anything.

This feature was needed to get the deprecated C API to compile, though after the latest edits I made it actually might not be necessary anymore.

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

lib/wasmer-types/src/features.rs
```diff
@@ -36,7 +36,9 @@ impl Features {
     pub fn new() -> Self {
         Self {
             threads: false,
-            reference_types: false,
+            // We may not want to have reference types on by default, but we do
+            // at least during development
+            reference_types: true,
```
nlewycky, line 41: Use our cfg feature?

lib/wasmer-types/src/features.rs
```diff
@@ -245,7 +247,7 @@ mod test_features {
             default,
             Features {
                 threads: false,
-                reference_types: false,
+                reference_types: true,
```
nlewycky, line 250: CFG feature here too

nlewycky (approved):
MHO, we should land this as soon as possible then continue development. The only blockers IMO are:
* everything that isn't really done here or is a hack or is commented out code that we don't want to just remove right now should be marked TODO so that we can find it easily after it lands
* the text that's currently in the PR description with the checkmarks for what's left to be done should be moved into a file that we can update and refer to.

lib/api/src/externals/function.rs
```diff
@@ -911,7 +934,7 @@ impl<T: VMDynamicFunction> VMDynamicFunctionCall<T> for VMDynamicFunctionContext
                 ret.write_value_to(values_vec.add(i));
             }
             Ok(())
-        }));
+        })); // We get an extern ref drop at the end of this block that we don't want
```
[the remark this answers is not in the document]
MarkMcCaskey: Oh this comment is about the efficiency of our ref counting, we were getting a drop here (and had to do an inc because of it), this comment is more just about how due to a drop at that point, we're wasting work (incing and then decing, when we could just do neither)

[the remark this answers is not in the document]
MarkMcCaskey: I'll revisit it and do something here

lib/api/src/externals/table.rs
```diff
@@ -27,7 +27,7 @@ pub struct Table {
 fn set_table_item(
     table: &dyn RuntimeTable,
     item_index: u32,
-    item: VMCallerCheckedAnyfunc,
+    item: TableReference,
```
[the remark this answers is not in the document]
MarkMcCaskey: Yeah, I agree it's weird, it's because tables can only contain references... and the term `TableElement` was already in use by some other type. I agree that we should probably change it before shipping it, especially because this type is more public than the other one

lib/engine-jit/src/engine.rs
```diff
@@ -299,4 +310,9 @@ impl JITEngineInner {
     pub fn signatures(&self) -> &SignatureRegistry {
         &self.signatures
     }
+
+    /// Shared func metadata registry.
+    pub(crate) fn func_data(&self) -> &Arc<FuncDataRegistry> {
```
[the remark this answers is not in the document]
MarkMcCaskey: Yeah, I think we use this pattern in the trait already is why I went with it (to be more consistent), returning `&Arc<T>` is more efficient (caller gets to decide if they need to do the allocation and atomic increment to make a new Arc; they don't have to if they just want to access the internal type) but also more brittle from an API perspective because it means that this type must always hold on to an Arc internally that lives long enough. It's definitely a trade-off 🤷

lib/vm/src/instance/mod.rs
```diff
@@ -84,11 +86,14 @@ pub(crate) struct Instance {
 
     /// Passive elements in this instantiation. As `elem.drop`s happen, these
     /// entries get removed.
-    passive_elements: RefCell<PrimaryMap<ElemIndex, Option<Box<[VMCallerCheckedAnyfunc]>>>>,
+    passive_elements: RefCell<HashMap<ElemIndex, Box<[VMFuncRef]>>>,
 
     /// Passive data segments from our module. As `data.drop`s happen, entries
     /// get removed. A missing entry is considered equivalent to an empty slice.
-    passive_data: RefCell<PrimaryMap<DataIndex, Option<Arc<[u8]>>>>,
+    passive_data: RefCell<HashMap<DataIndex, Arc<[u8]>>>,
```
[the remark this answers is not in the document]
MarkMcCaskey: I can, but the tests showing that that patch that landed are wrong only exist in this branch. We did not test the behavior before reference types, which is why it needed to be fixed here. We can independently fix it too, and I think that might be a good idea

lib/vm/src/instance/mod.rs
```diff
@@ -795,6 +891,47 @@ pub struct InstanceHandle {
     instance: InstanceRef,
 }
 
+/// TOOD: move this function somewhere else
```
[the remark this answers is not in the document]
MarkMcCaskey: haha, funny typo. Yeah I use `TODO(reftypes)` in a few places (maybe all gone now) but I think that's even better for finding things specific to reftypes here

lib/compiler-cranelift/src/func_environ.rs
```diff
@@ -595,11 +942,11 @@ impl<'module_environment> BaseFuncEnvironment for FuncEnvironment<'module_enviro
         ty: Type,
     ) -> WasmResult<ir::Value> {
         Ok(match ty {
-            Type::FuncRef => pos.ins().iconst(self.pointer_type(), 0),
-            // Type::ExternRef => pos.ins().null(self.reference_type(ty)),
+            Type::FuncRef => pos.ins().null(self.reference_type()), //iconst(self.pointer_type(), 0),
```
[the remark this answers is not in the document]
MarkMcCaskey: No need, I think it's done, that commented code is very old

MarkMcCaskey:
bors try

Issue #2025, opened 2021-01-15
Using metering middleware via c-api


satrobit:
### Summary

I'm trying to add metering to wasm modules using wasmer c-api but I'm not able to find the necessary headers for it. Is it even exposed to the c-api? If it is, can you point me in the right direction?

MarkMcCaskey:
Hello!

That's a good question! We currently do not have a metering API available to the C API.

It's certainly a possibility -- the hard part would be designing what that API would look like in a `wasm-c-api` style.

satrobit:
It'd be awesome to have this feature in the C API. It's indeed challenging to come up with an elegant design that is compatible with `wasm-c-api` but I think it's worth the effort. I'll probably hack something together for now but I'll share my findings if any clean design comes to my mind.

Hywan:
Hello @satrobit,

It took a little bit of time, but here is the metering C API, https://wasmerio.github.io/wasmer/crates/wasmer_c_api/wasm_c_api/unstable/middlewares/metering/index.html. I'm copy pasting the example for the record:

```c
// Define our “cost function”.
uint64_t cost_function(wasmer_parser_operator_t wasm_operator) {
    switch(wasm_operator) {
        // `local.get` and `i32.const` cost 1 unit.
        case LocalGet:
        case I32Const:
            return 1;

        // `i32.add` costs 2 units.
        case I32Add:
            return 2;

        // The other operations are free.
        default:
            return 0;
    }
}

int main() {
    // Create a new metering middleware, with our cost function.
    wasmer_metering_t* metering = wasmer_metering_new(10, cost_function);

    // Consume `metering` to produce a generic `wasmer_middleware_t` value.
    wasmer_middleware_t* middleware = wasmer_metering_as_middleware(metering);
     
    // Create a new configuration, and push the middleware in it.
    wasm_config_t* config = wasm_config_new();
    wasm_config_push_middleware(config, middleware);
     
    // Create the engine and the store based on the configuration.
    wasm_engine_t* engine = wasm_engine_new_with_config(config);
    wasm_store_t* store = wasm_store_new(engine);
     
    // Create the new WebAssembly module.
    wasm_byte_vec_t wat;
    wasmer_byte_vec_new_from_string(
        &wat,
        "(module\n"
        "  (type $add_t (func (param i32) (result i32)))\n"
        "  (func $add_two_f (type $add_t) (param $value i32) (result i32)\n"
        "    local.get $value\n"
        "    i32.const 1\n"
[...]

PR #2103, opened 2021-02-09
feat(c-api) Introduce the middleware (specifically metering) C API

nlewycky:
Edit by @Hywan: Fixes https://github.com/wasmerio/wasmer/issues/2025.


syrusakbary (commented):
I think the `wasmer_metering_points_value`, `wasmer_metering_points_is_exhausted` and `wasmer_metering_get_remaining_points` can be simplified.

Right know, to get the points you have to do: `wasmer_metering_get_remaining_points`, ask if `wasmer_metering_points_is_exhausted` and then `wasmer_metering_get_remaining_points`.
This ties very close to the inner Rust structures but made the usage harder.

I think one simpler thing that we can do is to have `wasmer_metering_get_remaining_points` return the remaining points integer directly if it's not exhausted and `-1` in case is exhausted.


## Tests

Apart from that, we will need to test the metering API. Probably you can use the `inline-c` crate macros for testing as we do it with other C-APi tests.

nlewycky:
> I think the `wasmer_metering_points_value`, `wasmer_metering_points_is_exhausted` and `wasmer_metering_get_remaining_points` can be simplified.
> 
> Right know, to get the points you have to do: `wasmer_metering_get_remaining_points`, ask if `wasmer_metering_points_is_exhausted` and then `wasmer_metering_get_remaining_points`.
> This ties very close to the inner Rust structures but made the usage harder.
> 
> I think one simpler thing that we can do is to have `wasmer_metering_get_remaining_points` return the remaining points integer directly if it's not exhausted and `-1` in case is exhausted.

Yep, that's a good change! The only adjustment to that is that instead of returning `-1` when exhausted, it should return a value that the user passes in, which is what `wasmer_metering_points_value` currently does.

syrusakbary:
> The only adjustment to that is that instead of returning -1 when exhausted, it should return a value that the user passes in, which is what wasmer_metering_points_value currently does.

I think that still makes the API a bit more complicated, and I'm failing to see the use case for a custom number that is not the "special case".
Exhausted points can only be one thing (is a state value) and with a -1 value it doesn't collide with the remaining points (since they are always > 0?).
So I still think not having to provide an artificial exhausted value is simpler than having to provide one.

What is the use case that you think we will cover with having the user to provide the exhausted value?

lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -88,6 +94,16 @@ impl Default for wasmer_engine_t {
     }
 }
 
+/// Opaque representing a middleware.
+///
+#[cfg(feature = "middlewares")]
+#[derive(Debug)]
+#[repr(C)]
```
Hywan, line 101: We want `wasmer_module_middle_t` to be an opaque type. No need to get a C representation.

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
```diff
@@ -0,0 +1,350 @@
+//! Unstable non-standard Wasmer-specific API that contains everything
+//! to create a the middleware metering API.
```
jubianchi, line 2:
```suggestion
//! to create the middleware metering API.
```

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
```diff
@@ -0,0 +1,350 @@
...
+//! # use inline_c::assert_c;
+//! # fn main() {
+//! #    (assert_c! {
+//! # #include "tests/wasmer_wasm.h"
+//! #
+//! int main() {
+//!     // Create a new metering middleware.
+//!     wasmer_metering_t* metering = wasmer_metering_new(10);
+//!
+//!     // Consume `metering` to produce a generic `wasmer_middle_t` value.
```
jubianchi, line 20:
```suggestion
//!     // Consume `metering` to produce a generic `wasmer_middleware_t` value.
```

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
```diff
@@ -0,0 +1,350 @@
...
+/// # Example
+///
+/// See module's documentation.
+#[no_mangle]
+pub unsafe extern "C" fn wasmer_metering_points_delete(
+    _metering_points: Option<Box<wasmer_metering_points_t>>,
+) {
+}
+
+/// Returns the number of remaining points if any, otherwise returned
```
jubianchi, line 162:
```suggestion
/// Returns the number of remaining points if any, otherwise returns
```

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
```diff
@@ -0,0 +1,350 @@
...
+/// the given `exhausted` value if points are exhausted.
+///
+/// # Example
+///
+/// See module's documentation.
+#[no_mangle]
+pub unsafe extern "C" fn wasmer_metering_points_unwrap_or(
```
jubianchi, line 169: This is a very Rust-style function (`unwrap_or`). 

Wouldn't it be better if:
* we rename this to something like `wasmer_metering_points_remaining`;
* we hard-code the exhausted value here (i.e `-1`);
* we either return the number or remaining points or `-1`

jubianchi (commented):
Seems great! 👍

Last comment I would add is it would be awesome to have a cost function example.

Hywan:
User-defined cost function isn't working yet :-). Working on it!

Hywan:
Everything is working! Here is a full example:

```c
// Define our “cost function”.
uint64_t cost_function(wasmer_parser_operator_t operator) {
    switch(operator) {
        // `local.get` and `i32.const` cost 1 unit.
        case LocalGet:
        case I32Const:
            return 1;

        // `i32.add` costs 2 units.
        case I32Add:
            return 2;

        // The other operations are free.
        default:
            return 0;
    }
}

int main() {
    // Create a new metering middleware, with our cost function.
    wasmer_metering_t* metering = wasmer_metering_new(10, cost_function);

    // Consume `metering` to produce a generic `wasmer_middle_t` value.
    wasmer_middleware_t* middleware = wasmer_metering_as_middleware(metering);
     
    // Create a new configuration, and push the middleware in it.
    wasm_config_t* config = wasm_config_new();
    wasm_config_push_middleware(config, middleware);
     
    // Create the engine and the store based on the configuration.
    wasm_engine_t* engine = wasm_engine_new_with_config(config);
     
    // Create the new WebAssembly module.
    wasm_byte_vec_t wat;
    wasmer_byte_vec_new_from_string(
        &wat,
        "(module\n"
        "  (type $add_t (func (param i32) (result i32)))\n"
        "  (func $add_two_f (type $add_t) (param $value i32) (result i32)\n"
        "    local.get $value\n"
        "    i32.const 1\n"
        "    i32.add)\n"
        "  (export \"add_two\" (func $add_two_f)))"
    );
    wasm_byte_vec_t wasm;
    wat2wasm(&wat, &wasm);

    wasm_module_t* module = wasm_module_new(store, &wasm);
    assert(module);
     
    // Instantiate the module.
    wasm_extern_vec_t imports = WASM_EMPTY_VEC;
    wasm_trap_t* traps = NULL;
    wasm_instance_t* instance = wasm_instance_new(store, module, &imports, &traps);
    assert(instance);
     
    // Here we go. At this step, we will get the `add_two` exported function, and
    // call it.
    wasm_extern_vec_t exports;
    wasm_instance_exports(instance, &exports);
    assert(exports.size >= 1);
    assert(wasm_extern_kind(exports.data[0]) == WASM_EXTERN_FUNC);

    const wasm_func_t* add_two = wasm_extern_as_func(exports.data[0]);
    assert(add_two);

    wasm_val_t arguments[1] = { WASM_I32_VAL(41) };
    wasm_val_t results[1] = { WASM_INIT_VAL };

    wasm_val_vec_t arguments_as_array = WASM_ARRAY_VEC(arguments);
    wasm_val_vec_t results_as_array = WASM_ARRAY_VEC(results);

    // Let's define a value when points are exhausted.
    uint64_t is_exhausted = -1;

    // Let's call `add_two` for the first time!
    {
        wasm_trap_t* trap = wasm_func_call(add_two, &arguments_as_array, &results_as_array);
        assert(trap == NULL);
        assert(results[0].of.i32 == 42);

        // There is 6 points left!
        wasmer_metering_points_t* metering_points = wasmer_metering_get_remaining_points(instance);
        assert(wasmer_metering_points_unwrap_or(metering_points, is_exhausted) == 6);
        assert(wasmer_metering_points_is_exhausted(metering_points) == false);
        wasmer_metering_points_delete(metering_points);
    }

    // Let's call `add_two` for the second time!
    {
        wasm_trap_t* trap = wasm_func_call(add_two, &arguments_as_array, &results_as_array);
        assert(trap == NULL);
        assert(results[0].of.i32 == 42);

        // There is 2 points left!
        wasmer_metering_points_t* metering_points = wasmer_metering_get_remaining_points(instance);
        assert(wasmer_metering_points_unwrap_or(metering_points, is_exhausted) == 2);
        assert(wasmer_metering_points_is_exhausted(metering_points) == false);
        wasmer_metering_points_delete(metering_points);
    }

    // Let's call `add_two` for the third time!
    {
        wasm_trap_t* trap = wasm_func_call(add_two, &arguments_as_array, &results_as_array);
        // Oh, it failed!
        assert(trap != NULL);

        // There is 0 point left… they are exhausted.
        wasmer_metering_points_t* metering_points = wasmer_metering_get_remaining_points(instance);
        assert(wasmer_metering_points_unwrap_or(metering_points, is_exhausted) == is_exhausted);
        assert(wasmer_metering_points_is_exhausted(metering_points) == true);
        wasmer_metering_points_delete(metering_points);
    }
     
    wasm_instance_delete(instance);
    wasm_module_delete(module);
    wasm_store_delete(store);
    wasm_engine_delete(engine);

    return 0;
}
```

Hywan:
bors try

Hywan:
Thinking about https://github.com/wasmerio/wasmer/pull/2103#discussion_r585886375. I'm suggesting the new following API (that removes the `wasmer_metering_points_t` type):

```c
// Create a metering middleware.
wasmer_metering_new(initial_value, cost_function);

// Transform metering to generic middle.
wasmer_metering_as_middleware(metering);

// Register the middleware.
wasm_config_push_middleware(config, middleware);

// Check remaining points. Returned value is 0 if exhausted, otherwise > 0.
uint32_t wasm_metering_get_remaining_points(instance);
```

That way, we don't need the `wasmer_metering_points_t` type, so it removes the following functions:

* `wasmer_metering_points_delete`,
* `wasmer_metering_points_is_exhausted`,
* `wasmer_metering_points_unwrap_or`.

Thoughts?

Hywan:
bors try

Hywan:
bors try

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
```diff
@@ -0,0 +1,300 @@
...
+//!     }
+//!
+//!     // Let's call `add_two` for the third time!
+//!     {
+//!         wasm_trap_t* trap = wasm_func_call(add_two, &arguments_as_array, &results_as_array);
+//!         // Oh, it failed!
+//!         assert(trap != NULL);
+//!
+//!         // There is 0 point left… they are exhausted.
+//!         assert(wasmer_metering_get_remaining_points(instance) == 0);
```
jubianchi, line 116: Using `0` as the exhausted value may be unclear: did we only consume all the points we had? did we consume more than what we had before?

In the first case, it's not an error AFAIR: we consumed everything but not more than what we had. In the latter case, it's an error, we tried to consume more points than we had.

I think, we should either:
* have `-1` to tell we did not had enough points
* reintroduce a second function to tell if we exhausted or not.

Hywan: Hmm I see. I don't want to return -1 because we are dealing with `u64` already. Supporting negative value will reduce the set of values we can return. Maybe we could continue to return 0, and add a function like `wasmer_metering_points_is_exhausted(wasm_instance_t*)`. Thoughts?

syrusakbary: I kind of prefer using -1. I'm not sure the metering points will ever reach super ultra big values (specially in blockchain contract context).

Hywan:
bors try

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
```diff
@@ -0,0 +1,300 @@
...
+//!
+//!         // `i32.add` costs 2 units.
+//!         case I32Add:
+//!             return 2;
+//!
+//!         // The other operations are free.
+//!         default:
+//!             return 0;
+//!     }
+//! }
```
nlewycky, line 32: Optional: It might be worth using a more realistic cost function for the example?
a) the four Const operations are 0
b) F32Div and F64Div are cost 2
c) everything else is cost 1
?

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
```diff
@@ -0,0 +1,300 @@
...
+        MeteringPoints::Remaining(value) => value,
+        MeteringPoints::Exhausted => 0,
+    }
+}
+
+/// Set a new amount of points for the given metering middleware.
+///
+/// # Example
+///
+/// This example is pointless as the number of points aren't updated
```
nlewycky, line 226: Is that a TODO then?

Otherwise it's not pointless, it's just that its point is only to illustrate the `wasmer_metering_set_remaining_points` function.

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
```diff
@@ -0,0 +1,300 @@
...
+/// uint64_t cost_function(wasmer_parser_operator_t operator) {
+///     switch(operator) {
+///         default:
+///             return 0;
+///     }
+/// }
+///
+/// int main() {
+///     // Set the initial amount of points to 10.
+///     wasmer_metering_t* metering = wasmer_metering_new(7, cost_function);
```
nlewycky, line 246: ten you say

lib/c-api/src/wasm_c_api/unstable/middlewares/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+use super::super::engine::wasm_config_t;
+use std::sync::Arc;
+use wasmer::ModuleMiddleware;
+
+#[cfg(all(feature = "middlewares", not(feature = "compiler")))]
+compile_error!("The `middlewares` feature requires the `compiler` feature to be turned on");
+
+/// Opaque representing any kind of middleware.
+///
+/// It's used by `wasm_config_push_middleware`. A specific middleware
```
nlewycky, line 15:
```suggestion
/// Used by `wasm_config_push_middleware`. A specific middleware
```

Or just not mention what it's used by?

lib/c-api/src/wasm_c_api/unstable/parser/operator.rs
```diff
@@ -0,0 +1,1017 @@
+use wasmer::wasmparser::Operator;
```
nlewycky, line 1: Clearly this file should be produced by a proc-macro that runs over wasmparser.

No action required. Please.

lib/engine-native/src/artifact.rs
```diff
@@ -317,7 +317,7 @@ impl NativeArtifact {
         if is_cross_compiling {
             Self::from_parts_crosscompiled(metadata, shared_filepath)
         } else {
-            let lib = Library::new(&shared_filepath).map_err(to_compile_error)?;
+            let lib = unsafe { Library::new(&shared_filepath).map_err(to_compile_error)? };
```
nlewycky, line 320: Technically this is part of a different PR?

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
Hywan: After discussion with @nlewycky, we decided to return `u64::MAX` so that we don't lose any values from `u64` by returning -1. I've also added `wasmer_metering_points_are_exhausted` to differentiate between 0 or exhausted points. Tests are up-to-date.

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
Hywan: I've taken the example from `examples/metering.rs`. It fits well with the simple WAT module. Thoughts?

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
Hywan: :+1: Updated

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
Hywan: :+1:

lib/c-api/src/wasm_c_api/unstable/middlewares/mod.rs
Hywan: :+1:

lib/c-api/src/wasm_c_api/unstable/parser/operator.rs
Hywan: 🤣

lib/engine-native/src/artifact.rs
Hywan: Yeah but I want to see green tests in this PR for the moment.

Hywan:
bors try

lib/c-api/build.rs
```diff
@@ -467,10 +471,33 @@ fn exclude_items_from_wasm_c_api(builder: Builder) -> Builder {
...
+        .exclude_item("wasm_named_extern_vec_new")
+        .exclude_item("wasm_named_extern_vec_new_empty")
+        .exclude_item("wasm_named_extern_vec_new_uninitialized")
+        .exclude_item("wasm_target_delete")
+        .exclude_item("wasm_target_new")
+        .exclude_item("wasm_target_t")
+        .exclude_item("wasm_triple_delete")
+        .exclude_item("wasm_triple_new")
+        .exclude_item("wasm_triple_new_from_host")
+        .exclude_item("wasm_triple_t")
```
Hywan, line 500: This must be removed.

Hywan (approved):
bors r+

lib/c-api/src/wasm_c_api/unstable/middlewares/metering.rs
```diff
@@ -0,0 +1,318 @@
...
+
+/// Function type to represent a user-defined cost function
+/// implemented in C.
+///
+/// # Example
+///
+/// See module's documentation.
+#[allow(non_camel_case_types)]
+pub type wasmer_metering_cost_function_t =
+    extern "C" fn(wasm_operator: wasmer_parser_operator_t) -> u64;
```
MarkMcCaskey, line 164: Technically should probably be an `unsafe` function, but it's a minor point here because we immediately wrap it anyways. Only really matters if it has greater contact with Rust code.

lib/c-api/src/wasm_c_api/unstable/middlewares/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+///
+/// This function takes ownership of `middleware`.
+///
+/// This is a Wasmer-specific function.
+///
+/// # Example
+///
+/// See the documentation of the [`metering`] module.
+#[no_mangle]
+pub extern "C" fn wasm_config_push_middleware(
```
MarkMcCaskey, line 35: I'd assume middleware happens at a higher level than `wasm_config`, as `wasm_config` is used for setting up the initial engine / store. Not a huge deal, but it means that it's harder to share a store with modules using different middlewares. (I'm just assuming we allow a more flexible usage in the Rust API)

MarkMcCaskey (commented):
looks good!

gottoouut:
Bye

PR #2106, opened 2021-02-10
Add a cargo-deny list.

nlewycky:
# Description
I recommend reading each of the individual commits in this PR. They start off by adding the original template, modifying that template so that it's as strict as possible but still passes (mostly loosening requirements), then modifying Wasmer where it was easy to do so in order to tighten the requirements.

# Review

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,9 +8,9 @@ checksum = "955f37ac58af2416bac687c8ab66a4ccba282229bd7422a28d2281a5e66a6116"
 
 [[package]]
 name = "addr2line"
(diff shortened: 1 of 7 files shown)


syrusakbary (approved):

syrusakbary (commented):
Let's also add a check in the CI (so we check cargo deny as part of future PRs)

nlewycky:
> Let's also add a check in the CI (so we check cargo deny as part of future PRs)

Yep! I was planning to do that in a separate PR. Should I add it in this one?

PR #2108, opened 2021-02-11
Don't #include inside extern "C" blocks.

nlewycky:
Also free the malloc'd error string in print_wasmer_error().

Run the C code through clang-format.


lib/cli/src/c_gen/object_file_header.rs
```diff
@@ -285,11 +288,9 @@ pub fn generate_header_file(
         value: HELPER_FUNCTIONS.to_string(),
     });
 
-    let inner_c = generate_c(&c_statements);
+    c_statements.push(CStatement::LiteralConstant {
+        value: "\n#ifdef __cplusplus\n}\n#endif\n\n".to_string(),
+    });
```
MarkMcCaskey, line 293: I don't remember much about the abstractions here but it might be nice to have the CPlusPlus guard be a thing that can work on nested data / trees so we don't have to remember to close it ourselves.

Don't let that block this PR though

lib/cli/src/commands/wasmer_create_exe_main.c
```diff
@@ -14,21 +10,17 @@ extern "C" {
...
-}
-#endif
-
-void print_wasmer_error()
-{
+static void print_wasmer_error() {
   int error_len = wasmer_last_error_length();
   printf("Error len: `%d`\n", error_len);
-  char* error_str = (char*) malloc(error_len);
+  char *error_str = (char *)malloc(error_len);
```
MarkMcCaskey, line 16: I will never believe that `char *name` is more readable or accurate than `char* name` 😆

MarkMcCaskey (approved):
Looks good -- thanks!

lib/cli/src/commands/wasmer_create_exe_main.c
nlewycky: `char x, *y, f()` is a char, a pointer-to-char and a function that returns a char. What a language! :laughing:

Hywan (approved):

PR #2114, opened 2021-02-11
Add cargo-deny action as part of our CI.

nlewycky:
This adds cargo-deny with its own separate workflow file so that we can change it independently from other workflows. It runs at the same times that our lint workflow runs.

### DIFF
--- a/.github/workflows/cargo-deny.yaml
+++ b/.github/workflows/cargo-deny.yaml
@@ -1,0 +1,15 @@
+on:
+  push:
+    branches:
+      - '**'
+      - '!master'
+
+name: cargo-deny
+
+jobs:
+  deny-check:
+    name: cargo-deny
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: EmbarkStudios/cargo-deny-action@v1


.github/workflows/cargo-deny.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+name: cargo-deny
+
+jobs:
+  deny-check:
+    name: cargo-deny
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: EmbarkStudios/cargo-deny-action@v1
```
MarkMcCaskey, line 15: My only comment is that I'm unclear if Rust version matters for cargo-deny, if so we may want to specify, otherwise 👍

MarkMcCaskey (approved):

.github/workflows/cargo-deny.yaml
nlewycky: I don't know how to tell whether Rust version matters?

I mimicked the examples on their homepage plus two examples of real users with workflows set up. That doesn't answer the question of whether the Rust version matters though.

MarkMcCaskey: There's no options on that action to specify, so it's probably fine 🤷 ; if it's a problem we can deal with it later

PR #2117, opened 2021-02-11
chore(c-api) Formalize API prefixes

Hywan:
# Description

The new rule is the following:

* `wasm_` for the standard C API,
* `wasmer_` or `wasi_` for the Wasmer non-standard C API.

For all symbols inside the `unstable` module, the renaming `wasm_` to
`wasmer_` is done without deprecations. It was clear that those API
were unstable.

For all the other symbols, symbols have been renamed to `wasmer_` but
the old symbols have been kept with deprecation warnings.

Special note: The `wasm_named_extern_t` type (and associated
functions) was in `wasi` by mistake. Its place was in the `unstable`
module. This patch also fixes that.

The `wasm_declare_vec_*` macros have been updated to support a default
prefix, or a user-defined prefix. It's now possible to write
`wasm_declare_boxed_vec!(foo);` to get all the API prefixed by `wasm_`
(as previously), or `wasm_declare_boxed_vec!(foo, wasmer);` to prefix
with `wasmer_`.

A user not using symbols from the `unstable` module will continue to
get working code, modulo some deprecations, after this patch.

# Review


Hywan:
bors try

syrusakbary (commented):
I think this need a deeper review.
In any case I don't think `wat2wasm` should be renamed to `wasmer_wat2wasm` `wat2wasm` is more readable and clear.

Also, we need to make sure that none of the `wasm_X` renamed to `wasmer_X` are part of the official wasm-c-api.

Hywan:
I've checked all that already :-). And we agreed with the rest of the team to rename `wat2wasm` to `wasmer_wat2wasm` to avoid naming conflict with potentially other projects.

nlewycky (approved):
Recommend waiting for @MarkMcCaskey too.

nlewycky:
> I've checked all that already :-). And we agreed with the rest of the team to rename `wat2wasm` to `wasmer_wat2wasm` to avoid naming conflict with potentially other projects.

Yeah, I do wish we could keep the `wat2wasm` name but C there is no namespacing of symbols in C. If wasmer as a library wants to provide `wat2wasm` then you can't write your own program with your own wat2wasm in it, nor can you use two libraries that both have wat2wasm. It's customary for C libraries to prefix all of their symbols with their project name to avoid conflicts.

> Also, we need to make sure that none of the `wasm_X` renamed to `wasmer_X` are part of the official wasm-c-api.

Yes this is important. It falls into a few categories: `wasm_cpu_*` `wasm_target_*` and `wasm_triple_*` which are not standard, `wasm_config_*`, creation and deletion is standard but anything else is not, wasm module names are not standard, `wasm_named_extern_*` is not standard and that's it.

lib/c-api/build.rs
```diff
@@ -469,31 +469,33 @@ fn exclude_items_from_wasm_c_api(builder: Builder) -> Builder {
...
-        .exclude_item("wasm_target_new")
-        .exclude_item("wasm_target_t")
-        .exclude_item("wasm_triple_delete")
-        .exclude_item("wasm_triple_new")
-        .exclude_item("wasm_triple_new_from_host")
-        .exclude_item("wasm_triple_t")
         .exclude_item("wasmer_compiler_t")
+        .exclude_item("wasmer_config_set_compiler")
+        .exclude_item("wasmer_config_set_engine")
+        .exclude_item("wasmer_config_set_target")
```
MarkMcCaskey, line 475: I mentioned it on Slack but I'll say it here for the record, I think there's a lot of value in having the string "wasm_config" appear in symbols that operate on this type to make it easier to find uses by text search.

Thus I tend to prefer `wasmer_wasm_config_*` or `wasm_config_*`.

Notably `wasmer_config_*` is the only set of functions/methods that don't follow the pattern `<type_name>_<operation>`. I think that inconsistency is enough to warrant further review, aside from my previous concerns about grepability of the identifiers.

I'm open to other ideas here too though!

lib/c-api/src/wasm_c_api/unstable/module.rs
```diff
@@ -139,7 +139,7 @@ pub unsafe extern "C" fn wasm_module_name(
 /// # }
 /// ```
 #[no_mangle]
-pub unsafe extern "C" fn wasm_module_set_name(
+pub unsafe extern "C" fn wasmer_module_set_name(
```
MarkMcCaskey, line 142: Oh I was wrong,this also breaks the rule of `<type_name>_<operation>`

lib/c-api/src/wasm_c_api/unstable/wasi.rs
```diff
@@ -0,0 +1,199 @@
...
+) -> Option<&wasm_extern_t> {
+    Some(named_extern?.r#extern.as_ref())
+}
+
+/// Non-standard function to get the imports needed for the WASI
+/// implementation with no particular order. Each import has its
+/// associated module name and name, so that it can be re-order later
+/// based on the `wasm_module_t` requirements.
+#[no_mangle]
+pub unsafe extern "C" fn wasi_get_unordered_imports(
```
MarkMcCaskey, line 150: Perhaps worth considering but maybe we should put `WASI` related functions in their own header files. I'll defer to NIck's judgement about header file / C API organization but just a thought. This doesn't have to be done in this PR, but you separating out WASI into its own sub-module here made me think of that

lib/c-api/build.rs
Hywan: For me, `wasmer_wasm_config_*` is a non-sense. `wasm_config_*` functions aren't standard (except for `_new` and `_delete`). We are writing non-standard functions, so the prefix is `wasmer_` regarding our new rules.

We can however add `#[doc(alias = "…")]` to ease findings in the documentation's search engine. Thoughts?

lib/c-api/src/wasm_c_api/unstable/wasi.rs
Hywan: Since this function is unstable, I prefer to keep it here. But the function is re-exported inside the `wasi` module.

Note that it doesn't really matter since everything is flat inside the `.h` file at the end of the process.

lib/c-api/build.rs
MarkMcCaskey: Hmm, I think that `wasmer_config_*` leads to big problems if you follow it through logically:

1. A type called `wasm_config_t` exists because it's declared by wasm.h
2. A function called `wasm_config_new` exists to operate on this type, it only has the `wasm_` prefix because of where it's declared
3. Because we're adding operations in our own namespace, `wasmer`, we then add functions as `wasmer_config_*` referring to a global, unnamespaced `config` object
4. No other types named `*_config_t` can exist, no other operations on any other config type can exist

That doesn't sound like the outcome we want. In fact it destroys the entire idea behind namespaces.

To me the only logically consistent names are `wasm_config_*` (saying it's okay for us to extend their namespace) or `wasmer_wasm_config_*` (saying it's not okay for us to extend their namespace). `wasmer_config_*` causes namespace conflicts (and is also very confusing in my opinion).

syrusakbary: I agree with @MarkMcCaskey point here.

My 2 cents: I'd vote for using `wasm_config_*`

Hywan: Let's go for `wasm_config_set*` then :-).

PR #2123, opened 2021-02-15
feat(makefile) Simplify the prelude + `build-capi` includes all available compilers

Hywan:
# Description

This PR contains 2 parts. Don't be afraid, it's very small changes, and it's easy to follow.

## How to Review

I recommend to review the PR patch by patch, it's way simpler than reading the final `diff`.
Note that the matrix (the table in the `Makefile` documentation) is updated patch after patch,
so this part should ideally be reviewed in the final `diff`.

The final `Makefile` is here, https://github.com/Hywan/wasmer/blob/feat-c-api-build-lib-with-all-compilers-and-engines/Makefile, and it's even easier to read now.

## Part 1, The Matrix

The set of patches writes a matrix: platform x architecture x compiler x engine x libc, and for each combination, says if it's supported or not.

This set of patches updates the “prelude” of the `Makefile` to make it clearler and easier to read, regarding the matrix.

1. adb7c59, This patch removes the `UNAME_S` and `ARCH` variables. It replaces
   them with `IS_WINDOWS`, `IS_LINUX`, `IS_DARWIN`, `IS_AMD64`, and
   `IS_AARCH64` variables. It makes the code easier to read and simpler
   to maintain.

   This patch also documents the large matrix of platform x architecture
   x compiler x engine x lib support.

2. 659bdb0, This patch removes a possible duplication of the `llvm` value inside
   the `compilers` variable if both LLVM 10 and 11 are available.

   This patch also checks for LLVM 11 before LLVM 10.

3. 7b38752, This patch moves all the assignments of `compilers` in the same place
    for the sake of clarity.

4. 006c653, This patch renames the `test_compilers_engines` variable to
    `compilers_engines`. We can use this variable for test purposes, but more
    importantly, it defines all the pair (compiler, engine), make it that clear.

    I've decided to separate this patch from the followings to facilitate the review.

5. 67a1dfe, This patch handles the (cranelift, *) pairs in a single place.

6. f3c729d, This patch handles the (llvm, *) pairs in a single place.

7. 1054e51, This patch handles the (singlepass, *) pairs in a single place.

8. 07e48b4 , This patch adds the `HAS_<compiler>` variables, it makes the code
    easier to read.

9. 06d543a, 660daba, 95dcb04, 6721769, and 62c7889 are small improvements.

10. 0a0b7f8, This patch introduces `ENABLE_<compiler>` variables, that can be
     overriden by the user to force to enable or to disable a compiler. They replace
     the `HAS_<compiler>` variables.

## Part 2, `libwasmer` includes all available compilers

1. 387a753, This patch updates the `build-capi` rule to _no longer uses Cranelift by default_,
   but to include/enable all the available compilers.

2. f846a0d, This patch explains `test-capi` with a little bit more details.

    This patch also introduces the `test-capi-all` rule. Why? `test-capi` will continue to
    build and to test the C API with a single compiler enabled, so that it's the default one
    that will be used by the tests. _But_ we also want to test the `libwasmer` we are giving
    to the users! That's what `test-capi-all` does. Notice that the `test-capi` rule depends
    `test-capi-all`, so we don't need to update the CI.

Closes https://github.com/wasmerio/wasmer/pull/2063

# Review


Hywan:
bors try

Makefile
```diff
@@ -1,112 +1,296 @@
...
+##
+# The Cranelift case.
+##
+
+ifeq ($(HAS_CRANELIFT), 1)
+	compilers_engines += cranelift-jit
+
+	ifeq ($(IS_WINDOWS), 0)
+		ifeq ($(IS_AMD64), 1)
+			ifneq ($(LIBC, musl))
```
jubianchi, line 188:
```suggestion
			ifneq ($(LIBC), musl)
```

jubianchi (changes requested):

Hywan:
bors try

Makefile
```diff
@@ -1,112 +1,296 @@
...
+#
+#####
+
+
+IS_DARWIN := 0
+IS_LINUX := 0
+IS_WINDOWS := 0
+IS_AMD64 := 0
+IS_AARCH64 := 0
+LIBC ?=
```
jubianchi, line 75:
```suggestion
LIBC :=
```

Now that #2003 changed the build process for `musl` we can safely change the operator here. The only thing I'm wondering is will it break the pipeline in your PR? I don't know.

So two choices here:
* you change it in your PR
* I rebase #2003 on top of your patch and then do this change on my side.

Hywan:
Windows is failing with https://github.com/wasmerio/wasmer/runs/1904034596#step:14:21. It's not related to this PR. The other runs are still active, and it runs well for the moment!

Hywan:
Integration tests are failing. I'm investigating.

Hywan:
bors try

Hywan:
I don't know exactly _yet_ why integration tests are failing. A quick solution would be to build a `libwasmer` with Cranelift only, without LLVM, and run the integration tests with that.

Makefile
```diff
@@ -175,15 +159,24 @@ endif
...
+# The Cranelift case.
+##
 
-	ifeq ($(IS_WINDOWS), 0)
-		# `cranelift-native`
-		ifneq ($(LIBC), musl)
-			# Native engine doesn't work on Windows and musl yet.
+compilers_engines += cranelift-jit
+
+ifeq ($(IS_WINDOWS), 0)
```
nlewycky: MHO we should stop testing for "not OS <X>" and instead test for "is OS <X>".

Makefile
```diff
@@ -105,6 +105,10 @@ endif
 #
 #####
 
+HAS_CRANELIFT := 0
+HAS_LLVM := 0
+HAS_SINGLEPASS := 0
```
nlewycky, line 111: I have a preference for naming these `ENABLE_` and then using `?=` so that the user can override and specify `ENABLE_CRANELIFT=1 make` on a new or different OS or CPU which we don't yet support with autodetection.

Makefile
```diff
@@ -468,9 +468,26 @@ test-packages:
...
-# link the tests against. cargo test doesn't know that the tests will be running
-test-capi: $(foreach compiler_engine,$(compilers_engines),test-capi-$(compiler_engine))
+# We want to run all the tests for all available compilers. The C API
+# and the tests rely on the fact that one and only one default
+# compiler will be selected at compile-time. Therefore, if we want to
+# test exhaustively for all available compilers, we need to build and
+# to test the C API with a different compiler each time.
+#
+# That's exactly what `test-capi` does: it runs `test-capi-*` rules
+# that, prior to testing, builds the C API with `build-capi-*` sibling
```
nlewycky, line 475: I think the explanation would be clearer if it mentioned the steps it does in the order it did them, rather than #2 runs test-capi-* but first #1 it runs build-capi-*.

Makefile
```diff
@@ -468,9 +468,26 @@ test-packages:
...
+# only `cargo build`.
+#
+# Finally, `test-capi` calls `test-capi-all` that runs the tests for
+# the library built with `build-capi`, which is the one we will
+# deliver to the users.
+test-capi: $(foreach compiler_engine,$(compilers_engines),test-capi-$(compiler_engine)) test-capi-all
+
+test-capi-all: build-capi
+	cargo test --manifest-path lib/c-api/Cargo.toml --release \
+		--no-default-features --features deprecated,wat,jit,native,object-file,wasi $(capi_default_features) $(compiler_features) -- --nocapture
```
nlewycky, line 487: Is the `--nocapture` here intentional or debug cruft?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 ## **[Unreleased]**
 
 ### Added
+- [#2123](https://github.com/wasmerio/wasmer/pull/2123) `libwasmer` comes with all available compilers per targets instead of Cranelift only.
```
nlewycky, line 11:
```suggestion
- [#2123](https://github.com/wasmerio/wasmer/pull/2123) `libwasmer` comes with all available compilers per target instead of Cranelift only.
```

Makefile
```diff
@@ -72,7 +72,7 @@ IS_LINUX := 0
 IS_WINDOWS := 0
 IS_AMD64 := 0
 IS_AARCH64 := 0
-LIBC ?=
+LIBC :=
```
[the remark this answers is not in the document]
Hywan: Correct. Reverted!

Makefile
```diff
@@ -1,13 +1,101 @@
...
+# |          |              |            |             | musl       |         no |
+# |          |              |            | Object File | glibc      |        yes |
+# |          |              |            |             | musl       |         no |
+# |          |              |            |             |            |            |
+# |          |              | LLVM       | JIT         | glibc      |        yes |
+# |          |              |            | Native      | glibc      |        yes |
+# |          |              |            | Object File | glibc      |        yes |
+# |          |              |            |             |            |            |
+# |          |              | Singlepass | JIT         | glibc      |        yes |
+# |          |              |            | Native      | glibc      |        yes |
```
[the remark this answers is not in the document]
Hywan: Yes, the matrix has been updated in next commits.

Makefile
```diff
@@ -1,13 +1,101 @@
...
+	IS_WINDOWS := 1
 else
-	# We can assume, if in windows it will likely be in x86_64
-	ARCH := x86_64
-	UNAME_S :=
-	LIBC ?=
+	# Platform
+	if ($(shell uname -s), Darwin)
+		IS_DARWIN := 1
+	else
```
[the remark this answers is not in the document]
Hywan: Good idea. I've updated the PR to raise an error with `$(error)` if the platform isn't supported.

Makefile
```diff
@@ -1,13 +1,101 @@
...
+		IS_LINUX := 1
+	endif
+
+	# Architecture
+	if ($(shell uname -m), x86_64)
+		IS_AMD64 := 1
+	else
```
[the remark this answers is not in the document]
Hywan: Agreed. Updated!

Makefile
```diff
@@ -1,13 +1,101 @@
...
+		IS_AARCH64 := 1
+	endif
+
+	# Libc
+	LIBC ?= $(shell ldd 2>&1 | grep -o musl | head -n1)
```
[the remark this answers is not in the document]
Hywan: This code isn't new. It's from the existing `Makefile`: https://github.com/wasmerio/wasmer/blob/efeef22d4c847214cd44069f11e7541ce48ead01/Makefile#L5. I assumed it was correct.

I tried and yes, `ldd` with no argument always fail. cc @jubianchi any input as I think you're the author of this?

Makefile
```diff
@@ -171,39 +171,37 @@ endif
...
 #####
 
-# Compilers and engines to test again.
-#
-# In the form "$(compiler)-$(engine)" which `compiler` and `engine` combinations to test
-# in `make test`.
-test_compilers_engines :=
+# The engine is part of a pair of kind (compiler, engine). All the
+# pairs are stored in the `compilers_engines` variable.
+compilers_engines :=
```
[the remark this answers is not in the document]
Hywan: The `compilers_engines` set of pairs is only used by `test-*` rules, true, but I thought it was more meaningful to name the variable after what it contains, and not after what it is used for.

If you prefer to name the variable `test_compilers_engines`, I'm fine keeping it!

Makefile
Hywan: Yup. I was hesitant. `Makefile` has no logical OR operator. We need to use a trick like:

```
ifneq (, $(filter 1, $(IS_DARWIN) $(IS_LINUX)))
```

When you know it, that's OK, but at first glance, it feels weird.

Hywan: (code updated)

Makefile
Hywan: OK, that's a new feature then. The `HAS_*` “constants” are set if we detect the specific compiler. If we want the user to set them, then we should also update the `compilers` variable to include the compiler identifier for the `build` and `test` rules.

I'm updating the code accordingly. It's outside the goal of this PR but I think it's an interesting feature.

Hywan: Done.

We can use `ENABLE_CRANELIFT=1 ENABLE_LLVM=0 make build-capi` to force to enable Cranelift and to force to disable LLVM when running `build-capi`. I've tested all the combinations, and it works.

Makefile
Hywan: Agreed. Done!

Makefile
Hywan: Intentional. That's a pattern we use in all `test-capi-*` rules. It helps to display the outputs of `wasm-c-api` tests, and errors from other tests if they fail.

CHANGELOG.md
Hywan: Thanks!

Hywan:
bors try

Hywan:
Audit is failing, but it's not related to this PR. I'm ignoring it for the moment.

Makefile
```diff
@@ -1,13 +1,101 @@
...
+# |          |              |            | Native      | glibc      |        yes |
+# |          |              |            | Object File | glibc      |        yes |
+# |          |              |            |             |            |            |
+# |          |              | LLVM       | JIT         | glibc      |         no |
+# |          |              |            | Native      | glibc      |         no |
+# |          |              |            | Object File | glibc      |         no |
+# |          |              |            |             |            |            |
+# |          |              | Singlepass | JIT         | glibc      |        yes |
+# |          |              |            | Native      | glibc      |        yes |
+# |          |              |            | Object File | glibc      |        yes |
```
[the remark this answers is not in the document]
Hywan: I tried something. Let me know if it works for you! See bed96ed.

Hywan:
bors try

Hywan:
Since I'm modifying `.github/workflows/main.yaml`, `bors` can no longer run because this PR is from a branch of my fork; not the original repository.

Ive created https://github.com/wasmerio/wasmer/pull/2124 to run the tests.

MarkMcCaskey (commented):
Not reviewing any of the specifics, just want to say again that I don't think we can ship libwasmer with LLVM until we fix the linking issues regarding LLVM. Otherwise users will need zlib, curses, and potentially other libraries even when not building against LLVM. If this is an unfixable problem, which I don't think anyone is saying it is, compiler-llvm (or maybe every compiler) should be shipped in its own static library.

But my general take is, we can't ship this without causing a major regression in user experience until that issue is solved, so I'd vote not to ship libwasmer with all compilers yet.

Hywan:
I expect ncurses and zlib to not be required because it's absent of our LLVM custom builds, but it might be true for others.

Hywan:
@MarkMcCaskey With https://github.com/wasmerio/wasmer/pull/2123/commits/ed34e48e42f4624569225a5ed6d03fa1f5316875, LLVM is always removed from the C API when built with `make build-capi` or `make test-capi`.

Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

Hywan:
Run is here https://github.com/wasmerio/wasmer/runs/1912092129

Hywan:
PR is green, modulo our sccache issue on Windows.

nlewycky:
Update to pick up fix to sccache issue.

nlewycky:
bors try

syrusakbary:
bors try

syrusakbary:
bors try-

syrusakbary:
bors try

syrusakbary:
> Not reviewing any of the specifics, just want to say again that I don't think we can ship libwasmer with LLVM until we fix the linking issues regarding LLVM. Otherwise users will need zlib, curses, and potentially other libraries even when not building against LLVM. If this is an unfixable problem, which I don't think anyone is saying it is, compiler-llvm (or maybe every compiler) should be shipped in its own static library.

@jubianchi can you check if that still the case (zlib/curses requirement when embedding LLVM in the wasmer c api)?
I thought we fixed by using our custom llvm builds, but that might not be the case

Hywan:
bors try

Makefile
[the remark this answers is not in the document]
jubianchi: on alpine, calling `ldd` without any argument produces the following output:

```
musl libc (x86_64)
Version 1.2.2
Dynamic Program Loader
Usage: /lib/ld-musl-x86_64.so.1 [options] [--] pathname
```

With this out put we have what we need to find if we are on musl or not.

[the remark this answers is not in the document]
Hywan: Good for me then.

jubianchi (approved):

PR #2125, opened 2021-02-16
chore(cargo) Fix RUSTSEC-2021-0023.

Hywan:
# Description

This patch updates `rand_core` from 0.6.1 to 0.6.2 as a vulnerability
has been discovered. See https://rustsec.org/advisories/RUSTSEC-2021-0023.

This issue has been reported by our Audit workflow in the CI.

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - [#2113](https://github.com/wasmerio/wasmer/pull/2113) Bump minimum supported Rust version to 1.49
 
 ### Fixed
+- [#2125](https://github.com/wasmerio/wasmer/pull/2125) Fix RUSTSEC-2021-0023; `rand_core`: incorrect check on buffer length when seeding RNGs.
 - [#2097](https://github.com/wasmerio/wasmer/pull/2097) Fix how string's length is computed in `wasm_cpu_features_add` in the C API.
 - [#2101](https://github.com/wasmerio/wasmer/pull/2101) cflags emitted by `wasmer config --pkg-config` are now correct.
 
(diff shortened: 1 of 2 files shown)


Hywan:
bors try

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - [#2113](https://github.com/wasmerio/wasmer/pull/2113) Bump minimum supported Rust version to 1.49
 
 ### Fixed
+- [#2125](https://github.com/wasmerio/wasmer/pull/2125) Fix RUSTSEC-2021-0023; `rand_core`: incorrect check on buffer length when seeding RNGs.
```
MarkMcCaskey, line 17: This doesn't affect us, might want to mention that in the changelog, or remove this from the changelog entirely.

MarkMcCaskey (approved):
Looks good! But I don't think we use seeded RNG anywhere so I don't think this bug affects us at all -- it's good to upgrade though!

CHANGELOG.md
Hywan: I can remove it from the changelog, yes.

Hywan, line 17:
```suggestion
```

Hywan:
We need to upgrade because our Audit workflow is failing due to that.

MarkMcCaskey:
> We need to upgrade because our Audit workflow is failing due to that.

I think it's the right thing to do in any case! Even though we're not affected by that bug right now, we easily could decide that we want to use seeded RNG in the future, so it's best to make sure it works! Thanks for the PR!

MarkMcCaskey:
Merging manually because of multiple CI breakages: this change seems safe and will unblock part of CI

PR #2131, opened 2021-02-18
Add test that shows local memory/table lifetime bug

MarkMcCaskey:
The issue is that we expect these things to work while the store/engine is alive but critical data is stored in the `Instance` itself and the `Instance` does not live as long as the engine/store.

# Review


Hywan:
@syrusakbary Is it closed by #2281?

epilys:
No longer relevant since objects are now owned by the Context/Store and the actual `Memory`, `Global` etc are id handles to the allocated objects (see #2892)

PR #2133, opened 2021-02-19
fix: Use space-indent when calling functions outside of rules

jubianchi:
This should fix failures in:
* https://github.com/wasmerio/wasmer-nightly/pull/14/checks?check_run_id=1938612785
* https://github.com/wasmerio/wasmer-nightly/pull/13/checks?check_run_id=1938633577

This will unblock #2003


Hywan (approved):
bors r+

PR #2141, opened 2021-02-25
feat(types) Avoid allocating a `Vec` when calling `FunctionType::new`

Hywan:
# Description

Sequel of https://github.com/wasmerio/wasmer/pull/2036. We can go further by defining `Params` and `Results` of `FunctionType::new` as `Into<Box<[Type]>>` directly rather than `Into<Vec<Type>>`. It simplifies the code: `params.into()` rather than `params.into().into_boxed_slice()`. I assume it doesn't allocate an intermediate vector.

Since `From<Box<[T]>>` is implemented for `Vec<T>`, I reckon it's not a BC break.

# Review

### DIFF
--- a/lib/api/src/import_object.rs
+++ b/lib/api/src/import_object.rs
@@ -145,20 +145,6 @@ impl IntoIterator for ImportObject {
 
 impl fmt::Debug for ImportObject {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        enum SecretOption {
-            None,
-            Some,
-        }
-
-        impl fmt::Debug for SecretOption {
-            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-                match self {
-                    Self::None => write!(f, "None"),
-                    Self::Some => write!(f, "Some(...)"),
-                }
-            }
-        }
-
         enum SecretMap {
             Empty,
             Some(usize),
--- a/lib/wasmer-types/src/types.rs
+++ b/lib/wasmer-types/src/types.rs
@@ -238,12 +238,12 @@ impl FunctionType {
     /// Creates a new Function Type with the given parameter and return types.
     pub fn new<Params, Returns>(params: Params, returns: Returns) -> Self
     where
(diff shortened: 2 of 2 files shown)


Hywan:
bors try

nlewycky (approved):

Hywan:
The CI failure has nothing related to this PR. Weird.

lib/api/src/import_object.rs
```diff
@@ -145,20 +145,6 @@ impl IntoIterator for ImportObject {
...
-
-        impl fmt::Debug for SecretOption {
-            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-                match self {
-                    Self::None => write!(f, "None"),
-                    Self::Some => write!(f, "Some(...)"),
-                }
-            }
-        }
-
```
MarkMcCaskey, line 161: This change seems entirely unrelated. Not sure if you intended to commit it

MarkMcCaskey (commented):
> Since From<Box<[T]>> is implemented for Vec<T>, I reckon it's not a BC break.

`From<Box<[T]>>` for `Vec<T>` means that we can go from `Box<T> -> Vec<T>`, did you mean `Into<Box<[T]>>` for `Vec<T>`? If so that's better but I think it's still a breaking change because for a given type `GT`, if `GT` implements `Into<Vec<T>>` but not `Into<Box<[T]>>` the caller code now has to do `gt.into().into()` (or because we call one of the `into`s they'd have to change from `FunctionType::new(g, ...)` to `FunctionType::new(g.into(), ...)`

I think this change is probably a good one but I do think it's probably a breaking change. We have lots of (generally very minor like this) breaking changes coming anyways though, we just need to decide when we want to ship a release to get them all in

lib/api/src/import_object.rs
Hywan: It's deadcode, raising a warning from `rustc`.

MarkMcCaskey (approved):
approved barring breaking changes, as you mentioned we still need to figure out how to manage them

PR #2149, opened 2021-02-28
Update engine-native to look for clang-11

coolreader18:
# Description
It was confusing getting our CI to work since using the `apt.llvm.org/llvm.sh` to install llvm11 doesn't install a clang or clang-10 binary, so I updated `engine-native` to look for clang-11 instead of clang-10, since that would make it easy to just install one version of clang on CI and get over with it.

# Review

### DIFF
--- a/lib/engine-native/src/engine.rs
+++ b/lib/engine-native/src/engine.rs
@@ -177,8 +177,8 @@ impl Engine for NativeEngine {
 #[derive(Clone, Copy)]
 pub(crate) enum Linker {
     None,
-    // this should be kept in sync with the llvm version required for wasmer-compiler-llvm
     Clang11,
+    Clang10,
     Clang,
     Gcc,
 }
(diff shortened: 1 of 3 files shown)


lib/engine-native/src/engine.rs
```diff
@@ -193,16 +177,40 @@ impl Engine for NativeEngine {
 #[derive(Clone, Copy)]
 pub(crate) enum Linker {
     None,
-    Clang10,
+    // this should be kept in sync with the llvm version required for wasmer-compiler-llvm
```
nlewycky, line 180: Actually, this does not need to be kept in sync with the compiler-llvm version! I'm not sure what the oldest version that works is, but whatever version the user has installed up to some minimum (that is probably around clang 6) will work fine. Since we know clang-10 works, I'd appreciate it if we could keep support for it as we add clang-11.

syrusakbary (approved):

coolreader18:
I think it would be good/convenient, though, if supported `clang-*` were kept up to date with the llvm version, so that we don't need to install multiple llvm toolchains.

Hywan:
Thanks!

PR #2154, opened 2021-03-01
Implement reference types in compiler-llvm.

nlewycky:


lib/compiler-llvm/src/abi/x86_64_systemv.rs
```diff
@@ -316,26 +315,7 @@ impl Abi for X86_64SystemV {
...
-                    );
-                    builder.build_bitcast(value, intrinsics.f64_ty, "")
-                }
-                Type::V128 => {
-                    assert!(value.get_type() == intrinsics.i128_ty.as_basic_type_enum());
-                    value
-                }
-                Type::ExternRef => unimplemented!("externref in the llvm backend"),
-                Type::FuncRef => unimplemented!("funcref in the llvm backend"),
+                _ => panic!("should only be called to repack 32-bit values"),
```
MarkMcCaskey, line 318: Just some other bug you noticed?

lib/compiler-llvm/src/object_file.rs
```diff
@@ -54,15 +54,65 @@ where
...
+    libcalls.insert("truncf".to_string(), LibCall::TruncF32);
+    libcalls.insert("trunc".to_string(), LibCall::TruncF64);
+    libcalls.insert("wasmer_f32_ceil".to_string(), LibCall::CeilF32);
+    libcalls.insert("wasmer_f64_ceil".to_string(), LibCall::CeilF64);
+    libcalls.insert("wasmer_f32_floor".to_string(), LibCall::FloorF32);
+    libcalls.insert("wasmer_f64_floor".to_string(), LibCall::FloorF64);
+    libcalls.insert("wasmer_f32_nearest".to_string(), LibCall::NearestF32);
+    libcalls.insert("wasmer_f64_nearest".to_string(), LibCall::NearestF64);
+    libcalls.insert("wasmer_f32_trunc".to_string(), LibCall::TruncF32);
+    libcalls.insert("wasmer_f64_trunc".to_string(), LibCall::TruncF64);
```
MarkMcCaskey, line 72: Are we using these?

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -9344,6 +9344,192 @@ impl<'ctx, 'a> LLVMFunctionCodeGenerator<'ctx, 'a> {
...
+             ***************************/
+            Operator::RefNull { .. } => {
+                self.state.push1(self.intrinsics.funcref_ty.const_null());
+            }
+            Operator::RefIsNull => {
+                let value = self.state.pop1()?.into_pointer_value();
+                let is_null = self.builder.build_is_null(value, "");
+                let is_null = self
+                    .builder
+                    .build_int_z_extend(is_null, self.intrinsics.i32_ty, "");
```
MarkMcCaskey, line 9359: Is this because LLVM doesn't use condition flags / special registers? I'm surprised it doesn't also have the concept of a bool type for this then

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -9344,6 +9344,192 @@ impl<'ctx, 'a> LLVMFunctionCodeGenerator<'ctx, 'a> {
...
+                    .intrinsics
+                    .i32_ty
+                    .const_int(segment as u64, false)
+                    .as_basic_value_enum();
+                self.builder.build_call(
+                    self.intrinsics.elem_drop,
+                    &[self.ctx.basic(), segment],
+                    "",
+                );
+            }
```
MarkMcCaskey, line 9470: Interesting, singlepass was also missing this. It must be part of the Wasm spec now though because the ref types proposal assumes it's implemented and it's not part of the ref types proposal 🤔

lib/compiler-llvm/src/translator/intrinsics.rs
```diff
@@ -39,10 +39,8 @@ pub fn type_to_llvm_ptr<'ctx>(
         Type::F32 => Ok(intrinsics.f32_ptr_ty),
         Type::F64 => Ok(intrinsics.f64_ptr_ty),
         Type::V128 => Ok(intrinsics.i128_ptr_ty),
-        ty => Err(CompileError::Codegen(format!(
-            "type_to_llvm: unimplemented wasmer_types type {:?}",
-            ty
-        ))),
+        Type::FuncRef => Ok(intrinsics.funcref_ty.ptr_type(AddressSpace::Generic)),
+        Type::ExternRef => Ok(intrinsics.funcref_ty.ptr_type(AddressSpace::Generic)),
```
MarkMcCaskey, line 43: Typo? This says `ExternRef` is a pointer to funcref_ty

lib/compiler-llvm/src/translator/intrinsics.rs
```diff
@@ -56,10 +54,8 @@ pub fn type_to_llvm<'ctx>(
         Type::F32 => Ok(intrinsics.f32_ty.as_basic_type_enum()),
         Type::F64 => Ok(intrinsics.f64_ty.as_basic_type_enum()),
         Type::V128 => Ok(intrinsics.i128_ty.as_basic_type_enum()),
-        ty => Err(CompileError::Codegen(format!(
-            "type_to_llvm: unimplemented wasmer_types type {:?}",
-            ty
-        ))),
+        Type::FuncRef => Ok(intrinsics.funcref_ty.as_basic_type_enum()),
+        Type::ExternRef => Ok(intrinsics.funcref_ty.as_basic_type_enum()),
```
MarkMcCaskey, line 58: here too

lib/compiler-llvm/src/translator/intrinsics.rs
```diff
@@ -460,6 +465,145 @@ impl<'ctx> Intrinsics<'ctx> {
...
+                None,
+            ),
+            table_fill: module.add_function(
+                "wasmer_table_fill",
+                void_ty.fn_type(
+                    &[
+                        ctx_ptr_ty.as_basic_type_enum(),
+                        i32_ty_basic,
+                        i32_ty_basic,
+                        funcref_ty.as_basic_type_enum(),
```
MarkMcCaskey, line 506: here as well, is `funcref_ty` the thing we're using for all reference types? Might be better to give it a different name when it's used in opaque contexts, as we do deref a funcref_ty in call indirect

lib/compiler-llvm/src/translator/intrinsics.rs
```diff
@@ -460,6 +465,145 @@ impl<'ctx> Intrinsics<'ctx> {
...
+                        i32_ty_basic,
+                        i32_ty_basic,
+                    ],
+                    false,
+                ),
+                None,
+            ),
+            func_ref: module.add_function(
+                "wasmer_func_ref",
+                funcref_ty.fn_type(&[i8_ptr_ty_basic, i32_ty_basic], false),
```
MarkMcCaskey, line 593: Does this not take a context pointer? I thought they all did because that's what the backends did by default

lib/compiler-llvm/src/translator/intrinsics.rs
```diff
@@ -460,6 +465,145 @@ impl<'ctx> Intrinsics<'ctx> {
...
+                "wasmer_elem_drop",
+                void_ty.fn_type(&[i8_ptr_ty_basic, i32_ty_basic], false),
+                None,
+            ),
+            throw_trap: module.add_function(
+                "wasmer_raise_trap",
+                void_ty.fn_type(&[i32_ty_basic], false),
+                None,
+            ),
+
```
MarkMcCaskey, line 606: check these too

MarkMcCaskey (approved):
Other than a few small issues, if it's passing all the spectests then this looks good to me!

Feel free to leave feedback based on anything you learned implementing this. This actually turned out to be a shallower task than I originally anticipated, so it won't help as much with reviewing the major internal changes.

lib/compiler-llvm/src/translator/code.rs
nlewycky: build_is_null returns an `i1` (a one-bit integer, so a bool), but webassembly only has 32 and 64 bit integers. This instruction is specified as producing a 32-bit integer, so we extend it here.

lib/compiler-llvm/src/object_file.rs
nlewycky: Nope!

lib/compiler-llvm/src/abi/x86_64_systemv.rs
nlewycky: On x86-64 this function is only called to cast 32-bit values, and should only be called to cast 32-bit values. I think this was copied out of the aarch64 code. I found this because I was auditing for places where ExternRef/FuncRef weren't handled and then investigated.

lib/compiler-llvm/src/translator/intrinsics.rs
nlewycky: Nope! intrinsics does not have a separate externref_ty. We could all the name and just make it equal to funcref_ty in intrisics.rs?

lib/compiler-llvm/src/translator/intrinsics.rs
nlewycky: Yep that's a bug, thanks!

lib/compiler-llvm/src/translator/intrinsics.rs
nlewycky: raise_trap is correct.

lib/compiler-llvm/src/translator/intrinsics.rs
MarkMcCaskey: I think it may make sense to be more precise here. I'd be in favor of a `reference_ty` or `anyref_ty` or something to indicate it's an opaque pointer that we don't know / care about the specifics of, when that's the case like in `table.set`.

The main reason is just that we use `funcref_ty` semantically somewhere and it's also a bit confusing to read.

lib/compiler-llvm/src/abi/aarch64_systemv.rs
```diff
@@ -140,8 +140,7 @@ impl Abi for Aarch64SystemV {
                         Type::I32 | Type::F32 => 32,
                         Type::I64 | Type::F64 => 64,
                         Type::V128 => 128,
-                        Type::ExternRef => unimplemented!("externref in the llvm backend"),
-                        Type::FuncRef => unimplemented!("funcref in the llvm backend"),
+                        Type::ExternRef | Type::FuncRef => 64, /* pointer */
```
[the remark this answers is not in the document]
nlewycky: The presence of VMOffsets is optional in this function. The issue is that we don't have a VMContext when producing trampolines, so we don't have a VMOffset for it either.

lib/compiler-llvm/src/translator/intrinsics.rs
nlewycky: We already have an `anyref_ty` and in fact `funcref_ty` is defined as a pointer-to-anyref_ty!

What's an anyref_ty anyways? Is that the right name for what it is?

MarkMcCaskey: Yeah, I think that name might be confusing. `funcref` should be a pointer to a specific struct. I believe that's called a `VMCallerCheckedAnyfunc` I'd expect an `anyref` to be an opaque pointer. It's probably safe to assume that this opaque pointer at least requires alignment of 1 word, unsure if that really matters, but it should be a safe assumption.

lib/compiler-llvm/src/abi/aarch64_systemv.rs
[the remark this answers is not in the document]
MarkMcCaskey: Yeah, it's fine. Really not a big deal because we do this thing in other places in compiler-llvm where we don't use data from `VMOffsets` so things may get out of sync. I think it's fine to not worry about it for now and we can update compiler-llvm later if we decide that that's a high enough priority thing to do.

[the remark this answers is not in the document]
MarkMcCaskey: Really, a lot of `VMOffsets` stuff is static. In this case I don't know if we actually need a `VMOffsets` but, yeah, generally speaking getting the size of things doesn't require anything other than knowing the size of a pointer on the target architecture. Maybe that means it needs to be split into two pieces

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -9344,6 +9344,213 @@ impl<'ctx, 'a> LLVMFunctionCodeGenerator<'ctx, 'a> {
                 size.add_attribute(AttributeLoc::Function, self.intrinsics.readonly);
                 self.state.push1(size.try_as_basic_value().left().unwrap());
             }
+            /***************************
+             * Reference types.
+             * https://github.com/WebAssembly/reference-types/blob/master/proposals/reference-types/Overview.md
+             ***************************/
+            Operator::RefNull { .. } => {
+                self.state.push1(self.intrinsics.funcref_ty.const_null());
+            }
```
MarkMcCaskey, line 9353: This function takes a `type` which will tell you if it's a funcref or an externref

nlewycky: Good catch, thanks. Fixed!

PR #2157, opened 2021-03-02
Simplify WasmPtr code

MarkMcCaskey:
`std::mem::size_of::<T>()` already accounts for end padding. The `align_of` code
was not only a no-op, it would be incorrect if it wasn't.

Thanks to Matt Itamenye on the wasmerians slack for pointing this out!

This PR does not change any behavior, only removes misleading code.


Hywan (approved):
You may want to take a look at [`Layout`](https://doc.rust-lang.org/std/alloc/struct.Layout.html), but `mem::size_of` does the job in this case.

nlewycky (approved):

syrusakbary (approved):

PR #2161, opened 2021-03-03
feat(llvm): Make NaN canonicalization configurable

jubianchi:
compiler-llvm now uses the experimental.constrained intrinsics to ensure
correct behavior on FP operations when full-canonicalization is
disabled.
    
This patch requires TheDan64/inkwell#247

# Review

### DIFF
--- a/lib/c-api/build.rs
+++ b/lib/c-api/build.rs
@@ -474,6 +474,7 @@ fn exclude_items_from_wasm_c_api(builder: Builder) -> Builder {
         .exclude_item("wasi_get_unordered_imports")
         .exclude_item("wasi_get_wasi_version")
         .exclude_item("wasi_version_t")
+        .exclude_item("wasm_config_canonicalize_nans")
         .exclude_item("wasm_config_push_middleware")
         .exclude_item("wasm_config_set_compiler")
         .exclude_item("wasm_config_set_engine")
--- a/lib/c-api/src/wasm_c_api/engine.rs
+++ b/lib/c-api/src/wasm_c_api/engine.rs
@@ -105,6 +105,7 @@ pub struct wasm_config_t {
     compiler: wasmer_compiler_t,
     #[cfg(feature = "middlewares")]
     pub(super) middlewares: Vec<wasmer_middleware_t>,
+    pub(super) nan_canonicalization: bool,
     pub(super) features: Option<Box<wasmer_features_t>>,
     pub(super) target: Option<Box<wasmer_target_t>>,
 }
@@ -480,6 +481,10 @@ pub extern "C" fn wasm_engine_new_with_config(
             #[cfg(feature = "middlewares")]
             for middleware in config.middlewares {
                 compiler_config.push_middleware(middleware.inner);
+            }
+
+            if config.nan_canonicalization {
+                compiler_config.canonicalize_nans(true);
             }
 
             let inner: Arc<dyn Engine + Send + Sync> = match config.engine {
--- a/lib/compiler-cranelift/src/config.rs
+++ b/lib/compiler-cranelift/src/config.rs
@@ -185,6 +185,14 @@ impl CompilerConfig for Cranelift {
         self.enable_verifier = true;
     }
 
+    fn enable_nan_canonicalization(&mut self) {
+        self.enable_nan_canonicalization = true;
+    }
+
+    fn canonicalize_nans(&mut self, enable: bool) {
+        self.enable_nan_canonicalization = enable;
+    }
+
     /// Transform it into the compiler
     fn compiler(self: Box<Self>) -> Box<dyn Compiler> {
         Box::new(CraneliftCompiler::new(*self))
--- a/lib/compiler-llvm/Cargo.toml
+++ b/lib/compiler-llvm/Cargo.toml
@@ -25,7 +25,8 @@ rayon = "1.5"
 loupe = "0.1"
 
 [dependencies.inkwell]
-version = "=0.1.0-beta.2"
+git = "https://github.com/wasmerio/inkwell"
+branch = "add-metadata-support"
 default-features = false
 features = ["llvm11-0", "target-x86", "target-aarch64"]
 
--- a/lib/compiler-llvm/Cargo.toml
+++ b/lib/compiler-llvm/Cargo.toml
@@ -25,8 +25,8 @@ rayon = "1.5"
 loupe = "0.1"
 
 [dependencies.inkwell]
-git = "https://github.com/wasmerio/inkwell"
-branch = "add-metadata-support"
+package = "wasmer_inkwell"
+version = "0.2.0-alpha.2"
 default-features = false
 features = ["llvm11-0", "target-x86", "target-aarch64"]
 
--- a/lib/compiler-llvm/src/abi/aarch64_systemv.rs
+++ b/lib/compiler-llvm/src/abi/aarch64_systemv.rs
@@ -85,8 +85,8 @@ impl Abi for Aarch64SystemV {
             [] => (
                 intrinsics.void_ty.fn_type(
                     param_types
-                        .map(|v| v.unwrap().into())
-                        .collect::<Vec<BasicMetadataTypeEnum>>()
+                        .map(|v| v.map(Into::into))
+                        .collect::<Result<Vec<BasicMetadataTypeEnum>, _>>()?
                         .as_slice(),
                     false,
                 ),
@@ -97,8 +97,8 @@ impl Abi for Aarch64SystemV {
                 (
                     type_to_llvm(intrinsics, single_value)?.fn_type(
                         param_types
-                            .map(|v| v.unwrap().into())
-                            .collect::<Vec<BasicMetadataTypeEnum>>()
+                            .map(|v| v.map(Into::into))
+                            .collect::<Result<Vec<BasicMetadataTypeEnum>, _>>()?
                             .as_slice(),
                         false,
                     ),
@@ -110,8 +110,8 @@ impl Abi for Aarch64SystemV {
                 (
                     context.struct_type(&[f32_ty, f32_ty], false).fn_type(
                         param_types
-                            .map(|v| v.unwrap().into())
-                            .collect::<Vec<BasicMetadataTypeEnum>>()
+                            .map(|v| v.map(Into::into))
+                            .collect::<Result<Vec<BasicMetadataTypeEnum>, _>>()?
                             .as_slice(),
                         false,
                     ),
@@ -123,8 +123,8 @@ impl Abi for Aarch64SystemV {
                 (
                     context.struct_type(&[f64_ty, f64_ty], false).fn_type(
                         param_types
-                            .map(|v| v.unwrap().into())
-                            .collect::<Vec<BasicMetadataTypeEnum>>()
+                            .map(|v| v.map(Into::into))
+                            .collect::<Result<Vec<BasicMetadataTypeEnum>, _>>()?
                             .as_slice(),
                         false,
                     ),
@@ -138,8 +138,8 @@ impl Abi for Aarch64SystemV {
                         .struct_type(&[f32_ty, f32_ty, f32_ty], false)
                         .fn_type(
                             param_types
-                                .map(|v| v.unwrap().into())
-                                .collect::<Vec<BasicMetadataTypeEnum>>()
+                                .map(|v| v.map(Into::into))
+                                .collect::<Result<Vec<BasicMetadataTypeEnum>, _>>()?
                                 .as_slice(),
                             false,
                         ),
@@ -153,8 +153,8 @@ impl Abi for Aarch64SystemV {
                         .struct_type(&[f32_ty, f32_ty, f32_ty, f32_ty], false)
                         .fn_type(
                             param_types
-                                .map(|v| v.unwrap().into())
-                                .collect::<Vec<BasicMetadataTypeEnum>>()
+                                .map(|v| v.map(Into::into))
+                                .collect::<Result<Vec<BasicMetadataTypeEnum>, _>>()?
                                 .as_slice(),
                             false,
                         ),
@@ -176,8 +176,8 @@ impl Abi for Aarch64SystemV {
                     [32, 32] => (
                         intrinsics.i64_ty.fn_type(
                             param_types
-                                .map(|v| v.unwrap().into())
-                                .collect::<Vec<BasicMetadataTypeEnum>>()
+                                .map(|v| v.map(Into::into))
+                                .collect::<Result<Vec<BasicMetadataTypeEnum>, _>>()?
                                 .as_slice(),
                             false,
                         ),
@@ -192,8 +192,8 @@ impl Abi for Aarch64SystemV {
                     | [32, 32, 32, 32] => (
                         intrinsics.i64_ty.array_type(2).fn_type(
                             param_types
-                                .map(|v| v.unwrap().into())
-                                .collect::<Vec<BasicMetadataTypeEnum>>()
+                                .map(|v| v.map(Into::into))
+                                .collect::<Result<Vec<BasicMetadataTypeEnum>, _>>()?
                                 .as_slice(),
                             false,
                         ),
@@ -225,8 +225,8 @@ impl Abi for Aarch64SystemV {
                         (
                             intrinsics.void_ty.fn_type(
                                 param_types
-                                    .map(|v| v.unwrap().into())
-                                    .collect::<Vec<BasicMetadataTypeEnum>>()
+                                    .map(|v| v.map(Into::into))
+                                    .collect::<Result<Vec<BasicMetadataTypeEnum>, _>>()?
                                     .as_slice(),
                                 false,
                             ),
--- a/lib/compiler-llvm/src/abi/aarch64_systemv.rs
+++ b/lib/compiler-llvm/src/abi/aarch64_systemv.rs
@@ -4,7 +4,7 @@ use inkwell::{
     attributes::{Attribute, AttributeLoc},
     builder::Builder,
     context::Context,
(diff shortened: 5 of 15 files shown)


jubianchi:
bors try

jubianchi:
bors try

lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -99,6 +99,7 @@ pub struct wasm_config_t {
     engine: wasmer_engine_t,
     #[cfg(feature = "compiler")]
     compiler: wasmer_compiler_t,
+    nan_canonicalization: bool,
```
MarkMcCaskey, line 102: If this is for the spec-test compliant part of nan canonicalization and not the "canonicalize everywhere" thing we also offer, I'd make it more explicit that this will make the Wasm VM behave in a non-spec compliant way. Probably in the name of the function or something, include the word "unsafe", like what React does.

jubianchi:
bors try

lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -269,6 +270,45 @@ pub extern "C" fn wasm_config_set_engine(config: &mut wasm_config_t, engine: was
...
+///     wasm_engine_delete(engine);
+///
+///     return 0;
+/// }
+/// #    })
+/// #    .success();
+/// # }
+/// ```
+#[no_mangle]
+pub extern "C" fn wasm_config_enable_nan_canonicalization(config: &mut wasm_config_t) {
```
syrusakbary, line 308: It might be a good idea to add a second variable (enable: true/false), in case someone want to overwrite the default value (and that default value is true)

lib/compiler-singlepass/src/config.rs
```diff
@@ -19,7 +19,7 @@ impl Singlepass {
     /// specified.
     pub fn new() -> Self {
         Self {
-            enable_nan_canonicalization: true,
+            enable_nan_canonicalization: false,
```
syrusakbary, line 22: I think since singlepass is mainly directed for blockchains, which need deterministic results by default, this value should be true by default.

lib/compiler/src/compiler.rs
```diff
@@ -36,6 +36,12 @@ pub trait CompilerConfig {
         // in case they create an IR that they can verify.
     }
 
+    /// Enable NaN canonicalization.
+    fn enable_nan_canonicalization(&mut self) {
+        // By default we do nothing, each backend will need to customize this
+        // in case they create an IR that they can verify.
+    }
```
syrusakbary, line 43: Receive another variable, and make the function not have a default implementation.

```suggestion
    fn enable_nan_canonicalization(&mut self, enable: bool);
```

tests/ignores.txt
```diff
@@ -34,6 +34,9 @@ cranelift::spec::simd::simd_i8x16_sat_arith on aarch64
 cranelift::spec::simd::simd_lane on aarch64
 cranelift::spec::skip_stack_guard_page on aarch64
 
+# TODO(https://github.com/wasmerio/wasmer/pull/2161): Need to review NaN canonicalization
+llvm::spec::float_exprs
+llvm::spec::simd::simd_f64x2_arith
```
syrusakbary, line 39: TODO: Don't merge this. When testing, we should enable the canonical flag.

lib/compiler/src/compiler.rs
nlewycky: If it takes a bool to enable or disable, maybe `set_nan_canonicalization` would be preferred? Or `set_canonicalize_all_nans`?

syrusakbary: Yeah, perhaps naming can be better in that case!

syrusakbary: For the Wasm features we simply put the name directly (without set/enable).
https://github.com/wasmerio/wasmer/blob/master/lib/wasmer-types/src/features.rs#L107-L110

Perhaps
`canonicalize_nans(&mut self, enable: bool);` ?

lib/c-api/src/wasm_c_api/engine.rs
Hywan: This function must move into the `unstable/engine.rs` module for the moment.
Also, I propose the following API: `wasm_config_nan_canonicalization(config;: &mut wasm_config_t, enable: bool)`. Thoughts?

Finally, this function should be added inside the `build.rs` script, inside the `exclude_items_from_wasm_c_api` function, so that it's not present in the `wasmer.h*` header files.

lib/compiler-llvm/src/config.rs
```diff
@@ -53,7 +53,7 @@ impl LLVM {
     /// specified.
     pub fn new() -> Self {
         Self {
-            enable_nan_canonicalization: false,
+            enable_nan_canonicalization: true,
```
syrusakbary, line 56: Lets have it false by default

tests/wast/wasmer/nan-canonicalization.wast
```diff
@@ -153,45 +153,45 @@
...
 (assert_return (invoke "nan-canonicalization-f64-local" (i64.const 0x7ff8000000000001)) (i64.const 0x7ff8000000000001))
-(assert_return (invoke "nan-canonicalization-f64-local-cncl" (i64.const 0x7ff8000000000001)) (i64.const 0x7ff8000000000000))
+(assert_return (invoke "nan-canonicalization-f64-local-cncl" (i64.const 0x7ff8000000000001)) (i64.const 0x7ff8000000000001))
 (assert_return (invoke "nan-canonicalization-f64-func-call" (i64.const 0x7ff8000000000001)) (i64.const 0x7ff8000000000001))
-(assert_return (invoke "nan-canonicalization-f64-func-call-cncl" (i64.const 0x7ff8000000000001)) (i64.const 0x7ff8000000000000))
+(assert_return (invoke "nan-canonicalization-f64-func-call-cncl" (i64.const 0x7ff8000000000001)) (i64.const 0x7ff8000000000001))
 (assert_return (invoke "nan-canonicalization-f64-func-call-indirect" (i64.const 0x7ff8000000000001)) (i64.const 0x7ff8000000000001))
-(assert_return (invoke "nan-canonicalization-f64-func-call-indirect-cncl" (i64.const 0x7ff8000000000001)) (i64.const 0x7ff8000000000000))
+(assert_return (invoke "nan-canonicalization-f64-func-call-indirect-cncl" (i64.const 0x7ff8000000000001)) (i64.const 0x7ff8000000000001))
```
syrusakbary, line 194: Will this change work with the other compilers??

syrusakbary: Ok, I think we have to replace all this numbers with:
* `0x7ff8000000000001` => `nan`
* `0x7ff8000000000000` => `nan:canonical`

jubianchi:
bors try

jubianchi:
bors try

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -3495,201 +3261,542 @@ impl<'ctx, 'a> LLVMFunctionCodeGenerator<'ctx, 'a> {
...
-                let ((v1, i1), (v2, i2)) = self.state.pop2_extra()?;
-                let (v1, v2) = (v1.into_float_value(), v2.into_float_value());
-                let res = self.builder.build_float_add(v1, v2, "");
-                self.state.push1_extra(
-                    res,
-                    (i1.strip_pending() & i2.strip_pending()) | ExtraInfo::pending_f32_nan(),
-                );
+                let (v1, v2) = self.state.pop2()?;
+                let v1 = self.canonicalize_nans(v1);
+                let v2 = self.canonicalize_nans(v2);
```
nlewycky, line 3266: Using `ExtraInfo` we used to be able to detect when the NaN is certainly already canonical and not re-canonicalize it here

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -3495,201 +3261,542 @@ impl<'ctx, 'a> LLVMFunctionCodeGenerator<'ctx, 'a> {
...
+                let res = self
+                    .builder
+                    .build_call(
+                        self.intrinsics.add_f32,
+                        &[
+                            v1,
+                            v2,
+                            inkwell::values::BasicValueEnum::MetadataValue(
+                                self.intrinsics.fp_rounding_md,
```
nlewycky, line 3275: It'd be nice if this was named to say what the rounding mode is. Similarly for exception.

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -3835,21 +3942,15 @@ impl<'ctx, 'a> LLVMFunctionCodeGenerator<'ctx, 'a> {
...
+                let v1 = self.v128_into_f32x4(v1);
+                let v2 = self.v128_into_f32x4(v2);
+                let v1 = self
+                    .canonicalize_nans(v1.as_basic_value_enum())
+                    .into_vector_value();
+                let v2 = self
+                    .canonicalize_nans(v2.as_basic_value_enum())
+                    .into_vector_value();
 
                 let v1_is_nan = self.builder.build_float_compare(
```
nlewycky, line 3955: This much be replaced with a call to `llvm.experimental.constrained.fcmp`:
> If any FP operation in a function is constrained then they all must be constrained. This is required for correct LLVM IR.
-- https://llvm.org/docs/LangRef.html#constrained-floating-point-intrinsics

Please audit for any other non-constrainted floating point operations.

lib/compiler-llvm/src/translator/intrinsics.rs
```diff
@@ -70,11 +71,44 @@ pub struct Intrinsics<'ctx> {
...
+    pub max_f32: FunctionValue<'ctx>,
+    pub max_f64: FunctionValue<'ctx>,
+    pub max_f32x4: FunctionValue<'ctx>,
+    pub max_f64x2: FunctionValue<'ctx>,
+
+    pub min_f32: FunctionValue<'ctx>,
+    pub min_f64: FunctionValue<'ctx>,
+    pub min_f32x4: FunctionValue<'ctx>,
+    pub min_f64x2: FunctionValue<'ctx>,
+
```
nlewycky, line 111: > All function calls done in a function that uses constrained floating point intrinsics must have the `strictfp` attribute.
> All function definitions that use constrained floating point intrinsics must have the `strictfp` attribute.
-- https://llvm.org/docs/LangRef.html#constrained-floating-point-intrinsics

Please add the `strictfp` attribute and place it in all places where it's required, which includes the trampolines.

nlewycky (commented):
I don't understand the changes to llvm at all. The constrained FP functions don't appear to have any documented difference in how they handle NaNs than the non-constainted operations. They only let you specify rounding mode and floating-point-exception handling. Also, this removed all the work done to track the state of our FP values to know whether they may contain NaNs and therefore don't need to be canonicalized, instead always inserting unnecessary canonicalization operations. What am I not seeing?

lib/c-api/build.rs
```diff
@@ -476,6 +476,7 @@ fn exclude_items_from_wasm_c_api(builder: Builder) -> Builder {
         .exclude_item("wasi_version_t")
         .exclude_item("wasm_config_push_middleware")
         .exclude_item("wasm_config_set_compiler")
+        .exclude_item("wasm_config_enable_nan_canonicalization")
```
Hywan, line 479:
```suggestion
        .exclude_item("wasm_config_canonicalize_nans")
```

lib/compiler-cranelift/src/config.rs
```diff
@@ -197,6 +188,14 @@ impl CompilerConfig for Cranelift {
         self.enable_verifier = true;
     }
 
+    fn enable_nan_canonicalization(&mut self) {
```
Hywan, line 191: Why not an API like `enable_nan_canonicalization(&mut self, enable: bool)`, just like we have for `enable_simd` and probably others?

jubianchi: my main concern here is consistency.

On `CompilerConfig` all the other things are enabled through `enable_<something>(&mut self)` functions.

If I decide to add a `enable: bool` parameter to `enable_nan_canonicalization` then I should add the same to all the other `enable_<something>` which is (strictly speaking) a breaking change to a public API that is most likely used a lot.

This is why I introduce a new way to switch this feature on or off. But I agree, the name can be better, something like `toggle_nan_canonicalization(&mut self, enable: bool)` would be better. and then, we can add `enable_<something>(&mut self, enable: bool)`.

WDYT?

tests/compilers/wast.rs
```diff
@@ -45,7 +45,14 @@ pub fn run_wast(wast_path: &str, compiler: &str) -> anyhow::Result<()> {
...
     );
-    let try_nan_canonicalization = wast_path.contains("nan-canonicalization");
+
+    cfg_if::cfg_if! {
+        if #[cfg(feature = "test-llvm")] {
+            let try_nan_canonicalization = wast_path.contains("nan-canonicalization") || wast_path.contains("f32") || wast_path.contains("f64") || wast_path.contains("float");
+        } else {
+            let try_nan_canonicalization = wast_path.contains("nan-canonicalization");
+        }
+    }
```
syrusakbary, line 55: I think all compilers should behave the same way if a feature X is enabled or disabled.
So I don't think is a good idea to change the features depending on the compiler used (other than support or not support something)

lib/compiler-llvm/Cargo.toml
```diff
@@ -25,7 +25,8 @@ rayon = "1.5"
 loupe = "0.1"
 
 [dependencies.inkwell]
-version = "=0.1.0-beta.2"
+git = "https://github.com/wasmerio/inkwell"
+branch = "add-metadata-support"
```
syrusakbary, line 29: Let's make sure a version is published to creates.io (we can also create `wasmer-inkwell`) before merging

lib/compiler-cranelift/src/config.rs
Hywan: Let's keep things as they are for the moment.

syrusakbary:
bors try

lib/compiler-llvm/src/abi/aarch64_systemv.rs
```diff
@@ -83,34 +83,51 @@ impl Abi for Aarch64SystemV {
 
         Ok(match sig.results() {
             [] => (
-                intrinsics
-                    .void_ty
-                    .fn_type(&param_types.collect::<Result<Vec<_>, _>>()?, false),
+                intrinsics.void_ty.fn_type(
+                    param_types
+                        .map(|v| v.unwrap().into())
```
MarkMcCaskey, line 88: The semantics here have changed from returning an error value if any of these things fail to just panicking. You can probably preserve the collect to `Result<Vec<_>, _>` behavior if you want with something like

```param_types.map(|v| v.and_then(|v| Ok(v.into()))).collect::<Result<Vec<BasicMetadataTypeEnum>, _>()?```

File: lib/compiler-llvm/src/translator/code.rs [hunk 17]
```diff
@@ -2228,7 +2225,15 @@ impl<'ctx, 'a> LLVMFunctionCodeGenerator<'ctx, 'a> {
                 }
                 */
 
-                let call_site = self.builder.build_call(func, &params, "");
+                let call_site = self.builder.build_call(
+                    func,
+                    params
+                        .iter()
+                        .map(|v| (*v).into())
```
MarkMcCaskey, line 2232: ⛳ You can also express this as:

```Rust
params.iter().copied().map(Into::into).collect::<Vec<BasicMetadataValueEnum>().as_slice()
```

if you prefer

MarkMcCaskey: You  can also make `From` impls that accept `&T` as well so you don't have to worry about it being `T` or `&T`, then you can remove the `copied` call here

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -4388,7 +4377,7 @@ impl<'ctx, 'a> LLVMFunctionCodeGenerator<'ctx, 'a> {
                 let input = self.state.pop1()?;
                 let res = self
                     .builder
-                    .build_call(self.intrinsics.sqrt_f32, &[input], "")
+                    .build_call(self.intrinsics.sqrt_f32, &[input.into()], "")
```
MarkMcCaskey, line 4380: Does it make sense to change the return value of `pop1` etc.? Do we ever use the non-`into`d version?

lib/compiler-llvm/src/translator/intrinsics.rs
```diff
@@ -320,94 +320,82 @@ impl<'ctx> Intrinsics<'ctx> {
...
-        let i1_ty_basic = i1_ty.as_basic_type_enum();
-        let i32_ty_basic = i32_ty.as_basic_type_enum();
-        let i64_ty_basic = i64_ty.as_basic_type_enum();
-        let f32_ty_basic = f32_ty.as_basic_type_enum();
-        let f64_ty_basic = f64_ty.as_basic_type_enum();
-        let i8x16_ty_basic = i8x16_ty.as_basic_type_enum();
-        let i16x8_ty_basic = i16x8_ty.as_basic_type_enum();
-        let f32x4_ty_basic = f32x4_ty.as_basic_type_enum();
-        let f64x2_ty_basic = f64x2_ty.as_basic_type_enum();
-        let i8_ptr_ty_basic = i8_ptr_ty.as_basic_type_enum();
```
MarkMcCaskey, line 295: We could also cache the most common ones here like we did before

lib/compiler-llvm/src/translator/code.rs
jubianchi: Actually, we can't. Doing this will just make us move the `into()` calls to other places.

For now, I think we can keep everything as is.

Hywan (approved):
Good work!

PR #2162, opened 2021-03-04
Update dependencies. Use Clap 3.0.0-beta2 instead of StructOpt

syrusakbary:
# Description

Update dependencies. Use Clap 3.0.0-beta2 instead of StructOpt


syrusakbary:
bors try

syrusakbary:
bors try

syrusakbary:
bors try-

syrusakbary:
bors try

lib/c-api/Cargo.toml
```diff
@@ -93,4 +93,4 @@ cranelift-backend = ["cranelift"]
 llvm-backend = ["llvm"]
 
 [build-dependencies]
-cbindgen = "0.15"
+cbindgen = "0.18"
```
Hywan, line 96: Please don't update `cbindgen` along with other deps. It's a big change that we must monitor carefully :-).

Hywan (commented):
Why migrating from StructOpt to Clap? Does it reduce the size of the `wasmer` binary?

syrusakbary:
> Why migrating from StructOpt to Clap? Does it reduce the size of the wasmer binary?

It seems is not being maintained enough, and we can have this functionality in clap itself.

lib/c-api/Cargo.toml
Hywan: Nevervind, I'm reviewing the changes now.

Hywan (approved):

syrusakbary:
Merging manually as test have already passed

lib/c-api/Cargo.toml
MarkMcCaskey: Following up on this, because we  ended up reverting back to structopt having things unrelated to this PR in the diff does make it a bit harder to read

PR #2163, opened 2021-03-04
fix: C API location and links are now correct

jubianchi:


Hywan (approved):
bors r+

Makefile
```diff
@@ -642,10 +642,9 @@ endif
 endif
 
 package-docs: build-docs build-docs-capi
-	mkdir -p "package/docs"
 	mkdir -p "package/docs/c/runtime-c-api"
 	cp -R target/doc package/docs/crates
-	cp -R lib/c-api/doc/deprecated/html/ package/docs/c/runtime-c-api
+	cp -R lib/c-api/doc/deprecated/html/* package/docs/c
```
Hywan, line 647: Could be great to copy in the `package/docs/c/deprecated/` directory. It will make it clear that it's about the deprecated API.

Can we also copy the Rustdoc for the new C API?

syrusakbary (approved):

Makefile
jubianchi: I think we should keep the actual path and use it to:
* describe the deprecated API and make it clear it's deprecated
* describe the new C API with some little introduction/examples and link from that to docs.wasmer.io and the rustdoc one

Makefile
```diff
@@ -642,10 +642,9 @@ endif
 endif
 
 package-docs: build-docs build-docs-capi
-	mkdir -p "package/docs"
 	mkdir -p "package/docs/c/runtime-c-api"
```
syrusakbary, line 645:
```suggestion
	mkdir -p "package/docs/c"
```

PR #2168, opened 2021-03-05
Added partial support for no-std environments

KernelFreeze:
# Description
Added dynamic imports and feature flags to allow support for no-std environments in `wasmer-compiler` and `singlepass-compiler` as a first step for supporting no-std in Wasmer.

# Review


Hywan:
Thanks for your patch, I'll review it soon :-).

lib/compiler-singlepass/src/codegen_x64.rs
```diff
@@ -1724,7 +1729,7 @@ impl<'a> FuncGen<'a> {
 
     pub fn get_state_diff(&mut self) -> usize {
         if !self.machine.track_state {
-            return std::usize::MAX;
+            return crate::lib::std::usize::MAX;
```
MarkMcCaskey, line 1732: I'm curious, can we just do `use crate::lib::std` at the top of the file and have this work? My understanding is that explicit imports take precedence over the prelude. That said, I don't actually know if `std` is considered part of the prelude or its own crate.

MarkMcCaskey (commented):
I skimmed through the changes and they seem reasonable -- thanks for the PR!

For some background, we haven't been working toward keeping Wasmer no-std supporting because we don't have a clear use case for it yet and it adds some overhead to the code we maintain. That said, if there are use cases for no-std Wasmer then we'd like to support it. And at least so far, the changes in this PR don't seem too burdensome.

On a related note to that, I think we'll need a test of some kind in CI or we'll likely break no-std support the day after this PR lands. A test could be a good opportunity to show-case some of the uses of no-std support, though that may be harder than it initially seems. I suspect we may run into issue with some of the OS-specific stuff in `wasmer-vm`, for example, when trying to use Wasmer with no-std. To not block this PR on something as (potentially) hard as getting no-std wasmer working in a test, perhaps we can just test that these crates compile with the `no-std` feature enabled (`cargo check`) or that the unit tests of these crates pass with `no-std`. 

I'd like to hear your thoughts on this!

KernelFreeze:
> For some background, we haven't been working toward keeping Wasmer no-std supporting because we don't have a clear use case for it yet and it adds some overhead to the code we maintain. That said, if there are use cases for no-std Wasmer then we'd like to support it. And at least so far, the changes in this PR don't seem too burdensome.

Hey, thank you for your quick reply! I have a use case that can be quite interesting, running Wasmer in a Rust Kernel, creating an environment to run WASM in bare-metal hardware. See [etheryal-kernel](https://github.com/etheryal/etheryal-kernel) for more details.

> On a related note to that, I think we'll need a test of some kind in CI or we'll likely break no-std support the day after this PR lands. A test could be a good opportunity to show-case some of the uses of no-std support, though that may be harder than it initially seems. I suspect we may run into issues with some of the OS-specific stuff in `wasmer-vm`, for example, when trying to use Wasmer with no-std.

I can try to add an integration test of some sort using no-std, if not possible, we could just make it compile with `core` feature flag, so it disables any std integration.

lib/compiler-singlepass/src/codegen_x64.rs
KernelFreeze: That works and looks way better!

syrusakbary:
> I can try to add an integration test of some sort using no-std, if not possible, we could just make it compile with core feature flag, so it disables any std integration.

Sounds good!

syrusakbary:
@KernelFreeze the progress is looking great. Let us know once the PR is ready to review :)

KernelFreeze:
Sure! Thank you :)

KernelFreeze:
Hey! I don't have time at the moment, so I will reopen another pull request when I'm ready. Thank you!

PR #2171, opened 2021-03-05
Fix lint warnings

Hywan:
# Description

WIP

# Review

### DIFF
--- a/lib/derive/src/lib.rs
+++ b/lib/derive/src/lib.rs
@@ -159,43 +159,41 @@ fn derive_struct_fields(data: &DataStruct) -> (TokenStream, TokenStream) {
                                     self.#name.initialize(#name);
                             }
                         }
-                    } else {
-                        if let Some(identifier) = identifier {
-                            let mut access_expr = quote_spanned! {
-                                f.span() =>
-                                    instance.exports.get_with_generics::<#inner_type, _, _>(#identifier)
+                    } else if let Some(identifier) = identifier {
+                        let mut access_expr = quote_spanned! {
+                            f.span() =>
+                                instance.exports.get_with_generics::<#inner_type, _, _>(#identifier)
+                        };
+                        for alias in aliases {
+                            access_expr = quote_spanned! {
+                                f.span()=>
+                                    #access_expr .or_else(|_| instance.exports.get_with_generics::<#inner_type, _, _>(#alias))
                             };
-                            for alias in aliases {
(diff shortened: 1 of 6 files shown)


Hywan:
bors try

lib/vm/src/instance/ref.rs
```diff
@@ -156,7 +156,7 @@ impl Clone for InstanceRef {
         Self {
             strong: self.strong.clone(),
             instance_layout: self.instance_layout,
-            instance: self.instance.clone(),
+            instance: self.instance,
```
Hywan, line 159: `self.instance` is of type `NonNull<T>`. It implements `Copy`, it's then useless to clone it.

syrusakbary (approved):

Hywan:
Let's merge it for the moment. I'll update one crate after another in the coming days.

bors r+

lib/derive/src/lib.rs
```diff
@@ -159,43 +159,41 @@ fn derive_struct_fields(data: &DataStruct) -> (TokenStream, TokenStream) {
                                     self.#name.initialize(#name);
                             }
                         }
-                    } else {
-                        if let Some(identifier) = identifier {
-                            let mut access_expr = quote_spanned! {
-                                f.span() =>
-                                    instance.exports.get_with_generics::<#inner_type, _, _>(#identifier)
+                    } else if let Some(identifier) = identifier {
```
MarkMcCaskey, line 162: I don't think this transformation is always a good idea. It's true that they're semantically equivalent at runtime, this transformation is a bit more explicit about whats happening in some ways, and sometimes it does improve code clarity, but sometimes an

```
else { 
  if ... { }
}
```

is forward-thinking in that it indicates that the `else` block is a distinct case that should be handleded as 1 unit. Currently there's only 1 handler in there so it can be simplified to an `else if` but the inverse transformation may be required in the future to make the code clean.

Perhaps that's over-thinking it, but I'm not sure I buy that this transformation is a good idea in the abstract.

-----

Specific to this PR: I think this change is fine, thanks for cleaning up some of the formatting too -- it's too bad there's no way to opt-in to rustfmt formatting macros. Historical note: my primary editor at the time kept choking on this crate (probably due to some combination of the LSP client I was using and the version of rust analyzer) and I had to use a different editor to write large sections of it and the second editor had its own ideas about formatting which is probably why it's so messy here.

MarkMcCaskey (approved):

Issue #2180, opened 2021-03-11
Loading NativeArtifact can be faster


ailisp:
Thanks for proposing a new feature!

### Motivation

Currently loading a NativeArtifact is two step:
1. load ELF, internally it's a mmap, very fast
2. deserialize ModuleMetadata from part of ELF data, slow

A simple print instant::now/elapsed() show the total loading takes 90% (3.5ms of 4ms) of the entire NativeArtifact::deserialization_from_file_unchecked, and longer when wasm file is bigger. So this is suboptimal

### Proposed solution

ModuleMetadata can also be replaced with a mem mapped data structure, as it's consist of struct or struct of vectors (primarymap). For example, I tried this lib: https://github.com/heremaps/flatdata/tree/master/flatdata-rs and loads a artificially constructed, similar structured and similar sized to wasmer Artifact, it takes only 0.2ms to load, and doesn't take longer when I enlarge size of vector in struct to 100x.

### Alternatives

Other mmap based crate maybe considerered

### Additional context

There is some HashMap and IndexMap in ModuleInfo, however they are all less than 1000 elements in a 2M contract (lib/tests/assets/qjs.wasm), which should be ok by replace with a sorted vector and lookup via binary search is faster than hashmap in this number of elements

Hywan:
Thank you for the proposal. That's an interesting thing to try.

For the moment, we use [`libloading`](https://crates.io/crates/libloading) to load the library. Behind the scene, [it calls `dlopen` to open the library, with the flags `RTLD_LAZY | RTLD_LOCAL`](https://docs.rs/libloading/0.7.0/src/libloading/os/unix/mod.rs.html#174-177). Maybe we can do better. I know `RTLD_NOLOAD` and `RTLD_NODELETE` can improve things quite a lot but not sure it will impact our generated shared library objects as they don't have dependencies. Need to test!

Are you willing to try something with a PR :-)?

syrusakbary:
@Hywan I believe @ailisp is mentioning not dlopen loading time issues, but bincode deserialization of `ModuleMetadata` when the artifact is partially loaded (already dlopen-ed).

@ailisp, I think your suggestion is really great. Once the PR #2183 is merged it should be really trivial to add a much faster deserialization mechanism. I'm happy with using `flatdata-rs` that you commented. But I'm also curious how it will perform also with [borsh-rs](https://github.com/near/borsh-rs).

ailisp:
[...]

PR #2183, opened 2021-03-12
Passive elements and data are now PrimaryMaps

syrusakbary:
# Description

This PR moves some fields in the `ModuleInfo` to stop using `HashMap` and start using `PrimaryMap` as the contents are not sparse but fully dense.
The PR is incentivized by #2180 as it's partially needed to stop using HashMap when possible. It doesn't completely solve it, as the `function_names` field is still a `HashMap`, but it should be trivial to solve that in a PR that improves the deserialization time.

# Review

### DIFF
--- a/lib/vm/src/instance/mod.rs
+++ b/lib/vm/src/instance/mod.rs
@@ -31,7 +31,6 @@ use memoffset::offset_of;
 use more_asserts::assert_lt;
 use std::any::Any;
 use std::cell::{Cell, RefCell};
-use std::collections::HashMap;
 use std::convert::{TryFrom, TryInto};
 use std::ffi;
 use std::fmt;
--- a/lib/vm/src/instance/mod.rs
+++ b/lib/vm/src/instance/mod.rs
@@ -82,11 +82,11 @@ pub(crate) struct Instance {
     /// Passive elements in this instantiation. As `elem.drop`s happen, these
     /// entries get removed. A missing entry is considered equivalent to an
     /// empty slice.
-    passive_elements: RefCell<HashMap<ElemIndex, Box<[VMCallerCheckedAnyfunc]>>>,
+    passive_elements: RefCell<PrimaryMap<ElemIndex, Box<[VMCallerCheckedAnyfunc]>>>,
 
     /// Passive data segments from our module. As `data.drop`s happen, entries
     /// get removed. A missing entry is considered equivalent to an empty slice.
-    passive_data: RefCell<HashMap<DataIndex, Arc<[u8]>>>,
+    passive_data: RefCell<PrimaryMap<DataIndex, Arc<[u8]>>>,
 
     /// Hosts can store arbitrary per-instance information here.
     host_state: Box<dyn Any>,
@@ -600,7 +600,7 @@ impl Instance {
         let table = self.get_table(table_index);
         let passive_elements = self.passive_elements.borrow();
         let elem = passive_elements
-            .get(&elem_index)
+            .get(elem_index)
             .map_or_else(|| -> &[VMCallerCheckedAnyfunc] { &[] }, |e| &**e);
 
         if src
@@ -625,7 +625,7 @@ impl Instance {
         // https://webassembly.github.io/reference-types/core/exec/instructions.html#exec-elem-drop
 
         let mut passive_elements = self.passive_elements.borrow_mut();
-        passive_elements.remove(&elem_index);
+        passive_elements[elem_index] = Box::new([]);
         // Note that we don't check that we actually removed an element because
         // dropping a non-passive element is a no-op (not a trap).
     }
@@ -718,9 +718,7 @@ impl Instance {
 
         let memory = self.get_memory(memory_index);
         let passive_data = self.passive_data.borrow();
-        let data = passive_data
-            .get(&data_index)
-            .map_or(&[][..], |data| &**data);
+        let data = passive_data.get(data_index).map_or(&[][..], |data| &**data);
 
         if src
             .checked_add(len)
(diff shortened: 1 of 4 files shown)


lib/vm/src/instance/mod.rs
```diff
@@ -746,7 +744,7 @@ impl Instance {
     /// Drop the given data segment, truncating its length to zero.
     pub(crate) fn data_drop(&self, data_index: DataIndex) {
         let mut passive_data = self.passive_data.borrow_mut();
-        passive_data.remove(&data_index);
+        passive_data[data_index] = Arc::new([]);
```
MarkMcCaskey, line 746: same here

MarkMcCaskey (commented):
First of all thanks for the description of what this PR is doing / what it's for.

I'm not sure if I believe that the data here is not sparse. It doesn't start sparse but with `elem_drop` it looks like it could be become sparse.

Given the diff I've just seen, it looks like the change is correct, but I'm a bit concerned about the change in semantics of this type. There could easily be code not included in this diff which doesn't do the `is_empty` check and is broken by this change. I think using `Option` instead of a sentinel value will be less error prone and more explicit. There's no runtime cost to using `Option<Box<T>>` instead of `Box<T>`, but the users _will_ now have to deal with a potentially `None` value, which they were theoretically doing before.

lib/vm/src/instance/mod.rs
```diff
@@ -625,7 +625,7 @@ impl Instance {
         // https://webassembly.github.io/reference-types/core/exec/instructions.html#exec-elem-drop
 
         let mut passive_elements = self.passive_elements.borrow_mut();
-        passive_elements.remove(&elem_index);
+        passive_elements[elem_index] = Box::new([]);
```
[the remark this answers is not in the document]
syrusakbary: Yeah I was thinking about that. I think perhaps a better structure is `MaybeUninit` (instead of Box), as it will probably require no changes on the access from the generated compiled code from the compilers.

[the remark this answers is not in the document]
MarkMcCaskey: `Option<Box<T>>` is FFI safe! It's in fact 100% identical to `Box<T>`! We use it in the C API all the time.

That said, I actually don't know if `Box<[T]>` in general has a defined layout but 🤷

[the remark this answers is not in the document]
MarkMcCaskey: (The reason is that `Box<T>` can never be null. `Option<Box<T>>` is just how you say "nullable (heap) pointer" in Rust. Rust optimizes that and `None` is equal to nullptr and Some(ptr) is normal `Box<T>` internally)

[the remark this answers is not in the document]
syrusakbary: Awesome!

lib/vm/src/instance/mod.rs
```diff
@@ -599,9 +597,16 @@ impl Instance {
...
+        let elem = passive_elements.get(elem_index).map_or_else(
+            || -> &[VMCallerCheckedAnyfunc] { &[] },
+            |e| {
+                if let Some(e) = e {
+                    &**e
+                } else {
+                    &[]
+                }
+            },
+        );
```
MarkMcCaskey, line 609: I wrote a long comment walking through how I simplified this and tried a few different things but the comment got deleted by Firefox / GitHub so,

```Rust
        let elem = passive_elements.get(elem_index).map_or(&[], |e| e.map_or(&[], &**e));
```

Should be fine. There's also a way where we combine the 2 options as one and only handle the failure case once, but I think it's a little less readable in this context.

`map_or` is fine because the alternative is cheap: `map_or_else` is only needed when we have an expensive to compute alternative.

MarkMcCaskey: Actually I'll include it:
```Rust
        let elem = passive_elements.get(elem_index).and_then(|e| e.as_ref().map(|e| &**e)).unwrap_or(&[]);
```

In case you're interested. It may be able to be simplified more with `as_deref`

maybe

```Rust
        let elem = passive_elements.get(elem_index).and_then(|e| e.as_ref().as_deref()).unwrap_or(&[]);
```

lib/vm/src/instance/mod.rs
```diff
@@ -718,9 +723,13 @@ impl Instance {
...
-        let data = passive_data
-            .get(&data_index)
-            .map_or(&[][..], |data| &**data);
+        let data = passive_data.get(data_index).map_or(&[][..], |data| {
+            if let Some(data) = data {
+                &**data
+            } else {
+                &[][..]
+            }
+        });
```
MarkMcCaskey, line 732: This can be cleaned up in the same ways as above.

lib/vm/src/instance/mod.rs
```diff
@@ -823,7 +832,13 @@ impl InstanceHandle {
             .map(|m| m.vmglobal())
             .collect::<PrimaryMap<LocalGlobalIndex, _>>()
             .into_boxed_slice();
-        let passive_data = RefCell::new(module.passive_data.clone());
+        let passive_data = RefCell::new(
+            module
+                .passive_data
+                .values()
+                .map(|data| Some(data.clone()))
```
MarkMcCaskey, line 839: Consider using `.cloned()` instead, it does exactly this

MarkMcCaskey (approved):
Looks good! the `get` logic can be cleaned up a bit but otherwise 👍

lib/vm/src/instance/mod.rs
syrusakbary: We are going from a non `Option` to an `Option` so we need to wrap in a `Some`?

MarkMcCaskey: Oh I see, good call!

You can do it in two lines as `.values().cloned().map(Some)`, which is a bit more explicit but what you have is fine!

lib/vm/src/instance/mod.rs
syrusakbary: I went to the second suggestion since it only has one `&[]`! Thanks!

PR #2190, opened 2021-03-17
Rkyv read native artifact from archive

ailisp:
# Description

As mentioned in #2180, load a native artifact is slow because of deserialization ModuleMetadata. This draft PR show how fast it can be when serialize it archive to a [u8] that can be directly read back and interprete as nested struct/HashMap/PrimaryMap without a deserialization. Run
```
cargo bench -p wasmer-cache --features singlepass
```
from root dir, will see a lot of timeing in the end:
```
...
3.447994ms // bincode::deserialize
rkyv 22ns
rkyv deser 386.539µs
```

This PR use a forked rkyv, for impl rkyv traits on PrimaryMap and IndexMap. 
TODO:
- [x] in fork rkyv, impl rkyv traits on IndexMap is not guanranteed keep indexmap's order, just a quick hack to see timing (reuse ArchivedHashMap). need to impl a real ArchivedIndexMap
- ~[ ] replace each use of `ModuleMetadata` and children to `ArchivedModuleMetadata` or something like:~
```
enum ModuleMetadataWrap {
    Original(ModuleMetadata),
    Archived(ArchivedModuleMetadata)
}
```
~and field access to `ModuleMetadataWrap` method calls~
- [x] rkyv deserialize is not as fast as rkyv archive, but fast enough and advantage is minimum change to wasmer's struct field access code, so I plan to use rkyv deserialize
- [x] add a test to ensure rkyv does deserialize correctly. I added assert to see deserialized obj is Eq to original, need to turn it to a unit test.
- [x] move impl rkyv::* for PrimaryMap/IndexMap inside this PR, instead of using a fork of rkyv, using a remote-derive technique

# Review


ailisp:
I'm in question about whether we should prefer rkyv deserialize over archive. The deser approach require minimum change (just the change in this PR, PR is almost done except for moving impl rkyv in PrimaryMap/IndexMap inside wasmer). The later is fast, but more importantly, just found its speed is **consistent**. `386.539µs deser` is a good enough time, but test multiple times it become volatile, and worst case takes more than 1ms, which is bad. While archive approach is always <30ns.
Archive need to replace `ModuleMetadata` with 
```
enum ModuleMetadataWrap {
    Original(ModuleMetadata),
    Archived(ArchivedModuleMetadata)
}
```
And all access of ModuleMetadata as method calls. When access ModuleMetadata's field of PrimaryMap/IndexMap type, also need implement entire ArchivedPrimaryMap/ArchivedIndexMap  methods that mirror all used methods of PrimaryMap/IndexMap.

syrusakbary:
@ailisp we just merged #2195, which moves `PrimaryMap` and `SecondaryMap` into Wasmer codebase. That way we can implement our own serializers and customize more our needs.
If you adapt this PR to latest `master` probably it would not require a fork of `rkyv` (but just the crates.io one).

Also, I think is critical that serialization/deserialization remains constant regardless of the platform/architecture.

That means that serializing the same content from macOS (with the m1 / arm64 chip) should give the same result as doing it from Linux (with a typical x86_64 chip). This is critical to allow cross-compilation from other devices. That means compiling the file in one system, and running it in another.
Serde/bincode assures this. And I'm sure that rkyv does as well... but I'm not that sure about archive. Could you give some light on this?

Note: if we can assure constant serialization from different architectures with the archive approach, I'm happy to choose that one!

Thanks!

ailisp:
>  we just merged #2195, which moves PrimaryMap and SecondaryMap into Wasmer codebase

That's great to hear! would you also plan to move IndexMap in it? It's okay if not, I can Implement traits on some WasmerIndexMapWrapper trait.

> Also, I think is critical that serialization/deserialization remains constant regardless of the platform/architecture.
That means that serializing the same content from macOS (with the m1 / arm64 chip) should give the same result as doing it from Linux (with a typical x86_64 chip) - and it should be the same for 32 bit architectures. This is critical to allow cross-compilation from other devices. That means compiling the file in one system, and running it in another.


Good point! I'll do some test. author mentioned major caveat is it is using platform native endian and didn't fully test for cross platform. Other than it theoratically work. Archive is internally encode all pointers to a relative offset as u64, that indicate where object is located in the resulting [u8]. I consider little endian is common enough (mac/linux/windows x86 and ARM) and author also gives an solution to handle cross platform endian: https://davidkoloski.me/rkyv/faq.html#how-does-rkyv-handle-endianness

Also I suppose the constant serialization means "the deserialized back object is Eq" when serialize from any platform and deserialize in any platform, instead of "serialized bytes is same byte by byte". Because in test I saw bincode doesn't give same byte when serialize the same ModuleMetadata twice.

syrusakbary:
Just thought of this.
It might be useful to research the usage of flex buffers (schema less iteration of flat buffers): https://crates.io/crates/flexbuffers

ailisp:
@syrusakbary I discovered you have some commented lines of using flexbuffers in wasmer, how is that going?
I also discovered this lib but it doesn't support fast access hashmap and indexmap, and you have to deserialize to access, but deserialize it won't be very fast because the overhead bring by serde (we benched any serde based deserialization is slower than borsh, and rkyv deserialize is faster than borsh)

lib/vm/src/module.rs
```diff
@@ -115,6 +118,142 @@ pub struct ModuleInfo {
     pub num_imported_globals: usize,
 }
 
+/// Mirror version of ModuleInfo that can derive rkyv traits
+#[derive(RkyvSerialize, RkyvDeserialize, Archive)]
+pub struct RkyvModuleInfo {
```
syrusakbary, line 123: `ArchivedModuleInfo` seems like a better name?

lib/wasmer-types/src/archives.rs
```diff
@@ -0,0 +1,120 @@
...
+        Ok(PrimaryMap {
+            elems,
+            unused: PhantomData,
+        })
+    }
+}
+
+#[derive(Serialize, Deserialize, Archive)]
+/// Rkyv Archivable IndexMap
+pub struct RkyvIndexMap<K: Hash + Eq + Archive, V: Archive> {
```
syrusakbary, line 94: `ArchivedIndexMap` seems like a better name

lib/engine-native/src/artifact.rs
```diff
@@ -545,13 +555,25 @@ impl NativeArtifact {
...
         let metadata_slice: &'static [u8] =
             slice::from_raw_parts(&size[10] as *const u8, metadata_len as usize);
-        let metadata: ModuleMetadata = bincode::deserialize(metadata_slice)
+
+        let mut pos: [u8; 8] = Default::default();
+        pos.copy_from_slice(&metadata_slice[0..8]);
+        let pos: u64 = u64::from_le_bytes(pos);
+        let archived = archived_value::<ModuleMetadata>(&metadata_slice[8..], pos as usize );
+        let mut deserializer = SharedDeserializerAdapter::new(AllocDeserializer);
+        let metadata = RkyvDeserialize::deserialize(archived, &mut deserializer)
```
syrusakbary, line 564: Can we move this into `ModuleMetadata::deserialize_from_archive`? (it will make easier to plug-in other ways of deserialization)

lib/vm/Cargo.toml
```diff
@@ -21,6 +21,7 @@ more-asserts = "0.2"
 cfg-if = "0.1"
 backtrace = "0.3"
 serde = { version = "1.0", features = ["derive", "rc"] }
+rkyv = "0.4.1"
```
syrusakbary, line 24: We should make this an optional dependency

lib/engine-native/Cargo.toml
```diff
@@ -24,6 +24,11 @@ leb128 = "0.2"
 libloading = "0.7"
 tempfile = "3.1"
 which = "4.0"
+rkyv = "0.4.1"
+
+[dev-dependencies]
+wasmer-compiler-singlepass = { version = "1.0.2", path = "../compiler-singlepass" }
+#wasmer = { version = "1.0.2", path = "../api" }
```
syrusakbary, line 31: Remove me

lib/engine-native/src/artifact.rs
```diff
@@ -545,13 +555,25 @@ impl NativeArtifact {
         })?;
         let metadata_slice: &'static [u8] =
             slice::from_raw_parts(&size[10] as *const u8, metadata_len as usize);
-        let metadata: ModuleMetadata = bincode::deserialize(metadata_slice)
+
```
syrusakbary, line 558: Since endianness can vary between platforms, can we add an extra check on the `triple.endianness()` assuring that the target triple endianness is the same as host endianness, and returning an error otherwise?

This will shield us from `rkyv` serialization issues in the future when the data is encoded in one endianness in the host but expected as other endianness in when deserialized in the target.

lib/vm/src/module.rs
ailisp: `derive(Archive)` will generate a `ArchivedModuleInfo` so i can't use that name. Also this mean to be a before-archive ModuleInfo-equivalant struct, maybe "RkyvArchivableModuleInfo" or "ModuleInfoForArchive"?

lib/wasmer-types/src/archives.rs
ailisp: Same as above, which name do you prefer?

lib/engine-native/src/artifact.rs
ailisp: Good point!

lib/vm/Cargo.toml
ailisp: Sounds good, will make serde as default, right now it replaced serde

lib/vm/src/module.rs
syrusakbary: But the `ModuleInfo` struct is not deriving from `Archive`, hence `ArchivedModuleInfo ` should be available.
So I guess it will be fine?

ailisp: Ah yes, it should be available. however this shouldn't be named so, because it hasn't been archived.
ArchivableModuleInfo
RkyvArchivableModuleInfo
ModuleInfoForArchive
or ModuleInfoBeforeArchive?

Note derive(archive) would prefix Archived to this name, and we're likely need to refer this ArchivedXXX name, therefore (Archived)ArchivableModuleInfo maybe the best?

lib/vm/Cargo.toml
ailisp: Was also consider make it optional in engine-native and user can select to compile, but in that case it has to expose a "use-bincode"/"use-rkyv" feature in lib/api, which seems make wasmer configuration over complicated there, so i only make this optional in lib/vm

lib/vm/src/module.rs
syrusakbary: Let's go with `ArchivableModuleInfo` 👍

lib/wasmer-types/src/archives.rs
syrusakbary: Let's go with `ArchivableIndexMap`

lib/engine-native/Cargo.toml
```diff
@@ -24,6 +24,10 @@ leb128 = "0.2"
 libloading = "0.7"
 tempfile = "3.1"
 which = "4.0"
+rkyv = "0.4.1"
+
+[dev-dependencies]
+wasmer-compiler-singlepass = { version = "1.0.2", path = "../compiler-singlepass" }
```
syrusakbary, line 30:
```suggestion
rkyv = "0.4.1"
```

lib/api/src/store.rs
```diff
@@ -126,3 +126,30 @@ pub trait StoreObject {
...
+        let binary = std::fs::read("../../lib/c-api/tests/assets/qjs.wasm").unwrap();
+        let artifact = NativeArtifact::new(&engine, &binary, store.tunables()).unwrap();
+
+        let artifact_serialized = artifact.serialize().unwrap();
+        let artifact_deserialized =
+            unsafe { NativeArtifact::deserialize(&engine, &artifact_serialized) }.unwrap();
+
+        assert_eq!(artifact.metadata(), artifact_deserialized.metadata());
+    }
+}
```
syrusakbary, line 155: We are already testing serialization/deserialization in two places, so we should be good to remove this:
* examples https://github.com/wasmerio/wasmer/blob/master/examples/engine_headless.rs
* per compiler/engine variation: https://github.com/wasmerio/wasmer/blob/master/tests/compilers/serialize.rs

The latter will test in all the available compilers/engines. This makes tests more modular!

```suggestion
```

syrusakbary:
bors try

lib/compiler/Cargo.toml
```diff
@@ -21,16 +21,18 @@ serde = { version = "1.0", features = ["derive"], optional = true }
...
-default = ["std", "enable-serde"]
+default = ["std", "enable-serde", "enable-rkyv"]
 # This feature is for compiler implementors, it enables using `Compiler` and
 # `CompilerConfig`, as well as the included wasmparser.
 # Disable this feature if you just want a headless engine.
 translator = ["wasmparser"]
 std = ["wasmer-types/std"]
 core = ["hashbrown", "wasmer-types/core"]
 enable-serde = ["serde", "serde_bytes", "wasmer-types/enable-serde"]
+enable-rkyv = ["rkyv"]
```
Hywan, line 35: Shouldn't they be exclusive? Either `enable-serde` or `enable-rkyv`?

lib/engine-native/src/artifact.rs
```diff
@@ -545,13 +546,19 @@ impl NativeArtifact {
...
         let mut engine_inner = engine.inner_mut();
 
         Self::from_parts(&mut engine_inner, metadata, shared_path, lib)
             .map_err(DeserializeError::Compiler)
     }
+
+    /// Used in test deserialize metadata is correct
+    pub fn metadata(&self) -> &ModuleMetadata {
+        &self.metadata
+    }
```
Hywan, line 561:
```suggestion
    /// Used in test deserialize metadata is correct
    #[cfg(test)]
    pub fn metadata(&self) -> &ModuleMetadata {
        &self.metadata
    }
```

If this function is only used for testing purposes, maybe it's better that way, thoughts?

lib/wasmer-types/src/indexes.rs
```diff
@@ -25,60 +36,149 @@ entity_impl!(LocalMemoryIndex);
...
 #[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)]
 #[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+#[cfg_attr(
+    feature = "enable-rkyv",
+    derive(RkyvSerialize, RkyvDeserialize, Archive)
+)]
+#[cfg_attr(
+    feature = "enable-rkyv",
+    archive(derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug))
+)]
```
Hywan, line 166: I'm wondering at some point whether we should not add this inside the `entity_impl!` macro directly 🤔.

Hywan (commented):
Excellent PR for the moment! Keep going!

lib/api/src/store.rs
ailisp: Sounds good, i actually tried to unit test ser/de of an artifact, but seems difficult as without compiler it's hard to construct an artifact. To test seraialize a module would also trigger serialize the artifact

lib/compiler/Cargo.toml
ailisp: It is not, because structs defined in lib/compiler are shared in all artifacts, but only native artifact switched to use rkyv to deserialize

lib/engine-native/src/artifact.rs
ailisp: Tried, however rust seems not happy to compile, even though it's only used in test:
```
error[E0599]: no method named `metadata` found for struct `NativeArtifact` in the current scope
   --> lib/api/src/store.rs:153:29
    |
153 |         assert_eq!(artifact.metadata(), artifact_deserialized.metadata());
    |                             ^^^^^^^^ private field, not a method

error[E0599]: no method named `metadata` found for struct `NativeArtifact` in the current scope
   --> lib/api/src/store.rs:153:63
    |
153 |         assert_eq!(artifact.metadata(), artifact_deserialized.metadata());
    |                                                               ^^^^^^^^ private field, not a method

error: aborting due to 2 previous errors
```
(lib/api/src/store.rs:153:29 is in `#[cfg(test)]` and `#[test]` 🤔 

I'll probably drop this, given will test the artifact's enclosing module == deserialized module, by deriving Eq in artifact, so doesn't need pub access artifact.metadata anymore

lib/wasmer-types/src/indexes.rs
ailisp: It makes sense to add so. entity_impl! are all on some struct(integer), and they can trivially derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)

ailisp: This was required because rkyv macro generate ArchivedCustomSectionIndex struct on CustomSectionIndex, then used ArchivedCustomSectionIndex and primary map key, but this key need Copy, ..., traits, so `archive(derive(Copy...))` is to let rkyv macro add a `derive(Copy,...)` on generated ArchivedCustomSectionIndex definition

ailisp:
It seems no tests is run on CI, so i run this test locally:
```
cargo test --test compilers serialize::test_serialize --features singlepass,test-cranelift,test-native
cargo test --test compilers serialize::test_deserialize --features singlepass,test-cranelift,test-native
```
And they both pass. Also attempt test-singlepass, but only serialize pass, and deserialize failed at execution same reason as master, so I assume that is unrelated to this PR.

MarkMcCaskey:
bors try

syrusakbary:
bors try-

syrusakbary:
bors try

syrusakbary:
@ailisp any insight on why rkyv causes a fail when deserializing? It's a bit red flag if this causes us not being able to upgrade rkyv to future versions

syrusakbary:
Tests seem to be failing:
```
$ cargo test -p wasmer-types --release
   Compiling wasmer-types v1.0.2 (/Users/runner/work/wasmer/wasmer/lib/types)
error[E0282]: type annotations needed
   --> lib/types/src/entity/packed_option.rs:171:37
    |
171 |         assert_eq!(some.expand(), x.into());
    |                                   --^^^^--
    |                                   | |
    |                                   | cannot infer type for type parameter `T` declared on the trait `Into`
    |                                   this method call resolves to `T`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0282`.
error: could not compile `wasmer-types`
```

ailisp:
> @ailisp any insight on why rkyv causes a fail when deserializing? It's a bit red flag if this causes us not being able to upgrade rkyv to future versions

Looking into it

ailisp:
> Tests seem to be failing:

I'm looking it now, so far it looks very weird as this is a all thing implemented in same source file unit test and in this particular file no line was changed, but it does failed only when `enable-rkyv`

ailisp:
rkyv 0.4.3 fail is also fixed in https://github.com/wasmerio/wasmer/pull/2190/commits/bb9945e521474f36e35a0a7b6ada3fd4b5dcac40 and https://github.com/wasmerio/wasmer/pull/2190/commits/2a3efeb1d28da9f3ea739ed917ac379919cbf156, root cause is 0.4.3 introduced an alignment check: https://github.com/djkoloski/rkyv/blob/b4e757d6e6fad5b8c8d4fcbec55b67cb5c637784/rkyv/src/lib.rs#L755

ailisp:
@syrusakbary ptal :)

syrusakbary:
Sorry for the delay!

The PR is looking great. I'd like to do a few things before merging:
* Add support for rkyv in the JIT engine
* Fix conflicts

ailisp:
@syrusakbary conflict is resolved. And sorry last commit actually hasn't fixed the debug_assert failure on rkyv 0.4.3, i accidentally run test with --release and that hides the failure. I fixed it in https://github.com/wasmerio/wasmer/pull/2190/commits/44339c18fe782c19ad25a2ae63bd11e0cb4fe011 and verified by:
```
cargo test --features cranelift,test-cranelift,test-native
```
It will be nice if the CI is back!

As for "Add support for rkyv in the JIT engine", it's going to be a big change too, would you like it in another PR?

syrusakbary:
> As for "Add support for rkyv in the JIT engine", it's going to be a big change too, would you like it in another PR?

Let's do it on other PR!

bors r+

syrusakbary (approved):

syrusakbary:
@ailisp it seems code linting is failing. Once is fixed we should be good to merge

ailisp:
@syrusakbary ah looks like there is a CI :D

syrusakbary:
It seems like cargo deny check is failing

lib/vm/src/module.rs
```diff
@@ -6,21 +6,28 @@
...
+use wasmer_types::ArchivableIndexMap;
 use wasmer_types::{
     CustomSectionIndex, DataIndex, ElemIndex, ExportIndex, ExportType, ExternType, FunctionIndex,
     FunctionType, GlobalIndex, GlobalInit, GlobalType, ImportIndex, ImportType, LocalFunctionIndex,
     LocalGlobalIndex, LocalMemoryIndex, LocalTableIndex, MemoryIndex, MemoryType, SignatureIndex,
     TableIndex, TableInitializer, TableType,
 };
 
-#[derive(Debug, Clone, MemoryUsage)]
+#[derive(Debug, Clone, MemoryUsage, RkyvSerialize, RkyvDeserialize, Archive)]
```
syrusakbary, line 30: We probably want this as a cfg flag:

```suggestion
#[derive(Debug, Clone, MemoryUsage)]
#[cfg_attr(
    feature = "enable-rkyv",
    derive(RkyvSerialize, RkyvDeserialize, Archive)
)]
```

syrusakbary:
@ailisp we need to merge this suggestion https://github.com/wasmerio/wasmer/pull/2190#discussion_r617149172

Otherwise tests will fail:
https://github.com/wasmerio/wasmer/runs/2396189543

ailisp:
Btw do you use some script / tool to run all tests locally? Sorry to have you start test back and forth

syrusakbary:
> Btw do you use some script / tool to run all tests locally? Sorry to have you start test back and forth

This are the main tests run:

```shell
make test
make test-capi
make test-integration
```

ailisp:
@syrusakbary I run cargo fmt --all on two computers give me different format result. So there are two reasons. One of it is I had a `~/.config/rustfmt/rustfmt.toml` on one computer, so i delete one, but it still differs. So the other reason is I had one rustc 1.49, anther 1.51. Do you consider have a fixed `rustfmt.toml` and `rust-toolchain` file like we did in nearcore?

https://github.com/near/nearcore/blob/master/rustfmt.toml
https://github.com/near/nearcore/blob/master/rust-toolchain

Or, it is helpful to ask contributors to always upgrade to latest stable rust (as used in ci), before run 
```
make lint
make test
make test-capi
make test-integration
```

ailisp:
now make lint, test and test-capi pass, make test-integration fail at a same reason as master, looks unrevelant to this PR:
```
failures:

---- object_file_engine_works stdout ----
Error: Failed to compile wasm with Wasmer

Caused by:
    No such file or directory (os error 2)
thread 'object_file_engine_works' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /home/bo/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/test/src/lib.rs:194:5
```

syrusakbary:
> Do you consider have a fixed rustfmt.toml and rust-toolchain file like we did in nearcore?

Let's open a PR for `rust-toolchain` (we don't need a custom rustfmt.toml)

syrusakbary:
I've looked and for some reasons tests are failing in Aarch64, but I believe is completely unrelated to this PR.

Merging manually

syrusakbary:
@ailisp if you can also open another PR integrating rkyv into the JIT engine that would be great

ailisp:
> @ailisp if you can also open another PR integrating rkyv into the JIT engine that would be great

Sounds good. We decide to use JIT engine now

PR #2195, opened 2021-03-18
feat(types) Remove dependency to `cranelift-entity`

Hywan:
# Description

Don't be afraid, everything is cool.

`cranelift-entity` is nice, but we are limited by it. We want to go
further for some features. Those features are very likely to be
refused by Cranelift because it's out of their scope.

This patch basically copies the code from `cranelift-entity` inside
`wasmer-types`. The code has been modified to fit in our own
“`no_std`” strategy. The code will be modified deeper in upcoming
patches and most of the entities are likely to radically change. The
goal of this patch is strictly to not break our code while porting the
new code.

# Review

- [ ] Do we need to update the `ATTRIBUTIONS.md` file?

### DIFF
--- a/lib/wasmer-types/src/entity/mod.rs
+++ b/lib/wasmer-types/src/entity/mod.rs
@@ -1,3 +1,6 @@
+// This file contains code from external sources.
+// Attributions: https://github.com/wasmerio/wasmer/blob/master/ATTRIBUTIONS.md
+
 /// A type wrapping a small integer index should implement `EntityRef` so it can be used as the key
 /// of an `SecondaryMap` or `SparseMap`.
 pub trait EntityRef: Copy + Eq {
--- a/lib/wasmer-types/src/entity/mod.rs
+++ b/lib/wasmer-types/src/entity/mod.rs
@@ -1,0 +1,92 @@
+/// A type wrapping a small integer index should implement `EntityRef` so it can be used as the key
+/// of an `SecondaryMap` or `SparseMap`.
+pub trait EntityRef: Copy + Eq {
+    /// Create a new entity reference from a small integer.
+    /// This should crash if the requested index is not representable.
+    fn new(_: usize) -> Self;
+
+    /// Get the index that was used to create this entity reference.
+    fn index(self) -> usize;
+}
+
+/// Macro which provides the common implementation of a 32-bit entity reference.
+#[macro_export]
+macro_rules! entity_impl {
+    // Basic traits.
+    ($entity:ident) => {
+        impl $crate::entity::EntityRef for $entity {
+            fn new(index: usize) -> Self {
+                debug_assert!(index < ($crate::lib::std::u32::MAX as usize));
+                $entity(index as u32)
+            }
+
+            fn index(self) -> usize {
+                self.0 as usize
+            }
+        }
+
+        impl $crate::entity::packed_option::ReservedValue for $entity {
+            fn reserved_value() -> $entity {
+                $entity($crate::lib::std::u32::MAX)
+            }
+
+            fn is_reserved_value(&self) -> bool {
+                self.0 == $crate::lib::std::u32::MAX
+            }
+        }
+
+        impl $entity {
+            /// Create a new instance from a `u32`.
+            #[allow(dead_code)]
+            pub fn from_u32(x: u32) -> Self {
+                debug_assert!(x < $crate::lib::std::u32::MAX);
+                $entity(x)
+            }
+
+            /// Return the underlying index value as a `u32`.
+            #[allow(dead_code)]
+            pub fn as_u32(self) -> u32 {
+                self.0
+            }
+        }
+    };
+
+    // Include basic `Display` impl using the given display prefix.
+    // Display a `Block` reference as "block12".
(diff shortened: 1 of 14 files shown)


Hywan:
bors try

Hywan:
bors try

lib/wasmer-types/src/entity/mod.rs
```diff
@@ -0,0 +1,92 @@
...
+                f: &mut $crate::lib::std::fmt::Formatter,
+            ) -> $crate::lib::std::fmt::Result {
+                write!(f, concat!($display_prefix, "{}"), self.0)
+            }
+        }
+
+        impl $crate::lib::std::fmt::Debug for $entity {
+            fn fmt(
+                &self,
+                f: &mut $crate::lib::std::fmt::Formatter,
```
MarkMcCaskey, line 71: Does this macro expand to the right thing in other crates?

lib/wasmer-types/src/entity/packed_option.rs
```diff
@@ -0,0 +1,171 @@
+//! Compact representation of `Option<T>` for types with a reserved value.
+//!
+//! Small Cranelift types like the 32-bit entity references are often used in tables and linked
```
MarkMcCaskey, line 3: Do we want to generalize the doc comments? We use these types for more than just interacting with Cranelift.

lib/wasmer-types/src/entity/packed_option.rs
```diff
@@ -0,0 +1,171 @@
...
+//! to represent `None`.
+
+use crate::lib::std::fmt;
+use crate::lib::std::mem;
+
+#[cfg(feature = "enable-serde")]
+use serde::{Deserialize, Serialize};
+
+/// Types that have a reserved value which can't be created any other way.
+pub trait ReservedValue {
```
MarkMcCaskey, line 17: Do we use any of the types in this module in Wasmer? If not, perhaps we can just not include them

lib/wasmer-types/src/entity/secondary_map.rs
```diff
@@ -0,0 +1,320 @@
...
+        }
+    }
+
+    /// Returns the number of elements the map can hold without reallocating.
+    pub fn capacity(&self) -> usize {
+        self.elems.capacity()
+    }
+
+    /// Get the element at `k` if it exists.
+    #[inline(always)]
```
MarkMcCaskey, line 86: A bit odd to see these on `SecondaryMap` and not on `PrimaryMap` but 🤷 . I find it easy to believe that these should always be inlined and that Rust probably also agrees and doesn't need to be told.

Though I think the `#[inline]` attributes _do_ matter for cross-crate inlining without LTO? With LTO they don't matter, but without LTO they do. At least that's what I remember, not super confident in that memory though.

Anyways, no action needed here. Just pointing it out.

MarkMcCaskey (approved):
Looks good to me.

👍 on the attributions. I'm in favor of being explicit about where the code came from and what the license is

lib/wasmer-types/src/entity/mod.rs
Hywan: Yes. The `lib` module is now public. `wasmer-compiler-llvm` uses this macro, and `cargo check` + tests are OK.

lib/wasmer-types/src/entity/packed_option.rs
Hywan: :+1:

lib/wasmer-types/src/entity/packed_option.rs
Hywan: Will check. I've already removed some types that we aren't using.

lib/wasmer-types/src/entity/secondary_map.rs
Hywan: Yes. That's part of the rewrite I want to do later. Let's keep the same code for the moment, and let's optimize later as planned :-). Thanks for pointing that out!

syrusakbary:
Note: on the attributions we should follow our current structure. Which is:
* No need to change `ATTRIBUTIONS.md` since the following is still true: `wasmtime: for their API and internal documentation, as well as some internal implementations`
* [ ] We need to update [`wasmer-types/README.md`](https://github.com/wasmerio/wasmer/blob/master/lib/wasmer-types/README.md) adding an Acknowledgments section with the cranelift-entity reference.
```
### Acknowledgments

This project borrowed some of the code for the entity structure from [cranelift-entity](https://crates.io/crates/cranelift-entity).
We decided to move it here to help on serialization/deserialization and also to ease the integration with other tools like loupe.

Please check [Wasmer ATTRIBUTIONS](https://github.com/wasmerio/wasmer/blob/master/ATTRIBUTIONS.md) to further see licenses and other attributions of the project. 
```

* [ ] We need the following content to the headers in the new added files:

```
// This file contains code from external sources.
// Attributions: https://github.com/wasmerio/wasmer/blob/master/ATTRIBUTIONS.md

```

lib/wasmer-types/src/entity/packed_option.rs
Hywan: I confirmed it's used.

Issue #2193, opened 2021-03-18
can not free the linear memory which is created by create_vm_memory


bxq2011hust:
### Describe the bug

I use wasmer as a engine to run wasm binary, but When I continue to execute, the memory will continue to grow.
At first I thought there was a bug in the deprecated C API, then I changed my code to use the new C API to run wasm, the memory still grow.

I use google profile to trace memory usage, the below is a result, it seems the linear memory is not freed, though I have called any delete function as the example showed.
```cpp
        wasm_memory_delete(memory);
        wasm_exporttype_vec_delete(&exportTypes);
        wasm_extern_vec_delete(&exports);
        wasm_instance_delete(instance);
        wasm_module_delete(module);
        wasm_store_delete(store);
        wasm_engine_delete(engine);
```

### Steps to reproduce

1. call wasm in a process over and over again

### Expected behavior

when stop the wasm engine, the memory should free

### Actual behavior

when the wasm execution stopped, the memory usage of process is not go down

### Additional context

![image](https://user-images.githubusercontent.com/15418097/111562397-1b3f9b00-87d1-11eb-85a9-54c92cbc6b9a.png)

MarkMcCaskey:
Thanks for the bug report!

That's correct -- as you've probably seen, the Wasm C API essentially has ref-counting / GC semantics on when things get freed. I believe this is to make the API more generic across all possible Wasm VMs. Though if you're freeing everything you have access to / everything that could be keeping the memory alive and that memory is not being freed, then that is definitely a bug.

I went to look into this and found 2 memory leaks in our `memory` example as well as an ownership bug which I filed at #2197. Unfortunately I can't get valgrind to work on my version of MacOS right now and can't verify that this fixed the leak and I don't have much time to devote to this right now, but can you take a look at #2196 and see if that resolves your problem?

I suspect the `exports` variable is keeping the memory alive -- the other memory leak (the WAT text bytes) shouldn't be related at all.

bxq2011hust:
Thanks you guys for quick response, I'm trying the new commit now to see if the memory stop grow

bxq2011hust:
I read the #2196 , it seems the PR only modify doc and tests. I try the new commit with the code below, the memory is still growing up, I think that I use Wasmer C API as the example shown. Is there anything I missed?

https://github.com/bxq2011hust/hera/blob/620a6f79b3aaef451bbe9cac7691648e71c6d3e8/src/wasmc.cpp#L1237-L1498

[...]

PR #2196, opened 2021-03-18
Update `memory` Wasm C API example

MarkMcCaskey:
Related to #2193

While working on this, I noticed that our implementation of the Wasm C API diverges from what `wasm.h` says in that we return owned data for `wasm_extern_as_*` and we free it with

```
    wasm_memory_delete(memory);
    wasm_func_delete(mem_size);
    wasm_func_delete(set_at);
    wasm_func_delete(get_at);
```

this data should not be `own`ed and we should not free it. I'll file a separate issue for this to follow up on.

### DIFF
--- a/lib/c-api/README.md
+++ b/lib/c-api/README.md
@@ -43,6 +43,7 @@ int main(int argc, const char* argv[]) {
     wasm_byte_vec_new(&wat, strlen(wat_string), wat_string);
     wasm_byte_vec_t wasm_bytes;
     wat2wasm(&wat, &wasm_bytes);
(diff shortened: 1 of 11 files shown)


Hywan (approved):

Hywan:
Did you check the doctests (written with `inline-c`)?

MarkMcCaskey:
> Did you check the doctests (written with `inline-c`)?

No, I didn't check for this bug anywhere else. Good call though, it's probably best to do it before merging this PR so we don't forget.

jubianchi (approved):

PR #2201, opened 2021-03-19
feat: Implement `MemoryUsage` for `Instance`

Hywan:
# Description

This patch implements `loupe::MemoryUsage` for `wasmer::Instance`.

~~This PR includes https://github.com/wasmerio/wasmer/pull/2200. To review unique patches: https://github.com/Hywan/wasmer/compare/feat-memory-usage-module...feat-memory-usage-instance?expand=1~~

# Review

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ wasmer-cache = { version = "1.0.2", path = "lib/cache", optional = true }
 wasmer-types = { version = "1.0.2", path = "lib/wasmer-types" }
 wasmer-middlewares = { version = "1.0.2", path = "lib/middlewares", optional = true }
 cfg-if = "1.0"
-loupe = { path = "../loupe/crates/loupe" }
 
 [workspace]
 members = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,7 @@ criterion = "0.3"
 lazy_static = "1.4"
 wasmer-engine-dummy = { path = "tests/lib/engine-dummy" }
 tempfile = "3.1"
+loupe = "0.1"
 
 [features]
 # Don't add the compiler features in default, please add them on the Makefile
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,6 @@ criterion = "0.3"
 lazy_static = "1.4"
 wasmer-engine-dummy = { path = "tests/lib/engine-dummy" }
 tempfile = "3.1"
-loupe = "0.1"
 
 [features]
 # Don't add the compiler features in default, please add them on the Makefile
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ wasmer-cache = { version = "1.0.2", path = "lib/cache", optional = true }
 wasmer-types = { version = "1.0.2", path = "lib/wasmer-types" }
 wasmer-middlewares = { version = "1.0.2", path = "lib/middlewares", optional = true }
 cfg-if = "1.0"
+loupe = { path = "../loupe/crates/loupe" }
 
 [workspace]
 members = [
--- a/lib/api/src/exports.rs
+++ b/lib/api/src/exports.rs
@@ -3,7 +3,7 @@ use crate::import_object::LikeNamespace;
 use crate::native::NativeFunc;
 use crate::WasmTypeList;
 use indexmap::IndexMap;
-use loupe_derive::MemoryUsage;
+use loupe::MemoryUsage;
 use std::fmt;
 use std::iter::{ExactSizeIterator, FromIterator};
 use std::sync::Arc;
--- a/lib/api/src/exports.rs
+++ b/lib/api/src/exports.rs
@@ -3,6 +3,7 @@ use crate::import_object::LikeNamespace;
 use crate::native::NativeFunc;
 use crate::WasmTypeList;
 use indexmap::IndexMap;
+use loupe_derive::MemoryUsage;
 use std::fmt;
 use std::iter::{ExactSizeIterator, FromIterator};
 use std::sync::Arc;
@@ -61,7 +62,7 @@ pub enum ExportError {
 /// the types of instances.
 ///
 /// TODO: add examples of using exports
-#[derive(Clone, Default)]
+#[derive(Clone, Default, MemoryUsage)]
 pub struct Exports {
     map: Arc<IndexMap<String, Extern>>,
 }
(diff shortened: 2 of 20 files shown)


Cargo.toml
```diff
@@ -26,6 +26,7 @@ wasmer-cache = { version = "1.0.2", path = "lib/cache", optional = true }
 wasmer-types = { version = "1.0.2", path = "lib/wasmer-types" }
 wasmer-middlewares = { version = "1.0.2", path = "lib/middlewares", optional = true }
 cfg-if = "1.0"
+loupe = { path = "../loupe/crates/loupe" }
```
MarkMcCaskey, line 29: Commenting for completeness: this line needs to change before we can ship this

This applies to all `Cargo.toml` files edited with this change

lib/api/src/exports.rs
```diff
@@ -3,6 +3,7 @@ use crate::import_object::LikeNamespace;
 use crate::native::NativeFunc;
 use crate::WasmTypeList;
 use indexmap::IndexMap;
+use loupe_derive::MemoryUsage;
```
MarkMcCaskey, line 6: You can reexport this through `loupe` so users only need to depend on one crate and things work much more smoothly.

In fact, I'd recommend making it a default feature of `loupe` to reexport the derive macro and you can opt-out by disabling default features.

lib/api/src/externals/function.rs
```diff
@@ -22,21 +23,22 @@ use wasmer_vm::{
 };
 
 /// A function defined in the Wasm module
-#[derive(Clone, PartialEq)]
+#[derive(Clone, PartialEq, MemoryUsage)]
 pub struct WasmFunctionDefinition {
     // Address of the trampoline to do the call.
+    #[memoryusage(ignore)]
```
MarkMcCaskey, line 29: Consider `#[loupe(ignore)]` or `#[loupe(skip)]`. I think it's more common to use the crate name as a the unique first element in the attribute than the name of the trait. For example serde, structopt, `WasmerEnv`.

Serde even does `#[serde(skip_serializing)]` and `#[serde(skip_deserializing)]` to disambiguate the two

lib/compiler/src/function.rs
```diff
@@ -35,7 +36,7 @@ pub struct CompiledFunctionFrameInfo {
 
 /// The function body.
 #[cfg_attr(feature = "enable-serde", derive(Deserialize, Serialize))]
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, MemoryUsage)]
```
MarkMcCaskey, line 39: serde is disabled here because it requires `std` I think. If `MemoryUsage` does too, we probably want it to be grouped with the `serde` things, not with the rest. Though the naming of `enable-serde` makes me more uncertain what the intent is here.

lib/vm/src/instance/ref.rs
```diff
@@ -208,3 +210,13 @@ impl Drop for InstanceRef {
...
     }
 }
+
+impl MemoryUsage for InstanceRef {
+    fn size_of_val(&self, tracker: &mut dyn MemoryUsageTracker) -> usize {
+        mem::size_of_val(self) + self.strong.size_of_val(tracker) - mem::size_of_val(&self.strong)
+            + self.instance_layout.size_of_val(tracker)
+            - mem::size_of_val(&self.instance_layout)
+            + self.as_ref().size_of_val(tracker)
+            - mem::size_of_val(&self.instance)
```
MarkMcCaskey, line 220: Is it possible to encapsulate the semantics of this into the loupe library itself more directly? It just looks very confusing without context

lib/vm/src/vmcontext.rs
```diff
@@ -336,6 +345,16 @@ unsafe impl Send for VMMemoryDefinition {}
 /// correctness in a multi-threaded context is concerned.
 unsafe impl Sync for VMMemoryDefinition {}
 
+impl MemoryUsage for VMMemoryDefinition {
+    fn size_of_val(&self, tracker: &mut dyn MemoryUsageTracker) -> usize {
+        if tracker.track(self.base as *const _ as *const ()) {
+            POINTER_BYTE_SIZE * (self.current_length as usize)
+        } else {
+            0
+        }
```
MarkMcCaskey, line 354: Without context this reads as

"If we're not tracking anything at this address, then the size is 0, otherwise it's this size"

I realize it's the opposite, but it reads that way 😄

lib/wasmer-types/src/entity/secondary_map.rs
```diff
@@ -279,6 +281,21 @@ where
...
+    fn size_of_val(&self, tracker: &mut dyn MemoryUsageTracker) -> usize {
+        mem::size_of_val(self)
+            + self
+                .elems
+                .iter()
+                .map(|value| value.size_of_val(tracker) - mem::size_of_val(value))
+                .sum::<usize>()
+    }
+}
+
```
MarkMcCaskey, line 298: I think you probably also need to do `self.default.size_of_val(tracker)`, but that's a small nit pick

lib/vm/src/vmcontext.rs
MarkMcCaskey: I don't know if there's anything actionable to do here, just thought I'd share this feedback

lib/api/src/exports.rs
Hywan: That's a good idea!

lib/api/src/externals/function.rs
Hywan: I agree. Let's move to `#[loupe(skip_memoryusage)]`.

lib/compiler/src/function.rs
Hywan: `enable-serde` isn't related to `std` or `core`/`alloc`. It's just we allowed to run without Serde in the past. Not sure it's still useful today though 🤔.

lib/vm/src/instance/ref.rs
Hywan: I will try to come with an API for that.

lib/vm/src/vmcontext.rs
Hywan: The name `track` doesn't express well what it does. Here is its documentation:

> When first called on a given address returns true, else returns false.

If you can come with a better name, I'll take it!

lib/wasmer-types/src/entity/secondary_map.rs
Hywan: :+1:

MarkMcCaskey (approved):

syrusakbary (approved):

PR #2207, opened 2021-03-23
Simplifies a few pattern-matches

huitseeker:
... which have a direct combinator equivalent on Option / Result.

Tool-aided by [comby-rust](https://github.com/huitseeker/comby-rust).

### DIFF
--- a/benches/static_and_dynamic_functions.rs
+++ b/benches/static_and_dynamic_functions.rs
@@ -1,7 +1,6 @@
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
 
 use wasmer::*;
-use wasmer_engine_jit::JIT;
 
 static BASIC_WAT: &str = r#"(module
     (func $multiply (import "env" "multiply") (param i32 i32) (result i32))
@@ -147,7 +146,7 @@ pub fn run_basic_dynamic_function(store: &Store, compiler_name: &str, c: &mut Cr
     );
 }
 
-fn run_static_benchmarks(c: &mut Criterion) {
+fn run_static_benchmarks(_c: &mut Criterion) {
     #[cfg(feature = "llvm")]
     {
         let store = Store::new(&JIT::new(wasmer_compiler_llvm::LLVM::new()).engine());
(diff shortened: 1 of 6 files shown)


Hywan (approved):
LGTM.

Hywan:
Is it ready to merge?

syrusakbary (approved):

huitseeker:
@Hywan AFAICT, yes

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -303,15 +303,11 @@ impl FuncTranslator {
...
-                Ok(
-                    if let Some(Symbol::LocalFunction(local_func_index)) =
-                        symbol_registry.name_to_symbol(name)
-                    {
-                        Some(RelocationTarget::LocalFunc(local_func_index))
-                    } else {
-                        None
+                Ok(symbol_registry.name_to_symbol(name).map(
+                    |Symbol::LocalFunction(local_func_index)| {
+                        RelocationTarget::LocalFunc(local_func_index)
```
MarkMcCaskey, line 308: Huh, I didn't know you could pattern match in `map` like this and have it work 🤔

MarkMcCaskey (approved):
lgtm

lib/compiler-llvm/src/translator/code.rs
Hywan: Huh? I missed this one. Are we sure about this rewrite?

MarkMcCaskey: I think if it compiles then it's probably correct. It looks like something that may not compile to me because `map` isn't exhaustive, but if it is, then I think it'll probably do what we want

MarkMcCaskey:
bors try

MarkMcCaskey:
Alright, looks like there's a number of different compilation issues (different CI failures failed for different reasons too), we'll need to get this all compiling before we can ship it.

MarkMcCaskey:
bors try

huitseeker:
@MarkMcCaskey Apologies, took me a while to figure out the interleaving of features in the project.

MarkMcCaskey:
> @MarkMcCaskey Apologies, took me a while to figure out the interleaving of features in the project.

No problem, we should definitely document our build process somewhere as we diverge from what's standard in Rust a bit -- namely that `cargo build` won't do a proper build without more features (this is because of the need for different defaults / sets of features enabled on different platforms and architectures and this is hard or impossible to express cleanly without an external tool like a Makefile).

I just filed an issue, #2213, so we can document this more explicitly going forward!

PR #2208, opened 2021-03-23
Implement to/from casting for `extern` w/ no allocation

### LINKED ISSUE #2197: Mismatch with Wasm C API on `wasm_extern_as_*` functions return value
Found while making #2196, the `wasm_extern_as_*` functions do not return `own`ed values. They should be like casts, not allocating new data that needs to be freed. This change is best done by refactoring what `wasm_extern_t` looks like on the Rust side, in my opinion. It shouldn't be too hard but it is a breaking change.

MarkMcCaskey:
Resolves #2197

This should be good to go now.

The only question is how we want to deal with a breaking change this big. @Hywan and I previously discussed this, it seems other implementers of the wasm C API may have the same bug this is fixing

[New Changelog rendered](https://github.com/wasmerio/wasmer/blob/fix/c-api-allocation/lib/c-api/CHANGELOG.md)
[Primary Changelog rendered](https://github.com/wasmerio/wasmer/blob/fix/c-api-allocation/CHANGELOG.md)


Hywan (commented):
Looks good so far! Thanks for tackling this :-).

lib/c-api/src/wasm_c_api/externals/mod.rs
```diff
@@ -6,16 +6,150 @@ mod table;
...
+
+/// All elements in this union must be `repr(C)` and have a
+/// `CApiExternTag` as their first element.
+#[allow(non_camel_case_types)]
+pub(crate) union wasm_extern_inner {
+    function: mem::ManuallyDrop<wasm_func_t>,
+    memory: mem::ManuallyDrop<wasm_memory_t>,
+    global: mem::ManuallyDrop<wasm_global_t>,
+    table: mem::ManuallyDrop<wasm_table_t>,
+}
```
Hywan, line 27: :+1:

lib/c-api/src/wasm_c_api/externals/mod.rs
```diff
@@ -6,16 +6,150 @@ mod table;
...
 pub use memory::*;
-use std::sync::Arc;
+use std::mem;
 pub use table::*;
-use wasmer::{Extern, Instance};
+use wasmer::{Extern, ExternType};
 
 #[allow(non_camel_case_types)]
-#[derive(Clone)]
+#[repr(transparent)]
```
Hywan, line 14: :+1:

lib/c-api/src/wasm_c_api/externals/mod.rs
```diff
@@ -6,16 +6,150 @@ mod table;
...
+        assert_eq!(size_of::<wasm_extern_t>(), size_of::<wasm_global_t>());
+        assert_eq!(size_of::<wasm_extern_t>(), size_of::<wasm_table_t>());
+    }
+}
+
+impl Drop for wasm_extern_inner {
+    fn drop(&mut self) {
+        unsafe {
+            let tag = self.function.tag;
+            dbg!(tag);
```
Hywan, line 47: Please remove that :-).

lib/c-api/src/wasm_c_api/externals/mod.rs
```diff
@@ -6,16 +6,150 @@ mod table;
...
+            match tag {
```
Hywan, line 48:
```suggestion
            match self.function.tag {
```

lib/c-api/src/wasm_c_api/externals/mod.rs
```diff
@@ -6,16 +6,150 @@ mod table;
...
+                CApiExternTag::Function => mem::ManuallyDrop::drop(dbg!(&mut self.function)),
+                CApiExternTag::Global => mem::ManuallyDrop::drop(dbg!(&mut self.global)),
+                CApiExternTag::Table => mem::ManuallyDrop::drop(&mut self.table),
+                CApiExternTag::Memory => mem::ManuallyDrop::drop(dbg!(&mut self.memory)),
```
Hywan, line 52: Be careful with the `dbg!` here:

```suggestion
                CApiExternTag::Function => mem::ManuallyDrop::drop(&mut self.function),
                CApiExternTag::Global => mem::ManuallyDrop::drop(&mut self.global),
                CApiExternTag::Table => mem::ManuallyDrop::drop(&mut self.table),
                CApiExternTag::Memory => mem::ManuallyDrop::drop(&mut self.memory),
```

lib/c-api/src/wasm_c_api/externals/mod.rs
```diff
@@ -31,67 +165,41 @@ pub unsafe extern "C" fn wasm_extern_copy(r#extern: &wasm_extern_t) -> Box<wasm_
...
-) -> Option<Box<wasm_extern_t>> {
-    let func = func?;
-
-    Some(Box::new(wasm_extern_t {
-        instance: func.instance.clone(),
-        inner: Extern::Function(func.inner.clone()),
-    }))
+pub unsafe extern "C" fn wasm_func_as_extern(func: Option<&wasm_func_t>) -> Option<&wasm_extern_t> {
+    std::mem::transmute::<Option<&wasm_func_t>, Option<&wasm_extern_t>>(func)
 }
```
Hywan, line 170: I woud prefer to get `pub extern "C" fn…` here, and an `unsafe { … }` block inside the funciton, with a comment that explains why it's safe to transmute in this case :-).

We may want to add a `debug_assert!` to ensure the size/layout is correct between the both?

Hywan (approved):
Very well done!

I see you've updated the tests. Did you check the doctests too?

I'm approving the PR despites there is `dbg!` calls that need to be removed :-).

Issue #1127, opened 2020-01-09
Yielding from host calls


bkolobara:
I'm currently embedding wasmer into a Rust project. So far I'm really happy with it and I made great progress.

One feature I need though is being able to call from WASM back to Rust and then suspend the executing of WASM until some IO finishes. I'm basically trying to embed wasmer into an async/await environment. From the perspective of wasm it would be a blocking call (runtime suspended). [Lucet exposed an API to do this](https://docs.rs/lucet-runtime/0.4.1/lucet_runtime/#yielding-and-resuming), but I couldn't find anything similar in wasmer.

What would be the best approach to implement something like this? I would also be happy to contribute some code if someone pointed me in the right direction. Thanks!

syrusakbary:
That would be a great addition. We would love to support yielding from host calls.

I think first, we need to figure out a good API to use it (we can use this issue to make proposals) and then just create a PR to implement it.

There are a few ways we can get it working with the following green-threads/fibers approaches:
* Via  [libfringe](https://github.com/edef1c/libfringe). Cons: only support in *Nix platforms
* Via [mioco](https://github.com/dpc/mioco). Cons: unmantained
* Via [genawaiter](https://github.com/whatisaphone/genawaiter).

After reviewing Lucet API and all the different yielding libraries implementations, it seems the simplest way to achieve it is via async generators (genawaiter), as it will support any platform and since it relies on native async/await and it's implementation is close to zero-cost. 

Here's an example API that I have in mind. Thoughts? @bkolobara @MarkMcCaskey 

```rust
pub enum Factorial {
    Multiply(u64, u64),
    Result(u64),
}

#![wasmer_generator]
pub unsafe extern "C" fn factorial(
    &mut vmctx,
    n: u64,
) -> u64 {
    let result = if n <= 1 {
        1
    } else {
        let n_rec = factorial(vmctx, n - 1);
        vmctx.yield(Factorial::Multiply(n, n_rec))
    };
    vmctx.yield(Factorial::Result(result))
    result
}

let import_object = imports! {
  factorial => factorial
};

// Run the Wasm
let instance = instantiate(WASM, &import_object)?;

let result = instance
        .dyn_func("run")?
        .call(&[Value::I32(42)])?;

let mut factorials = vec![];

while let Yield(val) = result.resume() {
    match k {
        Factorial::Multiply(n, n_rec) => {
            // guest wants us to multiply for it
            res = inst.resume_with(n * n_rec);
        }
[...]

PR #2219, opened 2021-03-29
RFC Host yielding API

kaimast:
Hi,

This is not intended to be merged into master yet, but I am curious about your thoughts on the approach. Essentially this pull request tries to fix/implement #1127 using the `async-wormhole` crate.

After enabling the `async` feature and you can call a function like so:
```Rust
let instance = Instance::new(&module, import_object);

let stack = switcheroo::stack::EightMbStack;
instance.call_with_stack("my_func", stack);
```

This then executes the function in the userspace thread. You can yield host calls by exposing the Yielder to your WasmerEnv.
```Rust
use wasmer::{LazyInit, Yielder};

#[ derive(WasmEnv) ]
sturct MyEnv {
     #[wasmer(yielder)]
      yielder: LazyInit<Yielder>
}
```

And then in the function call you can access the Yielder (which is just a clonable wrapper around `async-wormhole`'s API).
```Rust
fn my_host_func(env: &MyEnv) {
     let yielder = env.yielder.get_ref().unwrap().get();

     yielder.async_syspend(async {
             my_async_func().await;
     });
}
```


Hywan (commented):
I salute the amount of work to make all that working correctly! Thanks!

I'll need some times to review this PR carefully and to think about the feature design exactly :-). I reckon the feature is great, but it will take time to evaluate the design/the approach.

lib/api/src/env.rs
```diff
@@ -76,6 +76,10 @@ pub trait WasmerEnv: Clone + Send + Sync {
     fn init_with_instance(&mut self, _instance: &Instance) -> Result<(), HostEnvInitError> {
         Ok(())
     }
+
+    #[ cfg(feature="async") ]
+    /// Set the thread-specific yielder
+    fn set_yielder(&mut self, _: *const std::ffi::c_void) {}
```
MarkMcCaskey, line 82: This probably needs to be `unsafe` and have a `# Safety` section describing how to correctly call and/or implement this function

lib/api/src/externals/function.rs
```diff
@@ -269,10 +324,25 @@ impl Function {
                 Env::init_with_instance(&mut *env.ctx.env, instance)
             };
 
+        #[ cfg(feature="async") ]
+        let (host_env, metadata) =  {
+            let host_env_set_yielder_fn = |env_ptr, yielder_ptr| {
+                let env: &mut Env = unsafe { std::mem::transmute(env_ptr) };
```
MarkMcCaskey, line 330: It's nice to be super explicit about what exactly is being transmuted from what to what to make bugs and intent more obvious. I'd prefer all `transmute` being `transmute::<A,B>` with explicit types whenever possible.

lib/api/src/externals/function.rs
```diff
@@ -67,6 +67,53 @@ pub struct Function {
...
+            host_env_clone_fn,
+            host_env_set_yielder_fn,
+            host_env_drop_fn,
+        )
+    };
+
+    (env, metadata)
+}
+
+#[ cfg(not(feature="async")) ]
```
MarkMcCaskey, line 116: I'd prefer having the `cfg` logic internal to the original function rather than having 2 copies just because it's so precise, dangerous, and error-prone. The reason this function exists in the first place is to try to avoid duplicating this kind of code.

lib/api/src/externals/function.rs
```diff
@@ -386,8 +456,18 @@ impl Function {
         let function = inner::Function::<Args, Rets>::new(func);
         let address = function.address();
 
-        let (host_env, metadata) =
-            build_export_function_metadata::<Env>(env, Env::init_with_instance);
+        #[ cfg(feature="async") ]
+        let (host_env, metadata) = {
+            let host_env_set_yielder_fn = |env_ptr, yielder_ptr| {
+                let env: &mut Env = unsafe { std::mem::transmute(env_ptr) };
```
MarkMcCaskey, line 462: same here

lib/api/src/externals/yielder.rs
```diff
@@ -0,0 +1,27 @@
...
+        Self{ inner }
+    }
+
+    /// Get the `AsyncYielder`
+    pub fn get(&self) -> &mut AsyncYielder<Result<Box<[Val]>, RuntimeError>> {
+        let yielder: &mut AsyncYielder<Result<Box<[Val]>, RuntimeError>>= unsafe { std::mem::transmute(self.inner) };
+
+        yielder
+    }
+}
```
MarkMcCaskey, line 23: Either `new` needs to be `unsafe` or `get` needs to be unsafe (or possibly both)

lib/api/src/instance.rs
```diff
@@ -154,6 +155,46 @@ impl Instance {
...
+
+            let mut results = vec![Val::null(); func.result_arity()];
+
+            match &func.definition {
+                FunctionDefinition::Wasm(wasm) => {
+                    let res = func.call_wasm(&wasm, params, &mut results);
+
+                    if let Err(e) = res {
+                        return Err(e);
+                    }
```
MarkMcCaskey, line 185: ⛳ 
```Rust
let res = func.call_wasm(&wasm, params, &mut results)?;
```

lib/api/src/module.rs
```diff
@@ -263,9 +263,11 @@ impl Module {
...
     ) -> Result<InstanceHandle, InstantiationError> {
         unsafe {
-            let instance_handle =
-                self.artifact
-                    .instantiate(self.store.tunables(), resolver, Box::new(()))?;
+            let instance_handle = self.artifact.instantiate(
+                self.store.tunables(),
+                resolver,
+                Box::new(()),
+            )?;
```
MarkMcCaskey, line 270: Looks like unrelated formatting changes, we use 1 behind latest stable (though I think we may not have updated to 1.50 in this repo yet), would be good to run that version of rustfmt on the code

lib/derive/src/parse.rs
```diff
@@ -13,6 +13,11 @@ pub enum WasmerAttr {
         identifier: Option<LitStr>,
         span: Span,
     },
+    #[ cfg(feature="async") ]
+    Yielder {
+        identifier: Option<LitStr>,
+        span: Span,
+    }
```
MarkMcCaskey, line 20: Minor nit (I haven't review all the derive macro changes yet) but we probably want to unconditionally parse this attribute and error if the feature isn't enabled or otherwise generate code that will provide a useful error message in that case. Don't worry about doing all that work now for this experimental PR, just writing this down for future reference.

MarkMcCaskey (commented):
Interesting PR, thanks for going through and doing it. I don't understand how it all works yet, I'll probably have to look into what async-wormhole is doing to figure that out.

It's nice to have a proof of concept for this. I'm not currently well-informed enough to know what trade-offs this approach makes and if it's the right set of trade-offs for wasmer, but it seems minimally invasive...

Though changes like this in a system as complex as a Wasm VM can have a lot of side effects that we'll need to think through as well. I'll try to follow up on this more later once I've done some more research

kaimast:
Thanks for taking a look. I should also note that the commit history is kind of a mess right now and the pull request makes some unnecessary changes, mainly because I pulled these changes out of a larger branch.

kaimast:
> Some thoughts on the API design.
> 
> I believe a more ergonomic API would be:
> ## Call async function
> 
> ```rust
> let instance = Instance::new(&module, import_object);
> 
> // It can be EightMbStack or OneMbStack
> // This call is optional (and it can be EightMbStack by default if not called)
> store.set_stack(EightMbStack);
> 
> instance.call_async("my_func");
> ```
> 
> ## Function definition
> 
> We can allow functions to be defined async, and detect them with the Rust type system (automatically wrapping them).
> 
> ```rust
> async fn my_host_func(env: &MyEnv) {
>     my_async_func().await;
> }
> ```
> 
> I'm not super convinced of having the yielder in the Env, but I'm not super sure what are the alternatives. Thoughts?
> 
> Apart from this, we will need to add tests and benchmark the calls (similarly as we do benchmarking of [other function calls](https://github.com/wasmerio/wasmer/blob/master/benches/static_and_dynamic_functions.rs))

I like these changes a lot. Will close this pull request for now and reopen once I have time to implement them.

Thanks for all the feedback!

PR #2224, opened 2021-03-31
Cranelift: enable SIMD based on actual Wasm features

jubianchi:
# Review


syrusakbary (approved):

jubianchi:
bors try

MarkMcCaskey (changes requested):
I messaged about this on Slack, but I don't think we want `Features::enable_simd` to be related to this other use of simd which is entirely unrelated as I understand it. If we want to control low-level compiler details with `Features`, I think we need to add new fields, ideally in a new sub-struct to keep things organized

MarkMcCaskey (approved):
I think the title is misleading now, but my previous objection seems to no longer be in the PR. I do think we need to be careful about separating the ideas of codegen simd from Wasm simd though

jubianchi:
bors try

syrusakbary (approved):

PR #2225, opened 2021-04-01
Make Windows +crt-static logic c-api specific

MarkMcCaskey:
This seems to fix the wapm build issue we're seeing in wasmer nightly CI while not affecting the C API that we produce for windows


MarkMcCaskey:
bors try

syrusakbary (approved):

MarkMcCaskey:
This fix doesn't actually work, the rustflags don't apply to the C API when moved there. Need to find a new solution

Hywan (approved):
Ideally, we should create a `Makefile` just for the C API, but this patch is fine for the moment :-).

PR #2236, opened 2021-04-06
Fix where quoting happens with `RUSTFLAGS` in Makefile

MarkMcCaskey:
This is blocking release via CI (as seen by wasmer nightly when the cache is cleared when cross-compiling on Windows). Tested locally and it appears to work, the issue is that:

`RUSTFLAGS="string 1" "string 2"` (what we were doing before) causes the command to fail as it tries to execute the second string as a command or something. The solution is to not quote when building up `RUSTFLAGS` and to quote when using it.

Note: it's possible that an existing env var with `RUSTFLAGS` using quotes inside it could break this but 🤷 .

Note2: `RUSTFLAGS='"string 1" "string 2"'` does not work -- I tried that first.

# Review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -366,9 +366,9 @@ build-wasmer-debug:
 # incremental = false
 # codegen-units = 1
 # rpath = false
-build-wasmer-headless-minimal: RUSTFLAGS += "-C panic=abort" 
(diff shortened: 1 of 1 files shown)


syrusakbary (approved):

MarkMcCaskey:
Merged manually because CI passed but was invalidated by another manual merge into master

Issue #2122, opened 2021-02-15
Undefined behavior in `EmEnv::set_memory`


caelunshun:
### Describe the bug

`EmEnv::set_memory` is implemented as:

```rust
pub fn set_memory(&mut self, memory: Memory) {
        let ptr = Arc::as_ptr(&self.memory) as *mut _;
        unsafe {
            *ptr = Some(memory);
        }
    }
```

There are two problems with this code:
* casting `*const Memory` to `*mut Memory` creates a mutable pointer from an immutable pointer, which is UB.
* the `Arc` could have been cloned and sent to another thread, in which case writing to `ptr` is a data race.

Hywan:
Maybe we should use [`Arc::get_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut) or `Arc::make_mut` in this case. Thoughts @MarkMcCaskey?

protheory8:
How does one solve the second problem?

MarkMcCaskey:
Now that we have derive support with https://docs.rs/wasmer/1.0.2/wasmer/trait.WasmerEnv.html , I believe this should be solvable with that. It'll just take refactoring some code to make it work. If the memory is imported and not exported from Emscripten then it may be a bit more work because the `WasmerEnv` derive macro doesn't support grabbing imports yet.

I don't think there's an easy solution though; the current behavior is a consequence of the way the code was structured previously and an attempt to avoid spending too much time on it. I'd be surprised if there's a solution that doesn't involve rewriting a decent chunk of `emscripten/lib.rs`.

I can take a quick look again and see if there's a straight-forward solution that I missed, but otherwise, I probably won't be able to prioritize this for a while.

PR #2239, opened 2021-04-10
Fix unsound mut pointer to potentially cloned Arc

rickwebiii:
# Description

Fix unsound behavior when EmEnv is cloned. Since it impls Clone, that's a valid and legal thing to do. If you clone an EmEnv, the Arcs both point at the the same Option<Memory>. It is unsound to take a mutable reference or pointer to an object while there exists another reference.

With this change, we remove the unsafe block. If EmEnv has been cloned, the Option<Memory> gets cloned in the instance you're trying to write memory to. The old instance will point at whatever it previously had.

# Review

### DIFF
--- a/lib/emscripten/src/lib.rs
+++ b/lib/emscripten/src/lib.rs
@@ -92,10 +92,7 @@ impl EmEnv {
     }
 
     pub fn set_memory(&mut self, memory: Memory) {
-        let ptr = Arc::as_ptr(&self.memory) as *mut _;
-        unsafe {
-            *ptr = Some(memory);
-        }
+        Arc::make_mut(&mut self.memory).replace(memory);
     }
 
     /// Get a reference to the memory


MarkMcCaskey (commented):
Thanks for the patch and looking into this! Unfortunately I don't think this solves the issue.  Looking at the [`Arc::make_mut` docs](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.make_mut) we see that:
> If there are other Arc or Weak pointers to the same allocation, then make_mut will create a new allocation and invoke clone on the inner value to ensure unique ownership. This is also referred to as clone-on-write.

Tracing through the logic of `EmEnv` just now it looks like this will not have the intended effect of setting the memory for all the functions using `EmEnv`. Meaning that none of the imported functions will be able to access memory with this change.

Unfortunately, I don't think there is an easy fix for this without restructuring a lot of the code in `emscripten/src/lib.rs`. I can't remember why this doesn't just use  a `Mutex` but I think there's a reason that doesn't work too. I'll spend a bit of time investigating it again to regain some context here and follow up here or on the issue with what I've learned.

rickwebiii:
Yeah, there's quite a bit to unpack here and this probably needs some more holistic approach. Would you like me to create an issue?

MarkMcCaskey:
I think there's already an issue open at #2122 ; will follow up there later today

PR #2241, opened 2021-04-14
Fix UB in setting memory in emscripten EmEnv

### LINKED ISSUE #2122: Undefined behavior in `EmEnv::set_memory`

MarkMcCaskey:
Fixes the UB and another issue caused by the reftypes PR merging in

Resolves #2122


MarkMcCaskey:
bors try

MarkMcCaskey:
bors try-

syrusakbary (approved):

PR #2245, opened 2021-04-16
Add support for `cargo fuzz fmt` to output testcases.

nlewycky:


nlewycky:
I think there's a bug here where the termination limit from wasm-smith isn't being applied in equivalence_jit...

fuzz/fuzz_targets/jit_singlepass.rs
```diff
@@ -20,9 +20,25 @@ impl Config for NoImportsConfig {
         false
     }
 }
+#[derive(Default, Arbitrary)]
+struct WasmSmithModule(ConfiguredModule<NoImportsConfig>);
+impl std::fmt::Debug for WasmSmithModule {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&wasmprinter::print_bytes(self.0.to_bytes()).unwrap())
+    }
+}
```
syrusakbary, line 29: Can we move this into a common file?

nlewycky: Agreed. Unfortunately, I don't know how to make the directory structure and set things up so that fuzzers share common library code. I'm not saying it's hard, but I did look into it once before.

syrusakbary (approved):

Issue #2243, opened 2021-04-14
Support Absolute Symlinks In WASI


TheLostLambda:
### Motivation

Hello again! Wasmer has become a core part of a plugin system I maintain for [Zellij](https://github.com/zellij-org/zellij), but one of my plugins (a file-browser) chokes on absolute symlinks!

```
thread 'wasm' panicked at 'not implemented: Absolute symlinks are not yet supported': /home/tll/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-wasi-1.0.2/src/state/mod.rs:812
```

### Proposed solution

I'd love to turn the "are not yet" into an "are now"! I'm no expert with the code-base here, but I'd be happy to help if it's nothing *too* tricky to implement.

### Alternatives

Just don't use absolute symlinks, I suppose, but that's a tall ask of the end-user

MarkMcCaskey:
Thanks for the issue! That's great to hear about Zellij!

I don't think implementing the absolute symlink logic will be particularly difficult at all, the main consideration is just the sandboxing of WASI itself and making sure that there aren't ways to get access to files outside of the preopened directories beyond what's reasonable. I don't think the behavior of symlinks outside of preopened dirs has been defined clearly by WASI anywhere, but I think it's probably safe to assume that it should act as if it's in that preopened directory.

I think we can probably enable reading absolute symlinks without too many sandbox escape considerations, though I'd like to review a few things before being comfortable shipping it. Creating absolute pathed symlinks sounds a lot more complicated, although technically all paths a WASI program can see are relative to a preopened directory, so maybe the notion of creating an "absolute path" symlink from WASI doesn't even make sense...

The places that need the most scrutiny are the core `Kind` enum for wasi files to make sure the abstractions are coherent with symlinks to places outside of a preopened dir and the symlinking logic itself. Aside from that, anywhere in `wasi/src/syscalls/mod.rs` where symlinks are dealt with directly should probably be reviewed again too just in case.

TheLostLambda:
Sounds great! Thank you for all of the pointers as well! I can probably start seriously taking a look at things in the next week or so. It sounds like a good plan to treat the symlinks the same as they would be treated outside of WASI, and perhaps creating them could be a separate PR (as I agree that would need significantly more thought and discussion).

Thanks again for the guidance and I'll keep you updated on any progress I make!

MarkMcCaskey:
[...]

PR #2247, opened 2021-04-21
Clean up core WASI FS code

MarkMcCaskey:
Started looking into #2243 and noticed a few things that could be cleaned up and fixed.

The first fix just makes the code more explicit and removes a mutable variable. This is good because it also makes it more obvious that the final branch can never exit, it always returns.

The second fix aligns our path stripping behavior to that of WASI libc (we take the max, not the first, and favor later preopens over newer ones).

# Review

### DIFF
--- a/lib/wasi/src/state/mod.rs
+++ b/lib/wasi/src/state/mod.rs
@@ -784,7 +784,7 @@ impl WasiFs {
                             let file_type = metadata.file_type();
                             // we want to insert newly opened dirs and files, but not transient symlinks
                             // TODO: explain why (think about this deeply when well rested)
-                            let mut should_insert = false;
+                            let should_insert;
 
                             let kind = if file_type.is_dir() {
                                 should_insert = true;
@@ -803,6 +803,7 @@ impl WasiFs {
                                     fd: None,
                                 }
                             } else if file_type.is_symlink() {
+                                should_insert = false;
                                 let link_value = file.read_link().ok().ok_or(__WASI_EIO)?;
                                 debug!("attempting to decompose path {:?}", link_value);
 
@@ -957,18 +958,15 @@ impl WasiFs {
         Ok(cur_inode)
     }
 
-    /// Splits a path into the first preopened directory that is a parent of it,
-    /// if such a preopened directory exists, and the rest of the path.
-    ///
-    /// NOTE: this behavior seems to be not the same as what libpreopen is
-    /// doing in WASI.
-    ///
-    /// TODO: evaluate users of this function and explain why this behavior is
-    /// not the same as libpreopen or update its behavior to be the same.
+    /// Splits a path into the preopened directory that has the largest prefix of
+    /// the given path, if such a preopened directory exists, and the rest of the path.
     fn path_into_pre_open_and_relative_path(
         &self,
         path: &Path,
     ) -> Result<(__wasi_fd_t, PathBuf), __wasi_errno_t> {
+        let mut max_prefix_len = 0;
+        let mut max_stripped_path = None;
+        let mut max_po_fd = None;
         // for each preopened directory
         for po_fd in &self.preopen_fds {
             let po_inode = self.fd_map[po_fd].inode;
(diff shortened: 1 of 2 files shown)


lib/wasi/src/state/mod.rs
```diff
@@ -978,46 +976,23 @@ impl WasiFs {
...
             }
         }
-
-        let mut out = PathBuf::new();
-        for p in stack.iter().rev() {
-            out.push(p);
+        if max_prefix_len == 0 {
+            Err(__WASI_EINVAL) // this may not make sense depending on where it's called
+        } else {
+            Ok((max_po_fd.unwrap(), max_stripped_path.unwrap().to_owned()))
```
Earthmark, line 994: (I am very new, but I figure I should introduce myself by bein useful with a code review!)

I wonder if it is possible to structure this so the unwraps are not required. It seems like this function pivots once a candidate is found, would it be possible to fold the three parameters into a single Result value, and have the last step here be over the existing result methods instead of using unwrap?

The unwrap was a red flag for me, and I almost posted about the empty-list case, however these unwraps are safe because the max_prefix_len check and that if it's not null, these two are populated. Recognizing that requires me to mull over the state machine to figure out that the three are correlated, where as I feel the code could directly tell me that.

Either way, still works the same, just a clarity for me!

lib/wasi/src/state/mod.rs
```diff
@@ -957,18 +958,15 @@ impl WasiFs {
...
-    /// Splits a path into the first preopened directory that is a parent of it,
-    /// if such a preopened directory exists, and the rest of the path.
-    ///
-    /// NOTE: this behavior seems to be not the same as what libpreopen is
-    /// doing in WASI.
-    ///
-    /// TODO: evaluate users of this function and explain why this behavior is
-    /// not the same as libpreopen or update its behavior to be the same.
+    /// Splits a path into the preopened directory that has the largest prefix of
+    /// the given path, if such a preopened directory exists, and the rest of the path.
```
Earthmark, line 962: The wording of this comment confused me for a bit, this is just my perspective and additional data points should be gathered.

Either way it's a massive nitpick, feel free to ignore this.

With this though, my mind interperted the second sentence fragment "if such a preopened directory exists" to be the final part of the sentence, and then the "and the rest of the path" was confusing and didn't correctly link in my mind that it was talking about the second tuple value.

Instead, can I request that the topics of the "path of the preopened directory" and the "rest of the path" be located near eachother in the comment, with the error case being near the end of the line? That way it mirrors the structure of the result value, so reading the return value and comment is a similar read.

Hywan (approved):

lib/wasi/src/state/mod.rs
MarkMcCaskey: Yeah, this is a good observation! I agree, but it's a bit tricky here.

Sorry by the way -- I missed these comments before -- reviewing them now, I'll make a new PR with any updates I make. Thanks!

MarkMcCaskey: In theory, we can solve this problem with a typed state machine, I'm just not sure it's really worth it given the size of this function. If it were larger, I'd definitely be in favor though. The thing is it feels like I'll at least double the amount of code to do it that way...

Like

```Rust
enum BaseFdAndRelPath {
  None,
  BestMatch  {
    fd: __wasi_fd_t,
    rel_path: PathBuf,
    max_seen: usize,
  }
}

Impl  BaseFdAndRelPath {
  fn max_seen(&self) -> usize {
     match self {
       Self::None => 0,
       Self::BestMatch { max_seen, .. } => max_seen,
     }
  }
}
```

Actually, that's not too bad with the method... I'll try it out

MarkMcCaskey: Something like https://doc.rust-lang.org/std/cmp/fn.max_by.html could also work here, unfortunately it's not stable. That would lead to simplified code too.  I'm pretty happy with the result I have, I do think the code is more clear now! Thanks!

The above `Enum` that I posted could definitely be made generic and reused with generic data for finding the most X of a collection of things, could probably even move control flow into a method on this type

MarkMcCaskey: Pushed it up to #2325

> (I am very new, but I figure I should introduce myself by bein useful with a code review!)

Nice to meet you and thanks for the suggestions! We have a community slack as well where you can ping us if we ever miss something like this. If you're interested in hacking on some Wasm / Wasmer stuff, feel free to reach out!

PR #2248, opened 2021-04-21
Add rust-toolchain to have same cargo fmt on ci and locally

ailisp:
# Description

I was using rustc 1.49 which gives a different result in cargo fmt than ci checked. Also based on
https://github.com/actions-rs/toolchain#the-toolchain-file, to make ci always use rust-toolchain specified rust

# Review


ailisp:
oops my editor auto reformat these yamls

rust-toolchain
```diff
@@ -0,0 +1 @@
+1.51.0
```
MarkMcCaskey, line 1: We're currently staying 1 behind latest stable as wasmer crates get used as a libraries and it makes the user experience a bit better. We used to get a lot of error reports and confusion about wasmer only working on nightly when we were using the latest stable. The reason is that the error message you'll get when using rustc version N-1 to compile code using a feature stabilized in version N says something like "feature X is not stable, you can use it on the nightly compiler with #[feature(X)]". This understandably leads people to believe that wasmer only works on nightly.

It's also nice as a library to not rush our users to update too much. In the future we may try to support many Rust versions, but for now 1 behind latest stable is a good compromise.

MarkMcCaskey (approved):
Thanks for the PR! I wasn't aware of the `rust-toolchain` file

MarkMcCaskey:
> oops my editor auto reformat these yamls

Haha I was wondering about that! Don't worry about fixing it up, those changes seem fine

rust-toolchain
ailisp: That is good to know how 1 behind latest stable helps! I'll update it to 1.50

MarkMcCaskey (approved):

MarkMcCaskey:
Merging manually because it's coming from a fork and GitHub prevents merging changes to `workflows` via apps (so bors can't be used)

MarkMcCaskey:
Had to revert because of errors before it runs in CI, will fix it in a branch in this repo so we can test it before merging

PR #2249, opened 2021-04-22
Make Carnelift unwind feature optional

syrusakbary:
# Description

Make unwind feature optional

### DIFF
--- a/lib/compiler-cranelift/src/translator/unwind.rs
+++ b/lib/compiler-cranelift/src/translator/unwind.rs
@@ -1,17 +1,19 @@
 //! A `Compilation` contains the compiled function bodies for a WebAssembly
 //! module.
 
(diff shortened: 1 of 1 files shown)


Hywan:
Why do we want that? :-)

syrusakbary:
> Why do we want that? :-)

Just so the crate compiles if the feature is disabled.

nlewycky:
How is this tested? I realize this makes it compile when the feature is disabled, but does it work?

syrusakbary:
> How is this tested? I realize this makes it compile when the feature is disabled, but does it work? 

I tested manually, it was the easiest path. Do you have thoughts on how this should be tested.

nlewycky:
What was the manual test? Can we run spectests with cranelift and `unwind` feature disabled? If you have a command line for that, we could add it to the Makefile?

PR #2250, opened 2021-04-22
Improved frame_info to be always processed

syrusakbary:
Since we are going to switch serialization methods, the new one doesn’t have a slowdown if we read the data directly.

Thus, we can always operate with processed frame infos


syrusakbary:
bors try

syrusakbary:
bors try

lib/engine-jit/src/artifact.rs
```diff
@@ -141,20 +137,26 @@ impl JITArtifact {
...
+            DeserializeError::CorruptedBinary("Can't read metadata size".to_string())
+        })?;
+        let metadata_slice: &'static [u8] =
+            std::slice::from_raw_parts(&inner_bytes[12] as *const u8, metadata_len as usize);
 
-        // let r = flexbuffers::Reader::get_root(bytes).map_err(|e| DeserializeError::CorruptedBinary(format!("{:?}", e)))?;
-        // let serializable = SerializableModule::deserialize(r).map_err(|e| DeserializeError::CorruptedBinary(format!("{:?}", e)))?;
+        let length = metadata_slice.len();
+        // println!("[DES] Metadata: First 16 bytes {:?}.\nLast 16 bytes: {:?}\nTotal length: {}", &metadata_slice[..16], &metadata_slice[length-16..], length);
+        // println!("[DES] First 50 bytes {:?}.\nLast 50 bytes: {:?}", &bytes[..50], &bytes[inner_bytes.len()-50..]);
```
nlewycky, line 157:
```suggestion
```

lib/engine-jit/src/artifact.rs
```diff
@@ -325,15 +327,32 @@ impl Artifact for JITArtifact {
...
+        let mut metadata_binary = vec![0; 12];
+        let mut writable = &mut metadata_binary[..];
+        let serialized_data = self.serializable.serialize()?;
+        let length = serialized_data.len();
+        leb128::write::unsigned(&mut writable, length as u64).expect("Should write number");
+
+        serialized.extend(metadata_binary);
+
+        let align = std::mem::align_of::<SerializableModule>() as u64;
+        // println!("[SER] Metadata: First 16 bytes {:?}.\nLast 16 bytes: {:?}\nTotal length: {}", &serialized_data[..16], &serialized_data[length-16..], length);
```
nlewycky, line 342:
```suggestion
```

lib/engine-jit/src/artifact.rs
```diff
@@ -325,15 +327,32 @@ impl Artifact for JITArtifact {
...
+        let (serialized, offset) = append_data(serialized, &serialized_data, align);
+        // println!("[SER] First 50 bytes {:?}.\nAlign: {} - Offset: {}\nLast 50 bytes: {:?}", &serialized[..50], align, offset, &serialized[serialized.len()-50..]);
```
nlewycky, line 344:
```suggestion
```

lib/engine-jit/src/artifact.rs
```diff
@@ -325,15 +327,32 @@ impl Artifact for JITArtifact {
...
         Ok(serialized)
     }
 }
+
+pub fn append_data(mut prev_data: Vec<u8>, data: &[u8], align: u64) -> (Vec<u8>, u64) {
+    let align = align as usize;
```
nlewycky, line 350: This code requires that align be a power of two, please assert that.

lib/engine-jit/src/artifact.rs
```diff
@@ -325,15 +327,32 @@ impl Artifact for JITArtifact {
...
+
+        let align = std::mem::align_of::<SerializableModule>() as u64;
+        // println!("[SER] Metadata: First 16 bytes {:?}.\nLast 16 bytes: {:?}\nTotal length: {}", &serialized_data[..16], &serialized_data[length-16..], length);
+        let (serialized, offset) = append_data(serialized, &serialized_data, align);
+        // println!("[SER] First 50 bytes {:?}.\nAlign: {} - Offset: {}\nLast 50 bytes: {:?}", &serialized[..50], align, offset, &serialized[serialized.len()-50..]);
         Ok(serialized)
     }
 }
+
+pub fn append_data(mut prev_data: Vec<u8>, data: &[u8], align: u64) -> (Vec<u8>, u64) {
```
nlewycky, line 349: Maybe this should have a different name like "pad_and_extend"?

lib/engine-jit/src/artifact.rs
```diff
@@ -325,15 +327,32 @@ impl Artifact for JITArtifact {
...
+        let mut writable = &mut metadata_binary[..];
+        let serialized_data = self.serializable.serialize()?;
+        let length = serialized_data.len();
+        leb128::write::unsigned(&mut writable, length as u64).expect("Should write number");
+
+        serialized.extend(metadata_binary);
+
+        let align = std::mem::align_of::<SerializableModule>() as u64;
+        // println!("[SER] Metadata: First 16 bytes {:?}.\nLast 16 bytes: {:?}\nTotal length: {}", &serialized_data[..16], &serialized_data[length-16..], length);
+        let (serialized, offset) = append_data(serialized, &serialized_data, align);
```
nlewycky, line 343: Isn't `offset` unused? Remove it and simplify `append_data`'s signature?

lib/engine-jit/src/artifact.rs
```diff
@@ -141,20 +137,26 @@ impl JITArtifact {
     }
 
     /// Deserialize a JITArtifact
-    pub fn deserialize(jit: &JITEngine, bytes: &[u8]) -> Result<Self, DeserializeError> {
+    pub unsafe fn deserialize(jit: &JITEngine, bytes: &[u8]) -> Result<Self, DeserializeError> {
```
MarkMcCaskey, line 140: `unsafe` functions should have a `# Safety` section saying how the function should be used (i.e. why is it unsafe, what does the user need to manually check)

lib/types/src/archives.rs
```diff
@@ -55,6 +55,51 @@ where
...
+//     fn deserialize(&self, deserializer: &mut D) -> Result<SecondaryMap<K, V>, D::Error> {
+//         let elems: Vec<_> = self.0.deserialize(deserializer)?;
+//         let default = self.1.deserialize(deserializer)?;
+//         Ok(SecondaryMap {
+//             elems,
+//             default,
+//             unused: PhantomData,
+//         })
+//     }
+// }
```
nlewycky, line 101: ???

lib/types/src/lib.rs
```diff
@@ -90,7 +90,7 @@ pub use types::{
 };
 
 #[cfg(feature = "enable-rkyv")]
-pub use archives::{ArchivableIndexMap, ArchivedPrimaryMap};
+pub use archives::{ArchivableIndexMap, ArchivedPrimaryMap}; // ArchivedSecondaryMap
```
nlewycky, line 93: ?

lib/engine-jit/src/artifact.rs
```diff
@@ -141,20 +137,26 @@ impl JITArtifact {
...
             ));
         }
 
-        let inner_bytes = &bytes[Self::MAGIC_HEADER.len()..];
+        let mut inner_bytes = &bytes[Self::MAGIC_HEADER.len()..];
+
+        let metadata_len = leb128::read::unsigned(&mut inner_bytes).map_err(|_e| {
+            DeserializeError::CorruptedBinary("Can't read metadata size".to_string())
+        })?;
+        let metadata_slice: &'static [u8] =
```
MarkMcCaskey, line 152: Hmm, I don't like the `&'static` here. This is very easy to mess up and cause memory issues. We must require `'static` on bytes if we're returning from this function and calling that data `'static`

lib/engine-jit/src/artifact.rs
```diff
@@ -141,20 +137,26 @@ impl JITArtifact {
...
+            std::slice::from_raw_parts(&inner_bytes[12] as *const u8, metadata_len as usize);
```
nlewycky, line 153: How is this safe code? Why isn't the pointer invalidated / dangling?

lib/engine-jit/src/artifact.rs
```diff
@@ -325,15 +327,32 @@ impl Artifact for JITArtifact {
...
+        let (serialized, offset) = append_data(serialized, &serialized_data, align);
+        // println!("[SER] First 50 bytes {:?}.\nAlign: {} - Offset: {}\nLast 50 bytes: {:?}", &serialized[..50], align, offset, &serialized[serialized.len()-50..]);
         Ok(serialized)
     }
 }
+
+pub fn append_data(mut prev_data: Vec<u8>, data: &[u8], align: u64) -> (Vec<u8>, u64) {
+    let align = align as usize;
+    let mut offset = prev_data.len();
+    if offset & (align - 1) != 0 {
```
MarkMcCaskey, line 352: We don't check if `align` is a power of 2 anywhere, I'm assuming this code will behave incorrectly if an `align` that's not a power of 2 is passed in.

This type of code is also very error prone

lib/engine-jit/src/artifact.rs
MarkMcCaskey, line 349: Typically I'd expect a function like this to take a `&mut Vec<u8>` rather than take and return one. Not a huge deal, but it's a bit surprising

lib/engine-jit/src/artifact.rs
MarkMcCaskey, line 350: There's not much reason to use `u64` over `usize` in this function, if we're talking about offsets into an array, `usize` is the most natural type

lib/engine-jit/src/serialize.rs
```diff
@@ -38,9 +32,61 @@ pub struct SerializableCompilation {
...
+        let pos = serializer
+            .serialize_value(self)
+            .map_err(to_serialize_error)? as u64;
+        let mut serialized_data = serializer.into_inner().into_inner();
+        serialized_data.extend_from_slice(&pos.to_le_bytes());
+        if cfg!(target_endian = "big") {
+            serialized_data.extend_from_slice(&[b'b']);
+        } else if cfg!(target_endian = "little") {
+            serialized_data.extend_from_slice(&[b'l']);
+        }
```
MarkMcCaskey, line 58: What exactly is this information for? We unconditionally wrote something in little endian above this statement.

lib/engine-jit/src/serialize.rs
```diff
@@ -38,9 +32,61 @@ pub struct SerializableCompilation {
...
+        Ok(serialized_data)
+    }
+
+    pub unsafe fn deserialize(metadata_slice: &[u8]) -> Result<Self, DeserializeError> {
```
MarkMcCaskey, line 62: `unsafe` functions must have a `# Safety` section explaining how they can be used

lib/engine-jit/src/serialize.rs
```diff
@@ -38,9 +32,61 @@ pub struct SerializableCompilation {
...
+        let archived = Self::archive_from_slice(metadata_slice)?;
+        Self::deserialize_from_archive(archived)
+    }
+
+    unsafe fn archive_from_slice<'a>(
```
MarkMcCaskey, line 67: `unsafe` functions must have a `# Safety` section explaining how they can be used

lib/engine-jit/src/serialize.rs
```diff
@@ -38,9 +32,61 @@ pub struct SerializableCompilation {
...
+            || (cfg!(target_endian = "little") && endian == b'b')
+        {
+            return Err(DeserializeError::Incompatible("incompatible endian".into()));
+        }
+        pos.copy_from_slice(&metadata_slice[metadata_slice.len() - 9..metadata_slice.len() - 1]);
+        let pos: u64 = u64::from_le_bytes(pos);
+        Ok(archived_value::<SerializableModule>(
+            &metadata_slice[..metadata_slice.len() - 9],
+            pos as usize,
+        ))
```
MarkMcCaskey, line 82: Double check that we don't need to do bounds checking here to make sure we actually have enough data

lib/engine/src/serialize.rs
```diff
@@ -1,105 +0,0 @@
-use loupe::MemoryUsage;
-use serde::de::{Deserializer, Visitor};
-use serde::ser::Serializer;
```
MarkMcCaskey, line 3: `serde` may still be good to keep around because it's very generic and widely used

lib/engine-jit/src/artifact.rs
nlewycky: Oh, the function is `unsafe`. It's not clear to me that `'static` lifetime is correct for this reference. Maybe @MarkMcCaskey can take a look?

lib/types/src/archives.rs
```diff
@@ -55,6 +55,51 @@ where
...
+
```
MarkMcCaskey, line 102: Perhaps leave a comment above this explaining what this is and why it's commented out

lib/types/src/lib.rs
MarkMcCaskey, line 93: Clean this up before shipping

lib/vm/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 //! Runtime library support for Wasmer.
 
-#![deny(missing_docs, trivial_numeric_casts, unused_extern_crates)]
```
MarkMcCaskey, line 3: 🙀 

I think we should be testing this in CI, not in the crate unconditionally during development, but especially in `vm`, we need everything to be properly documented

lib/engine-jit/src/artifact.rs
MarkMcCaskey: 100% agreed, `'static` makes me real concerned, left feedback in my comments below

lib/engine-jit/src/serialize.rs
syrusakbary: Rkyv is endian-dependant. We write the endianness of the metadata so when deserializing we check the endianness of the serialized data and fail in case it varies from the host.

lib/types/src/lib.rs
syrusakbary: Fixed

lib/types/src/archives.rs
syrusakbary: I removed it

lib/engine/src/serialize.rs
syrusakbary: Our serialization format should be universal. That means that if someone serializes with Wasmer as a library, the result when reading from anywhere else should succeed.
Because of that, the serialization mechanism should be obscure to our users outside Wasmer.

lib/engine-jit/src/artifact.rs
syrusakbary: `u64` will shield us in `32` bit systems?

lib/engine-jit/src/artifact.rs
syrusakbary: Fixed

lib/engine-jit/src/artifact.rs
syrusakbary: Fixed by checking align

lib/engine-jit/src/artifact.rs
syrusakbary: Fixed

lib/types/src/lib.rs
syrusakbary: Fixed

lib/types/src/archives.rs
syrusakbary: Fixed

lib/engine-jit/src/artifact.rs
syrusakbary: Fixed

lib/engine-jit/src/artifact.rs
syrusakbary: Fixed

djkoloski:
The blocking bug has been fixed as of 0.6.1. rkyv issue: djkoloski/rkyv#116

lib/engine-jit/src/serialize.rs
syrusakbary: Only pub functions though :)

syrusakbary:
Thanks for the super quick fix @djkoloski!

lib/engine-jit/src/serialize.rs
syrusakbary: Fixed!

lib/engine-jit/src/serialize.rs
syrusakbary: Fixed! New comment added

syrusakbary:
bors try

lib/cli/src/commands/run.rs
```diff
@@ -224,7 +224,9 @@ impl Run {
         let module_result: Result<Module> = if !self.disable_cache && contents.len() > 0x1000 {
             self.get_module_from_cache(&store, &contents, &engine_type, &compiler_type)
         } else {
-            Module::new(&store, &contents).map_err(|e| e.into())
+            let module = Module::new(&store, &contents)?;
+            let _serialized = module.serialize()?;
```
Hywan: Why running `serialize` here?

lib/engine-jit/Cargo.toml
```diff
@@ -11,16 +11,16 @@ readme = "README.md"
 edition = "2018"
 
 [dependencies]
-wasmer-types = { path = "../types", version = "1.0.2" }
-wasmer-compiler = { path = "../compiler", version = "1.0.2", features = ["translator"] }
-wasmer-vm = { path = "../vm", version = "1.0.2" }
+wasmer-types = { path = "../types", version = "1.0.2", features = ["enable-rkyv"] }
+wasmer-compiler = { path = "../compiler", version = "1.0.2", features = ["translator", "enable-rkyv"] }
+wasmer-vm = { path = "../vm", version = "1.0.2", features = ["enable-rkyv"] }
```
Hywan, line 16: If `rkyv` is optional, shouldn't we add an optional feature here that enables `enable-rkyv` on `wasmer-types`, `wasmer-compiler` and `wasmer-vm`? Something like:

```toml
[features]
enable-rkyv = ["wasmer-types/enable-rkyv", …]
```

lib/engine-jit/Cargo.toml
```diff
@@ -11,16 +11,16 @@ readme = "README.md"
...
 wasmer-engine = { path = "../engine", version = "1.0.2" }
 # flexbuffers = { path = "../../../flatbuffers/rust/flexbuffers", version = "0.1.0" }
 region = "2.2"
 serde = { version = "1.0", features = ["derive", "rc"] }
-serde_bytes = { version = "0.11" }
-bincode = "1.3"
 cfg-if = "0.1"
+leb128 = "0.2"
+rkyv = "0.6.1"
```
Hywan, line 23: `rkyv` is required here, but it's optinal in `wasmer-compiler`. Why?

Hywan (commented):
The code looks good! I've a question about the required vs. optinal feature `enable-rkyv`. How do you want to deal with that? `serde` has always been optional, so I would expect `rkyv` to be optional too. Inside `wasmer-cli`, we can force `rkyv` though, as it provides much better performance.

lib/cli/src/commands/run.rs
syrusakbary: This was for debugging! I just reverted that commit!

lib/engine-jit/Cargo.toml
syrusakbary: In the case of engine jit, we actually need rkyv for serializing/deserializing (same way, serde was not optional but mandatory).

lib/engine-jit/Cargo.toml
syrusakbary: The engine chooses how to serialize things.
There might be an engine that chooses to serialize with serde, and other that chooses with rkyv. Because of that the serialization mechanism is optional in compiler. But in an engine is mandatory (and unique)

syrusakbary:
> I would expect rkyv to be optional too. Inside wasmer-cli, we can force rkyv though, as it provides much better performance

Engines serialization mechanisms should be fixed. If you compile with the `wasmer-cli` you can expect an embedded user of Wasmer can run `Module::deserialize_from_file(...)` on that exact file. Because of that serialization methods for an engine should be fixed and not movable.

lib/engine-jit/src/artifact.rs
```diff
@@ -141,21 +137,26 @@ impl JITArtifact {
     }
 
     /// Deserialize a JITArtifact
-    pub fn deserialize(jit: &JITEngine, bytes: &[u8]) -> Result<Self, DeserializeError> {
+    ///
+    /// # Safety
+    /// This function is unsafe because rkyv reads directly without validating
+    /// the data.
```
MarkMcCaskey, line 143: Minor nit, but it's useful (at least to me) to also phrase it in a way that is proactive for the prospective caller of the function. i.e. I'm writing some code or reviewing some code and I want to call this function / see if a call to this function is correct; what invariant should I be looking for to check that this function is being called correctly? The answer is probably that the serialized data came from a known valid source.

What's our threat model for maliciously constructed serialization data? Perhaps we need a signature system to mitigate some of these threats. Or at least a hash to fight corruption. Especially if corruption will lead to UB in our system

Either way, I think this is a potentially interesting part of the system from a security / product perspective so it'd be good to think this through and be explicit about what we expect in the safety comments.

lib/engine-jit/src/artifact.rs
```diff
@@ -141,21 +137,26 @@ impl JITArtifact {
...
-        // let serializable = SerializableModule::deserialize(r).map_err(|e| DeserializeError::CorruptedBinary(format!("{:?}", e)))?;
+        let mut inner_bytes = &bytes[16..];
 
-        let serializable: SerializableModule = bincode::deserialize(inner_bytes)
-            .map_err(|e| DeserializeError::CorruptedBinary(format!("{:?}", e)))?;
+        let metadata_len = leb128::read::unsigned(&mut inner_bytes).map_err(|_e| {
+            DeserializeError::CorruptedBinary("Can't read metadata size".to_string())
+        })?;
+        let metadata_slice: &[u8] =
+            std::slice::from_raw_parts(&bytes[32] as *const u8, metadata_len as usize);
```
MarkMcCaskey, line 157: For maintainability it might be a good idea to use named constants or a struct with a bunch of `const` methods on it so that we update all these numbers correctly when we need to. Last time I reviewed this code, I don't believe this was 32.

lib/engine-jit/src/artifact.rs
```diff
@@ -141,21 +137,26 @@ impl JITArtifact {
...
             return Err(DeserializeError::Incompatible(
                 "The provided bytes are not wasmer-jit".to_string(),
             ));
         }
 
-        let inner_bytes = &bytes[Self::MAGIC_HEADER.len()..];
-
-        // let r = flexbuffers::Reader::get_root(bytes).map_err(|e| DeserializeError::CorruptedBinary(format!("{:?}", e)))?;
-        // let serializable = SerializableModule::deserialize(r).map_err(|e| DeserializeError::CorruptedBinary(format!("{:?}", e)))?;
+        let mut inner_bytes = &bytes[16..];
```
MarkMcCaskey, line 151: Goes for here too, and just in general... This can be done as a follow up to this PR if you really want to ship it, but for the long term health of this code, using named constants can be really helpful

lib/engine-jit/src/artifact.rs
```diff
@@ -325,15 +326,34 @@ impl Artifact for JITArtifact {
...
-        // Ok(s.take_buffer())
-        let bytes = bincode::serialize(&self.serializable)
-            .map_err(|e| SerializeError::Generic(format!("{:?}", e)))?;
-
         // Prepend the header.
         let mut serialized = Self::MAGIC_HEADER.to_vec();
-        serialized.extend(bytes);
+
+        serialized.resize(32, 0);
+        let mut writable_leb = &mut serialized[16..];
```
MarkMcCaskey, line 333: And those same named constants could be used here too

lib/engine-jit/src/artifact.rs
```diff
@@ -325,15 +326,34 @@ impl Artifact for JITArtifact {
...
+        assert_eq!(offset, 32);
+
         Ok(serialized)
     }
 }
+
+/// It pads the data with the desired alignment
+pub fn pad_and_extend(prev_data: &mut Vec<u8>, data: &[u8], align: u64) -> u64 {
+    // We assert that align is a power of 2
+    debug_assert_eq!(align & (align - 1), 0);
```
MarkMcCaskey, line 350: Luckily Rust has you covered [`u64::is_power_of_two`](https://doc.rust-lang.org/std/primitive.u64.html#method.is_power_of_two):

⛳ 
```Rust
debug_assert!(align.is_power_of_two());
```

Alternatively, if you know that `std::mem::align_of` always returns a power of two, rather than taking `align` as a runtime parameter, you can take the type as a generic.

So you'd update the function to look like:

```Rust
pub fn pad_and_extend<T>(prev_data: &mut Vec<u8>, data: &[u8]) -> u64 {
   let align = std::mem::align_of::<T>() as u64;
   ...
}
```

Which would then prevent the function from ever being called incorrectly in this way. (you could obviously pass the wrong type and get the wrong align but 🤷 )

You may then need to change the caller to explicitly fill the generic parameter, so like:
```Rust
let offset = pad_and_extend::<SerializableModule>(&mut serialized, &serialized_data);
```

lib/engine-jit/src/artifact.rs
MarkMcCaskey: Because it's referring to an offset in memory it doesn't matter. On a 32bit system you only have 32 bits of addressable memory. A `usize` is defined to be the size big enough to cover all addressable memory, so when dealing with offsets it's a natural size. It makes sense to cast it to a u64 before writing it if you want it to take up a constant 8 bytes on all platforms, but when dealing with pointers / offsets, a `usize` will always be good enough and do the right thing on all platforms.

lib/engine-jit/src/artifact.rs
```diff
@@ -325,15 +326,34 @@ impl Artifact for JITArtifact {
...
+    let align = align as usize;
+    let mut offset = prev_data.len();
+    if offset & (align - 1) != 0 {
+        offset += align - (offset & (align - 1));
+        prev_data.resize(offset, 0);
+    }
+    prev_data.extend(data);
+    offset as u64
+}
```
MarkMcCaskey, line 359: This function in general is a great function for unit tests, it's just way too easy to have subtle bugs with logic like this.

lib/engine-jit/src/serialize.rs
```diff
@@ -38,9 +32,76 @@ pub struct SerializableCompilation {
...
+    ) -> Result<&'a ArchivedSerializableModule, DeserializeError> {
+        if metadata_slice.len() < 9 {
+            return Err(DeserializeError::Incompatible(
+                "invalid serialized data".into(),
+            ));
+        }
+        let mut pos: [u8; 8] = Default::default();
+        let endian = metadata_slice[metadata_slice.len() - 1];
+        if (cfg!(target_endian = "big") && endian == b'l')
+            || (cfg!(target_endian = "little") && endian == b'b')
```
MarkMcCaskey, line 88: Be aware that this logic may not do exactly what you want in degenerate cases.

Having the byte `b'j'` and `target_endian = "big"` is a success case here. (i.e. it only checks for well-defined failure: assuming only 2 possible bytes without validating it)

lib/vm/src/libcalls.rs
```diff
@@ -35,13 +35,17 @@
 //!   }
 //!   ```
 
+#![allow(missing_docs)] // For some reason lint fails saying that `LibCall` is not documented, when it actually is
```
MarkMcCaskey, line 38: If I had to guess, it's one of the derive macros. Derive macros can edit code streams... I don't know if they can add variants to an enum (probably not) but they can definitely make new `impl YourType { fn my_function() {  ... } }` blocks which could be failing the doc thing.

I think we shouldn't be asserting docs statically, but as part of a CI lint step in any case but until then, the narrower the `allow` directive is, the better.

MarkMcCaskey (commented):
Looks good! Just a few minor things, but glad things are working

lib/engine-jit/src/serialize.rs
MarkMcCaskey: I don't know about that. What the `unsafe` keyword means is "this function has invariants that cannot be checked by the compiler, you must manually ensure that they're true or the safety model of Rust breaks down".

An `unsafe` function without documentation explaining how it can be used is kind of nonsense (I suppose it means "you can't ever call this function"), even if it's not public.

lib/engine-jit/src/artifact.rs
djkoloski: If you do need validation capabilities, rkyv supports archive validation with [`check_archived_value`](https://docs.rs/rkyv/0.6.1/rkyv/validation/fn.check_archived_value.html). You'll incur a performance hit (see [benches](https://github.com/djkoloski/rust_serialization_benchmark) for details) but it's optional so you could put it behind a command line argument.

lib/engine-jit/src/artifact.rs
syrusakbary: Love this!

lib/engine-jit/src/artifact.rs
syrusakbary: Added!

lib/engine-jit/src/artifact.rs
syrusakbary: Fixed!

lib/engine-jit/src/artifact.rs
syrusakbary: Fixed!

lib/vm/src/libcalls.rs
syrusakbary: I'll keep it as it is now as I don't have any other way of solving it for now :(

lib/engine-jit/src/serialize.rs
syrusakbary: Fixed!

lib/engine-jit/src/artifact.rs
syrusakbary: Done!

lib/engine-jit/src/artifact.rs
syrusakbary: Fixed!

lib/engine-jit/src/artifact.rs
syrusakbary: Fixed!

syrusakbary: I tried to add optional validation, but the scope was bigger than I envisioned (bytecheck dependency, adding an attr to almost all structs, ...), so it will be better to do it in a PR apart

lib/engine-jit/src/artifact.rs
```diff
@@ -325,15 +331,51 @@ impl Artifact for JITArtifact {
...
+mod tests {
+    use super::pad_and_extend;
+
+    #[test]
+    fn test_pad_and_extend() {
+        let mut data: Vec<u8> = vec![];
+        let offset = pad_and_extend::<i64>(&mut data, &[1, 0, 0, 0, 0, 0, 0, 0]);
+        assert_eq!(offset, 0);
+        let offset = pad_and_extend::<i32>(&mut data, &[2, 0, 0, 0]);
+        assert_eq!(offset, 8);
```
MarkMcCaskey, line 373: I'm a bit confused why this is 8 instead of 0

lib/engine-jit/src/artifact.rs
```diff
@@ -325,15 +331,51 @@ impl Artifact for JITArtifact {
...
+        let offset = pad_and_extend::<i64>(&mut data, &[3, 0, 0, 0, 0, 0, 0, 0]);
+        assert_eq!(offset, 16);
```
MarkMcCaskey, line 375: Similarly, I'd expect this to be 8, not 16

lib/engine-jit/src/serialize.rs
```diff
@@ -38,9 +31,91 @@ pub struct SerializableCompilation {
...
+    /// from memory.
+    /// Right now we are not doing any extra work for validation, but
+    /// `rkyv` has an option to do bytecheck on the serialized data before
+    /// serializing (via `rkyv::check_archived_value`).
+    pub unsafe fn deserialize(metadata_slice: &[u8]) -> Result<Self, DeserializeError> {
+        let archived = Self::archive_from_slice(metadata_slice)?;
+        Self::deserialize_from_archive(archived)
+    }
+
+    unsafe fn archive_from_slice<'a>(
```
MarkMcCaskey, line 87: Still need a safety section on this (see https://github.com/wasmerio/wasmer/pull/2250/files#r623939328)

lib/engine-jit/src/artifact.rs
MarkMcCaskey: Oh nevermind, I misunderstood, I got confused between the slice after `data` and `data` itself. This looks correct 👍

lib/engine-jit/src/serialize.rs
syrusakbary: Fixed!

lib/engine-jit/Cargo.toml
Hywan: :+1:

PR #2257, opened 2021-04-24
Added a comment to WasmPtr about structs.

Earthmark:
# Description
This adds a comment about how to use compound c-like structs using WasmPtr.

# Review
This is possibly too small for a changelog entry.

Example tests ensuring this behavior should possibly be added.

I'm also curious as to if this is correct, cause it seems to be working in a side project but I'm unsure as to if it's a good idea!


syrusakbary:
Thanks for the PR @Earthmark! It looks great, let's wait to see what @MarkMcCaskey thinks!

lib/api/src/ptr.rs
```diff
@@ -32,6 +32,31 @@ pub struct Item;
 ///     derefed_ptr.set(inner_val + 1);
 /// }
 /// ```
+/// 
+/// This type can also be used with primitive-filled structs, but be careful of
+/// guarentees required by ValueType.
```
Hywan, line 37:
```suggestion
/// guarantees required by `ValueType`.
```

MarkMcCaskey (approved):
looks good to me

Earthmark:
I was not expecting it linted the comments in that way, I'll resolve this later tonight.

Earthmark:
I'm not sure what I need to do to get it going, I don't think the linux build was related.

Earthmark:
Well, I didn't realize it actually ran the comment code, I'll get that fixed.

Earthmark:
wasn't sure if that was a test pass or not, either way the compiler bug should be resolved now. I'm trying to validate this locally but it follows the same conventions as the comment right above.

Earthmark:
I thought I got a pass on this locally, I understand why it failed but I think I had the wrong command locally. I'll figure out why my local didn't fall then push the fix.

Earthmark:
Alright, now I'm aware of the `cargo test --doc` command, and I will certainly use this in other projects. Thank y'all for your patience.

syrusakbary:
No worries at all. Let’s try again!

bors d+

Issue #2187, opened 2021-03-15
can't represent 65536 pages


nlewycky:
### Describe the bug

Memory minimum and maximum sizes range 0..65536 inclusive. The verifier blocks 65537 and higher. However, wasmer only works on 65535 and lower.

Testcase:
```
(module (memory (;0;) 65536))
```
fails:
```
$ target/release/wasmer --cranelift x.wat 
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: TryFromIntError(())', lib/vm/src/memory.rs:262:74
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The line of code in question is:
```
        let mem_length = memory.minimum.bytes().0.try_into().unwrap();
```
and the error is that 65536*65536 is 4294967296, which is the first value too large to fit in a `u32`.

Wasmer needs to either remove the problematic conversions to bytes keeping the value as pages, use a larger datatype for the byte count, or store it in a 'max-1' format (assuming '0' is unnecessary since we store the absence of a memory differently.)

PR #2258, opened 2021-04-25
Allow 65536 pages

k-nasa:
# Description

fix: https://github.com/wasmerio/wasmer/issues/2187

> Memory minimum and maximum sizes range 0..65536 inclusive. The verifier blocks 65537 and higher. However, wasmer only works on 65535 and lower.

I fixed this issue.

# Review


k-nasa:
I would like to add a test to detect this issue. But I don't understand structure of tests in this repository so I don't know where to add it. Can you give me a signpost?

nlewycky:
Thank you for the patch! A `.wast` type test in `tests/wast/wasmer` should do it.

lib/vm/src/memory.rs
```diff
@@ -259,6 +259,7 @@ impl LinearMemory {
         let mapped_pages = memory.minimum;
         let mapped_bytes = mapped_pages.bytes();
 
+        println!("{}", minimum_bytes);
```
syrusakbary, line 262: Delete debug print

k-nasa: 🙈  I fixed it. 755639c

k-nasa:
Thanks @nlewycky! I was able to add test.

lib/types/src/units.rs
```diff
@@ -91,8 +91,8 @@ impl From<usize> for Bytes {
     }
 }
 
-impl From<u32> for Bytes {
-    fn from(other: u32) -> Self {
+impl From<u64> for Bytes {
+    fn from(other: u64) -> Self {
```
Hywan, line 95: Please keep both implementations, otherwise it's a regression :-).

lib/vm/src/vmcontext.rs
```diff
@@ -340,7 +340,7 @@ pub struct VMMemoryDefinition {
     pub base: *mut u8,
 
     /// The current logical size of this linear memory in bytes.
-    pub current_length: u32,
+    pub current_length: u64,
```
Hywan, line 343: I wonder if we should not create a type alias to represent the memory size, something like `pub type MemorySize = u64`. Thoughts @nlewycky, @MarkMcCaskey, @syrusakbary?

nlewycky: I don't think so in this case, we aren't going to be changing this type or confusing it with other values much.

nlewycky (commented):
This:
https://sourcegraph.com/github.com/wasmerio/wasmer/-/blob/lib/compiler-llvm/src/translator/intrinsics.rs#L620:13
also need to be updated from `i32` to `i64` or else we'll compute the wrong value, and that may require some further changes (llvm is a strongly-typed IR and so any place it's used with another i32 will fail the verifier).

Similarly singlepass loads a 32-bit value into a register here:
https://sourcegraph.com/github.com/wasmerio/wasmer/-/blob/lib/compiler-singlepass/src/codegen_x64.rs#L1280

I'm not sure what the equivalent is in cranelift, possibly the type of the global_value in make_heap in src/func_environ.rs (not src/translator/func_environ.rs)?

Amanieu:
Superseded by #2677

PR #2262, opened 2021-04-26
Make parallelism optional for singlepass backend

### LINKED ISSUE #2261: Consider making `rayon` dependency of singlepass backend optional
### Motivation

At near, we are using the singlepass backend because we care about determinism and predictability. In particular, we generally avoid using thread *parallelism*, to avoid situations where two *concurrent* bits of functionality decide to take advantage of all cores at the same time and starve each other for CPU. So we need to be able to run wasmer with just a single thread. 

We can already do this by configuring rayon's global theradpool, but that's not convenient (you need custom setup code for each 
 `main` and `#[test]`), and feels a bit roundabout. Even with `RAYON_NUM_THREADS` there's still a (degenerate) thread pool, and unnecessary inter-thread situation

### Proposed solution

I took a quick look at the code, and it seems that making `rayon` optional should be fairly straightforward. Something like:

```toml
[dependencies]
rayon = { version="1.0", optional = true}

[features]
default = ["rayon"]
```

```rust
let iter = some().itereator().chain().
#[cfg(feature = "rayon")]
let iter = iter.par_iter();
```

There's even https://github.com/cuviper/rayon-cond, but I am not sure it is worth it. 

### Alternatives

Controlling this at runtime via `RAYON_NUM_THREADS` or `rayon::ThreadPoolBuilder::build_global` is already possible, just inconvenient

matklad:
# Description

closes #2261

# Review

### DIFF
--- a/lib/compiler-singlepass/Cargo.toml
+++ b/lib/compiler-singlepass/Cargo.toml
@@ -15,7 +15,7 @@ edition = "2018"
 wasmer-compiler = { path = "../compiler", version = "1.0.2", features = ["translator"], default-features = false }
 wasmer-vm = { path = "../vm", version = "1.0.2" }
 wasmer-types = { path = "../types", version = "1.0.2", default-features = false, features = ["std"] }
-rayon = "1.5"
+rayon = { version = "1.5", optional = true }
 hashbrown = { version = "0.9", optional = true }
 serde = { version = "1.0", features = ["derive"] }
 more-asserts = "0.2"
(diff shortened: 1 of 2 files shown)


matklad:
Something I haven't thought about before writing the code: should this be a compile time or runtime feature? I just brute-codded compile-time, I guess we can add runtime feature if there's a need for that as well, and, even with a runtime feature, ability to not compile rayon is a plus. 

In other words, the current solution is what works today, and it's forward compatible with the potentially more general future soln.

syrusakbary:
This looks great! It should be good to merge, before then can we make this strategy also available in the other two compilers so they all have the same feature?

Thanks!

matklad:
>before then can we make this strategy also available in the other two compilers so they all have the same feature?

I'd rather keep this PR scoped to singlepass backend, as that's my immediate need (and I feel that singlethreadedness makes the most sense for the singlepass) :-)

nlewycky:
> (and I feel that singlethreadedness makes the most sense for the singlepass) :-)

FWIW, I work on all three compilers and I keep a local patch around to make them single-threaded because otherwise I can't debug a bug in any of them. My patch is much worse than yours though, I was adding a Mutex inside the parallel part. Just throwing in my 2¢ that it makes sense for the other compilers too.

syrusakbary (approved):

syrusakbary:
Merging manually, tests are passing

PR #2264, opened 2021-04-27
Middleware refactor

syrusakbary:
# Description
In order to be able to refactor the LLVM experimental compiler, receiving a callback instead of a serialized module metadata, we need to make compilers act on non-mutable input.
This is something that compilers should do, so this PR moves towards the scenario of pure compilation.

# Review

### DIFF
--- a/lib/compiler-cranelift/src/compiler.rs
+++ b/lib/compiler-cranelift/src/compiler.rs
@@ -53,8 +53,8 @@ impl CraneliftCompiler {
 
 impl Compiler for CraneliftCompiler {
     /// Get the middlewares for this compiler
-    fn get_middlewares(&self) -> Vec<Arc<dyn ModuleMiddleware>> {
-        self.config.get_middlewares()
+    fn get_middlewares(&self) -> &[Arc<dyn ModuleMiddleware>] {
+        &self.config.middlewares
     }
 
     /// Compile the module using Cranelift, producing a compilation result with
--- a/lib/compiler-cranelift/src/compiler.rs
+++ b/lib/compiler-cranelift/src/compiler.rs
@@ -25,7 +25,8 @@ use wasmer_compiler::CompileError;
 use wasmer_compiler::{CallingConvention, ModuleTranslationState, Target};
 use wasmer_compiler::{
     Compilation, CompileModuleInfo, CompiledFunction, CompiledFunctionFrameInfo,
-    CompiledFunctionUnwindInfo, Compiler, Dwarf, FunctionBody, FunctionBodyData, ModuleMiddleware,
+    CompiledFunctionUnwindInfo, Compiler, Dwarf, FunctionBinaryReader, FunctionBody,
+    FunctionBodyData, MiddlewareBinaryReader, ModuleMiddleware, ModuleMiddlewareChain,
     SectionIndex,
 };
 use wasmer_types::entity::{EntityRef, PrimaryMap};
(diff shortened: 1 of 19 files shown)


Hywan:
I like the direction of this PR. Is it ready for a review?

syrusakbary:
@Hywan yup, it's ready to review!

bors try

syrusakbary:
bors try-

lib/compiler-cranelift/src/config.rs
```diff
@@ -186,6 +186,11 @@ impl Cranelift {
 
         settings::Flags::new(flags)
     }
+
+    /// Get Middlewares
+    pub fn get_middlewares(&self) -> Vec<Arc<dyn ModuleMiddleware>> {
+        self.middlewares.clone()
+    }
```
MarkMcCaskey, line 193: Consider using `&[Arc<dyn ModuleMiddleware>]` it won't clone every `Arc` or allocate a new container when this is called. Callers can then get vecs of it with `.to_vec()`.

The primary reason to use `Vec` over `&[]` in a trait method is that you don't want to impose on the implementors that they should actually be storing the data that they're returning. `Vec` lets you just build it and return it, `&[]` requires that the data actually lives on the thing the method is being called on (and that the data is stored in such a way that you can return a `&[]` to it, i.e. linear storage like a Vec or an array).

lib/compiler-cranelift/src/compiler.rs
```diff
@@ -118,15 +121,20 @@ impl Compiler for CraneliftCompiler {
                 // if generate_debug_info {
                 //     context.func.collect_debug_info();
                 // }
+                let mut reader =
+                    MiddlewareBinaryReader::new_with_offset(input.data, input.module_offset);
+                reader.set_middleware_chain(
+                    self.config
+                        .middlewares
+                        .generate_function_middleware_chain(*i),
```
MarkMcCaskey, line 129: I haven't seen the whole PR yet but just looking at this, I'm not quite sure what this does. If I had to guess, it's compiler codegen implemented as a middleware

lib/engine-jit/src/artifact.rs
```diff
@@ -84,10 +84,15 @@ impl JITArtifact {
 
         let compiler = inner_jit.compiler()?;
 
+        let mut module = (*compile_info.module).clone();
+        let middlewares = compiler.get_middlewares();
+        middlewares.apply_on_module_info(&mut module);
```
nlewycky, line 89: This `module` is a clone of the one in `translation`. That means that if the middleware introduces different memory or table styles, our maps `memory_styles` and `table_styles` will be out of date. I suppose the only way to do this today would be for the middleware to add a memory to a module that doesn't have one, or remove memory from one that does. Either way, maybe we should apply the middleware after `let translation = ` above, before we use `translation.module`?

MarkMcCaskey (approved):
I don't have enough context on middleware itself to have much of a strong opinion about this change, it looks reasonable to me though.

All the Rust stuff looks good, though it might be better to change from `Vec` to `&[]` if it makes sense to.

lib/engine-jit/src/artifact.rs
syrusakbary: I'm just moving the logic, but your comment is completely on point (it was an issue also in previous code).
I'll try to move it up!

syrusakbary: Fixed it!

lib/compiler-cranelift/src/config.rs
syrusakbary: Fixed!

PR #2266, opened 2021-04-27
Disable aarch64 linux in CI for now

MarkMcCaskey:
We lost access to the servers and it's blocking CI, disabling so CI works while we fix it


syrusakbary (approved):

syrusakbary:
Everything passed, so merging manually

PR #2268, opened 2021-04-28
Update spectests

MarkMcCaskey:
Updating spectests in Wasmer, we disable all the ones that are now failing. We shouldn't merge this until we investigate the breakages and fix some of them. It's likely that we're passing most of the statements in most of these files, but due to some new changes are not passing all of them, thus merging it in as-is will reduce our overall test coverage.

Additionally, when running the update script, some feature specific things failed to update like bulk memory. But now that bulk memory has landed as a stable feature of Wasm it should be included in the general `spec` tests.


lib/vm/src/instance/mod.rs
```diff
@@ -1300,21 +1298,6 @@ cfg_if::cfg_if! {
...
 
-fn check_table_init_bounds(instance: &Instance) -> Result<(), Trap> {
-    let module = Arc::clone(&instance.module);
-    for init in &module.table_initializers {
-        let start = get_table_init_start(init, instance);
-        let table = instance.get_table(init.table_index);
-
-        let size = usize::try_from(table.size()).unwrap();
-        if size < start + init.elements.len() {
-            return Err(Trap::new_from_runtime(TrapCode::TableSetterOutOfBounds));
```
syrusakbary, line 1311: Is `TableSetterOutOfBounds` being used somewhere else? If not, can we remove it from the Enum?

lib/vm/src/instance/mod.rs
```diff
@@ -1351,23 +1334,6 @@ unsafe fn get_memory_slice<'instance>(
...
-fn check_memory_init_bounds(
-    instance: &Instance,
-    data_initializers: &[DataInitializer<'_>],
-) -> Result<(), Trap> {
-    for init in data_initializers {
-        let start = get_memory_init_start(init, instance);
-        unsafe {
-            let mem_slice = get_memory_slice(init, instance);
-            if mem_slice.get_mut(start..start + init.data.len()).is_none() {
-                return Err(Trap::new_from_runtime(TrapCode::HeapSetterOutOfBounds));
```
syrusakbary, line 1363: Is `HeapSetterOutOfBounds ` being used somewhere else? If not, can we remove it from the Enum?

lib/vm/src/instance/mod.rs
MarkMcCaskey: Yeah, it's still in use! Good call though. I'm assuming I'll get warnings for unused variants / code (though actually not if it's public...). I don't currently see any warnings though

lib/vm/src/instance/mod.rs
MarkMcCaskey: This looks like it might not be used anywhere else... I'll revisit this before shipping, good call!

syrusakbary (approved):

lib/vm/src/instance/mod.rs
MarkMcCaskey: Okay, so I'm going to keep this for now because while it currently is not used at all, judging by the fact that the other one (for tables) is used in `Table::copy`, I suspect that during the implementation of bulk memory we may need it again. And if so it'll be easier to keep it than remove it and readd it. If it's not needed for bulk memory, then we can actually delete it though!

syrusakbary: Ok, can you add a TODO over those enums mentioning that precisely? (that way will be easier to remove if we forgot)

syrusakbary (approved):

lib/vm/src/instance/mod.rs
MarkMcCaskey: good idea! done!

PR #2269, opened 2021-04-28
Add `--tests` flag to core test commands

MarkMcCaskey:
This stops the behavior of linking every example every time we compile tests.
This saves a non-negligible amount of time.

This has some side effects like not testing benchmarks as tests, but we are not affected by that due to our project layout. So this change should be a no-op that speeds up the compilation part of tests. The speedup should be visible in CI, I'll run `bors try` to compare.


MarkMcCaskey:
bors try

nlewycky (approved):
Looks like the examples are covered by `test-examples` instead.

PR #2272, opened 2021-04-29
Experimental Native Compilation Refactor

syrusakbary:
# Description

This PR refactors the experimental native compilation, making it much more resilient.
* It speeds up LLVM Native compilation by bypassing the single-threaded linking via LLVM, and emitting directly mutliple object files (one per module). Therefore we speed it up via multi-threaded linking in via the `lld` linker (we can embed it in the future)
* It refactors the output of the experimental native compilation to be just a struct with two fields:
  1. A list of object file contents
  2. The compilation frame info for each of the functions (containing the function length)
* It parses the function size per module and outputs it in the frame info.

This PR is also designed in mind to make much easier the creation and parsing of SourceMaps that we can use later to detect traps in an exact location

# Review


lib/compiler-llvm/src/compiler.rs
```diff
@@ -80,110 +77,95 @@ impl LLVMCompiler {
...
         );
-        let metadata_gv =
-            merged_module.add_global(metadata_init.get_type(), None, "WASMER_METADATA");
-        metadata_gv.set_initializer(&metadata_init);
-        metadata_gv.set_linkage(Linkage::DLLExport);
-        metadata_gv.set_dll_storage_class(DLLStorageClass::Export);
-
-        if self.config().enable_verifier {
-            merged_module.verify().unwrap();
-        }
```
syrusakbary, line 177: This was not needed any longer since we already do verification on `translate_to_module`

lib/engine-object-file/src/artifact.rs
```diff
@@ -488,6 +506,11 @@ impl Artifact for ObjectFileArtifact {
 
     /// Serialize a ObjectFileArtifact
     fn serialize(&self) -> Result<Vec<u8>, SerializeError> {
-        Ok(self.module_bytes.clone())
+        unimplemented!();
+        // let mut builder = ar::Builder::new(Vec::new());
+        // let header = Header::new(format!(b"baz.txt").to_vec(), file_size)
+        // builder.append(&header, file_content).unwrap();
+
+        // Ok(self.module_bytes.clone())
```
syrusakbary, line 514: The idea here is to create an archive file with all the object files. I'll wait to see @nlewycky thoughts on this.
I think llvm-ld can link properly archive files (also with `--whole-archive` option, if needed), so I think this would solve the limit file issue when Linking, as well as the CMD call size < 128Kb

lib/engine-native/src/serialize.rs
```diff
@@ -33,32 +37,22 @@ pub struct ModuleMetadata {
...
-    pub fn get_symbol_registry<'a>(&'a self) -> ModuleMetadataSymbolRegistry<'a> {
+    pub fn get_symbol_registry(&self) -> ModuleMetadataSymbolRegistry {
         ModuleMetadataSymbolRegistry {
-            prefix: &self.prefix,
+            prefix: self.prefix.clone(),
         }
     }
 
-    pub fn serialize(&mut self) -> Result<Vec<u8>, CompileError> {
+    pub fn serialize(&self) -> Result<Vec<u8>, CompileError> {
```
syrusakbary, line 55: The serializer should never be mutable. This fixes that

lib/engine-native/src/serialize.rs
```diff
@@ -33,32 +37,22 @@ pub struct ModuleMetadata {
     pub compile_info: CompileModuleInfo,
     pub prefix: String,
     pub data_initializers: Box<[OwnedDataInitializer]>,
-    // The function body lengths (used to find function by address)
-    pub function_body_lengths: PrimaryMap<LocalFunctionIndex, u64>,
+    pub frame_infos: PrimaryMap<LocalFunctionIndex, CompiledFunctionFrameInfo>,
```
syrusakbary, line 40: This is how we are going to be able to move traps forward!

lib/engine-object-file/src/artifact.rs
nlewycky: In the future we should link `lld` in directly instead of running it as a subprocess which will also let us bypass any file descriptor and command line limits by simply not using command lines or files.

lib/compiler-llvm/src/compiler.rs
```diff
@@ -80,110 +77,95 @@ impl LLVMCompiler {
...
+        let function_bodies_and_frame_infos = function_body_inputs
+            .into_iter()
+            .par_bridge()
+            .map_init(
+                || {
+                    let target_machine = self.config().target_machine(target);
+                    FuncTranslator::new(target_machine)
+                },
+                |func_translator, (i, input)| {
+                    // let symbol = Symbol::LocalFunction(i);
```
nlewycky, line 92:
```suggestion
```

lib/compiler/src/compiler.rs
```diff
@@ -16,6 +17,17 @@ use wasmer_types::entity::PrimaryMap;
...
+/// The output of the experimental native compilation.
+/// It returns the object files that the linker will link together after
+/// along with the frame infos necessary for fully traps support.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ExperimentalNativeCompilation {
+    /// The kind of emitted object file
+    pub object_files: Vec<Vec<u8>>,
+    /// The frame infos of the files
+    pub frame_infos: PrimaryMap<LocalFunctionIndex, CompiledFunctionFrameInfo>,
+}
```
nlewycky, line 29: How is the frame info not part of the object file??

lib/compiler-llvm/src/compiler.rs
nlewycky: No action required. It was there to check the module after changing the IR with `merged_module.add_global`.

lib/compiler/src/compiler.rs
```diff
@@ -16,6 +17,17 @@ use wasmer_types::entity::PrimaryMap;
 use wasmer_types::{Features, FunctionIndex, LocalFunctionIndex, SignatureIndex};
 use wasmparser::{Validator, WasmFeatures};
 
+/// The output of the experimental native compilation.
+/// It returns the object files that the linker will link together after
+/// along with the frame infos necessary for fully traps support.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ExperimentalNativeCompilation {
+    /// The kind of emitted object file
```
nlewycky, line 25:
```suggestion
    /// The contents of emitted object files
```

lib/engine-native/src/artifact.rs
```diff
@@ -311,8 +332,9 @@ impl NativeArtifact {
         );
 
         let linker = engine_inner.linker().executable();
-        let output = Command::new(linker)
-            .arg(&filepath)
+        let mut command = Command::new(linker);
+        let output = command
+            .args(&object_filepaths)
```
nlewycky, line 337: Why this change?

lib/engine-native/src/serialize.rs
```diff
@@ -33,32 +37,22 @@ pub struct ModuleMetadata {
...
-            prefix: &self.prefix,
-        };
-        (compile_info, symbol_registry)
-    }
-
-    pub fn get_symbol_registry<'a>(&'a self) -> ModuleMetadataSymbolRegistry<'a> {
+    pub fn get_symbol_registry(&self) -> ModuleMetadataSymbolRegistry {
         ModuleMetadataSymbolRegistry {
-            prefix: &self.prefix,
+            prefix: self.prefix.clone(),
```
nlewycky, line 51: Could we not clone this? Can we return a `&ModuleMetadataSymbolRegistry` instead?

lib/engine-object-file/src/serialize.rs
```diff
@@ -20,19 +46,55 @@ pub struct ModuleMetadataSymbolRegistry {
...
+        metadata_slice: &'a [u8],
+    ) -> Result<&'a ArchivedModuleMetadata, DeserializeError> {
+        let mut pos: [u8; 8] = Default::default();
+        let endian = metadata_slice[metadata_slice.len() - 1];
+        if (cfg!(target_endian = "big") && endian == b'l')
+            || (cfg!(target_endian = "little") && endian == b'b')
+        {
+            return Err(DeserializeError::Incompatible("incompatible endian".into()));
+        }
+        pos.copy_from_slice(&metadata_slice[metadata_slice.len() - 9..metadata_slice.len() - 1]);
```
nlewycky, line 83: The last eight bytes? What if `metadata_slice` is too short?

lib/compiler/src/compiler.rs
syrusakbary: This frame info’s are Wasmer ones, where we keep track of everything.
They live in the CompiledModule info which live in the metadata (which is a proprietary format of Wasmer). That’s why it has to live here and not the object file

lib/engine-native/src/serialize.rs
syrusakbary: Cloning it makes things easier to reason about, and the code simpler. I think is a bit better tradeoff?

lib/engine-native/src/artifact.rs
syrusakbary: Fixed!

lib/cli/src/commands/create_exe.rs
```diff
@@ -284,23 +284,42 @@ impl LinkCode {
...
             .additional_libraries
             .iter()
             .map(|lib| format!("-l{}", lib));
-        let command = command.args(link_aganist_extra_libs);
-        let output = command.arg("-o").arg(&self.output_path).output()?;
+        let command = command
+            .args(link_aganist_extra_libs)
+            .arg("-o")
+            .arg(&self.output_path);
+        println!("COMMAND: {:?}", command);
```
syrusakbary, line 321: TODO: Remove

lib/engine-object-file/src/serialize.rs
syrusakbary: Fixed!

lib/engine-object-file/src/artifact.rs
```diff
@@ -198,74 +197,94 @@ impl ObjectFileArtifact {
...
+                    &target,
+                    &metadata.compile_info,
+                    module_translation.as_ref().unwrap(),
+                    function_body_inputs,
+                )?;
+                let mut obj = get_object_for_target(&target_triple).map_err(to_compile_error)?;
+                let compiled_function_infos = compilation.get_frame_info();
+                // .values()
+                // .map(|frame_info| SerializableFunctionFrameInfo::Processed(frame_info.clone()))
+                // .collect::<PrimaryMap<LocalFunctionIndex, _>>();
```
nlewycky, line 257:
```suggestion
```

lib/engine-native/src/serialize.rs
```diff
@@ -33,58 +36,84 @@ pub struct ModuleMetadata {
...
+    /// serializing (via `rkyv::check_archived_value`).
     pub unsafe fn deserialize(metadata_slice: &[u8]) -> Result<Self, DeserializeError> {
         let archived = Self::archive_from_slice(metadata_slice)?;
         Self::deserialize_from_archive(archived)
     }
 
+    /// # Safety
+    ///
+    /// This method is unsafe.
+    /// Please check `ModuleMetadata::deserialize` for more details.
```
nlewycky, line 98: "The `metadata_slice` must be an archive produced by ModuleMetadata::serialize"?

lib/engine-native/src/serialize.rs
```diff
@@ -33,58 +36,84 @@ pub struct ModuleMetadata {
...
         ModuleMetadataSymbolRegistry {
-            prefix: &self.prefix,
+            prefix: self.prefix.clone(),
         }
     }
 
-    pub fn serialize(&mut self) -> Result<Vec<u8>, CompileError> {
+    /// Serialize the Metadata into bytes
+    /// The bytes will have the following format:
+    /// RKYV serialization (any length) + POS (8 bytes) + Endian (1 byte)
```
nlewycky, line 67: POS?

lib/engine-native/src/serialize.rs
```diff
@@ -33,58 +36,84 @@ pub struct ModuleMetadata {
...
-        };
-        (compile_info, symbol_registry)
+cfg_if::cfg_if! {
+    if #[cfg(target_endian = "big")] {
+        const HOST_ENDIAN: u8 = b'b';
+    } else if #[cfg(target_endian = "little")] {
+        const HOST_ENDIAN: u8 = b'l';
     }
+    else {
+        compile_error!("Endian not supported by the host");
```
nlewycky, line 54:
```suggestion
        compile_error!("Endian isn't big or little?");
```

lib/engine-object-file/src/serialize.rs
```diff
@@ -1,17 +1,53 @@
...
 use wasmer_types::{FunctionIndex, LocalFunctionIndex, OwnedDataInitializer, SignatureIndex};
 
+cfg_if::cfg_if! {
+    if #[cfg(target_endian = "big")] {
+        const HOST_ENDIAN: u8 = b'b';
+    } else if #[cfg(target_endian = "little")] {
+        const HOST_ENDIAN: u8 = b'l';
+    }
+    else {
+        compile_error!("Endian not supported by the host");
```
nlewycky, line 25: Wish we didn't have a copy of this from native.

lib/compiler/src/compiler.rs
```diff
@@ -16,6 +17,17 @@ use wasmer_types::entity::PrimaryMap;
 use wasmer_types::{Features, FunctionIndex, LocalFunctionIndex, SignatureIndex};
 use wasmparser::{Validator, WasmFeatures};
 
+/// The output of the experimental native compilation.
+/// It returns the object files that the linker will link together after
+/// along with the frame infos necessary for fully traps support.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ExperimentalNativeCompilation {
+    /// The contents of emitted object files
+    pub object_files: Vec<Vec<u8>>,
```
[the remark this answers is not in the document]
syrusakbary: Yeah, the order is deterministic as the current implementation (per llvm-compilation), but IMHO should not be required since it's just an implementation detail?

lib/engine-native/src/serialize.rs
syrusakbary: Yeah, that's better!

lib/compiler-llvm/src/object_file.rs
```diff
@@ -37,12 +37,61 @@ fn map_goblin_err(error: goblin::error::Error) -> CompileError {
...
     pub custom_sections: CustomSections,
     pub eh_frame_section_indices: Vec<SectionIndex>,
 }
 
+pub fn get_function_body_size(data: &[u8]) -> Result<u64, CompileError> {
+    use object::Object;
+    use object::ObjectSection;
+
+    let in_object = object::File::parse(&data).map_err(map_object_err)?;
+    let section = in_object.section_by_name("wasmer_function");
```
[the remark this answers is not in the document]
syrusakbary: I tried. `FUNCTION_SECTION` includes the `__TEXT`, which we don't need here. So the only viable option is to create another variable next to it (which I'm about to push now)

lib/compiler/src/compiler.rs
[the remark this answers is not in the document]
nlewycky: Please comment that the order is irrelevant but deterministic. It helps the next person understand this data.

lib/engine-native/src/serialize.rs
```diff
@@ -33,58 +36,85 @@ pub struct ModuleMetadata {
...
     }
 
+    /// Deserialize the Metadata from a slice.
+    /// The slice must have the following format:
+    /// RKYV serialization (any length) + POS (8 bytes) + Endian (1 byte)
+    ///
+    /// # Safety
+    ///
+    /// This method is unsafe since it deserializes data directly
+    /// from memory.
```
nlewycky, line 86:
```suggestion
```

lib/engine-native/src/serialize.rs
```diff
@@ -33,58 +36,85 @@ pub struct ModuleMetadata {
...
 
+    /// # Safety
+    ///
+    /// Please check `ModuleMetadata::deserialize` for the safety details.
     unsafe fn archive_from_slice<'a>(
         metadata_slice: &'a [u8],
     ) -> Result<&'a ArchivedModuleMetadata, DeserializeError> {
+        if metadata_slice.len() < 9 {
+            return Err(DeserializeError::Incompatible(
+                "invalid serialized data".into(),
```
nlewycky, line 105: Optional: "too short to be a serialized module"?

lib/engine-object-file/src/artifact.rs
```diff
@@ -443,7 +465,52 @@ impl Artifact for ObjectFileArtifact {
...
+        // The function sizes migth not be completely accurate.
+        // Because of that, we (reverse) order all the functions by their pointer.
+        // [f9, f7, f6, f8...] and calculate their potential function body size by
+        // getting the diff in pointers between functions.
+        let mut prev_pointer = usize::MAX;
+
+        let fp = self.finished_functions.clone();
+        let mut function_pointers = fp.into_iter().collect::<Vec<_>>();
+        // Sort the keys by the values in reverse order (function pointers)
+        // This way we can get the maximu function lengths (since functions can't collide in memory)
```
nlewycky, line 483:
```suggestion
        // This way we can get the maximum function lengths (since functions can't collide in memory)
```

lib/engine-object-file/src/serialize.rs
```diff
@@ -20,19 +56,80 @@ pub struct ModuleMetadataSymbolRegistry {
...
+    }
+
+    /// Deserialize the Metadata from a slice.
+    /// The slice must have the following format:
+    /// RKYV serialization (any length) + POS (8 bytes) + Endian (1 byte)
+    ///
+    /// # Safety
+    ///
+    /// This method is unsafe since it deserializes data directly
+    /// from memory.
```
nlewycky, line 86:
```suggestion
    /// The `metadata_slice` must be an archive produced by
    /// `ModuleMetadata::serialize`.
```

nlewycky (approved):

lib/engine-native/src/serialize.rs
syrusakbary: It should be obscure for the user that info

syrusakbary:
bors try

PR #2276, opened 2021-04-29
Update rkyv to version 0.6.0

MarkMcCaskey:
- Due to a security vulnerability, we have to update rkyv. https://rustsec.org/advisories/RUSTSEC-2021-0054
- Rkyv requires Rust 1.51, so I had to bump that too.
- Rkyv's API changed and I had to update our code that used it

It's fine that we're not 1 behind latest stable because we're close to the next release of Rust. I'm not very confident in the changes I made, I looked through the rkyv book and tried to find examples but I don't like the part where we have to `assume_init` without checking any kind of failure code... it looks like buggy code to me.


syrusakbary (approved):

syrusakbary:
Looks great, thanks for the quick fix!
bors r+

PR #2277, opened 2021-04-29
Update wasmparser to 0.77

MarkMcCaskey:
Resolves multiple failures in #2268 

Primarily this is just supporting the new SIMD ops. I also snuck in a change to our WAST test runner to fix another issue...

When updating the C API, I chose to namespace all the `wasmparser` Operators we're using because without that, the error messages are really quite bad. `SomeStringOfCharacters` in a match statement will match anything (it's bound as a variable name) if there's no enum variant that matches. `::` is not valid in variable names so namespacing it makes it unambiguous that we're trying to match a specific thing, not bind a variable and the error messages we get are much clearer.


MarkMcCaskey:
bors try

lib/c-api/src/wasm_c_api/unstable/parser/operator.rs
```diff
@@ -508,510 +520,522 @@ pub enum wasmer_parser_operator_t {
 
 impl<'a> From<&Operator<'a>> for wasmer_parser_operator_t {
     fn from(operator: &Operator<'a>) -> Self {
-        use Operator::*;
+        use Operator as O;
```
syrusakbary, line 523: I like this!

syrusakbary (approved):
Looks great!

syrusakbary:
We should be ready to go once #2276 lands!

PR #2280, opened 2021-04-30
Unify test running of compilers

MarkMcCaskey:
The idea behind this PR is to spend less time compiling in CI and just run all of our tests in parallel directly.

There are still a couple of rough edges, but it works. It's higher priority to do this for the C API but here it is for our core tests. This should improve caching and the performance of our CI tests.

We still require 2 compilation steps rather than 1 for this part, but hopefully we can still see some of the benefits of this

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -136,12 +136,16 @@ jobs:
           path: |
             ~/.cargo/registry
             ~/.cargo/git
-          key: ${{ matrix.build }}-${{ matrix.target }}-cargo-${{ hashFiles('Cargo.lock') }}-v1
+          key: ${{ matrix.build }}-${{ matrix.target }}-cargo-${{ hashFiles('**/Cargo.lock') }}-v1
+          restore-keys: |
+            ${{ matrix.build }}-${{ matrix.target }}-cargo-
       - uses: actions/cache@v2
         if: matrix.use_sccache
         with:
           path: ${{ runner.tool_cache }}/cargo-sccache
           key: ${{ matrix.build }}-${{ matrix.target }}-sccache-bin-${{ env.CARGO_SCCACHE_VERSION }}-v1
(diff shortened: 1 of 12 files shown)


MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

syrusakbary:
I was thinking on having something similar, but via a macro that tests can be decorated.

So something like this:
```rust
#[test]
#[variadic_test(engine = ["jit", "native], compiler = ["singlepass", "cranelift", "llvm"])]
fn my_test(engine: &str, compiler: &str) {
}
```

Which will expand to something like:
```rust
mod my_test {
    use super::*;
    fn my_test(engine: &str, compiler: &str) {
    }
    mod jit {
        mod singlepass {
            #[test]
            fn test() {
                super::super::my_test("jit", "singlepass"); 
            }
            ...
        }
    }
    mod native {
        ...
    }
}
```

MarkMcCaskey:
When I first saw that I didn't like it because macros complicate things, but I actually think that's a really good idea and would make this a lot cleaner

Would just need a way to opt out specific combinations of things like `engine+compiler`

syrusakbary:
Similar to quickcheck, but with fixed values:
https://github.com/BurntSushi/quickcheck

PS: this is caused because Rust doesn't have a proper test framework :(
https://rust-lang.github.io/rfcs/2318-custom-test-frameworks.html

MarkMcCaskey:
> PS: this is caused because Rust doesn't have a proper test framework :(

haha I'd probably call that a test framework framework, but I agree, that would be nice to have!

syrusakbary:
Ok, someone might have already done most of the work for us!
https://altsysrq.github.io/proptest-book/proptest/tutorial/macro-proptest.html

syrusakbary:
I think we can close this PR in favor of #2321

PR #2281, opened 2021-05-01
Refactor vm

syrusakbary:
# Description

This PR simplifies the VM code. and improves it's resiliency:
* [x] Removed unnecessary Exported structs
* [x] Reuse same VM data when available. Fixing #2131 
* [x] Removed unnecessary Function Definition in the API 
* [ ] Fixes the NativeEngine by automatically deallocating the libraries when the module is no longer used

# Review

### DIFF
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -18,7 +18,7 @@ use std::sync::Arc;
 use wasmer_engine::{Export, ExportFunction, ExportFunctionMetadata};
 use wasmer_vm::{
     raise_user_trap, resume_panic, wasmer_call_trampoline, ImportInitializerFuncPtr,
-    VMCallerCheckedAnyfunc, VMDynamicFunctionContext, VMExportFunction, VMFuncRef, VMFunctionBody,
+    VMCallerCheckedAnyfunc, VMDynamicFunctionContext, VMFunction, VMFuncRef, VMFunctionBody,
     VMFunctionEnvironment, VMFunctionKind, VMTrampoline,
 };
 
@@ -218,7 +218,7 @@ impl Function {
                         )
                     },
                 )),
-                vm_function: VMExportFunction {
+                vm_function: VMFunction {
                     address,
                     kind: VMFunctionKind::Dynamic,
                     vmctx,
@@ -311,7 +311,7 @@ impl Function {
             definition: FunctionDefinition::Host(HostFunctionDefinition { has_env: true }),
             exported: ExportFunction {
                 metadata: Some(Arc::new(metadata)),
-                vm_function: VMExportFunction {
+                vm_function: VMFunction {
                     address,
                     kind: VMFunctionKind::Dynamic,
                     vmctx,
@@ -365,7 +365,7 @@ impl Function {
                 // TODO: figure out what's going on in this function: it takes an `Env`
                 // param but also marks itself as not having an env
                 metadata: None,
-                vm_function: VMExportFunction {
+                vm_function: VMFunction {
                     address,
                     vmctx,
                     signature,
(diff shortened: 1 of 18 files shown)


syrusakbary:
bors try

syrusakbary:
bors try-

syrusakbary:
bors try

syrusakbary:
bors try

lib/api/src/externals/function.rs
```diff
@@ -1285,6 +1167,7 @@ mod inner {
 
     /// An empty struct to help Rust typing to determine
     /// when a `HostFunction` does not have an environment.
+    #[derive(Clone)]
```
nlewycky, line 1170: Cloning an empty struct? Why? Is the struct just used in the type system? Or is it compared by pointer?

lib/api/src/native.rs
```diff
@@ -115,98 +109,89 @@ macro_rules! impl_native_traits {
...
-                                // we know it's not overlapping because `using_rets_array` is false
-                                std::ptr::copy_nonoverlapping(src_pointer,
-                                                              rets_list,
-                                                              num_rets);
-                            }
+                            // TODO: we can probably remove this copy by doing some clever `transmute`s.
+                            // we know it's not overlapping because `using_rets_array` is false
+                            std::ptr::copy_nonoverlapping(src_pointer,
+                                                            rets_list,
+                                                            num_rets);
```
nlewycky, line 149: Indentation looks wrong? Does this pass lint?

lib/vm/src/export.rs
```diff
@@ -103,21 +103,21 @@ impl VMExportTable {
...
+    fn from(table: VMTable) -> Self {
         Self::Table(table)
     }
 }
 
 /// A memory export value.
-#[derive(Debug, Clone)]
-pub struct VMExportMemory {
+#[derive(Debug, Clone, MemoryUsage)]
+pub struct VMMemory {
```
nlewycky, line 120: No action required. FWIW I think that naming these things VM* is a bug. It's not a new bug introduced in your PR. VM* should mean "used by the VM" as in "the generated code uses this" which implies "must be repr(C) and contain things that are available in repr(C)" which would exclude Arc for instance (I've been arguing we should make a VMArc).

nlewycky (approved):

lib/api/src/externals/function.rs
syrusakbary: It's required for `WasmerEnv` to be clonable

lib/api/src/native.rs
syrusakbary: Yeah, weird. This change was made by the linter

lib/api/src/externals/function.rs
```diff
@@ -649,34 +570,19 @@ impl Function {
     /// assert_eq!(sum.call(&[Value::I32(1), Value::I32(2)]).unwrap().to_vec(), vec![Value::I32(3)]);
     /// ```
     pub fn call(&self, params: &[Val]) -> Result<Box<[Val]>, RuntimeError> {
-        let mut results = vec![Val::null(); self.result_arity()];
-
-        match &self.definition {
-            FunctionDefinition::Wasm(wasm) => {
-                self.call_wasm(&wasm, params, &mut results)?;
-            }
-            // TODO: we can trivially hit this, look into it
```
MarkMcCaskey, line 658: Probably shouldn't remove this comment unless it's been fixed by this PR

lib/api/src/native.rs
MarkMcCaskey: The body of macros are not formatted by rustfmt. If the formatting looks wrong, it has to be fixed manually. It was likely caused by your editor if you didn't do it

MarkMcCaskey (commented):
Some of the refactoring here looks really good, I didn't realize some of that was possible. The large PR and touching so many critical pieces made me concerned that this PR introduced a large breaking change that got past our tests but upon further review I don't think that's the case (at least my initial concerns were wrong). I haven't been able to fully review all the changes in this PR yet though.

PR #2286, opened 2021-05-03
Remove goblin, use object.

nlewycky:


lib/compiler-llvm/src/object_file.rs
```diff
@@ -212,75 +158,86 @@ where
...
-                    ));
-                }
-                goblin::elf::reloc::R_X86_64_GOTPC64 => {
-                    return Err(CompileError::Codegen(
-                        "unimplemented PIC relocation R_X86_64_GOTPC64".into(),
-                    ));
-                }
+            let kind = match (reloc.kind(), reloc.size()) {
+                (object::RelocationKind::Absolute, 64) => RelocationKind::Abs8,
+                (object::RelocationKind::Elf(24), 0) => RelocationKind::X86PCRel8,
```
syrusakbary, line 168: Can we have here the proper static variable here instead of `24`? (I think the `object` has defined most of those)

syrusakbary (commented):
Could you update the `wasmer-object` crate so it depends on object `0.24` instead of `0.23` ?

lib/compiler-llvm/src/object_file.rs
nlewycky: Done.

nlewycky:
> Could you update the `wasmer-object` crate so it depends on object `0.24` instead of `0.23` ?

Done.

syrusakbary (approved):

PR #2292, opened 2021-05-05
wasmer-middlewares: don't force enable wasmer default features

vavrusa:
# Description

wasmer-middlewares doesn't need to enable all default features from wasmer
when it's included. This avoid having to build cranelift crates when llvm is used
otherwise etc.

# Review


syrusakbary (approved):
Good point. Merging

PR #2298, opened 2021-05-06
Generalize these strings, stop chasing changing error messages.

nlewycky:
A runtime error encountered during instantiation has to be one of the kinds of errors that can occur during instantiation, out of bounds covers the cases like initializing memory or table entries that don't exist.

### DIFF
--- a/fuzz/fuzz_targets/jit_cranelift.rs
+++ b/fuzz/fuzz_targets/jit_cranelift.rs
@@ -48,14 +48,7 @@ fuzz_target!(|module: WasmSmithModule| {
         Ok(_) => {}
         Err(e) => {
             let error_message = format!("{}", e);
-            if error_message
-                .contains("RuntimeError: memory out of bounds: data segment does not fit")
-                || error_message
-                    .contains("RuntimeError: table out of bounds: elements segment does not fit")
-                || error_message.contains(
-                    "RuntimeError: out of bounds table access: elements segment does not fit",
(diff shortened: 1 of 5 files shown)


syrusakbary (approved):

fuzz/fuzz_targets/jit_cranelift.rs
```diff
@@ -48,14 +48,7 @@ fuzz_target!(|module: WasmSmithModule| {
...
             let error_message = format!("{}", e);
-            if error_message
-                .contains("RuntimeError: memory out of bounds: data segment does not fit")
-                || error_message
-                    .contains("RuntimeError: table out of bounds: elements segment does not fit")
-                || error_message.contains(
-                    "RuntimeError: out of bounds table access: elements segment does not fit",
-                )
-            {
+            if error_message.contains("RuntimeError: ") && error_message.contains("out of bounds") {
```
Hywan, line 51: Maybe we could use `error.message.starts_with` rather than `.contains` for `"RuntimeError: "` (https://doc.rust-lang.org/std/primitive.str.html#method.starts_with), thoughts?

Hywan: Not a big deal though.

nlewycky: https://github.com/wasmerio/wasmer/pull/2301

PR #2305, opened 2021-05-11
Improve traps

syrusakbary:
# Description

This PR moves the trap code forward so we can start refactoring easily in subsequent PRs:
* Adding a `is_wasm_pc` method to the traps/engine (so we only trap Wasm-related signals/...)
* Making `Store` generic over `TrapInfo` (so users can implement their custom trap handlers)
* Removed `Interrupt` (non-used) and `VMOutOfMemory` (moved into `RuntimeError`) cases in `TrapCode`, so it's deterministic

# Review


syrusakbary:
bors try

syrusakbary:
bors try

lib/engine/src/trap/frame_info.rs
```diff
@@ -42,6 +42,15 @@ pub struct GlobalFrameInfo {
     ranges: BTreeMap<usize, ModuleInfoFrameInfo>,
 }
 
+/// Returns whether the `pc`, according to globally registered information,
+/// is a wasm trap or not.
+pub fn is_wasm_pc(pc: usize) -> bool {
+    let frame_info = FRAME_INFO.read().unwrap();
```
Hywan, line 48: Is there a way to handle the `unwrap` properly here? It seems on a critical path to me.

lib/engine/src/trap/frame_info.rs
```diff
@@ -42,6 +42,15 @@ pub struct GlobalFrameInfo {
...
+    let module_info = frame_info.module_info(pc);
+    let is_wasm_pc = module_info.is_some();
+    is_wasm_pc
```
Hywan, line 51: It's very likely that Clippy will suggest:

```suggestion
    module_info.is_some()
```

lib/vm/src/instance/mod.rs
```diff
@@ -397,7 +393,7 @@ impl Instance {
     }
 
     /// Invoke the WebAssembly start function of the instance, if one is present.
-    fn invoke_start_function(&self) -> Result<(), Trap> {
+    fn invoke_start_function(&self, trap_info: &dyn TrapInfo) -> Result<(), Trap> {
```
Hywan, line 396: I wonder whether it should be better to write this as such:

```suggestion
    fn invoke_start_function<T: TrapInfo>(&self, trap_info: &T) -> Result<(), Trap> {
```

lib/vm/src/instance/mod.rs
```diff
@@ -1000,9 +995,6 @@ impl InstanceHandle {
             VMBuiltinFunctionsArray::initialized(),
         );
 
-        // Ensure that our signal handlers are ready for action.
-        init_traps();
```
Hywan, line 1004: The initialization has been moved in `Store`. Fine to me. I wonder what will happen if we have multiple stores, but that's because I'm ignorant of this code.

Another question: Is there a possible way to check whether the traps have been initialized? I think it's worth adding a check here to prevent against a bad usage of the API. Thought?
Edit: With [`Once::is_completed`](https://doc.rust-lang.org/std/sync/struct.Once.html#method.is_completed), it's possible to check whether the traps have been initialized correctly.

Hywan (commented):
Looks good to me!

lib/vm/src/instance/mod.rs
syrusakbary: `init_traps` is moved into the store. So it should be fine with multiple stores.

> Another question: Is there a possible way to check whether the traps have been initialized? I think it's worth adding a check here to prevent against a bad usage of the API.

The current API already prevents any misusage. Traps are initialized as soon as you init the store, so there should be no need for extra APIs or pushing the responsibility into the user

lib/vm/src/trap/traphandlers.rs
```diff
@@ -739,47 +753,146 @@ impl Drop for CallThreadState {
...
-        PTR.with(|p| {
-            let _r = Reset(p, p.replace(ptr));
-            closure()
-        })
+        // Note that this extension of the lifetime to `'static` should be
+        // safe because we only ever access it below with an anonymous
+        // lifetime, meaning `'static` never leaks out of this module.
+        let ptr = unsafe {
+            mem::transmute::<*const CallThreadState<'_>, *const CallThreadState<'static>>(state)
+        };
```
nlewycky, line 867: Why is the extension to `'static` lifetime needed at all? Should that be obvious from the code and I'm just missing it, or should the comment answer that?

lib/vm/src/instance/mod.rs
syrusakbary: I tried that after reading your suggestion, had to revert.
In the `Engine` a trait the `finish_instantiation` function (which calls this function under the hood) would need to take a generic parameter which causes the code to not compile.

lib/vm/src/trap/trapcode.rs
```diff
@@ -61,17 +61,8 @@ pub enum TrapCode {
...
-    /// This trap is resumable.
-    Interrupt = 11,
-
     /// An atomic memory access was attempted with an unaligned pointer.
     UnalignedAtomic = 12,
-
-    /// A trap indicating that the runtime was unable to allocate sufficient memory.
-    VMOutOfMemory = 13,
-    // /// A user-defined trap code.
-    // User(u16),
```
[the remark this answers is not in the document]
syrusakbary: Fixed!

lib/vm/src/trap/traphandlers.rs
```diff
@@ -427,55 +444,66 @@ pub enum Trap {
     /// A user-raised trap through `raise_user_trap`.
     User(Box<dyn Error + Send + Sync>),
 
-    /// A trap raised from machine code generated from Wasm
+    /// A trap raised from the Wasm generated code
+    ///
+    /// Note: this trap is deterministic (assuming a deterministic host implementation)
     Wasm {
-        /// The program counter in generated code where this trap happened.
+        /// The program counter in JIT code where this trap happened.
```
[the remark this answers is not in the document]
syrusakbary: Fixed

lib/vm/src/trap/traphandlers.rs
```diff
@@ -427,55 +444,66 @@ pub enum Trap {
...
+    Lib {
         /// Code of the trap.
         trap_code: TrapCode,
         /// Native stack backtrace at the time the trap occurred
         backtrace: Backtrace,
     },
+
+    /// A trap indicating that the runtime was unable to allocate sufficient memory.
+    ///
+    /// Note: this trap is undeterministic, since it depends on the host system.
```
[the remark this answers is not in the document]
syrusakbary: Fixed

lib/vm/src/trap/traphandlers.rs
syrusakbary: Fixed

lib/vm/src/trap/traphandlers.rs
```diff
@@ -739,47 +753,146 @@ impl Drop for CallThreadState {
...
+    }
+
+    /// Opaque state used to help control TLS state across stack switches for
+    /// async support.
+    pub struct TlsRestore(raw::Ptr);
+
+    impl TlsRestore {
+        /// Takes the TLS state that is currently configured and returns a
+        /// token that is used to replace it later.
+        ///
```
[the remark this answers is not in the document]
syrusakbary: Added

lib/api/src/store.rs
```diff
@@ -69,6 +72,18 @@ impl PartialEq for Store {
     }
 }
 
+unsafe impl TrapInfo for Store {
```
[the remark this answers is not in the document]
syrusakbary: Renamed it to `TrapHandler`

lib/api/src/store.rs
```diff
@@ -69,6 +72,18 @@ impl PartialEq for Store {
     }
 }
 
+unsafe impl TrapHandler for Store {
+    #[inline]
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+    fn custom_trap_handler(&self, _call: &dyn Fn(&TrapHandlerFn) -> bool) -> bool {
+        // Setting a custom handler is implemented for now, please open an issue
```
nlewycky, line 81:
```suggestion
        // Setting a custom handler is unimplemented for now, please open an issue
```

Also, this sort of "please open an issue in github repo" is a very odd thing to do, I don't like it. I don't this needs an comment (I think my previous review asked for a comment, but I meant for the definition of the `trait TrapHandler` which already had a comment anyways?).

lib/engine/src/trap/frame_info.rs
```diff
@@ -42,6 +42,14 @@ pub struct GlobalFrameInfo {
     ranges: BTreeMap<usize, ModuleInfoFrameInfo>,
 }
 
+/// Returns whether the `pc`, according to globally registered information,
+/// is a wasm trap or not.
+pub fn is_wasm_pc(pc: usize) -> bool {
+    let frame_info = FRAME_INFO.read().expect("can't acquire the FrameInfo lock");
```
nlewycky, line 48: When could this happen anyways? Would it be enough to have the same code running in two threads crashing at the same time, triggering the `is_wasm_pc` in both threads?

lib/vm/src/trap/traphandlers.rs
```diff
@@ -399,8 +417,9 @@ pub unsafe fn raise_lib_trap(trap: Trap) -> ! {
 ///
 /// # Safety
 ///
-/// Only safe to call when wasm code is on the stack, aka `wasmer_call` or
-/// `wasmer_call_trampoline` must have been previously called.
+/// Only safe to call when wasm code is on the stack, aka `catch_traps` must
+/// have been previously called. Additionally no Rust destructors can be on the
```
nlewycky, line 421: "been previously called" --> "been previously called and not returned" here too, and in a few other comments too.

lib/vm/src/trap/traphandlers.rs
```diff
@@ -547,106 +570,104 @@ where
...
     })?;
     Ok(global_results.assume_init())
 }
 
 /// Temporary state stored on the stack which is registered in the `tls` module
 /// below for calls into wasm.
-pub struct CallThreadState {
-    unwind: Cell<UnwindReason>,
+pub struct CallThreadState<'a> {
+    unwind: UnsafeCell<MaybeUninit<UnwindReason>>,
```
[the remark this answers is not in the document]
nlewycky: Maybe add an explicit "Poison" state to the UnwindReason and then initialize to that?

lib/vm/src/trap/traphandlers.rs
```diff
@@ -547,106 +572,104 @@ where
...
-                Some(prev) => (*prev).any_instance(func),
-                None => false,
+            UnwindReason::Panic(panic) => {
+                debug_assert_eq!(ret, 0);
+                std::panic::resume_unwind(panic)
             }
         }
     }
 
     fn unwind_with(&self, reason: UnwindReason) -> ! {
```
nlewycky, line 669: Shouldn't this be `mut self` since this is a mutating function?

lib/vm/src/trap/traphandlers.rs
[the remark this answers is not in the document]
nlewycky: I made a pidgin copy of the problem in the rust playground, this code works:
```
use std::cell::UnsafeCell;
pub struct Foo {
    unwind: UnsafeCell<Option<UnwindReason>>,
}
enum UnwindReason {
    Foo(u8),
    Bar
}
impl Foo {
    fn a(mut self) {
        match unsafe { &*self.unwind.get() } {
            Some(UnwindReason::Foo(_unwind)) => {},
            Some(UnwindReason::Bar) => {},
            None => {}
        }
    }
    fn b(&self, reason: UnwindReason) {
        *(unsafe { &mut *self.unwind.get() }) = Some(reason);
    }
}
```

lib/api/src/store.rs
MarkMcCaskey: Yeah, probably better to at least open the issue ourselves so discussion gets centralized and we can update people about it there.

lib/engine/src/trap/frame_info.rs
MarkMcCaskey: No, this is https://doc.rust-lang.org/std/sync/type.LockResult.html, the reason we just `unwrap` these is because it's not recoverable in general. I think the `expect` message is a bit misleading here as it sounds like it's doing a non-blocking acquire on the lock and failing, but it's really that the data covered by the lock cannot be guaranteed to be in a valid state anymore.

Panicking while holding a lock poisons the lock (because the data is in an undefined state), the best thing to do > 99% of the time is to propagate the panic by just `unwrap`ing it because the system is likely in an unrecoverable state, at least as far as that data is concerned.

syrusakbary: Cool, I'll move it back into `unwrap`

lib/vm/src/trap/traphandlers.rs
syrusakbary: I tried to deviate minimally on this from prev implementation on this. I tried it after your comment though, and it is technically impossible because of this:
```
   --> lib/vm/src/trap/traphandlers.rs:644:39
    |
644 |         let ret = tls::set(&mut self, || closure(&mut self))?;
    |                   -------- ---------  ^^              ---- second borrow occurs due to use of `self` in closure
    |                   |        |          |
    |                   |        |          second mutable borrow occurs here
    |                   |        first mutable borrow occurs here
    |                   first borrow later used by call
```

lib/api/src/store.rs
syrusakbary: Just implemented it!

syrusakbary:
bors try

nlewycky (approved):

syrusakbary:
Merging manually to avoid CI since tests have already passed and there are no new changes since last tests

PR #2307, opened 2021-05-12
Update cranelift, impl low hanging fruit SIMD opcodes

MarkMcCaskey:
Bumps to clif 0.73 and implements the new SIMD opcodes that are 0 effort to do. (I thought i8x16popcnt would be zero effort too, but popcnt doesn't work on vectors in cranelift). Also updates `ignores.txt` due to the newly passing spectests


syrusakbary (approved):

lib/compiler-cranelift/src/translator/code_translator.rs
```diff
@@ -2715,10 +2846,10 @@ fn optionally_bitcast_vector(
...
 #[inline(always)]
 fn is_non_canonical_v128(ty: ir::Type) -> bool {
-    matches!(
-        ty,
-        B8X16 | B16X8 | B32X4 | B64X2 | I64X2 | I32X4 | I16X8 | F32X4 | F64X2
-    )
+    match ty {
+        B8X16 | B16X8 | B32X4 | B64X2 | I64X2 | I32X4 | I16X8 | F32X4 | F64X2 => true,
+        _ => false,
+    }
```
MarkMcCaskey, line 2852: Any reason for this change? `matches!` is nice because it's more restricted in what it can do, so when you see it you know there's nothing complicated going on

MarkMcCaskey:
Looks like we're getting some verifier errors with your changes, probably need to rewrite those or disable them for now

lib/compiler-cranelift/src/translator/code_translator.rs
syrusakbary: This is a mimic of the latest changes here:

https://github.com/bytecodealliance/wasmtime/blob/main/cranelift/wasm/src/code_translator.rs#L2764-L2770

The smaller the diff, the better

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

MarkMcCaskey:
bors try

PR #2309, opened 2021-05-12
Initial commit of a wrapper for lld.

nlewycky:


lib/lld-wrapper/src/link.cc
```diff
@@ -0,0 +1,17 @@
...
+
+#include <lld/Common/Driver.h>
+#include <lld/Common/ErrorHandler.h>
+
+extern "C" {
+
+void wasmer_lld_wrapper_link(const char **filenames, uint32_t count) {
+  llvm::ArrayRef<const char *> files(filenames, count);
+
+  if (!lld::elf::link(files, false, llvm::outs(), llvm::errs()))
```
syrusakbary, line 13: This linking only supports elf I assume.

Can we adapt the kind of linking depending on the provided format (coff in windows, macho in MacOS)?

lib/lld-wrapper/src/link.cc
```diff
@@ -0,0 +1,17 @@
...
+    lld::exitLld(0);
```
syrusakbary, line 14: What this command will do? (exit the process?)

nlewycky: Yep. To quote the docs:
```
// Generic entry point when using LLD as a library, safe for re-entry, supports
// crash recovery. Returns a general completion code and a boolean telling
// whether it can be called again. In some cases, a crash could corrupt memory
// and re-entry would not be possible anymore. Use exitLld() in that case to
// properly exit your application and avoid intermittent crashes on exit caused
// by cleanup.
```

We probably just want to not use this API at all, but for now this is how it's going to be.

PR #2312, opened 2021-05-14
Attempt to solve cyclic ref in WasmerEnv

### LINKED ISSUE #2304: LazyInit<Memory> does not call munmap
### Describe the bug
I noticed that when a host function's environment has a `LazyInit<Memory> `as a member, the exported memory is not cleaned up properly.

I dug through the Wasmer code last night but could not figure out what was causing this. Probably when resolving imports the Environment is cloned and then not destroyed, which keeps a dangling reference to the Memory.

(This code is really hard to read as it passed around a raw pointer to the environment and a bunch of function pointers to clone and drop it. Might be worth investigating if this could be done in a nicer way.)

### Steps to reproduce
You can checkout my branch and run the modified example.

```
git clone git@github.com:kaimast/wasmer.git
cd wasmer
git checkout lazy-init
cargo run --example imported-function-env --features=cranelift
```

This will instantiate a module and call a function 10000 times sequentially. Run `htop` (or similar) and observe the virtual memory usage of the program.

### Expected behavior
All memory should be unmapped after an instance is dropped.

### Actual behavior
On my system this simple example reserves 25 Terabytes in memory... and a process can run out of virtual memory in a few minutes.

MarkMcCaskey:
Trying to fix #2304

I'm not sure we want to go with the strategy in this PR, it's very low level and unsafe. The idea is to decrement the strong count of `InstanceRef` for each export in the Env because in the context of host functions, anything coming from the same instance is guaranteed to be alive. We must preserve the property that cloning these types then increments the `InstanceRef` count though, so we can't get rid of it entirely. Additionally we have to handle the case of something being re-exported, so we can't assume that if something is exported from an Instance that it's owned by that Instance.

The PR currently gets stuck in an infinite loop during instantiation.

# Review

### DIFF
--- a/lib/api/src/env.rs
+++ b/lib/api/src/env.rs
@@ -87,6 +87,11 @@ pub trait WasmerEnv: Clone + Send + Sync {
     fn init_with_instance(&mut self, _instance: &Instance) -> Result<(), HostEnvInitError> {
         Ok(())
     }
+
+    // TODO: this probably can't reasonably have a default unless we want to default
+    // to leaking memory
+    /// Used to prevent memory leaks with cycles when using `WasmerEnv` with exports.
+    unsafe fn dec_strong_count_if_from_same_instance(&mut self, _instance: &Instance) {}
 }
 
 impl WasmerEnv for u8 {}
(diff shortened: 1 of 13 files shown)


MarkMcCaskey:
Closing this in favor of #2327 as #2327 is essentially an iteration on this that's more complete and simpler

PR #2318, opened 2021-05-15
Improved traphandler code

syrusakbary:
# Description

Improved traphandler code by simplifying the TrapHandler and making explicit it has to be Send + Sync.

# Review

### DIFF
--- a/lib/api/src/store.rs
+++ b/lib/api/src/store.rs
@@ -40,9 +40,8 @@ impl Store {
     }
 
     /// Retrieves the [`TrapHandler`] pointer if is already set for the `Store`.
-    pub fn trap_handler(&self) -> Option<*const TrapHandler<'static>> {
-        let trap_handler = (*self.trap_handler).as_ref()?;
-        Some(&**trap_handler as *const _)
+    pub fn trap_handler(&self) -> Arc<Option<Box<TrapHandler<'static>>>> {
+        self.trap_handler.clone()
     }
 
     /// Creates a new `Store` with a specific [`Engine`] and [`Tunables`].
--- a/lib/api/src/store.rs
+++ b/lib/api/src/store.rs
@@ -1,12 +1,11 @@
 use crate::tunables::BaseTunables;
 use loupe::MemoryUsage;
-use std::any::Any;
 use std::fmt;
-use std::sync::{Arc, RwLock};
+use std::sync::Arc;
 #[cfg(all(feature = "compiler", feature = "engine"))]
 use wasmer_compiler::CompilerConfig;
 use wasmer_engine::{is_wasm_pc, Engine, Tunables};
-use wasmer_vm::{init_traps, TrapHandler, TrapHandlerFn};
+use wasmer_vm::{init_traps, TrapHandler};
 
 /// The store represents all global state that can be manipulated by
 /// WebAssembly programs. It consists of the runtime representation
@@ -23,7 +22,7 @@ pub struct Store {
     engine: Arc<dyn Engine + Send + Sync>,
     tunables: Arc<dyn Tunables + Send + Sync>,
     #[loupe(skip)]
-    trap_handler: Arc<RwLock<Option<Box<TrapHandlerFn>>>>,
+    pub(crate) trap_handler: Arc<Option<Box<TrapHandler<'static>>>>,
 }
 
 impl Store {
@@ -35,10 +34,15 @@ impl Store {
         Self::new_with_tunables(engine, BaseTunables::for_target(engine.target()))
     }
 
-    /// Set the trap handler in this store.
-    pub fn set_trap_handler(&self, handler: Option<Box<TrapHandlerFn>>) {
-        let mut m = self.trap_handler.write().unwrap();
-        *m = handler;
+    /// Sets a [`TrapHandler`] for the `Store`
+    pub fn set_trap_handler(&mut self, handler: Option<Box<TrapHandler<'static>>>) {
+        self.trap_handler = Arc::new(handler);
+    }
+
+    /// Retrieves the [`TrapHandler`] pointer if is already set for the `Store`.
+    pub fn trap_handler(&self) -> Option<*const TrapHandler<'static>> {
+        let trap_handler = (*self.trap_handler).as_ref()?;
+        Some(&**trap_handler as *const _)
     }
 
     /// Creates a new `Store` with a specific [`Engine`] and [`Tunables`].
@@ -53,7 +57,7 @@ impl Store {
         Self {
             engine: engine.cloned(),
             tunables: Arc::new(tunables),
-            trap_handler: Arc::new(RwLock::new(None)),
+            trap_handler: Arc::new(None),
         }
     }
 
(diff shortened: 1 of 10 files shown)


Hywan (commented):
> Improved traphandler code

Can you elaborate please about how the code is improved now? What was the issue with the previous version of the code?

Thanks!

syrusakbary:
> Can you elaborate please about how the code is improved now? What was the issue with the previous version of the code?

Before it was depending on a Trait just for using a function. It makes the system on depending on a Lock when it was not really necessary. It is much simpler to depend on a function!

lib/api/src/store.rs
```diff
@@ -23,7 +22,7 @@ pub struct Store {
     engine: Arc<dyn Engine + Send + Sync>,
     tunables: Arc<dyn Tunables + Send + Sync>,
     #[loupe(skip)]
```
Hywan, line 24: It may be possible to remove `#[loupe(skip)]` here I guess (if `TrapHandler` can be inspected).

lib/api/src/store.rs
```diff
@@ -35,10 +34,15 @@ impl Store {
...
-    pub fn set_trap_handler(&self, handler: Option<Box<TrapHandlerFn>>) {
-        let mut m = self.trap_handler.write().unwrap();
-        *m = handler;
+    /// Sets a [`TrapHandler`] for the `Store`
+    pub fn set_trap_handler(&mut self, handler: Option<Box<TrapHandler<'static>>>) {
+        self.trap_handler = Arc::new(handler);
+    }
+
+    /// Retrieves the [`TrapHandler`] pointer if is already set for the `Store`.
+    pub fn trap_handler(&self) -> Option<*const TrapHandler<'static>> {
```
Hywan, line 43: I don't know but I don't like returning a pointer here. Because in other structs, we store a pointer, and we lost track of the lifetime and every other kind of guarantees provided by Rust.

Could we return an `Option<&TrapHandler<'static>>` instead?
Or even better, since it's behind an `Arc`, we can return `Arc<Option<Box<TrapHandler<'static>>>` already.

Thoughts?

lib/api/src/store.rs
Hywan: Edit: Nah, it's not possible because `TrapHandler` is a type alias to `dyn Fn`. Forget about it!

lib/api/src/store.rs
syrusakbary: We can't make it `Option<&TrapHandler<'static>>` because it has to be stored in `CallThreadState` and we will need to unsafely cast the pointer later to be able to compile later.

I somehow prefer `Option<*const TrapHandler<'static>>` over `Arc<Option<Box<TrapHandler<'static>>>` (since the lifetime of `TraphHandler` is always static there should be no issue?)

lib/engine/src/artifact.rs
```diff
@@ -161,7 +161,7 @@ pub trait Artifact: Send + Sync + Upcastable + MemoryUsage {
     /// See [`InstanceHandle::finish_instantiation`].
     unsafe fn finish_instantiation(
         &self,
-        trap_handler: &dyn TrapHandler,
+        trap_handler: Option<*const TrapHandler<'static>>,
```
MarkMcCaskey, line 164: Just a note about style, it's a little bit odd to see `Option<*const T>` as this is essentially a doubly nullable pointer. I think I'd prefer using `Option<NonNull<T>>` here as that uses the null representation of the pointer itself as the `Option::None` variant. Not a huge concern though

MarkMcCaskey: Alternatively, when used in the argument position, I think it's fine to just use `*const TrapHandler<'static>` here too. And using the null representation instead of `Option`. I like `Option` because it forces the receiver to check if it's null though

lib/api/src/store.rs
MarkMcCaskey: Yeah, I agree with Ivan that I find the use of a raw pointer a bit concerning, in large part because I've read over this PR twice and can't convince myself that it's entirely correct. With `CallThreadState` no longer bound by the lifetime of the trap handler, how do we know that `CallThreadState` can't stay around longer than the `Store` and eventually call freed memory as a function. Calling freed memory as a function is one of the highest risk security exploits a system can have (because it can fairly directly lead to arbitrary code execution), so I'm definitely in favor of keeping the lifetime constraints or using `Arc` here to make sure that can't happen. Thinking in terms of not just how the system is now, but in terms of all the possible changes that might happen to it, I think we want any extra guarantees we can get here about `TrapHandler` always being valid in the context of `CallThreadState`.

syrusakbary: Aand... done! :)

lib/engine/src/artifact.rs
syrusakbary: Used `Arc` at the end :)

lib/api/src/store.rs
```diff
@@ -35,10 +34,14 @@ impl Store {
         Self::new_with_tunables(engine, BaseTunables::for_target(engine.target()))
     }
 
-    /// Set the trap handler in this store.
-    pub fn set_trap_handler(&self, handler: Option<Box<TrapHandlerFn>>) {
-        let mut m = self.trap_handler.write().unwrap();
-        *m = handler;
+    /// Sets a [`TrapHandler`] for the `Store`
+    pub fn set_trap_handler(&mut self, handler: Option<Box<TrapHandler<'static>>>) {
+        self.trap_handler = Arc::new(handler);
```
MarkMcCaskey, line 39: So I like this, but lets make sure the new semantics are what we intended. In the old code, calling `set_trap_handler` would globally update the trap handler in all `Store`s. In the new code it only updates the trap handler for this specific store (and Stores `Clone`d from it). All the old trap handlers will continue to work here because they're kept alive by the `Arc`. (This is actually a great example of how using the raw pointer could have caused problems, the moment `set` is called again, the trap handler being used could have been invalid.)

If the move from `Store`'s having 1 trap handler to `Store`s having N trap handlers in a tree-like structure is intentional then 👍 , if not then we probably need to rethink some of this.

And if this is intentional we may want to call it out more in the doc comments.

MarkMcCaskey: Also side note but if you want to remove the extra `Box` in `Arc<Option<Box<TrapHandler<'static>>>` let me know, I think it should be possible to do so in many cases.

lib/vm/src/instance/mod.rs
```diff
@@ -1015,7 +1018,7 @@ impl InstanceHandle {
     /// Only safe to call immediately after instantiation.
     pub unsafe fn finish_instantiation(
         &self,
-        trap_handler: &dyn TrapHandler,
+        trap_handler: Arc<Option<Box<TrapHandler<'static>>>>,
```
MarkMcCaskey, line 1021: Generally for any of these functions, but you can make them take`&Arc`, too,  especially if they're not storing the `Arc` (in the cases where they just want to read it, you can even pass `Option<&TrapHandler<'static>>` I think by deref'ing the `Arc` and using `as_ref` on the `Option`)

lib/api/src/store.rs
Hywan: Thanks!

lib/engine/src/artifact.rs
Hywan: :+1:

lib/api/src/store.rs
Hywan: I agree with the first comment :-).

heyjdp:
Closing PR, fixed by other PR #2892

PR #2319, opened 2021-05-15
Added call unchecked to the functions

### LINKED ISSUE #1922: Is it possible to avoid syscalls during Function::call?

syrusakbary:
# Description

This PR resolves #1922 by adding a new method to `Store`: `catch_traps` and two new methods to `Function` and `NativeFunction`: `call_unchecked` so unsafe unchecked calls can be done.

```rust
unsafe {
  store.catch_traps(|| {
    assert_eq!(sum.call_unchecked(&[Value::I32(1), Value::I32(2)]).unwrap().to_vec(), vec![Value::I32(3)]);
  });
}
```

## Benchmarking

This are the timings in of `make bench` in my laptop (macOS M1 2020):
* Static function: 138ns -> 3ns
* Dynamic function: 240ns -> 100ns

```
basic static func llvm
                        time:   [136.39 ns 137.16 ns 138.08 ns]

basic static func llvm [unchecked]
                        time:   [3.0288 ns 3.0381 ns 3.0485 ns]

basic static func with many args llvm
                        time:   [149.08 ns 149.28 ns 149.50 ns]

basic static func with many args llvm [unchecked]
                        time:   [15.919 ns 16.085 ns 16.290 ns]

basic dynfunc llvm
                        time:   [242.52 ns 244.49 ns 247.13 ns]

basic dynfunc llvm [unchecked]
                        time:   [101.92 ns 103.36 ns 104.93 ns]
```

### DIFF
--- a/benches/static_and_dynamic_functions.rs
+++ b/benches/static_and_dynamic_functions.rs
@@ -106,7 +106,10 @@ pub fn run_basic_static_function(store: &Store, compiler_name: &str, c: &mut Cri
     );
 
     c.bench_function(
-        &format!("basic static func with many args {} [unchecked]", compiler_name),
+        &format!(
+            "basic static func with many args {} [unchecked]",
+            compiler_name
+        ),
         |b| unsafe {
             store
                 .catch_traps(|| {
--- a/benches/static_and_dynamic_functions.rs
+++ b/benches/static_and_dynamic_functions.rs
@@ -50,7 +50,7 @@ pub fn run_basic_static_function(store: &Store, compiler_name: &str, c: &mut Cri
     });
 
     c.bench_function(
-        &format!("basic static func [unchecked] {}", compiler_name),
+        &format!("basic static func {} [unchecked]", compiler_name),
         |b| unsafe {
             store
                 .catch_traps(|| {
@@ -106,7 +106,7 @@ pub fn run_basic_static_function(store: &Store, compiler_name: &str, c: &mut Cri
     );
 
     c.bench_function(
-        &format!("basic static func with many args {}", compiler_name),
+        &format!("basic static func with many args {} [unchecked]", compiler_name),
         |b| unsafe {
             store
                 .catch_traps(|| {
@@ -146,7 +146,7 @@ pub fn run_basic_dynamic_function(store: &Store, compiler_name: &str, c: &mut Cr
     });
 
     c.bench_function(
-        &format!("basic dynfunc [unchecked] {}", compiler_name),
+        &format!("basic dynfunc {} [unchecked]", compiler_name),
         |b| unsafe {
             store
                 .catch_traps(|| {
--- a/benches/static_and_dynamic_functions.rs
+++ b/benches/static_and_dynamic_functions.rs
@@ -49,6 +49,20 @@ pub fn run_basic_static_function(store: &Store, compiler_name: &str, c: &mut Cri
         })
     });
 
+    c.bench_function(
+        &format!("basic static func [unchecked] {}", compiler_name),
+        |b| unsafe {
+            store
+                .catch_traps(|| {
+                    b.iter(|| {
+                        let result = black_box(f.call_unchecked(4, 6).unwrap());
+                        assert_eq!(result, 10);
+                    })
+                })
+                .unwrap();
+        },
+    );
+
     let dyn_f_many: &Function = instance.exports.get("add20").unwrap();
     let f_many: NativeFunc<
         (
@@ -90,6 +104,27 @@ pub fn run_basic_static_function(store: &Store, compiler_name: &str, c: &mut Cri
             })
         },
     );
+
+    c.bench_function(
+        &format!("basic static func with many args {}", compiler_name),
+        |b| unsafe {
+            store
+                .catch_traps(|| {
+                    b.iter(|| {
+                        let result = black_box(
+                            f_many
+                                .call_unchecked(
+                                    1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18,
+                                    19, 20,
+                                )
+                                .unwrap(),
+                        );
+                        assert_eq!(result, 210);
+                    })
+                })
+                .unwrap();
+        },
+    );
 }
 
 pub fn run_basic_dynamic_function(store: &Store, compiler_name: &str, c: &mut Criterion) {
@@ -102,12 +137,28 @@ pub fn run_basic_dynamic_function(store: &Store, compiler_name: &str, c: &mut Cr
     let instance = Instance::new(&module, &import_object).unwrap();
 
     let dyn_f: &Function = instance.exports.get("add").unwrap();
+
     c.bench_function(&format!("basic dynfunc {}", compiler_name), |b| {
         b.iter(|| {
             let dyn_result = black_box(dyn_f.call(&[Val::I32(4), Val::I32(6)]).unwrap());
             assert_eq!(dyn_result[0], Val::I32(10));
         })
     });
+
+    c.bench_function(
+        &format!("basic dynfunc [unchecked] {}", compiler_name),
+        |b| unsafe {
+            store
+                .catch_traps(|| {
+                    b.iter(|| {
+                        let dyn_result =
+                            black_box(dyn_f.call_unchecked(&[Val::I32(4), Val::I32(6)]).unwrap());
+                        assert_eq!(dyn_result[0], Val::I32(10));
+                    })
+                })
+                .unwrap();
+        },
+    );
 
     let dyn_f_many: &Function = instance.exports.get("add20").unwrap();
     c.bench_function(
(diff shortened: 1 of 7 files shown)


benches/static_and_dynamic_functions.rs
```diff
@@ -146,42 +200,52 @@ pub fn run_basic_dynamic_function(store: &Store, compiler_name: &str, c: &mut Cr
...
     }
 }
 
-fn run_dynamic_benchmarks(_c: &mut Criterion) {
+fn run_dynamic_benchmarks(c: &mut Criterion) {
     #[cfg(feature = "llvm")]
     {
-        let store = Store::new(&JIT::new(wasmer_compiler_llvm::LLVM::new()).engine());
+        let store =
+            Store::new(&wasmer_engine_jit::JIT::new(wasmer_compiler_llvm::LLVM::new()).engine());
```
Hywan, line 232: Why is it necessary to use the full name here? The module (here `JIT`) is re-exported from `wasmer` as far as I know.

lib/api/src/externals/function.rs
```diff
@@ -573,7 +582,56 @@ impl Function {
...
             let mut results = vec![Val::null(); self.result_arity()];
-            self.call_wasm(trampoline, params, &mut results)?;
+            self.call_wasm(trampoline, params, &mut results, false)?;
+            return Ok(results.into_boxed_slice());
+        }
+
+        unimplemented!("The function definition isn't supported for the moment");
+    }
+
+    /// Call the `Function` function (unchecked).
```
Hywan, line 592: This comment doesn't explain what the difference between `call` and `call_unchecked`. Yes, we need to wrap `call_unchecked` in `Store::catch_traps`, OK, but it doesn't explain why I should `call` rather than `call_unchecked` and vice versa. Can you update the comment please :-)?

lib/api/src/native.rs
```diff
@@ -109,6 +109,23 @@ macro_rules! impl_native_traits {
         {
             /// Call the typed func and return results.
             pub fn call(&self, $( $x: $x, )* ) -> Result<Rets, RuntimeError> {
+                self.do_call(false, $( $x, )* )
+            }
+
+            /// Call the typed func and return results (unchecked)
```
Hywan, line 115: Same comment here, what's the difference between `call` and `call_unchecked`? Maybe we should explain this concept in the `Function` module once, and refers to this explanation everytime we need to (like here).

lib/api/src/store.rs
```diff
@@ -41,6 +41,19 @@ impl Store {
         *m = handler;
     }
 
+    /// Catches any Wasm traps that happen within the execution of `closure`,
+    /// returning them as a `Result`.
+    ///
+    /// # Safety
+    ///
+    /// Highly unsafe since `closure` won't have any dtors run.
```
Hywan, line 49:
```suggestion
    /// Highly unsafe since `closure` won't have any destructors run.
```

Hywan: Please, can you add an example of `catch_traps` + `Function::call_unsafe` here? That's the perfect place to provide an example (if possible, a doctest).

lib/vm/src/trap/traphandlers.rs
```diff
@@ -538,9 +539,40 @@ pub unsafe fn wasmer_call_trampoline(
...
+        trampoline,
+    )(vmctx, callee, values_vec);
+}
+
 /// Catches any wasm traps that happen within the execution of `closure`,
 /// returning them as a `Result`.
 ///
+/// # Safety
+///
 /// Highly unsafe since `closure` won't have any dtors run.
```
Hywan, line 576:
```suggestion
/// Highly unsafe since `closure` won't have any destructors run.
```

Hywan (commented):
If I understand correctly, this PR provides a new API to call function with an “unchecked” fashion. A “checked” call runs `catch_traps` everytime a function is called with `wasmer_call_trampoline`. However, an “unchecked” call doesn't run `catch_traps` as it must be run manually by the user with `Store::catch_traps` for example.

What's the point of providing this then? I mean: When a user would like to make “unchecked” calls? The question is answered in the next paragraphs.

Another question: Could we do something like this:

```rust
store.catch_traps(|| {
    function.call_unchecked(…);
    function.call_unchecked(…);
})
```

I suppose yes based on what I see in the benchmark. It means that we can call multiple functions and catch the trap as soon as one is raised. The performance trick is that it “sets up” the trap handler only once, rather than once per call.

OK. That's smart, and that's what boosts the performance!

So…

What if we improve that, and remove this “unsafe API”, and replace it by a new “batch API”. It's probably a lot more thinking, but that's really what we are trying to achieve here!

We could simply provide an API like:

```rust
store.batch_functions(|| {
    function1.call(…);
    do something else
    function2.call(…);
})
```

where `Store::bach_functions` “configures” `vm::trap::traphandlers` to run unchecked calls, and wraps everything inside `catch_traps`.

Pros:

* It removes all the newly introduced unsafe API, like `call_unchecked`,
* It removes the need to handle “unchecked” vs. “checked” API,
* It removes cognitive overload for the users and new contributors,
* It simplifies the API for the users,
* It is less error-prone as it is impossible for the users to forget to wrap “unchecked calls” within `Store::catch_traps` (which is very unexpected and weird to find it in `Store` by the way, but I get why you did it).

Cons:

* The only way I can imagine to “configure” the trap handler is to introduce a global variable. I don't like it. Other solutions exist but at the cost of a potentially larger refactoring (with a `TrapHandler` builder, or holding the `trap_handler` instance somewhere inside the `Store` for example).

Thoughts?

benches/static_and_dynamic_functions.rs
syrusakbary: Yeah, good call. Not sure why that was needed.

I added the change because `cargo bench` was actually not working before and it suggested that change. I can check again.

syrusakbary:
> What if we improve that, and remove this “unsafe API”, and replace it by a new “batch API”. It's probably a lot more thinking, but that's really what we are trying to achieve here!
> 
> We could simply provide an API like:
> 
```rust
store.batch_functions(|| {
    function1.call(…);
    do something else
    function2.call(…);
})
```

`catch_traps` and `call_unchecked` are highly unsafe.
If the `do something else` code has to run any destructor, and `function2.call` traps... the destructor will not be run. (this is documented in the `#Safety` section in the docs for `call_unchecked`.

Hywan:
Then we can mark `Store::batch_functions` as `unsafe`. But the idea is to create an API that reflects the intent, what it is designed for, for the users. Here, the intent is to allow multiple calls.

If “batch” is too specific, we could do something have a more “generic” API like “fast calls”, which will continue to be unsafe of course, but it's better than “unchecked” from my point of view.

Thoughts?

lib/api/src/externals/function.rs
```diff
@@ -478,16 +480,23 @@ impl Function {
...
-            wasmer_call_trampoline(
-                &self.store,
-                self.exported.vm_function.vmctx,
-                trampoline,
-                self.exported.vm_function.address,
-                values_vec.as_mut_ptr() as *mut u8,
-            )
-        } {
-            return Err(RuntimeError::from_trap(error));
+        unsafe {
```
MarkMcCaskey, line 483: I think the unsafe block can be moved into just one of the branches here, that makes it more obvious what could be going wrong here

lib/api/src/externals/function.rs
```diff
@@ -431,11 +431,13 @@ impl Function {
         &self.store
     }
 
+    #[inline(always)]
     fn call_wasm(
         &self,
         trampoline: VMTrampoline,
         params: &[Val],
         results: &mut [Val],
+        unchecked: bool,
```
MarkMcCaskey, line 440: If this function acts as a proxy for unchecked calling, it needs to be `unsafe` if we don't `catch_traps` ourselves inside this function

lib/api/src/externals/function.rs
```diff
@@ -478,16 +480,23 @@ impl Function {
...
+            if !unchecked {
+                wasmer_call_trampoline(
+                    &self.store,
+                    self.exported.vm_function.vmctx,
+                    trampoline,
+                    self.exported.vm_function.address,
+                    values_vec.as_mut_ptr() as *mut u8,
+                )?
+            } else {
+                wasmer_call_trampoline_unchecked(
```
MarkMcCaskey, line 493: It may be possible to represent, in the typesystem, a token that allows calling a function in an unchecked way. The only way to get such a token would be from the `catch_traps` function. This would allow us to not have to mark these functions as unsafe because they'd only be callable in that context.

Here's what I'm thinking:

```Rust
/// A token indicating permission to call Wasm functions without an extra trap handler.
#[derive(Debug)] // this type must not be `Copy` or `Clone`
pub struct CallUncheckedToken(PhantomData<()>); // some zero-sized private internal value so no one can construct this type outside of the file that it's declared in.

fn call_unchecked<F>(f: F) where F: FnOnce(&CallUncheckedToken, ...) -> ... {
   let call_unchecked_token = CallUncheckedToken
   catch_traps(||  f.call_unchecked(&call_unchecked_token, ...))
}
```

And this ensures that all `call_unchecked`s happen in a context where a trap catcher exists. It does mean users have to pass that token around and I'm not 100% sure how trap handling works, does it matter if there are function calls on the stack between the `catch_traps` and the function call itself? If so, then this solution doesn't work, for example given a stack trace from newest to oldest:

```
some_wasm_function_that_traps
host_function2
host_function1
catch_traps
...
```

We'll unwind through those host functions to get to `catch_traps`. If that's not a problem, this solution works. If it is a problem, this solution does not work.

lib/api/src/native.rs
```diff
@@ -109,6 +109,23 @@ macro_rules! impl_native_traits {
...
+            /// This method has to be wrapped using [`Store::catch_traps`] in order to be safe.
+            /// Otherwise traps caused by Wasm code will not be caught by the runtime and will
+            /// be spilled to caller context.
+            pub unsafe fn call_unchecked(&self, $( $x: $x, )* ) -> Result<Rets, RuntimeError> {
+                self.do_call(true, $( $x, )* )
+            }
+
+            /// Call the typed func and return results.
+            #[inline(always)]
+            fn do_call(&self, unchecked: bool, $( $x: $x, )* ) -> Result<Rets, RuntimeError> {
```
MarkMcCaskey, line 128: Same thing, this function should also be marked `unsafe`

lib/api/src/store.rs
```diff
@@ -41,6 +41,19 @@ impl Store {
         *m = handler;
     }
 
+    /// Catches any Wasm traps that happen within the execution of `closure`,
+    /// returning them as a `Result`.
+    ///
+    /// # Safety
+    ///
+    /// Highly unsafe since `closure` won't have any destructors run.
```
MarkMcCaskey, line 49: Won't have any destructors run on the closure itself or the things contained inside it?

lib/vm/src/trap/traphandlers.rs
```diff
@@ -524,6 +524,7 @@ impl Trap {
 ///
 /// Wildly unsafe because it calls raw function pointers and reads/writes raw
 /// function pointers.
+#[inline(always)]
```
MarkMcCaskey, line 527: Seeing lots of `inline(always)` so I'll say what I remember about them:

`#[inline]`: consider this function for cross-crate inlining even without LTO (has no effect if LTO is enabled. has no effect on code within the same crate [citation needed (this may be true if compiling with high optimization settings but not in general)])

`#[inline(always)]`: inline this function in all contexts even if the optimizer thinks it'll make things slower.

`#[inline(always)]`  is a pretty sharp tool because Rust does inlining optimization by default if it thinks it'll make things faster. It can be useful when you really need it, but you usually don't.

https://nnethercote.github.io/perf-book/inlining.html for more info about this

lib/vm/src/trap/traphandlers.rs
```diff
@@ -538,10 +539,41 @@ pub unsafe fn wasmer_call_trampoline(
...
+    )(vmctx, callee, values_vec);
+}
+
 /// Catches any wasm traps that happen within the execution of `closure`,
 /// returning them as a `Result`.
 ///
-/// Highly unsafe since `closure` won't have any dtors run.
+/// # Safety
+///
+/// Highly unsafe since `closure` won't have any destructors run.
```
MarkMcCaskey, line 576: If you clarify the other doc comment with the same phrasing, please update it here too!

heyjdp:
Closing PR, this will be solved by Wasmer v3.0

PR #2320, opened 2021-05-15
Remove unnecessary code

syrusakbary:
# Description
Cranelift has already fixed the SystemV ABI, so it’s safe to remove the WASI Wrappers

# Review

### DIFF
--- a/lib/engine/src/resolver.rs
+++ b/lib/engine/src/resolver.rs
@@ -173,17 +173,7 @@ pub fn resolve_imports(
                         // TODO: We should check that the f.vmctx actually matches
                         // the shape of `VMDynamicFunctionImportContext`
                     }
-                    VMFunctionKind::Static => {
-                        // The native ABI for functions fails when defining a function natively in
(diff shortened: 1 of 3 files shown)


Hywan (approved):
Please, do not forget to update the `CHANGELOG`!

Good job :-).

MarkMcCaskey (approved):
Looks good to me! One thing to note before shipping this though, I believe Singlepass on aarch64 had a very similar problem: more than N args (for an N low enough for WASI's syscalls to trigger) caused bad things to happen (i.e. unitialized memory being passed as args due to a hard-to-fix bug in singlepass)

syrusakbary:
> Looks good to me! One thing to note before shipping this though, I believe Singlepass on aarch64 had a very similar problem: more than N args (for an N low enough for WASI's syscalls to trigger) caused bad things to happen (i.e. unitialized memory being passed as args due to a hard-to-fix bug in singlepass)

Yup. I think we can address that in another PR in singlepass (throwing an error when generating code if the signature is greater than "x")

PR #2321, opened 2021-05-15
Improve test framework and execution speed

syrusakbary:
# Description

This PR focuses on improving the speed that we run the tests:
* [x] Unified C API tests. No more recompilation will be needed for testing different compilers. We bundle all in one object and test based on a environment variable `TEST`
* [x] Simplified Rust tests so only one compilation is needed. Not counting compilation time, here are the new testing times:
  - Building times:
      * test crates: ~3 mins
      * capi: ~2 mins
  - Singlepass: `cargo test --features singlepass`
    * jit: 3 seconds (170 passed; 0 failed; 342 ignored;)
  - Cranelift: `cargo test --features cranelift`
      * jit: 25 seconds (232 passed; 2 failed; 24 ignored; 0 measured; 258 filtered out;)
      * native: 12 minutes (256 passed, 2 failed; 24. ignored)
   - LLVM `cargo test --features llvm`
      * jit: 42 seconds (238 passed; 0 failed; 18 ignored; 0 measured; 256 filtered out)
      * native: (haven't go through everything yet)
  - CAPI: `make test-capi`. Total ~4 min.
      * singlepass+jit: 
      * cranelift+jit
      * cranelift+native
      * llvm+jit
      * llvm+native
* [x] Refactored a bit `ignores.txt`, now it is possible to target multiple features at the same time (OS, chipset, engine, compiler, ...) (small caveat: tests have to be "namespaced" because of some `proc_macro2` issues with Span and `source_file()`.

If I had to guess I think we are going to see at least a 80% improvement on test times, and a much nicer way of testing (just with cargo). There are around 1-3secs on build time that will be increased from the macros I created but the tradeoff is great.

You now want to tests all the simd tests for the llvm compiler at the same time (both engines)? No problem:

```bash
cargo test --features llvm --tests -- simd
```

You want to run absolutely all the tests with all compilers and engines at the same time? (excluding package tests and c-api of course). No prob:

```bash
cargo test --features singlepass,cranelift,llvm
```

Also, do you want to test the capi in all compilers and engines at the same time? Wait no more:

```bash
make test-capi
```

# Review


syrusakbary:
bors try

Makefile
```diff
@@ -614,7 +576,7 @@ else
 		cp wapm-cli/$(TARGET_DIR)/wapm package/bin/ ;\
 	fi
 ifeq ($(IS_DARWIN), 1)
-	codesign -s - package/bin/wapm
+	# codesign -s - package/bin/wapm || true
```
Hywan, line 579: To be removed from this diff, I suppose?

Makefile
```diff
@@ -636,7 +598,7 @@ ifeq ($(IS_WINDOWS), 1)
 else
 	cp $(TARGET_DIR)/wasmer package/bin/
 ifeq ($(IS_DARWIN), 1)
-	codesign -s - package/bin/wasmer
+	# codesign -s - package/bin/wasmer || true
```
Hywan, line 601: Same, I expect this to be removed from this diff, right?

lib/c-api/tests/wasm.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include "../wasm.h"
+#include "../wasmer_wasm.h"
+#include <stdio.h>
+#include <string.h>
+
+wasm_engine_t *wasm_engine_new() {
+  wasm_config_t *config = wasm_config_new();
+
+  char *wasmer_test_compiler = getenv("TEST");
```
Hywan, line 15: Can we rename this variable as `WASMER_C_API_TEST_CONFIGURATION`, or something like this; because just `TEST` is meaningless :-). Thanks!

tests/compilers/native_functions.rs
```diff
@@ -77,34 +76,34 @@ fn native_function_works_for_wasm() -> Result<()> {
...
+// )]
+// #[compiler_test(native_functions)]
+// fn native_with_env_host_function_closure_panics(config: crate::Config) {
+//     let store = config.store();
+//     let state = 3;
+//     let env = 4;
+//     Function::new_native_with_env(&store, env, move |_env: &i32, _: i32| {
+//         println!("{}", state);
+//     });
+// }
```
Hywan, line 102: Why are they commented now?

Makefile
```diff
@@ -513,22 +507,10 @@ test-cranelift: $(foreach cranelift_engine,$(filter cranelift-%,$(compilers_engi
...
+	cargo test --all
 	cargo test --manifest-path lib/compiler-cranelift/Cargo.toml --release --no-default-features --features=std
 	cargo test --manifest-path lib/compiler-singlepass/Cargo.toml --release --no-default-features --features=std
-	cargo test --manifest-path lib/cli/Cargo.toml $(compiler_features) --release
-	cargo test -p wasmer-cache --release
-	cargo test -p wasmer-engine --release
-	cargo test -p wasmer-derive --release
-	cargo check --manifest-path fuzz/Cargo.toml $(compiler_features) --release
-
+	cargo test --manifest-path lib/cli/Cargo.toml $(compiler_features)
```
MarkMcCaskey, line 513: Is removing `--release` intentional here?

The reason I ask is that debug builds are entirely separate and it means we'll recompile everything at least twice if we use both debug and release builds

Makefile
```diff
@@ -513,22 +507,10 @@ test-cranelift: $(foreach cranelift_engine,$(filter cranelift-%,$(compilers_engi
...
-	cargo test -p wasmer-engine-jit --release --no-default-features
-	cargo test -p wasmer-compiler --release
+	cargo test --all
 	cargo test --manifest-path lib/compiler-cranelift/Cargo.toml --release --no-default-features --features=std
 	cargo test --manifest-path lib/compiler-singlepass/Cargo.toml --release --no-default-features --features=std
-	cargo test --manifest-path lib/cli/Cargo.toml $(compiler_features) --release
-	cargo test -p wasmer-cache --release
-	cargo test -p wasmer-engine --release
-	cargo test -p wasmer-derive --release
-	cargo check --manifest-path fuzz/Cargo.toml $(compiler_features) --release
```
MarkMcCaskey, line 530: I believe `fuzz` is not part of the workspace so is now not being tested with this change

lib/c-api/tests/wasm.h
MarkMcCaskey: I agree, also `TEST` has a high risk of collision, having `WASMER_` in front reduces the chances of random looking breakage in case that env var already exists

tests/compilers/config.rs
```diff
@@ -0,0 +1,146 @@
...
+                    engine = engine.features(features.clone())
+                }
+                Box::new(engine.engine())
+            }
+            #[cfg(not(any(feature = "native", feature = "jit")))]
+            engine => panic!(
+                "The {:?} Engine is not enabled. Please enable it using the features",
+                engine
+            ),
+        }
```
MarkMcCaskey, line 85: The `cfg` logic doesn't look entirely correct here, if you try to compile only one engine, I'd expect a compiler error about inexhaustive match. Perhaps change `not(any(engine))` into `not(all(engine))`

tests/compilers/config.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    pub fn engine_headless(&self) -> Box<dyn WasmerEngine> {
+        match &self.engine {
+            #[cfg(feature = "native")]
+            Engine::Native => Box::new(wasmer_engine_native::Native::headless().engine()),
+            #[cfg(feature = "jit")]
+            Engine::JIT => Box::new(wasmer_engine_jit::JIT::headless().engine()),
+            #[cfg(not(any(feature = "native", feature = "jit")))]
```
MarkMcCaskey, line 94: same here

tests/compilers/config.rs
```diff
@@ -0,0 +1,146 @@
...
+            }
+            #[cfg(feature = "singlepass")]
+            Compiler::Singlepass => {
+                let mut compiler = wasmer_compiler_singlepass::Singlepass::new();
+                compiler.canonicalize_nans(canonicalize_nans);
+                compiler.enable_verifier();
+                self.add_middlewares(&mut compiler);
+                Box::new(compiler)
+            }
+            #[allow(dead_code)]
```
MarkMcCaskey, line 131: I prefer this 👍

tests/lib/compiler-test-derive/src/ignores.rs
```diff
@@ -0,0 +1,195 @@
...
+        let file = File::open(path).unwrap();
+        let reader = BufReader::new(file);
+        let mut patterns = Vec::new();
+
+        for line in reader.lines() {
+            let line = line.unwrap();
+            // If the line has a `#` we discard all the content that comes after
+            let line = if line.contains('#') {
+                let l: Vec<&str> = line.split('#').collect();
+                l.get(0).unwrap().to_string()
```
MarkMcCaskey, line 78: ⛳ `l.get(0).unwrap()` -> `l[0]`

tests/lib/compiler-test-derive/src/ignores.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub fn build_from_path(path: PathBuf) -> Ignores {
+        let file = File::open(path).unwrap();
+        let reader = BufReader::new(file);
+        let mut patterns = Vec::new();
+
+        for line in reader.lines() {
+            let line = line.unwrap();
+            // If the line has a `#` we discard all the content that comes after
+            let line = if line.contains('#') {
+                let l: Vec<&str> = line.split('#').collect();
```
MarkMcCaskey, line 77: New method just for this use case... [`split_once`](https://doc.rust-lang.org/std/string/struct.String.html#method.split_once)... unfortunately  it just stabilized in 1.52, so I suppose we can't use it yet

tests/lib/compiler-test-derive/src/ignores.rs
```diff
@@ -0,0 +1,195 @@
...
+                            os = Some(alias.to_string());
+                        }
+                        "jit" | "native" => {
+                            engine = Some(alias.to_string());
+                        }
+                        "cranelift" | "llvm" | "singlepass" => {
+                            compiler = Some(alias.to_string());
+                        }
+                        other => {
+                            panic!("Alias {:?} not currently supported", other);
```
MarkMcCaskey, line 108: Minor point, it might be nice to keep a counter of which line number we're looking at so we can report it in errors. Not high priority, but it's a nice thing to have

tests/lib/compiler-test-derive/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+    let ignores = crate::ignores::Ignores::build_from_path(ignores_txt_path);
+
+    let should_ignore = |test_name: &str, compiler_name: &str, engine_name: &str| {
+        let compiler_name = compiler_name.to_lowercase();
+        let engine_name = engine_name.to_lowercase();
+        // We construct the path manually because we can't get the
+        // source_file location from the `Span` (it's only available in nightly)
+        let full_path = format!(
+            "{}::{}::{}::{}",
+            quote! { #path },
```
MarkMcCaskey, line 61: This is surprising to me, I thought `quote!` was for making ASTs of Rust tokens; does this do the right thing? If I ran across this code in the wild, I'd be very appreciative of a comment explaining the idea here because it looks like it shouldn't work, but I'm guessing it does

MarkMcCaskey (commented):
Looks very good! Great job with the design of this and getting things working

tests/lib/compiler-test-derive/src/ignores.rs
MarkMcCaskey: This pattern appears a number of times but I only called it out once. It's slightly simpler but semantically the same. `get(0).unwrap()` is more explicit to the reader about the potential to failure though (by being more text and grabbing more attention), but otherwise they're the same

tests/lib/compiler-test-derive/src/ignores.rs
MarkMcCaskey: Oh, also, if you just want the first thing, you can just call `.next().unwrap()` rather than building a Vec and ignoring the rest

tests/lib/compiler-test-derive/src/lib.rs
syrusakbary: It took me a bit to get this working to be honest. It was not trivial to pass from a TokenStream to a string actually.
But yeah, I fully tested it and it's working :)

tests/lib/compiler-test-derive/src/ignores.rs
syrusakbary: We can do `lsplit(, 2)` that would be probably better

tests/lib/compiler-test-derive/src/ignores.rs
syrusakbary: Good call

Makefile
syrusakbary: Yup, published by mistake. Good call!

lib/c-api/tests/wasm.h
syrusakbary: `WASMER_C_API_TEST_CONFIGURATION` is like insanely large. I usually was testing by hand and typing that much was a pain. But I can prefix it with `WASMER_CAPI_CONFIG`

Makefile
```diff
@@ -489,22 +483,22 @@ build-capi-headless-all: capi-setup
...
-	cargo test --release --tests $(compiler_features) --features "test-cranelift test-jit"
+	cargo test --release --tests $(compiler_features) -- cranelift::jit
 
 test-llvm-native:
-	cargo test --release --tests $(compiler_features) --features "test-llvm test-native"
+	cargo test --release --tests $(compiler_features) -- llvm::native
 
 test-llvm-jit:
-	cargo test --release --tests $(compiler_features) --features "test-llvm test-jit"
+	cargo test --release --tests $(compiler_features) -- llvm::jit
```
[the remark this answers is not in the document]
syrusakbary: Agreed, I'll preserve this commands in case they are useful. But when we do global testing `make test` will do it performantly!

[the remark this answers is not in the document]
syrusakbary: Actually, looking back at it is not that trivial. I'll try to do it but it might not work because of some implications on the Makefile. If it doesn't work we can do that in a PR apart

Makefile
syrusakbary: Let's add it then! :)

tests/lib/compiler-test-derive/src/ignores.rs
syrusakbary: Done!

Makefile
```diff
@@ -513,22 +507,10 @@ test-cranelift: $(foreach cranelift_engine,$(filter cranelift-%,$(compilers_engi
...
 test-packages:
-	cargo test -p wasmer --release
-	cargo test -p wasmer-vm --release
-	cargo test -p wasmer-types --release
-	cargo test -p wasmer-wasi --release
-	cargo test -p wasmer-object --release
-	cargo test -p wasmer-engine-native --release --no-default-features
-	cargo test -p wasmer-engine-jit --release --no-default-features
-	cargo test -p wasmer-compiler --release
+	cargo test --all
```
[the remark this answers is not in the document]
syrusakbary: 1. Done
2. Fixed
3. I don't think testing without the default features makes sense right now. It might make sense to do it once we add `no-std` support, but since all the other features are additive testing should be ok for all :)

tests/lib/compiler-test-derive/src/ignores.rs
syrusakbary: Fixed

tests/compilers/config.rs
syrusakbary: I adapted the other code!

tests/compilers/native_functions.rs
syrusakbary: It was because the `compiler_test` macro didn't support properly the function decorators.
I just added support and put the code as it was!

lib/c-api/tests/wasm.h
syrusakbary: Fixed

Makefile
syrusakbary: Fixed

lib/engine-object-file/src/builder.rs
```diff
@@ -76,13 +76,13 @@ mod tests {
     #[cfg(feature = "compiler")]
     use std::sync::Arc;
     #[cfg(feature = "compiler")]
-    use wasmer_compiler::{Compiler, FunctionMiddlewareGenerator};
+    use wasmer_compiler::{Compiler, ModuleMiddleware};
```
syrusakbary, line 79: This was actually not being tested, so I caught the bug when testing in all packages :)

syrusakbary:
Linting is failing because an external package (which is weird).
Other than that there are some multithreading issues that now arise because tests are being run in parallel (but using the same `static HITS` variable):

```
    imports::dynamic_function::cranelift::jit
    imports::dynamic_function::cranelift::native
    imports::dynamic_function::llvm::native
    imports::dynamic_function::singlepass::jit
    imports::static_function::cranelift::native
    imports::static_function::llvm::jit
    imports::static_function::llvm::native
    imports::static_function::singlepass::jit
    imports::static_function_with_results::cranelift::native
    imports::static_function_with_results::llvm::jit
    imports::static_function_with_results::llvm::native
    imports::static_function_with_results::singlepass::jit
```

If you can take a look @MarkMcCaskey to help me resolve this that would be great.

bors try

MarkMcCaskey:
bors try

MarkMcCaskey (approved):
looks good to me!

syrusakbary:
DESKTOP-VN7S460

syrusakbary:
Merging manually as I disabled the failinc c-api tests in Windows by hand

PR #2325, opened 2021-05-19
Clean up some internal WASI code

MarkMcCaskey:
Shout out to @Earthmark on GitHub for the suggestions

Comments left on #2247, I completely forgot about the PR and didn't see any of the comments until this morning when I saw that the PR had been shipped.

-----

This PR addresses some community feedback by rewording a doc comment and updating the internals of a function using 3 mutable variables to find the best fit to only use 1 mutable variable. This change increases readability of the code by making it obvious that all 3 variables are conveying the same bit of information: that we found a match.

### DIFF
--- a/lib/wasi/src/state/mod.rs
+++ b/lib/wasi/src/state/mod.rs
@@ -816,7 +816,7 @@ impl WasiFs {
                                 symlink_count += 1;
                                 Kind::Symlink {
                                     base_po_dir: pre_open_dir_fd,
-                                    path_to_symlink: relative_path,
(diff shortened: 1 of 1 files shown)


Hywan (approved):
Looks good to me :-). Thanks!

Earthmark:
Thank you that is much clearer for me to understand, the comment guides me to the right understanding and the internal flow now shows me that the three values are bound!
I also didn't realize you could declare enums inside a function, in my mind the solution was an optional tuple but your solution is much more straightforward to read.

Hope y'all have a good day

PR #2326, opened 2021-05-19
Simplify `InstanceRef`

MarkMcCaskey:
I think the reason we didn't do this originally is that [`Arc::as_ptr`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.as_ptr) was only stabilized in 1.45. Either way, this removes a lot of very complicated code and replaces it with much simpler code.

I was looking into something else but I saw `Arc<AtomicUsize>` and wondered why we don't just drop the `AtomicUsize` entirely and just put everything in the `Arc`... this is the result of that.

# Review

### DIFF
--- a/lib/vm/src/instance/ref.rs
+++ b/lib/vm/src/instance/ref.rs
@@ -6,7 +6,11 @@ use std::ptr::{self, NonNull};
 use std::sync::Arc;
 
 /// Dynamic instance allocation.
+///
+/// This structure has a C representation because `Instance` is
+/// dynamically-sized, and the `instance` field must be last.
 #[derive(Debug)]
+#[repr(C)]
 struct InstanceInner {
     /// The layout of `Instance` (which can vary).
     instance_layout: Layout,
@@ -37,14 +41,7 @@ impl InstanceInner {
         ptr::drop_in_place(instance_ptr);
         std::alloc::dealloc(instance_ptr as *mut u8, self.instance_layout);
     }
-    // TODO: do we need this?
-    /*
-    /// Get the number of strong references pointing to this
-    /// `InstanceRef`.
-    pub fn strong_count(&self) -> usize {
-        self.strong.load(atomic::Ordering::SeqCst)
-    }
-    */
+
     /// Get a reference to the `Instance`.
     #[inline]
     pub(crate) fn as_ref(&self) -> &Instance {
@@ -112,11 +109,7 @@ impl MemoryUsage for InstanceInner {
 /// share an [`Instance`] between an [`InstanceHandle`] and the module
 /// exports, so that one can drop a [`InstanceHandle`] but still being
 /// able to use the exports properly.
-///
-/// This structure has a C representation because `Instance` is
-/// dynamically-sized, and the `instance` field must be last.
 #[derive(Debug, PartialEq, Clone, MemoryUsage)]
-#[repr(C)]
 pub struct InstanceRef(Arc<InstanceInner>);
 
 impl InstanceRef {
--- a/lib/vm/src/instance/ref.rs
+++ b/lib/vm/src/instance/ref.rs
@@ -9,6 +9,11 @@ use std::sync::Arc;
 ///
 /// This structure has a C representation because `Instance` is
 /// dynamically-sized, and the `instance` field must be last.
+///
+/// This `InstanceRef` must be freed with [`InstanceInner::deallocate_instance`]
+/// if and only if it has been set correctly. The `Drop` implementation of
+/// [`InstanceInner`] calls its `deallocate_instance` method without
+/// checking if this property holds, only when `Self.strong` is equal to 1.
 #[derive(Debug)]
 #[repr(C)]
 struct InstanceInner {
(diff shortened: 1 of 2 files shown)


lib/vm/src/instance/ref.rs
```diff
@@ -109,114 +52,102 @@ impl InstanceRef {
...
+    }
+
+    /// Get a reference to the `Instance`.
+    #[inline]
+    pub(crate) fn as_ref(&self) -> &Instance {
+        (&*self.0).as_ref()
+    }
+
+    #[inline]
+    /// Only succeeds if ref count is 1.
```
Hywan, line 140:
```suggestion
    /// Only succeeds if ref count is 1.
    #[inline]
```

lib/vm/src/instance/ref.rs
```diff
@@ -109,114 +52,102 @@ impl InstanceRef {
...
+    pub(super) fn as_mut(&mut self) -> Option<&mut Instance> {
+        Some(Arc::get_mut(&mut self.0)?.as_mut())
+    }
+
+    #[inline]
+    /// # Safety
+    /// It is the caller's responsibility to ensure exclusivity and synchronization of the
+    /// instance.
```
Hywan, line 148:
```suggestion
    /// # Safety
    /// It is the caller's responsibility to ensure exclusivity and synchronization of the
    /// instance.
    #[inline]
```

Hywan (approved):
I believe it's correct! Thanks for the clean up. That's great that `Arc::as_ref` is now stable, wasn't aware of that.

lib/vm/src/instance/ref.rs
```diff
@@ -109,114 +52,102 @@ impl InstanceRef {
...
+///
+/// Consequently, one must not share `Instance` but
+/// `InstanceRef`. It acts like an Atomically Reference Counter
+/// to `Instance`. In short, `InstanceRef` is roughly a
+/// simplified version of `std::sync::Arc`.
+///
+/// This `InstanceRef` must be freed with [`InstanceRef::deallocate_instance`]
+/// if and only if it has been set correctly. The `Drop` implementation of
+/// [`InstanceRef`] calls its `deallocate_instance` method without
+/// checking if this  property holds, only when `Self.strong` is equal to 1.
```
Hywan, line 102: This part of the comment must be updated please :-)!

lib/vm/src/instance/ref.rs
```diff
@@ -109,114 +52,102 @@ impl InstanceRef {
...
+        Some(Arc::get_mut(&mut self.0)?.as_mut())
+    }
+
+    #[inline]
+    /// # Safety
+    /// It is the caller's responsibility to ensure exclusivity and synchronization of the
+    /// instance.
+    pub(super) unsafe fn as_mut_unchecked(&mut self) -> &mut Instance {
+        let ptr: *mut InstanceInner = Arc::as_ptr(&self.0) as *mut _;
+        (&mut *ptr).as_mut()
```
Hywan, line 151: Two possibilities here.

1️⃣  The first one is the following:

```suggestion
        ptr.as_mut().unwrap().as_mut()
```

not tested, but I was thinking about [`ptr::as_mut`](https://doc.rust-lang.org/std/primitive.pointer.html#method.as_mut)

Hywan, line 151: 2️⃣ The second possibility is the following:

```suggestion
        Arc::as_mut(&self.0).as_mut()
```

Hywan: 3️⃣  There is another possibility!

```suggestion
        self.as_mut().unwrap()
```

Thoughts?

MarkMcCaskey: We can't use `as_mut` because `as_mut` can fail. This function can never fail, I'll look into the other way though

Hywan: Then possibility number 2 seems appropriate: https://github.com/wasmerio/wasmer/pull/2326#discussion_r636045237

MarkMcCaskey: For what it's worth, the unstable method https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut_unchecked (what we really want) is very similar internally to what the code is currently doing, so I think it's correct as-is.

lib/vm/src/instance/ref.rs
MarkMcCaskey: Thanks!  For sure on the bottom part, I actually thought about removing the 

> In short, `InstanceRef` is roughly a simplified version of `std::sync::Arc`.</span></span>

part because it's not true in implementation anymore but it's still true in API, which is arguably more important for the doc comment to express. `InstanceRef`'s methods are pretty much just a subset of `Arc`'s methods.

lib/vm/src/instance/ref.rs
Hywan: Sorry, it's `Arc::get_mut`, https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut

PR #2327, opened 2021-05-19
Fix cyclic ref in WasmerEnv

### LINKED ISSUE #2304: LazyInit<Memory> does not call munmap

MarkMcCaskey:
This is an alternative to #2312 to fix #2304 

The core idea is to have a `WeakInstanceRef` too. The main concern is that `Memory` (from a `WasmerEnv`) should behave like it has a strong instance ref if it's cloned. Currently we always convert weak `VMMemory`s into strong ones and if we can't, we panic.

1 other edge case not handled yet: we need to make sure that when the `InstanceRef` is weak that everything still works as expected (for example when the `Instance` it points to is gone). Or maybe we don't. `WasmerEnv`s are always passed by `&` and `Clone`ing forces them to upgrade into `Strong` or panic. So there's no real case where this can happen except for when you're calling into a host function from Wasm while the Instance is freed. Doing that means that you're already executing uninitialized memory as code so it shouldn't be possible / if it is, the concern is not with`WasmerEnv`.

# Review


lib/vm/src/export.rs
```diff
@@ -52,7 +52,23 @@ pub struct VMFunction {
...
+        Self {
+            address: self.address.clone(),
+            vmctx: self.vmctx.clone(),
+            signature: self.signature.clone(),
+            kind: self.kind.clone(),
+            call_trampoline: self.call_trampoline.clone(),
+            instance_ref: self
+                .instance_ref
+                .as_ref()
+                .map(|v| WeakOrStrongInstanceRef::Strong(v.get_strong().unwrap())),
```
syrusakbary, line 69: Cloning it automatically make ti strong?

lib/api/tests/export.rs
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        let f: NativeFunc<(), ()> = instance.exports.get_native_function("call_host_fn")?;
+        f.call()?;
+        f
+    };
+    f.call()?;
+
+    Ok(())
+}
```
syrusakbary, line 264: We are testing the strong/weak behavior directly (`is_strong_instance_ref`) which is great.

But can we have an extra test that verifies the behavior is fixed without checking for strong/weak fields? (that means, without assuming the design the API)

lib/vm/src/instance/ref.rs
```diff
@@ -155,3 +155,75 @@ impl InstanceRef {
...
+    /// Get a strong `InstanceRef`. A return Value of `None` means that the
+    /// `InstanceRef` has been freed and cannot be accessed.
+    pub fn get_strong(&self) -> Option<InstanceRef> {
+        match self {
+            Self::Weak(weak) => weak.upgrade(),
+            Self::Strong(strong) => Some(strong.clone()),
+        }
+    }
+    /// Get a weak `InstanceRef`.
+    pub fn get_weak(&self) -> WeakInstanceRef {
```
syrusakbary, line 212: This method is not used outside this Trait, can we make it private?

lib/api/src/externals/global.rs
```diff
@@ -208,6 +208,13 @@ impl Global {
     pub fn same(&self, other: &Self) -> bool {
         Arc::ptr_eq(&self.vm_global.from, &other.vm_global.from)
     }
+
+    /// Check if the global holds a strong `InstanceRef`.
+    /// None means there's no `InstanceRef`, strong or weak.
+    // TODO: maybe feature gate this, we only need it for tests...
```
syrusakbary, line 214: This is only used in tests. As such I would expect to only be visible there (can we delete it from here and move the logic somewhere else?)

lib/api/src/externals/table.rs
```diff
@@ -146,6 +146,13 @@ impl Table {
     pub fn same(&self, other: &Self) -> bool {
         Arc::ptr_eq(&self.vm_table.from, &other.vm_table.from)
     }
+
+    /// Check if the table holds a strong `InstanceRef`.
+    /// None means there's no `InstanceRef`, strong or weak.
+    // TODO: maybe feature gate this, we only need it for tests...
```
syrusakbary, line 152: Same as other comments

lib/api/src/externals/memory.rs
```diff
@@ -248,6 +248,13 @@ impl Memory {
     pub fn same(&self, other: &Self) -> bool {
         Arc::ptr_eq(&self.vm_memory.from, &other.vm_memory.from)
     }
+
+    /// Check if the memory holds a strong `InstanceRef`.
+    /// None means there's no `InstanceRef`, strong or weak.
+    // TODO: maybe feature gate this, we only need it for tests...
```
syrusakbary, line 254: Same as other comments

lib/api/src/externals/function.rs
```diff
@@ -712,6 +712,17 @@ impl Function {
     fn closures_unsupported_panic() -> ! {
         unimplemented!("Closures (functions with captured environments) are currently unsupported with native functions. See: https://github.com/wasmerio/wasmer/issues/1840")
     }
+
+    /// Check if the function holds a strong `InstanceRef`.
+    /// None means there's no `InstanceRef`, strong or weak.
+    // TODO: maybe feature gate this, we only need it for tests...
```
syrusakbary, line 718: Same as other comments

lib/api/src/externals/mod.rs
```diff
@@ -72,6 +72,15 @@ impl<'a> Exportable<'a> for Extern {
         // Since this is already an extern, we can just return it.
         Ok(_extern)
     }
+
+    fn into_weak_instance_ref(&mut self) {
```
syrusakbary, line 76: I'm not super fan of this function name. I think we can just have `into_weak` (the API user should not care about how is implemented?)

lib/api/src/native.rs
```diff
@@ -196,6 +196,12 @@ macro_rules! impl_native_traits {
                 }
 
             }
+            /// Check if the function holds a strong `InstanceRef`.
+            /// None means there's no `InstanceRef`, strong or weak.
+            // TODO: maybe feature gate this, we only need it for tests...
```
syrusakbary, line 201: Same as other comments

lib/vm/src/instance/ref.rs
```diff
@@ -155,3 +155,75 @@ impl InstanceRef {
...
+    /// Convert the existing type directly into a weak reference.
+    pub fn into_weak(&mut self) {
+        let new = self.get_weak();
+        *self = Self::Weak(new);
+    }
+
+    /// Check if the reference contained is strong.
+    pub fn is_strong(&self) -> bool {
+        matches!(self, WeakOrStrongInstanceRef::Strong(_))
+    }
```
syrusakbary, line 228: This is only used in tests. I believe we can move the logic out of this trait (perhaps into the tests?)

lib/vm/src/export.rs
syrusakbary: It's a bit weird this behavior. Perhaps we want to be more explicit having a custom method for this saying: `into_strong` that makes it strong.

lib/vm/src/instance/ref.rs
```diff
@@ -155,3 +155,75 @@ impl InstanceRef {
...
+    pub fn upgrade(&self) -> Option<InstanceRef> {
+        let inner = self.0.upgrade()?;
+        Some(InstanceRef(inner))
+    }
+}
+
+impl MemoryUsage for WeakInstanceRef {
+    fn size_of_val(&self, tracker: &mut dyn MemoryUsageTracker) -> usize {
+        mem::size_of_val(self)
+            + if let Some(ir) = self.upgrade() {
```
syrusakbary, line 182: Getting the size of a value makes it strong? That seems wrong.

lib/vm/src/instance/ref.rs
```diff
@@ -155,3 +155,75 @@ impl InstanceRef {
...
+    /// A weak instance ref.
+    Weak(WeakInstanceRef),
+    /// A strong instance ref.
+    Strong(InstanceRef),
+}
+
+impl WeakOrStrongInstanceRef {
+    /// Get a strong `InstanceRef`. A return Value of `None` means that the
+    /// `InstanceRef` has been freed and cannot be accessed.
+    pub fn get_strong(&self) -> Option<InstanceRef> {
```
syrusakbary, line 205: `get_strong` is always used as `WeakOrStrongInstanceRef::Strong(v.get_strong().unwrap())`.

Because of this, I think it might be better to have a `try_into_strong()` method that returns a `Option<WeakOrStrongInstanceRef>` and will also simplify this code.

lib/api/tests/export.rs
MarkMcCaskey: It's not exactly easy to test because it's a memory leak test, it is possible to run some logic in a loop and if it crashes due to OOM then it triggers. My concern about that is that changes to the system (like how much virtual / real memory is available) may change when/if that test passes / fails.

lib/vm/src/export.rs
MarkMcCaskey: Yeah, was unsure to what extent we want to encapsulate that logic, but this is the secret sauce that makes this solution invisible to the user and behave correctly in every situation. Weak externs automatically become strong when they're shared (via cloning)

lib/api/src/externals/global.rs
MarkMcCaskey: Unfortunately it can't be, these accessors access private info. It has to be a method on the type

lib/api/src/externals/mod.rs
MarkMcCaskey: Yeah, the reason for the extra precision is due to overlap in ownership. It's not really a weak version of the thing it's being called on. That thing can still keep things alive. The only thing being made weak here is an internal `InstanceRef`. So I do think to that extent, it does matter, it's generally not something users should have to interact with unless they're doing low level things.

lib/vm/src/instance/ref.rs
MarkMcCaskey: Yeah, weak references can't be accessed. The only way to look into this type and see how much memory might be   being used is to try to make it into a strong reference. If it can become a strong reference then that memory exists and we can count it, if it's not then we can't.

MarkMcCaskey: And to clarify it doesn't mutate the internal value, it just needs to upgrade the reference in order to count correctly.  This logic should be implemented inside of loupe itself on the `Weak` type, not here.

lib/vm/src/instance/ref.rs
MarkMcCaskey: Yeah, it could be a utility function in tests / in the methods of all the types that use it, it seems like a useful method to have on a type like this though

lib/vm/src/instance/ref.rs
MarkMcCaskey: Yeah it could be, but it seems like a useful method to have on a type like this.

lib/vm/src/instance/ref.rs
MarkMcCaskey: Nice observation about how it's used, we could use `TryInto` or something for these conversions

lib/api/src/externals/global.rs
MarkMcCaskey: One solution to this would be to define a trait with these methods on it and export it in like `wasmer::test` and mark that module as `#[hidden]`so it doesn't show up in docs. Then these methods will only be available if you import `wasmer::test::SomeTrait`

lib/api/src/externals/mod.rs
MarkMcCaskey: Looking at this function again, we probably need to make it unsafe because you can probably get use after free with it. I'll investigate this with a test and maybe update it

lib/api/src/externals/function.rs
```diff
@@ -712,6 +712,18 @@ impl Function {
...
     }
+
+    /// Get access to the backing VM value for this extern. This function is for
+    /// tests it should not be called by users of the Wasmer API.
+    ///
+    /// # Safety
+    /// This function is unsafe to call outside of tests for the wasmer crate
+    /// because there is no stability guarantee for the returned type and we may
+    /// make breaking changes to it at any time or remove this method.
+    #[doc(hidden)]
```
syrusakbary, line 723: Good call on the `unsafe` to discourage usage

lib/api/src/externals/function.rs
```diff
@@ -725,6 +737,26 @@ impl<'a> Exportable<'a> for Function {
...
+    fn clone(&self) -> Self {
+        let mut exported = self.exported.clone();
+        exported.vm_function.upgrade_instance_ref().unwrap();
+
+        Self {
+            store: self.store.clone(),
+            exported,
+        }
+    }
 }
```
syrusakbary, line 760: Love this

syrusakbary (approved):

kaimast:
Thanks for fixing this!

PR #2343, opened 2021-05-25
docs: point to wasmer-zig integration

kubkon:
# Description

Even though @Luukdegram and I just started working on Wasmer integration in Zig, I thought it might be a good idea to point everyone to the project early so that we involve as many contributors as possible, and perhaps, open a nice comms channel between Zig and Wasmer. So, this short PR is doing just that: pointing to [wasmer-zig] integration in the docs.

A couple of words on the integration and the direction we want this to go. Currently, the integration shows off a minimal working build system and example rewritten verbatim from the official Wasmer C API tutorial ([the link]). For the end result, we want this integration to be on par with Rust's, C's, or Go's integration hosted by Wasmer. While we are here, this might be a daft question so excuse me if it is, but is there any reason why you prefer exposing `wasm.h` primitives such as `wasm_instance_t` in your integration rather than `wasmer_instance_t`? In general, what would be the ideal direction to take here in your opinion? Currently, wasmer-zig exposes the latter and I'm wondering if we should expose one or the other, or (probably ideally) both?

[wasmer-zig]: https://github.com/kubkon/wasmer-zig
[the link]: https://wasmerio.github.io/wasmer/c/

# Review

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ qjs >
 | ![R logo] | [**R**][R integration] | *no published package* | [Docs][r docs] |
 | ![Postgres logo] | [**Postgres**][Postgres integration] | *no published package* | [Docs][postgres docs] |
 |  | [**Swift**][Swift integration] | *no published package* | |
+| ![Zig logo] | [**Zig**][Zig integration] | *no published package* | |
 
 [👋 Missing a language?](https://github.com/wasmerio/wasmer/issues/new?assignees=&labels=%F0%9F%8E%89+enhancement&template=---feature-request.md&title=)
 
@@ -169,6 +170,9 @@ qjs >
 [postgres docs]: https://github.com/wasmerio/wasmer-postgres#usage--documentation
 
 [swift integration]: https://github.com/AlwaysRightInstitute/SwiftyWasmer
+
+[zig logo]: https://raw.githubusercontent.com/ziglang/logo/master/zig-favicon.png
+[zig integration]: https://github.com/kubkon/wasmer-zig
 
 ## Contribute
 


kubkon:
> We are moving away from the old API (https://wasmerio.github.io/wasmer/c/) to adopt the new standard Wasm-C-API.
> We need to update the doc URL that you pointed to the newer API, so things match regarding the docs (and we don't point people to the old API).
> 
> Most of the Wasm-C-API should be fully covered and working:
> 
> * Examples: https://github.com/wasmerio/wasmer/blob/master/lib/c-api/examples/
> * Examples in Docs: https://docs.wasmer.io/integrations/examples (check "C/C++" tab)
> * Wasm-C-API examples: https://github.com/wasmerio/wasmer/tree/master/lib/c-api/tests/wasm-c-api/example

Ah, gotcha, so I was looking at the old API. Thanks for the links, I'll make sure we implement the new API which seems to match that I see in Go integration for instance. Nice.

syrusakbary (approved):

syrusakbary:
Merging manually as we don't need bors for docs changes

README.md
```diff
@@ -170,6 +171,9 @@ qjs >
 
 [swift integration]: https://github.com/AlwaysRightInstitute/SwiftyWasmer
 
+[zig logo]: https://raw.githubusercontent.com/ziglang/logo/master/zig-favicon.png
+[zig integration]: https://github.com/kubkon/wasmer-zig
```
Hywan, line 175: Why not using https://github.com/zigwasm/wasmer-zig?

Hywan: Ah, it's been probably moved recently.

Hywan:
Thanks @kubkon, that's awesome! If you need any help, please ping me, I'll be happy to help you, as @MarkMcCaskey and I are maintaining the C API of Wasmer, based on the standard Wasm C API.

README.md
kubkon: Yep, correct! We introduced `zigwasm` org only today. I'll push an update to Wasmer's README shortly.

kubkon:
> Thanks @kubkon, that's awesome! If you need any help, please ping me, I'll be happy to help you, as @MarkMcCaskey and I are maintaining the C API of Wasmer, based on the standard Wasm C API.

Thanks, I definitely will!

PR #2348, opened 2021-05-26
Copied trap handlers so this will compile for `aarch64-linux-android`

dgrijalva-squire:
Attempting to compile Wasmer for android fails with an 'unsupported platform' error. These two additional cases are copied from their neighbors, but properly match against the `aarch64-linux-android` target. I don't know enough about what this code is actually doing to know if this is 'right', but it does appear to work.


syrusakbary:
Looks right! Merging

syrusakbary (approved):

PR #2358, opened 2021-05-28
feat: Attempt to help the users when migrating

Hywan:
This PR is a bad attempt to help the users when migrating from 1.x to 2.x. Some features have been renamed, like `jit` to `universal`, and `native` to `dylib`.

We have several solutions:

1. Either we generate a `#[deprecated]` message, but we need to alias many types (`pub type JIT = Universal; pub type JITEngine = UniversalEngine;` etc.), and to raise `#[deprecated]` notes when used. Note that it's not possible to raise a deprecation note if the type isn't used. And it's not possible to raise a deprecation note for a feature being used only (at least without the help of dirty hacks),
2. Generate a compile error that guides the user to migrate.

This PR addresses the problem with the solution number 2. But I don't like it.

In an ideal world, every crate inside Wasmer must provide a `migrations` module (or `versions`, or `changes`, or `semver`, or `changelog`, … whatever!) which itself contains modules like `1_x_to_2_x.rs`. Each of those modules would create the type aliasing, bridges between one version to another if possible, and more importantly, would contain the documentation to guide the users to migrate from one version to another. This documentation would be part of the official crate documentation, so will be readable at docs.rs for example.

What do you think?

### DIFF
--- a/lib/api/src/lib.rs
+++ b/lib/api/src/lib.rs
@@ -370,5 +370,18 @@ pub use wasmer_engine_universal::{Universal, UniversalArtifact, UniversalEngine}
 #[cfg(feature = "dylib")]
 pub use wasmer_engine_dylib::{Dylib, DylibArtifact, DylibEngine};
 
+#[cfg(any(feature = "jit", feature = "native"))]
+compile_error!(
+    r#"At least one deprecated feature has been used, and cannot not
+be used any longer. Please refer to the following table:
+
+| Deprecated feature | New feature to use instead |
+|--------------------|----------------------------|
+| `jit`              | `universal`                |
+| `native`           | `dylib`                    |
+
+"#
+);
+
 /// Version number of this crate.
 pub const VERSION: &str = env!("CARGO_PKG_VERSION");
--- a/lib/c-api/src/lib.rs
+++ b/lib/c-api/src/lib.rs
@@ -43,3 +43,16 @@ pub mod deprecated;
 pub mod error;
 mod ordered_resolver;
 pub mod wasm_c_api;
+
+#[cfg(any(feature = "jit", feature = "native"))]
+compile_error!(
+    r#"At least one deprecated feature has been used, and cannot not
+be used any longer. Please refer to the following table:
+
+| Deprecated feature | New feature to use instead |
+|--------------------|----------------------------|
+| `jit`              | `universal`                |
+| `native`           | `dylib`                    |
+
+"#
+);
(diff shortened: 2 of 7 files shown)


lib/api/src/lib.rs
```diff
@@ -370,5 +370,18 @@ pub use wasmer_engine_universal::{Universal, UniversalArtifact, UniversalEngine}
...
+be used any longer. Please refer to the following table:
+
+| Deprecated feature | New feature to use instead |
+|--------------------|----------------------------|
+| `jit`              | `universal`                |
+| `native`           | `dylib`                    |
+
+"#
+);
+
```
syrusakbary, line 385: We don't need to make of this a breaking change (making people to update their code mandatory).

We can actually do it in a better way. How?
Since we are just renaming we can do:

```rust
#[cfg(feature = "jit")]
#[deprecated(
    since = "1.1.0",
    note = "Please use the `universal` feature instead"
)]
pub type JIT = Universal;

#[cfg(feature = "native")]
#[deprecated(
    since = "1.1.0",
    note = "Please use the `native` feature instead"
)]
pub type Native = Dylib;

```

lib/c-api/src/lib.rs
```diff
@@ -43,3 +43,16 @@ pub mod deprecated;
...
+    r#"At least one deprecated feature has been used, and cannot not
+be used any longer. Please refer to the following table:
+
+| Deprecated feature | New feature to use instead |
+|--------------------|----------------------------|
+| `jit`              | `universal`                |
+| `native`           | `dylib`                    |
+
+"#
+);
```
syrusakbary, line 58: Same as suggestion inside `api`.

lib/cli/src/lib.rs
```diff
@@ -30,3 +30,16 @@ pub mod utils;
...
+    r#"At least one deprecated feature has been used, and cannot not
+be used any longer. Please refer to the following table:
+
+| Deprecated feature | New feature to use instead |
+|--------------------|----------------------------|
+| `jit`              | `universal`                |
+| `native`           | `dylib`                    |
+
+"#
+);
```
syrusakbary, line 45: We should add a `--jit` and `--native` flag inside clap structs and mark them as deprecated there (similar as we do with the backends).

lib/api/src/lib.rs
Hywan: It's more than that, it's also about the `*Artifact` type, the `*Engine` type, the `*EngineInner` type etc.

syrusakbary:
Closing in favor of already merged #2377

PR #2356, opened 2021-05-27
feat: Rename `wasmer-engine-native` to `wasmer-engine-dylib`

Hywan:
This (small) patch renames the `wasmer-engine-native` crate into
`wasmer-engine-dylib`. We believe the new name conveys
better the intent of the engine.

I will update the `CHANGELOG` later.

PR #2340, opened 2021-05-25
feat: Rename `wasmer-engine-jit` to `wasmer-engine-universal`

Hywan:
This (small) patch renames the `wasmer-engine-jit` crate into
`wasmer-engine-universal`. We believe the new name conveys
better the intent of the engine.

I will update the `CHANGELOG` later.

PR #2364, opened 2021-05-31
feat: Rename `wasmer-engine-object-file` to `wasmer-engine-staticlib`.

Hywan:
# Description

This PR renames the `wasmer-engine-object-file` crate into `wasmer-engine-staticlib`. We believe it conveys a better meaning of what it does. This PR is similar to its siblings, #2340 and #2356.


Hywan:
bors try

lib/engine-staticlib/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "wasmer-engine-object-file"
+name = "wasmer-engine-staticlib"
 version = "1.0.2"
 authors = ["Wasmer Engineering Team <engineering@wasmer.io>"]
 description = "Wasmer Object File Engine"
```
jubianchi, line 5: Should we change the description here?

jubianchi (approved):

lib/engine-staticlib/Cargo.toml
Hywan: We should indeed! Thanks!

PR #2366, opened 2021-05-31
chore: Removing the `deprecated` crates

Hywan:
# Description

This PR removes the `lib/deprecated/` directory which contains the `wasmer-runtime` and `wasmer-runtime-core` deprecated crates. They were present only to transition from 0.x to 1.x. With the coming 2.x version, we no longer need them.


Hywan:
bors try

jubianchi (approved):

PR #2376, opened 2021-06-02
doc(readme) Fix the heading levels

Hywan:
# Description

This patch fixes the heading levels.


syrusakbary:
I did intentionally. It was easier to read based on Github.

You can look on how Deno also uses this strategy.
https://github.com/denoland/deno

Hywan:
Sorry but… it's not because other projects are doing it wrong that it's an excuse. Headings have a semantics. The level has a semantics.

syrusakbary:
I don't think they are doing it wrong. I believe they are doing it right because it improves the readability of their README.

Nothing mandates that after a h1 there needs to be h3 or h4 or h2. It's really about the importance of what is showcased in that section. In this case, I don't think they are h2 sections, but more like h3.

MarkMcCaskey:
Is it possible to use CSS in GitHub markdown? I know they support some HTML. If so, we could style an `##` as an `<h3>`.  I think for readability as HTML (as screen readers do) or as raw markdown as people who clone the repo may, `##` is better than `###`  there, but I find it very understandable that `###` looks better

PR #2383, opened 2021-06-03
Revert update to clap-beta back to structopt

MarkMcCaskey:
Clap-beta hasn't had a release in a while and likely won't for a while longer. We're affected by bugs that have been fixed and not released. We shouldn't be depending on a beta library anyways when there's a stable equivalent. It's very understandable that the clap developers don't want to be burdened with publishing bug fixes when making major changes something in-progress -- they'd probably just end up with way more bug reports overall for things they never intended to support.

Revert #2162 

more context https://github.com/clap-rs/clap/discussions/2487#discussioncomment-821162

# Review

### DIFF
--- a/tests/integration/cli/tests/version.rs
+++ b/tests/integration/cli/tests/version.rs
@@ -1,0 +1,62 @@
+use anyhow::bail;
+use std::process::Command;
+use wasmer_integration_tests_cli::WASMER_PATH;
+
+const WASMER_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+#[test]
+fn version_string_is_correct() -> anyhow::Result<()> {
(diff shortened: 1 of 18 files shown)


tests/integration/cli/tests/version.rs
```diff
@@ -0,0 +1,62 @@
+use anyhow::bail;
+use std::process::Command;
+use wasmer_integration_tests_cli::WASMER_PATH;
+
+const WASMER_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+#[test]
+fn version_string_is_correct() -> anyhow::Result<()> {
```
syrusakbary, line 8: Thanks for this!

syrusakbary (approved):

PR #2386, opened 2021-06-04
fix unwrap error when given a no export functions wasm

chenyukang:
# Description
When the wasm file contains no export functions, an unwrap crash will trigger:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', lib/cli/src/commands/run.rs:344:52
```

This fix will give a more detailed output.

### DIFF
--- a/lib/cli/src/commands/run.rs
+++ b/lib/cli/src/commands/run.rs
@@ -341,13 +341,17 @@ impl Run {
                     let suggested_command = format!(
                         "wasmer {} -i {} {}",
                         self.path.display(),
-                        suggested_functions.get(0).unwrap(),
+                        suggested_functions.get(0).unwrap_or(&String::new()),
                         args.join(" ")
                     );
-                    let suggestion = format!(
-                        "Similar functions found: {}.\nTry with: {}",
-                        names, suggested_command
-                    );
+                    let suggestion = if suggested_functions.len() == 0 {
+                        String::from("Can not find any export functions.")
+                    } else {
+                        format!(
+                            "Similar functions found: {}.\nTry with: {}",
+                            names, suggested_command
+                        )
+                    };
                     match e {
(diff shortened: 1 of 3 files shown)


syrusakbary (approved):
Great catch! Merging!

chenyukang:
@syrusakbary 
I found another issue when adding testcase, my code is not executed when I run: 

```bash 
 cargo test --package wasmer-integration-tests-cli --test run -- run_no_start_wasm_report_error --exact --nocapture
```
This is because WASMER_PATH only uses the release path, but I run it with debug version.

```rust
pub const WASMER_PATH: &str = concat!(
    env!("CARGO_MANIFEST_DIR"),
    "/../../../target/release/wasmer"
);
```
To keep it consistent, we should fix this: 

```rust
#[cfg(feature = "debug")]
pub const WASMER_PATH: &str = concat!(
    env!("CARGO_MANIFEST_DIR"),
    "/../../../target/debug/wasmer"
);

#[cfg(not(feature = "release"))]
pub const WASMER_PATH: &str = concat!(
    env!("CARGO_MANIFEST_DIR"),
    "/../../../target/release/wasmer"
);
```

syrusakbary:
I think that’s a good change/fix! Happy to merge it on this PR as well

MarkMcCaskey (approved):
Thanks!

MarkMcCaskey:
The one thing I'd say, is that the `debug` feature is not set automatically when doing a debug build... So we may want to have a function which checks for both a debug or release build and uses the newest one? Or a more proper fix, maybe the `make test-integration`  makefile command should depend on `build-wasmer`  so it's always there.

I don't think the `debug` feature exists in the the integration tests crate, so I think the `not(feature = "debug")` will be triggered unconditionally

Hywan:
:+1: thanks for the PR!

Issue #2387, opened 2021-06-04
`make install` fails


FGasper:
### Describe the bug

I can build wasmer from the source tree on CentOS 7, but the install fails:

```sh
> make install
-----------
INFORMATION
-----------

/bin/bash: rustup: command not found
Host Target: ``.
Enabled Compilers: cranelift, singlepass.
Testing the following compilers & engines:
* API: cranelift-universal cranelift-dylib singlepass-universal
* C-API: cranelift-universal cranelift-dylib
Cargo features:
* Compilers: `--features cranelift,singlepass`.


--------------
RULE EXECUTION
--------------


install -Dm755 target/release/wasmer /usr/local/bin/wasmer
for header in lib/c-api/*.{h,hh}; do install -Dm644 "$header" /usr/local/include/$(basename $header); done
install: cannot stat ‘lib/c-api/*.hh’: No such file or directory
make: *** [install-capi-headers] Error 1
```

### Steps to reproduce

Run `make install`.

### Expected behavior

It should build then install Wasmer.

### Actual behavior

It installs the main wasmer binary but fails when installing the C API headers.

FGasper:
Additional context:
```sh
> for header in lib/c-api/*.{h,hh}; do echo $header; done
lib/c-api/wasmer.h
lib/c-api/wasmer_wasm.h
lib/c-api/wasm.h
lib/c-api/*.hh
```

PR #2388, opened 2021-06-04
Fix `make install` when *.hh is empty.

FGasper:
Issue #2387

# Description

# Review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -637,7 +637,7 @@ install-wasmer:
 	install -Dm755 target/release/wasmer $(DESTDIR)/bin/wasmer
 
 install-capi-headers:
-	for header in lib/c-api/*.{h,hh}; do install -Dm644 "$$header" $(DESTDIR)/include/$$(basename $$header); done
+	for header in lib/c-api/*.h; do install -Dm644 "$$header" $(DESTDIR)/include/$$(basename $$header); done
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -637,7 +637,7 @@ install-wasmer:
 	install -Dm755 target/release/wasmer $(DESTDIR)/bin/wasmer
 
 install-capi-headers:
-	for header in lib/c-api/*.{h,hh}; do install -Dm644 "$$header" $(DESTDIR)/include/$$(basename $$header); done
+	shopt -s nullglob; for header in lib/c-api/*.{h,hh}; do install -Dm644 "$$header" $(DESTDIR)/include/$$(basename $$header); done
```
syrusakbary, line 640: This will fail in macos (as `shopt` is not shipped there). Perhaps we can remove the `.hh` option for now

FGasper: It's part of the shell on Linux. Interesting.

Hywan, line 640:
```suggestion
	for header in lib/c-api/*.h; do install -Dm644 "$$header" $(DESTDIR)/include/$$(basename $$header); done
```

As we no longer ship any `.h` file, we can simplify this line as suggested.

Hywan (approved):
Looks good to me :-)!

bors r+

PR #2398, opened 2021-06-06
Fix 'variable does not need to be mutable' warnings

imor:
# Description

When I built wasmer on Windows using the `cargo build` command, I got a couple of _the variable doesn't need to be mutable_ warnings. This PR fixes those warnings.

# Review

As this is an internal change, it shouldn't require an entry in the CHANGELOG.md file.


Hywan:
Looks good to me, thanks!

bors r+

Hywan (approved):
bors r+

PR #2411, opened 2021-06-08
feat(wasi-types) Extract types from `wasi` to a new `wasi-types` crate

Hywan:
# Description

This has been discussed with @MarkMcCaskey many times in the past already.

Because we are very likely to re-use those types in other crates (it
already happens in private repos), we want to extract the WASI
types from the `wasi` crate.

This patch extracts the `wasi/src/syscalls/types.rs` module into its
own `wasi-types` crate. This new crate takes this opportunity to
classify the numerous types in submodules. All exported public types
are flatten.

What's missing is the documentation, let's address in another round of
PR.

It's really a copy-pasta patch, nothing fancy. We must agree on the module
namings, that's probably the most difficult part.

The `Makefile` has not been updated because there is nothing to test with this crate.

# Review


MarkMcCaskey (approved):
I gave my feedback on Slack, I'll summarize it below.

- Not a huge fan of splitting into so many files but it's fine.
- It'd be better if wasmer-wasi-types didn't depend on `wasmer`, I'd say it's worth it to convert all uses of `WasmPtr` (there are 1 or 2) to `u32`s to avoid this. There is a way to do a more proper fix but it's a lot of work and the work-around is super simple.

Otherwise, looks good to me 👍 

Another point of feedback, the WASI docs don't use C-style naming for the types like we use anymore, it may be worth it update it but we can do that later. I'm not sure to what extent a witx generated crate would be useful, but perhaps that's something we can do, too, to avoid having to keep up with it.

Hywan:
`wasmer-wasi-types` depends on `wasmer` to get `ValueType`. Can you tell me why it is so?

MarkMcCaskey:
@Hywan Is it possible to move `ValueType` to `wasmer-types`? I wouldn't expect it to be at the `wasmer` level at all to be honest, it's a very low level trait meant to convey essentially what `Copy` does with an extra twist: namely that a type is safe to convert to/from bytes freely and that all bit patterns are valid

Hywan:
Ideally, `wasi-types` must not fetch `wasmer` because it will fetch much more, and we can't compile everything to `wasm32-wasi`. I'll update the PR to move `ValueType` to `wasmer-types`, so that we can remove the dependency to `wasmer`.

Hywan:
Please @MarkMcCaskey, can you review https://github.com/wasmerio/wasmer/pull/2411/commits/cfd400853070667073818f5fa3f89c62bf21391a carefully please :-)?

Hywan:
bors try

lib/wasi-types/src/io.rs
```diff
@@ -0,0 +1,19 @@
+use wasmer_types::ValueType;
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[repr(C)]
+pub struct __wasi_ciovec_t {
+    pub buf: u32,
```
Hywan, line 6: This changed from `WasmPtr<u8, Array>` to `u32` to remove the need to use `WasmPtr` in `wasmer-wasi-types`. But I believe it's complicated to create this type now. Imagine you have the following code:

```rust
let mut vec: Vec<u8> = vec![0; 2];
let iov = __wasi_ciovec_t {
    buf: vec.as_mut_ptr(),
    buf_len: vec.len() as u32,
};
```

it fails to compile because `buf` is a `u32`, not a `*mut u8`. Casting a pointer to a usize is unstable for the moment (it requires the `const_raw_ptr_to_usize_cast` feature). It's not ideal.

Any recommandation @MarkMcCaskey?

Hywan: A workaround is to:

```rust
    buf: vec.as_mut_ptr() as usize as u32,
```

I'm not super happy with that though.

MarkMcCaskey: > Casting a pointer to a usize is unstable for the moment 

Really? I've done it all the time and never seen this. I wonder if it's because it's `*mut`?  🤔 

I think your work around is what I'd expect here. We can have a `WasmUsize` type or something, but even that is confusing because with Wasm64, Wasm will have both 64 and 32 bit pointers depending on context. I'm unsure to what extent that will happen in practice given the target triples are all `wasm32-*`, maybe it'll be fully separated into `wasm64-*` and `wasm32-*`.

Hywan: It's out of the scope of this PR, but yeah, I agree that we must use a type alias to represent the memory size, something like `wasm_usize`/`WasmUsize` is an excellent idea.

lib/wasi/src/ptr.rs
```diff
@@ -83,9 +83,4 @@ impl<T: Copy + ValueType> WasmPtr<T, Array> {
     pub unsafe fn get_utf8_str(self, memory: &Memory, str_len: u32) -> Option<&str> {
         self.0.get_utf8_str(memory, str_len)
     }
-
-    #[inline(always)]
-    pub fn get_utf8_string(self, memory: &Memory, str_len: u32) -> Option<String> {
-        self.0.get_utf8_string(memory, str_len)
-    }
```
MarkMcCaskey, line 90: I think we want to undelete this as I believe it's a publicly visible API. I'm not sure to what extent it's useful as this is the WASI WasmPtr, but if it's publicly visible, we probably shouldn't remove it

MarkMcCaskey (approved):
Looks good to me

lib/wasi/src/ptr.rs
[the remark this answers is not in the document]
Hywan: The `WasmPtr` type wasn't any longer public due to types moving to another crate. I've made the entire `ptr` module public, and I've restored this method.

PR #2412, opened 2021-06-09
Abstracted WASI FS

syrusakbary:
# Description

This PR is in progress (created so I can comment and give feedback)

# Review


lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+use std::fmt;
+use std::io::{self, Read, Seek, Write};
+use std::path::Path;
+use thiserror::Error;
+use tracing::debug;
+
+pub mod host_fs;
+
+pub trait FileSystem: Send + Sync + 'static {
+    fn read_dir(&self, path: &Path) -> Result<std::fs::ReadDir, FsError>;
```
syrusakbary, line 11: The `std::fs::ReadDir` type is very restrictive since it's not a type that we can actually generate if our filesystem is not a host one (think on having this over a tar file, the read_dir will need to be implemented by the FileSystem implementor).

I believe we can solve this by either return a iterator of strings or a custom readdir trait.

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+    fn create_dir(&self, path: &Path) -> Result<(), FsError>;
+    fn remove_dir(&self, path: &Path) -> Result<(), FsError>;
+    fn rename(&self, from: &Path, to: &Path) -> Result<(), FsError>;
+
+    fn remove_file(&self, path: &Path) -> Result<(), FsError>;
+    fn new_open_options(&self) -> OpenOptions;
```
syrusakbary, line 17: What is this?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+            io::ErrorKind::TimedOut => FsError::TimedOut,
+            io::ErrorKind::UnexpectedEof => FsError::UnexpectedEof,
+            io::ErrorKind::WouldBlock => FsError::WouldBlock,
+            io::ErrorKind::WriteZero => FsError::WriteZero,
+            io::ErrorKind::Other => FsError::IOError,
+            // if the following triggers, a new error type was added to this non-exhaustive enum
+            _ => FsError::UnknownError,
+        }
+    }
+}
```
syrusakbary, line 319: Can we have a `MemoryFS` also with tests?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+
+impl dyn VirtualFile + 'static {
+    #[inline]
+    pub fn downcast_ref<T: 'static>(&'_ self) -> Option<&'_ T> {
+        self.upcast_any_ref().downcast_ref::<T>()
+    }
+    #[inline]
+    pub fn downcast_mut<T: 'static>(&'_ mut self) -> Option<&'_ mut T> {
+        self.upcast_any_mut().downcast_mut::<T>()
+    }
```
syrusakbary, line 216: I believe this can simplified a bit if we have a `as_any` method? (and we move the downcasting back to the code that uses and needs it)

lib/engine/src/artifact.rs
```diff
@@ -179,7 +179,7 @@ pub trait Artifact: Send + Sync + Upcastable + MemoryUsage {
 }
 
 // Implementation of `Upcastable` taken from https://users.rust-lang.org/t/why-does-downcasting-not-work-for-subtraits/33286/7 .
-/// Trait needed to get downcasting from `WasiFile` to work.
+/// Trait needed to get downcasting from `VirtualFile` to work.
```
syrusakbary, line 182: Is this correct? Engines should not care about `WasiFile` or `VirtualFile`

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+        None
+    }
+}
+
+// Implementation of `Upcastable` taken from https://users.rust-lang.org/t/why-does-downcasting-not-work-for-subtraits/33286/7 .
+/// Trait needed to get downcasting from `VirtualFile` to work.
+pub trait Upcastable {
+    fn upcast_any_ref(&'_ self) -> &'_ dyn Any;
+    fn upcast_any_mut(&'_ mut self) -> &'_ mut dyn Any;
+    fn upcast_any_box(self: Box<Self>) -> Box<dyn Any>;
```
syrusakbary, line 190: This method is not used anywhere

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+    fn get_raw_fd(&self) -> Option<i32> {
+        None
+    }
+}
+
+// Implementation of `Upcastable` taken from https://users.rust-lang.org/t/why-does-downcasting-not-work-for-subtraits/33286/7 .
+/// Trait needed to get downcasting from `VirtualFile` to work.
+pub trait Upcastable {
+    fn upcast_any_ref(&'_ self) -> &'_ dyn Any;
+    fn upcast_any_mut(&'_ mut self) -> &'_ mut dyn Any;
```
syrusakbary, line 189: I believe a better name is `as_any_mut`

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+    /// Returns the underlying host fd
+    fn get_raw_fd(&self) -> Option<i32> {
+        None
+    }
+}
+
+// Implementation of `Upcastable` taken from https://users.rust-lang.org/t/why-does-downcasting-not-work-for-subtraits/33286/7 .
+/// Trait needed to get downcasting from `VirtualFile` to work.
+pub trait Upcastable {
+    fn upcast_any_ref(&'_ self) -> &'_ dyn Any;
```
syrusakbary, line 188: `as_any` ?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    /// Returns the number of bytes available.  This function must not block
+    fn bytes_available(&self) -> Result<usize, FsError>;
+
+    /// Used for polling.  Default returns `None` because this method cannot be implemented for most types
+    /// Returns the underlying host fd
+    fn get_raw_fd(&self) -> Option<i32> {
+        None
+    }
```
syrusakbary, line 182: This method is not used publicly, so I believe we can remove it.

At the same time by removing it we also assure that we shield from any specific host requirement in the Virtual FS.

syrusakbary: Small correction. Is indeed using in `state.rs` for polling.
I think we can abstract the polling as well into the Virtual FS.

That way we don't need to expose any host requirements in the VirtualFS itself.

lib/virtual-fs/src/lib.rs
MarkMcCaskey: Yeah, it's wip

lib/virtual-fs/src/lib.rs
MarkMcCaskey: Allows specifying how files can be opened, same as std::fs::OpenOptions

lib/virtual-fs/src/lib.rs
MarkMcCaskey: Yep, working on  it

lib/virtual-fs/src/lib.rs
MarkMcCaskey: Well, we're dealing with `Box<dyn VirtualFile>` (previously `Box<dyn WasiFile>`), due to the way this trait works, everything just works and users don't have to do anything but they can use these methods if they want, so I think it's pretty good. We can discuss possible improvements though

lib/engine/src/artifact.rs
MarkMcCaskey: Looks like the doc comment was out of date, we needed the same pattern for engines too, I did a bulk find and replace for `WasiFile` -> `VirtualFile` and didn't realize this file had been edited. We should fix up this doc comment to be about engines

lib/virtual-fs/src/lib.rs
MarkMcCaskey: It's a public method, we can review its utility but it's useful when working with `dyn Thing` and you might, as a user, care about getting your concrete thing out of the dynamic thing. This is a useful part of the `WasiFile` API which allows users to define whatever logic they want for `WasiFile`s

lib/virtual-fs/src/mem_fs.rs
```diff
@@ -0,0 +1,600 @@
...
+#[derive(Clone, Default)]
+pub struct MemFileSystem {
+    inner: Arc<Mutex<MemFileSystemInner>>,
+}
+
+#[derive(Default)]
+pub struct MemFileSystemInner {
+    // done for recursion purposes
+    fs: MemKind,
+    inodes: HashMap<u64, Box<dyn VirtualFile>>,
```
syrusakbary, line 40: I think inodes should not be needed for a MemoryFilesystem? Or is this designed with symlinks in mind?

Why do we need the MemFileHandle?

syrusakbary: It seems we have `MemFileHandle` and `MemFile` as two ways of having a file in the memfs. I'm still trying to wrap my head around on why do we need two?

lib/virtual-fs/src/mem_fs.rs
```diff
@@ -0,0 +1,600 @@
...
+        inner
+            .inodes
+            .get(&self.inode)
+            .as_ref()
+            .map(|file| file.created_time())
+            .unwrap_or_default()
+    }
+
+    fn set_created_time(&self, created_time: u64) {
+        let inner = self.fs.inner.lock().unwrap();
```
syrusakbary, line 527: If `set_created_time` uses `&mut self` we can move from interior mutability to exterior very easily (so no locks are needed)

lib/virtual-fs/src/vfs_fs.rs
```diff
@@ -0,0 +1,201 @@
...
+use std::sync::Arc;
+
+pub use crate::host_fs::{Stderr, Stdin, Stdout};
+
+#[derive(Clone)]
+pub struct VfsFileSystem {
+    inner: Arc<dyn vfs::FileSystem>,
+}
+
+impl FileSystem for VfsFileSystem {
```
syrusakbary, line 14: Why do we need a VfsFileSystem?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+use tracing::debug;
+
+pub mod host_fs;
+//pub mod vfs_fs;
+pub mod mem_fs;
+
+pub trait FileSystem: Send + Sync + 'static {
+    fn read_dir(&self, path: &Path) -> Result<std::fs::ReadDir, FsError>;
+    fn create_dir(&self, path: &Path) -> Result<(), FsError>;
+    fn remove_dir(&self, path: &Path) -> Result<(), FsError>;
```
syrusakbary, line 15: Can we have this as `&mut self`?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+    fn rename(&self, from: &Path, to: &Path) -> Result<(), FsError>;
```
syrusakbary, line 16: Can we have this as `&mut self`?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    fn remove_file(&self, path: &Path) -> Result<(), FsError>;
```
syrusakbary, line 18: Can we have this as `&mut self`?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+    append: bool,
+    truncate: bool,
+}
+
+impl OpenOptionsConfig {
+    pub const fn read(&self) -> bool {
+        self.read
+    }
+
+    pub const fn write(&self) -> bool {
```
syrusakbary, line 45: Can we have this as `&mut self`?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    /// the last time the file was modified in nanoseconds as a UNIX timestamp
+    fn last_modified(&self) -> u64;
+
+    /// the time at which the file was created in nanoseconds as a UNIX timestamp
+    fn created_time(&self) -> u64;
+
+    /// set the last time the file was accessed in nanoseconds as a UNIX timestamp
+    fn set_last_accessed(&self, _last_accessed: u64) {
+        debug!("{:?} did nothing in VirtualFile::set_last_accessed due to using the default implementation", self);
```
syrusakbary, line 137: Can we have this as `&mut self`?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    /// set the last time the file was modified in nanoseconds as a UNIX timestamp
+    fn set_last_modified(&self, _last_modified: u64) {
+        debug!("{:?} did nothing in VirtualFile::set_last_modified due to using the default implementation", self);
```
syrusakbary, line 142: Can we have this as `&mut self`?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    /// set the time at which the file was created in nanoseconds as a UNIX timestamp
+    fn set_created_time(&self, _created_time: u64) {
```
syrusakbary, line 146: Can we have this as `&mut self`?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+            self
+        );
+    }
+
+    /// the size of the file in bytes
+    fn size(&self) -> u64;
+
+    /// Change the size of the file, if the `new_size` is greater than the current size
+    /// the extra bytes will be allocated and zeroed
+    fn set_len(&mut self, new_size: u64) -> Result<(), FsError>;
```
syrusakbary, line 158: Can we have this as `&mut self`?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    /// Request deletion of the file
+    fn unlink(&mut self) -> Result<(), FsError>;
```
syrusakbary, line 161: Can we have this as `self` (so it's consumed)?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    /// Store file contents and metadata to disk
+    /// Default implementation returns `Ok(())`.  You should implement this method if you care
+    /// about flushing your cache to permanent storage
+    fn sync_to_disk(&self) -> Result<(), FsError> {
+        Ok(())
+    }
```
syrusakbary, line 168: I think this only make sense for some kind of filesystems. Can we move it to other layer?

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    /// Moves the file to a new location
+    /// NOTE: the signature of this function will change before stabilization
+    // TODO: stablizie this in 0.7.0 or 0.8.0 by removing default impl
+    fn rename_file(&self, _new_name: &std::path::Path) -> Result<(), FsError> {
```
syrusakbary, line 173: Can we have this as `&mut self`?

lib/virtual-fs/src/lib.rs
syrusakbary: Thinking more, I think perhaps `unlink` and `rename` can be in the `FileSystem` api rather than in the `VirtualFile` itself. It makes things a bit easier to reason about I think

lib/virtual-fs/src/mem_fs.rs
MarkMcCaskey: It's a bit of a band-aid to cover up a mismatch: the API is written in terms of `Box<dyn VirtualFile>` but because it's also the backing storage we need to share it. So the internal file implementation here (which is also implemented in terms of `VirtualFile` but doesn't have to be) is shared with `Arc` and we return a `Box<dyn VirtualFile>` for the API itself. This can be fixed after more iteration, I just didn't want to make too many big changes all at once

lib/virtual-fs/src/mem_fs.rs
MarkMcCaskey: In this case we need still need it because it's shared. Interior mutability wouldn't work unless we use atomics. Which _is_ possible here but probably not worth it

lib/virtual-fs/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+use thiserror::Error;
+use tracing::debug;
+
+pub mod host_fs;
+//pub mod vfs_fs;
+pub mod mem_fs;
+
+pub trait FileSystem: Send + Sync + 'static {
+    fn read_dir(&self, path: &Path) -> Result<std::fs::ReadDir, FsError>;
+    fn create_dir(&self, path: &Path) -> Result<(), FsError>;
```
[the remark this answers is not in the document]
MarkMcCaskey: Yeah, it could be, I'm not sure it's any better though, then users of `dyn FileSystem` will need to do the sharing themselves, which is a fine pattern, this way is kind of simpler for users of the API though

lib/virtual-fs/src/lib.rs
MarkMcCaskey: Yeah, it's due to WasiFile that it is this way, I agree that it's probably possible to move some of that logic away, it's really just a clean up for the file itself. I think `self` is problematic in trait objects, `Drop::drop` takes self by `&mut` so  I think this is fine

lib/virtual-fs/src/lib.rs
MarkMcCaskey: Well it's fine for filesystems to not implement it, if this is in WasiFile then it's fine. But WASI needs a method like this to call for things to be emulated accurately

lib/virtual-fs/src/lib.rs
syrusakbary: Ok, I added a side comment on other place... perhaps we can delete this method and use only the filesystem one (`remove_file`)

lib/virtual-fs/src/lib.rs
syrusakbary: Why not move it to the WASI filesystem layer?

MarkMcCaskey: How would we? This is the layer through which WASI is interacting with the filesystem

MarkMcCaskey: I think it's important to have not only these types of methods, but a lot more if we want this to be generally useful as a drop-in replacement abstraction for real file system usage...

lib/virtual-fs/src/lib.rs
MarkMcCaskey: Yeah, I'm in no rush to do this yet (as I haven't done it) because I think depending on how we want to compose different types of VirtualFiles and file systems it may still make sense to have this as-is.

MarkMcCaskey: Working out the exact lines between the abstractions is still a work in progress

lib/virtual-fs/src/vfs_fs.rs
MarkMcCaskey: We don't, strictly, but the code is pretty much written. We'll probably delete it before shipping but I'll keep it around until the PR starts being less of draft in case we decide it is  something that we want.

lib/virtual-fs/src/lib.rs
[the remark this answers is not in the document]
MarkMcCaskey: Hmm, interesting, so the use case for Wasmer absolutely needs locking due to how `WasmerEnv` works and how we'd need  to make things work in WASI and emscripten... that's a good point though, if it's for use in Wasm, maybe it's worth it to move the locking to the API user as needed

lib/virtual-fs/src/vfs_fs.rs
syrusakbary: Yeah, I was trying to find a use case for it and was not able to find any. Let's keep the file up until merging just in case it results useful

lib/virtual-fs/src/mem_fs.rs
MarkMcCaskey: Yeah if we make `OpenOptions` have a lifetime then maybe that can work  if we change the whole api to use `&mut` and require locking from API users (like our existing wasmer code)

Issue #2414, opened 2021-06-10
Unresolved external symbol when compiling for 32bit windows


MidasLamb:
### Describe the bug
I'm compiling for windows 32 bit from a 64bit machine by doing 
```sh
cargo +stable-i686-pc-windows-msvc --target i686-pc-windows-msvc r
```

And I get following error: 
```sh
= note: libwasmer_vm-dc29bfe30745a26c.rlib(wasmer_vm-dc29bfe30745a26c.wasmer_vm.7cxdr9x9-cgu.15.rcgu.o) : error LNK2001: unresolved external symbol ___rust_probestack
          libwasmer_vm-dc29bfe30745a26c.rlib(wasmer_vm-dc29bfe30745a26c.wasmer_vm.7cxdr9x9-cgu.8.rcgu.o) : error LNK2001: unresolved external symbol ___rust_probestack
          C:\Users\Midas\Desktop\projects\dbs2\server\target\i686-pc-windows-msvc\debug\deps\server.exe : fatal error LNK1120: 1 unresolved externals
```

### Steps to reproduce
Have the following program: 
```rust
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    use wasmer::{imports, Instance, Module, Store, Value};
    // some other unrelated stuff, only having the use above is sufficient.
   println!("Hello world")
}
```
while having this in the dependencies in Cargo.toml
```toml
wasmer = {git = "https://github.com/wasmerio/wasmer.git", branch = "master", default-features = false, features = ["default-cranelift", "default-universal", "wat"]}
```
and try to run it with:
```sh
cargo +stable-i686-pc-windows-msvc --target i686-pc-windows-msvc r
```

### Expected behavior
Compilation succeeds and be able to run it.

### Actual behavior
Compilation fails during linking (`cargo check` succeeds)

I tried to solve this myself but I have no idea why this happens so I'm stuck :/

Hywan:
That's because we don't support Windows 32bits in our probestack module. It fallbacks to the last `else` block: https://github.com/wasmerio/wasmer/blob/144300e0bc51d01290680a46cd30a6ca76ed0359/lib/vm/src/probestack.rs#L20-L53 hence the error.

[...]

PR #2415, opened 2021-06-11
Add the `_chkstk` for 32 bit windows with msvc.

MidasLamb:
# Description
Add the `_chkstk` (singular underscore) for 32 bit windows when compiled with msvc.

### DIFF
--- a/lib/vm/src/probestack.rs
+++ b/lib/vm/src/probestack.rs
@@ -25,8 +25,18 @@ cfg_if::cfg_if! {
         extern "C" {
             pub fn __chkstk();
         }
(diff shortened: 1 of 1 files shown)


MidasLamb:
Fixes #2414

MidasLamb:
Also note, I have no idea what this routine is supposed to do, so it might be that `_chkstk` is not the right `extern`, but this at least compiles and links without failing.

syrusakbary:
This looks good. Thanks for the fix @MidasLamb

syrusakbary (approved):

Issue #2226, opened 2021-04-02
Failed to run file with imports from both "wasi_unstable" and "wasi_snapshot_preview1"


jsimnz:
### Describe the bug
Wasmer failing to run a wasm file with mixed WASI imports.

I have a compiled file with the following imports:
```
(import "wasi_unstable" "fd_write" (func $runtime.fd_write (type 1)))
(import "wasi_unstable" "clock_time_get" (func $runtime.clock_time_get (type 2)))
(import "wasi_snapshot_preview1" "environ_sizes_get" (func $__wasi_environ_sizes_get (type 3)))
(import "wasi_snapshot_preview1" "proc_exit" (func $__wasi_proc_exit (type 0)))
(import "wasi_snapshot_preview1" "environ_get" (func $__wasi_environ_get (type 3)))
```

I'm not sure why my specific compiler (tinygo) produces mixed output, but I believe it should work regardless.

When trying to run the following command
```
wasmer run main.wat
```

It returns the following error:
```
error: failed to run `main.wat`
╰─> 1: Error while importing "wasi_unstable"."fd_write": unknown import. Expected Function(FunctionType { params: [I32, I32, I32, I32], results: [I32] })
```

However, If I manually edit the `.wat` file, and change the imports to all use _either_ "wasi_unstable" or "wasi_snapshot_preview1", it works without issue, only with mixed versions does it produce the error.

I am using the following version
```sh
echo "`wasmer -V`" : wasmer 1.0.2
```

### Steps to reproduce
Run a wasm file with mixed `WASI` version imports.

### Expected behavior
Run the file and produce the correct program output without issue

### Actual behavior
Returns the following error:
```
error: failed to run `main.wat`
╰─> 1: Error while importing "wasi_unstable"."fd_write": unknown import. Expected Function(FunctionType { params: [I32, I32, I32, I32], results: [I32] })
```

### Additional Context
I have verified that the original wasm file, with the mixed WASI import versions runs succesfully using the wasmtime runtime.

PR #2424, opened 2021-06-16
Fix: get wasi versions right

uuhan:
# Description

The following sample code should have the right wasi_version.

```wast
(module
  (type (;0;) (func (param i32 i32 i32 i32) (result i32)))
  (type (;1;) (func (param i32 i32 i32 i32)))
  (import "wasi_snapshot_preview1" "fd_write" (func (;0;) (type 0)))
  (import "env" "abort" (func (;1;) (type 1)))
)
```

Related Issue: #2226

### DIFF
--- a/lib/cli/src/commands/run.rs
+++ b/lib/cli/src/commands/run.rs
@@ -156,41 +156,37 @@ impl Run {
             use wasmer_wasi::WasiVersion;
 
             let wasi_versions = Wasi::get_versions(&module);
-            match wasi_versions {
-                Some(wasi_versions) if !wasi_versions.is_empty() => {
-                    if wasi_versions.len() >= 2 {
-                        let get_version_list = |versions: &BTreeSet<WasiVersion>| -> String {
-                            versions
-                                .iter()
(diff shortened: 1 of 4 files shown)


uuhan:
OK, I find another related pr [#1028](https://github.com/wasmerio/wasmer/pull/1028) whick introduces a `strict` parameter.

So, what is the expected beheavior? If I have two module named `wasi_snapshot_preview1` and `env`, wasmer compilants `wasi_snapshot_preview1` not found.

jcaesar:
uuhan: If you have symbols from both `env.` and `wasi_snapshot_preview1.`, then that is probably code compiled by a not-that-old version of emscripten (e.g. > 1.39.20, iirc), which mixes its own custom interface with `wasi_snapshot_preview1`-like calls. Problem with that is: The emscripten `env.` stuff is not sandboxed, but wasmer implements wasi calls sandboxed.
So to support that kind of emscripten-generated binary, you'd need a non-sandboxed version of wasi, which is an entirely different problem…

PR #2428, opened 2021-06-18
Don't fail make install if WASMER_INSTALL_PREFIX isn't set

### LINKED ISSUE #2413: WASMER_INSTALL_PREFIX will cause confusion in install
### Summary

As a normal case in building from source on Linux/Unix. I will use 
```
make 
sudo make install
```
to build and install from source code.

But follow these steps, there will be an error like this by default: 
```
target/release/wasmer config --pkg-config | install -Dm644 /dev/stdin "/usr/local"/lib/pkgconfig/wasmer.pc
unset WASMER_DIR # Make sure WASMER_INSTALL_PREFIX is set during build
error: failed to retrieve the wasmer config
│   1: failed to retrieve the WASMER_DIR environment variables
╰─> 2: environment variable not found
```

Then, I need to find the document, and the document is not clear about this,  https://github.com/wasmerio/wasmer/blob/master/PACKAGING.md.
Actually, we need to add `WASMER_INSTALL_PREFIX` in the `make` stage, not `make install` stage.

By default, `makefile` use `DISTDIR` as the target directory. 
https://github.com/wasmerio/wasmer/blob/144300e0bc51d01290680a46cd30a6ca76ed0359/Makefile#L632

I saw the `WASMER_INSTALL_PREFIX` is introduced by this commit: https://github.com/wasmerio/wasmer/commit/80ec06ffab9d7af017a5a5490d153a2230f9374a

I don't get the point for this variable, and I think it makes some confusion.  

If we want to configure the path, we need to add two environment variables, `WASMER_INSTALL_PREFIX` to recompile the whole project.

jcaesar:
# Description
Fixes #2413, also improves PACKAGING.md

Alternatively, I could make it so `make install` errors out with a nice and clean error message if that variable is not set (and tells people that they're not supposed to make install themselves?), or default WASMER_INSTALL_PREFIX to `/usr/local` (at the danger of being missed, resulting in incorrect installs).

# Review
- [ ] Add a short description of the change to the CHANGELOG.md file: Seems like an irrelevant fix - Necessary?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -657,8 +657,13 @@ install-misc:
 	install -Dm644 LICENSE "$(DESTDIR)"/share/licenses/wasmer/LICENSE
 
 install-pkgconfig:
-	unset WASMER_DIR # Make sure WASMER_INSTALL_PREFIX is set during build
-	target/release/wasmer config --pkg-config | install -Dm644 /dev/stdin "$(DESTDIR)"/lib/pkgconfig/wasmer.pc
+	# Make sure WASMER_INSTALL_PREFIX is set during build
+	unset WASMER_DIR; \
+	if pc="$$(target/release/wasmer config --pkg-config 1>/dev/null 2>/dev/null)"; then \
+		echo "$$pc" | install -Dm644 /dev/stdin "$(DESTDIR)"/lib/pkgconfig/wasmer.pc; \
+	else \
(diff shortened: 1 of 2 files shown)


Hywan (approved):
Looks good to me. Thanks! Waiting for another review from @syrusakbary for example.

Hywan:
> Add a short description of the change to the CHANGELOG.md file: Seems like an irrelevant fix - Necessary?

Not needed in this case I believe.

PR #2442, opened 2021-06-24
Improved WasmPtr, added WasmCell

syrusakbary:
# Description

This PR improves a bit our WasmPtr API by making it a bit more safer and reliable.

## Breaking changes
* `WasmPtr`.`deref` will now return `WasmCell<'a, T>` instead of `&'a Cell<T>`
* `WasmPtr`.`deref_mut` is no needed any longer

### DIFF
--- a/lib/api/src/cell.rs
+++ b/lib/api/src/cell.rs
@@ -42,13 +42,13 @@ use core::ptr;
 /// See the [module-level documentation](self) for more.
 #[derive(Clone)]
 #[repr(transparent)]
-pub struct WasmCell<T: ?Sized> {
-    inner: *const Cell<T>,
+pub struct WasmCell<'a, T: ?Sized> {
+    inner: &'a Cell<T>,
 }
 
-unsafe impl<T: ?Sized> Send for WasmCell<T> where T: Send {}
+unsafe impl<T: ?Sized> Send for WasmCell<'_, T> where T: Send {}
 
-unsafe impl<T: ?Sized> Sync for WasmCell<T> {}
+unsafe impl<T: ?Sized> Sync for WasmCell<'_, T> {}
 
 // impl<T: Copy> Clone for WasmCell<T> {
 //     #[inline]
@@ -67,16 +67,16 @@ unsafe impl<T: ?Sized> Sync for WasmCell<T> {}
 //     }
 // }
 
-impl<T: PartialEq + Copy> PartialEq for WasmCell<T> {
+impl<T: PartialEq + Copy> PartialEq for WasmCell<'_, T> {
     #[inline]
     fn eq(&self, other: &WasmCell<T>) -> bool {
         true
     }
 }
 
-impl<T: Eq + Copy> Eq for WasmCell<T> {}
+impl<T: Eq + Copy> Eq for WasmCell<'_, T> {}
 
-impl<T: PartialOrd + Copy> PartialOrd for WasmCell<T> {
+impl<T: PartialOrd + Copy> PartialOrd for WasmCell<'_, T> {
     #[inline]
     fn partial_cmp(&self, other: &WasmCell<T>) -> Option<Ordering> {
         self.inner.partial_cmp(&other.inner)
@@ -103,7 +103,7 @@ impl<T: PartialOrd + Copy> PartialOrd for WasmCell<T> {
     }
 }
 
-impl<T: Ord + Copy> Ord for WasmCell<T> {
+impl<T: Ord + Copy> Ord for WasmCell<'_, T> {
     #[inline]
     fn cmp(&self, other: &WasmCell<T>) -> Ordering {
         self.get().cmp(&other.get())
@@ -117,7 +117,7 @@ impl<T: Ord + Copy> Ord for WasmCell<T> {
 //     }
 // }
 
-impl<T> WasmCell<T> {
+impl<'a, T> WasmCell<'a, T> {
     /// Creates a new `WasmCell` containing the given value.
     ///
     /// # Examples
@@ -128,7 +128,7 @@ impl<T> WasmCell<T> {
     /// let c = WasmCell::new(5);
     /// ```
     #[inline]
-    pub const fn new(cell: *const Cell<T>) -> WasmCell<T> {
+    pub const fn new(cell: &'a Cell<T>) -> WasmCell<'a, T> {
         WasmCell {
             inner: cell,
         }
@@ -202,7 +202,7 @@ impl<T> WasmCell<T> {
     // }
 }
 
-impl<T: Copy> WasmCell<T> {
+impl<T: Copy> WasmCell<'_, T> {
     /// Returns a copy of the contained value.
     ///
     /// # Examples
@@ -216,11 +216,17 @@ impl<T: Copy> WasmCell<T> {
     /// ```
     #[inline]
     pub fn get(&self) -> T {
-        unsafe { (*self.inner).get() }
+        self.inner.get()
+    }
+
+    /// Get an unsafe mutable pointer to the inner item
+    /// in the Cell.
+    pub unsafe fn get_mut(&self) -> &mut T {
+        &mut *self.inner.as_ptr()
     }
 }
 
-impl<T: Sized> WasmCell<T> {
+impl<T: Sized> WasmCell<'_, T> {
     /// Sets the contained value.
     ///
     /// # Examples
@@ -234,6 +240,6 @@ impl<T: Sized> WasmCell<T> {
     /// ```
     #[inline]
     pub fn set(&self, val: T) {
-        unsafe { (*self.inner).set(val) };
+        self.inner.set(val);
     }
 }
--- a/lib/api/src/cell.rs
+++ b/lib/api/src/cell.rs
@@ -130,6 +130,20 @@ impl<'a, T: Copy> WasmCell<'a, T> {
 
     /// Get an unsafe mutable pointer to the inner item
     /// in the Cell.
+    ///
+    /// # Safety
+    ///
+    /// This method is highly discouraged to use. We have it for
+    /// compatibility reasons with Emscripten.
+    /// It is unsafe because changing an item inline will change
+    /// the underlying memory.
+    ///
+    /// It's highly encouraged to use the `set` method instead.
+    #[deprecated(
+        since = "2.0.0",
+        note = "Please use the memory-safe set method instead"
+    )]
+    #[doc(hidden)]
     pub unsafe fn get_mut(&self) -> &'a mut T {
         &mut *self.inner.as_ptr()
     }
--- a/lib/api/src/cell.rs
+++ b/lib/api/src/cell.rs
@@ -1,10 +1,7 @@
 pub use std::cell::Cell;
 
 use core::cmp::Ordering;
-use core::fmt::{self, Debug, Display};
-use core::mem;
-use core::ops::{Deref, DerefMut};
-use core::ptr;
+use core::fmt::{self, Debug};
 use std::fmt::Pointer;
 
 /// A mutable Wasm-memory location.
--- a/lib/api/src/cell.rs
+++ b/lib/api/src/cell.rs
@@ -7,6 +7,7 @@ use core::fmt::{self, Debug, Display};
 use core::mem;
 use core::ops::{Deref, DerefMut};
 use core::ptr;
+use std::fmt::Pointer;
 
 /// A mutable memory location.
 ///
@@ -225,6 +226,14 @@ impl<T: Copy> WasmCell<'_, T> {
         &mut *self.inner.as_ptr()
     }
 }
+
+impl<T: Debug> Debug for WasmCell<'_, T> {
+    #[inline]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.inner.fmt(f)
+    }
+}
+
 
 impl<T: Sized> WasmCell<'_, T> {
     /// Sets the contained value.
--- a/lib/api/src/cell.rs
+++ b/lib/api/src/cell.rs
@@ -94,7 +94,7 @@ impl<T: PartialOrd + Copy> PartialOrd for WasmCell<'_, T> {
 impl<T: Ord + Copy> Ord for WasmCell<'_, T> {
     #[inline]
     fn cmp(&self, other: &WasmCell<T>) -> Ordering {
-        self.get().cmp(&other.get())
+        self.inner.cmp(&other.inner)
     }
 }
 
@@ -114,7 +114,7 @@ impl<'a, T> WasmCell<'a, T> {
     }
 }
 
-impl<T: Copy> WasmCell<'_, T> {
+impl<'a, T: Copy> WasmCell<'a, T> {
     /// Returns a copy of the contained value.
     ///
     /// # Examples
@@ -133,7 +133,7 @@ impl<T: Copy> WasmCell<'_, T> {
 
     /// Get an unsafe mutable pointer to the inner item
     /// in the Cell.
-    pub unsafe fn get_mut(&self) -> &mut T {
+    pub unsafe fn get_mut(&self) -> &'a mut T {
         &mut *self.inner.as_ptr()
     }
 }
--- a/lib/api/src/cell.rs
+++ b/lib/api/src/cell.rs
@@ -1,0 +1,239 @@
+pub use std::cell::Cell;
+
+
+
+use core::cmp::Ordering;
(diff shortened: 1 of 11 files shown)


lib/api/src/cell.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// let five = c.get();
+    /// ```
+    #[inline]
+    pub fn get(&self) -> T {
+        self.inner.get()
+    }
+
+    /// Get an unsafe mutable pointer to the inner item
+    /// in the Cell.
+    pub unsafe fn get_mut(&self) -> &mut T {
```
MarkMcCaskey, line 136: We should have a `# Safety` section for all `unsafe` functions. This one can mention that the caller should guarantee themselves that Wasm isn't executing and nothing else has a reference (mutable or otherwise) to the data being  pointed to by the returned reference

lib/api/src/ptr.rs
```diff
@@ -103,7 +104,7 @@ impl<T: Copy + ValueType> WasmPtr<T, Item> {
     /// If you're unsure what that means, it likely does not apply to you.
     /// This invariant will be enforced in the future.
     #[inline]
-    pub fn deref<'a>(self, memory: &'a Memory) -> Option<&'a Cell<T>> {
+    pub fn deref<'a>(self, memory: &'a Memory) -> Option<WasmCell<T>> {
```
MarkMcCaskey, line 107: Since we bind a lifetime explicitly here, we should probably set it in `WasmCell` here.. I'm not sure about the inference rules if `self` doesn't have a `&`, but when it does, it'd by default bind it to that. I'm guessing it's inferring unrelated lifetimes

MarkMcCaskey (approved):
I think this is workable if we're good with the breakage in API and the performance decrease for the array method. I think this is a reasonable path forward: we can always add more APIs later for better performance.

lib/api/src/ptr.rs
syrusakbary: Fixed!

lib/api/src/cell.rs
syrusakbary: Done

MarkMcCaskey (changes requested):
need to fix the major bugs first

lib/types/src/memory_view.rs
```diff
@@ -64,6 +64,33 @@ where
             _phantom: PhantomData,
         }
     }
+
+    /// Creates a subarray view from this MemoryView.
+    pub fn subarray(&self, start: u32, end: u32) -> Self {
+        assert!(start <= end);
+        assert!(end < self.length as u32);
```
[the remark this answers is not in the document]
syrusakbary: Oh this really means:

```suggestion
        assert!(end < (self.length as u32));
```

Is this clearer?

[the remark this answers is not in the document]
MarkMcCaskey: It's not about being clear, the logic doesn't handle a max length wasm memory, it's a logic error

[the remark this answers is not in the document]
MarkMcCaskey: > Bounds checking wasm memory should not be done with u32s.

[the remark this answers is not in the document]
syrusakbary: Got it!

lib/api/src/ptr.rs
```diff
@@ -151,66 +152,72 @@ impl<T: Copy + ValueType> WasmPtr<T, Array> {
...
-            Some(cell_ptrs)
-        }
+        Some(
+            (0..length)
+                .map(|i| unsafe {
+                    let cell_ptr = align_pointer(
+                        memory
+                            .view::<u8>()
+                            .as_ptr()
+                            .add((self.offset + i * item_size) as usize)
```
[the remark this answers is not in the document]
syrusakbary: I'm not sure I'm not sure if I understand this comment. This is going from u32 to usize, so we should be safe?

lib/types/src/memory_view.rs
```diff
@@ -64,6 +64,33 @@ where
...
+        Self {
+            ptr: unsafe { self.ptr.offset(start as isize) },
```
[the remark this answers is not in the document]
syrusakbary: Fixed I used ptr.add instead

lib/api/src/ptr.rs
```diff
@@ -151,66 +152,72 @@ impl<T: Copy + ValueType> WasmPtr<T, Array> {
...
-    pub fn deref(self, memory: &Memory, index: u32, length: u32) -> Option<&[Cell<T>]> {
+    pub fn deref(self, memory: &Memory, index: u32, length: u32) -> Option<Vec<WasmCell<'a, T>>> {
         // gets the size of the item in the array with padding added such that
         // for any index, we will always result an aligned memory access
-        let item_size = mem::size_of::<T>();
-        let slice_full_len = index as usize + length as usize;
-        let memory_size = memory.size().bytes().0;
+        let item_size = mem::size_of::<T>() as u32;
+        let slice_full_len = index + length;
+        let memory_size = memory.size().bytes().0 as u32;
```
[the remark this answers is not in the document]
syrusakbary: Got it. Now it should be fixed

lib/types/src/memory_view.rs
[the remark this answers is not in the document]
MarkMcCaskey: So you _can_ do it with `u32`s in a lot of places, it's just way harder and error prone, so I'd strongly recommend against it if at all possible... In fact anywhere we use `usize` for wasm memory may be problematic in 32bit contexts. Especially in Wasm ptr code, though, it's best to just use a u64 for 32bit wasm memory math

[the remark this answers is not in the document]
syrusakbary: Yeah, I just did the conversion on the `end` instead. Is more resilient

lib/api/src/ptr.rs
```diff
@@ -151,65 +131,40 @@ impl<T: Copy + ValueType> WasmPtr<T, Array> {
...
+                        memory
+                            .view::<u8>()
+                            .as_ptr()
+                            .add((self.offset as usize + i * item_size))
+                            as usize,
+                        mem::align_of::<T>(),
+                    ) as *const Cell<T>;
+                    WasmCell::new(&*cell_ptr)
+                })
+                .collect::<Vec<_>>(),
```
MarkMcCaskey, line 166: ⛳ 
This can be written with a smaller unsafe block and doing less work as:

```Rust
let cell_ptr = align_pointer(memory.view::<u8>().as_ptr().add(self.offset as usize) as usize, mem::align_of::<T>()) as *const Cell<T>;
let cell_ptrs = &std::slice::from_raw_parts(cell_ptr, slice_full_len)
            [index as usize..slice_full_len];
cell_ptrs.iter().map(|ptr| WasmCell::new(unsafe { &*cell_ptr })).collect::<Vec<_>>()
```

lib/types/src/memory_view.rs
```diff
@@ -64,6 +64,33 @@ where
             _phantom: PhantomData,
         }
     }
+
+    /// Creates a subarray view from this MemoryView.
+    pub fn subarray(&self, start: u32, end: u32) -> Self {
```
MarkMcCaskey, line 69: We can probably use a https://doc.rust-lang.org/std/ops/struct.Range.html here and be able to do it  as `x.subarray(start..end)`; or some generic type using https://doc.rust-lang.org/std/ops/trait.RangeBounds.html for full flexibility

lib/types/src/memory_view.rs
```diff
@@ -64,6 +64,33 @@ where
...
+        assert!(start <= end);
+        assert!((end as usize) < self.length);
+        Self {
+            ptr: unsafe { self.ptr.add(start as usize) },
```
MarkMcCaskey, line 73: ptr.add accounts for size, does the bounds checking code above account for size? If length is already in terms of `T` and not `u8` then this should be good 👍

lib/types/src/memory_view.rs
```diff
@@ -64,6 +64,33 @@ where
...
+        assert!(
+            src.len() == self.length,
+            "The source length must match the MemoryView length"
+        );
+        unsafe {
+            for (i, byte) in src.iter().enumerate() {
+                *self.ptr.offset(i as isize) = *byte;
+            }
+        }
+    }
```
MarkMcCaskey, line 93: This writes to shared memory without synchronization, this method seems like it should be `unsafe`. 🤔 I mean technically data races are allowed in Wasm linear memory, I suppose we should just never allow any kind of `&` into wasm memory to be correct on the Rust side

lib/wasi/src/syscalls/mod.rs
```diff
@@ -76,16 +75,18 @@ fn write_bytes<T: Write>(
 fn read_bytes<T: Read>(
     mut reader: T,
     memory: &Memory,
-    iovs_arr_cell: &[Cell<__wasi_iovec_t>],
+    iovs_arr_cell: Vec<WasmCell<__wasi_iovec_t>>,
```
MarkMcCaskey, line 78: This should take a `&[]` to be more generic, it's not relevant to this function that the data is  stored in a `Vec`

lib/wasi/src/syscalls/mod.rs
```diff
@@ -66,7 +65,7 @@ fn write_bytes_inner<T: Write>(
 fn write_bytes<T: Write>(
     mut write_loc: T,
     memory: &Memory,
-    iovs_arr_cell: &[Cell<__wasi_ciovec_t>],
+    iovs_arr_cell: Vec<WasmCell<__wasi_ciovec_t>>,
```
MarkMcCaskey, line 68: same here

lib/wasi/src/syscalls/mod.rs
```diff
@@ -47,7 +46,7 @@ pub use windows::*;
 fn write_bytes_inner<T: Write>(
     mut write_loc: T,
     memory: &Memory,
-    iovs_arr_cell: &[Cell<__wasi_ciovec_t>],
+    iovs_arr_cell: Vec<WasmCell<__wasi_ciovec_t>>,
```
MarkMcCaskey, line 49: same here

lib/wasi/src/syscalls/mod.rs
```diff
@@ -76,16 +75,18 @@ fn write_bytes<T: Write>(
...
 ) -> Result<u32, __wasi_errno_t> {
     let mut bytes_read = 0;
 
     for iov in iovs_arr_cell {
         let iov_inner = iov.get();
-        let bytes = WasmPtr::<u8, Array>::new(iov_inner.buf).deref(memory, 0, iov_inner.buf_len)?;
-        let mut raw_bytes: &mut [u8] =
-            unsafe { &mut *(bytes as *const [_] as *mut [_] as *mut [u8]) };
-        bytes_read += reader.read(raw_bytes).map_err(|_| __WASI_EIO)? as u32;
+        let mut raw_bytes: Vec<u8> = vec![0; iov_inner.buf_len as usize];
```
MarkMcCaskey, line 84: To slightly mitigate the extra slowdown we can move the vector outside the loop, call `raw_bytes.clear()` on every loop to reset it and then `raw_bytes.resize(iov_inner.buf_len as usize, 0)`  this will prevent reallocation over and over again. Note that `resize` never shrinks the underlying allocation so it'll just make it bigger if needs to.

Though honestly this is a prime candidate for optimization, I think this code is likely the hardest hit by these API changes.

lib/wasi/src/syscalls/mod.rs
```diff
@@ -76,16 +75,18 @@ fn write_bytes<T: Write>(
...
+        bytes_read += reader.read(&mut raw_bytes).map_err(|_| __WASI_EIO)? as u32;
+        memory
+            .view::<u8>()
+            .subarray(iov_inner.buf, iov_inner.buf + iov_inner.buf_len)
```
MarkMcCaskey, line 88: This will now panic given inputs that it would not panic under before, we would just return the correct error before.

I think this is a bit of a regression in API usability

lib/wasi/src/syscalls/mod.rs
```diff
@@ -2515,18 +2516,19 @@ pub fn proc_raise(env: &WasiEnv, sig: __wasi_signal_t) -> __wasi_errno_t {
...
-        getrandom::getrandom(u8_buffer)
-    };
+    let mut u8_buffer = vec![0; buf_len as usize];
+    let res = getrandom::getrandom(&mut u8_buffer);
     match res {
-        Ok(()) => __WASI_ESUCCESS,
+        Ok(()) => {
+            memory
+                .view::<u8>()
+                .subarray(buf, buf + buf_len)
```
MarkMcCaskey, line 2528: Same here, we will now panic given inputs that we previously handled correctly

lib/wasi/src/syscalls/mod.rs
MarkMcCaskey: To clarify: the issue is u32 overflow. This new API pushes it onto the caller when before we just handled it internally

lib/types/src/memory_view.rs
syrusakbary: Yeah, length is in terms of `T`.

lib/types/src/memory_view.rs
syrusakbary: Documented!

lib/wasi/src/syscalls/mod.rs
syrusakbary: Fixed. I transformed them to `usize`

lib/wasi/src/syscalls/mod.rs
syrusakbary: Fixed. I transformed them to `usize`

lib/wasi/src/syscalls/mod.rs
syrusakbary: Good call!

syrusakbary: Done

lib/wasi/src/syscalls/mod.rs
syrusakbary: Fixed

lib/wasi/src/syscalls/mod.rs
syrusakbary: Fixed

lib/wasi/src/syscalls/mod.rs
syrusakbary: Fixed

lib/types/src/memory_view.rs
syrusakbary: Done!

lib/api/src/ptr.rs
syrusakbary: Done!

syrusakbary:
bors try

syrusakbary:
bors try

lib/types/src/memory_view.rs
```diff
@@ -64,6 +67,44 @@ where
...
+            Bound::Unbounded => self.length,
+            Bound::Included(end) => *end,
+            Bound::Excluded(end) => *end - 1,
+        };
+        Self {
+            ptr: unsafe { self.ptr.add(start) },
+            length: (end - start),
+            _phantom: PhantomData,
+        }
+    }
```
Hywan, line 88: We must ensure it doesn't overflow outside the memory bounds.

lib/types/src/memory_view.rs
```diff
@@ -64,6 +67,44 @@ where
...
+    /// This function will efficiently copy the memory from within the wasm
+    /// module’s own linear memory to this typed array.
+    ///
+    /// # Safety
+    ///
+    /// This method is unsafe because the caller will need to make sure
+    /// there are no data races when copying memory into the view.
+    pub unsafe fn copy_from(&self, src: &[T]) {
+        assert!(
+            src.len() == self.length,
```
Hywan, line 101: If we change for `src.len() >= self.length`, we could be able to write from a slice that is smaller than the memory view. It can be useful. Ideally, we should also support an offset. I don't know what's your idea here: Do we want to provide a richer API, or just a super basic one?

lib/types/src/memory_view.rs
syrusakbary: Fixed!

lib/types/src/memory_view.rs
syrusakbary: Fixed, I actually remove the assertion. If the `src.len() < self.length` we actually just copy it's totality, and if it's bigger we get the smaller slice.

syrusakbary:
bors try

syrusakbary:
I assume everything is good to go now, given that all feedback is addressed and it was previously approved. Merging manually :)

Issue #2439, opened 2021-06-22
Trap messages get an extra null terminator


liamappelbe:
### Describe the bug
Trap messages get an extra null terminator, even if they're already null terminated.

I'm using wasmer-c-api version 1.0.2 on Windows.

### Steps to reproduce
As a work around to #2438, I'm trying to set the trap message to a unique string to use as a key in a map. I create a trap using `wasm_trap_new`, and pass it a `wasm_byte_vec_t` like `[100, 97, 114, 116, 58, 99, 49, 119, 109, 119, 97]`. This trap is returned from 
a function created using `wasm_func_new_with_env`. When I catch the trap returned by `wasm_func_call`, the `wasm_byte_vec_t` is `[100, 97, 114, 116, 58, 99, 49, 119, 109, 119, 97, 0]`.

I also tried explicitly null terminating the `wasm_byte_vec_t`, and then the returned vector is `[100, 97, 114, 116, 58, 99, 49, 119, 109, 119, 97, 0, 0]`.

This means that the returned message doesn't match the one I'm using as a key in my map. I'll work around it for now by just dropping any null terminators before using it as a map key.

### Expected behavior
The `wasm_byte_vec_t` message on the trap returned by `wasm_func_call` should be the same as the one returned by the callback function.

### Actual behavior
The `wasm_byte_vec_t` gets an extra 0 at the end.

syrusakbary:
Thanks for the great analysis @liamappelbe, we'll take a look at it soon!

Hywan:
The null byte is added here:

https://github.com/wasmerio/wasmer/blob/6773109fe0cf2fea78376939e084c097d37a9629/lib/c-api/src/wasm_c_api/trap.rs#L34-L47

Hywan:
It comes from the standard `wasm.h` file:

https://github.com/WebAssembly/wasm-c-api/blob/c9d31284651b975f05ac27cee0bab1377560b87e/include/wasm.h#L378

which we use and is up-to-date:

https://github.com/wasmerio/wasmer/blob/6773109fe0cf2fea78376939e084c097d37a9629/lib/c-api/tests/wasm-c-api/include/wasm.h#L378-L384

The Trap API uses `wasm_message_t`, which is like `wasm_name_t` but the message is null terminated.

Hywan:
Classical way to create a trap message is the following:

```c
  own wasm_message_t trap_message;
[...]

PR #2444, opened 2021-06-25
 fix(c-api) Trap's messages are always null terminated

Hywan:
# Description

`wasm_trap_new` expects a `wasm_message_t`. It's a type alias to
`wasm_name_t` with the exception that it represents a null-terminated
string.

When calling `wasm_trap_new`, no check was present to ensure the
string was well-formed. That's a first issue. But in the best
scenario, the string was correctly formed and was
null-terminated. This string was transformed to a Rust `String` —with
the null byte!— and passed to `RuntimeError`.

Then in `wasm_trap_message`, another null byte was pushed at the end
of the message. It's been introduced in
#1947. It results in a
doubly-null-terminated string, which is incorrect.

This patch does the following:

1. It checks that the string given to `wasm_trap_new` contains a
   null-terminated string or not, and will act accordingly. Note that
   it's possible to pass a non-null-terminated string, and it will
   still work because this detail is vicious. The idea is to get a
   well-formed `RuntimeError` in anycase.
    * If no null byte is found, the string is passed to `RuntimeError`
    as a valid Rust string,
    * If a null byte is found at the end of the string, a new string is
    passed to `RuntimeError` but without the final null byte,
    * If a null byte is found but not at the end, it's considered as an
    error,
    * If the string contains invalid UTF-8 bytes, it's considered as an
    error.
2. It updates `wasm_trap_message` to always add a null byte at the end
   of the returned owned string.
3. It adds test cases when passing a null-terminated or a
   non-null-terminated string to `wasm_trap_new` and to compare the
   results to `wasm_trap_message`.

It fixes https://github.com/wasmerio/wasmer/issues/2439.

# Review

### DIFF
--- a/lib/c-api/CHANGELOG.md
+++ b/lib/c-api/CHANGELOG.md
@@ -8,6 +8,11 @@ Looking for changes to the Wasmer CLI and the Rust API? See our [Primary Changel
 
 ## **[Unreleased]**
 
+### Fixed
+- [#2444](https://github.com/wasmerio/wasmer/pull/2444) Trap's messages are always null terminated.
+
+## 2.0.0 - 2020/06/16
+
 ## 2.0.0-rc1 - 2020/06/02
 
 ### Added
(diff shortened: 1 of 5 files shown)


Hywan:
bors try

syrusakbary (approved):

syrusakbary:
Merging directly as bors already passed

PR #2455, opened 2021-07-08
cargo: don't pull in both universal and dylib engines

vavrusa:
Selecting any compiler would pull in both engines due to */compiler feature propagation and there is no way to disable it.
Depending just on wasmer-compiler/translator should fix the issue and allow depending on just one engine.

# Description

# Review

### DIFF
--- a/lib/api/Cargo.toml
+++ b/lib/api/Cargo.toml
@@ -46,8 +46,6 @@ maintenance = { status = "actively-developed" }
 default = ["wat", "default-cranelift", "default-universal"]
 compiler = [
     "wasmer-compiler/translator",
-    "wasmer-engine-universal/compiler",
-    "wasmer-engine-dylib/compiler",
 ]
 engine = []
 universal = [


vavrusa:
It's not ideal because it requires depending on chosen engine directly to be able to set the compiler features, e.g. `wasmer-engine-dylib = { version = "2.0.0", feature = ["compiler"] }`. I'm not sure if there's a good way to make a feature matrix in Cargo, so it'd have to have features like `llvm-dylib, llvm-universal, ...`. I guess it would be better to make a Headless a type of Compiler to be able to have a `<engine>::new(..)` method accepting either an actual compiler or headless.

Hywan:
Hello,

Nor `wasmer-compiler-llvm`, or `wasmer-compiler-cranelift`, or `wasmer-compiler-singlepass` has a dependency to an engine (be `wasmer-engine-dylib`, `wasmer-engine-staticlib` or `wasmer-engine-universal`).

I'm not sure I understand your question. Can you provide more information please?

Hywan:
Hello,

Thank you for your issue.
The crates may be pulled but not compiled. Is it an issue to pull them?

vavrusa:
It seems like it does because compiler flag sets /compiler flag on both engines which seems to pull them and compile. Perhaps I'm doing something wrong?

```
$ cargo new --lib test-lib && cd test-lib
$ cat Cargo.toml
...
[dependencies]
wasmer = { version = "2.0", default-features = false, features = ["llvm", "dylib"] }

$ cargo build
...
   Compiling wasmer-types v2.0.0
   Compiling wasmer-compiler v2.0.0
   Compiling wasmer-engine v2.0.0
   Compiling wasmer-object v2.0.0
   Compiling wasmer-compiler-llvm v2.0.0
   Compiling wasmer-engine-dylib v2.0.0
   Compiling wasmer-engine-universal v2.0.0
   Compiling wasmer v2.0.0
   Compiling test-lib v0.1.0 (/Users/vavrusa/cf-repos/test-lib)
    Finished dev [unoptimized + debuginfo] target(s) in 1m 56s
```

Both `universal` and `dylib` were built?

syrusakbary:
@vavrusa the issue you commented is completely on point. The downside is that both engine crates will be compiled either you want it or not.

The issue with fixing it like this PR is doing is that it will make developers need to specify via a feature which engine and compiler they want to use. Rather than one or the other, which makes it for a worse development experience. Meaning: the current flow will not work.


On the way that Wasmer is designed, you can actually pull the engines directly yourself and use it in Wasmer normally.

Such as:

```toml
[dependencies]
wasmer = { version = "2.0", default-features = false }
wasmer-engine-dylib = { version = "2.0", features = ["compiler"] }
wasmer-compiler-llvm = "2.0"
```

And then use it like:
```rust
use wasmer_compiler_llvm::LLVM;
use wasmer_engine_dylib::Dylib;

let compiler = LLVM::new();
// Put it into an engine and add it to the store
let store = Store::new(&Dylib::new(compiler).engine());
```

I know that this solution is also not as ideal. I'm going to think more about this to find perhaps other possible solutions.

vavrusa:
This is actually a good enough workaround! I'm going to decline this PR, I agree it's not great. I think ideally engine shouldn't depend on the compiler flag, as headless should be just an empty compiler trait implementation.

PR #2462, opened 2021-07-13
feat: Add the `stale` bot

Hywan:
# Description

This patch configures (and thus finishes the installation) of the
“stale” bot, learn more at https://probot.github.io/apps/stale/.

Note: The bot/app has been installed already.

# Review

- ~[ ] Add a short description of the change to the CHANGELOG.md file~ not needed

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -6,7 +6,7 @@ daysUntilClose: 30
 exemptLabels:
   - "🐞 bug"
 # Label to use when marking an issue as stale
-staleLabel: "🚫 won't fix"
+staleLabel: "🏚 stale"
 # Comment to post when marking an issue as stale. Set to `false` to disable
 markComment: >
   This issue has been automatically marked as stale because it has not had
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 15
```
syrusakbary, line 4: I would be more conservative at the beginning.
I think this is a better setting:
```suggestion
daysUntilStale: 365
# Number of days of inactivity before a stale issue is closed
daysUntilClose: 30
```

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - "🐞 bug"
+# Label to use when marking an issue as stale
+staleLabel: "🚫 won't fix"
```
syrusakbary, line 9: Won't fix and stale are not the same. I'd like to have a label that we can filter after and use to just filter stale issues.
```
staleLabel: "🏚 stale"
```

Hywan:
```suggestion
staleLabel: "🏚 stale"
```

PR #2474, opened 2021-07-16
Fix windows cache permission

### LINKED ISSUE #2454: Wasmer cli requires admin privilege on Windows due to cache
### Describe the bug

Trying to run a wasm module with the wasmer cli in a normal user level cli results in:
```sh
> wasmer exorbitant.wasm
error: failed to run `exorbitant.wasm`
│   1: module instantiation failed (engine: universal, compiler: cranelift)
╰─> 2: Access is denied. (os error 5)
```
Windows 10 using Wasmer 2.0 installed from the Windows Installer in the release.

### Steps to reproduce

From a user-privilege cli run: `wasmer <some wasm file>`

### Expected behavior

I expect the cli to be able to run and cache build results in user accessible folders. I do not expect to need admin privileges to run wasmer. 

### Additional context

Workarounds:
1. In a user-privilege cli disable the cache:
    ```sh
    wasmer run --disable-cache exorbitant.wasm
    ```

2. Run from a cmd with admin privledges:
    1. Press start and type cmd
    2. Right click on the "Command Prompt" option and choose "Run as administrator" (approve the UAC prompt as needed)
    3. Run wasmer in the admin prompt

Edit: The wasm file used: https://wapm.io/package/rajsite/exorbitant

chenyukang:
# Description
Fix issue #2454 , default `WASMER_CACHE_DIR` is not right for Windows.
Add more detailed messages for creating cache directory failure.

close #2454.

### DIFF
--- a/lib/cache/src/filesystem.rs
+++ b/lib/cache/src/filesystem.rs
@@ -63,8 +63,15 @@ impl FileSystemCache {
             }
         } else {
             // Create the directory and any parent directories if they don't yet exist.
-            create_dir_all(&path)?;
-            Ok(Self { path, ext: None })
+            let res = create_dir_all(&path);
+            if res.is_err() {
+                Err(io::Error::new(
+                    io::ErrorKind::Other,
+                    format!("failed to create cache directory: {}", path.display()),
+                ))
+            } else {
+                Ok(Self { path, ext: None })
+            }
         }
     }
 
--- a/scripts/windows-installer/wasmer.iss
+++ b/scripts/windows-installer/wasmer.iss
@@ -17,8 +17,6 @@ DisableWelcomePage=no
 [Registry]
 Root: HKCU; Subkey: "Environment"; ValueType:string; ValueName: "WASMER_DIR"; \
     ValueData: "{app}"; Flags: preservestringtype
-Root: HKCU; Subkey: "Environment"; ValueType:string; ValueName: "WASMER_CACHE_DIR"; \
-    ValueData: "{app}\cache"; Flags: preservestringtype
 
 [Files]
 Source: "..\..\package\bin\*"; DestDir: "{app}\bin"
@@ -27,9 +25,6 @@ Source: "..\..\package\lib\*"; DestDir: "{app}\lib"
 Source: "..\..\package\LICENSE"; DestDir: "{app}"
 Source: "..\..\package\ATTRIBUTIONS"; DestDir: "{app}"
 Source: "wax.cmd"; DestDir: "{app}\bin"
-
-[Dirs]
-Name: "{app}\cache"
 
 [Code]
 const EnvironmentKey = 'SYSTEM\CurrentControlSet\Control\Session Manager\Environment';
--- a/scripts/windows-installer/wasmer.iss
+++ b/scripts/windows-installer/wasmer.iss
@@ -17,6 +17,8 @@ DisableWelcomePage=no
 [Registry]
 Root: HKCU; Subkey: "Environment"; ValueType:string; ValueName: "WASMER_DIR"; \
     ValueData: "{app}"; Flags: preservestringtype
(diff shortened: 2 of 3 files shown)


lib/cache/src/filesystem.rs
```diff
@@ -63,8 +63,15 @@ impl FileSystemCache {
             }
         } else {
             // Create the directory and any parent directories if they don't yet exist.
-            create_dir_all(&path)?;
-            Ok(Self { path, ext: None })
+            let res = create_dir_all(&path);
+            if res.is_err() {
+                Err(io::Error::new(
+                    io::ErrorKind::Other,
```
Hywan, line 69: Why not using `io::ErrorKind::PermissionDenied` here? (cf https://doc.rust-lang.org/std/io/enum.ErrorKind.html)

Hywan (commented):
Thank you for the PR! One last question before approving the PR.

lib/cache/src/filesystem.rs
chenyukang: I was also hesitant to choose between `PermissionDenied` and `Other`, the reason I choose `Other` is considering we have some other reasons for failing to create a directory ....

`PermissionDenied` is also OK, since I mainly want to print a detailed message when the cache directory is failed to create.

scripts/windows-installer/wasmer.iss
```diff
@@ -17,8 +17,6 @@ DisableWelcomePage=no
 [Registry]
 Root: HKCU; Subkey: "Environment"; ValueType:string; ValueName: "WASMER_DIR"; \
     ValueData: "{app}"; Flags: preservestringtype
-Root: HKCU; Subkey: "Environment"; ValueType:string; ValueName: "WASMER_CACHE_DIR"; \
-    ValueData: "{app}\cache"; Flags: preservestringtype
```
syrusakbary, line 21: If not provided, `WASMER_CACHE_DIR` will be `WASMER_DIR + "/cache"`

In this case, I think we should go to the value that worked before:

```suggestion
Root: HKCU; Subkey: "Environment"; ValueType:string; ValueName: "WASMER_CACHE_DIR"; \
    ValueData: "{%USERPROFILE}\.wasmer\cache"; Flags: preservestringtype
```

scripts/windows-installer/wasmer.iss
```diff
@@ -28,9 +26,6 @@ Source: "..\..\package\LICENSE"; DestDir: "{app}"
 Source: "..\..\package\ATTRIBUTIONS"; DestDir: "{app}"
 Source: "wax.cmd"; DestDir: "{app}\bin"
 
-[Dirs]
-Name: "{app}\cache"
```
syrusakbary, line 32:
```suggestion
[Dirs]
Name: "{%USERPROFILE}\.wasmer"
```

scripts/windows-installer/wasmer.iss
chenyukang: I think it's better not to set a register default value for it on Windows. We can keep the consistent usage style with Unix/Linux OS.

If users want to specify a cache directory, they can set an environment variable WASMER_CACHE_DIR, otherwise, env::temp_dir() will be used and it's safe for caching files.

/// Get the cache dir
pub fn get_cache_dir() -> PathBuf {
    match env::var("WASMER_CACHE_DIR") {
        Ok(dir) => {
            let mut path = PathBuf::from(dir);
            path.push(VERSION);
            path
        }
        Err(_) => {
            // We use a temporal directory for saving cache files
            let mut temp_dir = env::temp_dir();
            temp_dir.push("wasmer");
            temp_dir.push(VERSION);
            temp_dir
        }
    }
}
What do you think? @syrusakbary

lib/cache/src/filesystem.rs
Hywan: Good to me.

syrusakbary (approved):
Looks great, merging!

PR #1939, opened 2020-12-15
Attempt to simplify build on linux

MarkMcCaskey:
Work in progress

Resolves #1928 (if it works)

# Review

- [ ] Add a short description of the the change to the CHANGELOG.md file

PR #2475, opened 2021-07-16
fix(api,c-api,vm) Move `.cargo/config` to `Cargo.toml`s

### LINKED ISSUE #1928: Undefined reference to `wasmer_raise_trap` when using native runtime
Hi,

I have some wasm code that compiles and runs fine using llvm/jit but does not run with llvm/native. I get the following error.
```
thread 'main' panicked at 'Failed to compile wasm: Compile(Codegen("/tmp/wasmer_nativeLFymlg.so: undefined symbol: wasmer_raise_trap"))', src/main.rs:27:19
```

It looks like wasmer is not linking something correctly, or maybe I need to define these symbols?

Hywan:
# Description

Rework of https://github.com/wasmerio/wasmer/pull/1939/.

We need to pass the `-Wl,-E` arguments to the linker when compiling `wasmer`, `wasmer_c_api` and `wasmer_vm`. It was done generally with a `.cargo/config.toml` file _but_ it was local to our project, i.e. it was not used when Wasmer was a dependency of another project.

So we move the content of `.cargo/config.toml` to the respective `Cargo.toml` files.

Fixes #1939.
Fixes #1928.

# Review

- [ ] ~Add a short description of the change to the CHANGELOG.md file~ not necessary I guess.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,0 @@
-[target.'cfg(target_os = "linux")']
-rustflags = [
-    # Put the VM functions in the dynamic symbol table.
-    "-C", "link-arg=-Wl,-E",
-]
--- a/lib/api/Cargo.toml
+++ b/lib/api/Cargo.toml
@@ -42,6 +42,11 @@ anyhow = "1.0"
 [badges]
 maintenance = { status = "actively-developed" }
 
+[target.'cfg(target_os = "linux")']
+rustflags = [
+  "-C", "link-arg=-Wl,-E"
+]
+
 [features]
 default = ["wat", "default-cranelift", "default-universal"]
 compiler = [
--- a/lib/c-api/Cargo.toml
+++ b/lib/c-api/Cargo.toml
@@ -47,6 +47,11 @@ paste = "1.0"
 field-offset = "0.3.3"
 inline-c = "0.1.5"
 
+[target.'cfg(target_os = "linux")']
+rustflags = [
+  "-C", "link-arg=-Wl,-E"
+]
+
 [features]
 default = [
     "wat",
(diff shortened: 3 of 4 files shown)


Hywan:
bors try

syrusakbary:
I had no idea this was possible.
I think we might need to add to the `wasmer_vm` as well

Hywan:
Actually I'm not sure it's possible. I may have confused https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies with https://doc.rust-lang.org/cargo/reference/config.html?highlight=rustflags#configuration-format.

Hywan:
I believe we must write `build.rs` scripts that emit a https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname.

Hywan:
Er, we can't configure `RUSTFLAGS` with a `build.rs` script, and my previous idea of using `cargo-rustc-link-lib` is wrong.

Hywan:
Reference issue, https://github.com/rust-lang/cargo/issues/7984

Hywan:
What we need exactly is https://github.com/rust-lang/cargo/pull/9557: The `rustc-link-arg` option.

I'm going to try with `rustc-cdylib-link-arg` which is stable now, but is going to be renamed `rustc-link-arg-cdylib` in the future if I've understood everything correctly.

Hywan:
bors try

Hywan:
bors try

Hywan:
Nah, we need to wait :-).

syrusakbary:
Should we close this PR @Hywan ?

Hywan:
We can keep it open and wait on Rust.

PR #2482, opened 2021-07-19
doc(makefile) Make `sed -i` cross-platform to fix `wasmer-c-api`'s doc.

Hywan:
# Description

When generating the documentation for `wasmer-c-api`, we need to
modify its `Cargo.toml` file. For that, we use `sed`. But we must use
the `-i` argument (for `in-place` modifications). Unfortunately, it's
different between macOS and Linux. This patch makes it cross-platform
so that it runs on macOS and on Linux (incl. the CI).

Sequel of https://github.com/wasmerio/wasmer/pull/2457/.

# Review

- ~[ ] Add a short description of the change to the CHANGELOG.md file~ internal stuff

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -329,9 +329,25 @@ $(info --------------)
 $(info )
 $(info )
 
-############
-# Building #
-############
+#####
+#
+# Configure `sed -i` for a cross-platform usage.
+#
+#####
+
+SEDI ?=
+
+ifeq ($(IS_DARWIN), 1)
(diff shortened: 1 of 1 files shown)


Hywan:
I'm auto-merging this PR because: (i) it works on my machine (macOS), (ii) it will fix the `gh-pages` branch through the `.github/workflows/documentation.yml` workflow, so we don't need to run the tests here.

PR #2460, opened 2021-07-12
Added Wasmer Js API

syrusakbary:
# Description

This PR adds support for compiling Wasmer into JS/Wasm via wasm-bindgen, so 3rd party dependencies can use Wasmer to target the web also.

Things working:
* [x] Type introspection: we have to decide how to do it, either run our own parser on the data or rely on the future APIs for introspect Wasm types (this APIs are yet available in the browser, but a proposal is on the way: [wasm-js-types proposal](https://github.com/WebAssembly/js-types)).
* [x] Wasm-pack wasmer-js tests
* [x] Assert 0 warnings in lint
* [x] Allow for custom JS errors (via `RuntimeError`)
* Externals
   * [x] Functions with and without Environments
      * [x] Native functions
      * [x] Dynamic functions
   * [x] Memory
   * [x] Globals
   * [x] Tables

# Review

- [ ] Add a short description of the change to the CHANGELOG.md file

PR #2489, opened 2021-07-23
feat(api) Merge `js-api` into `api`

Hywan:
# Description

This patch takes the entire `wasmer-js` crate and merges it into the
`wasmer` crate.

Inside the `lib/api/src/` directory, there are 2 new directories:

1. a new `sys` directory, which contains the usual `wasmer` crate
   implementation,
2. a new directory `js`, which contains the implementation of
   `wasmer-js`.

The `Cargo.toml` file is still compatible. The `default` feature
fallbacks to `sys-default`, which enables the `sys` feature. All
features related to compilers or engines or anything else prior this
patch, activates the `sys` feature.

Parallel to that, there is a `js-default` and `js` features.

The `Cargo.toml` file is extensively documented to explain what are
dependencies, dev-dependencies, features and other sections related to
`sys` or to `js`.

There is a bug with `wasm_bindgen_test` where it doesn't compile or
look for tests in `tests/*/<test>.rs`. The hack is to name files
`tests/js_<test>.rs`. Ugly, but it works.

# How to review?

I think browsing https://github.com/wasmerio/wasmer/tree/js-api-into-api/lib/api/src can provide a nice overview.

Then the biggest contribution of this patch is the new `Cargo.toml`, https://github.com/wasmerio/wasmer/blob/js-api-into-api/lib/api/Cargo.toml. The trick is

https://github.com/wasmerio/wasmer/blob/9a5a10ff5273df9b52ad93b652af95f2d6d81765/lib/api/Cargo.toml#L43-L47

to avoid conflicts with `wasm-bindgen`. Dependencies are grouped in 3 categories: Shared, ones for `sys`,  ones for `js`.

The second biggest contribution is the `lib.rs`, https://github.com/wasmerio/wasmer/blob/js-api-into-api/lib/api/src/lib.rs. The trick to support `wasm-bindgen` correctly, is to declare the crate-type programmatically in `lib.rs` rather than in `Cargo.toml` as such: 

https://github.com/wasmerio/wasmer/blob/9a5a10ff5273df9b52ad93b652af95f2d6d81765/lib/api/src/lib.rs#L29-L30

Then, there is compiler errors that are raised if some features aren't correctly used, like:

https://github.com/wasmerio/wasmer/blob/9a5a10ff5273df9b52ad93b652af95f2d6d81765/lib/api/src/lib.rs#L458-L461

The rest is only moving files.

# Next steps

As soon as this PR is merged, and probably as soon as https://github.com/wasmerio/wasmer/pull/2460 is merged, then we could imagine refactoring `lib/api/` to remove duplicated code. Now it is possible :-).


Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

syrusakbary (approved):

PR #2491, opened 2021-07-24
Base support for WASI in wasmer-js

syrusakbary:
This PR adds support for WASI targeting Javascript


syrusakbary:
bors try

syrusakbary:
bors try

syrusakbary:
bors try

syrusakbary:
bors try

lib/vfs/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+edition = "2018"
+
+[dependencies]
+libc = { version = "^0.2", default-features = false, optional = true }
+thiserror = "1"
+tracing = { version = "0.1" }
+typetag = { version = "0.1", optional = true }
+serde = { version = "1.0", default-features = false, features = ["derive"], optional = true }
+
+[features]
```
Hywan, line 16:
```suggestion
[features]
default = ["host_fs"]
```

Let's define `host_fs` as the default file system, which is already the case and the expected behavior.

lib/vfs/src/host_fs.rs
```diff
@@ -0,0 +1,648 @@
...
+        fs::remove_dir(path).map_err(Into::into)
+    }
+    fn rename(&self, from: &Path, to: &Path) -> Result<(), FsError> {
+        fs::rename(from, to).map_err(Into::into)
+    }
+
+    fn remove_file(&self, path: &Path) -> Result<(), FsError> {
+        fs::remove_file(path).map_err(Into::into)
+    }
+    fn new_open_options(&self) -> OpenOptions {
```
Hywan, line 43:
```suggestion
    }

    fn new_open_options(&self) -> OpenOptions {
```

lib/vfs/src/host_fs.rs
```diff
@@ -0,0 +1,648 @@
...
+    }
+
+    fn metadata(&self, path: &Path) -> Result<Metadata, FsError> {
+        fs::metadata(path)
+            .and_then(fs_metadata_to_metadata)
+            .map_err(Into::into)
+    }
+}
+
+fn fs_metadata_to_metadata(metadata: fs::Metadata) -> Result<Metadata, std::io::Error> {
```
Hywan, line 54: Can't we replace this by a `TryFrom` implementation instead? It would be more elegant.

lib/vfs/src/host_fs.rs
```diff
@@ -0,0 +1,648 @@
...
+    fn open(
+        &mut self,
+        path: &Path,
+        conf: &OpenOptionsConfig,
+    ) -> Result<Box<dyn VirtualFile>, FsError> {
+        // TODO: handle create implying write, etc.
+        let read = conf.read();
+        let write = conf.write();
+        let append = conf.append();
+        let mut oo = std::fs::OpenOptions::new();
```
Hywan, line 116:
```suggestion
        let mut oo = fs::OpenOptions::new();
```

lib/vfs/src/host_fs.rs
```diff
@@ -0,0 +1,648 @@
...
+    }
+
+    #[cfg(unix)]
+    fn get_raw_fd(&self) -> Option<i32> {
+        use std::os::unix::io::AsRawFd;
+        Some(self.inner.as_raw_fd())
+    }
+    #[cfg(not(unix))]
+    fn get_raw_fd(&self) -> Option<i32> {
+        unimplemented!(
```
Hywan, line 351: I believe we may be able to use [`AaRawHandle`](https://doc.rust-lang.org/std/os/windows/io/trait.AsRawHandle.html) and [`RawHandle`](https://doc.rust-lang.org/std/os/windows/io/type.RawHandle.html) here. It's the size of a pointer (`type RawHandle = HANDLE;` and `type HANDLE = *mut c_void`), so we should update `i32` that represents a file descriptor.

Ideally, we should come with a type to represent a file descriptor:

```rust
type FileDescriptor = usize;

#[cfg(unix)]
impl From<RawFd> for FileDescriptor { … }

#[cfg(windows)]
impl From<RawHandle> for FileDescriptor { … }
```

lib/vfs/src/host_fs.rs
```diff
@@ -0,0 +1,648 @@
...
+
+        host_file_bytes_available(host_fd)
+    }
+
+    #[cfg(unix)]
+    fn get_raw_fd(&self) -> Option<i32> {
+        use std::os::unix::io::AsRawFd;
+        Some(self.inner.as_raw_fd())
+    }
+    #[cfg(not(unix))]
```
Hywan, line 349:
```suggestion
    #[cfg(windows)]
```

lib/vfs/src/host_fs.rs
```diff
@@ -0,0 +1,648 @@
...
+        {
+            use std::os::unix::fs::FileTypeExt;
+            (
+                filetype.is_char_device(),
+                filetype.is_block_device(),
+                filetype.is_socket(),
+                filetype.is_fifo(),
+            )
+        }
+        #[cfg(not(unix))]
```
Hywan, line 68:
```suggestion
        #[cfg(windows)]
```

lib/vfs/src/host_fs.rs
```diff
@@ -0,0 +1,648 @@
...
+        // success
+        0 => Ok(bytes_found.try_into().unwrap_or(0)),
+        libc::EBADF => Err(FsError::InvalidFd),
+        libc::EFAULT => Err(FsError::InvalidData),
+        libc::EINVAL => Err(FsError::InvalidInput),
+        _ => Err(FsError::IOError),
+    }
+}
+
+#[cfg(not(unix))]
```
Hywan, line 372:
```suggestion
#[cfg(windows)]
```

lib/vfs/src/host_fs.rs
```diff
@@ -0,0 +1,648 @@
...
+        let host_fd = self.get_raw_fd().unwrap();
+
+        host_file_bytes_available(host_fd)
+    }
+    #[cfg(unix)]
+    fn get_raw_fd(&self) -> Option<i32> {
+        use std::os::unix::io::AsRawFd;
+        Some(io::stdout().as_raw_fd())
+    }
+    #[cfg(not(unix))]
```
Hywan, line 460:
```suggestion
    #[cfg(windows)]
```

lib/vfs/src/host_fs.rs
```diff
@@ -0,0 +1,648 @@
...
+        let host_fd = self.get_raw_fd().unwrap();
+
+        host_file_bytes_available(host_fd)
+    }
+    #[cfg(unix)]
+    fn get_raw_fd(&self) -> Option<i32> {
+        use std::os::unix::io::AsRawFd;
+        Some(io::stderr().as_raw_fd())
+    }
+    #[cfg(not(unix))]
```
Hywan, line 551:
```suggestion
    #[cfg(windows)]
```

and so on

lib/vfs/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+use std::io::{self, Read, Seek, Write};
+use std::path::{Path, PathBuf};
+use thiserror::Error;
+use tracing::debug;
+
+#[cfg(feature = "host_fs")]
+pub mod host_fs;
+//pub mod vfs_fs;
+#[cfg(feature = "mem_fs")]
+pub mod mem_fs;
```
Hywan, line 13: I think both features are mutually exclusive.

We should add something like this:

```rust
#[cfg(all(feature = "host_fs", feature = "mem_fs"))]
compile_error!("…");
```

Thoughts?

lib/vfs/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+use std::ffi::OsString;
+use std::fmt;
+use std::io::{self, Read, Seek, Write};
+use std::path::{Path, PathBuf};
+use thiserror::Error;
+use tracing::debug;
+
+#[cfg(feature = "host_fs")]
+pub mod host_fs;
+//pub mod vfs_fs;
```
Hywan, line 11:
```suggestion
```

lib/vfs/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+
+    /// the time at which the file was created in nanoseconds as a UNIX timestamp
+    fn created_time(&self) -> u64;
+
+    /// the size of the file in bytes
+    fn size(&self) -> u64;
+
+    /// Change the size of the file, if the `new_size` is greater than the current size
+    /// the extra bytes will be allocated and zeroed
+    fn set_len(&mut self, new_size: u64) -> Result<(), FsError>;
```
Hywan, line 149: Usually, this method is named `truncate`. I think it's convey the intend much better. Thought?

lib/vfs/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+            io::ErrorKind::Other => FsError::IOError,
+            // if the following triggers, a new error type was added to this non-exhaustive enum
+            _ => FsError::UnknownError,
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct ReadDir {
+    // TODO: to do this properly we need some kind of callback to the core FS abstraction
```
Hywan, line 312: Need to be addressed, right?

lib/vfs/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+impl ReadDir {
+    pub fn new(data: Vec<DirEntry>) -> Self {
+        Self { data, index: 0 }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct DirEntry {
+    pub path: PathBuf,
+    // weird hack, to fix this we probably need an internal trait object or callbacks or something
```
Hywan, line 326: Why is it a weird hack?

lib/vfs/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+    pub metadata: Result<Metadata, FsError>,
+}
+
+impl DirEntry {
+    pub fn path(&self) -> PathBuf {
+        self.path.clone()
```
Hywan, line 332: Why cloning?

lib/vfs/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    pub fn metadata(&self) -> Result<Metadata, FsError> {
+        self.metadata.clone()
```
Hywan, line 336: Why cloning?

lib/vfs/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+    pub fn file_name(&self) -> OsString {
+        self.path
+            .file_name()
+            .unwrap_or_else(|| self.path.as_os_str())
+            .to_owned()
+    }
+}
+
+#[derive(Clone, Debug)]
+// TODO: review this, proper solution would probably use a trait object internally
```
Hywan, line 353: I don't understand what's wrong with this code. Can you elaborate please?

Hywan: Oh I see, maybe we should implement `Metadata` as a trait, and not as a struct. And I think it's a good idea indeed!

lib/vfs/src/vfs_fs.rs
```diff
@@ -0,0 +1,183 @@
+use crate::*;
+use std::io::{Read, Seek, Write};
+use std::path::Path;
+use std::sync::Arc;
+
+pub use crate::host_fs::{Stderr, Stdin, Stdout};
+
+#[derive(Clone)]
+pub struct VfsFileSystem {
```
Hywan, line 9: What's the difference between a virtual file system file system (I'm just expandind the name here :-p), and an in-memory file system?

lib/wasi-experimental-io-devices/Cargo.toml
```diff
@@ -14,7 +14,7 @@ edition = "2018"
 maintenance = { status = "experimental" }
 
 [dependencies]
-wasmer-wasi = { version = "2.0.0", path = "../wasi" }
+wasmer-wasi = { version = "2.0.0", path = "../wasi", default-features=false }
```
Hywan, line 17: Why don't we want the default features?

lib/wasi/Cargo.toml
```diff
@@ -10,22 +10,52 @@ license = "MIT"
...
 logging = ["tracing/log"]
-disable-all-logging = ["tracing/release_max_level_off", "tracing/max_level_off"]
\ No newline at end of file
+disable-all-logging = [
+    "tracing/release_max_level_off",
+    "tracing/max_level_off"
+]
+host_fs = ["wasmer-vfs/host_fs"]
+mem_fs = ["wasmer-vfs/mem_fs"]
+js = [
```
Hywan, line 47: Just like we did with `wasmer` (`lib/api/`), I think it's good to differentiate between `sys` and `js`. Thought?

lib/wasi/Cargo.toml
```diff
@@ -10,22 +10,52 @@ license = "MIT"
...
+]
+test-js = [
+    "js",
+    "wasmer/js-default",
+    "wasmer/wat"
+]
+
+[dev-dependencies]
+wasm-bindgen = "0.2"
+wasm-bindgen-test = "0.3.0"
```
Hywan, line 60:
```suggestion
[target.'cfg(target_arch = "wasm32")'.dependencies]
wasm-bindgen = "0.2"

[target.'cfg(target_arch = "wasm32")'.dev-dependencies]
wasm-bindgen-test = "0.3.0"
```

`wasm-bindgen-test` is a `dev-dependencies`. And both are for `wasm32` only.

lib/wasi/src/macros.rs
```diff
@@ -31,6 +31,6 @@ macro_rules! wasi_try {
 /// corrupted.
 macro_rules! get_input_str {
     ($memory:expr, $data:expr, $len:expr) => {{
-        wasi_try!($data.get_utf8_str($memory, $len), __WASI_EINVAL)
+        wasi_try!($data.get_utf8_string($memory, $len), __WASI_EINVAL)
```
Hywan, line 34: Why not keeping `get_utf8_str`? It returns a `Cow`, so you can easily transform it into a `String` if needed.

lib/wasi/src/ptr.rs
```diff
@@ -82,7 +82,16 @@ impl<T: Copy + ValueType> WasmPtr<T, Array> {
...
+        self,
+        memory: &'a Memory,
+        str_len: u32,
+    ) -> Option<std::borrow::Cow<'a, str>> {
+        self.0.get_utf8_str(memory, str_len).map(Into::into)
+    }
+
+    #[inline(always)]
+    pub unsafe fn get_utf8_string(self, memory: &Memory, str_len: u32) -> Option<String> {
+        self.0.get_utf8_string(memory, str_len)
```
Hywan, line 95: Not tested, but I reckon it can be written as follows:

```suggestion
        self.get_utf8_str(memory, str_len)?.to_owned()
```

What do you think?

lib/wasi/src/state/builder.rs
```diff
@@ -37,15 +38,18 @@ pub struct WasiStateBuilder {
...
-    stdin_override: Option<Box<dyn WasiFile>>,
+    stdout_override: Option<Box<dyn VirtualFile>>,
+    stderr_override: Option<Box<dyn VirtualFile>>,
+    stdin_override: Option<Box<dyn VirtualFile>>,
+    fs_override: Option<Box<dyn wasmer_vfs::FileSystem>>,
 }
 
 impl std::fmt::Debug for WasiStateBuilder {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // TODO: update this when stable
```
Hywan, line 52: ⚠️ There is a todo here that must be addressed at the end of the review process.

lib/wasi/src/state/builder.rs
```diff
@@ -76,7 +80,7 @@ pub enum WasiStateCreationError {
     #[error("wasi filesystem setup error: `{0}`")]
     WasiFsSetupError(String),
     #[error(transparent)]
-    WasiFsError(WasiFsError),
+    FsError(FsError),
```
Hywan, line 83: The other variants are named `WasiFs*`, so let's keep `WasiFsError` here?

lib/wasi/src/state/builder.rs
```diff
@@ -359,29 +385,35 @@ impl WasiStateBuilder {
             }
         }
 
-        // self.preopens are checked in [`PreopenDirBuilder::build`]
+        // using `take` here is a bit of a code smell
```
Hywan, line 388: Yeah, totally agree. If the `build` method was taking ownership of `self` rather than a mutable reference, that would not be an issue because `self` wouldn't be any longer usable. But in this case, it changes the behavior of the object. I don't like it at all :-). It must be changed!

lib/wasi/src/state/mod.rs
```diff
@@ -163,12 +168,111 @@ pub struct WasiFs {
...
+        } else if #[cfg(feature = "mem_fs")] {
+            Box::new(wasmer_vfs::mem_fs::MemFileSystem::default())
+        } else {
+            Box::new(FallbackFileSystem::default())
+        }
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct FallbackFileSystem;
```
Hywan, line 189: `FallbackFileSystem` is not required if at least `host_fs` _or_ `mem_fs` is required. One of them is required. But they are mutually exclusive.

I think it's the expected behavior here. WASI without a FS is… useless…, right?

lib/wasi/src/syscalls/mod.rs
```diff
@@ -90,9 +96,10 @@ fn read_bytes<T: Read>(
         bytes_read += reader.read(&mut raw_bytes).map_err(|_| __WASI_EIO)? as u32;
         unsafe {
             memory
-                .view::<u8>()
+                .uint8view()
```
Hywan, line 99: Let's keep `view::<u8>` here, it's just simple.

Hywan (changes requested):

Hywan:
This PR adds the following:

* It introduces `wasmer-vfs` (from @MarkMcCaskey I suppose). Please, present this crate a little bit and provide examples :-).
* It introduces the `--verbose` option to `wasmer` CLI,
* It allows to transform `ImportObject` to a JavaScript object,
* It removes `serde` from `wasmer-c-api`, which wasn't exposed to the public API,
* It uses `wasmer-vfs` in `wasmer-wasi`,
* It introduces JavaScript in `wasmer-wasi`?

I didn't review `memfs` and `vfsfs` because I don't know the difference betweem them, and I suspect one of them are going to be removed :-).

It's also too much to review. Can we split this PR in 2 parts:

1. `wasmer-vfs` + VFS in WASI,
2. WASI in JavaScript.

please?

Can't we run all WASI tests with `wasm-bindgen-tests`? That would be terrible.

lib/vfs/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "wasmer-vfs"
+version = "2.0.0"
+description = "Wasmer Virtual FileSystem"
```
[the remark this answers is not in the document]
syrusakbary: I want the VirtualFS implementation to be separated from WASI, and usable standalone.
I think is better without the WASI mention

lib/wasi/src/state/builder.rs
syrusakbary: In this case I believe it will not be a WasiFsError but a FileSystem error. I agree we need some consistency on the error names though

lib/wasi/src/state/mod.rs
syrusakbary: Not at all.. some commands (like cowsay) operate without a filesystem... they just do compute :)

lib/vfs/src/lib.rs
Hywan: See https://github.com/wasmerio/wasmer/pull/2522 to solve this comment.

lib/vfs/src/lib.rs
Hywan: See https://github.com/wasmerio/wasmer/pull/2523 to solve this comment.

lib/vfs/Cargo.toml
Hywan: See https://github.com/wasmerio/wasmer/pull/2523 to solve this comment.

lib/vfs/Cargo.toml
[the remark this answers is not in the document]
Hywan: OK.

lib/cli/Cargo.toml
```diff
@@ -35,11 +35,12 @@ wasmer-engine-universal = { version = "2.0.0", path = "../engine-universal", opt
 wasmer-engine-dylib = { version = "2.0.0", path = "../engine-dylib", optional = true }
 wasmer-engine-staticlib = { version = "2.0.0", path = "../engine-staticlib", optional = true }
 wasmer-vm = { version = "2.0.0", path = "../vm" }
-wasmer-wasi = { version = "2.0.0", path = "../wasi", default-features = false, optional = true }
+wasmer-wasi = { version = "2.0.0", path = "../wasi", default-features = false, features = ["host_fs"], optional = true }
```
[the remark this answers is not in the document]
Hywan: OK, so previously we were using the default features of `wasmer-wasi`. Now, `wasmer-wasi` has `host_fs` as the default feature, so that the previous behavior is kept, which is the correct way to do.

So… we don't need to update `wasmer-c-api` and `wasmer-cli` here. It's been fixed in https://github.com/wasmerio/wasmer/pull/2523.

lib/c-api/Cargo.toml
```diff
@@ -32,7 +32,7 @@ wasmer-engine-universal = { version = "2.0.0", path = "../engine-universal", opt
 wasmer-engine-dylib = { version = "2.0.0", path = "../engine-dylib", optional = true }
 wasmer-engine-staticlib = { version = "2.0.0", path = "../engine-staticlib", optional = true }
 wasmer-middlewares = { version = "2.0.0", path = "../middlewares", optional = true }
-wasmer-wasi = { version = "2.0.0", path = "../wasi", optional = true }
+wasmer-wasi = { version = "2.0.0", path = "../wasi", default-features = false, features = ["host_fs"], optional = true }
```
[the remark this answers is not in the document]
Hywan: See https://github.com/wasmerio/wasmer/pull/2491#discussion_r689562448, which is similiar to this situation.

tl;dr: Fixed in https://github.com/wasmerio/wasmer/pull/2523.

lib/vfs/src/host_fs.rs
Hywan: See https://github.com/wasmerio/wasmer/pull/2524 to fix it. It explains why `not(unix)` is not equivalent to `windows` (tl;dr it's because there is the `wasm` target family, which is new).

lib/vfs/src/host_fs.rs
Hywan: See https://github.com/wasmerio/wasmer/pull/2491#discussion_r689570334. tl;dr it's fixed by https://github.com/wasmerio/wasmer/pull/2524.

lib/vfs/src/host_fs.rs
Hywan: See https://github.com/wasmerio/wasmer/pull/2491#discussion_r689570334. tl;dr it's fixed by https://github.com/wasmerio/wasmer/pull/2524.

lib/vfs/src/host_fs.rs
Hywan: See https://github.com/wasmerio/wasmer/pull/2491#discussion_r689570334. tl;dr it's fixed by https://github.com/wasmerio/wasmer/pull/2524.

lib/vfs/src/host_fs.rs
Hywan: See https://github.com/wasmerio/wasmer/pull/2491#discussion_r689570334. tl;dr it's fixed by https://github.com/wasmerio/wasmer/pull/2524.

lib/cli/src/commands/run.rs
```diff
@@ -64,6 +64,9 @@ pub struct Run {
     #[structopt(long = "debug", short = "d")]
     debug: bool,
 
+    #[structopt(short, long, parse(from_occurrences))]
```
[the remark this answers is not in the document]
Hywan: I've not noticed the `short` option already. Marking this comment as resolved.

Testing to confirm:

```shell
$ ./target/release/wasmer run -h | rg verbose
    -v, --verbose                         
```

The `-v` option is the short version of `--verbose`.

lib/vfs/src/host_fs.rs
Hywan: Because of the orphan Rust rule, it's better to implement `TryInto` rather than `TryFrom`.

Hywan: Solved in https://github.com/wasmerio/wasmer/pull/2525.

lib/vfs/src/host_fs.rs
Hywan: Draft here, https://github.com/wasmerio/wasmer/pull/2528

Hywan:
`wasmer-vfs` with serde enabled doesn't run.

Hywan:
bors try

syrusakbary:
bors try-

syrusakbary:
bors try

Hywan:
bors try

lib/api/src/sys/externals/memory.rs
```diff
@@ -226,6 +226,11 @@ impl Memory {
         unsafe { MemoryView::new(base as _, length as u32) }
     }
 
+    /// A uint8view
+    pub fn uint8view(&self) -> MemoryView<u8> {
+        self.view()
+    }
+
```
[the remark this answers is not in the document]
Hywan: Method is removed.

lib/wasi/src/macros.rs
syrusakbary: I forgot why, we can try to use `get_utf8_str` but I believe I did this because the other didn't work in both js and sys

lib/wasi/src/state/mod.rs
```diff
@@ -1428,16 +1572,18 @@ impl WasiState {
 
     /// Turn the WasiState into bytes
     pub fn freeze(&self) -> Option<Vec<u8>> {
-        bincode::serialize(self).ok()
+        todo!("temporarily disabled")
+        //bincode::serialize(self).ok()
```
Hywan, line 1576: Why? cc @syrusakbary

lib/wasi/src/syscalls/mod.rs
```diff
@@ -2512,10 +2498,15 @@ pub fn poll_oneoff(
...
+//     // RuntimeError::raise(Box::new(WasiError::Exit(code)));
+// }
 pub fn proc_exit(env: &WasiEnv, code: __wasi_exitcode_t) {
-    debug!("wasi::proc_exit, {}", code);
-    RuntimeError::raise(Box::new(WasiError::Exit(code)));
-    unreachable!();
+    // panic!("Exit: {}", code);
+    // Err(WasiError::Exit(code))
+    // RuntimeError::raise(Box::new(WasiError::Exit(code)));
 }
```
Hywan, line 2510: We need to clarify this.

Hywan:
bors try

lib/wasi/src/syscalls/mod.rs
Hywan: Fixed in https://github.com/wasmerio/wasmer/pull/2491/commits/650352ddeca4e130fcfe45312f4a9ccdd1cea841.

lib/wasi/src/state/mod.rs
syrusakbary: We can revert and make bincode an optional dependency if `enable-serde` feature is set

lib/wasi/src/state/mod.rs
```diff
@@ -1428,16 +1572,18 @@ impl WasiState {
...
     }
 
     /// Get a WasiState from bytes
-    pub fn unfreeze(bytes: &[u8]) -> Option<Self> {
-        bincode::deserialize(bytes).ok()
+    pub fn unfreeze(_bytes: &[u8]) -> Option<Self> {
+        todo!("temporarily disabled")
+        //bincode::deserialize(bytes).ok()
```
syrusakbary, line 1582: We can revert and make bincode an optional dependency if enable-serde feature is set

lib/wasi/src/state/builder.rs
syrusakbary: Resolution: make `build` consume `self` (so we don't need to wrap in an `Arc` or other paths)

Hywan: It's not possible because it will break existing code, like:

```
let mut builder = WasiStateBuilder::new();
builder.stdin(…);
builder.set_fs(…);
```

it won't compile anymore because `stdin` will consume `self`.

Except if we introduce a BC break here, it's not possible to solve this problem. I've decided to write what's happening in the documentation when calling `build` or `finalize` multiple times. I think it's the best trade-off for the moment.

See https://github.com/wasmerio/wasmer/pull/2491/commits/4eaa864ae8b08daa3dd97f97c633ad0e6a6b6583.

lib/wasi/src/state/mod.rs
Hywan: Partially solved in https://github.com/wasmerio/wasmer/pull/2491/commits/a12ec28029ea902f13795e5d573f49b82a670556.

`wasmer-wasi` with `enable-serde` doesn't compile for the moment, but it's not new. Let's address that in another PR.

Hywan (approved):

Hywan:
I'm not fully satisfied with the state of this PR, but it's way better than it was initially. Let's fix the other pain points in other PR.

bors r+

Issue #483, opened 2019-06-06
The installer script should register Wasmer as a new binfmt_misc interpreter on Linux


lilianmoraru:
### Motivation

On Linux, `Wasmer` could use `binfmt_misc` to allow users to run WASM binaries like any other binary, without even knowing the `wasmer` name or flags.

### Proposed solution

The installer script should register `Wasmer` as a new `binfmt_misc` interpreter.

### Additional context

QEMU uses this feature to run for example ARM binaries on x86_64 PCs.
The package on Ubuntu that adds these configurations for QEMU is called `qemu-user-binfmt`.
In order to make this work, `binfmt-support` might be required.

jcaesar:
I guess this would need some kind of wrapper (script), as just adding `:wasm:M::\x00asm\x01\x00\x00::/usr/bin/wasmer:` to `/usr/lib/binfmt.d/wasm.conf` only works correctly if you don't pass any arguments.

Hywan:
@jcaesar Can you please open a PR? It seems to be a really good feature to provide!

jcaesar:
Could, but I learned some new stuff and now I kind-of don't want to, the way it is currently. 
* I'd like to support passing arguments like `--dir` (probably from environment variables)
* I'd like for this to work in docker / lxc / ...

The latter point is a bit tricky. You need to set the F and C flag on the binfmt_misc spec, and then make sure that whatever interpreter you specified is a static executable. The question is... do you want to ship an extra executable just for this? (qemu does.) Or do you want to mess up wasmer-cli's command structure so that it will first check if `$1` is a wasm binary and execute it if so?

jcaesar:
(Experimented a bit. binfmt-interpreting wasm files in docker containers is definitely doable: https://github.com/jcaesar/universal-docker-containers#experiment And of course I'm not the first person to have [this idea](https://github.com/bytecodealliance/wasmtime/issues/196#issuecomment-577747819).)

Hywan:
>  do you want to ship an extra executable just for this?

Yeah, nothing against that.

> Or do you want to mess up wasmer-cli's command structure so that it will first check if $1 is a wasm binary and execute it if so?

It depends how complex that is, but I'm not against neither.

jcaesar:
> It depends how complex that is, but I'm not against neither.

[...]

PR #2499, opened 2021-08-01
Register binfmt

jcaesar:
# Description
Specialized command line parsing for when running wasmer/cli as a binfmt_misc "interpreter", and a new subcommand to manage the binfmt_misc registration, as described in #483.

Effect: e.g. this works
```bash
wget https://registry-cdn.wapm.io/contents/liftm/dust-wasi/0.5.4-3/dust.wasm -O dust
chmod a+x dust
./dust
```

What you may want to consider during review:
* I have not seen that trick of copying the binary to `/tmp` to register it from a specially named path anywhere else. It might be a bad idea.
* By default `.` and nothing else is preopened. Am I correct in the assumption that there is no way of giving WASI full filesystem access (because it has no concept of a working directory).
 * Should it also register `.wat`?

# Review


lib/cli/src/commands.rs
```diff
@@ -11,11 +11,15 @@ mod self_update;
 mod validate;
 #[cfg(feature = "wast")]
 mod wast;
+#[cfg(unix)]
```
syrusakbary, line 14: Binfmt can't be installed on macOS, so we should use the linux target directly.

```suggestion
#[cfg(target_os = "linux")]
```

lib/cli/src/commands.rs
```diff
@@ -11,11 +11,15 @@ mod self_update;
...
+mod binfmt;
 
 #[cfg(feature = "compiler")]
 pub use compile::*;
 #[cfg(all(feature = "staticlib", feature = "compiler"))]
 pub use create_exe::*;
 #[cfg(feature = "wast")]
 pub use wast::*;
+#[cfg(unix)]
```
syrusakbary, line 23:
```suggestion
#[cfg(target_os = "linux")]
```

syrusakbary:
Replying inline:

> I have not seen that trick of copying the binary to /tmp to register it from a specially named path anywhere else. It might be a bad idea.

Is there any way we can reuse the installed wasmer binary without copying?

> By default . and nothing else is preopened. Am I correct in the assumption that there is no way of giving WASI full filesystem access (because it has no concept of a working directory).

I think giving access to `.` by default is good (rather than the whole filesystem)


> Should it also register .wat?

Wat is fully a a text-format so it will hard to recognize it properly (if we take on account comments and so, the first bytes might vary).

jcaesar:
> Is there any way we can reuse the installed wasmer binary without copying?

Even when doing some very terrible things to command line parsing, I don't think so. With binfmt_misc, it doesn't seem like you get any extra environment variables, and you can only get the command line to be either of
* `path-to-registered-interpreter` `executable-wasm-file` `original-arg-1` `original-arg-2` ...
* `path-to-registered-interpreter` `executable-wasm-file` `original-arg-0` `original-arg-1` `original-arg-2` ...

But I don't see any way of distinguishing that from a normal wasmer invocation, where you might pass `--dir . --` or similar. Consider: a wasm executable may have entirely the same command line parameter format that wasmer has.

Without changing the behavior of invoking wasmer normally, I can only think of
* Add a second executable that generates a command line that wasmer can understand (but that won't work through lxc / docker / chroots / ...)
* Set up the install script to create a hard link / ref link on install
* Add a second executable that contains another wasmer, exclusively for binfmt_misc

> Wat is fully a a text-format so it will hard to recognize it properly 

It would be recognized by the `.wat` filename extension (and the +x bit). Not sure if that's "hard".
(Of course it is also weird but possible to require files to start with a `;; execute with wasmer` comment and match on that.)

syrusakbary:
> But I don't see any way of distinguishing that from a normal wasmer invocation

I think it should be possible to create a new subcommand: `binfmt-run` that does a specialized version of `wasmer run`.
That way, the binary will not needed to be copied. Other way is just create a separate binary that executes `wasmer` under the hood (basically calling `Command::`. Thoughts?

> It would be recognized by the .wat filename extension (and the +x bit). Not sure if that's "hard".

Cool, let's roll with that for now then! (running executable `.wat`). Somehow I thought binfmt would only work if the header of the binary matched certain bytes.

jcaesar:
> I think it should be possible to create a new subcommand: binfmt-run that does a specialized version of wasmer run.
That way, the binary will not needed to be copied.

I don't see how you would get binfmt_misc to pass you this `binfmt-run` as `$1`, binfmt_misc doesn't support passing extra arguments.

(The format of registering with binfmt_misc is `:name:type:offset:magic:mask:interpreter:flags`, where `interpreter` is a file path and doesn't take arguments. You can't put a `/usr/bin/wasmer binfmt-run` into there (It will look for a file named `wasmer\ binfmt-run`.))

> Other way is just create a separate binary that executes wasmer under the hood

Yes. This is what I meant by:

>> Add a second executable that generates a command line that wasmer can understand 

But then you can't register wasmer on the "host" and run it in a docker container, or vice versa. Personally, I kind of want that feature. For example, try this:
```
docker run --rm --privileged -v /proc/sys/fs/binfmt_misc/:/binfmt_misc/ liftm/wasmer:binfmt-experiment
```
([backing Dockerfile](https://github.com/jcaesar/wasmer/blob/register-binfmt-fromdocker/binfmt.Dockerfile), in case you'd rather build from source.)
Run only that, and you'll have the binfmt setup to run a wasm executable in any container on the machine.
(I'm only using docker because I think it's the currently most commonly used. This works over bwrap, systemd-nspawn, plain lxc, chroots, ...)


I'm curious, is there anything that really speaks against the copying (other than probably being unexplored solution space)?

syrusakbary:
> I'm curious, is there anything that really speaks against the copying (other than probably being unexplored solution space)?

My main concern is that things will get outdated when you run `wasmer self-update` or install wasmer in other way, introducing unintended side effects.

Here's my proposal:
Create another binary file `lib/cli/src/bin/wasmer_binfmt.rs` that simply executes `Command::...` with `wasmer binfmt-run ...` under the hood. Thoughts?

jcaesar:
> My main concern is that things will get outdated when you run wasmer self-update or install wasmer in other way

That is a problem, indeed. The qemu interpreters I've used so far all handled this by being installed through the distro and having their registrations re-done on install/update. I could of course get the self-update command to be aware of binfmts, but there is a caveat about feature interaction and complexity. Hm.

(Note to self: Remove `F` flag to avoid the binary being preopened at registration time.)

jcaesar:
Hm, if the "copy self to /tmp" thing is through, then this entire PR is unnecessary.

All you need is two files

* `/etc/binfmt.d/wasmer.conf`
```
:wasm32:M::\x00asm\x01\x00\x00::/usr/bin/wasmer-binfmt:P
:wasm32-wat:E::wat::/usr/bin/wasmer-binfmt:P
```
* `/usr/bin/wasmer-binfmt`
```bash
#!/usr/bin/env sh
set -eu

wasmer="$(dirname "$0")/wasmer"
if ! test -x "$wasmer"; then
	if ! wasmer="$(which wasmer 2>/dev/null)"; then
		echo Could not locate wasmer binary 1>&2
		exit 137
	fi
fi

dir="${WASMER_BINFMT_MISC_PREOPEN-.}"
wasm="$1"
shift
cm="$1"
shift

exec "$wasmer" run --dir "$dir" --command-name "$cm" "$wasm" -- "$@"
```

but how to install those as part of the install script is something that the wasmer team will know better than I do.

Adding a `wasmer binfmt register` subcommand is unnecessary, and it is a relatively unusual/odd thing to do.

syrusakbary:
> Adding a wasmer binfmt register subcommand is unnecessary, and it would be a relatively unusual/odd thing to do.

What about calling `wasmer binfmt register` just automatically upon Wasmer installation in Linux, which creates the files with the contents you mentioned? I think that might be a bit better than just appending the `wasmer.conf` and other files in the bash install script?

That way we can also have a `wasmer binfmt unregister` that just uninstalls it.

Do you think that's a reasonable approach?

> In case it's useful, the wasmer-install repo lives here: https://github.com/wasmerio/wasmer-install

jcaesar:
> What about calling wasmer binfmt register just automatically upon Wasmer installation in Linux,

That solves the "self-update won't update the binfmt registration" problem with this PR, but the command needs to be re-executed after every boot.
So either install a systemd service, or use systemd-binfmt by placing a file in `/etc/binfmt.d/`.

Systemd service installation is a bit meh..
```bash
if test $OS == "linux" && which systemctl 1>/dev/null 2>/dev/null; then
cat <<EoF >/etc/systemd/system/wasmer-binfmt.service
[Unit]
Description=Set up wasmer to handle execution of wasm binaries
DefaultDependencies=no
Conflicts=shutdown.target
After=proc-sys-fs-binfmt_misc.automount
After=proc-sys-fs-binfmt_misc.mount
Before=sysinit.target shutdown.target
ConditionPathIsReadWrite=/proc/sys/

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=$INSTALL_DIRECTORY/bin/wasmer binfmt reregister
ExecStop=$INSTALL_DIRECTORY/bin/wasmer binfmt unregister
EoF
systemctl daemon-reload
systemctl enable wasmer-binfmt
systemctl restart wasmer-binfmt
fi
``` 

> That way we can also have a wasmer binfmt unregister that just uninstalls it.

Like so?
https://github.com/wasmerio/wasmer/pull/2499/files#diff-daba8471e1ba6c8dab8454a5c1979594cfd27e8903dc8ce4f2ecede97f35e99dR16

jcaesar:
> All you need is two files

Looked into this a bit more: The C flag (use security context of interpreted binary - I think this mostly applies to SUID/SGID) requires the O flag (pre-open the interpreted binary so it can be read even if the mount namespace differs(?)). However, a script starting with `#!` itself needs another interpreter, and O will only allow one interpreted file. So for the wrapper/exec wasmer approach, it would be better to make the wrapper a normal ELF binary. (But I'm still not in favor of that. ;))

lib/cli/src/commands/binfmt.rs
```diff
@@ -0,0 +1,153 @@
...
+            Register | Reregister => {
+                temp_dir = tempfile::tempdir().context("Make temporary directory")?;
+                seccheck(temp_dir.path())?;
+                let bin_path_orig: PathBuf = env::args_os()
+                    .nth(0)
+                    .map(Into::into)
+                    .filter(|p: &PathBuf| p.exists())
+                    .context("Cannot get path to wasmer executable")?;
+                let bin_path = temp_dir.path().join("wasmer-binfmt-interpreter");
+                fs::copy(&bin_path_orig, &bin_path).context("Copy wasmer binary to temp folder")?;
```
ptitSeb, line 71: I don't understand why the specialized binfmt version of wasmer is copied to temp. The binfmt magic will not survive a reboot like that. why not just create a symlink of wasmer to the new name in the same folder?

jcaesar: The binfmt registration is kernel side and does not survive a reboot anyway. (And the file in `/tmp` is deleted once `temp_dir` is dropped. This is allowed because of the `F` flag in `...:PFC`, which also makes sure the interpreter works across mount namespaces / chroots / ...) There needs to be some kind of systemd/openrc/anacron/... service that re-does the registration on every boot. I can add that to the installer script if you want, but it doesn't fit well with the current "install to user's home folder".

ptitSeb: I see.
I guess we can provide hint in doc to do an actual system installation, with binfmt script to be used by systemd binfmt deamon as an example.

ptitSeb (approved):

PR #2516, opened 2021-08-11
iOS Support - via .dylib & Headless Runtime

NathHorrigan:
# Description
This PR makes some fairly simple changes to feature flags so that forks originally intended `*-apple-darwin` now work for both `aarch64-apple-ios`/`x86_64-apple-ios`.

# Review


lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -23,7 +23,6 @@ use wasmer_engine_universal::Universal;
 ///
 /// This is a Wasmer-specific type with Wasmer-specific functions for
 /// manipulating it.
-#[cfg(feature = "compiler")]
```
syrusakbary, line 26: Undelete this :)

lib/api/Cargo.toml
```diff
@@ -28,6 +28,7 @@ thiserror = "1.0"
 more-asserts = "0.2"
 # - Optional shared dependencies.
 wat = { version = "1.0", optional = true }
+wasmparser = { version = "0.78", default-features = false, optional = true }
```
syrusakbary, line 31: This shouldn't be needed

lib/c-api/Cargo.toml
```diff
@@ -23,6 +23,8 @@ crate-type = ["cdylib", "rlib", "staticlib"]
 # We rename `wasmer` to `wasmer-api` to avoid the conflict with this
 # library name (see `[lib]`).
 wasmer-api = { version = "2.0.0", path = "../api", default-features = false, package = "wasmer" }
+wasmparser = { version = "0.78", default-features = false, optional = false }
+
```
syrusakbary, line 27: We can revert this, it shouldn't be needed

lib/c-api/src/wasm_c_api/engine.rs
NathHorrigan: Which is used by this method:
```
pub extern "C" fn wasmer_is_compiler_available(compiler: wasmer_compiler_t) -> bool {
    match compiler {
        wasmer_compiler_t::CRANELIFT if cfg!(feature = "cranelift") => true,
        wasmer_compiler_t::LLVM if cfg!(feature = "llvm") => true,
        wasmer_compiler_t::SINGLEPASS if cfg!(feature = "singlepass") => true,
        _ => false,
    }
}
```

lib/api/Cargo.toml
NathHorrigan: Removed.

lib/c-api/Cargo.toml
NathHorrigan: Removed.

syrusakbary (approved):

syrusakbary:
bors try

NathHorrigan:
It's failing with:
```
error[E0432]: unresolved imports `wasmparser`, `Operator`
168
   --> lib/c-api/src/wasm_c_api/unstable/parser/operator.rs:1:5
169
    |
170
1   | use wasmparser::Operator;
171
    |     ^^^^^^^^^^ help: a similar path exists: `wasmer_api::wasmparser`
172
...
173
523 |         use Operator as O;
174
    |             ^^^^^^^^^^^^^
```

NathHorrigan:
bors try

Makefile
```diff
@@ -483,6 +483,10 @@ build-capi-headless-all: capi-setup
 	RUSTFLAGS="${RUSTFLAGS}" cargo build --manifest-path lib/c-api/Cargo.toml --release \
 		--no-default-features --features universal,dylib,staticlib,wasi
 
+build-capi-headless-ios: capi-setup
```
Hywan, line 486:
```suggestion
build-capi-headless-dylib-ios: capi-setup
```

Since it uses the same features than `build-capi-headless-dylib`, I think it's important to get `dylib` in the target/recipe name. Thoughts?

examples/engine_headless_ios.rs
```diff
@@ -0,0 +1,117 @@
...
+//! workflow, but keep in mind the compiler isn't required after the
+//! compilation step.
+//!
+//! You can run the example directly by executing in Wasmer root:
+//!
+//! ```shell
+//! cargo run --example engine-headless --release --features "cranelift"
+//! ```
+//!
+//! Ready?
```
Hywan, line 46: Please, can you update the comment to explain why it is different for iOS?

examples/engine_headless_ios.rs
```diff
@@ -0,0 +1,117 @@
...
+    local.get $y
+    i32.add)
+  (export "sum" (func $sum_f)))
+"#
+            .as_bytes(),
+        )?;
+
+        let compiler_config = Cranelift::default();
+        let triple = Triple::from_str("aarch64-apple-ios")
+            .map_err(|error| RuntimeError::new(error.to_string()))?;
```
Hywan, line 83: So this is cross-compilation + headless.

Does it need an example for that? I'm OK but we should rename it to `examples/platform_ios_headless.rs` simply like this I guess. What do you think?

examples/engine_headless_ios.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        // Let's define a CPU feature.
+        let mut cpu_feature = CpuFeature::set();
+        cpu_feature.insert(CpuFeature::from_str("sse2")?);
```
Hywan, line 87: `sse2` is for `x86_64`. I see that you're compiling for `aarch64-apple-ios`, so `sse2` won't be supported. You can simply remove that line.

```suggestion
```

examples/engine_headless_ios.rs
```diff
@@ -0,0 +1,117 @@
...
+        let mut dylib_file = Path::new("./sum.dylib");
+        module.serialize_to_file(dylib_file)?;
+        dylib_file
+    };
+
+    Ok(())
+}
+
+#[test]
+#[cfg(not(any(windows, target_arch = "aarch64", target_env = "musl")))]
```
Hywan, line 114: We want it to run on `aarch64`. You must update this please :-).

lib/c-api/src/wasm_c_api/engine.rs
Hywan: Please, keep `#[cfg(feature = "compiler")]` here. But add:

```
use super::super::engine:{wasm_config_t, wasmer_engine_t};
#[cfg(feature = "compiler")]
use super::super::engine::wasmer_compiler_t;
```

in `lib/c-api/src/wasm_c_api/unstable/engine.rs`. It's a bug here.

Thank you very much!

lib/c-api/src/wasm_c_api/unstable/parser/operator.rs
```diff
@@ -1,4 +1,4 @@
-use wasmer_api::wasmparser::Operator;
+use wasmparser::Operator;
```
Hywan, line 1: Why this change? `wasmparser` is not a dependency of `wasmer-c-api`.

lib/engine-dylib/src/artifact.rs
```diff
@@ -291,12 +291,32 @@ impl DylibArtifact {
...
+        // Get the location of the 'ld' linker for clang
+        let fuse_linker = {
+            let ld_install = which::which("ld");
+            if ld_install.is_ok() {
+                ld_install.unwrap().into_os_string().into_string().unwrap()
+            } else {
+                "ld".to_string()
+            }
+        };
+
```
Hywan, line 312: Please configure and add this flag if and only if the `target_triple` is for `aarch64-apple-ios` :-).

lib/engine-dylib/src/artifact.rs
```diff
@@ -773,4 +793,29 @@ impl Artifact for DylibArtifact {
...
+            Command::new("install_name_tool")
+                .arg("-id")
+                .arg(&filename)
+                .arg(&filename)
+                .current_dir(&absolute_path);
+        }
+
+        Ok(())
+    }
```
Hywan, line 820: Please can you explain to me why this new function is required :-)?

Hywan (changes requested):
Thank you very much for your work! I believe we are on a good track. I hope my review provides relevant inputs :-).

lib/engine-dylib/src/artifact.rs
NathHorrigan: Yes! More specifics can be found on the #wasmer channel on Slack but it was caused by the serialised file referencing the original tempfile created by `DylibArtifact`:

```
nathanhorrigan@Nathans-MBP ~/C/w/examples (master)> otool -L sum.dylib
sum.dylib:
	/var/folders/5_/73q9bnl90d106ddc7smgrq7h0000gn/T/wasmer_dylib_dB8mZ6.dylib (compatibility version 0.0.0, current version 0.0.0)
```

The above renames the identifier in the `.dylib` file to match the new file instead of the old. I am still however having issues getting Xcode to load the `.dylib`...

NathHorrigan: Current Xcode error:
```
dyld: Library not loaded: sum.dylib
  Referenced from: /private/var/containers/Bundle/Application/F45CCF73-A561-4E89-9B58-1AD0693A9A18/Wasm3Example.app/Wasm3Example
  Reason: image not found
dyld: launch, loading dependent libraries
DYLD_LIBRARY_PATH=/usr/lib/system/introspection
DYLD_INSERT_LIBRARIES=/Developer/usr/lib/libBacktraceRecording.dylib:/Developer/usr/lib/libMainThreadChecker.dylib:/Developer/Library/PrivateFrameworks/DTDDISupport.framework/libViewDebuggerSupport.dylib
```

Makefile
NathHorrigan: Yes that's a good idea!

examples/engine_headless_ios.rs
NathHorrigan: It was @syrusakbary suggestion but I think it's quite useful as it gives a place to direct people who query compiling wasm to `.dylib` files for iOS. The current `engine_headless` example assumes that it can deserialize the file again which is a false assumption unless the developer has an M1 Mac (And even then I'm not sure if it works, I don't have one).

I'll rename the example.

NathHorrigan:
@Hywan - Do you know of any examples of loading a file in C++ and passing to `wasm_module_deserialize`. Is there a certain format that it is expecting?

Hywan:
> @Hywan - Do you know of any examples of loading a file in C++ and passing to `wasm_module_deserialize`. Is there a certain format that it is expecting?

You should need to load the file (in a `uint8_t[]`) and pass it to `wasm_module_deserialize`. You have an example here https://docs.rs/wasmer-c-api/2.0.0/wasmer_c_api/wasm_c_api/module/fn.wasm_module_deserialize.html, or you can check our C API used in Go for example, in https://github.com/wasmerio/wasmer-go/blob/4d063a16fde35e8cd780db2897d20ee1a94be14c/wasmer/module.go#L27-L33.

NathHorrigan:
Thanks @Hywan 

I'm currently getting this error: `Incompatible pointer types assigning to 'wasm_byte_t *' (aka 'char *') from 'uint8_t **' (aka 'unsigned char **')` from this source:

```
std::string wasm_path = get_resources_dir() + "/sum.dylib";
std::vector<uint8_t> dylib = read_vector_from_disk(wasm_path.c_str());
uint8_t* wasm_bytes = dylib.data();
  
wasm_engine_t* engine = wasm_engine_new();
wasm_byte_vec_t imported_bytes;
imported_bytes.size = dylib.size();
imported_bytes.data = &wasm_bytes;
```

CleanShot 2021-08-12 at 13 24 05@2x

NathHorrigan:
I've tried loading the file as `vector<char>` and that will compile but I get a `NULL` rest from `wasm_module_deserialize`.

Any Suggestions?

Hywan:
Try with:

```cpp
imported_bytes.data = (wasm_byte_t *) wasm_bytes;
```

Currently, you're assigning a reference to `wasm_bytes`, but it's already a pointer, so you're giving a `uint8_t **` rather than a `uint8_t *`.

lib/engine-dylib/src/artifact.rs
Hywan: Please explain this inside the code :-).

NathHorrigan:
Thanks @Hywan - That worked but I'm still getting a `NULL` response from `wasm_module_deserialize`.

Not sure if that's a file format issue or something else entirely, I may have to recompile the capi and investigate...


CleanShot 2021-08-12 at 15 21 46@2x
CleanShot 2021-08-12 at 15 22 00@2x

NathHorrigan:
I can confirm I now having the serialize the .dylib on iOS. I had to update the magic number detection 🤦

NathHorrigan:
Thanks so much for your help @Hywan

I'll start cleaning up this branch now.

examples/engine_headless_ios.rs
NathHorrigan: Removed

examples/engine_headless_ios.rs
NathHorrigan: Removed

lib/c-api/src/wasm_c_api/engine.rs
NathHorrigan: I added those compiler conditions, I had to add some others to fix more build errors though?

lib/c-api/src/wasm_c_api/unstable/parser/operator.rs
NathHorrigan: I've reverted this but I did have to make change in `lib/api/src/sys/mod.rs` so `wasmer_api::wasmparser` is available when the `compiler` feature isn't enabled.

lib/engine-dylib/src/artifact.rs
NathHorrigan: Is this better @Hywan?? We need to be able to find the linker for `x86_64-apple-ios` && `aarch64-apple-ios`.

lib/engine-dylib/src/artifact.rs
NathHorrigan: I've added a description.

examples/engine_headless_ios.rs
NathHorrigan: I've updated this description.

NathHorrigan:
This is ready for another set of review, @syrusakbary @Hywan 

Thanks Guys!

syrusakbary:
bors try

syrusakbary:
The build is failing because of `test_engine_headless` temporary files for the serialized artifacts doesn't have an extension.

The failing line is:
https://github.com/wasmerio/wasmer/blob/a36259ba89d83beb21ddfc5558ae03563009a80e/lib/engine-dylib/src/artifact.rs#L815

lib/engine-dylib/src/artifact.rs
```diff
@@ -773,4 +794,41 @@ impl Artifact for DylibArtifact {
...
+                .into_os_string()
+                .into_string()
+                .unwrap();
+
+            Command::new("install_name_tool")
+                .arg("-id")
+                .arg(format!("@executable_path/{}", &filename))
+                .arg(&filename)
+                .current_dir(&absolute_path)
+                .output()?;
```
syrusakbary, line 829: This only works when serializing to a file, but I assume if you change the location of that dylib file, it will stop working?
If the answer is no, then we should probably run this command after the linker has done their job creating the `dylib` (the `Artifact::new` function).

examples/platform_ios_headless.rs
```diff
@@ -0,0 +1,77 @@
...
+use std::path::Path;
+use std::str::FromStr;
+use tempfile::NamedTempFile;
+use wasmer::imports;
+use wasmer::wat2wasm;
+use wasmer::Instance;
+use wasmer::Module;
+use wasmer::RuntimeError;
+use wasmer::Store;
+use wasmer::Value;
```
syrusakbary, line 25: Can we do all this imports in the same line?

lib/api/src/sys/mod.rs
```diff
@@ -47,13 +47,13 @@ pub use crate::sys::types::{
 pub use crate::sys::types::{Val as Value, ValType as Type};
 pub use crate::sys::utils::is_wasm;
 pub use target_lexicon::{Architecture, CallingConvention, OperatingSystem, Triple, HOST};
-#[cfg(feature = "compiler")]
```
syrusakbary, line 50: We still want this line :)

Hywan:
Please, can you explain how we can test this also?

NathHorrigan:
@Hywan  - I can write a test for this approach, yes. Would a test that compiles a .dylib and verifies the output by grepping `file` and `otool` commands be sufficient?

Hywan:
I would like a full scenario where I can finally test on iOS, even an emulator. I want to be double check your PR, and maybe implement it as part of our CI.

NathHorrigan:
I'll have a think about it. If I had to add an Xcode project to the produced dylib, where would be the best place?

syrusakbary:
> If I had to add an Xcode project to test the produced dylib, where would be the best place?

I think we may need to create a new directory in the integration tests:
https://github.com/wasmerio/wasmer/tree/master/tests/integration

`/tests/integration/ios/`

Of course, we will only test this on the CI when running the tests on macOS.

lib/engine-dylib/src/artifact.rs
NathHorrigan: This command will stop working if you rename the file, Because we prefix it with  `@executable_path/{file_name}`, It doesn't matter where it goes because it is just referencing itself of the same name inside the compiler source.

Unfortunately  this wouldn't worked inside `Artifact::new` because it has that tempfile name that will inevitably change.

lib/api/src/sys/mod.rs
NathHorrigan: Reverting this breaks the `wasmer_api::wasmparser` import that you didn't want me to change previously?

NathHorrigan: `wasmer/lib/c-api/src/wasm_c_api/unstable/parser/operator.rs` relies on `wasmer_api::wasmparser` but when building for headless the compiler flag isn't set and so the above mod file doesn't expose `wasmparser` which means this file won't compile. This issue is present in the current `main` branch, this fixes the existing issue.

It's either this or we add `wasmparser` as a dependency to `wasm_c_api`

examples/platform_ios_headless.rs
NathHorrigan: Updated.

lib/api/src/sys/mod.rs
NathHorrigan: @syrusakbary - What do you think of the above?

NathHorrigan:
@Hywan, @syrusakbary 

I've added an iOS project that when run will:
- Build a wasmer binary for source for target Mac to compile a wasm file from the XCode project to a .dylib
- Build a headless wasmer library and link it to the project
- Link the new .dylib
- Run the .dylib example (the normal sum function)

I've also written a basic swift test that checks everything is loaded correctly and added a test to cargo that can run the iOS project on the iOS simulator within the current wasmer test suite.

lib/engine-dylib/src/artifact.rs
```diff
@@ -812,7 +812,8 @@ impl Artifact for DylibArtifact {
 
         You need to be running MacOS for the following to actually work though.
         */
-        if path.extension().unwrap() == "dylib" {
+        let has_extension = path.extension().is_some();
+        if has_extension && path.extension().unwrap() == "dylib" {
```
syrusakbary, line 816:
```suggestion
        if path.extension().unwrap_or_default() == "dylib" {
```

examples/platform_ios_headless.rs
```diff
@@ -0,0 +1,72 @@
...
+local.get $y
+i32.add)
+(export "sum" (func $sum_f)))
+"#
+        .as_bytes(),
+    )?;
+
+    // Create a compiler for iOS
+    let compiler_config = Cranelift::default();
+    let triple = Triple::from_str("x86_64-apple-ios")
```
syrusakbary, line 41: Shouldn't this be `aarch64-apple-ios` ? As far as I understand iOS runs only on the Apple silicon (ARM) ?

lib/engine-dylib/src/artifact.rs
NathHorrigan: This complained because it returns an OSString and not str. This seamed liked the cleanest solution?

examples/platform_ios_headless.rs
NathHorrigan: Sorry yes. Accidentally changed when testing some things. `x86_64-apple-ios` is a thing though as it is the compile target for iOS simulator on Intel Macs.

syrusakbary:
The PR is shaping good. I tried to run the integration tests locally but I couldn't (just because I'm not super used to the iOS ecosystem).

Could we have the integration tests plugged in into our Github CI? I can help, but since I can't get the tests to run locally I'm not sure how to make them work.

In general we can do it like:
1. Create a `test-integration-ios` command in the Makefile (so users can do `make test-integration-ios` locally) - it should depend on `build-capi-headless-ios` so the c-api it depends on is built
2. Generate a new `sum.dylib` file
3. Run the ios integration tests: `cargo test -p wasmer-integration-tests-ios`

examples/platform_ios_headless.rs
syrusakbary: I've just fixed it

lib/api/src/sys/mod.rs
syrusakbary: I've just fixed it

NathHorrigan:
Thanks @syrusakbary for the commits!

I've added the iOS integration tests to the Github actions file. The iOS project has build scripts embedded in it which will run `make build-capi-headless-ios` by default so it was fairly easy to add. If you could give `bors try` a go?

To fix your local issue could you open the project in the Xcode GUI with:
`open tests/integration/ios/DylibExample/DylibExample.xcodeproj/`

I suspect that your issue is just caused by an older version of Xcode that doesn't have the latest iOS and iPhone simulators.

syrusakbary:
Oh, I can’t see the updates to the GitHub CI. Did you pushed them?

NathHorrigan:
One sec, Github is refusing my push

NathHorrigan:
That should be ready for you now @syrusakbary !

syrusakbary:
bors try

NathHorrigan:
I also updated the Xcode build scripts so you shouldn't have any issues on your M1 machine :)

.github/workflows/main.yaml
```diff
@@ -254,6 +254,7 @@ jobs:
         run: |
           export WASMER_DIR=`pwd`/package
           make test-integration
+          make test-integration-ios
```
syrusakbary, line 257: This probably need to be just when we are in the macOS environment

NathHorrigan: The tests have a flag on them to prevent them running on Linux. Does that work?

syrusakbary: Yeah, I guess that works

syrusakbary:
bors try

syrusakbary:
bors try-

syrusakbary:
bors try

syrusakbary:
bors try-

syrusakbary:
bors try

syrusakbary:
Bors is weirdly failing:
```
{{:badmatch,   {:error, :merge_branch, 403,    "{\"message\":\"Resource not accessible by integration\",\"documentation_url\":\"https://docs.github.com/rest/reference/repos#merge-a-branch\"}",    "8D04:1534:5623A3:D8B0BD:611ED99C"}},  [    {BorsNG.GitHub, :merge_branch!, 2,     [file: 'lib/github/github.ex', line: 185]},    {BorsNG.Worker.Attemptor, :start_attempt, 2,     [file: 'lib/worker/attemptor.ex', line: 204]},    {BorsNG.Worker.Attemptor, :handle_cast, 2,     [file: 'lib/worker/attemptor.ex', line: 61]},    {:gen_server, :try_dispatch, 4,     [file: 'gen_server.erl', line: 695]},    {:gen_server, :handle_msg, 6,     [file: 'gen_server.erl', line: 771]},    {:proc_lib, :init_p_do_apply, 3,     [file: 'proc_lib.erl', line: 226]}  ]}
--
```

NathHorrigan:
That's very strange, according to Google that is an error caused when the userr calling Bors doesn't have permission. Do you wanna try again?

NathHorrigan:
@syrusakbary  ^^

syrusakbary:
bors try

syrusakbary:
bors try-

syrusakbary:
bors try

syrusakbary:
I think is because github actions file got changed. I'll close this PR and reopen as a new one from Wasmer (also I'll add you as contributor so you can directly push to the `feature/ios-headless-support` branch in this repo (without needing a fork)

PR #2517, opened 2021-08-11
WIP of signalless Singlepass

olonho:


syrusakbary:
bors try

lib/vm/src/vmoffsets.rs
```diff
@@ -453,13 +457,24 @@ impl VMOffsets {
...
     /// [`VMContext`]: crate::vmcontext::VMContext
     pub fn size_of_vmctx(&self) -> u32 {
-        self.vmctx_builtin_functions_begin()
+        self.vmctx_trap_handler_begin()
             .checked_add(
-                VMBuiltinFunctionIndex::builtin_functions_total_number()
+                self.num_trap_handlers
                     .checked_mul(u32::from(self.pointer_size))
                     .unwrap(),
             )
```
syrusakbary, line 480: This seems wrong

syrusakbary:
I will delay the code-review once spectests pass, to make sure this approach can work properly.

I'm confident that if tests pass that we will accept the PR.
Once tests pass with bors, I'll do a deeper review on the code structure and changes.

olonho:
bors try

olonho:
Unfortunately `bors try` doesn't work for me, but ` make test-singlepass` passes now.

lib/vm/src/vmoffsets.rs
olonho: Why? We just add one more field to vmctx and trap handler is the last field now.

syrusakbary:
bors try

olonho:
bors try

olonho:
Compilation issue on Windows is unrelated, but fixed anyway.

lib/compiler-singlepass/src/codegen_x64.rs
```diff
@@ -325,6 +327,28 @@ impl<'a> FuncGen<'a> {
         self.mark_instruction_address_end(offset);
     }
 
+    fn emit_trap(&mut self, code: TrapCode) {
```
olonho, line 330: Maybe a bit fancier would be if the trap PC would reflect the actual place in JITed code, not catch-all labels. Fix is pretty simple, we just need to init RDI to trap location before jump, but not sure if it's too important.

olonho: Alternative option is to inline trap_handler to the call site, not jump, but that would mean larger compiled binary, so not desirable.

syrusakbary:
bors try

lib/vm/src/trap/traphandlers.rs
```diff
@@ -385,12 +387,14 @@ static mut IS_WASM_PC: fn(usize) -> bool = |_| false;
...
+    unsafe {
         IS_WASM_PC = is_wasm_pc;
-        platform_init();
-    });
+        if !HANDLERS_INSTALLED && install_handlers {
+            platform_init();
+            HANDLERS_INSTALLED = true;
+        }
+    }
 }
```
syrusakbary, line 398: Rather than having this function receive an argument about installing handlers or not, it should be simply not be called. It should have the same arguments as before.

If the name is confusing, we can rename it to `init_trap_handlers`

lib/vm/src/trap/traphandlers.rs
```diff
@@ -1007,3 +1012,28 @@ pub fn lazy_per_thread_init() -> Result<(), Trap> {
         }
     }
 }
+
+extern "C" fn signal_less_trap_handler(pc: *const u8, trap: TrapCode) {
+    let jmp_buf = tls::with(|info| {
```
syrusakbary, line 1017: Why do we need tls here?

lib/vm/src/vmoffsets.rs
```diff
@@ -53,6 +53,8 @@ pub struct VMOffsets {
     pub num_local_memories: u32,
     /// The number of defined globals in the module.
     pub num_local_globals: u32,
+    /// Number of trap handlers in the module, 0 or 1 depending if signal-less mode is used.
+    pub num_trap_handlers: u32,
```
syrusakbary, line 57: This should be a boolean. I can't imaging why there will be more than one.

Something like `has_trap_handlers` seem like a much better option.

lib/vm/src/trap/traphandlers.rs
syrusakbary: BTW, we already have the `wasmer_raise_trap` method.

Why not calling this one?
https://github.com/wasmerio/wasmer/blob/master/lib/vm/src/libcalls.rs#L670-L673

lib/vm/src/trap/traphandlers.rs
olonho: Done.

lib/vm/src/trap/traphandlers.rs
olonho: Implemented this way to ensure we pass proper program counter. Can change, if you believe it makes sense.

lib/vm/src/vmoffsets.rs
olonho: Indeed, fixed.

lib/vm/src/trap/traphandlers.rs
syrusakbary: Can you expand on why do we need tls here @olonho?

olonho: This is the location where jump buffer is stored and where we read trap info from. Not sure if properly understand the question. `raise_user_trap` for example is written in the same way.

olonho:
@syrusakbary so anything else left for merge of this PR? All comments were addressed.

matklad:
friendly ping @Hywan @syrusakbary ^

PR #2523, opened 2021-08-16
feat(vfs) Make `host_fs` the default, and exclusive to `mem_fs`

Hywan:
# Description

This patch makes `host_fs` part of the default features. It also makes `host_fs` mutually eclusive to `mem_fs`. Finally, it generates a final compiler error if none of them have been enabled.

Finally (promise), it fixes the `Cargo.toml` file of `wasmer-cli` and `wasmer-c-api` because the changes weren't required. See https://github.com/wasmerio/wasmer/pull/2523/commits/5b53c5497a57483450a85fd3e116e5658d1611db for more explanations.

Address a feedback in https://github.com/wasmerio/wasmer/pull/2491.


lib/vfs/src/lib.rs
```diff
@@ -6,6 +6,12 @@ use std::path::{Path, PathBuf};
 use thiserror::Error;
 use tracing::debug;
 
+#[cfg(all(not(feature = "host_fs"), not(feature = "mem_fs")))]
+compile_error!("At least the `host_fs` or the `mem_fs` feature must be enabled. Please, pick one.");
+
+#[cfg(all(feature = "host_fs", feature = "mem_fs"))]
+compile_error!("The `host_fs` and `mem_fs` features are mutually exclusive.");
```
syrusakbary, line 13: I don't think we need this.
Also, people can use both mem_fs and host_fs together in the same implementation (think of mounting two together on a unionfs)... so we need to delete this.

Hywan: Agreed. Resolved!

PR #2524, opened 2021-08-16
fix(vfs) Update `#[cfg(not(unix))]` to `#[cfg(windows)]`.

Hywan:
# Description

The `cfg(unix)` or `cfg(windows)` syntax are shortcuts to
`cfg(target_family = "unix")` and `cfg(target_family = "windows")`.

If we check the list of target families,
https://doc.rust-lang.org/reference/conditional-compilation.html#target_family,
we have:

* `unix`,
* `windows`,
* `wasm`.

So by saying `not(unix)`, we may say `windows`, but also `wasm`. In
this case, we specifically want `windows`. So let's change `not(unix)`
to `windows`, which is clearer and more correct.

Address feedbacks in https://github.com/wasmerio/wasmer/pull/2491.

### DIFF
--- a/lib/vfs/src/host_fs.rs
+++ b/lib/vfs/src/host_fs.rs
@@ -66,7 +66,7 @@ fn fs_metadata_to_metadata(metadata: fs::Metadata) -> Result<Metadata, std::io::
                 filetype.is_fifo(),
             )
         }
-        #[cfg(not(unix))]
+        #[cfg(windows)]
         {
             (false, false, false, false)
         }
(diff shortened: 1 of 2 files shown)


syrusakbary:
Ummm... I'm not that sure about this one.

syrusakbary:
Yeah, thinking more about this, the unix path is implemented.

And the rest of paths aren't,  so it's more precise saying `not(unix)` than saying `windows)

syrusakbary:
Closing this PR :)

PR #2530, opened 2021-08-17
test(vfs) Write tests for `wasmer-vfs`

Hywan:
# Description

 feat(vfs) Rewrite the in-memory filesystem.

The initial in-memory filesystem, aka `mem_fs`, suffered from several
design issues: dead-locks, data duplications, non-optimal look
up/insertions/deletions algorithms, many unimplemented API etc. This
commit is the first step for the rewrite of `mem_fs` to adopt a new
design.

The new design is closer to `ext4` philosophy. A filesystem is a slab
(think an optimised memory allocations vector,
https://en.wikipedia.org/wiki/Slab_allocation) where indexes represent
inodes, and values are of kind `Node`.

A `Node` is now no longer directly recursive (it's indirectly recursive):

```rust
enum Node {
    File { … },
    Directory { children: Vec<Inode>, … },
}
```

To represent children, a node directory doesn't contain `Node` but
`Inode`. To traverse directories, we must jump from the filesystem'
slab, to the node, to the filesystem' slab etc. This design allows
very quick insertions and deletions (it's a O(1) operation, just
insert in the slab and update the parent's children). It also allows
to traverse the filesystem much quicker because we don't need to
traverse all the nodes, only the children of the currently traversed
directory, which reduces the iterations dramatically.

Apart from this new design, the code is adapted largely, with no
`unwrap`, more error handlings, no more `unimplemented!()`s or
`todo!()`s etc.

Some implementations were also wrong, like:

* Reading from `Stdin` never moved an internal cursor, so the user was
  reading the same data again and again.

* It was possible to read from or to write into a file that doesn't
  have the appropriate permissions (e.g. writing into a file with
  read-only permissions),

* The `append` and `truncate` flags on files weren't supported,

* No path canonicalization,

* and so on.

Address feedbacks in https://github.com/wasmerio/wasmer/pull/2491.


Hywan:
bors try

Hywan:
bors try

Hywan:
bors try

Hywan:
musl is passing. Now let's fix `test-js` (not related to this PR, just that `wasm-pack` has changed something, or `wasm-bindgen-test`).

bors try

Hywan:
bors try

PR #2547, opened 2021-08-31
fix(engine-dylib): do not keep temp file and delete it automatically

### LINKED ISSUE #2501: Temp files are not removed in dylib engine
### Describe the bug

dylib creates tempfile to be able to "dlopen" later, but those files are not removed later. This could be an issue for a long running program which constantly (re)loads wasm files, occupying lots of unnecessary disk spaces.

### Steps to reproduce

Use dylib engine to load a wasm file, then check /tmp for temp files.

### Expected behavior

Temp file is deleted when not being used.

### Actual behavior

Temp file persists even program restarts.

bnjjj:
close #2501

# Description
This PR plan to fix the issue #2501 
I know the implementation using Drop trait is not the best way to implement it. Maybe I need more knowledge about engine lifecycle or adding a trait method to clean any artifacts we create when using an engine. Feel free to give me more details about how I could implement this in a better way.

### DIFF
--- a/lib/engine-dylib/src/artifact.rs
+++ b/lib/engine-dylib/src/artifact.rs
@@ -48,6 +48,7 @@ use wasmer_vm::{
 #[derive(MemoryUsage)]
 pub struct DylibArtifact {
     dylib_path: PathBuf,
+    is_temporary: bool,
     metadata: ModuleMetadata,
     finished_functions: BoxedSlice<LocalFunctionIndex, FunctionBodyPtr>,
     #[loupe(skip)]
@@ -60,7 +61,9 @@ pub struct DylibArtifact {
 
 impl Drop for DylibArtifact {
     fn drop(&mut self) {
-        std::fs::remove_file(&self.dylib_path).expect("cannot delete the temporary artifact");
+        if self.is_temporary {
+            std::fs::remove_file(&self.dylib_path).expect("cannot delete the temporary artifact");
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


lib/engine-dylib/src/artifact.rs
```diff
@@ -58,6 +58,12 @@ pub struct DylibArtifact {
     frame_info_registration: Mutex<Option<GlobalFrameInfoRegistration>>,
 }
 
+impl Drop for DylibArtifact {
+    fn drop(&mut self) {
+        std::fs::remove_file(&self.dylib_path).expect("cannot delete the temporary artifact");
```
syrusakbary, line 63: This will drop the file always, however there is only one scenario where we should do it.

`deserialize` (which creates a temporal file and then calls`deserialize_from_file_unchecked`) should delete the temporal file, however `deserialize_from_file` (which doesn't create a temporal file) should *not* delete the file (since it was provided by the user), we can provably have a new field specifying if the file is temporal or not and delete it on such case.

bnjjj: Ok got it thank you, btw what's your policies regarding to error in this case ? Do you prefer to panic as I do here or just display a log ?

syrusakbary: I think it might be better to gracefully dismiss the error in case the worst case scenario is reached, so logging seems like a better option than a panic.

bnjjj: That's what I think too. Updated, thanks :)

syrusakbary: That's great, will review now!

syrusakbary (approved):
Looks great! Merging soonish (as soon as bors test and merges)

bnjjj:
Sorry I had to add a feature on dependency `tracing` for log in this crate

syrusakbary:
No worries.

bors r+

PR #2548, opened 2021-09-01
Fix probestack

ptitSeb:
# Description
The `skip_stack_gard_page` were failing.

The patch first add a trampoline for the PROBEGUARD function on x86_64 build, as this one comes from some runtime and can be too far away from generated code for a regular 32bits relative address (this fix a segfault when a generated function needs a call to probeguard).
The patch also add a special case for StackOverflow error, and doesn't check if the error comes from Wasm code (as the stack overflow will comes from a runtime library, generated by the probeguard function).

### DIFF
--- a/lib/compiler-cranelift/src/compiler.rs
+++ b/lib/compiler-cranelift/src/compiler.rs
@@ -25,12 +25,18 @@ use wasmer_compiler::CompileError;
 use wasmer_compiler::{CallingConvention, ModuleTranslationState, Target};
 use wasmer_compiler::{
     Compilation, CompileModuleInfo, CompiledFunction, CompiledFunctionFrameInfo,
-    CompiledFunctionUnwindInfo, Compiler, CustomSection, CustomSectionProtection, Dwarf,
+    CompiledFunctionUnwindInfo, Compiler,Dwarf,
     FunctionBinaryReader, FunctionBody, FunctionBodyData, MiddlewareBinaryReader, ModuleMiddleware,
-    ModuleMiddlewareChain, Relocation, RelocationKind, RelocationTarget, SectionBody, SectionIndex,
+    ModuleMiddlewareChain, SectionIndex,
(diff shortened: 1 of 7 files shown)


lib/compiler-cranelift/src/compiler.rs
```diff
@@ -102,6 +109,24 @@ impl Compiler for CraneliftCompiler {
...
+        let mut custom_sections = PrimaryMap::new();
+
+        #[cfg(target_arch = "x86_64")]
+        let probestack_trampoline = CustomSection {
+            protection: CustomSectionProtection::ReadExecute,
+            bytes: SectionBody::new_with_vec(vec![0xff, 0x25, 0x00, 0x00, 0x00, 0x00]),
+            relocations: vec![Relocation {
+                kind: RelocationKind::Abs8,
+                reloc_target: RelocationTarget::LibCall(LibCall::Probestack),
+                offset: 6,
```
syrusakbary, line 121: Can we comment giving context why is `6` the offset?

lib/compiler-cranelift/src/compiler.rs
```diff
@@ -102,6 +109,24 @@ impl Compiler for CraneliftCompiler {
             }
         };
 
+        let mut custom_sections = PrimaryMap::new();
+
+        #[cfg(target_arch = "x86_64")]
+        let probestack_trampoline = CustomSection {
+            protection: CustomSectionProtection::ReadExecute,
+            bytes: SectionBody::new_with_vec(vec![0xff, 0x25, 0x00, 0x00, 0x00, 0x00]),
```
syrusakbary, line 117: Can we comment what this means on assembly? `jmp` ?

syrusakbary (approved):
Other than my comments it looks great

syrusakbary:
Not sure if it's flaky, let's try again.

bors r+

PR #2552, opened 2021-09-03
fix(vm) Fix the memory_grow test on Windows (with a bigger page guard)

ptitSeb:
# Description
Fix the memory_grow test under Windows.

The test was failling randomly, at the step with a dynamic memory block of size 0, when trying a read/write at pagesize offset, so just outside the pagegard in that case.
It seems windows memory manager may map a writable page just after the newly alocated block, so using a "2 pages" guard size on windows to avoid this issue.

### DIFF
--- a/lib/api/src/sys/tunables.rs
+++ b/lib/api/src/sys/tunables.rs
@@ -52,13 +52,19 @@ impl BaseTunables {
 
         // Allocate a small guard to optimize common cases but without
         // wasting too much memory.
+        // The Windows memory manager seems more laxed than the other ones
+        // And a guard of just 1 page may not be enough is some borderline cases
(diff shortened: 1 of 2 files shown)


syrusakbary (approved):

lib/api/src/sys/tunables.rs
```diff
@@ -52,13 +52,19 @@ impl BaseTunables {
 
         // Allocate a small guard to optimize common cases but without
         // wasting too much memory.
+        // The Windows memory manager seems more laxed than the other ones
+        // And a guard of just 1 page may not be enough is some borderline cases
+        // So using 2 pages for guard on this plateform
```
Hywan, line 57:
```diff
- plateform
+ platform
```

:-)

PR #2558, opened 2021-09-06
Added support for LLVM 12

syrusakbary:
# Description

Added support for LLVM 12

# Review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,14 +133,18 @@ ifneq ($(ENABLE_LLVM), 0)
 		LLVM_VERSION := $(shell llvm-config --version)
 
 		# If findstring is not empty, then it have found the value
-		ifneq (, $(findstring 11,$(LLVM_VERSION)))
+		ifneq (, $(findstring 12,$(LLVM_VERSION)))
+			compilers += llvm
+		else ifneq (, $(findstring 11,$(LLVM_VERSION)))
 			compilers += llvm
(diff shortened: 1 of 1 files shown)


ptitSeb:
bors try

ptitSeb:
bors try

ptitSeb:
bors try

Makefile
```diff
@@ -324,6 +329,9 @@ $(info Cargo features:)
 $(info   * Compilers: `$(bold)$(green)${compiler_features}$(reset)`.)
 $(info Rust version: $(bold)$(green)$(shell rustc --version)$(reset).)
 $(info NodeJS version: $(bold)$(green)$(shell node --version)$(reset).)
+ifeq ($(ENABLE_LLVM), 1)
+        $(info LLVM version: $(bold)$(green)$(shell llvm-config --version)$(reset).)
```
syrusakbary, line 333: The command available might be `llvm-config-12` or other one. It might be better to create a `LLVM_VERSION` var and just use it here?

ptitSeb: Well, the first choice is using generic `llvm-config`, and if now found, using version specific ones. In that case, the LLVM_VERSION variable is not set, but that can be fixed easily, to be used later.
I'll make the change so the shown version is more coherent with the actually selected one.

syrusakbary: Sounds good!

ptitSeb (approved):

PR #2569, opened 2021-09-10
refactor(api): add Send + Sync trait when creating an instance

bnjjj:
# Description

I'm currently implementing wasmer inside Krustlet (https://github.com/krustlet/krustlet/pull/677). Then I had to force the thread safety of trait object used to create a wasmer instance. I know it could be a breaking change and maybe I miss some backgrounds about this. So feel free to let me know what do you think and how we could have a better implementation if needed. I'm open to any suggestions.


Amanieu:
LGTM, could you rebase to resolve the conflicts?

bnjjj:
done

PR #2574, opened 2021-09-15
Feature/singlepass windows

ptitSeb:
# Description

Added Windows x86_64 support to SinglePass Compiler.

Compared to System V ABI, the Windows ABI changed the way argument are send to function in the following way:
1. Only the 4 1st arguments can be send in register.
2. The argument will go in XMM reg is Float or Double, or in GPR for anything else
3. Regs are RCX/XMM0, RDX/XMM1, R8/XMM2 and R9/XMM3
4. Other argument go to the stack
5. There is 32bytes also reserved in the stack to shadow the 4 arguments if the callee function needs too

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -172,7 +172,7 @@ ifneq ($(ENABLE_SINGLEPASS), 0)
 	ifeq ($(ENABLE_SINGLEPASS), 1)
 		compilers += singlepass
 	# … otherwise, we try to check whether Singlepass works on this host.
-	else ifneq (, $(filter 1, $(IS_DARWIN) $(IS_LINUX)))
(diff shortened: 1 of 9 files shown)


syrusakbary (commented):
Right now we are using compilation flags `#[cfg(target_os = "windows")]` to determine the output.

To allow cross-compilation it would be great to have it depending on a runtime flag (we can add the `compiler::Target` to the `Singlepass` struct, and check the target OS there)

ptitSeb:
bors try

ptitSeb:
bors try

ptitSeb:
bors try

ptitSeb:
bors try

ptitSeb:
bors try

PR #2583, opened 2021-09-27
Improved wasmer-js api

syrusakbary:
Improved Wasmer-JS API

* Create an `init_envs` function in the instance (to allow creation of the Wasm Instance in JS, and setting up on Rust/Wasmer)
* Improved serde dependency on subcrates


syrusakbary:
bors try

Issue #2584, opened 2021-09-27
ca5dec1 breaks MinGW build


lucianoiam:
Recent commits do not build on MinGW

Compilation fails with missing symbols, see output below

```
rustc 1.53.0 (53cb7b09b 2021-06-17)
MINGW64_NT-6.1-7601 vm-PC 3.2.0-340.x86_64 2021-04-02 09:40 UTC x86_64 Msys
```

### Steps to reproduce

```
 RUSTFLAGS="-C target-feature=+crt-static" cargo build       --manifest-path lib/c-api/Cargo.toml --no-default-features       --features wat,universal,dylib,staticlib,wasi,middlewares       --features cranelift
```

### Expected behavior

Example output when building from source of commit 45f56d0 ( immediately previous to ca5dec1 )

```
    Finished dev [unoptimized + debuginfo] target(s) in 1m 21s
```

### Actual behavior

```
  = note: C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/10.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: \\vmware-host\Shared Folders\src_win\wasmer\target\debug\deps\libwasmer_vm-d4cfd2554963633f.rlib(handlers.o): in function `wasmer_register_setjmp':
          Z:\src_win\wasmer\lib\vm/src/trap/handlers.c:50: undefined reference to `__intrinsic_setjmp'
          collect2.exe: error: ld returned 1 exit status
```

### Additional context

Is MinGW an officially supported target?  the above report + [this one](https://github.com/wasmerio/wasmer/issues/2483) mean there is no MinGW support unless building from older source.

Thanks.

ptitSeb:
Ok, using a VS intrinsic is probably a bad idea on MinGW. Do you know if I can use `setjmp3` with a NULL 3rd parameter or something else on MinGW?

ptitSeb:
Yeah, I'll do that. I "just" need to setup a MSYS2 config to test this...

lucianoiam:
Maybe points 1) and 4) [here](https://github.com/lucianoiam/hiphop/blob/master/doc/windows.txt) can save some time. It contains the original links if needed.

ptitSeb:
Well, my problem is I cannot have both rustc toolchain for MSVC and MINGW on Windows system.

lucianoiam:
[...]

PR #2585, opened 2021-09-27
fix(compiler)-Use regular setjmp function for MinGW

ptitSeb:
Fix build issue with MinGW-64 toolchain.


ptitSeb:
Will fix #2584

syrusakbary (approved):

PR #2592, opened 2021-10-01
Add ImportObject::get_namespace_exports

Amanieu:
This allows the contents of an existing namespace to be added to by
extracting an `Exports` from it, adding to that `Exports` and then
replacing the existing namespace with the modified `Exports`.


lib/api/src/sys/import_object.rs
```diff
@@ -16,6 +17,12 @@ pub trait LikeNamespace {
     fn get_namespace_export(&self, name: &str) -> Option<Export>;
     /// Gets all exports in the namespace.
     fn get_namespace_exports(&self) -> Vec<(String, Export)>;
+    /// Returns the contents of this namespace as an `Exports`.
+    ///
+    /// This is used by `ImportObject::get_namespace_exports`.
+    fn as_exports(&self) -> Option<Exports> {
+        None
+    }
```
syrusakbary, line 25: Why `Option`? Also, why a default impl?

Amanieu: For backward compatibility with any existing implementations of `LikeNamespace` outside wasmer.

Amanieu: If backwards compatibility was not a concern then I would remove `LikeNamespace` entirely and use `Exports` directly in `ImportObject`.

syrusakbary: Makes sense

syrusakbary (approved):
Not sure if I got the hack, but it looks good to me 👍

ptitSeb:
Oh, still fd_rename_path on windows x64 that fails?

syrusakbary:
Let's try again

bors r+

PR #2646, opened 2021-11-03
Redesign the API for accessing the contents of a Wasm memory

Amanieu:
# Description

- Accessing the contents of a memory is now done through the `read`/`read_uninit`/`write` methods of `Memory`. These functions are thread-safe even when there are concurrent modifications to the memory.
- The `ValueType` trait has been reworked to also support zeroing out padding bytes in a type. This is necessary to avoid leaking uninitialized data into Wasm memory.
- `ValueType` can now be safely derived with `#[derive(ValueType)]`. `unsafe impl` is no longer needed.
- `WasmPtr` now dereferences to `WasmRef` and `WasmSlice` instead of giving raw access to a `Cell`.
- `WasmPtr` is now generic over the memory size, allowing for 32-bit and 64-bit `WasmPtr`s. This is done through a `MemorySize` trait and `Memory32`/`Memory64` types which implement it.
- `WasmRef` and `WasmSlice` types are added. Unlike `WasmPtr` these types contain a reference to a `Memory`.
- `MemoryView` has been removed, it is superseded by `WasmSlice`.

# Review


lib/api/src/js/externals/memory.rs
```diff
@@ -253,46 +231,7 @@ impl Memory {
...
-    pub fn view<T: ValueType>(&self) -> MemoryView<T> {
-        unimplemented!("The view function is not yet implemented in Wasmer Javascript");
-    }
-
-    /// A theoretical alais to `Self::view::<u8>` but it returns a `js::Uint8Array` in this case.
-    ///
-    /// This code is going to be refactored. Use it as your own risks.
-    #[doc(hidden)]
-    pub fn uint8view(&self) -> js_sys::Uint8Array {
+    pub(crate) fn uint8view(&self) -> js_sys::Uint8Array {
```
syrusakbary, line 234: Can we cache the result in an option in the memory?

Doing this operation might be heavy if we do a lot of reads `js_sys::Uint8Array::new(&self.vm_memory.memory.buffer())`

syrusakbary (commented):
The PR implementation looks good to me.

As a side comment, I think it might be worth to have `read` and `write` operations on a `MemoryBuffer` (new abstraction) instead than in a `Memory`.
That way, we will be a bit closer to the js API and we will solve easily the overhead of caching/creating a `uint8view`.

Amanieu:
Could we just create a view when the memory is created and keep that view alive for the whole lifetime of the memory?

syrusakbary:
> Could we just create a view when the memory is created and keep that view alive for the whole lifetime of the memory?

That sounds good. In the case of JS, if you want to access the memory buffer, you can do `memory.buffer` (which is always acccessible).

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Memory/Memory

Funnily enough, in JS you need to cast again the buffer into a `Uint8Array` or similar to really access the data (read/write).
https://developer.mozilla.org/en-US/docs/Web/API/ArrayBufferView


If we could follow a similar API in Rust that would be ideal. Thoughts?

syrusakbary (approved):
I still think we may want to create a buffer/view abstraction on the future to not operate on read/writes directly on memory, but I'm ok with the current implementation.

PR #2669, opened 2021-11-08
Use upstream version of sccache

Amanieu:


Amanieu:
bors try

Cargo.toml
```diff
@@ -159,6 +159,8 @@ coverage = []
 
 [profile.dev]
 split-debuginfo = "unpacked"
+[profile.release]
+debug = true
```
ptitSeb, line 163: Is that intentional?

ptitSeb (approved):

PR #2675, opened 2021-11-09
Added support for overriding the WASI syscall implementations

john-sharratt:
# Description

## Wasmer WASI syscall override

- Added a syscall proxy feature so that anyone using Wasmer can modify            
  the sys calls without forking the wasmer source code.
- Added better mappings of errors to WASI error codes
- Vastly improved the performance of memfs file writing (nasty slow loop)

See here for WASMER in a terminal working with some of the syscalls that were not yet implemented but that I could implement with this proxy:

https://sh.tokera.com/


syrusakbary:
> Added a syscall proxy feature so that anyone using Wasmer can modify
the sys calls without forking the wasmer source code.

I believe this is already possible? Right now if you are using `generate_import_object_snapshot0`(or similar), you can replace the imports with your own implementations. Would that work for you? I think this will simplify the PR by removing the need of a proxy

john-sharratt:
Well not without exposing the native implementations as basically I only override a few of the syscalls you have not yet implemented such as poll, time and yielding. If I use the generate_import_object_snapshot0 method it means I have to reimplement everything including the interface to the file system abstraction which you developed.

Is there another way you know of overriding just one or two of the syscalls?

syrusakbary:
Perhaps the `ChainedNamedResolver` will do it? You will need to create one ImportObject and then chain it to the one Wasmer WASI provides

john-sharratt:
Hmm... ok I'll take a look at it. Thanks
There's a speed up on the VFS VirtualFile write on MemFS also in there b.t.w - it was really slow because of the loop for some reason.

lib/vfs/src/lib.rs
```diff
@@ -22,6 +22,17 @@ pub type Result<T> = std::result::Result<T, FsError>;
...
+{
+    pub fn new(fd: usize) -> FileDescriptor {
+        FileDescriptor(fd)
+    }
+
+    pub fn fd(&self) -> usize {
+        self.0
+    }
+}
+
```
syrusakbary, line 35: FileDescriptor is intentionally opaque for the user. The internal implementations (host, mem), may use or not a numeric fd

lib/wasi/src/syscalls/windows.rs
```diff
@@ -5,42 +5,40 @@ use wasmer::WasmCell;
...
-    __WASI_ESUCCESS
+    Ok(resolution_val)
 }
 
 pub fn platform_clock_time_get(
     clock_id: __wasi_clockid_t,
     precision: __wasi_timestamp_t,
-    time: WasmCell<__wasi_timestamp_t>,
-) -> __wasi_errno_t {
+) -> Result<__wasi_timestamp_t, __wasi_errno_t> {
```
syrusakbary, line 29: This is changing the spec of the WASI `platform_clock_time_get` signature. Did it got updated and we missed it?

john-sharratt: My bad - I should not have changed this file - will put it back

lib/vfs/src/lib.rs
john-sharratt: Yeah - the reason for this is I actually implemented the polling functionality for WASI that was set as unimplemented. Basically the way I did it was to expose it here:

`impl VirtualFile for Fd {
    fn last_accessed(&self) -> u64 {
        0
    }
    fn last_modified(&self) -> u64 {
        0
    }
    fn created_time(&self) -> u64 {
        0
    }
    fn size(&self) -> u64 {
        0
    }
    fn set_len(&mut self, _new_size: wasi_types::__wasi_filesize_t) -> Result<(), WasiFsError> {
        Err(WasiFsError::PermissionDenied)
    }

    fn unlink(&mut self) -> Result<(), WasiFsError> {
        Ok(())
    }

    fn bytes_available(&self) -> Result<usize, WasiFsError> {
        Ok(0)
    }

    fn get_fd(&self) -> Option<FileDescriptor> {
        Some(FileDescriptor::new(self.raw.id))
    }
}`

...which is the tok terminal project I then used the proxy override functionality to put in a poll implementation without breaking the backwards compatibility with the other users of Wasmer..
https://github.com/john-sharratt/ate/blob/master/tokterm/src/wasi.rs#L238

But you are right we should not really expose the internals - I could drop the polling implementation entirely as it was just for compatibility reasons I implemented it but its not really mandatory.

Alternatively if you have another plan to poll happy to help.

lib/wasi/src/syscalls/windows.rs
john-sharratt: So looked into it and what I'd done is to hide more of the details of the WASI syscall from the proxies:

Basically the memory copies to WASM memory are done sooner than the implementations themselves...
https://github.com/john-sharratt/wasmer/blob/master/lib/wasi/src/syscalls/mod.rs#L82

Really I guess the choice is do you allow consumers to proxy or not?

If the answer is no... then I could look into using the ChainedNamedResolver as you suggested.
If the answer is yes... then I could look into reducing the abstraction leakage as much as possible and moving the memory copies into a nice place so that they don't bloat the code.

Not sure there is a right answer and its your hard work so just let me know how I can help.

Cheers

syrusakbary: > If the answer is yes... then I could look into reducing the abstraction leakage as much as possible and moving the memory copies into a nice place so that they don't bloat the code.

That sounds like a great tradeoff, if we can avoid the proxy abstraction and use ChainedResolver instead I think that will simplify things

> I've done and polling functionality thats not yet implemented on your mainline but needed for more advanced programs like wapm itself

I'd love to have that upstream as well (even the experimental wasi-net)... I really love the concept of having a self-hosted wapm published in wapm 😊

john-sharratt:
I will pull out this merge request and do a new one that merges the functionality rather than attempts to proxy - Will work on this today. Cheers

PR #2677, opened 2021-11-10
Support 32-bit memories with 65536 pages

### LINKED ISSUE #2187: can't represent 65536 pages

Amanieu:
Fixes #2187

### DIFF
--- a/lib/api/src/sys/externals/memory.rs
+++ b/lib/api/src/sys/externals/memory.rs
@@ -131,7 +131,7 @@ impl Memory {
     pub fn data_size(&self) -> u64 {
         let definition = self.vm_memory.from.vmmemory();
         let def = unsafe { definition.as_ref() };
-        def.current_length.into()
+        def.current_length.try_into().unwrap()
     }
 
     /// Returns the size (in [`Pages`]) of the `Memory`.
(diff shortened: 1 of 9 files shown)


ptitSeb (approved):

ptitSeb:
Windows test not passing because the Swapfile on the test system is not big enough. Should the test be disabled on Windows or can the CI modified?

Amanieu:
Let's just disable it on Windows.

PR #2681, opened 2021-11-15
fix singlepass slow compilation of many function wasm by using dynasm VecAssembler

ailisp:
# Description

We found in certain benchmark (a wasm with 150_000 no-op functions), compilation is slowed down by dynasmrt's mmapand mprotect syscalls, however, those mmap isn't used in wasmer and can be replaced by simpler VecAssembler. In this [bench](https://github.com/near/wasmer/pull/71) it's a 10x speedup in singlepass backend compilation time.

# Review

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,6 +53,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.12.1"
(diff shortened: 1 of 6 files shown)


syrusakbary (approved):

ptitSeb:
@ailisp can you do a `cargo fmt --all` and push the change, so the linter is happy and we can merge this?

PR #2689, opened 2021-11-17
Implemented missing features required for tokera.sh

john-sharratt:
Implemented a bunch of features in wasmer that allow a more complete experience on the browser environment.

In particular the following critical functions were missing
- the ability to kill a process using a hook in the syscalls - this is needed for the ctrl-c functionality
- polling was not implemented which is needed for `quickjs`
- error codes were defaulting to EIO which was hiding the WOULDBLOCK error which is needed for async IO. (none blocking sockets) which itself is needed for networking in WASI
- there was a missing permission needed to create symbolic links
- the logging was too verbose and the IO copy function too slow for browser operations
- there was no clock_time_get implemented which is needed for things like timeouts and fixed sleeping periods

The previous proxy implementation is no longer needed


syrusakbary:
I like the PR in general with one exception that I'm not very sure about: the `tick` function.

I think the feature of being able to halt/stop the VM is a feature of the VM, and not the systemcalls.
In browser environments, I think we can use the fact that webworkers can be stopped (and thus, the executing instances).

Thoughts @john-sharratt ?

john-sharratt:
Yeah this is true the web worker stop is a much better approach but it does take away some options for future plans for this capability that will be harder elsewhere.

Basically there is a problem in WASM at the moment with multi-threading and concurrency, nearly everything is synchronous and most WASM modules are multithreaded (except emscripten of course). So how then is it possible to make asynchronous IO operations?

I found a way that works quite nicely, essentially the callbacks are executed back into the WASM from the same thread that goes into an idle state. Effectively this is a bit like how async is implemented in rust and in Java so its actually quite easy to implement without changing too much; only problem is that it needs some form of premption.

I guess a bit of a compromise could be sought here that reduces the tick() sprinkling but still allows for a async based IO and callbacks into single threaded WASM modules (which is needed for decent web socket implementation and RFP that invokes other modules).... something like this.

1) remove all the tick()'s except for those that put the WASM module into an idle state (yield, blocking reads, etc..) - should be only about 5 or 6 syscalls.
2) implement the callback and termination code in these more limited hooks.
3) also implement a forced web-worker termination in the terminal that triggers after a set timeout or by using kill flags (-9 in linux).
4) modify the client and server side logic so that it does not do any lock or event spinning (code needs to go idle to receive callbacks).

john-sharratt:
So removed the infamous tick() function :-)

Also removed the terminate atomic that the process was watching to self terminate - instead I invoke a new method added terminate() which allows the caller to terminate the process from the outside rather than from the inside. It uses the already implemented RuntimeError::raise.

The on_tick callback was replaced with an on_yield callback - I need to explain what this is and what it does.

Basically with WASM there is a bit of a problem in that its essentially single threaded without any concurrency - when you run the process it goes until the function exits. This is ok for alot of use-cases but not for other async things that run in the background such as:
- timers
- web sockets
- async file IO
- sleeping

So what I did was I faked async operations by picking up the IO tasks in the background whenever the WASM module yields execution - effectively it gives an opportunity for callbacks to call back into the process which is basically how wasm-bus is now implemented.

This got me thinking then...

```rust
impl {
    fn work(...) {
        let driver = async move {
            loop {
                let msg = {
                    let mut queue = pool.queue.lock().unwrap();
                    queue.pop_front()
                };
                
                //...
                            task();
                //...

                let _ = rx.recv().await;
            }
        };
        wasm_bindgen_futures::spawn_local(driver);
    }
}
```

This is basically how the WebWorker thread pool launches processes and basically what its doing is running synchronous code within the Javascript async engine, which is ok because a web worker is not the GUI thread so everything runs ok still but its not ideal as it means I have to do all my async IO in another thread and write logic that shifts it back and forth - plus I have to do the on_yield trick to get things feeding back into the WASM execution scope.

...but what if we did something different, what if we can call WASM functions as ASYNC. My thoughts on how to do it are two options:
1) We implement a poll function on top of "async fn main" (bit like the WASI spec states it)
2) We make it so that when the process yields execution that the thread is "parked" just like normal threads in Linux does it.

The second option being a bit harder to implement but in my view its better because...
a) It allows the async engine to be independent of the hosting
b) It works with sync and async code
c) It 'should' work with tokio
d) Its not tied to WASI specs so works with wasm32-unknown-unknown

So while I think its a good idea to try to get whats already built merged upstream (let me know what you want changed) - how hard do you think it would be to implement option 2?

Cheers

syrusakbary (approved):
Some minor nits, but we can merge and solve them async

lib/wasi/src/lib.rs
```diff
@@ -72,26 +74,64 @@ pub enum WasiError {
     UnknownWasiVersion,
 }
 
+pub(crate) const TERMINATE_NOOP: u32 = u32::MAX;
```
syrusakbary, line 77: TODO: Remove

lib/wasi/src/state/mod.rs
```diff
@@ -791,7 +793,7 @@ impl WasiFs {
                                 }
                             } else if file_type.is_symlink() {
                                 should_insert = false;
-                                let link_value = file.read_link().ok().ok_or(__WASI_EIO)?;
+                                let link_value = file.read_link().map_err(|e| map_io_err(e))?;
```
syrusakbary, line 796: TODO:

```suggestion
                                let link_value = file.read_link().map_err(map_io_err)?;
```

lib/wasi/src/state/mod.rs
```diff
@@ -1191,7 +1193,7 @@ impl WasiFs {
                 match &mut inode.kind {
                     Kind::File { handle, .. } => {
                         if let Some(file) = handle {
-                            file.flush().map_err(|_| __WASI_EIO)?
+                            file.flush().map_err(|e| map_io_err(e))?
```
syrusakbary, line 1196: TODO
```suggestion
                            file.flush().map_err(map_io_err)?
```

syrusakbary:
Hey @john-sharratt, some checks were failing so we were unable to merge the PR for our upcoming release (Wasmer 2.1).

However, we can include in the next release once the tests are passing :)

john-sharratt:
Ah I see all these failed tests now and that's quite a of them! Next release definitely sounds like a wise choice to get this all to a higher quality without rushing it in.

Practical question, If I push code to this branch will re-run the tests right?
I don't remember seeing them before and not as familiar with GitHub as I should be.

john-sharratt:
Ah I found it don't worry my bad.
I'll fix this all up and make the changes you suggested.

lib/wasi/src/lib.rs
john-sharratt: done

lib/wasi/src/lib.rs
```diff
@@ -72,26 +74,64 @@ pub enum WasiError {
...
+            let delta = std::time::Instant::now() - start;
+            if delta > duration { break; }
+            let remaining = duration - delta;
+            std::thread::sleep(remaining.min(std::time::Duration::from_millis(10)));
+            if std::time::Instant::now() - start >= duration { break; }
+        }
+    }
+
+    // Terminates this thread
+    pub fn terminate(&self, code: u32) {
```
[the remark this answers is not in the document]
john-sharratt: I deleted the function from the pull request as its not needed anymore anyway - I use the runtime engine to raise this from the terminal instead

john-sharratt:
Resolved the merge conflicts and fixed a bug in the memory optimized memory copies which were causing a segmentation fault

john-sharratt:
Merge from master appears to break wasm32.
See https://github.com/wasmerio/wasmer/issues/2733

syrusakbary:
@john-sharratt can you update the PR solving the conflicts? (I changed the base from `master` to `wasmer3` so we can merge the PR)

john-sharratt:
righto - will take a look - there is another merge request after this one - kept it separate so that the pull request isn't too big.
will drop the second merge request once this one is committed.

john-sharratt:
Merge conflicts resolved... need to do some testing on it but its "Kingsday" here so will be between things.

This is the next branch we need to merge...
https://github.com/john-sharratt/wasmer/tree/granular_locks

It's got a changes that will need reviewing first of course

john-sharratt:
Resolved the merge conflicts on the second merge request thats to come
https://github.com/john-sharratt/wasmer/pull/1

I will fill out the details of the merge request to explain what was changed

lib/api/src/js/externals/memory.rs
```diff
@@ -243,6 +243,22 @@ impl Memory {
         self.view.clone()
     }
 
+    /// A theoretical alais to `Self::view::<u8>` but it returns a `js::Uint8Array` in this case.
+    ///
+    /// This code is going to be refactored. Use it as your own risks.
```
epilys, line 248:
```suggestion
    /// This code is going to be refactored. Use it as your own risk.
```

lib/vfs/src/lib.rs
```diff
@@ -177,7 +182,28 @@ pub trait VirtualFile: fmt::Debug + Send + Write + Read + Seek + 'static + Upcas
...
+    }
+
+    /// Returns the number of bytes available.  This function must not block
+    /// Defaults to `None` which means the number of bytes is unknown
+    fn bytes_available_write(&self) -> Result<Option<usize>> {
+        Ok(None)
+    }
+
+    // Indicates if the file has been open or close. This function must not block
+    // Defaults to a status of constantly open
```
epilys, line 203:
```suggestion
    /// Indicates if the file is opened or closed. This function must not block
    /// Defaults to a status of being constantly open
```

lib/wasi/src/lib.rs
```diff
@@ -72,31 +74,62 @@ pub enum WasiError {
...
     #[wasmer(export)]
     memory: LazyInit<Memory>,
 }
 
-impl WasiEnv {
-    pub fn new(state: WasiState) -> Self {
-        Self {
-            state: Arc::new(Mutex::new(state)),
-            memory: LazyInit::new(),
+/// The wasi thread dereferences into the wasi environment
```
epilys, line 89:
```suggestion
/// The WASI thread dereferences into the WASI environment
```

lib/wasi/src/lib.rs
```diff
@@ -72,31 +74,62 @@ pub enum WasiError {
...
+impl Deref for WasiThread {
+    type Target = WasiEnv;
+
+    fn deref(&self) -> &WasiEnv {
+        &self.env
+    }
+}
+
+impl WasiThread {
+    // Returns the unique ID of this thread
```
epilys, line 99:
```suggestion
    /// Returns the unique ID of this thread
```

lib/wasi/src/lib.rs
```diff
@@ -72,31 +74,62 @@ pub enum WasiError {
...
+    pub fn thread_id(&self) -> u32 {
+        self.id
+    }
+
+    // Yields execution
```
epilys, line 104:
```suggestion
    /// Yields execution
```

lib/wasi/src/lib.rs
```diff
@@ -72,31 +74,62 @@ pub enum WasiError {
...
+    pub fn yield_now(&self) {
+        if let Some(callback) = self.on_yield.as_ref() {
+            callback(self);
+        }
+        std::thread::yield_now();
+    }
+
+    // Sleeps for a period of time
```
epilys, line 112:
```suggestion
    /// Sleeps for a period of time
```

lib/wasi/src/lib.rs
```diff
@@ -72,31 +74,62 @@ pub enum WasiError {
...
+            self.yield_now();
+            let delta = std::time::Instant::now() - start;
+            if delta > duration {
+                break;
+            }
+            let remaining = duration - delta;
+            std::thread::sleep(remaining.min(std::time::Duration::from_millis(10)));
+            if std::time::Instant::now() - start >= duration {
+                break;
+            }
```
epilys, line 125: `Duration` math ops might panic. Better use the `checked_` versions which return `None` on error/bug:

https://doc.rust-lang.org/std/time/struct.Instant.html#method.checked_duration_since

lib/wasi/src/lib.rs
```diff
@@ -116,12 +149,77 @@ impl WasiEnv {
...
     }
 
+    /// Get a reference to the memory
+    pub fn memory(&self) -> &Memory {
+        self.memory_ref()
+            .expect("Memory should be set on `WasiThread` first")
+    }
+
+    // Copy the lazy reference so that when its initialized during the
+    // export phase that all the other references get a copy of it
```
epilys, line 165:
```suggestion
    /// Copy the lazy reference so that when its initialized during the
    /// export phase that all the other references get a copy of it
```

lib/wasi/src/lib.rs
```diff
@@ -133,138 +231,136 @@ impl WasiEnv {
...
 
-    pub(crate) fn get_memory_and_wasi_state(
-        &self,
-        _mem_index: u32,
-    ) -> (&Memory, MutexGuard<WasiState>) {
-        let memory = self.memory();
-        let state = self.state.lock().unwrap();
-        (memory, state)
+    // Copy the lazy reference so that when its initialized during the
+    // export phase that all the other references get a copy of it
```
epilys, line 240:
```suggestion
    /// Copy the lazy reference so that when it's initialized during the
    /// export phase, all the other references get a copy of it
```

lib/wasi/src/lib.rs
```diff
@@ -133,138 +231,136 @@ impl WasiEnv {
...
+            "path_symlink" => Function::new_native_with_env(store, thread.clone(), path_symlink),
+            "path_unlink_file" => Function::new_native_with_env(store, thread.clone(), path_unlink_file),
+            "poll_oneoff" => Function::new_native_with_env(store, thread.clone(), legacy::snapshot0::poll_oneoff),
+            "proc_exit" => Function::new_native_with_env(store, thread.clone(), proc_exit),
+            "proc_raise" => Function::new_native_with_env(store, thread.clone(), proc_raise),
+            "random_get" => Function::new_native_with_env(store, thread.clone(), random_get),
+            "sched_yield" => Function::new_native_with_env(store, thread.clone(), sched_yield),
+            "sock_recv" => Function::new_native_with_env(store, thread.clone(), sock_recv),
+            "sock_send" => Function::new_native_with_env(store, thread.clone(), sock_send),
+            "sock_shutdown" => Function::new_native_with_env(store, thread.clone(), sock_shutdown),
```
epilys, line 310: `thread` is owned, so the last evaluated `Function` constructor could skip the `clone`. This depends on the order the macro evaluates its parameters, but the following suggestion might be the correct one:

```suggestion
            "sock_shutdown" => Function::new_native_with_env(store, thread, sock_shutdown),
```

lib/wasi/src/lib.rs
```diff
@@ -133,138 +231,136 @@ impl WasiEnv {
...
+            "path_symlink" => Function::new_native_with_env(store, thread.clone(), path_symlink),
+            "path_unlink_file" => Function::new_native_with_env(store, thread.clone(), path_unlink_file),
+            "poll_oneoff" => Function::new_native_with_env(store, thread.clone(), poll_oneoff),
+            "proc_exit" => Function::new_native_with_env(store, thread.clone(), proc_exit),
+            "proc_raise" => Function::new_native_with_env(store, thread.clone(), proc_raise),
+            "random_get" => Function::new_native_with_env(store, thread.clone(), random_get),
+            "sched_yield" => Function::new_native_with_env(store, thread.clone(), sched_yield),
+            "sock_recv" => Function::new_native_with_env(store, thread.clone(), sock_recv),
+            "sock_send" => Function::new_native_with_env(store, thread.clone(), sock_send),
+            "sock_shutdown" => Function::new_native_with_env(store, thread.clone(), sock_shutdown),
```
epilys, line 363:
```suggestion
            "sock_shutdown" => Function::new_native_with_env(store, thread, sock_shutdown),
```

lib/wasi/src/state/types.rs
```diff
@@ -218,13 +219,60 @@ pub(crate) fn poll(
...
+                }
+                _ => {}
+            }
+        }
+        let revents = builder.build();
+        if revents != 0 {
+            ret += 1;
+        }
+        seen_events[n] = revents;
+    }
```
epilys, line 269: Since the slices have the same lengths at this point, you can do something like:

```suggestion
for ((file, events), seen_events) in files.iter().zip(events.iter()).zip(seen_events.iter_mut())
{
    let mut builder = PollEventBuilder::new();

    let can_read = file.bytes_available_read()?.map_or(false, |s| s > 0);
    let can_write = file.bytes_available_write()?.map_or(false, |s| s > 0);
    let is_closed = !file.is_open();

    for event in iterate_poll_events(events) {
        match event {
            PollEvent::PollIn if can_read => {
                builder = builder.add(PollEvent::PollIn);
            }
            PollEvent::PollOut if can_write => {
                builder = builder.add(PollEvent::PollOut);
            }
            PollEvent::PollHangUp if is_closed => {
                builder = builder.add(PollEvent::PollHangUp);
            }
            PollEvent::PollInvalid if is_closed => {
                builder = builder.add(PollEvent::PollInvalid);
            }
            PollEvent::PollError if is_closed => {
                builder = builder.add(PollEvent::PollError);
            }
            _ => {}
        }
    }
    let revents = builder.build();
    if revents != 0 {
        ret += 1;
    }
    *seen_events = revents;
}
```

lib/wasi/src/state/types.rs
```diff
@@ -218,13 +219,60 @@ pub(crate) fn poll(
...
-    _events: &[PollEventSet],
-    _seen_events: &mut [PollEventSet],
-) -> Result<(), FsError> {
-    unimplemented!("VirtualFile::poll is not implemented for non-Unix-like targets yet");
+    files: &[&dyn VirtualFile],
+    events: &[PollEventSet],
+    seen_events: &mut [PollEventSet],
+) -> Result<u32, FsError> {
+    if !(files.len() == events.len() && events.len() == seen_events.len()) {
+        return Err(FsError::InvalidInput);
```
epilys, line 229: It'd probably be a good idea to do a `debug!` print here to show the reason why it failed, since `InvalidInput` doesn't have a human readable description.

john-sharratt:
@Amanieu I just pushed a commit on this that seems to fix a regression issue that comes from the view caching you implemented - it seems that when accessing a cached view object from within a WebAssembly execution context the view length is zero and inaccessible, the only solution I could find was to create a new view every time like the original code.

Here is the commit
https://github.com/wasmerio/wasmer/pull/2689/commits/00bd90ad4a8534d667dc2fc498b1fc5666c09c9d

john-sharratt:
@Amanieu There's a regression issue on the wasmer3 branch which break qjs.wasm however I can't find the cause only the commit that caused it - lodged a bug here
https://github.com/wasmerio/wasmer/issues/2865

In the mean-time I will release the code with the regression issue so that I can keep going on the network code

john-sharratt:
Confirmed that https://github.com/wasmerio/wasmer/issues/2865 was fixed with https://github.com/wasmerio/wasmer/pull/2870

```
tokera.sh:/$ qjs
QuickJS - Type "\h" for help
qjs > 1 + 2
1 + 2
3
qjs > ^C
tokera.sh:/$
```

john-sharratt:
ouch!

```
john@LappyWorld:/prog/wasmer$ git merge upstream/wasmer3
Auto-merging CHANGELOG.md
Auto-merging Cargo.lock
CONFLICT (content): Merge conflict in Cargo.lock
Auto-merging Cargo.toml
Auto-merging lib/api/Cargo.toml
Auto-merging lib/api/src/js/externals/memory.rs
CONFLICT (content): Merge conflict in lib/api/src/js/externals/memory.rs
Auto-merging lib/api/src/js/mod.rs
Auto-merging lib/api/src/lib.rs
Auto-merging lib/api/src/sys/mod.rs
Auto-merging lib/api/src/sys/tunables.rs
Auto-merging lib/c-api/src/wasm_c_api/unstable/wasi.rs
CONFLICT (content): Merge conflict in lib/c-api/src/wasm_c_api/unstable/wasi.rs
Auto-merging lib/c-api/src/wasm_c_api/wasi/mod.rs
Auto-merging lib/cli/src/commands/run/wasi.rs
CONFLICT (content): Merge conflict in lib/cli/src/commands/run/wasi.rs
Auto-merging lib/compiler-singlepass/Cargo.toml
CONFLICT (content): Merge conflict in lib/compiler-singlepass/Cargo.toml
Auto-merging lib/compiler-singlepass/src/compiler.rs
Auto-merging lib/compiler-singlepass/src/config.rs
Auto-merging lib/compiler/Cargo.toml
Auto-merging lib/compiler/src/module.rs
Auto-merging lib/compiler/src/relocation.rs
Auto-merging lib/compiler/src/trap.rs
Auto-merging lib/emscripten/src/env/unix/mod.rs
CONFLICT (content): Merge conflict in lib/emscripten/src/env/unix/mod.rs
Auto-merging lib/emscripten/src/syscalls/mod.rs
Auto-merging lib/emscripten/src/syscalls/unix.rs
Auto-merging lib/engine-dylib/Cargo.toml
Auto-merging lib/engine-dylib/src/artifact.rs
Auto-merging lib/engine-dylib/src/engine.rs
Auto-merging lib/engine-dylib/src/serialize.rs
Auto-merging lib/engine-staticlib/Cargo.toml
Auto-merging lib/engine-staticlib/src/artifact.rs
Auto-merging lib/engine-universal/Cargo.toml
Auto-merging lib/engine-universal/src/artifact.rs
CONFLICT (content): Merge conflict in lib/engine-universal/src/artifact.rs
Auto-merging lib/engine-universal/src/engine.rs
Auto-merging lib/engine/Cargo.toml
Auto-merging lib/engine/src/artifact.rs
CONFLICT (content): Merge conflict in lib/engine/src/artifact.rs
Auto-merging lib/engine/src/lib.rs
Auto-merging lib/types/Cargo.toml
CONFLICT (content): Merge conflict in lib/types/Cargo.toml
Auto-merging lib/types/src/entity/boxed_slice.rs
Auto-merging lib/types/src/entity/primary_map.rs
Auto-merging lib/types/src/entity/secondary_map.rs
Auto-merging lib/types/src/lib.rs
CONFLICT (content): Merge conflict in lib/types/src/lib.rs
Auto-merging lib/types/src/module.rs
Auto-merging lib/types/src/trapcode.rs
CONFLICT (content): Merge conflict in lib/types/src/trapcode.rs
Auto-merging lib/types/src/types.rs
Auto-merging lib/types/src/vmoffsets.rs
CONFLICT (content): Merge conflict in lib/types/src/vmoffsets.rs
Auto-merging lib/vfs/src/lib.rs
Auto-merging lib/vm/Cargo.toml
Auto-merging lib/vm/src/export.rs
Auto-merging lib/vm/src/instance/mod.rs
Auto-merging lib/vm/src/lib.rs
CONFLICT (content): Merge conflict in lib/vm/src/lib.rs
Auto-merging lib/vm/src/libcalls.rs
CONFLICT (content): Merge conflict in lib/vm/src/libcalls.rs
Auto-merging lib/vm/src/memory.rs
CONFLICT (content): Merge conflict in lib/vm/src/memory.rs
Auto-merging lib/vm/src/table.rs
CONFLICT (content): Merge conflict in lib/vm/src/table.rs
Auto-merging lib/vm/src/vmcontext.rs
Auto-merging lib/wasi/src/lib.rs
CONFLICT (content): Merge conflict in lib/wasi/src/lib.rs
Auto-merging lib/wasi/src/macros.rs
CONFLICT (content): Merge conflict in lib/wasi/src/macros.rs
Auto-merging lib/wasi/src/syscalls/legacy/snapshot0.rs
CONFLICT (content): Merge conflict in lib/wasi/src/syscalls/legacy/snapshot0.rs
Auto-merging lib/wasi/src/syscalls/mod.rs
CONFLICT (content): Merge conflict in lib/wasi/src/syscalls/mod.rs
Auto-merging lib/wasi/src/syscalls/wasm32.rs
CONFLICT (content): Merge conflict in lib/wasi/src/syscalls/wasm32.rs
Auto-merging lib/wasi/src/utils.rs
Auto-merging tests/lib/engine-dummy/Cargo.toml
Auto-merging tests/lib/engine-dummy/src/artifact.rs
Auto-merging tests/lib/wast/src/wasi_wast.rs
CONFLICT (content): Merge conflict in tests/lib/wast/src/wasi_wast.rs
Automatic merge failed; fix conflicts and then commit the result.
```

john-sharratt:
[cry]

Not sure if this is because of the merge conflicts being resolved incorrectly or if its new code.... but it seems to be rather broken now

```
tokera.sh:/$ qjs
tokera.sh:/$ wapm install python/python
[INFO] Installing python/python@0.1.0
```

```
INFO tokweb/src/pool.rs:347 worker started (index=1, type=Shared)
327.main.js:648 WARN term-lib/src/bus/thread.rs:375 wasm-bus::call-reply - data copy failed - memory access out of bounds
```

lib/engine-universal/src/artifact.rs
```diff
@@ -4,6 +4,10 @@
 use crate::engine::{UniversalEngine, UniversalEngineInner};
 use crate::link::link_module;
 use enumset::EnumSet;
+<<<<<<< HEAD
+use std::mem;
+=======
+>>>>>>> upstream/wasmer3
```
Amanieu, line 10: Merge conflict.

lib/types/src/trapcode.rs
```diff
@@ -12,7 +12,6 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 /// A trap code describing the reason for a trap.
-///
```
Amanieu, line 15: Leftover from merge?

lib/types/src/vmoffsets.rs
```diff
@@ -5,7 +5,6 @@
 //! module.
 
 #![deny(broken_intra_doc_links)]
-
```
Amanieu, line 8: Leftover from merge?

lib/api/src/js/externals/memory.rs
```diff
@@ -243,6 +243,22 @@ impl Memory {
         self.view.clone()
     }
 
+    /// A theoretical alias to `Self::view::<u8>` but it returns a `js::Uint8Array` in this case.
+    ///
+    /// This code is going to be refactored. Use it as your own risk.
+    #[doc(hidden)]
+    pub fn uint8view_with_byte_offset_and_length(
```
Amanieu, line 250: This function doesn't seem to actually be called anywhere? Did this get lost with the merge?

Amanieu (commented):
Sorry about the breakage, I rebased the wasmer3 branch on top of the latest changes from master which unfortunately led to many merge conflicts.

john-sharratt:
The rebase failed multiple times as it was just too complex to do... I ended up reimplmenting it using branch compares and too many hours on GIT.

Anyhow the new pull request is here and it now includes the other pull request that was waiting behind this one so its quite a big commit.
https://github.com/wasmerio/wasmer/pull/2887

heyjdp:
Closing PR, superseded by PR #2887

PR #2184, opened 2021-03-12
add a compiler test to check for deterministic

xofyarg:
This is to test #2173, the empty test always pass while the table one
fails sometimes.

# Description

# Review

- [ ] Add a short description of the the change to the CHANGELOG.md file

PR #2692, opened 2021-11-22
Make module serialization deterministic

### LINKED ISSUE #2173: Reproducible builds
### Summary

At version 0.17, the compiler is able to generate the same output with the same input. But after upgrading to 1.0.2, the compilation results differ each time even the wasm file is the same. I tried LLVM and cranelift. I'm using the default options, not sure if I missed anything.

Amanieu:
Fixes #2173

Includes the tests from #2184.

### DIFF
--- a/fuzz/fuzz_targets/deterministic.rs
+++ b/fuzz/fuzz_targets/deterministic.rs
@@ -6,8 +6,8 @@ use wasmer::{CompilerConfig, Engine, Module, Store};
 use wasmer_compiler_cranelift::Cranelift;
 use wasmer_compiler_llvm::LLVM;
 use wasmer_compiler_singlepass::Singlepass;
-use wasmer_engine_jit::JIT;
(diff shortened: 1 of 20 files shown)


fuzz/fuzz_targets/deterministic.rs
```diff
@@ -0,0 +1,77 @@
...
+    // compile_and_compare(
+    //     "jit-singlepass",
+    //     JIT::new(compiler.clone()).engine(),
+    //     &wasm_bytes,
+    // );
+    // compile_and_compare(
+    //     "native-singlepass",
+    //     Native::new(compiler).engine(),
+    //     &wasm_bytes,
+    // );
```
syrusakbary, line 76: Can we re-enable this ones?

syrusakbary (approved):

syrusakbary:
It seems tests failed because of not enough memory in the VM. We can try to merge

bors r+

PR #2693, opened 2021-11-23
Add extra validation to compilers & engines

### LINKED ISSUE #1567: Singlepass should panic when run in non-AVX chips
When singlepass runs in non-AVX chips, a panic should be triggered.

### LINKED ISSUE #2590: Staticlib engine causes rust panic when creating a wasm instance
### Describe the bug

`wasmer 2.0.0 | rustc 1.54.0-nightly (3e99439f4 2021-05-17) | x86_64`

When making a wasm module instance with the `staticlib` engine, wasmer panics:
```
thread 'main' panicked at 'index out of bounds: the len is 0 but the index is 0', /Users/GW/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-vm-2.0.0/src/instance/mod.rs:1079:29
```

Here is a test project.
[wasmer-bug.zip](https://github.com/wasmerio/wasmer/files/7255691/wasmer-bug.zip)

When digging in to the stack trace, at `lookup_by_declaration wasmer-vm-2.0.0/src/instance/mod.rs:1079`, 
`instance-ref.functions.elems.length` is 0. When instantiating a module with `universal` or `dylib`, the length instead shows up as 2.

### Steps to reproduce

For example,
1. Open the test project: [wasmer-bug.zip](https://github.com/wasmerio/wasmer/files/7255691/wasmer-bug.zip)
2. Run with `cargo run -- --engine universal test.wasm`
3. See an output value
4. Run with `cargo run -- --engine staticlib test.wasm`
5. See the panic dump

### Expected behavior

That wasmer does not panic when using staticlib

### Actual behavior
That wasmer panics when using staticlib

### Additional context

None that I can think of.

Amanieu:
- Cranelift and singlepass now properly cross-compile with no dependency on the host target.
- Staticlib engine now panics if you try to run a freshly compiled module.
- CPU features used when a module was compiled are now checked against the host CPU features during instantiation.

Fixes #1567
Fixes #2590

### DIFF
--- a/lib/compiler-cranelift/src/compiler.rs
+++ b/lib/compiler-cranelift/src/compiler.rs
@@ -21,6 +21,7 @@ use gimli::write::{Address, EhFrame, FrameTable};
 use loupe::MemoryUsage;
 use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
 use std::sync::Arc;
+use target_lexicon::{Architecture, OperatingSystem};
 use wasmer_compiler::CompileError;
 use wasmer_compiler::{CallingConvention, ModuleTranslationState, Target};
 use wasmer_compiler::{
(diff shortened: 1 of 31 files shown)


lib/compiler-cranelift/src/compiler.rs
```diff
@@ -111,28 +110,36 @@ impl Compiler for CraneliftCompiler {
...
-                // 6 is the size of the jmp instruction. The relocated address must follow
-                offset: 6,
-                addend: 0,
-            }],
+        let probestack_trampoline_relocation_target = if target.triple().operating_system
+            == OperatingSystem::Linux
+            && matches!(
+                target.triple().architecture,
+                Architecture::X86_32(_) | Architecture::X86_64
+            ) {
```
ptitSeb, line 118: this code will not work on 32bits x86. the ff 25 00 00 00 00 translate to jmp [0x00000000].
Or the offset needs to be 2 instead of 6 in 32bits

ptitSeb (approved):

PR #2694, opened 2021-11-24
JS API improvement

syrusakbary:
This PR improves the JS api by adding a way to extend the ImportObject with a JS Object.

This is needed to fulfill the wasmer-js new API improvements.


lib/api/src/js/instance.rs
```diff
@@ -112,29 +112,40 @@ impl Instance {
     /// Is expected that the function [`Instance::init_envs`] is run manually
     /// by the user in case the instance has any Wasmer imports, so the function
     /// environments are properly initiated.
-    pub fn from_module_and_instance(module: &Module, instance: WebAssembly::Instance) -> Self {
+    #[doc(hidden)]
```
Amanieu, line 115: Why is this hidden? Should it be a private (non-pub) function instead? Is this actually safe to call or does the lack of env initialization make it potentially unsound?

lib/api/src/js/js_import_object.rs
```diff
@@ -0,0 +1,71 @@
...
+/// of a `Module` with imports that are provided directly from
+/// Javascript with a JS Object.
+#[derive(Clone, Default)]
+pub struct JsImportObject {
+    module_imports: HashMap<(String, String), ExternType>,
+    object: js_sys::Object,
+}
+
+unsafe impl Send for JsImportObject {}
+unsafe impl Sync for JsImportObject {}
```
Amanieu, line 14: JS objects are not Send/Sync, so this is technically unsound. However I believe this is only properly handled in #2646 and it might be acceptable to let this slide for now.

lib/api/src/js/js_import_object.rs
```diff
@@ -0,0 +1,71 @@
...
+
+impl JsImportObject {
+    /// Create a new `JsImportObject`.
```
Amanieu, line 17: There is no explanation of what the `module` parameter does. From the implementation I am guess that it is used to assign a type for each import, but this is not obvious from the documentation.

lib/api/src/js/js_import_object.rs
syrusakbary: I added a side comment to reflect this

Amanieu (approved):

syrusakbary:
Merging directly since last tests passed and the only change was lint (one space removal on docs)

PR #2706, opened 2021-12-01
Refactor Singlepass

ptitSeb:
# Description
Refactor SinglePass compiler to ease the implementation of future architectures.


Amanieu (approved):

lib/compiler-singlepass/src/machine_x64.rs
```diff
@@ -6938,7 +6939,7 @@ mod test {
 
         machine.release_locations_keep_state(&locs);
     }
-}
```
Amanieu: Could you find a way to make this test work properly again, or delete it entirely if you really can't. Don't leave commented code around.

ptitSeb: I haven't found any practical way to get this test running with the changes I have made.

PR #2709, opened 2021-12-02
Install LLVM for the macos ARM build

syrusakbary:
Install LLVM for the macos ARM build (as LLVM seems to not be included in the binary release)


heyjdp:
Closing PR, no longer relevant

PR #2723, opened 2021-12-14
fix for #2721

### LINKED ISSUE #2721: Pointers to static variables are sometimes passed incorrectly with the singlepass compiler
### Describe the bug

When building machine code for a wasm function which calls a host function and passes in a pointer to a `'static` string, the offset to the string data is passed through incorrectly (`0` is passed).

### Steps to reproduce

`make test-compilers` at https://github.com/wasmerio/wasmer/compare/master...nagisa:test-for-argpassing

### Expected behavior

The pointer argument received by the host function `banana` should be an offset to the first byte of UTF-8 encoding of the string `"bananapeach"` within the memory.

The behaviour should be consistent between backends.

### Actual behavior

Test when run with the `singlepass` backend passes through an offset 0 (even though the wasm code itself has a proper offset encoded within it!)

### Additional information

This is a regression introduced by 8f855cf9ccce1d0c0587f9a34c8e912331643e26.

ptitSeb:
# Description
Fix for #2721, regression introduce with Windows backend for Singlepass.


lib/compiler-singlepass/src/machine_x64.rs
```diff
@@ -1807,15 +1807,11 @@ impl Machine for MachineX86_64 {
...
-
-                // Use R9 as the temporary register here, since:
-                // - It is a temporary register that is not used for any persistent value.
-                // - This register as an argument location is only written to after `sort_call_movs`.'
+                // Push R9 value slot to be exchange with `mov`.
+                self.assembler.emit_push(Size::S64, Location::GPR(GPR::R9));
                 self.reserve_unused_temp_gpr(GPR::R9);
                 self.move_location(Size::S64, loc, Location::GPR(GPR::R9));
-                self.move_location(
+                self.assembler.emit_xchg(
```
Amanieu, line 1814: On x86, the `xchg` instruction always does an atomic swap, which is very slow. Try to use a normal `mov` instead.

ptitSeb: I know, but I need to save the value of the register.

Amanieu: It's probably cheaper to add another pair of `push`/`pop` for `r9` than to use `xchg`:
```
push rax (dummy)
push r9
mov r9, #imm64
mov [sp + 8], r9
pop r9
```

nagisa:
Please retain the authorship for the test case. Either cherry-pick the commit from my branch or if you'd rather not do so for some reason, I can submit a PR with the changes I've made as a separate PR once the fix here lands.

lib/compiler-singlepass/src/machine_x64.rs
ptitSeb: No it's not.
I have done this simple program:
```
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>
#include <time.h>


int main(int argc, const char* argv)
{
    struct timespec t1, t2;

    clock_gettime(CLOCK_MONOTONIC, &t1);
    #define NUM 100000
    for(int i=0; i<NUM; ++i)
        asm volatile ("\n\
            push %%r9\n\
            mov $0x123456789abcdef, %%r9\n\
            xchg (%%rsp), %%r9\n\
            pop  %%r9\n\
        "::);
    clock_gettime(CLOCK_MONOTONIC, &t2);
    uint64_t t = t1.tv_sec*1000000000LL + t1.tv_nsec;
    t = (t2.tv_sec*1000000000LL + t2.tv_nsec) - t;
    printf("Xchg: %llu nsec for %d occurences\n", t, NUM);
    for(int i=0; i<NUM; ++i)
        asm volatile ("\n\
            push %%rax\n\
            push %%r9\n\
            mov $0x123456789abcdef, %%r9\n\
            mov 0x8(%%rsp), %%r9\n\
            pop  %%r9\n\
            pop %%rax\n\
        "::);
    clock_gettime(CLOCK_MONOTONIC, &t2);
    t = t1.tv_sec*1000000000LL + t1.tv_nsec;
    t = (t2.tv_sec*1000000000LL + t2.tv_nsec) - t;
    printf("Push: %llu nsec for %d occurences\n", t, NUM);
}
```
And while I have some variation, the xchg method is always faster:
```
seb@Seb-PC:~/git/wasmer$ ./test
Xchg: 1911800 nsec for 100000 occurences
Push: 2216700 nsec for 100000 occurences
seb@Seb-PC:~/git/wasmer$ ./test
Xchg: 1156300 nsec for 100000 occurences
Push: 1452800 nsec for 100000 occurences
seb@Seb-PC:~/git/wasmer$ ./test
Xchg: 1156000 nsec for 100000 occurences
Push: 1452800 nsec for 100000 occurences
seb@Seb-PC:~/git/wasmer$ ./test
Xchg: 1345700 nsec for 100000 occurences
Push: 1668700 nsec for 100000 occurences
seb@Seb-PC:~/git/wasmer$ ./test
Xchg: 1499400 nsec for 100000 occurences
Push: 1898400 nsec for 100000 occurences
seb@Seb-PC:~/git/wasmer$ ./test
Xchg: 1185400 nsec for 100000 occurences
Push: 1493200 nsec for 100000 occurences
seb@Seb-PC:~/git/wasmer$ ./test
Xchg: 1276400 nsec for 100000 occurences
Push: 1598700 nsec for 100000 occurences
seb@Seb-PC:~/git/wasmer$ ./test
Xchg: 1276800 nsec for 100000 occurences
Push: 1596000 nsec for 100000 occurences
```
The implicit LOCK is not that slow, compare to a write+read in memory that the push/pop.

ptitSeb:
> 

Ok, I'll see what I can do so you keep the authorship, no problem.

Amanieu (approved):

PR #2724, opened 2021-12-14
Added access to the raw Wasm instance

syrusakbary:
In order to solve this issue in wasmer-js we need to expose access to the underlying WebAssembly instance:

https://github.com/wasmerio/wasmer-js/issues/284

### DIFF
--- a/lib/api/src/js/instance.rs
+++ b/lib/api/src/js/instance.rs
@@ -173,6 +173,12 @@ impl Instance {
     pub fn store(&self) -> &Store {
         self.module.store()
     }
+
+    /// Returns the inner WebAssembly Instance
+    #[doc(hidden)]
+    pub fn raw(&self) -> &WebAssembly::Instance {
+        &self.instance
+    }
 }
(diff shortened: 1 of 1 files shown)


lib/api/src/js/instance.rs
```diff
@@ -173,6 +173,12 @@ impl Instance {
     pub fn store(&self) -> &Store {
         self.module.store()
     }
+
+    /// Returns the inner WebAssembly Instance
+    #[doc(hidden)]
```
Amanieu, line 178: Is this API meant to be publicly usable? In that case it shouldn't be hidden from the documentation.

syrusakbary: Because it differs from the SYS API, I rather keep it hidden so people don't rely on it.

Although, we need it for wasmer-js

Amanieu (approved):

syrusakbary:
Merging directly to fasten the process (no existing code has been changed)

PR #2727, opened 2021-12-15
fix(compiler) Fix singlpass release_locations function

ptitSeb:
# Description
Fix a mistake in singlepass refactor


syrusakbary:
Can we have a test verifying the fix?

ptitSeb:
> Can we have a test verifying the fix?

I don't have any error with this, but there still was a mistake that slipped in durring the refactor.

Amanieu (approved):

PR #2734, opened 2021-12-26
Remove llvm from docs.rs feature set (trouble with building)

jcaesar:
Followup on #2605. It [currently](https://docs.rs/crate/wasmer/2.1.1/builds/480618) breaks the build on docs.rs, sorry for that.

There should be a way to get it added to the installed dependencies, but when I tried to reproduce [the environment](https://github.com/rust-lang/crates-build-env/blob/master/linux/packages.txt#L1015), it was already installed, yet there was a problem with `could not find native static library Polly, perhaps an -L flag is missing?`. I figure it's more trouble than it's worth.

### DIFF
--- a/lib/api/Cargo.toml
+++ b/lib/api/Cargo.toml
@@ -155,4 +155,4 @@ js-default = ["js", "std", "wasm-types-polyfill"]
 wasm-types-polyfill = ["js", "wasmparser"]
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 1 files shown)


jcaesar:
Thinking about it: I have no idea how https://docs.rs/llvm-sys/120.2.1/llvm_sys/ builds just fine…
Maybe there is some cleverer way of doing this.

Amanieu (approved):

PR #2743, opened 2022-01-04
Fix generation of wasmer.pc file

### LINKED ISSUE #2740: wasmer 2.1.1 build issues
I attempted to write a Gentoo ebuild for wasmer-2.1.1 based on the information in PACKAGING.md and found a couple of issues.

- wasmer does not honor WASMER_INSTALL_PREFIX.
- The wasmer.pc file that is installed is empty.

I attempted to run the make commands manually and saw the same issues.

Here are the steps I ran manually. If this is fixed it should solve the packaging issue.

```
$ export WASMER_INSTALL_PREFIX=/usr
$ make
$ DESTDIR=/tmp/test make install
```

After you run these steps, you will see that the installation is in /tmp/test instead of /tmp/test/usr. Also, you will see that the wasmer.pc file is empty.

Thanks for your time,

William

Amanieu:
Partially fixes #2740

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -713,7 +713,7 @@ install-misc:
 install-pkgconfig:
 	# Make sure WASMER_INSTALL_PREFIX is set during build
 	unset WASMER_DIR; \
-	if pc="$$(target/release/wasmer config --pkg-config 1>/dev/null 2>/dev/null)"; then \
+	if pc="$$(target/release/wasmer config --pkg-config 2>/dev/null)"; then \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -713,7 +713,7 @@ install-misc:
 install-pkgconfig:
 	# Make sure WASMER_INSTALL_PREFIX is set during build
 	unset WASMER_DIR; \
-	if pc="$$(target/release/wasmer config --pkg-config 1>/dev/null 2>/dev/null)"; then \
```
ptitSeb, line 716: `--pkg-config 1>/dev/null 2>&1` can also be used.

ptitSeb (approved):

PR #2746, opened 2022-01-07
Fix invoking `wasmer binfmt register` from `$PATH`

jcaesar:
# Description
Currently, invoking `wasmer binfmt register` from `$PATH` fails. 
```
$ wasmer binfmt register 
error: Cannot get path to wasmer executable
```

(Currently, only invoking it by absolute path (`$(which wasmer) binfmt register`, e.g. as would be from a systemd unit) works.)

I also added a commit that changes some bit twiddling with file modes to a nicer semantic check, but it does add a dependency. Please tell me if you don't want that.

# Review

- [ ] Add a short description of the change to the CHANGELOG.md file - Not necessary?

### DIFF
--- a/lib/cli/Cargo.toml
+++ b/lib/cli/Cargo.toml
@@ -55,6 +55,9 @@ fern = { version = "0.6", features = ["colored"], optional = true }
 log = { version = "0.4", optional = true }
 tempfile = "3"
 
+[target.'cfg(target_os = "linux")'.dependencies]
+unix_mode = "0.1.3"
+
 [features]
 # Don't add the compiler features in default, please add them on the Makefile
 # since we might want to autoconfigure them depending on the availability on the host.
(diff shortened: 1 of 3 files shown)


lib/cli/Cargo.toml
```diff
@@ -54,6 +54,7 @@ cfg-if = "1.0"
 fern = { version = "0.6", features = ["colored"], optional = true }
 log = { version = "0.4", optional = true }
 tempfile = "3"
+unix_mode = "0.1.3"
```
syrusakbary, line 57: This should be required only in unix systems

jcfj: Good point, the entire module is `#[cfg(target_os = 'linux')]`. Fixed.

jcaesar: Indeed, the entire module is #[cfg(target_os = 'linux')]. Ty, fixed.

syrusakbary (approved):

PR #2747, opened 2022-01-07
Use a standard header for metadata in all serialized modules

Amanieu:
This header includes an ABI version which will reject any serialized
modules from an incompatible version of Wasmer. Also, it replaces all of
the custom per-engine code for encoding the metadata length.


ptitSeb (approved):
I guess all previously generated modules will be incompatible. That's kind-of a breaking change.

lib/engine-universal/src/artifact.rs
```diff
@@ -46,11 +44,9 @@ pub struct UniversalArtifact {
 }
 
 impl UniversalArtifact {
-    const MAGIC_HEADER: &'static [u8; 22] = b"\0wasmer-universal\0\0\0\0\0";
-
```
syrusakbary, line 50: I'd like to preserve this. I see one thing as the metadata header, and the other is the file header.
Meaning: I think it will be useful to separate the file format from the metadata serialized version.

Amanieu: The metadata header itself already has a magic number and a length, so I felt that this was somewhat redundant.

Amanieu: Also in the case of the universal engine, there is *only* metadata and nothing else.

syrusakbary: What I mean is that the magic is only relevant for the file format (thus, for the universal engine).
We don't need magic in the metadata itself, since the metadata field name in the dynamic library is already the identifier.

Therefore, I think a better abstraction is to put the magic in the engine artifact and the rest (metadata version, length, ...) on the metadata as the PR is doing

Amanieu: The metadata header needs to be 16 bytes since the serialized metadata is required to be 16-byte aligned. Since the length only uses 8 bytes, we have 8 bytes left that we might as well use for a magic.

Replacing the custom header of wasmer-universal with the standard `MetadataHeader` was done to make the code more maintainable by centralizing header-related code.

webmaster128:
> I guess all previously generated modules will be incompatible. That's kind-of a breaking change.

#2781 discusses the stability of the module format. There also a module breaking change between Wasmer 2.0.0 and  2.1.0 is shown.

PR #2748, opened 2022-01-11
Use trampolines for all libcalls in engine-universal and engine-dylib

### LINKED ISSUE #1722: probestack broken with engine-native and compiler-llvm

Amanieu:
In both of these engines, the compiled code may be loaded in memory far
from the Wasmer runtime which means that libcalls may not be reachable
through the normal relocation types. Instead a trampoline is needed to
allow reaching any address in the 64-bit address space.

In the case of engine-dylib, this is even worse since the symbols are
not exported by the executable without some special linker flags. The
solution here is to manually patch in the addresses at load time into
a data table of function pointers.

Fixes #1722


Amanieu:
This is still incomplete: LLVM is bypassing the trampoline generation since it generates the object file directly.

syrusakbary:
I think @ptitSeb will be the best one to review this

Amanieu:
Ready for review

syrusakbary:
This LGTM, @ptitSeb can you take a look to approve it or give feedback?

lib/compiler-cranelift/src/func_environ.rs
```diff
@@ -1132,9 +1132,7 @@ impl<'module_environment> BaseFuncEnvironment for FuncEnvironment<'module_enviro
         Ok(func.import_function(ir::ExtFuncData {
             name,
             signature,
-            // We currently allocate all code segments independently, so nothing
-            // is colocated.
-            colocated: false,
+            colocated: true,
```
syrusakbary, line 1135: I might be missing some context on what colocated means. Why is this necessary?

Amanieu: This is necessary to avoid text relocations. It means that all libcalls are reachable through normal call instructions (+/- 2G on x86, +/- 128M on aarch64) so it doesn't need to load the target from memory and do an indirect call. This is true since we generate trampolines that are near the compiled code.

The text relocations are due to the way Cranelift generates indirect calls: it places the target as a pointer in the .text section (since it puts everything in .text) and marks it with a relocation so the linker fills in the libcall address at load time.

ptitSeb (approved):

PR #2766, opened 2022-01-24
Remove an attempt to reserve a GPR when no GPR clobbering is occurring

nagisa:
# Description

This PR adds a test for another regression introduced by https://github.com/wasmerio/wasmer/commit/8f855cf9ccce1d0c0587f9a34c8e912331643e26. What happens here is that the `local.get 0` instructions reserve the relevant registers and when it comes a time to codegen a function call, the code fails to move one of the `imm64` value to the stack region used to pass arguments.

Turns out we had asserts which became unnecessary as part of an earlier fix submitted in https://github.com/wasmerio/wasmer/pull/2723 – we no longer clobber any registers when we move `imm64` to a stack, but we still attempted to reserve a temporary register for clobbering. Turns out there are situations where reserving an R9 can fail, and this just happened to be one of them. We don't reserve R9 anywhere else in the codebase, luckily, making scope of this kind of bug fairly limited.

As a future improvement I think it would make sense to see if we can `pick_gpr` or `pick_temp_gpr` here and avoid the `xchg` altogether if there are registers available for use, but I wanted to reduce the scope of this particular PR as much as possible in hopes of landing it more smoothly.

# Review

### DIFF
--- a/lib/compiler-singlepass/src/machine_x64.rs
+++ b/lib/compiler-singlepass/src/machine_x64.rs
@@ -1807,16 +1807,22 @@ impl Machine for MachineX86_64 {
     fn push_location_for_native(&mut self, loc: Location) {
         match loc {
             Location::Imm64(_) => {
-                // Push R9 value slot to be exchange with `mov`.
+                // x86_64 does not support `mov imm64, mem`. We must first place the immdiate value
+                // into a register and then write the register to the memory. Now the problem is
+                // that there might not be any registers available to clobber. In order to make
+                // this work out we spill a register thus retaining both the original value of the
+                // register and producing the required data at the top of the stack.
+                //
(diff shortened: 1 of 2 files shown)


ptitSeb (approved):

matklad:
as a meta, the issue was found by our fuzzing here:

https://github.com/near/nearcore/blob/96a4d1a2f76ec10280bf79ba3a5fa6035081e158/runtime/near-vm-runner/fuzz/fuzz_targets/runner.rs#L14-L17

syrusakbary:
It seems tests are failing in master because of `tracing-subscriber` using a non-safe version of `thread-local`.

We've just updated `Cargo.lock` dependencies on master to fix it, so if you could rebase with the latest commits, tests should pass and we should be good to merge.

Issue #1727, opened 2020-10-15
cranelift reports wrong trap types with dylib engine


nlewycky:
Testcase:
```
(module
  (func (export "div_s") (param $x i64) (param $y i64) (result i64) (i64.div_s (local.get $x) (local.get $y)))
)
(assert_trap (invoke "div_s" (i64.const 1) (i64.const 0)) "integer divide by zero")
```

Actual behaviour:
```
$ wasmer/target/debug/wasmer wast --cranelift --native ov.wast
error: failed to test the wast `ov.wast`
│   1: tests failed
╰─> 2: Failed directives on ov.wast:
         • expected 'integer divide by zero', got 'RuntimeError: call stack exhausted' (4:1)
```

The cause of the bug is that the native engine doesn't serialize the `CompiledFunctionFrameInfo`, which includes the Traps information.

This doesn't affect compiler-llvm because that emits calls to the wasmer_raise_trap libcall instead of using Traps in the frame info.

The fix is to start serializing it and loading it back in, such as with serde. However, Traps is expressed in terms of offsets in the code, and the linker may change those offsets as part of linking ("linker relaxation") and then a serialized listing in Traps would be incorrect. We can work around that if we link with BFD and Gold linkers that support the `--no-relax` flag to disable linker relaxation. However, LLD doesn't have this flag (it also may not perform relaxation at all, I don't know).

The other way is to serialize it in a manner that doesn't use serde, we emit the CodeOffsets in the traps as data which refers to those offsets in the code via relocations. If the linker performs relaxation then it has to update those relocations too (or it has to not relax that sequence, either way).

Note that ignoring the relaxation problem will produce a system that appears to work most of the time, including for the minimal test above. To test that, you'd need to produce an input wasm file that causes cranelift to emit a relaxable assembly sequence that the linker then actually does relax when building the shared object, and when run we hit a trap that is after the relaxed assembly sequence. It may be easier to hand-craft a .o file that triggers the problem.

jubianchi:
This bug exists with both Cranelift and SinglePass.

```
$ wasmer/target/debug/wasmer wast --singlepass --native ov.wast
error: failed to test the wast `ov.wast`
│   1: tests failed
╰─> 2: Failed directives on ov.wast:
         • expected 'integer divide by zero', got 'RuntimeError: call stack exhausted' (4:1)
```

nlewycky:
> However, LLD doesn't have this flag (it also may not perform relaxation at all, I don't know).

[...]

PR #2768, opened 2022-01-27
Fix serialize FrameInfo on Dylib engine (for #1727)

ptitSeb:
# Description

Add serialization of Frame Info (except when using llvm) for the Dylib engine, so Trap Information is still valid and the correct trap are triggered by the program.


Amanieu:
We have some tests in ignore.txt that are disabled due to #1727, can you try re-enabling them?

ptitSeb:
> We have some tests in ignore.txt that are disabled due to #1727, can you try re-enabling them?

Ah yes. I have remove the tests from the `ignore.txt`, they are working.

syrusakbary (approved):
Approving just in case (not sure if bors will work? I think it should, but I'm not sure!)

PR #2775, opened 2022-01-31
Support SSE 4.2 for the Singlepass compiler

### LINKED ISSUE #2767: Support SSE legacy instructions for the SInglepass compiler
### Motivation

I am trying to run a block chain node on my home server and the chain uses wasmer for its smart contracts. My server uses the legacy SSE 4.2 instructions so when trying to compile the wasm code using the singlepass compiler, I get an error because only the newer AVX instructions are supported.

We should support these legacy instructions as many machines out there still use them and new computers are still being sold with chips that use them. 

### Proposed solution

I looked at `emitter_x64.rs` and I see that the AVX instructions used mostly have an SSE counterpart. Non-destructive instructions were introduced in AVX. For SSE, to achieve the same behavior, requires two instructions: A move and then the destructive instruction. 

Some ops such as blendvps only have support on SSE 4.1, so it will have to mimicked for older SSE versions if those are to be supported.

batconjurer:
Closes #2767 

# Description
This PR adds support for the SSE 4.2 instruction set in the singlepass compiler. Previously, only AVX was supported. This implementation may also work for SSE 4.1, but this should be double checked. It does use the `blendvps` instruction indroduced in SSE 4.1, so it cannot support any earlier SSE versions at present for sure. 

The following are the main changes:
 - The assembler has changed from a Vec<X64Relocation> to a struct containing this as well as the corresponding CPU feature (indicating AVX/SSE4.2/...)
 - A set of SSE function macros have been added. To handle non-destructive operations, additional moves may be made before executing the corresponding SSE instruction if `src1 != dst`
 - A minor change to the `EmitterX64` trait. It now has a function for returning a possible CPU feature.
 - The implementation of `EmitterX64` for the assembler has obviously changed. It chooses the emitted assembly based on the CPU feature.

### DIFF
--- a/lib/compiler-singlepass/src/emitter_x64.rs
+++ b/lib/compiler-singlepass/src/emitter_x64.rs
@@ -2,13 +2,13 @@ use crate::common_decl::Size;
 use crate::location::Location as AbstractLocation;
 pub use crate::location::Multiplier;
 pub use crate::machine::{Label, Offset};
+use crate::machine_x64::AssemblerX64;
 pub use crate::x64_decl::{GPR, XMM};
 use dynasm::dynasm;
 use dynasmrt::{
-    x64::X64Relocation, AssemblyOffset, DynamicLabel, DynasmApi, DynasmLabelApi, VecAssembler,
+    AssemblyOffset, DynamicLabel, DynasmApi, DynasmLabelApi,
 };
-
-type Assembler = VecAssembler<X64Relocation>;
+use wasmer_compiler::CpuFeature;
 
 /// Force `dynasm!` to use the correct arch (x64) when cross-compiling.
 /// `dynasm!` proc-macro tries to auto-detect it by default by looking at the
@@ -17,7 +17,7 @@ type Assembler = VecAssembler<X64Relocation>;
 macro_rules! dynasm {
     ($a:expr ; $($tt:tt)*) => {
         dynasm::dynasm!(
-            $a
+            $a.inner
             ; .arch x64
             ; $($tt)*
         )
@@ -58,6 +58,7 @@ pub enum GPROrMemory {
 }
 
 pub trait EmitterX64 {
+    fn get_simd_arch(&self) -> Option<&CpuFeature>;
     fn get_label(&mut self) -> Label;
     fn get_offset(&self) -> Offset;
     fn get_jmp_instr_size(&self) -> u8;
@@ -71,7 +72,7 @@ pub trait EmitterX64 {
 
     fn emit_nop(&mut self);
 
-    /// A high-level assembler method. Emits an instruction sequence of length `n` that is functionally
+    /// A high-level inner method. Emits an instruction sequence of length `n` that is functionally
     /// equivalent to a `nop` instruction, without guarantee about the underlying implementation.
     fn emit_nop_n(&mut self, n: usize);
 
@@ -122,6 +123,7 @@ pub trait EmitterX64 {
     fn emit_vmovapd(&mut self, src: XMMOrMemory, dst: XMMOrMemory);
     fn emit_vxorps(&mut self, src1: XMM, src2: XMMOrMemory, dst: XMM);
     fn emit_vxorpd(&mut self, src1: XMM, src2: XMMOrMemory, dst: XMM);
+
 
     fn emit_vaddss(&mut self, src1: XMM, src2: XMMOrMemory, dst: XMM);
     fn emit_vaddsd(&mut self, src1: XMM, src2: XMMOrMemory, dst: XMM);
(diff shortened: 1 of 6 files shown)


lib/compiler-singlepass/src/emitter_x64.rs
```diff
@@ -471,154 +475,290 @@ macro_rules! jmp_op {
...
             }
         }
     }
 }
 
-impl EmitterX64 for Assembler {
+macro_rules! sse_round_fn {
+    ($ins:ident, $mode:expr) => {
+        |emitter: &mut AssemblerX64, precision: Precision, src1: XMM, src2: XMMOrMemory, dst: XMM| {
+            match src2 {
```
ptitSeb, line 743: `vroundss` and `vroundsd` are always used with src1 == src2.
Here the use of `vround` instead of simple `round` makes no sense, as xmm regs only carry a single Float or Double value.
So, the the `move_src_to_dst` is useless (maybe put an `assert!(src1==src2)` to be safe.

lib/compiler-singlepass/src/emitter_x64.rs
```diff
@@ -471,154 +475,290 @@ macro_rules! jmp_op {
...
+                XMMOrMemory::XMM(x) => {
+                    move_src_to_dst(emitter, precision, src1, dst);
+                    dynasm!(emitter ; $ins Rx((src1 as u8)), Rx((x as u8)), $mode)
+                }
```
ptitSeb, line 747: I think you want to reference `dst` and not `src1` here

lib/compiler-singlepass/src/emitter_x64.rs
```diff
@@ -471,154 +475,290 @@ macro_rules! jmp_op {
...
+                XMMOrMemory::Memory(base, disp) => {
+                    move_src_to_dst(emitter, precision, src1, dst);
+                    dynasm!(emitter ; $ins Rx((src1 as u8)), [Rq((base as u8)) + disp], $mode)
```
ptitSeb, line 750: same here (`dst` instead of src1`)

lib/compiler-singlepass/src/emitter_x64.rs
```diff
@@ -471,154 +475,290 @@ macro_rules! jmp_op {
...
+    ($ins:ident) => {
+        |emitter: &mut AssemblerX64, precision: Precision, src1: XMM, src2: GPROrMemory, dst: XMM| {
+            match src2 {
+                GPROrMemory::GPR(x) => {
+                    move_src_to_dst(emitter, precision, src1, dst);
+                    dynasm!(emitter ; $ins Rx((dst as u8)), Rq((x as u8)))
+                },
+                GPROrMemory::Memory(base, disp) => {
+                    move_src_to_dst(emitter, precision, src1, dst);
+                    dynasm!(emitter ; $ins Rx((src1 as u8)), QWORD [Rq((base as u8)) + disp])
```
ptitSeb, line 660: `dst` instead of `src1`?

lib/compiler-singlepass/src/emitter_x64.rs
```diff
@@ -471,154 +475,290 @@ macro_rules! jmp_op {
...
     }
 }
 
+macro_rules! sse_i2f_64_fn {
+    ($ins:ident) => {
+        |emitter: &mut AssemblerX64, precision: Precision, src1: XMM, src2: GPROrMemory, dst: XMM| {
+            match src2 {
+                GPROrMemory::GPR(x) => {
+                    move_src_to_dst(emitter, precision, src1, dst);
+                    dynasm!(emitter ; $ins Rx((dst as u8)), Rq((x as u8)))
```
ptitSeb, line 656: what if `dst` == `src2`?

lib/compiler-singlepass/src/emitter_x64.rs
```diff
@@ -471,154 +475,290 @@ macro_rules! jmp_op {
...
+    ($ins:ident, $mode:expr) => {
+        |emitter: &mut AssemblerX64, precision: Precision, src1: XMM, src2: XMMOrMemory, dst: XMM| {
+            match src2 {
+                XMMOrMemory::XMM(x) => {
+                    move_src_to_dst(emitter, precision, src1, dst);
+                    dynasm!(emitter ; $ins Rx((dst as u8)), Rx((x as u8)), $mode)
+                }
+                XMMOrMemory::Memory(base, disp) => {
+                    move_src_to_dst(emitter, precision, src1, dst);
+                    dynasm!(emitter ; $ins Rx((src1 as u8)), [Rq((base as u8)) + disp], $mode)
```
ptitSeb, line 598: `dst` instead of `src1`?

lib/compiler-singlepass/src/emitter_x64.rs
```diff
@@ -471,154 +475,290 @@ macro_rules! jmp_op {
...
     }
 }
 
+macro_rules! sse_fn {
+    ($ins:ident) => {
+        |emitter: &mut AssemblerX64, precision: Precision, src1: XMM, src2: XMMOrMemory, dst: XMM| {
+            match src2 {
+                XMMOrMemory::XMM(x) => {
+                    move_src_to_dst(emitter, precision, src1, dst);
+                    dynasm!(emitter ; $ins Rx((dst as u8)), Rx((x as u8)))
```
ptitSeb, line 580: what if `dst`== `src2`?

lib/compiler-singlepass/src/emitter_x64.rs
batconjurer: If we want the src1 register to remain undisturbed, this is not possible unless a scratch register is used. And it is not clear to me how to do that without an argument for the scratch register being passed in. Otherwise, we will have to clobber src1 in order to get the result in src2.

ptitSeb: I think is simple if would be enough to handle the case where dst and src2 are the same XMM register.
Something like
```RUST
let src = if dst != src2 {
  move_src_to_dst(emitter, precision, src1, dst);
  x
} else {
 src1
};
dynasm!(emitter ; $ins Rx((dst as u8)), Rq((src as u8)));
```
Should be enough?

batconjurer: This only works if `$ins` is commutative, which isn't always the case. Subtraction for instance

ptitSeb: right. But that still doesn't solve the case were src2==dst. maybe at least just add a `assert!(x!=dst);` than, to be safe.

syrusakbary:
bors try

ptitSeb:
Can you rebase or merge your branch with current source code? The issue with the windows test has been fixed on master.

ptitSeb:
sorry, code linter is complaining. you need to run `cargo fmt --all`.

Issue #2777, opened 2022-02-01
Singlepass v2 on ARM fails to compile


syrusakbary:
The program `main.wasm` which is a Go database (TiDB) is able to run on other compilers but when run with singlepass it outputs the following:

Download the file here: https://registry-cdn.wapm.io/contents/lucklove/tidb/0.1.6/main.wasm

```
$ wasmer run main.wasm --singlepass
thread '<unnamed>' panicked at 'singlepass can't emit LDR S64, Memory(X29, -104), Memory(X8, 0)', lib/compiler-singlepass/src/emitter_arm64.rs:382:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread '<unnamed>' panicked at 'singlepass can't emit LDR S64, Memory(X29, -72), Memory(X8, 0)', lib/compiler-singlepass/src/emitter_arm64.rs:382:18
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: ImpossibleRelocation(Dynamic(DynamicLabel(2)))', lib/compiler-singlepass/src/machine_arm64.rs:1891:35
```

ptitSeb:
You need to be patient, it may take a long time for the other compiler to build this one.

I suggest you read this blog entry: https://wasmer.io/posts/wasmer-2.2 at the end, there are some compile benchmark.

Zhangyx24:
Thanks a lot !

PR #2778, opened 2022-02-01
Fix Singlepass f32_load and f64_load function, and introduce emit_bond_label_far as direct bcond may not work all the time (for #2777)

ptitSeb:
# Description
Aarch64 singlepass fix for #2777 :
* Change implementation of f32_load and f64_load to be more flexible
* Instroduce the `emit_bcond_label_far` emitter to alow use of bcond on "farther" reloc. It does slow down a little bit (like less than 5% on less powerfull devices), but it's cecessary to be able to run larger wasm file. Also, beeing "singlepass", no real smart way alay to detect if a "far" version is needed or not (so "far" are used everywhere).


Amanieu:
I'm a bit concerned about the performance, but <5% seems acceptable for now.

bors r+

PR #2787, opened 2022-02-11
Fix singlepass codegen regression

### LINKED ISSUE #2782: OOB table access with singlepass backend
### Describe the bug

singlepass backend somehow emits instructions that can cause unexpected OOB in table access.
llvm and cranelift backend doesn't reproduce this issue. Also this issue has been introduced since 2.2.0-rc1

```console
wasmer 2.2.0-rc1 | rustc 1.56.1 (59eed8a2a 2021-11-01) | x86_64
```

### Steps to reproduce

[VarArgs.swift.wasm.zip](https://github.com/wasmerio/wasmer/files/8011496/VarArgs.swift.wasm.zip)

```console
$ wasmer run --singlepass VarArgs.swift.wasm
```

### Expected behavior

```console
$ wasmer run --singlepass VarArgs.swift.wasm
The answer to life and everything is 42, 42, -42, 3.141593
dig it: 0  0 -1  1 -2  2 -3  3 -4  4 -5  5 -6  6 -7  7 -8  8 -9  9 -10 10 -11 11
pointers: '0x12345670' '0x12345671' '0x12345672' '0x12345673' '0x12345674'
a 1.1 1 2.2 1 3.3 1 a
b 1 1.1 1 2.2 1 3.3 1 b
c 1 2 1 3 1 4444444444444444 1 c
d 1 1 2 1 3 1 4444444444444444 1 d
e 10 20 10 30 10 4040404040404040 10 e
f 10 10 20 10 30 10 4040404040404040 10 f
rdar-32547102: 0.0 1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0 10.0 11.0
a 1.1 1 2.2 1 4.5 1 1.1 a
b 1 1.1 1 2.2 1 4.5 1 1.1 b
done.
```

### Actual behavior

```console
$ wasmer run --singlepass VarArgs.swift.wasm
error: failed to run `VarArgs.swift.wasm`
│   1: RuntimeError: undefined element: out of bounds table access
╰─▶ 2: table_get_oob
```

### Additional context

This issue was found in SwiftWasm's CI: https://github.com/swiftwasm/swift/pull/4203

ptitSeb:
# Description
Issue was with the conversion Hash -> vec of the used GPR and SIMD registers before mass Push/Pop. Because a register might be allocated/deallocated in between, the order of the resulting vector might change.
The fix use a simple `sort` on the vector before push/pop elements, to be sure the order is deterministic.
The PR also contains a few ARM64 machine change, so the test can be run also on Aarch64 Linux & macOS (tested to be working too).

Fixes #2782

### DIFF
--- a/lib/compiler-singlepass/src/machine_x64.rs
+++ b/lib/compiler-singlepass/src/machine_x64.rs
@@ -1626,17 +1626,13 @@ impl Machine for MachineX86_64 {
         self.used_gprs.insert(gpr);
     }
 
-    fn push_used_gpr(&mut self) -> usize {
-        let mut used_gprs = self.get_used_gprs();
-        used_gprs.sort();
+    fn push_used_gpr(&mut self, used_gprs: &Vec<GPR>) -> usize {
         for r in used_gprs.iter() {
             self.assembler.emit_push(Size::S64, Location::GPR(*r));
         }
         used_gprs.len() * 8
     }
-    fn pop_used_gpr(&mut self) {
-        let mut used_gprs = self.get_used_gprs();
-        used_gprs.sort();
+    fn pop_used_gpr(&mut self, used_gprs: &Vec<GPR>) {
         for r in used_gprs.iter().rev() {
             self.assembler.emit_pop(Size::S64, Location::GPR(*r));
         }
@@ -1684,9 +1680,7 @@ impl Machine for MachineX86_64 {
         assert_eq!(self.used_simd.remove(&simd), true);
     }
 
-    fn push_used_simd(&mut self) -> usize {
-        let mut used_xmms = self.get_used_simd();
-        used_xmms.sort();
+    fn push_used_simd(&mut self, used_xmms: &Vec<XMM>) -> usize {
         self.adjust_stack((used_xmms.len() * 8) as u32);
 
         for (i, r) in used_xmms.iter().enumerate() {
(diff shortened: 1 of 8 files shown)


syrusakbary:
In some tests blockchains were having a lot of arguments in the functions.
In this case I think sorting might become expensive. Rather than that, can we just use a structure that allows for deterministic pop/push without sorting?

ptitSeb:
> In some tests blockchains were having a lot of arguments in the functions. In this case I think sorting might become expensive. Rather than that, can we just use a structure that allows for deterministic pop/push without sorting?

The vector only have a handfull of value: less than 10. I don't think it's changes much. But if there is a better structure, why not (but egain, it's a very small set)

syrusakbary:
Perhaps we can use IndexSet for determinism? (other ideas/crates that you know @Amanieu ?)

https://docs.rs/indexmap/1.6.1/indexmap/set/struct.IndexSet.html

lib/compiler-singlepass/src/machine_x64.rs
```diff
@@ -1626,17 +1626,13 @@ impl Machine for MachineX86_64 {
         self.used_gprs.insert(gpr);
     }
 
-    fn push_used_gpr(&mut self) -> usize {
-        let mut used_gprs = self.get_used_gprs();
-        used_gprs.sort();
+    fn push_used_gpr(&mut self, used_gprs: &Vec<GPR>) -> usize {
         for r in used_gprs.iter() {
```
Amanieu, line 1630: `HashSet` still has a randomized iteration order so this makes the codegen of singlepass non-deterministic (but still correct).

ptitSeb: True, but do we need to generate the same code on each run?

ptitSeb:
bors

ptitSeb:
bors try

lib/compiler-singlepass/src/machine_x64.rs
ptitSeb: Fair enough. I'll remove the hashmaps

syrusakbary:
bors try

Amanieu (approved):

PR #2788, opened 2022-02-11
(tests) Disabling some of the Tests that use Trap backtrace on Cranel…

ptitSeb:
# Description
 Disabling some of the Tests that use Trap backtrace on Cranelift+Windows, it's broken on github CI, until investigation and a proper fix


Amanieu:
Alternatively we can change `.github/workflows/test-sys.yaml` to use `windows-2019` instead of `windows-latest`.

ptitSeb:
> Alternatively we can change `.github/workflows/test-sys.yaml` to use `windows-2019` instead of `windows-latest`.

Well, taht's probably a better solution than yes. Can you do change, and will check on the other PR if it build correctly after that.

PR #2799, opened 2022-02-19
Add riscv64 support

Ast-x64:
This adds riscv64 support for `wasmer-vm`.
This requires `libc >= 0.2.118` so `Cargo.lock` is modified.


syrusakbary:
Thanks for the PR. I believe @tnayuki was working also on RISC-V support which had a bit broader implications on the engines and so on.

https://github.com/tnayuki/wasmer/pull/1

Ast-x64:
Thanks for your reply. I'll close this pr.

PR #2800, opened 2022-02-20
RISC-V support

tnayuki:
# Description

Add RISC-V support in compiler LLVM and universal & dylib engine.

# Review

### DIFF
--- a/lib/compiler-llvm/src/config.rs
+++ b/lib/compiler-llvm/src/config.rs
@@ -100,6 +100,13 @@ impl LLVM {
     }
 
     fn target_triple(&self, target: &Target) -> TargetTriple {
+        let architecture = if target.triple().architecture
+            == Architecture::Riscv64(target_lexicon::Riscv64Architecture::Riscv64gc)
+        {
+            target_lexicon::Architecture::Riscv64(target_lexicon::Riscv64Architecture::Riscv64)
+        } else {
+            target.triple().architecture
+        };
         // Hack: we're using is_pic to determine whether this is a native
         // build or not.
         let operating_system = if target.triple().operating_system
@@ -125,7 +132,7 @@ impl LLVM {
             target_lexicon::BinaryFormat::Elf
         };
         let triple = Triple {
-            architecture: target.triple().architecture,
+            architecture: architecture,
             vendor: target.triple().vendor.clone(),
             operating_system,
             environment: target.triple().environment,
@@ -158,6 +165,14 @@ impl LLVM {
                 info: true,
                 machine_code: true,
             }),
+            Architecture::Riscv64(_) => InkwellTarget::initialize_riscv(&InitializationConfig {
+                asm_parser: true,
+                asm_printer: true,
+                base: true,
+                disassembler: true,
+                info: true,
+                machine_code: true,
+            }),
             // Architecture::Arm(_) => InkwellTarget::initialize_arm(&InitializationConfig {
             //     asm_parser: true,
             //     asm_printer: true,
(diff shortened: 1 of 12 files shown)


lib/compiler-llvm/src/config.rs
```diff
@@ -179,16 +194,47 @@ impl LLVM {
...
+
+                *((my_target_machine.target_machine as *mut u8).offset(0x410) as *mut u64) = 5;
+                std::ptr::copy_nonoverlapping(
+                    "lp64d\0".as_ptr(),
+                    (my_target_machine.target_machine as *mut u8).offset(0x418),
+                    6,
+                );
+
+                std::mem::transmute(my_target_machine)
+            }
```
syrusakbary, line 234: This will need to be in llvm-sys or inkwell, right?

tnayuki: Yes, but LLVM C API to change ABI doesn't seem to exist, so it should also be added.

syrusakbary: Thanks for the context.
Do you think we can create an issue on the LLVM project and add an extra comment in this source with a reference to the URL that new LLVM issue?

ptitSeb (commented):
I have compiled and tested that on a StarFive board running Fedora, and I have some test that are failling. Do you have tests failling also on your side?

tnayuki:
@ptitSeb Sorry for late response
Which tests failed on your board?

> I have compiled and tested that on a StarFive board running Fedora, and I have some test that are failling. Do you have tests failling also on your side?

tnayuki:
@ptitSeb on my HiFive Unmatched board, 
deterministic::deterministic_empty, deterministic::deterministic_table, traps::test_trap_stack_overflow::llvm::universal and traps::test_trap_trace::llvm::universal were failed, but I have not properly investigated the cause.

By the way some tests that seem to use filesystem were ignored.

ptitSeb:
> @ptitSeb on my HiFive Unmatched board, deterministic::deterministic_empty, deterministic::deterministic_table, traps::test_trap_stack_overflow::llvm::universal and traps::test_trap_trace::llvm::universal were failed, but I have not properly investigated the cause.
> 
> By the way some tests that seem to use filesystem were ignored.

Yes, the traps one are also failing on my side. But I get a segfault on simd tests. The CPU doesn't have any SIMD extension, but I was assuming llvm should use non-simd path? Likewise, I haven't analysed yet.
I use a custom built LLVM-13.0.1, as the Fedora image of the VisionFive only have llvm-11 available.

ptitSeb:
Also, wasmer source have evolved lately, with some new mecanism around trap and stack handling. You should try to update your fork.
Unfortunatly, the stack switcher corosensei is not compatible with Risc-V for now, so this will need to be added too (@Amanieu ), or find a way to have a generic fallback mecanism for unsupported platforms.

syrusakbary:
@ptitSeb I think corosensei it is compatible with RISC-V?

ptitSeb:
> @ptitSeb I think corosensei it is compatible with RISC-V?

There is a `riscv.rs` source file yes, but it's not shown on the readme as a supported arch.

shaggygi:
[Question]:  I recently came across this and was just curious.  Does this PR mean there will be Wasmer support for the likes of ESP32 MCUs? Thanks for any info.

lib/compiler-llvm/src/config.rs
tnayuki: Sorry for the late reply as I have been busy with my day job😢 
I checked LLVM again, and I found that it seems by this change the correct ABI will be specified in LLVM 15, so I will try it.
https://github.com/llvm/llvm-project/commit/4e115b7d881136947c083e12f62010bc6b1d3f00

PR #2801, opened 2022-02-20
Fix the broken link for WASI ABI

ken-matsui:
# Description

The link for WASI ABI was moved, so I fixed it towards the correct link. According to the [`WASI/legacy`](https://github.com/WebAssembly/WASI/blob/main/legacy/README.md), the link will be broken again after 2023. Therefore, I would suggest either removing the section or using the specific SHA for the link like the following:
https://github.com/WebAssembly/WASI/blob/8a381fb9124b3aca29755ce8e255d2eeaddb16cb/legacy/application-abi.md

Please let me know if you would like to use the SHA-based link.

# Review


syrusakbary:
This looks good! Merging

Issue #2736, opened 2021-12-28
engine-dylib with singlepass compiler fails on Mac with ld: illegal text-relocation


tzemanovic:
Hello, I'm trying to use engine-dylib together with the singlepass compiler.

### Describe the bug

The simple wasm code in https://github.com/wasmerio/wasmer/blob/master/examples/engine_dylib.rs#L27 works fine, but when I tried to use some WASM code compiled from Rust instead, I noticed that this can in some cases fail in the linker with "ld: illegal text-relocation":

```
Shared object file generator failed with:
stderr:Apple clang version 12.0.0 (clang-1200.0.32.29)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
 \"/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ld\" -demangle -lto_library /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/libLTO.dylib -dynamic -dylib -arch x86_64 -platform_version macos 10.15.0 10.15.6 -syslibroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk -o /var/folders/8n/xzxk5c154qzg_chndqmzgfdr0000gp/T/wasmer_dylib_aOgbIS.dylib -L/usr/local/lib /var/folders/8n/xzxk5c154qzg_chndqmzgfdr0000gp/T/wasmer_dylib_07B9ou.o -undefined dynamic_lookup -lSystem /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/12.0.0/lib/darwin/libclang_rt.osx.a
ld: illegal text-relocation to '_wasmer_function__61' in /var/folders/8n/xzxk5c154qzg_chndqmzgfdr0000gp/T/wasmer_dylib_07B9ou.o from '_wasmer_function__0' in /var/folders/8n/xzxk5c154qzg_chndqmzgfdr0000gp/T/wasmer_dylib_07B9ou.o for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Concretely, it seems that the text-relocation is applied on code that uses function pointers.

### Steps to reproduce

[...]

PR #2802, opened 2022-02-23
Feat/singlepass dylib

ptitSeb:
# Description
Enable Dylib+singlepass on Linux and macOS. Works fine, but larger wasm file may still fail to link on macOS/Aarch64.
For #2736


lib/engine-dylib/src/artifact.rs
```diff
@@ -384,7 +384,7 @@ impl DylibArtifact {
             // We are explicit on the target when the host system is
             // Apple Silicon, otherwise compilation fails.
             if target_triple_str == "arm64-apple-darwin" {
-                vec![format!("--target={}", target_triple_str)]
+                vec![form at!("--target={}", target_triple_str)]
```
Amanieu, line 387: typo?

lib/engine-dylib/src/artifact.rs
```diff
@@ -402,6 +402,7 @@ impl DylibArtifact {
 
         let notext = match (target_triple.operating_system, target_triple.architecture) {
             (OperatingSystem::Linux, Architecture::X86_64) => vec!["-Wl,-z,notext"],
+            (OperatingSystem::MacOSX, Architecture::X86_64) => vec!["-Wl,-read_only_relocs,suppress"],
```
Amanieu, line 405: We actually want to remove both of these options. They currently tell the linker to allow text relocations which we *don't* want for PIC code.

syrusakbary:
I don't have enough context on this PR to approve myself. So we'll wait for @Amanieu review!

Amanieu:
I tested this branch and it seems to generate a .so file with zero runtime relocations, which is good. However I don't see why the `no-text` flag is still needed. Additionally, this PR has a lot of unnecessary changes like:
- `emit_ldr_label` and `emit_nop` that are never used.
- Leftover change to traphandlers.c
- It still includes the linker flags that we don't want. This should never trigger if we are indeed emitting code with zero relocations.

ptitSeb:
> * It still includes the linker flags that we don't want. This should never trigger if we are indeed emitting code with zero relocations.

The cross compile test is failling. It compile using dylib-cranelift targetting x86_64-linux-musl and fail with a relocation of type `R_X86_64_64` for `wasmer_vm_probestack` in text. It only happens for musl. It seems to fails if I actually cross compile: if I cange the example to build to x86_64-linux-gnu on an x86_64 linux box, it works, but the same change on an arm64 box the sample fails. It seems to only happens when `lld` is used. `ld`/`gold` don't have the issue.

Amanieu:
Can you dump the object files that are passed to the linker in both the success and failing cases? Also the arguments passed to the linker. The `wasmer_vm_probestack` symbol should be defined in the same object file, in a separate section that is added by engine-dylib.

ptitSeb:
Yes, I was thinking on the same thing, check the difference, if any, on both object files.

Amanieu:
Actually the `R_X86_64_64` relocation seems strange. That's an absolute 64-bit relocation. I would have expected a `R_X86_64_PC32` instead which is used for the offset in a call instruction.

I think that I know what the problem is, can you try applying this commit from #2807: https://github.com/wasmerio/wasmer/pull/2807/commits/657bbae518a649ecc98b06571a78f30fdb1fe023

You might also need to apply this fix: https://github.com/wasmerio/wasmer/commit/a2809145a1af7efa6cffc565f2144475e8af5305

ptitSeb:
Yes, it works with both commit.

ptitSeb:
So I'll wait until the Stack/Corosensei branch is merged...

lib/object/src/module.rs
```diff
@@ -309,6 +311,62 @@ pub fn emit_compilation(
                     RelocationEncoding::Generic,
                     32,
                 ),
+                Reloc::Arm64Movw0 => (
```
Amanieu, line 314: Are these still needed?

Amanieu (approved):

lib/object/src/module.rs
ptitSeb: Probably not

Amanieu: Remove them then it's good to r+.

Amanieu (approved):

PR #2807, opened 2022-02-25
Run Wasm code on a separate stack

### LINKED ISSUE #2562: macOS function calls can be optimized
Right now we are using `sigsetjmp` when calling functions from host to wasm, which adds significant overhead.
We can either use `setjmp` or macho exception handling directly.

Performance from `sigsetjmp` to `setjmp` reported here: https://github.com/wasmerio/wasmer/pull/2102 (from 140.14ns to 29.684ns in some examples).

### LINKED ISSUE #2757: Improve stack overflow handling
Wasmer's handling of stack overflows is rather poor: it assumes that any segfault that isn't tied to a Wasm memory access is a potential stack overflow and will `longjmp` back to the last host->wasm call. However this can potentially lead to deadlocks or memory corruption if the stack overflow happens in an unexpected place, such as the middle of `malloc`.

The solution is to not use the host stack when running Wasm code. If Wasm code executes on a separate stack then we can safely catch segfaults on the guard page of this stack and reset back to the last point where the secondary stack was entered. Any host code with non-trivial state or locking (e.g. `malloc`) must switch back to the host stack first for the duration of that call.

Amanieu:
This uses the [corosensei](https://crates.io/crates/corosensei) crate to
run Wasm code on a separate stack from the main thread stack.

In trap handlers for stack overflows and memory out of bounds accesses,
we can now check whether we are executing on the Wasm stack and reset
execution back to the main thread stack when returning from the trap
handler.

When Wasm code needs to perform an operation which may modify internal
data structures (e.g. growing a memory) then execution must switch back
to the main thread stack using on_host_stack. This is necessary to avoid
leaving internal data structure in an inconsistent state when a stack
overflow happens.

In the future, this can also be used to suspend execution of a Wasm
module (#1127) by modeling it as an async function call.

Fixes #2757
Fixes #2562


lib/vm/src/trap/handlers.c
```diff
@@ -1,66 +0,0 @@
...
-  }
-  *buf_storage = &buf;
-  body(payload);
-  return 1;
-}
-
-void wasmer_unwind(void *JmpBuf) {
-  platform_jmp_buf *buf = (platform_jmp_buf*) JmpBuf;
-  platform_longjmp(*buf, 1);
-}
```
syrusakbary, line 66: Nice work removing the dependency on this c file!

syrusakbary (commented):
Should we eliminate `vm/probestack.rs` along with the `PROBESTACK` const, `wasmer_vm_probestack`, and `LibCall::Probestack` (in `vm/libcalls.rs`) ?

They seem to not be used any longer (although I'm not really sure)

Amanieu:
> Should we eliminate `vm/probestack.rs` along with the `PROBESTACK` const, `wasmer_vm_probestack`, and `LibCall::Probestack` (in `vm/libcalls.rs`) ?
> 
> They seem to not be used any longer (although I'm not really sure)

Cranelift still uses it for stack probing.

lib/vm/src/trap/handlers.c
ptitSeb: That C file had some hack to allow correct debugging on mac. It would be nice if that hack culd be put somewhere else, as a comment or something optionnal?

Amanieu: You can use an LLDB command to mask signals:
```
process handle SIGBUS --notify false --stop false
```

ptitSeb: no, that's not the issue. It's not about signal trapping.
that ine of code
`task_set_exception_ports(mach_task_self(), EXC_MASK_BAD_ACCESS, MACH_PORT_NULL, EXCEPTION_DEFAULT, 0);`
allow EXC_BAD_ACCESS exception to be routed and trigger segfault. Without it, you get stuck at the exception and can debug anything.

Amanieu: Is there a reason why this line is hidden behind an `#if 0` by default? It seems that [mono](https://github.com/mono/mono/pull/20321/files) always does this.

ptitSeb: No specific reason, I just let it disabled unless activated for debug to avoid changing previous behaviour. But I suppose it can just be enabled.

tests/ignores.txt
```diff
@@ -39,13 +39,15 @@ singlepass multi_value_imports::dynamic # Singlepass doesn't support multivalue
...
-dylib             spec::skip_stack_guard_page # Missing trap information in dylibs
+# Also neither LLVM nor Cranelift currently implement stack probing on AArch64.
+# https://github.com/wasmerio/wasmer/issues/2808
+cranelift+aarch64 spec::skip_stack_guard_page
+llvm+aarch64      spec::skip_stack_guard_page
+# Stack probing doesn't seem to be working correctly on Windows with Cranelift
+cranelift+windows spec::skip_stack_guard_page
+# TODO: Needs more investigation
+cranelift+macos
+llvm+macos
```
syrusakbary, line 50: We need to make sure we un-do this before merging the PR.
Perhaps you mean to skip a specific test @Amanieu ?

Amanieu: Oops, my mistake.

ptitSeb (approved):
we need to remember to add some workaround for the macOS `task_set_exception_ports(mach_task_self(), EXC_MASK_BAD_ACCESS, MACH_PORT_NULL, EXCEPTION_DEFAULT, 0);` debuggin issue...

syrusakbary:
Some context on why we did cancel:

> We still need to add the macOS debugging workaround and publish a new version of corosensei.

Once that's addressed we should be good to merge

PR #2812, opened 2022-03-03
change host_state drop order for InstanceHandle

YunSuk-Yeo:
# Description

This PR is following https://github.com/wasmerio/wasmer/pull/2806
We still met `prev.start > max` assertion error even after we pathed #2806. 
#2086 seem fixed drop order problem on `Module`, but not solved it when we use `Instance`.

I found some other place which can cause drop order problem when we use `Instance`.
After we patch this, I couldn't see more `prev.start > max` assertion.

This instance handle will be stay in https://github.com/wasmerio/wasmer/blob/7fbaf6fc0563694d56d51106589a5b31f6c83582/lib/vm/src/instance/mod.rs#L100-L101 of `Instance`.

# Review


syrusakbary:
I think this fix makes sense for me. @Amanieu could you review it as well?

lib/api/src/sys/module.rs
```diff
@@ -268,7 +268,7 @@ impl Module {
             let instance_handle = self.artifact.instantiate(
                 self.store.tunables(),
                 resolver,
-                Box::new((self.store.clone(), self.artifact.clone())),
+                Box::new((self.artifact.clone(), self.store.clone())),
```
syrusakbary, line 271: Actually it might make more sense to just replace this with:

```suggestion
                Box::new((self.clone(),))
```

That way we move the drop responsibility to the module itself. Can you try if this also fix your issue @YunSuk-Yeo ?

YunSuk-Yeo: okay, let me try.

YunSuk-Yeo: This one is working, so updated PR.

Issue #2817, opened 2022-03-11
Recommended pattern to setup workspace have multiple binary depends on different feature sets


kwonoj:
### Summary

Hi, this time's question is regarding organizing workspaces. 

Here's simplified example https://github.com/kwonoj/wasmer-js-test/tree/workspace-features-resolution

```
//workspace
 - A (compiled to wasm)
   - runner_base
 - B (native binary)
   - runner_base
 - runner_base
   - [features = "js"]
      - wasmer/js
   - [features = "native"]
      - wasmer/sys
```

A and B are orthogonal, no direct overlaps except sharing some of common dependencies. Common dependency opt to in to specific feature of wasmer depends on build target. This makes build failures by cargo, with errors like 

```
    |
233 |                 crate::Function::get_self_from_extern(_extern)?.native().map_err(|_| crate::sys::exports::ExportError::IncompatibleType)
    |                        ^^^^^^^^ ambiguous name
    |
    = note: ambiguous because of multiple glob imports of a name in the same module
note: `Function` could refer to the struct imported here
   --> /home/ojkwon/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-2.2.0/src/lib.rs:467:9
    |
467 | pub use sys::*;
    |         ^^^^^^
    = help: consider adding an explicit import of `Function` to disambiguate
note: `Function` could also refer to the struct imported here
   --> /home/ojkwon/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-2.2.0/src/lib.rs:473:9
```

Which looks like cargo try to resolve wasmer for every feature for each build target somewhat sound similar to https://github.com/rust-lang/cargo/issues/4463, but cannot say for sure if this is the exact case or wasmer have some internal behavior raises this.

Is there a recommended pattern to workaround this in easy way?

Thanks.

### Additional details

Provide any additional details here.

syrusakbary:
Hey @kwonoj, I'm unable to access the repo posted: https://github.com/kwonoj/wasmer-js-test/tree/workspace-features-resolution

So I'll reply a bit more blindly.
I think the issue might be that `runner_base` is depending on `wasmer` with the default features activated.
Perhaps using wasmer without any default feature and then the "native" feature will use `wasmer/sys` and "js" `wasmer/js` should just work?

It might be also doable to do it via target features.

To put an example in our repo, the wasmer-wasi integration can target both the native system and the JS (just depending on its features) and it works in the desired way when used as a dependency.

[...]

PR #2821, opened 2022-03-17
build(wasmer/cache): opt 'sys' into default feature

kwonoj:
# Description
Context: this is originated from https://github.com/wasmerio/wasmer/issues/2817#issuecomment-1064752540 , where SWC (https://github.com/swc-project/swc) trying to setup its package dependencies orthogonally requires two distinct features of wasmer (js / sys), encounters feature selection getting merged by cargo. We have workaround it by manual opt in into each features, but the common dependency we have imports wasmer-cache, which ends up to enabling `sys` mandatory. It is also possible to do some workaround by having a feature with make dep as optional, but I feel it'd be better if `cache` allows to opt out from feature.

I am honestly not sure if this is acceptable or recommended approach to resolve issue. Please feel free to close if that's the case. 

# Review


syrusakbary:
I thought about this further, and compiling `wasmer-cache`  without the default features (that is without `wasmer/sys`) will make Rust compilation fail, as it depends on serialization/deserialization being present in the `wasmer` crate, however those methods are not present in `wasmer/js`.

To make this work when the default options are not present, the solution is simple: implement serialization/deserialization in `wasmer/js`, so `wasmer-cache` can be used also from JS. How?
* Each time a new module is being created with `Module::new`, we store the `[u8]` bytes inside the an attribute in the Module struct
* `Module::deserialize` (returning a `wasmer::Module`) takes as input an array of bytes and just calls `Module::new` on them.
* `Module::serialize` just returns the field on the struct with the module bytes contents
* Copy the `SerializeError` and `DeserializeError` to `wasmer/js`: https://github.com/wasmerio/wasmer/blob/master/lib/engine/src/error.rs#L10-L40

We can do that as part of the PR and we should be good to merge! :)

kwonoj:
Thanks for the guide, I'll peek if I can spin some time to shape PR to correctly include those.

kwonoj:
I tried to add a few paths as much to follow suggestions, open to reflect fixes / suggestions.

lib/cache/Cargo.toml
```diff
@@ -25,6 +25,9 @@ wasmer-engine-universal = { path = "../engine-universal", version = "=2.2.1" }
 wasmer-engine-dylib = { path = "../engine-dylib", version = "=2.2.1" }
 
 [features]
+default = ["sys"]
+sys = ["wasmer/sys-default"]
+js = ["wasmer/js-default", "wasmer/js-serializable-module"]
```
syrusakbary, line 30: I don't think we should enforce sys or js by default. We can simplify this such as:

```suggestion
default = ["wasmer/js-serializable-module"]
```

This should work without adding any feature dependency on wasmer (that is, that the wasmer crate will not automatically use `js` or `sys`, it will just use the `wasmer/js-serializable-module` feature when using js but do nothing on `sys`, which is good.

lib/cache/src/filesystem.rs
```diff
@@ -1,3 +1,4 @@
+#![cfg_attr(not(feature = "sys"), allow(unused))]
```
syrusakbary, line 1: Rather than this requiring the `sys` feature, can we create a new feature called `filesystem` that is enabled by default?

Why this? Because some targets like wasm32-wasi (which are not `sys`) could still make use of this without depending on sys.

syrusakbary:
Thanks @kwonoj, I added some feedback!

lib/cache/Cargo.toml
kwonoj: I was curious about this

> wasmer crate will not automatically use js or sys

Wouldn't it make existing build as breaking, that upper deps having wasmer-cache or transitive deps need to manually opt in `sys` if they're using default? Or it'll be fine since any pkg having `wasmer` already may opt in js / sys?

kwonoj:
Thanks for the help, updated as suggested.

lib/cache/Cargo.toml
```diff
@@ -25,6 +25,8 @@ wasmer-engine-universal = { path = "../engine-universal", version = "=2.2.1" }
 wasmer-engine-dylib = { path = "../engine-dylib", version = "=2.2.1" }
 
 [features]
+default = ["wasmer/js-serializable-module"]
```
syrusakbary, line 28: We should make filesystem also the default to make sure we don't incur in breaking changes for people using the wasmer-cache crate.

```suggestion
default = ["wasmer/js-serializable-module", "filesystem"]
```

syrusakbary (approved):

lib/cache/Cargo.toml
syrusakbary: Oh, I think I was not super clear. I think the `filesystem` should be enabled by default (that way there are no breaking changes)
However, we should not enforce sys or js by default (wasmer-cache should remain agnostic).
In any case, I did applied a small suggestion and the PR looks good!

syrusakbary:
Linting is failing, but once is solved we should be good to merge.
You can lint locally with: `cargo fmt --all`

kwonoj:
Thanks, applied fmt.

PR #2828, opened 2022-03-27
Fix JsImportObject resolver

kateinoigakukun:
# Description

The `JsImportObject.get_export` implementation is apparently broken. The second argument of `WASI.instantiate` of `@wasmer/wasi` doesn't work at all due to this issue.

# Review


syrusakbary (approved):
Thanks for the fix!

PR #2829, opened 2022-03-27
Improve error message oriented from JS object

kateinoigakukun:
# Description

`JsValue.as_string` converts the value to String iff the value is a
String itself, so the error message always failovers to empty string.
`JsValue` impls `Debug` using `JSON.stringify`, and it gives better
description of the error.

# Review


syrusakbary (approved):
Nice fix!

syrusakbary:
bors r +

PR #2834, opened 2022-04-03
Update cranelift to 0.82

Amanieu:
This adds SIMD support on the Cranelift backend.


syrusakbary:
Tests seem to be failing

Amanieu:
The failure seems to be due to the wasmparser upgrade, I'm looking into it.

syrusakbary (approved):
We should be good to merge once the tests pass

Amanieu:
It seems that the wasmparser upgrade may cause a breaking change in the C API. Specifically in this file: https://github.com/wasmerio/wasmer/blob/master/lib/c-api/src/wasm_c_api/unstable/parser/operator.rs

Is this an acceptable breaking change? The API is marked as unstable.

Amanieu:
Nevermind I managed to not make it a breaking change.

syrusakbary:
Tests passed, so I'm merging directly as I already reviewed it

PR #2835, opened 2022-04-04
Feat/wasmer.wasm

ptitSeb:
# Description
Allow WASMER to be built as WASM, with compile ability (only singlepass compiler and universal engine for now)


syrusakbary:
Side question: what is the diff between api/sys and api/wasm ?  (in code, what are the things unused? Both look pretty similar to me but I might be missing something!)

ptitSeb:
> Side question: what is the diff between api/sys and api/wasm ? (in code, what are the things unused? Both look pretty similar to me but I might be missing something!)

Many "runtime"' only parts are just empty, like `native.rs`, but also many files from externals, like globals. It's actualy a mix between `js` and `sys`, but with some parts removed and left as "unimplemented".

syrusakbary:
Cranelift is now updated in master. Can you update the PR and also see if we can get cranelift working?

Amanieu:
I'm concerned about the amount of code duplication in this PR. It introduces a whole new copy of the `vm` and `api` crates which are full of unimplemented stubs. I would prefer a cleaner structure where `wasmer.wasm` is a separate tool that only depends on the compiler crates without depending on the rest of the engine.

ptitSeb:
> I'm concerned about the amount of code duplication in this PR. It introduces a whole new copy of the `vm` and `api` crates which are full of unimplemented stubs. I would prefer a cleaner structure where `wasmer.wasm` is a separate tool that only depends on the compiler crates without depending on the rest of the engine.

Note that they are unimplemented because they will not work in a wasm context (not because I'm lazy).

syrusakbary:
Closing this PR as it was done and merged in #2851

PR #2836, opened 2022-04-04
Use Undefined instruction with payload to carry over TrapInformation on Singlepass

ptitSeb:
# Description
Use Undefined instruction with payload to carry over TrapInformation on without using an vector or hashmap on Singlepass.
This allow for slightly more compact binary and slightly faster trap decoding on runtime.


lib/vm/src/trap/traphandlers.rs
```diff
@@ -753,13 +818,17 @@ impl<T> TrapHandlerContextInner<T> {
...
+            trap_code
+        } else {
+            maybe_fault_address.map(|addr| {
+                if self.coro_trap_handler.stack_ptr_in_bounds(addr) {
+                    TrapCode::StackOverflow
+                } else {
+                    TrapCode::HeapAccessOutOfBounds
+                }
+            })
+        };
```
syrusakbary, line 831: This will simplify code a bit:

```suggestion
        let signal_trap = trap_code.unwrap_or_else(|| {
            maybe_fault_address.map(|addr| {
                if self.coro_trap_handler.stack_ptr_in_bounds(addr) {
                    TrapCode::StackOverflow
                } else {
                    TrapCode::HeapAccessOutOfBounds
                }
            })
        });
```

lib/vm/src/trap/traphandlers.rs
```diff
@@ -153,13 +154,60 @@ cfg_if::cfg_if! {
...
+                    }
+                    if cfg!(target_arch = "aarch64") {
+                        let addr = (*siginfo).si_addr() as usize;
+                        val = if read_unaligned(addr as *mut u32)&0xffff0000 == 0 {
+                            read_unaligned(addr as *mut u8)
+                        } else {
+                            0
+                        }
+                    }
+                    if val&0xc0 == 0xc0 {
```
syrusakbary, line 182: Can we put this into a static const? (it will make easier to understand what `0xc0` is)

lib/vm/src/trap/traphandlers.rs
ptitSeb: Oh, sure, i'll do that

lib/vm/src/trap/traphandlers.rs
ptitSeb: 0xc0 is a magic const, but I can do that yes, no problem

lib/vm/src/trap/traphandlers.rs
ptitSeb: Ah, in fact no, I don't want to unwrap yet.

syrusakbary: > Ah, in fact no, I don't want to unwrap yet.

Not sure if I understood. I think the suggestion is syntactically the same, just easier to read? Or is there anything I missed?

ptitSeb: `signal_trap`, like `trap_code` is an `Option<TrapCode>`, not a `TrapCode` as `unwrap_...` will produce.

syrusakbary: Oh, got it... then I think the operation should be `or_else`. Let me know if that works.

```suggestion
        let signal_trap = trap_code.or_else(|| {
            maybe_fault_address.map(|addr| {
                if self.coro_trap_handler.stack_ptr_in_bounds(addr) {
                    TrapCode::StackOverflow
                } else {
                    TrapCode::HeapAccessOutOfBounds
                }
            })
        });
```

ptitSeb: Ah, yes, that works :)

syrusakbary:
Tests seem to be failing

lib/vm/src/trap/traphandlers.rs
```diff
@@ -153,13 +159,60 @@ cfg_if::cfg_if! {
...
+                    }
+                    if cfg!(target_arch = "aarch64") {
+                        let addr = (*siginfo).si_addr() as usize;
+                        val = if read_unaligned(addr as *mut u32)&0xffff0000 == 0 {
+                            read_unaligned(addr as *mut u8)
+                        } else {
+                            0
+                        }
+                    }
+                    if val&MAGIC == MAGIC {
```
Amanieu, line 187: And use the `.and_then` method on `Option` here.

Amanieu (commented):
You can probably also remove `TrapCode::StackOverflow` and `TrapCode::HeapAccessOutOfBounds` from singlepass since those trap codes are now detected in the signal handler.

lib/vm/src/trap/traphandlers.rs
```diff
@@ -153,13 +159,60 @@ cfg_if::cfg_if! {
...
                 _ => None,
             };
-
+            let trap_code = match signum {
+                // check if it was cased by a UD and if the Trap info is a payload to it
+                libc::SIGILL => {
+                    let mut val: u8 = 0;
+                    if cfg!(target_arch = "x86_64") {
+                        let addr = (*siginfo).si_addr() as usize;
+                        val = if read_unaligned(addr as *mut u8)&0xf0 == 0x40
```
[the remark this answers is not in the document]
ptitSeb: ok

lib/vm/src/trap/traphandlers.rs
```diff
@@ -153,13 +159,60 @@ cfg_if::cfg_if! {
                 }
                 _ => None,
             };
-
+            let trap_code = match signum {
+                // check if it was cased by a UD and if the Trap info is a payload to it
+                libc::SIGILL => {
+                    let mut val: u8 = 0;
```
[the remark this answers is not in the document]
ptitSeb: done

lib/vm/src/trap/traphandlers.rs
ptitSeb: and done

lib/vm/src/trap/traphandlers.rs
```diff
@@ -587,8 +686,14 @@ thread_local! {
...
 struct TrapHandlerContext {
     inner: *const u8,
-    handle_trap:
-        fn(*const u8, usize, usize, Option<usize>, &mut dyn FnMut(TrapHandlerRegs)) -> bool,
+    handle_trap: fn(
+        *const u8,
+        usize,
+        usize,
+        Option<usize>,
+        Option<TrapCode>,
```
Amanieu, line 694: To avoid code duplication, could you pass a `maybe_invalid_inst_address: Option<usize>` to the common trap handling code and do the instruction decoding there?

Amanieu (commented):
Apart from that, LGTM.

lib/vm/src/trap/traphandlers.rs
ptitSeb: I don't like the idea of have an unprocess address going through all those call. I can simply create a local function to process the illegal intruction and extract the trap information to avoid code duplication, and keep the current mecanism with the `Option<TrapCode>` that for me makes more sense.
What do you think @Amanieu ?

Amanieu: Sure, that is fine too.

Amanieu:
bors retry

ptitSeb:
Those TimeOut seems to be more and more frequent now. There is an issue with the CI.

PR #2846, opened 2022-04-14
Replace ImportObject with new type Imports

epilys:
TODO:

- [x] Update all references to `ImportObject` in documentation
- [x] Update migration document with interface changes


lib/api/src/sys/module.rs
```diff
@@ -276,7 +276,7 @@ impl Module {
 
     pub(crate) fn instantiate(
         &self,
-        resolver: &dyn Resolver,
+        resolver: &crate::Imports,
```
epilys, line 279: This should be a value, not a ref.

tests/compilers/issues.rs
```diff
@@ -214,9 +214,9 @@ fn call_with_static_data_pointers(mut config: crate::Config) -> Result<()> {
         "gas",
         Function::new_native_with_env(&store, env.clone(), gas),
     );
-    let mut imports = ImportObject::new();
-    imports.register("env", exports);
-    let instance = Instance::new(&module, &imports)?;
+    let mut imports = Imports::new();
+    imports.import_namespace("env", exports.get_namespace_externs());
```
syrusakbary, line 218: Can we go back into the `register` nomenclature? (`register_namespace`).
Also, if we could have an `Into` from Exports on that function, it would make the following possible:

```rust
    imports.register_namespace("env", exports);
```

syrusakbary:
Can we create a document named: `Migration to 3.0` (similar to https://github.com/wasmerio/wasmer/blob/master/docs/migration_to_1.0.0.md) where we start describing the breaking changes and how people can upgrade to 3.0?

It will also make reviews on the API breaking changes for each PR a bit simpler :)

tests/compilers/issues.rs
epilys: Will do!

lib/api/src/sys/imports.rs
```diff
@@ -0,0 +1,343 @@
...
+            .collect();
+        if ret.is_empty() {
+            None
+        } else {
+            Some(ret)
+        }
+    }
+
+    /// Returns a list of all externs defined in all namespaces.
+    pub fn externs_vec(&self) -> Vec<(String, String, Extern)> {
```
Amanieu, line 100: I'd remove this method. If you need access to the contents of an `Imports` then you should use the iterator interface.

lib/api/src/sys/instance.rs
```diff
@@ -118,12 +113,52 @@ impl Instance {
     /// Those are, as defined by the spec:
     ///  * Link errors that happen when plugging the imports into the instance
     ///  * Runtime errors that happen when running the module `start` function.
-    pub fn new(
-        module: &Module,
-        resolver: &(dyn Resolver + Send + Sync),
-    ) -> Result<Self, InstantiationError> {
+    pub fn new(module: &Module, imports: Imports) -> Result<Self, InstantiationError> {
```
Amanieu, line 116: In hindsight we probably shouldn't take `Imports` by value and take it by reference like it was before instead. There's no benefit to taking it by value and it could be reused to make more instances.

syrusakbary: I think we needed to be owned for the JS use case?

epilys: Maybe we could encourage `Clone`ing in this case?

Amanieu: I can't remember why we decided it would be needed for JS. I don't think the reasoning applies any more, but I could be wrong.

Amanieu: Please change this back to taking `Imports` by reference.

docs/migration_to_3.0.0.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+You can also build the `Imports` object manually:
+
+```rust
+let mut import_object: Imports = Imports::new();
+import_object.define("env", "host_function", host_function);
+let instance = Instance::new(&module, &import_object).expect("Could not instantiate module.");
+```
+
```
syrusakbary, line 63: Before we used to have the `ChainLeftNamedResolver` (or similar, forgot the exact name). We should probably put an example on how to merge two import objects (left to right and right to left)

epilys: We hadn't talked about that, only that `*Resolver` traits are removed, that's why this wasn't included. This can implemented by using `IntoIterator`:

```rust
let imports1: Imports = todo!();
let mut imports2: Imports = todo!();

for (ns, name, ext) in imports1.into_iter() {
    imports2.define(&ns &name, ext);
}
```

Shall I add this into the documentation?

We can also add `chain_back` and `chain_front` methods for ease, or even implement [`std::ops::Add`](https://doc.rust-lang.org/stable/std/ops/trait.Add.html) and [`std::ops::AddAssign`](https://doc.rust-lang.org/stable/std/ops/trait.AddAssign.html):

```rust
let imports1: Imports = todo!();
let mut imports2: Imports = todo!();

imports2 += imports1;
```

But I don't know if this functionality is worth the extra API.

Amanieu: You could implement the `Extend` trait, which would allow you to call `imports.extend(other_imports)` since `Imports` implements `IntoIterator`.

lib/api/src/js/exports.rs
```diff
@@ -193,6 +192,32 @@ impl Exports {
...
+        self.map
+            .iter()
+            .map(|(k, v)| (k.clone(), v.to_export()))
+            .collect()
+    }
+
+    /// safa
+    pub fn as_exports(&self) -> Option<Exports> {
+        Some(self.clone())
+    }
```
Amanieu, line 220: Can you remove these from the JS API as well?

lib/api/src/js/imports.rs
```diff
@@ -2,173 +2,207 @@
...
-        let mut out = VecDeque::new();
-        let guard = self.map.lock().unwrap();
-        let map = guard.borrow();
-        for (name, ns) in map.iter() {
-            for (id, exp) in ns.get_namespace_exports() {
-                out.push_back(((name.clone(), id), exp));
+    /// Resolve and return a vector of imports in the order they are defined in the `module`'s source code.
+    ///
+    /// This means the returned `Vec<Export>` might be a subset of the imports contained in `self`.
+    pub fn imports_for_module(&self, module: &Module) -> Result<Vec<Export>, InstantiationError> {
```
Amanieu, line 132: We should be returning a `Vec<Extern>` here. Again, `Export` should be removed from the public API.

lib/api/src/js/imports.rs
```diff
@@ -2,173 +2,207 @@
...
+    /// let mut import_object = Imports::new();
+    /// import_object.get_export("ns", "name");
     /// ```
-    pub fn get_export(&self, module: &str, name: &str) -> Option<Export> {
-        let guard = self.map.lock().unwrap();
-        let map_ref = guard.borrow();
-        if map_ref.contains_key(module) {
-            let namespace = map_ref[module].as_ref();
-            return namespace.get_namespace_export(name);
+    pub fn get_export(&self, ns: &str, name: &str) -> Option<Export> {
```
Amanieu, line 58: `Export` should be removed from the public API. We should only give `Extern`.

lib/api/src/sys/imports.rs
```diff
@@ -1,184 +1,172 @@
...
     /// # Usage
-    /// ```ignore
-    /// # use wasmer_vm::{ImportObject, Instance, Namespace};
-    /// let mut import_object = ImportObject::new();
+    /// ```no_run
+    /// # use wasmer::Imports;
+    /// let mut import_object = Imports::new();
     /// import_object.get_export("module", "name");
     /// ```
     pub fn get_export(&self, module: &str, name: &str) -> Option<Export> {
```
Amanieu, line 55: Same here, change to return an `Extern` instead.

lib/api/src/sys/imports.rs
```diff
@@ -1,184 +1,172 @@
...
-        let mut out = VecDeque::new();
-        let guard = self.map.lock().unwrap();
-        let map = guard.borrow();
-        for (name, ns) in map.iter() {
-            for (id, exp) in ns.get_namespace_exports() {
-                out.push_back(((name.clone(), id), exp));
+    /// Resolve and return a vector of imports in the order they are defined in the `module`'s source code.
+    ///
+    /// This means the returned `Vec<Export>` might be a subset of the imports contained in `self`.
+    pub fn imports_for_module(&self, module: &Module) -> Result<Vec<Export>, LinkError> {
```
Amanieu, line 132: Same here, return `Vec<Extern>` instead.

Issue #2744, opened 2022-01-05
Make loupe dependency optional


syrusakbary:
We should make loupe dependency optional based on a feature, following up from this comment: https://github.com/wasmerio/wasmer/issues/2735#issuecomment-1005277369

Issue #2731, opened 2021-12-21
Use of supertrait bounds makes it tedious to implement some traits


nagisa:
### Describe the bug

This issue is not limited to, but perhaps best exemplified by the `Tunables` trait. This trait seems like something that one might want to implement (there's even an example to do so!), but in order to implement this trait it is also necessary to implement the external trait `loupe::MemoryUsage`. Since `loupe::MemoryUsage` itself is not re-exported by `wasmer`, implementing `Tunables` suddenly requires importing an entirely unrelated third party crate.

If the user does not utilize the `MemoryUsage` functionality themselves, then the motivation for implementing this trait disappears entirely (other than the fact that it gatekeeps the developer from implementing the trait of interest itself).

### Steps to reproduce

Try to implement `Tunables` for your own type.

### Expected behavior

`Tunables` and other similar traits should be implementable without requiring the developer to implement traits/functionality irrelevant to them.

I think one of the straightforward options here would be to have a 

```rust
impl MemoryUsage for dyn Tunables + Send + Sync {
    fn size_of_val(&self, tracker: &mut dyn MemoryUsageTracker) -> usize
        ...
    }
}
```

or somesuch which then delegates to some method of the `Tunables` trait.

syrusakbary:
It makes a lot of sense. I was thinking that perhaps would be a good idea to make `loupe::MemoryUsage` dependency optional across all crates. Thoughts?

nagisa:
Hm, if `MemoryUsage` was made optional, then perhaps simply re-exporting `MemoryUsage` from something like `wasmer-types` could be good enough for people who actively utilize that functionality, and for the rest of us it would no longer be a concern at all.

syrusakbary:
Adding the 2.x milestone and @Amanieu so we can move this things forward as part of the next release

Amanieu:
We don't use `MemoryUsage` internally, it's purely available as a way for users to estimate the memory usage of their Wasm instances. I think the best solution here would be to just re-export the `loupe` crate as `wasmer::loupe`. This allows users to just import `wasmer::loupe::MemoryUsage` and then `#[derive(MemoryUsage)]` to satisfy the supertrait requirement.

> Adding the 2.x milestone and @Amanieu so we can move this things forward as part of the next release

The timing is rather unfortunate since the 2.1.1 release just went out an hour ago.

syrusakbary:
#2744 is the follow-up for this issue

PR #2868, opened 2022-05-04
Remove loupe dep for wasmer 3.0

epilys:
Closes https://github.com/wasmerio/wasmer/issues/2731
Closes https://github.com/wasmerio/wasmer/issues/2744


webmaster128:
If I understand correctly, this unconditionally kills the loupe feature. I see the issues that loupe can bring. However, this is a problem as we need this in order to get the in-memory size of a `Module`, which is something we do in order to cache as many modules in memory as possible, given a configured module bound.

E.g. https://github.com/CosmWasm/cosmwasm/blob/v1.0.0-beta8/packages/vm/src/cache.rs#L289-L298

It would be great if we could recosider this change.

epilys:
> If I understand correctly, this unconditionally kills the loupe feature. I see the issues that loupe can bring. However, this is a problem as we need this in order to get the in-memory size of a `Module`, which is something we do in order to cache as many modules in memory as possible, given a configured cache memory limit.
> 
> E.g. https://github.com/CosmWasm/cosmwasm/blob/v1.0.0-beta8/packages/vm/src/cache.rs#L289-L298
> 
> It would be great if we could recosider this change.

We appreciate your comment. Since there are people dependent on this feature we will consider adding it back. However notice that this branch is for the next major release Wasmer 3.0, not the current major release. We will try to fit in loupe or similar functionality back to the API when it's done refactored; if you wish make a ticket requesting a memory size capability for the new runtime version and we will keep it in our todo list!

Issue #2758, opened 2022-01-19
Use rkyv Archive types directly instead of deserializing


Amanieu:
Wasmer uses rkyv as its serialization framework but performs a full serialization when loading a module and therefore misses out on rkyv's main feature which is zero-copy deserialization. Rkyv works by creating a separate [`Archive`](https://docs.rs/rkyv/latest/rkyv/trait.Archive.html) type for every serializable type. These types can be used straight from the byte buffer without any processing.

Wasmer should switch to using the `Archive` version of serializable types in the VM to improve the speed to module loading by entirely avoiding deserialization overhead.

nagisa:
I've been working on implementing a similar change in one of the forks. This is definitely doable and produces significant speed-ups[^1], but the separation and isolation between `Artifact`, `Engine`, `Instance` and `ModuleInfo` is not super great. And so implementing a change like this forces significant API changes (e.g. removing `module_ref` getters, which touch everything else). Such a change is further complicated also by prevalent use of `Arc<dyn ...>` types. One more complication is that it isn't super easy to construct rkyv data structures from scratch which is something that would be necessary when the `Artifact` is compiled from wasm in the first place.

There are definitely some preparation steps that can help to implement this more easily:

* Make sure constructing an `Artifact` does not allocate into an `Engine` (e.g. `UniversalArtifact` does this – move those to instantiation time)
* Remove the `ModuleInfo` from any structure other than implementers of `Artifact` that contain it – obtain this information by passing a reference to an `Artifact` to the methods (all the way to the public facing API)

[^1]: Even when `Artifact` become a `Vec<u8>` of the serialized data (to ensure it remains “owned”; artifact that borrows backing data is a further can of worms), which forces _one_ copy of the artifact data can result in speed ups in order of 3 to 5x from deserialize to execute wasm module's entrypoint.

nagisa:
So, we had more time to work on this on our end, and while I don't anticipate having time to replicate similar changes upstream, I can at least write down some more information to whoever attempts this. Downstream we have removed a fair amount of functionality which made implementation of this feature _significantly_ more straightforward. Of note is, for example, that we removed the ability to obtain the list of `Export`s and `Import`s, since they are tightly coupled with having access to the `ModuleInfo`.

There have also been some changes from what I've written in the comment above, so just ignore that one ^^

1. **rkyv integration**: Use of `rkyv` probably wants to become unconditional (i.e. not a feature);
[...]

PR #2869, opened 2022-05-05
Wasmer3 rkyv archives WIP

epilys:
This PR fixes:
* #2758
* #2916 (partially)

# Review

- [ ] Document API changes in documentation and migration document


lib/vm/Cargo.toml
```diff
@@ -21,7 +21,7 @@ more-asserts = "0.2"
 cfg-if = "1.0"
 backtrace = "0.3"
 serde = { version = "1.0", features = ["derive", "rc"] }
-rkyv = { version = "0.7.20", optional = true }
+rkyv = { version = "0.7.38", features = ["indexmap"] }
```
Amanieu, line 24: It seems strange that the vm crate needs any serializable types. The vm crate is not used by the compiler, so how could the compiler include any of these types in the serialized artifact?

ptitSeb:
@epilys can you try some `make build-wasmer-wasm` from time to time, to make sure wasm build is not broken

epilys:
> @Amanieu We'll want to add back the `create-exe` code in the future once the universal engine is capable of emitting object files.

You're right to note this. Since this PR touches/changes/removes lots of stuff before merge we need to write down what needs to be re-introduced after this change.

epilys:
The artifact serialization doesn't include the magic header and the length anymore. `rkyv` provides a validation option through a feature flag. The safe {de,}serialization methods should use this.

Amanieu:
Deserialization can't be made safe: you're loading arbitrary executable code and there's no way to verify it. We shouldn't try to validate the input data, the magic header is just there to provide a version number so that internal format can evolve.

epilys:
@Amaniue the comment says about rkyv's validation feature through [the bytecheck crate](https://github.com/rkyv/bytecheck) which if I understand correctly validates the byte alignments etcetera is correct, not the fact that you're loading arbritary executable code which is offtopic here. In my local environment, deserializing a random file as a serialized artifact leads to SIGSEGV/invalid memory reference errors but compiling with validation rejects it with an error. It's just a layer of validation, not a 100% catch-all solution.

Amanieu:
Sure but we can get the same level of safety without the performance overhead by just using the magic header.

epilys:
So you're basically saying we should keep the magic header. ACK.

PR #2872, opened 2022-05-06
lib/c-api: fix env_finalizer being ran everytime WrapperEnv was dropped

### LINKED ISSUE #2871: Finalizer is called twice when using the wasm_func_new_with_env function in the C API
### Describe the bug

A finalizer passed to wasm_func_new_with_env will be called twice, instead of as expected once. This can cause issues.

### Steps to reproduce

- Create a function with an env and finalizer
- Import the function into an instance
- Delete the function through wasm_func_delete
- Delete the instance through wasm_instance_delete

A full example in C can be found below.

### Expected behavior

The finalizer should be called exactly once. 

### Actual behavior

The finalizer is called both when the function is deleted, and when the instance is deleted.

### Additional context

This is problematic as I am using the finalizer to clean up the environment object I created for the function. As the finalizer is called twice, this means the deletion of the object is attempted twice, leading to a use after free.

### Full C example
```
 #include <wasm.h>
 #include <wasmer.h>
 #include <stdio.h>

wasm_trap_t* empty_callback(void* enc, const wasm_val_vec_t* params, wasm_val_vec_t* result){ return 0; }

void finalizer(void* env){ printf("Finalizer called\n"); }

int main(int argc, const char* argv[]) {
    wasm_engine_t* engine = wasm_engine_new();

    wasm_functype_t* host_func_type = wasm_functype_new_0_0();
    wasm_func_t* host_func = wasm_func_new_with_env( store, host_func_type, empty_callback, engine, finalizer);
    wasm_functype_delete(host_func_type);

    const char *wat_string =
        "(module\n"
        "  (func $host_function (import \"\" \"host_function\")))";

    wasm_byte_vec_t wat;
    wasm_byte_vec_new(&wat, strlen(wat_string), wat_string);
    wasm_byte_vec_t wasm_bytes;
    wat2wasm(&wat, &wasm_bytes);
    wasm_byte_vec_delete(&wat);

    wasm_module_t* module = wasm_module_new(store, &wasm_bytes);
    wasm_byte_vec_delete(&wasm_bytes);

    wasm_extern_t* externs[] = {
        wasm_func_as_extern(host_func),
    };

    wasm_extern_vec_t import_object = WASM_ARRAY_VEC(externs);
    wasm_instance_t* instance = wasm_instance_new(store, module, &import_object, NULL);
    wasm_func_delete(host_func);
    wasm_module_delete(module);
    wasm_instance_delete(instance);
    wasm_store_delete(store);
    wasm_engine_delete(engine);
}
```
Compile through: ``clang example_file.c -I$WASMER_DIR/include $WASMER_DIR/lib/libwasmer.a -lm``

Observe that "Finalizer called" is printed twice.

epilys:
We would like to run the environment finalizer callback only when the
last reference to env pointer is dropped, which is when the Arc<_>
reference count is one.

Closes #2871

### DIFF
--- a/lib/c-api/src/wasm_c_api/externals/function.rs
+++ b/lib/c-api/src/wasm_c_api/externals/function.rs
@@ -6,7 +6,7 @@ use super::CApiExternTag;
 use std::convert::TryInto;
 use std::ffi::c_void;
 use std::mem::MaybeUninit;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use wasmer_api::{Function, RuntimeError, Val};
 
 #[derive(Debug, Clone)]
@@ -110,7 +110,7 @@ pub unsafe extern "C" fn wasm_func_new_with_env(
     #[repr(C)]
     struct WrapperEnv {
         env: *mut c_void,
-        env_finalizer: Arc<Option<wasm_env_finalizer_t>>,
+        env_finalizer: Arc<Mutex<Option<wasm_env_finalizer_t>>>,
     }
 
     impl wasmer_api::WasmerEnv for WrapperEnv {}
@@ -122,8 +122,12 @@ pub unsafe extern "C" fn wasm_func_new_with_env(
 
     impl Drop for WrapperEnv {
         fn drop(&mut self) {
-            if let Some(env_finalizer) = *self.env_finalizer {
-                unsafe { (env_finalizer)(self.env as _) }
+            if let Ok(mut guard) = self.env_finalizer.lock() {
+                if Arc::strong_count(&self.env_finalizer) == 1 {
+                    if let Some(env_finalizer) = guard.take() {
+                        unsafe { (env_finalizer)(self.env as _) };
+                    }
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -122,8 +122,12 @@ pub unsafe extern "C" fn wasm_func_new_with_env(
 
     impl Drop for WrapperEnv {
         fn drop(&mut self) {
-            if let Some(env_finalizer) = *self.env_finalizer {
-                unsafe { (env_finalizer)(self.env as _) }
+            if Arc::strong_count(&self.env_finalizer) == 1 {
+                if let Ok(mut guard) = self.env_finalizer.lock() {
```
syrusakbary, line 126: Should we lock before the strong count to prevent double drops?

epilys: There's not gonna be a double drop, because the callback is behind a mutex. I think what you mean to say is that strong_count might return 2 and before that is compared to 1, another instance has dropped. So the only failing scenario is the callback not running at all. I will switch the lock to be first.

syrusakbary: Yup, that's what I was referring. Thanks!

syrusakbary (approved):

Issue #2637, opened 2021-10-28
Remove Module::artifact


syrusakbary:
### Describe the bug

Remove `Module::artifact` method since it is too dangerous (see #1943 for more
details). This method has initially been introduced as a hack

This should be done after #2916

syrusakbary:
This is fixed in #3029

PR #2876, opened 2022-05-09
Removed Module::Artifact (for #2637)

ptitSeb:
# Description
Removed Module::Artifact (that was hidden already).
For ticket #2637


Amanieu:
Both `symbol_registry` and `metadata_length` still seem like internal implementation details that we don't want to expose in the public API.

@syrusakbary maybe you have some more context here?

ptitSeb:
so they should be "hidden"? like artifact() was before?

syrusakbary:
> Both symbol_registry and metadata_length still seem like internal implementation details that we don't want to expose in the public API.

I agree with this. We are removing artifact but adding two other public things that can be even more dangerous.
If I remember why artifact was needed in Wasmer, was for compiling in the static lib. Perhaps we can fully remove the method and see if we can find other ways of making the dangling code to work, not sure.

ptitSeb:
The idea with this was:
1. remove the possibility to get the Artifact from the module, to avoid mis-used of this
2. Added as small as possible function to not break functionnality. I added those 2 functions while keeping in mind that the different engines will be merge in the future, and so that those method could make sense in an unified way. 

If you don't want to hide the method and remove them, them I'm confused: what kind of mecanism do you propose?

ptitSeb:
What if I change to:
````
    #[doc(hidden)]
    pub fn metadata_prefix(&self) -> Option<String> {
        self.artifact.metadata_prefix()
    }

    #[doc(hidden)]
    pub fn metadata_length(&self) -> usize {
        self.artifact.metadata_length()
    }
```
Would that seems less dangerous and acceptable?

PR #2881, opened 2022-05-12
bors.toml: Update status checks

epilys:
bors.toml wasn't updated with the new workflows merged in 86818452411dbeb290610f58f7ef23debf6a4ea9
and bors was timing out waiting on statuses that weren't there.

### DIFF
--- a/bors.toml
+++ b/bors.toml
@@ -4,10 +4,9 @@ status = [
     "Test on linux-x64",
     # "Test on linux-musl-x64",
     # "Test on linux-aarch64",
+    "Test on macos-arm64",
     "Test on macos-x64",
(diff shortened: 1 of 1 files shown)


Amanieu:
Let's give it a try, I think bors should pick up the updated bors.toml.

bors r+

epilys:
[Oh indeed:](https://bors.tech/documentation/getting-started/)

> Note: that bors reads bors.toml from the pull requests it’s merging, not the one in master, so changes to the file get checked before they land. You can therefore verify whether bors is working by using bors r+ to try and land the PR that enables it!

That means we need to rebase `wasmer3` again btw.

PR #2891, opened 2022-05-17
Require minimum rust version in Cargo.toml instead of rust-toolchain file

silwol:
# Description

# Review


syrusakbary:
I'm not sure if we want this. We need to double check that people using wasmer (as a lib) will not be enforced to use a specific Rust version

silwol:
The question is, what is the reason for having a `rust-toolchain` in the first place? This is what interfered with my rust setup which is actually pretty mainline, so I expect this to cause issues for others as well if they want to start developing. My thought was that adding a MSRV is close enough to what the `rust-toolchain` file does right now. If there is no hard requirement on a MSRV, we can of course leave that out or think about other options.

syrusakbary:
> The question is, what is the reason for having a rust-toolchain in the first place?

Basically, when people clone the project and contribute to it they must know what is the minimum version expected to work to compile Wasmer and run it's tests with.

When people use the library, they should not enforced to use a specific version of Rust.

epilys:
[From Cargo docs:](https://doc.rust-lang.org/nightly/cargo/reference/manifest.html#the-rust-version-field)

> Setting the rust-version key in [package] will affect all targets/crates in the package, including test suites, benchmarks, binaries, examples, etc.

As for the libraries that use `wasmer` as a dependency, [apparently Cargo does not use `rust-version` for dependencies](https://github.com/rust-lang/cargo/issues/9930).

It does not provide any value to `wasmer` to require a specific version because we're not (I think) using features that don't exist on earlier versions. To have this documented for ourselves we can build it for all versions after let's say 1.56 and see if there's something unexpected.

jcaesar:
> Basically, when people clone the project and contribute to it they must know what is the minimum version expected to work to compile Wasmer and run it's tests with.

You do want this.
* The rust-toolchain file will be ignored if you're not using rustup 
* The rust-toolchain file doesn't set a minimum, but the default version which will be auto-installed even if you have a newer rust
* Cargo-toml's rust-version sets a minimum and gives nice error message:
```none
error: package `bar v0.1.0 (/src/bar)` cannot be built because it requires rustc 1.59 or newer, while the currently active rustc version is 1.56.1
```
Caveat here is that rust before 1.56 will ignore rust-version:
```none
warning: `rust-version` is not supported on this version of Cargo and will be ignored
```
This could be fixed by also bumping the edition to 2021.

Anyway: wasmer-vm depends on corosensei, which already sets `rust-version = "1.59"` and `edition = "2021"`. So this PR doesn't really change the situation much, but makes the situtation a bit clearer. ¯\\\_(つ)\_/¯

> As for the libraries that use wasmer as a dependency, apparently Cargo does not use rust-version for dependencies

You're misinterpreting this. The rust-version isn't used as a constraint for dependency resolution. The `rust-version` on dependencies is definitely checked.

PR #2892, opened 2022-05-20
[WIP] Implement new Context API for Wasmer 3.0

### LINKED ISSUE #1522: Ability to create a “null” funcref
When creating a `Table`, a `TableType { ty: Type::FuncRef, … }` type is required (as first argument), but an initial value too (as second argument). I see `Value::null()` that creates an `ExternRef`, but how to create a “null” function reference?

We can't. We need a new API for that :-).

### LINKED ISSUE #1630: Instance host_state not required to be Sync, yet can be accessed from multiple threads at the same time.

### LINKED ISSUE #1632: InstanceHandler::memory_grow performs unsynchronized write to shared memory

### LINKED ISSUE #1734: Explicit Send and Sync
### Motivation

Currently, it is unclear which types are meant to be Send and/or Sync. For example:
- `Memory` is Sync, but `Instance` is not
- `Function`is Sync but `NativeFunc` is not
The syncness is also due to auto impl which can change with the implementation.

### Proposed solution

Add `Sync`, `!Sync`, `Send` and `!Send` implementation for all major types, including `Instance`, `Module`, `Function`, `NativeFunc`, `Exports`, `Memory` and others I may have missed. Basically, any type with Arc inside, I think. Sync and Send would then be part of the public API.

### LINKED ISSUE #1840: Support closures as host functions tracking issue

### LINKED ISSUE #2544: Relaxed (no Send + Sync) requirements for WasmerEnv for single-threaded applications
### Motivation

I'm writing a program to run routines from 1 or more modules *sequentially*.  While I have what I need to do this, it's a bit awkward using `Arc<T>` and `Arc<Mutex<T>>` simply because it's required for fields of types deriving from `WasmerEnv` that I want to share between host and modules.

### Proposed solution

Either an alternative trait to WasmerEnv that doesn't derive from Send/Sync, or some configuration or switch that disables the requirement on build.  There might be a "smarter" solution wherein a type could derive Send and Sync *explicitly* in addition to WasmerEnv and be used in multi-threaded contexts; or not derive Send and Sync and only be used in single-threaded contexts.  In general, I think that would be preferable, however it would make the change a breaking change; besides, I don't even know if that would be possible.

### Additional context

I understand that this is probably a low-reward change with a potentially high cost (I'm guessing this would require changes throughout the Wasmer project).  It is perhaps just an aesthetic concern -- there might be a performance gain, but I don't have any benches to suggest it would be significant -- but I think that using `Rc<T>` and `Rc<RefCell<T>>` would make the code clearer to any potential consumers of my code that it's intended for single-threaded use only.

### LINKED ISSUE #2816: Wasmer/js raises `cannot read property 'buffer'` when try to access memory via Env
### Describe the bug

Note: I'm filing this as a bug since I was not able to figure out why one passes and the other is not, but it is totally possible I made a mistakes in my code. Please change category if needed. 🙏 

I'm trying to make existing wasm execution logic to be compatible on `wasmer/js` features as well, to allow run wasm binary from its executor / runner compiled into wasm itself. I made poc work, however in certain case wasmer raises `TypeError: Cannot read property 'buffer' of undefined` when it try to access memory from imported fn's env and I am not able to figure out what exactly triggers this.

#### Passing poc

```
fn run() {
    let wasmer_store = Store::default();
    let module = Module::new(&wasmer_store, &WASM_BINARY).unwrap();
    let transform_result: Arc<Mutex<Vec<u8>>> = Arc::new(Mutex::new(vec![]));

    let set_transform_result_fn_decl = Function::new_native_with_env(
        &wasmer_store,
        HostEnvironment {
            memory: LazyInit::default(),
            transform_result: transform_result.clone(),
        },
        set_transform_result,
    );

    let imports = imports! {
        "env" => {
            "__set_transform_result" => set_transform_result_fn_decl,
        }
    };

    let instance = Instance::new(&module, &imports).unwrap();

    let input_ptr = write_bytes_into_guest(&instance, &input_serialized);

    let transform_fn = instance
        .exports
        .get_native_function::<(i32, i32), i32>("__guest_transform")
        .unwrap();

    transform_fn.call(input_ptr.0, input_ptr.1).unwrap();
}
```

#### Failing POC

```
fn load_plugin() -> Result<(Instance, Arc<Mutex<Vec<u8>>>), Error> {
    let wasmer_store = Store::default();
    let module = Module::new(&wasmer_store, &WASM_BINARY);

    return match module {
        Ok(module) => {
            let transform_result: Arc<Mutex<Vec<u8>>> = Arc::new(Mutex::new(vec![]));

            let set_transform_result_fn_decl = Function::new_native_with_env(
                &wasmer_store,
                HostEnvironment {
                    memory: LazyInit::default(),
                    transform_result: transform_result.clone(),
                },
                set_transform_result,
            );

            let imports = imports! {
                "env" => {
                    "__set_transform_result" => set_transform_result_fn_decl,
                }
            };

            let instance = Instance::new(&module, &imports).unwrap();

            Ok((instance, transform_result))
        },
        Err(err) => panic!("should not be here"),
    }
}

fn run() {
    let (instance, transform_result) = load_plugin().unwrap();

    let input_ptr = write_bytes_into_guest(&instance, &input_serialized);

    let transform_fn = instance
        .exports
        .get_native_function::<(i32, i32), i32>("__guest_transform")
        .unwrap();


    transform_fn.call(input_ptr.0, input_ptr.1).unwrap();
}
```

Codes are exactly same, the only difference is failing cases have separate fn to load / instantiate module then return created instance (with cloned env properties).

1. Is there something I did setup mistakenly cuases this?
2. If so, what's the reason passing case works properly? Somethings gets dropped with `load_plugin` maybe?
3. What is the correct / recommended pattern in this case?

Thanks!

### Steps to reproduce

1. Clone https://github.com/kwonoj/wasmer-js-test
5. Run ./build.sh && node index.js
6. See test_success() passes (console out JSON object), test_fail() raises exception

### Actual behavior

```
wasm-bindgen: imported JS function that was not marked as `catch` threw an error: Cannot read property 'buffer' of undefined

Stack:
TypeError: Cannot read property 'buffer' of undefined
    at /home/ojkwon/github_wsl/wasmer-js-test/host_runner/pkg/host_runner.js:460:30
    at logError (/home/ojkwon/github_wsl/wasmer-js-test/host_runner/pkg/host_runner.js:222:18)
    at module.exports.__wbg_buffer_5e74a88a1424a2e0 (/home/ojkwon/github_wsl/wasmer-js-test/host_runner/pkg/host_runner.js:459:68)
    at js_sys::WebAssembly::Memory::buffer::hf044a38285fed8aa (<anonymous>:wasm-function[8773]:0x2f99f6)
    at wasmer::js::externals::memory::Memory::size::hfd9a6285e7a4b573 (<anonymous>:wasm-function[1247]:0x1b53e1)
    at wasmer::js::ptr::WasmPtr<T,wasmer::js::ptr::Array>::deref::h310443bb52737f93 (<anonymous>:wasm-function[587]:0x14acbc)
    at host_runner::set_transform_result::hac06130700517b90 (<anonymous>:wasm-function[866]:0x17f185)
    at core::ops::function::Fn::call::h4ed7026923862d80 (<anonymous>:wasm-function[7064]:0x2d468f)
    at <Func as wasmer::js::externals::function::inner::HostFunction<(A1,A2),Rets,wasmer::js::externals::function::inner::WithEnv,Env>>::function_body_ptr::func_wrapper::{{closure}}::h558bdfb380e59924 (<anonymous>:wasm-function[4322]:0x28659d)
    at <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once::h5553e26885b27801 (<anonymous>:wasm-function[4896]:0x29a416)
```

### LINKED ISSUE #2838: Improve usabiliy of WasmerEnv trait
### Motivation

[`init_with_instance`](https://github.com/wasmerio/wasmer/blob/baebb436e25f72657feb77fd059d1172ed2a011f/lib/api/src/sys/env.rs#L91) provides the host env with a `&Instance`, where `Instance` is `Clone`, making it very tempting for people to simply clone the instance and store it on the host env for later use, which seems to create a memory leak. 

### Proposed solution

`init_with_instance` could  take an argument that would enforce the implicit constraints of the method. 

For example a `&InitialiazedInstanceProxy`, which would not be `Clone`, and would have a method like `get_export_with_generics_weak`, wich would proxy internally to `instance.exports.get_with_generics_weak`.

### Alternative

Clarify this issue in the documentation. Good idea anyway, since people can still easily store an `Instance` on their host env outside of `init_with_instance`.

### Additional context

The temptation to store `Instance` on the host env is especially tempting in the light of using `wasmer_middlewares::metering`, since helper functions like [`get_remaining_points`](https://github.com/wasmerio/wasmer/blob/8f2e49d52823cb7704d93683ce798aa84b6928c8/lib/middlewares/src/metering.rs#L293) take an `&Instance` as argument, and it's not obvious how to get access to it from host functions unless stored on the env. 

Example of mistake made in the wild and fix: https://github.com/massalabs/massa-sc-runtime/pull/121

### LINKED ISSUE #2856: Stack var overlapped for multithread
### Describe the bug
Given the following guest rust code and host-runtime implementation:
```rust
extern "C" {
    fn ocall(thread: i32, nth: i32, p: i32);
}

#[inline(never)]
fn use_more_stack(n: i32) -> i32 {
    let mut bug_buffer = [n; 32];
    unsafe {
        ocall(n, 2, bug_buffer.as_mut_ptr() as i32);
    }
    bug_buffer[0]
}

#[no_mangle]
extern "C" fn entry(thread: i32) {
    let mut b = [thread; 1];
    unsafe {
        ocall(thread, 1, b.as_mut_ptr() as i32);
    }
    let bug_thread_num = use_more_stack(thread);
    unsafe {
        ocall(thread, 3, bug_thread_num);
    }
}
```
```rust
use wasmer::{imports, Function, Instance, Module, Store};

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let wasm_bytes = include_bytes!("../guest/target/wasm32-unknown-unknown/release/guest.wasm");
    let store = Store::default();

    println!("Compiling module...");

    fn ocall(thread: i32, nth: i32, p: i32) {
        println!("thread {}, {}-nth ocall, p={}", thread, nth, p);
        std::thread::sleep(std::time::Duration::from_millis(50));
    }

        "env" => {
            "ocall" => Function::new_native(&store, ocall),
        }
    };

    println!("Instantiating module...");

    let mut threads = vec![];
    for i in 0..3 {
        let handle = std::thread::spawn(move || {
            let entry = instance
                .exports
                .get_function("entry")
                .unwrap()
                .native::<i32, ()>()
                .unwrap();
            entry.call(i).unwrap();
        });
        threads.push(handle);
    }
    for handle in threads {
        handle.join().unwrap();
    }
    Ok(())
}
```

`cargo run` usually output as expected:
```
$ cargo run --release
    Finished release [optimized] target(s) in 0.10s
     Running `target/release/play-wasmer`
Compiling module...
Instantiating module...
thread 0, 1-nth ocall, p=1048572
thread 1, 1-nth ocall, p=1048556
thread 2, 1-nth ocall, p=1048540
thread 0, 2-nth ocall, p=1048400
thread 1, 2-nth ocall, p=1048272
thread 2, 2-nth ocall, p=1048144
thread 0, 3-nth ocall, p=0
thread 2, 3-nth ocall, p=2
thread 1, 3-nth ocall, p=1
```

But sometimes, the output is somthing like:
```
$ cargo run --release
    Finished release [optimized] target(s) in 0.04s
     Running `target/release/play-wasmer`
Compiling module...
Instantiating module...
thread 0, 1-nth ocall, p=1048572
thread 2, 1-nth ocall, p=1048556
thread 1, 1-nth ocall, p=1048540
thread 0, 2-nth ocall, p=1048400
thread 2, 2-nth ocall, p=1048272
thread 1, 2-nth ocall, p=1048272
thread 0, 3-nth ocall, p=0
thread 2, 3-nth ocall, p=1
thread 1, 3-nth ocall, p=1
```

Which means the `bug_buffer` in guest get overlapped memory for thread `1` and thread  `2`.

```sh
wasmer: command not found
 | rustc 1.60.0 (7737e0b5c 2022-04-04) | x86_64
```

wasmer in Cargo.lock:
```
[[package]]
name = "wasmer"
version = "2.1.1"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "23f0188c23fc1b7de9bd7f8b834d0b1cd5edbe66e287452e8ce36d24418114f7"
```

### Steps to reproduce
```
git clone https://github.com/kvinwang/wasmer-stack-overlap-poc.git
cd wasmer-stack-overlap-poc/guest
cargo build --release --target wasm32-unknown-unknown
cd ..
cargo run # again and again until the issue appears.
```

### LINKED ISSUE #2866: Please implement the directly call funcref from guest table feature
### Motivation

I'm building plugin system for my rust app using wasmer crate, and I need to call funcref from guest table in host directly without calling call_callback function from guest

### Proposed solution

This feature is implemented in wasmtime crate (https://github.com/bytecodealliance/wasmtime)

### Additional context
Note: error message when i try to call funcref from guest table directly
```
thread 'main' panicked at 'not implemented: Native function definitions can't be directly called from the host yet', MY_CARGO_PATH\wasmer-2.2.1\src\sys\externals\function.rs:540:17
```

### LINKED ISSUE #2873: Dropping all references to a `Global` export leads to a use-after-free in returned `NativeFunc`s
### Describe the bug

When a `NativeFunc` is returned from a function, &&a

```sh
wasmer 1.2.1 (crate) | rustc 1.60.0 (7737e0b5c 2022-04-04) | arm64
```

### Steps to reproduce
This can be reproduce with a fresh Rust project and the following code.

```rust
use wasmer::{self, Store, imports, Instance, NativeFunc, Module, Global};

fn make_function() -> (Global, NativeFunc<(), i64>) {
    let store = Store::default();
    let wasm = wasmer::wat2wasm(r#"
    (module
        (import "global" "foo" (global $global_foo i64))
        (func $bar (result i64) (global.get $global_foo))
        (export "bar" (func $bar)))
    "#.as_bytes()).unwrap();

    let global = Global::new(&store, wasmer::Value::I64(123));
    let module = Module::new(&store, wasm).unwrap();
    let imports = imports! { "global" => { "foo" => global.clone() } };
    let instance = Instance::new(&module, &imports).unwrap();
    let func: NativeFunc<(), i64> = instance.exports.get_native_function("bar").unwrap();
    (global, func)
}

fn main() {
    let (global, func) = make_function();
    dbg!(func.call().unwrap());
    drop(global);
    dbg!(func.call().unwrap());
}
```

### Expected behavior
The above code outputs
```
[src/main.rs:22] func.call().unwrap() = 123
[src/main.rs:24] func.call().unwrap() = 123
```

### Actual behavior
The above code outputs
```
[src/main.rs:22] func.call().unwrap() = 123
[src/main.rs:24] func.call().unwrap() = 101649545396320
```

I feel like `NativeFunc` instances either shouldn't be able to outlive their instance's exports or should keep a reference to them internally.

### Additional context

My actual use case doesn't require any of the allocation / reference counting performed by `Global`, and it would be nice if I could simply provide a value for immutable global exports.

### LINKED ISSUE #2893: Late initialization of the host environments
The problem occurs when the `start function` and call the ABI, it looks like the host environment isn't yet initialized and functions like
`env.remaining_points.as_ref()` return an error.

We can reproduce with that kind of code:

```rust
    let resolver: ImportObject = imports! {
        "my_api" => {
            // other stuff
            "foo" => Function::new_native_with_env(&store, env.clone(), foo),
        },
    };

    // [....]
    fn foo(env: &Env) {
        env.remaining_points.as_ref().unwrap() // <-- panic if foo is called in the start function
    }
```

Amanieu:
### Overview

This PR reworks the way Wasmer manages objects used by instances, specifically: `Instance`, `Memory`, `Table`, `Function`, `Global` and `ExternRef`. Previously these were tracked using reference counting, but this cannot work fully because Wasm code can manipulate references to functions (and by implication the instance that function is from) using tables. 

The new approach uses a `Context` type which owns all of the objects listed above: objects are only freed when the `Context` is dropped. The `Instance`, `Memory`, `Table`, `Function`, `Global` and `ExternRef` types are now simply wrappers around an integer which indexes a vector in the `Context`. As a result, any function on those "handle" types now need to take a `&Context` or `&mut Context` as a parameter to access the actual data in the context. This also vastly simplifies Wasmer's thread-safety story: since all mutating operations require a `&mut Context`, no synchronization or internal locks are needed.

Additionally, the way host functions are represented is also changed. The `WasmerEnv` trait is removed since it was unergonomic and hard to use correctly. Instead, the state for host functions is stored directly as the `T` in `Context<T>`. This state is initialized when the context is created, and can be accessed through the `data` and `data_mut` methods on `Context<T>`. All host functions receive a `ContextMut<'_, T>` as their first argument, which gives them access to the host state of the context. This is much better than the old approach since it provides mutable access without locks and allows all host functions to share the same state.

This design is heavily inspired by [this RFC](https://github.com/bytecodealliance/rfcs/blob/main/accepted/new-api.md) from Wasmtime.

### TODO

- [ ] #2908
- [x] Update wasmer-wasi to use the new API
- [x] #2909
- [x] #2913
- [ ] #2912 
- [ ] #2911 
- [ ] #2910

Fixes #2893
Fixes #2873 
Fixes #2866
Fixes #1840 
Fixes #1734
Fixes #1632
Fixes #1522
Fixes #1630 
Fixes #2838
Fixes #2856 
Fixes #2544
Fixes #2816


lib/api/src/sys/context.rs
```diff
@@ -0,0 +1,191 @@
...
+pub(crate) struct ContextInner<T> {
+    pub(crate) objects: ContextObjects,
+    pub(crate) store: Store,
+    pub(crate) data: T,
+}
+
+/// A context containing a set of WebAssembly instances, along with host state.
+///
+/// All WebAssembly instances must exist within a context. In the majority of
+/// cases each intance will have its own context, but it is possible to have
```
syrusakbary, line 17:
```suggestion
/// cases each instance will have its own context, but it is possible to have
```

examples/hello_world.rs
```diff
@@ -63,25 +67,26 @@ fn main() -> anyhow::Result<()> {
...
-    let instance = Instance::new(&module, &import_object)?;
+    let instance = Instance::new(&mut context, &module, &import_object)?;
 
     // We get the `TypedFunction` with no parameters and no results from the instance.
     //
     // Recall that the Wasm module exported a function named "run", this is getting
     // that exported function from the `Instance`.
-    let run_func: TypedFunction<(), ()> = instance.exports.get_native_function("run")?;
+    let run_func: TypedFunction<(), ()> =
+        instance.exports.get_native_function(&mut context, "run")?;
```
syrusakbary, line 85: Can we mark the function `get_native_function` as deprecated, and points to use a new `get_typed_function` instead?

syrusakbary:
We need to fix: #3012 

> When using wasmer as a dependency for another project, it fails upon instantiation with this message:
> Error: Compilation error: The UniversalEngine is not compiled in.
> This appears to be caused by the wasmer-compiler/universal_engine feature not being enabled.

PR #2898, opened 2022-05-23
Make rayon optional in cranelift compiler

ptitSeb:
# Description
Add cranelift to the list of compiler available in wasmer-compiler build from wasm-wasi32 target


epilys (approved):

syrusakbary:
It seems tests are passing, so we should be good to merge directly :)

PR #2900, opened 2022-05-24
CI: Install bash in alpine container

epilys:
Alpine container was updated and bash is missing from the default
packages. Install it explicitly.


syrusakbary:
Merging directly as it's blocking other PRs

PR #2902, opened 2022-05-25
Migrated WASI and WAST to Wasmer 3.0 API

ptitSeb:
# Description
Migrated WASI and WAST to the new Wasmer 3.0 Context API

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -585,7 +585,7 @@ test-wasi-unit:
 	$(CARGO_BINARY) test $(CARGO_TARGET) --manifest-path lib/wasi/Cargo.toml --release
 
 test-wasi:
-	$(CARGO_BINARY) test $(CARGO_TARGET) --release --tests $(compiler_features) -- wasi::wasitests
+	$(CARGO_BINARY) ltest $(CARGO_TARGET) --release --tests $(compiler_features) -- wasi::wasitests
 
 test-examples:
 	$(CARGO_BINARY) test $(CARGO_TARGET) $(compiler_features) --features wasi --examples
--- a/Makefile
+++ b/Makefile
@@ -585,7 +585,7 @@ test-wasi-unit:
 	$(CARGO_BINARY) test $(CARGO_TARGET) --manifest-path lib/wasi/Cargo.toml --release
 
 test-wasi:
-	$(CARGO_BINARY) ltest $(CARGO_TARGET) --release --tests $(compiler_features) -- wasi::wasitests
+	$(CARGO_BINARY) test $(CARGO_TARGET) --release --tests $(compiler_features) -- wasi::wasitests
 
 test-examples:
 	$(CARGO_BINARY) test $(CARGO_TARGET) $(compiler_features) --features wasi --examples
--- a/lib/wasi/src/lib.rs
+++ b/lib/wasi/src/lib.rs
@@ -47,7 +47,7 @@ pub use crate::state::{
 };
 pub use crate::syscalls::types;
 pub use crate::utils::{get_wasi_version, get_wasi_versions, is_wasi_module, WasiVersion};
-pub use wasmer::{Context, ContextMut};
+use wasmer::{Context, ContextMut};
 #[deprecated(since = "2.1.0", note = "Please use `wasmer_vfs::FsError`")]
 pub use wasmer_vfs::FsError as WasiFsError;
 #[deprecated(since = "2.1.0", note = "Please use `wasmer_vfs::VirtualFile`")]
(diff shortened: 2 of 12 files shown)


lib/wasi/src/syscalls/legacy/snapshot0.rs
```diff
@@ -22,17 +23,18 @@ pub fn fd_filestat_get(
     let new_buf: WasmPtr<types::__wasi_filestat_t> = buf.cast();
 
     // Copy the data including the extra data
-    let new_filestat_setup: types::__wasi_filestat_t = wasi_try_mem!(new_buf.read(memory));
+    let new_filestat_setup: types::__wasi_filestat_t = wasi_try_mem!(new_buf.read(&ctx, memory));
```
syrusakbary, line 26: Does this compile? From what I read on the `WasmPtr` wasmer3 API the signature (https://github.com/wasmerio/wasmer/blob/wasmer3/lib/api/src/sys/ptr.rs#L186) only requires the memory reference.
Am I missing something?

syrusakbary, line 26: Does this compile? From what I read on the `WasmPtr` wasmer3 API the signature (https://github.com/wasmerio/wasmer/blob/wasmer3/lib/api/src/sys/ptr.rs#L186) only requires the memory reference.
Am I missing something?

ptitSeb: It does build, and `make test-wasi` on linux x86_64 also works.

image

From what I have seen, all functions require the Context.

ptitSeb: You are looking at the wrong branch: you need to look in https://github.com/wasmerio/wasmer/tree/context_api for the new API

syrusakbary: ahá! Thanks

syrusakbary:
I think right now `WasiEnv` is using a Mutex that will no longer be needed since the refactor.
We should probably create an issue and work on it after this PR and the one PR for improving WASI #2887 is merged

ptitSeb:
> I think right now WasiEnv is using a Mutex that will no longer be needed since the refactor.
> We should probably create an issue and work on it after this PR and the one PR for improving WASI https://github.com/wasmerio/wasmer/pull/2887 is merged

yes, I have kept the `pub state: Arc<Mutex<WasiState>>,` for now, to ease the migration, but I was thinking of just using a simple Rc instead of Arc Mutex?

syrusakbary:
> yes, I have kept the pub state: Arc<Mutex<WasiState>>, for now, to ease the migration, but I was thinking of just using a simple Rc instead of Arc Mutex

I don't think we need RC either? (rather than cloning the env, we just pass it and let the Context own it?)

Makefile
```diff
@@ -585,7 +585,7 @@ test-wasi-unit:
 	$(CARGO_BINARY) test $(CARGO_TARGET) --manifest-path lib/wasi/Cargo.toml --release
 
 test-wasi:
-	$(CARGO_BINARY) test $(CARGO_TARGET) --release --tests $(compiler_features) -- wasi::wasitests
+	$(CARGO_BINARY) ltest $(CARGO_TARGET) --release --tests $(compiler_features) -- wasi::wasitests
```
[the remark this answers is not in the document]
ptitSeb: oops, forgot to remove that. Sorry.

lib/wasi/src/lib.rs
```diff
@@ -71,200 +72,148 @@ pub enum WasiError {
...
     /// Shared state of the WASI system. Manages all the data that the
     /// executing WASI program can see.
-    ///
-    /// Be careful when using this in host functions that call into Wasm:
-    /// if the lock is held and the Wasm calls into a host function that tries
-    /// to lock this mutex, the program will deadlock.
     pub state: Arc<Mutex<WasiState>>,
-    #[wasmer(export)]
-    memory: LazyInit<Memory>,
+    pub memory: Option<Memory>,
```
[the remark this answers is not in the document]
ptitSeb: once, so if the memory is not None, than panic?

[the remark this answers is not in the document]
ptitSeb: and remove the "pub"?

lib/wasi/src/lib.rs
```diff
@@ -47,6 +47,7 @@ pub use crate::state::{
 };
 pub use crate::syscalls::types;
 pub use crate::utils::{get_wasi_version, get_wasi_versions, is_wasi_module, WasiVersion};
+pub use wasmer::{Context, ContextMut};
```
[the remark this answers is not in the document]
ptitSeb: ok

ptitSeb:
> > yes, I have kept the pub state: Arc<Mutex>, for now, to ease the migration, but I was thinking of just using a simple Rc instead of Arc Mutex
> 
> I don't think we need RC either? (rather than cloning the env, we just pass it and let the Context own it?)

I used Rc<RefCell<WasiState>> to keep the same usage as with Arc<Mutex<WasiState>>. It cannot be a own for state, as it needs to be mutably borrowed all the time.

PR #2933, opened 2022-06-06
Rename NativeFunc to TypedFunction

silwol:
# Description

# Review


syrusakbary (changes requested):
We need to still make `NativeFunc` type alias work (in case someone wants to import, we should just show a deprecation warning)

syrusakbary (approved):

PR #2938, opened 2022-06-07
cli: self update from Github API WIP

epilys:
When doing `wasmer self-update`, fetch the latest release from Github API instead of reusing `wasmer-install` shell script. This way you can avoid redownloading the same version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -132,6 +132,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dc19a4937b4fbd3fe3379793130e42060d10627a360f2127802b10b87e7baf74"
 
 [[package]]
+name = "base64"
+version = "0.13.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


syrusakbary (commented):
I do like centralizing the logic of updating/installing the wasmer client in one place, otherwise it will become hard to update or to install.

If we merge this PR and we want to use other method for updating/installing that is not based on Github releases, then we will need to update both the source code of Wasmer and of Wasmer-install.
So I think the best tradeoff for now is to make sure that the install script works well even when a version of wasmer is already installed.

Thoughts @epilys ?

syrusakbary:
Side comment: reqwest and such also add some overhead for both compiling and the binary executable size (for example, the wasmer headless binary might no longer be 800kb)

epilys:
> Side comment: reqwest and such also add some overhead for both compiling and the binary executable size (for example, the wasmer headless binary might no longer be 800kb)

I will see how small a self-container install binary can be, and if we can use it for wasmer-install.

Issue #2926, opened 2022-06-01
Fix Clippy warnings in Wasmer 3


syrusakbary:
Fix clippy warnings in wasmer 3.
Once we fix it, we should add it to the CI so there are no warnings in the future.

PR #2942, opened 2022-06-09
Fix clippy lints

silwol:
Fixes: #2926

# Description
* Fix all clippy lints
* Adjust `Makefile` so that the `cargo clippy` commands exit with non-zero when warnings are found

# TODO before this is ready
- [x] Describe safety in [lib/engine/src/tunables.rs:67](https://github.com/silwol/wasmer/blob/dev/2926-fix-clippy-warnings-in-wasmer3/lib/engine/src/tunables.rs#L67)
- [x] Describe safety in [lib/types/src/values.rs:70](https://github.com/silwol/wasmer/blob/dev/2926-fix-clippy-warnings-in-wasmer3/lib/types/src/values.rs#L70)
- [x] Describe safety in [lib/types/src/values.rs:76](https://github.com/silwol/wasmer/blob/dev/2926-fix-clippy-warnings-in-wasmer3/lib/types/src/values.rs#L76)
- [x] Describe safety in [lib/emscripten/src/utils.rs:126](https://github.com/silwol/wasmer/blob/dev/2926-fix-clippy-warnings-in-wasmer3/lib/emscripten/src/utils.rs#L126)
I don't know why it needs to be unsafe, can you research why?
  - The reason is the call to `unsafe slice::from_raw_parts_mut(addr, count as usize)` in line 135. Not sure if this function actually encapsulates the behavior good enough, but if it does, we could just put a `unsafe{…}` around this call and remove the `unsafe` from the function. (silwol)
- [x] Describe safety in [lib/emscripten/src/utils.rs:141](https://github.com/silwol/wasmer/blob/dev/2926-fix-clippy-warnings-in-wasmer3/lib/emscripten/src/utils.rs#L141)
We need to double check the safety of the previous method, as this one depends on it (and I believe that's why it's unsafe)
  - That appears to be the case, once the previous method is no longer marked unsafe, the `unsafe` here is no longer required either. (silwol)
- [x] Describe safety in [lib/c-api/src/wasm_c_api/wat.rs:13](https://github.com/silwol/wasmer/blob/dev/2926-fix-clippy-warnings-in-wasmer3/lib/c-api/src/wasm_c_api/wat.rs#L13)
Not sure why it's unsafe, can you research?
  - Ok, `extern "C"`, that's why. No idea why I missed that initially. :-) (silwol)
- [x] Describe safety in [lib/api/src/sys/externals/function.rs:866](https://github.com/silwol/wasmer/blob/dev/2926-fix-clippy-warnings-in-wasmer3/lib/api/src/sys/externals/function.rs#L866)
Not sure why it's unsafe, can you research?
  - I have no clear idea, just the gut feeling that it has something to do with handling the native types. Really would appreciate if we had a better description than what I could guess. (silwol)

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Looking for changes that affect our C API? See the [C API Changelog](lib/c-api/C
 
 ## **Unreleased**
 
+### Fixed
+- [#2942](https://github.com/wasmerio/wasmer/pull/2942) Fix clippy lints.
 
 ## 2.3.0 - 2022/06/06
 
--- a/Makefile
+++ b/Makefile
@@ -745,11 +745,11 @@ install-wasmer-headless-minimal:
 update-testsuite:
 	git subtree pull --prefix tests/wast/spec https://github.com/WebAssembly/testsuite.git master --squash
 
-lint-packages: RUSTFLAGS += -D dead-code -D nonstandard-style -D unused-imports -D unused-mut -D unused-variables -D unused-unsafe -D unreachable-patterns -D bad-style -D improper-ctypes -D unused-allocation -D unused-comparisons -D while-true -D unconditional-recursion -D bare-trait-objects # TODO: add `-D missing-docs` # TODO: add `-D function_item_references` (not available on Rust 1.47, try when upgrading)
+lint-packages: RUSTFLAGS += -D dead-code -D nonstandard-style -D unused-imports -D unused-mut -D unused-variables -D unused-unsafe -D unreachable-patterns -D bad-style -D improper-ctypes -D unused-allocation -D unused-comparisons -D while-true -D unconditional-recursion -D bare-trait-objects -D function_item_references # TODO: add `-D missing-docs`
 lint-packages:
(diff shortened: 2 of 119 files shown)


syrusakbary:
> Describe safety in [lib/emscripten/src/utils.rs:126](https://github.com/silwol/wasmer/blob/dev/2926-fix-clippy-warnings-in-wasmer3/lib/emscripten/src/utils.rs#L126)

```
# Safety
This method is unsafe because it operates directly with the slice of memory represented by the address
```

> Describe safety in [lib/emscripten/src/utils.rs:141](https://github.com/silwol/wasmer/blob/dev/2926-fix-clippy-warnings-in-wasmer3/lib/emscripten/src/utils.rs#L141)

```
# Safety
This method is unsafe because it uses `allocate_on_stack` which is unsafe
```

> Describe safety in [lib/api/src/sys/externals/function.rs:866](https://github.com/silwol/wasmer/blob/dev/2926-fix-clippy-warnings-in-wasmer3/lib/api/src/sys/externals/function.rs#L866)

```
# Safety
This trait is unsafe given the nature of how values are written and read from the native stack
```

syrusakbary (approved):

epilys (approved):

PR #2943, opened 2022-06-09
Fix build error on some archs by using c_char instead of i8

silwol:
# Description

# Review


syrusakbary:
How do we know this works? Can we have a test for it?

ptitSeb:
Is this because char are signed on x86 and unsigned on arm64?

silwol:
> Is this because char are signed on x86 and unsigned on arm64?

Exactly. With this change, it uses the `c_char` directly, independently of the platform it builds on, and therefore inherits whatever type the alias points to.

ptitSeb (approved):

Issue #2921, opened 2022-06-01
Use custom LLVM builds


syrusakbary:
Right now, the default LLVM builds require the library `libtinfo.so.5` which is not always available (for example, in musl).

We should go back to using https://github.com/wasmerio/llvm-custom-builds in our codebase to remove this dependency.

This issue is a follow-up from #2822

PR #2949, opened 2022-06-14
Switch back to using custom LLVM builds

silwol:
Closes: #2921

# Description
Switch back to using the custom LLVM builds from [wasmerio/llvm-custom-builds](https://github.com/wasmerio/llvm-custom-builds).

This will remove the dependency of binaries to `libtinfo5`, verified that in a VM.

# Review


epilys (changes requested):
This seems to result in only `linux-amd64` and `macos-amd64` using the llvm build (you can see that by going to the `make build-wasmer` step and see what compiler feature flags it ends up using

https://github.com/wasmerio/wasmer/actions/runs/2493814170

epilys (changes requested):
Also, on the llvm builds repository the new build artifacts aren't uploaded as Releases, hence the urls used here are from November 2021

silwol:
> Also, on the llvm builds repository the new build artifacts aren't uploaded as Releases, hence the urls used here are from November 2021

The release tag is from November 2021, but that only contained the source tarballs. The change in [Adjust CI for reactivation of custom llvm builds](https://github.com/wasmerio/llvm-custom-builds/commit/1398c3f36f5a09d778ab15896ca4f517f0ecfea1) built the releases and attached the artifacts to the release tag, see [llvm-custom-builds release 13.x](https://github.com/wasmerio/llvm-custom-builds/releases/tag/13.x).

silwol:
> This seems to result in only `linux-amd64` and `macos-amd64` using the llvm build (you can see that by going to the `make build-wasmer` step and see what compiler feature flags it ends up using

Looking into that…

epilys (approved):

PR #2951, opened 2022-06-14
bors.toml: add checks

epilys:


ptitSeb:
This was removed from bors on purpose, because testing was already done before the r+ command was done

epilys:
Yeah but I noticed yesterday it'd still merge a PR even if the testing wasn't finished.

epilys:
Bah, this will fail since this commit doesn't change any part of the code hence all the other tests won't run hence bors will timeout waiting on the code tests to be run... what a mess. I'll just close the PR and give up.

PR #2973, opened 2022-06-24
Port C API to new Context API

epilys:
~~Not done: tests~~


syrusakbary (changes requested):
We shouldn't change the signatures of `wasm_global_type`, `wasm_global_set`, ... from `store` to `ctx` since those signatures are given by the headers in the official wasm-c-api and we must preserve them.
You can check the official headers here: https://github.com/WebAssembly/wasm-c-api

I propose to attach the `ctx` to the `store` itself (when it's created on the c-api) and that way there will be no breaking changes on the API.

epilys:
> We shouldn't change the signatures of `wasm_global_type`, `wasm_global_set`, ... from `store` to `ctx` since those signatures are given by the headers in the official wasm-c-api and we must preserve them. You can check the official headers here: https://github.com/WebAssembly/wasm-c-api
> 
> I propose to attach the `ctx` to the `store` itself (when it's created on the c-api) and that way there will be no breaking changes on the API.

Resolved

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -159,23 +87,18 @@ pub unsafe extern "C" fn wasm_func_new_with_env(
             .take()
             .into_iter()
             .map(TryInto::try_into)
-            .collect::<Result<Vec<Val>, _>>()
+            .collect::<Result<Vec<Value>, _>>()
             .expect("Result conversion failed");
 
         Ok(processed_results)
     };
+    let mut lck = ctx.lock().unwrap();
```
syrusakbary, line 95: We shouldn't lock the context. The Wasm-C-API rely on soundness for using their API. Meaning: it puts the responsibility on the user to use safely on different threads, not on the API itself.

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -221,17 +146,23 @@ pub unsafe extern "C" fn wasm_func_call(
 
 #[no_mangle]
 pub unsafe extern "C" fn wasm_func_param_arity(func: &wasm_func_t) -> usize {
-    func.inner.ty().params().len()
+    let ctx = func.context.as_ref().unwrap();
+    let lck = ctx.lock().unwrap();
```
syrusakbary, line 150: Same, ctx shouldn't be locked

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -221,17 +146,23 @@ pub unsafe extern "C" fn wasm_func_call(
...
+    func.inner.ty(&lck.inner).params().len()
 }
 
 #[no_mangle]
 pub unsafe extern "C" fn wasm_func_result_arity(func: &wasm_func_t) -> usize {
-    func.inner.ty().results().len()
+    let ctx = func.context.as_ref().unwrap();
+    let lck = ctx.lock().unwrap();
```
syrusakbary, line 157: Same, ctx shouldn't be locked

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -221,17 +146,23 @@ pub unsafe extern "C" fn wasm_func_call(
...
+    func.inner.ty(&lck.inner).results().len()
 }
 
 #[no_mangle]
 pub extern "C" fn wasm_func_type(func: Option<&wasm_func_t>) -> Option<Box<wasm_functype_t>> {
     let func = func?;
+    let ctx = func.context.as_ref().unwrap();
+    let lck = ctx.lock().unwrap();
```
syrusakbary, line 165: Same, ctx shouldn't be locked

lib/c-api/src/wasm_c_api/externals/global.rs
```diff
@@ -1,48 +1,56 @@
...
-    let store = store?;
     let global_type = global_type?;
+    let store = store?;
+    let ctx = store.context.as_mut()?;
     let val = val?;
 
     let global_type = &global_type.inner().global_type;
     let wasm_val = val.try_into().ok()?;
-    let store = &store.inner;
+    let mut lck = ctx.lock().unwrap();
```
syrusakbary, line 43: Same, ctx shouldn't be locked

syrusakbary (changes requested):
I added a few comments over locking the context, but it's intended for almost any usage of the `ctx.lock()` in the PR.

In general, we shouldn't rely on locking the context. The Wasm C-API is designed to push that responsibility into the user code, not on the API side

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -1,27 +1,31 @@
...
+use std::rc::Rc;
+use wasmer_api::{Function, RuntimeError, Value};
 
 #[derive(Debug, Clone)]
 #[allow(non_camel_case_types)]
 #[repr(C)]
 pub struct wasm_func_t {
     pub(crate) tag: CApiExternTag,
     pub(crate) inner: Box<Function>,
+    pub(crate) context: Option<Rc<RefCell<wasm_context_t>>>,
```
syrusakbary, line 20: Why do we need to add the new `context` field?
`Function` should already have a reference to the store. And we can retrieve the context from that function store... so why duplicate it? (this comment should apply also for `Global`, ...)

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -159,23 +91,17 @@ pub unsafe extern "C" fn wasm_func_new_with_env(
...
-            .collect::<Result<Vec<Val>, _>>()
+            .collect::<Result<Vec<Value>, _>>()
             .expect("Result conversion failed");
 
         Ok(processed_results)
     };
+    let function = Function::new(&mut ctx.inner, func_sig, inner_callback);
+    drop(ctx);
+    let mut retval = Box::new(wasm_func_t::new(function));
+    retval.context = store.context.clone();
```
syrusakbary, line 102: We can just skip this ?

lib/c-api/src/wasm_c_api/externals/function.rs
```diff
@@ -194,16 +120,20 @@ pub unsafe extern "C" fn wasm_func_call(
 ) -> Option<Box<wasm_trap_t>> {
     let func = func?;
     let args = args?;
+    if func.context.is_none() {
+        crate::error::update_last_error(wasm_store_t::CTX_ERR_STR);
+    }
+    let mut ctx = func.context.as_ref()?.borrow_mut();
```
syrusakbary, line 126:
```suggestion
    let store_ctx = func.store().context;
    if store_ctx.is_none() {
        crate::error::update_last_error(wasm_store_t::CTX_ERR_STR);
    }
    let mut ctx = store_ctx.as_ref()?.borrow_mut();
```

lib/c-api/src/wasm_c_api/externals/function.rs
epilys: Because that's the point of the `Context` API, `Function` doesn't have a reference to the store because it's owned by the `Context`/store. `wasmer_api::Function` is just a handle (an integer) to the `context`'s vectors of objects.

syrusakbary (approved):

PR #2974, opened 2022-06-24
Context api tests

ptitSeb:
# Description
Migrate compiler tests and metering to the new Context API


lib/middlewares/src/metering.rs
```diff
@@ -418,21 +411,21 @@ mod tests {
...
         add_one.call(&mut ctx.as_context_mut(), 1).unwrap();
         assert_eq!(
-            get_remaining_points(ctx.as_context_mut(), &instance),
+            get_remaining_points(&mut ctx.as_context_mut(), &instance),
             MeteringPoints::Remaining(2)
         );
 
         // Third call fails due to limit
-        assert!(add_one.call(&mut ctx.as_context_mut(), 1).is_err());
+        assert!(add_one.call(1).is_err());
```
syrusakbary, line 426: Is this call valid?

ptitSeb: No. That's strange. Something must have gone wrong when I  rebased the branch over context-api.

epilys (changes requested):
Squash fixes with previous commits

syrusakbary (changes requested):
Let's change all the calls to `&mut ctx.as_context_mut()` to `&mut ctx` (and from `&ctx.as_context_ref()` to `&ctx`, if any of this exists)

ptitSeb:
> Squash fixes with previous commits

Done

ptitSeb:
> Let's change all the calls to `&mut ctx.as_context_mut()` to `&mut ctx` (and from `&ctx.as_context_ref()` to `&ctx`, if any of this exists)

Done

syrusakbary (approved):

PR #2976, opened 2022-06-26
Upgrade enumset minimum version to one that compiles

palfrey:
# Description

enumset dependency was on 1.0, but only [1.0.2 has a fix for compiling with newer syn](https://github.com/Lymia/enumset/blob/master/RELEASES.md#version-102-2021-01-25). This PR fixes that, and adds a makefile target for testing the minimal version scenario. It probably really should be added to the CI, but not sure where in the complicated setup here!

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,14 @@ Looking for changes that affect our C API? See the [C API Changelog](lib/c-api/C
 ## **Unreleased**
 
 ### Changed
-- #2946 Remove dylib,staticlib engines in favor of a single Universal engine
+- [#2946](https://github.com/wasmerio/wasmer/pull/2946) Remove dylib,staticlib engines in favor of a single Universal engine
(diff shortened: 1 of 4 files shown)


epilys:
Hello and thank you for your PR!

[Cargo actually chooses the latest patch version when you specify a dependency like that](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility) (i.e. `1.0`) and indeed you can see at our Cargo.lock it actually uses `1.0.11`. Have you had a problem with this dependency when using those crates on by themselves e.g. on crates.io?

palfrey:
The actual issue I'm trying to fix is https://github.com/swc-project/swc/issues/5041 but I'm having to dig through all its dependencies with this issue first!

The cargo semver stuff does the right thing most of the time, but sometimes someone has set a really strict version requirement somewhere else (and is compatible with the lower version) and then your code here will fail because the version listed doesn't actually work.

I agree that this isn't a case that crops up most of the time, and hence the thinking around double checking this in CI.

PR #2999, opened 2022-07-01
Support --invoke option for emscripten files without _start function

fschutt:
When running emscripten files, the CLI now supports the `--invoke` option to invoke a function by name.

# Review

### DIFF
--- a/lib/emscripten/src/lib.rs
+++ b/lib/emscripten/src/lib.rs
@@ -23,9 +23,9 @@ use std::f64;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex, RwLock};
 use wasmer::{
-    imports, namespace, Exports, Function, FunctionType, Global, Imports, Instance, LazyInit,
-    Memory, MemoryType, Module, Pages, RuntimeError, Store, Table, TableType, TypedFunction, Val,
-    ValType, WasmPtr, WasmerEnv,
+    imports, namespace, ExportError, Exports, Function, FunctionType, Global, Imports, Instance,
+    LazyInit, Memory, MemoryType, Module, Pages, RuntimeError, Store, Table, TableType,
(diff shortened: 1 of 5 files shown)


lib/emscripten/src/lib.rs
```diff
@@ -313,24 +313,51 @@ pub fn set_up_emscripten(instance: &mut Instance) -> Result<(), RuntimeError> {
...
+    instance: &Instance,
+    main_func_names: &[&'a str],
+) -> Result<&'a str, ExportError> {
+    let mut function_name = None;
+    let mut last_err = None;
+
+    for func_name in main_func_names.iter() {
+        match instance.exports.get::<Function>(func_name) {
+            Ok(_) => {
+                function_name = Some(func_name);
```
epilys, line 330: You can just return here immediately

lib/emscripten/src/lib.rs
```diff
@@ -384,8 +410,12 @@ pub fn run_emscripten_instance(
             .get(&ep)
             .map_err(|e| RuntimeError::new(e.to_string()))?;
         func.call(&[Val::I32(arg as i32)])?;
+    } else if let Ok(name) = emscripten_get_main_func_name(instance, &main_func_names) {
```
epilys, line 413: Ignoring the error case from `emscripten_get_main_func_name`; maybe it should return an Option then instead?

epilys (changes requested):

epilys:
@fschutt btw can you share the emscripten files you used for testing?

lib/emscripten/src/lib.rs
jcaesar: Isn't a Result preferable for a public library function?

fschutt: I'd stick with `Result` because `emscripten_call_main` also has a result. But I would document the function better.

PR #3001, opened 2022-07-02
Fix context capi ci errors

epilys:
Some `examples/` binaries fail on CI and I didn't notice because on my panopticon macos the binaries weren't allowed to run at all. Fixed them and now all test pass 🎉


epilys:
bors cancel

PR #3005, opened 2022-07-04
Refactored the WasiEnv object so that its cleaner

john-sharratt:
Refactored WasiEnv to turn Option<Memory> into Memory and to move the initialization code into the builder.


lib/api/src/js/module.rs
```diff
@@ -520,6 +520,49 @@ impl Module {
...
+    /// # }
+    /// ```
+    pub fn externs(&self) -> Result<Exports, InstantiationError> {
+        let exports = module
+            .exports()
+            .map(|export_type| {
+                let name = export_type.name();
+                let extern_type = export_type.ty().clone();
+                let js_export =
+                    js_sys::Reflect::get(&instance_exports, &name.into()).map_err(|_e| {
```
syrusakbary, line 551: Where is `instance_exports` defined? Does this code compile?

syrusakbary (commented):
Usually to get the memory you need to get the extern which requires the module to be instantiated (basically, you need an instance to then get the memory).

I'm a bit confused on how you are able to get the memory just with the module, before having the instance, and I'm not sure why this works (does it work?) at all. Could you give some extra  insight here @john-sharratt ?

Issue #2984, opened 2022-06-29
Add manual CI action for public Rust API changes


epilys:
Using https://github.com/Enselic/cargo-public-api

PR #3008, opened 2022-07-05
Add check-public-api.yaml workflow

fschutt:
WIP PR: Adds a new CI check that uses `cargo public-api` to track changes in the API between `master` and the last deployed version on crates.io.

Fixes https://github.com/wasmerio/wasmer/issues/2984

# Review

### DIFF
--- a/.github/workflows/check-public-api.yaml
+++ b/.github/workflows/check-public-api.yaml
@@ -1,0 +1,47 @@
+name: Check public API
+
+env:
+  RUST_BACKTRACE: 1
+on:
+  workflow_dispatch:
+    inputs:
+      compareMasterWithVersion:
+        required: false
+        description: 'Version to compare with (i.e. "1.0.2")'
+jobs:
+  setup:
+    name: Set up
+    runs-on: ubuntu-latest
+    outputs:
+      VERSION: ${{ steps.setup.outputs.VERSION }}
+      SHOULD_CHECK_API: ${{ steps.setup.outputs.SHOULD_CHECK_API }}
+    steps:
+      
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install latest nightly
+        uses: actions-rs/toolchain@v1
+        with:
+            toolchain: nightly
+            profile: minimal
+            override: true
+
+      - name: Generate API change report
+        shell: bash
+        run: |
+          cargo install cargo-public-api
+          CARGO_LATEST_VERSION = git tag --list | tail -n 1
+          LATEST_VERSION=${INPUT_COMPAREMASTERWITHVERSION:=$CARGO_LATEST_VERSION}
+          cargo public-api --manifest-path=lib/api/Cargo.toml --diff-git-checkouts $LATEST_VERSION master > diff.txt
+      
+      - name: Archive change report
+        uses: actions/upload-artifact@v3
+        with:
+          name: api-diff-report
+          path: |
+            diff.txt
+
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Looking for changes that affect our C API? See the [C API Changelog](lib/c-api/C
 
 ### Changed
 - [#3017](https://github.com/wasmerio/wasmer/pull/3017) Fix typo in README.md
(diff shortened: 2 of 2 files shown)


.github/workflows/check-public-api.yaml
```diff
@@ -0,0 +1,52 @@
+name: Check public API
+
+env:
+  RUST_BACKTRACE: 1
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
```
epilys, line 9: We don't want this to break on every commit, better remove the on push and on pull_request events and put workflow_dispatch (that's how github calls manual actions)

fschutt: Yes, this is a draft PR, this part will be removed later on, I just needed to trigger it once in the PR.

Also I need it to be `master`, not `main`

fschutt:
@epilys I have no idea why this workflow doesn't run, works 1:1 for me on my local machine.

fschutt:
So the workflow runs now @epilys - https://github.com/wasmerio/wasmer/actions/runs/2649376904 contains the artifact containing the auto-generate API diff.

Now that this works, I'd change the workflow to only get triggered on `vX.Y.Z` release tags.

.github/workflows/check-public-api.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+env:
+  RUST_BACKTRACE: 1
+on:
+  push:
+    branches:
+      - 'master'
+    tags:
+      # this is _not_ a regex, see: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
+      - '[0-9]+.[0-9]+.[0-9]+*'
```
epilys, line 11: This should probably be removed, if a tag has been pushed to master by that point it's "too late". We should think about how to integrate this manual check into the release workflow.

epilys (changes requested):

.github/workflows/check-public-api.yaml
```diff
@@ -0,0 +1,59 @@
+name: Check public API
+
+env:
+  RUST_BACKTRACE: 1
+on:
+  push:
+    tags:
+      # this is _not_ a regex, see: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
+      - '[0-9]+.[0-9]+.[0-9]+*'
```
epilys, line 9:
```suggestion
```

.github/workflows/check-public-api.yaml
```diff
@@ -0,0 +1,59 @@
...
+        id: setup
+        shell: bash
+        run: |
+          VERSION=${GITHUB_REF/refs\/tags\//}
+          echo ::set-output name=VERSION::${VERSION}
+          SHOULD_CHECK_API=$(echo $VERSION | grep -c '^[0-9]\+\.[0-9]\+\.[0-9]\+\(-\([a-zA-Z]\+\)\?[0-9]*\)\?$' || true)
+          echo ::set-output name=SHOULD_CHECK_API::${SHOULD_CHECK_API}
+          echo $VERSION
+          echo $SHOULD_CHECK_API
+
```
epilys, line 34:
```suggestion
```

.github/workflows/check-public-api.yaml
```diff
@@ -0,0 +1,59 @@
...
+            toolchain: nightly
+            profile: minimal
+            override: true
+
+      - name: Generate API change report
+        run: |
+          cargo install cargo-public-api
+          cargo install ripgrep
+          LATEST_VERSION=$(cargo search wasmer | rg --only-matching --replace '$1' 'wasmer = "(.*)"')
+          cargo public-api --manifest-path=lib/api/Cargo.toml --diff-git-checkouts $LATEST_VERSION master > diff.txt
```
epilys, line 50: Maybe make `LATEST_VERSION` a workflow input? And if it's not specified make the default the last tagged release, which you can get with `git tag --list | tail -n 1` which doesn't make `rg` necessary.

.github/workflows/check-public-api.yaml
fschutt: I removed it in the latest changeset.

epilys (approved):
Nice!

PR #3018, opened 2022-07-14
Context api refactor and capi

syrusakbary:
This PR improves the C-API in the context API refactor. Some tests are passing, but not all.

We need now to polish it @epilys


lib/c-api/tests/wasm-c-api/example/callback.c
```diff
@@ -50,9 +48,9 @@ own wasm_trap_t* print_callback(
 
 // A function closure.
 own wasm_trap_t* closure_callback(
-  const wasm_val_vec_t* args, wasm_val_vec_t* results
+  void* env, const wasm_val_vec_t* args, wasm_val_vec_t* results
 ) {
-  int i = *(int*) wasm_store_data_get(global_store);
+  int i = *(int*)env;
```
syrusakbary, line 53: We can't modify the wasm-c-api example

PR #3032, opened 2022-07-21
C-API: add functions to overwrite stdin / stdout / stderr handlers

### LINKED ISSUE #2334: [c-api] add more functions for stdin/stderr/stdout control
### Motivation

Currently the c-api propose a limited set of possibilities when it comes to WASI and redirection over stdin/stderr/stdout.
As I understand it, the `wasi_config_t` will always inherit `stdin` from the host program (see https://github.com/wasmerio/wasmer/pull/2059).

### Proposed solution

Since the `wasi_config_t` has access to a `WasmStateBuilder`, one could replace any of the program pipes with a valid `WasiFile`. This could be a `HostFile` or a `Pipe`. `stderr` and `stdout` can already be captured using the `wasi_config_capture_{stdout,stderr}` functions, but there is no way to replace the inherited `stdin`.

A function could be added to replace the stdin from the `WasiStateBuilder` with a `Pipe`, example:

https://github.com/wasmerio/wasmer/blob/4b24038efe04a09419ac5754ab903b5acf961570/lib/c-api/src/wasm_c_api/wasi/mod.rs#L156-L160

A function could be added to then write to this stdin (with less unwraps):

https://github.com/wasmerio/wasmer/blob/4b24038efe04a09419ac5754ab903b5acf961570/lib/c-api/src/wasm_c_api/wasi/mod.rs#L277-L288

This would allow the instantiation of the same module with different inputs. The code above is from [this commit](https://github.com/Pangoraw/wasmer/commit/4b24038efe04a09419ac5754ab903b5acf961570) on my fork, I can open a PR if the proposed change is accepted.

### Alternatives

Since any `WasiFile` is valid, one could also provide file-backed outputs and inputs, similar to wasmtime's [`wasi_config_set_stdin_file(wasi_config_t *config, const char *path)`](https://docs.wasmtime.dev/c-api/wasi_8h.html#adc6be10a42fb437e3ce014d248550018)

### Additional context

An example usage where I control the stdin input given to cowsay.wasm using a wrapper written in Julia:

image

epilys:
Fixes #2334

# Review

- [x] Add functional test for this **(important!)**


fschutt:
@epilys as you requested, I separated the fixes from clippy and `make lint` into a new PR, which can be merged after this PR has been merged. The other tests should work, so this PR can be merged now.

[patches.zip](https://github.com/wasmerio/wasmer/files/9267750/patches.zip)

epilys:
@syrusakbary this needs a manual merge as well

syrusakbary (changes requested):
I think we may need a meeting to discuss this further. I see as an issue that we are operating just in terms of specific structs (STDIN, STDOUT, STDERR), rather than in terms of pipes.

I'd like the API to be pipe based, as I think it will be more resilient also for other kind of filesystem things. Let's discuss

fschutt:
You can store pipe objects inside of the structs, the current approach is much more flexible than locking the approach to pipes. I could add code to add a `wasi_new_console_out_pipe()` to use pipes, if you want, but I wouldn't change the overall API.

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -10,20 +10,617 @@ use super::{
...
+}
+
+extern "C" fn wasi_console_out_delete_null(_: *const c_void, _: usize, _: usize) -> i64 {
+    0
+}
+
+unsafe extern "C" fn wasi_console_out_read_memory(
+    ptr: *const c_void,    /* = *Pipe */
+    sizeof: usize,         /* = sizeof(Pipe) */
+    alignof: usize,        /* = alignof(Pipe) */
```
syrusakbary, line 353: We can delete this

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -10,20 +10,617 @@ use super::{
...
+) -> i64 {
+    0
+}
+
+extern "C" fn wasi_console_out_delete_null(_: *const c_void, _: usize, _: usize) -> i64 {
+    0
+}
+
+unsafe extern "C" fn wasi_console_out_read_memory(
+    ptr: *const c_void,    /* = *Pipe */
```
syrusakbary, line 351: Why don't we have directly: `ptr: *mut Pipe`

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -10,20 +10,617 @@ use super::{
...
+
+/// Creates a new `wasi_console_out_t` which uses a memory buffer
+/// for backing stdin / stdout / stderr
+#[no_mangle]
+pub unsafe extern "C" fn wasi_console_out_new_memory() -> *mut wasi_console_out_t {
+    use std::mem::ManuallyDrop;
+
+    let data = Pipe::new();
+    let mut data = ManuallyDrop::new(data);
+    let ptr: &mut Pipe = &mut data;
```
syrusakbary, line 459: This can be improved

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -10,20 +10,538 @@ use super::{
...
+unsafe extern "C" fn wasi_console_out_delete_memory(ptr: *const c_void /* = *Pipe */) -> i64 {
+    let ptr = ptr as *const Pipe;
+    let _: Pipe = std::mem::transmute_copy(&*ptr); // dropped here, destructors run here
+    0
+}
+
+/// Creates a new `wasi_console_out_t` which uses a memory buffer
+/// for backing stdin / stdout / stderr
+#[no_mangle]
+pub unsafe extern "C" fn wasi_console_out_new_memory() -> *mut wasi_console_out_t {
```
syrusakbary, line 376: `wasi_pipe_new`

PR #3034, opened 2022-07-22
Implemented multi-threading for both JS and SYS, plus other WASIX implementations

john-sharratt:
# Description

# Review


lib/api/src/sys/externals/memory.rs
```diff
@@ -72,47 +72,6 @@ impl Memory {
...
-    }
-
-    /// Returns the size (in bytes) of the `Memory`.
-    pub fn data_size(&self, store: &impl AsStoreRef) -> u64 {
-        self.buffer(store).len.try_into().unwrap()
-    }
-
-    /// Retrieve a slice of the memory contents.
-    ///
-    /// # Safety
```
syrusakbary, line 91: We shouldn't modify the API of non-shared memory. We should either:
* create a new `SharedMemory` that is able to lock
* preserve the current memory API and depending if the memory is shared o no, to work on the MutexGuard or not

lib/api/src/js/instance.rs
```diff
@@ -19,8 +23,10 @@ use std::fmt;
 #[derive(Clone)]
 pub struct Instance {
     _handle: StoreHandle<WebAssembly::Instance>,
+    thread_seed: Arc<AtomicU32>,
```
syrusakbary, line 26: What is this for?

lib/api/src/js/imports.rs
```diff
@@ -130,7 +131,11 @@ impl Imports {
     /// Resolve and return a vector of imports in the order they are defined in the `module`'s source code.
     ///
     /// This means the returned `Vec<Extern>` might be a subset of the imports contained in `self`.
-    pub fn imports_for_module(&self, module: &Module) -> Result<Vec<Extern>, InstantiationError> {
+    pub fn imports_for_module(
+        &self,
+        _store: &mut impl AsStoreMut,
+        module: &Module
```
syrusakbary, line 137: Why do we need to nave a new `_store` param, if we are not using it?

lib/api/src/sys/externals/memory.rs
```diff
@@ -172,61 +131,13 @@ impl Memory {
...
+    /// In order to access the memory you must first lock it
+    /// (this will prevent it from growing)
+    pub fn lock<'a>(&'a self, store: &'a impl AsStoreRef) -> MemoryGuard<'a> {
+        let view = self.handle.get(store.as_store_ref().objects()).vmmemory();
+        let def = unsafe { view.as_ref() };
+        MemoryGuard {
+            _view: view,
             base: def.base,
             len: def.current_length,
             marker: PhantomData,
```
syrusakbary, line 143: I't like to name this `buffer` and return a `ArrayBuffer` (which is similar as a memory view) in case memory is not shared, and return a `SharedArrayBuffer` in case it's shared

lib/vm/src/table.rs
```diff
@@ -79,6 +81,31 @@ pub struct VMTable {
...
+
+    /// Returns the number of allocated elements.
+    pub fn size(&self) -> u32 {
+        unsafe {
+            let td_ptr = self.get_vm_table_definition();
+            let td = td_ptr.as_ref();
+            td.current_elements
+        }
+    }
+}
```
syrusakbary, line 99: Why do we need to protect the table at all?

lib/api/src/sys/externals/memory.rs
syrusakbary: Similarly to JS:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer

lib/api/src/sys/function_env.rs
```diff
@@ -40,16 +40,18 @@ impl<T> FunctionEnv<T> {
...
     where
         T: Any + Send + 'static + Sized,
     {
         self.handle
             .get_mut(store.objects_mut())
             .as_mut()
-            .downcast_mut::<T>()
-            .unwrap()
+            .map(|a| a.downcast_mut::<T>())
+            .flatten()
```
syrusakbary, line 54: I think `.unwrap().unwrap()` will be faster to execute?

lib/api/src/sys/function_env.rs
```diff
@@ -29,7 +29,7 @@ impl<T> FunctionEnv<T> {
     }
 
     /// Get the data as reference
-    pub fn as_ref<'a>(&self, store: &'a impl AsStoreMut) -> &'a T
+    pub fn as_ref<'a>(&self, store: &'a impl AsStoreRef) -> &'a T
```
syrusakbary, line 32: Ooops, thanks for the fix!

lib/api/src/js/instance.rs
```diff
@@ -19,8 +23,10 @@ use std::fmt;
...
     module: Module,
-    #[allow(dead_code)]
+    /// Reactors are used to wake up the program
+    pub reactors: Reactors,
```
syrusakbary, line 29: Shouldn't the programs use SharedArrayBuffer to wake up? (I might be confusing things, so it would be good to have a bit of better details on for which cases `Reactors` are used)

lib/api/src/sys/mem_access.rs
```diff
@@ -146,9 +146,9 @@ impl<'a, T: ValueType> fmt::Debug for WasmRef<'a, T> {
 ///
 /// This wrapper safely handles concurrent modifications of the data by another
 /// thread.
-#[derive(Clone, Copy)]
+#[derive(Clone)]
 pub struct WasmSlice<'a, T: ValueType> {
-    buffer: MemoryBuffer<'a>,
+    view: MemoryGuard<'a>,
```
syrusakbary, line 151: I'm not super sure about this changes. Once we have two kinds of Buffer (non shared and shared), I think this might need to vary more

lib/api/src/js/instance.rs
john-sharratt: The thread ID is abstracted so that real thread ID's are not leaked into the sandbox (security considerations) - this seed is used to generate incrementing threads - it could be moved into WasiEnv but then threading starts to become a WASI only feature while I think its likely other things will use threading (unless we just use WASI code for none WASI ABI's)

lib/api/src/sys/externals/memory.rs
john-sharratt: SharedMemory would still require guards that abstract the access to the memory where none shared memory could be made to be a zero cost abstraction it will still mandate some changes in the way Wasmer API accesses memory. So the main difference becomes if they implement the lock inside or not. Also we would need to figure out a way to decide when to use SharedMemory and when not to use it. With WASI we don't know if its multithreaded or not upfront until its actually used (having import memory is not conclusive as one may wish to pass in memory but be singlethreaded). Hence in the WASI mode we would have to assume all programs are multithreaded and hence need SharedMemory. Beyond this there is no technical limitation to implementing what you say and can be made optional on the "enable_threads" argument

lib/api/src/js/imports.rs
john-sharratt: The store is needed here (for sys) - JS doesn't use it but I kept the interface consistent.
https://github.com/wasmerio/wasmer/blob/af5663beadb9c827841f52730bb90132bb37e910/lib/api/src/sys/imports.rs#L149

It's used for the situation that imported memory is needed (multithreading) - it will create the memory

lib/api/src/sys/externals/memory.rs
john-sharratt: Sure no worries - makes sense

lib/vm/src/table.rs
john-sharratt: There are race conditions that need to be protected against because the table was converted to a COW (Copy-on-write) pattern. In normal single-threaded code this will be close to a zero cost abstraction but for multi-threaded code it will copy the VMTable before modifying it. For performance reasons I implemented the copying as COW to make it a shallow copy rather than a deep copy as I assume that a large number of the tables do not change between threads

lib/api/src/sys/function_env.rs
john-sharratt: It probably compiles to the same assembly code - I can change it if you like though no worries

lib/api/src/js/instance.rs
john-sharratt: That would make the scope of a SharedArrayBuffer functionality wider then...well shared buffers. We should try and follow the "separation of concerns" principle when we can. Do you think we need to create a superset object/class? and if so what would you name it?

The reactors are needed because WASIX provides reactors (code execution stops until work arrives to be done) - this is essential for building asynchronous programs (for instance with Tokio). This will also be important for low overhead density of web assembly applications running on the same physical server, if there are web assembly programs running that are just in a spin lock / futex waiting on syscall to finish (e.g. poll) then it will not be possible to push them into a warm state (loaded but no CPU cycles)

lib/api/src/sys/externals/memory.rs
john-sharratt: Drastically simplified the changes around memory much closer to the original code. I had made an incorrect assumption which means the multithreading risks are much lower than I thought thus requiring much less protection. Wasmer uses mmap which means it does not change the base pointer address during grow operations, further it never shrinks in size... with those two constraints together it means there is no need to protect against a grow operation in parallel with normal memory operation, we only need to protect against concurrent grow operations which is only in one small part of the code. A simple RwLock around mmap is thus all that's needed.

lib/api/src/sys/mem_access.rs
epilys: the threading extension proposes a new Shared Memory object. We will need two kinds of buffer, shared and not shared.

lib/api/src/common/reactors.rs
```diff
@@ -0,0 +1,88 @@
+use std::collections::VecDeque;
+use std::sync::Mutex;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering;
+use std::sync::mpsc;
+use std::sync::Arc;
+use std::task::Waker;
+use std::time::Duration;
```
syrusakbary, line 8: This doesn't need to live inside of Wasmer. It can be a side utility/module inside of wasmer-wasi atm (since it's only used there)

lib/api/src/js/thread.rs
```diff
@@ -0,0 +1,44 @@
+use std::{sync::{Arc, mpsc, Mutex}, time::Duration};
+
```
syrusakbary, line 2: This doesn't need to live inside of Wasmer. It can be a side utility/module inside of wasmer-wasi atm (since it's only used there)

syrusakbary: Or we can move it to WASI

john-sharratt:
This is now replaced by...
https://github.com/wasmerio/wasmer/pull/3116

lib/api/src/sys/externals/memory.rs
john-sharratt: This is now implemented with VMSharedMemory instead and preserves the original API where possible

lib/api/src/js/instance.rs
john-sharratt: this is now moved away from Instance and put into WasiStateThreading instead

lib/api/src/js/imports.rs
john-sharratt: This is no longer needed and is removed from the new patch I made

lib/api/src/sys/externals/memory.rs
john-sharratt: this is no longer needed as the changes were simplified

lib/vm/src/table.rs
john-sharratt: the VMTable does not need to be multithread safe anymore so this was removed - instead we follow closer the threads proposal and make a new instance instead

lib/api/src/js/instance.rs
john-sharratt: I moved this away from the instance and put it in Wasi instead - its much cleaner now

lib/api/src/sys/mem_access.rs
john-sharratt: Now implemented it with Shared and Owned memory objects closer to the threading proposal

lib/api/src/common/reactors.rs
john-sharratt: I have indeed now moved this to wasmer-wasi instead

lib/api/src/js/thread.rs
john-sharratt: This is now moved to WASI

PR #3038, opened 2022-07-25
Re-introduce create-exe to wasmer-cli v3.0

syrusakbary:
### Usage

```shell
$ make && make package-capi
...
$ export WASMER_DIR=`pwd`/package
$  ./target/release/wasmer create-exe qjs.wasm -o qjs   
Compiler: cranelift
Target: x86_64-unknown-linux-gnu
Using path `[..]/wasmer/package/lib/libwasmer.a` as libwasmer path.
✔ Native executable compiled successfully to `qjs`.
$ ./qjs
QuickJS - Type "\h" for help
qjs > ^C
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1911,9 +1911,9 @@ dependencies = [
 
 [[package]]
 name = "rkyv"
-version = "0.7.38"
+version = "0.7.39"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "517a3034eb2b1499714e9d1e49b2367ad567e07639b69776d35e259d9c27cca6"
+checksum = "cec2b3485b07d96ddfd3134767b8a447b45ea4eb91448d0a35180ec0ffd5ed15"
 dependencies = [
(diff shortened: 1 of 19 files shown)


lib/api/src/sys/module.rs
```diff
@@ -149,7 +149,12 @@ impl Module {
     /// the WebAssembly text format (if the "wat" feature is enabled for
     /// this crate).
     pub fn from_binary(store: &impl AsStoreRef, binary: &[u8]) -> Result<Self, CompileError> {
-        Self::validate(store, binary)?;
+        if let Err(err) = Self::validate(store, binary) {
+            if let Ok(module) = unsafe { Self::deserialize(store, binary) } {
+                return Ok(module);
+            }
```
syrusakbary, line 155: From binary should only take `.wasm` / `wat` formats (not the serialized artifact of Wasmer/Engine).

epilys: Leftover from a refactor, replaced it

PR #3057, opened 2022-08-01
wasmer-cli: create-obj command

epilys:


lib/cli/src/c_gen/staticlib_header.rs
```diff
@@ -0,0 +1,298 @@
...
+        current_offset += sizeof(dynamic_function_trampoline_pointers);
+
+        wasm_byte_vec_t module_byte_vec = {
+                .size = buffer_size,
+                .data = memory_buffer,
+        };
+        return module_byte_vec;
+}
+
+wasm_module_t* wasmer_staticlib_engine_new(wasm_store_t* store, const char* wasm_name) {
```
syrusakbary, line 61: Should we have this named differently? (`wasmer_new_module_from_staticlib`)

lib/compiler/Cargo.toml
```diff
@@ -43,7 +44,7 @@ default = ["std", "enable-serde" ]
 # `CompilerConfig`, as well as the included wasmparser.
 # Disable this feature if you just want a headless engine.
 translator = ["wasmparser"]
-compiler = ["translator"]
+compiler = ["translator", "wasmer-object"]
```
syrusakbary, line 47: Why compiler requires `wasmer-object`? Should we have a new flag?

lib/compiler/src/engine/artifact.rs
```diff
@@ -26,7 +41,15 @@ use wasmer_vm::{FunctionBodyPtr, MemoryStyle, TableStyle, VMSharedSignatureIndex
 use wasmer_vm::{InstanceAllocator, InstanceHandle, StoreObjects, TrapHandlerFn, VMExtern};
 
 /// A compiled wasm module, ready to be instantiated.
-pub struct Artifact {
+pub enum Artifact {
+    /// FIXME
+    Universal(UniversalArtifact),
+    /// FIXME
+    Static(StaticArtifact),
```
syrusakbary, line 48: I don't think this is the optimal way of doing this. Basically both of them shared the finished functions, signatures and so on.

The only different thing is `frame_info_registration` and `_symbol_registry`. But that's only needed when using loading the artifact. In case of `frame_info_registration` is already optional, and perhaps we can make _symbol_registry is only needed at loading time (not really needed to store it)

lib/compiler/Cargo.toml
epilys: Might be a mistake, will recheck

lib/compiler/src/engine/artifact.rs
epilys: No it isn't the optimal way, but it was the fastest way to get it working, next step is making it prettier as you say

syrusakbary (approved):

syrusakbary:
Once lint is fixed we should be good to merge

Issue #2847, opened 2022-04-15
test wast::spec::unreachable::cranelift::universal fails on aarch64-apple-darwin target


epilys:
### Describe the bug

Starting from commit 4c946257cfe8223fc75f9411fe6856bc0b79d42c, test `wast::spec::unreachable::cranelift::universal` fails on M1 macos:

```sh
$ echo "`rustc -V` | `uname -rm`"
rustc 1.59.0 (9d1b2106e 2022-02-23) | 21.4.0 arm64
```

```shell
running 1 test                                                                                                                                                        
test wast::spec::unreachable::cranelift::universal ... FAILED                                                                                                         
                                                                                                                                                                      
failures:                                                                                                                                                             
                                                                                                                                                                      
---- wast::spec::unreachable::cranelift::universal stdout ----                                                                                                        
Running wast `tests/wast/spec/unreachable.wast`                                    
Error: Failed directives on tests/wast/spec/unreachable.wast:                      
  • expected 'unreachable', got 'RuntimeError: call stack exhausted' (231:1)       
                                                                                                                                                                      
thread 'wast::spec::unreachable::cranelift::universal' panicked at 'assertion failed: `(left == right)`                                                               
  left: `1`,                                                                                                                                                          
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library
/test/src/lib.rs:186:5                                                             
stack backtrace:                                                                   
   0: rust_begin_unwind                                                                                                                                               
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/std/src/panicking.rs:498:5                                                                    
   1: core::panicking::panic_fmt                                                                                                                                      
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/panicking.rs:116:14                                                                  
   2: core::panicking::assert_failed_inner                                                                                                                            
   3: core::panicking::assert_failed                                                                                                                                  
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/panicking.rs:154:5
   4: test::assert_test_result                                                     
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/test/src/lib.rs:186:5                                                                         
   5: compilers::wast::spec::unreachable::cranelift::universal::{{closure}}        
[...]

PR #3058, opened 2022-08-01
Fix trap tracking

ptitSeb:
# Description
Ajusted how trap tracking information are retrieve at runtime, fixing cranelift+aarch64 trap test like #2847 

There are still issues on aarch64+macos with the pretty print of trap.


lib/compiler/src/engine/trap/frame_info.rs
```diff
@@ -207,7 +203,7 @@ pub fn register(
     ) in finished_functions.iter()
     {
         let start = **start as usize;
-        let end = start + len;
+        let end = start + len - 1;
```
syrusakbary, line 206: Can we add a code comment here making explicit why is `-1` so we remember it easily for the future?

syrusakbary (approved):

PR #3059, opened 2022-08-01
Add support for packing PiritaFiles into standalone .exe

fschutt:
If the file given to `wasmer create-exe [file]` is a PiritaFile, the resulting exe will contain all compiled atoms and filesystems and run as a standalone binary.

#### TODO

- [ ] Omit linker errors with duplicated functions in object files
- [ ] Write unit tests


syrusakbary:
Note for the future, before merging the PR (optimization): we’ll need to remove the Wasm atoms from the generated executable

fschutt:
@syrusakbary The atoms are already removed from the filesystem. The atoms are in a separate volume, there aren't part of the filesystem volumes.

Issue #3025, opened 2022-07-19
Path bytes returned from `wasi::fd_prestat_dir_name` appended by `0`


Mo-Fatah:
### Describe the bug
I used [this wasi](https://docs.rs/wasi/latest/wasi/) crate to compare the behavior of `WASI` implementations between `wasmtime` and `wasmer`. 
 Initially, I used [wasi::fd_prestat_get](https://docs.rs/wasi/0.11.0+wasi-snapshot-preview1/wasi/fn.fd_prestat_get.html) and [wasi::fd_prestat_dir_name](https://docs.rs/wasi/0.11.0+wasi-snapshot-preview1/wasi/fn.fd_prestat_dir_name.html) .

Apparently all path/dir/file names (in bytes) returned from `fd_prestat_dir_name` are appended by `0`. 

I am using 
```
wasmer 2.3.0 | rustc 1.61.0 (fe5b13d68 2022-05-18) | x86_64
```

### Steps to reproduce
I wrote a simple program [here](https://github.com/Mo-Fatah/wasi-testsuite/tree/main/engines-diff/end-of-path) that prints either `wasmer` or `wasmtime` based on the directory name returned from `fd_prestat_dir_name`.

1- The program takes an environment variable which is the directory name , this will be [arg](https://github.com/Mo-Fatah/wasi-testsuite/blob/13ccd4d9ebd6d326dcb487ebadf2917088c07223/engines-diff/end-of-path/src/main.rs#L9)

2- Then, I used `fd_prestat_get` and `fd_prestat_dir_name` to access the directory passed to the engine through `WASI` interface which will be stored in [dst](https://github.com/Mo-Fatah/wasi-testsuite/blob/13ccd4d9ebd6d326dcb487ebadf2917088c07223/engines-diff/end-of-path/src/main.rs#L27) variable.

3- I compared the dir name bytes which passed to the program (`arg`) against the name bytes returned from the `WASI` interface
https://github.com/Mo-Fatah/wasi-testsuite/blob/13ccd4d9ebd6d326dcb487ebadf2917088c07223/engines-diff/end-of-path/src/main.rs#L37-L46
 
### Expected behavior
Assuming that the directory name is `scratch_dir` and it is located in the root of the program.
- Executing with `wasmtime` by 
  `wasmtime --dir scratch_dir ./target/wasm32-wasi/debug/end-of-path.wasm scratch_dir` 
the `arg` bytes and `dst` bytes will be exactly the same (as they should be)
```
[...]

PR #3061, opened 2022-08-02
Removed trailing zero in WASI::fd_prestat_dir_name name return (for #3025)

ptitSeb:
# Description
Removed trailing zero in WASI::fd_prestat_dir_name returned name (for #3025)


epilys:
[Posix defines character strings as null terminated](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_276), personally I'd like all path strings that go out into a module to be null terminated. What do you think?

ptitSeb:
I agree, but this is WASI here, not POSIX. check the ticket mentioned for more detail. Not sure the change should be done. But if we don't it, we should close the mentioned ticket as a "will not be done"

lib/wasi/src/syscalls/mod.rs
```diff
@@ -959,13 +959,11 @@ pub fn fd_prestat_dir_name<M: MemorySize>(
...
     match guard.deref() {
         Kind::Dir { .. } | Kind::Root { .. } => {
-            // TODO: verify this: null termination, etc
             let path_len: u64 = path_len.into();
-            if (inode_val.name.len() as u64) < path_len {
+            if (inode_val.name.len() as u64) <= path_len {
                 wasi_try_mem!(path_chars
                     .subslice(0..inode_val.name.len() as u64)
                     .write_slice(inode_val.name.as_bytes()));
-                wasi_try_mem!(path_chars.index(inode_val.name.len() as u64).write(0));
```
syrusakbary, line 968: We might still want to terminate it by 0 (but just not pass it to the user).
Basically, all the PR as is but, reverting this to what it was:
```
                wasi_try_mem!(path_chars.index(inode_val.name.len() as u64).write(0));
```

Thoughts?

ptitSeb: we might overwite memory here! The memory is allocated client side, for the result of `prestat_fd`, where the 0 is not counted anymore on this PR.
We cannot have both. It with or without the appended 0

syrusakbary: Ok!

syrusakbary (approved):

Issue #3033, opened 2022-07-21
The execution result of the program differs from that of the native program.


abc767234318:
### Describe the bug

```
#include <dirent.h>
#include <stdio.h>
#include <errno.h>

int main(int argc, char **argv) {
  DIR *d;

  char * target = ".";
  if (argc == 2) {
      target = argv[1];
  }
  struct dirent *dir;
  d = opendir(target);
  if (d) {
    while ((dir = readdir(d)) != NULL) {
      printf("%s\n", dir->d_name);
    }

    printf("errno: %d\n", errno);
    closedir(d);
  }

  return(0);
}
```
This is a source code first mentioned in wasmtime issues NO.2493.
My project files are as follows:
[wasmtime-2493.zip](https://github.com/wasmerio/wasmer/files/9159635/wasmtime-2493.zip)

```sh
echo "`wasmer -V` | `rustc -V` | `uname -m`"
wasmer 2.3.0 | rustc 1.62.0 (a8314ef7d 2022-06-27) | x86_64
```

### Steps to reproduce
```
wasmer run ls.wasm --dir=./testfolder -- testfolder | wc -l
```

### Expected behavior
When I compile using g++, the program results in the following
```
203
```

### Actual behavior
```
201
```

syrusakbary:
This indeed seems like a bug, this is how wasmtime fixed it:
https://github.com/bytecodealliance/wasmtime/issues/2493

ptitSeb:
Should be fixed in current master, can you check @abc767234318 so the ticket can be closed?

ptitSeb:
Fixed on my side, closing the ticket now.

PR #3065, opened 2022-08-02
Added '.' and '..' special folder t WASI fd_readdir return (for #3033)

ptitSeb:
# Description
Simply adding "." and ".." to the returnned list from an WASI::fd_readdir call

### DIFF
--- a/lib/wasi/src/syscalls/mod.rs
+++ b/lib/wasi/src/syscalls/mod.rs
@@ -1321,6 +1321,10 @@ pub fn fd_readdir<M: MemorySize>(
                             (entry.name.to_string(), stat.st_filetype, stat.st_ino)
                         }),
                 );
+                // adding . and .. special folders
+                // TODO: inode
+                entry_vec.push((".".to_string(), __WASI_FILETYPE_DIRECTORY, 0));
+                entry_vec.push(("..".to_string(), __WASI_FILETYPE_DIRECTORY, 0));
                 entry_vec.sort_by(|a, b| a.0.cmp(&b.0));
                 entry_vec
             }


syrusakbary:
This should fix #3033

lib/wasi/src/syscalls/mod.rs
```diff
@@ -1321,6 +1321,10 @@ pub fn fd_readdir<M: MemorySize>(
                             (entry.name.to_string(), stat.st_filetype, stat.st_ino)
                         }),
                 );
+                // adding . and .. special folders
+                // TODO: inode
+                entry_vec.push((".".to_string(), __WASI_FILETYPE_DIRECTORY, 0));
+                entry_vec.push(("..".to_string(), __WASI_FILETYPE_DIRECTORY, 0));
```
syrusakbary, line 1327: Ummm... what about the root? We shouldn't add `..` to readdir in the root.
Can you also add a test case for this? Thanks

ptitSeb: But, isn't Root the next `match` here?

syrusakbary: Also, side comment. Do you have insight on why our fix is quite different from wasmtime's one?

https://github.com/bytecodealliance/wasmtime/pull/2494/files#diff-df2ef1370349b8385454adfabfe871fc51fd3c12debc7b0691b98ca19c5724deR309-R335

ptitSeb: What I hesitated is adding "." to a root readdir

ptitSeb: the root cause was completly different: it's not a buffer size issue here, it's the missing "." and "..". All the 201 folder appears correctly, just missing the special folders

lib/wasi/src/syscalls/mod.rs
```diff
@@ -1321,6 +1321,10 @@ pub fn fd_readdir<M: MemorySize>(
                             (entry.name.to_string(), stat.st_filetype, stat.st_ino)
                         }),
                 );
+                // adding . and .. special folders
+                // TODO: inode
```
syrusakbary, line 1325: What is this TODO?

ptitSeb: same as line 1310

lib/wasi/src/syscalls/mod.rs
ptitSeb: So no, adding "." to root is a bad idea, it breaks things.

syrusakbary (approved):

Issue #3066, opened 2022-08-02
Documentation for wasmer 3 fails to build for docs.rs


silwol:
### Describe the bug
https://docs.rs/ contains documentation built for all crates that are published on https://crates.io/. Their infrastructure attempts to build the documentation, but that failed for wasmer 3.0.0-alpha.4. I consider API documentation availability on https://docs.rs/ very important, as it's the go-to location for many who want to develop something against a crate API. Therefore we should get it working with one of the 3.0 pre-releases in order to be certain we have a proper docs.rs rendering for the final release.

### Additional context
The log of the failed build can be found [here](https://docs.rs/crate/wasmer/3.0.0-alpha.4/builds/601770). Pasting here because I don't know anything about retention of the logs at https://docs.rs/
```
# rustc version
rustc 1.64.0-nightly (9067d5277 2022-07-28)
# docs.rs version
docsrs 0.6.0 (f9ffabd 2022-07-24)

# build log
[...]

PR #3069, opened 2022-08-04
Remove native feature entry from docs.rs metadata

silwol:
The native feature is no longer present, but was still in the list of
features to be enabled when building the documentation for docs.rs,
causing the build there to fail. Removal should re-enable the build of
the documentation for docs.rs.

Closes: #3066

### DIFF
--- a/lib/api/Cargo.toml
+++ b/lib/api/Cargo.toml
@@ -126,4 +126,4 @@ static-artifact-load = ["wasmer-compiler/static-artifact-load"]
 static-artifact-create = ["wasmer-compiler/static-artifact-create"]
 
 [package.metadata.docs.rs]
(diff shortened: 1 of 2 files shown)


epilys (approved):
Good,

It'd also be a good idea to add a new commit adding these new features in the docs:

```
wasmer-artifact-load 
wasmer-artifact-create 
static-artifact-load 
static-artifact-create
```

silwol:
> It'd also be a good idea to add a new commit adding these new features in the docs:

These reference the respective feature with the same name in `wasmer-compiler`, I assume it makes sense to enable them inside that crate as well, otherwise we might attempt reference types inside `wasmer-compiler` that are not rendered there.

epilys (approved):

PR #3076, opened 2022-08-05
Add support for cross-compiling in create-exe with zig cc WIP

### LINKED ISSUE #3042: Add support for Zig cc when compiling with `create-exe`
We should add support for `zig cc` when compiling (if zig is available) instead of using clang, as it's better for cross-compilation.

epilys:
zig version must be at least 0.10.0-dev.3431+4a4f3c50c

Closes #3042

# Description

# Review


lib/cli/src/commands/create_exe.rs
```diff
@@ -446,3 +534,195 @@ impl LinkCode {
...
+                } else {
+                    false
+                }
+            });
+            assets.retain(|a| {
+                if let Some(name) = a["name"].as_str() {
+                    match target_triple.vendor {
+                        wasmer_types::Vendor::Apple => !name.contains("windows"),
+                        wasmer_types::Vendor::Pc => !name.contains("apple"),
+                        _ => true,
```
syrusakbary, line 607: Ummm... this seems somewhat incorrect?

What are we computing here? It's a bit confusing for me

epilys: It is confusing yeah. Apple vendor does not necessarily mean that the given triple includes the vendor name, because `zig` has its own triple format and it doesn't include the vendor (for example `x86_64-macos-none` is a valid zig triple but in llvm/clang it's `x86_64-apple-darwin`). So the checks here are reversed because they can only single out artifacts, not find them all.

syrusakbary: What I mean is... what is the true/false value signaling for? I still have no idea. It's true that is Linux?

epilys: If the vendor is Apple, it should definitely not include substring "windows". If the vendor is Pc, it should definitely not include "apple". In any other case just retain it anyways.

PR #3083, opened 2022-08-09
Disable wasm build in build CI

epilys:
Wasm build was added unnecessarily in the build github action in
4710a2af2e5100dd46733e0799a8d3f0418ebeb9

The wasm build is not currently part of the binary assets of a release,
so this commits disable it.

By the way, the wasm build fails because the wasm32-wasi target is
not added with rustup before the `make build-wasmer-wasm` is called.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -196,11 +196,11 @@ jobs:
         shell: bash
         run: |
           make build-wasmer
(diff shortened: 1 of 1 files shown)


ptitSeb:
was build was added to avoid it being broken by code change

epilys:
> wasm build was added to avoid it being broken by code change

Makes sense, maybe it'd be better to put it in the test workflows.

PR #3090, opened 2022-08-10
Added version to the wasmer cli

syrusakbary:
WAPM update checks (and our install script) were failing because `wasmer --version` was not working (we accidentally removed version on #3079 )

This PR brings it back


syrusakbary:
Merging manually as tests passed

PR #3097, opened 2022-08-12
MemoryView lifetime tied to memory and not StoreRef

ptitSeb:
MemoryView lifetime tied to memory and not StoreRef

This make using MemoryView easier to use in Rust (with the Borrow checker) and still have some failsafe about the View lifetime.


syrusakbary (approved):

syrusakbary:
We should be good to merge once tests pass

PR #3101, opened 2022-08-12
CI/build.yaml: add libwasmer headless in default distribution

epilys:


syrusakbary:
I believe current build-capi-headless includes the wasmer artifact and the static one, right? How big is the `.a` in that case?

We need to include only the static one

PR #3102, opened 2022-08-12
Reenable tests for wasm32-wasi and make tests not fail

fschutt:
Continouusly testing whether wasmer-build-wasm fails is necessary because we need to ship it in the cloudcompiler package, see https://github.com/wasmerio/wasmer/commit/842d511173d10dcf2cd738e294299d294603a036

This PR re-enables the CI test, but this time with `rustup target add wasm32-wasi` in front, so that the tests don't fail.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -196,11 +196,12 @@ jobs:
         shell: bash
         run: |
           make build-wasmer
-     #- name: Build Wasmer binary on Wasm32-WASI without LLVM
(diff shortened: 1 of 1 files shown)


epilys (changes requested):
This should be in test, not builds

builds action should take as little time as possible hence should not have useless builds

fschutt:
Okay after talking to Syrus, it seems we are already testing it in the `test-sys.yaml` file.

PR #3114, opened 2022-08-17
Implemented shared memory for Wasmer in preparation for multithreading

john-sharratt:
This adds shared (and custom) memory for Wasmer


john-sharratt:
Added Into<VMMemory> so its a cleaner API.

PR #3115, opened 2022-08-17
Fix static object signature deserialization

syrusakbary:
The current code was overly complex regarding signatures.

This PR simplifies it and removes the trailing `get_with_generics_weak` function in the exports


syrusakbary:
Tests passed, so merging manually

PR #3116, opened 2022-08-18
Multithreading, full networking and RPC for WebAssembly

john-sharratt:
Behold! Multithreading, networking and RPC!


lib/api/src/js/exports.rs
```diff
@@ -53,6 +53,9 @@ pub enum ExportError {
     /// This error arises when an export is missing
     #[error("Missing export {0}")]
     Missing(String),
+    /// This error arises when an export is missing
+    #[error("Serialization failed {0}")]
+    SerializationFailed(String),
```
syrusakbary, line 58: Not sure if I understand when this can be triggered.
When an export can be missing?

lib/api/src/js/instance.rs
```diff
@@ -110,10 +107,17 @@ impl Instance {
             })
             .collect::<Result<Exports, InstantiationError>>()?;
 
+        // If the memory is imported then also export it for backwards compatibility reasons
+        // (many will assume the memory is always exported) - later we can remove this
+        if exports.get_memory("memory").is_err() {
+            if let Some(memory) = externs.iter().filter(|a| a.ty(store).memory().is_some()).next() {
+                exports.insert("memory", memory.clone());
+            }
+        }
```
syrusakbary, line 116: I really don't think we should do this. This should live somewhere else (not on the JS API layer, as it's not JS or VM specific)

lib/api/src/js/exports.rs
john-sharratt: ah this is leftovers - this is not used anymore - we should remove it

lib/api/src/js/instance.rs
john-sharratt: that will trigger some increased coding because of these...

image

Those if we are ok to modify the interfaces that assume memory is an export (perhaps we need to put in some form of abstraction) then we can get rid of these.

lib/api/src/js/instance.rs
```diff
@@ -87,11 +84,11 @@ impl Instance {
     pub fn from_module_and_instance(
         mut store: &mut impl AsStoreMut,
         module: &Module,
+        externs: Vec<Extern>,
         instance: StoreHandle<WebAssembly::Instance>,
```
ptitSeb, line 88: That's another API change that should go in 3.0?

lib/api/src/js/module.rs
```diff
@@ -152,7 +186,8 @@ impl Module {
     /// Opposed to [`Module::new`], this function is not compatible with
     /// the WebAssembly text format (if the "wat" feature is enabled for
     /// this crate).
-    pub fn from_binary(_store: &impl AsStoreRef, binary: &[u8]) -> Result<Self, CompileError> {
+    pub fn from_binary(_store: &impl AsStoreRef, binary: impl IntoBytes) -> Result<Self, CompileError> {
+        let binary = binary.into_bytes();
         //
```
ptitSeb, line 191: and all the `IntoBytes` changes can also be consider API-breaking changes? (and so should go in 3.0?)

lib/api/src/sys/function_env.rs
```diff
@@ -29,7 +29,7 @@ impl<T> FunctionEnv<T> {
     }
 
     /// Get the data as reference
-    pub fn as_ref<'a>(&self, store: &'a impl AsStoreMut) -> &'a T
+    pub fn as_ref<'a>(&self, store: &'a impl AsStoreRef) -> &'a T
```
ptitSeb, line 32: Should also go in 3.0

lib/api/src/sys/instance.rs
```diff
@@ -128,6 +128,14 @@ impl Instance {
...
             })
             .collect::<Exports>();
+        
+        // If the memory is imported then also export it for backwards compatibility reasons
+        // (many will assume the memory is always exported) - later we can remove this
+        if exports.get_memory("memory").is_err() {
+            if let Some(memory) = externs.iter().filter(|a| a.ty(store).memory().is_some()).next() {
+                exports.insert("memory", memory.clone());
+            }
+        }
```
ptitSeb, line 139: does this change is also a break-change on the API level?

lib/api/src/sys/ptr.rs
```diff
@@ -76,7 +76,7 @@ impl<T, M: MemorySize> WasmPtr<T, M> {
 
     /// Get the offset into Wasm linear memory for this `WasmPtr`.
     #[inline]
-    pub fn offset(self) -> M::Offset {
+    pub fn offset(&self) -> M::Offset {
         self.offset
```
ptitSeb, line 80: API breaking change?

lib/api/src/js/module.rs
syrusakbary: Why we want to own the bytes @john-sharratt ?

lib/api/src/js/module.rs
```diff
@@ -165,16 +200,28 @@ impl Module {
     /// This is safe since the JS vm should be safe already.
     /// We maintain the `unsafe` to preserve the same API as Wasmer
     pub unsafe fn from_binary_unchecked(
-        _store: &impl AsStoreRef,
-        binary: &[u8],
+        store: &impl AsStoreRef,
+        binary: impl IntoBytes,
```
syrusakbary, line 204: Why we want to own the bytes @john-sharratt ?

lib/api/src/js/module.rs
```diff
@@ -210,8 +257,9 @@ impl Module {
     /// This validation is normally pretty fast and checks the enabled
     /// WebAssembly features in the Store Engine to assure deterministic
     /// validation of the Module.
-    pub fn validate(_store: &impl AsStoreRef, binary: &[u8]) -> Result<(), CompileError> {
-        let js_bytes = unsafe { Uint8Array::view(binary) };
+    pub fn validate(_store: &impl AsStoreRef, binary: impl IntoBytes) -> Result<(), CompileError> {
```
syrusakbary, line 260: Why we want to own the bytes @john-sharratt ?

docs/migration_to_3.0.0.md
```diff
@@ -139,14 +139,13 @@ import_object.define("env", "host_function", host_function);
...
-For WASI, don't forget to import memory to `WasiEnv`
+For WASI, don't forget to initialize it
 
 ```rust
 let mut wasi_env = WasiState::new("hello").finalize()?;
 let import_object = wasi_env.import_object(&mut store, &module)?;
 let instance = Instance::new(&mut store, &module, &import_object).expect("Could not instantiate module.");
-let memory = instance.exports.get_memory("memory")?;
-wasi_env.data_mut(&mut store).set_memory(memory.clone());
+wasi_env.initialize(&mut store, &instance).unwrap();
```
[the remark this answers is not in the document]
john-sharratt: yes i think it would be good to bring this forward - i think this way is cleaner

lib/api/src/js/function_env.rs
```diff
@@ -36,7 +36,7 @@ impl<T> FunctionEnv<T> {
     }
 
     /// Get the data as reference
-    pub fn as_ref<'a>(&self, store: &'a impl AsStoreMut) -> &'a T
+    pub fn as_ref<'a>(&self, store: &'a impl AsStoreRef) -> &'a T
     where
```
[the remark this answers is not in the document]
john-sharratt: yeah this is a easy one to bring in

lib/api/src/js/function_env.rs
```diff
@@ -112,6 +112,11 @@ impl<T: Send + 'static> FunctionEnvMut<'_, T> {
         self.func_env.as_mut(&mut self.store_mut)
     }
 
+    /// Borrows a new immmutable reference
+    pub fn as_ref(&self) -> FunctionEnv<T> {
+        self.func_env.clone()
+    }
+
```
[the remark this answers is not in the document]
john-sharratt: yeah if we move the other ones we might as well

lib/api/src/js/instance.rs
john-sharratt: not sure on this one as it will drag a bunch of other code with it - might be too much

lib/api/src/js/module.rs
john-sharratt: In order to pass it between threads - in particular it needs to be shipped to the other WebWorkers in the browser as they can not share the modules between each other, only the compiled bytes can be shared.

lib/api/src/sys/function_env.rs
john-sharratt: agreed

lib/api/src/sys/instance.rs
john-sharratt: this probably needs some more thought on it anyway as it assumes the first exported memory but in the future there could be multiple - it does not change the API but does change the behavior.

lib/api/src/sys/ptr.rs
john-sharratt: API change but i dont think its breaking

lib/api/src/js/module.rs
john-sharratt: so the compiled code can be shared to the main thread, compiled there and fed to the other web workers. its a complex setup to get it working because of the JS limitations

john-sharratt: if there is an easier way it would be good to drop the Bytes but it is tricky, thats for sure

lib/api/src/js/module.rs
john-sharratt: see above

lib/api/src/js/module.rs
syrusakbary: In the case of JS, the bytes are already stored in `Vec<u8>`. So it shall be possible to share the compiled code?

lib/api/src/js/instance.rs
ptitSeb: Does that mean the signature of the function can be rolled back to without the externs, and we forget about this change?

PR #3117, opened 2022-08-18
Add tests for wasmer-cli create-{exe,obj} commands

### LINKED ISSUE #3071: Add tests for wasmer-cli create-{exe,obj} commands

epilys:
Unified generated objects interfaces: both static and serialized objects
expose the same function signature to create the module:

  wasm_module_t* wasmer_object_module_new(wasm_store_t* store, const char* module_name);

Closes #3071


tests/integration/cli/tests/create_exe.rs
```diff
@@ -160,3 +223,187 @@ fn create_exe_works_with_file() -> anyhow::Result<()> {
...
+}
+
+#[test]
+fn create_obj_default() -> anyhow::Result<()> {
+    create_obj(vec![], "ymbols", "symbols")
+}
+
+#[test]
+fn create_obj_symbols() -> anyhow::Result<()> {
+    create_obj(vec!["--object-format", "symbols"], "ymbols", "symbols")
```
syrusakbary, line 321: `ymbols` is this a typo?

epilys: No, the output contains `Symbols` hence `ymbols` will match uppercase and lowercase if it ever changes

syrusakbary: Will it match bymbols as well?

lib/cli/src/commands/create_exe.rs
```diff
@@ -21,8 +21,7 @@ use wasmer_object::{emit_serialized, get_object_for_target};
 pub type PrefixerFn = Box<dyn Fn(&[u8]) -> String + Send>;
 
 const WASMER_MAIN_C_SOURCE: &[u8] = include_bytes!("wasmer_create_exe_main.c");
-#[cfg(feature = "static-artifact-create")]
-const WASMER_STATIC_MAIN_C_SOURCE: &[u8] = include_bytes!("wasmer_static_create_exe_main.c");
```
fschutt, line 25: Isn't the `wasmer_static_create_exe_main.c` file necessary for the other code / the `--object-format symbols`? If this is removed, isn't it going to impact the `create-exe-pirita` work?

lib/cli/src/commands/create_exe.rs
```diff
@@ -307,9 +306,39 @@ impl CreateExe {
                         .map_err(|err| anyhow::anyhow!(err.to_string()))?;
                     writer.flush()?;
                     drop(writer);
+                    // Write down header file that includes deserialize function
+                    {
+                        let mut writer = BufWriter::new(File::create("static_defs.h")?);
```
fschutt, line 311: Please use absolute file paths, not relative file paths to the current directory. I.e. `blah_dir.join("static_defs.h")`

lib/cli/src/commands/create_exe.rs
```diff
@@ -307,9 +306,39 @@ impl CreateExe {
...
+                        writer.write_all(WASMER_DESERIALIZE_HEADER.as_bytes())?;
+                        writer.flush()?;
+                    }
+
+                    // write C src to disk
+                    let c_src_path = Path::new("wasmer_main.c");
```
fschutt, line 317: Same here

lib/cli/src/commands/create_exe.rs
```diff
@@ -307,9 +306,39 @@ impl CreateExe {
...
+                    #[cfg(not(windows))]
+                    let c_src_obj = PathBuf::from("wasmer_main.o");
+                    #[cfg(windows)]
+                    let c_src_obj = PathBuf::from("wasmer_main.obj");
 
-                    let cli_given_triple = self.target_triple.clone();
-                    self.compile_c(wasm_object_path, cli_given_triple, output_path)?;
```
fschutt, line 312: I think it's better to have this code in a separate function because afaik, the `create-exe-pirita` code also needs to call `compile_c`, so that way we can avoid duplication.

lib/cli/src/commands/create_exe.rs
```diff
@@ -666,6 +659,7 @@ fn run_c_compile(
         .arg("-O2")
         .arg("-c")
         .arg(path_to_c_src)
+        .arg("-I.")
```
fschutt, line 662: Please do not include the current directory. The reason for this is that if there are accidentally stray header files in the current directory (from which wasmer gets executed from), cc will prefer the header in the current directory, which then leads to linker errors. This happened to me because I had a stray `static_defs.h` in the current directory.

The solution is to make a separate directory for all the header files in the compilation process.

fschutt (changes requested):
Added a few comments regarding style + some questions

lib/cli/src/commands/create_exe.rs
epilys: The deserialize functions for both serialized and symbol object formats now have the same signature, and they no longer require different main.c files.

lib/cli/src/commands/create_exe.rs
epilys: It's literally the same thing.

lib/cli/src/commands/create_exe.rs
epilys: compile_c just called LinkCode blah blah. Having it as a different function doesn't let you set flags etc easily, so no.

lib/cli/src/commands/create_exe.rs
epilys: The current directory is the tempfile created one, not the one you execute create-exe in. This temp current directory should include the generated static_defs.h.

lib/cli/src/commands/create_exe.rs
syrusakbary: I agree with the take from Felix, we should not rely on changing the current working environment for having things to work. We need to stop using `set_current_dir` as only make other functions work as a "side effect". We need to keep things as functional as possible

lib/cli/src/commands/create_exe.rs
syrusakbary: Agreed

lib/cli/src/commands/create_exe.rs
syrusakbary: Your solution relies again on the `set_working_dir` which I really believe we shouldn't be using. So I also agree with Felix here

epilys:
View diff to previous PR version: 

https://github.com/wasmerio/wasmer/compare/create-exe-tests-v1..6bb3ae4b499d7bc7d973f3118794db7f8ded2d53

fschutt (approved):

syrusakbary (approved):

PR #3119, opened 2022-08-18
Added LinearMemory trait

ptitSeb:
Added LinearMemory trait, to describe memory used in runtime.


lib/api/src/sys/externals/memory.rs
```diff
@@ -60,6 +60,13 @@ impl Memory {
         })
     }
 
+    /// Create a memory object from an existing memory and attaches it to the store
+    pub fn new_from_existing(new_store: &mut impl AsStoreMut, memory: VMMemory) -> Self {
```
syrusakbary, line 64: Can this be `Into<VMMemory>`? That way, any implementation that implements the `LinearMemory` trait will work

lib/types/src/memory.rs
```diff
@@ -113,10 +121,89 @@ unsafe impl MemorySize for Memory64 {
...
     fn offset_to_native(offset: Self::Offset) -> Self::Native {
         offset as Self::Native
     }
     fn native_to_offset(native: Self::Native) -> Self::Offset {
         native as Self::Offset
     }
 }
+
+/// Represents memory that is used by the WebAsssembly module
+pub trait LinearMemory
```
syrusakbary, line 134: Can we move this, along with the `VMMemoryDefinition` back to the `wasmer-vm` crate?
The JS implementation will not make use of it (and can't make use of it), so the VM seems a much better place

lib/api/src/js/externals/memory.rs
```diff
@@ -116,6 +100,21 @@ impl Memory {
...
+        store: &mut impl AsStoreMut,
+        js_memory: js_sys::WebAssembly::Memory,
+        ty: MemoryType,
+    ) -> Result<Self, MemoryError> {
+        let vm_memory = VMMemory::new(js_memory, ty);
+        Ok(Self::from_vm_export(store, vm_memory))
+    }
+
+    /// Create a memory object from an existing memory and attaches it to the store
+    pub fn new_from_existing(new_store: &mut impl AsStoreMut, memory: VMMemory) -> Self {
```
syrusakbary, line 114: `Into<VMMemory>` to support the proposal for sys

lib/api/src/sys/externals/memory.rs
ptitSeb: but about the store parameter?

syrusakbary:
@ptitSeb can you add tests using a custom memory implementation?

syrusakbary:
@ptitSeb I think we want the custom memory to be used from the Tunables to make sure the wasmer API is well-rounded so people can use it properly

lib/api/src/sys/tunables.rs
```diff
@@ -174,4 +174,138 @@ mod tests {
...
+            Ok(VMMemory::from_custom(memory))
+        }
+        unsafe fn create_vm_memory(
+            &self,
+            _ty: &MemoryType,
+            _style: &MemoryStyle,
+            _vm_definition_location: NonNull<VMMemoryDefinition>,
+        ) -> Result<VMMemory, MemoryError> {
+            let memory = VMTinyMemory::new().unwrap();
+            Ok(VMMemory::from_custom(memory))
```
syrusakbary, line 268: Can we do `Ok(memory.into())` (as I believe the `From` is implemented)

lib/api/src/sys/tunables.rs
```diff
@@ -174,4 +174,138 @@ mod tests {
...
+        fn table_style(&self, _table: &TableType) -> TableStyle {
+            TableStyle::CallerChecksSignature
+        }
+        fn create_host_memory(
+            &self,
+            _ty: &MemoryType,
+            _style: &MemoryStyle,
+        ) -> Result<VMMemory, MemoryError> {
+            let memory = VMTinyMemory::new().unwrap();
+            Ok(VMMemory::from_custom(memory))
```
syrusakbary, line 259: Can we do `Ok(memory.into())` (as I believe the `From` is implemented)

lib/vm/src/vmcontext.rs
```diff
@@ -303,120 +303,77 @@ mod test_vmglobal_import {
     }
 }
 
-/// The fields compiled code needs to access to utilize a WebAssembly linear
```
syrusakbary, line 306: We should probably keep `VMMemoryDefinition` inside of the `vmcontext`

ptitSeb: done

syrusakbary (approved):

PR #3122, opened 2022-08-20
Update Cargo.lock dependencies

epilys:
plotters 0.3.2 was yanked, issued `cargo update` to resolve.


epilys:
Manual Runtime test action because it wasn't automatically triggered https://github.com/wasmerio/wasmer/actions/runs/2895862408

PR #3123, opened 2022-08-21
Lower libwasmer headless size

epilys:
Brought libwasmer-headless.a from 22MiB to 7.2MiB (on my machine)

By using lto optimization flags.

A qjs.wasm executable using the headless engine now weights 4.6MiB:

% du -sh qjs-he*.out
 20M    qjs-headfull.out
4.6M    qjs-headless.out

### DIFF
--- a/lib/wasi/Cargo.toml
+++ b/lib/wasi/Cargo.toml
@@ -10,9 +10,6 @@ license = "MIT"
 readme = "README.md"
 edition = "2018"
 
-[lib]
-crate-type = ["cdylib", "rlib"]
-
 [dependencies]
 cfg-if = "1.0"
 thiserror = "1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,6 +38,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "android_system_properties"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d7ed72e1635e121ca3e79420540282af22da58be50de153d36f81ddc6b83aa9e"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "ansi_term"
 version = "0.12.1"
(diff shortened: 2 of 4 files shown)


lib/wasi/Cargo.toml
```diff
@@ -10,9 +10,6 @@ license = "MIT"
 readme = "README.md"
 edition = "2018"
 
-[lib]
-crate-type = ["cdylib", "rlib"]
```
syrusakbary, line 14: I believe this is needed for wasm-bindgen / wasmer-js (wasi)

epilys: This is necessary for the lto though. Do you know what command/build fails for wasm-bindgen/wasmer-js if this is removed?

epilys: We can add the crate-type as a special case.

epilys: **Runtime tests - JS** have passed by the way.

syrusakbary:
This relies in the LLVM LTO flags and using llvm bitcode to let LLVM optimize things. I believe there might be better ways to do it as `wasmer-headless` have proven, but I think this solution might be good enough for now

epilys:
> I believe there might be better ways to do it as `wasmer-headless` have proven, but I think this solution might be good enough for now

This is literally the same solution, except for using `xargo` to build a custom sysroot/std. Check out how wasmer-headless is built in `.github/workflows/build.yml`

syrusakbary:
> This is literally the same solution, except for using xargo to build a custom sysroot/std. Check out how wasmer-headless is built in .github/workflows/build.yml

Afaik the solution in this PR relies on LLVM bitcode and using LTO while the one for `wasmer-headless` doesn't (or at least, not directly). The one from xargo is also able to get `wasmer-headless` down to 2Mb, so perhaps there are also differences on the things included in each of the binaries

Issue #2955, opened 2022-06-14
Wasmer cache can throw panic if cache file has been corrupted


grishasobol:
https://github.com/wasmerio/wasmer/blob/66c1c00df0704ba2ba59c250ac571586d5010c79/lib/engine-universal/src/artifact.rs#L85-L103

Hello guys, as I can see in code above, you work with `string slice` without any checks before. As a result this may cause problems  in some cases. For example: if we would corrupt file in compiled wasm cache, then this part of code may throw panic, which is not suitable behaviour. I think the best way is to return error in the case.

epilys:
I meant a `validate: bool` argument. You might not want to revalidate an immutable bytes slice after a first validation. Would that satisfy your usecase?

grishasobol:
> I meant a `validate: bool` argument. You might not want to revalidate an immutable bytes slice after a first validation. Would that satisfy your use case.

Yep, if you will use `validate = true` by default in wasmer functions which uses this method, or propagate validate argument up for all functions which uses this method.

epilys:
Propagating the option makes sense to me. We're redesigning the API for the next major release, 3.0, at the moment. So we will take this into consideration, thanks for writing the ticket!

silwol:
If we strive to follow the [API guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html), I think it makes sense to rather expose a `deserialize_unchecked(…)` method that may panic, and a `deserialize(…)` method that does checking and returns a `Result<…>`. The latter may call the former of course, or both call an additional private function that has a `validate: bool` parameter - depending on what makes more sense for the implementation.

epilys:
Actually keeping this open because it might get lost in #2925. But it depends on it.

PR #3130, opened 2022-08-23
Remove panics from Artifact::deserialize

fschutt:
Fixes https://github.com/wasmerio/wasmer/issues/2955

#### Review

### DIFF
--- a/lib/compiler/src/engine/artifact.rs
+++ b/lib/compiler/src/engine/artifact.rs
@@ -117,29 +117,26 @@ impl Artifact {
             ));
         }
 
-        if bytes.len() < ArtifactBuild::MAGIC_HEADER.len() {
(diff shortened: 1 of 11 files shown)


lib/compiler/src/engine/artifact.rs
```diff
@@ -573,6 +593,16 @@ impl Artifact {
         ))
     }
 
+    fn get_byte_slice(input: &[u8], start: usize, end: usize) -> Option<&[u8]> {
```
syrusakbary, line 596: This can just return a Result and then the code will be greatly simplified:

```
.ok_or(DeserializeError::InvalidByteLength {
                    expected: cur_offset + WORD_SIZE,
                    got: bytes.len(),
                })
```

syrusakbary:
Tests are failing, once they pass we should be good to go

PR #3136, opened 2022-08-24
Fix cache path install

fschutt:
If the cache path didn't exist, downloading the `x86_64-linux-gnu.tar.gz` tarball would fail because the `/cache` directory did not exist, this PR fixes that.


syrusakbary:
I think this PR has extra commits?

PR #3150, opened 2022-08-30
Use WASI generated types instead of manual ones

syrusakbary:
Wasmer has the crate `wasmer-wasi-types` which is a manual implementation of the types used in WASI.
Ideally, we would use `wit-bindgen` to generate those instead of generating the types by hand.


fschutt:
Superseded by #3177.

PR #3153, opened 2022-08-31
SharedMemory & Atomics

ptitSeb:
# Description
Enabled SharedMemory and the Atomics extension proposal

- [x] Enable Atomic extension by default
- [x] Fix "imports" tests https://github.com/wasmerio/wasmer/issues/3154
- [x] Add function for memory.atomic.wait32, memory.atomic.wait64 and memory.atomic.notify opcodes https://github.com/wasmerio/wasmer/issues/3155
- [x] Add support for the new wait/notify opcodes in Cranelift compiler #3156 
- [x] Add support for the new wait/notify opcodes in LLVM compiler #3157
- [x] Add support for atomic access opcodes in AArch64/Singlepass compiler #3159
- [x] Add support for the new wait/notify opcodes in Singlepass compiler #3158
- [x] Fix Atomic issues on x86_64 Singlepass compiler not related to Wait/Notify opcodes #3161
- [x] Fix Atomic issues on Cranelift compiler not related to Wait/Notify opcodes #3162
- [x] Fix Atomic issues on LLVM compiler not related to Wait/Notify opcodes #3163
- [x] Fix the ticket #3167 on Cranelift

For #3304


lib/api/src/sys/imports.rs
```diff
@@ -111,6 +112,36 @@ impl Imports {
...
+            self.define(
+                "env",
+                "memory",
+                Memory::new_from_existing(store, memory.clone().into()),
+            );
+            Some(memory)
+        } else {
+            None
+        }
+    }
```
syrusakbary, line 143: I don't think this should be in the public Wasmer API

ptitSeb: It should be addressed with commit [7cb4b0a](https://github.com/wasmerio/wasmer/pull/3153/commits/7cb4b0acd507fe1645f84f9637514cbf3c68257c)

lib/api/src/sys/imports.rs
```diff
@@ -130,16 +130,21 @@ impl Imports {
...
+                .contains_key(&("env".to_string(), "memory".to_string()))
+            {
+                ret.define(
+                    "env",
+                    "memory",
+                    Memory::new_from_existing(store, memory.clone().into()),
+                );
+            }
+        };
+        ret
```
syrusakbary, line 147: I still like to have this code outside of the API. It's veeeery tied to a specific need (and that's why it should be out of the API crate)

ptitSeb: Im' sory but I don't understand what you are asking. You want me to remove completly "import_shared_memory"?

syrusakbary: Yes. I believe we can achieve the same with a function method `into_shared` on the `Memory` export

ptitSeb: Ok, done

lib/vm/src/instance/mod.rs
```diff
@@ -776,6 +782,205 @@ impl Instance {
...
+        //let memory = unsafe { import.definition.as_ref() };
+        //if ! memory.shared {
+        // We should trap according to spec, but official test rely on not trapping...
+        //}
+
+        // fetch the notifier
+        let key = (memory_index.as_u32(), dst);
+        let mut conds = self.conditions.lock().unwrap();
+        let mut cnt = 0u32;
+        if conds.contains_key(&key) {
```
theduke, line 972: This is the same code as above, right?
I'd put it in a separate function to avoid duplication.

Also: double hashing/acquiring ( as above).

lib/vm/src/instance/mod.rs
```diff
@@ -776,6 +782,205 @@ impl Instance {
...
+        &mut self,
+        memory_index: LocalMemoryIndex,
+        dst: u32,
+        count: u32,
+    ) -> Result<u32, Trap> {
+        //let memory = self.memory(memory_index);
+        //if ! memory.shared {
+        // We should trap according to spec, but official test rely on not trapping...
+        //}
+
```
theduke, line 938: This seems to be missing an alignment check.

(Spec says to trap on non-aligned notify)

Or is that happening earlier somewhere?

lib/vm/src/instance/mod.rs
```diff
@@ -776,6 +782,205 @@ impl Instance {
...
+        memory_index: MemoryIndex,
+        dst: u32,
+        count: u32,
+    ) -> Result<u32, Trap> {
+        //let import = self.imported_memory(memory_index);
+        //let memory = unsafe { import.definition.as_ref() };
+        //if ! memory.shared {
+        // We should trap according to spec, but official test rely on not trapping...
+        //}
+
```
theduke, line 967: Alignment? (see above)

tests/wast/wasmer/atomic_load.wast
```diff
@@ -0,0 +1,9 @@
+(module
+    (memory 1)
+    (func (export "atomic_load")
+        i32.const 0xffff_fff0
+        i32.atomic.load offset=16
+        drop
+    )
+)
+(assert_trap (invoke "atomic_load") "out of bound")
```
theduke, line 9: I don't think the WAST tests can actuallly test concurrency, right?

Would be nice to have a custom test that actually runs multiple threads and tests notify/wait behaviour.

theduke (changes requested):

lib/vm/src/instance/mod.rs
ptitSeb: Not sure, I'll check if it's not a borrow checker hell.

lib/vm/src/instance/mod.rs
ptitSeb: I thought it was done earlier, but I need to check as  it's a per compiler thing.

lib/vm/src/instance/mod.rs
```diff
@@ -776,6 +782,205 @@ impl Instance {
...
+    // once unparked, the waiter thread will remove it's mark on the HashMap
+    // timeout / awake is tracked with a boolean in the HashMap
+    // because `park_timeout` doesn't gives any information on why it returns
+    fn do_wait(&mut self, index: u32, dst: u32, timeout: i64) -> u32 {
+        // fetch the notifier
+        let key = (index, dst);
+        let mut conds = self.conditions.lock().unwrap();
+        conds.entry(key).or_insert_with(Vec::new);
+        let v = conds.get_mut(&key).unwrap();
+        v.push((current(), false));
```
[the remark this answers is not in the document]
theduke: Yeah, I think a test is overkill.

It's still good to have the limit for Deploy, because otherwise a single deployment could eat up all the memory by just spawning lots of waiter threads in a loop.

Well, we'll actually we'll always have max thread limits, so that's also not a huge problem.
Still doesn't hurt to adhere to the spec, `HashMap::len()` is not expensive.

tests/wast/wasmer/atomic_load.wast
ptitSeb: Yeah, WAST will not really test concurency. We need a new kind of test suite for this, with probably some new infrastructure (so I would prefer another separate PR)

theduke:
I assume you validated the alignment checking for notify?

ps: I noticed that the official .wast suite doesn't seem to test notify alignment, so we might want to send them an issue to add that. (unless I missed it).

theduke (approved):
Thanks for the NotifyMap!

Issue #3152, opened 2022-08-30
Standardize js and sys tests


syrusakbary:
In the Wasmer API, we have two kind of tests: `js` and `sys` ones depending on the feature used.

However, both are testing the Wasmer API (which should be resilient across the js or sys feature).
In general, we should unify the `crates/api/tests` for both js and sys (and using the `test` or `wasm_bindgen` macro depending on the target/feature).
That way we test our that our public API is resilient

fschutt:
Ideally this should be done after https://github.com/wasmerio/wasmer/pull/3129 is merged, but I can start working on it

syrusakbary:
Closing

PR #3164, opened 2022-08-31
Synchronize between -sys and -js tests

fschutt:
Fixes https://github.com/wasmerio/wasmer/issues/3152.


lib/api/tests/instance.rs
```diff
@@ -0,0 +1,89 @@
...
+
+
+#[cfg(feature = "js")]
+#[cfg_attr(feature = "js", wasm_bindgen_test)]
+fn exports_work_after_multiple_instances_have_been_freed_js() {
+    exports_work_after_multiple_instances_have_been_freed().unwrap();
+}
+
+#[cfg_attr(feature = "sys", test)]
+fn exports_work_after_multiple_instances_have_been_freed() -> Result<()> {
```
syrusakbary, line 15: Could we do this with a macro?

```rust
#[universal_test]
fn exports_work_after_multiple_instances_have_been_freed() -> Result<()> {
```

The macro could expand into something like:
```rust
#[wasm_bindgen_test]
fn exports_work_after_multiple_instances_have_been_freed_js() {
    fn exports_work_after_multiple_instances_have_been_free().unwrap();
}

#[test]
fn exports_work_after_multiple_instances_have_been_freed() -> Result<()> {
    // ...
}
```

That way the code is not as repetitive

syrusakbary (approved):
Looks good once tests pass!

syrusakbary (approved):

PR #3177, opened 2022-09-07
Auto-generate wasi-types from .wit files

### LINKED ISSUE #2972: Update Wasmer-WASI types with wit-bindgen bindings
Once #2911 is done, we should update Wasmer-WASI with those bindings

fschutt:
Fixes #2972.

Supersedes #3150.


Michael-F-Bryan:
Instead of doing your code generation as part of a `build.rs` script, you might want to consider moving it elsewhere so you don't impact build times. It's also generally considered [bad manners](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script) for build scripts to change files outside of `$OUT_DIR` (I *think* `cargo` will warn you about this when you try to publish).

The rust-analyzer project uses a pattern where they'll generate the code in-memory during a test and compare that to whatever is on disk. If the generated code has changed, the files are immediately updated and the test is failed with a message like *"generated.rs was out of date. Please commit the updated files and re-run this test"*. That way you can't accidentally merge outdated files because CI will fail.

Here is an example where I use the pattern in one of my projects:

https://github.com/Michael-F-Bryan/scad-rs/blob/4dbff0c30ce991105f1e649e678d68c2767e894b/crates/codegen/src/syntax/mod.rs#L15-L32

fschutt:
@syrusakbary okay, tests are finally passing, but I wouldn't merge it just yet because I'd need to review whether the old types are the same as the new ones esp. for ones marked `#[repr(u16)]`

lib/wasi-types-generated/wit-clean/output.wit
```diff
@@ -0,0 +1,1750 @@
...
+}
+
+enum bus-event-type {
+    noop,
+    exit,
+    call,
+    result,
+    fault,
+    close,
+}
```
syrusakbary, line 805: I believe this should be part of wasix itself? @john-sharratt

lib/wasi/src/syscalls/wasix32.rs
```diff
@@ -2,6 +2,12 @@
 use crate::{WasiEnv, WasiError, WasiState, WasiThread};
 use wasmer::{FunctionEnvMut, Memory, Memory32, MemorySize, StoreMut, WasmPtr, WasmSlice};
 use wasmer_wasi_types::*;
```
syrusakbary, line 4: Do we still need `wasmer_wasi_types` ? Why?

fschutt: Not all types can be ported to WIT, especially not types that use generics.

syrusakbary: Kind of make sense... 
I see the file: https://github.com/wasmerio/wasmer/blob/wasi-types-generation-2/lib/wasi-types/src/lib.rs

I think it might make sense to move that into the `wasmer-wasi` repo.
Although I see some types that might be feasible to move to wit?
* `subscription::EventType`
* `net::*` (most of the net types?)


I also think the bus type names would make sense to have different types if it's 32 or 64 bits (so it doesn't need to be generic).
cc @john-sharratt

fschutt: @syrusakbary The `subscription` module is portable (I missed that somehow), but net uses arrays, which WIT does not support (that's why I couldn't port it).

syrusakbary: Ok. Let's port the subscription module. And for the rest of things that can't be ported let's just have a `types.rs` files inside `wasmer-wasi` instead of a new crate

fschutt:
Currently we can't merge this because of https://github.com/bytecodealliance/wit-bindgen/issues/323

@syrusakbary There are only a few types that are `repr(u16)`, I'll just backport them, so that we can merge this PR.

syrusakbary (approved):

PR #3178, opened 2022-09-08
Feat enhanced tinytunable test

ptitSeb:
# Description
Added a new test for Custom Tunables, with a Store and Instance creation.


lib/vm/src/memory.rs
```diff
@@ -410,4 +439,11 @@ where
 
     /// Attempts to clone this memory (if its clonable)
     fn try_clone(&self) -> Option<Box<dyn LinearMemory + 'static>>;
+
+    #[doc(hidden)]
```
syrusakbary, line 443: It would be great if we can add a comment to say why is unsafe.
Example
```
    /// # Safety
    /// This function is unsafe because WebAssembly specification requires that data is always set at initialization time.
    /// It should be the implementors responsibility to make sure this respects the spec
```

syrusakbary (approved):
One small minor nit, but looks great overall!

syrusakbary (changes requested):
Requested a small change, sorry for the double re-review

syrusakbary (approved):

PR #3200, opened 2022-09-27
Return 0 instead of BrokenPipe if WasiPipe cannot be read from

### LINKED ISSUE #3171: WasiPipe should return 0 (no bytes read), not BrokenPipe
https://github.com/wasmerio/wasmer/blob/aedbdccad8b6a0b9f1093dd2a2baff5c746b7350/lib/wasi/src/state/pipe.rs#L110

fschutt:
This is the correct behaviour, since the pipe isn't "broken", it just can't be read from anymore

Fixes #3171.

### DIFF
--- a/lib/wasi/src/state/pipe.rs
+++ b/lib/wasi/src/state/pipe.rs
@@ -106,12 +106,12 @@ impl Read for WasiPipe {
                 }
             }
             let rx = self.rx.lock().unwrap();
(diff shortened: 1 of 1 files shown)


syrusakbary:
Can we add a test into the `wasi-wast` test suite? https://github.com/wasmerio/wasmer/tree/master/tests/wasi-wast 

Thanks!

Issue #3125, opened 2022-08-22
WASI: file system rights doesn't work


Mo-Fatah:
### Describe the bug
I am able to call many WASI APIs on a `file descriptor` that doesn't have the right to, and it works without any errors. I wrote an example [here](https://github.com/Mo-Fatah/wasi-testsuite/blob/main/engines-diff/wasmer-fs-rights/src/main.rs) where I was able to call `fd_write`, `fd_seek` and `fd_read` on a `file descriptor` that doesn't have any base rights and no errors appeared and I was able to read and write to this file.

I am using 
```
wasmer 2.3.0 | rustc 1.61.0 (fe5b13d68 2022-05-18) | x86_64
```

### Steps to reproduce
- You can run the [example](https://github.com/Mo-Fatah/wasi-testsuite/blob/main/engines-diff/wasmer-fs-rights/src/main.rs) I mentioned above, or you can try to use (for example ) `fd_read` on a file descriptor that doesn't have `fd_read` rights and it will work without errors.
### Expected behavior
- The result should be an error, `NOT_CAPABLE` error I assume.
### Actual behavior
- The read and write operations is performed without any problems.

### Additional context
According to the latest WASI standard, this is supposed to be a security issue. I don't know if the wasmer team is handling capabilities differently, or this is indeed a bug.

syrusakbary:
This seems a bug in our side, we need to investigate further. Thanks for opening the issue @Mo-Fatah

syrusakbary:
We tried to fix it, but CI is failing. Which indicates that there might be another bug in the Wasmer-WASI implementation, that we need to further investigate (we'll post our findings as we learn them on this issue)

PR #3201, opened 2022-09-27
Add failing test for debugging file descriptor and return 0 on BrokenPipe instead of Error

fschutt:
This is a WIP PR for https://github.com/wasmerio/wasmer/issues/3125

The compiled wasm file should panic with an "EACESS" error, currently it panics with "EBADF" 

cc @syrusakbary


fschutt:
@syrusakbary This PR depends on #3145 because #3145 changes the WasiPipe API. It would be better to merge that first, then continue here.

Also, updating the compiler version seems to have broken a lot of test cases for some reason.

fschutt:
Superseded by #3240

Issue #3210, opened 2022-09-30
popcnt stuck in a loop on M1 and singlepass compiler


grishasobol:
### Describe the bug
Using M1 + singlepass compiler causes stucking in loop for that code for example:
```wat
(module
  (func $popcnt (export "popcnt") (param i32) (result i32)
      local.get 0
      i32.popcnt
  )
)
```
If param == 1, then function `popcnt` call execution lasts forever.

### Environment
Use wasmer in Cargo.toml like this:
```
wasmer = { version = "2.3.0", default-features = false, features = ["default-singlepass", "default-universal", "wat", "sys"] }
```

```sh
$ rustc -V
rustc 1.65.0-nightly (17cbdfd07 2022-09-13)

$ uname -a
... MacBook-Pro.local 21.6.0 Darwin Kernel Version 21.6.0 ... RELEASE_ARM64_T6000 arm64
```

### Steps to reproduce
1. Clone repo https://github.com/grishasobol/wasmer-fail .
2. Make `cargo test` on machine with M1 processor (see Environment above).
3. Tests must stuck in a loop.

### Additional context
This happen because this assembler execution on M1 result conflicts with Aarch64 documentation:
```asm
mov w1, 32
mov w2, 1
lsl w3, w2, w1
```
There will be `1` in `w3`, but due documentation must be `0`. (The same for 64 registers)

You can reproduce this:
1. take file [arm.s.zip](https://github.com/wasmerio/wasmer/files/9685436/arm.s.zip)
2. build it
```sh
$ as -o arm.o arm.s && ld -o arm arm.o -lSystem -syslibroot `xcrun -sdk macosx --show-sdk-path` -e _start -arch arm64
```
3. run it and print return code
```sh
$ ./arm ; echo $?
1
```

syrusakbary:
This should be fixed by #3211

PR #3211, opened 2022-09-30
fix popcnt for aarch64

grishasobol:
# Description
Fix problem with infinite `popcnt` WASM instruction execution on M1 processor. See details in corresponded issue: https://github.com/wasmerio/wasmer/issues/3210

# Review
1) Make fix for `i32.popcnt`. Use additional execution branch if `src` register equal to `1`:
```
if src == 0 => goto exit_label
if src == 1 => goto src_is_one_label

loop_label:
... // popcnt loop body

goto exit_label

src_is_one_label:
mov dest, 1

exit_label:
```


ptitSeb:
Thank you for the bug report, the detailled analysis, and the PR :)

I have checked ARM64 documentation, and in fact, the LSL will behave like you describe on all Aarch64 platform, not only M1 cpu, so you can remove the Apple specific compile tag.
The doc list:
```
result = ShiftReg(n, shift_type, UInt(operand2) MOD datasize);
```
so we see the modulo datasize for the shift value.


As for the fix, I would prefer some more simple, like changing
```rust
        self.assembler.emit_clz(Size::S32, src, tmp)?; // clz src => tmp
        self.assembler
            .emit_add(Size::S32, tmp, Location::Imm8(1), tmp)?; // inc tmp
        self.assembler.emit_lsl(Size::S32, src, tmp, src)?; // src << tmp => src
        self.assembler.emit_cbnz_label(Size::S32, src, label_loop)?; // if src!=0 goto loop
```
to
```rust
        self.assembler.emit_clz(Size::S32, src, tmp)?; // clz src => tmp
        self.assembler.emit_lsl(Size::S32, src, tmp, src)?; // src << tmp => src
        self.assembler.emit_lsl(Size::S32, src, Location::Imm8(1), src)?; // src << 1 => src
        self.assembler.emit_cbnz_label(Size::S32, src, label_loop)?; // if src!=0 goto loop
```
for the 32bits popcnt, to workaround the 31+1=32 issue.

I tested this with your test on a aarhc64/linux box, and the change is both needed and working.

grishasobol:
Sure)) your solution is much better. Will do it

ptitSeb (approved):

ptitSeb:
@grishasobol can you please run `cargo fmt --all` and push the diff, the lint is failing, sorry.

ptitSeb (approved):

PR #3215, opened 2022-10-04
Update wasmer --version logic, integrate wapm-cli

### LINKED ISSUE #3189: Include Build ID in wasmer --version
It's better to print both the version and the git ref + date when the binary was built when printing `wasmer --version`

### LINKED ISSUE #3208: Unify wasmer-cli and wapm-cli into one tool
The goal is to unify the `wasmer` and `wapm` tools into one tool so you can do:

```sh
$ wasmer run python/python
```

TODO:

- create a new crate `wasmer/registry`
- implement wapm functionality cleanly
- remove Lockfile, just use one `.wasmer/cache` directory
- publish `wasmer-registry` as a library, then test & integrate with wasmer-cli

fschutt:
- [x] Update clap
- [x] Integrate wapm-cli logic
- [x] Polish CLI handling and UX

Fixes #3208.
Fixes #3189.


lib/cli/build.rs
```diff
@@ -16,6 +16,8 @@ pub fn main() {
             "cargo:rustc-env=WASMER_BUILD_GIT_HASH_SHORT={}",
             &git_hash[..5]
         );
+    } else {
+        println!("cargo:rustc-env=WASMER_BUILD_GIT_HASH_SHORT=??????");
```
fschutt, line 20: This is the only thing I'm unsure about - for some reason the git hash is not available on -musl, not sure if it's 100% necessary, this is more of a quick hack

lib/cli/src/cli.rs
```diff
@@ -204,5 +233,244 @@ pub fn wasmer_main() {
...
+            if let Ok(mut sv) = split_version(&package) {
+                let mut package_download_info = None;
+                if !sv.package.contains('/') {
+                    let sp = spinner::SpinnerBuilder::new(format!(
+                        "Looking up command {} ...",
+                        sv.package
+                    ))
+                    .spinner(vec![
+                        "⣾", "⣽", "⣻", "⢿", "⡿", "⣟", "⣯", "⣷", " ", "⠁", "⠂", "⠄", "⡀", "⢀", "⠠",
+                        "⠐", "⠈",
```
syrusakbary, line 249: Since this is used/reused more times in the codebase, can we add it to a global var or isolate into a common function?

lib/cli/src/cli.rs
```diff
@@ -204,5 +233,244 @@ pub fn wasmer_main() {
...
+                        .into_run_args(local_package_wasm_path, Some(package.manifest))
+                        .execute();
+                }
+
+                // else: local package not found
+                let sp =
+                    spinner::SpinnerBuilder::new(format!("Installing package {} ...", sv.package))
+                        .spinner(vec![
+                            "⣾", "⣽", "⣻", "⢿", "⡿", "⣟", "⣯", "⣷", " ", "⠁", "⠂", "⠄", "⡀", "⢀",
+                            "⠠", "⠐", "⠈",
```
syrusakbary, line 297: Same (refactor so this is reused)

Michael-F-Bryan (commented):
I don't think I have enough experience with the codebase to comment on the user-facing changes, so I've mainly focused on code quality/maintainability things.

Something I'd strongly recommend is running `cargo clippy` every now and then. I've tried to point out a couple bits of unidiomatic code, but in general `clippy` (and `cargo clippy --fix`) is super useful for cleaning up your code.

lib/cli/src/cli.rs
```diff
@@ -204,5 +233,244 @@ pub fn wasmer_main() {
...
+            [p] => (p.trim().to_string(), None),
+            _ => {
+                return Err(anyhow!("Invalid version / command: {s:?}"));
+            }
+        };
+        version = Some(v);
+        command
+    } else {
+        None
+    };
```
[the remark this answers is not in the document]
fschutt: I agree, a regex would be better here. The first block splits for `@`, the second blocks look for `:` in the `package` or `version`.

lib/cli/src/cli.rs
```diff
@@ -204,5 +233,244 @@ pub fn wasmer_main() {
...
+        println!("Registry  Package  Version  Commands ");
+        println!("======================================");
+        println!();
+    } else {
+        let mut table = Table::init(rows);
+        table.set_titles(row!["Registry", "Package", "Version", "Commands"]);
+        table.set_format(*format::consts::FORMAT_NO_LINESEP_WITH_TITLE);
+        table.set_format(*format::consts::FORMAT_NO_COLSEP);
+        let _ = table.printstd();
+    }
```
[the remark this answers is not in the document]
fschutt: Not like I wanted it to, for some reason it didn't print the headers at all. I can try again, but when having no packages installed, it didn't print anything, which is why this code exists.

lib/cli/src/commands/run.rs
```diff
@@ -21,6 +21,150 @@ mod wasi;
...
+    debug: bool,
+
+    #[cfg(feature = "debug")]
+    #[clap(long = "verbose")]
+    verbose: Option<u8>,
+
+    /// Application arguments
+    #[clap(value_name = "ARGS")]
+    args: Vec<String>,
+}
```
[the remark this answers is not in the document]
fschutt: It's a copy of the `Run` struct, but without the `path` field and yes, we sadly do need the `cfg()`s. It's just so that I can `derive(Parser)` and let clap handle the CLI arguments, but without expecting the `path` field.

I'll see if I can remove / shorten this by just re-using `Run`.

lib/registry/src/lib.rs
```diff
@@ -0,0 +1,956 @@
...
+    let namespace = name
+        .split('/')
+        .next()
+        .ok_or(format!("missing namespace for {name:?}"))?;
+    let name = name
+        .split('/')
+        .nth(1)
+        .ok_or(format!("missing name for {name:?}"))?;
+    let install_dir =
+        get_global_install_dir(registry_host).ok_or(format!("no install dir for {name:?}"))?;
```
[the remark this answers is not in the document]
fschutt: Yeah, I don't think this would've passed CI anyways since clippy runs in CI.

lib/cli/build.rs
```diff
@@ -1,4 +1,26 @@
+use chrono::prelude::*;
+use std::process::Command;
+
 pub fn main() {
-    println!("cargo:rerun-if-changed=build.rs");
-    println!("cargo:rerun-if-env-changed=WASMER_INSTALL_PREFIX");
```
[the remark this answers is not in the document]
fschutt: Right now the build script is run on every build, which is what I wanted because it sets the date of compilation on every single run, that's why I removed those lines.

lib/registry/src/lib.rs
```diff
@@ -0,0 +1,956 @@
...
+        };
+
+        let host = match host {
+            Some(s) => s,
+            None => continue 'outer,
+        };
+
+        let root_dir = match get_global_install_dir(&host) {
+            Some(o) => o,
+            None => continue 'outer,
```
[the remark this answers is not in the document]
fschutt: Okay, but I personally don't like splitting everything across mini-functions, sometimes it is indeed easier to read and more explicit.

lib/cli/src/cli.rs
```diff
@@ -204,5 +233,244 @@ pub fn wasmer_main() {
...
+                            .into_run_args(buf, Some(package.manifest))
+                            .execute();
+                    }
+                    Err(e) => {
+                        println!("{e}");
+                        return Ok(());
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
Michael-F-Bryan: Yeah, I understand that the logic gets a lot harder because we're essentially trying to do half of `clap`'s job. 

My point is more that we've got a 180-line function that has 5-6 levels of indentation and which is performing several very different responsibilities.

I know you said you prefer to keep code together instead of splitting it out, but for someone that wasn't there when the code was written, long sprawling functions tend to make code really hard to read and maintain. They also tend to attract technical debt because you get into the *"this function is already pretty messy so it'll be fine to just add another if-statement"* mentality instead of continuously refactoring.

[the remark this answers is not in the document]
fschutt: I split it into multiple functions in https://github.com/wasmerio/wasmer/pull/3215/commits/471fd367209626455e5c9f48134d7cb5a0513570

fschutt:
@syrusakbary I think you can review + merge now

lib/cli/src/cli.rs
```diff
@@ -204,5 +237,462 @@ pub fn wasmer_main() {
...
+
+    let mut run_args = RunWithoutFile::try_parse_from(args_without_package.iter())?;
+    run_args.command_name = sv.command.clone();
+
+    run_args
+        .into_run_args(package_dir, sv.command.as_deref())?
+        .execute()
+}
+
+fn start_spinner(msg: String) -> SpinnerHandle {
```
syrusakbary, line 427: This should be part of `commands/run.rs`

lib/cli/src/cli.rs
```diff
@@ -204,5 +237,462 @@ pub fn wasmer_main() {
...
+        .execute()
+        .map_err(|e| ExecuteLocalPackageError::DuringExec(e.context(anyhow::anyhow!("{}", sv))))
+}
+
+fn try_autoinstall_package(
+    args: &[String],
+    sv: &SplitVersion,
+    package: Option<PackageDownloadInfo>,
+    force_install: bool,
+) -> Result<(), anyhow::Error> {
```
syrusakbary, line 394: This should be part of `commands/run.rs`

lib/cli/src/cli.rs
```diff
@@ -204,5 +237,462 @@ pub fn wasmer_main() {
...
+// before vs. during execution
+enum ExecuteLocalPackageError {
+    BeforeExec(anyhow::Error),
+    DuringExec(anyhow::Error),
+}
+
+fn try_execute_local_package(
+    args: &[String],
+    sv: &SplitVersion,
+) -> Result<(), ExecuteLocalPackageError> {
```
syrusakbary, line 358: This should be part of `commands/run.rs`

lib/cli/src/cli.rs
```diff
@@ -204,5 +237,462 @@ pub fn wasmer_main() {
...
+    let _ = std::io::stdout().flush();
+    Err(anyhow::anyhow!("command {sv} not found"))
+}
+
+// We need to distinguish between errors that happen
+// before vs. during execution
+enum ExecuteLocalPackageError {
+    BeforeExec(anyhow::Error),
+    DuringExec(anyhow::Error),
+}
```
syrusakbary, line 353: This should be part of `commands/run.rs`

lib/cli/src/cli.rs
```diff
@@ -204,5 +237,462 @@ pub fn wasmer_main() {
...
+        Err(ExecuteLocalPackageError::DuringExec(e)) => return Err(e),
+        _ => {}
+    }
+
+    println!("finding local package {} failed", sv);
+    // else: local package not found - try to download and install package
+    try_autoinstall_package(args, &sv, package_download_info, r.force_install)
+}
+
+fn try_lookup_command(sv: &mut SplitVersion) -> Result<PackageDownloadInfo, anyhow::Error> {
```
syrusakbary, line 325: This should be part of `commands/run.rs`

lib/cli/src/cli.rs
```diff
@@ -204,5 +237,462 @@ pub fn wasmer_main() {
...
-    PrettyError::report(options.execute());
+    // Check if the file is a package name
+    if let WasmerCLIOptions::Run(r) = &options {
+        return try_run_package_or_file(&args, r);
+    }
+
+    options.execute()
+}
+
+fn try_run_package_or_file(args: &[String], r: &Run) -> Result<(), anyhow::Error> {
```
syrusakbary, line 248: This should be part of `commands/run.rs`

lib/cli/src/cli.rs
```diff
@@ -160,6 +165,7 @@ impl WasmerCLIOptions {
             Self::CreateObj(create_obj) => create_obj.execute(),
             Self::Config(config) => config.execute(),
             Self::Inspect(inspect) => inspect.execute(),
+            Self::List => print_packages(),
```
syrusakbary, line 168: This should be a new command (and file), like config, inspect and others

fschutt:
It seems that the integration tests sometimes fail because of failing network access / backend slowness. In extreme cases you'd have to restart the failing tests. I think that `test_is_registry_present` fails too fast.

syrusakbary (approved):

PR #3218, opened 2022-10-05
Seal `HostFunctionKind`

uint:
# Description
I noticed the `HostFunctionKind` trait was clearly intended to be sealed, but it actually isn't. This PR completes the sealing pattern. Feels like this is important to include in the 3.0.0 release to avoid subtle breaking changes.

# Review


syrusakbary (approved):
Great call! Merging

PR #3221, opened 2022-10-10
Fix #3197

### LINKED ISSUE #3197: Incompatible import type for resized table
### Describe the bug

A WASM module that imports a table with a minimum size can only be instantiated with an imported table that was created with a compatible table type. Importing a table that was created with a lower minimum size and resized using `Table::grow` to a compatible size results in an error `incompatible import type`.

Observed using Wasmer 2.3.0 and 3.0.0-beta.

### Steps to reproduce

Using Wasmer 3.0.0-beta:

```rust
use wasmer::{imports, Instance, Module, Store, Table, TableType, Type, Value};

const WAT:&str = r#"(module (table (import "env" "table") 100 funcref))"#;

fn wasmer3() -> anyhow::Result<()> {
    let mut store = Store::default();
    let module = Module::new(&store, WAT)?;
    let ty = TableType::new(Type::FuncRef, 0, None);
    let table = Table::new(&mut store, ty, Value::FuncRef(None))?;
    table.grow(&mut store, 100, Value::FuncRef(None))?;
    let imports = imports! {"env" => {"table" => table}};
    let _instance = Instance::new(&mut store, &module, &imports)?;
    Ok(())
}
```

### Expected behavior

After calling `table.grow` the size of `table` is compatible with the WASM import declaration, and the instance is created.

### Actual behavior

Error while importing `"env"."table"`: incompatible import type. Expected `Table(TableType { ty: FuncRef, minimum: 100, maximum: None })` but received `Table(TableType { ty: FuncRef, minimum: 0, maximum: None })`.

### Additional context

The WASM specification states that the limits of the table limits must match (https://www.w3.org/TR/wasm-core-2/valid/types.html#tables), and the minimum limit is updated when growing a table (https://www.w3.org/TR/wasm-core-2/exec/modules.html#grow-table).

In `wasmtime` and Javascript in Firefox/Chrome/V8 the instantiation is successful in the corresponding code:

#### Javascript (works)

```js
const module = await WebAssembly.compileStreaming(fetch('test1.wasm'));
const table = new WebAssembly.Table({ initial: 0, element: "anyfunc" });
table.grow(100);
const instance = await WebAssembly.instantiate(module, {env: {table: table}});
console.log(instance);
```

#### Wasmtime 1.0.0 (works)

```rust
use wasmtime::{Engine, Instance, Module, Store, Table, TableType, Val, ValType};

fn wasmtime() -> anyhow::Result<()> {
    let engine = Engine::default();
    let mut store = Store::new(&engine, ());
    let module = Module::new(&engine, WAT)?;
    let ty = TableType::new(ValType::FuncRef, 0, None);
    let table = Table::new(&mut store, ty, Val::FuncRef(None))?;
    table.grow(&mut store, 100, Val::FuncRef(None))?;
    let _instance = Instance::new(&mut store, &module, &[table.into()])?;
    Ok(())
}
```

#### Wasmer 2.3.0 (fails)

```rust
use wasmer::{Store, Module, Instance, Table, TableType, Type, Value, imports};

fn wasmer2() -> anyhow::Result<()> {
    let store = Store::default();
    let module = Module::new(&store, WAT)?;
    let ty = TableType::new(Type::FuncRef, 0, None);
    let table = Table::new(&store, ty, Value::FuncRef(None))?;
    table.grow(100, Value::FuncRef(None))?;
    let imports = imports!{"env" => {"table" => table}};
    let _instance = Instance::new(&module, &imports)?;
    Ok(())
}
```

fschutt:
Fixes #3197


fschutt:
@syrusakbary wasmtime has a different architecture in this section, which is why they can just return `self.vec.len()` for the minimum, while we use a "descriptor", so we have to update it every time the table is grown or shrunk.

syrusakbary:
We need to address this based on today's latest review (check the length, not just the descriptor minimum, on link time)

syrusakbary (approved):
It looks great, although we forgot to update the `is_memory_compatible` with the same fix as we did in the table

fschutt:
@syrusakbary yeah I wasn't sure if I should do that

syrusakbary:
As long as that's done we should be good to merge

lib/vm/src/table.rs
```diff
@@ -88,6 +88,11 @@ impl VMTable {
         unsafe { Self::new_inner(table, style, None) }
     }
 
+    /// Returns the size of the table
+    pub fn get_runtime_size(&self) -> u32 {
```
Michael-F-Bryan, line 92: Should we be using `usize` here, or is that a wasm32 thing?

lib/api/src/sys/externals/table.rs
```diff
@@ -211,3 +211,18 @@ impl<'a> Exportable<'a> for Table {
         }
     }
 }
+
+#[test]
+fn test_table_grow_3197() {
```
Michael-F-Bryan, line 216: Let's add a link for lazy people like me :stuck_out_tongue: 

```suggestion
/// Check the example from <https://github.com/wasmerio/wasmer/issues/3197>.
#[test]
fn test_table_grow_issue_3197() {
```

Michael-F-Bryan (approved):
LGTM :+1: 

I've got a couple queries, but they're more so I can understand the decision (e.g. `u32` vs `usize`) or to make troubleshooting `test_table_grow_3197()` easier if it ever fails down the track.

lib/vm/src/table.rs
fschutt: No, u32 is correct, this is a wasm32 thing I think.

lib/api/src/sys/externals/table.rs
```diff
@@ -211,3 +211,18 @@ impl<'a> Exportable<'a> for Table {
...
+
+    const WAT: &str = r#"(module (table (import "env" "table") 100 funcref))"#;
+
+    let mut store = Store::default();
+    let module = Module::new(&store, WAT).unwrap();
+    let ty = TableType::new(Type::FuncRef, 0, None);
+    let table = Table::new(&mut store, ty, Value::FuncRef(None)).unwrap();
+    table.grow(&mut store, 100, Value::FuncRef(None)).unwrap();
+    let imports = imports! {"env" => {"table" => table}};
+    let _instance = Instance::new(&mut store, &module, &imports).unwrap();
```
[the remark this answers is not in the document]
fschutt: I can't do this because `is_table_compatible` is not exposed in wasmer_types, but I added a comment explaining what the test does. Plus we have the linked issue.

fschutt:
@syrusakbary I have no idea why the Linux CI doesn't run, otherwise tests should pass now.

PR #3222, opened 2022-10-10
Add function to retrieve function name from wasm_frame_t

### LINKED ISSUE #3166: How to get function name from `wasm_frame_t`?
I'm trying to make my trap handling code show a user friendly stack trace. I'd at least like to show which functions are hit in the trace.

Looking at the C API functions for `wasm_frame_t`, I can get the function index, function offset, and module offset, for a given frame. Function index sounds useful, but what is it an index into? Is there a way of looking up the name of a function from its index?

fschutt:
Fixes #3166.


fschutt:
Should be done like https://docs.wasmtime.dev/c-api/trap_8h_source.html

fschutt:
@syrusakbary @Michael-F-Bryan review + merge?

Michael-F-Bryan:
It doesn't look like we provide a function which gives you a `*const c_char` from a `*const wasm_name_t`. Does that mean we expect the caller to cast a `wasm_name_t` to a `*const char` so they can read it?

Jumping around wasmtime's docs, [it seems like](https://docs.wasmtime.dev/c-api/wasm_8h.html#a5a6e9ac85b29c62a9fb679a974a7cdaa) `wasm_name_t` is just a typedef for [`wasm_byte_vec_t`](https://docs.wasmtime.dev/c-api/structwasm__byte__vec__t.html), which is pretty much the C version of an `&str` reference pointing into our `wasm_frame_t`.

I'm not sure whether `frame.info.function_name()` and `frame.info.module_name()` return borrowed `&str`s or owned `String`s, but if they're borrowed it means we can drop the `wasm_name_delete()` destructor.

fschutt:
> I'm not sure whether frame.info.function_name() and frame.info.module_name() return borrowed &strs or owned Strings, but if they're borrowed it means we can drop the wasm_name_delete() destructor.

It doesn't matter because we need to convert them to `CString` anyway (trailing 0). So we need to allocate.

> Does that mean we expect the caller to cast a wasm_name_t to a *const char so they can read it?

No, it returns a `wasm_name_t`, which has to be deleted by the user because it allocates Rust memory (see above). So I wanted to make that clear in the API, so that we have `wasm_frame_module_name` and `wasm_frame_function_name` for creation of the `wasm_name_t` and `wasm_name_delete` for deletion. The API is not that hard.

Users can print `printf("%s", wasm_name_t.name);` just like normal.

Michael-F-Bryan (approved):

syrusakbary (changes requested):

PR #3224, opened 2022-10-11
Implement ExternRef in wasmer-js

### LINKED ISSUE #3175: Have ExternRef working in wasmer-js
Currently both wasmer-js and wasmer-sys don't support ExternRef. Since the browser JS VM already support extern refs, we just need to wrap the JS Wasm API to support it in wasmer-js

fschutt:
Fixes #3175.


lib/api/src/js/value.rs
```diff
@@ -33,7 +34,7 @@ pub enum Value {
     F64(f64),
 
     /// An `externref` value which can hold opaque data to the wasm instance itself.
-    //ExternRef(Option<ExternRef>),
+    ExternRef(Option<Arc<ExternRef>>),
```
syrusakbary, line 37: Why an Arc here?

syrusakbary: It seems to me that the `ExternRef` should be the one storing the `Arc` (if really needed) (and hidden from the user)

Issue #3228, opened 2022-10-14
Rename syscall should replace target_path when it it exists


Yesterday17:
https://github.com/wasmerio/wasmer/blob/19e7e7af2727792ffae1e092e99585dc486a05b2/lib/wasi/src/syscalls/mod.rs#L2662-L2673

> If newpath already exists, it will be atomically replaced, so
> that there is no point at which another process attempting to
> access newpath will find it missing.  However, there will
> probably be a window in which both oldpath and newpath refer to
> the file being renamed.
>
> -- https://man7.org/linux/man-pages/man2/rename.2.html

syrusakbary:
This should be fixed by #3230

PR #3230, opened 2022-10-14
Remove test if dest file exist on path_rename wasi syscall (for #3228)

ptitSeb:
# Description
The test was added "just in case", but is not conforment with WASI spec, were path_rename should befave like POSIX `renameat`


lib/wasi/src/syscalls/mod.rs
```diff
@@ -2657,18 +2657,6 @@ pub fn path_rename<M: MemorySize>(
...
         true
     ));
-    if state
-        .fs
-        .get_inode_at_path(
-            inodes.deref_mut(),
-            new_fd,
-            target_path.to_str().as_ref().unwrap(),
-            true,
-        )
```
syrusakbary, line 2667: I believe we added this check because just checking would create the inode if didn't exist.

Perhaps we can just remove the if/return and make it:

```suggestion
        fs.get_inode_at_path(
            inodes.deref_mut(),
            new_fd,
            target_path.to_str().as_ref().unwrap(),
            true,
        );
```

syrusakbary (changes requested):
We need to add a test verifying this is fixed

lib/wasi/src/syscalls/mod.rs
ptitSeb: I don't think it's necessary, I think the test was really added for a wrong reason (avoid overwrite of destination while it should allow it). But creating the node doesn't do any harm, so lets do that.

lib/vfs/src/mem_fs/filesystem.rs
```diff
@@ -189,13 +197,25 @@ impl crate::FileSystem for FileSystem {
...
+            match inode_dest {
+                Some((position, inode_of_file)) => {
+                    // Remove the file from the storage.
+                    fs.storage.remove(inode_of_file);
+
+                    // Remove the child from the parent directory.
+                    fs.remove_child_from_node(inode_of_to_parent, position)?;
+                }
+                None => (),
+            }
```
Michael-F-Bryan, line 217:
```suggestion
            if let Some((position, inode_of_file)) =  inode_dest {
                // Remove the file from the storage.
                fs.storage.remove(inode_of_file);

                // Remove the child from the parent directory.
                fs.remove_child_from_node(inode_of_to_parent, position)?;
            }
```

lib/wasi/src/syscalls/mod.rs
```diff
@@ -2659,18 +2659,12 @@ pub fn path_rename<M: MemorySize>(
...
-        .is_ok()
-    {
-        return Errno::Exist;
-    }
+    state.fs.get_inode_at_path(
```
Michael-F-Bryan, line 2662:
```suggestion
    // Create the destination inode if the file exists.
    let _ = state.fs.get_inode_at_path(
```

lib/wasi/src/syscalls/mod.rs
```diff
@@ -2659,18 +2659,12 @@ pub fn path_rename<M: MemorySize>(
...
+        inodes.deref_mut(),
```
Michael-F-Bryan, line 2663: We can avoid this (and needing to import `DerefMut`) by taking a mutable reference and letting the compiler auto-deref.

```suggestion
        &mut inodes,
```

lib/wasi/src/syscalls/mod.rs
```diff
@@ -2659,18 +2659,12 @@ pub fn path_rename<M: MemorySize>(
...
+        new_fd,
+        target_path.to_str().as_ref().unwrap(),
```
Michael-F-Bryan, line 2665: What happens when the target path isn't a UTF-8 string?

lib/wasi/src/syscalls/mod.rs
```diff
@@ -2829,10 +2824,15 @@ pub fn path_symlink<M: MemorySize>(
...
             .fs
             .get_parent_inode_at_path(inodes.deref_mut(), fd, old_path_path, true));
-    let depth = wasi_try!(state
+    let depth = state
         .fs
-        .path_depth_from_fd(inodes.deref(), fd, source_inode))
-        - 1;
+        .path_depth_from_fd(inodes.deref(), fd, source_inode);
+
+    // depth == -1 means folder is not relative
```
Michael-F-Bryan, line 2831: Is it worth mentioning the ticket number or failing test that caused this change?

```suggestion
    // depth == -1 means folder is not relative. See issue #3233.
```

tests/wasi-wast/wasi/tests/path_rename.rs
```diff
@@ -74,6 +74,88 @@ fn run_with_toplevel_dir() {
...
+        let shiba_string = "「柴犬」という名前は中央高地で使われていたもので、文献上では、昭和初期の日本犬保存会の会誌「日本犬」で用いられている。一般的には、「柴」は小ぶりな雑木を指す。
+由来には諸説があり、
+
+    柴藪を巧みにくぐり抜けて猟を助けることから
+    赤褐色の毛色が枯れ柴に似ている（柴赤）ことから
+    小さなものを表す古語の「柴」から
+
+の3つの説が代表的。";
+        let shiba_bytes: Vec<u8> = shiba_string.bytes().collect();
+        f.write_all(&shiba_bytes[..]).unwrap();
```
Michael-F-Bryan, line 103:
```suggestion
        f.write_all(shiba_string.as_bytes()).unwrap();
```

tests/wasi-wast/wasi/tests/path_rename.rs
```diff
@@ -74,6 +74,88 @@ fn run_with_toplevel_dir() {
...
+    {
+        let mut f = std::fs::OpenOptions::new()
+            .create_new(true)
+            .write(true)
+            .open(&file_to_rename_to)
+            .unwrap();
+
+        let lorem_string = "lorem ispum";
+        let lorem_bytes: Vec<u8> = lorem_string.bytes().collect();
+        f.write_all(&lorem_bytes[..]).unwrap();
```
Michael-F-Bryan, line 115:
```suggestion
        f.write_all(lorem_string.as_bytes()).unwrap();
```

tests/wasi-wast/wasi/tests/path_rename.rs
```diff
@@ -162,6 +244,7 @@ fn run_with_different_sub_dirs() {
 
 fn main() {
     run_with_toplevel_dir();
+    run_with_toplevel_dir_overwrite();
     run_with_sub_dir();
     run_with_different_sub_dirs();
 }
```
Michael-F-Bryan, line 250: Out of curiosity, why do we use a `main()` function that invokes each test sequentially instead of Rust's builtin test harness?

Michael-F-Bryan (commented):
LGTM :+1: 

We've already discussed most of my comments over Zoom and I'm happy with the PR. There are a couple suggestions, but they're mostly cosmetic changes.

lib/wasi/src/syscalls/mod.rs
Michael-F-Bryan: We use `deref_mut()` in other places within the same file, so this change is probably not going to help.

lib/wasi/src/syscalls/mod.rs
Michael-F-Bryan: The rest of the code uses `unwrap()` when handling paths, so to *properly* handle paths that aren't UTF-8 we'd need to rewrite a lot of code.

tests/wasi-wast/wasi/tests/path_rename.rs
ptitSeb: Because the test is ran both as native and wasm, and the native ran output is used to build the actual .wast test (to build the expected behaviour automaticaly)

Michael-F-Bryan (approved):

syrusakbary (approved):

PR #3237, opened 2022-10-19
Inherit cargo metadata

### LINKED ISSUE #3231: Use cargo version / authors from workspace
I just learned it's possible to simplify our version updating by using a similar strategy to:

https://github.com/bytecodealliance/wasmtime/blob/main/crates/cranelift/Cargo.toml#L3-L4

This would simplify the way that we update versions

Michael-F-Bryan:
# Description

This resolves #3231 by inheriting common metadata from the workspace's `Cargo.toml` file. This feature is [quite new](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds), so it also requires bumping the MSRV from 1.61 to 1.64.

# Review


theduke:
@Michael-F-Bryan after the WASIX merge the MSRV is now `1.64`, so we could re-activate this PR - would be great for maintenance, especially as we are now adding a whole bunch of crates.

Michael-F-Bryan:
Superseded by #3918.

PR #3244, opened 2022-10-20
Feat riscv llvm

ptitSeb:
# Description
Add Risc-V support for the LLVM compiler


syrusakbary:
Cranelift recently added also support for RISC-V, it might be interesting to also enable and test it @ptitSeb 
https://github.com/bytecodealliance/wasmtime/pull/4271

ptitSeb:
> Cranelift recently added also support for RISC-V, it might be interesting to also enable and test it @ptitSeb [bytecodealliance/wasmtime#4271](https://github.com/bytecodealliance/wasmtime/pull/4271) [bytecodealliance/wasmtime#5044](https://github.com/bytecodealliance/wasmtime/pull/5044)

Yes, I have seen that, but I think I'll separate it from this ticket and create 2 more tickets: 1. update cranelift 2. add riscv support

Makefile
```diff
@@ -14,21 +14,24 @@ SHELL=/usr/bin/env bash
...
 # | Cranelift  | Linux    | amd64        | glibc |
 # | LLVM       | Darwin   | aarch64      | musl  |
-# | Singlepass | Windows  |              |       |
+# | Singlepass | Windows  | riscv        |       |
 # |------------|----------|--------------|-------|
 #
 # Here is what works and what doesn't:
 #
-# * Cranelift works everywhere,
+# * Cranelift works everywhere except Linux/`riscv`,
```
syrusakbary, line 22: Probably want to remove this since I assume Cranelift also works now in RISC-V?

lib/compiler-cranelift/src/config.rs
```diff
@@ -117,11 +117,13 @@ impl Cranelift {
...
 
-        builder.finish(self.flags())
+        let is_riscv = matches!(target.triple().architecture, Architecture::Riscv64(_));
+
+        builder.finish(self.flags(is_riscv))
     }
 
     /// Generates the flags for the compiler
-    pub fn flags(&self) -> settings::Flags {
+    pub fn flags(&self, is_riscv: bool) -> settings::Flags {
```
syrusakbary, line 126: Since this is a public function, I think it will be better to pass the `&Target` to this function (similarly to isa)

```suggestion
    pub fn flags(&self, target: &Target) -> settings::Flags {
        let is_riscv = matches!(target.triple().architecture, Architecture::Riscv64(_));
```

syrusakbary (approved):
Other than my suggestions, the PR looks good

Makefile
ptitSeb: Right, it's a leftover before I added Cranelift-riscv compatibility.

lib/compiler-cranelift/src/config.rs
ptitSeb: Ok, done (not sure why `flags` is public)

deny.toml
```diff
@@ -214,9 +214,9 @@ allow-registry = ["https://github.com/rust-lang/crates.io-index"]
 # List of URLs for allowed Git repositories
 allow-git = []
 
-#[sources.allow-org]
+[sources.allow-org]
 # 1 or more github.com organizations to allow git sources for
-#github = [""]
+github = ["TheDan64"]
```
theduke, line 219: nit: Comment for why this is allows would be good.

I already wondered with other settings in `deny.toml.`

lib/compiler-llvm/src/config.rs
```diff
@@ -177,16 +192,47 @@ impl LLVM {
...
+                    _ => self.code_model(),
+                },
             )
-            .unwrap()
+            .unwrap();
+
+        if let Architecture::Riscv64(_) = triple.architecture {
+            // TODO: totally non-portable way to change ABI
+            unsafe {
+                pub struct MyTargetMachine {
```
theduke, line 218: Any particular reason to wrap this in a struct rather than just writing to the memory directly?

If sticking with the struct

* should have `#[repr(c)]` 
* a comment explaining what's going would be nice

tests/ignores.txt
```diff
@@ -28,6 +29,34 @@ cranelift+aarch64+macos    traps::start_trap_pretty
 # https://github.com/wasmerio/wasmer/issues/2808
 cranelift+aarch64 spec::skip_stack_guard_page
 llvm+aarch64      spec::skip_stack_guard_page
+cranelift+riscv64 spec::skip_stack_guard_page
+llvm+riscv64      spec::skip_stack_guard_page
+
+# riscv support is still early, function call ABI needs some work
+llvm+riscv64      static_function::llvm::universal
+llvm+riscv64      static_function_with_env::llvm::universal
+llvm+riscv64      static_function_with_results::llvm::universal
```
theduke, line 38: We should document the RISC-V limitations and behaviour somewhere, maybe in a separate file in `docs/`, and also point this out in the release notes.

lib/wasi/src/lib.rs
```diff
@@ -258,6 +258,7 @@ impl WasiRuntimeError {
     }
 }
 
+#[allow(clippy::result_large_err)]
```
theduke, line 261: Do we really want to plaster this everywhere instead of, eg, boxing the offending error variant?

theduke (commented):
RISC-V here we come, landing this will be awesome!

Just a few nitpicks.

Regarding the large error: the error is not in any hot path, so the large error should be fine, just having the annotation everywhere is pretty annoying.

Because users will also get this error if they use clippy, not just us.

lib/wasi/src/lib.rs
theduke: Yeah, if it's a public error we can't change it anyway.

deny.toml
ptitSeb: This is not needed anymore, we are using regular version of inkwell now

lib/compiler-llvm/src/config.rs
ptitSeb: Added some more comment.
This must not have the `#[repr(c)]` decorator, as the point is to have the same memory representation

tests/ignores.txt
ptitSeb: Ok, I added a file, but not sure what should be put inside, if the details are enough or not.

lib/compiler-llvm/src/config.rs
theduke: Super hacky, but what works that works.

tests/ignores.txt
theduke: I think that's a good intro.

If users have specific questions they can ask and we can amend the docs.

:+1:

theduke (approved):
RISC-V here we come.

PR #3254, opened 2022-10-25
Release 3.0.0-rc.1

fschutt:

PR #3255, opened 2022-10-25
Remove the "#" from some changelog URLs

Michael-F-Bryan:
I'm guessing this was just a copy/paste bug from #3254 - we were linking to `https://github.com/wasmerio/wasmer/pull/#3248` instead of `https://github.com/wasmerio/wasmer/pull/3248` (note the extra `#`).


fschutt:
I already did that myself, but thanks for the PR

PR #3258, opened 2022-10-26
Migrate pirita / native executables feature from wasmer-private

fschutt:


fschutt:
Ideally this should be one single patch based on https://github.com/wasmerio/wasmer-private/pull/40 so that it can be applied / reverted later on.

lib/cli/src/commands/wasmer_create_exe_main.c
```diff
@@ -1,16 +1,24 @@
...
 
 #define own
 
-// TODO: make this define templated so that the Rust code can toggle it on/off
-#define WASI
 
-extern wasm_module_t* wasmer_object_module_new(wasm_store_t* store, const char* module_name) asm("wasmer_object_module_new");
+#define WASI
 
+#ifdef WASI_PIRITA
```
syrusakbary, line 14: Where `WASI_PIRITA` is being defined?

lib/cli/src/commands/create_exe.rs
```diff
@@ -427,27 +345,192 @@ impl CreateExe {
...
+                v.clone().canonicalize().unwrap_or(v.clone())
+            } else {
+                {
+                    let libwasmer_path = if target_triple
+                        .clone()
+                        .unwrap_or(Triple::host())
+                        .operating_system
+                        == wasmer_types::OperatingSystem::Windows
+                    {
+                        "lib/wasmer.lib"
```
syrusakbary, line 412: Could we test if create exe now works in windows? (since we now have the wasmer.lib file)

lib/c-api/Cargo.toml
```diff
@@ -39,6 +39,12 @@ thiserror = "1"
 typetag = { version = "0.1", optional = true }
 paste = "1.0"
 
+[dependencies.pirita]
+path = "../../../pirita/crates/pirita"
```
syrusakbary, line 43: What will happen if the directory is not found, since the dependency is optional... will it fail?

fschutt: Yes. We fixed that by just copy-pasting the entire pirita codebase into this repository.

It does compile when the /pirita is in the correct path, but I didn't want to inline the entire pirita repo in this commit. So that we have two commits at least: one with the diff of all the create-exe / create-obj / run-pirita changes and one with the /pirita repo.

I am still not sure how to solve the issue of not being able to depend on private repositories in CI.

fschutt: I've fixed this with a `ssh://` URL - it works locally but we need to look at how to integrate it in the CI via Deploy Keys / SSH keys

lib/wasi/src/runners/emscripten.rs
```diff
@@ -0,0 +1,118 @@
...
+    if !is_emscripten_module(module) {
+        return Err(anyhow!("Atom {name:?} is not an emscripten module"));
+    }
+
+    #[cfg(feature = "wasi")]
+    if wasmer_wasi::Wasi::has_wasi_imports(&module) {
+        return Err(anyhow!(
+            "Unsupported: atom {name:?} has both emscripten and WASI imports."
+        ));
+    }
```
syrusakbary, line 83: This should not be the case. A emscripten package can have wasi imports as well. We should delete this

lib/cli/src/commands/wasmer_deserialize_module.h
```diff
@@ -10,7 +10,7 @@ extern "C" {
 extern size_t WASMER_MODULE_LENGTH asm("WASMER_MODULE_LENGTH");
 extern char WASMER_MODULE_DATA asm("WASMER_MODULE_DATA");
 
-wasm_module_t* wasmer_object_module_new(wasm_store_t* store, const char* module_name) {
+wasm_module_t* wasmer_module_new(wasm_store_t* store, const char* wasm_name) {
```
syrusakbary, line 13: The `wasmer_module_new` is not very descriptive. I would add back the `object` on it, since I think is more clear

fschutt: Should I also rename `wasm_module_t` to `wasm_object_module_t`?

syrusakbary: `wasm_module_t` is still the same right? (meaning: the returned type is stil a wasm module) So... no need to change `wasm_module_t`

lib/wasi/Cargo.toml
```diff
@@ -28,6 +28,11 @@ bincode = { version = "1.3", optional = true }
 chrono = { version = "^0.4", default-features = false, features = [ "wasmbind", "std", "clock" ], optional = true }
 derivative = { version = "^2" }
 bytes = "1"
+webc = { version = "3.0.1", optional = true, default-features = false, features = ["std", "mmap"] }
+serde_cbor = { version = "0.11.2", optional = true }
+nuke-dir = { version = "0.1.0", optional = true }
```
syrusakbary, line 33: This dependency is not used anywhere

Issue #3264, opened 2022-10-27
Openning a file from local folder, when not mounted, should result in error `ENOTCAPABLE` and no file openned


ptitSeb:
With no folder preopenned, openning a file from current folder should result in an error `ENOTCAPABLE` and no file openned.

here is the code

```C
#include <assert.h>
#include <errno.h>
#include <stdio.h>

int main() {
  FILE* file = fopen("fopen-working-directory.c", "r");
  assert(file == NULL);
  assert(errno == ENOTCAPABLE);

  return 0;
}
```
[fopen-working-directory.zip](https://github.com/wasmerio/wasmer/files/9880613/fopen-working-directory.zip)

And attached the wasm file, that expect to be run with `fopen-working-directory.c` existing in current folder.

Issue #3263, opened 2022-10-27
[WASI] trying to fopen a folder should result in `ENOTCAPABLE` error and no handle


ptitSeb:
According to wasi.fyi libc test suite, trying to fopen anything from a parent of a preopenned folder should result in a `ENOTCAPABLE` error and no file open. 
While wasmer forbid the operation, the error code doesn't seems correct.
`ENOTCAPABLE` should be the value 93

C code is
```C
#include <assert.h>
#include <errno.h>
#include <stdio.h>

int main() {
  FILE* file = fopen("fopen-directory-parent-directory/..", "r");
  assert(file == NULL);
  assert(errno == ENOTCAPABLE);

  return 0;
}
```

Attached is the wasm, that should be run with an existing `fopen-directory-parent-directory` folder  in the pre-openned folder
[fopen-directory-parent-directory.zip](https://github.com/wasmerio/wasmer/files/9880580/fopen-directory-parent-directory.zip)

syrusakbary:
This was fixed with #3266

PR #3266, opened 2022-10-28
Return ENotCapable error when accessing unknown files on root (for #3263 and #3264)

ptitSeb:
Fixes for wasm.fyi tests:
 * When accessing bad path from the root, return ENOTCAPABLE error instead of ENOENT (root can only contains pre-openned folder, no other capability)
 * Accessing ROOT for other things then entering folder should result in ENOTCAPABLE


syrusakbary (approved):
We need to add also a test

ptitSeb:
> We need to add also a test

This is a chalange here: the original tests are based on C code and around specific syscall error code return or Exit Code value. There is no infrastructure to do that on wasmer tests

lib/cli/src/error.rs
```diff
@@ -25,8 +26,16 @@ impl PrettyError {
...
+                let trapcode = runtime.map(|e| e.clone().to_trap());
                 eprintln!("{:?}", PrettyError { error });
-                1
+                match trapcode {
+                    #[cfg(target_os = "windows")]
+                    Some(_) => 3,
+                    #[cfg(not(target_os = "windows"))]
+                    Some(_) => 128 + libc::SIGABRT,
+                    _ => 1,
+                }
```
syrusakbary, line 38:
```suggestion
                let runtime: Option<&RuntimeError> = error.downcast_ref();
                let trapcode = runtime.map(|e| e.clone().to_trap());
                eprintln!("{:?}", PrettyError { error });
                match trapcode {
                    Some(err) => {
                        match err.downcast::<WasiError>() {
                            Ok(WasiError::Exit(exit_code)) => {
                                // We should exit with the provided exit code
                                std::process::abort();
                            }
                            _ => 1,
                        }
                    }
                    _ => 1,
                }
```

syrusakbary (approved):

PR #3276, opened 2022-11-03
Remove unnecessary checks to test internet connection

fschutt:


syrusakbary (commented):
Can you also update the query to get the packages as we chatted in the morning?

PR #3283, opened 2022-11-07
Added Attach/Detach instance from a FunctionEnv (allow use of Instance from a function callback)

ptitSeb:
# Description
Allows the FunctionEnv to be attached/detached an instance. 
After it's created, simply use `env.attach_instance(&mut store, &instance)`. Only one instance can be attached. Use `detach_instance(&mut store)` to detach any instance attached. To get the attached instance, use `env.get_instance(&store)`, or, for a `FunctionEnvMut`, simply `env.get_instance()`. This allows the use of instance from a function Callback.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -251,6 +251,11 @@ path = "examples/metering.rs"
 required-features = ["cranelift"]
 
 [[example]]
+name = "metering-callback"
+path = "examples/metering_callback.rs"
+required-features = ["cranelift"]
+
+[[example]]
 name = "imports-exports"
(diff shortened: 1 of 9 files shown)


theduke:
My main concern with this is what I mentioned yesterday: we are essentially exposing hidden unsafety in the public API.

At least `attach_instance`, the instance getters etc would have to be marked as `unsafe fn`, because the caller must ensure that the Instance stays alive and is attached correctly.

But I had a different idea that I didn't consider yesterday, will discuss offline.

PR #3289, opened 2022-11-08
Fix polling in WasiPipe

fschutt:
> @john-sharratt: the polling functionality you have implemented won't work as its not 
> actually polling the receiver - you need to implement it like this: 
> https://github.com/john-sharratt/wasmer/blob/fe14f086808431d6f64b540421798ac779cdce0a/lib/wasi/src/state/pipe.rs#L223
>
> Note: I've converted wasmer-os over to wasmer and now using WasiPipe hence its important this is implemented correctly. Especially as bash uses Polling on stdin which is connected via this pipe.

### DIFF
--- a/lib/wasi/src/state/pipe.rs
+++ b/lib/wasi/src/state/pipe.rs
@@ -384,12 +384,29 @@ impl VirtualFile for WasiPipe {
     fn unlink(&mut self) -> Result<(), FsError> {
         Ok(())
     }
-    fn bytes_available_read(&self) -> Result<Option<usize>, FsError> {
-        Ok(Some(
-            self.read_buffer
-                .as_ref()
-                .map(|s| s.len())
-                .unwrap_or_default(),
-        ))
+
+    /// Returns the number of bytes available.  This function must not block
+    /// Defaults to `None` which means the number of bytes is unknown
+    fn bytes_available_read(&self) -> wasmer_vfs::Result<Option<usize>> {
+        loop {
+            {
+                let read_buffer = self.read_buffer.lock().unwrap();
+                if let Some(inner_buf) = read_buffer.as_ref() {
+                    let buf_len = inner_buf.len();
+                    if buf_len > 0 {
+                        return Ok(Some(buf_len));
+                    }
+                }
+            }
+            let rx = self.rx.lock().unwrap();
+            if let Ok(data) = rx.try_recv() {
+                drop(rx);
+
+                let mut read_buffer = self.read_buffer.lock().unwrap();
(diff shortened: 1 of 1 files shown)


fschutt:
TODO: when this is done and bash / dash is working, we can test it with `echo blah | cat`

lib/wasi/src/state/pipe.rs
```diff
@@ -235,18 +235,27 @@ impl WasiPipe {
...
-                    let reader = buf.as_ref();
-                    let read = read_bytes(reader, memory, iov).map(|_| buf_len as usize)?;
-                    buf.advance(read);
-                    return Ok(read);
+            if let Ok(mut read_buffer) = self.read_buffer.lock() {
+                if let Some(buf) = read_buffer.as_mut() {
+                    let buf_len = buf.len();
+                    if buf_len > 0 {
+                        let reader = buf.as_ref();
+                        let read = read_bytes(reader, memory, iov).map(|_| buf_len as usize)?;
```
theduke, line 243: Seems like you are throwing away the actual amount of bytes read, which can be lower than the buffer size.

lib/wasi/src/state/pipe.rs
```diff
@@ -305,19 +316,21 @@ impl Seek for WasiPipe {
...
-                        let read = reader.read(buf).map(|_| buf_len as usize)?;
-                        inner_buf.advance(read);
-                        return Ok(read);
+            if let Ok(mut read_buffer) = self.read_buffer.lock() {
+                if let Some(inner_buf) = read_buffer.as_mut() {
+                    let buf_len = inner_buf.len();
+                    if buf_len > 0 {
+                        if inner_buf.len() > buf.len() {
+                            let mut reader = inner_buf.as_ref();
+                            let read = reader.read_exact(buf).map(|_| buf.len())?;
```
theduke, line 325: Same thing as above.

lib/wasi/src/state/pipe.rs
```diff
@@ -305,19 +316,21 @@ impl Seek for WasiPipe {
...
+                            inner_buf.advance(read);
+                            return Ok(read);
+                        } else {
+                            let mut reader = inner_buf.as_ref();
+                            let read = reader.read(buf).map(|_| buf_len as usize)?;
```
theduke, line 330: As above.

theduke (commented):
Not familiar enough with this code to provide a comprehensive review.

Did notice one problematic pattern, though.

theduke:
This will be subsumed by #3422 .

PR #3290, opened 2022-11-09
Limit the use of clone when handling Compilation object

ptitSeb:
# Description
Small improvment to limit the number of clone when handling Compilation object.

Based on https://github.com/near/wasmer/pull/140

For #3305


lib/compiler/src/artifact_builders/artifact_builder.rs
```diff
@@ -85,25 +83,35 @@ impl ArtifactBuild {
...
+            .iter()
+            .map(|(_, section)| section.relocations.clone())
+            .collect::<PrimaryMap<SectionIndex, _>>();
         let libcall_trampolines_section = make_libcall_trampolines(target);
         custom_section_relocations.push(libcall_trampolines_section.relocations.clone());
         let libcall_trampolines = custom_sections.push(libcall_trampolines_section);
         let libcall_trampoline_len = libcall_trampoline_len(target) as u32;
 
         let serializable_compilation = SerializableCompilation {
-            function_bodies: compilation.get_function_bodies(),
```
syrusakbary, line 99: If get_function_bodies and other methods are no longer used, perhaps can we eliminate them totally?

ptitSeb: I forgot to removed the commented code, it's done now

syrusakbary (approved):

PR #3297, opened 2022-11-10
Implement wasmer login

fschutt:


lib/cli/src/commands/login.rs
```diff
@@ -0,0 +1,35 @@
+use clap::Parser;
+use dialoguer::Input;
+
+/// Subcommand for listing packages
+#[derive(Debug, Clone, Parser)]
+pub struct Login {
+    /// Registry to log into (default: wapm.io)
+    #[clap(name = "REGISTRY")]
+    pub registry: Option<String>,
```
Michael-F-Bryan, line 9: Should this be an argument? That way people could run `wasmer login --registry wapm.dev my-secret-token`. Having multiple optional positional arguments feels like it could be ambiguous.

It also lets us do something like this:

```rs
#[clap(long, env, default_value = "wapm.io")]
pub registry: String,
```

That way the only place that hard codes the default registry is that `clap` attribute.

lib/cli/src/commands/login.rs
```diff
@@ -0,0 +1,35 @@
...
+    #[clap(name = "TOKEN")]
+    pub token: Option<String>,
+}
+
+impl Login {
+    fn get_token_or_ask_user(&self) -> Result<String, std::io::Error> {
+        match self.token.as_ref() {
+            Some(s) => Ok(s.clone()),
+            None => Input::new()
+                .with_prompt("Please paste the login token from https://wapm.io/me:\"")
```
Michael-F-Bryan, line 20: We might want to use `self.registry` here instead of hard-coding `https://wapm.io/me`.

lib/registry/src/config.rs
```diff
@@ -0,0 +1,329 @@
...
+    /// A mocked version of the standard function for integration tests
+    #[cfg(feature = "integration_tests")]
+    pub fn save(&self) -> anyhow::Result<()> {
+        let config_serialized = toml::to_string(&self)?;
+        crate::integration_tests::data::RAW_CONFIG_DATA.with(|rcd| {
+            *rcd.borrow_mut() = Some(config_serialized);
+        });
+
+        Ok(())
+    }
```
Michael-F-Bryan, line 269: Is this a good opportunity to remove the `feature = "integration_tests"` thing? We should be able to use the `tempfile` crate to get directories we can save to.

lib/registry/src/config.rs
```diff
@@ -0,0 +1,329 @@
...
+                let default_dir = Self::get_current_dir()
+                    .ok()
+                    .unwrap_or_else(|| PathBuf::from("/".to_string()));
+                #[cfg(feature = "dirs")]
+                let home_dir =
+                    dirs::home_dir().ok_or(GlobalConfigError::CannotFindHomeDirectory)?;
+                #[cfg(not(feature = "dirs"))]
+                let home_dir = std::env::var("HOME")
+                    .ok()
+                    .unwrap_or_else(|| default_dir.to_string_lossy().to_string());
```
Michael-F-Bryan, line 308: Is it worth just using the `dirs` crate unconditionally and dropping the feature flag?

lib/registry/src/graphql.rs
```diff
@@ -0,0 +1,251 @@
...
+{
+    let client = {
+        let builder = Client::builder();
+
+        #[cfg(not(target_os = "wasi"))]
+        let builder = if let Some(proxy) = proxy::maybe_set_up_proxy()? {
+            builder.proxy(proxy)
+        } else {
+            builder
+        };
```
Michael-F-Bryan, line 189: It should be fine to run `proxy::maybe_set_up_proxy()` on WASI, shouldn't it? As far as I can see, we're only reading environment variables, and those work fine on WASI.

lib/registry/src/graphql.rs
```diff
@@ -0,0 +1,251 @@
...
+        res = res.timeout(t);
+    }
+
+    let res = res.send()?;
+    let response_body: Response<R> = res.json()?;
+    if let Some(errors) = response_body.errors {
+        let error_messages: Vec<String> = errors.into_iter().map(|err| err.message).collect();
+        return Err(anyhow::anyhow!("{}", error_messages.join(", ")));
+    }
+    Ok(response_body.data.expect("missing response data"))
```
Michael-F-Bryan, line 225: Under what circumstances would we typically get a missing response data? If this is a generic utility used when executing any GraphQL query, it might be useful to turn this into an error.

Michael-F-Bryan (changes requested):

lib/cli/src/commands/login.rs
Michael-F-Bryan: Yeah, that makes sense. 

I'm just looking at what some other tools do:

```console
$ cargo login -h
Save an api token from the registry locally. If token is not specified, it will be read from stdin.

Usage: cargo login [OPTIONS] [token]

Arguments:
  [token]

Options:
  -q, --quiet                Do not print cargo log messages
      --registry <REGISTRY>  Registry to use
      ...
  -h, --help                 Print help information

Run `cargo help login` for more detailed information.

$ npm login -h
Add a registry user account

Usage:
npm adduser

Options:
[--registry <registry>] [--scope <@scope>]
[--auth-type <legacy|web|sso|saml|oauth|webauthn>]

aliases: login, add-user

Run "npm help adduser" for more info

$ yarn login -h

  Usage: yarn [command] [flags]

  Stores registry username and email.

  Options:

    -v, --version                       output the version number
    ...
    --registry <url>                    override configuration registry
    ...
    -h, --help                          output usage information
  Visit https://yarnpkg.com/en/docs/cli/login for documentation about this command.
```

fschutt: Okay, then let's be consistent with other tools.

lib/cli/src/commands/login.rs
fschutt: ok done

lib/registry/src/config.rs
fschutt: ok done

lib/registry/src/config.rs
fschutt: okay

lib/registry/src/graphql.rs
fschutt: I think there was some issue on WASI, but it doesn't matter because we don't compile wasmer to wasi anyway

lib/registry/src/graphql.rs
fschutt: No, I don't think it makes sense to strongly type this error: it only fails if - for example - the backend is down and isn't returning any data.

fschutt: Sorry, I just saw that you just wanted me to turn this into an error: yeah, that's okay

Michael-F-Bryan (changes requested):

lib/registry/src/config.rs
```diff
@@ -0,0 +1,333 @@
...
+    /// A mocked version of the standard function for integration tests
+    #[cfg(test)]
+    pub fn save(&self) -> anyhow::Result<()> {
+        let config_serialized = toml::to_string(&self)?;
+        RAW_CONFIG_DATA.with(|rcd| {
+            *rcd.borrow_mut() = Some(config_serialized);
+        });
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
fschutt: Okay, I've added that.

lib/registry/src/config.rs
```diff
@@ -0,0 +1,333 @@
...
+            }
+            Err(_e) => Ok(Self::default()),
+        }
+    }
+
+    pub fn get_current_dir() -> std::io::Result<PathBuf> {
+        #[cfg(target_os = "wasi")]
+        if let Some(pwd) = std::env::var("PWD").ok() {
+            return Ok(PathBuf::from(pwd));
+        }
```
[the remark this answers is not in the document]
fschutt: https://github.com/WebAssembly/wasi-filesystem/issues/24

I would leave this as is for now, since it worked for `wapm`, it'll work for wasmer.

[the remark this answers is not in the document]
fschutt: Also, this is only triggering when `PWD` is set for WASI

Michael-F-Bryan:
@fschutt, sorry, I think I was pushing you down the wrong path by carrying on about that `#[cfg(test)]` stuff. I think this should be good to merge if we roll back to the commit before fcaf2552568eaee0387de8f97edc351063cafb71 and get CI passing.

lib/registry/src/graphql.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    let form = Form::new()
+        .text("query", query.query.to_string())
+        .text("operationName", query.operation_name.to_string())
+        .text("variables", vars);
+
+    let form = form_modifier(form);
+
+    let user_agent = format!(
+        "wapm/{} {} {}",
```
syrusakbary, line 145: Let's change this to wasmer

lib/registry/src/graphql.rs
```diff
@@ -0,0 +1,251 @@
...
+        env!("CARGO_PKG_VERSION"),
+        whoami::platform(),
+        whoami_distro(),
+    );
+
+    let mut res = client
+        .post(registry_url)
+        .multipart(form)
+        .bearer_auth(env::var("WAPM_REGISTRY_TOKEN").unwrap_or_else(|_| login_token.to_string()))
```
syrusakbary, line 154: Let's use WASMER_TOKEN and fall back to WAPM_REGISTRY_TOKEN and then the login_token

lib/registry/src/graphql.rs
```diff
@@ -0,0 +1,251 @@
...
+        let builder = Client::builder();
+
+        let builder = if let Some(proxy) = proxy::maybe_set_up_proxy()? {
+            builder.proxy(proxy)
+        } else {
+            builder
+        };
+        builder.build()?
+    };
+
```
syrusakbary, line 134: This logic is repeated multiple times

lib/registry/src/graphql.rs
```diff
@@ -0,0 +1,251 @@
...
+pub fn whoami_distro() -> String {
+    whoami::os().to_lowercase()
+}
+
+#[cfg(not(target_os = "wasi"))]
+pub fn whoami_distro() -> String {
+    whoami::distro().to_lowercase()
+}
+
+pub fn execute_query_modifier_inner_check_json<V, F>(
```
syrusakbary, line 113: I don't think we need this funciton. Where and why is used?

fschutt: It is being used in the `test_if_registry_present` function. I've added a comment.

PR #3299, opened 2022-11-11
Fix "create-exe" for windows-x86_64 target

fschutt:


lib/cli/src/commands/create_exe.rs
```diff
@@ -562,26 +564,71 @@ impl CreateExe {
...
+            }
+
+            #[cfg(target_os = "windows")]
+            {
+                // TODO: discover include path?
+                default_include_path = Some(
+                    "C:\\Program Files (x86)\\Windows Kits\\10\\Include\\10.0.17134.0\\ucrt"
+                        .to_string(),
+                );
+            }
```
syrusakbary, line 589: Why would we need this if we are cross-compiling?

lib/cli/src/commands/create_exe.rs
```diff
@@ -1504,39 +1564,61 @@ mod http_fetch {
...
+                            if let Err(err) = std::fs::copy(&download_path, &cache_path) {
                                 eprintln!(
                                     "Could not store tarball to cache path `{}`: {}",
                                     cache_path.display(),
                                     err
                                 );
                             } else {
+                                eprintln!("copying to /Development");
+                                let _ = std::fs::File::create(
+                                    "/Users/fs/Development/wasmer-windows.tar.gz",
```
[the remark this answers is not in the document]
fschutt: just for debugging

lib/cli/src/commands/create_exe.rs
fschutt: I was not aware that `-isystem` exists, so yesterday I tried fixing the issue "manually". The issue was that `#include <stdio.h>` for example wasn't found because the zig cc command doesn't include the default system headers.

I will remove this block and replace it with `-isystem`, hopefully that will work.

lib/cli/src/commands/create_exe.rs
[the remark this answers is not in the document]
fschutt: It's removed now, I just needed it to debug why the downloaded .tar.gz was always 0 bytes large.

fschutt:
This is currently stuck on `zig cc` not properly linking the object files:

```
"/Users/fs/.wasmer/utils/zig/0.11.0-dev.183+ceb9fedb4/zig" "cc" -target x86_64-windows-gnu -L. -l:wasmer.lib -I. "-isystem" ./python.obj ./wasmer_main.c ./volume.o -o ./python.exe
```

[allfiles.zip](https://github.com/wasmerio/wasmer/files/10016030/allfiles.zip)
[allfiles 2.zip](https://github.com/wasmerio/wasmer/files/10016033/allfiles.2.zip)

fschutt:
Windows build to test:

[x86_64-windows-gnu-python.zip](https://github.com/wasmerio/wasmer/files/10020473/x86_64-windows-gnu-python.zip)

fschutt:
@syrusakbary At this point the error is fixed for every system except the `x86_64-windows` target: `zig build-exe` doesn't output a target file,but doesn't output an error either. It works for every other target (the original error of `stdio.h not found` is fixed by switching from `zig cc` to `zig build-exe`).

All I can do at this point is to report a bug to zig and ask for help.

[zigbug.zip](https://github.com/wasmerio/wasmer/files/10022578/zigbug.zip)

lib/cli/src/commands/create_exe.rs
```diff
@@ -557,35 +548,51 @@ impl CreateExe {
             header_code_path = std::env::current_dir()?;
         }
 
+        println!("Output result to: {}", output_path.display());
```
Michael-F-Bryan, line 551: I think you forgot to delete this :stuck_out_tongue:

lib/cli/src/commands/create_exe.rs
```diff
@@ -557,35 +548,51 @@ impl CreateExe {
...
+            cmd.arg(&c_src_path);
+            cmd.arg(libwasmer_path.join(&lib_filename));
+            if zig_triple.contains("windows") {
+                let mut libwasmer_parent = libwasmer_path.clone();
+                libwasmer_parent.pop();
+                cmd.arg(libwasmer_parent.join("winsdk/ADVAPI32.lib"));
+                cmd.arg(libwasmer_parent.join("winsdk/BCRYPT.lib"));
+                cmd.arg(libwasmer_parent.join("winsdk/KERNEL32.lib"));
+                cmd.arg(libwasmer_parent.join("winsdk/USERENV.lib"));
+                cmd.arg(libwasmer_parent.join("winsdk/WS2_32.lib"));
```
Michael-F-Bryan, line 588: Do we have a list of these files?

It might be a good idea to pull them into a constant at the top of the file, that way we can set up tests to make sure all the `*.lib` files are always included with Wasmer.

lib/cli/src/commands/create_exe.rs
```diff
@@ -557,35 +548,51 @@ impl CreateExe {
...
-                cmd_mut = cmd_mut.arg(volume_obj.clone());
+                cmd.arg(volume_obj.clone());
             }
 
-            cmd_mut
-                .arg("-o")
-                .arg(&output_path)
-                .output()
-                .context("Could not execute `zig`")?
+            println!("{:?}", cmd);
```
Michael-F-Bryan, line 594: We can delete this one, too.

lib/cli/src/commands/create_exe.rs
```diff
@@ -813,22 +820,22 @@ impl CreateExe {
...
                         let mut cmd_mut: &mut Command = cmd
                             .arg("cc")
                             .arg("--verbose")
                             .arg("-target")
                             .arg(&zig_triple)
-                            .arg(&format!("-L{}", libwasmer_path.display()))
-                            .arg(&format!("-l:{}", lib_filename))
-                            .arg(&format!("-I{}", include_dir.display()));
+                            .arg(&format!("-I{}", include_dir.display()))
+                            .arg("-lc");
```
Michael-F-Bryan, line 830: Do we need to do the `if windows { cmd.arg("-lc++") } else { cmd.arg("-lc") }` thing here, too?

lib/cli/src/commands/create_exe.rs
```diff
@@ -1357,30 +1362,36 @@ impl LinkCode {
...
+        // Increases rate-limiting in GitHub CI
+        let auth = std::env::var("GITHUB_TOKEN");
+        let mut response = Request::new(&uri);
+
+        if let Ok(token) = auth {
+            response.header("Authorization", &format!("Bearer {token}"));
+        }
+
+        let response = response
+            .header("User-Agent", "wasmerio")
```
Michael-F-Bryan, line 1382: I'm not sure if people check user agents much any more, but we can use `env!("CARGO_PKG_XXX")` variables to include more information if we want. 

For example, 

```rust
const USER_AGENT: &str = concat!(
  env!("CARGO_PKG_NAME"),
  " v"
  env!("CARGO_PKG_VERSION"),
);

```

lib/cli/src/commands/create_exe.rs
```diff
@@ -1357,30 +1362,36 @@ impl LinkCode {
...
         if response.status_code() != StatusCode::new(200) {
-            return Err(anyhow!(
-                "Github API replied with non-200 status code: {}",
-                response.status_code()
-            ));
+            eprintln!(
+                "Warning: Github API replied with non-200 status code: {}. Response: {}",
+                response.status_code(),
+                String::from_utf8_lossy(&writer),
+            );
```
Michael-F-Bryan, line 1394: It might be better to use `log::warn!()` here. That way the warning can go into our logging system, which might end up being saved to a log file or sent to analytics.

lib/cli/src/commands/create_exe.rs
```diff
@@ -1443,31 +1454,64 @@ mod http_fetch {
...
+
+            if let Ok(mut entries) = paths {
+                entries.retain(|p| p.to_str().map(|p| p.ends_with(".tar.gz")).unwrap_or(false));
+
+                // create-exe on Windows is special: we use the windows-gnu64.tar.gz (GNU ABI)
+                // to link, not the windows-amd64.tar.gz (MSVC ABI)
+                if target_triple.operating_system == OperatingSystem::Windows {
+                    entries.retain(|p| {
+                        p.to_str()
+                            .map(|p| p.contains("windows") && p.contains("gnu64"))
```
Michael-F-Bryan, line 1472: Does `create-exe` support 32-bit Windows?

lib/cli/src/commands/create_exe.rs
```diff
@@ -1482,142 +1526,123 @@ mod http_fetch {
...
+        };
+
+        let filename = browser_download_url
+            .split('/')
+            .last()
+            .unwrap_or("output")
+            .to_string();
+
+        let download_tempdir = tempdir::TempDir::new("wasmer-download")?;
+        let download_path = download_tempdir.path().to_path_buf().join(&filename);
```
Michael-F-Bryan, line 1568: You should be able to drop the `.to_path_buf()` and call `download_tempdir.path().join(...)` without the intermediate copy.

lib/cli/src/commands/create_exe.rs
```diff
@@ -1482,142 +1526,123 @@ mod http_fetch {
...
+
+        let mut file = std::fs::File::create(&download_path)?;
+        eprintln!(
+            "Downloading {} to {}",
+            browser_download_url,
+            download_path.display()
+        );
```
Michael-F-Bryan, line 1575: See my previous comment about using a logger instead of print statements.

lib/cli/src/commands/create_exe.rs
```diff
@@ -1482,142 +1526,123 @@ mod http_fetch {
...
+
+        let uri = Uri::try_from(browser_download_url.as_str())?;
+        let mut response = Request::new(&uri)
+            .header("User-Agent", "wasmer")
+            .send(&mut file)
+            .map_err(anyhow::Error::new)
+            .context("Could not lookup wasmer artifact on Github.")?;
+        if response.status_code() == StatusCode::new(302) {
```
Michael-F-Bryan, line 1583: It might be better to check against the `StatusCode::FOUND` constant rather than the number.

We can also tell `reqwest` to follow redirects automatically with [`ClientBuilder::redirect()`](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.redirect) and drop the explicit redirect handling. By default, it will follow up to 10 steps before giving up.

lib/registry/src/lib.rs
```diff
@@ -643,77 +642,94 @@ pub fn get_global_install_dir(
...
+    let target_targz_path = tempdir.path().join("package.tar.gz");
+
+    let mut resp = reqwest::blocking::get(url)
+        .map_err(|e| anyhow::anyhow!("failed to download {url}: {e}"))?;
+
+    if !target_targz_path.exists() {
+        // create all the parent paths, only remove the created directory, not the parent dirs
+        let _ = std::fs::create_dir_all(&target_targz_path);
+        let _ = std::fs::remove_dir(&target_targz_path);
+    }
```
Michael-F-Bryan, line 717: Is this necessary? I thought `TempDir::new()` would always give you the name of an empty directory, so `tempdir.path().join("package.tar.gz")` will never exist.

lib/registry/src/lib.rs
```diff
@@ -643,77 +642,94 @@ pub fn get_global_install_dir(
...
+                target_targz_path.display()
+            )
+        })?;
+
+        resp.copy_to(&mut file)
+            .map_err(|e| anyhow::anyhow!("{e}"))?;
+    }
+
+    try_unpack_targz(target_targz_path.as_path(), target_path, strip_toplevel)
+        .context(anyhow::anyhow!("Could not download {url}"))?;
```
Michael-F-Bryan, line 732: I think you meant `.with_context(|| anyhow::anyhow!("Could not download {url}"))` here. 

The `context()` method accepts a value, which means `anyhow::anyhow!()` will be capturing backtraces and creating errors every time we reach this line. Even if there were no errors.

tests/integration/cli/tests/run.rs
```diff
@@ -21,6 +21,66 @@ fn test_no_start_wat_path() -> String {
...
+        "x86_64-darwin",
+        "x86_64-linux-gnu",
+        "aarch64-linux-gnu",
+        // TODO: this test depends on the latest release -gnu64.tar.gz
+        // to be present, but we can't release the next release before
+        // the integration tests are passing, so this test depends on itself
+        //
+        // We need to first release a version without -windows being tested
+        // then do a second PR to test that Windows works.
+        // "x86_64-windows-gnu",
```
Michael-F-Bryan, line 40: Is it worth adding a link to the ticket so we don't forget about this?

lib/cli/src/commands/create_exe.rs
fschutt: No, it's just the libraries that I found out that libwasmer.a depends on.

There is no way to find out the libraries, except for using Google to lookup which DLL the missing functions belong to.

lib/cli/src/commands/create_exe.rs
fschutt: According to GitHub, the User-Agent should be the organization name, and there is no way to get that from env vars.

lib/cli/src/commands/create_exe.rs
fschutt: No, we don't support that.

lib/cli/src/commands/create_exe.rs
Michael-F-Bryan: Oh no, what I meant was instead of copy-pasting the `cmd.arg(libwasmer_parent.join(...))` line for every library, we should store them in a constant and use a loop. 

```rust
const VENDORED_LIBRARIES: &[&str] = &[
    "ADVAPI32.lib",
    "BCRYPT.lib",
    ...
];
```

That way we can add a test to the `wasmer` CLI that makes sure all the libraries are embedded in the installer. It'd be annoying if the list of libraries we tell the compiler to link against and the list of libraries bundled with `wasmer` got out of sync. We probably wouldn't notice until weeks later when our first user wants to do networking and finds we forgot to give them `WS2_32.lib`.

lib/cli/src/commands/create_exe.rs
Michael-F-Bryan: Awesome!

lib/cli/src/commands/create_exe.rs
Michael-F-Bryan: Can we use `log::debug!()`, then? End users probably don't want to see the full command-line we are using to generate EXEs.

lib/cli/src/commands/create_exe.rs
Michael-F-Bryan: Ah thank, I didn't know that.

tests/integration/cli/tests/run.rs
fschutt: No because I know I need to fix this before 3.0.0, so the ticket would only be up for a few hours.

lib/cli/src/commands/create_exe.rs
fschutt: migrated to reqwest

lib/cli/src/commands/create_exe.rs
fschutt: I've refactored this so we don't need this code block at all anymore.

Also this code block is impossible to hit because cross-compilation + ObjectFormat::Serialized doesn't work in wasmer.

lib/cli/src/commands/create_exe.rs
fschutt: I've made a loop with `fs::read_dir` now, so we don't need to hardcode any paths, it will just link to every object in the `/winsdk` folder.

lib/cli/src/commands/create_exe.rs
fschutt: yes, I migrated to `log::debug`

PR #3302, opened 2022-11-14
Some Refactor of Singlepass compiler to have better error and cpu features handling

ptitSeb:
# Description

* Some refactor of Singlepass compiler to have a better handling of Target CPU Feature and Error propagation
* Added support for TZCNT and LZCNT opcode on x86_64 backend (based on https://github.com/near/wasmer/pull/145)
For #3305


lib/compiler-singlepass/src/compiler.rs
```diff
@@ -228,7 +207,7 @@ impl Compiler for SinglepassCompiler {
             .values()
             .collect::<Vec<_>>()
             .into_par_iter_if_rayon()
-            .map(|func_type| gen_std_trampoline(func_type, target, calling_convention))
+            .map(|func_type| gen_std_trampoline(func_type, target, calling_convention).unwrap())
```
syrusakbary, line 210: This will now panic if the chipset is not supported (for example, when cross compiling). Do we really want this?

ptitSeb: it was panicking before too.

Also, note that this is only on building the tranpoline, that will not panic if chipset is unsupported, as it's more permissive.

(Anyway, there is no easy way to get the Error to interupt the rayon iterator)

syrusakbary (approved):

theduke:
I gave it a quick pass, but I don't have anything valuable to add.

PR #3316, opened 2022-11-17
Implement wasmer whoami

fschutt:
Note: has to be rebased on the `wasmer login` branch, otherwise the integration test won't work


lib/cli/src/cli.rs
```diff
@@ -151,6 +153,10 @@ enum WasmerCLIOptions {
     #[clap(name = "binfmt")]
     Binfmt(Binfmt),
 
+    /// Shows the current logged in user for the current active registry
+    #[clap(name = "whoami")]
```
Michael-F-Bryan, line 157: fyi - if you don't use the `#[clap(name = "...")]` attribute, `clap` will default to using the lowercased version of the enum variant for a subcommand's name (i.e. `whoami`) so we don't need to add the attribute.

lib/cli/src/commands/whoami.rs
```diff
@@ -0,0 +1,18 @@
...
+    /// Which registry to check the logged in username for
+    #[clap(long, name = "registry")]
+    pub registry: Option<String>,
+}
+
+impl Whoami {
+    /// Execute `wasmer whoami`
+    pub fn execute(&self) -> Result<(), anyhow::Error> {
+        let (registry, username) = wasmer_registry::whoami(self.registry.as_deref())?;
+        println!("logged into registry {registry:?} as user {username:?}");
```
Michael-F-Bryan, line 15: Can you switch this over to a `log` statement?

lib/registry/src/lib.rs
```diff
@@ -894,6 +894,48 @@ pub fn install_package(
...
+    use graphql_client::GraphQLQuery;
+
+    #[cfg(test)]
+    let config = PartialWapmConfig::from_file(test_name);
+    #[cfg(not(test))]
+    let config = PartialWapmConfig::from_file();
+
+    let config = config
+        .map_err(|e| anyhow::anyhow!("{e}"))
+        .context(anyhow::anyhow!("{registry:?}"))?;
```
Michael-F-Bryan, line 911: We can drop the `map_err()` because the `context()` method should automatically wrap an error in an `anyhow::Error`. 

Also, we need to use `with_context()` instead of `context()` if constructing the context requires function calls. This is similar to [Clippy's lint](https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call) about unconditionally making expensive calls with side-effects when they should only be triggered during an error.

```suggestion
        .with_context(|| format!("{registry:?}"))?;
```

lib/registry/src/lib.rs
```diff
@@ -894,6 +894,48 @@ pub fn install_package(
...
+    let login_token = config
+        .registry
+        .get_login_token_for_registry(&registry)
+        .ok_or_else(|| anyhow::anyhow!("not logged into registry {:?}", registry))?;
+
+    let q = WhoAmIQuery::build_query(who_am_i_query::Variables {});
+    let response: who_am_i_query::ResponseData =
+        crate::graphql::execute_query(&registry, &login_token, &q)
+            .map_err(|e| anyhow::anyhow!("{e}"))
+            .context(anyhow::anyhow!("{registry:?}"))?;
```
Michael-F-Bryan, line 927: See my previous suggestion about `with_context()`.

```suggestion
           .with_context(|| format!("{registry:?}"))?;
```

tests/integration/cli/tests/run.rs
```diff
@@ -81,6 +81,53 @@ fn test_cross_compile_python_windows() -> anyhow::Result<()> {
...
+        );
+    }
+
+    assert_eq!(
+        stdout,
+        "logged into registry \"https://registry.wapm.dev/graphql\" as user \"ciuser\"\n"
+    );
+
+    Ok(())
+}
```
Michael-F-Bryan, line 129: I don't think this test is a good idea.

It assumes the `$WAPM_DEV_TOKEN` variable is set and that user's name is `ciuser`, so the only way for integration tests to pass locally is if you have the `ciuser` token on your computer, which is a massive no-no (no human developers should have access to CI tokens).

We're also trying to contact wapm.dev, which makes the test slow and flaky (e.g. wapm.dev could be having a bad day and always error out).

Michael-F-Bryan (changes requested):
Is there a different way to check whether the `whoami` command succeeds without requiring access to the `ciuser` token?

Developers shouldn't have access to the `ciuser` token because it means they can masquerade as a system process (which screws up the audit trail), and that token has admin rights on WAPM so we'd be screwed if the wrong people got their hands on it.

lib/cli/src/cli.rs
fschutt: Yeah I just did it to stay consistent with the current implementation, I know that it does do that.

Should I remove it?

lib/cli/src/commands/whoami.rs
fschutt: In this case it needs to be always printed, with log it would prefix a `LOG ...` statement which I'd really not want

Michael-F-Bryan: Good point!

lib/cli/src/cli.rs
fschutt: okay I removed it

tests/integration/cli/tests/run.rs
fschutt: Yeah well with network requests, that's kind of a given, otherwise it wouldn't be an integration test. In the worst case the test fails because wapm.dev is down, then we just need to restart the tests on GitHub.

The other part, I think I can require `WAPM_DEV_TOKEN ` if `GITHUB_TOKEN` is present (true for integration tests in the CI, false for running tests locally).

Michael-F-Bryan (approved):
I think checking for `$GITHUB_TOKEN` and assuming the test passes is a pretty good trade-off for now :+1:

PR #3319, opened 2022-11-17
Disable 'Test integration CLI' on CI for the Windows platform as it's not working at all

ptitSeb:
# Description
Disable 'Test integration CLI' on CI for the Windows platform as it's not working at all
To be re-enabled later when it's fixed for this platform


fschutt:
I submitted a tracking issue, so we can track it after the release.

PR #3324, opened 2022-11-18
Implement wasmer init and wasmer publish

fschutt:


lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,346 @@
...
+            .lines()
+            .collect::<Vec<_>>()
+            .join("\r\n");
+
+            std::fs::write(target_file, &toml_string)?;
+        }
+
+        Ok(())
+    }
+}
```
Michael-F-Bryan, line 357: Are you able to refactor `Init::execute()` and introduce some higher-level abstractions?

I'm not trying to be pedantic here, there's lots of stuff going on with lots of implicit assumptions and it's not immediately apparent what `wasmer init` will actually generate.

For example, if people can choose templates then I'd expect there to be separate functions for setting up JavaScript and Python. Or possibly some `Template` struct which is populated with the source for Python/JS-specific files.

Similarly, we spend a lot of lines special-casing `Cargo.toml`, so that should be split out into its own set of functions for retrieving information from `cargo` and constructing the `wapm_toml::Manifest`.

Michael-F-Bryan (changes requested):

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,346 @@
...
+            (false, _, true) => BinOrLib::Lib,
+            _ => BinOrLib::Bin,
+        };
+
+        let module_name = package_name.split('/').next().unwrap_or(&package_name);
+
+        let modules = vec![wapm_toml::Module {
+            name: module_name.to_string(),
+            source: if cargo_toml.is_some() {
+                Path::new(&format!("target/wasm32-wasi/release/{module_name}.wasm")).to_path_buf()
```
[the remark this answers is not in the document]
fschutt: ok done

Michael-F-Bryan (approved):
I think we've got the core of `wasmer init` here. Let's merge it once CI passes and make a follow-up PR refactoring `Init::execute()`.

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+    /// If the `manifest-path` is a Cargo.toml, use that file to initialize the wapm.toml
+    #[clap(long, name = "manifest-path")]
+    pub manifest_path: Option<PathBuf>,
+    /// Add default dependencies for common packages (currently supported: `python`, `js`)
+    #[clap(long, name = "template")]
+    pub template: Option<String>,
+    /// Include file paths into the target container filesystem
+    #[clap(long, name = "include")]
+    pub include: Vec<String>,
+    /// Directory of the output file name. wasmer init will error in the target dir
```
Michael-F-Bryan, line 43: Small typo.

```suggestion
    /// Directory of the output file name. wasmer init will error if the target dir
```

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+    pub namespace: Option<String>,
+    /// Version of the initialized package
+    #[clap(long, name = "version")]
+    pub version: Option<semver::Version>,
+    /// If the `manifest-path` is a Cargo.toml, use that file to initialize the wapm.toml
+    #[clap(long, name = "manifest-path")]
+    pub manifest_path: Option<PathBuf>,
+    /// Add default dependencies for common packages (currently supported: `python`, `js`)
+    #[clap(long, name = "template")]
+    pub template: Option<String>,
```
Michael-F-Bryan, line 39: Can we use [the `#[derive(clap::ValueEnum)]` attribute] here? It'll let us specify the template using an enum (less string comparisons = better).

`clap` will will automatically set `possible_values = ["js", "python"]` so people can see which values are available, and it'll error out if they enter the wrong thing (possibly with a suggestion if they do a typo).

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+                    .ok_or_else(|| anyhow::anyhow!("no dir name"))?;
+                s.join(WASMER_TOML_NAME)
+            }
+        };
+
+        if target_file.exists() && !self.overwrite {
+            return Err(anyhow::anyhow!(
+                "wapm project already initialized in {}",
+                target_file.display()
+            ));
```
Michael-F-Bryan, line 108: It doesn't help much here because the message's length and current indentation level mean the return is spread over multiple lines, but sometimes `anyhow::bail!()` is more concise.

```suggestion
            anyhow::bail!(
                "wapm project already initialized in {}",
                target_file.display(),
            );
```

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+
+        let cargo_toml = if manifest_path.exists() {
+            use anyhow::Context;
+
+            let mut metadata = MetadataCommand::new();
+            metadata.manifest_path(&manifest_path);
+            metadata.no_deps();
+            metadata.features(CargoOpt::AllFeatures);
+
+            println!("{:#?}", metadata.cargo_command());
```
Michael-F-Bryan, line 134: Should this be turned into a debug log statement?

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+                license_file: package.license_file.clone().map(|f| f.into_std_path_buf()),
+                workspace_root: metadata.workspace_root.into_std_path_buf(),
+                build_dir: metadata
+                    .target_directory
+                    .into_std_path_buf()
+                    .join("wasm32-wasi"),
+            })
+        } else {
+            None
+        };
```
Michael-F-Bryan, line 168: Can we pull all of this out into its own `read_cargo_toml()` function? Moving "boring" logic out into its own functions helps with readability because I can skim through `Init::execute()` and see all the interesting bits without having to care about exactly how we get a `MiniCargoTomlPackage`.

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+
+        let package_name = cargo_toml
+            .as_ref()
+            .map(|p| &p.name)
+            .unwrap_or(&package_name);
+
+        let namespace = self.namespace.clone().unwrap_or_else(|| {
+            let username = wasmer_registry::whoami(None, None).ok().map(|o| o.1);
+            username
+                .or_else(|| package_name.split('/').next().map(|s| s.to_string()))
```
Michael-F-Bryan, line 178: This feels like a code smell. 

If the `package_name` came from the `cargo_toml` variable then it can never contain a `/`. Similarly, if you scroll up about 100 lines you'll see that `package_name` is constructed using *just* the `file_stem()` of the current directory. That means the `split()` will always return the entire `package_name` string and we don't need to deal with any `Option`s.

The deeper problem is that we shouldn't be using `string.split("/")` when dealing with package names because package names are more nuanced than just a string. Instead, we should try to use a high-level type and do something like `string.parse::<PackageName>()` so we can access its namespace.

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+            module: match bin_or_lib {
+                BinOrLib::Empty => None,
+                _ => Some(modules),
+            },
+            fs: self.get_filesystem_mapping(),
+            base_directory_path: target_file
+                .parent()
+                .map(|o| o.to_path_buf())
+                .unwrap_or_else(|| target_file.clone()),
+        };
```
Michael-F-Bryan, line 260: Can we pull all of this code out into its own `construct_manifest()` function? That way the `Init::execute()` function isn't filled with 100 lines of copying variables from `self` and `cargo_toml` into a `wapm_toml::Manifest`.

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+        let note =
+            "# See more keys and definitions at https://docs.wasmer.io/ecosystem/wapm/manifest";
+
+        // generate the wasmer.toml and exit
+        if !should_add_to_cargo_toml {
+            let toml_string = toml::to_string_pretty(&constructed_manifest)?
+                .replace("[dependencies]", &format!("{note}\r\n\r\n[dependencies]"))
+                .lines()
+                .collect::<Vec<_>>()
+                .join("\r\n");
```
Michael-F-Bryan, line 308: Should we be adding `\r` here?

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+
+            std::fs::write(target_file, &toml_string)?;
```
Michael-F-Bryan, line 310: Let's mention the filename when a write fails.

```suggestion
            std::fs::write(&target_file, &toml_string)
                .with_context(|| format!("Unable to write to \"{}\"", target_file.display()))?;
```

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+            eprintln!();
+            eprintln!("    cargo install cargo-wapm");
+            eprintln!();
+        }
+
+        let note =
+            "# See more keys and definitions at https://docs.wasmer.io/ecosystem/wapm/manifest";
+
+        // generate the wasmer.toml and exit
+        if !should_add_to_cargo_toml {
```
Michael-F-Bryan, line 303: Can we refactor from here until the end of the function to be something like this?

```rust
if should_add_to_cargo_toml {
  add_metadata_to_cargo_toml()?;
} else {
  save_wapm_toml()?;
}

Ok(())
```

That way it's obvious that there are two ways we can finish initializing the package, one of which is to add our WAPM metadata to `Cargo.toml`, and the other is to save the `wapm_toml::Manifest` to `wapm.toml`.

Also, having those two operations in their own functions makes `Init::execute()` easier to understand later on. 

For some context on why I'm so picky about helper functions... At the first company I worked at, the primary software developer was someone who would routinely write 500+ line functions with lots of nesting (I once counted 10+  and never refactored his code into more digestible chunks... the company ended up throwing his entire 200kloc program out after he retired and starting from scratch because it was completely unmaintainable, buggy, and nobody wanted to touch it with a ten foot pole. I used to absolutely dread fixing bugs in his code, and really don't want Wasmer's codebase to turn out that way.

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+                }
+                Some("python") => {
+                    let mut map = HashMap::default();
+                    map.insert("python".to_string(), "python/python@latest".to_string());
+                    map
+                }
+                _ => HashMap::default(),
+            }
+        })
+    }
```
Michael-F-Bryan, line 422: We can probably get rid of some nesting here:

```suggestion
    fn get_dependencies(&self) -> Option<HashMap<String, String>> {
        let mut map = HashMap::new();
        
        match self.template.as_deref() {
            Some("js") => {
                map.insert("python".to_string(), "quickjs/quickjs@latest".to_string());
            }
            Some("python") => {
                map.insert("python".to_string(), "python/python@latest".to_string());
            }
            _ => {},
            }
            
            Some(map)
    }
```

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+        }
+    }
+
+    fn get_bindings(target_file: &Path, bin_or_lib: BinOrLib) -> Option<wapm_toml::Bindings> {
+        match bin_or_lib {
+            BinOrLib::Bin | BinOrLib::Empty => None,
+            BinOrLib::Lib => target_file.parent().and_then(|parent| {
+                walkdir::WalkDir::new(parent)
+                    .min_depth(1)
+                    .max_depth(3)
```
Michael-F-Bryan, line 446: I don't know if we want to artificially limit the search depth to 1-3 levels, here.

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+                            Some(wapm_toml::Bindings::Wai(wapm_toml::WaiBindings {
+                                exports: None,
+                                imports: vec![e.path().to_path_buf()],
+                                wai_version: semver::Version::parse("0.1.0").unwrap(),
+                            }))
+                        } else {
+                            None
+                        }
+                    })
+                    .next()
```
Michael-F-Bryan, line 468: What happens when there are multiple `*.wai` or `*.wit` files in a folder? For example, a WebAssembly module could be exporting `whatever.wai`, but importing `logger.wai` and `fs.wai` from the host.

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+                        if is_wit {
+                            Some(wapm_toml::Bindings::Wit(wapm_toml::WitBindings {
+                                wit_exports: e.path().to_path_buf(),
+                                wit_bindgen: semver::Version::parse("0.1.0").unwrap(),
+                            }))
+                        } else if is_wai {
+                            Some(wapm_toml::Bindings::Wai(wapm_toml::WaiBindings {
+                                exports: None,
+                                imports: vec![e.path().to_path_buf()],
+                                wai_version: semver::Version::parse("0.1.0").unwrap(),
```
Michael-F-Bryan, line 462: WAI is currently at version `0.2.0` (technically `0.2.2`, but this will match any `0.2.x`).

```suggestion
                                wai_version: semver::Version::parse("0.2.0").unwrap(),
```

Michael-F-Bryan (changes requested):

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,647 @@
...
+        let modules = manifest.module.as_ref().ok_or(PublishError::NoModule)?;
+        let manifest_string = toml::to_string(&manifest)?;
+
+        let readme = package.readme.as_ref().and_then(|readme_path| {
+            let normalized_path = normalize_path(&manifest.base_directory_path, readme_path);
+            if builder
+                .append_path_with_name(&normalized_path, readme_path)
+                .is_err()
+            {
+                // TODO: Maybe do something here
```
Michael-F-Bryan, line 110: Should we do a `log::warn!()` here so the error doesn't get silently dropped?

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,647 @@
...
+                        .metadata()
+                        .map_err(|_| PublishError::MissingBindings {
+                            module: module.name.clone(),
+                            path: normalized_path.clone(),
+                        })?;
+                    builder
+                        .append_path_with_name(&normalized_path, &module.source)
+                        .map_err(|e| PublishError::ErrorBuildingPackage(module.name.clone(), e))?;
+                }
+            }
```
Michael-F-Bryan, line 153: Can we pull this into its own method?

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,647 @@
...
+
+        builder.finish().ok();
+        let tar_archive_data = builder.into_inner().map_err(|_| PublishError::NoModule)?;
+        let archive_name = "package.tar.gz".to_string();
+        let archive_dir = tempfile::TempDir::new()?;
+        let archive_dir_path: &std::path::Path = archive_dir.as_ref();
+        fs::create_dir(archive_dir_path.join("wapm_package"))?;
+        let archive_path = archive_dir_path.join("wapm_package").join(&archive_name);
+        let mut compressed_archive = fs::File::create(&archive_path).unwrap();
+        let mut gz_enc = GzEncoder::new(&mut compressed_archive, Compression::best());
```
Michael-F-Bryan, line 183: Instead of writing to a temporary directory, couldn't we just save to a `Vec<u8>` or write to the final `*.tar.gz` file directly?

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,647 @@
...
+
+        let namespace = self
+            .namespace
+            .as_deref()
+            .or_else(|| package.name.split('/').next())
+            .unwrap_or("")
+            .to_string();
+        if username != namespace {
+            return Err(anyhow::anyhow!("trying to publish package under the namespace {namespace:?}, but logged in as user {username:?}"));
+        }
```
Michael-F-Bryan, line 218: Won't this error out when someone is publishing to a namespace (e.g. when `ciuser` tries publishing to `wasmer`)?

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,647 @@
...
+                .map_err(|e| {
+                    MigrationError::TransactionFailed(migration_number, format!("{}", e))
+                })?;
+        }
+        2 => {
+            tx.execute_batch(include_str!("../../sql/migrations/0002.sql"))
+                .map_err(|e| {
+                    MigrationError::TransactionFailed(migration_number, format!("{}", e))
+                })?;
+        }
```
Michael-F-Bryan, line 369: Instead of writing a big `match` with lots of copy/paste, could we store this in an array constant at the top of the file?

```rust
const MIGRATIONS: &[(i32, &str)] = &[
  (0, include_str!("../../sql/migrations/0000.sql")),
  ...
];
```

Then, you'd do a `MIGRATIONS.iter().find_map(|(n, sql)| if n == migration_number { Some(sql) } else { None })` to find the right migration or handle errors appropriately.

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,647 @@
...
+        .next();
+
+    if let Some(res) = result {
+        Ok(res?)
+    } else {
+        Err(anyhow!("No active key found"))
+    }
+}
+
+mod interfaces {
```
Michael-F-Bryan, line 431: Should this be moved to its own file?

tests/integration/cli/tests/publish.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    if let Some(token) = wapm_dev_token {
+        cmd.arg("--token");
+        cmd.arg(token);
+    }
+
+    let output = cmd
+        .stdin(Stdio::null())
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
```
Michael-F-Bryan, line 52: Note that `Stdio::inherit()` will write *directly* to stdout/stderr, so the output from `wasmer publish` (or any other command we run) will get interleaved with the test runner's output. That could make it annoying to debug failing tests in CI.

Michael-F-Bryan (changes requested):

tests/integration/cli/tests/publish.rs
fschutt: The integration tests should be run with `--test-threads 1` (running the tests in parallel is relatively useless, since they run very quick). So this will make it possible to have nice output when looking at the logs. I would leave this as is, except for the final test that tests `output.stdout` (we can't use inherit here because we have to capture the stdout).

lib/wasm-interface/README.md
```diff
@@ -0,0 +1,88 @@
...
+  (func (import "wasi_unstable" "poll_oneoff") (param i32 i32 i32 i32) (result i32))
+  (func (import "wasi_unstable" "proc_exit") (param i32))
+  (func (import "wasi_unstable" "proc_raise") (param i32) (result i32))
+  (func (import "wasi_unstable" "random_get") (param i32 i32) (result i32))
+  (func (import "wasi_unstable" "sched_yield") (result i32))
+  (func (import "wasi_unstable" "sock_recv") (param i32 i32 i32 i32 i32 i32) (result i32))
+  (func (import "wasi_unstable" "sock_send") (param i32 i32 i32 i32 i32) (result i32))
+  (func (import "wasi_unstable" "sock_shutdown") (param i32 i32) (result i32))
+)
+```
```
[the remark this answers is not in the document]
fschutt: I'm not sure but for WASIX we'd probably have to do a rework, for now I just 1:1 copied the necessary code from wapm-cli.

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,647 @@
...
+
+    /// Gets the current time in our standard format
+    pub fn get_current_time_in_format() -> Option<String> {
+        use time::format_description::well_known::Rfc3339;
+        let cur_time = time::OffsetDateTime::now_utc();
+        cur_time.format(&Rfc3339).ok()
+    }
+}
+
+mod validate {
```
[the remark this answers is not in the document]
fschutt: I would like to keep it here, so that we have all of the publishing logic in one file (init.rs, login.rs, publish.rs). But it should be in a separate module so it doesn't interfere with anything else.

lib/cli/src/commands/publish.rs
fschutt: I would like to keep it here, so that we have all of the publishing logic in one file (init.rs, login.rs, publish.rs). But it should be in a separate module so it doesn't interfere with anything else.

lib/cli/src/commands/publish.rs
fschutt: Yeah, just be aware that this part of the code is copy-pasted, not my original code. I just copied + refactored it for now.

lib/cli/src/commands/publish.rs
fschutt: The file may be very big.

Also, I think that that the tar crate doesn't work on `Vec<u8>` at the moment, but I might be misremembering things.

lib/cli/src/commands/publish.rs
fschutt: Hm, yeah, we should have a query `can_publish_package` - I just wanted to avoid that the user has to wait for uploading and then at the end it errors out with "cannot publish package, you don't have rights".

lib/cli/src/commands/init.rs
fschutt: It's supposed to prevent accidentally searching things like `.git/objects/...` and build directories, which would maybe make it slow. `wasmer init` is not supposed to be perfect, it's just supposed to give a rough template that the user can edit.

lib/wasm-interface/README.md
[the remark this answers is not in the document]
theduke: I don't know the context here, but: for wasi there is also the `snapshot1` , which is also commonly used , not just `wasi_unstable`.

And yes, wasix will have a different namespace and additional functions.
Also, there is a 32bit and a 64bit version.
(Wasmer doesn't support 64bit yet though)

tests/integration/cli/tests/publish.rs
```diff
@@ -0,0 +1,167 @@
...
+    cmd.arg("target");
+    cmd.arg("add");
+    cmd.arg("wasm32-wasi");
+
+    let _ = cmd
+        .stdin(Stdio::null())
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .output()
+        .unwrap();
```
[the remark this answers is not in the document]
fschutt: ok done

Makefile
```diff
@@ -549,6 +549,7 @@ test-examples:
 	$(CARGO_BINARY) test $(CARGO_TARGET) --release $(compiler_features) --features wasi --examples
 
 test-integration-cli:
+	rustup target add wasm32-wasi
```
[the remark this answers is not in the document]
fschutt: ok done

tests/integration/cli/tests/publish.rs
fschutt: Fixed now, the tests should pass properly since they capture stdout.

lib/cli/src/commands/publish.rs
fschutt: Fixed

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+
+/// CLI args for the `wasmer init` command
+#[derive(Debug, Parser)]
+pub struct Init {
+    /// Initialize wapm.toml for a library package
+    #[clap(long, name = "lib")]
+    pub lib: bool,
+    /// Initialize wapm.toml for a binary package
+    #[clap(long, name = "bin")]
+    pub bin: bool,
```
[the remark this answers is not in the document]
fschutt: ok done

lib/cli/src/commands/init.rs
fschutt: It's for Windows, if we don't do it we risk problems on Windows not showing the Cargo.toml correctly.

lib/cli/src/commands/publish.rs
fschutt: I think we should leave this as is. For now, a tempdir has never caused problems.

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,472 @@
...
+                    .unwrap_or_else(|| std::env::current_exe().unwrap())
+                    .join("Cargo.toml");
+                cargo_toml_path
+                    .canonicalize()
+                    .unwrap_or_else(|_| cargo_toml_path.clone())
+            }
+        };
+
+        let cargo_toml = if manifest_path.exists() {
+            use anyhow::Context;
```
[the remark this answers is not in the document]
fschutt: ok done

lib/cli/src/commands/init.rs
fschutt: ok done

lib/cli/src/commands/init.rs
fschutt: ok done

lib/cli/src/commands/init.rs
fschutt: ok done

lib/cli/src/commands/publish.rs
fschutt: ok done

lib/wasm-interface/README.md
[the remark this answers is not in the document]
fschutt: For now I'll leave this as-is since I'm only porting the functionality we already have in wapm. If we notice problems after 3.2, we can still go in and edit the definitions again.

lib/cli/src/commands/publish.rs
fschutt: It's implemented now, but the functionality of --namespace is still a bit lacking since it doesn't allow you to change the namespace when uploading.

lib/cli/src/commands/init.rs
fschutt: This is for cases where the name is constructed from the `wasmer.toml`, in the `init` step, I am currently setting the default name to `_` (since that user doesn't exist, we reserve it for commands), so it's a good default.

The `--namespace` option exists so I can change the username quickly.

Makefile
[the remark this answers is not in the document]
Michael-F-Bryan: If it's handled by our CI's setup code, don't forget to remove this line from the `Makefile`.

lib/cli/src/commands/init.rs
Michael-F-Bryan: I've pushed a commit that fixes this to only add `\r` characters on Windows.

Michael-F-Bryan:
Hey @fschutt, I think there's an issue with the logic `wasmer init` uses to populate the `[package.metadata.wapm]` table.

When I ran `wasmer init` from inside the `lib/cli/` folder, it added the following to the end of `Cargo.toml`:

```toml
[package.metadata.wapm]
namespace = 'wasmer-workspace'
package = 'wasmer-workspace'
abi = 'wasi'
```

I don't think the `namespace` and `package` fields are meant to be the same thing. 

The `cargo wapm` subcommand will infer our package inferred from the `name` key under `[package]`, so we can remove that explicit `package = 'wasmer-workspace'`. Using `wasmer-workspace` is kinda wrong anyway because the "current" crate (i.e. the name from the `Cargo.toml` in the current directory) is `wasmer-cli`. It seems to have taken the workspace "root", which is just the package that happens to have a `[workspace]` table if you aren't using [virtual manifests](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-manifest).

Using the workspace root (i.e. `wasmer-workspace`) as the namespace doesn't seem very helpful. I would have expected the namespace to be whatever username you get from `wasmer whoami` or failing that, maybe the name of the current user.

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,545 @@
...
+        }
+    }
+
+    /// Returns the dependencies based on the `--template` flag
+    fn get_dependencies(template: Option<&Template>) -> HashMap<String, String> {
+        let mut map = HashMap::default();
+
+        match template {
+            Some(Template::Js) => {
+                map.insert("python".to_string(), "quickjs/quickjs@latest".to_string());
```
Michael-F-Bryan, line 334: I was testing `wasmer init` locally noticed we have a copy/paste bug here.

```toml
[dependencies]
python = 'quickjs/quickjs@latest'
```

```suggestion
                map.insert("quickjs".to_string(), "quickjs/quickjs@latest".to_string());
```

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,549 @@
...
+        let username = wasmer_registry::whoami(None, None).ok().map(|o| o.1);
+        username
+            .or_else(|| package_name.split('/').next().map(|s| s.to_string()))
+            .unwrap_or_else(|| "_".to_string())
+    });
+    let module_name = package_name
+        .split('/')
+        .last()
+        .unwrap_or(package_name)
+        .to_string();
```
Michael-F-Bryan, line 426: I think [my previous comment](https://github.com/wasmerio/wasmer/pull/3324#discussion_r1047083376) got automatically marked as outdated when we pulled `construct_manifest()` out into its own function.

> This feels like a code smell.
> 
> If the package_name came from the cargo_toml variable then it can never contain a /. Similarly, if you scroll up about 100 lines you'll see that package_name is constructed using just the file_stem() of the current directory. That means the split() will always return the entire package_name string and we don't need to deal with any Options.
> 
> The deeper problem is that we shouldn't be using string.split("/") when dealing with package names because package names are more nuanced than just a string. Instead, we should try to use a high-level type and do something like string.parse::<PackageName>() so we can access its namespace.

Makefile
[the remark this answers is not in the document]
fschutt: ok done

lib/cli/src/commands/init.rs
fschutt: Continued in the follow-up comment: https://github.com/wasmerio/wasmer/pull/3324#discussion_r1049349754

lib/cli/src/commands/init.rs
fschutt: I don't know how to get the imports / exports from a `.wai` file, do I need to parse the file?

fschutt: okay, I added an error message

lib/cli/src/commands/publish.rs
fschutt: This will be resolved with the new `wapm-toml` version.

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,605 @@
...
+    pub empty: bool,
+    /// Force overwriting the wapm.toml, even if it already exists
+    #[clap(long)]
+    pub overwrite: bool,
+    /// Don't display debug output
+    #[clap(long)]
+    pub quiet: bool,
+    /// Ignore the existence of cargo wapm / cargo wasmer
+    #[clap(long)]
+    pub no_cargo_wapm: bool,
```
syrusakbary, line 33: We should remove this until we have `cargo wasmer` and have this as a new PR to review separately

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,605 @@
...
+    /// Namespace to init with, default = current logged in user or _
+    #[clap(long)]
+    pub namespace: Option<String>,
```
syrusakbary, line 36: We shouldn't ask for the namespace in general.
There is one package name: `python/python` is a package name. The frontend should not care about separating the package name from the namespace. The frontend (cli) should only validate that the package name is valid as per regex. That's it

lib/cli/src/commands/init.rs
fschutt: okay, this is removed now

lib/cli/src/commands/init.rs
fschutt: okay, changed it to a `package-name` field

lib/cli/src/commands/init.rs
fschutt: I removed this check, now the package name = the module name

lib/cli/src/commands/publish.rs
fschutt: I removed this check now and use `--package-name` instead

fschutt:
@Michael-F-Bryan okay, this should be done now, I removed support for `cargo wapm`, so those problems with `namespace = ...` are gone now.

lib/cli/Cargo.toml
```diff
@@ -40,6 +40,8 @@ wasmer-types = { version = "=3.1.0", path = "../types" }
 wasmer-registry = { version = "=3.1.0", path = "../registry" }
 wasmer-object = { version = "=3.1.0", path = "../object", optional = true }
 wasmer-vfs  = { version = "=3.1.0", path = "../vfs", default-features = false, features = ["host-fs"] }
+wasmer-wasm-interface = { version = "3.0.1", path = "../wasm-interface" }
```
ptitSeb, line 43: isn't the version supposed to be "=3.1.0"?

fschutt: yeah thanks for catching

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,503 @@
...
+pub struct Init {
+    /// Initialize wapm.toml for a library package
+    #[clap(long, group = "crate-type")]
+    pub lib: bool,
+    /// Initialize wapm.toml for a binary package
+    #[clap(long, group = "crate-type")]
+    pub bin: bool,
+    /// Initialize an empty wapm.toml
+    #[clap(long, group = "crate-type")]
+    pub empty: bool,
```
ptitSeb, line 24: wouldn't it be better to have a 3 state type here, as the init is either for a binary, a lib or empty, but need to be 1 of that and not a combination? Or can it be both for bin & lib?

fschutt: It's not possible with clap: https://github.com/clap-rs/clap/issues/2621

But clap automatically watches that you can't specify both `--bin` and `--lib` for example

ptitSeb: Ah, with "group" thingy I assume. Nice.

rust-toolchain
```diff
@@ -1 +1,5 @@
-1.63
+# rust-toolchain.toml
+[toolchain]
+channel = "1.63"
+components = [ "rustfmt", "clippy" ]
+targets = [ "wasm32-unknown-unknown", "wasm32-wasi" ]
```
ptitSeb, line 5: Is that change necessary? why imposing wasm32 targets or rustfmt/clippy for people that just plan on building wasmer?

fschutt: Yeah it's a leftover from when I tested cargo wapm

ptitSeb (approved):

rust-toolchain
fschutt: ok done

fschutt: turns out it was necessary to run the integration tests. wasm32-wasi is not a very large download, so I'll just use `targets = ["wasm32-wasi"]`

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,679 @@
...
+        let (registry, username) =
+            wasmer_registry::whoami(self.registry.as_deref(), self.token.as_deref()).with_context(
+                || {
+                    anyhow::anyhow!(
+                        "could not find username / registry for registry = {:?}, token = {}",
+                        self.registry,
+                        self.token.as_deref().unwrap_or_default()
+                    )
+                },
+            )?;
```
syrusakbary, line 99: We already chatted about this.
We shouldn't check for authorization on the client when publishing.

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,679 @@
...
+
+        let registry_present =
+            wasmer_registry::test_if_registry_present(&registry).unwrap_or(false);
+
+        if !registry_present {
+            return Err(anyhow::anyhow!(
+                "registry {} is currently unavailable",
+                registry
+            ));
+        }
```
syrusakbary, line 109: This should happen here. Basically, in the same call when getting the signed url is where we can check if the registry is available or not, otherwise we hammer the GraphQL backend

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,679 @@
...
+
+        // If the package name is not a "/", try to prefix it with the current logged in user
+        if !Package::validate_package_name(&manifest.package.name) {
+            manifest.package.name = format!("{username}/{}", manifest.package.name);
+        }
```
syrusakbary, line 114: This is completely incorrect. We shouldn't do any modification of the manifest package name whatsoever

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,679 @@
...
+
+        // Validate the name again
+        if !Package::validate_package_name(&manifest.package.name) {
+            return Err(anyhow::anyhow!(
+                "Invalid package name {:?}",
+                manifest.package.name
+            ));
+        }
```
syrusakbary, line 122: We chatted about this on slack, we mentioned to not do any validation on the package name on the clients whatsoever (just for wasmer run as an optimization mechanism).

Makefile
```diff
@@ -401,7 +401,7 @@ else
 	strip --strip-unneeded target/$(HOST_TARGET)/release/wasmer-headless
 endif
 
-WAPM_VERSION = v0.5.3
+WAPM_VERSION = v0.5.9
```
syrusakbary, line 404: We need to remove this

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,503 @@
...
+                    .unwrap_or_else(|| std::env::current_dir().unwrap())
+                    .join("Cargo.toml");
+                cargo_toml_path
+                    .canonicalize()
+                    .unwrap_or_else(|_| cargo_toml_path.clone())
+            }
+        };
+
+        let cargo_toml = if manifest_path.exists() {
+            Some(parse_cargo_toml(&manifest_path)?)
```
syrusakbary, line 111: We should notify the user (via print) that the cli is generating the wasmer.toml using the settings from cargo

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,503 @@
...
+        // generate the wasmer.toml and exit
+        Self::write_wasmer_toml(&target_file, &constructed_manifest)
+    }
+
+    /// Writes the metadata to a wasmer.toml file
+    fn write_wasmer_toml(path: &PathBuf, toml: &wapm_toml::Manifest) -> Result<(), anyhow::Error> {
+        let toml_string = toml::to_string_pretty(&toml)?
+            .replace(
+                "[dependencies]",
+                &format!("{NOTE}{NEWLINE}{NEWLINE}[dependencies]"),
```
syrusakbary, line 152: Good touch

lib/cli/src/commands/publish.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3432/commits/46a2d28f30de284ddef3bff417c0faa6b7e627dc

lib/cli/src/commands/publish.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3432/commits/46a2d28f30de284ddef3bff417c0faa6b7e627dc

Makefile
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3432/commits/964b897133b2a311cfd1d6aafa7a916e99df428b

lib/cli/src/commands/publish.rs
fschutt: Done in https://github.com/wasmerio/wasmer/commit/46a2d28f30de284ddef3bff417c0faa6b7e627dc

lib/cli/src/commands/publish.rs
fschutt: Done in https://github.com/wasmerio/wasmer/commit/46a2d28f30de284ddef3bff417c0faa6b7e627dc

lib/cli/src/commands/init.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3432/commits/dd470adeffc4fd80cd1b29f99746d276c1088c42

PR #3353, opened 2022-11-22
Speed up CI

fschutt:
Work in progress PR to get the CI runs to be more manageable

Goals:

- migrate from bash / python scripts to cargo xtask to be able to use Rust to script the CI for reproducability
- add more `[jobs]` to be able to run multiple CI steps in parallel
- Reduce CI times to 20 minutes or less, not 1h 30min


syrusakbary (changes requested):
As we chatted, let's keep xtask/makefile completely separated from the PR.
This PR should mainly focus on improving the speed of CI

.github/workflows/test.yaml
```diff
@@ -0,0 +1,1552 @@
...
+            os: macos-11
+            target: x86_64-apple-darwin
+            llvm_url: 'https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang+llvm-13.0.0-x86_64-apple-darwin.tar.xz'
+            run_test: true
+            run_test_capi: true
+            use_sccache: true
+            run_integration_tests: true
+            run_ios_tests: true
+            run_build_docs: false
+          - build: macos-x86_64
```
ptitSeb, line 135: This look like the same step as macos-x64. Maybe one should macos-arm64 instead?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,1552 @@
...
+
+  build_wasmer: 
+    name: Build wasmer on ${{ matrix.build }}
+    runs-on: ${{ matrix.os }}
+    needs: setup
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - build: linux-x64
```
ptitSeb, line 116: linux-arm64 seems to have disapeared?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,1552 @@
...
+            os: macos-11
+            target: x86_64-apple-darwin
+            llvm_url: 'https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang+llvm-13.0.0-x86_64-apple-darwin.tar.xz'
+            run_test: true
+            run_test_capi: true
+            use_sccache: true
+            run_integration_tests: true
+            run_ios_tests: true
+            run_build_docs: false
+          - build: macos-x86_64
```
ptitSeb, line 273: macos-arm64?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,1552 @@
...
+
+  build_capi: 
+    name: Build C-API on ${{ matrix.build }}
+    runs-on: ${{ matrix.os }}
+    needs: setup
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - build: linux-x64
```
ptitSeb, line 254: linux-arm64 is mising?

lib/c-api/tests/wasmer-c-api-test-runner/src/lib.rs
```diff
@@ -23,52 +23,7 @@ impl Config {
...
-                            let path = entry.path();
-                            println!("    {:?}", path.file_name());
-                        }
-                    }
-                    Err(e) => {
-                        println!("error in reading config.wasmer_dir: {e}");
-                    }
-                };
-            }
+            assert!(std::path::Path::new(&config.wasmer_dir).exists());
```
ptitSeb, line 26: So, all this rust code (and `make_package()`) has been moved to the Makefile?

fschutt: Yes

.github/workflows/test.yaml
```diff
@@ -0,0 +1,1552 @@
+name: test-sys
+
+on:
+  push:
+    branches:
+      - '**'
+      - '!master'
```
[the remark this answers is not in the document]
fschutt: not sure, I think it was a copy-paste error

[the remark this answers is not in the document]
fschutt: fixed it

.github/workflows/test.yaml
fschutt: Yeah I think this was an error, I didn't see the duplicate `x86_64` when I experimented with runners.

.github/workflows/test.yaml
fschutt: It hasn't disappeared, we never test it on PRs. We only build it for releases. Would be good to add, however.

.github/workflows/test.yaml
fschutt: fixed

.github/workflows/test.yaml
fschutt: Added it now.

.github/workflows/test.yaml
fschutt: Added it now.

ptitSeb (approved):

.github/workflows/cargo-deny.yaml
```diff
@@ -1,15 +0,0 @@
...
-
-name: cargo-deny
-
-jobs:
-  deny-check:
-    name: cargo-deny
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: EmbarkStudios/cargo-deny-action@v1
```
syrusakbary, line 15: Why deleting this? How we are running this now? We need to run this (is a must)

.github/workflows/documentation.yaml
```diff
@@ -1,37 +0,0 @@
...
-        env:
-          LLVM_DIR: ${{ github.workspace }}/llvm-13
-          LLVM_URL: 'https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz'
-      - name: Build & package documentation
-        run: make package-docs
-      - name: Publish documentation
-        uses: peaceiris/actions-gh-pages@v3
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          publish_dir: ./package/docs
```
syrusakbary, line 37: How we are running this now?

syrusakbary: Note that this is a new workflow because only runs on `master`.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,1385 @@
...
+        env:
+          LLVM_DIR: ${{ github.workspace }}/llvm-13
+          LLVM_URL: 'https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz'
+      - name: Build & package documentation
+        run: make package-docs
+      - name: Publish documentation
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./package/docs
```
syrusakbary, line 133: This should be it's own workflow

.github/workflows/documentation.yaml
fschutt: @syrusakbary we are running it in a separate file, but we are never shipping the documentation

.github/workflows/test.yaml
fschutt: Well okay, I just wanted to make the doc tests pass on master

.github/workflows/documentation.yaml
fschutt: Merged it back into documentation.yaml

.github/workflows/test.yaml
fschutt: done

fschutt:
Results:

- parallelization: done
- sccache: -10 minutes, should be kept
- GitHub cache: obligatory speedup of ~20 minutes, should be kept
- wasmer fails to run on `windows-latest`, but not on `windows-2019`
- 8 cores: 2x speedup, but macos is the blocking part since we don't have 8-core runners, but we get faster failures
- cannot upgrade to macos-12, sadly
- bors: will only speed up merging, not being run in CI, removed for now since we can use GitHub auto-merge

Best Run: 35 minutes with 8 cores, sccache + GitHub cache

theduke (changes requested):
I honestly think the tests here can all just be in one Job per target platform. 

I doubt it would make any notable difference.

Especially on a cache hit, and considering that the integration tests will be running in parallel.

fschutt:
> I doubt it would make any notable difference.

It does make a massive difference, not because of running the tests, but because of compiling them. Let's say I compile `test-stage-1` and `test-stage-2` after each other. This is what [happens](https://github.com/wasmerio/wasmer/actions/runs/3686106983/jobs/6237905223):

```
...

test wast::wasmer::stack_overflow_sret::cranelift::universal ... ok

test result: ok. 902 passed; 0 failed; 247 ignored; 0 measured; 0 filtered out; finished in 134.28s

     Running tests/deprecated.rs (target/x86_64-unknown-linux-gnu/release/deps/deprecated-a290bfa9b2446b26)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

cargo test --target x86_64-unknown-linux-gnu --all --release --exclude wasmer-c-api --exclude wasmer-cli --exclude wasmer-compiler-cli --exclude wasmer-wasi-experimental-io-devices --exclude wasmer-integration-tests-cli --exclude wasmer-integration-tests-ios
   Compiling once_cell v1.16.0
   Compiling stable_deref_trait v1.2.0
   Compiling scopeguard v1.1.0
   Compiling fallible-iterator v0.2.0
   Compiling seahash v4.1.0

... 

   Compiling wasmer-cache v3.1.0 (/home/runner/work/wasmer/wasmer/lib/cache)
    Finished release [optimized] target(s) in 11m 51s
     Running unittests src/lib.rs (target/release/deps/compiler_test_derive-7a19caff8a8fb2a9)

running 2 tests
test ignores::tests::features_match ... ok
test tests::identity_for_no_unsafe ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/release/deps/macro_wasmer_universal_test-fa57d3d920156c3a)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/x86_64-unknown-linux-gnu/release/deps/test_generator-e67a2d19b59aeb61)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/main.rs (target/x86_64-unknown-linux-gnu/release/deps/wasi_test_generator-4bd3d9c46bfc69ee)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src/lib.rs (target/x86_64-unknown-linux-gnu/release/deps/wasmer-2d7ac58d5089f721)

running 25 tests
test sys::externals::function::inner::test_from_to_native_wasm_type::test_to_native ... ok
test sys::externals::function::inner::test_into_result::test_into_result_over_result ... ok

...


   Doc-tests wasmer-wast

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

...
```

On an average run, the time spent on testing is 45 minutes (without caching dependencies):

image

When split by stage, running in parallel, even with no cache, the tests finish in around 20 minutes (the 30+ minute test was blocked on retrieving the cache for some reason):

image

Things are going to be even faster with caching between commits. With the split, I can cache the differences between test stages (running `test --examples` has different dependencies than `test --all`), when merging them into one job, I can't cache the difference (otherwise, compilation of the second test would be instant).

This is the case even with the differences between `--debug` and `--release` tests, with the same feature flags.

Creating more runners is free, since we don't use private ones.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,607 @@
...
+            stage: 0,
+            description: 'Test compilers', 
+          },
+          {
+            stage: 1, 
+            description: 'cargo test --all',
+          },
+          {
+            stage: 2, 
+            description: 'Test compiler-cranelift',
```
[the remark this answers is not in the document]
fschutt: > The only reason I see is to make sure the compilers work fine without the other ones being compiled

That's the reason, yeah. Not sure how to move forward on that, should we continue to do this or not?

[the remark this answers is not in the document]
theduke: Probably best to ask @ptitSeb if it makes sense or not.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,607 @@
...
+      fail-fast: false
+      matrix:
+        stage: [
+          {
+            stage: 0,
+            description: 'Test compilers', 
+          },
+          {
+            stage: 1, 
+            description: 'cargo test --all',
```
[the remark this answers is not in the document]
fschutt: It's done by mistake, it should enable LLVM if it has `ENABLE_LLVM` set on the platform target.

We don't have LLVM on every platform, that's why.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,607 @@
...
+            stage: 4, 
+            description: 'Test lib/cli',
+          },
+          {
+            stage: 5, 
+            description: 'Test examples (debug)',
+          },
+          {
+            stage: 6,
+            description: 'Test examples (release)',
```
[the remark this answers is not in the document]
fschutt: Okay, removed the `--release` version.

.github/workflows/test.yaml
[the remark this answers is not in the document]
fschutt: So I merged everything into one `test-packages` step like it was before and use the same compiler feature flags now.

If there is still recompilation happening between the steps, we have to split it again.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,607 @@
...
+            stage: 2, 
+            description: 'Test compiler-cranelift',
+          },
+          {
+            stage: 3, 
+            description: 'Test compiler-singlepass',
+          },
+          {
+            stage: 4, 
+            description: 'Test lib/cli',
```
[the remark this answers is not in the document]
fschutt: As far as I can see from the Makefile, there are three major stages, with some sub-stages: `test packages` (cargo test), `test examples` (cargo test --examples) and `test compilers` (cannot be merged with the other steps).

One of the test also test without WASI, one with WASI.

The most I can merge is these three steps. Let's see if there's any recompilation happening between them (if they use the same feature sets), and otherwise revert the merging.

I can imagine that the CLI itself pulls in a lot more dependencies than just the ones necessary for testing wasmer, so that's probably why it's in a separate step.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,607 @@
...
+          },
+          {
+            stage: 5, 
+            description: 'Test examples (debug)',
```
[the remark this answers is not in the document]
fschutt: No, it does add significant amount of time, tested it now. I can't merge any stages, the best run was https://github.com/wasmerio/wasmer/pull/3353/commits/72064fcfab274a33d7b8834f3618ed342a0e390c.

.github/workflows/test.yaml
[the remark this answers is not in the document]
fschutt: It does add significant time to the compilation of the tests, I can't merge test stages together. My original solution is still the best one.

.github/workflows/test.yaml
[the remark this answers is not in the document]
fschutt: LLVM is tested on the respective platforms, it's just disabled on the runner you looked at.

.github/workflows/test.yaml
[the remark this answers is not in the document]
ptitSeb: Not sure it's worth the work. I would not pursuit this unless there are some other advantage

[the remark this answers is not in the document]
fschutt: Okay, so I'm going to merge this into one test with `--compilers cranelift,singlepass + llvm?`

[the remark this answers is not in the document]
fschutt: Oh I just saw why we do this: because the dependencies are different when we test `--manifest-path lib/compiler-cranelift` vs `--manifest-path lib/compiler-singlepass`. I'm going to leave this as is because of this.

syrusakbary (changes requested):
Can you review one by one all the tasks and make sure that they do what they are supposed to do?
For example:
* test wasm build doesn't seem to do anything. Correction: it's compiling to wasm32-wasi but it doesn't have a proper name for each of the commands, which confuses when seeing the action log
* [Build and test C-API on linux-aarch64](https://github.com/wasmerio/wasmer/actions/runs/3695985923/jobs/6261132872#logs) only does build (not testing)
* Windows GNU - I don't think this one has the right title (is probably building the c-api?)
* Build and test C-API on different platforms is actually only doing build (no test whatsoever)

Also, some tests seem to be duplicated (are they doing the same thing?):
* https://github.com/wasmerio/wasmer/actions/runs/3695985923/jobs/6261140388
* https://github.com/wasmerio/wasmer/actions/runs/3695985923/jobs/6261140117

Also, we renamed the tests from `test-xyz` to `test-stage-0` where I think actually "stage 0" actually explains nothing about what the stage is.
So we need to have proper descriptive names for each of the stages

syrusakbary (changes requested):
* [Test C-API on linux-x64](https://github.com/wasmerio/wasmer/actions/runs/3705065786/jobs/6278487619#logs) - and relative should be done inside [Build C-API + C-API integration tests on linux-x64](https://github.com/wasmerio/wasmer/actions/runs/3705065786/jobs/6278486625#logs) without triggering any extra recompilation (test and build should use the same compile flags)
* Cache doesn't seem to be working consistently across tasks
   * https://github.com/wasmerio/wasmer/actions/runs/3705065786/jobs/6278492142

fschutt:
Okay this is done, the bad timings are because of bad cache fetches. 
The CI is broken on Windows, but I need to fix that in a separate PR.

ptitSeb:
Please don't merge with a test still red.

theduke (approved):
This could still be improved quite a bit, but we are spending too much time on this.

I'm in favor of merging now, we can always iterate as we go.

*BUT*: CI needs to be green.

If there is a test that is known to fail, let's disable it for now and track fixing that functionality in an issue.

lib/c-api/tests/wasmer-c-api-test-runner/src/lib.rs
```diff
@@ -23,52 +25,7 @@ impl Config {
...
-                            let entry = entry.unwrap();
-                            let path = entry.path();
-                            println!("    {:?}", path.file_name());
-                        }
-                    }
-                    Err(e) => {
-                        println!("error in reading config.wasmer_dir: {e}");
-                    }
-                };
-            }
```
ptitSeb, line 71: Why is this code gone?

ptitSeb (approved):
Ok on my side once tests are green (probably need to disabled that new Windows test and fix it in a different PR)

lib/c-api/tests/wasmer-c-api-test-runner/src/lib.rs
fschutt: It was running `make package` manually as a hack. I removed that in favor of running `make package && make test-integration-cli` instead.

syrusakbary (changes requested):
Test C-API once build C-API is done should be close to instant (or very fast).

Right now is recompiling the c-api somehow? We should not do that

Also, the build c-api on windows should be part of the build matrix

Screenshot 2022-12-19 at 10 36 45 AM

fschutt:
Okay so since the last signoff there were a couple of changes:

- Linking the CRT statically on Windows caused the tests to recompile, even if the feature-flags were the same.
- build-capi-windows-gnu is now part of the build matrix
- tried to cache the windows-x64 build, but failed
- fixed all caches, but R2 is often times locking up and taking 30mins to just retrieve the cache. I'd like to test later on with S3 instead of R2, but for now it's good enough
- upgraded to ubuntu-22.04 to avoid GitHub brownouts of ubuntu-18.04

syrusakbary (commented):
There's something clearly wrong with the cache here:

https://github.com/wasmerio/wasmer/actions/runs/3741028179/jobs/6350105848

fschutt:
@syrusakbary Yes, I know, but I can't do anything about it, I just call the GH action with the proper parameters. If R2 decides to lock up, that's not really something I can debug. I can only try to use S3 to compare the reliability against R2.

fschutt:
I'm currently blocked because I can't create an AWS account because I lost my 2FA access to my bank account. I can get it back, but only after the Dec 25th. Sorry :/

lib/cli/src/commands/create_exe.rs
```diff
@@ -464,7 +464,7 @@ impl CreateExe {
 
         let file = files
         .iter()
-        .find(|f| f.ends_with("libwasmer.a")).cloned()
+        .find(|f| f.ends_with("libwasmer.a") || f.ends_with("wasmer.lib")).cloned()
         .ok_or_else(|| {
             anyhow!("Could not find libwasmer.a for {} target in the provided tarball path (files = {files:#?})", target)
```
syrusakbary, line 469:
```suggestion
            anyhow!("Could not find libwasmer.a / wasmer.lib for {} target in the provided tarball path (files = {files:#?})", target)
```

syrusakbary (commented):
Can we add integration tests on the steps defined here making sure that everything works as we laid out? (both for the wasm case and the webc case)

https://github.com/wasmerio/wasmer/issues/3436#issuecomment-1361688769

theduke (approved):

PR #3355, opened 2022-11-22
Fetch the pirita download URL

Michael-F-Bryan:
Add some more queries to the `wasmer-registry` crate.

I've also had to remove `lib/wasi-types/wit-bindgen` because it's a broken link to use `wai-bindgen` from crates.io, otherwise you get the following error when trying to use `wasmerio/wasmer` as a git dependency (I had to temporarily patch my `wasmer-registry` from crates.io in https://github.com/wasmerio/pirita/pull/62 to see if this PR would work).

```
    Updating git repository `https://github.com/wasmerio/wasmer`
error: failed to load source for dependency `wasmer-registy`

Caused by:
  Unable to update https://github.com/wasmerio/wasmer?branch=more-queries

Caused by:
  failed to update submodule `lib/wasi-types/wit-bindgen`

Caused by:
  no URL configured for submodule 'lib/wasi-types/wit-bindgen'; class=Submodule (17)
```

### DIFF
--- a/lib/registry/src/lib.rs
+++ b/lib/registry/src/lib.rs
@@ -47,6 +47,7 @@ pub struct PackageDownloadInfo {
     pub commands: String,
     pub manifest: String,
     pub url: String,
+    pub pirita_url: Option<String>,
 }
 
 pub fn get_package_local_dir(
@@ -317,6 +318,7 @@ pub fn query_command_from_registry(
     let package = command.package_version.package.display_name;
     let version = command.package_version.version;
     let url = command.package_version.distribution.download_url;
+    let pirita_url = command.package_version.distribution.pirita_download_url;
 
     Ok(PackageDownloadInfo {
         registry: registry_url.to_string(),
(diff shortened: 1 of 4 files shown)


lib/registry/src/lib.rs
```diff
@@ -317,6 +318,7 @@ pub fn query_command_from_registry(
     let package = command.package_version.package.display_name;
     let version = command.package_version.version;
     let url = command.package_version.distribution.download_url;
+    let pirita_url = command.package_version.distribution.pirita_download_url;
 
     Ok(PackageDownloadInfo {
```
syrusakbary, line 323: Why don't we return the type/content from `command.package_version`. That seems way easier than recreating a new struct each time

syrusakbary: I'll approve in any case, we can revisit this after

syrusakbary (approved):

lib/registry/src/lib.rs
Michael-F-Bryan: Yeah, you have a good point.

Something we should do is make all our GraphQL queries public. That way people can execute them directly if a high-level function (e.g. `wasmer_registry::query_command_from_registry()`) doesn't support exactly what you are looking for.

PR #3364, opened 2022-11-23
Added the actual LZCNT / TZCNT implementation

ptitSeb:
# Description
The actual implementation of LZCNT and TZCNT for the x86_64 Singlepass backend.
Somehow, it was missing from #3302


lib/compiler-singlepass/src/emitter_x64.rs
```diff
@@ -946,6 +946,43 @@ impl EmitterX64 for AssemblerX64 {
         Ok(())
     }
 
+    fn arch_has_xzcnt(&self) -> bool {
+        match &self.target {
+            Some(target) => {
+                target.cpu_features().contains(CpuFeature::LZCNT)
+                    && target.cpu_features().contains(CpuFeature::BMI1)
```
theduke, line 953: More curiosity than comment: 

I guess `xzcnt` means TZCOUNT and LZCOUNT.
TZCOUNT is part of BMI1, right?

Is there some good resource to look up CPUID mappings , other than the reference manuals? :D

lib/compiler-singlepass/src/emitter_x64.rs
```diff
@@ -946,6 +946,43 @@ impl EmitterX64 for AssemblerX64 {
...
+        dst: Location,
+    ) -> Result<(), CompileError> {
+        binop_gpr_gpr!(lzcnt, self, sz, src, dst, {
+            binop_mem_gpr!(lzcnt, self, sz, src, dst, {
+                codegen_error!("singlepass cannot emit lzcnt")
+            })
+        });
+        Ok(())
+    }
+    fn arch_emit_tzcnt(
```
theduke, line 972: Nit: missing newline.

theduke (commented):
General question: do we have codegen test suites that validate behaviour both with and without certain flags?

That would be a huge surface area to test though, so probably not?

lib/compiler-singlepass/src/emitter_x64.rs
ptitSeb: > TZCOUNT is part of BMI1, right?

Yes

> Is there some good resource to look up CPUID mappings , other than the reference manuals? :D

Not that I am aware off. I use Intel manuel and various web search when I have question like that.

lib/compiler-singlepass/src/emitter_x64.rs
ptitSeb: Ah, yeah, forgot that one.

PR #3370, opened 2022-11-24
Fix wasmer run not interpreting URLs correctly + display fixes

### LINKED ISSUE #3346: Wasmer not showing right stdout on CLI apps
Wasmer stdout is outputting the incorrect thing when used in a simple CLI app:

```
➜  ~ wasmer dynamite-bud/ayush

Hello, world!package dynamite-bud/ayush ...
```

It should output:

```
➜  ~ wasmer dynamite-bud/ayush
Hello, world!
```

### LINKED ISSUE #3366: wasmer run {url} interprets c:// file paths as URLs on Windows
### Describe the bug

I get the following error:

```
>  wapm run cowsay
error: error fetching c:\Users\username\wapm_packages\syrusakbary/cowsay@0.2.0\target/wasm32-wasi/release/cowsay.wasm: get_webc_bytes failed
```

```
> wasmer -V
wasmer 3.0.1
```

### Steps to reproduce
Run above command on Windows

### Expected behavior
Output from cowsay

### Actual behavior
Error, see above

fschutt:
Fixes #3366.
Fixes #3346.

Should land in the 3.0.2 release.


tests/integration/cli/tests/run.rs
```diff
@@ -546,3 +546,41 @@ fn run_no_start_wasm_report_error() -> anyhow::Result<()> {
...
+            "wapm install cowsay failed with: stdout: {}\n\nstderr: {}",
+            std::str::from_utf8(&output.stdout)
+                .expect("stdout is not utf8! need to handle arbitrary bytes"),
+            std::str::from_utf8(&output.stderr)
+                .expect("stderr is not utf8! need to handle arbitrary bytes")
+        );
+    }
+
+    Ok(())
+}
```
Michael-F-Bryan, line 586: Doesn't this test belong in the `wasmerio/wapm-cli` repo?

We also don't have any guarantee of which version of `wapm` people have installed (or even if it *is* installed), so this test could be a bit flaky, either in CI or for random contributors that run `cargo test`.

lib/cli/src/commands/run.rs
```diff
@@ -848,6 +844,18 @@ pub(crate) fn try_run_package_or_file(
         return r.execute();
     }
 
+    // c:// might be parsed as a URL on Windows
+    let url_string = format!("{}", r.path.display());
+    if url_string.starts_with("http") {
+        if let Ok(url) = url::Url::parse(&url_string) {
```
Michael-F-Bryan, line 850: Would it be better to parse as a URL first, then check if `url.scheme() == "http" || url.scheme() == "https"`? 

I'm normally a bit dubious about "stringly-typed" APIs where you are doing things like `string.contains()` instead of parsing into a more strongly-typed representation like `Url` where you can properly inspect its properties.

For an example of what I mean, running `wasmer run http-server` would always error out saying something like *"unable to send a GET request to http-server"* because `"http-server".starts_with("http")` is `true`.

lib/cli/src/commands/run.rs
```diff
@@ -645,17 +645,20 @@ impl Run {
...
-            .spinner(vec![
-                "⣾", "⣽", "⣻", "⢿", "⡿", "⣟", "⣯", "⣷", " ", "⠁", "⠂", "⠄", "⡀", "⢀", "⠠", "⠐", "⠈",
-            ])
-            .start(),
-    )
+    #[cfg(target_os = "windows")]
+    {
+        use colored::control;
+        let _ = control::enable_virtual_terminal(true);
+    }
```
Michael-F-Bryan, line 656: Should we be calling [`colored::control::set_virtual_terminal()`](https://docs.rs/colored/2.0.0/x86_64-pc-windows-msvc/colored/control/fn.set_virtual_terminal.html) here?

Michael-F-Bryan (changes requested):

tests/integration/cli/tests/run.rs
fschutt: It's just for the CI and no, this should be here so this doesn't happen again (wasmer run breaking wapm run). I could disable this test if GITHUB_TOKEN isn't set.

lib/cli/src/commands/run.rs
fschutt: According to the docs, yes: https://github.com/ad4mx/spinoff#note-for-windows-users

lib/cli/src/commands/run.rs
fschutt: done

tests/integration/cli/tests/run.rs
Michael-F-Bryan: Yep, makes sense :+1:

Michael-F-Bryan (approved):
Looks like the `colored` thing has been updated so this should be good to merge.

.github/workflows/test-sys.yaml
```diff
@@ -206,7 +206,7 @@ jobs:
         if: matrix.run_test && matrix.os != 'windows-2019'
         shell: bash
         run: |
-          make build-wasmer && make build-capi && make package-capi && make package && export WASMER_DIR=`pwd`/package && make test-integration-cli
+          curl https://get.wasmer.io -sSfL | sh && make build-wasmer && make build-capi && make package-capi && make package && export WASMER_DIR=`pwd`/package && make test-integration-cli
```
syrusakbary, line 209: I'd like not to depend on wapm on here. Mainly, because we are moving away from it, and it introduces technical debt

tests/integration/cli/tests/run.rs
```diff
@@ -546,3 +546,53 @@ fn run_no_start_wasm_report_error() -> anyhow::Result<()> {
     assert_eq!(result.contains("Can not find any export functions."), true);
     Ok(())
 }
+
+// Test that changes to wapm run don't break wasmer run
+#[test]
+fn test_wapm_run_works() -> anyhow::Result<()> {
```
syrusakbary, line 552: We should test wasmer run with the complex url (in windows) rather than depend on wapm

syrusakbary (changes requested):
Address comments

.github/workflows/test-sys.yaml
syrusakbary: To be specific: I'd not to like to depend on wapm in any case. We should see what command is wapm doing that make wasmer to fail and start from that case.

fschutt: Done, but we should have a check / test suite that runs before every release (not on every PR / commit) that tests the whole ecosystem: wasmer-js, wapm-cli, wasmer-python, etc. so that this doesn't happen again.

tests/integration/cli/tests/run.rs
fschutt: done, only CI has to pass now

tests/integration/cli/tests/run.rs
```diff
@@ -546,3 +546,46 @@ fn run_no_start_wasm_report_error() -> anyhow::Result<()> {
...
+        .arg("run")
+        .arg(root.join("qjs.wasm"))
+        .arg("--")
+        .arg("-q")
+        .current_dir(&root)
+        .output()?;
+
+    if !output.status.success() {
+        bail!(
+            "wapm install cowsay failed with: stdout: {}\n\nstderr: {}",
```
syrusakbary, line 582: This is not a good message

tests/integration/cli/tests/run.rs
```diff
@@ -546,3 +546,46 @@ fn run_no_start_wasm_report_error() -> anyhow::Result<()> {
     assert_eq!(result.contains("Can not find any export functions."), true);
     Ok(())
 }
+
+// Test that wasmer can run a complex path
+#[test]
+fn test_wasmer_run_complex_url() -> anyhow::Result<()> {
+    let temp_dir = tempfile::tempdir()?;
+    let path = temp_dir.path().to_path_buf();
+    let root = std::env::home_dir().unwrap_or(path);
```
syrusakbary, line 555: We shouldn't rely in the homedir to test. We should rely on files within the repository itself

fschutt: ok done

tests/integration/cli/tests/run.rs
fschutt: fixed

tests/integration/cli/tests/run.rs
```diff
@@ -546,3 +546,58 @@ fn run_no_start_wasm_report_error() -> anyhow::Result<()> {
...
+    }
+
+    println!(
+        "copying from {} to {}",
+        wasi_test_wasm_path(),
+        root.join("qjs.wasm").display()
+    );
+    std::fs::create_dir_all(&root).unwrap();
+    println!("copying...");
+    std::fs::copy(wasi_test_wasm_path(), root.join("qjs.wasm")).unwrap();
```
syrusakbary, line 575: You don't need to copy or create a temp dir, you can just use the qjs.wasm file normally (with display for windows)
: `wasi_test_wasm_path().display()`

.gitignore
```diff
@@ -12,6 +12,7 @@ api-docs-repo/
 /src/windows-installer/WasmerInstaller.exe
 /lib/c-api/wasmer.h
 .xwin-cache
+/temp-wasmer-run-complex-url
```
syrusakbary, line 15: Not needed

fschutt: ok done

tests/integration/cli/tests/run.rs
fschutt: ok done

syrusakbary (approved):

tests/integration/cli/tests/run.rs
```diff
@@ -6,19 +6,23 @@ use std::process::Command;
...
 
 fn wasi_test_wasm_path() -> String {
-    format!("{}/{}", C_ASSET_PATH, "qjs.wasm")
+    let slash = if C_ASSET_PATH.ends_with('/') { "" } else { "/" };
+    format!("{}{slash}{}", C_ASSET_PATH, "qjs.wasm")
 }
 
 fn test_no_imports_wat_path() -> String {
-    format!("{}/{}", ASSET_PATH, "fib.wat")
+    let slash = if ASSET_PATH.ends_with('/') { "" } else { "/" };
```
syrusakbary, line 19: This is not the correct way of joining paths. We should use the std os path join and return a Path

tests/integration/cli/tests/run.rs
```diff
@@ -546,3 +542,79 @@ fn run_no_start_wasm_report_error() -> anyhow::Result<()> {
...
+    println!(
+        "wasmer path {:?} exists: {:?}",
+        get_wasmer_path(),
+        get_wasmer_path().exists()
+    );
+    println!(
+        "qjs file path {:?} exists: {:?}",
+        Path::new(&wasm_test_path),
+        Path::new(&wasm_test_path).exists()
+    );
```
syrusakbary, line 594: We don't need this

syrusakbary (approved):

tests/integration/cli/tests/run.rs
fschutt: done

syrusakbary:
Merging manually as tests should pass

Issue #3372, opened 2022-11-24
Make a Module buildable from a AsEngineRef instead of AsStoreRef


ptitSeb:
### Motivation

To allows creating a module for multiple threads, it's dependancy to the AsStoreRef (that will not be shared accross threads) needs to be remove. An Engine should be enough to build a Module.

### Proposed solution

Create `AsEngineRef` for `Engine`, similar to `AsStoreRef` for the `Store`, and make the Module buildable from this instead of `AsStoreRef`
To avoid API breaking chnage,  a `Store` should also implement `AsEngineRef`.

### Additional context

This should allow better scenario of multi-thread usage, as described there: https://github.com/dsherret/slow-wasm-example/tree/slow_drop

ptitSeb:
Merged

PR #3378, opened 2022-11-25
Introduced EngineRef and AsEngineRef trait

ptitSeb:
# Description
Introduced `EngineRef` structure that is used to build a Module (instead of `StoreRef`) and the `AsEngineRef` trait.
`Store` implement this trait, so there will be no change in writen code even if it's an API change.

This allow to create a Module independantly of a Store, as it can be usefull for multi-thread scenario.

For #3372


lib/types/src/basetunables.rs
```diff
@@ -0,0 +1,58 @@
...
+pub struct BaseTunables {
+    /// For static heaps, the size in wasm pages of the heap protected by bounds checking.
+    pub static_memory_bound: Pages,
+
+    /// The size in bytes of the offset guard for static heaps.
+    pub static_memory_offset_guard_size: u64,
+
+    /// The size in bytes of the offset guard for dynamic heaps.
+    pub dynamic_memory_offset_guard_size: u64,
+}
```
syrusakbary, line 22: Can we move this to the compiler instead (same place where the engine lives?)

syrusakbary (changes requested):
Can we make the Engine also implement `EngineRef`? (that way will be a bit easier to use)

It seems we might also need to move/store tunables one level up to the Engine (instead of the store)

lib/types/src/basetunables.rs
ptitSeb: Oh, it's a left over of a try that I forgot to remove.

ptitSeb:
> Can we make the Engine also implement `EngineRef`? (that way will be a bit easier to use)
> 
> It seems we might also need to move/store tunables one level up to the Engine (instead of the store)

No, because Tunables are not linked to the Engine.

ptitSeb:
For refernce, using https://github.com/dsherret/slow-wasm-example/tree/slow_drop I modified the main.rs to this:
```rust
extern crate wasmer;

use wasmer::{imports, BaseTunables, Cranelift, EngineBuilder, EngineRef, Instance, Module, Store};

// run build.sh in root directory to create this
static WASM: &'static [u8] =
    include_bytes!("../../formatter/target/wasm32-unknown-unknown/release/formatter.wasm");

fn main() {
    println!("Starting...");
    let start = {
        let mut threads = Vec::new();
        let start = std::time::Instant::now();
        let compiler = Cranelift::default();
        let engine = EngineBuilder::new(compiler).engine();
        let tunables = BaseTunables::for_target(&engine.target());
        let engineref = EngineRef::new(&engine, &tunables);
        let module = Module::new(&engineref, &WASM).unwrap();
        for _ in 0..10 {
            let module = module.clone();
                // I haven't figure out how to create a module once and share it amongst the threads in wasmer 3
                let mut store = Store::default();
                let instance = Instance::new(&mut store, &module, &import_object).unwrap();
                let format = instance.exports.get_function("format").unwrap();
                let result = format.call(&mut store, &[]).unwrap();
                (store, instance, result)
            }));
        }
        let mut data = Vec::new();
        for thread in threads {
            data.push(thread.join().unwrap());
        }
        println!("Finished in {}ms...", start.elapsed().as_millis());
        eprintln!("Dropping...");
        std::time::Instant::now()
    };
    println!("Finished dropping in {}ms...", start.elapsed().as_millis());
}
```
I get performances similar to the wasmer 2.3 code

syrusakbary:
> No, because Tunables are not linked to the Engine.

That's what I mean, can we link Tunables to the Engine?

ptitSeb:
> > No, because Tunables are not linked to the Engine.
> 
> That's what I meant, can we link/move Tunables to the Engine?

I don't see how to do that without API Braking changes.

syrusakbary (approved):

PR #3380, opened 2022-11-26
Use cargo read-manifest instead of pkgid to get package version

NickCao:
cargo pkgid does not have a stable output format
see https://doc.rust-lang.org/cargo/commands/cargo-pkgid.html#description

# Description

# Review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -703,7 +703,7 @@ install-capi-headers:
 
 # Currently implemented for linux only. TODO
 install-capi-lib:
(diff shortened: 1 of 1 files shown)


NickCao:
For issue in the wild, see Package Contents of https://archlinux.org/packages/community/x86_64/wasmer/

NickCao:
> For issue in the wild, see Package Contents of https://archlinux.org/packages/community/x86_64/wasmer/

They are now working around the issue by simply deleting the files: https://github.com/archlinux/svntogit-community/blob/e05a5acf008ceca27f4b34f4ca698e4a066e3497/trunk/PKGBUILD#L33

syrusakbary:
So far, this haven't caused any issue and introducing another dependency like `jq` can cause other systems to not work. Closing the PR but thanks for bringing this to our attention

PR #3386, opened 2022-11-29
Restore Support For All Wasi Clock Types

### LINKED ISSUE #3384: `not yet implemented: could not serialize number 2 to enum Clockid` even with `wasi_snapshot_preview1` ABI
See https://github.com/ruby/ruby.wasm/issues/116

theduke:
- fix(wasi): Enables support for all Wasi clock types

# Description

Enables support for the wasi process and thread time clock types (2 and 3).

These were already implemented, but not correctly wired up due to a
duplication of the Clockid type.

This commit switches to using the correct Snapshot0Clockid everywhere.

In the future the "enum Clockid" should be removed, but that was not
done in this commit because currently the ./regenerate.sh script is
broken.

Closes #3384 .


theduke:
Fixes #3386 .

syrusakbary (approved):
This looks good to me. It may be a good idea to get the review from Felix as well on this

fschutt (approved):

Issue #3387, opened 2022-11-29
Fix and Update wasi-types Generation


theduke:
Fix the `./regenerate.sh` script for wasi types.

* Update to the new `wai-bindgen-cli`
* Use a pinned version of wai-binden-cli
* Re-generate bindings and make sure defintions haven't changed
* Remove `clockid` from `typenames.wit` 
  This has been replaced with snapshot0-clockid in #3386 
* Make sure the script deletes the unused `bindings.rs` file 
  
 Priority is high because this really needs to work properly.

PR #3389, opened 2022-11-29
Regenerate bindings using wai-bindgen (requires MSRV 1.64)

Michael-F-Bryan:
Note: This depends on https://github.com/wasmerio/wai/pull/20 and a v0.2.2 release of `wai-bindgen-cli`! It also requires bumping our MSRV to 1.64.

In https://github.com/wasmerio/wasmer/commit/c1b2b08bd19a4d524d056be86e2ad8ef193a60ca, we accidentally updated the `regenerate.sh` script to run `cargo install wai-bindgen` instead of `wai-bindgen-cli` (the crate with the executable). 

That version of the `wai-bindgen` CLI also didn't include https://github.com/wasmerio/wai/commit/fbea20fecaaa280841d35fcc5962f1ead572cfc4 which adds a `--force-generate-structs` flag to the `wai-bindgen rust-wasm ...` sub-command.

This PR updates the `regenerate.sh` script to prompt users to install `wai-bindgen-cli` with the desired version if they haven't already. I've also regenerated the types to make sure `--force-generate-structs` works as intended.

Closes https://github.com/wasmerio/wasmer/issues/3387.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,17 +100,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-trait"
-version = "0.1.58"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e805d94e6b5001b651426cf4cd446b1ab5f319d27bab5c644f61de0a804360c"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn",
-]
-
-[[package]]
 name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


theduke:
We need to pin wai-bindgen to a fixed version . 

Ideally we'd switch to Rust based generation with a dependency instead of the script. (xtask style).

theduke:
Closes #3387 .

Michael-F-Bryan:
Yep, agreed, this feels like the perfect place for a `cargo xtask`.

We do also have the slight problem that the `wai` project uses Rust 1.64 so it can inherit `Cargo.toml` metadata, whereas Wasmer's MSRV is 1.63.

```console
$ cargo check
error: package `wai-bindgen-gen-rust v0.2.1` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.63.0
```

theduke:
This was rewritten in #3422 to generate code with a Rust CLI instead.

PR #3393, opened 2022-11-29
WIP: Debug interaction between create-obj and create-exe

fschutt:


fschutt:
Not necessary anymore after #3430.

PR #3395, opened 2022-11-30
Fix create-exe to be able to cross-compile on Windows

### LINKED ISSUE #3374: Error with create-exe on Windows 11
### Describe the bug

I tried to build executable file via create-exe for my wasm file. It works on Docker/Linux.
But when I tried it on Windows 11, i got below error.

```
> wasmer.exe create-exe -o .\hello.exe .\hello.wasm
Compiler: cranelift
Target: x86_64-pc-windows-msvc
Format: Symbols
error: Failed to find libwasmer
```

Wasmer version is 3.0.1
```
> wasmer --version
wasmer 3.0.1
> rustc --version
rustc 1.65.0 (897e37553 2022-11-02)
```

I can't find `libwasmer` under .wasmer directory. There are only wasmer.dll. Do I need something extra install?

```
PS C:\Users\koduki\.wasmer> dir .\lib\

    Directory: C:\Users\koduki\.wasmer\lib

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---          2022/11/23    16:26       10873344 wasmer.dll
-a---          2022/11/23    16:26          70052 wasmer.dll.lib
-a---          2022/11/23    16:26       65993796 wasmer.lib
```

### Steps to reproduce

1. Install wamer via`iwr https://win.wasmer.io -useb | iex`
2. Install rust via `winget install rustlang.rustup`
3. Install Visual Studio 2022 and exec `vcvars64.bat`
4. exec via ` wasmer.exe create-exe -o .\hello.exe .\hello.wasm`

hello.wasm was build via Rust and it was confirmed `wamer run` Linux(Docker) / Windows, and success create-exe on Linux. 

### Expected behavior
Success build and create executable file

### Actual behavior

```
> wasmer.exe create-exe -o .\hello.exe .\hello.wasm
Compiler: cranelift
Target: x86_64-pc-windows-msvc
Format: Symbols
error: Failed to find libwasmer
```

fschutt:
Fixes #3374.

### DIFF
--- a/lib/cli/src/commands/create_exe.rs
+++ b/lib/cli/src/commands/create_exe.rs
@@ -419,7 +419,7 @@ impl CreateExe {
                         let files = untar(local_tarball.clone(), target_file_path.clone())?;
                         tarball_dir = target_file_path.canonicalize().unwrap_or(target_file_path);
                         files.iter().find(|f| f.ends_with("libwasmer.a")).cloned().ok_or_else(|| {
-                            anyhow!("Could not find libwasmer.a for {} target in the provided tarball path (files = {files:#?})", target)})?
+                            anyhow::bail!("Could not find libwasmer.a for {} target in the provided tarball path (files = {files:#?})", target)});
                     } else {
                         // check if the tarball for the target already exists locally
                         let local_tarball = std::fs::read_dir(get_libwasmer_cache_path()?)?
--- a/lib/cli/src/commands/create_exe.rs
+++ b/lib/cli/src/commands/create_exe.rs
@@ -402,24 +402,9 @@ impl CreateExe {
             let library = if let Some(v) = cross_subc.library_path.clone() {
                 v.canonicalize().unwrap_or(v)
             } else {
-                {
-                    let tarball_dir;
-                    let filename = if let Some(local_tarball) = cross_subc.tarball.as_ref() {
-                        let target_file_path = local_tarball
-                            .parent()
-                            .and_then(|parent| Some(parent.join(local_tarball.file_stem()?)))
-                            .unwrap_or_else(|| local_tarball.clone());
-
-                        let target_file_path = target_file_path
-                            .parent()
-                            .and_then(|parent| Some(parent.join(target_file_path.file_stem()?)))
-                            .unwrap_or_else(|| target_file_path.clone());
-
-                        let _ = std::fs::create_dir_all(&target_file_path);
-                        let files = untar(local_tarball.clone(), target_file_path.clone())?;
-                        tarball_dir = target_file_path.canonicalize().unwrap_or(target_file_path);
-                        files.iter().find(|f| f.ends_with("libwasmer.a")).cloned().ok_or_else(|| {
-                            anyhow::bail!("Could not find libwasmer.a for {} target in the provided tarball path (files = {files:#?})", target)});
+                let (filename, tarball_dir) =
+                    if let Some(local_tarball) = cross_subc.tarball.as_ref() {
+                        Self::find_filename(local_tarball, &target)
                     } else {
                         // check if the tarball for the target already exists locally
                         let local_tarball = std::fs::read_dir(get_libwasmer_cache_path()?)?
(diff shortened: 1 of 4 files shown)


lib/cli/src/commands/create_exe.rs
```diff
@@ -419,32 +418,102 @@ impl CreateExe {
                         let _ = std::fs::create_dir_all(&target_file_path);
                         let files = untar(local_tarball.clone(), target_file_path.clone())?;
                         tarball_dir = target_file_path.canonicalize().unwrap_or(target_file_path);
-                        files.iter().find(|f| f.contains(libwasmer_path)).cloned().ok_or_else(|| {
-                            anyhow!("Could not find libwasmer for {} target in the provided tarball path (files = {files:#?}, libwasmer_path = {libwasmer_path:?})", target)})?
+                        files.iter().find(|f| f.ends_with("libwasmer.a")).cloned().ok_or_else(|| {
```
Michael-F-Bryan, line 421: Can this change depending on OS? (e.g. we get `wasmer.dll` on Windows, or whatever)

lib/cli/src/commands/create_exe.rs
```diff
@@ -419,32 +418,102 @@ impl CreateExe {
...
+                            anyhow!("Could not find libwasmer.a for {} target in the provided tarball path (files = {files:#?})", target)})?
```
Michael-F-Bryan, line 422: You can use `anyhow::bail!()` to return an error.

```suggestion
                            anyhow::bail!("Could not find libwasmer.a for {} target in the provided tarball path (files = {files:#?})", target)});
```

Also, how many files does a tarball normally have? If it's more than a couple, we're going to be spamming the user's console. Would it be worth emitting some log statements intead?

lib/cli/src/commands/create_exe.rs
```diff
@@ -419,32 +418,102 @@ impl CreateExe {
...
+                                    }
+                                }
+
+                                if let OperatingSystem::Linux = target.operating_system {
+                                    if !p.file_name()?.to_str()?.contains("linux") {
+                                        return None;
+                                    }
+                                }
+
+                                Some(p)
```
Michael-F-Bryan, line 468: Can you pull this out into its own helper function?

lib/cli/src/commands/create_exe.rs
```diff
@@ -419,32 +418,102 @@ impl CreateExe {
...
-                                .unwrap_or_else(|_| target_file_path.clone());
-                            let files = untar(tarball, target_file_path)?;
-                            files.into_iter().find(|f| f.contains(libwasmer_path)).ok_or_else(|| {
-                                anyhow!("Could not find libwasmer for {} target in the fetched release from Github: you can download it manually and specify its path with the --cross-compilation-library-path LIBRARY_PATH flag.", target)})?
+                            let _ = std::fs::create_dir_all(&target_file_path);
+                            let files = untar(local_tarball.clone(), target_file_path.clone())?;
+                            tarball_dir =
+                                target_file_path.canonicalize().unwrap_or(target_file_path);
+                            files.iter().find(|f| f.ends_with("libwasmer.a")).cloned().ok_or_else(|| {
+                                anyhow!("Could not find libwasmer.a for {} target in the provided tarball path (files = {files:#?})", target)})?
```
Michael-F-Bryan, line 488: Hang on, didn't we already do this bit?

lib/cli/src/commands/create_exe.rs
```diff
@@ -419,32 +418,102 @@ impl CreateExe {
...
+
+                                tarball_dir = target_file_path
+                                    .canonicalize()
+                                    .unwrap_or_else(|_| target_file_path.clone());
+                                let files = untar(tarball, target_file_path)?;
+                                files.iter().find(|f| f.ends_with("libwasmer.a")).cloned().ok_or_else(|| {
+                                    anyhow!("Could not find libwasmer for {} target in the fetched release from Github: you can download it manually and specify its path with the --cross-compilation-library-path LIBRARY_PATH flag. (files = {files:#?}", target)})?
+                            }
+                            #[cfg(not(feature = "http"))]
+                            return Err(anyhow!("This wasmer binary isn't compiled with an HTTP request library (feature flag `http`). To cross-compile, specify the path of the non-native libwasmer or release tarball with the --library-path LIBRARY_PATH or --tarball TARBALL_PATH flag."));
```
Michael-F-Bryan, line 515: Can we remove the `http` feature altogether and just make it accessible all the time? 

A `wasmer` CLI without internet access seems kinda pointless. Plus `wasmer create-exe` will always depend on resources we've downloaded from the internet, won't it?

Also, I suspect `reqwest` and `tokio` will already sneak in via our other dependencies (e.g. `wasmer-registry`), so the conditional compilation won't be helping us with build times.

lib/cli/src/commands/create_exe.rs
fschutt: No, it's the same for all operating systems.

lib/cli/src/commands/create_exe.rs
fschutt: In general, no, it's just 10 - 15 files, it's not spamming plus it shouldn't happen anyway. In this case the println helped me to find the bug faster, so I'd like to keep it as is.

lib/cli/src/commands/create_exe.rs
fschutt: done

lib/cli/src/commands/create_exe.rs
fschutt: yeah, it's duplicated, I factored it into a function

lib/cli/src/commands/create_exe.rs
fschutt: ok done

lib/cli/src/commands/create_exe.rs
```diff
@@ -460,6 +442,70 @@ impl CreateExe {
...
+            .parent()
+            .and_then(|parent| Some(parent.join(local_tarball.file_stem()?)))
+            .unwrap_or_else(|| local_tarball.to_path_buf());
+
+        let target_file_path = target_file_path
+            .parent()
+            .and_then(|parent| Some(parent.join(target_file_path.file_stem()?)))
+            .unwrap_or_else(|| target_file_path.clone());
+
+        let _ = std::fs::create_dir_all(&target_file_path);
```
theduke, line 459: Any reason why the error here is ignored instead of returned?

lib/cli/src/commands/create_exe.rs
```diff
@@ -572,10 +618,6 @@ impl CreateExe {
                 cmd.arg("-lc");
             }
             cmd.arg("-lunwind");
-            cmd.arg("-OReleaseSafe");
```
theduke, line 575: Is the default optimization level in Zig debug?
That would mean we are always doing debug builds?

theduke (commented):
Not terribly familiar with this, but just two small questions, otherwise LGTM.

lib/cli/src/commands/create_exe.rs
fschutt: No, missed it. Sometimes I ignore errors if I don't want the command to crash in case something weird happens, but in this case we should crash instantly.

lib/cli/src/commands/create_exe.rs
fschutt: No, I should add this back

theduke (approved):
LGTM.

ASSSUMING that we don't want the stripping anymore, otherwise it should be added back in before merging.

lib/cli/src/commands/create_exe.rs
```diff
@@ -573,9 +621,6 @@ impl CreateExe {
             }
             cmd.arg("-lunwind");
             cmd.arg("-OReleaseSafe");
-            cmd.arg("-fstrip");
```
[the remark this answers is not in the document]
fschutt: Yes, they were causing issues with different zig versions. Stripping is not important, it was only saving ~1 MB on a 30MB binary since it doesn't strip the debug info from the libwasmer.a, but the debug info from the wasmer_main.o file (which is very small). So unless zig can do LTO, we're not going to get massive amounts of size reduction here.

PR #3400, opened 2022-12-01
Use the wasm_bindgen_downcast crate for downcasting JsValues

Michael-F-Bryan:
We need this because `@wasmer/wasi` gets minified, which breaks the `JsValue -> WasmerRuntimeError` downcast in the same way that was mentioned in https://github.com/rustwasm/wasm-bindgen/issues/2231#issuecomment-1167895291. 

The [`wasm-bindgen-downcast`](https://github.com/wasmerio/wasm-bindgen-downcast) crate works by associating a unique JavaScript `Symbol` with each Rust type implementing the `DowncastJS` trait. That gives us something we can check before downcasting a `JsValue` back to a `WasmerRuntimeError`.

The underlying issue is that `@wasmer/wasi` implements the `exit()` syscall by raising a `WasmerRuntimeError` with the exit code  as a JavaScript exception and catching it from the outside. To get the exit code back, we try to downcast the `JsValue` we caught back to a `WasmerRuntimeError`, but because the minifier renamed `"WasmerRuntimeError"` to `"H"`, the string-based `generic_of_jsval()` downcast doesn't work. 

Using symbols fixes this issue because a symbol is guaranteed to be unique and unchanging once you create it (`wasm-bindgen-downcast` stores it in a `once-cell`).

See https://github.com/wasmerio/wasmer-js/pull/310 for a much more in-depth explanation of the issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3208,9 +3208,9 @@ checksum = "6bdef32e8150c2a081110b42772ffe7d7c9032b606bc226c8260fd97e0976601"
 
 [[package]]
 name = "syn"
-version = "1.0.103"
+version = "1.0.105"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a864042229133ada95abf3b54fdc62ef5ccabe9515b64717bcb9a1919e59445d"
+checksum = "60b9b43d45702de4c839cb9b51d9f529c5dd26a4aff255b42b1ebc03e88ee908"
 dependencies = [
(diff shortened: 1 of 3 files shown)


syrusakbary (approved):

MTRNord:
Hi is it intentional that https://github.com/wasmerio/wasm-bindgen-downcast is not public? The crates.io page claims it to be MIT and Apache2 licensed however the github repo is not public.

PR #3403, opened 2022-12-02
Add wasm_importtype_copy to C API

### LINKED ISSUE #3322: [c-api] could not load symbol "wasm_importtype_copy":
### Describe the bug

The c-api does not export the function `wasm_importtype_copy` even tough it is defined in the `wasm.h` header file.

```sh
# wasmer 3.0.0-rc.2 | rustc 1.61.0 (fe5b13d68 2022-05-18) | x86_64
```

### Steps to reproduce

1. Run `make build-capi`
2. Test with `nm target/release/libwasmer.so | grep wasm_importtype_copy` which will fail

### Expected behavior

The function should be implemented as per the `wasm.h` header.

### Actual behavior

The function is not implemented.

### Additional context

I suspect that adding the implementation can be fixed by adding this simple line:
```patch
From 49703d0d0db0a100c810660316560f05e165415d Mon Sep 17 00:00:00 2001
From: Paul <9824244+Pangoraw@users.noreply.github.com>
Date: Thu, 17 Nov 2022 23:02:45 +0100
Subject: [PATCH] implement wasm_importtype_copy

---
 lib/c-api/src/wasm_c_api/types/import.rs | 1 +
 1 file changed, 1 insertion(+)

diff --git a/lib/c-api/src/wasm_c_api/types/import.rs b/lib/c-api/src/wasm_c_api/types/import.rs
index dc2e891d7..8be6b99c1 100644
--- a/lib/c-api/src/wasm_c_api/types/import.rs
+++ b/lib/c-api/src/wasm_c_api/types/import.rs
@@ -11,6 +11,7 @@ pub struct wasm_importtype_t {
 }
 
 wasm_declare_boxed_vec!(importtype);
+wasm_impl_copy!(importtype);
 
 #[no_mangle]
 pub extern "C" fn wasm_importtype_new(
-- 
2.38.1
```

fschutt:
Fixes #3322.

### DIFF
--- a/lib/c-api/src/wasm_c_api/types/import.rs
+++ b/lib/c-api/src/wasm_c_api/types/import.rs
@@ -11,6 +11,7 @@ pub struct wasm_importtype_t {
 }
 
 wasm_declare_boxed_vec!(importtype);
+wasm_impl_copy!(importtype);
 
 #[no_mangle]
 pub extern "C" fn wasm_importtype_new(


fschutt:
I tested manually that it works, I don't think we need a unit / integration test here.

Before: 

```sh
$ nm target/release/libwasmer.so | grep wasm_importtype_copy
$
``` 

After:

```sh

$ nm target/release/libwasmer.so | grep wasm_importtype_copy
00000000000d6248 T _wasm_importtype_copy
$
```

PR #3407, opened 2022-12-05
Fix error message when trying to run a nonexistent path

### LINKED ISSUE #3404: wasmer run can't run absolute paths
Wasmer run can't run absolute paths:

```
$ wasmer run /Users/syrusakbary/Development/abc/program.wasm
error: Package abc/program.wasm not found in registries ["users"].

Errors:
  https:///Users/syrusakbary/Development/abc/program/wapm/graphql: error sending query: Error sending GetPackagesQuery: error sending request for url (https://users/syrusakbary/Development/abc/program)
```

fschutt:
Fixes #3404.


lib/cli/src/cli.rs
```diff
@@ -344,14 +344,34 @@ fn test_split_version() {
     );
 }
 
+#[derive(Debug)]
+pub(crate) enum SplitVersionError {
+    InvalidCommandName(anyhow::Error, Option<anyhow::Error>),
+    Other(anyhow::Error),
+}
+
+impl std::error::Error for SplitVersionError {}
```
Michael-F-Bryan, line 353: Can you implement the `source()` method here? That way the underlying error will be added to the error chain when pretty-printing errors.

Michael-F-Bryan (changes requested):
I think it's time to refactor the `SplitVersion` type to be something more closely matched to how we treat `SplitVersion` in practice.

We're actually reusing `SplitVersion` for several different concepts, so it'd be better to split them up into their own bespoke types. From what I can tell, depending on the context, `SplitVersion` might mean one of the following:

- The path to a package folder/file on disk
- A package stored on a registry (where registry URL and namespace are optional)
- A package stored on an arbitrary URL
- A particular command inside a package

Where certain fields are assumed to be `None` (or ignored) depending on how it is being used.

In my opinion, it'd be better to take a more data-oriented approach. Maybe something like this?

```rust
struct PackageSpec {
  location: PackageLocation,
  version: Option<semver::Version>,
}

enum PackageLocation {
  Url(url::Url),
  Registry {
    registry: String,
    namespace: Namespace,
    package_name: String,
  }
}

enum Namespace {
  Global,
  Underscore,
  Organisation(String),
}

struct Command {
  package: PackageSpec,
  command: Option<String>,
}
```

I know it's a pain, but we shouldn't keep adding extra logic and semantics onto  the same struct.

ptitSeb (approved):

syrusakbary (changes requested):
I'd like to have few more tests:
1. Testing that a proper formatted package name actually is searched for in a registry. 
2. Assure that anything else that doesn't look like a package is not searched for in a registry and is treated like a file
3. Assure that non-urls (such as wapm.io without the https://) are actually treated like files

lib/cli/src/cli.rs
```diff
@@ -344,14 +344,34 @@ fn test_split_version() {
     );
 }
 
+#[derive(Debug)]
+pub(crate) enum SplitVersionError {
+    InvalidCommandName(anyhow::Error, Option<anyhow::Error>),
```
theduke, line 349: `anyhow::Error` is designed to be a top level error type.

Nesting it within other error types doesn't really make sense.

You can either implement a concrete enum with variants that hold the original upstream errors, 
or you can have custom error(s) for specific conditiions that are then wrapped in anyhow and can be checked via downcasting.

But embedding `anhow::Error` doesn't make sense.

theduke, line 349: Apart from the other comment, a `(Error, Option<Error>)` tuple is not descriptive at all.
I'd use a struct variant with more descriptive names for the fields.

lib/cli/src/cli.rs
```diff
@@ -364,6 +384,13 @@ impl FromStr for SplitVersion {
 
         let mut no_version = false;
 
+        if s.starts_with('/') || s.contains("..") {
```
theduke, line 387: `./test.wasm` is a valid path.
As is `hello/./test.wasm`.

This doesn't seem to be handled.
But I don't know the wider implications for the rest of the code.

lib/cli/src/commands/run.rs
```diff
@@ -859,9 +859,22 @@ pub(crate) fn try_run_package_or_file(
     let package = format!("{}", r.path.display());
 
     let mut is_fake_sv = false;
-    let mut sv = match SplitVersion::parse(&package) {
```
theduke, line 862: I would approach this differently: 

* first, try to normalize the argument with `Path::canonicalize()` 
* Then, check if it's a valid file (.wasm or .wat), and run it if that's the case
   (there is already similar logic above, but only for cases where there is a `wapm.toml`)

* If that doesn't work, try to parse it as a package identifier

theduke (changes requested):

theduke:
As a general note: we really need a solid parser for package identifiers. 
It's quite tricky to parse, and we also need that in deploy.

Should probably go into the Pirita crate.

@Michael-F-Bryan @fschutt

lib/cli/src/commands/run.rs
syrusakbary: To add on top. If is not "parseable" as a package identifier, it should not fall back into the registry.

Basically, here's the logic:
* Does the file exists? If so, try to run it
* If the file doesn't exist. Does it look like a package name? If so, transform it to a url
* Is the provided path an url? Try to download it
* Is the provided path not an url, show an error mentioning the file path doesn't work

syrusakbary:
> It's quite tricky to parse, and we also need that in deploy.
> Should probably go into the Pirita crate.

I think I disagree with this. Probably on the registry crate?

Michael-F-Bryan:
> It should probably go on the registry crate?

Maybe. 

On one hand, package naming is something inherent to the way we communicate with WAPM and the `wasmer-registry` crate is about communicating with WAPM, but on the other hand I don't want projects to start pulling in `tokio` and `reqwests` and a GraphQL client just because they want a `PackageName` struct... and hiding those dependencies behind a `#[cfg]` is not the answer :wink: 

On the other hand, the `webc` and `pirita` crates are more targeted towards the WEBC format and still have their own non-trivial dependencies.

How about we leave it as a private type in a dedicated `wasmer_cli::package_name` module until we can think of a better long-term home?

lib/cli/src/cli.rs
fschutt: I refactored the error handling, so this does no longer apply.

tests/integration/cli/tests/run.rs
```diff
@@ -156,6 +156,34 @@ fn run_whoami_works() -> anyhow::Result<()> {
     Ok(())
 }
 
+#[test]
+fn run_wasi_works_non_existent() -> anyhow::Result<()> {
+    let output = Command::new(get_wasmer_path())
+        .arg("run")
+        .arg("does/not/exist")
```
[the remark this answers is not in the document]
fschutt: It doesn't parse anymore in the refactored error handling.

fschutt:
@Michael-F-Bryan parsing can currently not be in any other crate than the CLI crate because the parser needs to know which commands are prohibited (for example the command `init` would not be parseable as a valid command).

If that restriction was lifted (could be done by refactoring), I can put it into a separate crate such as a `guid` crate that handles all parsing (and then we can use it from wasmer, pirita and deploy at the same time).

Michael-F-Bryan:
@fschutt, I don't think implementation details from `wasmer some-package` should be bleeding into the parsing code for package names. That would mean the `wasmer` CLI effectively gets baked into the definition of a package name because every time we add a new subcommand we need to add another name to the blacklist.

> If that restriction was lifted (could be done by refactoring), I can put it into a separate crate such as a guid crate that handles all parsing (and then we can use it from wasmer, pirita and deploy at the same time).

This is my preferred option.

Ideally the command-line parsing and package name parsing would be completely separate, where our command-line parsing would do its normal parsing, then if it gets some sort of "unknown subcommand" error it'll re-parse the command-line arguments as a `wasmer_cli::commands::run::Run` command.

lib/cli/src/commands/run.rs
```diff
@@ -92,94 +90,154 @@ fn is_dir(e: &walkdir::DirEntry) -> bool {
...
-                        println!(
-                            "warning: cannot map {alias:?} to {}: directory does not exist",
-                            real_dir.display()
-                        );
-                    }
-                    continue;
-                }
+    /// Registry to run the FILE from (optional)
+    #[clap(short = 'r', long, name = "registry")]
+    pub(crate) registry: Option<String>,
```
theduke, line 102: Why whould we have a separate registry flag if one can provide a URL?

lib/cli/src/commands/run.rs
```diff
@@ -92,94 +90,154 @@ fn is_dir(e: &walkdir::DirEntry) -> bool {
...
-        Ok(Run {
-            path: pathbuf,
+        let argv = std::env::args_os().collect::<Vec<_>>();
+        let (_interpreter, executable, original_executable, args) = match &argv[..] {
+            [a, b, c, d @ ..] => (a, b, c, d),
+            _ => {
+                bail!("Wasmer binfmt interpreter needs at least three arguments (including $0) - must be registered as binfmt interpreter with the CFP flags. (Got arguments: {:?})", argv);
+            }
+        };
+        // TODO: Optimally, args and env would be passed as an UTF-8 Vec.
```
theduke, line 141: why not just use `std::env::args()`, which gives a `String` iterator?

lib/cli/src/split_version.rs
```diff
@@ -0,0 +1,726 @@
...
+                "invalid scheme {}",
+                url.scheme()
+            )));
+        }
+        Ok(Self::Url(url))
+    }
+
+    /// Try parsing `s` as a package (`namespace/user`) or a file (`./namespace/user`)
+    pub fn try_parse_package(s: &str) -> Result<Self, SplitVersionError> {
+        let re1 = regex::Regex::new(r#"(.*)/(.*)@(.*):(.*)"#).unwrap();
```
theduke, line 343: These regexes have lot's of problems.

First, they'll need to handle the whole input, like `"^ ..... $`.
Then, `(.*)/...`  the `.*` is greedy, so it would also include slashes if the later parts match.
You'll want either a non-greedy match `(.*?)` , or be more explicit, by excluding slashes, like `[^/]+`. 
Meaning: everything except a slash, and at least once.

But actually you'll want to be more specific, like `[a-zA-Z0-9\-_]+`.

This is why I said we really need a proper parser for this...

lib/cli/src/split_version.rs
```diff
@@ -0,0 +1,726 @@
...
+            }
+            Err(e) => {
+                errors.push(e);
+            }
+        }
+
+        Err(SplitVersionMultiError {
+            original: s.to_string(),
+            errors,
+        })
```
theduke, line 626: I don't really think it makes sense to collect multiple errors here. 
Should just return something like "InvalidPackageSpecifier" here, that's enough.

theduke (changes requested):

lib/cli/src/commands/run.rs
fschutt: Because we might want to install a package from `wapm.dev` / `wapm.io` instead of going through `wasmer login` / `wapm config set registry.url`

lib/cli/src/split_version.rs
fschutt: The Regex just do the rough parsing. The input is checked for alphanumeric and slashes later on (a string `my/package/version` would pass the Regex but get discarded later on).

lib/cli/src/commands/run.rs
theduke: But that would just be `https://wapm.dev/ns/name` then?

lib/cli/src/split_version.rs
fschutt: No, this is on purpose, so we can print something like:

```sh
wasmer run qjsj.temp
error: Invalid value "qjsj.temp" for '<FILE>': 
    parsing as file: file does not exist: qjsj.temp
    parsing as URL: relative URL without a base
    parsing as package: Invalid package: "qjsj.temp"
    parsing as command: invalid command: qjsj.temp


For more information try --help
```

This way it's much easier to debug if something fails. There are cases where you can't distinguish between a package and a filepath, so if we error on one case, this might cause problems when debugging such cases.

lib/cli/src/commands/run.rs
theduke: Downloading a package doesn't require login.

lib/cli/src/split_version.rs
theduke: I don't think that's very helpful for the user, rather just confusing.

I'd just print a generic error message like "error: 

Invalid run target "qjsj.temp".
You can specify a local path, run packages from the Wasmer registry with ............................

lib/cli/src/commands/run.rs
fschutt: I used `std::env::args()` now, not sure why it used `args_os()` originally.

lib/cli/src/commands/run.rs
fschutt: It is useful if you want to specify the registry for a package, but you don't want to mess with the `wapm.config`. Ex. by default the default registry in prod is `wapm.io`, but the test are always locked to `wapm.dev`, since some packages don't exist on .io.

`wasmer run {url}` only downloads + runs webc files, not tar.gz packages.

I also don't think you can specify versions yet like `/name@version`

For now I'd like to leave it as is, at least until we've migrate the website documentation from `wapm run` to `wasmer run`.

lib/cli/src/split_version.rs
```diff
@@ -0,0 +1,733 @@
...
+            })
+        }
+    );
+
+    // These test fail in the CI thanks to GitHub inserting some \\ \\ on Windows. Disabling for now.
+    // Technically this case is tested by the integration tests
+    #[cfg(not(target_os = "windows"))]
+    {
+        assert_eq!(
+            SplitVersion::parse(concat!(env!("CARGO_MANIFEST_DIR"), "/Cargo.toml")).unwrap(),
```
syrusakbary, line 694: We need to use static paths

lib/cli/src/split_version.rs
```diff
@@ -0,0 +1,733 @@
...
+    /// Parse a version from a String
+    pub fn parse(s: &str) -> Result<SplitVersion, SplitVersionMultiError> {
+        s.parse()
+    }
+}
+
+impl FromStr for SplitVersion {
+    type Err = SplitVersionMultiError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
syrusakbary, line 572: From str should not do any io

lib/cli/src/split_version.rs
```diff
@@ -0,0 +1,733 @@
...
+                original: concat!(env!("CARGO_MANIFEST_DIR"), "/Cargo.toml").to_string(),
+                inner: SplitVersionInner::File {
+                    path: concat!(env!("CARGO_MANIFEST_DIR"), "/Cargo.toml").to_string()
+                },
+            }
+        );
+        assert_eq!(
+            SplitVersion::parse(env!("CARGO_MANIFEST_DIR")).unwrap_err(),
+            SplitVersionMultiError {
+                original: env!("CARGO_MANIFEST_DIR").to_string(),
```
syrusakbary, line 705: We should not use real paths for testing the split

lib/cli/src/split_version.rs
```diff
@@ -0,0 +1,733 @@
...
+        Err(SplitVersionMultiError {
+            original: s.to_string(),
+            errors,
+        })
+    }
+}
+
+#[test]
+fn test_split_version() {
+    assert_eq!(
```
syrusakbary, line 632: We should tests all the possible outputs

lib/cli/src/split_version.rs
```diff
@@ -0,0 +1,733 @@
...
+
+/// Struct containing all combinations of file sources:
+///
+/// - URLs
+/// - namespace/package
+/// - command
+/// - local file
+///
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SplitVersion {
```
syrusakbary, line 21: This is not a good name

lib/cli/src/commands/run.rs
syrusakbary: We need to follow this logic

syrusakbary: Note for us in the future: we can enforce that namespaces exists always (for now), as a tradeoff towards simplification

syrusakbary: If no namespace is given, we assume the namespace as being `_`

lib/cli/src/split_version.rs
```diff
@@ -0,0 +1,733 @@
...
+        }
+        Ok(Self::Url(url))
+    }
+
+    /// Try parsing `s` as a package (`namespace/user`) or a file (`./namespace/user`)
+    pub fn try_parse_package(s: &str) -> Result<Self, SplitVersionError> {
+        let re1 = regex::Regex::new(r#"(.*)/(.*)@(.*):(.*)"#).unwrap();
+        let re2 = regex::Regex::new(r#"(.*)/(.*)@(.*)"#).unwrap();
+        let re3 = regex::Regex::new(r#"(.*)/(.*):(.*)"#).unwrap();
+        let re4 = regex::Regex::new(r#"(.*)/(.*)"#).unwrap();
```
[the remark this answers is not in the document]
fschutt: Implemented now, although I had to drop support for the `:command` syntax

lib/cli/src/split_version.rs
fschutt: Okay, this is now just one regex.

lib/cli/src/split_version.rs
fschutt: Okay, done.

lib/cli/src/split_version.rs
fschutt: Okay, removed.

fschutt:
Superseded by #3416.

Issue #3343, opened 2022-11-21
Singlepass crash/assertion on Apple M1


sagacity:
### Describe the bug

I am seeing heap_get_oob errors using the 'Singlepass' compiler on Apple M1. It does not appear on Cranelift and does not seem to appear on Linux.

This happens on the latest rustc 1.67.0-nightly (a28f3c88e 2022-11-20) on arm64. The repro case seems to work fine on rustc 1.65.0 (stable), but in more complicated scenarios I've also been getting "out of bounds memory access" errors with rustc 1.65.0. I have not yet been able to get these reproduced, though.

### Steps to reproduce

I have a reproduction repo with two steps: https://github.com/sagacity/wasmer-m1-singlepass-crash

### Expected behavior

Deserializing goes fine. The `host` application runs and exits cleanly.

### Actual behavior

A "HeapAccessOutOfBounds" trap is raised during the deserializing of a simple struct.

### Additional context

I'm not sure whether it's really wasmer being at fault. Perhaps the WASM code generated by rustc is faulty in some way. I'd be happy to do some more testing, but I could use some pointers in getting there :)

Additionally, I noticed that if the struct that's being deserialized is below a certain size, there are no crashes. In my reproduction scenario this happens if the `MyHack` struct does not contain the `inner4` field (so just inner/inner2/inner3).

Cross-compiling triggers an assertion inside `emitter_x64.rs` so I have not been able to test if cross-compilation would have solved the issue.

ptitSeb:
Thank you for the detailled ticket. I reproduce the issue locally, I'll have a look to betterunderstand what's going on here.

sagacity:
I've also tried with rustc 1.59, 1.61 and 1.63 and they give the same crash. I will spend some time seeing if I can downgrade some dependencies to see if that helps at all.

sagacity:
Ok! The reason I was a bit suspicious that it would be rustc miscompilation is that we've seen similar deserialization issues when running this kind of code in Chrome (which is not using wasmer afaik). But perhaps those issues are unrelated.

ptitSeb:
After some more testing, the crash also appear in Linux:/ARM64, so it's not something macOS ABI specific, but a bug in the Singlepass/ARM64 backend.

syrusakbary:
This is now fixed in master. Closing the issue

PR #3415, opened 2022-12-08
Fix singlepass for Aarch64

ptitSeb:
# Description
Following debug of ticket #3343 , some issues in the singlepass were found.
In particular, some ADD/SUB opcode involving XSp with large value were not using the correct opcode, and `sub xsp, xsp, x17` were encoded as `neg xzr, x17` (a.k.a. `sub xzr, xzr, x17`). By forcing the dummy `UXTX` modifier, the correct opcode is selected.
Changed both SUB and ADD emiters.
Some alignement function specific to macOS ABI were also adjusted, and Local variable storage/retreival optimized a bit.

### DIFF
--- a/lib/compiler-singlepass/src/emitter_arm64.rs
+++ b/lib/compiler-singlepass/src/emitter_arm64.rs
@@ -3257,24 +3257,16 @@ pub fn gen_std_trampoline_arm64(
                 #[allow(clippy::single_match)]
                 match calling_convention {
                     CallingConvention::AppleAarch64 => {
-                        match sz {
-                            Size::S8 => (),
-                            Size::S16 => {
-                                if caller_stack_offset & 1 != 0 {
-                                    caller_stack_offset = (caller_stack_offset + 1) & !1;
-                                }
-                            }
-                            Size::S32 => {
-                                if caller_stack_offset & 3 != 0 {
-                                    caller_stack_offset = (caller_stack_offset + 3) & !3;
-                                }
-                            }
-                            Size::S64 => {
-                                if caller_stack_offset & 7 != 0 {
-                                    caller_stack_offset = (caller_stack_offset + 7) & !7;
-                                }
-                            }
+                        let sz = 1 << match sz {
+                            Size::S8 => 0,
+                            Size::S16 => 1,
+                            Size::S32 => 2,
+                            Size::S64 => 3,
                         };
+                        // align first
+                        if sz > 1 && caller_stack_offset & (sz - 1) != 0 {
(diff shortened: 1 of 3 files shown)


syrusakbary (approved):
Can we add a test verifying the fix? (triggering what was failing)

Otherwise it looks good

lib/compiler-singlepass/src/emitter_arm64.rs
```diff
@@ -1336,13 +1342,13 @@ impl EmitterARM64 for Assembler {
                 let src1 = src1.into_index() as u32;
                 let src2 = src2.into_index() as u32;
                 let dst = dst.into_index() as u32;
-                dynasm!(self ; add X(dst), X(src1), X(src2));
+                dynasm!(self ; add X(dst), X(src1), X(src2), UXTX);
```
theduke, line 1345: Just checking that I understand correctly:

UXTX forces a specific instruction encoding, and thus prevents us from special-casing the instruction on the value, correct?

theduke (approved):
As Syrus said, a test would be great.
We should try to get into the habit of adding regression tests for each bug we fix.

Adding the Rust reproduction sounds good.

Would be nice to also have a minimal `.wat`  test case, but probably not worth spending extra time on that if it's not trivial to recreate.

In the future we might also want to have some (singlepass) codegen snapshot tests.
As in: test that a specific input compiles to the expected Assembly.

lib/compiler-singlepass/src/emitter_arm64.rs
ptitSeb: Correct. the `sub reg, reg, reg` opcode as 2 form. The "extended register" and the "shifted register'. Only the extended register can be used with XSP, the shifted register will use XZR. The UXTX for the use of the "extended register" form, but in practice does nothing special (UXTX is Zero Extend from 64bits to 64bits).

theduke: :+1: 

Thanks for the explanation, I'm enjoying brushing up on my Assembly knowledge.

ptitSeb:
> As Syrus said, a test would be great. We should try to get into the habit of adding regression tests for each bug we fix.
> 
> Adding the Rust reproduction sounds good.
> 
> Would be nice to also have a minimal `.wat` test case, but probably not worth spending extra time on that if it's not trivial to recreate.
> 
> In the future we might also want to have some (singlepass) codegen snapshot tests. As in: test that a specific input compiles to the expected Assembly.

Yeah, well.
I have already spent 6h trying to get a "working" minimal test. A test for an architecture (Aarch64) that is NOT tested on the CI.
I'll add the test, but I'm not convinced this is good investment of time.

PR #3416, opened 2022-12-09
Download and install packages via .tar.gz URLs and improve installation error message

fschutt:
Fixes https://github.com/wasmerio/wasmer/issues/3404.


tests/integration/cli/tests/run.rs
```diff
@@ -508,6 +508,115 @@ fn run_no_imports_wasm_works() -> anyhow::Result<()> {
...
+    if output.stdout != b"hello\n".to_vec() {
+        panic!("failed to run https://wapm.io/python/python for the second time");
+    }
+
+    // package should be cached
+    assert!(std::time::Instant::now() - time < std::time::Duration::from_secs(1));
+
+    Ok(())
+}
+
```
theduke, line 619: Needs tests for:

* Running local file
* Running with a url

lib/cli/src/commands/run.rs
```diff
@@ -586,6 +557,19 @@ impl Run {
             .collect::<Result<Vec<_>>>()?;
         Ok(func.call(ctx, &invoke_args)?)
     }
+}
+
+impl Run {
+    /// Executes the `wasmer run` command
+    pub fn execute(&self) -> Result<(), anyhow::Error> {
+        // downloads and installs the package if necessary
```
theduke, line 565: I think we said that we always need to prefer local files if they exist, and only then check for remote packages.

Just needs an additional check here if the file exists, and then prefer the local file.

lib/cli/src/package_source.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+impl PackageSource {
+    /// Parses a package source and transforms it to a URL or a File
+    pub fn parse(s: &str) -> Result<Self, String> {
+        // If the file is a http:// URL, run the URL
+        if let Ok(url) = url::Url::parse(s) {
+            if url.scheme() == "http" || url.scheme() == "https" {
+                return Ok(Self::Url(url));
```
theduke, line 39: We probably don't want to accept just any URL, but only urls that conform to the package spec.

Just requires checking the `url.path().strip_prefix('/')` with `Package::from_str()` and report something like an "invalid URL" error.

theduke (changes requested):
Code is much cleaner now!

Just some relatively small modifications.

lib/cli/src/commands/run.rs
fschutt: Ok, done in https://github.com/wasmerio/wasmer/pull/3416/commits/575373c55a0a1f98212a6dd20f3b37cfccf255a4

lib/cli/src/package_source.rs
fschutt: We want to only accept `http://` URLs because the other Urls were conflicting with Windows' `C://` formatting (the original source of the bug that started this whole PR).

fschutt: If you mean, only accepting very narrow `https://wapm.dev/namespace/package` and for example, not `https://wapm.dev/bb0df255edf049905a5d23b03e59a0c240d9723c`, that might cause unexpected issues later on. I would not do this.

tests/integration/cli/tests/run.rs
fschutt: Running with a URL is already in the test?

Running with a local file is already tested in every other integration test (for example the `test_no_start_wat_path` runs the local `no_start.wat` file).

theduke: Ah, right. I overlooked the `.arg("https://wapm.io/python/python")` test.

lib/cli/src/package_source.rs
theduke: That's a valid argument.

It's a better UX if we get a sensible error, because we'll later get something like a `could not parse ` error when you use a URL that doesn't return a webc.

But that's probably fine.

fschutt: @theduke If the URL doesn't return a webc file, it will print the first 100 bytes:

https://github.com/wasmerio/wasmer/pull/3416/files#diff-d83774b7aa5a38e272bd0772c3f95c406608bf13c31c699652c9ce0356c2e2a1R1192-R1198

theduke (approved):

theduke:
I tested "wrong" urls.
Gives  a reasonable error:


```
❯ cargo run -F cranelift,webc_runner -- https://google.com/
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
     Running `/home/theduke/dev/github.com/wasmerio/wasmer/target/debug/wasmer 'https://google.com/'`
error: could not install package from URL https://google.com/
│   1: Could not unpack file downloaded from https://google.com/
╰─▶ 2: failed to unpack /tmp/download.7TKmZ2Pv9hN0/package.tar.gz: xz error: Invalid XZ magic, expected [253, 55, 122, 88, 90, 0]
```

syrusakbary (approved):
Looks great also from my side. Let's merge and release 3.1 once tests pass

PR #3419, opened 2022-12-12
Fix CHANGELOG generation to list by PR merged date, not created date

fschutt:


theduke (approved):
I guess our changelogs were always wrong until now, so the updating of the history is fine, as are the code changes.

But we really need to change our changelog workflow in the future.

fschutt:
@theduke not necessarily wrong, just incomplete also the order of items changed, making the diff a lot bigger than it should have been, but at least it's complete now

PR #3422, opened 2022-12-13
WASIX

### LINKED ISSUE #3308: Add full WASIX support to Wasmer runtime
Merge latest WASIX developpement
Add testing, similar to WASI tests

The objective is to be able to run WASIX compiled program like dash, bash or a WebServer with the Wasmer runtime localy.

### LINKED ISSUE #3387: Fix and Update wasi-types Generation

theduke:
This pull request implements full WASIX support for Wasmer.

It includes many changes, which will be summarized when the PR is ready for review.

Closes #3387 . (subsumes #3389 )
Closes #3308 

Followup issues:

- #3594 
- #3602
- #3603 
- #3604 
- #3605
- #3606
- #3607
- #3608 
- #3609
- #3615 
- #3616
- #3617
- #3618


lib/api/src/js/trap.rs
```diff
@@ -252,6 +257,30 @@ impl std::error::Error for RuntimeError {
...
+            None => {
+                // We simply relay the js value
+                Err(js)
+            }
+        }
+    } else {
+        Err(js)
+    }
+}
+
```
syrusakbary, line 283: I believe this is no longer needed. Can you confirm @Michael-F-Bryan ?

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -328,23 +328,15 @@ pub unsafe extern "C" fn wasi_env_new(
 #[no_mangle]
 pub extern "C" fn wasi_env_delete(_state: Option<Box<wasi_env_t>>) {}
 
-/// Set the memory on a [`wasi_env_t`].
-#[no_mangle]
-pub unsafe extern "C" fn wasi_env_set_memory(env: &mut wasi_env_t, memory: &wasm_memory_t) {
-    let mut store_mut = env.store.store_mut();
-    let wasi_env = env.inner.data_mut(&mut store_mut);
-    wasi_env.set_memory(memory.extern_.memory());
-}
```
syrusakbary, line 337: This is a breaking change. Do we want this instead of adapting the code?

lib/cache/src/filesystem.rs
```diff
@@ -98,7 +98,13 @@ impl Cache for FileSystemCache {
...
         };
         let path = self.path.join(filename);
-        Module::deserialize_from_file(store, path)
+        let ret = Module::deserialize_from_file(store, path.clone());
+        if ret.is_err() {
+            // If an error occurs while deserializing then we can not trust it anymore
+            // so delete the cache file
+            let _ = std::fs::remove_file(path);
+        }
+        ret
```
syrusakbary, line 107: I'm not sure we should handle this here.
Basically, let's say I put a file by mistake (that is not wasm) and try to load it. This will remove it.
We don't want that

lib/vm/src/memory.rs
```diff
@@ -116,6 +116,27 @@ impl WasmMmap {
 
         Ok(prev_pages)
     }
+
+    /// Copies the memory
+    /// (in this case it performs a copy-on-write to save memory)
+    pub fn fork(&mut self) -> Result<WasmMmap, MemoryError> {
```
syrusakbary, line 122: I think this should be a clone instead?

lib/api/src/js/export.rs
```diff
@@ -45,6 +52,52 @@ impl VMMemory {
     pub(crate) fn try_clone(&self) -> Option<VMMemory> {
         Some(self.clone())
     }
+
+    /// Copies this memory to a new memory
+    pub fn fork(&self) -> Result<VMMemory, wasmer_types::MemoryError> {
```
syrusakbary, line 57: Let's call it clone?

lib/wasi/src/syscalls/wasi/fd_read.rs
```diff
@@ -0,0 +1,364 @@
...
+                                }
+
+                                let mut data = Vec::with_capacity(max_size);
+                                unsafe { data.set_len(max_size) };
+                                let amt = handle.read(&mut data[..]).await.map_err(|err| {
+                                    let err = From::<std::io::Error>::from(err);
+                                    match err {
+                                        Errno::Again => {
+                                            if is_stdio {
+                                                Errno::Badf
```
theduke, line 171: @ptitSeb helped fix a test that was failing by adding this check.

I assume this will break consumers, because reading from an empty pipe is not a wrong thing to do and shouldn't return an error. 

I'll investigate if this is problematic for dash/bash etc.

theduke: @john-sharratt

ptitSeb: Check the rust sources: https://github.com/rust-lang/rust/blob/415d8fcc3e17f8c1324a81cf2aa7127b4fcfa32e/library/std/src/io/stdio.rs#L96-L117

You'll see that the ebadf error is the expected and return an Ok (with the `handle_ebadf` function).

lib/api/src/sys/imports.rs
```diff
@@ -121,6 +123,37 @@ impl Imports {
             .insert((ns.to_string(), name.to_string()), val.into());
     }
 
+    /// Imports (any) shared memory into the imports.
+    /// (if the module does not import memory then this function is ignored)
+    #[cfg(feature = "compiler")]
```
ptitSeb, line 128: I find strange to tie this function to the "compiler" feature, as it looks like a Runtime one.

lib/vfs/src/union_fs.rs
```diff
@@ -0,0 +1,1078 @@
+//! Another implementation of the union that uses paths,
+//! its not as simple as TmpFs. not currently used but was used by
+//! the previoulsy implementation of Deploy - now using TmpFs
```
ptitSeb, line 3: Do we need to keep this one if it's not used?

lib/wasi-types/src/wasi/wasix_manual.rs
```diff
@@ -0,0 +1,187 @@
...
+pub union EventUnion {
+    pub clock: u8,
+    pub fd_readwrite: EventFdReadwrite,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[repr(C)]
+pub struct StackSnapshot {
+    pub user: u64,
+    pub hash: u128,
```
theduke, line 107: FIXME: we are currently using C struct "de/serialization" for a few WASIX syscalls. 

This is mostly fine, but not for `u128`, which has different alignment requirements depending on the platform.
This should be replaced with 32 `u8` , or 2 `u64` or whatever.

lib/wasi-types/src/wasi/wasix_manual.rs
```diff
@@ -0,0 +1,187 @@
+use std::mem::MaybeUninit;
+
+use wasmer::ValueType;
+
+use super::{
+    Errno, EventFdReadwrite, Eventtype, Signal, Snapshot0SubscriptionClock, SubscriptionClock,
+    SubscriptionFsReadwrite, Userdata,
+};
+
+pub type WasiHash = u128;
```
theduke, line 10: FIXME: we are currently using C struct "de/serialization" for a few WASIX syscalls. 

This is mostly fine, but not for `u128`, which has different alignment requirements depending on the platform.
This should be replaced with 32 `u8` , or 2 `u64` or whatever.

lib/wasi/src/syscalls/wasi/fd_read.rs
theduke: :+1:

lib/api/src/sys/module.rs
```diff
@@ -267,11 +267,11 @@ impl Module {
...
     /// ```
     pub unsafe fn deserialize(
-        engine: &impl AsEngineRef,
+        store: &impl AsStoreRef,
         bytes: impl IntoBytes,
     ) -> Result<Self, DeserializeError> {
         let bytes = bytes.into_bytes();
-        let artifact = engine.as_engine_ref().engine().deserialize(&bytes)?;
+        let artifact = store.as_store_ref().engine().deserialize(&bytes)?;
         Ok(Self::from_artifact(artifact))
```
ptitSeb, line 275: Why puting store back instead of engine?

lib/c-api/src/wasm_c_api/wasi/mod.rs
ptitSeb: But then, how are you supposed to setup memory with a the C api?

lib/cli/Cargo.toml
```diff
@@ -154,6 +155,7 @@ singlepass = [
 cranelift = [
     "wasmer-compiler-cranelift",
     "compiler",
+    "wasmer-wasi/compiler-cranelift"
```
ptitSeb, line 158: I guess this `wasmer-wasi/compiler-cranelift` should disapear. There is no reasn to have a compiler dependancy on Wasi

theduke: The problem is this: 

There is a "virtual" `wasmer run` command that is available inside of WASIX instances.

For this reason there is a new `ModuleCache` type, which can retrieve WebC images from a registry and compile them into a module so it can be executed as a "WasiProcess". 

This also needs to be available in the CLI so you can do the same thing when (for example) running bash or dash.

I agree that this should be refactored, and I wanted to so so for months.
This can be handled by something like a 

```rust
trait ModuleBuilder {
  fn compile(..) ...
}
```

That way we can drop the compiler features.

I would also argo that `ModuleCache`, `BinPackage`, `BinFactory`, `CmdWasmer` (`lib/wasi/os/command/builtins/cmd_wasmer`) etc don't actually belong in the wasi crate, but in a separate utility crate , which can be used by the CLI and  by Deploy.

Just didn't have any cycles to do that yet.

theduke: Note also the `Console` abstraction, which runs a given WebC package with a TTY.

Same thing applies to that.

lib/api/src/sys/module.rs
theduke: This is aligned with master again.

lib/wasi/src/runners/wasi.rs
```diff
@@ -98,9 +98,6 @@ pub(crate) fn exec_module(
 ) -> Result<(), anyhow::Error> {
     let import_object = wasi_env.import_object(store, module)?;
     let instance = Instance::new(store, module, &import_object)?;
-    let memory = instance.exports.get_memory("memory")?;
-    wasi_env.data_mut(store).set_memory(memory.clone());
-
```
syrusakbary, line 103: We are missing calling `wasi_env.initialize(...)` function here

lib/wasi/src/bin_factory/module_cache.rs
```diff
@@ -0,0 +1,284 @@
...
+use bytes::Bytes;
+use wasmer::{AsEngineRef, Module};
+use wasmer_wasi_types::wasi::Snapshot0Clockid;
+
+use super::BinaryPackage;
+use crate::{syscalls::platform_clock_time_get, VirtualTaskManager, WasiRuntimeImplementation};
+
+pub const DEFAULT_COMPILED_PATH: &'static str = "~/.wasmer/compiled";
+pub const DEFAULT_WEBC_PATH: &'static str = "~/.wasmer/webc";
+pub const DEFAULT_CACHE_TIME: std::time::Duration = std::time::Duration::from_secs(30);
```
syrusakbary, line 12: This code should not live here. For two main reasons:
1. It couples a given path when people may be using `wasmer-wasi` as a library
2. It's tied into other logic (server/deploy) that should probably live there

lib/wasi/src/bin_factory/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+
+pub use self::{
+    binary_package::*,
+    exec::{spawn_exec, spawn_exec_module},
+    module_cache::ModuleCache,
+};
+use crate::{os::command::Commands, WasiRuntimeImplementation, WasiState};
+
+#[derive(Debug, Clone)]
+pub struct BinFactory {
```
syrusakbary, line 24: What is `BinFactory` really?

lib/wasi/src/http/client.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// TODO: use types from http crate?
+pub struct HttpRequest {
+    pub url: String,
+    pub method: String,
+    pub headers: Vec<(String, String)>,
+    pub body: Option<Vec<u8>>,
+    pub options: HttpRequestOptions,
+}
```
syrusakbary, line 55: I agree, we should use types from the http crate. I think it will be way cleaner

lib/wasi/src/os/console/mod.rs
```diff
@@ -0,0 +1,277 @@
...
+    os::task::{control_plane::WasiControlPlane, process::WasiProcess},
+    runtime::{RuntimeStderr, RuntimeStdout},
+    state::Capabilities,
+    TtyFile, VirtualTaskManagerExt, WasiEnv, WasiRuntimeImplementation, WasiState,
+};
+
+//pub const DEFAULT_BOOT_WEBC: &'static str = "sharrattj/bash";
+pub const DEFAULT_BOOT_WEBC: &'static str = "sharrattj/dash";
+//pub const DEFAULT_BOOT_USES: [&'static str; 2] = [ "sharrattj/coreutils", "sharrattj/catsay" ];
+pub const DEFAULT_BOOT_USES: [&'static str; 0] = [];
```
syrusakbary, line 37: We should have no ties into specific packages here.
It points that perhaps we are not doing properly a separation of concerns here

lib/wasi/src/os/command/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+        runtime: Arc<dyn WasiRuntimeImplementation + Send + Sync + 'static>,
+        compiled_modules: Arc<ModuleCache>,
+    ) -> Self {
+        let mut cmd = Self::new();
+        let cmd_wasmer =
+            builtins::cmd_wasmer::CmdWasmer::new(runtime.clone(), compiled_modules.clone());
+        cmd.register_command(cmd_wasmer);
+
+        cmd
+    }
```
syrusakbary, line 55: Can someone explain this TODO?

lib/wasi/src/os/task/control_plane.rs
```diff
@@ -0,0 +1,206 @@
...
+        Arc, RwLock,
+    },
+};
+
+use crate::{WasiProcess, WasiProcessId};
+
+#[derive(Debug, Clone)]
+pub struct WasiControlPlane {
+    state: Arc<State>,
+}
```
syrusakbary, line 14: Can we name this thing somewhat different? In my mind a control plane is related to higher-infra stuff.
Thoughts? What would be a better name?

lib/wasi/src/os/task/process.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// User data associated with thread local data
+    pub thread_local_user_data: HashMap<TlKey, TlUser>,
+    /// Seed used to generate thread local keys
+    pub thread_local_seed: TlKey,
+    /// Signals that will be triggered at specific intervals
+    pub signal_intervals: HashMap<Signal, WasiSignalInterval>,
+    /// Represents all the process spun up as a bus process
+    pub bus_processes: HashMap<WasiProcessId, Box<BusSpawnedProcess>>,
+    /// Indicates if the bus process can be reused
+    pub bus_process_reuse: HashMap<Cow<'static, str>, WasiProcessId>,
```
syrusakbary, line 109: Why are all this fields public, in an inner type? It seems wrong

lib/wasi/src/runtime/host_ws.rs
```diff
@@ -0,0 +1,90 @@
+// TODO: move to more appropriate submodule
```
syrusakbary, line 1: Address TODO

lib/wasi/src/runtime/mod.rs
```diff
@@ -0,0 +1,347 @@
...
+impl From<WasiThreadError> for Errno {
+    fn from(a: WasiThreadError) -> Errno {
+        match a {
+            WasiThreadError::Unsupported => Errno::Notsup,
+            WasiThreadError::MethodNotFound => Errno::Inval,
+            WasiThreadError::MemoryCreateFailed => Errno::Fault,
+            WasiThreadError::InvalidWasmContext => Errno::Noexec,
+        }
+    }
+}
```
syrusakbary, line 59: Can we move this where WasiThreads are defined? (I think it was near the control plane?)

lib/wasi/src/runtime/stdio.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+impl RuntimeStdout {
+    pub fn new(runtime: Arc<dyn crate::WasiRuntimeImplementation + Send + Sync + 'static>) -> Self {
+        Self {
+            runtime,
+            writing: None,
+        }
+    }
+}
```
syrusakbary, line 30: What is this? Don't we have the pipes already defined? Why are we not using them?

lib/wasi/src/runtime/stdio.rs
```diff
@@ -0,0 +1,256 @@
...
+#[derive(Derivative)]
+#[derivative(Debug)]
+pub struct RuntimeStderr {
+    runtime: Arc<dyn crate::WasiRuntimeImplementation + Send + Sync + 'static>,
+    #[derivative(Debug = "ignore")]
+    writing: Option<(
+        Pin<Box<dyn Future<Output = io::Result<()>> + Send + Sync + 'static>>,
+        u64,
+    )>,
+}
```
syrusakbary, line 144: Same

lib/wasi/src/runtime/term.rs
```diff
@@ -0,0 +1,120 @@
...
+#[cfg(unix)]
+pub fn io_result(ret: libc::c_int) -> std::io::Result<()> {
+    match ret {
+        0 => Ok(()),
+        _ => Err(std::io::Error::last_os_error()),
+    }
+}
+
+#[cfg(unix)]
+pub fn set_mode_no_echo() -> std::fs::File {
```
syrusakbary, line 21: This function is only used when `#[cfg(feature = "host-termios")]`. So perhaps should we place all the term into `#[cfg(feature = "host-termios")]` mode and raise a compiler error if not in unix?

lib/wasi/src/runtime/ws.rs
```diff
@@ -0,0 +1,24 @@
+#[cfg(feature = "async_ws")]
+use async_trait::async_trait;
+
+use crate::WasiRuntimeImplementation;
+
+// This ABI implements a general purpose web socket
+#[cfg_attr(feature = "async_ws", async_trait)]
+pub trait WebSocketAbi {
```
syrusakbary, line 8: I have no idea on what this accomplishes

lib/wasi/src/state/builder.rs
```diff
@@ -41,16 +40,20 @@ pub(crate) fn create_wasi_state(program_name: &str) -> WasiStateBuilder {
...
+    map_commands: HashMap<String, PathBuf>,
     vfs_preopens: Vec<String>,
+    compiled_modules: Arc<ModuleCache>,
     #[allow(clippy::type_complexity)]
     setup_fs_fn: Option<Box<dyn Fn(&mut WasiInodes, &mut WasiFs) -> Result<(), String> + Send>>,
     stdout_override: Option<Box<dyn VirtualFile + Send + Sync + 'static>>,
     stderr_override: Option<Box<dyn VirtualFile + Send + Sync + 'static>>,
     stdin_override: Option<Box<dyn VirtualFile + Send + Sync + 'static>>,
-    fs_override: Option<Box<dyn wasmer_vfs::FileSystem>>,
+    fs_override: Option<WasiFsRoot>,
```
syrusakbary, line 56: It shouldn't be a root? Perhaps I'm missing the context on `WasiFsRoot`. Can someone shed some light here?

lib/wasi/src/state/builder.rs
```diff
@@ -152,6 +171,55 @@ impl WasiStateBuilder {
         self
     }
 
+    /// Adds a container this module inherits from
+    pub fn use_webc<Name>(&mut self, webc: Name) -> &mut Self
```
syrusakbary, line 175: Should it be a full webc package instead of a `Name` ?

lib/wasi/src/state/func_env.rs
```diff
@@ -0,0 +1,213 @@
...
+                    let _ = thread_local_destroy.call(
+                        store,
+                        user_data_low as i32,
+                        user_data_high as i32,
+                        val_low as i32,
+                        val_high as i32,
+                    );
+                }
+            }
+        }
```
syrusakbary, line 208: I think this function logic should live on the `WasiEnv` itself, since it's mainly operating with it

lib/wasi/src/state/mod.rs
```diff
@@ -1938,17 +345,43 @@ impl WasiState {
...
-    if file_type.is_dir() {
-        Filetype::Directory
-    } else if file_type.is_file() {
-        Filetype::RegularFile
-    } else if file_type.is_symlink() {
-        Filetype::SymbolicLink
-    } else {
-        Filetype::Unknown
+#[derive(Debug, Clone)]
+pub struct WasiDummyWaker;
```
syrusakbary, line 368: What is this? Can we have code comments on the use?

tests/lib/wast/Cargo.toml
```diff
@@ -19,6 +19,7 @@ wast = "38.0"
 serde = "1"
 tempfile = "3"
 thiserror = "1.0"
+tokio = { version = "1", features = [ "io-util", "rt" ], default_features = false }
```
syrusakbary, line 22: Why is tokio needed for tests? Can we use other task manager

lib/wasi/src/bin_factory/binary_package.rs
```diff
@@ -0,0 +1,135 @@
...
+#[derive(Derivative, Clone)]
+#[derivative(Debug)]
+pub struct BinaryPackage {
+    pub package_name: Cow<'static, str>,
+    pub when_cached: Option<u128>,
+    pub ownership: Option<Arc<dyn Any + Send + Sync + 'static>>,
+    #[derivative(Debug = "ignore")]
+    pub entry: Option<Cow<'static, [u8]>>,
+    pub hash: Arc<Mutex<Option<String>>>,
+    pub wapm: Option<String>,
```
syrusakbary, line 67: What is this and how it differs from `package_name`?

lib/wasi-types/Cargo.toml
```diff
@@ -13,14 +13,19 @@ edition = "2018"
...
-wit-parser = { package = "wasmer-wit-parser", version = "0.1.1" }
+wasmer = { default-features = false, path = "../api", version = "=3.1.0" }
 wasmer-types = { path = "../types", version = "=3.1.0" }
 wasmer-derive = { path = "../derive", version = "=3.1.0" }
-wasmer = { path = "../api", version = "=3.1.0", default-features=false }
+wai-bindgen-gen-rust = "0.2.1"
+wai-bindgen-rust = { version = "0.2.1", default-features = false, features = ["macros"] }
+wai-bindgen-gen-rust-wasm = "0.2.1"
+wai-bindgen-gen-core = "0.2.1"
+wai-parser = "0.2.1"
```
syrusakbary, line 23: Thanks for this

lib/vfs/src/host_fs.rs
```diff
@@ -533,76 +525,156 @@ impl VirtualFile for Stdout {
...
     }
-}
 
-/// A wrapper type around Stderr that implements `VirtualFile` and
-/// `Serialize` + `Deserialize`.
-#[derive(Debug, Default)]
-#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
-pub struct Stderr;
+    fn poll_write_ready(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<io::Result<usize>> {
+        Poll::Ready(Ok(8192))
```
syrusakbary, line 537: Why this number and not other? Should it be a const?

lib/vfs/src/lib.rs
```diff
@@ -1,42 +1,54 @@
...
-//#[cfg(all(feature = "mem-fs", feature = "enable-serde"))]
-//compile_warn!("`mem-fs` does not support `enable-serde` for the moment.");
-
+pub mod arc_file;
+pub mod arc_fs;
+pub mod builder;
+pub mod empty_fs;
 #[cfg(feature = "host-fs")]
 pub mod host_fs;
-#[cfg(feature = "mem-fs")]
```
ptitSeb, line 16: so "mem-fs" is always present? That's a good idea.

lib/wai-bindgen-wasmer/src/le.rs
```diff
@@ -0,0 +1,195 @@
...
+    )*)
+}
+
+primitives! {
+    u8 i8
+    u16 i16
+    u32 i32
+    u64 i64
+    f32 f64
+}
```
ptitSeb, line 164: No u128 here?

lib/wasi-types/src/wasi/wasix_manual.rs
ptitSeb: if you want to be free about any alignment, choose 32 `u8`, `u64` might have some on some platforms (especially 32bits)

lib/wasi/src/runtime/ws.rs
theduke: Allows environments to provide different Websocket client implementations.

I already commented previously that I think this should be integrated with my added `wasix_http_client`, but otherwise not be so deeply integrated into the whole code base.

lib/wasi/src/state/builder.rs
theduke: `WasiFsRoot` is just a simple enum to distinguish between a dynamic backing filesystem and a concrete `TmpFs`. 

I bet this materialized because John needed it to union the file system with others for merging WebC volumes. 

I agree that this should just be stripped out.

We have a few options of how to go about this.

* always use a TmpFs as the root, and "mount" other file systems in the TmpFs as needed
* Always use an `Arc<dyn FileSystem>`   (note  that it needs to be arc now with thread spawning etc)
    In that case we probably want to add an `as_any(&self) -> Option<&dyn Any>` method to the trait to enable downcasting
   
Note: I'm also removing the `.uses()` method from the WasiEnv builder and moving all the webc related setup code to a single place (BinFactory), so we can initially just use `dyn`.
But we might want to dynamically union in new filesystems in the future for Deploy, just to put this on the radar.

lib/wasi/src/state/builder.rs
theduke: Yes, in fact it should probably use the same parsing logic Felix implemented for the CLI, or just always require a full WebC specifier (this is why I wanted a parser in `webc`!)

But as above, I'm also moving this to a central location for webc handling.

lib/wasi/src/state/mod.rs
theduke: Will add some comments.

In short: if you want to manually poll a future outside of an async runtime, you have to provide a "fake" no-op waker, because `poll()` requires a `Context` argument , which can only be constructed with a waker.
( I implemented the same thing at least twice before)

lib/wasi/src/bin_factory/binary_package.rs
theduke: I'm working on more or less completely rewriting this anyway.

lib/wasi-types/Cargo.toml
theduke: Gladly.

lib/vfs/src/lib.rs
theduke: Yep, doesn't require any extra dependencies, so it makes sense to have it always available.

lib/wai-bindgen-wasmer/src/le.rs
theduke: This was just moved over from the `wai` repo.

`u128` is not part of the component model ABI, so no point to have it here.

lib/wasi/src/os/task/process.rs
theduke: I think I already added a comment above that these should be private. 

( I very recently refactored the `WasiThread` in the same vain - reduce redundant separate Arcs, make fields private)

lib/wasi/src/runtime/host_ws.rs
theduke: This file wasn't actually included in the module tree!

I deleted it.

I preserved the code in `sys-utils/src/tungstenite_websocket`, but with a FIXME.
Should probably just be removed and replaced with WAI based websocket bindings in the future.

lib/wasi/src/runtime/mod.rs
theduke: Moved to `os::task::thread` submodule in a155340b1caf50935f2741e614b35e68f5ad4b02 .

lib/wasi/src/runtime/stdio.rs
theduke: This is due to a very awkward split where currently some STDIO functionality is placed on the `WasiRuntimeImplementation` trait.

It absolutely does not belong there, and should be refactored.

Haven't done so until now because it affects code across a few implementations (wasmer-web, Deploy, etc)

lib/wasi/src/runtime/term.rs
theduke: As commented above:

A)  this stuff does not belong on the Runtime trait at all
B) Even if it did, the config flags make no sense. This is a trait, so the code should just live in environment  specific implementations (like a Unix impl)

I bet this is the result of refactoring a struct into a trait...

lib/wasi/src/state/mod.rs
theduke: Moved to utils submodule and added a doc comment in 9aa6a07e7ba139be580b5739b6321217484a9f55 .

lib/wai-bindgen-wasmer/src/le.rs
theduke: Going to resolve this, feel free to comment if unclear.

lib/cli/src/commands/wasmer_static_create_exe_main.c
```diff
@@ -0,0 +1,218 @@
+#include "wasmer.h"
```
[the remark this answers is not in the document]
theduke: This seems to be a merge/rebase artifact.

Was added by Felix previously, so probably just a left-over.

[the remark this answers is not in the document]
theduke: The culprit is probably this change by @john-sharratt :

```
 #ifdef WASI
-  // Read the exports.
-  wasm_extern_vec_t exports;
-  wasm_instance_exports(instance, &exports);
-  wasm_memory_t* mem = NULL;
-  for (size_t i = 0; i < exports.size; i++) {
-    mem = wasm_extern_as_memory(exports.data[i]);
-    if (mem) {
-      break;
-    }
-  }
-
-  if (!mem) {
-    fprintf(stderr, "Failed to create instance: Could not find memory in exports\n");
+  if (!wasi_env_initialize_instance(wasi_env, store, instance)) {
+    fprintf(stderr, "Failed to initialize env\n");
     print_wasmer_error();
     return -1;
   }
-  wasi_env_set_memory(wasi_env, mem);

   own wasm_func_t *start_function = wasi_get_start_function(instance);
   if (!start_function) {
```

lib/cli/src/commands/run/wasi.rs
```diff
@@ -64,7 +84,7 @@ impl Wasi {
     pub fn get_versions(module: &Module) -> Option<BTreeSet<WasiVersion>> {
         // Get the wasi version in strict mode, so no other imports are
         // allowed.
-        get_wasi_versions(module, true)
+        get_wasi_versions(module, false)
```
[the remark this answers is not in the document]
theduke: Adjusted the commend.

lib/cli/src/commands/wasmer_static_create_exe_main.c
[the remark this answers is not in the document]
theduke: @ptitSeb do you know where this code was moved to so I can re-apply the change?

lib/wasi/src/os/task/control_plane.rs
theduke: Happy to hear suggestions.

This doesn't currently hold any actual logic, just the data, so `ProcessTree` ? 

But we should probably move some of the process and thread spawning logic in here, so `ProcessManager`?

tests/lib/wast/Cargo.toml
theduke: Because the crate contains wasi tests, and our wasi implementation now requires tokio.

lib/wasi/src/lib.rs
```diff
@@ -28,420 +30,195 @@ compile_error!(
...
-        #[cfg(feature = "wasix")]
-        if is_wasix_module(module) {
-            self.data_mut(store)
-                .state
-                .fs
-                .is_wasix
-                .store(true, std::sync::atomic::Ordering::Release);
-        }
+/// The default stack size for WASIX
+pub const DEFAULT_STACK_SIZE: u64 = 1_048_576u64;
```
ptitSeb, line 157: So 1Mb is the minimum memory requirement for WASIX?

lib/wasi/src/os/command/builtins/cmd_wasmer.rs
```diff
@@ -0,0 +1,154 @@
...
+
+use wasmer::{FunctionEnvMut, Store};
+use wasmer_vbus::{BusSpawnedProcess, SpawnOptions, VirtualBusError};
+use wasmer_wasi_types::wasi::Errno;
+
+use crate::{
+    bin_factory::{spawn_exec, BinaryPackage, ModuleCache},
+    syscalls::stderr_write,
+    VirtualTaskManager, VirtualTaskManagerExt, WasiEnv, WasiRuntimeImplementation,
+};
```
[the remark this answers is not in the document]
ptitSeb: but for that to work, we would need hidden syscall probably (like to fetch webc packages)

lib/wasi/src/os/console/txt/about.md
```diff
@@ -0,0 +1,9 @@
+# Wasmer Terminal
+
+This terminal is an Wasmer powered terminal hosted in a browser which implements
+a basic operating system and is natively integrated with ATE and WAPM.
+
+For more information try:
+
+about wasmer
+about deploy
```
ptitSeb, line 9: We have "deploy" mentionned here, is that ok?

lib/wasi/src/os/console/txt/about_deploy.md
```diff
@@ -0,0 +1,14 @@
+# wasmer.sh
```
ptitSeb, line 1: I don't understand why this file is here.

lib/cli/src/commands/wasmer_static_create_exe_main.c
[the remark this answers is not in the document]
ptitSeb: I don't think this file is relevent anymore. All the code should be in `lib/cli/src/commands/wasmer_create_exe_main.c` already.

lib/wasi/src/runtime/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+#[cfg(feature = "sys")]
+pub type ArcTunables = std::sync::Arc<dyn wasmer::Tunables + Send + Sync>;
+
+/// Represents an implementation of the WASI runtime - by default everything is
+/// unimplemented.
+#[allow(unused_variables)]
+pub trait WasiRuntimeImplementation
+where
+    Self: fmt::Debug + Sync,
+{
```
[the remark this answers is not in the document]
syrusakbary: The write stdout/ stderr and read stdin should rely on the Pipes and not in this type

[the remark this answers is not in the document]
theduke: As mentioned, currently the various `RuntimeImplementation`s carry additional state , especially in Deploy and wasmer-web.

That would only be viable to remove if we have cutoms envs for these, but that would be a larger refactoring task.

lib/vfs/src/host_fs.rs
john-sharratt: it should be const indeed

as for why this exact number itself - it should be aligned with a page size (4K) (for better caching performance) and be of reasonable size. network stacks have shown that the size of modern packets is too small (1500bytes) for a decent throughput thus quite often a switch to jumbo frames (9000 bytes packets) can have quite a significant improvement on bandwidth - a similar IO overhead will occur on piped data too however without testing its basically an educated guess.

given most code i've seen (like `tokio` ) just ignores the poll result numbers and hits the file descriptor with non blocking writes until it returns `E_AGAIN` the use of the number is hard measure - still its better than putting 1 as there could be nasty edge cases if one actually does put some logic on that number (for instance, wait for at least 32bits before writing a length prefix on the pipe).

either way another educated guess is also welcome for this number and it indeed should be a const.

lib/cli/src/commands/wasmer_static_create_exe_main.c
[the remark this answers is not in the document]
john-sharratt: yeah this was added while the env intitialize code was still being refactored in the mainline - its just a left over artifact of working in parallel.

lib/vbus/src/lib.rs
```diff
@@ -190,43 +274,157 @@ pub trait VirtualBusScope: fmt::Debug + Send + Sync + 'static {
 
 pub trait VirtualBusInvokable: fmt::Debug + Send + Sync + 'static {
     /// Invokes a service within this instance
+    #[allow(unused_variables)]
     fn invoke(
         &self,
-        topic: String,
+        topic_hash: u128,
```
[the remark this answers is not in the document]
john-sharratt: 16 byte hashes are fairly common these days but a strongly typed struct may be better was just trying to keep the dependencies to a minimum - but either way we can probably ditch `vbus` if we add a way to do RPC calls to bindings - then this entire trait can go into the code blackhole

lib/wasi/src/lib.rs
john-sharratt: there are a number of references to a default WASM stack size of 1MB however its not per say a standard I would say.
https://github.com/rustwasm/wasm-pack/issues/479

`wasix` could run with less stack, even very small stacks however it would not work so well with libc and call recursion so I would recommend keeping it at least 1MB to reduce the chances of hitting stack overflows on "normal" call stacks in libc. another option is to try and detect the stack size that the WASM program is compiled against - this might be possible by looking at the memory layout of the WASM file  - it might also be an overkill for now

lib/wasi/src/os/console/txt/about.md
john-sharratt: leftovers - it can be removed

lib/wasi/src/runtime/mod.rs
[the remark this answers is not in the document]
john-sharratt: There are quite so differences between how one implements the runtime on the browser vs on x86 native - hence an abstraction is needed - this is that abstraction - sometimes abstractions add more complexity than they are worth, in this case I believe the abstraction to be well justified. custom envs for different hosting environment would be way more complex unless an abstraction is added to custom envs but then that would be the same as this anyway

lib/wasi/src/runners/wasi.rs
theduke: Added back in the `exec_module` function.

lib/cli/src/commands/wasmer_static_create_exe_main.c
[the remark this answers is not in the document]
theduke: Removed.

lib/wasi/src/os/console/txt/about.md
theduke: Removed.

lib/wasi/src/os/console/txt/about_deploy.md
theduke: Removed.

lib/wasi/src/runtime/mod.rs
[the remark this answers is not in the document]
theduke: Yes I concur. 

There are *some* things that need to change about the trait, for example: all the TTY and STDIO things **really** don't belong on here.

But for the other aspects I concur that it makes sense.

lib/vfs/src/host_fs.rs
theduke: Factored it out to a const and added a doc comment.

tests/lib/wast/Cargo.toml
theduke: @syrusakbary did that clear it up?
We basically have tokio as a hard dependency for WASI now.

They only way to not have that dependency would be to split out WASI tests into a separate crate.

lib/wasi/src/os/command/builtins/cmd_wasmer.rs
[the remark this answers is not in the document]
theduke: I don't think we want to push the Wasmerception too far.

It would be nice to have compilation etc available inside Wasmer, so for this having a published Wasmer package could make sense.

But for downloading the webc we  want the environment (aka Deploy or the browser)  to do it for caching. 
And running it must be done by the environment anyway. 

So we could have an "embedded Wasmer" build that does everything except support "run", but I believe that's quite a bit off, right?

For now I will move the custom command to a `wasmer-utils` crate, so it can be used from there for Deploy and `wasmer-web` and the CLI.

lib/vm/src/memory.rs
theduke: Outdated, we changed this do duplicate().

theduke:
One WASIX tests was failing because the host vfs readdir() sorted by modification time.

I'm curios why it did that in the first place, but sorting by modification time is problematic in a test, and not very deterministic.
(I assume the sorting is there in the first place to give some determinism)

I changed the readdir implementation to sort by path, which seems more sensible to me.

theduke:
Another tests is failing because the C API WASI implementation doesn't use the tokio executor - I will remedy that.

lib/cli/Cargo.toml
theduke: Compiler specific features have been removed.

lib/wasi/src/runtime/stdio.rs
theduke: All of this stuff was removed.

lib/wasi/src/runtime/stdio.rs
theduke: Was removed.

lib/wasi/src/runtime/term.rs
theduke: Refactored the code so these functions live in a submodule gated by the dependency.

lib/wasi/src/runtime/ws.rs
theduke: Websocket support was removed.

lib/api/src/js/trap.rs
theduke: Removed

lib/api/src/sys/imports.rs
theduke: This is the same on master.

lib/vfs/src/mem_fs/file_opener.rs
```diff
@@ -1,20 +1,322 @@
...
+                fs_lock.add_child_to_node(inode_of_parent, inode_of_file)?;
+
+                inode_of_file
+            }
+        };
+        Ok(())
+    }
+
+    /// Inserts a arc file into the file system that references another file
+    /// in another file system (does not copy the real data)
```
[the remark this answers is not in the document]
theduke: Gone now.

lib/vfs/src/special_file.rs
```diff
@@ -0,0 +1,98 @@
...
+use std::task::{Context, Poll};
+
+use crate::VirtualFile;
+use tokio::io::{AsyncRead, AsyncSeek, AsyncWrite};
+use wasmer_wasi_types::wasi::Fd;
+
+/// A "special" file is a file that is locked
+/// to one file descriptor (i.e. stdout => 0, stdin => 1), etc.
+#[derive(Debug)]
+pub struct SpecialFile {
```
syrusakbary, line 15: I believe this have a specific naming in unix? (although forgot which is the correct one!)

examples/wasi.rs
```diff
@@ -37,28 +37,11 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     let module = Module::new(&store, wasm_bytes)?;
 
-    println!("Creating `WasiEnv`...");
-    // First, we create the `WasiEnv`
-    let wasi_env = WasiState::new("hello")
+    // Run the module.
+    let mut wasi_env = WasiEnv::builder("hello")
         // .args(&["world"])
         // .env("KEY", "Value")
-        .finalize(&mut store)?;
```
syrusakbary, line 45: I think is still a good idea to separate the build of the state, and the initialize funciton.

I'm happy to have only two, but I don't think we should have one that does the two

lib/api/src/sys/native.rs
```diff
@@ -48,10 +45,6 @@ impl<Args: WasmTypeList, Rets: WasmTypeList> Clone for TypedFunction<Args, Rets>
     }
 }
 
-thread_local! {
-    static ON_CALLED: Cell<Option<OnCalledHandler>> = Cell::new(None);
```
syrusakbary, line 52: Thanks for this, I also deleted it on the unification branch

lib/api/src/sys/mem_access.rs
```diff
@@ -313,6 +313,14 @@ impl<'a, T: ValueType> WasmSlice<'a, T> {
         self.buffer.write(self.offset, bytes)
     }
 
+    /// Reads this `WasmSlice` into a `slice`.
+    #[inline]
+    pub fn copy_to_slice(self, buf: &mut [MaybeUninit<u8>]) -> Result<usize, MemoryAccessError> {
```
syrusakbary, line 318: This should be consistent with the naming in js: `read_to_slice`:

```
    pub fn read_to_slice<'b>(
        self,
        buf: &'b mut [MaybeUninit<u8>],
    ) -> Result<usize, MemoryAccessError> {
```

lib/sys-utils/tests/fd_mmap_memory.rs
```diff
@@ -0,0 +1,124 @@
+// // TODO: make this a sensible test before merge NOTE: commented out since it's mostly here for demonstration purposes.
+// use std::sync::Arc;
```
syrusakbary, line 2: Do we want to remove this file from the PR since is all commented out?

lib/vfs/Cargo.toml
```diff
@@ -7,23 +7,40 @@ license = "MIT"
...
+static-fs = ["webc", "anyhow"]
 enable-serde = [
     "serde",
     "typetag"
 ]
 no-time = []
+std = [ "wasmer-types/std" ]
+core = [ "wasmer-types/core" ]
+sys = [ "wasmer-wasi-types/sys" ]
+js = [ "wasmer-wasi-types/js" ]
```
syrusakbary, line 46: We should not include nor leak wasmer types into vfs. VFS should be agnostic of wasmer logic & types

lib/vfs/src/pipe.rs
```diff
@@ -0,0 +1,434 @@
...
+            Poll::Ready(Ok(0))
+        } else {
+            Poll::Ready(Ok(8192))
+        }
+    }
+}
+
+/// Pipe pair of (a, b) Pipes that are connected together
+#[derive(Clone, Debug)]
+pub struct BidiPipe {
```
syrusakbary, line 299: Is this not normally called a stream duplex?

lib/vfs/src/union_fs.rs
syrusakbary: I think itmight be useful to have specially for running on the frontend

lib/wai-bindgen-wasmer/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "wai-bindgen-wasmer"
```
syrusakbary, line 2: Shouldn't this live in `wai`? (I'm ok for this living here, but want to clarify just in case!)

lib/wasi-types/src/lib.rs
```diff
@@ -1,46 +1,69 @@
...
-        .join("\r\n");
-    let generated_str = generated_str.replace("mod output {", "pub mod output {");
-    let bindings_target = bindings_target
-        .lines()
-        .map(|s| s.to_string())
-        .collect::<Vec<String>>()
-        .join("\r\n");
-    pretty_assertions::assert_eq!(generated_str, bindings_target); // output.wit out of date? regenerate bindings.rs
-}
+// TODO: re-enable once WAI bindings generator generates correct code.
```
syrusakbary, line 9: Reenable?

examples/wasi.rs
ptitSeb: What could be the use of a seprate WasiState for the end user?

lib/api/src/js/exports.rs
```diff
@@ -53,6 +53,9 @@ pub enum ExportError {
     /// This error arises when an export is missing
     #[error("Missing export {0}")]
     Missing(String),
+    /// This error arises when an export is missing
```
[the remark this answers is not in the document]
ptitSeb: Yep, the ExportError::SerializationFailed was not there on the preparation PR and isnt' used here. Removing it...

lib/wai-bindgen-wasmer/Cargo.toml
theduke: We talked about this quite a bit with @Michael-F-Bryan .

The problem is that we now actually use the WAI bindings for WASI.
That means we need to be able to make changes to the bindgen crate and point it towards the latest version of Wasmer.
Thats not possible if the crate is not in the Wasmer repo.

examples/wasi.rs
theduke: Note that `WasiState` is currently a private implementation detail. 

To properly initialize a state we need the module and the FunctionEnv, because there are functional dependencies between them. 

The API is intended to prevent misuse, because it's rather easy to use a built state incorrectly.

lib/vfs/src/special_file.rs
theduke: It's often called special file, sometimes device file.

theduke: But the system here just pins to one specific `Fd`, so it's a bit more general than "device file".

I think the name is fine.

lib/vfs/src/builder.rs
```diff
@@ -0,0 +1,181 @@
...
+
+pub struct RootFileSystemBuilder {
+    default_root_dirs: bool,
+    default_dev_files: bool,
+    add_wasmer_command: bool,
+    stdin: Option<Box<dyn VirtualFile + Send + Sync>>,
+    stdout: Option<Box<dyn VirtualFile + Send + Sync>>,
+    stderr: Option<Box<dyn VirtualFile + Send + Sync>>,
+    tty: Option<Box<dyn VirtualFile + Send + Sync>>,
+}
```
ptitSeb, line 18: Why is there a tty & the 3 std stream?

lib/api/src/sys/exports.rs
```diff
@@ -57,6 +57,9 @@ pub enum ExportError {
     /// This error arises when an export is missing
     #[error("Missing export {0}")]
     Missing(String),
+    /// This error arises when an export is missing
+    #[error("Serialization failed {0}")]
+    SerializationFailed(String),
```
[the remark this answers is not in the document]
theduke: Rebase artifact, I removed it.

lib/wasi/src/bin_factory/mod.rs
ptitSeb: I was about to ask the same question. Is that related to the spawn syscall, to handle new binary being spawned or something like that?

lib/wasi/src/fs/inode_guard.rs
```diff
@@ -0,0 +1,753 @@
+use std::{
```
ptitSeb, line 1: What's the use of the inode_guard? This seems quite complex, with listeners and events.

lib/wasi/src/syscalls/wasix/callback_thread.rs
```diff
@@ -0,0 +1,35 @@
+use super::*;
+use crate::syscalls::*;
+
+// FIXME: remove , since it's no longer used.
+/// ### `callback_spawn()`
```
ptitSeb, line 5: So this need to be removed? Shall I do it?

lib/wasi/src/syscalls/wasix/callback_signal.rs
```diff
@@ -0,0 +1,54 @@
+use super::*;
+use crate::syscalls::*;
+
+// FIXME: remove , since it's no longer used.
+/// ### `callback_signal()`
```
ptitSeb, line 5: So this need to be removed? Shall I do it?

lib/wasi/src/syscalls/wasix/callback_reactor.rs
```diff
@@ -0,0 +1,35 @@
+use super::*;
+use crate::syscalls::*;
+
+// FIXME: remove , since it's no longer used.
```
ptitSeb, line 4: So this need to be removed? Shall I do it?

lib/wasi/src/syscalls/wasix/call_reply.rs
```diff
@@ -0,0 +1,23 @@
+use super::*;
+use crate::syscalls::*;
+
+// FIXME: remove , since it's no longer used.
```
ptitSeb, line 4: So this need to be removed? Shall I do it?

lib/wasi/src/syscalls/wasix/call_fault.rs
```diff
@@ -0,0 +1,13 @@
+use super::*;
+use crate::syscalls::*;
+
+// FIXME: remove , since it's no longer used.
```
ptitSeb, line 4: So this need to be removed? Shall I do it?

lib/wasi/src/syscalls/wasix/call_close.rs
```diff
@@ -0,0 +1,10 @@
+use super::*;
+use crate::syscalls::*;
+
+// FIXME: remove , since it's no longer used.
```
ptitSeb, line 4: So this need to be removed? Shall I do it?

lib/wasi/src/syscalls/wasix/bus_subcall.rs
```diff
@@ -0,0 +1,26 @@
+use super::*;
+use crate::syscalls::*;
+
+// FIXME: remove , since it's no longer used.
```
ptitSeb, line 4: So this need to be removed? Shall I do it?

lib/wasi/src/syscalls/wasix/callback_thread_local_destroy.rs
```diff
@@ -0,0 +1,35 @@
+use super::*;
+use crate::syscalls::*;
+
+// FIXME: remove , since it's no longer used.
```
ptitSeb, line 4: So this need to be removed? Shall I do it?

lib/wasi/src/syscalls/wasix/bus_call.rs
```diff
@@ -0,0 +1,26 @@
+use super::*;
+use crate::syscalls::*;
+
+// FIXME: remove , since it's no longer used.
```
ptitSeb, line 4: So this need to be removed? Shall I do it?

lib/wasi/src/syscalls/wasix/bus_close.rs
```diff
@@ -0,0 +1,12 @@
+use super::*;
+use crate::syscalls::*;
+
+// FIXME: remove , since it's no longer used.
```
ptitSeb, line 4: So this need to be removed? Shall I do it?

lib/wasi-types/src/wasi/wasix_manual.rs
theduke: WasiHash was removed, because VBUS was removed completely.

lib/wasi/src/http/client.rs
theduke: We should to this later, not pre-merge.

lib/sys-utils/tests/fd_mmap_memory.rs
theduke: Id rather keep it in so we are less likely to forget about fixing it up when we have time.

lib/vfs/src/union_fs.rs
theduke: Let's keep it for now.

theduke: Let's keep it for now.

theduke: I'm cheekily going to resolve, reopen if you disagree.

lib/cache/src/filesystem.rs
theduke: It makes sense that the cache path is owned , we add and delete things in there anyway.

So I'm resolving.

lib/vfs/Cargo.toml
theduke: It's agnostic but it's mostly using the various error types. 

We can have separate definitions and map between them, but that's honestly something I'd rather do post merge.

lib/wai-bindgen-wasmer/Cargo.toml
theduke: Going to resolve since it should be here, but feel free to re-open.

lib/wasi/src/syscalls/wasix/bus_close.rs
theduke: Already did so before seeing your comments.

lib/wasi/src/syscalls/wasix/call_fault.rs
theduke: Removed

lib/wasi/src/syscalls/wasix/call_close.rs
theduke: Removed

lib/wasi/src/syscalls/wasix/bus_subcall.rs
theduke: Removed

lib/wasi-types/src/lib.rs
theduke: No, because WAI currently produces code that requires some cleanup.
We already have lots of other fixes, so we should probably add that as a last change and also create an issue to track this in WAI.

lib/wai-bindgen-wasmer/Cargo.toml
syrusakbary: SGTM

lib/vfs/Cargo.toml
syrusakbary: I think this is something that can leak outside if we don't solve on this PR (specially because the features can cause things way more complex than they should)
So I'd push for solving this on this PR

lib/vbus/src/lib.rs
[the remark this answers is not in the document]
theduke: VBUS was removed.

lib/wasi/src/syscalls/wasix/callback_thread_local_destroy.rs
john-sharratt: Yes, please go ahead

lib/wasi/src/syscalls/wasix/callback_reactor.rs
john-sharratt: Yes, please go ahead

lib/wasi/src/syscalls/wasix/callback_signal.rs
john-sharratt: Yes, please go ahead

lib/wasi/src/syscalls/wasix/callback_thread.rs
john-sharratt: Yes, please go ahead

lib/wasi/src/syscalls/wasix/callback_reactor.rs
john-sharratt: No this is needed for the multithreading code, we can only remove it after we move over to the new threading spec which can not be done till.its stabilized

lib/wasi/src/syscalls/wasix/callback_signal.rs
john-sharratt: No this is needed for the multithreading code, we can only remove it after we move over to the new threading spec which can not be done till.its stabilized

lib/wasi/src/syscalls/wasix/callback_thread.rs
john-sharratt: No this is needed for the multithreading code, we can only remove it after we move over to the new threading spec which can not be done till.its stabilized

john-sharratt: This sets the name of the exported function that should be invoked when a thread is spawned, it is called by libc and rust std library

lib/wasi/src/syscalls/wasix/callback_signal.rs
john-sharratt: This sets the name of the exported function that should be invoked when a signal is sent to a process, it is called by libc

lib/wasi/src/syscalls/wasix/callback_reactor.rs
john-sharratt: This sets the name of the exported function that should be invoked when a thread is spawned, it is called by libc and rust std library. Although it is not used anymore it is a dependency with the rust std library for wasix until we refactor

lib/wasi/src/fs/inode_guard.rs
john-sharratt: Its used by the poll syscall which needs to hold multiple locks on multiple files in a safe and performant way. Yes its complex but thats because `poll_oneoff` is complex which has been made considerably simplier and more stable then it was. Ultimately these guards and the pill create an async bridge back to the threads in the runtime using an wasm version of mio and tokio. Not fun to implement but definately worth it as we get some impressive async performance now in wasix

john-sharratt: Performance by the way which is hard to beat in other runtimes. @syrusakbary we should definately benchmark this and put it on the website

lib/vfs/src/builder.rs
john-sharratt: Because tty files in /dev/tty act different to /dev/stdout. Stdout inherits stdio from the parent while tty bypasses the process tree and connects stdio directly to the virtual operating system

lib/wasi-types/src/lib.rs
john-sharratt: The wai bindings have quite a bit more work before they substitute for the full syscalls of wasi and wasix. Delaying the merge because of this would be quite a delay therefore I agree with @theduke

lib/vfs/src/pipe.rs
john-sharratt: I think Duplex does indeed sound better

lib/wasi/src/syscalls/wasix/callback_thread_local_destroy.rs
john-sharratt: Just like the other callbacks, we need this so libc and the rust std library can tell the runtime what exported function to call when a thread is terminated and needs to cleanup local thread storage

theduke: I have removed the FIXME comments.

lib/wasi/src/syscalls/wasix/callback_signal.rs
theduke: Removed the FIXME.

lib/wasi/src/syscalls/wasix/callback_reactor.rs
theduke: Removed the FIXME

lib/wasi/src/syscalls/wasix/callback_thread.rs
theduke: Removed the FIXME

lib/vfs/src/pipe.rs
theduke: Yeah I agree, I renamed it to DuplexPipe. 

I stuck with pipe though because I think it fits better with our terminology, and it actually uses two pipes internally, so...

lib/vfs/Cargo.toml
theduke: Yeah, I don't disagree... 

I'll tacke it.

lib/wasi/src/state/func_env.rs
theduke: The problem is that this cleanup needs to call back into the instance.
The rest of the code just collects the necessary arguments for the call. 

Doesn't make much sense to me to move it out.

lib/vfs/Cargo.toml
theduke: VFS was cleaned up!

lib/vfs/src/special_file.rs
syrusakbary: Ok, after thinking more about this, I think we should name this `Device` and then stdin/stderr/stdout are just special kind of devices?

examples/wasi.rs
syrusakbary: Still, it doesn't allow to set custom imports that may be provided by the user.
We need composability, as is also needed in wasmer-js and in the wai generated bindings for WASI modules.

In general, we should have something like:

```rust
    let import_object = wasi_env.import_object(&mut store, &module)?;
    let instance = Instance::new(&mut store, &module, &import_object)?;

   wasi_env.initialize(instance);
```

examples/wasi.rs
```diff
@@ -15,9 +15,11 @@
 //!
 //! Ready?
```
syrusakbary, line 16: Duplicate this test and do it the old way

theduke: I added an example with the manual setup flow.

lib/wasi-types/src/wasi/wasix_manual.rs
theduke: Now tracked in https://github.com/wasmerio/wasmer/issues/3602

lib/wasi/src/bin_factory/module_cache.rs
theduke: Tracked in https://github.com/wasmerio/wasmer/issues/3603

lib/wasi/src/bin_factory/mod.rs
theduke: In particular it handles bundling `webc` packages, including merging dependencies.

theduke: I have a partial rewrite of this to make it cleaner, but that should be merged after the main PR.

lib/wasi/src/fs/inode_guard.rs
theduke: Treating this as resolved , since it's just a question - @ptitSeb

lib/vfs/src/builder.rs
theduke: Assumed to be answered - it's required. (@ptitSeb )

lib/wasi-types/src/lib.rs
theduke: Tracked in https://github.com/wasmerio/wasmer/issues/3604

lib/wasi/src/os/console/mod.rs
theduke: Removed the default, now has to be supplied as an argument.

lib/wasi/src/os/command/builtins/cmd_wasmer.rs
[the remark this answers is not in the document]
theduke: Tracked in #3606

lib/wasi/src/os/command/mod.rs
theduke: This is due to a rather cumbersome situation where the `CmdWasmer` requires some additional context.

This would/will be fixed as part of the work in #3606

lib/wasi/src/os/task/control_plane.rs
theduke: Didn't get any suggestions - name is fine for now.

lib/wasi/src/os/task/process.rs
theduke: Refactor tracked in #3607

lib/wasi/src/state/builder.rs
theduke: Asked and answered

lib/wasi/src/state/builder.rs
theduke: Part of #3603

lib/wasi/src/state/func_env.rs
theduke: Really needs to be on the FunctionEnv.

tests/lib/wast/Cargo.toml
theduke: Assumed as answered - it's required.

lib/wasi/src/bin_factory/binary_package.rs
theduke: Part of #3605

lib/wasi/src/lib.rs
theduke: Assumed to be answered.

lib/wasi/src/runtime/mod.rs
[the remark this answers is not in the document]
theduke: Will stay as is - the runtime trait makes sense at the moment.

lib/vfs/src/special_file.rs
theduke: As mentioned, I don't think `DeviceFile` makes sense here, since it doesn't need to be a device at all.

examples/wasi.rs
theduke: Still possible - a new example is at `examples/wasi_manual_setup.rs`.

lib/sys-utils/tests/fd_mmap_memory.rs
theduke: #3422

tests/ignores.txt
```diff
@@ -123,9 +124,187 @@ wasitests::nightly_2022_10_18::mem_fs::readlink
 wasitests::nightly_2022_10_18::mem_fs::unix_open_special_files
 wasitests::nightly_2022_10_18::mem_fs::writing
 
+wasitests::snapshot1::tmp_fs::close_preopen_fd
```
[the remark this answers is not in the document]
theduke: #3609

syrusakbary (approved):

PR #3426, opened 2022-12-15
WASIX Preparation

theduke:
This commit extracts changes to core libraries made in the WASIX branch.

It is not reasonable to extract the partial commit history, so this is
just a batch commit.

The history will make a bit more sense again if we decide to merge the
WASIX branch with full commit history.


theduke:
@fschutt 

The Linux runtime test fails with:

```
test test_wasmer_create_exe_pirita_works ... compiling to target x86_64-unknown-linux-gnu
packaging /package to .tar.gz: /tmp/.tmpxSTqhk/link.tar.gz
packaging done
tmp tar gz path: /tmp/.tmpxSTqhk/link.tar.gz - exists: true
running: "/home/runner/work/wasmer/wasmer/tests/integration/cli/../../../target/x86_64-unknown-linux-gnu/release/wasmer" "create-exe" "/tmp/.tmp2IYt78/python.wasmer" "--tarball" "/tmp/.tmpxSTqhk/link.tar.gz" "--target" "x86_64-unknown-linux-gnu" "-o" "/tmp/.tmp2IYt78/python"
compilation ok!
Error: python_exe_output_path /tmp/.tmp2IYt78/python does not exist
```

That's probably unrelated to this PR, right?

lib/api/src/js/export.rs
```diff
@@ -45,6 +52,52 @@ impl VMMemory {
     pub(crate) fn try_clone(&self) -> Option<VMMemory> {
         Some(self.clone())
     }
+
+    /// Copies this memory to a new memory
+    pub fn fork(&self) -> Result<VMMemory, wasmer_types::MemoryError> {
```
syrusakbary, line 57: Let's use clone. I think is a more accurate naming

lib/api/src/js/externals/memory.rs
```diff
@@ -221,6 +263,12 @@ impl Memory {
     pub fn is_from_store(&self, store: &impl AsStoreRef) -> bool {
         self.handle.store_id() == store.as_store_ref().objects().id()
     }
+
+    /// Copies this memory to a new memory
+    pub fn fork(&mut self, store: &impl AsStoreRef) -> Result<VMMemory, MemoryError> {
```
syrusakbary, line 268: Let's use clone. I think is a more accurate naming

lib/api/src/sys/tunables.rs
```diff
@@ -120,6 +120,16 @@ mod tests {
         fn try_clone(&self) -> Option<Box<dyn LinearMemory + 'static>> {
             None
         }
+        fn fork(&mut self) -> Result<Box<dyn LinearMemory + 'static>, MemoryError> {
```
syrusakbary, line 123: Let's use clone. I think is a more accurate naming

lib/vm/src/memory.rs
```diff
@@ -116,6 +116,27 @@ impl WasmMmap {
 
         Ok(prev_pages)
     }
+
+    /// Copies the memory
+    /// (in this case it performs a copy-on-write to save memory)
+    pub fn fork(&mut self) -> Result<Self, MemoryError> {
```
syrusakbary, line 122: Let's use clone. I think is a more accurate naming

lib/vm/src/memory.rs
```diff
@@ -270,16 +279,22 @@ impl VMOwnedMemory {
...
     /// Converts this owned memory into shared memory
     pub fn to_shared(self) -> VMSharedMemory {
         VMSharedMemory {
             mmap: Arc::new(RwLock::new(self.mmap)),
             config: self.config,
         }
     }
+
+    /// Copies this memory to a new memory
+    pub fn fork(&mut self) -> Result<Self, MemoryError> {
```
syrusakbary, line 292: Let's use clone. I think is a more accurate naming

lib/vm/src/memory.rs
```diff
@@ -316,8 +331,26 @@ impl LinearMemory for VMOwnedMemory {
     fn try_clone(&self) -> Option<Box<dyn LinearMemory + 'static>> {
         None
     }
+
+    /// Copies this memory to a new memory
+    fn fork(&mut self) -> Result<Box<dyn LinearMemory + 'static>, MemoryError> {
```
syrusakbary, line 336: Let's use clone. I think is a more accurate naming

lib/vm/src/memory.rs
```diff
@@ -341,6 +374,15 @@ impl VMSharedMemory {
     ) -> Result<Self, MemoryError> {
         Ok(VMOwnedMemory::from_definition(memory, style, vm_memory_location)?.to_shared())
     }
+
+    /// Copies this memory to a new memory
+    pub fn fork(&mut self) -> Result<Self, MemoryError> {
```
syrusakbary, line 379: Let's use clone. I think is a more accurate naming

lib/vm/src/memory.rs
```diff
@@ -379,9 +421,15 @@ impl LinearMemory for VMSharedMemory {
...
 
-    /// Owned memory can not be cloned (this will always return None)
+    /// Shared memory can always be cloned
     fn try_clone(&self) -> Option<Box<dyn LinearMemory + 'static>> {
-        None
+        Some(Box::new(self.clone()))
+    }
+
+    /// Copies this memory to a new memory
+    fn fork(&mut self) -> Result<Box<dyn LinearMemory + 'static>, MemoryError> {
```
syrusakbary, line 430: Let's use clone. I think is a more accurate naming

lib/vm/src/memory.rs
```diff
@@ -445,6 +493,11 @@ impl LinearMemory for VMMemory {
     unsafe fn initialize_with_data(&self, start: usize, data: &[u8]) -> Result<(), Trap> {
         self.0.initialize_with_data(start, data)
     }
+
+    /// Copies this memory to a new memory
+    fn fork(&mut self) -> Result<Box<dyn LinearMemory + 'static>, MemoryError> {
```
syrusakbary, line 498: Let's use clone. I think is a more accurate naming

lib/vm/src/memory.rs
```diff
@@ -497,12 +550,17 @@ impl VMMemory {
...
+    pub fn from_custom<IntoVMMemory>(memory: IntoVMMemory) -> Self
     where
-        IntoVMMemory: Into<VMMemory>,
+        IntoVMMemory: Into<Self>,
     {
         memory.into()
     }
+
+    /// Copies this memory to a new memory
+    pub fn fork(&mut self) -> Result<Box<dyn LinearMemory + 'static>, MemoryError> {
```
syrusakbary, line 561: Let's use clone. I think is a more accurate naming

lib/vm/src/memory.rs
```diff
@@ -555,4 +613,7 @@ where
 
         initialize_memory_with_data(memory, start, data)
     }
+
+    /// Copies this memory to a new memory
+    fn fork(&mut self) -> Result<Box<dyn LinearMemory + 'static>, MemoryError>;
```
syrusakbary, line 618: Let's use clone. I think is a more accurate naming

lib/api/src/js/module.rs
```diff
@@ -207,16 +180,34 @@ impl Module {
     /// This is safe since the JS vm should be safe already.
     /// We maintain the `unsafe` to preserve the same API as Wasmer
     pub unsafe fn from_binary_unchecked(
-        _store: &impl AsStoreRef,
-        binary: &[u8],
+        store: &impl AsStoreRef,
+        binary: impl IntoBytes,
```
syrusakbary, line 184: We chatted about this before with @ptitSeb and @john-sharratt . Binary should be `&[u8]`

lib/api/src/js/module.rs
```diff
@@ -194,7 +163,11 @@ impl Module {
     /// Opposed to [`Module::new`], this function is not compatible with
     /// the WebAssembly text format (if the "wat" feature is enabled for
     /// this crate).
-    pub fn from_binary(_store: &impl AsStoreRef, binary: &[u8]) -> Result<Self, CompileError> {
+    pub fn from_binary(
+        _store: &impl AsStoreRef,
+        binary: impl IntoBytes,
```
syrusakbary, line 168: We chatted about this before with @ptitSeb and @john-sharratt . Binary should be `&[u8]`

lib/api/src/js/module.rs
```diff
@@ -274,10 +268,71 @@ impl Module {
...
+            // the error for us, so we don't need to handle it
+        }
+        Ok((
+            crate::StoreHandle::new(
+                store.as_store_mut().objects_mut(),
+                WebAssembly::Instance::new(&self.module, &imports_object)
+                    .map_err(|e: JsValue| -> RuntimeError { e.into() })?,
+            ),
+            import_externs,
+        ))
```
syrusakbary, line 335: I'm not sure this changes are really needed. Why do we need this?

lib/api/src/js/module.rs
```diff
@@ -558,14 +636,33 @@ impl Module {
...
     }
 
+    /// Returns true if the module is still ok - this will be
+    /// false if the module was passed between threads in a
+    /// way that it became undefined (JS does not share objects
+    /// between threads except via a post_message())
+    pub fn is_ok(&self) -> bool {
+        let val = JsValue::from(&self.module);
+        !val.is_undefined() && !val.is_null()
+    }
```
syrusakbary, line 646: Rather than constructing the Module, shouldn't we just do this check whenever we cross the boundaries and before the Rust `Module` is constructed?

lib/api/src/js/module.rs
```diff
@@ -558,14 +636,33 @@ impl Module {
...
+    // /// Get the custom sections of the module given a `name`.
+    // ///
+    // /// # Important
+    // ///
+    // /// Following the WebAssembly spec, one name can have multiple
+    // /// custom sections. That's why an iterator (rather than one element)
+    // /// is returned.
+    // pub fn custom_sections<'a>(&'a self, name: &'a str) -> impl Iterator<Item = Arc<[u8]>> + 'a {
+    //     unimplemented!();
+    // }
```
syrusakbary, line 657: Delete me

lib/api/src/js/native.rs
```diff
@@ -64,11 +66,34 @@ macro_rules! impl_native_traits {
...
+                        if let Some(callback) = store_mut.inner.on_called.take() {
+                            match callback(store_mut) {
+                                Ok(wasmer_types::OnCalledAction::InvokeAgain) => { continue; }
+                                Ok(wasmer_types::OnCalledAction::Finish) => { break; }
+                                Ok(wasmer_types::OnCalledAction::Trap(trap)) => { return Err(RuntimeError::user(trap)) },
+                                Err(trap) => { return Err(RuntimeError::user(trap)) },
+                            }
+                        }
+                        break;
+                    }
```
syrusakbary, line 94: I need extra context on why do we need to do this. Is not obvious and I'm not sure we really want this?

@john-sharratt @theduke

lib/api/src/js/module.rs
theduke: Done.

lib/api/src/js/module.rs
```diff
@@ -254,17 +238,20 @@ impl Module {
...
+            }
         }
     }
 
     pub(crate) fn instantiate(
         &self,
         store: &mut impl AsStoreMut,
         imports: &Imports,
-    ) -> Result<WebAssembly::Instance, RuntimeError> {
+    ) -> Result<(crate::StoreHandle<WebAssembly::Instance>, Vec<Extern>), RuntimeError> {
```
theduke, line 254: This is a breaking change.

Need to investigate why it is needed.

lib/api/src/js/module.rs
theduke: I restored it to the previous state.

This was botched up during a rebase.

lib/vm/src/mmap.rs
```diff
@@ -256,6 +314,185 @@ impl Mmap {
...
+        trace!("memory copy finished (size={})", len);
+        Ok(new_mmap)
+    }
+}
+
+/// Rust implementation of libc::copy_file_range.
+///
+/// Needed because that function is not available on all platforms.
+// TODO: better implementation! (this is very quick, low effort)
+// TODO: this function needs tests!
```
syrusakbary, line 443: Add tests

lib/vm/src/mmap.rs
```diff
@@ -256,6 +314,185 @@ impl Mmap {
...
+        #[cfg(feature = "tracing")]
+        trace!("memory copy finished (size={})", len);
+        Ok(new_mmap)
+    }
+}
+
+/// Rust implementation of libc::copy_file_range.
+///
+/// Needed because that function is not available on all platforms.
+// TODO: better implementation! (this is very quick, low effort)
```
syrusakbary, line 442: See if we want to improve the implementation

lib/types/src/store.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// Represents a snapshot of parts of the store that mutate
+/// (such as globals and tables)
+#[derive(Debug, Default)]
+pub struct StoreSnapshot {
+    /// Global values at the time the snapshot was taken
+    pub globals: HashMap<u32, u128>,
+}
+
+impl StoreSnapshot {
```
syrusakbary, line 12: I'm a bit unclear on what this is for

syrusakbary (commented):
I think it should be possible to move the `OnCalledAction` outside of the API layer (perhaps move it to WASI). I think this might simplify the modifications on the main crate @theduke

theduke:
> I think it should be possible to move the `OnCalledAction` outside of the API layer (perhaps move it to WASI). I think this might simplify the modifications on the main crate @theduke

Agreed.
I'll see what can be done.

lib/types/src/store.rs
theduke: This is a copy of all the globals in a module.

Used for forking to copy over the global state.

I guess this is in here because the Store doesn't have a public api to access all globals.

.github/workflows/lint.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: 1.63
+          toolchain: 1.64
```
ptitSeb, line 26: Only the lint pass needs an upgrade to rust 1.64?

Makefile
```diff
@@ -367,7 +367,7 @@ check-capi: capi-setup
 		--no-default-features --features wat,compiler,wasi,middlewares $(capi_compiler_features)
 
 build-wasmer:
-	$(CARGO_BINARY) build $(CARGO_TARGET) --release --manifest-path lib/cli/Cargo.toml $(compiler_features) --features="webc_runner" --bin wasmer
+	$(CARGO_BINARY) build $(CARGO_TARGET) --release --manifest-path lib/cli/Cargo.toml --features cranelift,singlepass --features="webc_runner" --bin wasmer
```
ptitSeb, line 370: Mmm, that shuld not stay like that right?

fschutt: yeah, it was just for testing, because my Linux VM doesn't support building with LLVM

lib/vm/src/mmap.rs
```diff
@@ -256,6 +314,186 @@ impl Mmap {
...
+                };
+
+                // The shallow copy failed so we have to do it the hard way
+                let mut off_in: libc::off_t = 0;
+                let mut off_out: libc::off_t = 0;
+
+                cfg_if::cfg_if! {
+                    if #[cfg(not(any(target_env = "musl", target_vendor = "apple")))]
+                    {
+                        let ret = libc::copy_file_range(self.fd.0, &mut off_in, fd.0, &mut off_out, len, 0);
```
fschutt, line 364: This is causing the CI failure on linux-x64:

```
ld.lld --error-limit=0 -O3 -z
stack-size=16777216 --gc-sections --eh-frame-hdr -znow -m
aarch64linux -o /media/psf/Home/Development/wasmer/tests/integration/cli/debug/python
/home/parallels/.cache/zig/o/e87e9a86d6703349199f944fa8e74a29/Scrt1.o
/home/parallels/.cache/zig/o/e7ccbcf4c6260284654d7e852778c92c/crti.o -dynamic-linker
/lib/ld-linux-aarch64.so.1 /media/psf/Home/Development/wasmer/tests/integration/cli/debug/python.o
/media/psf/Home/Development/wasmer/tests/integration/cli/debug/libwasmer.a
/media/psf/Home/Development/wasmer/tests/integration/cli/debug/volume.obj
/home/parallels/.cache/zig/o/b0b7a26d74bbb6c7060ad372b02cbda0/wasmer_main.o
--as-needed /home/parallels/.cache/zig/o/f9abe4dcff02854b182b64b8691cfec5/libunwind.a
/home/parallels/.cache/zig/o/59a82fa70408f42b72de3e9040755e78/libm.so.6
/home/parallels/.cache/zig/o/59a82fa70408f42b72de3e9040755e78/libpthread.so.0
/home/parallels/.cache/zig/o/59a82fa70408f42b72de3e9040755e78/libc.so.6
/home/parallels/.cache/zig/o/59a82fa70408f42b72de3e9040755e78/libdl.so.2
/home/parallels/.cache/zig/o/59a82fa70408f42b72de3e9040755e78/librt.so.1
/home/parallels/.cache/zig/o/59a82fa70408f42b72de3e9040755e78/libld.so.2
/home/parallels/.cache/zig/o/59a82fa70408f42b72de3e9040755e78/libutil.so.1
/home/parallels/.cache/zig/o/59a82fa70408f42b72de3e9040755e78/libresolv.so.2
/home/parallels/.cache/zig/o/1952c727694664ba6c3c698fa8822f61/libc_nonshared.a
/home/parallels/.cache/zig/o/0acdc9e72b88424c9e3fd3a49be30b1a/crtn.o --allow-shlib-undefined
ld.lld: error: undefined symbol: copy_file_range
>>> referenced by wasmer_vm.bc74d6e7-cgu.3
>>>               wasmer_vm-992d0331437bc80b.wasmer_vm.bc74d6e7-cgu.3.rcgu.o:(wasmer_vm::mmap::Mmap::fork::h7e88e1f027f5f888)
                  in
                  archive /media/psf/Home/Development/wasmer/tests/integration/cli/debug/libwasmer.a
                  Error: python_exe_output_path debug/python does not
                  exist thread 'test_wasmer_create_exe_pirita_works'
                  panicked at 'assertion failed: `(left == right)`
```

However, I'm not sure what can be done about this. An easy fix would be to disable it on linux completely.

```suggestion
                    if #[cfg(not(any(target_family = "linux", target_vendor = "apple")))]
                    {
                        let ret = libc::copy_file_range(self.fd.0, &mut off_in, fd.0, &mut off_out, len, 0);
```

fschutt: Probably this has to do with zigs version of `libc` not being up-to-date.

I'd disable this for now since we have to test the fallback anyway.

Makefile
fschutt: Reverted in https://github.com/wasmerio/wasmer/pull/3426/commits/8b83cfcec88dea195595a6ff58d304d62c9db67b

lib/vm/src/mmap.rs
fschutt: I've verified that this would fix the CI issue on Linux.

theduke: I re-wrote the function and started using it on all unix platforms, so we should be good here.

Will re-open if the tests still have problems.

lib/api/src/js/export.rs
ptitSeb: I'm not in favor of `clone` either, as the operation is quite different from the `Clone` trait.

lib/api/src/js/export.rs
```diff
@@ -57,6 +110,29 @@ impl VMGlobal {
     pub(crate) fn new(global: Global, ty: GlobalType) -> Self {
         Self { global, ty }
     }
+
+    /// Saves the global value into the snapshot
+    pub fn save_snapshot(&self, index: usize, snapshot: &mut StoreSnapshot) {
+        if let Some(val) = self.global.as_f64() {
```
ptitSeb, line 116: What about Global that are of V128 type?

ptitSeb: Oh, not supported on JS, ok forget it then.

lib/api/src/js/export.rs
```diff
@@ -57,6 +110,29 @@ impl VMGlobal {
...
+            *entry = val as u128;
+        }
+    }
+
+    /// Restores the global value from the snapshot
+    pub fn restore_snapshot(&mut self, index: usize, snapshot: &StoreSnapshot) {
+        let index = index as u32;
+        if let Some(entry) = snapshot.globals.get(&index) {
+            if let Some(existing) = self.global.as_f64() {
+                let existing = existing as u128;
```
ptitSeb, line 127: this looks a bit mysterious to me. Are you sure you compare comparable things here (f64->u128 vs mem->u128 or did I miss something)?

lib/api/src/js/instance.rs
```diff
@@ -106,24 +105,43 @@ impl Instance {
     pub fn from_module_and_instance(
         mut store: &mut impl AsStoreMut,
         module: &Module,
-        instance: WebAssembly::Instance,
+        externs: Vec<Extern>,
```
ptitSeb, line 108: That looks like a breaking API change.

lib/api/src/sys/instance.rs
```diff
@@ -166,6 +175,14 @@ impl Instance {
             })
             .collect::<Exports>();
 
+        // If the memory is imported then also export it for backwards compatibility reasons
+        // (many will assume the memory is always exported) - later we can remove this
+        if exports.get_memory("memory").is_err() {
```
ptitSeb, line 180: I don't understand how this is required for backward compatibility while it's new code

lib/c-api/Cargo.toml
```diff
@@ -32,7 +32,7 @@ wasmer-middlewares = { version = "=3.1.0", path = "../middlewares", optional = t
 wasmer-wasi = { version = "=3.1.0", path = "../wasi", default-features = false, features = ["host-fs", "sys"], optional = true }
 wasmer-types = { version = "=3.1.0", path = "../types" }
 wasmer-vfs = { version = "=3.1.0", path = "../vfs", optional = true, default-features = false, features = ["static-fs"] }
-webc = { version = "3.0.1", optional = true }
+webc = { version = "0.4.1", optional = true }
```
ptitSeb, line 35: Really 3.0.1 -> 0.4.1?

fschutt: Yeah, this was a mistake from when I thought we should match the webc version to the wasmer version

lib/emscripten/src/memory.rs
```diff
@@ -110,6 +110,8 @@ pub fn sbrk(mut ctx: FunctionEnvMut<EmEnv>, increment: i32) -> i32 {
         increment,
         total_memory
     );
+    let _ = dynamictop_ptr;
```
ptitSeb, line 113: What's the point of this?

lib/vm/src/mmap.rs
```diff
@@ -66,15 +99,37 @@ impl Mmap {
...
             // Allocate a single read-write region at once.
             let ptr = unsafe {
                 libc::mmap(
                     ptr::null_mut(),
                     mapping_size,
                     libc::PROT_READ | libc::PROT_WRITE,
-                    libc::MAP_PRIVATE | libc::MAP_ANON,
-                    -1,
+                    flags,
+                    fd.0,
```
ptitSeb, line 132: Doesn't the systematic file mapping will slowdown things (compared to previous Private Anonymous map)?

And I'm still unsure of the purpose of this.

ptitSeb (commented):
Pushed a few comments...

lib/api/src/js/export.rs
theduke: Renamed to `duplicate()`.

lib/api/src/sys/tunables.rs
theduke: Renamed to `duplicate()`.

lib/vm/src/memory.rs
theduke: Renamed to `duplicate()`.

lib/vm/src/memory.rs
theduke: Renamed to `duplicate()`.

lib/vm/src/memory.rs
theduke: Renamed to `duplicate()`.

lib/vm/src/memory.rs
theduke: Renamed to `duplicate()`.

lib/vm/src/memory.rs
theduke: Renamed to `duplicate()`.

lib/vm/src/memory.rs
theduke: Renamed to `duplicate()`.

lib/vm/src/memory.rs
theduke: Renamed to `duplicate()`.

lib/vm/src/memory.rs
theduke: Renamed to `duplicate()`.

lib/api/src/js/module.rs
theduke: We need to ask @john-sharratt .

lib/api/src/js/module.rs
theduke: @john-sharratt I agree that this `is_ok()` is quite awakard. 

Can we implement that a bit cleaner?

lib/vm/src/mmap.rs
theduke: Has one test now.

Should get more in the future, but it'll do for now.

Especially since we might refactor this quite a bit to remove the file layer dependency.

lib/vm/src/mmap.rs
theduke: And good enough until we run into performance issues.

lib/c-api/Cargo.toml
theduke: Yes, `0.4.1` is the latest version.

Downgrading a major version of a dependency is still a breaking change.

We are stuck in an awkward position here.

We can publich a `3.x` with the latest changes, but that would probably violate semver for `webc`.

Or we can publish a `4.x` webc and upgrade Wasmer to that, but technically a dependency major version upgrade is a breaking change.

lib/api/src/js/instance.rs
syrusakbary: We probably don't want this

lib/api/src/sys/instance.rs
syrusakbary: We should not have the logic on the main API

lib/api/src/js/externals/function.rs
```diff
@@ -402,11 +415,34 @@ impl Function {
...
+                        Ok(wasmer_types::OnCalledAction::Trap(trap)) => {
+                            return Err(RuntimeError::user(trap))
+                        }
+                        Err(trap) => return Err(RuntimeError::user(trap)),
+                    }
+                }
+                break;
+            }
+            r?
+        };
```
syrusakbary, line 445: We can move this into the wasix library itself, I think it'll be cleaner

lib/types/src/memory.rs
```diff
@@ -59,6 +59,8 @@ pub unsafe trait MemorySize: Copy {
         + PartialOrd<Self::Offset>
         + Clone
         + Copy
+        + Sync
```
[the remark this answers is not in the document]
syrusakbary: I'm ok with this breaking change

lib/types/src/trapcode.rs
```diff
@@ -120,6 +120,18 @@ impl FromStr for TrapCode {
...
+#[derive(Debug)]
+pub enum OnCalledAction {
+    /// Will call the function again
+    InvokeAgain,
+    /// Will return the result of the invocation
+    Finish,
+    /// Traps with an error
+    Trap(Box<dyn std::error::Error + Send + Sync>),
+}
+
```
syrusakbary, line 134: Move this to WASI (up to discussion)

lib/api/src/js/externals/memory.rs
```diff
@@ -86,19 +86,26 @@ impl Memory {
...
+        let vm_memory = VMMemory::new(Self::new_internal(ty.clone())?, ty);
+        Ok(Self::from_vm_export(store, vm_memory))
+    }
+
+    pub(crate) fn new_internal(ty: MemoryType) -> Result<js_sys::WebAssembly::Memory, MemoryError> {
         let descriptor = js_sys::Object::new();
-        js_sys::Reflect::set(&descriptor, &"initial".into(), &ty.minimum.0.into()).unwrap();
-        if let Some(max) = ty.maximum {
-            js_sys::Reflect::set(&descriptor, &"maximum".into(), &max.0.into()).unwrap();
+        #[allow(unused_unsafe)]
```
syrusakbary, line 95: Probably only one decorator is required (there are two unused_unsafe here)

lib/api/src/sys/externals/function.rs
```diff
@@ -713,16 +803,19 @@ impl<'a> Exportable<'a> for Function {
 }
 
 /// Host state for a dynamic function.
+#[cfg(feature = "compiler")]
```
syrusakbary, line 806: We should not require compiler in the function externals anywhere (as the trampolines should be present when running a precompiled module)

lib/api/src/sys/instance.rs
```diff
@@ -129,6 +130,14 @@ impl Instance {
...
+            if let Some(memory) = externs.iter().find(|a| a.ty(store).memory().is_some()) {
+                exports.insert("memory", memory.clone());
+            }
+        }
```
syrusakbary, line 139: I don't think this is right

syrusakbary (commented):
We need deep review of this with @john-sharratt, including:
* OnCalledAction optimization outside of the core
* JS breaking changes

Apart from this, we need:
* Tests for snapshots @theduke 
* Move snapshotting outside of the core wasmer crate (but enable it with an iterator over the Store object, so we are able to filter the globals and do operations on them)

.github/workflows/lint.yaml
theduke: Actually, the new webc needs 1.64.

Do we know why that is, @fschutt ?

But if our policy is current + last 2 versions then bumping to 1.64 would be fine, sine 1.66 just released.

fschutt: @theduke as far as I know, it needs 1.64 for inheriting cargo metadata, but you'd need to ask @Michael-F-Bryan about this, he did the version bumping

ptitSeb: My issue here is that just lint.yaml is bumped, not any other yaml that are stying in 1.63
This is inconsistant, so it's either not needed on linter or missing everywhere else.

theduke: Oh, well, the `rust-toolchain` file also has the bump!

ptitSeb: Ok, I pused a commit to bumped the rust toolchain to 1.64 for all CI step then.

lib/c-api/Cargo.toml
ptitSeb: I vote for an upgrade to 4.x. we cannot stay in this situation.

lib/vm/src/mmap.rs
ptitSeb: Even for a suspend on disk, I think the performance penalty is not worth it. The suspend on disk is a "rare" operation compare to read/write memory.

theduke: Suspend to disk will be very common on Deploy servers, because many instances will be idle for a long time, so we want to evict them from memory.

You can actually configure a mapping so it will only sync to disk when there is memory pressure.
I'm not sure yet how well this will work in practice but that's the plan.

Either way it should not be the default.

We either need a way to configure the mmap behaviour, or we just duplicate the OwnedMemory and SharedMemory in deploy.

theduke: Probably better to just duplicate for now, it's not a whole lot of code.

theduke: Actually, that might require us adding a way to downcast memory to a concrete type (by adding something like `LinearMemory::as_any(&self) -> Option<&dyn Any>`.

lib/types/src/memory.rs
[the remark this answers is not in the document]
theduke: Alright, we'll let this slip through, since no-one would implement this trait anyway...

lib/api/src/js/module.rs
theduke: Crate local, so not a breaking change...

lib/vm/src/store.rs
```diff
@@ -104,6 +105,19 @@ impl StoreObjects {
...
     }
 
+    /// Return an immutable iterator over all globals
+    pub fn iter_globals(&self) -> Iter<VMGlobal> {
+        self.globals.iter()
+    }
+
+    /// Set a global, at index idx. Will panic if idx is out of range
+    pub fn set_global(&self, idx: usize, val: u128) {
+        assert!(idx >= self.globals.len());
```
theduke, line 115: Shouldn't this be `<=` ?

In general, it seems unexpected that this API signature would panic.

Should probably return a `Result<(), ?>` instead.

lib/vm/src/store.rs
```diff
@@ -104,6 +105,19 @@ impl StoreObjects {
...
+        unsafe {
+            self.globals[idx].vmglobal().as_mut().val.u128 = val;
```
theduke, line 117: Since this circumvents the type checks that we get with `Global`, maybe we should call the function `set_global_unchecked`, and make it clear in the function comment that the caller is responsible for writing a correct value for the globals type.

lib/vm/src/store.rs
theduke: Or rather `<`.

ptitSeb: fixed

lib/api/src/sys/externals/function.rs
theduke: Changes were reverted.

lib/c-api/Cargo.toml
theduke: I have asked Felix to release `4.0` .

lib/api/src/sys/instance.rs
john-sharratt: without exporting it (for things compiled without `atomics`) we would need a way of passing the memory to WASIX in some other construct

lib/api/src/js/externals/function.rs
john-sharratt: Even if we move this, there still needs to be changes made to this main loop or `asyncify` will not work - it needs to be able to detect if the main loop is replaying a stack or recording a stack

lib/vm/src/mmap.rs
john-sharratt: as discussed on notion - no problem with having multiple implementations of the memory but for Deploy we need high density (and so a backing file)

lib/c-api/Cargo.toml
john-sharratt: yeah i think 4.0 is a better way forward

lib/api/src/sys/instance.rs
john-sharratt: we could always try removing this and seeing if the regression fails - I'm not 100% sure its needed looking at the code again - I seem to remember regression issues when I tried to take it out (hence the comment)

lib/api/src/js/instance.rs
john-sharratt: this was to reduce code duplication here but we could try and take it out - wasmer-js is working

lib/api/src/js/module.rs
john-sharratt: this is no longer used and can be removed - we now share modules between threads properly which reduces memory usage significantly

lib/api/src/js/module.rs
john-sharratt: The imported/exported memory does not have the correct properties (incorrect size and shared flag) hence when using shared memory its failing - the only way to fix it is to resolve the import and use the correct memory properties. this regression issue was only found in WASIX on the browser as the other areas don't mind that they don't match up

lib/types/src/trapcode.rs
john-sharratt: see earlier comment this is needed for `asyncify` to work - it can't be moved into WASI completely as the main loop needs to change hence this was an attempt to at least keep the changes needed to a minimum in the main API.

lib/c-api/Cargo.toml
fschutt: @theduke it's released now: https://crates.io/crates/webc/4.0.0

lib/api/src/js/instance.rs
theduke: @john-sharratt I already started removing this one!

lib/api/src/js/module.rs
theduke: Methods removed.

lib/c-api/Cargo.toml
theduke: Was released as 4.0, and the corresponding Cargo.toml files were updated.

NOTE: need to remember that we need a breaking change.

lib/compiler-cranelift/Cargo.toml
```diff
@@ -38,3 +38,4 @@ wasm = ["std", "unwind"]
 unwind = ["cranelift-codegen/unwind", "gimli"]
 std = ["cranelift-codegen/std", "cranelift-frontend/std", "wasmer-compiler/std", "wasmer-types/std"]
 core = ["hashbrown", "cranelift-codegen/core", "cranelift-frontend/core"]
+verbose = []
```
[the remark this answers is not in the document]
john-sharratt: yeah this isn't needed anymore thanks to the release compile optimization trick you showed me here:
https://github.com/wasmerio/deploy/blob/c3b0ba3cd89736641c70f00c99bb41c2ad7e9d8b/Cargo.toml#L29

It's safe to remove now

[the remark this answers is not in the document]
theduke: Removed.

lib/vm/src/mmap.rs
theduke: FD MMap implementation was removed from `vm`. 

A separate memory implementation was added in `wasi`.

syrusakbary:
Clarification:
* We will move the asyncify stuff (removing the internals `wasmer_types::OnCalledAction`) into it's own create (that will be reused by WASI) before the official release
* Hide out any OnCalledAction or related stuff from the lib

lib/api/src/js/externals/function.rs
syrusakbary: Note for the future: We agreed to move this logic out of the API after merging the PR and WASIX merges

lib/api/src/sys/native.rs
```diff
@@ -87,15 +94,110 @@ macro_rules! impl_native_traits {
...
-                unsafe {
-                    wasmer_vm::wasmer_call_trampoline(
-                        store.as_store_ref().signal_handler(),
-                        anyfunc.vmctx,
-                        anyfunc.call_trampoline,
-                        anyfunc.func_ptr,
-                        args_rets.as_mut_ptr() as *mut u8,
-                    )
-                }?;
+
```
syrusakbary, line 97: Can we add a new comment linking to the asyncify refactor afterwards:

```
// TODO: issue #XYZ 
```

lib/api/src/js/externals/function.rs
```diff
@@ -402,11 +415,34 @@ impl Function {
             let js_value = param.as_jsvalue(&store.as_store_ref());
             arr.set(i as u32, js_value);
         }
-        let result = js_sys::Reflect::apply(
-            &self.handle.get(store.as_store_ref().objects()).function,
-            &wasm_bindgen::JsValue::NULL,
-            &arr,
-        )?;
+
```
syrusakbary, line 418: Can we add a new comment linking to the asyncify refactor afterwards:

```
// TODO: issue #XYZ 
```

lib/api/src/js/native.rs
```diff
@@ -64,11 +66,34 @@ macro_rules! impl_native_traits {
             pub fn call(&self, mut store: &mut impl AsStoreMut, $( $x: $x, )* ) -> Result<Rets, RuntimeError> where
             $( $x: FromToNativeWasmType + crate::js::NativeWasmTypeInto, )*
             {
-                let params_list: Vec<JsValue> = vec![ $( JsValue::from_f64($x.into_raw(&mut store))),* ];
```
syrusakbary, line 67: Can we add a new comment linking to the asyncify refactor afterwards:

```
// TODO: issue #XYZ 
```

lib/api/src/sys/externals/function.rs
```diff
@@ -401,24 +430,61 @@ impl Function {
...
-        let vm_function = self.handle.get(store.as_store_ref().objects());
-        if let Err(error) = unsafe {
-            wasmer_call_trampoline(
-                store.as_store_ref().signal_handler(),
-                vm_function.anyfunc.as_ptr().as_ref().vmctx,
-                trampoline,
-                vm_function.anyfunc.as_ptr().as_ref().func_ptr,
-                values_vec.as_mut_ptr() as *mut u8,
-            )
-        } {
```
syrusakbary, line 414: Can we add a new comment linking to the asyncify refactor afterwards:

```
// TODO: issue #XYZ 
```

lib/api/src/sys/mod.rs
```diff
@@ -42,7 +42,7 @@ pub use wasmer_derive::ValueType;
 pub use wasmer_types::is_wasm;
 pub use wasmer_types::{
     CpuFeature, ExportType, ExternType, FunctionType, GlobalType, ImportType, MemoryType,
-    Mutability, TableType, Target, Type,
+    Mutability, OnCalledAction, TableType, Target, Type,
```
syrusakbary, line 45: Can we add a new comment linking to the asyncify refactor afterwards:

```
// TODO: issue #XYZ 
```

lib/api/src/sys/store.rs
```diff
@@ -309,6 +326,17 @@ impl<'a> StoreMut<'a> {
     pub(crate) unsafe fn from_raw(raw: *mut StoreInner) -> Self {
         Self { inner: &mut *raw }
     }
+
+    /// Sets the unwind callback which will be invoked when the call finishes
+    pub fn on_called<F>(&mut self, callback: F)
```
syrusakbary, line 331: Can we add a new comment linking to the asyncify refactor afterwards:

```
// TODO: issue #XYZ 
```

lib/types/src/lib.rs
```diff
@@ -105,7 +105,7 @@ pub use value::{RawValue, ValueType};
 pub use crate::libcalls::LibCall;
 pub use crate::memory::MemoryStyle;
 pub use crate::table::TableStyle;
-pub use crate::trapcode::TrapCode;
+pub use crate::trapcode::{OnCalledAction, TrapCode};
```
syrusakbary, line 108: Can we add a new comment linking to the asyncify refactor afterwards:

```
// TODO: issue #XYZ 
```

lib/types/src/trapcode.rs
syrusakbary: Can we add a new comment linking to the asyncify refactor afterwards:

```
// TODO: issue #XYZ 
```

lib/api/Cargo.toml
```diff
@@ -117,7 +118,7 @@ jit = ["engine"]
 js = ["wasm-bindgen", "js-sys"]
 js-default = ["js", "std", "wasm-types-polyfill"]
 
-wasm-types-polyfill = ["js", "wasmparser"]
+wasm-types-polyfill = ["js", "wasmparser", "js-serializable-module" ]
```
syrusakbary, line 121: Why we require the module to be serializable when we do the polyfill? It shouldn't be required

lib/api/src/js/imports.rs
```diff
@@ -172,11 +171,17 @@ impl Imports {
         for (ns, exports) in namespaces.into_iter() {
             let import_namespace = js_sys::Object::new();
             for (name, ext) in exports {
-                js_sys::Reflect::set(&import_namespace, &name.into(), &ext.as_jsvalue(store))
-                    .expect("Error while setting into the js namespace object");
+                #[allow(unused_unsafe)]
+                unsafe {
```
syrusakbary, line 175: Why we are doing this? `#[allow(unused_unsafe)]` and then `unsafe {`. Doesn't make that much sense.
Perhaps I'm missing something?

lib/api/src/js/imports.rs
```diff
@@ -233,29 +238,40 @@ impl Imports {
...
                 let import_namespace = js_sys::Object::new();
-                js_sys::Reflect::set(
-                    &import_namespace,
-                    &name.into(),
-                    &extern_.as_jsvalue(&store.as_store_ref()),
-                )
-                .unwrap();
-                js_sys::Reflect::set(&imports_object, &namespace.into(), &import_namespace.into())
+                #[allow(unused_unsafe)]
+                unsafe {
```
syrusakbary, line 261: Why we are doing this? `#[allow(unused_unsafe)]` and then `unsafe {`. Doesn't make that much sense.
Perhaps I'm missing something?

lib/api/src/js/instance.rs
```diff
@@ -106,24 +105,43 @@ impl Instance {
...
+        // (many will assume the memory is always exported) - later we can remove this
+        if exports.get_memory("memory").is_err() {
+            if let Some(memory) = externs
+                .iter()
+                .filter(|a| a.ty(store).memory().is_some())
+                .next()
+            {
+                exports.insert("memory", memory.clone());
+            }
+        }
```
syrusakbary, line 143: We shouldn't have this logic in the main API. We should move it outside

lib/api/src/js/module.rs
```diff
@@ -254,17 +238,20 @@ impl Module {
     /// validation of the Module.
     pub fn validate(_store: &impl AsStoreRef, binary: &[u8]) -> Result<(), CompileError> {
         let js_bytes = unsafe { Uint8Array::view(binary) };
-        match WebAssembly::validate(&js_bytes.into()) {
-            Ok(true) => Ok(()),
-            _ => Err(CompileError::Validate("Invalid Wasm file".to_owned())),
+        #[allow(unused_unsafe)]
+        unsafe {
```
syrusakbary, line 242: Why we are doing this? `#[allow(unused_unsafe)]` and then `unsafe {`. Doesn't make that much sense.
Perhaps I'm missing something?

lib/api/src/js/module.rs
```diff
@@ -466,10 +528,13 @@ impl Module {
             return Err("The exports length must match the type hints lenght".to_owned());
         }
         for (i, val) in exports.iter().enumerate() {
-            let kind = Reflect::get(val.as_ref(), &"kind".into())
-                .unwrap()
-                .as_string()
-                .unwrap();
+            #[allow(unused_unsafe)]
+            let kind = unsafe {
```
syrusakbary, line 532: Why we are doing this? `#[allow(unused_unsafe)]` and then `unsafe {`. Doesn't make that much sense.
Perhaps I'm missing something?

lib/vm/src/function_env.rs
```diff
@@ -1,7 +1,11 @@
+use derivative::Derivative;
 use std::any::Any;
 
 /// Underlying FunctionEnvironment used by a `VMFunction`.
+#[derive(Derivative)]
+#[derivative(Debug)]
```
syrusakbary, line 6: What do we gain by making this Debug?

lib/wasi/src/runners/mod.rs
```diff
@@ -107,11 +107,11 @@ impl Bindings for WitBindings {
         container: &WapmContainer,
         value: &serde_cbor::Value,
     ) -> Result<Self, String> {
-        let value: webc::WitBindingsExtended =
+        let value: webc::BindingsExtended =
```
syrusakbary, line 110: This seems wrong. Am I missing something?

syrusakbary: cc @Michael-F-Bryan

theduke: That's the latest release.

https://docs.rs/webc/latest/webc/enum.BindingsExtended.html

lib/api/src/js/module.rs
theduke: I think this is due to js bindgen compilation quirks.

Some functions apparently are unsafe when compiled to wasm32-, but aren't unsafe when the target is native Rust.

So this prevents compilation warnings/errors when doing `cargo check` without `--target wasm32-*`.

syrusakbary: That's weird, because `Reflect::get` is a safe function for all targets (thus usage should also be safe).
https://rustwasm.github.io/wasm-bindgen/api/js_sys/Reflect/fn.get.html

I don't think this is really necessary and it bloats a bit the code?

theduke: I'm not sure where it comes in, exactly, but I did try to remove the annotation and then had "requires unsafe" errors.

lib/vm/src/function_env.rs
theduke: It's useful for debugging.

In general all public types should be `Debug` - that's also in the Rust API guidelines.

ptitSeb (approved):

lib/api/src/js/imports.rs
ptitSeb: `js_sys::Reflect::get(...)` is `extern "C"`. My guess is that it's not unsafe depending were the rust code for the function itself is defined relative to the crate built.

ptitSeb: So, the `#[allow(unused_unsafe)]` / `unsafe { ... }` can be remove, it will build fine. But IDE like VSCode are not happy

image

So I think it's better to keep it that way for now.

lib/vm/src/function_env.rs
syrusakbary: I don't think any VM structure should be debuggable to be honest. It's not meant to be used by anyone outside of wasmer

lib/compiler/src/compiler.rs
```diff
@@ -79,6 +79,12 @@ where
 
 /// An implementation of a Compiler from parsed WebAssembly module to Compiled native code.
 pub trait Compiler: Send {
+    /// Returns a descriptive name for this compiler.
+    // TODO!: Probably want to make this mandatory on the next major bump.
+    fn name(&self) -> &str {
+        "UNKNOWN"
+    }
```
syrusakbary, line 86: We should just not have this implemented, even if it's a breaking change.
Right now I don't know of any user/company using their custom compilers

lib/vm/src/function_env.rs
ptitSeb: Removing this makes `StoreObjects` non `Debug` too, wich contains many usefull Debug information. So it would be better to keep it that way.

lib/api/src/js/module.rs
syrusakbary: Can you add a test/example that reproduces this? I think we can simplify but without a base test/example that we can test against it will be hard

lib/api/src/js/native.rs
```diff
@@ -64,11 +66,35 @@ macro_rules! impl_native_traits {
...
+                let params_list: Vec<RawValue> = unsafe {
+                    vec![ $( RawValue { f64: $x.into_raw(store) } ),* ]
+                };
+                let params_list: Vec<JsValue> = params_list
+                    .into_iter()
+                    .map(|a| a.as_jsvalue(&store.as_store_ref()))
+                    .collect();
+                let results = {
+                    let mut r;
+                    //TODO: This loop is needed for async. It will be refactor with https://github.com/wasmerio/wasmer/issues/3451
```
syrusakbary, line 79: Can we add a space on the comment and change async to asyncify? (do this in all the comments)

```suggestion
// TODO: This loop is needed for asyncif. It will be refactored with https://github.com/wasmerio/wasmer/issues/3451
```

lib/api/src/js/module.rs
john-sharratt: `sharrattj/dash` should be able to reproduce this but the problem is that's going to be quite hard now that the merge is split between two as it won't work without the second commit.

basically if dash can run (including with coreutils commands) without this piece of code then we can delete it

lib/api/src/js/native.rs
ptitSeb: done

lib/api/src/sys/externals/function.rs
ptitSeb: done

lib/api/src/sys/mod.rs
ptitSeb: done

lib/api/src/sys/native.rs
ptitSeb: done

lib/api/src/sys/store.rs
ptitSeb: done

lib/types/src/lib.rs
ptitSeb: done

lib/types/src/trapcode.rs
ptitSeb: done

syrusakbary (approved):
Looks good.

lib/api/src/js/native.rs
ptitSeb: done

lib/compiler/src/compiler.rs
ptitSeb: done

lib/api/src/js/native.rs
theduke: Decided to keep it.
Code now contains a TODO notice.

lib/api/src/js/module.rs
theduke: Code now has a TODO to refactor this.

lib/api/src/js/imports.rs
theduke: Code now has comments that explain it.

lib/api/src/js/imports.rs
theduke: Discussed in sibling threads.

Code now has an explanation comment.

lib/api/src/sys/instance.rs
theduke: discussed, stays for now

lib/api/src/sys/instance.rs
theduke: stays for now

PR #3427, opened 2022-12-15
Fix cargo-deny failing on webc crate

fschutt:


theduke:
I guess it fails because of the unmaintained warnings?

Guess we'll need to whitelist those as well.

deny.toml
```diff
@@ -169,6 +169,7 @@ deny = [
 ]
 # Certain crates/versions that will be skipped when doing duplicate detection.
 skip = [
+    { name = "webc" },
```
theduke, line 172: Please add comments on why these are whitelisted.

theduke (approved):
Approved after the mentioned comments are added.

deny.toml
ptitSeb: done

theduke:
@fschutt I actually meant comments for all the crates, not just webc, because we will have no idea why they are whitelisted in a months time.

But we can just include that in the #3426 PR.

PR #3428, opened 2022-12-15
Implement wasmer config

### LINKED ISSUE #3383: Implement wasmer config
`wasmer config set/get` should work exactly the same way as `wapm config` (even using the same `~/.wasmer/wapm.toml config` file)

fschutt:
Fixes #3383.


ptitSeb:
I would prefer a simpler:
`config XXX` instead of  `config get XXX`
and
`set-config XXX YYY` instead of `config set XXX YYY`

But yeah, the switch from `config --XXX` is a bit strange to me, as it's the common way to query config on various linux tools (pkg-config, llvm-config, sdl2-config...).

That's just personnal taste, nothing blocking.

ptitSeb (approved):

lib/registry/src/config.rs
```diff
@@ -36,16 +36,32 @@ pub struct Proxy {
     pub url: Option<String>,
 }
 
-#[derive(Deserialize, Serialize, Debug, PartialEq, Eq, Default)]
+#[derive(Deserialize, Serialize, Debug, PartialEq, Eq)]
 pub struct UpdateNotifications {
     pub enabled: String,
```
theduke, line 41: I know you didn't change this here, but this should just be a `bool`.

serde-toml handles that just fine with `true` / `false`.

lib/registry/src/config.rs
```diff
@@ -36,16 +36,32 @@ pub struct Proxy {
...
+    fn default() -> Self {
+        UpdateNotifications {
+            enabled: "false".to_string(),
+        }
+    }
+}
+
+#[derive(Deserialize, Serialize, Debug, PartialEq, Eq)]
 pub struct Telemetry {
     pub enabled: String,
```
theduke, line 54: Should just be a bool

lib/cli/src/commands/config.rs
```diff
@@ -3,38 +3,111 @@ use anyhow::{Context, Result};
...
 
-    /// Libraries needed to link against Wasmer components.
-    #[clap(long, conflicts_with = "pkg-config")]
-    libs: bool,
+/// Set if update notifications are enabled
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Parser)]
+pub struct SetUpdateNotificationsEnabled {
+    /// Whether to enable update notifications
+    #[clap(name = "ENABLED", possible_values = ["true", "false"])]
+    pub enabled: String,
```
theduke, line 94: You can make this a bool, and use something like 

```
struct BoolString(bool);

impl std::str::FromStr for BoolString {
   ...
}
```

That way the other code becomes cleaner, and we can also accept something like `y` and `n` more easily.

theduke (dismissed):

lib/c-api/README.md
```diff
@@ -148,48 +148,48 @@ $ make test-capi
 
 `wasmer config` output various configuration information needed to compile programs which use Wasmer.
 
-### `wasmer config --pkg-config`
+### `wasmer config get pkg-config`
```
syrusakbary, line 151: We shouldn't change this in any case

lib/c-api/README.md
```diff
@@ -148,48 +148,48 @@ $ make test-capi
...
 It outputs the necessary details for compiling and linking a program to Wasmer,
 using the `pkg-config` format:
 
 ```bash
-$ wasmer config --pkg-config > $PKG_CONFIG_PATH/wasmer.pc
+$ wasmer config get pkg-config > $PKG_CONFIG_PATH/wasmer.pc
 ```
 
-### `wasmer config --includedir`
+### `wasmer config get includedir`
```
syrusakbary, line 160: We shouldn't change this

lib/c-api/README.md
```diff
@@ -148,48 +148,48 @@ $ make test-capi
...
 
 Directory containing Wasmer headers:
 
 ```bash
-$ wasmer config --includedir
+$ wasmer config get includedir
```
syrusakbary, line 165: We shouldn't change this

syrusakbary (changes requested):
`wasmer config --pkg-config` and others (`--include-dir`) should NOT live in `wasmer config get` in any case. `wasmer config get {KEY}` are for settings that can be configured by the user via the counterpart `wasmer config set {KEY} {VALUE}`.

We need to preserve old behavior

lib/registry/src/config.rs
fschutt: Okay, switched to a new key so it doesn't break wapm-cli.

lib/registry/src/config.rs
fschutt: Switched to a new key so it doesn't break wapm-cli

lib/cli/src/commands/config.rs
fschutt: ok done

lib/c-api/README.md
fschutt: Reverted in https://github.com/wasmerio/wasmer/pull/3428/commits/710b8242f1f770271c5ab0a132f2e152047cb7dc

lib/c-api/README.md
fschutt: Reverted in https://github.com/wasmerio/wasmer/pull/3428/commits/710b8242f1f770271c5ab0a132f2e152047cb7dc

lib/c-api/README.md
fschutt: Reverted in https://github.com/wasmerio/wasmer/pull/3428/commits/710b8242f1f770271c5ab0a132f2e152047cb7dc

tests/integration/cli/tests/config.rs
```diff
@@ -0,0 +1,196 @@
...
+        .arg("config")
+        .arg("--libs")
+        .output()?;
+
+    let lib_path = Path::new(env!("WASMER_DIR")).join("lib");
+    assert_eq!(
+        String::from_utf8(bindir.stdout).unwrap(),
+        format!("-L{} -lwasmer\n", lib_path.display())
+    );
+
```
syrusakbary, line 62: Can you test two flags at once?

Right now:

```bash
$ wasmer config --bindir --cflags
/Users/syrusakbary/.wasmer/bin
-I/Users/syrusakbary/.wasmer/include
```

But `--pkg-config` should not be used with other flags together:

```bash
$ wasmer config --bindir --cflags --pkg-config
error: The argument '--bindir' cannot be used with '--pkg-config'

USAGE:
    wasmer config --bindir --cflags

For more information try --help
```

fschutt: Implemented in https://github.com/wasmerio/wasmer/pull/3428/commits/e96c31318b56f7648271b702fe272da8d3a47327 and https://github.com/wasmerio/wasmer/pull/3428/commits/ca50d8218c19da7485014544606d187523ac8815

tests/integration/cli/tests/config.rs
```diff
@@ -0,0 +1,72 @@
+use anyhow::bail;
+use std::path::PathBuf;
+use std::process::Command;
+use wasmer_integration_tests_cli::get_wasmer_path;
+
+#[test]
+fn config_works() -> anyhow::Result<()> {
+    let output = Command::new(get_wasmer_path())
```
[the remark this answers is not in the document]
fschutt: Implemented in https://github.com/wasmerio/wasmer/pull/3428/commits/f1aa1d5f536573cd1ba8c5bc6dcaeeeddae27673

fschutt:
@syrusakbary should be ready to merge when you sign off

lib/cli/src/commands/config.rs
```diff
@@ -3,10 +3,22 @@ use anyhow::{Context, Result};
 use clap::Parser;
 use std::env;
 use std::path::PathBuf;
+use std::str::ParseBoolError;
+use wasmer_registry::WasmerConfig;
 
 #[derive(Debug, Parser)]
-/// The options for the `wasmer config` subcommand
+/// The options for the `wasmer config` subcommand: `wasmer config get prefix`
```
syrusakbary, line 10: What is `get prefix`?

lib/registry/src/config.rs
```diff
@@ -1,26 +1,25 @@
...
-    pub telemetry: Telemetry,
+    pub telemetry_enabled: bool,
 
     /// Whether or not updated notifications are enabled.
     #[serde(default)]
-    pub update_notifications: UpdateNotifications,
+    pub update_notifications_enabled: bool,
+
+    /// The registry that wapm will connect to.
+    pub registry: Registries,
```
syrusakbary, line 22: I have no idea what registries is here

lib/cli/src/commands/config.rs
```diff
@@ -37,13 +49,124 @@ pub struct Config {
...
+    #[clap(name = "update-notifications.enabled")]
+    UpdateNotificationsEnabled(SetUpdateNotificationsEnabled),
+    /// `proxy.url`
+    #[clap(name = "proxy.url")]
+    ProxyUrl(SetProxyUrl),
+}
+
+/// Set the current active registry URL
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Parser)]
+pub struct SetRegistryUrl {
```
syrusakbary, line 108: Where is this used? I don't see it anywhere?

syrusakbary: (same for the other setters)

fschutt: These were leftovers from when I was experimenting with clap on how to do CLI flags properly.

syrusakbary: Cool, then we shall remove them

lib/registry/src/config.rs
syrusakbary: What we should do here is:

Have a field tokens that associate a given registry with a token.

We should also have a new field called: current registry.


Config should be like

```json
tokens = [{
    registry: "https://registry.wapm.dev/graphql",
    token: "abc"
}, ...]
active_registry = "https://registry.wapm.io/graphql"
```

lib/cli/src/commands/config.rs
fschutt: It's not a leftover, sorry, it's necessary so I can `#[derive(clap::Parser)]`

syrusakbary: We are using them actually, and they are needed. Resolving this comment

lib/cli/src/commands/config.rs
```diff
@@ -37,13 +49,124 @@ pub struct Config {
...
+    #[clap(subcommand)]
+    Set(StorableConfigField),
+}
+
+/// Value that can be queried from the wasmer config
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Parser)]
+pub enum RetrievableConfigField {
+    /// Print the path to the configuration file
+    #[clap(name = "config.path")]
+    ConfigPath,
```
syrusakbary, line 68: Let's remove this. If we want it, we can do it via `wasmer config --path`

fschutt:
@syrusakbary the default config now looks like this:

```toml
wax_cooldown = 0
telemetry_enabled = false
update_notifications_enabled = false

[registry]
active_registry = "https://registry.wapm.dev/graphql"
tokens = []

[proxy]

```

1. Do we still need the `wax_cooldown`, since we don't have wax anymore?
2. We need to watch out that the switch from `waom.toml` to `wasmer.toml` is as smooth as possible:

https://github.com/wasmerio/wasmer/blob/58ee1740564d1c14a235e6cbaa68bf98147da3e8/lib/registry/src/config.rs#L206

... will fail if we do non-backwards compatible breaking changes to the `wasmer.toml`. Would be better to return `Self::default()` here in the error case.

syrusakbary:
> Do we still need the wax_cooldown, since we don't have wax anymore?

Probably not, but we should assure that when renaming the old `wapm.toml` to `wasmer.toml` the reading doesn't crash

syrusakbary (approved):

theduke (approved):
LGTM, relying on @syrusakbary sign off yesterday.

PR #3432, opened 2022-12-16
Amend changes to wasmer init

fschutt:
This reverts commit 250ebff4e8ce250e40c61d273e5627adf3c0f90d.

This is a continuation of https://github.com/wasmerio/wasmer/pull/3324 (to keep the git history in a proper format), so we can address review comments that came in after the PR was merged.


lib/registry/src/lib.rs
```diff
@@ -113,8 +115,9 @@ pub fn get_executable_file_from_path(
     package_dir: &PathBuf,
     command: Option<&str>,
 ) -> Result<(wapm_toml::Manifest, PathBuf), anyhow::Error> {
-    let wapm_toml = std::fs::read_to_string(package_dir.join("wapm.toml"))
-        .map_err(|_| anyhow::anyhow!("Package {package_dir:?} has no wapm.toml"))?;
+    let wapm_toml = std::fs::read_to_string(package_dir.join(GLOBAL_CONFIG_FILE_NAME))
```
syrusakbary, line 118: Let's name this wasmer_toml.

lib/cli/Cargo.toml
```diff
@@ -62,7 +63,7 @@ dirs = { version = "4.0" }
 serde_json = { version = "1.0" }
 target-lexicon = { version = "0.12", features = ["std"] }
 prettytable-rs = "0.9.0"
-wapm-toml = "0.2.0"
+wapm-toml = "0.4.0"
```
syrusakbary, line 66: Can we rename this to wasmer-toml ?

lib/registry/src/config.rs
```diff
@@ -4,6 +4,8 @@ use serde::Serialize;
 use std::collections::BTreeMap;
 use std::path::{Path, PathBuf};
 
+pub static GLOBAL_CONFIG_DATABASE_FILE_NAME: &str = "wapm.sqlite";
+
 #[derive(Deserialize, Default, Serialize, Debug, PartialEq, Eq)]
 pub struct PartialWapmConfig {
```
syrusakbary, line 10: This should be named Wasmer config isn't it?

tests/integration/cli/tests/fixtures/init3.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "wasmer"
+version = "0.5.6"
+description = "hello"
+
+[dependencies]
+
+[package.metadata.wapm]
```
syrusakbary, line 8: This file shouldn't exist is a `cargo wapm` one which we agreed not to have in this PR

tests/integration/cli/tests/fixtures/init2.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "wasmer"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
```
syrusakbary, line 6: This file shouldn't exist is a `cargo wapm` one which we agreed not to have in this PR

tests/integration/cli/tests/fixtures/init5.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "test-wasmer-init"
+version = "RANDOMVERSION1.RANDOMVERSION2.RANDOMVERSION3"
+edition = "2021"
+description = "description of package test-wasmer-init"
+
+[dependencies]
```
syrusakbary, line 7: This file shouldn't exist is a `cargo wapm` one which we agreed not to have in this PR

lib/registry/src/lib.rs
fschutt: This is for putting it in the package. You told me to use wasmer.toml for uploading, but put wapm.toml in the .tar.gz (to be compatible with the current model). Not sure if this will break downloading / running packages.

tests/integration/cli/tests/fixtures/init2.toml
syrusakbary: Note: unless is used for testing the `cargo.toml` naming parsing into `wasmer init`

tests/integration/cli/tests/fixtures/init3.toml
syrusakbary: Note: unless is used for testing the `cargo.toml` naming parsing into `wasmer init`

tests/integration/cli/tests/fixtures/init5.toml
syrusakbary: Note: unless is used for testing the `cargo.toml` naming parsing into `wasmer init`

lib/registry/src/config.rs
fschutt: I mean yeah if you want to break compatibility with wapm (settings would not be shared between wapm and wasmer)

lib/registry/src/package.rs
```diff
@@ -1,8 +1,17 @@
 use crate::PartialWapmConfig;
+use regex::Regex;
 use std::path::PathBuf;
 use std::{fmt, str::FromStr};
 use url::Url;
 
+const REGEX_FULL: &str = r#"^([a-zA-Z0-9\-_]+)/([a-zA-Z0-9\-_]+)(@([a-zA-Z0-9\.\-_]+*))?$"#;
```
syrusakbary, line 7: REGEX_PACKAGE_WITH_VERSION is more accurate

lib/registry/src/package.rs
```diff
@@ -1,8 +1,17 @@
...
+const REGEX_PACKAGE: &str = r#"^([a-zA-Z0-9\-_]+)/([a-zA-Z0-9\-_]+)$"#;
```
syrusakbary, line 8: Why we need this regex?

fschutt: `REGEX_PACKAGE` was for validating the package name when uploading (otherwise `name = "foo/bar@version"` would have been valid). Since we now always do it on the backend, it's not needed anymore.

lib/registry/src/config.rs
syrusakbary: What do you mean? `PartialWapmConfig` can be called `PartialWasmerConfig`.

lib/registry/src/publish.rs
```diff
@@ -0,0 +1,243 @@
...
+    archived_data_size: u64,
+    quiet: bool,
+) -> Result<(), anyhow::Error> {
+    let registry = match registry.as_ref() {
+        Some(s) => format_graphql(s),
+        None => {
+            #[cfg(not(test))]
+            let config = PartialWapmConfig::from_file();
+            #[cfg(test)]
+            let config = PartialWapmConfig::from_file("publish");
```
syrusakbary, line 45: What is this?

lib/registry/src/lib.rs
fschutt: @syrusakbary this will break wasmer run, because all packages on wapm use `wapm.toml`. I can either break it, or prefer wasmer.toml and look for wapm.toml as a fallback.

syrusakbary: I meant the naming of the variable `wapm_toml` seems incorrect. We should name it `wasmer_toml`

> prefer wasmer.toml and look for wapm.toml as a fallback.

In any case, this seems like the right approach for the file name

lib/registry/src/package.rs
syrusakbary: Ok, lets remove it then

syrusakbary: Where REGEX_FULL is being used? Shouldn't `wasmer run` in current master user it? If so, why there is a new regex?

tests/integration/cli/tests/fixtures/init3.toml
fschutt: I removed `init3` and `init5`, we don't need them anymore.

lib/registry/src/publish.rs
fschutt: This is a problem with cargo test: whenever I want to test `PartialWapmConfig::from_file` in an integration test, I have to give it the ID of the currently running test. Otherwise, I will either accidentally modify the global config.toml or (using a tempdir), subsequent calls to `PartialWapmConfig::from_file` cannot see the changes I made.

So for testing, I create a `config.test_1.toml`, `config.test_2.toml` file, etc. So that if I do:

```rust
#[test]  
fn test_1() {
    let completely_new_config_1 = PartialWapmConfig::from_file("test_1");
}

#[test]  
fn test_2() {
    let completely_new_config_2 = PartialWapmConfig::from_file("test_2");
}
```

... these test can run in parallel without interfering with each other (tempdir does not work as tempdir gets dropped at the end of the function call). I talked with Michael about this, if you have a better solution, I'm open for suggestions.

tests/integration/cli/tests/fixtures/init2.toml
fschutt: I removed init3 and init5, we don't need them anymore.

tests/integration/cli/tests/fixtures/init5.toml
fschutt: I removed init3 and init5, we don't need them anymore.

lib/registry/src/config.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3432/commits/c29377f298aa017144658446db4530d7d5d1ec66

lib/registry/src/package.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3432/commits/8e865a241168661521ed252b247e42711c340c20

lib/registry/src/package.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3432/commits/8e865a241168661521ed252b247e42711c340c20

lib/registry/src/lib.rs
fschutt: Okay, done, I've also fixed `std::fs::read("wapm.toml")` to use a new function to prefer wasmer.toml and use wapm.toml as a fallback.

lib/cli/Cargo.toml
fschutt: We can after https://github.com/wasmerio/wapm-toml/pull/8 has been merged and we've published a new version.

lib/registry/src/publish.rs
syrusakbary: This should be improved. Please chat with @Michael-F-Bryan or @theduke so they can provide ideas how to

theduke: Two options:

* use an env var to override the config path
  That could run into issues though with concurrent tests, since env vars apply to the whole process
* Make the tested methods accept a config path/full config as an argument

Michael-F-Bryan: I would make `PartialWapmConfig::from_file()` always take a path that everything is calculated relative to (e.g. `wasmer_home: &Path`). That way, the `wasmer` CLI can pass in the `$WASMER_HOME` directory for loading the config, while a testing environment would use the `wasmer.toml` from a temporary directory.

My thinking is that nothing in `lib/registry/src/config.rs` should ever call `PartialWapmConfig::get_file_location()` or `PartialWapmConfig::get_folder()` because that essentially hard-codes the Wasmer home directory. I'd really like to avoid using `#[cfg(test)]` anywhere except in a `#[cfg(test)] mod tests { ... }`.

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,517 @@
...
+
+    let metadata = metadata.exec();
+
+    let metadata = match metadata {
+        Ok(o) => o,
+        Err(e) => {
+            return Err(anyhow::anyhow!("failed to load metadata: {e}")
+                .context(anyhow::anyhow!("{}", manifest_path.display())));
+        }
+    };
```
Michael-F-Bryan, line 494:
```suggestion
    let metadata = metadata.exec()
        .with_context(|| format!("Unable to load metadata from \"{}\"", manifest_path.display()))?;
```

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,652 @@
...
+            validate::validate_directory(&manifest, &registry, cwd.clone())?;
+        }
+
+        builder.append_path_with_name(&manifest_path_buf, PACKAGE_TOML_FALLBACK_NAME)?;
+
+        let manifest_string = toml::to_string(&manifest)?;
+
+        let (readme, license) = construct_tar_gz(&mut builder, &manifest, &cwd)?;
+
+        builder.finish().ok();
```
Michael-F-Bryan, line 107: If finishing the builder fails here, should we be bubbling errors up to the user?

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,652 @@
...
+            &manifest_string,
+            &license,
+            &readme,
+            &archive_name,
+            &archive_path,
+            &maybe_signature_data,
+            archived_data_size,
+            self.quiet,
+        )
+        .map_err(on_error)
```
Michael-F-Bryan, line 155: Is it worth hoisting the `on_error()` cal into the main `cli` module so we only need to report errors to sentry once, rather than  needing to do it for every sub-command?

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,652 @@
...
+            Some(&minisign::PublicKey::from_base64(
+                &personal_key.public_key_value,
+            )?),
+            &private_key,
+            compressed_archive,
+            None,
+            None,
+        )?
+        .to_string()),
+    })
```
Michael-F-Bryan, line 313: Let's split this into separate statements so it's easier to see where we're generating the signature.

```suggestion
    let public_key = minisign::PublicKey::from_base64(
        &personal_key.public_key_value,
    )?;
    let signature = minisign::sign(
            Some(&public_key),
            &private_key,
            compressed_archive,
            None,
            None,
    )?;
    Ok(SignArchiveResult::Ok {
        public_key_id: personal_key.public_key_id,
        signature: signature.to_string(),
    })
```

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,652 @@
...
+use std::fs;
+use std::io::{self, Write};
+use std::path::{Path, PathBuf};
+use tar::Builder;
+use thiserror::Error;
+use time::{self, OffsetDateTime};
+use wasmer_registry::publish::SignArchiveResult;
+use wasmer_registry::{PartialWapmConfig, PACKAGE_TOML_FALLBACK_NAME};
+
+const CURRENT_DATA_VERSION: i32 = 3;
```
Michael-F-Bryan, line 14: We should probably write a test to make sure `CURRENT_DATA_VERSION` is in sync with the number of available migrations.

lib/cli/src/commands/publish.rs
```diff
@@ -0,0 +1,652 @@
...
+fn apply_migration(conn: &mut Connection, migration_number: i32) -> Result<(), MigrationError> {
+    let tx = conn
+        .transaction_with_behavior(TransactionBehavior::Immediate)
+        .map_err(|e| MigrationError::TransactionFailed(migration_number, format!("{}", e)))?;
+
+    let migrations = &[
+        (0, include_str!("../../sql/migrations/0000.sql")),
+        (1, include_str!("../../sql/migrations/0001.sql")),
+        (2, include_str!("../../sql/migrations/0002.sql")),
+    ];
```
Michael-F-Bryan, line 364: If we hoist this into its own `const` variable at the top of the file, `CURRENT_DATA_VERSION` just becomes `MIGRATIONS.len()`.

Michael-F-Bryan (changes requested):

lib/registry/src/publish.rs
fschutt: Okay, I migrated to `wasmer_dir: Path` instead of using `test_name`

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,517 @@
...
+                    .unwrap_or_else(|| std::env::current_dir().unwrap())
+                    .join("Cargo.toml");
+                cargo_toml_path
+                    .canonicalize()
+                    .unwrap_or_else(|_| cargo_toml_path.clone())
+            }
+        };
+
+        let cargo_toml = if manifest_path.exists() {
+            parse_cargo_toml(&manifest_path).ok()
```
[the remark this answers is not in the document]
fschutt: No, I think I didn't want to error out, wasmer init should error out as little as possible, the output can be "wrong" since it's just an init command

lib/cli/src/commands/init.rs
```diff
@@ -0,0 +1,517 @@
...
+                            .ok()?
+                            .file_stem()
+                            .and_then(|s| s.to_str())
+                            .map(|s| s.to_string())
+                    })
+                    .ok_or_else(|| anyhow::anyhow!("no current dir name"))?;
+                Ok((package_name, current_dir.join(WASMER_TOML_NAME)))
+            }
+            Some(s) => {
+                let _ = std::fs::create_dir_all(s);
```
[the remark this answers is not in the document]
fschutt: We should return an error here. Done in https://github.com/wasmerio/wasmer/pull/3432/commits/f9bc2d916f041ea8abe3019d2174f0d5b9f7ebd0

lib/cli/src/commands/publish.rs
fschutt: Yes, done in https://github.com/wasmerio/wasmer/pull/3432/commits/96f14164af80f1f0b078488f8b6c839508df4d18

lib/cli/src/commands/publish.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3432/commits/e9ceff9e1620aa33c0f1327c0ddfb9af4440b3e5

lib/cli/src/commands/publish.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3432/commits/e9ceff9e1620aa33c0f1327c0ddfb9af4440b3e5

lib/cli/src/commands/publish.rs
fschutt: We could although I'm not sure we'd want to report every failure of every subcommand to sentry. I think this is fine, we just want to see if the publishing fails (likely because of wapm having an error). I don't think we want to log every failure of every subcommand, in fact this is the only subcommand where this logging is implemented.

Michael-F-Bryan: Sounds reasonable to me. I imagine we'd get spammed quite a lot if we were reporting to sentry whenever a package lookup fails or we can't run a `*.wasm` file for whatever reason.

Michael-F-Bryan (approved):

Issue #3373, opened 2022-11-24
FreeBSD: wasmer hangs after running WASM file


MalloryA:
wasmer executes WASM files correctly, but after execution it hangs for around a minute before exiting.

```sh
% uname -a
FreeBSD athena 13.1-RELEASE-p3 FreeBSD 13.1-RELEASE-p3 GENERIC amd64
% echo "`./target/release/wasmer -V` | `rustc -V` | `uname -m`" 
wasmer 3.0.1 | rustc 1.63.0 (4b91a6ea7 2022-08-08) | amd64
% git show --oneline
a7defaf11a (HEAD -> master, origin/staging, origin/master, origin/HEAD) Merge #3365
```

### Steps to reproduce

(gmake is GNU Make from package `gmake`, rust is installed via rustup and not from the FreeBSD `rust` package)

```sh
% gmake build-wasmer
% fetch https://registry-cdn.wapm.io/contents/syrusakbary/cowsay/0.2.0/target/wasm32-wasi/release/cowsay.wasm
% time ./target/release/wasmer cowsay.wasm -- hello world          
```

### Expected behavior

The cow says "hello world" and the program immediately exits. It should take a fraction of a second (and it does when I run the same command on macOS).

### Actual behavior

The cow says "hello world" within a second, then the process hangs for 55 seconds. The fact that `time` reports `100% cpu` is unusual too.

```
% time ./target/release/wasmer cowsay.wasm -- hello world          
 _____________
< hello world >
 -------------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
               ||----w |
                ||     ||
./target/release/wasmer cowsay.wasm -- hello world  55.54s user 0.02s system 100% cpu 55.557 total
```

### Additional context

- wasmer can be compiled on FreeBSD/amd64 as of ~30 minutes ago: #3365

Michael-F-Bryan:
Thanks for reporting this @MalloryA. It might be something to do with our `WasiPipe` implementation reading stdout/stdin in a busy loop or pipes not being closed properly.

We're happy to review pull requests if someone finds a solution, but we don't currently have bandwidth to actively support BSD.

PR #3439, opened 2022-12-20
Use GNU/Linux frame registration code for FreeBSD too

MalloryA:
# Description

This change prevents a long hang when a program exits.

FreeBSD's `__deregister_frame` and `__register_frame` implementations are ported from GCC, which is why we should be using the same frame registration code for FreeBSD and GNU/Linux.

Link: https://github.com/freebsd/freebsd-src/blob/release/13.1.0/contrib/llvm-project/libunwind/src/UnwindLevel1-gcc-ext.c#L241-L299
Fixes: https://github.com/wasmerio/wasmer/issues/3373

# Review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Looking for changes that affect our C API? See the [C API Changelog](lib/c-api/C
 
 ## **Unreleased**
 
+## Fixed
+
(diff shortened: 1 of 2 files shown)


syrusakbary (approved):
Looks great. Merging!

PR #3448, opened 2022-12-30
Fix make install-capi-lib

### LINKED ISSUE #3445: Shared library doesn't have correct path
When compiling wasmer like this:
```
make WASMER_INSTALL_PREFIX=/usr ENABLE_LLVM=1
```
and then installing it like this:
```
DESTDIR=/tmp make install
```
I'll get a file tree such as:
```
/tmp/wasmer
├── bin
│   └── wasmer
├── include
│   ├── wasmer.h
│   ├── wasmer-README.md
│   ├── wasmer_wasm.h
│   └── wasm.h
├── lib
│   ├── libwasmer.so -> libwasmer.so.file:///home/svenstaro/src/wasmer/lib/c-api#wasmer-c-api@3.1.0
│   └── libwasmer.so.file:
│       └── home
│           └── svenstaro
│               └── src
│                   └── wasmer
│                       └── lib
│                           ├── c-api#wasmer-c-api@3 -> libwasmer.so.file:///home/svenstaro/src/wasmer/lib/c-api#wasmer-c-api@3.1.0
│                           ├── c-api#wasmer-c-api@3.1 -> libwasmer.so.file:///home/svenstaro/src/wasmer/lib/c-api#wasmer-c-api@3.1.0
│                           └── c-api#wasmer-c-api@3.1.0
```
The shared library now has an unexpected path and its symlink doesn't even work as it's not aware of the `DESTDIR`.

svenstaro:
# Description
Fixes #3445. The sed command was more complicated than it needed to be and it looked for the wrong character. I'm not sure why it broke, maybe cargo changed the output slightly?

# Review


NickCao:
I think https://github.com/wasmerio/wasmer/pull/3380 is a better solution, without resorting to regex.

svenstaro:
I agree but it needs jq installed which is an unexpected dependency that I didn't want to introduce. I thought fixing the current solution would make it easier for upstream to accept the contribution until something better would come along. However, I'm happy to view my PR as deprecated compared to yours.

svenstaro:
Can we merge this or the solution by @NickCao? Currently it's still broken.

syrusakbary:
Merging this once all tests pass @svenstaro

syrusakbary (approved):

Issue #3461, opened 2023-01-05
panic (left != right) left: XMM3,  right: XMM3'


AurelienFT:
### Describe the bug

I run a simple instance of WASM compiled using singlepass using the following features : 

```rust
    const FEATURES: Features = Features {
        threads: false,
        reference_types: false,
        simd: false,           
        bulk_memory: true,     
        multi_value: false,   
        tail_call: false,      
        module_linking: false, 
        multi_memory: false,   
        memory64: false,      
        exceptions: false,
        relaxed_simd: false,
        extended_const: false,
    };
```

### Steps to reproduce

I'm having an error on a VPS where I have this information about the CPU : 
```
***@***.***:~# cat /proc/cpuinfo
processor       : 0
vendor_id       : AuthenticAMD
cpu family      : 6
model           : 13
model name      : QEMU Virtual CPU version 2.5+
stepping        : 3
microcode       : 0x1000065
cpu MHz         : 1996.230
cache size      : 512 KB
physical id     : 0
siblings        : 8
core id         : 0
cpu cores       : 8
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 13
wp              : yes
flags           : fpu de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ht syscall nx lm nopl cpuid tsc_known_freq pni pclmulqdq ssse3 cx16 sse4_1 sse4_2 x2apic popcnt aes xsave rdrand hypervisor lahf_lm cmp_legacy abm 3dnowprefetch ssbd ibpb vmmcall arch_capabilities
bugs            : fxsave_leak sysret_ss_attrs null_seg spectre_v1 spectre_v2 spec_store_bypass
bogomips        : 3992.46
TLB size        : 1024 4K pages
clflush size    : 64
cache_alignment : 64
address sizes   : 40 bits physical, 48 bits virtual
...
```

### Expected behavior
Run without error

### Actual behavior
```rust
thread '' panicked at 'assertion failed: (left != right)
 left: XMM3,
 right: XMM3', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-compiler-singlepass-3.1.0/src/emitter_x64.rs:1752:40
```

### Additional context
I'm sorry I can't provide source that I'm using as it's part of a big project. I assume it's an error with the CPU architecture but if you can give me details it could be great.

[...]

PR #3465, opened 2023-01-10
Remove assert in sse_round_fn and handle case where src2 is in memory

sydhds:
In Singlepass SSE4.2 backend and for sse_round_fn function, some special cases were not handled:
* src1 != src2
* src2 is in memory

This should fix those cases.

This is a follow up of ticket https://github.com/wasmerio/wasmer/issues/3461


lib/compiler-singlepass/src/emitter_x64.rs
```diff
@@ -909,11 +909,12 @@ macro_rules! sse_round_fn {
         |emitter: &mut AssemblerX64, precision: Precision, src1: XMM, src2: XMMOrMemory, dst: XMM| {
             match src2 {
                 XMMOrMemory::XMM(x) => {
-                    assert_eq!(src1, x);
-                    move_src_to_dst(emitter, precision, src1, dst);
+                    move_src_to_dst(emitter, precision, x, dst);
```
ptitSeb, line 912: No like this.
You need to create an `if src1 == x` with that new code, and have the old code in the `else` branch.

ptitSeb (changes requested):
Please create a `if` /  `else` branch

ptitSeb (changes requested):
So, I reviewed the ROUNDSS/VROUNDSS opcode (for example, it's one of the opcode use here): https://www.felixcloutier.com/x86/roundss
And it's not what I imagined.
So the code should be:
```rust
if x != dst {
 move_src_to_dst(emitter, precision, src1, dst);
}
dynasm!(emitter ; $ins Rx((x as u8)), Rx((dst as u8)), $mode)
```
Basicaly a bigger change, but that version should be exact. It's not important for now, as ROUND functions are basically used as 2 operators opcode, never as 3-way in Singlepass.

ptitSeb (approved):

syrusakbary (approved):
It should already be approved. But approving again to see if that trigger tests

syrusakbary:
Merged manually, as I had no idea why tests are not being executed

theduke:
@syrusakbary Github Actions have been problematic the last two days, also on other repos.
Timeouts, network errors, jobs just sitting idle forever, ...

PR #3467, opened 2023-01-10
Fix wasmer-wasi-js compilation

syrusakbary:
Fix wasmer-wasi-js compilation

### DIFF
--- a/lib/derive/Cargo.toml
+++ b/lib/derive/Cargo.toml
@@ -17,5 +17,5 @@ proc-macro2 = "1"
 proc-macro-error = "1.0.0"
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


lib/derive/Cargo.toml
```diff
@@ -17,5 +17,5 @@ proc-macro2 = "1"
 proc-macro-error = "1.0.0"
 
 [dev-dependencies]
-wasmer = { path = "../api" }
+wasmer_types = { path = "../types", default-features=false }
```
syrusakbary, line 20:
```suggestion
wasmer-types = { path = "../types", default-features=false }
```

fschutt (approved):

PR #3468, opened 2023-01-10
Move wasi-types into wasi::types

fschutt:


theduke:
@fschutt what is/was the motiviation behind this one?

PR #3471, opened 2023-01-10
Rename `WasiState::new()` to `WasiState::builder()`

Michael-F-Bryan:
The typical convention is for `Foo::new()` to return an instance of `Foo` and `Foo::builder()` to return some type of builder object. Our `WasiState` type doesn't follow this convention (and explicitly `#[allow]`s the associated clippy warning), so the API can be confusing for anyone that hasn't used it before.

I've renamed the `WasiState::new()` function to `WasiState::builder()` and added a `#[deprecated]` attribute to the original `WasiState::new()` so it doesn't break Deploy.

### DIFF
--- a/lib/wasi/src/state/mod.rs
+++ b/lib/wasi/src/state/mod.rs
@@ -264,7 +264,14 @@ impl WasiState {
     /// Create a [`WasiStateBuilder`] to construct a validated instance of
     /// [`WasiState`].
     #[allow(clippy::new_ret_no_self)]
+    #[deprecated = "Use WasiState::builder()"]
     pub fn new(program_name: impl AsRef<str>) -> WasiStateBuilder {
+        WasiState::builder(program_name)
+    }
+
+    /// Create a [`WasiStateBuilder`] to construct a validated instance of
+    /// [`WasiState`].
+    pub fn builder(program_name: impl AsRef<str>) -> WasiStateBuilder {
         create_wasi_state(program_name.as_ref())
     }
 


lib/wasi/src/state/mod.rs
```diff
@@ -264,7 +264,14 @@ impl WasiState {
     /// Create a [`WasiStateBuilder`] to construct a validated instance of
     /// [`WasiState`].
     #[allow(clippy::new_ret_no_self)]
+    #[deprecated = "Use WasiState::builder()"]
```
theduke, line 267:
```suggestion
    #[deprecated(note = "Use WasiState::builder() instead", since = "3.2.0"]
```

lib/wasi/src/state/mod.rs
```diff
@@ -264,7 +264,14 @@ impl WasiState {
...
     pub fn new(program_name: impl AsRef<str>) -> WasiStateBuilder {
+        WasiState::builder(program_name)
+    }
+
+    /// Create a [`WasiStateBuilder`] to construct a validated instance of
+    /// [`WasiState`].
+    pub fn builder(program_name: impl AsRef<str>) -> WasiStateBuilder {
         create_wasi_state(program_name.as_ref())
```
theduke, line 275: While you are it it, can you do: `create_wasi_state()` -> `WasiStateBuilder::new()`?

theduke (changes requested):
Small additional change, then good to go.

theduke (approved):
LGTM

Do we have a solution for the nasty CI issues?

Michael-F-Bryan:
@theduke it looks like the CI issues are coming from the `wasix` branch itself and not related to this PR.

PR #3473, opened 2023-01-11
Fix wasm publish validation

syrusakbary:
Follow up from this PR: https://github.com/wasmerio/wapm-cli/pull/305

### DIFF
--- a/lib/cli/src/commands/publish.rs
+++ b/lib/cli/src/commands/publish.rs
@@ -640,7 +640,18 @@ mod validate {
         file_name: String,
     ) -> anyhow::Result<()> {
         use wasmparser::WasmDecoder;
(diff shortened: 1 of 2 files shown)


Michael-F-Bryan (approved):
Looks good to me :+1:

After talking with @syrusakbary this evening, I agree with the idea of enabling all WebAssembly features while validating. 

One option is to get the flags from `wasmer::Features::default()` instead of hard-coding them, but that would enforce validation using the current Wasmer defaults. We are actually trying to validate that each atom would be runnable by the end runner, though, and there is no way to know which features a runner would enable or disable (something that's essentially impossible to know).

Given that, I agree with being "optimistic" and allowing all WebAssembly features when validating packages before publishing - it's more of a sanity check, anyway.

PR #3479, opened 2023-01-16
This should fix CI build of CAPI Headless

ptitSeb:
# Description
Command lines arguments are not correctly built when the Makefile is used from the CI and build CAPI Headless.


Michael-F-Bryan (approved):
I wasn't able to run this locally because I don't have access to the `wasmer/aarch64` Docker image, but looking at the commands Make would be running, this should fix the issue on `master` :+1:

Michael-F-Bryan (approved):

Michael-F-Bryan (approved):

PR #3480, opened 2023-01-16
Ignore Create-exe with serialize test, something is wrong with the generated exe

ptitSeb:
Ignore Create-exe with serialize test, something is wrong with the generated exe the generated exe and needs further investigation.
But to not hold other more priority work, the tests are for now disabled to help the CI goes green again.

### DIFF
--- a/tests/integration/cli/tests/create_exe.rs
+++ b/tests/integration/cli/tests/create_exe.rs
@@ -483,7 +483,8 @@ fn create_exe_works_with_file() -> anyhow::Result<()> {
 
 // Ignored because of -lunwind linker issue on Windows
 // see https://github.com/wasmerio/wasmer/issues/3459
-#[cfg_attr(target_os = "windows", ignore)]
+//#[cfg_attr(target_os = "windows", ignore)]
+#[ignore]
 #[test]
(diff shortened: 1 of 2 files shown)


tests/integration/cli/tests/create_exe.rs
```diff
@@ -483,7 +483,8 @@ fn create_exe_works_with_file() -> anyhow::Result<()> {
 
 // Ignored because of -lunwind linker issue on Windows
 // see https://github.com/wasmerio/wasmer/issues/3459
-#[cfg_attr(target_os = "windows", ignore)]
```
theduke, line 486: Please add `TODO: ` comments here, and also create an issue (assigned to Felix) to fix this, otherwise we will definitely forget.

theduke (changes requested):

tests/integration/cli/tests/create_exe.rs
theduke: To clarify: TODO on both tests.

ptitSeb: Done

theduke (approved):
Still need to create the issue and reference the PR, but otherwise :+1:

PR #3483, opened 2023-01-17
Fix feature flags for make-build-wasmer-headless

fschutt:
Fixes master being red

### DIFF
--- a/lib/cli/Cargo.toml
+++ b/lib/cli/Cargo.toml
@@ -162,7 +162,7 @@ llvm = [
 debug = ["fern", "wasmer-wasi/logging"]
 disable-all-logging = ["wasmer-wasi/disable-all-logging"]
 headless = []
-headless-minimal = ["headless", "disable-all-logging", "wasi"]
+headless-minimal = ["headless", "disable-all-logging", "wasi", "compiler", "webc_runner", "static-artifact-create"]
 
 # Optional
 enable-serde = [
--- a/lib/wasi/Cargo.toml
+++ b/lib/wasi/Cargo.toml
@@ -50,7 +50,7 @@ tracing-wasm = "0.2"
 default = ["sys-default"]
 wasix = []
 
-webc_runner = ["webc", "serde_cbor", "anyhow", "serde", "wasmer/compiler", "wasmer/cranelift"]
+webc_runner = ["webc", "serde_cbor", "anyhow", "serde", "wasmer/compiler", "wasmer/cranelift", "wasmer-vfs/webc-fs"]
 webc_runner_rt_emscripten = ["wasmer-emscripten"]
 webc_runner_rt_wasi = []
 
--- a/lib/vfs/Cargo.toml
+++ b/lib/vfs/Cargo.toml
@@ -20,7 +20,7 @@ anyhow = { version = "1.0.66", optional = true }
 default = ["host-fs", "mem-fs", "webc-fs", "static-fs"]
 host-fs = ["libc"]
 mem-fs = ["slab"]
-webc-fs = ["webc", "anyhow"]
+webc-fs = ["webc", "mem-fs", "anyhow"]
 static-fs = ["webc", "anyhow", "mem-fs"]
 enable-serde = [
     "serde",


lib/cli/Cargo.toml
```diff
@@ -162,7 +162,7 @@ llvm = [
 debug = ["fern", "wasmer-wasi/logging"]
 disable-all-logging = ["wasmer-wasi/disable-all-logging"]
 headless = []
-headless-minimal = ["headless", "disable-all-logging", "wasi"]
+headless-minimal = ["headless", "disable-all-logging", "wasi", "compiler", "webc_runner", "static-artifact-create"]
```
ptitSeb, line 165: Is it really something we want, to have "headless-minimal" including "compiler". That doesn't sound minimal to me.

fschutt: It doesn't actually include any compiler, I'd personally like the `compiler` feature to be gone completely. If you look for `#[feature = "compiler"]` in the codebase, you can see that it's just controlling imports / exports.

fschutt: The `compiler` feature should be removed and always be enabled. There should only be feature flags for unnecessary dependencies and different targets. But otherwise, this can be replaced by `#[cfg(not(any(llvm, singlepass, cranelift)]` instead

ptitSeb: I don't think the "compiler" feature can be built for `wasm32-wasi` target

fschutt: It can, we currently have `--feature singlepass,cranelift` enabled on wasm32-wasi and both auto-enable the `compiler` feature

theduke: I think this meshes with the complaints I had about compilers not being available, @ptitSeb .

I'd also like the basic compiler APIs to always be available, even if the environment doesn't actually have a valid implementation. 

(I don't know the APIs well yet, so not aware of complications this would cause)

syrusakbary: We want to include compilers in `wasm32-wasi`, because that way we can do the cloud compilation in the future.

But that doesn't mean that we should require "compiler" for the headless minimal

syrusakbary:
```suggestion
headless-minimal = ["headless", "disable-all-logging", "wasi", "webc_runner", "static-artifact-create"]
```

lib/wasi/Cargo.toml
```diff
@@ -50,7 +50,7 @@ tracing-wasm = "0.2"
 default = ["sys-default"]
 wasix = []
 
-webc_runner = ["webc", "serde_cbor", "anyhow", "serde", "wasmer/compiler", "wasmer/cranelift"]
+webc_runner = ["webc", "serde_cbor", "anyhow", "serde", "wasmer/compiler", "wasmer/cranelift", "wasmer-vfs/webc-fs"]
```
syrusakbary, line 53: We should fix this. The runner should never require a compiler or cranelift to run.

```suggestion
webc_runner = ["webc", "serde_cbor", "anyhow", "serde", "wasmer-vfs/webc-fs"]
```

lib/wasi/Cargo.toml
```diff
@@ -50,9 +50,15 @@ tracing-wasm = "0.2"
...
-webc_runner = ["webc", "serde_cbor", "anyhow", "serde", "wasmer/compiler", "wasmer/cranelift"]
+webc_runner = ["webc", "serde_cbor", "anyhow", "serde", "wasmer-vfs/webc-fs"]
 webc_runner_rt_emscripten = ["wasmer-emscripten"]
 webc_runner_rt_wasi = []
+emscripten_cranelift = ["wasmer/cranelift"]
+wasi_cranelift = ["wasmer/cranelift"]
+emscripten_singlepass = ["wasmer/singlepass"]
+wasi_singlepass = ["wasmer/singlepass"]
+emscripten_llvm = ["wasmer/llvm"]
+wasi_llvm = ["wasmer/llvm"]
```
syrusakbary, line 61: This is not correct. Wasi should not depend on any compiler whatsoever

Cargo.toml
```diff
@@ -100,16 +100,16 @@ engine = ["universal"]
...
 compiler = [
     "wasmer/compiler",
     "wasmer-compiler/translator",
 ]
-singlepass = ["wasmer-compiler-singlepass", "compiler"]
-cranelift = ["wasmer-compiler-cranelift", "compiler"]
-llvm = ["wasmer-compiler-llvm", "compiler"]
+singlepass = ["wasmer-compiler-singlepass", "compiler", "wasmer-wasi/emscripten_singlepass", "wasmer-wasi/wasi_singlepass"]
+cranelift = ["wasmer-compiler-cranelift", "compiler", "wasmer-wasi/emscripten_cranelift", "wasmer-wasi/wasi_cranelift"]
+llvm = ["wasmer-compiler-llvm", "compiler", "wasmer-wasi/emscripten_llvm", "wasmer-wasi/wasi_llvm"]
```
syrusakbary, line 112: This changes are not correct

lib/wasi/src/runners/wasi.rs
```diff
@@ -42,8 +43,20 @@ impl crate::runners::Runner for WasiRunner {
...
+        #[cfg(feature = "wasi_singlepass")]
+        let mut store = Store::new(wasmer::Singlepass::default());
+        #[cfg(not(any(
+            feature = "wasi_cranelift",
+            feature = "wasi_llvm",
+            feature = "wasi_singlepass"
+        )))]
+        let mut store = {
+            return Err(anyhow::anyhow!("wasmer-wasi needs one of wasi_cranelift or wasi_llvm or wasi_singlepass features enabled").into());
+        };
```
syrusakbary, line 59: This is not correct.

```suggestion
        let mut store =  Store::default()
```

lib/c-api/Cargo.toml
```diff
@@ -78,14 +78,20 @@ compiler-headless = [
 ]
 singlepass = [
     "wasmer-compiler-singlepass",
+    "wasmer-wasi/emscripten_singlepass", 
+    "wasmer-wasi/wasi_singlepass",
```
syrusakbary, line 82: This is not correct

lib/c-api/Cargo.toml
```diff
@@ -78,14 +78,20 @@ compiler-headless = [
...
     "compiler",
 ]
 cranelift = [
     "wasmer-compiler-cranelift",
+    "wasmer-wasi/emscripten_cranelift", 
+    "wasmer-wasi/wasi_cranelift",
```
syrusakbary, line 88: This is not correct

lib/c-api/Cargo.toml
```diff
@@ -78,14 +78,20 @@ compiler-headless = [
...
     "compiler",
 ]
 llvm = [
     "wasmer-compiler-llvm",
+    "wasmer-wasi/emscripten_llvm", 
+    "wasmer-wasi/wasi_llvm",
```
syrusakbary, line 94: This is not correct

lib/wasi/src/runners/wasi.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/a483d047de82c89a295d3fbf7ce813dfdea1ed32

Cargo.toml
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/a483d047de82c89a295d3fbf7ce813dfdea1ed32

lib/c-api/Cargo.toml
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/a483d047de82c89a295d3fbf7ce813dfdea1ed32

lib/c-api/Cargo.toml
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/a483d047de82c89a295d3fbf7ce813dfdea1ed32

lib/c-api/Cargo.toml
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/a483d047de82c89a295d3fbf7ce813dfdea1ed32

lib/api/src/sys/store.rs
```diff
@@ -174,20 +174,19 @@ impl Default for Store {
...
+                                .engine()
+                        } else {
+                            EngineBuilder::headless()
+                                .engine()
+                        }
                     }
-            }
                 } else {
-                    compile_error!("No default engine chosen")
+                    compile_error!("No compiler [cranelift, llvm, singlepass] enabled, use Store::headless() instead of Store::default() for a headless engine")
```
syrusakbary, line 189: Revert to the previous text

lib/api/src/sys/store.rs
```diff
@@ -174,20 +174,19 @@ impl Default for Store {
...
+                    cfg_if::cfg_if! {
+                        if #[cfg(any(feature = "cranelift", feature = "llvm", feature = "singlepass"))]
+                        {
+                            let config = get_config();
+                            EngineBuilder::new(Box::new(config) as Box<dyn wasmer_compiler::CompilerConfig>)
+                                .engine()
+                        } else {
+                            EngineBuilder::headless()
+                                .engine()
+                        }
```
syrusakbary, line 186: I assume this is just formatting code?

lib/api/src/sys/store.rs
fschutt, line 189:
```suggestion
                    compile_error!("No default engine chosen")
```

fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/caeca9a7a82739152e4a9a4d724dc7ce1a080133

lib/cli/Cargo.toml
```diff
@@ -149,14 +149,17 @@ experimental-io-devices = [
 ]
 singlepass = [
     "wasmer-compiler-singlepass",
+    "wasmer/singlepass",
```
syrusakbary, line 152: This is not correct

lib/cli/Cargo.toml
```diff
@@ -149,14 +149,17 @@ experimental-io-devices = [
...
     "compiler",
 ]
 cranelift = [
     "wasmer-compiler-cranelift",
+    "wasmer/cranelift",
```
syrusakbary, line 157: This is not correct

lib/cli/Cargo.toml
```diff
@@ -149,14 +149,17 @@ experimental-io-devices = [
...
     "compiler",
 ]
 llvm = [
     "wasmer-compiler-llvm",
+    "wasmer/llvm",
```
syrusakbary, line 162: This is not correct

lib/cli/Cargo.toml
syrusakbary: Please undo

lib/cli/Cargo.toml
syrusakbary: Please undo

lib/cli/Cargo.toml
syrusakbary: Please undo

lib/cli/Cargo.toml
fschutt: If I remove this, it will error with "engine is not compiled in" when using `Store::default()`. This is the only way to pass the "singlepass" feature to the `api` crate, which is necessary for code that uses `#[cfg(feature = "singlepass")]`.

Otherwise `Store::default()` will error out. Not sure what the problem is from a dependency standpoint, if we want singlepass to be enabled in the CLI, it should obviously also be enabled for code that uses `Store::default()` (the dependencies are the same).

lib/cli/Cargo.toml
fschutt: Undone in https://github.com/wasmerio/wasmer/pull/3483/commits/8d3e7c0dadc638cec97265415a46c32f035ae073

lib/cli/Cargo.toml
fschutt: Undone in https://github.com/wasmerio/wasmer/pull/3483/commits/8d3e7c0dadc638cec97265415a46c32f035ae073

ptitSeb:
I don't understand why you needed all those lint adjustments. `make lint` works fine localy on my side, none of the adjustments are needed. I suspect you are using updated clippy or something like that localy @fschutt to have those needs.

lib/cli/src/commands/create_exe.rs
```diff
@@ -223,14 +224,31 @@ impl CreateExe {
...
+        } else if self.compiler.llvm {
+            CompilerType::LLVM
+        } else {
+            ct // default compiler type
+        };
+
+        let engine = EngineBuilder::new(get_config(&compiler_type)?).engine();
+
+        let tunables = BaseTunables::for_target(engine.target());
+        let store = Store::new_with_tunables(&engine, tunables);
```
syrusakbary, line 242: I'm a bit confused... I think the store returned by `get_store_for_target` already does this?

Am I missing something? Have you tried doing this?
```suggestion
        let (store, ct) = self.compiler.get_store_for_target(target.clone())?;
```

And if you did, and it's different... can you explore on why your own-constructed `Store` is different from the one returned by `get_store_for_target` ?

lib/cli/src/commands/create_exe.rs
```diff
@@ -300,6 +318,52 @@ impl CreateExe {
     }
 }
 
+fn get_config(
```
syrusakbary, line 321: This function is duplicating logic that should not be needed

syrusakbary (changes requested):
Address the comments

lib/cli/src/commands/create_exe.rs
fschutt: @syrusakbary the store returned by `get_store_for_target` emits the wrong `UnwindInfo` type on Windows (windows-native, but we need DWARF unwind info, even on Windows). I need to do `Cranelift::default()`, `Singlepass::default()`, etc. instead. The `get_store_for_target` doesn't work for `create-exe`.

lib/cli/src/commands/create_exe.rs
fschutt: It is needed, but I could move it somewhere else so it's reusable.

lib/cli/src/commands/create_exe.rs
syrusakbary: > the store returned by `get_store_for_target` emits the wrong UnwindInfo type on Windows (windows-native, but we need DWARF unwind info, even on Windows)

I don't think this is true. I believe you were using `Store::default()` somewhere else and that caused a mismatch when using the store generated by `get_store_for_target`.
In any case, Can you explain what's the difference between your code and `get_store_for_target`? Basically, and following your assumption, I'd like to know why `get_store_for_target` is generating wrong unwind info and your code is not

fschutt:
The bug disappears if I set the `Engine::new(compiler_config, target, features)` in `lib/compiler/src/engine/builder.rs:65` to change the `target.operating_system = OperatingSystem::Darwin` instead of `OperatingSystem::Windows` (alternative, removing `EngineBuilder::new().set_target(Some(...))` also works.

fschutt:
I've filed a bug in https://github.com/wasmerio/wasmer/issues/3508 and worked around this problem for now.

lib/cli/src/commands/create_exe.rs
fschutt: Removed in https://github.com/wasmerio/wasmer/pull/3483/commits/c61a5d7cf5891bcd3e8e2ea6c5535d3ebec2eeb5

lib/wasi/src/runners/emscripten.rs
```diff
@@ -1,24 +1,46 @@
...
-#[derive(Debug, Default, Clone, PartialEq, Eq, PartialOrd, Hash, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
 pub struct EmscriptenRunner {
     args: Vec<String>,
+    #[serde(skip, default)]
+    store: Store,
 }
 
 impl EmscriptenRunner {
+    /// Constructs a new `EmscriptenRunner` given an `Engine`
```
syrusakbary, line 24:
```suggestion
    /// Constructs a new `EmscriptenRunner` given an `Store`
```

lib/compiler/src/engine/artifact.rs
```diff
@@ -79,7 +90,18 @@ impl Artifact {
...
+    /// Compile a data buffer into a `ArtifactBuild`, which may then be instantiated.
+    #[cfg(feature = "compiler")]
+    pub fn get_module_info(
+        engine: &Engine,
+        data: &[u8],
+        tunables: &dyn Tunables,
+    ) -> Result<ModuleInfo, CompileError> {
+        let artifact = Self::get_artifact_build(engine, data, tunables)?;
+        Ok(artifact.create_module_info())
     }
```
syrusakbary, line 105: Why don't we move this into the `Engine`, so we have in the `Engine` impl:

```rust
    #[cfg(feature = "compiler")]
    pub fn get_module_info(
        data: &[u8],
    ) -> Result<ModuleInfo, CompileError> {
        // this is from `artifact_builder.rs`
        let environ = ModuleEnvironment::new();
        let translation = environ.translate(data).map_err(CompileError::Wasm)?;
        return translation.module;
    }
```

lib/compiler/src/engine/artifact.rs
```diff
@@ -56,6 +56,17 @@ impl Artifact {
...
+        let mut inner_engine = engine.inner_mut();
+        Self::from_parts(&mut inner_engine, artifact)
+    }
+
+    #[cfg(feature = "compiler")]
+    fn get_artifact_build(
+        engine: &Engine,
+        data: &[u8],
+        tunables: &dyn Tunables,
+    ) -> Result<ArtifactBuild, CompileError> {
```
syrusakbary, line 69: We can undo this

lib/cli/src/commands/create_exe.rs
```diff
@@ -961,22 +964,25 @@ fn get_module_infos(
...
 ) -> Result<BTreeMap<String, ModuleInfo>, anyhow::Error> {
     let mut entrypoint =
         get_entrypoint(directory).with_context(|| anyhow::anyhow!("get module infos"))?;
 
     let mut module_infos = BTreeMap::new();
     for (atom_name, atom_bytes) in atoms {
-        let store = Store::default();
-        let module = Module::from_binary(&store, atom_bytes.as_slice())
-            .map_err(|e| anyhow::anyhow!("could not deserialize module {atom_name}: {e}"))?;
+        let tunables = BaseTunables::for_target(store.engine().target());
```
syrusakbary, line 974: You don't need this

lib/cli/src/commands/create_exe.rs
```diff
@@ -961,22 +964,25 @@ fn get_module_infos(
...
+        let module_info =
+            Artifact::get_module_info(store.engine(), atom_bytes.as_slice(), &tunables)
```
syrusakbary, line 976:
```suggestion
        let module_info = Engine::get_module_info(atom_bytes.as_slice())
```

lib/cli/src/store.rs
```diff
@@ -313,7 +314,8 @@ impl StoreOptions {
     pub fn get_store_for_target(&self, target: Target) -> Result<(Store, CompilerType)> {
         let (compiler_config, compiler_type) = self.compiler.get_compiler_config()?;
         let engine = self.get_engine_with_compiler(target, compiler_config)?;
-        let store = Store::new(engine);
+        let tunables = BaseTunables::for_target(engine.target());
+        let store = Store::new_with_tunables(&engine, tunables);
```
syrusakbary, line 318: We should undo this

lib/wasi/src/runners/emscripten.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483

lib/compiler/src/engine/artifact.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/b26b48794cc7b00bb5f939dfdf40bba94e9ba86c

lib/compiler/src/engine/artifact.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/b26b48794cc7b00bb5f939dfdf40bba94e9ba86c

lib/cli/src/commands/create_exe.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/b26b48794cc7b00bb5f939dfdf40bba94e9ba86c

lib/cli/src/commands/create_exe.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/b26b48794cc7b00bb5f939dfdf40bba94e9ba86c

lib/cli/src/store.rs
```diff
@@ -105,7 +105,8 @@ impl CompilerOptions {
     pub fn get_store_for_target(&self, target: Target) -> Result<(Store, CompilerType)> {
         let (compiler_config, compiler_type) = self.get_compiler_config()?;
         let engine = self.get_engine(target, compiler_config)?;
-        let store = Store::new(engine);
+        let tunables = BaseTunables::for_target(engine.target());
+        let store = Store::new_with_tunables(&engine, tunables);
```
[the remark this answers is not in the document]
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/54f0f0cbe9aa4eea1afaf2bfd29b7053b2b51325

lib/cli/src/store.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/54f0f0cbe9aa4eea1afaf2bfd29b7053b2b51325

lib/cli/src/store.rs
```diff
@@ -30,15 +30,15 @@ pub struct StoreOptions {
 pub struct CompilerOptions {
     /// Use Singlepass compiler.
     #[clap(long, conflicts_with_all = &["cranelift", "llvm"])]
-    singlepass: bool,
+    pub singlepass: bool,
```
syrusakbary, line 33: This should not be pub

lib/cli/src/store.rs
```diff
@@ -30,15 +30,15 @@ pub struct StoreOptions {
...
 
     /// Use Cranelift compiler.
     #[clap(long, conflicts_with_all = &["singlepass", "llvm"])]
-    cranelift: bool,
+    pub cranelift: bool,
```
syrusakbary, line 37: This should not be pub

lib/cli/src/store.rs
```diff
@@ -30,15 +30,15 @@ pub struct StoreOptions {
...
 
     /// Use LLVM compiler.
     #[clap(long, conflicts_with_all = &["singlepass", "cranelift"])]
-    llvm: bool,
+    pub llvm: bool,
```
syrusakbary, line 41: This should not be pub

syrusakbary (approved):
Once the last nits are fixed, we should be good to merge

lib/cli/src/store.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/ff11cbd7059c0b138067cefb644713b584e1777c

lib/cli/src/store.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/ff11cbd7059c0b138067cefb644713b584e1777c

lib/cli/src/store.rs
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3483/commits/ff11cbd7059c0b138067cefb644713b584e1777c

lib/cli/src/commands/create_exe.rs
fschutt: `Store::default()` used a mix between the host-native and the target flags / features.

PR #3491, opened 2023-01-18
Fix make lint not working consistently

### LINKED ISSUE #3457: Harmonise the build and test CI workflows
In order to release a new version, we currently still use the old build.yaml file, which has the same steps as in the test.yaml file, except for the last "release" step, which is missing from test.yaml. The build.yaml is a lot slower than the test.yaml (45 minutes vs 20 minutes).

We should merge them into one workflow and copy the release step to the test.yaml.

### LINKED ISSUE #3489: make lint does not properly catch CI failures when committing changes into master
For some reason we can have a green master, but when I make changes off of that I suddenly get lots of "make lint" errors that I have to then fix in the next PR, which makes those PRs harder to review. We need to debug why this happens and fix it.

fschutt:
Fixes #3489.
Fixes #3457.


fschutt:
Note to @syrusakbary: If this PR is correct, then it should show all the "make lint" errors here (so we can check that it actually works).

theduke (changes requested):
A)

This does not just fix the lints, it does a completely separate change that should be done in a separate PR.

The important part is fixing the linting, and doing all the related fixes that break when clippy actually runs.

B) 

The important part here is figuring out why the lints don't work properly with the cache, so we don't run into this again.

C)

We need to run clippy for feature and target combinations, not just for linux and with a subset of flags.

Clippy has `--all-targets` and `--all-features` flags that do this. I'm not sure if that also properly tests for various combinations of features, probably not, but it's at least a significant improvement.

That will probably produce a significant amount of errors to fix, though, so you should focus on that.

theduke:
Running `cargo clippy --all-targets --all-features` actually doesn't produce too many problems, just some in `benches`, which is probably unused.

fschutt:
Okay, it seems that the issue is that I use `rustc 1.66 nightly` locally which comes with `clippy 1.66`. That's why I get errors that don't show up in the CI. The only thing useful from this PR is that we now run cargo clippy on all operating systems instead of just on Linux. I'll create a separate PR for refactoring the build.yaml / test.yaml / release steps.

fschutt:
Superseded by https://github.com/wasmerio/wasmer/pull/3499

Issue #3485, opened 2023-01-17
Memory-Leak when generating Instance


ubamrein:
### Describe the bug

I'm currently using wasmer to play arround with wasm-edgecomputing. For every request, I use the serialized bytes of a module to deserialize it, and execute the corresponding wasm file (I also tried the non serialized method before, but the memory leak was even worse).

The following code is the minimal amount of code I need, to reproduce a memory-leak:
```rust
let module_bytes = if let Some(wasm_module) = self.wasm_module.as_ref() {
    wasm_module // &Vec<u8>
} else {
    return Err("No wasm".into());
};

let mut store = Store::default();
let module = unsafe { Module::deserialize(&store, module_bytes.as_slice())? };
let instance = Instance::new(&mut store, &module, &Imports::default())?;
```

### Steps to reproduce
Execute the above code within a loop, as for example like this (note I use compressed wasm files but that should not matter in this context as it is only done once):

```rust 
use std::{thread::sleep, time::Duration};

use flate2::read::GzDecoder;
use wasmer::{Imports, Instance, Module, Store};

fn main() {
    env_logger::init();
    let m: Vec<u8> = {
        let wasm_file = include_bytes!("../example.gz");
        let store = Store::default();

        let wasm_file = decompress(wasm_file);

        let module = if let Ok(module) = Module::from_binary(&store, &wasm_file) {
            module
        } else {
            return;
        };
        module.serialize().unwrap().into()
    };

    loop {
        let mut store = Store::default();
        let module = unsafe { Module::deserialize(&store, m.as_slice()).unwrap() };
        let instance = Instance::new(&mut store, &module, &Imports::default()).unwrap();
        sleep(Duration::from_millis(1));
    }
}

fn decompress(archive: &[u8]) -> Vec<u8> {
    use std::io::prelude::*;
    let mut e = GzDecoder::new(archive);
    let mut buf = vec![];
    e.read_to_end(&mut buf).unwrap();
    buf
}
```

### Expected behavior

There should be no memory-leak, as no wasm code gets executed (possible leaks there) and there is literally nothing, except for the wasm file, being loaded.

### Actual behavior

You can verify that the memory usage grows and grows until it is OOM killed (which in the example above can take some time :)).

Leo-Besancon:
I have the same issue (taking the example given here: https://docs.rs/wasmer/latest/wasmer/#usage).

Valgrind shows the following output:
```
[...]

PR #3493, opened 2023-01-18
Fixed some memory leak issue with InstanceHandle

ptitSeb:
This is help with #3485 
This memory leak is also tracked there: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55183


syrusakbary (changes requested):
Rename: `InstanceHandle` to `VMInstance` and `write_instance` to `into_vminstance`

theduke:
I guess you are aware, but that rename is a breaking change and would require a major version bump for the vm crate.

ptitSeb:
> I guess you are aware, but that rename is a breaking change and would require a major version bump for the vm crate.

I don't think `InstanceHandle` is only used Internaly. So we should be safe.

theduke:
It's in the public API of `-vm`, so at least that create would need a bump.

theduke:
It's also in the public api, because of `Artifact::instantiate()`.

ptitSeb:
This also fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=55183 when tested localy

syrusakbary (approved):
Looks good 👍

Issue #3492, opened 2023-01-18
Rework filesystem mapping and partial overlap of --mapdir directories


fschutt:
Currently when mapping directories in WASI with `.map_dir()`, the order of mapping is important, as `/lib/api/blah` could overwrite another mapping. Currently the order of mappings is important, but we should fix this in the next release or at least when we migrate / rework the `[fs]` mapping in the new Manifest format.

Also, duplicated mapping to `/` (root folder) aren't allowed, which makes dependency management impossible if two packages use the same folder name. (for example, having a dependency on python, which maps to `/ -> /usr/bin/python` and a dependency on ruby mapping `/ -> /usr/bin/ruby`). In the end `/ruby.wasm` and `/python.wasm` should both work, the only danger are conflicting file names (should be recognized and rejected when uploading).

ptitSeb:
Related to #3476

Michael-F-Bryan:
This is also related to #3678.

syrusakbary:
Closing this in favor of https://github.com/wasmerio/wasmer/issues/3678

PR #3494, opened 2023-01-18
Update wasmer-toml version

fschutt:
Fixes two bugs: one is that we cannot run packages with a `CommandV1` command, since the `module` name was never found, the other bug is that the `[fs]` section is currently non-deterministic with regards to how filesystem mappings overlap each other (mapping `/`, `/lib`, `/lib/a/b/c` leads to a situation where the `/lib/a/b/c` would shadow the `/` mapping).

I've tracked the FS issues in https://github.com/wasmerio/wasmer/issues/3492 since they appeared in the hackathon. For now it's okay to use `IndexMap` to work around this issue, but we need to track it and fix it properly in the future.


lib/cli/src/commands/run.rs
```diff
@@ -121,7 +120,7 @@ impl RunWithPathBuf {
 
             #[cfg(feature = "wasi")]
             {
-                let default = HashMap::default();
+                let default = indexmap::IndexMap::default();
```
theduke, line 123: For the history books: can you 

* explain here in the PR what problem this solves
* Add comments in the code why `IndexMap` is used
* Open an issue to discuss how we want overlapping file system semantics to work

fschutt: See https://github.com/wasmerio/wasmer/issues/3492 - we need IndexMap to have a stable ordering for the [fs] mapping, otherwise overlapping filesystem mappings might not work. For example, in the `wasmer_toml`, we need to map `/lib`, then `/`, the order is important so that the mappings are mounted in the correct order. If we use `HashMap` here, then the order of the mounts are not preserved as they are written in the `wasmer.toml` file.

fschutt: @theduke done

theduke:
See https://github.com/wasmerio/wasmer/issues/3492 for related discussion.

theduke (approved):
LGTM

Issue #3481, opened 2023-01-16
Create-exe with Serialized object fail


ptitSeb:
### Describe the bug

CI tests `create_exe_serialized_works`  and `create_exe_with_object_input_serialized` have been disable because they are not working.

The generated EXE is somehow incorect, and just segfault during some Rust's Stdlib initialization.

Isolating a binary and running it with gdb gives this output:

````
─── Output/messages ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7d1eff3 in __GI_getenv (name=0x16b0810 "RUST_BACKTRACE") at ./stdlib/getenv.c:51
51      ./stdlib/getenv.c: No such file or directory.
─── Assembly ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 0x00007ffff7d1efe1  __GI_getenv+33 movzbl (%rdi),%eax
 0x00007ffff7d1efe4  __GI_getenv+36 mov    %rdi,%r13
 0x00007ffff7d1efe7  __GI_getenv+39 test   %al,%al
 0x00007ffff7d1efe9  __GI_getenv+41 je     0x7ffff7d1f088 <__GI_getenv+200>
 0x00007ffff7d1efef  __GI_getenv+47 cmpb   $0x0,0x1(%rdi)
 0x00007ffff7d1eff3  __GI_getenv+51 mov    0x0(%rbp),%rbx
 0x00007ffff7d1eff7  __GI_getenv+55 jne    0x7ffff7d1f030 <__GI_getenv+112>
 0x00007ffff7d1eff9  __GI_getenv+57 or     $0x3d,%ah
 0x00007ffff7d1effc  __GI_getenv+60 test   %rbx,%rbx
 0x00007ffff7d1efff  __GI_getenv+63 jne    0x7ffff7d1f015 <__GI_getenv+85>
[...]

Issue #3459, opened 2023-01-03
Create-exe: Cross-compiling from Windows to Windows doesn't work


fschutt:
This is a followup for https://github.com/wasmerio/wasmer/pull/3430 - currently the `test_cross_compile_python_windows` and `test_wasmer_create_exe_pirita_works` don't work on windows when compiling from Windows to Windows:

```
x86_64-windows-gnu target cranelift                                                                                                                                                                                   
Error: linking failed with: stdout: Compiler: cranelift                                                                                                                                                               
Target: x86_64-unknown-windows-gnu                                                                                                                                                                                    
Format: Symbols                                                                                                                                                                                                       
Using path `C:\Users\felix\wasmer\package\cache\wasmer-windows-gnu64\lib\libwasmer.a` as libwasmer path.                                                                                                              
                                                                                                                                                                                                                      
                                                                                                                                                                                                                      
stderr: C:\ProgramData\chocolatey\lib\zig\tools\zig-windows-x86_64-0.10.0\zig.exe clang C:\Users\felix\AppData\Local\Temp\.tmpnAprtL\wasmer_main.c -fno-caret-diagnostics -target x86_64-unknown-windows-gnu -nostdinc
 -fno-spell-checking -isystem C:\ProgramData\chocolatey\lib\zig\tools\zig-windows-x86_64-0.10.0\lib\include -isystem C:\ProgramData\chocolatey\lib\zig\tools\zig-windows-x86_64-0.10.0\lib\libc\include\x86_64-windows
-gnu -isystem C:\ProgramData\chocolatey\lib\zig\tools\zig-windows-x86_64-0.10.0\lib\libc\include\generic-mingw -isystem C:\ProgramData\chocolatey\lib\zig\tools\zig-windows-x86_64-0.10.0\lib\libc\include\x86_64-wind
ows-any -isystem C:\ProgramData\chocolatey\lib\zig\tools\zig-windows-x86_64-0.10.0\lib\libc\include\any-windows-any -Xclang -target-cpu -Xclang x86-64 -Xclang -target-feature -Xclang -16bit-mode -Xclang -target-fea
ture -Xclang -32bit-mode -Xclang -target-feature -Xclang -3dnow -Xclang -target-feature -Xclang -3dnowa -Xclang -target-feature -Xclang +64bit -Xclang -target-feature -Xclang -adx -Xclang -target-feature -Xclang -a
[...]

PR #3487, opened 2023-01-17
Fix create exe tests

fschutt:
Fixes #3481.

PR #3496, opened 2023-01-18
Fixed create-exe tests for object-format serialized

fschutt:
Fixes https://github.com/wasmerio/wasmer/issues/3481.

Continuation of https://github.com/wasmerio/wasmer/pull/3487 without all the clippy changes.


theduke:
Closes #3459 .

theduke (approved):
For poterity: module_start was added for WASIX, but is not actually needed anymore, so it's fine to remove.

BUT: @fschutt we need to figure out why the create-exe step was failing, so we can prevent this in the future.

Did the compiled binaries somehow still use the old serialization format without the new field, which made loading artifacts fail?
We need to know the cause.

PR #3498, opened 2023-01-18
Wasix Control Plane - Thread Limit + Cleanup

theduke:
This commit adds a total task (aka thread) limit for a WasiControlPlane.

The limit is configured at startup, and is checked whenever a new process or a new thread is created.

The major breaking change is that the new_process and new_thread methods are now fallible and return an error.

Also adds two tests that validate the limits are repsected.

Along with this feature the commit also does some cleanup and reafactoring:

* Simplify WasiControlPlane

It was using multiple Arcs with locks and atomics with no real justification. The state is only locked for very short periods, so the complicated setup doesn't yield any benefits.

The mutable state is consolidated behind a single RwLock, which holds the process id seed and the process map, and the whole type is behind a single Arc.

The reservations are removed, since they are not really useful if the state is only locked for very short periods.

Also adds todo notes for handling pid exhaustion and process de-registration.

* TaskJoinHandle

Adds a new TaskJoinHandle handle that is used by both processes and threads to track termination.

Replaces a complex, cryptic type.

* Simplify Thread state and make it private

Make internal thread state private to better encapsulate the code. Also restructures the type to be behind a single Arc instead of multiple redundant Arcs.

NOTE: should also be done for WasiProcess

# Description

### DIFF
--- a/lib/wasi/src/os/console/mod.rs
+++ b/lib/wasi/src/os/console/mod.rs
@@ -18,9 +18,9 @@ use tokio::sync::{mpsc, RwLock};
 use tracing::{debug, error, info, trace, warn};
 #[cfg(feature = "sys")]
 use wasmer::Engine;
-use wasmer_vbus::{BusSpawnedProcess, SpawnOptionsConfig};
+use wasmer_vbus::{BusSpawnedProcess, SpawnOptionsConfig, VirtualBusError};
 use wasmer_vfs::{FileSystem, RootFileSystemBuilder, SpecialFile, WasiPipe};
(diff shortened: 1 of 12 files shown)


syrusakbary:
Can the control plane be attached to the WasiEnv? And if so, can we customize the threads num via the WasiEnv builder?

theduke:
It already is attached through the `WasiProcess`, which is stored in the env.

Currently Deploy always goes through `Console` (though that should change soon), which constructs the control plane, so we just need to add a config option there. 

Just need to add an option to the console to make it configurable.

PR #3499, opened 2023-01-19
Run make lint on all operating systems

### LINKED ISSUE #3489: make lint does not properly catch CI failures when committing changes into master

fschutt:
Fixes #3489.

Continuation of https://github.com/wasmerio/wasmer/pull/3491.


.github/workflows/test.yaml
```diff
@@ -34,26 +34,78 @@ jobs:
...
   lint:
-    name: Code lint
-    runs-on: ubuntu-22.04
+    name: Code lint on ${{ matrix.metadata.build }}
+    runs-on: ${{ matrix.metadata.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        metadata: [
+          {
```
theduke, line 43: I'm not a big fan of adding 4 more extra jobs, we already have a lot of them.

Can we add the clippy tests to some other jobs that are relatively fast anyway?

fschutt: Done in https://github.com/wasmerio/wasmer/pull/3499/commits/6ceb8ff312b3754c87d1d8405e0e28341dcfd0ce

theduke (approved):
LGTM

fschutt:
@syrusakbary can be merged now, I don't think we need to spend more time on this

theduke:
@fschutt conflicts!

Makefile
```diff
@@ -849,9 +849,12 @@ update-testsuite:
 
 lint-packages: RUSTFLAGS += -D dead-code -D nonstandard-style -D unused-imports -D unused-mut -D unused-variables -D unused-unsafe -D unreachable-patterns -D bad-style -D improper-ctypes -D unused-allocation -D unused-comparisons -D while-true -D unconditional-recursion -D bare-trait-objects -D function_item_references # TODO: add `-D missing-docs`
 lint-packages:
-	RUSTFLAGS="${RUSTFLAGS}" cargo clippy --all --exclude wasmer-cli -- -D clippy::all
+	RUSTFLAGS="${RUSTFLAGS}" cargo clippy --all --exclude wasmer-cli --exclude wasmer-compiler-llvm $(compiler_features) -- -D clippy::all
```
theduke, line 852: This should only be excluded if LLVM is disabled, not unconditionally.

theduke (changes requested):

Makefile
fschutt: Done in https://github.com/wasmerio/wasmer/pull/3499/commits/c83b9e0455d1a906ba105258e6760a177d2816d9

Issue #3510, opened 2023-01-20
wasmer run corrupts package on the second run on the hackathon branch


fschutt:
In order to reproduce this bug, you need to checkout the `hackathon` branch, which has some changes of WASIX merged.

```
$ git checkout hackathon

$ make build-wasmer

$ ./target/release/wasmer run ciuser/customlambda 55
139583862445                                                           

$ ./target/release/wasmer run ciuser/customlambda 55
Fatal Python error: Py_Initialize: Unable to get the locale encoding
Traceback (most recent call last):
  File "lib/python3.6/encodings/__init__.py", line 31, in <module>
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 674, in exec_module
  File "<frozen importlib._bootstrap_external>", line 779, in get_code
  File "<frozen importlib._bootstrap_external>", line 487, in _compile_bytecode
EOFError: marshal data too short
error: failed to run `/Users/fs/.wasmer/checkouts/68747470733a2f2f7761706d2e696f2f6369757365722f637573746f6d6c616d626461@0.1.0/atoms/self::python@0.1.0:python`
│   1: RuntimeError: unreachable
╰─▶ 2: unreachable

$ ./target/release/wasmer -vV    
wasmer 3.1.0 (799ec 2023-01-20)
binary: wasmer-cli
commit-hash: 799ecbab394627594944e86196414b2347b7d2e6
commit-date: 2023-01-20
host: aarch64-apple-darwin
compiler: singlepass,cranelift
``` 

In the hackathon I could work around this bug by clearing the `~/.wasmer/checkouts` directory before every run, but obviously this is not a long-term solution.

I am not sure if this is fixed in master because of https://github.com/wasmerio/wasmer/issues/3509 - I can't really test it.

fschutt:
This bug also appears on master:

```
./target/release/wasmer run ciuser/customlambda --disable-cache 55
Fatal Python error: Py_Initialize: Unable to get the locale encoding
Traceback (most recent call last):
  File "lib/python3.6/encodings/__init__.py", line 31, in <module>
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
[...]

Issue #3470, opened 2023-01-10
SIGBUS and SIGSEGV a when compiling a modified version of php_cgi with cranelift


Michael-F-Bryan:
### Describe the bug

I am running into segfaults (`SIGSEGV` and `SIGBUS`) during the compilation of a particular WebAssembly file. 

This happens on both debug and release versions of the `wasix` branch (commit d34ce882e5e90ef500b51a2caacd5939e3397c82) on a M1 Mac, although the release binary triggers a `SIGSEGV` while the debug binary triggers a `SIGBUS`.

```console
$  ~/.cargo/bin/wasmer --version --verbose
wasmer 3.1.0 (d34ce 2023-01-10)
binary: wasmer-cli
commit-hash: d34ce882e5e90ef500b51a2caacd5939e3397c82
commit-date: 2023-01-10
host: aarch64-apple-darwin
compiler: singlepass,cranelift

$ rustc --version --verbose
rustc 1.68.0-nightly (659e169d3 2023-01-04)
binary: rustc
commit-hash: 659e169d37990b9c730a59a96081f2ef7afbe8f1
commit-date: 2023-01-04
host: aarch64-apple-darwin
release: 1.68.0-nightly
LLVM version: 15.0.6

$ uname -a
Darwin mbp 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:03:51 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T6000 arm64
```

### Steps to reproduce

First, install the `wasmer` CLI from the `wasix` branch.

```console
$ git checkout d34ce882e5e90ef500b51a2caacd5939e3397c82
$ cargo install --features cranelift,webc_runner,wasi,singlepass --path lib/cli --debug
```

Then, extract the `modified.wasm` file from [input.zip](https://github.com/wasmerio/wasmer/files/10386982/input.zip).

Now you can try to run it with `wasmer run`.

```console
$ ~/.cargo/bin/wasmer run --cranelift modified.wasm
[1]    48340 bus error  ~/.cargo/bin/wasmer run --cranelift modified.wasm
```

Running LLDB on the debug binary gave the following backtrace:

```
(lldb) run rename-import-namespace/modified.wasm
Process 45804 launched: '/Users/work/.cargo/bin/wasmer' (arm64)
Process 45804 stopped

* thread #1, name = 'main', queue = 'com.apple.main-thread', stop reason = signal SIGBUS
[...]

PR #3511, opened 2023-01-20
Incremented CURRENT_VERSION, so all cache will be invalidate and be rebuilt with the 3.2 version

### LINKED ISSUE #3509: wasmer run segfaults on current master branch
```
$ ./target/release/wasmer run ./lib/c-api/examples/assets/qjs.wasm
zsh: segmentation fault  ./target/release/wasmer run ./lib/c-api/examples/assets/qjs.wasm

$ ./target/release/wasmer -vV
wasmer 3.1.0 (580ba 2023-01-20)
binary: wasmer-cli
commit-hash: 580baeb5090ab825d082e4d5dbcafb9e880d3e06
commit-date: 2023-01-20
host: aarch64-apple-darwin
```

I have no more context on this bug, but it's a full segfault, no panic trace given. Not sure if this is because of `zsh` interaction, but even when doing `codesign -d -vvv --entitlements :- ./target/release/wasmer` it doesn't work on Mac.

ptitSeb:
This will force all old Cache to be purged and regenerated.
This should also fix #3509 , #3510 and #3470

### DIFF
--- a/lib/types/src/serialize.rs
+++ b/lib/types/src/serialize.rs
@@ -178,7 +178,7 @@ pub struct MetadataHeader {
 impl MetadataHeader {
     /// Current ABI version. Increment this any time breaking changes are made
     /// to the format of the serialized data.
-    const CURRENT_VERSION: u32 = 1;
+    const CURRENT_VERSION: u32 = 2;
 
     /// Magic number to identify wasmer metadata.
     const MAGIC: [u8; 8] = *b"WASMER\0\0";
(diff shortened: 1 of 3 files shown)


theduke (approved):

fschutt:
The issue is that create-exe links the objects using the `libwasmer.a` pulled from GitHub (which expects the old `METADATA_HEADER_VERSION`). That's why the tests fail because wasmer-cli generates objects with version 2, but `object_metadata_deserialize` from libwasmer.a still expects version 1.

We need to disable tests, release the alpha and the re-enable the tests.

syrusakbary (approved):

PR #3522, opened 2023-01-23
Fix --mapdir CLI flag not working on generated executables

fschutt:
We did not test that `--mapdir` actually works on the generated executables (doesn't work on 3.1.0 either).

Needs to be addressed before releasing the 3.1.0-alpha.1


fschutt:
@syrusakbary @theduke So I've fixed the PR for now, but we need to sync on how the `--mapdir` permissions should work, right now wasmer allows access to the root directory if any (!) `--mapdir` argument is given, meaning that any wasm file can read `/etc/passwd` if it wants to. These permissions should be more locked-down, but I'm not fixing that in this PR. This would need more design syncs and I'd like to hand off this PR to @theduke or @Michael-F-Bryan if possible, because discussing how to handle filesystem permissions with `HostFs` is likely going to take some time.

PR #3534, opened 2023-01-25
Use JS VM store instead of artificially replicate it

syrusakbary:
Use JS VM store instead of artificially replicate it.
Right now all the JS VMs use a single store (v8 isolates, for example) per process. We were replicating the store unnecessarily in Rust, when we can just leverage that store.

This PR should bring some relative speed ups on the outer interface of Wasmer (for example, when doing function calls) or getting exports of an instance

### DIFF
--- a/lib/api/src/js/store.rs
+++ b/lib/api/src/js/store.rs
@@ -78,14 +78,6 @@ impl fmt::Debug for Store {
     }
 }
 
-/// A trait represinting any object that lives in the `Store`.
-pub trait StoreObject {
-    /// Return true if the object `Store` is the same as the provided `Store`.
-    fn comes_from_same_store(&self, _store: &Store) -> bool {
-        true
-    }
-}
-
 impl AsStoreRef for Store {
     fn as_store_ref(&self) -> StoreRef<'_> {
         StoreRef { inner: &self.inner }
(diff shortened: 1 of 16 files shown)


lib/api/src/js/store.rs
```diff
@@ -266,23 +254,23 @@ mod objects {
 }
 
     impl_store_object! {
-        functions => VMFunction,
-        tables => VMTable,
+        // Note: we store the globals in order to be able to access them later via
+        // `StoreObjects::iter_globals`.
         globals => VMGlobal,
```
ptitSeb, line 259: but VMGlobal are not handled as a StoreObject anymore. Isn't there a risk of unsynchronisation here?

ptitSeb (commented):
My only concer is how VMGlobal can be not a StoreObject anymore, but still beeing reference as a StoreObject.

This PR make Store mostly useless in JS space.

ptitSeb (approved):

lib/api/src/js/store.rs
syrusakbary: > Isn't there a risk of unsynchronisation here?

No. The only risk is the fact that the globals are stored in the `Store` as a side effect. But they are stored and work nonetheless.
But I think we should probably remove that after merging WASIX

PR #3547, opened 2023-01-26
Remove setting memory for WASI compatibility

syrusakbary:
I found some dangling code that was no longer needed (since it's not present either in the sys counterpart).

Needs review @theduke @john-sharratt


theduke (approved):
We still very much require "memory" to be exported, but that is actually the case anyway for most of the common build systems targeting WASM, so this should be fine!

If anything breaks we can always re-add the logic in `-wasi`.

PR #3556, opened 2023-01-27
Use standard API for js and sys for Module. Added Engine in js

syrusakbary:
This PR aims to standardize the API for js/sys using a similar strategy as Rust is doing when offering the same API in different Operating Systems.

This PR aims to move things one at a time, so it's easier to test along the process.
APIs generalized:
* [x] `Store`
* [x] `Engine`
* [x] `Module`
* [x] `Instance`
* [x] `Imports`
* [x] `Exports`
* [x] Externs
  * [x] `Function`
  * [x] `NativeFunction`
  * [x] `Global`
  * [x] `Memory`
     * [x] `MemoryView`
  * [x] `Table`
* [x] `NativeType`
* [x] `FunctionEnv`
* [x] `WasmPtr`
* [x] `WasmRef`, `WasmSlice`, `WasmSliceIter` (from `mem_access.rs`)
* [ ] `Trap`
* [x] `Value`

Perhaps, also optional but maybe interesting

* [x] `ExternRef` - https://github.com/wasmerio/wasmer/pull/3224
  * [ ]  Real implementation


lib/api/src/errors.rs
```diff
@@ -0,0 +1,73 @@
+use crate::{LinkError, RuntimeError};
+use thiserror::Error;
+
+// /// An error while instantiating a module.
+// ///
+// /// This is not a common WebAssembly error, however
```
syrusakbary, line 6: Need to standarize errors in wasmer_types

lib/api/src/engine.rs
```diff
@@ -40,12 +52,13 @@ impl AsEngineRef for EngineRef<'_> {
...
+// impl<P> AsEngineRef for P
+// where
+//     P: Deref,
+//     P::Target: AsEngineRef,
+// {
+//     fn as_engine_ref(&self) -> EngineRef<'_> {
+//         EngineRef { inner: **self }
+//         (**self).as_engine_ref()
+//     }
+// }
```
syrusakbary, line 64: I might need your help on reenabling this @Michael-F-Bryan

Michael-F-Bryan: You're running into a coherence issue because we have a blanket `impl<P: Deref> EngineRef for P` and also explicitly implement `AsEngineRef` for `Engine` which is a type alias for `engine_imp::Engine`, which is a re-export of `wasmer_compiler::Engine`, which belongs to another crate. 

The problem is our `wasmer_compiler` crate could one day decide to implement `Deref` and return something which itself implements `AsEngineRef`, causing a conflict.

The fix is actually pretty simple, and we were probably going to do it anyway... Create a new `Engine` struct in `lib/api/src/engine.rs` which wraps `engine_imp::Engine` and implement `AsEngineRef` for *that* instead of doing `pub type Engine = engine_imp::Engine`.  That way `Engine` becomes part of the `wasmer` crate again and there are no coherence issues.

syrusakbary: This is re-enabled now. Closing

lib/api/src/errors.rs
syrusakbary: I standardize everything except the `InstantiationError` as that error depends on the `RuntimeError` which is different for sys and js (since the sys one includes a backtrace that can't be used on js)

john-sharratt (approved):
Looks good

lib/api/src/sys/engine.rs
```diff
@@ -0,0 +1,118 @@
...
+
+// impl From<CompilerConfig> for crate::engine::Engine {
+//     fn from(compiler: CompilerConfig) -> Self {
+//         Self(compiler.into())
+//     }
+// }
+
+/// The custom trait to access to all the `sys` function in the common
+/// engine.
+pub trait WasmerCompilerEngine {
```
theduke, line 68: This is an extension trait, which are usually suffixed with `Ext`.

maybe: `WasmerCompilerEngineExt`

The name is a bit clunky though, what about `NativeEngineExt`?
Or, if you want to highlight Wasmer because we might get additional engines in the futures, `WasmerEngineExt`.

lib/api/src/js/as_js.rs
```diff
@@ -0,0 +1,308 @@
...
+    }
+}
+
+impl AsJs for Imports {
+    type DefinitionType = crate::module::Module;
+
+    // Annotation is here to prevent spurious IDE warnings.
+    #[allow(unused_unsafe)]
+    fn as_jsvalue(&self, store: &impl AsStoreRef) -> wasm_bindgen::JsValue {
+        // /// Returns the `Imports` as a Javascript `Object`
```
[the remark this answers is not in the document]
syrusakbary: I want to remove it just at the end to make sure we have the right implementation (the commented one and the non-one are different ones but both valid)

lib/api/src/js/extern_ref.rs
```diff
@@ -0,0 +1,40 @@
+use std::any::Any;
+
+use crate::js::vm::VMExternRef;
+use crate::store::{AsStoreMut, AsStoreRef};
+use wasmer_types::RawValue;
+
+#[derive(Debug, Clone)]
+#[repr(transparent)]
+pub struct ExternRef;
```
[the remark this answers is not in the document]
syrusakbary: We should implement it outside of this PR, is not trivial

lib/api/src/sys/engine.rs
syrusakbary: Good call. I ended using `NativeEngineExt` which I think is more accurate

lib/api/src/js/as_js.rs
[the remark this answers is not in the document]
syrusakbary: Done

lib/api/src/engine.rs
```diff
@@ -0,0 +1,92 @@
...
+#[cfg(feature = "js")]
+pub(crate) use crate::js::engine::default_engine;
+
+/// The engine type
+#[derive(Clone, Debug)]
+pub struct Engine(pub(crate) engine_imp::Engine);
+
+impl Engine {
+    #[deprecated(
+        since = "4.0.0",
```
ptitSeb, line 19: Mmmm, so you want the 3.2 to become 4.0 it seems?

lib/api/src/engine.rs
```diff
@@ -0,0 +1,92 @@
...
+        note = "engine.cloned() has been deprecated in favor of engine.clone()"
+    )]
+    /// Returns the [`Tunables`].
```
ptitSeb, line 22: Returns [`Engine`]

lib/compiler/src/engine/trap/frame_info.rs
```diff
@@ -54,7 +54,7 @@ pub struct GlobalFrameInfoRegistration {
 struct ModuleInfoFrameInfo {
     start: usize,
     functions: BTreeMap<usize, FunctionInfo>,
-    module: ModuleInfo,
+    module: Arc<ModuleInfo>,
```
ptitSeb, line 57: Why is this Arc needed here?

lib/types/src/store_id.rs
```diff
@@ -0,0 +1,22 @@
...
+/// correct context.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub struct StoreId(NonZeroU64);
+
+impl Default for StoreId {
+    // Allocates a unique ID for a new context.
+    fn default() -> Self {
+        // No overflow checking is needed here: overflowing this would take
+        // thousands of years.
+        static NEXT_ID: AtomicU64 = AtomicU64::new(1);
```
ptitSeb, line 19: Maybe we should use `AtomicU64` or `AtomicUsize` consistantly, even if this end up beeing the same things on our platforms

ptitSeb (approved):

lib/api/src/engine.rs
syrusakbary: Good catch. Let's make it 3.2

syrusakbary:
```suggestion
        since = "3.2.0",
```

lib/api/src/engine.rs
syrusakbary:
```suggestion
    /// Returns the [`Engine`].
```

lib/compiler/src/engine/trap/frame_info.rs
syrusakbary: Basically, before we were cloning the `ModuleInfo` in multiple places, but that's no longer needed.

By having an Arc, we allow it to be cloned *only* when we set up the module frame info registration (is needed, as otherwise the lifetimes for accessing a pointer will make our life a bit of a hell there), so Arc was the simplest solution

lib/types/src/value.rs
```diff
@@ -43,6 +67,27 @@ macro_rules! partial_eq {
...
+                unsafe {
+                    match &from {
+                        RawValue { i32: i32 } => *i32 as _,
+                        // RawValue { f32: f32 } => *f32 as _,
+                        // RawValue { u32: u32 } => *u32 as _,
+                        // RawValue { i64: i64 } => *i64 as _,
+                        // RawValue { f64: f64 } => *f64 as _,
+                        // RawValue { u64: u64 } => *u64 as _,
+                        // _ => unimplemented!()
+                    }
```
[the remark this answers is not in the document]
syrusakbary: This implementation is a bit of a hack, because is using a Rust Union which is inheritnly unsafe.
Basically, when we had the different values for matching the union, Rust linter was complaining that the rest of match cases were not being accessed (because Rust don't know the type of the value)

lib/types/src/store_id.rs
syrusakbary: Fixed

theduke (approved):
This is a brilliant improvement for the API!

PR #3557, opened 2023-01-30
Fix make-test-integration CLI producing files when running locally

fschutt:
When running make test-integration-cli, I don't want it to produce files locally, but in the CI I want to see the debug output in case something goes wrong with Zig.

### DIFF
--- a/tests/integration/cli/tests/run.rs
+++ b/tests/integration/cli/tests/run.rs
@@ -164,8 +164,10 @@ fn test_cross_compile_python_windows() -> anyhow::Result<()> {
             output.arg("-o");
             output.arg(python_wasmer_path.clone());
             output.arg(format!("--{c}"));
-            output.arg("--debug-dir");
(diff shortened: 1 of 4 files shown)


tests/integration/cli/tests/run.rs
```diff
@@ -164,8 +164,10 @@ fn test_cross_compile_python_windows() -> anyhow::Result<()> {
             output.arg("-o");
             output.arg(python_wasmer_path.clone());
             output.arg(format!("--{c}"));
-            output.arg("--debug-dir");
-            output.arg(format!("{t}-{c}"));
```
theduke, line 168: Do we still want this on CI?
If so, are you setting it via env var?

fschutt: I want it on the CI, just not locally.

syrusakbary (approved):

syrusakbary (approved):

PR #3558, opened 2023-01-30
Reduce libwasmer-headless size

fschutt:
This reduces the libwasmer-headless size from 9MB to 2.4 MB by removing all panic info and using `abort()` instead, however we need to build with a custom libstd for that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -617,6 +617,8 @@ ifeq ($(IS_DARWIN), 1)
 endif
 endif
 
+package-capi-headless: build-capi-headless package-capi
+
 package-capi:
 	mkdir -p "package/include"
 	mkdir -p "package/lib"
--- a/Makefile
+++ b/Makefile
@@ -451,11 +451,11 @@ build-capi-llvm-universal:
 
 build-capi-headless:
 ifeq ($(CARGO_TARGET_FLAG),)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -695,39 +695,13 @@ package-capi:
 		cp target/$(HOST_TARGET)/release/libwasmer.a package/lib/libwasmer.a ;\
 	fi
 
-package-capi-headless: build-capi-headless
```
syrusakbary, line 698: Why did we removed this?

fschutt: This step is already included in `package-capi`, this step is useless.

However, if `package-capi-headless` runs after `package-capi`, it can happen that `$(TARGET_DIR)` is set to `target/release`, so it happened that it copied `target/release/libwasmer.a` to `libwasmer-headless`, not `target/headless/.../libwasmer.a`. So you suddenly get a `libwasmer-headless.a` that is a duplicate of `libwasmer.a`.

It's better to just have one `package-capi` step that also accounts for the `-headless` stuff.

fschutt: I added it back, but removed all of the steps for `package-capi-headless`, since they are now part of the `package-capi` step: https://github.com/wasmerio/wasmer/pull/3558/commits/a23c02f927470d2dda3106d12e551de086af5536

syrusakbary (approved):

PR #3562, opened 2023-02-01
Improved handling of wasmer-headless generation and use on local run

ptitSeb:
# Description
Change Makefile to better handle generation of libwasmer-headless localy.
Also include generation of of libwasmer-headless on test integration and "build all" config as it's now used by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -561,10 +561,10 @@ test-wasi-unit:
 test-wasi:
 	$(CARGO_BINARY) test $(CARGO_TARGET_FLAG) --release --tests $(compiler_features) -- wasi::wasitests
 
-test-integration-cli: build-wasmer build-capi build-capi-headless package distribution
+test-integration-cli: build-wasmer build-capi package-capi-headless package distribution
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -635,6 +635,10 @@ package-capi:
 		cp target/headless/$(CARGO_TARGET)/release/wasmer.dll package/lib/wasmer-headless.dll ;\
 	fi
 
+	if [ -f target/headless/$(HOST_TARGET)/release/wasmer.dll ]; then \
```
theduke, line 638: I assume the duplication of these copies is intentional to cover both native and cross compilation?

theduke (commented):
On question, but otherwise LGTM.

Makefile
ptitSeb: yes, but it's to cover CI vs Local also, because local use cross-compile to HOST_TARGET (or else the build would not work for some obscure reasons)

theduke (approved):

PR #3564, opened 2023-02-02
Updated Cranelift to 0.91

ptitSeb:
# Description
Update Cranelift to v0.91

### DIFF
--- a/lib/compiler-cranelift/src/translator/func_translator.rs
+++ b/lib/compiler-cranelift/src/translator/func_translator.rs
@@ -243,7 +243,7 @@ fn parse_function_body<FE: FuncEnvironment + ?Sized>(
     // If the exit block is unreachable, it may not have the correct arguments, so we would
     // generate a return instruction that doesn't match the signature.
     if state.reachable {
-        debug_assert!(builder.is_pristine());
+        //debug_assert!(builder.is_pristine());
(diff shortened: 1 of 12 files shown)


lib/compiler-cranelift/src/translator/func_translator.rs
```diff
@@ -243,7 +243,7 @@ fn parse_function_body<FE: FuncEnvironment + ?Sized>(
     // If the exit block is unreachable, it may not have the correct arguments, so we would
     // generate a return instruction that doesn't match the signature.
     if state.reachable {
-        debug_assert!(builder.is_pristine());
+        //debug_assert!(builder.is_pristine());
```
syrusakbary, line 246: What that is for?

syrusakbary (approved):

lib/compiler-cranelift/src/translator/func_translator.rs
ptitSeb: `builder.is_pristine()` is gone. The replacement is not really equivalent, I removed the assert. I can remove the code entirely instead of just commenting it

syrusakbary: Ok! Sounds good to me, we should be good to merge

Issue #3508, opened 2023-01-19
Engine::new().set_target(Some(Windows)) crashes Module::compile


fschutt:
When initializing the `Engine` from `CompilerOptions::get_engine()`, compiling a module crashes with
```
  3: wasmer_compiler::engine::code_memory::CodeMemory::allocate
   4: wasmer_compiler::engine::inner::EngineInner::allocate
   5: wasmer_compiler::engine::artifact::Artifact::from_parts
   6: wasmer_compiler::engine::artifact::Artifact::new
   7: wasmer_compiler::engine::inner::Engine::compile
   8: wasmer::sys::module::Module::from_binary
   9: wasmer_cli::commands::create_exe::get_module_infos
  10: wasmer_cli::commands::create_exe::CreateExe::execute
  11: wasmer_cli::cli::wasmer_main_inner
  12: wasmer_cli::cli::wasmer_main
```

... if the `target` is set to `Windows`.

Minimal example: 

```rust
use std::str::FromStr;

use wasmer::*; // v3.1.0

const WASM_BYTES: &[u8] = include_bytes!("qjs.wasm");

fn main() {
    let engine = EngineBuilder::new(Box::new(wasmer_compiler_cranelift::Cranelift::default()) as Box<_>)
    .set_target(Some(Target::new(Triple::from_str("x86_64-windows-gnu").unwrap(), [CpuFeature::SSE2].into_iter().collect())));
    let store = Store::new(engine.engine());
    let module = Module::new(&store, WASM_BYTES).unwrap();
    println!("{:#?}", module.name());
}
```

This can be disabled by setting the `target.operating_system` to anything but `Windows`.

PR #3573, opened 2023-02-03
Fix/compile not in memory

ptitSeb:
When compiling an artifact, don't load compiled result in memory if the the compilation target is not compatible with the host.

This will also help with #3508 because runtime unwind information will not be constructed when Cross-compiling.

Note that they are some small API-breaking changes here:
* A new InstantiationError::DifferentArchOS 
*  Artifact::from_parts(...) now takes a 3rd parameter (&Target of the parts)

### DIFF
--- a/lib/c-api/src/wasm_c_api/instance.rs
+++ b/lib/c-api/src/wasm_c_api/instance.rs
@@ -85,6 +85,12 @@ pub unsafe extern "C" fn wasm_instance_new(
 
             return None;
         }
+
+        Err(e @ InstantiationError::DifferentArchOS) => {
+            crate::error::update_last_error(e);
+
+            return None;
+        }
     };
 
     Some(Box::new(wasm_instance_t {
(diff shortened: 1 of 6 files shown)


lib/c-api/src/wasm_c_api/instance.rs
```diff
@@ -85,6 +85,12 @@ pub unsafe extern "C" fn wasm_instance_new(
 
             return None;
         }
+
+        Err(e @ InstantiationError::DifferentArchOS) => {
```
syrusakbary, line 89: I didn't know this syntax. Nice!

lib/types/src/compilation/target.rs
```diff
@@ -192,6 +192,13 @@ impl Target {
     pub fn cpu_features(&self) -> &EnumSet<CpuFeature> {
         &self.cpu_features
     }
+
+    /// Check if target is a native (eq to host) or not
+    pub fn is_native(&self) -> bool {
+        let host = Triple::host();
+        host.operating_system == self.triple.operating_system
+            && host.architecture == self.triple.architecture
```
syrusakbary, line 200: I believe you can do simple compare on the triple itself.

```suggestion
host == self.triple
```

ptitSeb: We take the risk of comparing Vendor for example, that should could lead to a different result while it should not.
I did this to be sure to have a minimum comparison.

syrusakbary (approved):

Issue #3567, opened 2023-02-02
UAF possible (though not likely in practice) with Memory::view


coolreader18:
### Describe the bug

Because the return type of `Memory::view` is tied to the lifetime of `self` and not the lifetime of the `store`, it's possible to use the `MemoryView` after the store (and therefore the actual memory buffer) has dropped.

### Steps to reproduce

```rust
let mut store = wasmer::Store::new(wasmer::Engine::headless());
let mem = wasmer::Memory::new(&mut store, wasmer::MemoryType::new(1, None, false))?;
let view = mem.view(&store);
view.write_u8(0x10, 42)?;
drop(store); // bye bye memories
let x = view.read_u8(0x10)?; // hello UAF :3
```

```
❯ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/testcase`
[1]    158432 segmentation fault (core dumped)  cargo run
```

### Additional context

wasmtime solves this issue by having the return of its equivalent function be tied just to the store parameter, not to the memory handle:
```rust
pub fn data<'a, T: 'a>(&self, store: impl Into<StoreContext<'a, T>>) -> &'a [u8]
```

Michael-F-Bryan:
I believe this is a duplicate of #3516.

Issue #3516, opened 2023-01-21
Segfault with `Memory::grow`/`MemoryView` interaction


Imberflur:
### Describe the bug

https://docs.rs/wasmer/3.1.0/wasmer/struct.MemoryView.html documents:
> After a memory is grown a view must not be used anymore.

However, it is possible to do this in safe code and instead of producing an error or panicking this results in a segfault. So there is a soundness issue with the interaction of `Memory::grow` and `Memory::view`.

`wasmer v3.1.0 | rustc 1.66.1 (90743e729 2023-01-10) | x86_64`

### Steps to reproduce

```rs
use wasmer::{Memory, MemoryType, Store};

fn main() {
    // The default configuration gives memory plenty of room to grow in-place.
    let tunables = wasmer::BaseTunables {
        static_memory_bound: 4.into(),
        static_memory_offset_guard_size: 1024 * 64 * 4,
        dynamic_memory_offset_guard_size: 1024 * 64 * 4,
    };
    let engine = wasmer::EngineBuilder::headless().engine();

    let mut store = Store::new_with_tunables(engine, tunables);

    let memory = Memory::new(&mut store, MemoryType::new(10, None, false)).unwrap();
    let view = memory.view(&store);
    memory.grow(&mut store, 1024).unwrap();
    view.write_u8(0, 4u8).unwrap();
}
```

```shell
> cargo r
fish: Job 1, 'cargo r' terminated by signal SIGSEGV (Address boundary error)
```

### Expected behavior

One of the following:
* Producing an error, panicking, or aborting.
* An API that statically prevents this, e.g. by having the `Store` borrowed for the lifetime of `MemoryView`.
* An unsafe API.

### Actual behavior

Access to freed memory in safe code. Segfault.

Michael-F-Bryan:
@ptitSeb was this resolved in #3580?

PR #3580, opened 2023-02-09
Make memoryView lifetime linked to StoreRef instead of Memory

ptitSeb:
On Sys (not on JS), make MemoryView lifetime linked to the StoreRef instead of the Memory.
For #3516 and #3567 

This change in the Lifetime makes the examples in both tickets impossible to build.


syrusakbary:
I'd like to rereview/merge this after we merge the unification PR: #3556

syrusakbary (approved):

Issue #3438, opened 2022-12-20
[C API] Memory leak


rt21q8:
### Describe the bug
When whole `wasmer` related cleanup calls are done, not whole memory is released. This scenario can be also observed on slightly modified example files contained in this repository. Details described below. Is this known/expected behavior? 

```sh
$ echo "`wasmer -V` | `rustc -V` | `uname -m`"
wasmer 3.1.0 | rustc 1.63.0 (4b91a6ea7 2022-08-08) | x86_64
```

### Steps to reproduce

1. Go to `lib/c-api/examples`.
2. Put all logic from `main` function into infinite `while` loop for example files like `wasi.c` or `imports-exports.c` or `exports-function.c`.
3. Rebuild modified examples.
4. Run new executables.
5. Observe memory usage for started process in tool like `top`.
### Expected behavior

Memory usage should not keep growing continuously (there is cleanup logic at the end of main/loop - are these all required calls?).
### Actual behavior
Memory keeps growing.

ptitSeb:
I think this has been fixed now, on master, with latest Instance memory leak. Can you check again and see if you still see memory leak?

ptitSeb:
oh, ok. So it's another leak. Thanks for the testing, I'll investigate on this later.

EclesioMeloJunior:
hello! I would like to mention that I've posted this issue https://github.com/wasmerio/wasmer-go/issues/393 on `wasmer-go` repository where I try to explain a memory leak I am facing that may be related to `wasmer-c-api`.

PR #3581, opened 2023-02-13
Fix WASI example that initialized wasi_config 2 times

ptitSeb:
This fix a small memory leak in a c-api example.

This is for #3438 
Note that I tested with valgrind and found no other memory leak in the c-api examples.


Michael-F-Bryan (approved):
Looks good to me :+1:

PR #3584, opened 2023-02-15
Wasix major fixes and tweaks

john-sharratt:
- Removed generation_arena which was causing some serious leakages of files and sockets
- Added OsError for NetworkErrors so that "Too Many Open Files" is properly passed
- Local networking will now cap at 10 sockets in the backlog
- Added the missing shutdown error code
- Removed the inodes lock around most of the WASI syscalls
- Fixed some race conditions in the event notifications for WASI
- The polling loop will now only notify a closed socket once
- Event notifications now uses Wakers rather than MPSC
- Some socket errors now return the right codes which prevents panics in WASM
- Fixed a bug where the file read and write guards might release the file before the lock
- The inode seed is now much safer preventing overlaps
- The fd seed is now much safer preventing overlaps
- Closing files is now implicit rather than explicit reducing possibliities for error
- Forking of file descriptors is now much simplier
- Polling events will now be returned in random order to prevent some race conditions
- Removed a number of memory allocations which were wasting memory and performance
- Sockets now only copy the send and recv data once rather than multiple times


lib/vnet/src/lib.rs
```diff
@@ -516,6 +516,9 @@ pub enum NetworkError {
     /// A call to write returned 0
     #[error("write returned 0")]
     WriteZero,
+    /// OS error
+    #[error("operating system error({0})")]
+    OsError(i32),
```
theduke, line 521: We are our own OS, I don't see how having an upstream OS error here makes sense. 
We should map OS errors to the appropriate error variant.

lib/wasi-local-networking/src/lib.rs
```diff
@@ -139,6 +139,12 @@ impl VirtualTcpListener for LocalTcpListener {
         &mut self,
         cx: &mut std::task::Context<'_>,
     ) -> std::task::Poll<Result<usize>> {
+        {
+            let backlog = self.backlog.lock().unwrap();
+            if backlog.len() > 10 {
```
theduke, line 144: So, if I understand correctly: here you are limiting the backlog to 10 connections.

But why are we even storing a backlog of connections?
That seems to invite all kinds of problems.

lib/wasi/src/fs/fd.rs
```diff
@@ -74,6 +70,21 @@ impl InodeVal {
     }
 }
 
+#[derive(Debug)]
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+pub struct NotificationInner {
+    /// Used for event notifications by the user application or operating system
+    /// (positive number means there are events waiting to be processed)
+    pub counter: AtomicU64,
```
theduke, line 78: None of this should be public, we should encapsulate all the logic in methods.

lib/wasi/src/fs/inode_guard.rs
```diff
@@ -237,209 +198,229 @@ impl<'a> Future for InodeValFilePollGuardJoin<'a> {
...
                 }
                 InodeValFilePollGuardMode::Socket { ref inner } => {
-                    let mut guard = inner.write().unwrap();
-                    guard.poll_write_ready(cx).map_err(net_error_into_io_err)
+                    let mut guard = inner.protected.write().unwrap();
+                    let res = guard.poll_write_ready(cx).map_err(net_error_into_io_err);
+                    match res {
+                        Poll::Ready(Err(err)) if is_err_closed(&err) => {
+                            tracing::trace!("socket write ready error (fd={}) - {}", fd, err);
+                            if std::mem::replace(&mut guard.notifications.closed, true) == false {
```
theduke, line 331: Using `replace` for a boolean is pretty odd, you can just dereference and set the value,  would be more readable.

theduke: Also, shouldn't all this code be on the guard poll_write_ready directly?

lib/wasi/src/fs/inode_guard.rs
```diff
@@ -237,209 +198,229 @@ impl<'a> Future for InodeValFilePollGuardJoin<'a> {
...
         }
         Poll::Pending
     }
 }
 
 #[derive(Debug)]
 pub(crate) struct InodeValFileReadGuard {
-    #[allow(dead_code)]
-    file: Arc<RwLock<Box<dyn VirtualFile + Send + Sync + 'static>>>,
     guard: RwLockReadGuard<'static, Box<dyn VirtualFile + Send + Sync + 'static>>,
```
theduke, line 413: I know this isn't new, but... I  do not .. like this ... at all...

Note: `tokio::sync::RwLock` has `Owned` variants of the guards that work for `Arc<RwLock<T>>`, created with `RwLock::read/write_owned()`.
That would be much better to use...
(the tokio locks still also have sync lock methods)

lib/vnet/src/lib.rs
john-sharratt: Fixed in https://github.com/wasmerio/wasmer/commit/4473be1b043cb772b2fa897835a75a6b2a8c2e56

lib/wasi-local-networking/src/lib.rs
john-sharratt: Yeah this is because TcpListener does not support polling via the `accept_ready_poll` method so we have to simulate it... we could quite easily reduce it to a single item and use an option instead if this is nicer?

theduke: Yeah I was also going to suggest just buffering a single one. :+1:

theduke: We don't want to have a bunch of idle TCP connections sitting around if the thread doesn't handle them.

lib/wasi/src/fs/fd.rs
john-sharratt: fixed in https://github.com/wasmerio/wasmer/commit/4f117a6b63222d403a1e1bed03ff5fb15b3527ed

lib/wasi/src/fs/inode_guard.rs
john-sharratt: that creates code duplication as it needs to be implemented for all socket implementations so it needs to stay here.
I prefer the std::mem::replace to be honest as its less lines of code

I made it slightly easier to read with this https://github.com/wasmerio/wasmer/commit/e82fc87a1d53f68b9a651b6afe96c3360fd30d1e

lib/wasi/src/fs/inode_guard.rs
john-sharratt: As discussed - updated here https://github.com/wasmerio/wasmer/commit/31c6eca32eeefaae422bf17c0048d1bd8b0d0300

lib/wasi/src/fs/inode_guard.rs
theduke: Not a big deal either way, that's a preference thing.

PR #3586, opened 2023-02-15
Make AsStoreRef and friends work for anything that implements Deref

Michael-F-Bryan:
# Description

While updating some WAI code, I ran into a weird bug where `&mut Store` didn't implement `AsEngineRef` because we've only got an `impl AsEngineRef for &Store`.

This PR uses the `Deref` (or `DerefMut`, when applicable) trait to implement things like `AsStoreRef`, `AsStoreMut`, and `AsEngineRef` for anything that can be dereferenced to an implementation of those traits. That means we'll have the same `&Store: AsEngineRef` impl, but we'll also pick up `AsEngineRef` implementations for `Box<T>`, `Arc<T>`, and so on.

### DIFF
--- a/lib/api/src/sys/store.rs
+++ b/lib/api/src/sys/store.rs
@@ -1,6 +1,9 @@
 use crate::sys::tunables::BaseTunables;
 use derivative::Derivative;
-use std::fmt;
+use std::{
+    fmt,
+    ops::{Deref, DerefMut},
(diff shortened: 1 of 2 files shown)


syrusakbary (approved):
Love it

theduke:
@Michael-F-Bryan nice !

PR #3590, opened 2023-02-17
Optimize getting byteLength in MemoryView::new

rlane:
This js_sys::Reflect call was taking about 20% of my [application's](https://oort.rs) CPU time. After this change MemoryView::new only consumes 1% of total CPU.


syrusakbary:
Great catch! Merging

syrusakbary (approved):

PR #3598, opened 2023-02-22
Wasix multithreading fix

john-sharratt:
There was a thread corruption issue causes when real threads spawned a thread
multiple times. This occurred because the thread context was being reused
from previously exited threads.

Also there was a bug in the `futex_wait` callback where it was not correctly
returning the woken event


theduke (approved):
Good call.

I thought something fishy was going on there when reading over the code when doing some other refactoring recently.

PR #3599, opened 2023-02-22
Move the `wcgi-runner` into Wasmer

### LINKED ISSUE #3600: Move the `wcgi_runner` crate to the Wasmer repository

Michael-F-Bryan:
As discussed previously, we're moving the `wcgi-runner` crate into the `wasmerio/wasmer` repo so it can be integrated into the `wasmer` CLI.

Fixes #3600.


lib/wcgi-runner/src/annotations.rs
```diff
@@ -0,0 +1,24 @@
+use wcgi_host::CgiDialect;
+
+// FIXME(@Michael-F-Bryan): Make this public in the webc crate
```
theduke, line 3: Is there anything preventing us from making this public and releasing a new version of webc?

Better to not carry around this legacy burden, because it will spread to every runner, and also to other products we are working on...

But this also feeds into layering concerns about webc. 
See below for more.

lib/wcgi-runner/src/builder.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /// Create a [`Runner`] that executes a WEBC file.
+    ///
+    /// If [`Builder::program`] was set, this will look for the command with
+    /// that name in the WEBC file. Otherwise, it will fall back to the WEBC
+    /// file's default entrypoint.
+    ///
+    /// This will infer the [`CgiDialect`] from the WEBC file's metadata
+    pub fn build_webc(self, webc: impl Into<Bytes>) -> Result<Runner, Error> {
```
theduke, line 151: We face a difficult layering problem with webc.
I also had to contend with the same issue in ... another product...

I arrived at the conclusion that what we want is:

* A webc runner that understands entrypoints and commands, extracts the required configuration and forwards to a concrete runners
* Concrete runners that are not webc-aware at all

If we don't do this we need to plaster webc logic all over every runner we implement, which feels like a lot of duplication and a layering violation to me.

theduke: Ignoring the above, I feel `new_webc` would be more appropriate here, and the method would also need to to take an`command: Option<String>`.

lib/wcgi-runner/examples/wcgi-runner.rs
```diff
@@ -0,0 +1,111 @@
...
+        address,
+        input,
+        env_all,
+        mapped_dirs,
+    } = Args::parse();
+
+    // Hack to initialize the global shared tokio task manager handle.
+    // Prevents "cannot drop runtime in async context" errors, because
+    // the default will be initialized to the current tokio context.
+    let rt = wasmer_wasi::runtime::task_manager::tokio::TokioTaskManager::default();
```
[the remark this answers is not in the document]
theduke: Note that we have the `VirtualTaskManager` trait to remain async runtime agnostic.

lib/wcgi-runner/examples/wcgi-runner.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// Example:
+    ///
+    /// --map-dir /www:./my-website
+    ///   => will make the ./my-website directory available for wazsm at /www
+    #[clap(
+        long = "mapdir",
+        name = "GUEST_DIR:HOST_DIR",
+        value_parser = parse_mapdir,
+    )]
+    mapped_dirs: Vec<(String, PathBuf)>,
```
theduke, line 71: We should stop propagating the `GUEST:HOST` paradigm. 
See comments on the builder below.

lib/wcgi-runner/src/builder.rs
```diff
@@ -0,0 +1,252 @@
...
+    {
+        for (guest_dir, host_dir) in mapping {
+            self.mapped_dirs.push((guest_dir.into(), host_dir.into()));
+        }
+
+        self
+    }
+
+    /// Pass in a [`Handle`] to a custom tokio runtime.
+    pub fn tokio_handle(self, handle: Handle) -> Self {
```
theduke, line 129: Would be preferable to remain runtime agnostic and use the `VirtualTaskManager` trait.

lib/wcgi-runner/src/context.rs
```diff
@@ -0,0 +1,226 @@
...
+                    })?;
+            }
+        }
+        Ok(root_fs)
+    }
+}
+
+/// Drive the request to completion by streaming the request body to the
+/// instance and waiting for it to exit.
+async fn drive_request_to_completion(
```
theduke, line 145: Don't we have tooling for this in `wcgi-host`?

lib/wcgi-runner/tests/integration.rs
```diff
@@ -0,0 +1,95 @@
+use std::path::Path;
+
+use bytes::Bytes;
+use http::{Request, StatusCode};
+use hyper::{body::HttpBody, Body};
+use tempfile::TempDir;
+use wcgi_host::CgiDialect;
+use wcgi_runner::Builder;
+
+const FERRIS_SAYS: &str = "https://registry-cdn.wapm.dev/packages/wasmer-examples/ferris-says/ferris-says-0.2.0-2f5dfb76-77a9-11ed-a646-d2c429a5b858.webc";
```
theduke, line 10: We are guilty of this in Deploy as well, but we should stop depending on packages in a registry for tests - this will lead to pain down the line.

Doesn't necessarily need to be fixed for this PR, but we should strive to have the tests not depend on hosted packages in an external registry, but rather just have a test package / wasm in the repo.

Note that we are also working on a mock Rust registry that will be able to serve locally configured webcs like the real registry to test loaders.

theduke, line 10: At the very least we shouldn't depend on a hard CDN path, but rather use `REGISTRY.com/namespace/package@version` with `Accept: application/webc`.

lib/wcgi-runner/src/runner.rs
```diff
@@ -0,0 +1,87 @@
...
+///     let runner = runner.clone();
+///     async move { Ok::<_, std::convert::Infallible>(runner) }
+/// });
+///
+/// hyper::Server::bind(&address).serve(make_service).await?;
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Clone)]
+pub struct Runner {
```
theduke, line 37: I'd name this `WcgiRunner`.

Don't want lots of different `Runner` structs in one repo, and it's more self-explanatory to see `WcgiRunner` when reading some code.

lib/wasi/src/runners/wcgi/handler.rs
```diff
@@ -0,0 +1,235 @@
...
+pub(crate) struct Handler {
+    pub(crate) program: Arc<str>,
+    pub(crate) env: Arc<HashMap<String, String>>,
+    pub(crate) args: Arc<[String]>,
+    pub(crate) mapped_dirs: Arc<[MappedDirectory]>,
+    pub(crate) task_manager: Arc<dyn VirtualTaskManager>,
+    pub(crate) module: Module,
+    pub(crate) dialect: CgiDialect,
+    #[derivative(Debug = "ignore")]
+    pub(crate) callbacks: Arc<dyn Callbacks>,
```
syrusakbary, line 40: Having everything as an arc here seems a bit wrong.
Why not create an inner type and wrap just only that type inside the Arc?

lib/cli/Cargo.toml
```diff
@@ -119,62 +140,62 @@ wast = ["wasmer-wast"]
...
 emscripten = ["wasmer-emscripten"]
 wat = ["wasmer/wat"]
-webc_runner = ["wasi", "wasmer-wasi/webc_runner", "wasmer-wasi/webc_runner_rt_wasi", "wasmer-wasi/webc_runner_rt_emscripten", "nuke-dir", "webc"]
+webc_runner = [
+    "wasi",
+    "wasmer-wasi/webc_runner",
+    "wasmer-wasi/webc_runner_rt_wasi",
+    "wasmer-wasi/webc_runner_rt_emscripten",
+    "wasmer-wasi/webc_runner_rt_wcgi",
+    "nuke-dir",
```
syrusakbary, line 149: I believe nuke-dir is no longer used. Can we remove it?

lib/cli/src/commands/run.rs
```diff
@@ -376,48 +372,46 @@ impl RunWithPathBuf {
...
+        let mut runner = wasmer_wasi::runners::wcgi::WcgiRunner::new(id);
+        runner.config().args(args);
+        if runner.can_run_command(id, command).unwrap_or(false) {
+            runner
+                .run_cmd(&container, id)
+                .context("WCGI runner failed")?;
         }
 
-        result.unwrap_or_else(|| Err(anyhow::anyhow!("neither emscripten or wasi file")))
+        anyhow::bail!("No runner");
```
syrusakbary, line 414: This error is not descriptive. Something like "No WCGI runner was found in the package with the name {X}" is way more helpful. Note: we may also want to specify when a runner is found, but is not wcgi

lib/wcgi-runner/tests/integration.rs
Michael-F-Bryan: I think it's the least-bad solution here. I've been in projects where they commit binaries to the repository and it makes using the repository a massive pain. Recompiling test binaries all the time will also slow CI down a lot.

syrusakbary (approved):

syrusakbary:
The PR looks hacikly good... let's merge once tests pass

lib/wcgi-runner/src/annotations.rs
Michael-F-Bryan: I've published this under the `webc::metadata::annotations` module.

lib/wcgi-runner/tests/integration.rs
Michael-F-Bryan: Done.

lib/wcgi-runner/src/runner.rs
Michael-F-Bryan: Done

lib/wasi/src/runners/wcgi/handler.rs
Michael-F-Bryan: Done.

lib/wcgi-runner/src/context.rs
Michael-F-Bryan: I guess we could pull it into the `wcgi-host` crate, but I feel like the code is going to be closely coupled enough to the way our `wasmer_wasi` implementation works that it won't be a very clean abstraction. 

In particular, there is enough juggling going on with the various pipes, setting up the `WasiEnvBuilder`, waiting on exit codes, and such that there's no easy way to abstract out the process of communicating with a WCGI binary. I feel like we'd end up creating a trait that only ever has one implementation because it's just not expressive enough to encompass everyone's needs.

theduke (commented):
If I'm reading the code correctly there is no way to run a local .wasm file with the runner, right?

How are you supposed to develop locally then?
For me that's a baseline requirement.

PR #3610, opened 2023-02-23
Wasix webc fs fix for inheritance

john-sharratt:
Fixed an issue where packages without commands were not loading the webc file system because it was package specific, for webc inheritence this does not make sense


john-sharratt:
wrong target

PR #3611, opened 2023-02-23
Wasix webc fs fix for inheritance

john-sharratt:
Fixed an issue where packages without commands were not loading the webc file system because it was package specific, for webc inheritence this does not make sense


theduke:
Subsumed by #3629 .

The command additions will be handled separately.

Issue #3592, opened 2023-02-19
`FunctionEnvMut` method for mutably borrowing both store and env


rachel-bousfield:
### Motivation
This feature would improve the ergonomics of `FunctionEnvMut<T>`.

In `wasmer 3.x` an `&mut impl AsStoreMut`, commonly a `&mut StoreMut` in practice, is needed to modify a `Global`.

If a `FunctionEnvMut<T>` carries a `Global`, then modification requires cloning it since `FunctionEnv::as_store_mut()` yields the needed `StoreMut`. This works fine if the `FunctionEnvMut<T>`'s underlying `T` is cheap to `Clone`, but for more complicated `FunctionEnv`'s, one need discard references to produced by `FunctionEnvMut::data_mut`.

This is awkward, since it means designing host functions like
```rs
fn host_foo(mut env: WasmEnvMut) {
    let global = env.data().global.clone().unwrap();
    let mut store = env.as_store_mut();
    global.set(&mut store, 0);
    
    let rest_of_state = env.data_mut();
    // can't use `store` anymore due to `rest_of_state`
    drop(rest_of_state)
    
    let mut store = env.as_store_mut();
    // back to being able to modify globals
}
```

### Proposed solution

We add a new method to `FunctionEnvMut<T>` that somehow returns both a `StoreMut` and the underlying `&mut T`. This would require deeper changes since these currently entail aliasing.

### Alternatives

Don't require `Global`, `Memory`, and the like need a `Store` for access as was the case in `wasmer 2.x`.

We found the upgrade to `wasmer 3.x` to be quite difficult due to the above. Needing a `Store` [required we re-architect our host functions](https://github.com/OffchainLabs/nitro/blob/69f76e461b32e307b0cf535648dfdc50d0ba6cdb/arbitrator/jit/src/gostack.rs#L23), using complex types to achieve what used to be given for free.

knarkzel:
Would be really nice to have this.

ptitSeb:
The new function is now on master. Can this ticket be closed?

Michael-F-Bryan:
This should be fixed now. @knarkzel and @rachel-bousfield let us know if it's still an issue and we can reopen the ticket.

PR #3612, opened 2023-02-23
Added conveniance function FunctionEnvMut::data_and_store_mut 

ptitSeb:
This allows to borrows mutable ref of both staore and data env from a FunctionEnvMut, making the writting of complex function s easier.

Should address #3592


lib/api/src/sys/function_env.rs
```diff
@@ -117,6 +117,17 @@ impl<T: Send + 'static> FunctionEnvMut<'_, T> {
...
+    /// Borrows a new mutable reference of both the attached Store and host state
+    pub fn data_and_store_mut(&mut self) -> (&mut T, StoreMut) {
+        let data = self.func_env.as_mut(&mut self.store_mut) as *mut T;
+        // telling the borrow check to close his eyes here
+        // this is still relatively safe to do as func_env are
+        // stored in a specific vec of Store, separate from the other objects
+        // and not really directly accessible with the StoreMut
+        let data = unsafe { &mut *data };
+        (data, self.store_mut.as_store_mut())
+    }
```
syrusakbary, line 130: Can we add it on the js version as well and add a `#[universal_test]` verifying the desired behavior? (that way, we test that it also works in js)

ptitSeb: It is on js (in the 2nd commit).
I'll find something for a test.

syrusakbary: Oh I missed the js impl. Let's just have an universal test

examples/imports_function_env_global.rs
```diff
@@ -0,0 +1,156 @@
...
+    // Once we have our counter we'll wrap it inside en `Env` which we'll pass
+    // to our imported functionsvia the FunctionEnv.
+    //
+    // This struct may have been anything. The only constraint is it must be
+    // possible to know the size of the `Env` at compile time (i.e it has to
+    // implement the `Sized` trait).
+    // The Env is then accessed using `data()` or `data_mut()` method.
+    #[derive(Clone)]
+    struct Env {
+        counter: Arc<Mutex<i32>>,
```
syrusakbary, line 81: Why this do need to be a mutex?

syrusakbary: Just read the comment

syrusakbary: But I'm not sure if I agree with it, we can discuss in a call

lib/api/src/sys/function_env.rs
```diff
@@ -117,6 +117,17 @@ impl<T: Send + 'static> FunctionEnvMut<'_, T> {
             func_env: self.func_env.clone(),
         }
     }
+
+    /// Borrows a new mutable reference of both the attached Store and host state
+    pub fn data_and_store_mut(&mut self) -> (&mut T, StoreMut) {
+        let data = self.func_env.as_mut(&mut self.store_mut) as *mut T;
+        // telling the borrow check to close his eyes here
```
theduke, line 124: After reviewing the API, I don't see any other (public) way  to get access to the `FunctionEnv` in `StoreInner`, so this should be safe from a public API standpoint. 

Of course internally we could still violate this. 

A thing I thought about yesterday : we could put the FunctionEnvs behind a `RefCell` to make this a bit more ssafer, but that would come with  some overhead.

But otherwise I think this is a not ideal, but acceptable use of unsafe.

examples/imports_function_env_global.rs
ptitSeb: It's based on the `import_function_env` example, it's really mostly copy/paste. But we can have a call about the examples for sure.

lib/api/src/sys/function_env.rs
syrusakbary: Yeah, agreed

syrusakbary (approved):

PR #3613, opened 2023-02-23
Add WCGI to wasmer run

Michael-F-Bryan:
> **Do Not Merge:** This PR branched off #3599 while we were waiting for #3422 to be merged. Once #3599 is merged, we can rebase onto `master` and get ready to merge this PR.

This adds the WCGI runner introduced in #3599 to the `wasmer` CLI.


lib/wcgi-runner/src/builder.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    /// Create a [`Runner`] that executes a WEBC file.
+    ///
+    /// If [`Builder::program`] was set, this will look for the command with
+    /// that name in the WEBC file. Otherwise, it will fall back to the WEBC
+    /// file's default entrypoint.
+    ///
+    /// This will infer the [`CgiDialect`] from the WEBC file's metadata
+    pub fn build_webc(self, webc: impl Into<Bytes>) -> Result<Runner, Error> {
```
syrusakbary, line 151: I don't think `build_webc` is a good name. Probably `use_webc` or `from_webc` are way better names

syrusakbary: Also, you may want to provide a command to execute, rather than infer it. And get the webc directly instead of the bytes.

That way there are two steps:
1. parsing the webc
2. detecting the command to run and what kind it is
3. Run the runner

lib/wcgi-runner/examples/wcgi-runner.rs
```diff
@@ -0,0 +1,111 @@
...
+        mapped_dirs,
+    } = Args::parse();
+
+    // Hack to initialize the global shared tokio task manager handle.
+    // Prevents "cannot drop runtime in async context" errors, because
+    // the default will be initialized to the current tokio context.
+    let rt = wasmer_wasi::runtime::task_manager::tokio::TokioTaskManager::default();
+
+    let mut builder = Runner::builder()
+        .map_dirs(mapped_dirs)
```
syrusakbary, line 29: I'd put here: with_wasi_config(CONFIG). Because that way it can be reused in the future with the WASI runner

lib/wcgi-runner/src/module_loader/cached.rs
```diff
@@ -0,0 +1,48 @@
...
+            cached: Mutex::new(None),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl<L: ModuleLoader> ModuleLoader for Cached<L> {
+    async fn load(&self, ctx: ModuleLoaderContext<'_>) -> Result<LoadedModule, Error> {
+        let mut cached = self.cached.lock().await;
+
```
syrusakbary, line 30: Why do we need this exactly? We already have the `wasmer-cache` crate that should handle this

Michael-F-Bryan: Regardless of whether we use a `Mutex<Option<LoadedModule>>` or a `wasmer_cache::FileSystemCache`, we still need to have some sort of struct which can be used as a `ModuleLoader`.

I used a mutex-based solution just because it's trivial to write and was good enough to get things implemented.

lib/wcgi-runner/src/builder.rs
Michael-F-Bryan: I deliberately named it that way so we can follow the `Runner::builder().foo().bar().build()` pattern. Using "use" or "from" would be very confusing. 

In this case, we have multiple ways of constructing the final runner (e.g. a WEBC file, a WebAssembly binary, or the filename for something on disk), so there are multiple `build_xxx()` methods.

lib/wcgi-runner/src/module_loader/cached.rs
syrusakbary: But the `wasmer_cache::FileSystemCache` is already provided by the CLI, right?
I don't think the caching mechanism should be defined in the runner (but plugged into it).

I think we might be doing an early optimization by having the `wcgi-runner` standalone instead of being first integrated on the cli first (which I think might be simpler)

lib/wcgi-runner/src/builder.rs
syrusakbary: I see the point that the `build_webc` actually returns a Runner (which I didn't realize before), so I kind of agree on that naming if we want to return a runner.

However, I think the webc shall be parsed beforehand and then chose the command to run.
For example, let's say you have a package with two commands:
* A (WASI command)
* B (WCGI command)

You don't know which runner you are going to run, until you have parsed the webc package. And at that point you already have the config and such. So passing a stream of bytes to parse again the webc seems wrong

Michael-F-Bryan:
After a bit of discussion today on Slack, I'm going to close this and move all changes to #3599. We need to redesign how the WCGI runner works because the current design doesn't match how we want `wasmer run` to work.

Issue #3578, opened 2023-02-07
Support for Wasm Coredump


xtuc:
### Motivation

When the Wasm instance traps, it's sometimes difficult to understand what happened. Post-mortem debugging using coredumps (which is extensively used in native environment) would be helpful for investigating and fixing crashes.

Demo from a developer's point of view: https://github.com/xtuc/wasmgdb/wiki/Demo.

### Proposed solution

Implement Wasm coredumps as specified by https://github.com/WebAssembly/tool-conventions/blob/main/Coredump.md. Note that the spec is early and subject to changes. Feedback very welcome!

### Additional context

Wasm coredump is especially useful for serverless environment where production binaries are stripped and/or have access to limited logging.

syrusakbary:
Hey @xtuc , thanks for opening the issue.

We would love to support wasm coredumps, but we are a bit constrained on resources.
However, if you would like to help us on that quest we would be more than welcome to accept any improvements/patches towards that goal!

xtuc:
@syrusakbary  I will give it a try. I believe I can use the [Backtrace](https://docs.rs/backtrace/0.3.67/backtrace/struct.Backtrace.html) information when a [RuntimeError](https://docs.rs/wasmer/latest/wasmer/struct.RuntimeError.html) occurs.
This should give us a initial / basic coredump implementation.

Please free to send any other pointers if you have something in mind.

syrusakbary:
Awesome, I believe RuntimeError and backtrace info as you mentioned should make wasm coredump possible.

If you have any questions upon implementing things please let us know on this thread and we will assist!

xtuc:
@syrusakbary  I opened https://github.com/wasmerio/wasmer/pull/3626 for adding basic coredump generation using the existing APIs. 

It's inpired by the change I did for wasmtime and wrote some documentation: https://github.com/bytecodealliance/wasmtime/blob/db931d7a93208738b8106e8d8c1a4afc8efaca6d/docs/examples-coredump.md.

xtuc:
Basic coredump generation has been merged (thanks!).

Now, to have the complete debugger experience, we need to collect the following information:
- Wasm locals of each stack frames; when discussing with wasmtime and having a look at the code myself, support in cranelift is going to take a significant effort. Would the implementation in the LLVM backend be easier for wasmer?
- Snapshot the Wasm instance memory; sounds relatively easy.

syrusakbary:
[...]

PR #3626, opened 2023-02-24
add basic coredump generation

xtuc:
This change adds a basic coredump generation after a WebAssembly trap was entered. The coredump includes rudimentary stack / process debugging information.

A new CLI argument is added to enable coredump generation:

```
wasmer --coredump-on-trap=/path/to/coredump/file module.wasm
```

Refs https://github.com/wasmerio/wasmer/issues/3578

### DIFF
--- a/lib/cli/src/commands/run.rs
+++ b/lib/cli/src/commands/run.rs
@@ -8,6 +8,8 @@ use crate::suggestions::suggest_function_exports;
 use crate::warning;
 use anyhow::{anyhow, Context, Result};
 use clap::Parser;
+use std::fs::File;
+use std::io::Write;
 use std::ops::Deref;
 use std::path::PathBuf;
 #[cfg(feature = "cache")]
@@ -89,6 +91,10 @@ pub struct RunWithoutFile {
     #[clap(long = "verbose")]
     pub(crate) verbose: Option<u8>,
 
+    /// Enable coredump generation after a WebAssembly trap.
+    #[clap(name = "COREDUMP PATH", long = "coredump-on-trap", parse(from_os_str))]
+    coredump_on_trap: Option<PathBuf>,
+
     /// Application arguments
     #[clap(value_name = "ARGS")]
     pub(crate) args: Vec<String>,
@@ -154,7 +160,7 @@ impl RunWithPathBuf {
         if self.debug {
             logging::set_up_logging(self_clone.verbose.unwrap_or(0)).unwrap();
         }
-        self_clone.inner_execute().with_context(|| {
+        let invoke_res = self_clone.inner_execute().with_context(|| {
             format!(
                 "failed to run `{}`{}",
                 self_clone.path.display(),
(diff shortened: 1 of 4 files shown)


syrusakbary:
WOW. The PR ended being way simpler than I was imagining. I'll review it on Monday with @ptitSeb

lib/cli/src/commands/run.rs
```diff
@@ -89,6 +91,10 @@ pub struct RunWithoutFile {
     #[clap(long = "verbose")]
     pub(crate) verbose: Option<u8>,
 
+    /// Enable coredump generation after a WebAssembly trap.
+    #[clap(long = "coredump-on-trap", value_name = "PATH")]
+    coredump_on_trap: Option<String>,
```
syrusakbary, line 96: This should be a PathBuf.

```suggestion
    #[clap(name = "COREDUMP PATH", long = "coredump-on-trap", parse(from_os_str))]
    coredump_on_trap: Option<PathBuf>,
```

xtuc: Fixed, thanks!

lib/cli/src/commands/run.rs
```diff
@@ -164,7 +170,23 @@ impl RunWithPathBuf {
...
+                    Err(err)
+                } else {
+                    Err(err.context(format!("core dumped at {}", coredump_path.display())))
+                }
+            } else {
+                Err(err)
+            }
+        } else {
+            invoke_res
+        }
```
syrusakbary, line 189: This can be simplified a bit using `map_err`. That way we will not need to save the result `invoke_res` and then return it

syrusakbary:
Hey @xtuc, could you also add a documentation file on how to use wasm-coredump with wasmer? (end to end)

Thanks!

xtuc:
@syrusakbary  yes sure. This is what I wrote for wasmtime: https://github.com/bytecodealliance/wasmtime/blob/db931d7a93208738b8106e8d8c1a4afc8efaca6d/docs/examples-coredump.md. Where's the best place to write such a documentation for wasmer?

lib/cli/src/commands/run.rs
xtuc: I'm not so fan of this approach. To me, `map_err` should only be used to map/cast the error into another error, not for any I/O or application logic.

That being said, I'm fine doing that change if you think it's better.

syrusakbary: I tend to think the code will be a bit simpler, but I'm not strong either way.
Let's try to chime in more opinions to see what the other maintainers prefer!

syrusakbary:
> Where's the best place to write such a documentation for wasmer?

Probably having that markdown doc inside the `docs/en` folder should be good enough!

xtuc:
@syrusakbary  added some docs.

lib/cli/src/commands/run.rs
Michael-F-Bryan: I think it should be fine for now. 

We'll be refactoring the `run.rs` module at some point to pay off technical debt, so things may get shuffled around anyway.

syrusakbary (approved):

syrusakbary:
Looks great, merging.

PR #3627, opened 2023-02-25
wasi: Remove vbus leftovers and improve task joining

### LINKED ISSUE #3625: WASIX: Vbus and Thread Joining Cleanup
The WASIX PR #3422  has types and functionality of a former `vbus` crate.

This functionality was largely removed, but the leftovers need to be cleaned up.

The `BusSpawnedProcessJoin` handle which is used for watching and handling thread/process results also needs to be simplified and cleaned up.

Part of #3619

theduke:
* A new TaskStatus is introduced that specifies the current state of a task
  (usually a process or thread)
* Status tracking is refactored to use tokio::sync::watch instead of a
  homegrown solution which combined channels and mutexes
* The task status now not just tracks an exit code, but also the ocurred error
  This helps consumers of the task status to know that and error
  ocurred, and which one, which helps debugging / logging and potential
  recovery
* Remove the unused VBUS crate leftover types and traits
  Almost all gone, except for the VirtualBusError
* Move SignalHandlerAbi types to os::signal

Closes #3625


theduke:
Per @john-sharratt s request I ensured that `dash` still works correctly.
`bash` is broken in multiple ways, but that's also true on master, no change here.

syrusakbary (approved):

PR #3628, opened 2023-02-25
Registry: Implement App Publishing and Token Generation

theduke:
Does a little cleanup on the registry and adds a new RegistryClient that 
has functions for publishing an app and for creating a deployment token.

- refactor(registry): Clean up the graphql code structure
- docs(registry): Add minimal README for the registry
- chore(registry): Format GraphQL files with prettier
- registry: Add a RegistryClient
- feat(registry): Add Deploy app publishing and token generation

# Description


lib/registry/src/graphql/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    Ok(())
+}
+
+pub fn execute_query_modifier_inner<R, V, F>(
+    registry_url: &str,
+    login_token: &str,
+    query: &QueryBody<V>,
+    timeout: Option<Duration>,
+    form_modifier: F,
```
syrusakbary, line 87: Nice I like this

lib/registry/src/client.rs
```diff
@@ -0,0 +1,146 @@
...
+            .context("Query did not return data")?
+            .deploy_app_version;
+        let app = version.app.context("Query did not return expected data")?;
+
+        Ok(PublishDeployAppOutput {
+            app_id: app.id,
+            app_name: app.name,
+            version_id: version.id,
+            version_name: version.version,
+            owner_name: app.owner.global_name,
```
syrusakbary, line 143: I think this two functions should live outside of the client, as they are just executing mutations but not affecting the global state, while also being domain specific

lib/registry/src/graphql/mod.rs
theduke: That's actually old code, I just moved it from graphql.rs to graphql/mod.rs.

There is a much cleaner way to execute the `graphql-client` queries, no need for form data.
The new execution logic is in client.rs ( fn execute(), fn execute_checked() )

lib/registry/src/client.rs
theduke: I usually write API clients to be one struct with all the various query methods on the struct instead of separate functions. 

Both designs are valid, but the former is usually more convenient documentation and usability wise, because with all the queries as separate functions things get more boiler-plate heavy and browsing the documentation the available  queries is more cumbersome.

```rust
use registry::Client;

let client = Client::from_env()?;
let user = client.current_user().await?;
let app = client.publish_deploy_app(Vars{ ...}).await?;
let token = client.generate_token(&app.version_id).await?;
```

vs 

```rust
use registry::{
  Client,
  queries,
};

let client = Client::from_env()?;
let user = queries::current_user(&client).await?;
let app = queries::publish_deploy_app(&client, Vars{ ...}).await?;
let token = queries::generate_token(&client, &app.version_id).await?;
```

theduke: I can be convinced otherwise if you feel strongly about this, but I do usually prefer the first variant.

syrusakbary: I've reflected a bit on this and I still believe is good to move them out of the client.
Otherwise, the client then starts being responsive on how to do certain queries (but not all) and it's hard to draw the line between when a query should live in the client and when it shouldn't.

Not only that, the second one is more composable frees the concept of queries outside of the client, while the former doesn't

theduke: @syrusakbary the naming here is a bit difficult, I thought about also using public `queries` and `mutations` submodules, but users of the crate shouldn't have to care about GraphQL details like the queries vs mutations split.

They could also just be root exports, but that's currently suboptimal because of all the other stuff in the crate.

So my compromise was putting them under `api`. 

Happy about suggestions though.

syrusakbary (approved):

PR #3629, opened 2023-02-27
Webc Inheritance Fixes

theduke:
- Fixed an issue where packages without commands were not loading the webc file system because it was package specific, for webc inheritence this does not make sense
- Removed some trace info which is bloating the logs
- Added some debug information for packages that are fetched

This splits out the webc related fixes from #3611.
The command additions of that PR will be added separately.


ptitSeb (approved):
Looks good to me

PR #3635, opened 2023-03-01
Added some fixes for the browser version to work again

john-sharratt:
Modules and Memory can not be shared between threads in rust and must be passed between threads using javascript instead

### DIFF
--- a/lib/api/src/js/store.rs
+++ b/lib/api/src/js/store.rs
@@ -309,14 +309,26 @@ mod objects {
             self.globals.iter()
         }
 
+        /// Return an vector of all globals and converted to u128
+        pub fn as_u128_globals(&self) -> Vec<u128> {
+            self.iter_globals()
+                .map(|v| v.global.value().as_f64().unwrap() as u128)
+                .collect()
+        }
+
         /// Set a global, at index idx. Will panic if idx is out of range
         /// Safety: the caller should check taht the raw value is compatible
         /// with destination VMGlobal type
-        pub fn set_global_unchecked(&self, idx: usize, val: u128) {
+        pub fn set_global_unchecked(&self, idx: usize, new_val: u128) {
             assert!(idx < self.globals.len());
 
-            let value = JsValue::from(val);
-            self.globals[idx].global.set_value(&value);
+            let g = &self.globals[idx].global;
+            let cur_val = g.value().as_f64().unwrap();
+            let new_val = new_val as f64;
+            if cur_val != new_val {
+                let new_value = JsValue::from(new_val);
+                g.set_value(&new_value);
+            }
         }
     }
 
--- a/lib/vm/src/store.rs
+++ b/lib/vm/src/store.rs
@@ -108,6 +108,13 @@ impl StoreObjects {
         self.globals.iter()
     }
 
+    /// Return an immutable iterator over all globals and convert them to u128
+    pub fn as_u128_globals(&self) -> Vec<u128> {
+        self.iter_globals()
+            .map(|v| unsafe { v.vmglobal().as_ref().val.u128 })
+            .collect()
+    }
+
     /// Set a global, at index idx. Will panic if idx is out of range
     /// Safety: the caller should check taht the raw value is compatible
     /// with destination VMGlobal type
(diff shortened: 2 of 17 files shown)


lib/wasi/src/os/console/mod.rs
```diff
@@ -222,14 +222,19 @@ impl Console {
...
 
-        // TODO: fetching dependencies should be moved to the builder!
-        // if let Err(err) = env.uses(self.uses.clone()) {
-        //     tasks.block_on(async {
-        //         let _ = self.runtime.stderr(format!("{}\r\n", err).as_bytes()).await;
-        //     });
-        //     tracing::debug!("failed to load used dependency - {}", err);
-        //     return Err(crate::vbus::VirtualBusError::BadRequest);
-        // }
+        if let Err(err) = env.uses(self.uses.clone()) {
```
theduke, line 225:
```suggestion
        // TODO: fetching dependencies should be moved to the builder!
        if let Err(err) = env.uses(self.uses.clone()) {
```

lib/wasi/src/runtime/task_manager/mod.rs
```diff
@@ -67,7 +67,13 @@ pub trait VirtualTaskManager: std::fmt::Debug + Send + Sync + 'static {
     /// Starts an asynchronous task will will run on a dedicated thread
     /// pulled from the worker pool that has a stateful thread local variable
     /// It is ok for this task to block execution and any async futures within its scope
-    fn task_wasm(&self, task: Box<dyn FnOnce() + Send + 'static>) -> Result<(), WasiThreadError>;
+    fn task_wasm(
```
theduke, line 70: A comment why this method needs to look like this would be very helpful.

lib/wasi/src/bin_factory/module_cache.rs
```diff
@@ -187,6 +187,7 @@ impl ModuleCache {
     pub fn get_compiled_module(
         &self,
         #[cfg(feature = "sys")] engine: &impl wasmer::AsEngineRef,
+        #[cfg(feature = "js")] engine: &impl wasmer::AsStoreRef,
```
theduke, line 190: After the unification PR lands we could to drop this feature flag completely because JS now has an engine.

Not for this PR, but: JS should just use a completely different cache, because the requirements are different.

theduke (changes requested):

theduke:
LGTM apart from the globals accessor on the Store - needs to use `iter_globals`

lib/wasi/src/bin_factory/module_cache.rs
syrusakbary: Yeah, I completely removed this in the unification + this PR

lib/api/src/js/store.rs
```diff
@@ -309,14 +309,26 @@ mod objects {
             self.globals.iter()
         }
 
+        /// Return an vector of all globals and converted to u128
+        pub fn as_u128_globals(&self) -> Vec<u128> {
```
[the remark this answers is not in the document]
syrusakbary: I will correct this apart

lib/wasi/src/os/console/mod.rs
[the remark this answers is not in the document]
syrusakbary:
```suggestion
        // TODO: the Console only makes sense in the context of SSH and the terminal.
        // We should make this just take a WasiBuilder and the console related configs
        // and not add so much custom logic in here.
        if let Err(err) = env.uses(self.uses.clone()) {
```

lib/api/src/js/store.rs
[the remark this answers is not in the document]
theduke: To be fixed later.

lib/wasi/src/bin_factory/module_cache.rs
theduke: To be fixed with the unification.

lib/vm/src/store.rs
```diff
@@ -108,6 +108,13 @@ impl StoreObjects {
         self.globals.iter()
     }
 
+    /// Return an vector of all globals and converted to u128
+    pub fn as_u128_globals(&self) -> Vec<u128> {
```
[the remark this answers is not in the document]
theduke: To be fixed in the unification.

theduke (approved):

Issue #3615, opened 2023-02-23
Fix Disabled create-exe Integration Tests


theduke:
During the merge of #3422 some of the create-exe tests were disabled.

We need to determine the cause , fix it and re-enable the tests.

Disabled tests:

* create_exe_works_multi_command_args_handling
* test_cross_compile_python_windows
* create_exe_with_object_input_serialized
* create_exe_serialized_works

@ptitSeb

ptitSeb:
Fixed with #3636

PR #3636, opened 2023-03-01
Removed choice for object-format in create-exe and create-obj commands

ptitSeb:
Also removed the related tests.
There was no doc to change as the `--object-format` parameter was not documented.

Yhis also address #3615

### DIFF
--- a/lib/cli/src/commands/create_exe.rs
+++ b/lib/cli/src/commands/create_exe.rs
@@ -2,7 +2,6 @@
 
 use self::utils::normalize_atom_name;
 
-use super::ObjectFormat;
 use crate::common::normalize_path;
 use crate::store::CompilerOptions;
 use anyhow::{Context, Result};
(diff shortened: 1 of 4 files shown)


lib/cli/src/commands/create_exe.rs
```diff
@@ -433,21 +412,18 @@ pub(super) fn compile_pirita_into_directory(
...
+                )
+            })?;
 
-                Some(target_dir.join("include").join(format!(
-                    "static_defs_{}.h",
-                    utils::normalize_atom_name(&atom_name)
-                )))
-            }
-            ObjectFormat::Serialized => None,
+            Some(target_dir.join("include").join(format!(
```
theduke, line 423: Minor nitpick: this is a big expression for a `Some(_)`, I'd move that out to a variable.

theduke (approved):
One optional minor nitpick, but LGTM.

That's a diff I can get behind!

PR #3643, opened 2023-03-05
fix(wasi): Memory leak due to cyclical WasiControlPlane references

### LINKED ISSUE #3642: wasi: Memory Leak due to Cyclical WasiControlPlane References
`WasiProcess` holds `WasiControlPlane` which holds all `WasiProcess`es...

theduke:
Fixed by WasiProcess just storing a Weak<> handle to the control plane.

Note: This could/should also be changed so that the WasiProcess does not
have a handle to the control plane at all, but that requires some
slightly bigger refactoring that also changes or removes the SignalHandlerAbi

Closes #3642

### DIFF
--- a/lib/wasi/src/os/task/process.rs
+++ b/lib/wasi/src/os/task/process.rs
@@ -17,13 +17,12 @@ use wasmer_wasi_types::{
 };
 
 use crate::{
-    os::task::{control_plane::WasiControlPlane, signal::WasiSignalInterval},
-    syscalls::platform_clock_time_get,
-    WasiThread, WasiThreadHandle, WasiThreadId,
+    os::task::signal::WasiSignalInterval, syscalls::platform_clock_time_get, WasiThread,
+    WasiThreadHandle, WasiThreadId,
 };
 
 use super::{
-    control_plane::ControlPlaneError,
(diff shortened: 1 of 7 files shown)


lib/wasi/src/os/task/process.rs
```diff
@@ -232,7 +231,7 @@ impl WasiProcess {
             if self.waiting.load(Ordering::Acquire) > 0 {
                 let mut triggered = false;
                 for pid in children.iter() {
-                    if let Some(process) = self.compute.get_process(*pid) {
+                    if let Some(process) = self.compute.must_upgrade().get_process(*pid) {
```
ptitSeb, line 234: is that ok to use the `must_upgrade` in an `if`? What happens if the if is not true?

theduke: It will panic. 

Which should be perfectly fine, because this can only happen if you completely misuse the API.

`WasiControlPlane` is stored on `WasiEnv`, so this can never fail as long as the `WasiEnv` still exists. 
(Unless someone was to *replace* the WasiControlPlane , but that would also be serious misuse and worthy of a panic. )

ptitSeb (approved):

PR #3644, opened 2023-03-05
Improvements to the tracing and logging in wasmer

john-sharratt:
- Now using the fmt subscriber for improved formatting
- Using spans on all syscalls
- Added recorded fields on the spans instead trace and debug macros
- Adding timing on all syscall

### DIFF
--- a/lib/cli/src/logging.rs
+++ b/lib/cli/src/logging.rs
@@ -1,7 +1,7 @@
 //! Logging functions for the debug feature.
 
 /// Subroutine to instantiate the loggers
-#[cfg(feature = "tracing")]
+#[cfg(any(feature = "tracing", feature = "debug"))]
 pub fn set_up_logging(verbose: u8) -> Result<(), String> {
     use tracing_subscriber::prelude::*;
     use tracing_subscriber::{fmt, EnvFilter};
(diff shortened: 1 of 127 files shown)


syrusakbary (approved):
This looks great. I wished we did this earlier. Good work!

lib/cli/src/logging.rs
```diff
@@ -1,14 +1,44 @@
 //! Logging functions for the debug feature.
-use crate::utils::wasmer_should_print_color;
-use anyhow::Result;
-use fern::colors::{Color, ColoredLevelConfig};
-use std::time;
 
-/// The debug level
-pub type DebugLevel = log::LevelFilter;
+/// Subroutine to instantiate the loggers
+#[cfg(feature = "tracing")]
```
theduke, line 4: I think it's fine to just switch to tracing full stop and remove the old log system.

theduke (commented):
Very nice @john-sharratt !

One concern: since often the syscall args are in the spans now this has the potential to produce extremely noisy logs, we might want to move some of those back to just logging them in a first log message inside the syscall.

But we can just change that as we go along.

lib/cli/src/logging.rs
john-sharratt: the spans all have `skip-all` turned on by default - which means the number of fields have to be explicitly added and won't grow, but it might be an idea to add a feature flag for super noisy syscalls.

to be honest i think we should do the opposite though as I commended out the `thread_local_get` syscall debug and trace messages because they were so noisy... but then later (as in a few days ago) when I put the spans on i noticed there were millions of them now because of `tokio` going nuts with locals.

i think we should leave it noisy when in tracing mode, make it "useful" in debug level.... and optimize the code so that less syscalls are made. i mean the syscalls will be killing performance anyway

john-sharratt:
> Very nice @john-sharratt !
> 
> One concern: since often the syscall args are in the spans now this has the potential to produce extremely noisy logs, we might want to move some of those back to just logging them in a first log message inside the syscall.
> 
> But we can just change that as we go along.

the spans all have skip-all turned on by default - which means the number of fields have to be explicitly added and won't grow, but it might be an idea to add a feature flag for super noisy syscalls.

to be honest i think we should do the opposite though as I commended out the thread_local_get syscall debug and trace messages because they were so noisy... but then later (as in a few days ago) when I put the spans on i noticed there were millions of them now because of tokio going nuts with locals.

i think we should leave it noisy when in tracing mode, make it "useful" in debug level.... and optimize the code so that less syscalls are made. i mean the syscalls will be killing performance anyway

theduke:
@john-sharratt small misunderstanding:

All syscalls should be instrumented, totally.

But we can chose between either including the fields in `instrument()`, which means they will be re-printed on each log message inside the created span, or just logging the values once with a manual `trace!()` which means they won't be in the span. 

And I think we will mostly want to do the latter, except for very small stuff like maybe the `fd`.

john-sharratt:
> @john-sharratt small misunderstanding:
> 
> All syscalls should be instrumented, totally.
> 
> But we can chose between either including the fields in `instrument()`, which means they will be re-printed on each log message inside the created span, or just logging the values once with a manual `trace!()` which means they won't be in the span.
> 
> And I think we will mostly want to do the latter, except for very small stuff like maybe the `fd`.

Ah I understand - yeah that makes sense - we'll work towards that approach going forward

PR #3648, opened 2023-03-06
Fix CI and llvm detection

ptitSeb:
CI Test will only trigger on some specific branch now:
 - `main`
 - `v3.0.x`
 - `v3.1.x`
 - `with_ci-.*`

PR should still test also.

Also, detection of LLVM is improved when only `llvm-config` is available


theduke (approved):
Always nice to move stuff from the Github specific code to the makefile.

Branch config changes LGTM as well.

:+1:

PR #3649, opened 2023-03-06
Move the wcgi-runner into Wasmer (second attempt)

Michael-F-Bryan:
This is my second attempt at #3599.


Michael-F-Bryan:
We no longer need this second PR because I managed to find the issue in the original one.

PR #3650, opened 2023-03-06
Wasmer run-unstable

### LINKED ISSUE #3514: Respect main_args from the wasmer.toml file when running commands
Currently the `wasmer.toml` file has a `main_args` field for setting default arguments for the runner, however those `main_args` are not being respected when actually running the .wasm file (they're not passed in).

Current:

```
wasmer run ciuser/customlambda@0.2.27 customlambda.py 55
```

Goal:

```
wasmer run ciuser/customlambda@0.2.27 55
```

The necessary changes for this are already implemented in https://github.com/wasmerio/wasmer/pull/3464/commits/a3d3550ea841420ca74b59364c3585fb7f42fb31#diff-bc54a1b8f8ed51b1b4d84d6e94de7a3cb3a059311cac846e470dc28f0a136218R141-R146 we just need to port them to master.

Michael-F-Bryan:
This implements a second version of the `wasmer run` command to make the flow easier to follow and use WEBC runners.

In particular, it...

- Introduces a `wasmer run-unstable` sub-command which will eventually replace `wasmer run` and add a bunch of integration tests for it
- Deprecates several feature flags in the CLI (`cache`, `cache-blake3-pure`, `debug`, `wasi`, `emscripten`, `webc_runner`), making sure their functionality is always included and the flags themselves are no-ops
- Introduces a `wasmer_cli::WasmerHome` abstraction for interacting with the `$WASMER_DIR` directory and our integrating it with our caches
- Introduces an `impl<D> Filesystem for D where D: Deref<Target = impl Filesystem>` so `Box<dyn FileSystem>` and friends implement the `FileSystem` trait
- fixes a bug in `wasmer_vfs::WebcFileSystem` where you would OOM or have an infinite loop because reads would never increment the file's cursor
- Propagate environment variables and mapped directories to the WASI runner
- Wire up logging and error-handling middleware to the server started by the WCGI runner

We're calling it `wasmer run-unstable` instead of overwriting the `wasmer run` command for now until we get a chance to do more testing.

Fixes #3514.


lib/c-api/Cargo.toml
```diff
@@ -32,7 +32,7 @@ wasmer-middlewares = { version = "=3.2.0-alpha.1", path = "../middlewares", opti
 wasmer-wasi = { version = "=3.2.0-alpha.1", path = "../wasi", features = ["host-fs", "host-vnet"], optional = true }
 wasmer-types = { version = "=3.2.0-alpha.1", path = "../types" }
 wasmer-vfs = { version = "=3.2.0-alpha.1", path = "../vfs", optional = true, default-features = false, features = ["static-fs"] }
-webc = { version = "5.0.0-rc.5", optional = true }
+webc = { version = "4.0.0", optional = true }
```
syrusakbary, line 35: Roll back?

tests/integration/cli/tests/run2.rs
```diff
@@ -0,0 +1,339 @@
...
+
+        assert.success().stdout(contains("Hello, World!"));
+    }
+
+    #[test]
+    fn quickjs_as_url() {
+        let assert = Command::new(get_wasmer_path())
+            .arg("run2")
+            .arg("https://wapm.io/saghul/quickjs")
+            .arg("--entrypoint=quickjs")
```
syrusakbary, line 227: What happens if we don't provide an entrypoint (and there is only one command / or the package has already an entrypoint defined)? Will things still work?

Michael-F-Bryan: Yep!

https://github.com/wasmerio/wasmer/blob/bca161cb441b769722b7d7ca1ddf4d30aa5332e8/lib/cli/src/commands/run2.rs#L271-L287

I've also added an extra test using WABT to show the message you get when you need to specify the entrypoint.

https://github.com/wasmerio/wasmer/blob/bca161cb441b769722b7d7ca1ddf4d30aa5332e8/tests/integration/cli/tests/run2.rs#L51-L60

syrusakbary (approved):

PR #3653, opened 2023-03-07
Dash fixes and pthreads

john-sharratt:
This merge brings in support for pthreads using the WASI threads spec and fixes a number of issues with `dash` WASIX


john-sharratt:
Opening new pull request

Issue #3647, opened 2023-03-06
singlepass can't emit ROR


sidenaio:
Hello, we are using wasmer 2.3.0 to run contracts in our blockchain. Some of our users are running their nodes on Raspberry Pi and facing the error: 
`Mar  6 09:39:55 pi-server idena-node[214248]: thread '<unnamed>' panicked at 'singlepass can't emit ROR S64 GPR(X9) Imm64(32) GPR(X9)', /cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-compiler-singlepass-2.3.0/s
rc/emitter_arm64.rs:1442:18`
I've found a similar  issue https://github.com/wasmerio/wasmer/issues/2959. But this fix is not presented in wasmer 2.x.
Do you plan to patch wasmer 2.x or we need to migrate to wasmer 3.x?

ptitSeb:
It would be better to migrate to 3.x indeed, as there is no plan to update 2.x release for now. Is that doable for you?

sidenaio:
If 2.x and 3.x are compiling and running the wasm in the same way,  it will be ok. If there's some difference then we need to do a fork.

ptitSeb:
The API as evolved significantly between 2.x and 3.x. So migration is not just plug'n play I'm afraid.

sidenaio:
API is not a problem. We only need a consensus between nodes while running wasm.

ptitSeb:
Ah, then, I don't know. It might be simpler.

I checked, and I still need to add the correct emiter to fix the issue on master, so I'll do that shortly.

sidenaio:
Ok, if you need the wasm that produces the error, you can find hex of wasm file [here](https://scan.idena.io/contract/0xa877f4632dff78f8b87f835379f844e260d0245d).

ptitSeb:
The fix is on the master now

PR #3654, opened 2023-03-07
[SINGLEPASS] Add more ROR emitter to ARM64 backend (for #3647)

ptitSeb:
Add more ROR emitter to ARM64 backend of Singlepass compiler, it was missing Imm64 handling for 64bits operations.

Should fix ticket #3647


syrusakbary (approved):
Looks good. Should we add a new test for this?

PR #3655, opened 2023-03-07
[CI] Manually setup LLVM because Custom build doesn't include llvm-config.exe

ptitSeb:
Manually setup LLVM because Custom build doesn't include llvm-config.exe


theduke (approved):
Let's hope it builds now...

PR #3659, opened 2023-03-07
Fixed building with with just the sys feature

ptitSeb:
Fix the build with
```bash
cargo build --no-default-features --features=sys --manifest-path=lib/api/Cargo.toml
```

(used in CI)


Michael-F-Bryan (approved):
I verified that this builds on my local machine, so it should be good to merge as soon as CI is green :+1:

PR #3663, opened 2023-03-07
Dash fixes and pthreads

john-sharratt:
This merge brings in support for pthreads using the WASI threads spec and fixes a number of issues with dash WASIX


john-sharratt:
Note the snapshot tests won't work until I regenerate them all - which I'll do now but be aware they are not backwards compatible

lib/wasi/src/runtime/mod.rs
```diff
@@ -57,13 +59,33 @@ where
...
+#[derive(Clone, Derivative)]
+#[derivative(Debug)]
 pub struct PluggableRuntimeImplementation {
     pub rt: Arc<dyn VirtualTaskManager>,
     pub networking: DynVirtualNetworking,
     pub http_client: Option<DynHttpClient>,
     #[cfg(feature = "sys")]
     pub engine: Option<wasmer::Engine>,
+    #[derivative(Debug = "ignore")]
+    pub tty: Arc<dyn TtyBridge + Send + Sync>,
```
theduke, line 88: Only WASIX environment want a TTY, WASI doesn't need it.

lib/wasi/src/runtime/mod.rs
```diff
@@ -97,13 +123,21 @@ impl PluggableRuntimeImplementation {
                 let http_client = None;
             }
         }
+        cfg_if::cfg_if! {
```
theduke, line 126: I went through quite a bit of work to remove `#[cfg()]` code like this from this generic type. 

Let's please keep this out of here.
Construction code like the in the CLI can set this.

lib/wasi/src/runtime/mod.rs
```diff
@@ -141,4 +175,8 @@ impl WasiRuntime for PluggableRuntimeImplementation {
     fn task_manager(&self) -> &Arc<dyn VirtualTaskManager> {
         &self.rt
     }
+
+    fn tty(&self) -> Option<&dyn TtyBridge> {
+        Some(self.tty.as_ref())
```
theduke, line 180: Keep it optional.

lib/wasi-types/Cargo.toml
```diff
@@ -28,6 +28,7 @@ cfg-if = "1.0.0"
 anyhow = "1.0.66"
 byteorder = "1.3"
 time = "0.2"
+tracing = "0.1"
```
theduke, line 31: Was this added by mistake?

I don't see it used, and it shouldn't be in this crate anyway.

lib/wasi/src/bin_factory/exec.rs
```diff
@@ -166,17 +166,17 @@ pub fn spawn_exec_module(
                     start
                         .call(&mut store, &[])
                         .map_err(WasiRuntimeError::from)
-                        .map(|_| 0)
+                        .map(|_| Errno::Success)
                 } else {
                     debug!("wasi[{}]::exec-failed: missing _start function", pid);
-                    Ok(Errno::Noexec as u32)
+                    Ok(Errno::Noexec)
```
theduke, line 172: General comment on the `u32 => Errno` code changes:

I initially wanted to say this is nice for the code, BUT exit codes are generic, any number is a valid exit code.
By using `Errno` (which doesn't have an `other(x)` variant we restrict this and conversion would fail at some point. 

But just in general `Errno` != exit code .

theduke (changes requested):
Mostly looks good to me, apart from some smaller requests and one larger issue (Errno).

lib/wasi/src/bin_factory/exec.rs
john-sharratt: Exit codes are unfortunately not generic when it comes to LIBC and besides if you look at the code (in practice) Errno == exit code. Either way we can't have them diverge without breaking backwards compatibility with LIBC. This was causing serious issues in `dash` and `bash`

lib/wasi/src/runtime/mod.rs
john-sharratt: I don't agree with this  as it outsources a code problem to a consumer, effectively the consumer has to determine if their binary will need TTY before executing it rather than the runtime handling that for them. The consequences of us supporting TTY by convention are trival on our side.

P.s. WASI did also need this for quite some time as I (along with others) had to create a hack for getting TTY in WASI by using the `/dev/tty` FD which has been upstreamed into create like the `password` crate which need to switch to none echo mode for password entering (quite common).

lib/wasi/src/runtime/mod.rs
john-sharratt: Fair enough - I'll look into how to get rid of this

lib/wasi-types/Cargo.toml
john-sharratt: fixed

lib/wasi/src/runtime/mod.rs
john-sharratt: its optional again

lib/wasi/src/runtime/mod.rs
john-sharratt: removed

lib/wasi/src/runtime/mod.rs
john-sharratt: I've made this a command line argument now however I default it to TTY attached rather than TTY detached (i.e. the command line argument is --no-tty)

john-sharratt:
> Note the snapshot tests won't work until I regenerate them all - which I'll do now but be aware they are not backwards compatible

Snapshot tests are updated but there is a strange cranelift error that is not understood

lib/wasi/src/bin_factory/exec.rs
theduke: Sure, the exit codes need to be correct for certain failure conditions to propagate properly.

But code is still three to return any exit code it likes. 
So won't this  break code that relies on specific / non-standard exit codes?

lib/wasi/src/runtime/mod.rs
theduke: Oh, small mis-scommunication: I agree that the CLI and SSH can totally always enable the TTY. 

I was just saying it wouldn't be necessary for many WCGI or even many WASIX tasks that don't have interactivity by nature.

lib/wasi/src/bin_factory/exec.rs
john-sharratt: The unknown exit codes will be converted into Errno::Unknown if they are used by `proc_join` otherwise they will remain u32's

lib/wasi/Cargo.toml
```diff
@@ -72,6 +70,7 @@ optional = true
 
 [target.'cfg(unix)'.dependencies]
 libc = { version = "^0.2", default-features = false }
+termios = { version = "0.3" }
```
theduke, line 73: This should still be optional, no?
We don't need it by default, just if we want host terminal support.

theduke (approved):

lib/wasi/Cargo.toml
john-sharratt: Well it can't be optional if we don't have a feature flag so its either one or the other

lib/wasi/src/bin_factory/exec.rs
john-sharratt: @theduke this is a good compromise I think - it gives the strong typing needed for `proc_join` to stay in sync but also allows for arbitary exit codes
https://github.com/wasmerio/wasmer/pull/3663/commits/a41bf19da05b2d859720db14b980055d382a13fc

syrusakbary (approved):

PR #3664, opened 2023-03-08
feat(wasi): Add CapabilityThreading with max thread count

theduke:
Extends the wasi Capabilities type with a setting for threading that
configures the maximum thread count.


Michael-F-Bryan (approved):
Looks like we're just propagating `max_threads` through the config to the `WasiControlPlane`. Makes sense :+1:

ptitSeb (approved):

Issue #3645, opened 2023-03-05
Threads are enabled by default, contradicting the documentation


TethysSvensson:
### Describe the bug

The [documentation](https://docs.rs/wasmer/3.1.1/wasmer/struct.Features.html#method.threads) on `Features` for v3.1.1 says that `threads` are disabled by default, however that is not the case.

### Steps to reproduce

The following code crashes:

```rust
assert_eq!(wasmer::Features::new().threads, false);
```

### Expected behavior

That the documentation matches the behavior.

syrusakbary:
Good call. We need to fix this on the docs

ptitSeb:
Need to check also the CLI to be able to disable threads.

syrusakbary:
@ptitSeb we should enable them by default now (replicating browsers behavior), so we just need to update the docs

PR #3665, opened 2023-03-08
Fix doc, threads are enabled by default now

ptitSeb:
This is to address ticket #3645

### DIFF
--- a/lib/cli/src/common.rs
+++ b/lib/cli/src/common.rs
@@ -13,9 +13,9 @@ pub struct WasmFeatures {
     #[clap(long = "enable-simd")]
     pub simd: bool,
 
-    /// Enable support for the threads proposal.
-    #[clap(long = "enable-threads")]
-    pub threads: bool,
+    /// Disable support for the threads proposal.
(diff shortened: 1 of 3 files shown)


lib/cli/src/common.rs
```diff
@@ -13,9 +13,9 @@ pub struct WasmFeatures {
     #[clap(long = "enable-simd")]
     pub simd: bool,
 
-    /// Enable support for the threads proposal.
-    #[clap(long = "enable-threads")]
-    pub threads: bool,
```
theduke, line 18: We probably don't want to break CLI users that use this flag in scripts etc.

We could leave an `enable-threads` flag that is a no-op, with a comment saying it's deprecated, doesn't do anything,  and mention  ---disable-threads.

theduke (commented):
Backwards compat suggestion.

lib/cli/src/common.rs
ptitSeb: ah yeah, true

theduke (approved):

PR #3666, opened 2023-03-08
refactor(wasi): Move capabilities into root submodule

theduke:
Just a small restructuring that moves the capabilities submodule from
state into the root lib.

Done because it is now exported in full rather than just partially from
state::, and because it is a more generic module that will probably be moved
out to a separate crate soon.


ptitSeb (approved):
Linter is not happy

PR #3676, opened 2023-03-13
Switch `wasmer_vfs::host_fs` tests over to using temp directories

Michael-F-Bryan:
The unit tests in `wasmer_vfs::host_fs` currently write to the test directory. 

This PR switches them over to using a `tempfile::TempDir` because a) it's cleaner, and more importantly, b) it means `cargo watch` won't peg the CPU to 100% because running the tests updates files inside the `src/` folder, which in turn triggers a re-run of `cargo watch`, which causes the crate to be recompiled and all the tests to start again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5217,6 +5217,7 @@ dependencies = [
  "libc",
  "pin-project-lite",
  "slab",
+ "tempfile",
  "thiserror",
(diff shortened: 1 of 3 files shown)


ptitSeb (approved):
Looks good to me

PR #3677, opened 2023-03-13
Overlay FileSystem

Michael-F-Bryan:
This adds a `wasmer_vfs::overlay_fs::OverlayFileSystem` which allows you to "overlay" multiple `wasmer_vfs::FileSystem` implementations.

As opposed to the `wasmer_vfs::union_fs::UnionFileSystem`, which allows you to mount disjointed filesystems, the `OverlayFileSystem` has a single primary filesystem and zero or more secondary filesystems, where a items from a filesystem with higher precedence will shadow items from a lower precedence filesystem.

We want this as part of #3650.


lib/vfs/src/overlay_fs.rs
```diff
@@ -0,0 +1,573 @@
...
+
+                for fs in self.0.filesystems() {
+                    f.entry(&fs);
+                }
+
+                f.finish()
+            }
+        }
+
+        f.debug_struct("FileSystem")
```
theduke, line 307:
```suggestion
        f.debug_struct("OverlayFileSystem")
```

lib/vfs/src/overlay_fs.rs
```diff
@@ -0,0 +1,573 @@
...
+///     std::mem::size_of_val(&fs),
+///     std::mem::size_of::<(MemFS, HostFS)>(),
+/// );
+/// ```
+///
+/// A more complex example is
+#[derive(Clone, PartialEq, Eq)]
+pub struct OverlayFileSystem<P, S> {
+    primary: P,
+    secondaries: S,
```
theduke, line 48: I feel like this is not quite the right layer of abstraction.

I would instead have this just take a single secondary filesystem, not a list.
And in turn have another wrapper that composes multiple read-only filesystems in order.  

That also allows implementing more efficient special implementations (like for example multiple read-only WebCs) without then taking the additional overhead that always assuming a list of secondaries introduces here.

lib/vfs/src/filesystems.rs
```diff
@@ -0,0 +1,106 @@
+use crate::FileSystem;
+
+/// A chain of one or more [`FileSystem`]s.
+pub trait FileSystems<'a>: 'a {
```
theduke, line 4: Is this primarily just here to avoid writing `IntoIterator` bounds on the `OverlayFs`?

Maybe there is a reason I'm missing, but seems like this could just be cut.

I'd rather not take this into the public API - we'll probably want to be conservative with this specific crate.

lib/vfs/src/overlay_fs.rs
Michael-F-Bryan: Isn't a "wrapper that composes multiple read-only filesystems in order" almost exactly what `OverlayFileSystem` is, though?

The only difference is that this implementation would allow the first item to be mutable, but otherwise you've got a bunch of filesystems that look like they are overlayed on top of each other, using a particular order to solve precedence of operations.

lib/vfs/src/filesystem_ext.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    fn is_file(&self, path: impl AsRef<Path>) -> bool {
+        match self.metadata(path.as_ref()) {
+            Ok(meta) => meta.is_file(),
+            Err(_) => false,
+        }
+    }
+
+    fn create_dir_all(&self, path: impl AsRef<Path>) -> Result<(), FsError> {
```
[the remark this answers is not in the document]
Michael-F-Bryan: That's fine with me. 

The main reason I introduced an extension trait in the first place was that doing anything with a `FileSystem` would require several lines of boilerplate that you'd copy/paste every time. That was especially annoying in tests where using 4 lines when 1 is necessary can make them hard to follow.

I don't mind if we put them all on the main trait instead of using an extension trait.

lib/vfs/src/overlay_fs.rs
syrusakbary: I disagree on that take @theduke. I think multiple read-only fs is the optimal approach, also because of the usability aspect. Your comment is still addressable as a subcase of OverlayFS but with only one read-only layer instead of multiple.
I'm not sure if I see the overhead that you are referring to. Can you expand on that?

This strategy of using mutliple read-only is also what other VFS implementations do:
* Rust's vfs: https://docs.rs/vfs/latest/vfs/impls/overlay/struct.OverlayFS.html
* JS memfs UnionFs: https://github.com/streamich/unionfs/blob/master/src/union.ts#L338-L349 (they are a bit more messy on their naming)

lib/vfs/src/filesystems.rs
Michael-F-Bryan: It also allows you to use tuples instead of forcing you to store a `Vec<Box<Dyn FileSystem>>` when the secondaries are homogeneous, which would introduce two extra levels of indirection. 

The "one read-write and a list of read-only" semantics are also what we decided we wanted when Syrus and I were discussing during our Planning & Alignment meeting.

lib/vfs/src/overlay_fs.rs
theduke: You can always layer another more optimized abstraction into the list with a small bit of overhead, but I can see quite a few more optimized implementations where the list is mostly noise.

I think looking at things like overlayfs and the Linux kernel in general would be educatioinal.

theduke: (in the long term, not for this PR, obviouslyl)

syrusakbary: To keep things lean, please lets keep using `OverlayFS`. In the future we'll make sure we are all aligned on how to abstract the fs (I'm a bit opinionated on this front as you can see, but it comes from previous experience)

syrusakbary (approved):
Once the comments are addressed (except of the one suggesting a different design for OverlayFS) and the tests pass, I think we should be good to go.

Also, side comment: I think we should make WebC read only for now (basically, panic/fail on any write operation on it, and we will remove the memory filesystem on it -> https://github.com/wasmerio/wasmer/blob/444d59685780e3061c762a6bf2f8f7529a69bea2/lib/vfs/src/webc_fs.rs#L24 )
If things fully work then I think we should be good to go! (note: we can do that as part of other PR if that's easier)

lib/vfs/src/filesystem_ext.rs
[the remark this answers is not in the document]
syrusakbary: Move this as a standalone function in the crate instead of the Trait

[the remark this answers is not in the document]
Michael-F-Bryan: I've moved the functions to a module called `ops`.

Not being able to use method syntax takes away a lot of the ergonomics gains the extension trait gave us, so I'm not going to expose the functions as part of the public API.

lib/vfs/src/overlay_fs.rs
Michael-F-Bryan: Fixed.

lib/vfs/src/overlay_fs.rs
```diff
@@ -0,0 +1,573 @@
...
+
+            Ok(ReadDir::new(entries))
+        } else {
+            Err(FsError::BaseNotDirectory)
+        }
+    }
+
+    fn create_dir(&self, path: &Path) -> Result<(), FsError> {
+        match self.primary.create_dir(path) {
+            Err(e) if should_continue(e) => {}
```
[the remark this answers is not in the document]
Michael-F-Bryan: I've inlined the `should_continue()` function.

PR #3679, opened 2023-03-14
Vectored io fix

john-sharratt:
This patch fixes a bug with vectored IO which was causing duplicate data to be written in edge cases that had multiple buffers and EAGAIN return values.


john-sharratt:
This seems to have pulled in more than I wanted. Closing

PR #3680, opened 2023-03-14
Fix for the vectored IO

john-sharratt:
This patch fixes a bug with vectored IO which was causing duplicate data to be written in edge cases that had multiple buffers and EAGAIN return values.


syrusakbary (approved):

theduke (approved):
lgtm

PR #3681, opened 2023-03-15
(feat) Ability to supply your own shared tokio runtime

john-sharratt:
Modified the shared tokio runtime so that they can be set before they are lazy loaded thus allowing the caller to pass in their own runtime


john-sharratt:
@syrusakbary @theduke lets try and close this pull request out please - its needed as to fix a stability issue with concurrent runtimes

theduke:
The downside to this is that it will always require a separate runtime now.

Most users would just invoke this from an async context without knowing the internals, and thus be forced into always having two separate runtimes active at once.

theduke:
Let's merge this for now regardless.

theduke (approved):

PR #3683, opened 2023-03-15
Renamed some crates and changed their version to 0.1.0

ptitSeb:
Renamed:
```
wasmer-wasi => wasmer-wasix 0.1.0
wasmer-wasi-types =>wasmer-wasix-types  0.1.0
wasmer-sys-utils 0.1.0
wasmer-vfs => virtfs 0.1.0
wasmer-vnet => virtnet 0.1.0
wasmer-wasi-local-networking  => virtnet-native  0.1.0
wasi-experimental-io-devices => wasix-experimental-io-devices 0.1.0
```


ptitSeb:
> Did you not rename the directories on purpose?
> 
> The repo would become very inconsistent and confusing without aligning them with the crate names.
> 
> _But_ this will be a _very annoying_ change though for all open PRs, so if we do this we please need to merge #3653 first, because that's a large PR, and rebasing this PR on top of that one will be much easier than the other way around.
> 
> Alternative is to first merge without renaming the directories, try to get all large PRs in and then do the dir rename as a second PR.

Yes, that was on purpose. I propose the crates that have ben rename to `wasix` instead of `wasi` have the folder renamed. I'm less sure about the other that already have a different folder name. But I can do it too.

How closed is the #3653 to be merged?

theduke:
I think @john-sharratt is still fixing some stuff. 

If we want to get this beta out quickly we can merge without renaming the directories and do that as a followup - but I do think we should get that PR in first either way because the conflicts would be annoying regardless. 

> I'm less sure about the other that already have a different folder name.

I actually see it almost the other way around - it's more important for `virtfs` / `virtnet`, etc, because those crates won't be changed back again. The ones that we want to eventually rename back to the old names could theoretically keep their old names. But I think it's better to rename them as well to keep the repo consistent.

ptitSeb:
I'm fine with renaming all the foler. It's just that it will be an anoying PR to handle for sure.

lib/cli/Cargo.toml
```diff
@@ -32,16 +32,16 @@ wasmer-compiler-singlepass = { version = "=3.2.0-alpha.1", path = "../compiler-s
...
+virtnet-native = { version = "0.1.0", path = "../wasi-local-networking", optional = true }
 wasmer-wast = { version = "=3.2.0-alpha.1", path = "../../tests/lib/wast", optional = true }
 wasmer-cache = { version = "=3.2.0-alpha.1", path = "../cache", optional = true }
 wasmer-types = { version = "=3.2.0-alpha.1", path = "../types", features = ["enable-serde"] }
 wasmer-registry = { version = "=4.0.0", path = "../registry" }
 wasmer-object = { version = "=3.2.0-alpha.1", path = "../object", optional = true }
-wasmer-vfs  = { version = "=3.2.0-alpha.1", path = "../vfs", default-features = false, features = ["host-fs"] }
-wasmer-vnet  = { version = "=3.2.0-alpha.1", path = "../vnet" }
+virtfs  = { version = "0.1.0", path = "../vfs", default-features = false, features = ["host-fs"] }
+virtnet  = { version = "0.1.0", path = "../vnet" }
```
syrusakbary, line 44: virtual-net

syrusakbary (approved):

PR #3687, opened 2023-03-16
[WIP] Master with fixes

syrusakbary:
This PR aims to do a bit of cleanup of Wasmer codebase:

* [x] Fixed `sys`/`js` features in wasmer when used with dependencies
* [x] Use singlepass if available in macOS M1
* [x] Removed unused types from WASI
* [ ] Remove dependency of Wasmer/sys from WASI


syrusakbary:
This PR is mainly improving WASIX to not make use of VM internals, so it can be used easily by any runtime configuration.
This is required to improve some other things such as different runtime engine support and so on, and the rest of changes are trivial.

I'll proceed to merge since all tests are already passing and this is blocking other work

PR #3691, opened 2023-03-20
Don't error out on webc/wasi run when exiting normaly

ptitSeb:
Avoid the webc runner to print an error when WASI exit normaly.

### DIFF
--- a/lib/wasi/src/runners/wasi.rs
+++ b/lib/wasi/src/runners/wasi.rs
@@ -95,7 +95,12 @@ impl crate::runners::Runner for WasiRunner {
             builder.set_runtime(Arc::new(rt));
         }
 
-        builder.run(module)?;
+        let res = builder.run(module);
+        match res {
+            Ok(()) => Ok(()),
+            Err(crate::WasiRuntimeError::Wasi(crate::WasiError::Exit(_))) => Ok(()),
+            Err(e) => Err(e),
+        }?;
 
         Ok(())
     }


lib/wasi/src/runners/wasi.rs
```diff
@@ -95,7 +95,12 @@ impl crate::runners::Runner for WasiRunner {
             builder.set_runtime(Arc::new(rt));
         }
 
-        builder.run(module)?;
+        let res = builder.run(module);
+        match res {
+            Ok(()) => Ok(()),
+            Err(crate::WasiRuntimeError::Wasi(crate::WasiError::Exit(_))) => Ok(()),
```
Michael-F-Bryan, line 101: This means any exit code will indicate a success. What about modifying it to only report an exit code of `0` as `Ok(())`?

```suggestion
            Err(crate::WasiRuntimeError::Wasi(crate::WasiError::Exit(0))) => Ok(()),
```

Michael-F-Bryan (changes requested):

lib/wasi/src/runners/wasi.rs
Michael-F-Bryan: After talking with @ptitSeb, we don't want to make assumptions about the exit code so unconditionally returning `Ok(())` is fine.

Michael-F-Bryan (approved):

lib/wasi/src/runners/wasi.rs
ptitSeb: Exit Code !=0 are still normal exit code. Will not make any assumption on the exit code.

PR #3693, opened 2023-03-20
Validate Module Artifacts (CLI + WASIX ModuleCache)

theduke:
- wasix: Use validation when loading serialized modules
- cli: Use validation when loading module artifacts


syrusakbary (dismissed):
This is not solving the caching problem in the proper way.

We should be fixing this function:
https://github.com/wasmerio/wasmer/blob/master/lib/types/src/serialize.rs#L103-L118

To use: `rkyv::check_archived_value` instead (note that you may need to enable the `validation` feature in rkyv)

ptitSeb:
I think this https://github.com/wasmerio/wasmer/issues/3695 is relevant here.

theduke:
Important modifications: 

commit d4c65d3c4737c6e610e0efc166f3de3553ac7f9d (HEAD -> wasix-no-unsafe-archive, origin/wasix-no-unsafe-archive)
Author: Christoph Herzog <chris@theduke.at>
Date:   Thu Mar 30 15:36:00 2023 +0200

    Switch wasi module cache and CLI to use validated module deserialization

    Prevents undefined behaviour when loading modules.

    This is a much saner default option, since loading modules without
    validation can cause UB.

commit 6c7d17281e6c13107cc469e79d4e73fca00c9fc8
Author: Christoph Herzog <chris@theduke.at>
Date:   Thu Mar 30 15:32:51 2023 +0200

    feat: Implement validation for serialized modules

    Implement validation of rykv serialized modules with the "validation"
    feature which uses bytecheck.

    Bumps the MAGIC VERSION of the serialization header , because some
    format changes were required.
    (mostly setting an explicit repr() for enums)

    Also updates various APIs to add unsafe methods as
    "dangerous_deserialize_unchecked", deprecate the old unsafe "deserialize" methods
    and to add new "deserialize_checked" methods

theduke:
I rather like the `dangerous_` prefix, because it really makes the user notice "hey, really don't do this unless you are sure you want it". 

But I don't mind much either way.
Thoughts, @syrusakbary ?

lib/api/src/module.rs
```diff
@@ -217,6 +217,21 @@ impl Module {
...
     }
 
+    /// See [`Self::dangerous_deserialize_unchecked`].
+    ///
+    /// # Safety
+    /// See [`Self::dangerous_deserialize_unchecked`].
+    #[deprecated(
+        since = "3.2.0",
+        note = "Use `deserialize_checked` or `dangerous_deserialize_unchecked` instead"
+    )]
```
syrusakbary, line 227: I don't think we need the dangerous here since it's already unsafe? We can just keep using the normal deserialize?

syrusakbary: Yeah, I also thought this was code in the VM. Let's not do a breaking change regarding deserialize

theduke: I really feel very few users should use this function, unless they are really, really sure they know what they are doing - so it should also have a scary name.

syrusakbary: Let's put all breaking changes for 4.0

syrusakbary (commented):
I have a better take. Rather than making `deserialize` unsafe and deprecated, we will make `deserialize` to call `deserialize_checked` and `deserialize_unchecked` as `dangerous_deserialize_unchecked`. Thoughts @theduke ?

theduke:
I don't think we want to change the behaviour of this. 

I'll just revert the renames and we can tackle that for 4.0.

lib/compiler/src/engine/inner.rs
```diff
@@ -200,12 +200,38 @@ impl Engine {
...
     }
 
+    /// Deserializes a WebAssembly module
     #[cfg(not(target_arch = "wasm32"))]
+    pub fn deserialize_checked(&self, bytes: &[u8]) -> Result<Arc<Artifact>, DeserializeError> {
+        #[cfg(not(target_arch = "wasm32"))]
+        {
+            Ok(Arc::new(Artifact::deserialize_checked(self, bytes)?))
+        }
+        #[cfg(target_arch = "wasm32")]
```
ptitSeb, line 210: how can this `cfg` be used, the function has the oposite config.

lib/compiler/src/engine/inner.rs
```diff
@@ -200,12 +200,38 @@ impl Engine {
         Ok(Arc::new(Artifact::deserialize(self, bytes)?))
     }
 
+    /// Deserializes a WebAssembly module
     #[cfg(not(target_arch = "wasm32"))]
+    pub fn deserialize_checked(&self, bytes: &[u8]) -> Result<Arc<Artifact>, DeserializeError> {
+        #[cfg(not(target_arch = "wasm32"))]
```
ptitSeb, line 206: the `cfg` is probably not usefull here.

ptitSeb:
This is quite the oposite as the "dangerous_" one now. I would have created `_unchecked` high level variant with the current code, left the current one to just call the `_unchecked` and mark the current one as deprecated.
Else, I'm unsure many will une the newly `_checked` version.

But I'm fine with current version, is just that I'm unsure it's enough this time.

lib/compiler/src/engine/inner.rs
theduke: Thanks!

This was just an artifact from reverting the renaming.

lib/compiler/src/engine/inner.rs
theduke: As above, just an artifact of the revert.

ptitSeb (approved):

ptitSeb:
fine with me

theduke:
`4.0` shouldn't be too far off anyway.

ptitSeb:
> `4.0` shouldn't be too far off anyway.

Yes, but `4.0` is too late to remove function if they have not been marked as `deprecated` before.
That means it's for the `5.0` for now.

PR #3694, opened 2023-03-21
Added unit tests for the opcodes version of futex used by http server

john-sharratt:


john-sharratt:
note: this pull request is meant to fail, until the opcodes are fixed for futex

PR #3700, opened 2023-03-22
Added back all removed function for Engine to avoid API breaking changes

ptitSeb:
Added back all removed function for Engine to avoid API breaking changes

Compared with 3.1.1 API, adding back the missing function.

Note that all function from `NativeEngineExt` are now shadowed by this.  So those function are not marked as deprecated to avoid the use of `NativeEngineExt::tunables(...)` for example.
Needed to removed the use of `NativeEngineExt` from many crates to satisfy clipper as a consequence.


theduke (approved):
Annoying, but necessary.

Let's also create a followup issue to remove the methods in the `4.0` milestone so we don't forget.

PR #3706, opened 2023-03-24
[CI] Add RISCV in test and build

ptitSeb:
Added RISCV in both test and build Actions


.github/workflows/build.yml
```diff
@@ -405,8 +405,73 @@ jobs:
...
+          CROSS_DOCKER_IN_DOCKER: true
+          CARGO_TARGET: riscv64gc-unknown-linux-gnu
+          PKG_CONFIG_PATH: /usr/lib/riscv64-linux-gnu/pkgconfig
+          PKG_CONFIG_ALLOW_CROSS: true
+          ENABLE_LLVM: 0
+      - name: Dist
+        run: |
+          make distribution
+        env:
+          CARGO_BINARY: docker run -v /var/run/docker.sock:/var/run/docker.sock -v ${GITHUB_WORKSPACE}:/project -w /project wasmer/riscv64:latest cargo
```
theduke, line 458: This is a nice trick.

theduke (approved):
Didn't do a thorough review, but I didn't notice anything problematic. LGTM.

Docker cross-compilation is always annoying...

ptitSeb:
The RISCV branch depend on using `wapm-targz-to-pirita` v0.2.0, but then, it need #3690 to be merge to be able to run all test without error (or else https://github.com/wasmerio/wasmer/blob/master/lib/wasi/src/runners/container.rs#L36 will be triggered)

.github/workflows/web.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Install Rust
+      uses: actions/checkout@v3
+      with:
+        toolchain: ${{ matrix.rust }}
+
+    - name: Show Rust version
+      run: |
+        rustc --version
+
+    - name: Check formatting
```
theduke: Is formatting already covered by the regular formatting check?
Or is that also just run for specific crates?

.github/workflows/web.yaml
```diff
@@ -0,0 +1,48 @@
...
+      shell: bash
+      run: |
+        cd lib/wasi-web
+        cargo fmt --check
+
+    - name: Check
+      continue-on-error: true
+      env:
+        CARGO_NET_GIT_FETCH_WITH_CLI: "true"
```
theduke: Optional: do we also want to add this to the makefile and use `make XXX` instead so we have local parity for checks?

.github/workflows/web.yaml
theduke: Whoops, wrong PR.

PR #3710, opened 2023-03-26
Implemented an asyncify based implementation of asynchronous threading

john-sharratt:
This pull request implements `asynchronous threading`

In essence what this does is allow the runtime to exit a WASM thread and resume it where it left off after an `async` condition passes - when coupled with the WASIX syscalls this allows for all waits and IO operations to unload their WASM threads when going idle.

In future this will allow a designer to `move` threads from one machine to another

Performed some benchmarks and asynchronous threading gives approximately a 6% increase in performance thus the benefits of parking idle threads appears to out way the overhead of managing them (tested with multiple runs).

before:
```
john@AlienWorld:/prog/deploy$ wrk -t 10 -c 600 -d 60 http://localhost:9080/
Running 1m test @ http://localhost:9080/
  10 threads and 600 connections
    Latency   203.43ms  335.09ms   1.87s    85.30%
    Req/Sec     1.38k   777.50     4.85k    71.05%
  750093 requests in 1.00m, 1.07GB read
Requests/sec:  12483.79
Transfer/sec:     18.24MB
```

after:
```
john@AlienWorld:/prog/deploy$ wrk -t 10 -c 600 -d 60 http://localhost:9080/
Running 1m test @ http://localhost:9080/
  10 threads and 600 connections
    Latency   125.93ms  209.33ms   1.41s    87.05%
    Req/Sec     1.38k   592.88     4.32k    73.30%
  798846 requests in 1.00m, 1.14GB read
Requests/sec:  13292.74
Transfer/sec:     19.42MB
```


syrusakbary (commented):
It looks good from me for the API side

lib/api/src/ptr.rs
```diff
@@ -60,7 +60,7 @@ pub type WasmPtr64<T> = WasmPtr<T, Memory64>;
 #[repr(transparent)]
 pub struct WasmPtr<T, M: MemorySize = Memory32> {
     offset: M::Offset,
-    _phantom: PhantomData<*mut T>,
```
theduke, line 63: There used to be some obscure effects on type layout and optimizations when using full types, hence `PhantomData` is usually used with pointers rather than actual types.

Any specific reason you changed this?

lib/wasi/src/runtime/task_manager/mod.rs
```diff
@@ -70,6 +83,21 @@ pub trait VirtualTaskManager: std::fmt::Debug + Send + Sync + 'static {
...
 
+    /// Starts an WebAssembly task will will run on a dedicated thread
+    /// pulled from the worker pool that has a stateful thread local variable
+    /// After the trigger has successfully completed
+    fn resume_wasm_after_trigger(
+        &self,
+        task: Box<dyn FnOnce(Store, Module) + Send + 'static>,
+        store: Store,
+        module: Module,
+        #[allow(clippy::all)] trigger: Box<
```
theduke, line 94: As the clippy lint says ... let's move that out to a separate type definition `type WhateverTrigger = Box<...>`.

lib/wasi/src/runtime/task_manager/tokio.rs
```diff
@@ -146,6 +146,32 @@ impl VirtualTaskManager for TokioTaskManager {
...
+        module: Module,
+        trigger: Box<
+            dyn FnOnce(Store) -> Pin<Box<dyn Future<Output = TaskResumeAction> + Send + 'static>>
+                + Send
+                + 'static,
+        >,
+    ) -> Result<(), WasiThreadError> {
+        let trigger = trigger(store);
+        let handle = self.0.clone();
+        self.0.spawn(async move {
```
theduke, line 163: Why is resuming bound so tightly to threading?

We shouldn't need to spawn a thread in order to run something resumable.

theduke (commented):
I'm not so happy about putting more logic into the logic in `binfactory/exec.rs`, I really wanted to move away from that to make the logic less dependent on it.

Calls like `WasiEnvBuilder::run()` do not go through that flow.

We should but this into some reusable function , that preferably also does not require spawning threads.

lib/wasi/src/syscalls/mod.rs
```diff
@@ -297,75 +325,203 @@ where
...
 /// to (optionally) timeout the IO, force exit the process, callback signals and pump
 /// synchronous IO engine
-pub(crate) fn __asyncify_light<T, Fut>(
-    env: &WasiEnv,
+///
+/// This will either return the `ctx` as the asyncify has completed successfully
+/// or it will return an WasiError which will exit the WASM call using asyncify
+/// and instead process it on a shared task
+///
+pub(crate) fn __asyncify_with_deep_sleep<M: MemorySize, Fut>(
```
theduke, line 395: This behaviour needs to be opt-in and specified in `WasiEnvBuilder`. 

Good place to put th esetting might be in `ControlPlane`?

lib/wasi/src/runtime/task_manager/mod.rs
john-sharratt: fixed

lib/api/src/ptr.rs
john-sharratt: indeed that is the case

lib/wasi/src/runtime/task_manager/tokio.rs
john-sharratt: This is required because of how the `postMessage` and friends are implemented in the JS version of the runtime. Suggest when we migrate the runtime to JS that we also refactor this area.

I created an issue for it here:
https://github.com/wasmerio/wasmer/issues/3714

lib/wasi/src/syscalls/mod.rs
john-sharratt: I have added an opt-out in `ControlPlane`

john-sharratt:
Quite some refactoring to get the snapshot tests working - there were a number of edge cases that were failed

john-sharratt:
> Only did a supferficial review, more in-depth after we discuss it.
> 
> Also:
> 
> * needs some tests

I've split the snapshot tests so there are explicit tests `with-async-threads` and `without-async-threads` - this will make it clearer when there are future failures if the async threads code is causing it or not.

All tests should now be green

lib/cli/src/commands/run/wasi.rs
```diff
@@ -225,21 +249,161 @@ impl Wasi {
...
+        let tasks = run.ctx.data(&store).tasks().clone();
+
+        // The return value is passed synchronously and will block until the result is returned
+        // this is because the main thread can go into a deep sleep and exit the dedicated thread
+        let (tx, rx) = std::sync::mpsc::channel();
+
+        // We run it in a blocking thread as the WASM function may otherwise hold
+        // up the IO operations
+        tasks.task_dedicated(Box::new(move || {
+            Self::run_with_deep_sleep(run, store, tx, None);
```
theduke, line 263: This should only be needed when the functionality is enabled, but:

I don't see the point of supporting sleep for the main thread in the CLI. 
What's the motivation?
Just for testing?
Child threads can make sense when opted into, but otherwise this just seems to add code comlexity.

lib/cli/src/commands/run/wasi.rs
```diff
@@ -225,21 +249,161 @@ impl Wasi {
...
+        let ctx = run.ctx;
+        if let Some((mut rewind_state, trigger_res)) = rewind_state {
+            if rewind_state.is_64bit {
+                if let Err(exit_code) = rewind_state
+                    .rewinding_finish::<Memory64>(ctx.env.clone().into_mut(&mut store), trigger_res)
+                {
+                    tx.send(Ok(exit_code.raw())).ok();
+                    return;
+                }
+                let res = rewind::<Memory64>(
```
theduke, line 286: This is way too much logic to expose/put into the CLI.

This should be handled by a helper function in the `wasi` crate.

lib/cli/src/commands/run/wasi.rs
```diff
@@ -76,6 +83,10 @@ pub struct Wasi {
     #[clap(long = "no-tty")]
     pub no_tty: bool,
 
+    /// Disables asynchronous threading
+    #[clap(long = "no-async-threads")]
+    pub no_async_threads: bool,
```
theduke, line 88: In the spirit of being opt-in this should be `enable_async_threads`.

lib/cli/src/commands/run/wasi.rs
```diff
@@ -225,21 +249,161 @@ impl Wasi {
...
+        let ret: Result<i32> = match result {
             Ok(_) => Ok(0),
             Err(err) => {
-                let err: anyhow::Error = match err.downcast::<WasiError>() {
-                    Ok(WasiError::Exit(exit_code)) => {
-                        return Ok(exit_code.raw());
+                match err.downcast::<WasiError>() {
+                    Ok(WasiError::Exit(exit_code)) => Ok(exit_code.raw()),
+                    Ok(WasiError::DeepSleep(deep)) => {
+                        let pid = ctx.data(&store).pid();
```
theduke, line 365: Yeah, reading this, as mentioned above, we really shouldn't do this for the main thread...

lib/wasi/src/bin_factory/exec.rs
```diff
@@ -195,6 +190,124 @@ pub fn spawn_exec_module(
     Ok(join_handle)
 }
 
+/// Finishes with a module and notifies an errcode
```
theduke, line 193: This function seems pretty redundant.

lib/wasi/src/bin_factory/exec.rs
```diff
@@ -195,6 +190,124 @@ pub fn spawn_exec_module(
...
+    rewind_state: Option<(RewindState, Result<(), Errno>)>,
+) {
+    let env = ctx.data(&store);
+    let pid = env.pid();
+    let thread = env.thread.clone();
+    let tasks = env.tasks().clone();
+    thread.set_status_running();
+
+    // If we need to rewind then do so
+    if let Some((mut rewind_state, trigger_res)) = rewind_state {
```
theduke, line 213: This should only be done if async threads are enabled.

lib/wasi/src/capabilities.rs
```diff
@@ -31,4 +31,8 @@ pub struct CapabilityThreadingV1 {
     ///
     /// [`None`] means no limit.
     pub max_threads: Option<usize>,
+
+    /// Flag that indicates if asynchronous threading is disabled
+    /// (default = false)
+    pub disable_asynchronous_threading: bool,
```
theduke, line 37:
```suggestion
    pub enable_asynchronous_threading: bool,
```

lib/wasi/src/fs/inode_guard.rs
```diff
@@ -115,9 +127,12 @@ impl<'a> InodeValFilePollGuardJoin<'a> {
     pub(crate) fn fd(&self) -> u32 {
         self.fd
     }
+    pub(crate) fn peb(&self) -> PollEventSet {
```
theduke, line 130: More descriptive name, please.

lib/wasi/src/lib.rs
```diff
@@ -111,49 +109,92 @@ pub use crate::{
...
-#[derive(Error, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub enum WasiError {
-    #[error("WASI exited with code: {0}")]
-    Exit(ExitCode),
-    #[error("The WASI version could not be determined")]
-    UnknownWasiVersion,
+/// Trait that will be invoked after the rewind has finished
+/// It is possible that the process will be terminated rather
+/// than restored at this point
+pub trait RewindPostProcess {
```
theduke, line 119: Maybe put this in `os/task/rewind.rs`?

We could collect related functionality there.

lib/wasi/src/lib.rs
```diff
@@ -111,49 +109,92 @@ pub use crate::{
...
-/// Represents the ID of a WASI calling thread
-#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct WasiCallingId(u32);
-
-impl WasiCallingId {
-    pub fn raw(&self) -> u32 {
-        self.0
-    }
+/// The rewind state after a deep sleep
+pub struct RewindState {
```
theduke, line 129: Yeah, lets put these into a separate module.

lib/wasi/src/lib.rs
```diff
@@ -337,19 +378,6 @@ lazy_static::lazy_static! {
     static ref CALLER_ID_SEED: Arc<AtomicU32> = Arc::new(AtomicU32::new(1));
 }
 
-/// Returns the current thread ID
-pub fn current_caller_id() -> WasiCallingId {
```
theduke, line 341: This function is currently public.

Is there a usecase for external usage?
If so we can't just remove it.

lib/wasi/src/lib.rs
```diff
@@ -838,6 +866,11 @@ fn mem_error_to_wasi(err: MemoryAccessError) -> Errno {
     }
 }
 
+fn mem_error_to_wasi_error(err: MemoryAccessError) -> WasiError {
```
theduke, line 869: `From<MemoryAccessError> for WasiError`

theduke: But this seems like it should be a `WasiRuntimeError`.

lib/wasi/src/os/task/control_plane.rs
```diff
@@ -39,12 +39,15 @@ impl WasiControlPlaneHandle {
 pub struct ControlPlaneConfig {
     /// Total number of tasks (processes + threads) that can be spawned.
     pub max_task_count: Option<usize>,
+    /// Flag that indicates if asynchronous threading is disable (opt-out)
+    pub disable_asynchronous_threading: bool,
```
theduke, line 43: `enable_`. ;)

lib/wasi/src/runtime/task_manager/mod.rs
```diff
@@ -70,6 +89,17 @@ pub trait VirtualTaskManager: std::fmt::Debug + Send + Sync + 'static {
         spawn_type: SpawnType,
     ) -> Result<(), WasiThreadError>;
 
+    /// Starts an WebAssembly task will will run on a dedicated thread
+    /// pulled from the worker pool that has a stateful thread local variable
+    /// After the trigger has successfully completed
+    fn resume_wasm_after_trigger(
```
theduke, line 95: Why do these two need to be on the task manager?

Do we actually need custom resume logic depending on the environment?

theduke: I know you mentioned postMessage, but I need some additional context there.

lib/wasi/src/syscalls/wasix/futex_wait.rs
```diff
@@ -57,6 +73,36 @@ where
     }
 }
 
+struct FutexAfter<M>
```
theduke, line 76: What's a `FutexAfter` => doc comment

theduke (changes requested):
This is a big PR with a lot of complextiy.

We really need to make sure that:

* the added functionality is completely optional and does not affect the default execution model
* the abstractions are reasonably clean

lib/cli/src/commands/run/wasi.rs
john-sharratt: the motivation was that asynchronous threading is higher performance (when under high number of threads) as shown with some performance test runs with minimal extra overhead therefore the default of "on" was possibly the preferred for option for consumers however I get your point that reduced capabilities may be the preferred as long as they can opt-in.

i've patched it and switched from opt-out to opt-in

lib/cli/src/commands/run/wasi.rs
john-sharratt: I've reverted all this code back to synchronous only, given its isolated to the CLI only anyway its not that big a deal to revert it.

lib/cli/src/commands/run/wasi.rs
john-sharratt: I've reverted all this code back to synchronous only, given its isolated to the CLI only anyway its not that big a deal to revert it.

lib/wasi/src/bin_factory/exec.rs
john-sharratt: The function is gone now since it was reverted.

lib/wasi/src/lib.rs
john-sharratt: no one is using it externally and it should never have been public - it was a leaky abstraction

lib/wasi/src/bin_factory/exec.rs
john-sharratt: The function has been removed

lib/wasi/src/os/task/control_plane.rs
john-sharratt: fixed

lib/wasi/src/syscalls/wasix/futex_wait.rs
john-sharratt: Added documentation for this struct

lib/wasi/src/lib.rs
john-sharratt: The function wasn't used anymore anyway so I deleted it

lib/wasi/src/runtime/task_manager/mod.rs
john-sharratt: I'll write the `wasmer-web` version and link the code here - that will be easier to understand

lib/cli/src/commands/run/wasi.rs
john-sharratt: thinking about this some more there may be a very good reason to do this....

`tiered compilation`

asynchronous threads allow a safe way to unload and reload a new module, hence it may become fairly trivial to build on top of this patch and introduce a mechanism where it defaults to `single-pass` and then upgrades to `cranelift` on the next rewind

theduke: Slight misunderstanding there: this wasn't a general remark, but just specifically about support deep sleep for the main thread in the CLI - which I don't think is necessary.

lib/wasi/src/lib.rs
theduke: Yeah then we should make it private. 

We now have decoupled `-wasix` from the other crates for versioning, so we can make breaking changes much more liberally.

lib/cli/src/commands/run/wasi.rs
john-sharratt: its need in the CLI to support tiered compilation from the CLI as otherwise the main thread will not be able to reload the module after its upgraded from `singlepass` to `cranelift`

lib/wasi/src/runtime/task_manager/mod.rs
john-sharratt: i've managed to remove this now with some refactoring

lib/wasi/src/bin_factory/exec.rs
john-sharratt: refactored, needs new review

lib/wasi/src/fs/inode_guard.rs
john-sharratt: refactored, needs new review

lib/wasi/src/lib.rs
john-sharratt: yeah might be good to create that but not in `/os` as its more runtime related

lib/wasi/src/lib.rs
john-sharratt: https://github.com/wasmerio/wasmer/pull/3710/commits/4b7ddf145b70bf832e4ba167b1390c27def0b3f3

lib/wasi/src/lib.rs
john-sharratt: https://github.com/wasmerio/wasmer/pull/3710/commits/4b7ddf145b70bf832e4ba167b1390c27def0b3f3

john-sharratt:
Tests have been added in snapshots

lib/api/src/js/externals/table.rs
```diff
@@ -116,4 +117,8 @@ impl Table {
     pub fn is_from_store(&self, _store: &impl AsStoreRef) -> bool {
         true
     }
+
+    pub fn store_id(&self) -> StoreId {
+        StoreId::unknown()
+    }
```
syrusakbary, line 123: Why this? Store's should be consistent per thread @john-sharratt

lib/cli/src/commands/run/wasi.rs
```diff
@@ -215,7 +238,7 @@ impl Wasi {
     /// Helper function for instantiating a module with Wasi imports for the `Run` command.
     pub fn instantiate(
         &self,
-        store: &mut impl AsStoreMut,
+        store: &mut Store,
```
syrusakbary, line 241: The other one keeps things a bit better if you wants to pass a function context instead of a store.
I'd like to keep the `&mut impl AsStoreMut`

lib/types/src/store_id.rs
```diff
@@ -11,6 +12,13 @@ use std::{
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 pub struct StoreId(NonZeroUsize);
 
+impl StoreId {
+    /// Represents an unknown store ID
+    pub fn unknown() -> Self {
+        Self(usize::MAX.try_into().unwrap())
+    }
```
syrusakbary, line 19: This causes `Store::unknown() == Store::unknown()` which seems a bit strange for me

lib/api/src/js/externals/table.rs
john-sharratt: It can be deleted - it was added because of the improved error messages which I have subsequently removed again - ill take it out

lib/types/src/store_id.rs
john-sharratt: let me take a look at that

lib/cli/src/commands/run/wasi.rs
john-sharratt: let me look into this

lib/api/src/js/externals/table.rs
john-sharratt: deleted

lib/types/src/store_id.rs
john-sharratt: deleted

lib/cli/src/commands/run/wasi.rs
john-sharratt: updated

syrusakbary (approved):
Looks good to me from the API side

lib/cli/src/commands/run/wasi.rs
```diff
@@ -225,21 +248,164 @@ impl Wasi {
...
+        rx.recv()
+            .expect("main thread terminated without a result, this normally means a panic occurred within the main thread")
+    }
+
+    // Runs the Wasi process (asynchronously)
+    pub fn run_with_deep_sleep(
+        run: RunProperties,
+        mut store: Store,
+        tx: Sender<Result<i32>>,
+        rewind_state: Option<(RewindState, Result<(), Errno>)>,
```
syrusakbary, line 273: It feels weird that we are passing a result as an argument.
I think we can pass an `Option<Errno>` instead?

lib/wasi/src/bin_factory/exec.rs
```diff
@@ -186,6 +135,130 @@ pub fn spawn_exec_module(
...
+        .map(|a| a.clone())
+        .ok()
+}
+
+/// Calls the module
+fn call_module(
+    ctx: WasiFunctionEnv,
+    mut store: Store,
+    handle: WasiThreadRunGuard,
+    rewind_state: Option<(RewindState, Result<(), Errno>)>,
```
syrusakbary, line 153: Same as other suggestion. Since the result is not used, perhaps we can pass an `Option<Errno>`

lib/wasi/src/syscalls/wasix/futex_wake_all.rs
```diff
@@ -6,7 +6,7 @@ use crate::syscalls::*;
 /// ## Parameters
 ///
 /// * `futex` - Memory location that holds a futex that others may be waiting on
-#[instrument(level = "trace", skip_all, fields(futex_idx = field::Empty, woken = field::Empty), ret)]
+//#[instrument(level = "trace", skip_all, fields(futex_idx = field::Empty, woken = field::Empty), ret)]
```
syrusakbary, line 9: No more tracing here? Perhaps better to remove?

lib/wasi/src/syscalls/wasix/futex_wait.rs
```diff
@@ -3,56 +3,95 @@ use std::task::Waker;
...
-struct FutexPoller<'a, M>
-where
-    M: MemorySize,
-{
-    env: &'a WasiEnv,
-    view: MemoryView<'a>,
+#[derive(Clone)]
+struct FutexPoller {
+    state: Arc<WasiState>,
+    woken: Arc<Mutex<bool>>,
```
syrusakbary, line 9: AtomicBool seems better suited here?

[std](https://doc.rust-lang.org/std/index.html)::[sync](https://doc.rust-lang.org/std/sync/index.html)::[atomic](https://doc.rust-lang.org/std/sync/atomic/index.html)::[AtomicBool](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicBool.html#)

lib/wasi/src/runtime/task_manager/mod.rs
```diff
@@ -2,43 +2,125 @@
...
+use crate::{capture_snapshot, InstanceSnapshot, WasiEnv, WasiFunctionEnv, WasiThread};
 
 #[derive(Debug)]
-pub struct SpawnedMemory {
-    pub ty: MemoryType,
+pub enum SpawnMemoryType<'a> {
+    CreateMemory,
+    CreateMemoryOfType(MemoryType),
+    ShareMemory(Memory, StoreRef<'a>),
+    CopyMemory(Memory, StoreRef<'a>),
```
syrusakbary, line 22: I don't love that we are storing `StoreRef` which is pretty much internal to the VM. This might break JSC again.
Are there other ways to achieve this?

john-sharratt: unsafe ones yes... the problem is that memory is a pointer to something that exists on in a single thread when running in the browser... the alternative is to change this to VMMemory

lib/wasi/src/syscalls/wasix/futex_wake_all.rs
john-sharratt: that's a mistake, ill undo that

lib/cli/src/commands/run/wasi.rs
john-sharratt: That's subjective as an Option does not normally represent an error but in this case it does. I'm not hung on it though so happy to change it if you like but its more a personal preference I think

lib/wasi/src/bin_factory/exec.rs
john-sharratt: See other comment

lib/wasi/src/syscalls/wasix/futex_wait.rs
john-sharratt: i managed to delete it completely instead

lib/wasi/src/syscalls/wasix/futex_wake_all.rs
john-sharratt: fixed

lib/wasi/src/runtime/task_manager/mod.rs
syrusakbary: `VMMemory` alternative will be non viable for JSC. Do you know if having a `Box<dyn AsStoreRef<'a>>` would work here @john-sharratt ?

john-sharratt: Putting it on the heap won't change that fact its still a lifetime limited memory reference - if you want a heap reference we would need to do something like `Arc<Mutex<Store>>` but that will be a mess

theduke (commented):
This is only partial, final review to come.

lib/cli/src/commands/run/wasi.rs
john-sharratt: agreed - I'll leave it like this for now

lib/api/src/externals/memory.rs
```diff
@@ -143,34 +134,44 @@ impl Memory {
...
     /// Attempts to clone this memory (if its clonable) in a new store
-    pub fn clone_in_store(
+    /// (cloned memory will be shared between those that clone it)
+    pub fn share_in_store(
         &self,
         store: &impl AsStoreRef,
         new_store: &mut impl AsStoreMut,
-    ) -> Option<Self> {
+    ) -> Result<Self, MemoryError> {
         if !self.ty(store).shared {
```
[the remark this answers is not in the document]
john-sharratt: thanks

lib/api/src/externals/memory.rs
```diff
@@ -143,34 +134,44 @@ impl Memory {
...
+            });
         }
         self.0
             .try_clone(&store)
             .map(|new_memory| Self::new_from_existing(new_store, new_memory))
+            .ok_or_else(|| MemoryError::Generic("memory is not clonable".to_string()))
     }
 
     /// Attempts to duplicate this memory (if its clonable) in a new store
-    pub fn duplicate_in_store(
```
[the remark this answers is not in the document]
john-sharratt: put this back in and added deprecation instead

lib/wasi-web/src/runtime.rs
```diff
@@ -103,47 +102,34 @@ impl<'g> Drop for WebRuntimeGuard<'g> {
...
-    }
-
     /// Invokes whenever a WASM thread goes idle. In some runtimes (like singlethreaded
     /// execution environments) they will need to do asynchronous work whenever the main
     /// thread goes idle and this is the place to hook for that.
-    async fn sleep_now(&self, time: Duration) {
+    fn sleep_now(
+        &self,
+        time: Duration,
+    ) -> Pin<Box<dyn Future<Output = ()> + Send + Sync + 'static>> {
```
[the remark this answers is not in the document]
john-sharratt: This needs to be `Send` - used in the asyncify code which can move the sleep to a new thread

lib/wasi/src/syscalls/mod.rs
```diff
@@ -297,32 +325,241 @@ where
...
+    mut after: After,
+) -> Result<AsyncifyAction<'_>, WasiError>
+where
+    After: RewindPostProcess + Send + Sync + 'static,
+{
+    // We build a wrapper around the polling struct that will
+    // check for a timeout
+    struct WorkWithTimeout {
+        inner: Pin<Box<AsyncifyFuture>>,
+        timeout: Option<Pin<Box<dyn Future<Output = ()> + Send + Sync>>>,
```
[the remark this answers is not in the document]
john-sharratt: needs to be `Send`

lib/wasi/src/runtime/task_manager/mod.rs
```diff
@@ -2,43 +2,125 @@
...
 }
 
-#[derive(Debug)]
-pub enum SpawnType {
-    Create,
-    CreateWithType(SpawnedMemory),
-    NewThread(Memory),
+pub type WasmResumeTask = dyn FnOnce(WasiFunctionEnv, Store, Result<(), Errno>) + Send + 'static;
+
+pub type WasmResumeTrigger =
```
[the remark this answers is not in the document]
john-sharratt: Needs to be `Send`

lib/wasi/src/runtime/task_manager/mod.rs
john-sharratt: For now this will have to do, left a TODO to look at it again

lib/api/src/externals/memory.rs
```diff
@@ -119,15 +119,6 @@ impl Memory {
         self.0.grow(store, delta)
     }
 
-    /// Copies the memory to a new store and returns a memory reference to it
-    pub fn copy_to_store(
```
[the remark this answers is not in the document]
john-sharratt: put this back in for now with deprecation

lib/wasi/src/syscalls/mod.rs
```diff
@@ -297,32 +325,241 @@ where
...
+}
+impl<'a, 'b, 'c> Future for AsyncifyPoller<'a, 'b, 'c> {
+    type Output = Result<Result<(), Errno>, WasiError>;
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let store = self.store;
+        let memory = self.memory.clone();
+        if let Poll::Ready(res) = self.work.as_mut().poll(cx) {
+            return Poll::Ready(Ok(res));
+        }
+        if let Some(forced_exit) = self.thread.try_join() {
```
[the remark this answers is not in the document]
john-sharratt: Signals are fired instead on thread however the signals alone can not contain metadata, hence the extra method call

syrusakbary (approved):
LGTM on the API side

lib/wasi-web/src/runtime.rs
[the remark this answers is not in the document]
theduke: `BoxFuture` includes `Send`, but doesn't include `Sync`, because it's not needed.

PR #3711, opened 2023-03-27
Allow publishing packages without a module

theduke:
Removes the check that requires a module to be specified when publishing
a package.

Removed because modules should not be mandatory - packages with just a
file system, or with a command that uses the module of a dependency are
perfectly valid.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1344,12 +1344,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "42703706b716c37f96a77aea830392ad231f44c9e9a67872fa5548707e11b11c"
 
 [[package]]
-name = "fuchsia-cprng"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


Michael-F-Bryan (approved):
LGTM 👍

theduke:
Closes #3712

theduke:
@Michael-F-Bryan the changes here seem to regress the Python tests again.

syrusakbary (approved):

Michael-F-Bryan:
@theduke I'm guessing the regression is because `wapm-targz-to-pirita` now generates WEBC v2 files and the `wasmer_wasix::runner::WapmContainer` can only deal with WEBC v1.

#3690 fixes that by removing the `WapmContainer` and using `webc::compat::Container` directly, but it's currently failing to load Python because it actually cares whether a path is relative or absolute and `webc::v1::WebC` would silently work with the relative paths given to it by Python.

theduke:
The package validation changes were included in #3716, since that also moves the validation code.

The webc version upgrades / deduplication should be handled in a dedicated PR.

PR #3715, opened 2023-03-28
Fix annotation deserializing for the Wasi and Wcgi runners

Michael-F-Bryan:
Here are a couple small fixes I added while making the [`wcgi-rust-template`](https://github.com/wasmerio/wcgi-rust-template) and [`wcgi-php-template`](https://github.com/wasmerio/wcgi-php-template) work.

The key difference is that `wapm2pirita` will nest all annotations in your `wasmer.toml` file under an extra level ([here](https://github.com/wasmerio/pirita/blob/03f0f159e244d9e2a6197c7ebd81e46463e236cc/crates/wapm-targz-to-pirita/src/transform.rs#L240-L251)). 

That means, instead of a `webc::metadata::CommandV2` looking like this

```json
{
  "runner": "https://webc.org/runner/wcgi",
  "annotations": {
    "wcgi": {
      "dialect": "rfc-3875"
    }
  }
}
```

and being able to call `command.get_annotations::<Wasi>("wasi")`, the annotations actually look like this:

```json
{
  "runner": "https://webc.org/runner/wcgi",
  "annotations": {
    "https://webc.org/runner/wcgi": {
      "wcgi": {
        "dialect": "rfc-3875"
      }
    }
  }
}
```

... and you need to create a temporary struct with the correct annotation types (e.g. `webc::metadata::annotations::Wasi` and `webc::metadata::annotations::Wcgi` in the WCGI case) so you can do `let MyAnnotations { wcgi, wasi } = command.get_annotations(webc::metadata::annotations::WCGI_BASE_URI)`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1661,7 +1661,7 @@ dependencies = [
  "atomic-polyfill",
  "hash32",
  "rustc_version 0.4.0",
- "spin 0.9.6",
+ "spin 0.9.7",
  "stable_deref_trait",
 ]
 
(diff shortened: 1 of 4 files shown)


ptitSeb (approved):
Looks good to me, but I'm not very relevent in this subject.

PR #3716, opened 2023-03-28
Move Package Build/Publish Logic to wasmer-registry

theduke:
Moves the code for building and publishing packages from wasmer.toml manifests 
from the CLI to the wasmer-registry crate to make the logic reusable.

- refactor: Move package build + publish logic to registry
- docs(cli): Fix doc comment on publish command


syrusakbary (approved):
Looks good to me once tests pass. I think we should get also a review from @Michael-F-Bryan prior to merging

PR #3718, opened 2023-03-29
Apply CGI environment variables in the correct order

syrusakbary:
This PR does the following:
* [x] Allows receiving a `--debug`/`-d` flag to the run-unstable subcommand
* [x] Removes the `SCRIPT_NAME` hack (it should be done by the package annotation)
* [x] Prints the server where things are running (I got so lost before that I had no idea in which port it was running)

### DIFF
--- a/lib/cli/src/commands/run_unstable.rs
+++ b/lib/cli/src/commands/run_unstable.rs
@@ -29,7 +29,6 @@ use wasmer_wasix::runners::{MappedDirectory, Runner, WapmContainer};
 use webc::metadata::Manifest;
 use webc_v4::DirOrFile;
 
-use crate::logging;
 use crate::{
     store::StoreOptions,
     wasmer_home::{DownloadCached, ModuleCache, WasmerHome},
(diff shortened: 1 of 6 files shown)


Michael-F-Bryan:
I've updated this PR to remove the `--debug` flag. We actually enable the logger already at the start of the function - you can increase the global log level by adding more `-v` arguments (default is warn) and can also use `$RUST_LOG` for fine-tuning.

I'd prefer not to pass `Config` to the callbacks if we can. For a situation like this, the caller already knows the address we're serving from because they set up the runner, so we can just reuse it.

I also reorganised the order we set environment variables in so we'll apply the CGI variables first and give users the ability to override them. I felt that was better than https://github.com/wasmerio/wcgi/pull/37 because it means you still get a useful default when you don't `$DOCUMENT_ROOT` and `$SCRIPT_FILENAME`, rather than erroring out because `mod_cgi.wasm` doesn't know which script to invoke.

theduke (approved):
Looks good to me.

PR #3723, opened 2023-03-30
Fix Wait/Notify opcode, the waiters hashmap is now on the Memory itself

ptitSeb:
The Wait/Notify opcodes needs a list a waiters to works. 
This list (an hashmap in fact) needs to be on a common object. It makes more sense to have the hashmap being part of the Memory object.

The `LinearMemory` memory trait has been expanded accordingly, with sensible default, that will behave as if the memory was not shared. 
The actuall hashmap in only implemented for `VMSharedMemory`


lib/vm/src/memory.rs
```diff
@@ -431,6 +436,16 @@ impl LinearMemory for VMSharedMemory {
         let forked = Self::duplicate(self)?;
         Ok(Box::new(forked))
     }
+
+    // Add current thread to waiter list
+    fn do_wait(&mut self, dst: u32, timeout: i64) -> u32 {
```
theduke, line 441: This is relatively minor, but can we keep using `NotifyLocation` here for better code clarity? Since we already have the type.

Would make the code more readable.

If you change it... since we convert the timeout to a `Duration` later anyway , I'd also take a `Duration` here so conversion happens at the earliest point.

lib/vm/src/memory.rs
```diff
@@ -431,6 +436,16 @@ impl LinearMemory for VMSharedMemory {
...
+        self.conditions.do_wait(dst, timeout)
+    }
+
+    /// Notify waiters from the wait list. Return the number of waiters notified
```
theduke, line 445: Same as above: `NotifyLocation` / `Duration`.

lib/vm/src/memory.rs
```diff
@@ -498,6 +513,16 @@ impl LinearMemory for VMMemory {
     fn duplicate(&mut self) -> Result<Box<dyn LinearMemory + 'static>, MemoryError> {
         self.0.duplicate()
     }
+
+    // Add current thread to waiter list
+    fn do_wait(&mut self, dst: u32, timeout: i64) -> u32 {
```
theduke, line 518: Same as above: NotifyLocation / Duration.

lib/vm/src/threadconditions.rs
```diff
@@ -0,0 +1,108 @@
...
+            conds.map.remove(&key);
+        }
+        ret
+    }
+
+    /// Notify waiters from the wait list
+    pub fn do_notify(&mut self, dst: u32, count: u32) -> u32 {
+        let key = NotifyLocation { address: dst };
+        let mut conds = self.lock_conditions().unwrap();
+        let mut cnt = 0u32;
```
theduke, line 88: nitpick: it's pretty easy to confuse `cnt` and `count`.
Maybe `count_woken` ?

theduke (commented):
One thing to fix, and some recommendations.

I like moving this to the memory.

*Note*: this can be a separate PR, but we also need some external way to know when a thread goes to sleep (wait) and is woken up (notify).

Moving this to the memory is actually great for that, because now we should be able to add it to the trait, so the behaviour of wait/notify can be customized.

You see any problems with that?

lib/vm/src/instance/mod.rs
```diff
@@ -861,7 +821,8 @@ impl Instance {
 
         if let Ok(mut ret) = ret {
             if ret == 0 {
-                ret = self.do_wait(memory_index.as_u32(), dst, timeout);
+                let memory = self.get_local_vmmemory_mut(memory_index);
+                ret = memory.do_wait(dst, timeout);
             }
             if ret == 0xffff {
```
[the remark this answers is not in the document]
theduke: (everywhere where 0xffff is used)

ptitSeb:
> 

I'm fine with that :)

lib/vm/src/threadconditions.rs
```diff
@@ -0,0 +1,108 @@
...
+        if let Some(v) = conds.map.get_mut(&key) {
+            for waiter in v {
+                if cnt < count && !waiter.notified {
+                    waiter.notified = true; // mark as was waiked up
+                    waiter.thread.unpark(); // wakeup!
+                    cnt += 1;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
ptitSeb: The removal from the list is done  in `do_wait`. This is because the `park_timeout` doesn't gives any info if the timeout has been reached or not, so I gather the information in an inderect whay with the notify flags. Also, I also need to remove waiters that have timed out anyway.

lib/vm/src/memory.rs
ptitSeb: Ok, I'll do that

lib/vm/src/threadconditions.rs
```diff
@@ -0,0 +1,108 @@
...
+#[derive(Debug, Default)]
+struct NotifyMap {
+    pub map: HashMap<NotifyLocation, Vec<NotifyWaiter>>,
+}
+
+/// HashMap of Waiters for the Thread/Notify opcodes
+#[derive(Debug)]
+pub struct ThreadConditions {
+    inner: Arc<Mutex<NotifyMap>>, // The Hasmap with the Notify for the Notify/wait opcodes
+}
```
syrusakbary, line 24: Have we taken a look on dashmap? it might be useful here (instead of locking)

https://github.com/xacrimon/dashmap

theduke: Yeah, this lock will probably be pretty hot in multi-threaded workloads.

We should also look into using `fnv` or https://github.com/rust-lang/rustc-hash for a small and fast hash function, no need to use a secure hash here.

lib/vm/src/memory.rs
ptitSeb: done

lib/vm/src/threadconditions.rs
ptitSeb: done and done

john-sharratt:
Hey guys - this is not asynchronous thus it will block execution and cause quite some performance bottlenecks on the WASI implementation with `tokio`. Have you put consideration into how the thread will be unloaded into an asynchronous context?

theduke:
@john-sharratt we will add functionality to extend the wait/notify behaviour in the `LinearMemory` trait, but I haven't in detail thought about how to propagate this . 

It will be difficult while also being able to snapshot the stack. 

Need to talk/think about this, but that can be done as a followup, not in this PR.

lib/vm/src/instance/mod.rs
```diff
@@ -888,14 +856,21 @@ impl Instance {
...
-
         if let Ok(mut ret) = ret {
             if ret == 0 {
-                ret = self.do_wait(memory_index.as_u32(), dst, timeout);
-            }
-            if ret == 0xffff {
-                // ret is 0xffff if there is more than 2^32 waiter in queue
-                return Err(Trap::lib(TrapCode::TableAccessOutOfBounds));
+                let memory = self.get_vmmemory_mut(memory_index);
+                let location = NotifyLocation { address: dst };
```
theduke, line 862: This is quite a lot of duplicated code for local/non local.

We should only duplicate the code for retrieving the memory, the other logic should be in a common function.

lib/vm/src/instance/mod.rs
```diff
@@ -920,11 +895,19 @@ impl Instance {
...
         if let Ok(mut ret) = ret {
             if ret == 0 {
-                ret = self.do_wait(memory_index.as_u32(), dst, timeout);
-            }
-            if ret == 0xffff {
-                // ret is 0xffff if there is more than 2^32 waiter in queue
-                return Err(Trap::lib(TrapCode::TableAccessOutOfBounds));
+                let memory = self.get_local_vmmemory_mut(memory_index);
+                let location = NotifyLocation { address: dst };
+                let timeout = if timeout < 0 {
```
theduke, line 900: As above, duplication.

lib/vm/src/instance/mod.rs
```diff
@@ -950,51 +933,37 @@ impl Instance {
...
         if let Ok(mut ret) = ret {
             if ret == 0 {
-                ret = self.do_wait(memory_index.as_u32(), dst, timeout);
-            }
-            if ret == 0xffff {
-                // ret is 0xffff if there is more than 2^32 waiter in queue
-                return Err(Trap::lib(TrapCode::TableAccessOutOfBounds));
+                let memory = self.get_vmmemory_mut(memory_index);
+                let location = NotifyLocation { address: dst };
+                let timeout = if timeout < 0 {
```
theduke, line 938: As above... we have the same code 4 times.

lib/vm/src/memory.rs
```diff
@@ -616,4 +643,15 @@ where
 
     /// Copies this memory to a new memory
     fn duplicate(&mut self) -> Result<Box<dyn LinearMemory + 'static>, MemoryError>;
+
+    /// Add current thread to the waiter hash, and wait until notified or timout.
+    /// Return 0 if the waiter has been notified, 2 if the timeout occured, or None if en error happened
+    fn do_wait(&mut self, _dst: NotifyLocation, _timeout: Option<Duration>) -> Option<u32> {
```
theduke, line 649: We should return a `Result<_, Trap` here to allow being specific about the trap. 

See some related comments on this.

lib/vm/src/instance/mod.rs
```diff
@@ -861,11 +821,19 @@ impl Instance {
...
+                    None
+                } else {
+                    Some(std::time::Duration::from_nanos(timeout as u64))
+                };
+                let waiter = memory.do_wait(location, timeout);
+                if waiter.is_none() {
+                    // ret is None if there is more than 2^32 waiter in queue or some other error
+                    return Err(Trap::lib(TrapCode::TableAccessOutOfBounds));
```
theduke, line 834: We need to make sure the behaviour here matches the spec, and that we trap with an appropriate code (for allowing proper debugging).

See some related comments below.

lib/vm/src/memory.rs
theduke: This will be especially important when we allow customizing this behaviour via a trait.

lib/vm/src/threadconditions.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl Clone for ThreadConditions {
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+        }
+    }
+}
+
```
theduke, line 109:
```
#[cfg(test)]
mod tests {

}
```

theduke (changes requested):

lib/vm/src/instance/mod.rs
theduke: Note: this currently matches the spec, because it only returns `None` if the limit is exceeded, but this would be pretty easy to mess up in a future refactor, and we also want to add trait extensibility.

lib/vm/src/memory.rs
ptitSeb: But what Trap should be returned for the non-implemented path?

lib/vm/src/instance/mod.rs
ptitSeb: I change to return a Result with custom error instead @theduke . Is that better now?

lib/vm/src/memory.rs
ptitSeb: I change to return a Result with custom error instead @theduke . Is that better now?

theduke (approved):
LGTM

lib/vm/src/instance/mod.rs
theduke: Yes!

PR #3724, opened 2023-03-30
ci: Cancel previous workflow runs when new commits are pushed

theduke:
Not having this introduces pointless bloat to the CI, and might delay newer runs if a
lot of branches are being built.

There's no value in finishing the previous runs.


ptitSeb (approved):
indeed

PR #3730, opened 2023-04-02
Added a supported stuff flags list thats detected when initializing a module

john-sharratt:
This pull request clears up some left overs from the WASIX merge and introduces an expanded list of support flags that can be used by runtimes and supporting infrastructure around those runtimes.


syrusakbary:
I don't think is longer needed, but we need follow up

theduke:
This is most likely stale and no longer needed.

Issue #3678, opened 2023-03-14
the `UnionFileSystem` and `mem_fs::FileSystem` don't allow mounting overlapping directories


Michael-F-Bryan:
### Describe the bug

As part of #3677, I wanted to mount different folders from the host system to overlapping folders in the guest (e.g. `/tmp/first/` goes to `/mapped/` and `/home/michael/` goes to `/mapped/nested/`). It doesn't look like this use case is supported by either `UnionFileSystem` or `mem_fs::FileSystem` at the moment.

### Steps to reproduce

Here are some tests we can add to the `wasmer-vfs` crate.

```rust
// lib/vfs/src/union_fs.rs

use crate::{FileSystem, FileSystemExt, unionFileSystem, mem_fs};

#[test]
fn mount_to_overlapping_directories() {
    let top_level = mem_fs::FileSystem::default();
    top_level.touch("/file.txt").unwrap();
    let nested = mem_fs::FileSystem::default();
    nested.touch("/another-file.txt").unwrap();

    let mut fs = UnionFileSystem::default();
    fs.mount(
        "top-level",
        "/",
        false,
        Box::new(top_level),
        Some("/top-level"),
    );
    fs.mount(
        "nested",
        "/",
        false,
        Box::new(nested),
        Some("/top-level/nested"),
    );

    assert!(fs.is_dir("/top-level"));
    assert!(fs.is_file("/top-level/file.txt"));
    assert!(fs.is_dir("/top-level/nested"));
    assert!(fs.is_file("/top-level/nested/another-file.txt"));
}
```

```rust
// lib/vfs/src/mem_fs/filesystem.rs

#[test]
fn mount_to_overlapping_directories() {
    let top_level = FileSystem::default();
    top_level.touch("/file.txt").unwrap();
    let nested = FileSystem::default();
    nested.touch("/another-file.txt").unwrap();
    let top_level: Arc<dyn crate::FileSystem + Send + Sync> = Arc::new(top_level);
    let nested: Arc<dyn crate::FileSystem + Send + Sync> = Arc::new(nested);

    let fs = FileSystem::default();
    fs.mount("/top-level".into(), &top_level, "/".into())
        .unwrap();
    fs.mount("/top-level/nested".into(), &nested, "/".into())
        .unwrap();

    assert!(fs.is_dir("/top-level"));
    assert!(fs.is_file("/top-level/file.txt"));
    assert!(fs.is_dir("/top-level/nested"));
[...]

PR #3737, opened 2023-04-03
Fixed FileSystem createdir when parent_inode is an ArcDirectory

ptitSeb:
Fixed FileSystem createdir when parent_inode is an ArcDirectory

The `inode_of_parent(...)` shoudl return `InodeResolution::Redirect(...)` when parent is an `ArcDirectory`

### DIFF
--- a/lib/vfs/src/mem_fs/filesystem.rs
+++ b/lib/vfs/src/mem_fs/filesystem.rs
@@ -605,8 +605,8 @@ impl FileSystemInner {
                     Some(Node::Directory(DirectoryNode { .. })) => {
                         Ok(InodeResolution::Found(inode_of_parent))
                     }
-                    Some(Node::ArcDirectory(ArcDirectoryNode { .. })) => {
-                        Ok(InodeResolution::Found(inode_of_parent))
+                    Some(Node::ArcDirectory(ArcDirectoryNode { fs, path, .. })) => {
+                        Ok(InodeResolution::Redirect(fs.clone(), path.clone()))
                     }
                     _ => Err(FsError::BaseNotDirectory),
                 }


lib/vfs/src/mem_fs/filesystem.rs
```diff
@@ -605,8 +605,8 @@ impl FileSystemInner {
                     Some(Node::Directory(DirectoryNode { .. })) => {
                         Ok(InodeResolution::Found(inode_of_parent))
                     }
-                    Some(Node::ArcDirectory(ArcDirectoryNode { .. })) => {
-                        Ok(InodeResolution::Found(inode_of_parent))
+                    Some(Node::ArcDirectory(ArcDirectoryNode { fs, path, .. })) => {
+                        Ok(InodeResolution::Redirect(fs.clone(), path.clone()))
```
theduke, line 609: Nice find. 

Let's either add a small unit test for this here, or create a follow up issue to add one.

theduke (approved):

syrusakbary (approved):

lib/vfs/src/mem_fs/filesystem.rs
ptitSeb: There isn't a single test with `ArcDirectory`. Not sure when this has been introduced, but it is test-less

ptitSeb: Added a mention of this is ticket https://github.com/wasmerio/wasmer/issues/3678 similar this.

theduke: I'll add an explicit issue, because otherwise we will forget.

PR #3739, opened 2023-04-04
Use cache in runners

syrusakbary:
Runners were missing cache, which caused loading times to be non-ideal when running containers.

This PR fixes it by adding a new function `with_compiler` to each of the Runners

### DIFF
--- a/lib/cli/src/commands/run_unstable.rs
+++ b/lib/cli/src/commands/run_unstable.rs
@@ -128,12 +128,15 @@ impl RunUnstable {
             .get(id)
             .with_context(|| format!("Unable to get metadata for the \"{id}\" command"))?;
 
-        // TODO(Michael-F-Bryan): Refactor the wasmer_wasi::runners::Runner
-        // trait So we can check whether a command is supported by a runner
-        // without needing to go through and instantiate each one.
-
         let (store, _compiler_type) = self.store.get_store()?;
-        match command.runner.as_ref() {
+        let runner_base = command
+            .runner
+            .as_str()
+            .split_once('@')
+            .map(|(base, version)| base)
+            .unwrap_or_else(|| command.runner.as_str());
+
+        match runner_base {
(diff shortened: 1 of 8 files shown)


lib/wasi/Cargo.toml
```diff
@@ -58,7 +58,7 @@ pin-project = "1.0.12"
 # Used by the WCGI runner
 hyper = { version = "0.14", features = ["server", "stream"], optional = true }
 wcgi = { version = "0.1.1", optional = true }
-wcgi-host = { version = "0.1.0", optional = true }
+wcgi-host = { version = "0.1.1", optional = true }
```
syrusakbary, line 61: This was a catch-up thing that I intentionally made part of the PR to simplify dependencies

ptitSeb (approved):
Looks good to me

lib/cli/src/commands/run_unstable.rs
```diff
@@ -129,41 +133,54 @@ impl RunUnstable {
...
-
-        let (store, _compiler_type) = self.store.get_store()?;
-        let mut runner = wasmer_wasix::runners::emscripten::EmscriptenRunner::new(store);
-        runner.set_args(self.args.clone());
-        if runner.can_run_command(id, command).unwrap_or(false) {
-            return runner
-                .run_cmd(container, id)
-                .context("Emscripten runner failed");
-        }
+        match command.runner.as_ref() {
```
Michael-F-Bryan, line 136: Let's use the constants defined in `webc::metadata:: annotations` instead of string literals. We don't need to worry about things like `runner = "wcgi"` because `wapm2pirita` will automatically convert that to the associated URI.

syrusakbary: Fixed

syrusakbary:
The tests that didn't pass are just flaky. Merging manually

PR #3740, opened 2023-04-04
Fix fuzz build

manunio:
# Description
While working on build failure issue: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=56795, some fuzz targets were not building, this pr attempts to fix that.


ptitSeb (approved):
Thank you

Issue #3746, opened 2023-04-05
Chinese Documentation Broken Links


dynamite-bud:
The links in the Chinese Documentation are broken.
Like the rust crate for chinese documentation links to: [link](https://wasmerio.github.io/wasmer/crates/wasmer)

ptitSeb:
I'll to fix as much as I can, but that will be for next beta

PR #3750, opened 2023-04-06
Fixes Chinese docs links (for #3746)

ptitSeb:
Fixes Chinese docs links.
Address ticket #3746


dynamite-bud (changes requested):
Can you please change the link for php wasm with the following https://wasmerio.github.io/wasmer-php/

That also seems to be broken at the current moment

dynamite-bud (approved):

PR #3753, opened 2023-04-07
Upgrade webc from webc 5.0.0-rc.5 to webc 5.0.0-rc.6

john-sharratt:


Michael-F-Bryan (approved):

ptitSeb (approved):
It's ok, but why not use `rc.7`?

Issue #2808, opened 2022-02-25
Stack probing on AArch64


Amanieu:
Currently neither LLVM nor Cranelift implement stack probing on AArch64, which means that malicious Wasm code could potentially write to memory past the end of the stack.

Related:
- Rust issue: https://github.com/rust-lang/rust/issues/77071
- LLVM patch: https://reviews.llvm.org/D96004

afonso360:
With https://github.com/bytecodealliance/wasmtime/pull/5353 merged, Cranelift now supports inline stack probing.

syrusakbary:
Yup, we probably need to use this setting to enable in cranelift:

```
set enable_probestack=true
set probestack_strategy=inline
```

theduke:
Not stale.

PR #3756, opened 2023-04-08
Aarch64 stackprobe

syrusakbary:
This PR addresses #2808  (only Cranelift) by using the inline stack option introduced in Cranelift 0.91.1

It also reenables most of the snapshot testing in macos


ptitSeb:
I suppose this test `cranelift+aarch64 spec::skip_stack_guard_page` can also be re-enable from `ignores.txt`

syrusakbary:
Done @ptitSeb , I think tests should be passing if you want to approve!

ptitSeb (approved):

PR #3757, opened 2023-04-08
WASI without VM internals

syrusakbary:
This PR removes the dependency of VM internals inside of WASI by using `Memory` instead of `VMMemory` or `VMSharedMemory`.

This makes possible to also use WASI in some environments where the internals might be different (JavascriptCore, to put an example), and also removes the exposure of sensible internals.

Basically, it makes the API more robust


syrusakbary:
@john-sharratt all snapshot tests are passing (including macos). I think we should be good to go!

john-sharratt (approved):

syrusakbary:
Merging directly as only one flaky test failed

PR #3758, opened 2023-04-08
More and better snapshot tests

john-sharratt:
Made the epoll test more deterministic and added a new test for testing multi-threaded memory access


syrusakbary (approved):
Good to go once tests pass

PR #3761, opened 2023-04-11
Fix error in Korean translation

code-yeongyu:
# Description

- The current version of Korean README says
```md
Secure by default. The access to file, network, environment cannot be enabled explicitly.
```
- And I changed the wrong translation to:
```md
Secure by default. Unless you set it explicitly, file, network, and environment cannot be accessed.
```


ptitSeb (approved):
Thank you

Issue #3762, opened 2023-04-11
Implement `NativeWasmTypeInto` for `u32` and `u64` with the `js` feature.


kajacx:
### Motivation

When using wasmer on the web (wasmer itself is compiled to WASM and then loads another WASM module) with the `js` feature, you cannot pass `u32` and `u64` as arguments from / to functions, because those types do not implement `NativeWasmTypeInto`.

### Proposed solution

Implement `NativeWasmTypeInto` for `u32` and `u64`.

### Alternatives

You can, of course, cast your unsigned integers to signed integers, but it is kind of annoying. Example:

```rust
// Plugin:
#[no_mangle]
pub fn add_ten(value: u32) -> u32 {
    value + 10
}

// Host, current workaround:
let mut value: u32 = 50;
let add_ten = instance
    .exports
    .get_typed_function::<i32, i32>(&store, "add_ten")
    .unwrap();
value = take_u32.call(&mut store, value as i32).unwrap() as u32;

// Host, with proposed change:
let mut value: u32 = 50;
let add_ten = instance
    .exports
    .get_typed_function::<u32, u32>(&store, "add_ten")
    .unwrap();
value = take_u32.call(&mut store, value).unwrap();
```

### Additional context

Not sure if implementing `NativeWasmTypeInto` for `u8` / `u16` / `i8` / `i16` makes sense in the same way, but that can be another feature request.

syrusakbary:
That sounds good to me. Do you want to work on a PR @kajacx ?

PR #3763, opened 2023-04-12
Implement NativeWasmTypeInto for u32 and u64

kajacx:
Implements feature request [#3762](https://github.com/wasmerio/wasmer/issues/3762)

# Description

Implements `NativeWasmTypeInto` and `NativeWasmType` for `u32` and `u64` on the web target (`js` feature). See the feature request for more detail.


kajacx:
I was not sure what `<u32 as NativeWasmType>::Abi` should be, so I left it as `Self` ([link](https://github.com/kajacx/wasmer/commit/64fda393fa7bef007543ce5009d4466dc0aa3c36#diff-117edb4fcb522fbe8b03425aad1caa110986c634f06ca686f65afce48d24c059R183)). If it should always be the same as the `WASM_TYPE`, then it probably wouldn't need to be 2 separate lines of code.

But if needs to be `i32` to match the `WASM_TYPE`, I can of course change it. Also, I was not sure how to test it, I did some manual testing, but my I was having some problems with wasm-bindgen, so I couldn't test it as thoroughly as I wanted. Is there an automated testing strategy for something like this? The test would probably need to run in the browser, not sure how that would work.

syrusakbary (approved):

kajacx:
I see the PR approved, but after some more testing, I found many things work when passing around big `u64` values and even big `i64` values. I will have to re-check everything and create a minimal reproducible example, but it seems that the `BigInt` conversion is not working correctly.

[This](https://github.com/wasmerio/wasmer/blob/2cb70e6d251290f3855e0fb4a20e0e1a721c2faa/lib/api/src/js/as_js.rs#L40) is the part where it goes wrong. I think this was added when I wanted to pass `u64` values around for fp-bindgen a few months back, but it doesn't seem to work so well with very large values like `u64::MAX - 5` for example.

kajacx:
@syrusakbary It seems that this has been already implemented on master in the meantime. See [api/src/native_type.rs](https://github.com/wasmerio/wasmer/blob/master/lib/api/src/native_type.rs#L102) and [types/src/lib.rs](https://github.com/wasmerio/wasmer/blob/master/lib/types/src/lib.rs#L187). I have only added a test to be sure it stays working.

PR #3767, opened 2023-04-14
Fix Snapshot0 fd_filestat_get and path_filestat_get

ptitSeb:
Change those 2 syscall  to not use clobbering technique 

This fixes flakyness of python,  that was observed on the CI test or localy

### DIFF
--- a/lib/wasi/src/syscalls/wasi/fd_filestat_get.rs
+++ b/lib/wasi/src/syscalls/wasi/fd_filestat_get.rs
@@ -1,5 +1,6 @@
 use super::*;
 use crate::syscalls::*;
+use crate::types::wasi::Snapshot0Filestat;
 
 /// ### `fd_filestat_get()`
 /// Get the metadata of an open file
@@ -15,7 +16,14 @@ pub fn fd_filestat_get<M: MemorySize>(
     fd: WasiFd,
     buf: WasmPtr<Filestat, M>,
 ) -> Errno {
-    fd_filestat_get_internal(&mut ctx, fd, buf)
+    let stat = wasi_try!(fd_filestat_get_internal(&mut ctx, fd));
+
+    let env = ctx.data();
+    let (memory, _) = env.get_memory_and_wasi_state(&ctx, 0);
+    let buf = buf.deref(&memory);
+    wasi_try_mem!(buf.write(stat));
+
+    Errno::Success
 }
 
 /// ### `fd_filestat_get()`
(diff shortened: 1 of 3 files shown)


lib/wasi/src/syscalls/wasi/fd_filestat_get.rs
```diff
@@ -26,22 +34,51 @@ pub fn fd_filestat_get<M: MemorySize>(
...
+/// ### `fd_filestat_get_old()`
+/// Get the metadata of an open file
+/// Input:
+/// - `Fd fd`
+///     The open file descriptor whose metadata will be read
+/// Output:
+/// - `Snapshot0Filestat *buf`
+///     Where the metadata from `fd` will be written
+#[instrument(level = "debug", skip_all, fields(%fd), ret)]
+pub fn fd_filestat_get_old<M: MemorySize>(
```
theduke, line 60: I'd put the `_old` functions in `snapshot0.rs` and instead make the `_internal` functions `pub(in crate::syscalls)` instead. 

To prevent confusion.

lib/wasi/src/syscalls/wasi/path_filestat_get.rs
```diff
@@ -92,3 +81,61 @@ pub(crate) fn path_filestat_get_internal(
...
+/// - `const char *path`
+///     String containing the file path
+/// - `u32 path_len`
+///     The length of the `path` string
+/// Output:
+/// - `__wasi_file_stat_t *buf`
+///     The location where the metadata will be stored
+#[instrument(level = "trace", skip_all, fields(%fd, path = field::Empty), ret)]
+pub fn path_filestat_get_old<M: MemorySize>(
+    ctx: FunctionEnvMut<'_, WasiEnv>,
```
theduke, line 101: As above

theduke (approved):
Uff, nice find!

lib/wasi/src/syscalls/wasi/fd_filestat_get.rs
ptitSeb: I have left them in the same file because they shared the same logic, and I'm afraid we might get desynchronized if we separate the "_old" one from the new one

lib/wasi/src/syscalls/wasi/path_filestat_get.rs
ptitSeb: I think it's safer to left them in the same file actualy

Issue #3760, opened 2023-04-09
RuntimeError: call stack exhausted


hungryzzz:
### Summary

When I use Wasmer to run a case which contains some recursive function calls, Wasmer will raise **_RuntimeError: call stack exhausted_**. The original Cpp program can run successfully. 

Can I adjust the call stack size of Wasmer to address this problem?

Thank you!

### Additional details

**wasmer version:** wasmer 3.2.0-alpha.1

**emscripten version:** 
emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.24 (68a9f990429e0bcfb63b1cde68bad792554350a5)
clang version 16.0.0 (https://github.com/llvm/llvm-project 277c382760bf9575cfa2eac73d5ad1db91466d3f)
Target: wasm32-unknown-emscripten
Thread model: posix

**emscripten compile cmd:**
`emcc -sENVIRONMENT=shell -O2 -s TOTAL_MEMORY=1024MB -s WASM=1 ackermann.cpp -o ackermann.wasm`

[case.zip](https://github.com/wasmerio/wasmer/files/11185871/case.zip)

Michael-F-Bryan:
@hungryzzz the stack size isn't configurable through Wasmer. This "call stack exhausted" may be because emscripten decides the stack size and the default size is quite small. Try passing something like `-s TOTAL_STACK=2MB` to `emcc` to modify this behaviour.

hungryzzz:
@Michael-F-Bryan Hi, I've compiled with the option `-s TOTAL_STACK=2MB`, but Wasmer still raises the same runtime error.
I am wondering that, when Wasmer runs a wasm case, what is the default size of the execution stack space allocated to the case?

theduke:
You should increase the stack size, because `5MB` is the default used by emscripten.

hungryzzz:
@theduke Thank you, I've tried it under `50MB` but it still failed in Wamser.

ptitSeb:
Can you run the program under a debugger to see if you have a segfault, and what would be the native backtrace when it occurs.?

ptitSeb:
Fixed on master. use the `--stack-size` parameter to set a bigger stack

PR #3786, opened 2023-04-19
Made Stack Size parametrable (for #3760)

ptitSeb:
Made Stack Size parametrable.

By default, Stacks are 1MB. It is generaly a good default value, but it might be to small for some specific case, where recursion is (ab)used a lot, like for ticket #3760 .
In some other special case, with a lot of thread, smaller value might also be prefered.


lib/vm/src/trap/traphandlers.rs
```diff
@@ -29,6 +29,18 @@ use wasmer_types::TrapCode;
 // On Arm64, the udf alows for a 16bits values, so we'll use the same 0xC? to store the trapinfo
 static MAGIC: u8 = 0xc0;
 
+static mut DEFAULT_STACK_SIZE: usize = 1024 * 1024;
```
Michael-F-Bryan, line 32: Do we need to use a `static mut` variable here? If we make it an `AtomicUsize` we can drop the `unsafe`.

Michael-F-Bryan (commented):
I think it should be fine to make this a global variable here instead of trying to thread the stack size through `Tumables`. 

We should probably use an atomic or `Mutex` for the variable so we can drop the `unsafe`, though.

lib/vm/src/trap/traphandlers.rs
ptitSeb: Yeah, it's much cleaner to use Atomic, and I don't think it will have much impact on performances...

theduke:
I disagree, we very well may want different stack sizes per instance. 

I can think of a concrete use case where users can configure the stack size per workload...

This global atomic approach won't work for this. 

A global atomic won't work when multiple threads are constructing instances. 

In general I'm not a fan of a global static setting.

This should be threaded through, even though it's annoying. 

Note that tunables isn't ideal here either, because that would require constructing a separate engine.

@ptitSeb , does the chosen stack size affect compilation?

theduke:
I'm not talking about changing the stack size for an existing instance, just configuring the size for a newly created one.

Michael-F-Bryan (approved):

Issue #3787, opened 2023-04-19
3.2.0 Release Docs Fail to Build on docs.rs


theduke:
https://docs.rs/crate/wasmer/latest

PR #3791, opened 2023-04-20
Fix doc api

ptitSeb:
This fix the building of the API Doc

It also add an option in the Makefile to build the API Doc
And a CI test too.

Address #3787


Michael-F-Bryan (approved):
After chatting with @ptitSeb, we're going to keep the `make build-docs-api` as-is and make the "proper" fix as part of our upcoming build system work.

theduke (approved):

Issue #3773, opened 2023-04-17
Passing i64 around always fails with the js feature


kajacx:
### Passing `i64` to and from a module on the web always fails with the `js` feature

When compiling wasmer itself to WASM using the `js` feature so that it can run in the browser, there is an error when you then load another plugin from wasmer and try to pass `i64` values around.

### Steps to reproduce

1. Clone [this repo](https://github.com/kajacx/wasmer-browser-u64/tree/add-i64-result-error) at the `add-i64-result-error` tag.
2. Run the `build.sh` script (after reading the entire source code, of course).
3. Open the webpage and see the result.

### Expected behavior

`i64` values should be added correctly inside the plugin and the values should be passed between the plugin and wasmer without a problem.

### Actual behavior

Calling the `add_i64` exported function always returns a runtime error, example:

```
RuntimeError { source: Js(JsValue(TypeError: can't convert 5 to BigInt
getImports/imports.wbg.__wbg_apply_aedce30790c00792/<@http://127.0.0.1:8094/hello_wasmer_u64.js:414:37
handleError@http://127.0.0.1:8094/hello_wasmer_u64.js:224:18
getImports/imports.wbg.__wbg_apply_aedce30790c00792@http://127.0.0.1:8094/hello_wasmer_u64.js:413:68
@http://127.0.0.1:8094/hello_wasmer_u64_bg.wasm:wasm-function[199]:0x21883
@http://127.0.0.1:8094/hello_wasmer_u64_bg.wasm:wasm-function[43]:0x2b1e
test_add_i64@http://127.0.0.1:8094/hello_wasmer_u64.js:212:14
@http://127.0.0.1:8094/:13:15
promise callback*@http://127.0.0.1:8094/:11:14
)) }
```

### Additional context

This was discovered when I was doing testing for [#3763](https://github.com/wasmerio/wasmer/pull/3763). Although this is a different error than I was experiencing there. That might be because I was using wasmer from source in that PR, whereas here I am using the currently released version `3.1.1`.

Also, I followed [this example](https://developer.mozilla.org/en-US/docs/WebAssembly/Rust_to_wasm) by Mozilla on how to run Rust on the browser using `wasm-pack`, if that's relevant.

ptitSeb:
Can you try again with current master, mainy things have changed on the Js part so it might just work now.

kajacx:
> Can you try again with current master, many things have changed on the Js part so it might just work now.

[...]

PR #3796, opened 2023-04-20
Fix returning negative i64 values on web target

kajacx:
# Description

Fixes [#3773](https://github.com/wasmerio/wasmer/issues/3773), which is a bug where returning a negative `i64` from a module caused a crash when wasmer itself was running inside the browser using the `js` feature.

Looking at [`js-sys`'s implementation of `TryFrom`](https://github.com/rustwasm/wasm-bindgen/blob/main/crates/js-sys/src/lib.rs#L1248), it looks like `BigInt` is converted to `JsValue` anyway before converting it further, so converting `BigInt` to `JsValue` just so it is converted back seems like extra work.

To be honest, this needs an automated test, but I don't really know to make one. Can node load wasmer with the `js` feature? That is worth looking into.


syrusakbary (commented):
Can we add tests verifying that the issue is fixed?

kajacx:
> Can we add tests verifying that the issue is fixed?

Yes, I wanted to do that, but I do know where to begin. I am looking into the [`tests`](https://github.com/wasmerio/wasmer/tree/master/tests) folder in the repo, but I don't see any tests for the `js` web target that I could easily add too.

The closest file I could find is [`types_functions.rs`](https://github.com/wasmerio/wasmer/blob/master/tests/compilers/typed_functions.rs), but I do not know if the automated test suit already runs this with the `js` feature or not.

kajacx:
Ok, I have tried adding [this test](https://github.com/kajacx/wasmer-browser-u64/blob/example-add-three-i64-test/hello-wasmer-u64/src/lib.rs). It is loading the WASM plugin from a `wat` text just like the other tests, which makes it simpler to work with.

The plugin exports a function that adds one, then calls the imported "add one" function and then adds one again to add three in total. This tests back and forth exchanging of values between the plugin and the host. Same idea as I did with [fp-bindgen](https://github.com/fiberplane/fp-bindgen/commit/de72d30de96a019de80c90069251a851a926101d#diff-9068ccd40e5b9740c264b07c406a1842e060557aca98cd914e78e93931e7b567).

I am testing it with interesting values that might overflow (near `0`, `i64::MIN` and `i64::MAX`).

I have verified that this test fails on the current `master` branch but succeeds with this fix.

But what I am missing is a way to run this automatically. It seems that node can run WASM, but I will have to try that.

syrusakbary:
Cool, feel free to add the test inside `lib/api/tests/import_function.rs` (you will need to replace `#[wasm_bindgen]` with `#[universal_test]`).

You can also that the test is working by running `make test-js-api` @kajacx

kajacx:
> Cool, feel free to add the test inside `lib/api/tests/import_function.rs` (you will need to replace `#[wasm_bindgen]` with `#[universal_test]`). Also, try to make the test as generic as possible (no wasm bindgen or js specific stuff such as `console_error_panic_hook`, as that's done automatically by the universal test)
> 
> You can also that the test is working by running `make test-js-api` @kajacx

Ok, I have tried that, but I cannot get it to work. There seems to be some problem with `i64` values on node. I checked out the current master, and ran the `make-test-api` command, but it gave the following error: ([full error message](https://pastebin.com/ntpPrFWG))

```
---- module::calling_host_functions_with_negative_values_works_js output ----
    error output:
        panicked at 'called `Result::unwrap()` on an `Err` value: "RuntimeError { source: Js(JsValue(TypeError: wasm function signature contains illegal type
TypeError: wasm function signature contains illegal type
```

After doing some searching, I have found [this article](https://github.com/emscripten-core/emsdk/issues/476) about bigint support in node, but I have not figures out how to pass the `--experimental-wasm-bigint` flag to node. Any help @syrusakbary? Thanks.

syrusakbary:
Tests passed in CI so I suspect that you may have an old version of Node installed in your system @kajacx

kajacx:
I am ill right now, so I will not be doing anything on the computer for a few weeks. You can merge the PR if it looks good, I will re-check it to my satisfaction later either way.

ptitSeb:
@syrusakbary it seems ready to be merged?

syrusakbary:
Yup, good to merge!

syrusakbary (approved):

Issue #3789, opened 2023-04-20
WASI-based runners don't pass arguments to the WebAssembly instance correctly


Michael-F-Bryan:
### Describe the bug

@ayys and I found a bug in the way command-line arguments are passed to the WebAssembly instance. It looks like some arguments are being passed in twice, which messes up things for the program being executed.

```console
$ wasmer --version --verbose
wasmer 3.1.0 (0e336 2022-12-12)
binary: wasmer-cli
commit-hash: 0e3369a9d906513eb90cfac2e3160d8c2ded4247
commit-date: 2022-12-12
host: aarch64-apple-darwin
compiler: singlepass,cranelift
$ rustc --version --verbose
rustc 1.65.0 (897e37553 2022-11-02)
binary: rustc
commit-hash: 897e37553bba8b42751c67658967889d11ecd120
commit-date: 2022-11-02
host: aarch64-apple-darwin
release: 1.65.0
LLVM version: 15.0.0
```

### Steps to reproduce

```console
$ wasmer run-unstable saghul/quickjs --command=qjs -- --eval 'console.log(scriptArgs)' hello world "hello world"
hello,world,hello world,--eval,console.log(scriptArgs),hello,world,hello world
```

The output should be just `--eval,console.log(scriptArgs),hello,world,hello world`.

It seems like command-line arguments are being sent through twice, but for whatever reason anything with an argument gets skipped.

### Additional context

This seems to be an issue in the way arguments are constructed for a WASI-based runner. We tried running `quickjs.wasm` directly and it behaved correctly.

A good place to start looking would be here:

https://github.com/wasmerio/wasmer/blob/8a343d0f9a1d3d4b4bfe622ed345253d780a3d76/lib/wasi/src/runners/wasi_common.rs#L70-L76

PR #3805, opened 2023-04-21
Wasi runner args

Michael-F-Bryan:
Turns out we were adding arguments to the `WasiEnvBuilder` twice when setting up WASI-based runners. I'm guessing I forgot to remove the original `add_args()` line when I split out `populate_args()`.

Fixes https://github.com/wasmerio/wasmer/issues/3789.


ptitSeb (approved):
look good to me

PR #3813, opened 2023-04-22
Better errors

syrusakbary:
This PR tries to simplify the error implementation in Wasmer, by moving the frames into `wasmer_types` and the RuntimeError directly to the `wasmer` api (instead of the compiler/engine)


syrusakbary:
All tests passed, and I tested them locally as well. Merging this as is mostly cleaning up stuff and not doing any significant change, and we require it for JSC

Issue #3794, opened 2023-04-20
Runners are unable to map to relative directories


Michael-F-Bryan:
### Describe the bug

A common thing you'll want to do when using runners is to map the current directory on the host to the current directory on the guest (e.g. `--mapdir .:.`). 

When we were re-working runners and wiring them up to `wasmer run`/`wasmer run-unstable` (https://github.com/wasmerio/wasmer/pull/3704, https://github.com/wasmerio/wasmer/pull/3599) we implemented host directory mapping by mounting each folder to a `virtual_fs::FileSystem`. However, the `virtual_fs::mem_fs::FileSystem` we mount to doesn't let you mount to the `.` directory, and will instead error out with `FsError::InvalidInput`.

The old implementation used to work because we set up the `WasiEnvBuilder` using `map_dir(".", "/")`, and when we mounted directories we would prepend all relative paths with `/` to make them absolute. 

```console
$ wasmer --version --verbose
wasmer 3.1.0 (0e336 2022-12-12)
binary: wasmer-cli
commit-hash: 0e3369a9d906513eb90cfac2e3160d8c2ded4247
commit-date: 2022-12-12
host: aarch64-apple-darwin
compiler: singlepass,cranelift
$ rustc --version --verbose
rustc 1.65.0 (897e37553 2022-11-02)
binary: rustc
commit-hash: 897e37553bba8b42751c67658967889d11ecd120
commit-date: 2022-11-02
host: aarch64-apple-darwin
release: 1.65.0
LLVM version: 15.0.0
```

### Steps to reproduce

```console
$ wasmer run-unstable wasmer/wapm2pirita --mapdir ".:$PWD"
error: WASI runner failed
╰─▶ 1: invalid input
```

### Expected behavior

The command should start correctly.

### Actual behavior

As part of setting up the WASI runner's filesystem and mounting host directories, we do a `create_dir_all()` to make sure the parent directory exists. This works perfectly fine when the guest path is absolute because we recursively try to create the parent until we reach `/`, which exists, at which point `create_dir_all()` does its thing. 

However, this doesn't work for relative paths (e.g. `./path/to/file.txt`) because we will reach `"."` (or `""`, it doesn't matter which), and the memfs doesn't let you create `.` because it is a "special" file.

### Additional context

This is another symptom of our `virtual_fs::FileSystem` and `WasiFs` abstractions not being aligned. Ideally, there would be one way of mapping directories greater guarantees around how relative paths are treated.

[...]

PR #3818, opened 2023-04-24
Added a quickfix for mounting relative directories

Michael-F-Bryan:
# Description

This adds a test for #3794 and a temporary fix that will make `wasmer run --mapdir .:.` work again.

The proper fix for #3794 will probably require an overhaul of the `virtual_fs::FileSystem` and `WasiFs` layers, which isn't planned for the immediate future.

### DIFF
--- a/lib/wasi/src/runners/wasi_common.rs
+++ b/lib/wasi/src/runners/wasi_common.rs
@@ -98,8 +98,7 @@ fn prepare_filesystem(
             );
 
             if guest.is_relative() {
-                guest = apply_relative_path_mounting_hack(&guest, builder)
-                    .context("Relative path quickfix failed")?;
+                guest = apply_relative_path_mounting_hack(&guest);
             }
 
             if let Some(parent) = guest.parent() {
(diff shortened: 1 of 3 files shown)


ptitSeb (approved):
Looks good to me

lib/wasi/src/runners/wasi_common.rs
```diff
@@ -132,6 +136,31 @@ fn prepare_filesystem(
     Ok(Box::new(OverlayFileSystem::new(root_fs, [container_fs])))
 }
 
+/// HACK: We need this so users can mount host directories at relative paths.
+/// This assumes that the current directory when a runner starts will be "/", so
+/// instead of mounting to a relative path, we just mount to "/$path".
+///
+/// This isn't really a long-term solution because there is no guarantee what
+/// the current directory will be. The WASI spec also doesn't require the
+/// current directory to be part of the "main" filesystem at all, we really
```
theduke, line 145: > *should* be mounting to a relative directory

Not quite sure what you mean there. 
Mounting to a relative directory doesn't make sense, because as you mentioned, the relative directory is an ephemeral thing that always changes in the instance environment.

As I see it, mounting only makes sense with both absolute host and guest paths. 
Auto-expanding a relative to an absolute path makes sense for convenience, but I don't see what else we should be doing.

lib/wasi/src/runners/wasi_common.rs
```diff
@@ -26,14 +26,13 @@ impl CommonWasiOptions {
...
-            builder.add_preopen_dir(path).map_err(Error::from)
-        })?;
+        let fs = prepare_filesystem(&self.mapped_dirs, container_fs, builder)?;
 
         builder.add_preopen_dir("/")?;
-        if fs.read_dir(".".as_ref()).is_ok() {
-            // Sometimes "." won't be mounted so preopening will fail.
-            builder.add_preopen_dir(".")?;
+
+        if self.mapped_dirs.iter().all(|m| m.guest != ".") {
```
theduke, line 33: What happens the user already specified a mount for `/` ? 
Seems like we should check for that and error out otherwise?

Michael-F-Bryan: It should be fine. This just makes sure `.` points to the root directory on startup and doesn't affect any existing mounts on `/`.

PR #3820, opened 2023-04-25
fd_write: Flush on every write to a file

### LINKED ISSUE #3790: `fflush` doesn't work on WASI
### Describe the bug

`fflush` provided by wasi-sdk doesn't work on wasmer.

```sh
wasmer -vV; rustc -vV
wasmer 3.2.0 (8a343d0 2023-04-18)
binary: wasmer-cli
commit-hash: 8a343d0f9a1d3d4b4bfe622ed345253d780a3d76
commit-date: 2023-04-18
sh: 1: rustc: not found
```

### Steps to reproduce

sample C program:

```c
#include <stdio.h>
#include <unistd.h>

int main(int argc, char **argv)
{
  char buf[5];
  int len = 5, ret;

  printf("input something\n");
  printf("> ");
  fflush(stdout);
  ret = read(0, buf, len);
  if (ret > 0) {
    printf("got %.*s (first 5 chars)\n", ret, buf);
  }
  return 0;
}
```

I built this using wasi-sdk 19.0. (`${WASI_SDK_PATH}/bin/clang --sysroot=${WASI_SDK_PATH}/share/wasi-sysroot -o out.wasm main.c`)

On host, prompt (`>`) is printed by `fflush`.

```console
$ gcc main.c ; ./a.out
input something
> hi
got hi
 (first 5 chars)
```

On wasmer, prompt (`>`) isn't printed (isn't flushed).

```console
$ wasmer /app/out.wasm
input something
```

### Expected behavior

fflush works and buffered strings are printed.

### Actual behavior

fflush doesn't work and buffered strings aren't printed.

### LINKED ISSUE #3814: Wasmer stdout is not good
Try to run spidermonkey js.wasm: https://firefoxci.taskcluster-artifacts.net/Wl0gB9DqTUOyBSenX32lXg/0/public/build/js.wasm

```
$ wasmer run js.wasm
1+1 // <---- This is what you input in the shell, and you press enter
js> 2
```

It should be:

```
$ wasmer run js.wasm
js> 1+1 // <---- This is what you input in the shell, and you press enter
2
js> 
```

mucinoab:
# Description

Previously we were only flushing the writes until we closed the file, now every write flushes.
This unfortunately will penalize performance, I am open to suggestions on how can this be avoided or minimized. 

cc @theduke @syrusakbary @ibrahim-akrab

Fixes #3790
Fixes #3814

/claim #3790

### DIFF
--- a/lib/wasi/src/syscalls/wasi/fd_write.rs
+++ b/lib/wasi/src/syscalls/wasi/fd_write.rs
@@ -145,8 +145,9 @@ fn fd_write_internal<M: MemorySize>(
                                         break;
                                     }
                                 }
-                                handle.flush().await.map_err(map_io_err)?;
-
+                                if is_stdio {
+                                    handle.flush().await.map_err(map_io_err)?;
+                                }
                                 Ok(written)
                             }
                         )?
--- a/lib/wasi/src/syscalls/wasi/fd_write.rs
+++ b/lib/wasi/src/syscalls/wasi/fd_write.rs
@@ -145,6 +145,8 @@ fn fd_write_internal<M: MemorySize>(
                                         break;
                                     }
                                 }
+                                handle.flush().await.map_err(map_io_err)?;
+
                                 Ok(written)
                             }
                         )?
--- a/lib/wasi/tests/stdio.rs
+++ b/lib/wasi/tests/stdio.rs
@@ -52,13 +52,13 @@ async fn test_stdout() {
         (export "memory" (memory 0))
 
         ;; Write 'hello world\n' to memory at an offset of 8 bytes
-        ;; Note the trailing newline which is required for the text to appear
-        (data (i32.const 8) "hello world\n")
+        ;; No trailing newline is required since all stdio writes are flushing
+        (data (i32.const 8) "hello world")
 
         (func $main (export "_start")
             ;; Creating a new io vector within linear memory
             (i32.store (i32.const 0) (i32.const 8))  ;; iov.iov_base - This is a pointer to the start of the 'hello world\n' string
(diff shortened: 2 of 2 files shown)


mucinoab:
I want to write a test. Is there any way to examine the state of the env at some point _during_ execution (and not _after_ the execution)? 
Here some relevant test I found: [stdio.rs test_stdout](https://github.com/wasmerio/wasmer/blob/bd75b2b615ac9dc8eb8081e2015fd675178ac8ce/lib/wasi/tests/stdio.rs#L42)

ptitSeb:
Can the flush be done only for stout/stderr, that would avoid the slowdown of flush for regular file.

ibrahim-akrab:
@ptitSeb I mean surely this can be done but then we'll just be waiting for the next issue of someone writing maybe a logger and writing all the logs to a file before uploading it to some api and not getting all the logs uploaded because wasmer didn't find it suitable to flush the data to the file despite the developer explicitly calling flush.

I think it's not a matter of performance penalization issue as much as program behavior correctness.

ptitSeb:
I would prefer that we separate issue. This is issue is specifically about `stdout`/`stderr`. Generic files are backed up by different virtual-file, that might or might not need flushing. There is a complex and efficient cache system in wasmer, and the point is not to nullify it's effect with this. The solution for generic file might be different, or there might be way to detect a flush.

The fix have to only touch `stdout`/`stderr` and not impact other files. If you don't do it, I'll do it after the merge, so I would prefer you do it directly.

ibrahim-akrab:
> you would need to rebase also.

Done

ptitSeb (approved):

PR #3825, opened 2023-04-27
Added support for JavascriptCore

syrusakbary:
Added support for JavascriptCore as a backend in Wasmer, and available through the `jsc` feature in the `wasmer` crate


lib/api/src/externals/table.rs
```diff
@@ -1,5 +1,7 @@
 #[cfg(feature = "js")]
 use crate::js::externals::table as table_impl;
+#[cfg(feature = "jsc")]
```
john-sharratt, line 3: If we introduce `jsc` as a feature, perhaps `js` needs to be renamed to `jsb` (JavascriptBrowser)

lib/api/src/jsc/extern_ref.rs
```diff
@@ -0,0 +1,39 @@
+use std::any::Any;
```
john-sharratt, line 1: Can any of these files be made common between `jsc` and `js`? Perhaps using a `#[path()]` attribute?

lib/api/src/jsc/externals/memory.rs
```diff
@@ -0,0 +1,329 @@
...
+// https://developer.mozilla.org/en-US/docs/Web/API/structuredClone
+// ```js
+// const memory = new WebAssembly.Memory({
+//   initial: 10,
+//   maximum: 100,
+//   shared: true // <--- It must be shared, otherwise structuredClone will fail
+// });
+// structuredClone(memory)
+// ```
+unsafe impl Send for Memory {}
```
john-sharratt, line 36: If we can avoid this it would be better

lib/api/src/jsc/externals/memory.rs
```diff
@@ -0,0 +1,329 @@
...
+            );
+            return Err(MemoryAccessError::HeapOutOfBounds);
+        }
+        unsafe {
+            volatile_memcpy_read(self.base.add(offset as usize), buf.as_mut_ptr(), buf.len());
+        }
+        Ok(())
+    }
+
+    pub(crate) fn read_uninit<'b>(
```
john-sharratt, line 224: Can we make this code common between `js` and `jsc`? It's an area that can easily get bugs so making its footprint smaller will help

lib/api/src/jsc/store.rs
```diff
@@ -0,0 +1,264 @@
...
+            vec![]
+            // self.iter_globals()
+            //     .map(|v| v.global.value().as_f64().unwrap() as u128)
+            //     .collect()
+        }
+
+        /// Set a global, at index idx. Will panic if idx is out of range
+        /// Safety: the caller should check taht the raw value is compatible
+        /// with destination VMGlobal type
+        pub fn set_global_unchecked(&self, idx: usize, new_val: u128) {
```
john-sharratt, line 100: If this is not implemented then won't it break threading and forking?

lib/cli/src/bin/wasmer-jsc.rs
```diff
@@ -0,0 +1,8 @@
+use wasmer_cli::cli::wasmer_main;
+
+#[cfg(not(feature = "jsc"))]
```
theduke, line 3: is this a leftover from testing?

lib/cli/Cargo.toml
```diff
@@ -18,6 +18,12 @@ path = "src/bin/wasmer.rs"
 doc = false
 required-features = ["compiler"]
 
+[[bin]]
+name = "wasmer-jsc"
+path = "src/bin/wasmer-jsc.rs"
+doc = false
```
theduke, line 24: As commented on the file, is this just a leftover from testing?

lib/api/src/externals/table.rs
ptitSeb: I would prefer not to change current feature flags

ptitSeb (approved):
If some of the code could be mutualised with the js backend, that would be great for future maintenance.

PR #3830, opened 2023-04-28
Added `epoll` to WASIX

john-sharratt:
POC for asynchronous IO built on top of asynchronous threading foundation which should give some big performance benefits when used correctly by a WASIX program.

Will look at this more once we have some time so this is just to keep the code here for now


lib/wasi/src/state/env.rs
```diff
@@ -694,6 +718,60 @@ impl WasiEnv {
...
+        let inner = env
+            .try_inner()
+            .ok_or_else(|| WasiError::Exit(Errno::Fault.into()))?;
+
+        let handler = match woken {
+            true => inner.waker_wake.clone(),
+            false => inner.waker_drop.clone(),
+        };
+        if let Some(handler) = handler {
+            let high = id / u32::MAX as WakerId;
```
ptitSeb, line 749: This doesn't look like a proper 64bits -> high/low 32bits extraction (because 0x1_0000
_0000 will give 1 / 1 instead of 1 / 0). Is that on purpose?

lib/wasi/src/state/waker.rs
```diff
@@ -0,0 +1,59 @@
...
+    RawWakerVTable::new(
+        |s| wasi_waker_clone(&*(s as *const WasiWaker)), // clone
+        |s| wasi_waker_wake(&*(s as *const WasiWaker)),  // wake
+        |s| (*(s as *const WasiWaker)).wake_now(),       // wake by ref (don't decrease refcount)
+        |s| drop(Arc::from_raw(s as *const WasiWaker)),  // decrease refcount
+    )
+};
+
+fn wasi_waker_into_waker(s: *const WasiWaker) -> Waker {
+    let raw_waker = RawWaker::new(s as *const (), &VTABLE);
```
ptitSeb, line 52: Mmmm, are you creating an object "manualy" here? That looks a bit hacky.

ptitSeb (approved):
Also, clippy is not happy

lib/wasi/src/state/env.rs
john-sharratt: Dropped the support for 64bits

lib/wasi/src/state/waker.rs
john-sharratt: that's how `cooked_waker` does it, I was trying to avoid adding yet another dependency
https://crates.io/crates/cooked-waker

so i just copied the native implementation instead
https://medium.com/@KevinBGreene/async-programming-in-rust-part-3-futures-and-async-await-b508f7e44abc

it's the "right" way to do it apparently, its just ugly and rather not `rust` like

lib/virtual-fs/src/host_fs.rs
```diff
@@ -16,10 +16,22 @@ use std::task::{Context, Poll};
 use std::time::{SystemTime, UNIX_EPOCH};
 use tokio::fs as tfs;
 use tokio::io::{AsyncRead, AsyncSeek, AsyncWrite, ReadBuf};
+use tokio::runtime::Handle;
 
-#[derive(Debug, Default, Clone)]
```
theduke, line 20: Putting a tokio handle on the FS, and then again on each file, and the trying to get the current env or using a handle seems... like an extremely leaky abstraction and hacky.

Can't we make the callers (eg syscalls) ensure that a runtime is available, and just make these methods async?

lib/virtual-net/src/host.rs
```diff
@@ -13,20 +13,21 @@ use std::ptr;
...
 use tokio::sync::mpsc;
 #[allow(unused_imports, dead_code)]
 use tracing::{debug, error, info, trace, warn};
 
 #[derive(Debug)]
 pub struct LocalNetworking {
-    // Make struct internals private.
-    // Can be removed once some fields are added (like permissions).
-    _private: (),
+    handle: Handle,
```
theduke, line 23: As above for the FS, this seems very hacky.

lib/wasi-types/src/wasi/bindings.rs
```diff
@@ -257,6 +257,8 @@ pub enum Errno {
     Memviolation,
     #[doc = " An unknown error has occured"]
     Unknown,
+    #[doc = " Invocation is pending until the waker is woken."]
+    Pending,
```
theduke, line 261: Could this be represented by `::Again` ?

lib/wasi-types/src/wasi/wasix_manual.rs
```diff
@@ -324,3 +324,150 @@ impl From<ExitCode> for i32 {
...
+        self as i32
+    }
+
+    fn from_native(n: Self::Native) -> Self {
+        match n {
+            0 => Self::Add,
+            1 => Self::Mod,
+            2 => Self::Del,
+
+            q => todo!("could not serialize number {q} to enum EpollCtl"),
```
theduke, line 411: I don't think we want a panic here...

Will just have to run the conversion on a plain number in the syscalls, I think?

lib/wasix/src/fs/fd.rs
```diff
@@ -97,6 +115,15 @@ pub enum Kind {
         /// Reference to the pipe
         pipe: Pipe,
     },
+    Epoll {
+        // List of events we are polling on
```
theduke, line 119: Can we move this out to a separate type, to also keep logic sort of constrained in methods?

lib/wasix/src/http/reqwest.rs
```diff
@@ -1,14 +1,25 @@
...
 use std::convert::TryFrom;
+use tokio::runtime::Handle;
 
 use super::{HttpRequest, HttpResponse};
 
-#[derive(Default, Clone, Debug)]
-pub struct ReqwestHttpClient {}
+#[derive(Clone, Debug)]
+pub struct ReqwestHttpClient {
+    handle: Handle,
```
theduke, line 10: Leakiness... like in the FS and net.

lib/wasix/src/runtime/task_manager/waker.rs
```diff
@@ -0,0 +1,70 @@
+use std::{
+    sync::{mpsc, Arc},
+    task::{Context, Poll, RawWaker, RawWakerVTable, Waker},
+};
+
+use futures::Future;
+
+pub struct InlineWaker {
+    waker_rx: mpsc::Receiver<()>,
+    waker_tx: mpsc::Sender<()>,
```
theduke, line 10: Can we split this into two types?
Primarily: can the actual waker just hold the sender, so that dropping it will also make the receiver fail?

lib/wasix/src/state/builder.rs
```diff
@@ -840,15 +841,18 @@ impl WasiEnvBuilder {
         // The return value is passed synchronously and will block until the result
         // is returned this is because the main thread can go into a deep sleep and
         // exit the dedicated thread
-        let (tx, rx) = std::sync::mpsc::channel();
+        let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
```
theduke, line 844: isn't this just a `sync::oneshot` ?

lib/wasix/src/syscalls/wasi/fd_read.rs
```diff
@@ -217,6 +208,13 @@ fn fd_read_internal<M: MemorySize>(
 
                     drop(guard);
 
+                    let nonblocking = fd_flags.contains(Fdflags::NONBLOCK);
+                    let timeout = socket
+                        .opt_time(TimeType::ReadTimeout)
+                        .ok()
+                        .flatten()
+                        .unwrap_or(Duration::from_secs(30));
```
theduke, line 216: Are these 30 secs just arbitrary?

lib/wasix/src/syscalls/wasix/epoll_ctl.rs
```diff
@@ -0,0 +1,179 @@
...
+    let fd_entry = wasi_try_ok!(env.state.fs.get_fd(epfd));
+
+    let inode = fd_entry.inode.clone();
+    let tasks = env.tasks().clone();
+    let mut inode_guard = inode.read();
+    match inode_guard.deref() {
+        Kind::Epoll {
+            subscriptions, tx, ..
+        } => {
+            if let EpollCtl::Del | EpollCtl::Mod = op {
```
theduke, line 46: As mentioned before, can we collect all the related business logic in a `EpollObject` or similar struct?

lib/wasix/src/syscalls/wasix/sock_recv.rs
```diff
@@ -113,8 +130,26 @@ fn sock_recv_internal<M: MemorySize>(
...
 
+                let nonblocking = waker.is_none() && fd.flags.contains(Fdflags::NONBLOCK);
+                let timeout = if waker.is_none() {
+                    Some(
+                        socket
+                            .opt_time(TimeType::ReadTimeout)
+                            .ok()
+                            .flatten()
+                            .unwrap_or(Duration::from_secs(30)),
+                    )
```
theduke, line 141: again, magic timeout.... that seems quite suboptimal?

theduke (changes requested):
I didn't dive into the details of the epoll code or the syscall changes yet, but here are some preliminary comments.

lib/virtual-fs/src/host_fs.rs
john-sharratt: The tokio handle is extremely brittle, this is a better way of doing it - keeps the tokio abstraction largely away from the WASM threads. Syscalls should not be aware of tokio at all... in the browser we won't be using the tokio handles anyway so it should not be in the syscalls

lib/virtual-net/src/host.rs
john-sharratt: As above

lib/wasi-types/src/wasi/bindings.rs
john-sharratt: It's leftovers, ill remove it

lib/wasi-types/src/wasi/wasix_manual.rs
john-sharratt: This is not my code, its how the template for it was coded - there are 29 instances of this in the code already on native WASIX deserialization

lib/wasix/src/http/reqwest.rs
john-sharratt: As above, this is better

lib/wasix/src/syscalls/wasi/fd_read.rs
john-sharratt: yes, an arbitrary default

lib/wasix/src/syscalls/wasix/epoll_ctl.rs
john-sharratt: that is better done in the fs refactor, I'll just keeping it within the existing technical debt

lib/wasix/src/syscalls/wasix/sock_recv.rs
john-sharratt: its a default, not having a default means sockets read forever, which is not good as a default

john-sharratt:
Superseded by https://github.com/wasmerio/wasmer/pull/4050

PR #3831, opened 2023-04-28
Introduce a package resolver

### LINKED ISSUE #3603: WASIX: Refactor Package Fetching and Module Caching
The `wasix` implementation currently supports fetching WebC packages and compiling them. 
This functionality is exposed to guest code via `proc_spawn` , and via a virtual Wasmer command.

This currently is too deeply integrated into the code base, and should be refactored.

Fetching modules should be abstracted behind a `ModuleResolver` trait, which can be supplied/configured via the `WasiRuntime` trait. 

There already is a branch with a partial solution, but the changes were not included in the `wasix` branch to speed up the merge.

Michael-F-Bryan:
# Description

This adds a `PackageResolver` getter to `WasiRuntime` and wires it up to the `ModuleCache`.

TL;DR:

- The `WasiRuntime` trait now has a `package_resolver()` method which gives you an `Arc<dyn PackageResolver>`
- A `PackageResolver` will resolve package identifiers to a `BinaryPackage` (expect the return type to change - eventually I want to return a `wasmer_wasix::runtime::resolver::ResolvedPackage`)
- The `PluggableRuntime` will use the "builtin" resolver from `lib/wasi/src/runtime/resolver/builtin.rs` by default, this just calls into the existing `wasmer_wasix::wapm::fetch_webc()` code which uses the WAPM backend and will cache `*.webc` files to disk
  - Both the cache directory and URL for the WAPM backend are configurable
  - there is a `BuiltinResolver::from_env()` constructor that will eventually set things up using `$WASMER_HOME/wasmer.toml` (I've just added a quick'n'dirty implementation for now)
- The `ModuleCache::get_webc()` module now just defers to `runtime.package_resolver().resolve_package(...)` because the PackageResolver implementation should be in charge of caching
- Now that `ModuleCache` doesn't keep the in-memory cache of resolved `BinaryPackage`s I've added a `with_cache()`  method that returns `InMemoryCache<Self>` so you can wrap an existing `PackageResolver` with an in-memory caching layer 
- Removed the `--include-webc` flag from `wasmer run` because it used to explicitly add items to the `ModuleCache`'s WEBC cache and all that code has been deleted (not sure if this should have been a hard delete or if we should warn people when they use it that `--include-webc` is now a no-op)
- Removed the `engine()` getter from the `WasiRuntime` trait so we can introduce a separate mechanism for module compilation and caching (nothing used this anyway, they just called `WasiRuntime::new_store()`)

CC: #3603.


.github/workflows/test.yaml
```diff
@@ -553,6 +554,7 @@ jobs:
       - name: Set up base deps on musl
         if: matrix.metadata.build == 'linux-musl'
         run: |
+            apk update
```
syrusakbary, line 557: Why this? It will usually slow down ci

lib/wasi/src/runtime/resolver/cache.rs
```diff
@@ -0,0 +1,228 @@
...
+    runtime::resolver::{PackageResolver, ResolverError, WebcIdentifier},
+};
+
+/// A resolver that wraps a [`PackageResolver`] in an in-memory LRU cache.
+#[derive(Debug)]
+pub struct InMemoryCache<R> {
+    resolver: R,
+    packages: Mutex<Vec<CacheEntry>>,
+    config: CacheConfig,
+}
```
syrusakbary, line 15: Why it needs to be LRU? If so, why do we need to implement us the LRU instead of reusing other package?

lib/cli/src/commands/run/wasi.rs
```diff
@@ -49,11 +49,6 @@ pub struct Wasi {
     #[clap(long = "use", name = "USE")]
     uses: Vec<String>,
 
-    /// List of webc packages that are explicitely included for execution
-    /// Note: these packages will be used instead of those in the registry
-    #[clap(long = "include-webc", name = "WEBC")]
-    include_webcs: Vec<String>,
```
john-sharratt, line 55: The removal of this parameter needs to be replaced with another way of making the integration tests deterministic or it will prevent them from properly regression testing our changes and will require the availability of registries in order for the tests to run successfully.

One way is to create a dummy local registry that the unit tests can point to which is probably the better way to do it, either way we shouldn't remove this code until the replacement is in place.

john-sharratt (changes requested):

.github/workflows/test.yaml
Michael-F-Bryan: Yep, but at the moment CI is erroring out on musl builds saying you need to run `apk update`. See [my comment](https://wasmerio.slack.com/archives/CDNJPS7SA/p1682942675958989) on Slack.

lib/cli/src/commands/run/wasi.rs
Michael-F-Bryan: Good point. I'm guessing I'll need to wire that into the `BuiltinResolver` somehow and then make sure `wasmer run` configures it's own `BuiltinResolver` and passes to to the `PluggableRuntime`.

lib/wasi/src/runtime/resolver/cache.rs
Michael-F-Bryan: The idea was so we can start adding cache invalidation to Deploy because at the moment, all modules and `BinaryPackage`s that end up in an in-memory cache will stay there until the program exits.

In retrospect, it's probably an overcomplication at the moment. I think I'll rip the LRU stuff out and use a `RwLock<HashMap>` that never gets cleaned up, just like what `ModuleCache` did before it.

lib/wasi/src/runtime/resolver/types.rs
```diff
@@ -0,0 +1,197 @@
...
+use crate::{
+    bin_factory::BinaryPackage,
+    http::HttpClient,
+    runtime::resolver::{cache::CacheConfig, InMemoryCache},
+};
+
+#[async_trait::async_trait]
+pub trait PackageResolver: std::fmt::Debug + Send + Sync {
+    /// Resolve a package, loading all dependencies.
+    async fn resolve_package(
```
theduke, line 12: I think this is too limited.

There are different use cases that need access to different things:

A) The manifest for a package. This could be needed during dependency resolution, or during other resolution in different contexts. Manifests could be cached efficiently as well, (like cargo does with the new HTTP index access method) , or directly loaded from local webc files. 

B) The raw webc 

C) A loaded `WebC` for directly accessing the manifest, the volumes, etc

D) A materialized binary package

lib/wasi/src/runtime/resolver/types.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub struct WebcIdentifier {
+    /// The package's full name (i.e. `wasmer/wapm2pirita`).
+    pub full_name: String,
+    pub locator: Locator,
+    /// A semver-compliant version constraint.
+    pub version: String,
```
theduke, line 61: Why isn't' the version an `Option<_>` ? Would be more expressive.

lib/wasi/src/runtime/resolver/types.rs
```diff
@@ -0,0 +1,197 @@
...
+    runtime::resolver::{cache::CacheConfig, InMemoryCache},
+};
+
+#[async_trait::async_trait]
+pub trait PackageResolver: std::fmt::Debug + Send + Sync {
+    /// Resolve a package, loading all dependencies.
+    async fn resolve_package(
+        &self,
+        pkg: &WebcIdentifier,
+        client: &(dyn HttpClient + Send + Sync),
```
theduke, line 15: A resolver might not need an http client at all. Or might need to use a specific http client.

This shouldn't be part of the API, but passed in during specific resolver construction.

lib/wasi/src/runtime/resolver/builtin.rs
```diff
@@ -0,0 +1,66 @@
...
+    runtime::resolver::{types::ResolverError, types::WebcIdentifier, PackageResolver},
+};
+
+/// The builtin package resolver, backed by WAPM.
+///
+/// Any downloaded assets will be cached on disk.
+#[derive(Debug, Clone)]
+pub struct BuiltinResolver {
+    cache_dir: PathBuf,
+    registry_endpoint: Url,
```
theduke, line 18: Note that supporting multiple registries could be a common use case, eg accessing both dev and prod registry packages. Also need to mix in some local package resolution on top. (for local dev)

lib/wasi/src/runtime/resolver/builtin.rs
```diff
@@ -0,0 +1,66 @@
...
+    bin_factory::BinaryPackage,
+    http::HttpClient,
+    runtime::resolver::{types::ResolverError, types::WebcIdentifier, PackageResolver},
+};
+
+/// The builtin package resolver, backed by WAPM.
+///
+/// Any downloaded assets will be cached on disk.
+#[derive(Debug, Clone)]
+pub struct BuiltinResolver {
```
theduke, line 16: `Builtin` is a suboptimal name, it doesn't really express what it does. 

Maybe `RegistryResolver` ?

lib/wasi/src/runtime/resolver/types.rs
theduke: Note that `WasiRuntime::http_client()` should also be changed to return an `Arc<>`,  or to have a `build_http_client()` instead that constructs a new, independent client.

Michael-F-Bryan: Yeah, I agree.

Depending on how we want to tackle this, we could a) pass in some sort of factory that'll give you a HTTP client (or TCP connection or whatever) that we got from `WasiRuntime::http_client()`, or we could b) require `PackageResolver` implementations to bring their own HTTP client.

I'm partial to the first approach because it makes the code more testable and lets the runtime manage all HTTP client instances.

lib/wasi/src/runtime/mod.rs
```diff
@@ -26,24 +32,11 @@ where
...
-    fn engine(&self) -> Option<wasmer::Engine> {
-        None
-    }
+    fn package_resolver(&self) -> Arc<dyn PackageResolver + Send + Sync>;
 
     /// Create a new [`wasmer::Store`].
     fn new_store(&self) -> wasmer::Store {
-        cfg_if::cfg_if! {
-            if #[cfg(feature = "sys")] {
-                if let Some(engine) = self.engine() {
```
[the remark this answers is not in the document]
Michael-F-Bryan: Sorry, I saw the `engine()` method was unused throughout the repo except for our default `new_store()` implementation and one line in the `ModuleCache` that I have since removed, so I thought it was dead code.

I've put it back in.

lib/wasi/src/runtime/resolver/types.rs
Michael-F-Bryan: I'm going to leave this for the next PR where we start to flesh out the `PackageResolver` trait and clean up the code that uses it.

john-sharratt (approved):

lib/wasi/src/runtime/resolver/types.rs
Michael-F-Bryan: Yeah, in the next PR I'll start fleshing the `PackageResolver` trait out a bit more.

lib/wasi/src/runtime/resolver/builtin.rs
Michael-F-Bryan: True. I'm going to leave those more complex use cases for now, though.

lib/wasi/src/runtime/resolver/builtin.rs
Michael-F-Bryan: Renamed.

theduke (approved):

syrusakbary:
(let’s not merge until we release Wasmer 3.3)

PR #3832, opened 2023-04-29
WASIX: Support Package Download Auth Tokens in ModuleCache

theduke:
- feat(wasix): Support package download auth through a token
- feat(wasix): Allow specifying a registry auth token for the ModuleCache


theduke:
This PR was made redundant by larger changes to package loading.

PR #3833, opened 2023-04-30
Fix Missing WaiterError export + Add notify/wait to fd_mmap memory

theduke:
- Make wasmer_vm::WaiterError non_exhaustive.
- vm: Make WaiterError and ThreadConditions public
- Implement notify/wait support for fd_mmap memory

Note: WaiterError is part of the public API in the LinearMemory trait, so not making it public was an oversight.


ptitSeb (approved):
Non exhaustive error, good call!

PR #3841, opened 2023-05-03
Introduce a module cache abstraction

Michael-F-Bryan:
This PR adds an abstraction for caching modules to the `WasiRuntime`, plus some implementations that can be composed to implement tiered module caching.

TL;DR:
- Introduce a `CompiledModuleCache` trait that has async `save()` and `load()` methods 
  - Both methods accept a `&dyn VirtualTaskManager` at the moment in case we need to spawn blocking operations, but that'll probably be moved around
- Add a `and_then()` combinator to `CompiledModuleCache` so you can layer your caching
- I've taken the three different layers of caching in `ModuleCache` and split them into their own types - `ThreadLocalCache`, `SharedCache`, and `OnDiskCache`
- There is a `wasmer_wasix::runtime::module_cache::default_cache()` constructor which chains the thread-local, `Arc<RwLock<HashMap>>`, and on-disk caches together
- The `ModuleCache` is now just a wrapper around an instance of that default cache

CC: #3603 and #3831.


lib/wasi/src/runtime/module_cache/types.rs
```diff
@@ -0,0 +1,78 @@
...
+/// same key to load or save will always result in the "same" module.
+///
+/// Implementations can also assume that [`CompiledModuleCache::load()`] will
+/// be called more often than [`CompiledModuleCache::save()`] and optimise
+/// their caching strategy accordingly.
+#[async_trait::async_trait]
+pub trait CompiledModuleCache: Debug + Send + Sync {
+    async fn load(&self, key: &str, engine: &Engine) -> Result<Module, CacheError>;
+
+    async fn save(&self, key: &str, module: &Module) -> Result<(), CacheError>;
```
syrusakbary, line 21: I just realized that we can make `load` and `save` perfectly sync. As I have not seen any await outside of calling load/save.

Which basically means that we can use wasmer-cache directly?

lib/wasi/src/runtime/module_cache/shared.rs
```diff
@@ -0,0 +1,42 @@
...
+use wasmer::{Engine, Module};
+
+use crate::runtime::module_cache::{CacheError, CompiledModuleCache};
+
+/// A [`CompiledModuleCache`] based on a
+/// <code>[Arc]<[RwLock]<[HashMap]<[String], [Module]>>></code> that can be
+/// shared.
+#[derive(Debug, Default, Clone)]
+pub struct SharedCache {
+    modules: Arc<RwLock<HashMap<String, Module>>>,
```
syrusakbary, line 13: We can use `DashMap` instead, it will be more performant
https://crates.io/crates/dashmap/2.1.0

lib/wasi/src/runtime/module_cache/shared.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    pub fn from_existing_modules(modules: Arc<RwLock<HashMap<String, Module>>>) -> Self {
+        SharedCache { modules }
+    }
+}
+
+#[async_trait::async_trait]
+impl CompiledModuleCache for SharedCache {
+    async fn load(&self, key: &str, _engine: &Engine) -> Result<Module, CacheError> {
+        let modules = self.modules.read().await;
```
syrusakbary, line 29: I don't think async here will be required if we use dashmap

lib/wasi/src/runtime/module_cache/shared.rs
```diff
@@ -0,0 +1,42 @@
...
+
+        modules.get(key).cloned().ok_or(CacheError::NotFound)
+    }
+
+    async fn save(&self, key: &str, module: &Module) -> Result<(), CacheError> {
+        let module = module.clone();
+        let key = key.to_string();
+
+        self.modules.write().await.insert(key, module);
```
syrusakbary, line 38: *and here

lib/wasi/src/runtime/module_cache/shared.rs
Michael-F-Bryan: Ooh, good thinking!

lib/wasi/src/runtime/module_cache/types.rs
theduke: No, there will very much be async use cases.

In a tokio context you will want to use `tokio::fs`, otherwise you need to spawn a blocking task.

In case of a shared/distributed cache we also need async.

lib/wasi/src/runtime/module_cache/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+
+pub(crate) use self::disabled::Disabled;
+
+/// Get a [`ModuleCache`] which should be good enough for most use
+/// cases.
+///
+/// The returned object will use a mix of in-memory and on-disk caching
+/// strategies.
+pub fn default_cache(cache_dir: &std::path::Path) -> impl ModuleCache + Send + Sync {
+    ThreadLocalCache::default()
```
theduke, line 24: `ThreadLocal` should not be a default, it is only required for the `web` target. (aka `js`, currently)

lib/wasi/src/runtime/module_cache/types.rs
```diff
@@ -0,0 +1,78 @@
...
+///
+/// Implementations can assume that cache keys are unique and that using the
+/// same key to load or save will always result in the "same" module.
+///
+/// Implementations can also assume that [`CompiledModuleCache::load()`] will
+/// be called more often than [`CompiledModuleCache::save()`] and optimise
+/// their caching strategy accordingly.
+#[async_trait::async_trait]
+pub trait ModuleCache: Debug {
+    async fn load(&self, key: &str, engine: &Engine) -> Result<Module, CacheError>;
```
theduke, line 19: As mentioned, I don't think this should be a generic string, but the sha256 hash of the wasm code.

An implementation should decide how to map a hash into a key, based on the engine.

lib/wasi/src/runtime/module_cache/shared.rs
Michael-F-Bryan: I've switched over to dashmap, but I'm planning to keep `ModuleCache` as an `async` trait.

lib/wasi/src/runtime/module_cache/types.rs
```diff
@@ -0,0 +1,141 @@
...
+///
+/// Implementations can also assume that [`ModuleCache::load()`] will
+/// be called more often than [`ModuleCache::save()`] and optimise
+/// their caching strategy accordingly.
+#[async_trait::async_trait]
+pub trait ModuleCache: Debug {
+    async fn load(&self, key: Key, engine: &Engine) -> Result<Module, CacheError>;
+
+    async fn save(&self, key: Key, module: &Module) -> Result<(), CacheError>;
+
```
theduke, line 26: Should probably have a `remove(hash)` and a `purge()/clear()`.

theduke (changes requested):

lib/wasi/src/runtime/module_cache/types.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// The item was not found.
+    #[error("Not found")]
+    NotFound,
+    #[error(transparent)]
+    Other(Box<dyn std::error::Error + Send + Sync>),
+}
+
+/// A 256-bit key used for caching.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct Key([u8; 32]);
```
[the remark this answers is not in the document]
syrusakbary: We already have a type in wasmer-cache.

The closer it is, the better: https://github.com/wasmerio/wasmer/blob/master/lib/cache/src/hash.rs

lib/wasi/src/runtime/mod.rs
```diff
@@ -34,6 +38,13 @@ where
 
     fn package_resolver(&self) -> Arc<dyn PackageResolver + Send + Sync>;
 
+    /// A cache for compiled modules.
+    ///
+    /// Caching is disabled by default.
+    fn module_cache(&self) -> Arc<dyn ModuleCache + Send + Sync> {
+        Arc::new(module_cache::Disabled)
```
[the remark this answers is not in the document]
Michael-F-Bryan: There's no guarantee that the returned module cache will be saved by the caller and reused everywhere, so returning a new instance of the default cache every time means the cached modules won't actually be shared or persisted.

The best way to avoid footguns is to not provide a default implementation and force the implementor to handle things themselves.

theduke: That's a valid point, but things like spawn_exec, dependency resolution (uses, etc) , runners will populate the cache.

The async threading PR also introduces some more common abstractions for running.

So many code paths would probably populate the cache without the user knowing about it.

lib/wasi/src/runtime/module_cache/types.rs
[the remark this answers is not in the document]
theduke: I guess all the things added here will move to the `-cache` crate anyway.

lib/wasi/src/runtime/module_cache/types.rs
Michael-F-Bryan: I'm tempted to say it's not the `ModuleCache` trait's responsibility to remove items from a cache or purge them because those extra methods won't be used by WASIX. 

There's nothing stopping us from creating a `DeployModuleCache` and adding `remove()`, `purge()`, and `clear()` as normal methods, though.

lib/wasi/src/runtime/module_cache/on_disk.rs
```diff
@@ -0,0 +1,167 @@
...
+                    dir=%parent.display(),
+                    error=&e as &dyn std::error::Error,
+                    "Unable to create the cache dir",
+                );
+            }
+        }
+
+        // PERF: We can reduce disk usage by using the weezl crate to
+        // LZW-encode the serialized module.
+        module
```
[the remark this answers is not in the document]
Michael-F-Bryan: Fixed.

lib/wasi/src/runtime/module_cache/on_disk.rs
```diff
@@ -0,0 +1,167 @@
+use std::path::{Path, PathBuf};
+
+use wasmer::{Engine, Module};
+
+use crate::runtime::module_cache::{CacheError, Key, ModuleCache};
+
+/// A cache that saves modules to a folder on disk using
+/// [`Module::serialize()`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OnDiskCache {
```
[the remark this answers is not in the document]
Michael-F-Bryan: I've renamed it to `FileSystemCache`.

lib/wasi/src/runtime/module_cache/filesystem.rs
```diff
@@ -0,0 +1,231 @@
...
+
+        if let Some(parent) = path.parent() {
+            if let Err(e) = std::fs::create_dir_all(parent) {
+                tracing::warn!(
+                    dir=%parent.display(),
+                    error=&e as &dyn std::error::Error,
+                    "Unable to create the cache directory",
+                );
+            }
+        }
```
syrusakbary, line 89: I think it would be good to do the "setup" only when creating the `FilesystemCache` (that is, calling `::new(cache_dir)`.

Similarly as we currently do in the FilesystemCache
https://github.com/wasmerio/wasmer/blob/master/lib/cache/src/filesystem.rs#L43-L78

lib/wasi/src/runtime/module_cache/types.rs
```diff
@@ -0,0 +1,164 @@
...
+    Deserialize(#[from] wasmer::DeserializeError),
+    #[error("Unable to read from \"{}\"", path.display())]
+    Read {
+        path: PathBuf,
+        #[source]
+        error: std::io::Error,
+    },
+    #[error("Unable to write to \"{}\"", path.display())]
+    Write {
+        path: PathBuf,
```
syrusakbary, line 79: The cache might not be able to write, and there might be no path (for example, if you are using a cache saved in redis via a key).
We should remove the path field here as is filesystem dependant

lib/wasi/src/runtime/module_cache/types.rs
```diff
@@ -0,0 +1,164 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum CacheError {
+    #[error("Unable to serialize the module")]
+    Serialize(#[from] wasmer::SerializeError),
+    #[error("Unable to deserialize the module")]
+    Deserialize(#[from] wasmer::DeserializeError),
+    #[error("Unable to read from \"{}\"", path.display())]
+    Read {
+        path: PathBuf,
```
syrusakbary, line 73: The cache might not be able to read, and there might be no path (for example, if you are using a cache saved in redis via a key).
We should remove the path field here as is filesystem dependant

lib/wasi/src/runtime/module_cache/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+/// # Platform-specific Notes
+///
+/// This will use the [`ThreadLocalCache`] when running in the browser.  Each
+/// thread lives in a separate worker, so sharing compiled modules in the
+/// browser requires using a custom [`ModuleCache`] built on top of
+/// [`postMessage()`][pm] and [`SharedArrayBuffer`][sab].
+///
+/// [pm]: https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage
+/// [sab]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer
+pub fn in_memory() -> impl ModuleCache + Send + Sync {
```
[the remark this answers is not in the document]
Michael-F-Bryan: I don't think we should do this. 

It would require either adding a `cache_dir: PathBuf` argument that is completely useless when running on the web, or we would need to duplicate `wasmer-registry`'s logic for determining the cache directory.

Whatever mechanism people use to discover the `in_memory()` method (auto-complete, API docs, etc.) will also show them there is a `FileSystemCache`, so I don't think discoverability will be a big deal.

lib/wasi/src/runtime/module_cache/types.rs
syrusakbary: I kind of like the idea of having `delete(key, &engine)` and `purge(&engine)`

lib/wasi/src/runtime/module_cache/types.rs
```diff
@@ -0,0 +1,164 @@
...
+///
+/// Implementations can assume that cache keys are unique and that using the
+/// same key to load or save will always result in the "same" module.
+///
+/// Implementations can also assume that [`ModuleCache::load()`] will
+/// be called more often than [`ModuleCache::save()`] and optimise
+/// their caching strategy accordingly.
+#[async_trait::async_trait]
+pub trait ModuleCache: Debug {
+    async fn load(&self, key: Key, engine: &Engine) -> Result<Module, CacheError>;
```
syrusakbary, line 24: Thinking more about this, probably this should return a `LoadCacheError` (enum with states: NotFound, ReadError, Generic), as is more specialized and `WriteError` and `SerializeError` are states that can't be reached here

lib/wasi/src/runtime/module_cache/types.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    async fn save(&self, key: Key, module: &Module) -> Result<(), CacheError>;
```
syrusakbary, line 26: Thinking more about this, probably this should return a `SaveCacheError` (enum with states: WriteError(wasmer::SerializeError), Generic(std::error::Error))

lib/wasi/src/runtime/module_cache/types.rs
Michael-F-Bryan: Users should use different variants if that is the case. The `CacheError::Other` variant is what the redis example would use.

This variant is primarily here so get a useful error message like `Error: Unable to write to "path/to/cache", Caused by: Is a directory` where you can associate the failing filesystem operation with a path.

lib/wasi/src/runtime/module_cache/disabled.rs
```diff
@@ -0,0 +1,18 @@
+use wasmer::{Engine, Module};
+
+use crate::runtime::module_cache::{CacheError, Key, ModuleCache};
+
+/// A cache that always fails.
+#[derive(Debug, Copy, Clone, PartialEq, Default)]
+pub(crate) struct Disabled;
```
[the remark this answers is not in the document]
Michael-F-Bryan: I've made `module_cache()` a required method of `WasiRuntime` to avoid footguns, so the `Disabled` cache implementation is no longer needed.

lib/wasi/src/runtime/mod.rs
Michael-F-Bryan: I've made this a required method. The `PluggableRuntime` uses the `in_memory()` cache by default, so people will still benefit from sharing their module cache across the runtime.

lib/wasi/src/runtime/module_cache/filesystem.rs
theduke: I think it's good to always ensure the parent dir exists. 

Otherwise you could delete the cache dir and deploy stops workings. 
(I do this regularly when working on deploy)

The overhead should be small enough to not matter.

lib/wasi/src/runtime/module_cache/types.rs
Michael-F-Bryan: I agree it could be useful, but I don't want to start adding extra complexity to the `ModuleCache` trait until we see a need for it. 

Also keep in mind that the `ModuleCache` trait doesn't need to include everything any cache implementation could ever need. It's perfectly fine for cache implementations to have their own inherent methods which do operations that make sense for that cache. It's only when you need to do things in a generic context that they need to be added to the trait.

lib/wasi/src/runtime/module_cache/types.rs
Michael-F-Bryan: See my response from your comment on the `Write` variant. TL;DR: These variants are for specific IO errors that may happen when you do filesystem operations, rather than a general *"some sort of read operation failed"* error.

lib/wasi/src/runtime/module_cache/types.rs
```diff
@@ -0,0 +1,164 @@
...
+    Write {
+        path: PathBuf,
+        #[source]
+        error: std::io::Error,
+    },
+    /// The item was not found.
+    #[error("Not found")]
+    NotFound,
+    #[error(transparent)]
+    Other(Box<dyn std::error::Error + Send + Sync>),
```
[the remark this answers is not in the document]
Michael-F-Bryan: The name, `Generic`, is a bit uhh... generic... When trait implementations need an escape hatch to supply their own error, it'll be called `Unknown` or `Other`.

lib/wasi/src/runtime/module_cache/types.rs
Michael-F-Bryan: Let's leave it as a single error for now. There's a good chance the `ModuleCache` trait's shape will change as we start getting feedback from production, and if we add `purge()` or `remove()` methods I don't really want the `wasmer_wasix::runtime::module_cache` to keep growing extra error types.

lib/wasi/src/runtime/module_cache/mod.rs
Michael-F-Bryan: This has been fixed, although I've chosen not to add the filesystem layer because that's a bit too automagical for my liking.

lib/wasi/src/runtime/module_cache/types.rs
Michael-F-Bryan: I'd prefer to leave this as a single error for now. Let's wait until the `ModuleCache` trait gets some real-world use before doing more tweaks.

lib/wasi/src/runtime/module_cache/types.rs
```diff
@@ -93,32 +104,26 @@ impl CacheError {
...
-    /// Generate a new [`Key`] based on the SHA-256 hash of some bytes.
-    pub fn sha256(data: impl AsRef<[u8]>) -> Self {
-        let mut hasher = Sha256::default();
-        hasher.update(data);
-        Key::new(hasher.finalize().into())
-    }
+    /// Generate a new [`ModuleCache`] based on the SHA-256 hash of some bytes.
+    pub fn sha256(wasm: impl AsRef<[u8]>) -> Self {
+        let wasm = wasm.as_ref();
+        debug_assert!(
```
theduke, line 119: Initial I wanted to say this could actually be a `Result<(), XXXError>`. 

Only complication is if we wanted to also allow caching of `.wat`, where we should cache the original input rather than a transformed `.wasm`.

But that could also be a separate method. 

===

In any case, I don't think `debug_assert!` is the way to go here: either it's a required invariant, in which case it should be an error, or we just remove it.

lib/wasi/src/runtime/module_cache/filesystem.rs
```diff
@@ -0,0 +1,237 @@
...
+
+        // Note: We'll first save to a temporary file in the same folder, then
+        // rename it when we are done serializing. Try to use a unique extension
+        // just in case we have concurrent saves of the same module.
+        static UNIQUE_ID: AtomicU64 = AtomicU64::new(0);
+        let extension = format!("tmp{}", UNIQUE_ID.fetch_add(1, Ordering::Relaxed));
```
theduke, line 101: Multiple Wasmer processes could trample over each other with this. 

I'd use a `Uuid::new_v4()` or a nanosecond timestamp.

lib/wasi/src/runtime/mod.rs
```diff
@@ -181,4 +190,8 @@ impl WasiRuntime for PluggableRuntime {
     fn tty(&self) -> Option<&(dyn TtyBridge + Send + Sync)> {
         self.tty.as_deref()
     }
```
theduke, line 192: Need a `set_module_cache()` here.

lib/cli/src/commands/run_unstable.rs
```diff
@@ -454,17 +454,24 @@ impl TargetOnDisk {
...
+        }
+
+        // If we can't figure out the file type based on its content, fall back
+        // to checking the extension.
+
+        match path.extension().and_then(|s| s.to_str()) {
+            Some("wat") => Ok(TargetOnDisk::Wat(path)),
+            _ => anyhow::bail!("Unable to determine how to execute \"{}\"", path.display()),
         }
     }
```
theduke, line 477: Run `run` and `run_unstable` still need to set up the file system cache.

Is that implemented?

theduke (changes requested):

lib/wasi/src/runtime/module_cache/types.rs
theduke: The question I ask myself with error types is: what invariants need to be handled with custom code paths ? 

Sometimes hard to tell before you use it. 

I think this is fine for now. 
(the main one I can think of is `deserialize`)

lib/wasi/src/runtime/module_cache/types.rs
Michael-F-Bryan: Removed.

lib/wasi/src/runtime/module_cache/filesystem.rs
Michael-F-Bryan: I've switched to the `tempfile` crate. That should let us write to and persist temporary files properly without worrying about concurrent processes trampling each other.

lib/wasi/src/runtime/mod.rs
Michael-F-Bryan: Done.

lib/cli/src/commands/run_unstable.rs
Michael-F-Bryan: I just wired it up.

PR #3843, opened 2023-05-03
Adapted the publishing script

ptitSeb:
Adaptation needed to the publish script.

### DIFF
--- a/scripts/publish.py
+++ b/scripts/publish.py
@@ -48,7 +48,7 @@ SETTINGS = {
         "wasmer-cli": "default,cranelift",
         "wasmer-wasix": "sys,wasmer/sys",
         "wasmer-wasix-types": "wasmer/sys",
-        "wasmer-wasix-experimental-io-devices": "wasmer-wasix/sys",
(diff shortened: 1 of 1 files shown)


scripts/publish.py
```diff
@@ -48,7 +48,7 @@
         "wasmer-cli": "default,cranelift",
         "wasmer-wasix": "sys,wasmer/sys",
         "wasmer-wasix-types": "wasmer/sys",
-        "wasmer-wasix-experimental-io-devices": "wasmer-wasix/sys",
+        "wasmer-wasix-experimental-io-devices": "wasmer-wasix/sys,wasmer/sys",
```
theduke, line 51: This is the only one with multiple features, does the code properly handle splitting? 
Or is this just passed to `-F`  or similar so it works fine?

Michael-F-Bryan (approved):

scripts/publish.py
ptitSeb: It's not the only one. `wasmer-cli` and `wasmer-wasix` also have 2 features. the while features test is used as-is with no issues.

PR #3853, opened 2023-05-10
Fix cargo update

ptitSeb:
Some crates should not be updated yet, and fixed linting issue with updated emuset.


theduke:
What's the failing `unused_uniti` lint?

ptitSeb:
It's the EnumSet derive.
image

theduke (approved):

PR #3854, opened 2023-05-11
deps: Upgrade clap to v4

theduke:
Upgrade clap from 3.x to 4.2.7 in both the main CLI and the compiler
CLI.

Needed as a prep work for future CLI additions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -56,10 +56,53 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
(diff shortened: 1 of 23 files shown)


theduke:
Damn, `clap` has a new dependency `anstream`, which doesn't compile to `wasi`.

ptitSeb (approved):
So, the `parse(from_os_str)` attribute is no longer neeed?

PR #3871, opened 2023-05-17
Added the ability to force preloaded packages to be latest

john-sharratt:
Needed to bring so that preloaded packages will always be used instead of looking for the latest version - used in certain use cases


theduke:
This can be handled by arranging the `Source`s in the new package resolution abstractions in #3852 .

PR #3877, opened 2023-05-18
wasix: Remove Legacy thread-local System

### LINKED ISSUE #3699: Implement WebAssembly local-threads support for WASIX
# Goal

Currently `wasix` uses syscalls for getting and setting local thread variables which adds quite some overhead - instead the TLS area should be used for this thus avoiding this overhead and removing some redundant syscalls.

# Impacted syscalls

The syscalls impacted are as follows:

```
        "thread_local_create" => Function::new_typed_with_env(&mut store, env, thread_local_create::<Memory32>),
        "thread_local_destroy" => Function::new_typed_with_env(&mut store, env, thread_local_destroy),
        "thread_local_set" => Function::new_typed_with_env(&mut store, env, thread_local_set),
        "thread_local_get" => Function::new_typed_with_env(&mut store, env, thread_local_get::<Memory32>),
```

# Rust Standard Library

The rust standard library must be modified in order to use TLS instead of the `wasix` syscalls - the locations of the thread local implementation is as follows:

https://github.com/wasmerio/rust/blob/wasix/library/std/src/thread/local.rs
https://github.com/wasmerio/rust/blob/wasix/library/std/src/sys/wasix/atomics/thread_local_key.rs

# Avoiding memory leaks

Particular attention must be made that when threads exit they do not cause memory leaks, at present this is done by creating a callback that the runtime executes after the thread exits `_thread_local_destroy`. This method will free any memory used by the local thread variables including invoking any destructors (for unsafe allocations). One could use the `pthreads` libc implementation to achieve the same thing rather than having this as a runtime responsibility.

# wasix-libc

Threading is currently implemented using `pthreads` in the libc fork here:
https://github.com/wasmerio/wasix-libc

One approach to solving this problem would be to implement rust local thread storage entirely using the pthreads implementation however this is not the only way approach and a fully rust native solution is also possible.

# Cargo build

WASIX applications can be compiled using this helpful toolchain fork:
https://github.com/wasmerio/cargo-wasix

# Testing

There is a fully working http server that uses local thread storage extensively and thus can be used for functional and non-functional tests - it is located here:
https://github.com/john-sharratt/static-web-server

theduke:
Remove deprecated thread_local functionality

thread_local support in wasix was previously implemented via custom
exports provided to modules.

This has been refactored to a more efficient (guest-internal)
implementation in both the wasix libc and the Rust fork.

We kept the exports and related functionality in for a while to allow for
modules to update, but now all the related code can safely be removed.

Closes #3699


theduke:
@dynamite-bud this is also relevant for you - you can remove the documentation for all the `thread-local-*` exports.

john-sharratt:
Unfortunately I had to remove one of the tests as the original source code was never kept - its `longjump2` but fortunately its not a realistic test anyway, it was doing longjumps inside rust by manually invoking the syscalls. anyway - all the rest of the tests are now fixed

john-sharratt (approved):

PR #3878, opened 2023-05-18
lib/wasi-types: add Socktype::Unused shim value

waynr:
# Description

In wasix-libc we derive the value of the C macro `SOCK_DGRAM` and `SOCK_STREAM` from the `sock_type` enum in [wasix-witx](https://github.com/john-sharratt/wasix-witx). This `sock_type` witx enum corresponds roughly to the Wasmer [`socktype` enum](https://github.com/wasmerio/wasmer/blob/fbe784a6893b62ca0f961153e9dfdbbbc2374d24/lib/wasi-types/schema/wasi/typenames.wit#L697-L702).

The problem with these corresponding enums as-written is that they result in a `SOCK_DGRAM` value of 0. This doesn't work for cpython (and potentially other programs depending on libc) because the value `0` is used for a cpython-specific macro that is used in a switch case arm alongside `SOCK_DGRAM`:

* [cpython's `GAI_ANY` macro defined as 0](https://github.com/python/cpython/blob/2f369cafeeb4a4886b00396abd8a5f33e555e1c3/Modules/getaddrinfo.c#L68)
* [cpython's `GAI_ANY` used next to `SOCK_DGRAM` in a switch block](https://github.com/python/cpython/blob/2f369cafeeb4a4886b00396abd8a5f33e555e1c3/Modules/getaddrinfo.c#L355-368)

This cpython hack seems likely to me to be replicated elsewhere in the open source world and implies that most libc implementations likely don't have any `SOCK_*` macros with the value `0`.

This PR is paired with a corresponding [wasix-witx PR](https://github.com/john-sharratt/wasix-witx/pull/1). Both PRs utilize the same hack: introduce a shim enum variant to ensure that none of the other variants take the value `0`.

### DIFF
--- a/lib/wasi-types/schema/wasi/typenames.wit
+++ b/lib/wasi-types/schema/wasi/typenames.wit
@@ -703,8 +703,8 @@ enum socktype {
     /// socktype::dgram enum variant, can't have that same value). the same
     /// hack is in wasix-libc.
     unused,
-    dgram,
     %stream,
+    dgram,
     raw,
     seqpacket,
 }
--- a/lib/wasi-types/schema/wasi/typenames.wit
+++ b/lib/wasi-types/schema/wasi/typenames.wit
@@ -695,6 +695,14 @@ record subscription-fs-readwrite {
 // }
 
 enum socktype {
+    /// the unused variant is used to maintain backward compatibility with
+    /// older libc-based applications that were written where there is no
+    /// socket type variant with a '0' value (for example, cpython has a
+    /// macro definition used in a case arm alongside SOCK_DGRAM with the value
+    /// '0', which means SOCK_DGRAM, ultimately corresponding to this
+    /// socktype::dgram enum variant, can't have that same value). the same
+    /// hack is in wasix-libc.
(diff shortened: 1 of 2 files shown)


theduke (commented):
This will break all wasix code we have compiled that uses sockets.

I do think it makes sense to have this instead of doing the mapping in `libc`, but we'll need to change the `wasix` crate and the libc in unison with this PR.

Thoughts, @john-sharratt ?

waynr:
> This will break all wasix code we have compiled that uses sockets.
> 
> I do think it makes sense to have this instead of doing the mapping in `libc`, because that creates confusing ambiguities, but we'll need to change the `wasix` crate and the libc in unison with this PR.
> 
> And then re-compile everything using sockets, including the snapshot tests.
> 
> Thoughts, @john-sharratt ?

John and I discussed this last night and came to the same conclusion as you, but we decided we would wait to merge it until we are reasonably certain no more breaking changes are needed for what I'm working on.

theduke:
This was included in #3906 .

PR #3885, opened 2023-05-19
Rename js feature to web

### LINKED ISSUE #3844: Rename `js` feature to `web`
Rename the `wasmer` crate `js` feature to `web` (we should still support `js` feature, but it should just depend on `web`.

syrusakbary:
JS feature might be a bit confusing, so we decided to rename the `js` feature to `web` (we'll keep supporting `js` for now for compatibility reasons).

Fixes #3844


.github/workflows/test.yaml
```diff
@@ -109,14 +109,14 @@ jobs:
...
           curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
-      - name: make test-js
+      - name: make test-web
         run: |
-          make test-js
+          make test-web
 
       # The no_std functionality doesn't work at the moment - no point in testing it.
-      # - name: make test-js-core
+      # - name: make test-web-core
```
ptitSeb, line 117: web-core? we don't keep js-core here?

lib/api/src/lib.rs
```diff
@@ -451,10 +451,10 @@ mod sys;
 #[cfg(feature = "sys")]
 pub use sys::*;
 
-#[cfg(feature = "js")]
+#[cfg(feature = "web")]
 mod js;
```
ptitSeb, line 455: the feature is web but the mod stay `js`?

ptitSeb (approved):

lib/api/src/lib.rs
syrusakbary: Good call. Let me fix that

syrusakbary (commented):
DO NOT MERGE YET. I'm wondering if web is a good name when Wasmer get's compiled for Node.js

PR #3896, opened 2023-05-23
This fixes wasi-web which is broken in master

john-sharratt:
This fixes wasi-web which is broken after the registry refactor, this was undetected because the tests continue on error


syrusakbary (approved):
Looks good from my side, but I think it may be better if @Michael-F-Bryan reviews

theduke:
Let's wait for #3894 to be merged .

PR #3899, opened 2023-05-23
enable integration tests with debug binary

waynr:


tests/integration/cli/src/assets.rs
```diff
@@ -25,7 +25,7 @@ pub const WASMER_TARGET_PATH_2: &str = concat!(
 pub const WASMER_TARGET_PATH: &str =
     concat!(env!("CARGO_MANIFEST_DIR"), "/../../../target/release/");
 #[cfg(not(feature = "debug"))]
-pub const WASMER_TARGET_PATH2: &str = concat!(
```
theduke, line 28: You can already overwrite the path with the `WASMER_TEST_WASMER_PATH` env var. 

This would make more sense as a separate `make` command to me, which sets the env var, instead of a feature flag..

theduke (commented):
What would be nice is if `--debug` used `cargo run` for the CLI, so it's automatically re-compiled.

tests/integration/cli/src/assets.rs
theduke: Valid point, I missed that the `#[cfg]` was already in.

Let's just merge this as is for now, we can improve it separately.

theduke (approved):

PR #3900, opened 2023-05-23
lib/virtual-net: tokio::net::lookup_host requires host:port format

waynr:
This PR updates the call to [`tokio::net::lookup_host`](https://docs.rs/tokio/latest/tokio/net/fn.lookup_host.html) used by [`syscalls::wasix::wasi_resolve`](https://github.com/wasmerio/wasmer/blob/master/lib/wasi/src/syscalls/wasix/resolve.rs#L23).

Prior to this PR when `tokio::net::lookup_host` is given a hostname without a port, name resolution fails with an `Errno::Inval`. When the port is added as shown in the `tokio::net::lookup_host` docs, it resolves successfully. This is kind of weird as it seems to me like the only thing necessary to look up a hostname should be the hostname itself -- the port shouldn't factor into it.

### DIFF
--- a/lib/virtual-net/src/host.rs
+++ b/lib/virtual-net/src/host.rs
@@ -92,7 +92,11 @@ impl VirtualNetworking for LocalNetworking {
         port: Option<u16>,
         dns_server: Option<IpAddr>,
     ) -> Result<Vec<IpAddr>> {
-        tokio::net::lookup_host(host)
+        let port = match port {
+            Some(p) => p,
+            None => 0,
+        };
+        tokio::net::lookup_host(format!("{}:{}", host, port))
             .await
             .map(|a| a.map(|a| a.ip()).collect::<Vec<_>>())
             .map_err(io_err_into_net_error)
--- a/lib/virtual-net/src/host.rs
+++ b/lib/virtual-net/src/host.rs
@@ -92,10 +92,20 @@ impl VirtualNetworking for LocalNetworking {
         port: Option<u16>,
         dns_server: Option<IpAddr>,
     ) -> Result<Vec<IpAddr>> {
-        tokio::net::lookup_host(format!("{}:{}", host, port.unwrap_or(0)))
-            .await
-            .map(|a| a.map(|a| a.ip()).collect::<Vec<_>>())
-            .map_err(io_err_into_net_error)
+        tokio::net::lookup_host(
+            host.chars()
+                .find_map(|c| {
+                    if c == ':' {
+                        Some(host.to_string())
+                    } else {
+                        None
+                    }
+                })
+                .unwrap_or(format!("{}:{}", host, port.unwrap_or(0))),
+        )
+        .await
+        .map(|a| a.map(|a| a.ip()).collect::<Vec<_>>())
+        .map_err(io_err_into_net_error)
     }
 }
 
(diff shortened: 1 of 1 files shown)


waynr:
I verified this with a very simple reproduction of the problem i was seeing:
```
use tokio::net;

#[tokio::main]
async fn main() -> std::io::Result<()> {
    for addr in net::lookup_host("wasmer.io").await? {
        println!("socket address is {}", addr);
    }
    println!("Hello, world!");
    Ok(())
}
```

When I run this the output looks like
```
./target/debug/tokio-net-lookup-host
Error: Error { kind: InvalidInput, message: "invalid socket address" }
```

However, when I simply append a `:0` to the host string it works:
```
use tokio::net;

#[tokio::main]
async fn main() -> std::io::Result<()> {
    for addr in net::lookup_host("wasmer.io:0").await? {
        println!("socket address is {}", addr);
    }
    println!("Hello, world!");
    Ok(())
}
```

```
./target/debug/tokio-net-lookup-host
socket address is 172.64.80.1:0
socket address is [2606:4700:3031::6815:16cc]:0
socket address is [2606:4700:3033::ac43:ceeb]:0
Hello, world!
```

lib/virtual-net/src/host.rs
```diff
@@ -92,7 +92,7 @@ impl VirtualNetworking for LocalNetworking {
         port: Option<u16>,
         dns_server: Option<IpAddr>,
     ) -> Result<Vec<IpAddr>> {
-        tokio::net::lookup_host(host)
```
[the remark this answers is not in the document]
waynr: With the fix in this PR, I don't see why anything outside of this function would need to change. This function still takes a hostname and an optional port. The hostname and port are combined into `<hostname>:<port>` format; if the given port is `None` then `0` is used, which works fine.

All this PR does is fix an implementation detail.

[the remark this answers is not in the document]
theduke: What I'm saying is: we need to check if the `wasix` crate / the Rust stdlib are currently doing that, and if so then fix them before merging this.

lib/virtual-net/src/host.rs
```diff
@@ -92,10 +92,20 @@ impl VirtualNetworking for LocalNetworking {
         port: Option<u16>,
         dns_server: Option<IpAddr>,
     ) -> Result<Vec<IpAddr>> {
-        tokio::net::lookup_host(format!("{}:{}", host, port.unwrap_or(0)))
-            .await
-            .map(|a| a.map(|a| a.ip()).collect::<Vec<_>>())
-            .map_err(io_err_into_net_error)
+        tokio::net::lookup_host(
+            host.chars()
```
theduke, line 96: Nit: can we move this out to a separate variable? Putting a large expression in a function is hard to read.

lib/virtual-net/src/host.rs
```diff
@@ -92,10 +92,20 @@ impl VirtualNetworking for LocalNetworking {
...
+                .find_map(|c| {
```
theduke, line 97: Nitpick: `if host.contains(':')`?

theduke (approved):

lib/virtual-net/src/host.rs
waynr: doh!

PR #3902, opened 2023-05-24
TmpFs Memory Usage Tracking and Limiting

### LINKED ISSUE #3865: vfs: Parameterize `memfs` With Custom Allocator
Allow using a custom allocator for the in-memory file data of `memfs` files.

This is needed for restricting and metering memory usage.

Since the std `Allocator` trait is still unstable, we'll use the [allocator-api2](https://crates.io/crates/allocator-api2) crate, which implements the `std` api externally.

theduke:
feat(virtual-fs): Add TmpFs memory tracking and limiting

Hacky implementation for tracking and limiting the memory consumption of
in-memory files of the TmpFs / mem_fs.

Ideally this would use a Vec with a custom allocator, but the allocator
APIs are currently restricted to nightly Rust.

To keep both code impact and performance impact low, a TrackedVec is
added that can hold a FsMemoryLimiter, which has callbacks for growing
and for shrinking memory.

Without the new "tracked" feature enabled, a stub impl is added, which
does not do any tracking , and hence has minimal performance impact.

This should be rewritten to a sane implementaiton soon as part of a
larger virtual-fs rewrite.

Closes #3865

### DIFF
--- a/lib/virtual-fs/src/limiter.rs
+++ b/lib/virtual-fs/src/limiter.rs
@@ -115,7 +115,7 @@ mod tracked_vec {
     impl Drop for TrackedVec {
         fn drop(&mut self) {
             if let Some(limiter) = &self.limiter {
-                limiter.on_shrink(self.data.len());
+                limiter.on_shrink(self.data.capacity());
             }
         }
     }
--- a/lib/virtual-fs/src/limiter.rs
+++ b/lib/virtual-fs/src/limiter.rs
@@ -1,0 +1,210 @@
+use std::sync::Arc;
+
+use crate::FsError;
+
+pub use self::tracked_vec::TrackedVec;
+
+/// Allows tracking and limiting the memory usage of a memfs [`FileSystem`].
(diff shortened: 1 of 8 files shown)


lib/virtual-fs/src/limiter.rs
```diff
@@ -0,0 +1,210 @@
...
+
+            Ok(Self {
+                data: other,
+                limiter: self.limiter.clone(),
+            })
+        }
+
+        pub fn resize(&mut self, new_len: usize, value: u8) -> Result<(), FsError> {
+            let old_capacity = self.data.capacity();
+            self.data.resize(new_len, value);
```
ptitSeb, line 86: the resize can only grow? can't a new_size smaller then previous size happens?

theduke: Resize will never reduce the capacity of the vec.
It can truncate, but truncate does change the capacity, since that would require a re-allocation.

See https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.resize .

ptitSeb (approved):

PR #3905, opened 2023-05-24
fix(virtual-fs): Fix horrible write performance for in-memory files

### LINKED ISSUE #3903: Fix Incredibly Slow write Implementation of TmpFs In-Memory Files
Currently writing to somewhere else than the end of a file always allocates a new buffer  in the mem fs.
That will have ... horrible performance.

Need to fix this.

This is bad enough to call it a bug.

https://github.com/wasmerio/wasmer/blob/458330edc6c62fa1d0fdc53084b319d27aade855/lib/virtual-fs/src/mem_fs/file.rs#L1323-L1327

theduke:
The write() implementation for in-memory files used by the TmpFs /
mem_fs was terrible: it used to re-allocate and copy the whole data if
the write position was not at the end of the file.

This re-writes the implementation to grow the buffer in place if
required.

Also adds some tests.

Closes #3903

### DIFF
--- a/lib/virtual-fs/src/mem_fs/file.rs
+++ b/lib/virtual-fs/src/mem_fs/file.rs
@@ -1304,39 +1304,22 @@ impl File {
 
 impl File {
     pub fn write(&mut self, buf: &[u8], cursor: &mut u64) -> io::Result<usize> {
-        match *cursor {
-            // The cursor is at the end of the buffer: happy path!
-            position if position == self.buffer.len() as u64 => {
-                self.buffer.extend_from_slice(buf)?;
-            }
-
-            // The cursor is at the beginning of the buffer (and the
-            // buffer is not empty, otherwise it would have been
-            // caught by the previous arm): almost a happy path!
-            0 => {
-                // FIXME(perf,theduke): make this faster, it's horrible!
-                let mut new_buffer = TrackedVec::with_capacity(
-                    self.buffer.len() + buf.len(),
-                    self.buffer.limiter().cloned(),
-                )?;
-                new_buffer.extend_from_slice(buf)?;
-                new_buffer.append(&mut self.buffer)?;
-
-                self.buffer = new_buffer;
-            }
-
-            // The cursor is somewhere in the buffer: not the happy path.
-            position => {
-                self.buffer.reserve_exact(buf.len())?;
-
-                // FIXME(perf,theduke): make this faster, it's horrible!
-                let mut remainder = self.buffer.split_off(position as usize)?;
-                self.buffer.extend_from_slice(buf)?;
-                self.buffer.append(&mut remainder)?;
-            }
+        if buf.is_empty() {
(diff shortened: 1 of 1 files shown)


lib/virtual-fs/src/mem_fs/file.rs
```diff
@@ -1304,39 +1304,25 @@ impl File {
...
-                let mut new_buffer = TrackedVec::with_capacity(
-                    self.buffer.len() + buf.len(),
-                    self.buffer.limiter().cloned(),
-                )?;
-                new_buffer.extend_from_slice(buf)?;
-                new_buffer.append(&mut self.buffer)?;
-
-                self.buffer = new_buffer;
-            }
+        if buf.is_empty() {
```
ptitSeb, line 1307: shouldn't the file by truncated in that case, depending on the write mode?

lib/virtual-fs/src/mem_fs/file.rs
```diff
@@ -1304,39 +1304,22 @@ impl File {
...
+        let start_cursor: usize = (*cursor) as usize;
+        let end_cursor = start_cursor + buf.len();
 
-                // FIXME(perf,theduke): make this faster, it's horrible!
-                let mut remainder = self.buffer.split_off(position as usize)?;
-                self.buffer.extend_from_slice(buf)?;
-                self.buffer.append(&mut remainder)?;
-            }
+        if end_cursor > self.buffer.len() {
+            self.buffer.resize(end_cursor, 0)?;
```
ptitSeb, line 1316: doesn't the file (well the vec) be srinked if it's smaller?

lib/virtual-fs/src/mem_fs/file.rs
```diff
@@ -1399,3 +1382,52 @@ impl ReadOnlyFile {
...
+        assert_eq!(cursor, 11);
+
+        cursor = 6;
+        file.write(b"the universe!", &mut cursor).unwrap();
+        assert_eq!(file.buffer.deref(), b"alter the universe!");
+        assert_eq!(cursor, 19);
+
+        file.write(b" mkay?", &mut cursor).unwrap();
+        assert_eq!(file.buffer.deref(), b"alter the universe! mkay?");
+    }
```
ptitSeb, line 1432: maybe a test with a seek and truncate could be interesting?

lib/virtual-fs/src/mem_fs/file.rs
theduke: The `buf` is only the newly written data.

lib/virtual-fs/src/mem_fs/file.rs
theduke: Definitely.

We should have a proper test suite for `trait FileSystem` implementations.

I only added a tests for `write()` because that's the only thing I changed.

theduke:
Superseded by #4165

PR #3906, opened 2023-05-24
WASIX updates after superset changes

john-sharratt:
- allow_multiple_wasi_versions argument is removed, this is no longer a warning
- WASIX bus syscalls are all removed
- fixes for the WASIX filetype
- socket types have been reordered
- new functions added for sock_accept_v2 and thread_spawn_v2
- added in tracing for imports
- removed the local thread variable syscalls
- updated the snapshot tests


theduke (commented):
Looks reasonable to me.

We'll need follow-up work, like removing the `is_wasix` flag from WasiFs and unify behaviour.

That might also require changes in libc etc though.

theduke (approved):

PR #3907, opened 2023-05-24
Choose subversion of tracing crate in wasix

kajacx:
The wasix crate needs version at least 0.1.36, because of method "record" here: https://docs.rs/crate/tracing/0.1.36/source/src/span.rs

```rust
    pub fn record<Q: ?Sized, V>(&self, field: &Q, value: V) -> &Self
    where
        Q: field::AsField,
        V: field::Value,
    { ... }
```

but sys-utils is using 0.1.37, so might as well use that.


Michael-F-Bryan (approved):

theduke:
Thanks!

PR #3918, opened 2023-05-26
Workspace metadata

Michael-F-Bryan:
This PR...

- Switches to using workspace metadata (@ptitSeb  this means the version number for our "core" crates only need to be changed in the root `/Cargo.toml` file - which part of the release script will I need to update?) 
- Bumps the MSRV to 1.67 (stable is 1.69 and we usually lag by 2 versions)
- Upgrades to the 2021 edition and fixes a bunch of resulting lints
- Adds common dependencies to the `[workspace.dependencies]` section


syrusakbary (approved):

lib/wasm-interface/Cargo.toml
```diff
@@ -1,12 +1,14 @@
 [package]
 name = "wasmer-wasm-interface"
 version = "4.0.0-alpha.1"
```
ptitSeb, line 3: Why keeping this one?

ptitSeb (approved):

lib/wasm-interface/Cargo.toml
Michael-F-Bryan: Good catch! I'm guessing I skipped that one because I thought it was one of the packages that gets versioned separately.

PR #3919, opened 2023-05-26
Handle non-200 status codes when downloading WEBC files

Michael-F-Bryan:
I've refactored the `wasmer_wasix::http` module to use more types from the `http` crate. This lets us implement a `HttpResponse::is_ok()` method in terms of the `http::StatusCode` we get back, rather than making each client implementation decide whether a request failed or not.

That means when a package has recently been released and the backend hasn't finished generating a `*.webc` file for it (which we get a 503 for), instead of seeing an error like this:

```shell
$ wasmer run-unstable https://wapm.io/wasmer/wasmer-io
error: Unable to resolve "https://wapm.io/wasmer/wasmer-io"
│   1: Unable to detect the WEBC version
╰─▶ 2: Expected the magic bytes to be "\x00webc", but found "The w"
```

You will see something like this:

```shell
$ wasmer run-unstable https://wapm.io/wasmer/wasmer-io
error: Unable to resolve "https://wapm.io/wasmer/wasmer-io"
│   1: GET request to "https://wapm.io/wasmer/wasmer-io" failed
|   2: 503 Service Unavailable (Retry after: 5 minutes)
```

### DIFF
--- a/lib/wasi/src/runtime/resolver/polyfills.rs
+++ b/lib/wasi/src/runtime/resolver/polyfills.rs
@@ -1,6 +1,10 @@
 use std::path::Path;
 
+use anyhow::Error;
+use http::{HeaderMap, StatusCode};
 use url::Url;
+
+use crate::http::{HttpResponse, USER_AGENT};
 
 /// Polyfill for [`Url::from_file_path()`] that works on `wasm32-unknown-unknown`.
 pub(crate) fn url_from_file_path(path: impl AsRef<Path>) -> Option<Url> {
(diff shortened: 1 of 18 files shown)


lib/wasi/src/runtime/resolver/polyfills.rs
```diff
@@ -25,6 +29,29 @@ pub(crate) fn url_from_file_path(path: impl AsRef<Path>) -> Option<Url> {
...
+
+pub(crate) fn http_error(response: &HttpResponse) -> Error {
+    let status = response.status;
+
+    if status == StatusCode::SERVICE_UNAVAILABLE {
+        if let Some(retry_after) = response
+            .headers
+            .get("Retry-After")
+            .and_then(|retry_after| retry_after.to_str().ok())
+        {
```
theduke, line 47: Let's add a tracing::debug message here so it doesn't seem like it has stalled out completely.

theduke (approved):
Nice!

lib/wasi/src/runtime/resolver/polyfills.rs
Michael-F-Bryan: Done.

PR #3920, opened 2023-05-26
Fix filesystem access when running python with `wasmer run-unstable .`

Michael-F-Bryan:
This fixes `wasmer run-unstable .` for CPython by...

- Upgrading to a version of `webc` that includes the fixes for https://github.com/wasmerio/pirita/pull/115 and https://github.com/wasmerio/pirita/pull/117
- Adding a (slow & inefficient) method for merging in the filesystem for a package supplied with the `--use` flag

# Context

When you are setting up a WASI instance and add a `BinaryPackage` to a `WasiEnv` (e.g. via `WasiEnvBuilder::use_webc()`), we will merge the package's filesystem (i.e. all its volumes, and those of its dependencies) into the root filesystem being constructed for our new WASI instance (`WasiFsRoot`).

When you use the sandboxed filesystem (a `virtual_fs::TmpFileSystem`) via `WasiEnvBuilder::set_sandbox_fs()`, this merging is done like normal because `TmpFileSystem` has a very efficient `union()` method.

However, we don't get that unioning behaviour when the user supplies their own filesystem instance. Instead, the WASI instance will use an opaque `WasiFsRoot::Backing` and the only way for that package's files to become available is by manually copying the bytes from each file across (that's what this PR is doing).

The Python use case wasn't working because #3852 left that manual copy operation as a TODO.


lib/wasi/tests/runners.rs
```diff
@@ -68,6 +67,8 @@ mod wasi {
         });
         let err = handle.join().unwrap().unwrap_err();
 
+        dbg!(&err);
+
```
syrusakbary, line 71: Remove me?

Michael-F-Bryan: Done. That `dbg!()` was added temporarily while troubleshooting so I could see which error we were getting. I've adjusted the assertion so we print out unknown error kinds.

lib/wasi/src/runtime/resolver/utils.rs
```diff
@@ -56,13 +56,39 @@ pub(crate) fn http_error(response: &HttpResponse) -> Error {
     Error::msg(status)
 }
 
+pub(crate) fn file_path_from_url(url: &Url) -> Result<PathBuf, Error> {
+    debug_assert_eq!(url.scheme(), "file");
+
+    // Note: The Url::to_file_path() method is platform-specific
+    cfg_if::cfg_if! {
+        if #[cfg(any(unix, windows, target_os = "redox", target_os = "wasi"))] {
```
ptitSeb, line 64: redox?

ptitSeb (approved):

lib/wasi/src/runtime/resolver/utils.rs
Michael-F-Bryan: MacOS counts as `unix`. The `#[cfg]` is copied straight from the `Url::from_file_path()` function:

https://github.com/servo/rust-url/blob/206d37851b9313ef3a6ecb83766d9bbc65d466be/url/src/lib.rs#L2422-L2440

PR #3924, opened 2023-05-29
Rename "wasmer run-unstable" to "wasmer run"

### LINKED ISSUE #3801: Rename "wasmer run-unstable" to "wasmer run"
### Motivation

We introduced a `wasmer run-unstable` command when we rewrote the `wasmer run` code as part of the 3.2.0 release (#3650). By the time 4.0 comes out we should have had plenty of time to test this command and promote it to "stable".

### Proposed solution

Delete the existing `wasmer run` command and rename `wasmer run-unstable` to `wasmer run`.

Michael-F-Bryan:
Now that we've been using the `wasmer run-unstable` flow for a while, it's time to promote it to `wasmer run`.

Closes #3801.


Makefile
```diff
@@ -398,6 +400,10 @@ check-capi:
 build-wasmer:
 	$(CARGO_BINARY) build $(CARGO_TARGET_FLAG) --release --manifest-path lib/cli/Cargo.toml $(compiler_features) --features="webc_runner" --bin wasmer
 
+
+build-wasmer-debug:
```
ptitSeb, line 404: This is already present in the Makefile, same command but with the added "debug" feature.

theduke: Ah, yeah, I missed that.

Michael-F-Bryan: I've reverted the changes so `Makefile` matches `master` again.

lib/wasi/src/state/builder.rs
```diff
@@ -792,40 +793,197 @@ impl WasiEnvBuilder {
...
     }
 }
 
+/// Extract the exit code from a `Result<(), WasiRuntimeError>`.
+///
+/// We need this because calling `exit(0)` inside a WASI program technically
+/// triggers [`WasiError`] with an exit code of `0`, but the end user won't want
+/// that treated as an error.
+fn wasi_exit_code(
+    mut result: Result<(), WasiRuntimeError>,
```
theduke, line 872: There already is a function for this on `WasiRuntimeError`: https://github.com/wasmerio/wasmer/blob/eec73b70f45e15e48be592009b12a692e4c5b4c1/lib/wasi/src/lib.rs#L194

theduke (approved):
Approved, with two relatively small change requests.

tests/integration/cli/tests/snapshot.rs
```diff
@@ -10,10 +10,17 @@ use anyhow::Error;
...
 use insta::assert_json_snapshot;
+use regex::Regex;
 
+use once_cell::sync::Lazy;
 use tempfile::NamedTempFile;
 use wasmer_integration_tests_cli::get_wasmer_path;
 
+/// Logs tend to include unstable things like timestamps and function call
+/// durations, so we use this regex to remove them from our snapshot output.
+static LOG_PATTERN: Lazy<Regex> =
```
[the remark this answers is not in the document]
theduke: I believe something like `RUST_LOG=never` works.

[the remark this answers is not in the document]
Michael-F-Bryan: Yeah, `tracing` lets you do `RUST_LOG=off` for setting `LogLevelFilter::Off` and that's what you should get if you leave `spec.debug_output` as `false`. My understanding is that flag is just to help you debug failing snapshot tests, and we don't want log messages to end up in the snapshots.

Using `RUST_LOG=never` is actually the equivalent of `RUST_LOG=never=trace` (i.e. set the `never` crate to `trace` level).

PR #3926, opened 2023-05-30
Update the URL for wabt's tarball

Michael-F-Bryan:
The WAPM backend is in the process of moving its storage to a different bucket so the tarball URLs have changed.


ptitSeb (approved):

ayys (approved):
The URL changed because we are moving away from the registry, so this makes sense.

Note: In the future, you won't be able to `assert_eq!` the URLs because we will soon move to signed URLs which will be different for each request.

ptitSeb (approved):

Michael-F-Bryan:
This PR is no longer relevant because [c8370ef](https://github.com/wasmerio/wasmer/commit/c8370ef57610eb99a045b3f3de7fb62322eac97e) removed a lot of the old code for "installing" packages from WAPM.

PR #3933, opened 2023-05-31
Fix deserialize calls in doc comments

chipshort:
The renaming in wasmer 4 should be reflected in the doc comments.
`deserialize_checked` -> `deserialize`
`deserialize` -> `deserialize_unchecked`
It was missed in a few places.


theduke (approved):
Thanks!

PR #3950, opened 2023-06-05
tests: Remove debug_output option from snapshot tests.

### LINKED ISSUE #3942: Remove the debug_output setting from snaspshot tests
This setting makes litlte sense, since debug output is inherently unreliable and depends on lots of conditions.

theduke:
This option didn't really make any sense to begin with, tracing log
output is highly variable and can change at any time depending on
internals.

It should not be included in test snapshot output.

Closes #3942


Michael-F-Bryan (approved):
I'm happy to remove the `debug_output` flag. If people want to debug things then, like we were discussing on Slack, the best way is to manually run the command that gets printed out and enable logging yourself.

Issue #3808, opened 2023-04-21
Add Test For Stack Size Configuration


theduke:
Add a test for stack size configuration introduced in #3786 .

This is easy to do with a Rust WASM test that just allocates a large array on the stack.

syrusakbary:
Related: https://github.com/wasmerio/wasmer/issues/3800

PR #3957, opened 2023-06-06
Added a test for small Stack (for #3808)

ptitSeb:
Added a test for small Stack

for ticket #3808 

Note that the test needed to use Singlepass compiler, as Cranelift would optimized the code (even at optimisation level "None")and remove all the unused local, and so avoiding the stack overflow.

### DIFF
--- a/lib/api/src/sys/tunables.rs
+++ b/lib/api/src/sys/tunables.rs
@@ -279,7 +279,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(feature = "singlepass")]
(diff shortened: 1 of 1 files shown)


theduke (approved):
Nice work!

lib/api/src/sys/tunables.rs
```diff
@@ -277,4 +277,171 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    #[cfg(all(feature = "singlepass", not(target_os = "windows")))]
```
theduke, line 282: Why does it not work on Windows?

PR #3971, opened 2023-06-08
Added the missing sock_accept for preview1

john-sharratt:


theduke (commented):
Should also have a test for this to make sure a preview1 compiled `.wasm` works.

theduke (approved):

PR #3976, opened 2023-06-09
Fix so that conditional_union works again

john-sharratt:


Michael-F-Bryan (approved):
This should work for what we need :+1:

In the long term, we'll want to make that `merge_filesystems()` operation cheap, but that's not easily doable with the current `WasiFsRoot` design.

Michael-F-Bryan:
This was superseded by #1009.

PR #3978, opened 2023-06-09
Regression issue with bash in the browser

john-sharratt:


lib/api/src/js/externals/memory.rs
```diff
@@ -181,7 +181,7 @@ impl Memory {
         new_store: &mut impl AsStoreMut,
     ) -> Option<Self> {
         self.try_clone(&store)
-            .and_then(|mut memory| memory.duplicate().ok())
+            .and_then(|memory| memory.try_clone())
```
syrusakbary, line 184: Are we sure about this?

PR #3980, opened 2023-06-09
Re-work logging and command-line argument parsing

### LINKED ISSUE #3948: Showing unnecessary WARN when running a package
Right now running:

```
➜  wasmer run python/python -- -c "for x in range(5): print(f'{x} square: {x*x}')"
2023-06-05T12:03:33.728123Z  WARN ThreadId(01) virtual_fs::overlay_fs: Attempting to write to a readonly file path=/lib/.DS_Store
2023-06-05T12:03:33.728169Z  WARN ThreadId(01) wasmer_wasix::state::env: Unable to merge the package's filesystem into the main one error=permission denied
0 square: 0
1 square: 1
2 square: 4
3 square: 9
4 square: 16
```

It should be:

```
➜  wasmer run python/python -- -c "for x in range(5): print(f'{x} square: {x*x}')"
0 square: 0
1 square: 1
2 square: 4
3 square: 9
4 square: 16
```

Michael-F-Bryan:
When discussing #3948, we decided that the `wasmer` command shouldn't show any log messages unless the user explicitly sets the `$RUST_LOG` environment variable. That way there's no chance that logging will mess up the output you might get from a program.

I've also re-worked the CLI parsing code to be vanilla clap rather than having a mix of manual argument parsing and two different clap parsers. This means we no longer need to update that big `match` statement every time a new command or alias is added, plus it makes the code a lot more readable. 

Fixes #3948.


lib/cli/src/commands/run.rs
```diff
@@ -58,8 +57,6 @@ static WASMER_HOME: Lazy<PathBuf> = Lazy::new(|| {
 /// The unstable `wasmer run` subcommand.
 #[derive(Debug, Parser)]
 pub struct Run {
-    #[clap(flatten)]
-    verbosity: clap_verbosity_flag::Verbosity<WarnLevel>,
```
theduke, line 62: We should keep the verbosity flag IMO.

Lots of users won't know about RUST_LOG.

lib/cli/src/cli.rs
```diff
@@ -218,6 +218,8 @@ pub fn wasmer_main() {
 }
 
 fn wasmer_main_inner() -> Result<(), anyhow::Error> {
+    crate::logging::set_up_logging();
```
theduke, line 221: Not sure about this.

The "do not show any logs if not enabled" restriction is primarily relevant for `run`, for the others having a default level of WARN actually makes sense.

theduke (changes requested):

lib/cli/src/cli.rs
Michael-F-Bryan: That'll make things pretty annoying to maintain because we'll need to remember to initialize logs for every sub-command in the CLI, plus any sub-commands we may add in the future.

What are your thoughts on still running `set_up_logging()` from `wasmer_main_inner()`, but we'll make the `verbosity` flags global so anyone can do `--verbose` for any command? Instead of going through `clap-verbosity-flags`, I might add some custom logic so no `--verbose` flag leaves logs off, one occurrence means INFO logs, two means DEBUG, and so on. With this setup it probably doesn't make sense to have a `--quiet` flag.

lib/cli/src/commands/run.rs
syrusakbary: Agreed

lib/cli/src/cli.rs
syrusakbary: I’m a bit more agnostic on the result here

lib/cli/src/commands/run.rs
Michael-F-Bryan: @theduke and @syrusakbary what are the semantics you want for those verbosity flags? Should `-vvv` increase the default tracing log level from WARN to TRACE, or should it be affecting how much human-friendly output we are showing (progress bars, etc.)?

Also, what do we want the `--quiet` flag to do? Using `clap_verbosity_flag`, adding one `-q` will decrease the default tracing log level from WARN to ERROR.

theduke (approved):
Nice.

Love getting rid of the command hacks!

syrusakbary (approved):
Looks good, but can we have a test testing that `wasmer awasifile.wasm` still works? Thanks!

Michael-F-Bryan:
@syrusakbary most of the `wasmer run` integration tests already check `wasmer run whatever.wasm` and `wasmer whatever.wasm`, so we should be all good on that front 👌 

For example,

https://github.com/wasmerio/wasmer/blob/2b867faf870d7120e38c14ec0b6d424faa4ef23f/tests/integration/cli/tests/run.rs#L339-L354

PR #3983, opened 2023-06-09
Introduce WAPM query caching to reduce `wasmer run`'s startup delay

### LINKED ISSUE #3962: 1 second delay when starting `wasmer run`
### Motivation

Every time we run `wasmer run`, there seems to be a 500-1000ms delay before the WASI instance starts.

I believe this is because of the `wasmer_wasix::runtime::resolver::WapmSource` querying WAPM every time, but I'll double-check the logs.

### Proposed solution

The best way to avoid hitting the WAPM backend every time we start up is to cache the result of queries that have been made in the past. This could either be the original GraphQL queries or the more high-level *"which versions satisfy `some/package@^1.0`?"*

### Additional context

This delay is causing a fair amount of flakiness in Wasmer's CI because the [`run_test_caching_works_for_packages_with_versions`](https://github.com/wasmerio/wasmer/blob/e85be593f8bb9f21552e5d81be1602ae079158dd/tests/integration/cli/tests/run.rs#L479-L513) test relies on measuring a command's run time to infer whether we hit the cache or not.

Michael-F-Bryan:
We've discovered that there is a ~1 second delay on startup when running something like `wasmer run python/python`. After some investigation with @dynamite-bud, we narrowed it down to a GraphQL query which will be executed every single time we try to fetch metadata from something that looks like a WAPM package.

https://github.com/wasmerio/wasmer/blob/f93f083a634c1b72dbe8dbb9f2e65bfab9a46944/lib/wasix/src/runtime/resolver/wapm_source.rs#L53-L78

The simplest way to solve this is by caching the response from this sort of GraphQL query, marking the cached value as stale if it's older than (for example) 10 minutes or we've just run `wasmer publish`. It'd be nice if we could use a sqlite database for the cache, but I'm not sure how that'd go on the web. 

Fixes #3962


lib/wasix/src/runtime/resolver/wapm_source.rs
```diff
@@ -144,6 +208,149 @@ fn decode_summary(pkg_version: WapmWebQueryGetPackageVersion) -> Result<PackageS
...
 
+/// A local cache for package queries.
+#[derive(Debug, Clone)]
+struct FileSystemCache {
+    cache_dir: PathBuf,
+    timeout: Duration,
+}
+
+impl FileSystemCache {
+    const DEFAULT_TIMEOUT: Duration = Duration::from_secs(10 * 60);
```
ptitSeb, line 219: So, cache is valid 10 min.
Could cache be disabled from command line also?

ptitSeb (approved):

lib/wasix/src/runtime/resolver/wapm_source.rs
ptitSeb: tere are indeed many cache, but not all are "internet" cache. For example, I don't think wapm cache should be mixed with compiler cache (there is a `--disable-cache` for that already).
But some control over the caches layer on command line would be nice for a later version.

lib/wasix/src/runtime/resolver/wapm_source.rs
```diff
@@ -144,6 +208,149 @@ fn decode_summary(pkg_version: WapmWebQueryGetPackageVersion) -> Result<PackageS
...
+
+    fn new(cache_dir: impl Into<PathBuf>) -> Self {
```
theduke, line 221: Can we make the timeout configurable?

I'd might need to play around with it on Deploy.

Michael-F-Bryan: I've added a `timeout: Duration` parameter to the `WapmSource::with_local_cache()` method :+1:

PR #3985, opened 2023-06-09
Display human-friendly progress when starting `wasmer run`

### LINKED ISSUE #3981: The `wasmer run` command doesn't show any output when starting
### Describe the bug

When running the `wasmer run` command on the latest `master` ([f93f083](https://github.com/wasmerio/wasmer/commit/f93f083a634c1b72dbe8dbb9f2e65bfab9a46944)) there is a 1-second delay on startup. The delay itself has its own ticket (#3962), but this highlights that the `wasmer run` command needs to show some sort of interactive output so users don't get impatient and think it has locked up.

The messages I'm thinking we should show are:

- Loading the initial target (e.g. a `*.webc` or `*.wasm` on disk or something remote) (spinner)
- The resolution stage (spinner)
- Packages as they are being downloaded from WAPM or the internet (progress bar)
- Initializing the WASI environment & runner (spinner)

Michael-F-Bryan:
This fixes #3981 by adding some spinners during certain parts of the startup process. 

I've mostly done this by creating a wrapper `wasmer_wasix::Runtime` implementation that will display progress bars during certain operations. Unfortunately, our current abstractions don't really expose the hooks we'd want for fine-grained progress updates like telling people how many bytes have been downloaded for a certain package, and some expensive operations (e.g. module compilation) aren't possible to hook into for reporting progress, but I guess it's better than nothing 🤷

### DIFF
--- a/lib/cli/src/commands/run.rs
+++ b/lib/cli/src/commands/run.rs
@@ -92,6 +92,12 @@ impl Run {
     }
 
     fn execute_inner(self, output: Output) -> Result<(), Error> {
+        let pb = ProgressBar::new_spinner();
+        pb.set_draw_target(output.draw_target());
+        pb.enable_steady_tick(TICK);
+
+        pb.set_message("Initializing the WebAssembly VM");
+
         let runtime = tokio::runtime::Builder::new_multi_thread()
             .enable_all()
             .build()?;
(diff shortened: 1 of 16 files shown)


lib/cli/src/commands/run.rs
```diff
@@ -104,12 +107,24 @@ impl Run {
             self.wasi
                 .prepare_runtime(store.engine().clone(), &self.wasmer_dir, handle)?;
 
+        let progress = MultiProgress::new();
+        progress.set_draw_target(output.draw_target());
+
+        // This is a slow operation, so let's temporarily wrap the runtime with
+        // something that displays progress
+        let monitoring_runtime = MonitoringRuntime::new(runtime, progress.clone());
+        let pb = progress.add(ProgressBar::new_spinner().with_message("Getting Ready..."));
```
ptitSeb, line 116: "Getting Ready" doesn't sound like a good message. Isn't there something more meaningful?

ptitSeb (approved):

lib/cli/src/commands/run.rs
Michael-F-Bryan: Fixed.

PR #3989, opened 2023-06-12
OverlayFS now has COW

john-sharratt:


lib/virtual-fs/src/lib.rs
```diff
@@ -333,7 +336,7 @@ pub trait VirtualFile:
     fn set_len(&mut self, new_size: u64) -> Result<()>;
 
     /// Request deletion of the file
-    fn unlink(&mut self) -> Result<()>;
+    fn unlink<'a>(&'a mut self) -> LocalBoxFuture<'a, Result<()>>;
```
theduke, line 339: The `LocalBoxFuture` here and elsewhere in the trait are problematic.

This means tasks that these futures can never be moved across threads, and hence never just be run in a regular tokio runtime.

Is that really what we want?

I assume this was done to not require shared state, like a clone  of the WasiEnv?

theduke (commented):
Mostly looks good to me.

Michael-F-Bryan:
As best I can tell, this is a summary of what this PR does:

- The `FileSystem::rename()` method is now async
- Introduces support for "whiteout files" - after a bit of googling (see [here](https://www.madebymikal.com/interpreting-whiteout-files-in-docker-image-layers/) and [here](https://askubuntu.com/questions/39826/what-does-whiteout-mean-in-a-copy-on-write-filesystem)), this is an idea Docker (technically AUFS) uses for their unionfs where "deleting" a file from a read-only filesystem will actually create a new file prepended with `.wh.` that tells the unionfs to treat the file as deleted
- Fixes the `conditional_union` so we merge a package's filesystem into the main filesystem when spawning (previously only worked when you used the "sandboxed" `WasiFsRoot`)

lib/virtual-fs/src/lib.rs
Michael-F-Bryan: We've moved everything over to `BoxFuture` now.

Michael-F-Bryan (approved):

PR #3991, opened 2023-06-12
Renamed most appearances of Wapm to Wasmer

syrusakbary:
Renamed most appearances of Wapm to Wasmer


Michael-F-Bryan (approved):
Everything looks good as far as I can tell :+1:

PR #3992, opened 2023-06-13
[CI][RISCV64] Try to use bookworm debian instead of sid

ptitSeb:
Try to use bookworm debian instead of sid


theduke (approved):
There is one potential concern here:

Is bookworm on a newer libc version?
If so, it would probably mean our pre-built wasmer for releases won't be compatible with older distros?

ptitSeb:
> There is one potential concern here:
> 
> Is bookworm on a newer libc version? If so, it would probably mean our pre-built wasmer for releases won't be compatible with older distros?

Bookworm is basically the 1st stable Debian version for RISCV64. We were using unstable sid before, so switching to stable is an improvment and staying on unstable or testing is a lesser choice imo

PR #3996, opened 2023-06-13
wasix: Switch Console to WasiRunner + Various WASI Improvements

theduke:
This PR switches the wasix Console to use the WasiRunner instead of custom 
setup.

This removes code duplication, and prevents potential resulting from differences
in bootstrapping for WASI environments.

To make this happen, the PR also extends the WasiRunner with an option to provide
a custom TmpFileSystem.

NOTE: this PR is based on https://github.com/wasmerio/wasmer/pull/3989, which 
must be merged first.

- chore: Remove unused name_map from WasiFs
- wasix: Allow passing a custom root fs to WasiRunner/CommonWasiOptions
- wasix: Switch the Console over to the WasiRunner
- chore: remove redundant webc_runner[_wasi] features
- feat(wasix): Extend Runtime trait with module loading/compiling


lib/wasix/src/lib.rs
```diff
@@ -169,6 +169,20 @@ pub enum SpawnError {
     /// Some other unhandled error. If you see this, it's probably a bug.
     #[error("unknown error found")]
     UnknownError,
+    #[error("runtime error")]
+    Runtime(#[from] WasiRuntimeError),
+    #[error("{0}")]
+    Other(String),
```
Michael-F-Bryan, line 175: Rather than adding a `Other(String)` variant, would it be better to modify `UnknownError` to be `UnknownError(#[cause] Box<dyn std::error::Error + Send + Sync>)`?

lib/wasix/src/runtime/mod.rs
```diff
@@ -76,6 +79,63 @@ where
     fn tty(&self) -> Option<&(dyn TtyBridge + Send + Sync)> {
         None
     }
+
+    /// Load a a Webassembly module, trying to use a pre-compiled version if possible.
+    fn load_module<'a>(&'a self, wasm: &'a [u8]) -> BoxFuture<'a, Result<Module, anyhow::Error>> {
```
Michael-F-Bryan, line 84: I'm sure we'll miss one or two places were we run `Module::new()`, but I'm happy with this addition 👍 

It also helps with #3985 because now we've got something that can be hooked into for showing a "Compiling..." spinner.

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -19,49 +19,27 @@ use crate::{runtime::SpawnMemoryType, Runtime, WasiEnv, WasiFunctionEnv};
...
-                );
-            }
-            module
-        }
-        (None, None) => {
-            error!("package has no entry [{}]", name,);
+    let wasm = match binary.entrypoint_bytes() {
+        Some(wasm) => wasm,
+        None => {
+            error!("package has no entrypoint command [{}]", name,);
```
Michael-F-Bryan, line 29: Can you modify this to add a bit more context?

```rust
tracing::error!(
  command=name,
  pkg.name=%binary.package_name,
  pkg.version=%binary.version,
  "Unable to spawn a command because its package has no entrypoint",
);
```

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -19,49 +19,27 @@ use crate::{runtime::SpawnMemoryType, Runtime, WasiEnv, WasiFunctionEnv};
...
             env.cleanup(Some(Errno::Noexec.into())).await;
             return Err(SpawnError::CompileError);
         }
     };
 
+    let module = match runtime.load_module_sync(wasm) {
```
Michael-F-Bryan, line 35: I think we might want to use the async version of `load_module()` here. The `spawn_exec()` function is async, so `load_module_sync()` will probably trigger a runtime-in-runtime panic from Tokio.

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -19,49 +19,27 @@ use crate::{runtime::SpawnMemoryType, Runtime, WasiEnv, WasiFunctionEnv};
...
+        Ok(module) => module,
+        Err(err) => {
+            error!("failed to compile module '{name}': {err}",);
```
Michael-F-Bryan, line 38: I normally like to use `tracing`'s structured logging because it has special cases for things implementing `std::error::Error`. Not doing string interpolation in the message also makes analysing bulk logs from production a lot easier because you can make the logger emit JSON, ingest those logs, then set up a query that filters for `record.message == "Failed to compile the module"`.

```rust
tracing::error!(
  command=name,
  error=&err as &dyn std::error::Error,
  "Failed to compile the module",
);
```

Michael-F-Bryan (changes requested):
we're setting up `WasiRunner` correctly as far as I can tell. The optional `root_fs` in `CommonWasiOptions::prepare_webc_env()` is a bit of a code smell, but I think that says more about the complexity of what we're trying to do and the limitations of our filesystem layer.

I've left a couple trivial suggestions about using structured logs a bit more because it'll help when spelunking through logs later on.

I don't think we should merge this without adding some tests to the `Console`, though. Try adding a test for something non-trivial like `python/python` to make sure the filesystem is being set up properly. Unit tests in the `console` module would be best, but I understand `Console` isn't in the most testable state at the moment.

lib/wasix/src/lib.rs
theduke: Yeah, though the primary motivation here is for `anyhow::Error`, so we'll a std error wrapper for anyhow. 

We should have a more extensive custom error type for `Runtime::load_module` though.

lib/wasix/src/bin_factory/exec.rs
theduke: I just copieed the old log I think , but yeah makes sense.

lib/wasix/src/bin_factory/exec.rs
theduke: :+1:

lib/wasix/src/bin_factory/exec.rs
theduke: yeah, as above, that trace already existed.

theduke:
> I don't think we should merge this without adding some tests to the Console, though.

Depending on when we merge this, I think we can also just remove the console now. 

It is only used by Deploy, and now it has become a very simple wrapper that does almost nothing, so I should be able to switch Deploy away from it.

lib/cli/src/commands/run.rs
```diff
@@ -566,29 +566,11 @@ impl ExecutableTarget {
...
-                            );
-                        }
-
-                        Module::new(&engine, &wasm)
-                            .with_context(|| format!("Unable to compile \"{}\"", path.display()))?
-                    }
-                };
+                pb.set_message("Compiling to WebAssembly");
+                let module = runtime
+                    .load_module_sync(&wasm)
```
Michael-F-Bryan, line 572: Yay 🚀

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -19,49 +19,36 @@ use crate::{runtime::SpawnMemoryType, Runtime, WasiEnv, WasiFunctionEnv};
...
 
+    let module = match runtime.load_module(wasm).await {
+        Ok(module) => module,
+        Err(err) => {
+            tracing::error!(
+                command = name,
+                error = &*err,
+                "Failed to compile the module",
+            );
+            return Err(SpawnError::CompileError);
```
Michael-F-Bryan, line 48: Do we need to do an `env.cleanup()` here?

lib/wasix/src/lib.rs
```diff
@@ -170,6 +169,20 @@ pub enum SpawnError {
     /// Some other unhandled error. If you see this, it's probably a bug.
     #[error("unknown error found")]
     UnknownError,
+    #[error("runtime error")]
+    Runtime(#[from] WasiRuntimeError),
+    #[error("{0}")]
```
Michael-F-Bryan, line 174: Let's use `#[error(transparent)]` here so we don't print out the same error message twice.

```suggestion
    #[error(transparent)]
```

lib/wasix/src/lib.rs
```diff
@@ -209,6 +222,33 @@ impl WasiRuntimeError {
...
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        Some(&*self.0)
+    }
+}
+
+impl From<anyhow::Error> for AnyhowStdError {
+    fn from(error: anyhow::Error) -> Self {
+        Self(error)
+    }
+}
```
Michael-F-Bryan, line 250: I don't think adding a wrapper around `anyhow::Error` is going to help us much here. The `anyhow` crate is 1.0 so we can rely on it being stable, which means introducing a newtype like this feels like we're adding friction unnecessarily.

There is a `impl From<Error> for Box<dyn std::error::Error + Send + Sync + 'static>` so we should be able to convert from `anyhow::Error` to `SpawnError::Other` directly:

```rust
impl From<anyhow::Error> for SpawnError {
  fn from(e: anyhow::Error) -> SpawnError {
    SpawnError::Other(e.into())
  }
}
```

lib/wasix/src/os/console/mod.rs
```diff
@@ -252,6 +199,40 @@ impl Console {
...
+
+        let builder = crate::runners::wasi::WasiRunner::new()
+            .with_envs(self.env.clone().into_iter())
+            .with_args(args)
+            .with_capabilities(self.capabilities.clone())
+            .prepare_webc_env(prog, &wasi_opts, &pkg, self.runtime.clone(), Some(root_fs))
+            // TODO: better error conversion
+            .map_err(|err| SpawnError::Other(Box::new(crate::AnyhowStdError(err))))?;
+
+        // TODO: no unwrap!
```
Michael-F-Bryan, line 223: I think this TODO is no longer relevant?

```suggestion
```

lib/wasix/src/os/console/mod.rs
```diff
@@ -298,3 +280,65 @@ impl Console {
             .ok();
     }
 }
+
+#[cfg(test, not(target_family = "wasm"))]
```
Michael-F-Bryan, line 284: Should we be using `all()` here?

```suggestion
#[cfg(all(test, not(target_family = "wasm")))]
```

lib/wasix/src/runners/wasi_common.rs
```diff
@@ -85,13 +87,15 @@ impl CommonWasiOptions {
...
+
 fn prepare_filesystem(
+    root_fs: TmpFileSystem,
     mapped_dirs: &[MappedDirectory],
     container_fs: Arc<dyn FileSystem>,
     builder: &mut WasiEnvBuilder,
-) -> Result<Box<dyn FileSystem + Send + Sync>, Error> {
-    let root_fs = RootFileSystemBuilder::default().build();
-
+) -> Result<ContainerFs, Error> {
```
Michael-F-Bryan, line 98: It might be better to return a `impl FileSystem + Send + Sync` here. Something I do quite frequently when debugging is to add a `let fs = virtual_fs::TraceFileSystem::new(fs);` at the bottom of the function so all filesystem calls get logged, and updating the return type every time would be annoying.

lib/wasix/tests/runners.rs
```diff
@@ -1,4 +1,6 @@
-#![cfg(feature = "webc_runner")]
+// Exclude runner tests from wasm targets for now, since they don't run properly
+// there.
+#![cfg(not(target_family = "wasm"))]
```
Michael-F-Bryan, line 3: I don't think we run the `wasmer-wasix` crate's unit tests on WebAssembly, do we?

Michael-F-Bryan (approved):
I've left some suggestions, but there's nothing blocking this PR from merging :+1:

lib/wasix/tests/runners.rs
theduke: The problem is that the test fails to compile due to `TokioTaskManager` not being available on wasm targets.

lib/wasix/src/lib.rs
theduke: I forget about that `From` impl every damn time...

lib/wasix/src/bin_factory/exec.rs
theduke: cleanup() is actually sort of redundant now, since the forking state was cleaned up. 
Especially while the module hasn't yet had a chance to open files, since open file cleanup is all that's happening now.

But let's leave it in for now.

theduke:
Commits from this PR were included in #4050  and hence are already merged.

PR #4002, opened 2023-06-16
[RISCV64][CI] Use simpler way to install riscv64 packages

ptitSeb:
Use simpler way to install riscv64 packages


Michael-F-Bryan (approved):
I'm not 100% familiar with how we set up CI for riscv, but everything looks pretty straightforward. I'm happy to merge this once CI goes green.

PR #4004, opened 2023-06-16
Ignoring archived package versions when querying the registry

### LINKED ISSUE #4000: Unable to run the wasmer/python package
### Describe the bug

When getting ready for the Wasmer 4.0-beta.3 release, we noticed that the `wasmer/python` package doesn't mount its filesystem properly.

### Steps to reproduce

```console
$ cargo run --release --features=compiler,cranelift --package=wasmer-cli -- run wasmer/python@3.12.0-build.5
    Finished release [optimized] target(s) in 0.72s
     Running `target/release/wasmer run 'wasmer/python@3.12.0-build.5'`
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Python path configuration:
  PYTHONHOME = (not set)
  PYTHONPATH = (not set)
  program name = 'python'
  isolated = 0
  environment = 1
  user site = 1
  safe_path = 0
  import site = 1
  is in build tree = 0
  stdlib dir = '/usr/local/lib/python3.12'
  sys._base_executable = ''
  sys.base_prefix = '/usr/local'
  sys.base_exec_prefix = '/usr/local'
  sys.platlibdir = 'lib'
  sys.executable = ''
  sys.prefix = '/usr/local'
  sys.exec_prefix = '/usr/local'
  sys.path = [
    '/usr/local/lib/python312.zip',
    '/usr/local/lib/python3.12',
    '/usr/local/lib/python3.12/lib-dynload',
  ]
Fatal Python error: init_fs_encoding: failed to get the Python codec of the filesystem encoding
Python runtime state: core initialized
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00080018 (most recent call first):
  <no Python frame>
```

### Additional context

This happens on e06601c3d32f5a58a44c5fc663f0ce6fc2379978.

For more context, see the discussion [on Slack](https://wasmerio.slack.com/archives/C02CK7101PG/p1686820333956909).

Michael-F-Bryan:
# Description

This PR updates the `WapmSource` to respect a package version's `isArchived` field. All packages with `isArchived` set will  now be ignored while determining which versions satisfy the query.

# Context

We were seeing #4000 because of a combination of semver quirks and not respecting a package version's `isArchived` flag:

- When running a package without a version constraint (i.e. `wasmer run wasmer/python` instead of `wasmer run wasmer/python@3.12.0`) we use the catch-all `*` version constraint to say "just give me the latest version".
- You need to explicitly ask for pre-release versions - they won't be picked up by `*`
- As far as semver is concerned, pre-releases (e.g. `3.12.0-build.5` or `3.12.0-beta`) are considered to have a lower version number than their released version (i.e. `3.12.0`).
- The `WapmSource` wasn't respecting the `isArchived` flag when looking for packages that satisfy a constraint

This all meant that when faced with the following versions for `wasmer/python`...

- `3.12.0-build.5`
- `3.12.0-build.2` (archived)
- `3.12.0` (archived)
- `3.12.0-build.4` (archived)
- `3.12.0-build.3` (archived)

... We'll always pick version `3.12.0` (which is broken) because it has the "highest" version number that matches `*`.

Hence why running `wasmer run wasmer/python` would fail but explicitly `wasmer run wasmer/python@3.12.0-build.5` would work.

Fixes #4000

### DIFF
--- a/lib/wasix/src/runtime/resolver/wapm_source.rs
+++ b/lib/wasix/src/runtime/resolver/wapm_source.rs
@@ -161,6 +161,7 @@ impl Source for WapmSource {
             Some(WapmWebQueryGetPackage { versions }) => versions,
             None => return Ok(Vec::new()),
         };
+        let mut archived_versions = Vec::new();
 
         for pkg_version in versions {
             tracing::trace!(?pkg_version, "checking package version");
@@ -171,6 +172,7 @@ impl Source for WapmSource {
                     pkg.version=%version,
                     "Skipping an archived version",
                 );
+                archived_versions.push(version);
                 continue;
             }
 
@@ -184,6 +186,25 @@ impl Source for WapmSource {
                             "Skipping version because its metadata couldn't be parsed"
                         );
                     }
+                }
+            }
+        }
+
+        if summaries.is_empty() {
+            match archived_versions.as_slice() {
+                [] => {
+                    // looks like this package couldn't be satisfied at all.
+                }
+                [version] => {
+                    anyhow::bail!(
+                        "The only version satisfying the constraint, {version}, is archived"
+                    );
+                }
+                [first, rest @ ..] => {
+                    let num_others = rest.len();
+                    anyhow::bail!(
+                        "Unable to satisfy the request, {first}, and {num_others} are all archived"
+                    );
                 }
             }
         }
--- a/lib/wasix/src/runtime/resolver/wapm_source.rs
+++ b/lib/wasix/src/runtime/resolver/wapm_source.rs
@@ -166,7 +166,15 @@ impl Source for WapmSource {
             tracing::trace!(?pkg_version, "checking package version");
             let version = Version::parse(&pkg_version.version)?;
 
-            if !pkg_version.is_archived && version_constraint.matches(&version) {
+            if pkg_version.is_archived {
+                tracing::debug!(
+                    pkg.version=%version,
+                    "Skipping an archived version",
+                );
+                continue;
+            }
+
+            if version_constraint.matches(&version) {
                 match decode_summary(pkg_version) {
                     Ok(summary) => summaries.push(summary),
                     Err(e) => {
--- a/lib/wasix/src/runtime/resolver/wapm_source.rs
+++ b/lib/wasix/src/runtime/resolver/wapm_source.rs
@@ -164,9 +164,9 @@ impl Source for WapmSource {
 
         for pkg_version in versions {
             tracing::trace!(?pkg_version, "checking package version");
-
             let version = Version::parse(&pkg_version.version)?;
-            if version_constraint.matches(&version) {
+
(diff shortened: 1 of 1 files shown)


lib/wasix/src/runtime/resolver/wapm_source.rs
```diff
@@ -86,9 +86,9 @@ impl Source for WapmSource {
 
         for pkg_version in versions {
             tracing::trace!(?pkg_version, "checking package version");
-
             let version = Version::parse(&pkg_version.version)?;
-            if version_constraint.matches(&version) {
+
+            if !pkg_version.is_archived && version_constraint.matches(&version) {
```
ptitSeb, line 91: so archived version can never be selected, even with the precise version match? Probably ok for most cases anyway.

ptitSeb (approved):

lib/wasix/src/runtime/resolver/wapm_source.rs
theduke, line 91: This is suboptimal for UX. 

Can we: 

* if the constraint matches the version exactly (`=1.1.1`), show an error that the given version was archived
* if the the constraint matches loosely (eg `1.1.1`), and there is no other available version compatible version, show an error that the only matching version was archived?

* produce a terminal debug message that an archived version was skipped, for debugging?

Michael-F-Bryan: Yeah, printing a debug message when we skip over an archived package version is a good idea :+1:

I'm quite hesitant to start poking around inside `semver::VersionReq` and adding special cases, though. While it might feel convenient, that also opens us up to some pretty magical behaviour.

My thinking is that it's easy to give people a single rule that lockfiles reuse previous dependency resolution results and archiving a package will always remove it from being considered when doing future dependency resolution. 

However, things get a lot muddier once you start saying archived versions are almost always ignored, but we'll also include it if you add a strict `=` constraint, or if you didn't precisely ask for that version but there happened to be no other versions in that range.

Michael-F-Bryan: We had a chat about this during today's standup and @syrusakbary feels the same way about adding edge cases.

The redirect for `Accept: application/webc` will always work for archived packages, so if you want you can always do `wasmer run https://wasmer.io/python/python@1.2.3` if you want.

theduke: > if the constraint matches the version exactly (=1.1.1), show an error that the given version was archived

What I wrote above is not that these situations should allow usage of the archived version, I was just referring to producing concrete errors, which state  specific version is archived (`=xxx`) and thus can't be used, or that the only matching version is archived, rather than just saying "could not find a matching version"

Michael-F-Bryan: Ah, gotcha. I've added something that should error out under those conditions.

PR #4005, opened 2023-06-16
Remove "build_linux_riscv64" as a pre-requisite for running CLI integration tests

Michael-F-Bryan:
# Description

We've started getting build issues in CI with the `build_riscv64` jobs because of changes in the Debian repositories. This PR removes `build_linux_riscv64` (both the C API and Wasmer CLI) as a pre-requisite for running the CLI integration tests.

We don't test the CLI for riscv at the moment, so this shouldn't be a problem.

This PR **doesn't** ignore the riscv builds completely, so `master` will still be red. It just means we can unblock PRs currently waiting for CI to pass. 

When #4002 is merged again, we should either revert this PR or update branch protections to make the job required.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -728,6 +728,7 @@ jobs:
           name: capi-macos-x64
           path: package/cache/wasmercache4
       - uses: actions/download-artifact@v3
+        if: false
         with:
           name: capi-linux-x64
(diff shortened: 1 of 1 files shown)


ptitSeb (approved):

Michael-F-Bryan:
This PR is no longer needed because riscv was fixed in #4002.

Issue #3793, opened 2023-04-20
`assertion failed: prev.start > max` on first invocation (without fs cache)


kdy1:
### Describe the bug

https://github.com/swc-project/swc/issues/7304

The Wasm plugin runner of SWC fails with an assertion failure.

<details>
    <summary> Versions of wasmer & deps </summary>

```
wasmer v3.2.0
├── bytes v1.4.0
│   └── serde v1.0.160
│       └── serde_derive v1.0.160 (proc-macro)
│           ├── proc-macro2 v1.0.56
│           │   └── unicode-ident v1.0.8
│           ├── quote v1.0.26
│           │   └── proc-macro2 v1.0.56 (*)
│           └── syn v2.0.15
│               ├── proc-macro2 v1.0.56 (*)
│               ├── quote v1.0.26 (*)
│               └── unicode-ident v1.0.8
├── cfg-if v1.0.0
├── derivative v2.2.0 (proc-macro)
│   ├── proc-macro2 v1.0.56 (*)
│   ├── quote v1.0.26 (*)
│   └── syn v1.0.109
│       ├── proc-macro2 v1.0.56 (*)
│       ├── quote v1.0.26 (*)
│       └── unicode-ident v1.0.8
├── indexmap v1.9.3
│   ├── hashbrown v0.12.3
│   │   └── ahash v0.7.6
│   │       ├── getrandom v0.2.9
│   │       │   ├── cfg-if v1.0.0
│   │       │   └── libc v0.2.141
│   │       ├── once_cell v1.17.1
│   │       └── serde v1.0.160 (*)
│   │       [build-dependencies]
│   │       └── version_check v0.9.4
│   ├── rayon v1.7.0
│   │   ├── either v1.8.1
│   │   └── rayon-core v1.11.0
│   │       ├── crossbeam-channel v0.5.8
│   │       │   ├── cfg-if v1.0.0
│   │       │   └── crossbeam-utils v0.8.15
│   │       │       └── cfg-if v1.0.0
│   │       ├── crossbeam-deque v0.8.3
│   │       │   ├── cfg-if v1.0.0
│   │       │   ├── crossbeam-epoch v0.9.14
│   │       │   │   ├── cfg-if v1.0.0
│   │       │   │   ├── crossbeam-utils v0.8.15 (*)
│   │       │   │   ├── memoffset v0.8.0
│   │       │   │   │   [build-dependencies]
│   │       │   │   │   └── autocfg v1.1.0
[...]

PR #4011, opened 2023-06-20
Move Artifact Register FrameInfo

ptitSeb:
Move Artifact Register FrameInfo to a more fitting place. Also register_frameinfo needs a `&mut self` now to reflect it change a state (even if it's a global static)

### DIFF
--- a/lib/compiler/src/engine/artifact.rs
+++ b/lib/compiler/src/engine/artifact.rs
@@ -301,7 +301,7 @@ impl Artifact {
             finished_dynamic_function_trampolines.into_boxed_slice();
         let signatures = signatures.into_boxed_slice();
 
-        Ok(Self {
+        let mut artifact = Self {
             id: Default::default(),
             artifact,
             allocated: Some(AllocatedArtifact {
@@ -312,7 +312,11 @@ impl Artifact {
                 frame_info_registration: Some(Mutex::new(None)),
                 finished_function_lengths,
             }),
-        })
+        };
+
+        artifact.register_frame_info();
+
+        Ok(artifact)
     }
 
     /// Check if the provided bytes look like a serialized `ArtifactBuild`.
@@ -379,7 +383,7 @@ impl Artifact {
     /// Register thie `Artifact` stack frame information into the global scope.
     ///
     /// This is required to ensure that any traps can be properly symbolicated.
-    pub fn register_frame_info(&self) {
+    pub fn register_frame_info(&mut self) {
         if let Some(frame_info_registration) = self
             .allocated
             .as_ref()
@@ -530,8 +534,6 @@ impl Artifact {
             .create_globals(context, &module)
             .map_err(InstantiationError::Link)?
             .into_boxed_slice();
-
-        self.register_frame_info();
 
         let handle = VMInstance::new(
             allocator,


lib/compiler/src/engine/artifact.rs
```diff
@@ -38,12 +37,13 @@ use wasmer_vm::{FunctionBodyPtr, MemoryStyle, TableStyle, VMSharedSignatureIndex
 use wasmer_vm::{InstanceAllocator, StoreObjects, TrapHandlerFn, VMConfig, VMExtern, VMInstance};
 
 pub struct AllocatedArtifact {
+    frame_info_registered: bool,
+    // frame_info_registered is not staying there but transfered to InnerEnginge / CodeMemory
```
theduke, line 41: Can you clarify the comment a bit?

Not sure what it means.

lib/compiler/src/engine/artifact.rs
```diff
@@ -38,12 +37,13 @@ use wasmer_vm::{FunctionBodyPtr, MemoryStyle, TableStyle, VMSharedSignatureIndex
 use wasmer_vm::{InstanceAllocator, StoreObjects, TrapHandlerFn, VMConfig, VMExtern, VMInstance};
 
 pub struct AllocatedArtifact {
+    frame_info_registered: bool,
```
theduke, line 40: Isn't a bool flag redundant if the registration itself is `Option`?

theduke (commented):
I don't feel comfortable approving here, because I don't know the code and context.

Is @syrusakbary familiar with this?

Otherwise will dig in a bit more to understand it properly.

lib/compiler/src/engine/artifact.rs
ptitSeb: No, because the FrameRegisteration is moved, and the Option is None again. Yet, the registration has been done and must not be done again. This is like that to avoid chnaging the API.

lib/compiler/src/engine/artifact.rs
```diff
@@ -378,47 +392,81 @@ impl ArtifactCreate for Artifact {
...
+        self.allocated
+            .as_mut()
+            .expect("It must be allocated")
+            .frame_info_registered = true;
+    }
+
+    /// The GlobalFrameInfoRegistration needs to be transfered to EngineInner if
+    /// register_frame_info has been used.
+    #[deprecated(since = "4.0.0", note = "done automaticaly by Artifact::from_parts.")]
+    pub fn transfert_frame_info_registration(&mut self) -> Option<GlobalFrameInfoRegistration> {
```
syrusakbary, line 456: Do you mean `transfer_frame_info_registration` ? (there's an extra `t`)

syrusakbary: I think we can name it `take_frame_info_registration`

syrusakbary (commented):
Would it be possible to have a test verifying that the collision is fixed?

lib/compiler/src/engine/artifact.rs
ptitSeb: oh, spelling...

ptitSeb: fixed

webmaster128:
With this change I cannot reproduce the issue #3793 anymore. I ran the project's tests for more than 1000 times now. Looking good 🎉

syrusakbary (approved):

PR #4013, opened 2023-06-20
Added a `WasmerEnv` abstraction to the CLI

### LINKED ISSUE #3959: `wasmer login` does not respect the registry.url config
### Describe the bug

When running `wasmer login`, I always get the wapm.io settings page URL, even when I have changed the registry.

```sh
$ wasmer -vV; rustc -vV

wasmer 4.0.0-beta.1 (b45f9a5 2023-06-01)
binary: wasmer-cli
commit-hash: b45f9a51845ed3ea55b61ef8774f7410ac4904f9
commit-date: 2023-06-01
host: aarch64-apple-darwin
rustc 1.70.0 (90c541806 2023-05-31)
binary: rustc
commit-hash: 90c541806f23a127002de5b4038be731ba1458ca
commit-date: 2023-05-31
host: aarch64-apple-darwin
release: 1.70.0
LLVM version: 16.0.2
```

### Steps to reproduce

```sh
$ wasmer config get registry.url
https://registry.wasmer.wtf/graphql

$ wasmer login
Please paste the login token from https://wapm.io/settings/access-tokens:
```

### Expected behavior
I expect the URL in wasmer-login to be `https://wasmer.wtf/settings/access-tokens`

### Actual behavior

I get the URL to wapm.io.

I think https://github.com/wasmerio/wasmer/blob/master/lib/cli/src/commands/login.rs#LL10C1-L10C45 this could be the reason for this.

### LINKED ISSUE #3993: Unable to Publish Package with `WASMER_TOKEN` env var
### Describe the bug

I encountered an issue while trying to publish a package to the Wasmer Registry using the `wasmer publish` command from the CI.

```sh
wasmer 4.0.0-beta.2 (18ab43b 2023-06-09)
binary: wasmer-cli
commit-hash: 18ab43bc60d4427676fb2adcee78ef016321b829
commit-date: 2023-06-09
host: aarch64-apple-darwin
rustc 1.70.0 (90c541806 2023-05-31)
binary: rustc
commit-hash: 90c541806f23a127002de5b4038be731ba1458ca
commit-date: 2023-05-31
host: aarch64-apple-darwin
release: 1.70.0
LLVM version: 16.0.2
```

### Steps to reproduce
1. Set the registry URL
```bash
$ wasmer config set registry.url wasmer.io
```
2. Export the Wasmer token
```bash
$ export WASMER_TOKEN="wap_..."
```
3. Publish the package
```bash
$ wasmer publish
error: cannot publish package: not logged into registry "https://registry.wasmer.io/graphql"
```
4. Set the registry URL again
```bash
$ wasmer config set registry.url wasmer.io
Successfully logged into registry "https://registry.wasmer.io/graphql" as user "..."

```
5. Retry publishing the package
```bash
$ wasmer publish
error: cannot publish package: not logged into registry "https://registry.wasmer.io/graphql"
```

### Expected behavior

I expected the package to publish both times, but it didn't.

### Actual behavior

I got an error that I am not logged in to the registry, even though I am.

### Additional context
I noticed that `WASMER_TOKEN` env var does not get added to the global wasmer.toml config file. This could be why I am getting this error.

Michael-F-Bryan:
This adds a `WasmerEnv` struct which abstracts over `$WASMER_DIR` and config used to communicate with the registry (i.e. graphql endpoint and tokens). 

Now, whenever you run a command that interacts with the registry in some way, you should always be able to use the `$WASMER_DIR`, `$WASMER_REGISTRY`, and `$WASMER_TOKEN` environment variables or their associated command-line arguments.

This makes the handling of things like `$WASMER_DIR`, `--token`, and `--registry` consistent across all our subcommands.

Fixes #3993, fixes #3959.

### DIFF
--- a/lib/cli/src/commands/login.rs
+++ b/lib/cli/src/commands/login.rs
@@ -1,86 +1,122 @@
+use std::path::PathBuf;
+
 use clap::Parser;
 #[cfg(not(test))]
 use dialoguer::Input;
-use wasmer_registry::WasmerConfig;
+
+use crate::{Registry, WasmerDir};
 
 /// Subcommand for listing packages
 #[derive(Debug, Clone, Parser)]
 pub struct Login {
-    /// Registry to log into (default: wasmer.io)
-    #[clap(long, default_value = "wasmer.io")]
-    pub registry: String,
-    /// Login token
-    #[clap(name = "TOKEN")]
+    /// Set Wasmer's home directory
+    #[clap(long, env = "WASMER_DIR", default_value = crate::WASMER_DIR.as_os_str())]
+    pub wasmer_dir: PathBuf,
+    /// The registry to fetch packages from (inferred from the environment by
+    /// default)
+    #[clap(long, env = "WASMER_REGISTRY")]
+    pub registry: Option<Registry>,
+    /// The API token to use when communicating with the registry (inferred from
+    /// the environment by default)
     pub token: Option<String>,
 }
 
 impl Login {
-    fn get_token_or_ask_user(&self) -> Result<String, std::io::Error> {
-        match self.token.as_ref() {
-            Some(s) => Ok(s.clone()),
-            None => {
-                let registry_host = wasmer_registry::format_graphql(&self.registry);
-                let registry_tld = tldextract::TldExtractor::new(tldextract::TldOption::default())
-                    .extract(&registry_host)
-                    .map_err(|e| {
-                        std::io::Error::new(
-                            std::io::ErrorKind::Other,
-                            format!("Invalid registry for login {}: {e}", self.registry),
-                        )
-                    })?;
-                let login_prompt = match (
-                    registry_tld.domain.as_deref(),
-                    registry_tld.suffix.as_deref(),
-                ) {
-                    (Some(d), Some(s)) => {
-                        format!("Please paste the login token from https://{d}.{s}/settings/access-tokens")
-                    }
-                    _ => "Please paste the login token".to_string(),
-                };
-                #[cfg(test)]
-                {
-                    Ok(login_prompt)
-                }
-                #[cfg(not(test))]
-                {
-                    Input::new().with_prompt(&login_prompt).interact_text()
-                }
-            }
+    fn get_token_or_ask_user(&self, wasmer_dir: &WasmerDir) -> Result<String, anyhow::Error> {
+        if let Some(token) = &self.token {
+            return Ok(token.clone());
         }
+
+        let registry_host = wasmer_dir.registry_endpoint()?;
+        let registry_tld = tldextract::TldExtractor::new(tldextract::TldOption::default())
+            .extract(registry_host.as_str())
+            .map_err(|e| {
+                std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    format!("Invalid registry for login {}: {e}", registry_host),
+                )
+            })?;
+        let login_prompt = match (
+            registry_tld.domain.as_deref(),
+            registry_tld.suffix.as_deref(),
+        ) {
+            (Some(d), Some(s)) => {
+                format!("Please paste the login token from https://{d}.{s}/settings/access-tokens")
+            }
+            _ => "Please paste the login token".to_string(),
+        };
+        #[cfg(test)]
+        {
+            Ok(login_prompt)
+        }
+        #[cfg(not(test))]
+        {
+            let token = Input::new().with_prompt(&login_prompt).interact_text()?;
+            Ok(token)
+        }
+    }
+
+    fn wasmer_dir(&self) -> WasmerDir {
+        WasmerDir::new(
+            self.wasmer_dir.clone(),
+            self.registry.clone(),
+            self.token.clone(),
+        )
     }
 
     /// execute [List]
     pub fn execute(&self) -> Result<(), anyhow::Error> {
-        let token = self.get_token_or_ask_user()?;
-        let wasmer_dir =
-            WasmerConfig::get_wasmer_dir().map_err(|e| anyhow::anyhow!("no wasmer dir: {e}"))?;
-        match wasmer_registry::login::login_and_save_token(&wasmer_dir, &self.registry, &token)? {
+        let wasmer_dir = self.wasmer_dir();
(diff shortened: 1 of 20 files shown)


syrusakbary:
This PR made me super confused.

I think this PR is more complicated than it should. It also ties the WASMER_TOKEN into a Wasmer dir? I don't think this make much sense, what am I missing?

lib/cli/src/commands/login.rs
```diff
@@ -1,86 +1,122 @@
...
 
 /// Subcommand for listing packages
 #[derive(Debug, Clone, Parser)]
 pub struct Login {
-    /// Registry to log into (default: wasmer.io)
-    #[clap(long, default_value = "wasmer.io")]
-    pub registry: String,
-    /// Login token
-    #[clap(name = "TOKEN")]
+    /// Set Wasmer's home directory
```
theduke, line 12: Why does this not also use `WasmerEnv`?

I get that the particular can be different, but that should be unifiable with some focused methods on WasmerEnv?

lib/cli/src/wasmer_env.rs
```diff
@@ -0,0 +1,127 @@
...
+use once_cell::sync::Lazy;
+use url::Url;
+use wasmer_registry::WasmerConfig;
+
+/// Command-line flags for determining the local "Wasmer Environment".
+///
+/// This is where you access `$WASMER_DIR`, the `$WASMER_DIR/wasmer.toml` config
+/// file, and specify the current registry.
+#[derive(Debug, Clone, PartialEq, clap::Parser)]
+pub struct WasmerEnv {
```
theduke, line 13: I'd love to unify the Edge CLI ( which currently uses a very similar approach, but with different code).

Can we put this into the `registry` crate directly, maybe under a feature flag to hard required  require `clap` for compiling `registry`?

I realize that's a bit awkward, but putting it into `registry`, or another helper crate is the only way to allow the Edge CLI to use it, since the main CLI depends on the Edge CLI.

lib/cli/src/wasmer_env.rs
```diff
@@ -0,0 +1,127 @@
...
+    #[clap(long, env = "WASMER_REGISTRY")]
+    registry: Option<Registry>,
+    /// The API token to use when communicating with the registry (inferred from
+    /// the environment by default)
+    #[clap(long, env = "WASMER_TOKEN")]
+    token: Option<String>,
+}
+
+impl WasmerEnv {
+    pub(crate) fn new(
```
theduke, line 28: nit: what's the point of a constructor that just takes all three fields and puts them in the struct unmodified?

lib/cli/src/wasmer_env.rs
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+}
+
+/// The default value for `$WASMER_DIR`.
+pub static WASMER_DIR: Lazy<PathBuf> =
+    Lazy::new(|| match wasmer_registry::WasmerConfig::get_wasmer_dir() {
+        Ok(path) => path,
+        Err(e) => {
+            if let Some(install_prefix) = std::env::var_os("WASMER_INSTALL_PREFIX") {
```
theduke, line 92: `WASMER_INSTALL_PREFIX` was previously primarily  provided at compile time through the build scripts, not at runtime, right?

In the old code it is used through `option_env!()`, which captures the compile time specified value.

This seems to break that behaviour.

lib/cli/src/wasmer_env.rs
```diff
@@ -0,0 +1,127 @@
...
+                PathBuf::from(install_prefix)
+            } else {
+                panic!("Unable to determine the wasmer dir: {e}");
+            }
+        }
+    });
+
+/// A registry as specified by the user.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Registry(String);
```
theduke, line 102: Can we make this a `url::Url`?

theduke (commented):
Love it.

The Edge CLI followed a very similar approach.

Left a few comments/discussion points.

lib/cli/src/wasmer_env.rs
Michael-F-Bryan: We want users to be able to do `--registry wasmer.wtf` and have `registry.graphql_endpoint()` return `https://registry.wasmer.wtf/graphql`, but `wasmer.wtf` isn't a valid URL.

We could write the `From<String>` impl to automatically call `wasmer_registry::format_graphql()` on whatever string users pass in. However, it'd mean the help text and other printed messages would show `https://registry.wasmer.wtf/graphql` even if the user specified `--registry wasmer.wtf` and that's not great for UX. 

Making `Registry` a newtype wrapper around a `String` is a bit of a compromise that lets us capture the user's original input while still getting the coerced GraphQL endpoint automatically.

lib/cli/src/wasmer_env.rs
Michael-F-Bryan: The `Login` command needs to construct a `WasmerEnv` manually because its command-line flags aren't compatible with the `#[flatten] wasmer_env: WasmerEnv` approach.

I'd prefer to not make `WasmerEnv`'s fields public just so `Login` can construct a `WasmerEnv` directly. We want other sub-commands to go through`WasmerEnv`'s accessor methods so we always fall back to values from the `~/.wasmer/wasmer.toml` config.

lib/cli/src/wasmer_env.rs
Michael-F-Bryan: Ah, I didn't notice it was using `option_env!()` to embed the prefix at compile-time. I'll fix that up.

lib/cli/src/wasmer_env.rs
Michael-F-Bryan: The `wasmer-registry` crate is already a bit of a dumping ground. I'm sure nobody will notice if we add yet another chunky dependency to it 😅

lib/cli/src/wasmer_env.rs
theduke: Yeah, putting it into the `FromStr` impl would have been my idea, and using the magic conversion function. 

The struct could also hold both the original value as string, and the derived value as a url, but that's also not necessarily ideal.

Fine for me as is.

ptitSeb (approved):

PR #4018, opened 2023-06-21
Canonicalize the host path passed to `--mapdir`

### LINKED ISSUE #4010: `--mapdir` seems to needs an absolute path for the host folder
While working on a PHP port, I had issue with the `--mapdir` argument.

Using command like
```bash
~/git/wasmer/target/release/wasmer run sapi/cli/php.wasm --mapdir /blah:./blah -- /blah/my.php
```
ended with an file not found error
```
PHP_MINIT_FUNCTION(spl)
Could not open input file: /blah/my.php
```
but using
```bash
~/git/wasmer/target/release/wasmer run sapi/cli/php.wasm --mapdir /blah:$(pwd)/blah -- /blah/my.php
```
Worked.

Michael-F-Bryan:
This PR patches the logic for `--mapdir` so we always use absolute paths when mounting directories.

Ideally we wouldn't need to do this canonicalizing because the code for mounting directories should be able to handle relative paths, but this quickfix should resolve things until we do our big filesystem layer overhaul.

Fixes #4010.

### DIFF
--- a/lib/wasix/src/runners/mod.rs
+++ b/lib/wasix/src/runners/mod.rs
@@ -23,7 +23,10 @@ use crate::runtime::{
 /// instance (the "guest").
 #[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct MappedDirectory {
+    /// The absolute path for a directory on the host filesystem.
     pub host: std::path::PathBuf,
+    /// The absolute path specifying where the host directory should be mounted
+    /// inside the guest.
     pub guest: String,
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2810,9 +2810,9 @@ checksum = "0ab1bc2a289d34bd04a330323ac98a1b4bc82c9d9fcb1e66b63caa84da26b575"
 
 [[package]]
 name = "openssl"
-version = "0.10.54"
+version = "0.10.55"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


ptitSeb (approved):

lib/wasix/src/runners/mod.rs
```diff
@@ -23,7 +23,10 @@ use crate::runtime::{
 /// instance (the "guest").
 #[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct MappedDirectory {
+    /// The absolute path for a directory on the host filesystem.
```
theduke, line 26: Should we have a constructor `MappedDirectory::new()` that ensures the struct only holds absolute paths? And make the fields private?

(not required for merge)

theduke (approved):
I thought some of my additions already added always canonicalizing to an absolute path , at least in the WASI runner.

But it makes sense to move this a bit higher up.

As mentioned previously, I don't think it makes any sense to have relative paths here apart from the CLI argument phase.

lib/wasix/src/runners/mod.rs
Michael-F-Bryan: Yeah, I was thinking about that as well. 

It'd be nice to enforce correctness through a constructor, but at the same time it's really nice to be able to destructure with `let MappedDir { host, guest } = ...` instead of needing to write `mapped_dir.host` and `mapped_dir.guest` all over the place.

PR #4019, opened 2023-06-21
Make sure `ExecutableTarget::from_file()` for `*.wasm` files checks the module cache

### LINKED ISSUE #4016: Running simple wasm file doesn't use Compiler Cache
The compilation cache seems to not be used when running simple .wasm file.
It's ok for package.

After a bisect, the regression cames from one of the following commit
```
23f27396004e824acf5eb734cca9d833b327ffc0
eed8ac542355a949f51aa4206fda2ce176f87749
18620acbb9d8140e16c2b1a1b0c4d87bdfcb6afb
2c161b66e93f469e9b5cf620761cb5ce2cfbfe9c
d9df1c1f1c607d47d18c20e343e1317f3867af2e
464bcd4c9ed9c8b258c5e4252cb6e522d6e121de
2b4e3be245b5b6bbf8c60f4d2438ebc70a4a22eb
7c4834f74f180fd33a7bb4522782d1d539c03bf4
1e543310e51078559f45c190941756f75c521720
```

For testing, remove your cache folder `~/.wasmer/cache` and run a simple `.wasm` file: the cache folder stays empty.

Michael-F-Bryan:
@ptitSeb noticed that running `wasmer run some-file.wasm` will recompile the `*.wasm` file every time. Apparently we forgot to check the module cache when loading a `*.wasm` target.

Fixes #4016.

### DIFF
--- a/lib/cli/src/commands/run.rs
+++ b/lib/cli/src/commands/run.rs
@@ -34,6 +34,7 @@ use wasmer_wasix::{
     bin_factory::BinaryPackage,
     runners::{MappedDirectory, Runner},
     runtime::{
+        module_cache::{CacheError, ModuleHash},
         package_loader::PackageLoader,
         resolver::{PackageSpecifier, QueryError},
     },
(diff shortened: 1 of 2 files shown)


lib/cli/src/commands/run.rs
```diff
@@ -575,7 +576,28 @@ impl ExecutableTarget {
                 let wasm = std::fs::read(path)?;
                 let engine = runtime.engine().context("No engine available")?;
                 pb.set_message("Compiling to WebAssembly");
-                let module = Module::new(&engine, wasm)?;
+
+                let tasks = runtime.task_manager();
+                let module_cache = runtime.module_cache();
+                let module_hash = ModuleHash::sha256(&wasm);
+
+                let module = match tasks.block_on(module_cache.load(module_hash, &engine)) {
```
theduke, line 584: Maybe leave a TODO/followup issue that this should switch to `Runtime::load_module` once #3996 is merged?

theduke (approved):

ptitSeb (approved):

lib/cli/src/commands/run.rs
Michael-F-Bryan: I've added a comment on #3996 so hopefully that PR will be able to do the fix immediately.

Issue #4015, opened 2023-06-20
The `wasmer-integration-tests-cli` test suite isn't hygienic


Michael-F-Bryan:
### Motivation

Almost all of the tests in the `wasmer-integration-tests-cli` test suite are non-hygienic. They all use the same `~/.wasmer/` directory, which means that

1. Integration tests need to be run sequentially (i.e. `--test-threads=1`) and a failure midway through can sometimes mess things up for subsequent tests (I've noticed caching is particularly bad for this)
2. After the integration tests are done, the global `~/.wasmer/wasmer.toml` file will be left in a broken state and you'll need to find your old access token and log in again

### Proposed solution

The way we solve this problem is by making sure each integration test uses its own independent wasmer directory. We can do this fairly trivially by setting the `$WASMER_DIR` environment variable to something unique like [`$CARGO_TARGET_TMPDIR/wasmer-integration-tests-cli/$test_name`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates) 

Cargo always sets provides integration tests with a `$CARGO_TARGET_TMPDIR` directory inside `target/` they can use as a scratch space. That also means CI will get caching for free.

Issue #3568, opened 2023-02-02
Split create-exe Integration Tests into Separate Test Crate


theduke:
### Motivation

Dumping the create-exe tests into the integration test CLI was not a great idea.

Running the tests locally now requires also building the whole C api, which slows down iteration speed a lot, and also isn't at all necessary for running the regular WASI or WASIX snapshot tests.

### Proposed solution

Move create-exe tests to a separate test crate that can be invoked separately.

PR #4020, opened 2023-06-21
Make CLI integration tests automatically run the most up-to-date version of the CLI

### LINKED ISSUE #4014: Make the `wasmer-integration-tests-cli` crate run the `wasmer` CLI deterministically
### Motivation

Currently, the integration test suite for our CLI invokes the `wasmer` CLI by blindly running whichever `wasmer` executable is in `target/release/` without checking that it was recompiled after any recent changes and was compiled with the correct feature flags.

That means developers need to remember to rebuild the CLI (in release mode!) before every test run, which is pretty error-prone. One strategy the integration tests for `wasmer run-unstable` took (called `wasmer run` as of #3801 ) is to do invoke the CLI via `cargo run`, but that often means tests take multiple seconds as Cargo does its thing.

It would be much better if the `wasmer-integration-tests-cli` crate could automatically make sure it always ran an up-to-date version of the `wasmer` CLI with a fixed set of feature flags.

### Proposed solution

One solution is to add the `wasmer-cli` as a dependency of `wasmer-integration-tests-cli` so we can make sure the crate is available and compiled with the right flags. The `wasmer-integration-tests-cli` would then grow a `src/bin/wasmer-cli-shim.rs` executable which we call using [`$CARGO_BIN_EXE_ wasmer-cli-shim`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates).

Michael-F-Bryan:
Something that always annoys me is that the CLI integration tests require you to remember to recompile the `wasmer` CLI in release mode every time you want to run the test suite. This tends to be quite a) error-prone and b) slow.

I'm introducing a `wasmer-cli-shim` executable so we can use the `$CARGO_BIN_EXE_<name>` environment variable `cargo` sets when compiling integration tests. This will make sure we always run an up-to-date CLI executable because `wasmer-cli-shim` will be recompiled whenever `wasmer-cli` changes.

As a happy side-effect, this means the old `run_unstable` tests no longer need to do their defensive `cargo run --package=wasmer-cli` and can run the `wasmer-cli-shim` executable directly.

Fixes #4014.

### DIFF
--- a/tests/integration/cli/tests/config.rs
+++ b/tests/integration/cli/tests/config.rs
@@ -1,6 +1,8 @@
 use std::path::{Path, PathBuf};
 use std::process::Command;
-use wasmer_integration_tests_cli::{get_repo_root_path, get_wasmer_path};
+use wasmer_integration_tests_cli::get_repo_root_path;
+
+const WASMER_EXE: &str = env!("CARGO_BIN_EXE_wasmer-cli-shim");
 
 fn get_wasmer_dir() -> Result<PathBuf, anyhow::Error> {
     if let Ok(s) = std::env::var("WASMER_DIR") {
@@ -33,7 +35,7 @@ fn wasmer_config_multiget() -> anyhow::Result<()> {
     let bin = format!("{}", bin_path.display());
     let include = format!("-I{}", include_path.display());
 
-    let output = Command::new(get_wasmer_path())
+    let output = Command::new(WASMER_EXE)
         .arg("config")
         .arg("--bindir")
         .arg("--cflags")
@@ -53,7 +55,7 @@ fn wasmer_config_multiget() -> anyhow::Result<()> {
 
 #[test]
 fn wasmer_config_error() -> anyhow::Result<()> {
-    let output = Command::new(get_wasmer_path())
+    let output = Command::new(WASMER_EXE)
         .arg("config")
         .arg("--bindir")
         .arg("--cflags")
@@ -65,9 +67,9 @@ fn wasmer_config_error() -> anyhow::Result<()> {
         .map(|s| s.trim().to_string())
         .collect::<Vec<_>>();
     #[cfg(not(windows))]
-    let expected_1 = "Usage: wasmer config --bindir --cflags";
+    let expected_1 = "Usage: wasmer-cli-shim config --bindir --cflags";
     #[cfg(windows)]
-    let expected_1 = "Usage: wasmer.exe config --bindir --cflags";
+    let expected_1 = "Usage: wasmer-cli-shim.exe config --bindir --cflags";
 
     let expected = vec![
         "error: the argument '--bindir' cannot be used with '--pkg-config'",
@@ -83,7 +85,7 @@ fn wasmer_config_error() -> anyhow::Result<()> {
 }
 #[test]
 fn config_works() -> anyhow::Result<()> {
-    let bindir = Command::new(get_wasmer_path())
+    let bindir = Command::new(WASMER_EXE)
         .arg("config")
         .arg("--bindir")
         .output()?;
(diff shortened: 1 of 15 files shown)


tests/integration/cli/tests/config.rs
```diff
@@ -65,9 +67,9 @@ fn wasmer_config_error() -> anyhow::Result<()> {
         .map(|s| s.trim().to_string())
         .collect::<Vec<_>>();
     #[cfg(not(windows))]
-    let expected_1 = "Usage: wasmer config --bindir --cflags";
+    let expected_1 = "Usage: wasmer-cli-shim config --bindir --cflags";
```
theduke, line 70: Seeing different command names opposed to the  the proper `wasmer` command would be somewhat confusing, wouldn't it?
Needs to be at least documented, maybe in the README.

theduke (commented):
I get the motivation, and it seems reasonable. 

However, this does seem to break a bunch of assumptions and workflows.

* this makes us build the test CLI in a different manner than the  "real CLI", because that is handled in the makefile, which could lead to drift in feature flags or various other compilation-affecting env vars/flags, which cause the integration to not exactly test the proper CLI

* Related to the above: integratin-cli specific dependencies  could lead to version drift between the shim and the full CLI, which could affect test results due to slightly changed behaviour. Image for example the integration cli using a higher but compatible version of a dependencies, which doesn't get unified for some reason or another. Not a problem if everything was a  workspace dependency, and probably a somewhat theoretical concern, because the CLI doesn't usually use exact versions, but it could still happen 

* are there no tests exercising llvm at the moment? the feature is not enabled

* As far as I understand, the full build currently also handles c api related building and packaing, which is required to succeed for all the build-exe related tests to pass, this hides that fact, and doesn't obviate the need for running `make` in order to get the tests to run properly 
  (note: I already suggested breaking out the build-exe tests into a separate step in the past, for somewhat related reasons)

* Prevents from running the API against a specific existing binary. no idea if that is useful 

I'm tentatively in favour despite this, but I think we'll need to hear what @ptitSeb  and @syrusakbary think, and if they are aware of any gotchas.

A lot of this can be mitigated by keeping the WASMER_PATH env var as an optional override, and only use the shim if it is not set.

theduke:
Addendum: I think the CLI should keep using the `make build-wasmer` built binary.

This can be used to make local iteration easier regardless.

Michael-F-Bryan:
Thanks for the input, @theduke! 

In retrospect, I'm thinking it might not be a good idea to proceed with this PR. 

The overall `make test-integration-cli` workflow is broken outside of CI for all intents and purposes anyway[^1], but adding yet another way to build the CLI won't help the situation. This approach also isn't a perfect solution so we'd probably exchange one set of footguns for a different one.

For now, I might just keep running my `cargo watch` command which rebuilds and sets the correct feature flags.

> this makes us build the test CLI in a different manner than the "real CLI", because that is handled in the makefile, which could lead to drift in feature flags or various other compilation-affecting env vars/flags, which cause the integration to not exactly test the proper CLI

Yeah, I guess we need to take a step back and ask what the integration tests are trying to achieve in the first place. 

Are they testing our compiler implementations, our WASIX implementation, CLI sub-command, or all of the above? 

> are there no tests exercising llvm at the moment? the feature is not enabled

That's correct. We only run the integration test suite once, so CI only does integration tests against one compiler (I'm guessing cranelift?).

> As far as I understand, the full build currently also handles c api related building and packaing, which is required to succeed for all the build-exe related tests to pass

Good point. In the long term, I'd prefer to split the `create-exe` tests out into a separate crate and build job (#3568).

> Prevents from running the API against a specific existing binary. no idea if that is useful

I don't think I've ever had a concrete situation where I wanted this. It might be useful for a hypothetical WASIX spec test suite, but when you are wanting to test the CLI you'll always want to re-compile so you know what code is being tested.

[^1]: 
    - Using `--nocapture` means you get bombarded with so much output (I just did a run and it generated 40760 lines of output) that it's unusable
    - the Makefile and our feature flags like to recompile the world (even with `sccache`)
    - the code for setting up the C API and running `create-exe` tests has never worked on any of my Windows/Linux/MacOS machine - you always get `#include "wasmer.h"` errors
    - we use `--test-threads=1` because of #4015 so the tests themselves are really slow
    - I could go on for hours...

PR #4024, opened 2023-06-22
Save compiled `*.wasm` files to the module cache when they are loaded directly

Michael-F-Bryan:
#4019 made sure we check the module cache when doing a `wasmer run whatever.wasm` but forgot to save the compiled module after a cache miss 😅

### DIFF
--- a/lib/cli/src/commands/run.rs
+++ b/lib/cli/src/commands/run.rs
@@ -585,7 +585,8 @@ impl ExecutableTarget {
                             );
                         }
 
-                        let module = Module::new(&engine, &wasm)
+                        let module = tracing::debug_span!("compiling_wasm")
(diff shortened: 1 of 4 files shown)


lib/cli/src/commands/run.rs
```diff
@@ -585,8 +585,13 @@ impl ExecutableTarget {
                             );
                         }
 
-                        Module::new(&engine, &wasm)
-                            .with_context(|| format!("Unable to compile \"{}\"", path.display()))?
+                        let module = tracing::debug_span!("compiling_wasm")
```
ptitSeb, line 588: isn't there another way that having an object being built from a `tracing::debug_span!` command? That feel odd.

ptitSeb (approved):

PR #4031, opened 2023-06-26
Speed up the module cache 6x by removing LZW compression

Michael-F-Bryan:
# Description

While troubleshooting slow start times for `wasmer run`, we noticed that using LZW compression when caching compiled modules to the filesystem (`wasmer_wasix::runtime::module_cache::FileSystemCache`) incurs about a very noticeable 6x performance penalty.

For perspective, compiling php.wasm with cranelift took about 4.75 seconds on a M1 Mac.

Without LZW compression:
- `FileSystemCache::save()`: 408ms, 142MB binary
- `FileSystemCache::load()`: 155ms

With LZW compression:
- `FileSystemCache::save()`: 2.4s, 72MB binary
- `FileSystemCache::load()`: 822ms

Check out the conversation [on Slack](https://wasmerio.slack.com/archives/C04193Z3RHS/p1687445504681669) for more context.

### DIFF
--- a/lib/wasix/src/runtime/module_cache/filesystem.rs
+++ b/lib/wasix/src/runtime/module_cache/filesystem.rs
@@ -1,6 +1,5 @@
 use std::{
-    fs::File,
-    io::{BufReader, BufWriter, Read, Seek, Write},
+    io::{BufWriter, Write},
     path::{Path, PathBuf},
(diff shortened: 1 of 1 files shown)


ptitSeb (approved):
seems fine

PR #4032, opened 2023-06-26
Make the CLI respect the `--token` flag

### LINKED ISSUE #4028: Token/Registry Flag Handling Broken With --token + --registry flags
The recently introduced new token + registry flag handling in the CLI currently does not work when both 

`--token` and `--registry` are supplied as flags.

It seems like it will always try to get the token from the config for the given registry, even if the `--token` flag is supplied, resulting in a "not logged in" error.

@Michael-F-Bryan

Michael-F-Bryan:
The `WasmerEnv::token()` getter introduced in #4013 never checks the `--token` flag, so we were always using the token from a user's global config file.

Fixes #4028

### DIFF
--- a/lib/registry/src/wasmer_env.rs
+++ b/lib/registry/src/wasmer_env.rs
@@ -88,8 +88,10 @@ impl WasmerEnv {
         // Fall back to the config file
 
         let config = self.config().ok()?;
-        let login = config.registry.current_login()?;
-        Some(login.token.clone())
+        let registry_endpoint = self.registry_endpoint().ok()?;
+        config
+            .registry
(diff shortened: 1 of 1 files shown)


lib/registry/src/wasmer_env.rs
```diff
@@ -81,6 +81,12 @@ impl WasmerEnv {
 
     /// The API token for the active registry.
     pub fn token(&self) -> Option<String> {
+        if let Some(token) = &self.token {
```
theduke, line 84: Can we have a regression test for this?

Michael-F-Bryan: Done.

theduke (approved):

PR #4036, opened 2023-06-26
Wasmer.sh browser tests

Michael-F-Bryan:


Michael-F-Bryan:
I'm going to abandon this PR.

Cypress won't work with wasmer.sh because it embeds your website in its runner using an `<iframe>`, and we couldn't find a way to set the iframe's security settings to allow wasmer.sh to use `SharedArrayBuffer`.

Issue #4027, opened 2023-06-23
The `--mapdir` command seems ignored when running a package


ptitSeb:
For this test, I created a folder `blah` in my current folder, and put a simple `index.php` in there with the following content:

```php
<?  phpinfo();
```

Try to run this file using a packaged version of `php` built for `wasm32-wasix`, using:

```bash
wasmer run php/php --net --mapdir /blah:$(pwd)/blah --  -S localhost:8001 -t /blah
```

The server will not start with the error:

```
Directory /blah does not exist.
```

While the same command using a local naked wasm file works.

Michael-F-Bryan:
As mentioned [on Slack](https://wasmerio.slack.com/archives/C02CK7101PG/p1688960922797099), I think this may be something do to with the `php.wasm` file rather than runners or the filesystem.

I'm reproducing everything on the current `master`, 07198b9e7d952c3135ff315bc16900647d3bab95, with the following local files:

```console
$ tree /tmp/app
/tmp/app
└── index.php

1 directory, 1 file

$ cat /tmp/app/index.php
<?  phpinfo();
```

When I ran `php/php@7.4.31` using `wasmer run`, I saw the same `Directory "/" does not exist` error.

```console
$ RUST_BACKTRACE=1 cargo run --package=wasmer-cli --features=compiler,singlepass,cranelift --release -- \
        run php/php@7 --net --cranelift --registry=wasmer.io --mapdir=/app:/tmp/app -- \
        -S localhost:8000 -t /app

Directory /app does not exist.
```

However, using the same `--mapdir` command and the `python/python` package to inspect the filesystem shows me that `/app/` does exist and you can freely access `/app/index.php`.

```console
$ RUST_BACKTRACE=1 cargo run --package=wasmer-cli --features=compiler,singlepass,cranelift --release -- \
        run python/python --net --cranelift --registry=wasmer.io --mapdir=/app:/tmp/app

Python 3.6.7 (default, Feb 14 2020, 03:17:48)
[Wasm WASI vClang 9.0.0 (https://github.com/llvm/llvm-project 0399d5a9682b3cef7 on generic
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.listdir("/")
['.app', '.private', 'app', 'bin', 'dev', 'etc', 'lib', 'tmp']
>>> os.listdir("/app/")
['index.php']
>>> open("/app/index.php").read()
'<?  phpinfo();\n'
```

[...]

PR #4038, opened 2023-06-27
Clean up the integration tests

### LINKED ISSUE #3698: Segfault in `wasmer run-unstable`'s integration tests on musl linux
### Describe the bug

The integration tests for `wasmer run-unstable` (#3650) are currently segfaulting... but only when run in CI and building for `x86_64-unknown-linux-musl`.

This is probably related to WASIX or our integration testing and not the `wasmer run-unstable` PR.

### Steps to reproduce

Push a commit to GitHub and watch the musl linux integration tests fail ([example](https://github.com/wasmerio/wasmer/actions/runs/4471026419/jobs/7855861497)).

You should see the following:

```
...
error: test failed, to rerun pass '-p wasmer-integration-tests-cli --test run_unstable'

Caused by:
  process didn't exit successfully: `/__w/wasmer/wasmer/target/x86_64-unknown-linux-musl/debug/deps/run_unstable-2b992cbc0342feb2 --nocapture --test-threads=1` (signal: 11, SIGSEGV: invalid memory reference)
```

### Additional context

To make sure #3650 is mergeable, we've disabled the tests in xxxx.

### LINKED ISSUE #4015: The `wasmer-integration-tests-cli` test suite isn't hygienic

Michael-F-Bryan:
This PR does a couple things to make our CLI integration test story a bit nicer:

- The run-unstable integration tests have been merged into `run.rs` because `wasmer run-unstable` no longer exists
- We now use the `--registry` flag from #4013 instead of running `wasmer login` and clobbering the tokens in the user's `$WASMER_DIR/wasmer.toml` (Fixes #4015)
- The run-unstable tests now run the `wasmer` executable directly instead of making a defensive `cargo run --features=...` 
  - The `cargo run` doesn't make sense any more because #4020 decided that we'll rely on `make integration-test-cli` or the user to provide an up-to-date `wasmer` executable
- Some create-exe tests were moved from `run.rs` into `create_exe.rs` where they belong 
- Moved the paths for all files used during testing into a single `fixtures` module
- Remove the `--test-threads=1` from `make test-integration-cli` so we use all cores when running integration tests
- Remove the `--no-capture` from `make test-integration-cli` to reduce the amount of spam you are bombarded with when running tests

After these changes, the *"Test Integration CLI"* CI step is reduced from taking about 30 minutes to 3 minutes.

It also looks like the MUSL segfaults are gone (i.e. fixes #3698).


tests/integration/cli/tests/run.rs
```diff
@@ -263,104 +274,117 @@ fn run_wasi_works_non_existent() -> anyhow::Result<()> {
...
         .arg("python/python")
-        .arg(format!("--mapdir=.:{}", asset_path().display()))
+        .arg(format!("--mapdir=/app:{}", asset_path().display()))
         .arg("--registry=wasmer.io")
-        .arg("test.py")
-        .env("WASMER_DIR", wasmer_dir.path())
-        .env("RUST_LOG", &rust_log)
+        .arg("/app/test.py")
+        .env("WASMER_CACHE_DIR", wasmer_dir.path())
+        .env("RUST_LOG", &*CACHE_RUST_LOG)
```
ptitSeb, line 289: This works? Why it doesn't when I try on php/php https://github.com/wasmerio/wasmer/issues/4027, is there something special to do on package itself?

tests/integration/cli/tests/run.rs
```diff
@@ -263,104 +274,117 @@ fn run_wasi_works_non_existent() -> anyhow::Result<()> {
...
 
-// FIXME: Re-enable. See https://github.com/wasmerio/wasmer/issues/3717
-#[ignore]
 #[test]
 fn run_test_caching_works_for_urls() {
+    let wasmer_dir = TempDir::new().unwrap();
+
     let assert = Command::new(get_wasmer_path())
+        .arg("run")
         .arg("https://wapm.io/python/python")
```
ptitSeb, line 356: still targetting "wapm.io"?

ptitSeb (approved):

tests/integration/cli/src/assets.rs
```diff
@@ -1,51 +1,70 @@
...
-    "/debug/"
-);
+pub fn wasmer_include_path() -> &'static Path {
+    Path::new(concat!(env!("CARGO_MANIFEST_DIR"), "/../../../lib/c-api/"))
+}
 
-/* env var TARGET is set by tests/integration/cli/build.rs on compile-time */
+pub fn wasmer_target_path() -> &'static Path {
+    let path = if cfg!(feature = "debug") {
+        concat!(env!("CARGO_MANIFEST_DIR"), "/../../../target/debug/")
```
ptitSeb, line 24: When building from the CLI, the path might be `target/x86_64-linux/release` or something similar as we define the target arch even when not cross-compiling

tests/integration/cli/tests/run.rs
Michael-F-Bryan: Ah good point, I'll go through and update everything to use either `wasmer.io` or `wasmer.wtf`.

tests/integration/cli/tests/run.rs
Michael-F-Bryan: I'm not entirely sure why it doesn't work. There are loads of tests that use `--mapdir` so directory mapping definitely works. You can even mount to `/`, but it'll break packages that bundle their own files (e.g. python and its standard library). 

Maybe it's an issue with PHP's filesystem access?

Makefile
```diff
@@ -622,12 +622,12 @@ test-wasi:
...
 	cp ./dist/wasmer.tar.gz ./link.tar.gz
 	rustup target add wasm32-wasi
-	WASMER_DIR=`pwd`/package $(CARGO_BINARY) test $(CARGO_TARGET_FLAG) --features webc_runner --no-fail-fast -p wasmer-integration-tests-cli -- --nocapture --test-threads=1
+	WASMER_DIR=`pwd`/package $(CARGO_BINARY) test $(CARGO_TARGET_FLAG) --features webc_runner --no-fail-fast -p wasmer-integration-tests-cli
 
 # Before running this in the CI, we need to set up link.tar.gz and /cache/wasmer-[target].tar.gz
 test-integration-cli-ci:
 	rustup target add wasm32-wasi
-	$(CARGO_BINARY) test $(CARGO_TARGET_FLAG) --features webc_runner -p wasmer-integration-tests-cli --  --test-threads=1
+	$(CARGO_BINARY) test $(CARGO_TARGET_FLAG) --features webc_runner -p wasmer-integration-tests-cli
```
syrusakbary, line 630: Nice!

Michael-F-Bryan:
I'm going to put this PR on hold until #4050 is merged. We're seeing a bunch of new runtime-in-runtime issues during the snapshot tests, which doesn't really make sense this PR hasn't touched any code in the main `wasmer-wasix` crate.

PR #4045, opened 2023-06-28
Use restore-keys in CI

Michael-F-Bryan:
# Description

I was sitting around waiting for some CI runs to complete and thought I'd look into why CI always takes forever and it turns out we are actually getting loads of cache misses. If you look around, you'll see loads of messages like [this](https://github.com/wasmerio/wasmer/actions/runs/5398883400/jobs/9806323087):

```
> Run whywaita/actions-cache-s3@v2
Cache not found for input keys: r22-wasmerio/wasmer-Linux-aafdfe74f38020d8b6611baeee368d5fccf4aabad21ebaad5279068c216d6e41-wasmer-make-test-integration-cli-linux-musl
```

Some key things this PR tries to fix:

- We don't use `restore-keys`, so if the cache key isn't found we'll always have a cache miss and rebuild the world from scratch, rather than falling back to a cache that may be partially applicable
- We put `hashFiles('Cargo.lock')` in the middle of the cache key even though it's something that changes all the time (this hinders `restore-keys`)
- The cache key doesn't include the current branch, so that means every PR will share the same cache for a particular job and constantly clobber each other

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -73,7 +73,10 @@ jobs:
           path: |
             ~/.cargo/*
             ./target/*
-          key: r22-${{ github.repository }}-${{ runner.os }}-${{ hashFiles('Cargo.lock') }}-wasmer-make-lint-linux-x64
+          key: r22-${{ github.repository }}-${{ runner.os }}-wasmer-make-lint-linux-x64-${{ hashFiles('Cargo.lock') }}
+          restore-keys: |
+            r22-${{ github.repository }}-${{ runner.os }}-wasmer-make-lint-linux-x64
+            r22-${{ github.repository }}-${{ runner.os }}
           aws-s3-bucket: wasmer-rust-artifacts-cache
           aws-access-key-id: ${{ secrets.CLOUDFLARE_ARTIFACTS_CACHE_ACCESS_TOKEN }}
           aws-secret-access-key: ${{ secrets.CLOUDFLARE_ARTIFACTS_CACHE_ACCESS_KEY }}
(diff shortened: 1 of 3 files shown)


Michael-F-Bryan:
I did some testing to see how this affects CI times.

The first build had no caching at all and took 1 hr 40 mins ([build](https://github.com/wasmerio/wasmer/actions/runs/5399590519)).

I then pushed an empty commit to re-trigger CI. As far as I can tell, each job got a cache hit and the overall CI run completed in 1 hr 4 mins ([build](https://github.com/wasmerio/wasmer/actions/runs/5402544227)).

I then made a trivial change to the `wasmer` crate so we would rebuild it and anything that depends on it. This is an exact cache hit and shows what you would expect during the course of a "normal" PR. As far as I can tell, each job got a cache hit and the overall CI run completed in 1 hr 7 mins ([build](https://github.com/wasmerio/wasmer/actions/runs/5402544227)).

I've just pushed a `cargo update` commit. This should mean we no longer get an exact cache hit because `hashFiles('Cargo.lock')` will be different, but `restore-keys` should kick in and give us *some* caching. I'll check the CI run's overall time in the morning.

Michael-F-Bryan:
It looks like we got cache misses on all jobs for [d512a8e](https://github.com/wasmerio/wasmer/pull/4045/commits/d512a8ebcbf753d6f94333eceadda21f3f2ea814) ([build](https://github.com/wasmerio/wasmer/actions/runs/5404191810)), which is very strange... It's almost like `restore-keys` isn't doing anything 🤔 

This is what the docs say:

> The cache action first searches for cache hits for key and the cache version in the branch containing the workflow run. If there is no hit, it searches for restore-keys and the version. If there are still no hits in the current branch, the cache action retries same steps on the default branch. Please note that the scope restrictions apply during the search. For more information, see "[Restrictions for accessing a cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache)."
>
> `restore-keys` allows you to specify a list of alternate restore keys to use when there is a cache miss on `key`. You can create multiple restore keys ordered from the most specific to least specific. The `cache` action searches the `restore-keys` in sequential order. When a key doesn't match directly, the action searches for keys prefixed with the restore key. If there are multiple partial matches for a restore key, the action returns the most recently created cache.

In theory, we *should* have had a cache hit on the [previous build](https://github.com/wasmerio/wasmer/actions/runs/5403544668). Even though the current cache key[^current] doesn't exist, the previous cache key[^prev] is a prefix for both the `r22-wasmerio/wasmer-Linux-wasmer-make-build-wasmer-linux-x64-wasmer` and the `r22-wasmerio/wasmer-Linux-wasmer-make-build-wasmer` restore keys.

I'm guessing the fork of `@actions/cache` that we are using might have a bug in it?

[^current]: r22-wasmerio/wasmer-Linux-wasmer-make-build-wasmer-linux-x64-wasmer-4045/merge-d08bac47cccf681b6dbfc4c23ed704c3f57c084215cc2fb462bd79281e0fcb2d

[^prev]: r22-wasmerio/wasmer-Linux-wasmer-make-build-wasmer-linux-x64-wasmer-4045/merge-2242bf728929c975bbb0829a4781b9687709c08c17c403dbb9853882316c6406

Issue #3930, opened 2023-05-30
Module instances are leaked


vapourismo:
### Describe the bug

Created module instances are leaked.

See culprits in Instruments:

```
 173.16 KB     100.0%	683	 	start
 173.16 KB     100.0%	683	 	 main
 173.16 KB     100.0%	683	 	  wasm_instance_new
 173.16 KB     100.0%	683	 	   wasmer::sys::instance::Instance::new_by_index::h3f778f08f235f2e5
 167.16 KB      96.5%	682	 	    wasmer::sys::module::Module::instantiate::h578fd9e49215478a
 167.16 KB      96.5%	682	 	     wasmer_compiler::engine::artifact::Artifact::instantiate::h8700c6ef8eb7c0a3
 138.33 KB      79.8%	227	 	      wasmer_vm::instance::allocator::InstanceAllocator::new::hbded1a5c0a954d87
  28.83 KB      16.6%	455	 	      wasmer_compiler::engine::tunables::Tunables::create_memories::h040daa7098df58ec
  21.28 KB      12.2%	227	 	       wasmer_vm::memory::VMMemory::from_definition::hb61ab796d08ad056
   4.00 KB       2.3%	1	 	       wasmer_vm::store::InternalStoreHandle$LT$T$GT$::new::hdfab1d2e9f056009
   4.00 KB       2.3%	1	 	        alloc::raw_vec::RawVec$LT$T$C$A$GT$::reserve_for_push::h119af399b1a83ede
   4.00 KB       2.3%	1	 	         alloc::raw_vec::finish_grow::hfeb597ee13fa8ad2
   4.00 KB       2.3%	1	 	          realloc
   6.00 KB       3.4%	1	 	    alloc::raw_vec::RawVec$LT$T$C$A$GT$::reserve_for_push::h9ed7b6522f5c4f3d
   6.00 KB       3.4%	1	 	     alloc::raw_vec::finish_grow::hb394925605ecddb5
   6.00 KB       3.4%	1	 	      realloc
```

Wasmer: 3.3.0
Rust: 1.66.0
Platform: macOS 13.4

### Steps to reproduce

Minimal C program to try with:

```c
#include <wasm.h>
#include <wasmer.h>
#include <stdio.h>
#include <unistd.h>

int main(void) {
  wasm_engine_t* engine = wasm_engine_new();
  wasm_store_t* store = wasm_store_new(engine);

  wasm_byte_vec_t text_code;
[...]

PR #4046, opened 2023-06-28
Add C API function to create Module from Engine instead of Store

vapourismo:
Makes it explicit that Wasmer doesn't need a Store to create a Module.

See related comments:
- https://github.com/wasmerio/wasmer/issues/3930#issuecomment-1572558077
- https://github.com/wasmerio/wasmer/issues/3930#issuecomment-1578239177


lib/c-api/src/wasm_c_api/unstable/module.rs
```diff
@@ -150,3 +152,30 @@ pub unsafe extern "C" fn wasmer_module_set_name(
 
     module.inner.set_name(name)
 }
+
+/// A WebAssembly module contains stateless WebAssembly code that has
+/// already been compiled and can be instantiated multiple times.
+///
+/// Creates a new WebAssembly Module given the configuration
+/// in the engine.
```
ptitSeb, line 160: The comment doesn't look in sync with the code

vapourismo: This was copied from `wasm_module_new`. Could you be more specific how exactly this is out of sync?

ptitSeb: You don't create a Module from a configuration here, you create a module from an Engine. That's what I meant.

vapourismo: I am not sure that is how it is meant to be interpreted. To me this means the engine configuration influences the the module creation.

See other documentation like this: https://github.com/wasmerio/wasmer/blob/bb79130ccffaec08061f7574e9955e825f7c5ec8/lib/api/src/module.rs#L48-L49

I suppose one can make this more explicit. How about:

> Creates a new WebAssembly Module using the provided engine, respecting its configuration.

vapourismo: I've changed this part of the docs.

ptitSeb (approved):

PR #4052, opened 2023-07-04
Selenium-style tests for wasmer.sh

### LINKED ISSUE #4051: Integration tests for wasmer.sh
## Motivation

The [wasmer.sh](https://wasmer.sh/) website currently has no integration tests and because it lives in a separate workspace from the rest of wasmer (in particular, the `wasmer-wasix` crate), breakages can happen often and will go unnoticed for quite a while.

We have [a CI job](https://github.com/wasmerio/wasmer/blob/f9ce24d504c7a14f27663741c235d80da60b1830/.github/workflows/web.yaml) that makes sure the code compiles, but it needs to include tests.

There are two main components that need testing:

- The `wasmer_web::WebRuntime` implementation, including
  - threading and processes
  - HTTP requests
  - package resolution and loading
  - module cache
- The interactive xterm console that users will see

## Additional context

I've made several attempts at adding integration tests to the `wasmer-web` crate in the past.

### Browser Testing using Cypress

In https://github.com/wasmerio/wasmer/pull/4036, we tried using Cypress to run UI tests in a browser.  The `xterm.js` package uses multiple `HtmlCanvasElement`s for things like the blinking prompt, terminal output, and so on, so we were planning to write tests by taking screenshots.

This eventually ran into issues because Cypress uses an `<iframe>` to embed the page under test into the test runner's webpage. This triggers a `Error: Failed to execute 'postMessage' on 'Worker': SharedArrayBuffer transfer requires self.crossOriginIsolated` error, and we couldn't figure out the correct incantation to make Chrome's security system happy. Neither adding `chromeWebSecurity: false` to `cypress.config.ts` or modifying WebPack dev server to set`cross-origin-XXX` headers seemed to make a difference. 

See [here](https://wasmerio.slack.com/archives/C04193Z3RHS/p1687791235955739) and [here](https://wasmerio.slack.com/archives/C04M230JEAK/p1687848901107969) on Slack for more context.

### Hand-written WebDriver tests (aka "Selenium")

Another approach that was taken was to use WebDriver Protocol and the [`fantoccini`](https://lib.rs/fantoccini) crate to manually spin up an instance of Chrome and do similar screenshot tests. This doesn't suffer from the same `<iframe>` issues because you are loading the page in a browser like normal and using browser APIs to interact with the page.

This proved to be quite difficult because `xterm.js`'s use of `HtmlCanvasElement` makes it hard to wait until something is ready. You need to somehow keep taking screenshots until they match some reference image. 

I ended up re-implementing a good proportion of a browser test suite and that seemed like overkill for just testing a bit of Rust works, so I thought I'd take the `wasmer-pack-test` approach instead.

### CLI testing using `wasm-pack test`

The `wasmer-pack` project comes with its own test runner which is able to run tests inside either NodeJS or your browser. The thinking was that even if we can't test the terminal (i.e. `xterm.js`), we can at least make sure the `wasmer_web::WebRuntime` works as expected.

This approach ran into a couple of issues:

- You need to remove the `#[wasm_bindgen(start)] pub fn main()` from `glue.rs` because it emits a `_start` symbol which conflicts with the one generated by the test harness, resulting in linker errors when `wasm-pack test` tries to build the integration test binary. 
  - That's easy enough to work around - just use `std::sync::Once` to make sure we run global initialization logic on first use
- Running in a Chome browser using `wasm-pack test --chrome` always fails because of `Failed to detect test as having been run. It might have timed out`
  - This happens because one of the `*.js` helpers our Rust code uses (`lib/wasi-web/js/fit.js`) is trying to do `import { Terminal } from "xterm"` and we get a `Uncaught TypeError: Failed to resolve module specifier "xterm". Relative references must start with either "/", "./", or "../"` in the browser... presumably because bundling isn't being done
- Running on Node fails because `wasm-bindgen` isn't able to generate bindings and there is no `wasm-pack test --target=web` flag like you have with `wasm-pack build`
```
Error: executing `wasm-bindgen` over the wasm file

Caused by:
    0: failed to generate bindings for import of `__wbindgen_placeholder__::__wbindgen_module`
    1: `wasm_bindgen::module` is currently only supported with `--target no-modules` and `--target web`
```

Michael-F-Bryan:
This fixes #4051 by using the `fantoccini` crate to spin up a browser and inject some JavaScript code that taps into `xterm.js` so we can start testing wasmer.sh.


tests/wasmer-web/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
```
theduke, line 1: This is debatable, but I'd put this under `wasmer-web/` instead of in the integration tests dir.

tests/wasmer-web/tests/wasi-web.rs
```diff
@@ -0,0 +1,40 @@
...
+
+#[macro_rules_attribute::apply(browser_test)]
+async fn pipe_between_commands(client: Client) {
+    // Wait for xterm to be ready
+    client.wait_for_xterm(str::contains(PROMPT)).await;
+
+    let output = client.execute_command("ls | wc", PROMPT).await;
+
+    assert_eq!(output, "      5       5      20\n");
+}
```
theduke, line 40: Can we add another test that runs `wasmer run`?

EG with `cowasy` or similar?

To check that the whole `wasmer run` flow works (package fetching/downloading/execution).

theduke (commented):
Looks sensible to me, though we should also implement  running the snapshot tests soon-ish.

tests/wasmer-web/tests/wasi-web.rs
Michael-F-Bryan: I've added a test using Python. That way we'll get "can't find the standard library" errors if filesystems don't work.

tests/wasmer-web/Cargo.toml
Michael-F-Bryan: Funnily enough, that's where I was originally putting them. 

A problem you can run into is that we're adding a bunch of extra dependencies just for communicating with a browser (i.e. `fantoccini` which pulls in `hyper`) and those dependencies won't be able to compile to `wasm32-unknown-unknown` with the rest of the crate. That means doing `wasm-pack test` in the `lib/wasi-web` directory won't work without a bunch of `#[cfg]`'s and fiddling around in `Cargo.toml`. 

It felt like putting them next to our other integration tests was the cleanest option.

theduke (approved):

Issue #4029, opened 2023-06-23
wasmer-wasix@4 does not allow to instantiate without http client


kwonoj:
Thanks for proposing a new feature!

### Motivation

We have limited wasix support in the runtime, by enabling some of wasix features like

```
[
"wasmer-wasix/sys",
  "wasmer-wasix/logging",
  "wasmer-wasix/host-fs",
  "wasmer-wasix/sys-poll",
  "wasmer-wasix/sys-thread",
  "wasmer-wasix/host-threads",
]
```

With wasmer@4, this no longer works since default_http_client does not return instance if `host-reqwest` is not enabled.   Instead, it raises build time error 

`panicked at 'Loading the builtin resolver should never fail: No HTTP client available'`

I would agree having a default network client is a reasonable default, but in certain cases like us bringing whole `reqwest` impl into binary is not something easily able to adapt. Would like to have a way to instantiate runtime without it.

### Proposed solution

Make a no-op client if backing http client is not enabled; or in a high level `networking` itself could be opt-in. 

Or, provide an interface to provide http client dynamically when init runtime, so the caller can provide no-op instance if they want.

### Alternatives

Unsure if there's a workaround without upstream changes.

### Additional context

Primary reason we want to avoid reqwest (and potentially others) is it is not trivial to have enabled for the all of the target platform we support, and also we have quite complex cargo feature conflict when bring those into.

theduke:
So the background here is that the runtime HTTP client is also used for fetching packages from the registry, to make things work in the browser. 

It should not be needed for just executing wasm files, and the `Runtime::http_client()` method already returns `Option<_>` to reflect that fact.

It's definitely possible that we try to force-use an http client in a situation where it isn't strictly neccessary.

In what situation are you running into issues?
Can you give a code example?


ps: networking is already opt in, there is a `UnsupportedVirtualNetworking` struct that implements `VirtualNetworking`.

kwonoj:
This is WIP branch shows the error: https://github.com/kwonoj/swc/tree/feat-wasmer-4 with `cargo test -p swc_plugin_runner --features plugin_transform_schema_v1 --features rkyv-impl --features ecma --features css`

kwonoj:
[...]

PR #4055, opened 2023-07-05
Use a no-op package loader by default in PluggableRuntime

Michael-F-Bryan:
# Description

This fixes the root cause of #4029 by switching `PluggableRuntime` from the `BuiltinPackageLoader` to an `UnsupportedPackageLoader` which will error out when you try to resolve a package tree. This should side-step the `.expect("Loading the builtin resolver should never fail")` @kwonoj was encountering.

I've also added default implementations for `Runtime::module_cache()` and `Runtime::package_loader()` which will return some sane values. In the case of `module_cache()`, we'll return a thread-local cache so users get some level of caching by default. For `package_loader()` I'm just returning `UnsupportedPackageLoader`.

Closes https://github.com/wasmerio/wasmer/issues/4029.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1898,9 +1898,9 @@ dependencies = [
 
 [[package]]
 name = "hermit-abi"
-version = "0.3.1"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


theduke:
Closes #4029

theduke (approved):

PR #4061, opened 2023-07-08
Use is-terminal to check tty

tottoto:
# Description

The motivation is to modernize the way to check tty. Since `isatty` and `atty` are not maintained, replaces them with `is-terminal` crate, which is used by rust ecosystems widely and the api was stabilized as `std::io::IsTerminal` at Rust 1.70. This reduces dependencies on outdated versions of the libraries they depend on as well.


theduke (approved):
We are actually already pulling in `is-terminal` through the edge CLI, so this is a good call.

Issue #3232, opened 2022-10-15
How can I got notified when method like `link` and `symlink` are called within WasiFs?


Yesterday17:
### Summary

I'm building a WASI-compatible system with a full virtual filesystem, hoping to redirect all fs-related operations to it. But I found some methods(like `link` and `symlink`) are handled by WasiFs directly, and not exposed to `FileSystem` trait.

Michael-F-Bryan:
After reviewing with @syrusakbary, we think there should be some sort of `symlink()` method added to [`wasmer_vfs::FileSystem`](https://docs.rs/wasmer-vfs/latest/wasmer_vfs/trait.FileSystem.html) for creating a symlink. For hard links, it might be easiest to set them up manually when constructing your `FileSystem` implementation.

PR #4062, opened 2023-07-08
implement symlinks for virtual-fs

ijsnow:
# Description

This is PR implements symlinks for the `virtual-fs` crate. It is a WIP and I'd appreciate any feedback/confirmation I'm heading in a direction that would get the implementation accepted! This would close https://github.com/wasmerio/wasmer/issues/3232. 

The main chunk of work I'd like reviewed is for `mem_fs::FileSystem`. I based the logic for the symlink functionality in the memory file system based on functionality I found playing around with `std::os::unix::fs::symlink` in [this gist.](https://gist.github.com/ijsnow/23ac53e85705ff889c900898384b49b7)

Remaining work outside of any feedback is to implement symlink for any of the file systems using `webc::v1::Volume` as I can't seem to find the repository containing the crate (is `pirita` private?).


ptitSeb:
pyrita is pivate for now, but that will change (very) soon.
Why using a feature for the symlink?

The PR looks ok for now, appart for the symlink behind a feature. And I see some tests already, which is good.

PR #4063, opened 2023-07-09
Fix docstring for MemoryView struct

vinnymeller:
It says to use Memory.grow() to get a view, but
Memory.view() is what returns a MemoryView.


ptitSeb (approved):

ptitSeb:
Good catch, thank you.

Issue #4033, opened 2023-06-26
The `wasmer` crate's `sys`, `js`, and `jsc` feature flags expose an inconsistent public API 


Michael-F-Bryan:
### Describe the bug

As part of #4031 I use `wasmer::Artifact::is_deserializable()` to detect whether a compiled module has been LZW compressed or not. The `make test-js` build is currently failing because `Artifact` is only available with the `sys` target, not `js` and `jsc`.

```
error[E0433]: failed to resolve: could not find `Artifact` in `wasmer`
   --> lib/wasix/src/runtime/module_cache/filesystem.rs:159:16
    |
159 |     if wasmer::Artifact::is_deserializable(&leading_bytes) {
    |                ^^^^^^^^ could not find `Artifact` in `wasmer`
```

[(GitHub Actions log)](https://github.com/wasmerio/wasmer/actions/runs/5373019395/jobs/9746986067)

The root cause is that we aren't selective when re-exporting types from underlying implementations. We'll keep having these issues as long as we're doing `pub use sys::*`.

https://github.com/wasmerio/wasmer/blob/77898a7767eb7057834fdc1a80b239e6b68cf44e/lib/api/src/lib.rs#L448-L464

I'm guessing this regression was originally introduced in #3556 and we didn't notice it when JSC support was added in #3825.

### Proposed Solution

The best way to fix this would be to change the way we re-export things to always re-export the same types. That way backend-specific items like `Artifact` won't be part of our universal `wasmer` API.

```rust
cfg_if::cfg_if! {
  if #[cfg(feature = "sys")] {
    pub mod sys;
    use sys as imp;
  } else if #[cfg(feature = "js")] {
    pub mod js;
    use js as imp;
  } else if #[cfg(feature = "jsc")] {
    pub mod jsc;
    use jsc as imp;
  } else {
    // Use a fallback implementation
     #[path = "unsupported/mod.rs"]
    mod imp;
  }
}

pub use imp::{Module, Instance, ...};
```

This has the added benefit that users will still be able to access items from the `sys`, `js`, and `jsc` modules as an escape hatch. We'll also get those nice *"Available on crate feature sys only"* annotations from rustdoc if we set `#![cfg_attr(docsrs, feature(doc_cfg))]`.

Given we literally *just* released Wasmer 4.0 with a bunch of breaking changes and this will remove items from `wasmer`'s public API, I don't think this is feasible.

### Alternative Solution 

[...]

PR #4064, opened 2023-07-10
Enable the `cfg_doc` feature on docs.rs

### LINKED ISSUE #4039: Enable the `cfg_doc` feature on docs.rs
### Describe the bug

The `wasmer` crate uses conditional compilation to add/remove a lot of functionality from the public API, but at the moment there's no easy way to tell what is available when you check out the docs. Rustdoc actually has a `doc_cfg` feature which will show a message like *"Available on crate feature fs only"*.

### Solution

We can fix this pretty easily by updating the `wasmer` crate's `lib.rs`.

```rs
#![cfg_attr(docsrs, feature(doc_cfg))]
```

[(from the `lib.rs` in tokio)](https://docs.rs/tokio/latest/src/tokio/lib.rs.html#19)

Michael-F-Bryan:
# Description

This PR makes sure docs.rs enables the `doc_cfg` and `auto_doc_cfg` features (see https://github.com/rust-lang/rust/issues/43781) when generating documentation.

I was checking this locally using the following command:

```console
$ RUSTFLAGS="--cfg=docsrs" RUSTDOCFLAGS="--cfg=docsrs" cargo +nightly d --package=wasmer-wasix --open
```

It doesn't 100% work for the `wasmer` crate because the way we use `#[cfg]` to switch between backing implementations screws things up, but the other crates with simpler conditional compilation stories are documented as expected. I'm not *too* concerned though, because #4033 will fix things properly.

Fixes #4039.

### DIFF
--- a/lib/api/Cargo.toml
+++ b/lib/api/Cargo.toml
@@ -162,3 +162,4 @@ features = [
     "wasmer-artifact-create",
     "wasmer-artifact-load",
 ]
+rustc-args = ["--cfg", "docsrs"]
--- a/lib/cli/Cargo.toml
+++ b/lib/cli/Cargo.toml
@@ -188,3 +188,6 @@ bin-dir = "bin/{ bin }"
 [package.metadata.binstall.overrides.x86_64-pc-windows-msvc]
 pkg-url = "{ repo }/releases/download/v{ version }/wasmer-windows-amd64.{ archive-format }"
 bin-dir = "bin/{ bin }.exe"
+
+[package.metadata.docs.rs]
+rustc-args = ["--cfg", "docsrs"]
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -158,7 +158,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: ${{ env.MSRV }}
+          toolchain: nightly
           target: x86_64-unknown-linux-gnu
       - run: cargo install toml-cli # toml-cli is required to run `make test-build-docs-rs`
       - name: make test-build-docs-rs
--- a/Makefile
+++ b/Makefile
@@ -454,10 +454,10 @@ test-build-docs-rs:
 			fi; \
 			printf "*** Building doc for package with manifest $$manifest_path ***\n\n"; \
 			if [ -z "$$features" ]; then \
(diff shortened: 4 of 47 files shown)


lib/cli/Cargo.toml
```diff
@@ -188,3 +188,6 @@ bin-dir = "bin/{ bin }"
 [package.metadata.binstall.overrides.x86_64-pc-windows-msvc]
 pkg-url = "{ repo }/releases/download/v{ version }/wasmer-windows-amd64.{ archive-format }"
 bin-dir = "bin/{ bin }.exe"
+
+[package.metadata.docs.rs]
+rustc-args = ["--cfg", "docsrs"]
```
theduke, line 193: Is there some workspace support for package.metadata yet? 

I guess even if there was, it wouldn't work for nested keys, and we have other bininstall related metadata?

lib/api/Cargo.toml
```diff
@@ -162,3 +162,4 @@ features = [
     "wasmer-artifact-create",
     "wasmer-artifact-load",
 ]
+rustc-args = ["--cfg", "docsrs"]
```
theduke, line 165: Do we want this?

Will this make the rebuild situation even worse due to RUSTFLAGS mismatch?

theduke (approved):

lib/api/Cargo.toml
Michael-F-Bryan: It should be fine because the `rustc-args` flag is only used by docs.rs's when it's building docs. It's hard to see because of how GitHub's UI collapses things, but `rustc-args` is under the `[package.metadata.docs.rs]` table.

lib/cli/Cargo.toml
Michael-F-Bryan: I know there is a `[workspace.metadata]` table, but I'm not sure if docs.rs's build system checks it. According to [the `cargo` docs](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-metadata-table), it's up to individual tools to use check that, though.

The `[workspace.package]` table only supports the following keys:

  |  
-- | --
authors | categories
description | documentation
edition | exclude
homepage | include
keywords | license
license-file | publish
readme | repository
rust-version | version

lib/api/Cargo.toml
theduke: Yeah I noticed this after, I initially thought it was in the main build section.

PR #4068, opened 2023-07-11
Enable tokio's "net" feature in the virtual-net crate

### LINKED ISSUE #4066: CI: Fix docs.rs Build Step
After #4064  was merged, master is now red due to the docs.rs build step failing: https://github.com/wasmerio/wasmer/actions/runs/5507170414/jobs/10036777838 .

Michael-F-Bryan:
# Description

The `virtual-net` crate doesn't enable Tokio's `net` feature so it won't compile by itself.

This fixes #4066.

### DIFF
--- a/lib/virtual-net/Cargo.toml
+++ b/lib/virtual-net/Cargo.toml
@@ -14,7 +14,7 @@ thiserror = "1"
 bytes = "1.1"
 async-trait = { version = "^0.1" }
 tracing = "0.1"
-tokio = { version = "1", features = [ "sync", "macros", "io-util", "signal" ], default_features = false, optional = true }
+tokio = { version = "1", features = [ "sync", "macros", "io-util", "signal", "net" ], default_features = false, optional = true }
 libc = { version = "0.2.139", optional = true }
 
 [features]


lib/virtual-net/Cargo.toml
```diff
@@ -14,7 +14,7 @@ thiserror = "1"
 bytes = "1.1"
 async-trait = { version = "^0.1" }
 tracing = "0.1"
-tokio = { version = "1", features = [ "sync", "macros", "io-util", "signal" ], default_features = false, optional = true }
+tokio = { version = "1", features = [ "sync", "macros", "io-util", "signal", "net" ], default_features = false, optional = true }
```
theduke, line 17: The `net` feature is not available in all environments.

This would cause difficulties for the browser builds, I think, and wouldn't be good for users that want another network stack, or no network stack.

We should instead make the `host-net` feature enable `tokio/net`.

theduke (changes requested):

lib/virtual-net/Cargo.toml
Michael-F-Bryan: @theduke I don't think that makes a difference because the `tokio` dependency is only included when `host-net` is enabled.

theduke: Right now you are unconditionally enabling the net feature, which is not what we want.

theduke (approved):

PR #4075, opened 2023-07-12
Fixed an issue where the socket address family was being corrupted due to padding fields being incorrect used.

john-sharratt:


lib/wasi-types/src/types.rs
```diff
@@ -233,6 +233,7 @@ pub mod net {
     #[repr(C)]
     pub struct __wasi_addr_t {
         pub tag: Addressfamily,
+        pub _padding: u8,
```
theduke, line 236: Please add a comment to the `_padding` field explaining why it is there.

lib/wasi-types/src/types.rs
```diff
@@ -246,6 +247,7 @@ pub mod net {
     #[repr(C)]
     pub struct __wasi_addr_port_t {
         pub tag: Addressfamily,
+        pub _padding: u8,
```
theduke, line 250: Please add a comment to the `_padding` field explaining why it is there.

lib/wasi-types/src/types.rs
```diff
@@ -259,6 +261,7 @@ pub mod net {
     #[repr(C)]
     pub struct __wasi_cidr_t {
         pub tag: Addressfamily,
+        pub _padding: u8,
```
theduke, line 264: Please add a comment to the `_padding` field explaining why it is there.

theduke (changes requested):

lib/wasi-types/src/types.rs
Michael-F-Bryan: @theduke if we are manually adding padding bytes, should we mark the struct as `#[repr(packed)]` and have an `assert_eq!(std::mem::size_of::<__wasi_addr_t>(), 42)` test?

theduke (approved):

ptitSeb (approved):

syrusakbary:
It seems this is already in master. Closing

PR #4080, opened 2023-07-14
Warn on email not validated

### LINKED ISSUE #4079: Give a warning when logging in with unverified account
### Motivation

Make it clear to users that they cannot perform certain actions (like publishing packages/apps) if their email is not yet verified.

### Proposed solution

When logging in, throw a warning along the lines of `Account {username} does not have a verified email. This means you have limit access to wasmer features. To verify your account, go to https://wasmer.io/settings/profile.` 

```bash
$ wasmer login
...
Login for Wasmer user "{username}" saved

WARNING: Account `{username}` does not have a verified email. This means you have limit access to wasmer features. 
To verify your account, go to https://wasmer.io/settings/profile.
```

Michael-F-Bryan:
This fixes #4079 by checking the `isEmailVerified` field when looking up information about the current user. If the email address isn't verified, commands like `wasmer whoami` and `wasmer login` will print a warning message.

### DIFF
--- a/lib/cli/src/commands/login.rs
+++ b/lib/cli/src/commands/login.rs
@@ -237,46 +237,49 @@ impl Login {
         let auth_state = match self.token.clone() {
             Some(token) => Ok(AuthorizationState::TokenSuccess(token)),
             None => {
-                let person_wants_to_login =
-                    match wasmer_registry::current_user(env.dir(), Some(registry.as_str()), None) {
-                        Result::Ok(Some(CurrentUser {
-                            registry,
-                            user,
-                            verified,
-                            ..
-                        })) => {
-                            println!(
-                                "You are already logged in as {:?} on registry {:?}",
-                                user, registry
-                            );
-                            if !verified {
-                                println!("Warning: Your email address still needs to be verified");
-                            }
-
-                            #[cfg(not(test))]
-                            {
-                                let login_again = Input::new()
-                                    .with_prompt(format!(
-                                        "{} {} - [y/{}]",
-                                        style("?").yellow().bold(),
-                                        style("Do you want to login with another user?")
-                                            .bright()
-                                            .bold(),
-                                        style("N").green().bold()
-                                    ))
-                                    .show_default(false)
-                                    .default(BoolPromptOptions::No)
-                                    .interact_text()?;
-
-                                login_again == BoolPromptOptions::Yes
-                            }
-                            #[cfg(test)]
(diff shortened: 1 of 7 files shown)


lib/cli/src/commands/whoami.rs
```diff
@@ -12,9 +12,24 @@ impl Whoami {
...
+        match wasmer_registry::current_user(self.env.dir(), Some(registry.as_str()), None)? {
+            Some(CurrentUser {
+                registry,
+                user,
+                verified,
+                ..
+            }) => {
+                println!("logged into registry \"{registry}\" as user \"{user}\"");
+                if !verified {
+                    println!("Warning: Your email address still needs to be verified");
```
ayys, line 25:
```suggestion
                    eprintln!("Warning: Please verify your email address to fully access all Wasmer features");
```

I am not familiar with the codebase so maybe using eprintln isn't the right way? idk, just a suggestion.

lib/cli/src/commands/login.rs
```diff
@@ -237,38 +237,49 @@ impl Login {
...
+                        user,
+                        verified,
+                        ..
+                    })) => {
+                        println!(
+                            "You are already logged in as {:?} on registry {:?}",
+                            user, registry
+                        );
+                        if !verified {
+                            println!("Warning: Please verify your email address to fully access all Wasmer features");
```
dynamite-bud, line 256: @Michael-F-Bryan If the person is not verified he can't create the tokens too. So he should not be able to login. And if he's not verified he also can't proceed with the login again path probably with that same account. In the case of person not being verified the login should just stop ?

dynamite-bud (changes requested):

lib/cli/src/commands/login.rs
dynamite-bud: @Michael-F-Bryan Okay this thing has been redacted by @syrusakbary. Only warning will suffice

dynamite-bud: As discussed, this behaviour is required and we want the person to be able to create tokens even not verified to improve the DX but show a warning like this.

dynamite-bud (approved):
Looks Good @Michael-F-Bryan

Michael-F-Bryan:
It turns out email validation wasn't necessary. From memory, it wasn't possible to get into a situation where the CLI is logged in but your email hasn't been validated.

PR #4083, opened 2023-07-18
For WASIX program, removed the mapdir(".", "/") (for #4027)

ptitSeb:
The initalisation code for WASIX program on the CLI ende the init of the filesystem with a `mapdir(".", "/")?` that was supposed to add an alias from `.` to `/`.
But the alias makes little sense as-is, and the side effect was to add a preopen folder "." which was breaking wasix-libc programs.
Removing this mapdir doesn't seems to have any adverse side effect (conanicalization of relative folder is not using this mapdir to work)"


syrusakbary:
This will be fixed in other way in another PR, we can close this

PR #4096, opened 2023-07-24
Fix benchmark

tottoto:
# Description

This pull request updates benchmark to be able to be run with `make bench`.


ptitSeb:
Nice! Thanks for the fix.

PR #4097, opened 2023-07-24
deps: Bumped wasmer-deploy-cli to 0.1.18

theduke:
Improvements:
* "app create static-site" now checks if app/package name is available
  and warns the user


theduke:
The CLI needs more work, we'll do the upgrade later.

Issue #4104, opened 2023-07-25
when FFmpeg wasix build exit, the OS prompt is not restored properly


ptitSeb:
### Describe the bug

Using the ffmpeg wasix build, like with `wasmer run wasmer/ffmpeg --mapdir /src:. -- -i /src/a.mp4 /src/a.avi` will do the coversion, but when it's over, the program is like it's waiting for an input, while it's over. The OS prompt is not showing.

### Steps to reproduce

```bash
wasmer run wasmer/ffmpeg --mapdir /src:. -- -i /src/a.mp4 /src/a.av
```

with a.mp4 existing...

### Expected behavior
Prompt showing at this end of the conversion

### Actual behavior

Prompt not showing at the end of the conversion

ptitSeb:
@john-sharratt  Here is a way to reproduce the issue with a more minimum sample:
create `main.c` with
```C
#include <unistd.h>
#include  <sys/time.h>

static int read_key()
{
    int ch;
    int n = 1;
    struct timeval tv;
    fd_set rfds;

    FD_ZERO(&rfds);
    FD_SET(0, &rfds);
    tv.tv_sec = 0;
    tv.tv_usec = 0;
    n = select(1, &rfds, NULL, NULL, &tv);
    if (n > 0) {
        n = read(0, &ch, 1);
        if (n == 1)
            return ch;

        return n;
    }
}

int main(int argc, const char* argv[])
{
  int ch = read_key();
  printf("key = %d, bye\n", ch);
}
```

Build with 

```bash
clang-15 --target=wasm32-wasi --sysroot=/home/seb/wasix-libc/sysroot32 -matomics -mbulk-memory -mmutable-globals -pthread -mthread-model posix -ftls-model=local-exec -fno-trapping-math -D_WASI_EMULATED_MMAN -D_WASI_EMULATED_SIGNAL -D_WASI_EMULATED_PROCESS_CLOCKS -O2 -g -flto -Wl,--shared-memory -Wl,--max-memory=4294967296 -Wl,--import-memory -Wl,--export-dynamic -Wl,--export=__heap_base -Wl,--export=__stack_pointer -Wl,--export=__data_end -Wl,--export=__wasm_init_tls -Wl,--export=__wasm_signal -Wl,--export=__tls_size -Wl,--export=__tls_align -Wl,--export=__tls_base -lwasi-emulated-mman -O2 -flto -g  main.c -o main.wasm
```

[...]

PR #4107, opened 2023-07-26
Added a forced shutdown on tokio runtimes as the STDIN blocks the shu…

john-sharratt:


ptitSeb:
Should solve https://github.com/wasmerio/wasmer/issues/4104

lib/cli/src/commands/run/wasi.rs
```diff
@@ -251,9 +251,9 @@ impl Wasi {
         &self,
         engine: Engine,
         env: &WasmerEnv,
-        handle: Handle,
+        runtime: tokio::runtime::Runtime,
```
ptitSeb, line 254: That's an API breaking change :O

lib/wasix/src/runtime/task_manager/tokio.rs
```diff
@@ -1,28 +1,67 @@
...
+            Self::Runtime(h, _) => h,
+        }
+    }
+}
+
 /// A task manager that uses tokio to spawn tasks.
 #[derive(Clone, Debug)]
 pub struct TokioTaskManager {
-    handle: Handle,
+    rt: RuntimeOrHandle,
```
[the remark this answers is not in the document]
john-sharratt: the construct should be backwards compatible while this field is private

lib/cli/src/commands/run/wasi.rs
john-sharratt: fixed, this is now backwards compatible

ptitSeb (approved):

PR #4117, opened 2023-07-31
Fixed an issue where inheritance was inverted

john-sharratt:
Package inheritance is inverted and thus files in the higher level packages are being overridden by files in the lower packages.


Michael-F-Bryan (commented):
Thanks for picking this up, John!

Can you update the filesystem test to make sure it works as expected?

Michael-F-Bryan (approved):

ptitSeb:
I think some tests on the new branch https://github.com/wasmerio/wasmer/actions/runs/5729906019/job/15542912665?pr=4115 are not working because of that, so I would say the tests are coming in next merge

Issue #4072, opened 2023-07-11
Misaligned pointer dereference on Mac M1 Pro & Mac M2 Pro


Eitu33:
We have been encountering the following issue when running wasmer 3.3 and wasmer 4.0 on M1 & M2:

```sh
thread 'execution' panicked at 'misaligned pointer dereference: address must be a multiple of 0x10 but is 0x109a5fc18', /Users/thomas/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wasmer-vm-4.0.0/src/trap/traphandlers.rs:219:28
```

```sh
rustc 1.72.0-nightly (f4b80cacf 2023-06-30)
binary: rustc
commit-hash: f4b80cacf93ca216c75f6ae12f4b9dec19eba42f
commit-date: 2023-06-30
host: aarch64-apple-darwin
release: 1.72.0-nightly
LLVM version: 16.0.5
```

This issue is _not_ reproduced on:
- `linux_x86_64`
- `win_x86_64`

After looking into the `traphandlers.rs` code, we noticed that there was a cast made towards the `libc::ucontext_t` type:
```rust
let ucontext = &mut *(context as *mut libc::ucontext_t);
```

But the used `libc::ucontext_t` is different depending on the platform, on linux we have:
```rust
libc::unix::linux_like::linux::gnu::b64::x86_64
pub struct ucontext_t // size = 936 (0x3A8), align = 0x8
```
Located at `libc-0.2.147/src/unix/linux_like/linux/gnu/b64/x86_64/mod.rs`

And for mac we have:
```rust
libc::unix::bsd::apple::b64::aarch64::align
pub struct ucontext_t // size = 880 (0x370), align = 0x10
```
Located at `libc-0.2.147/src/unix/bsd/apple/b64/aarch64/align.rs`

The context that is being casted into `ucontext_t` at `L219` has its alignment set at `0x08`:
```rust
context: *mut c_void // size = 8, align = 0x8
```

This produces a misalignment between the expected `0x10` and the provided `0x08` on these types of mac architectures.

This is currently a big issue for us, if help is required we would be glad to contribute.

theduke:
@ptitSeb probably a duplicate of https://github.com/wasmerio/wasmer/issues/4059 ?

bilboquet:
> @ptitSeb probably a duplicate of #4059 ?

I tested issue #4072 with @Eitu33 who reported it.
So I'm just giving my 2 cents.
[...]

PR #4120, opened 2023-08-02
Added proper definition of ucontext for macos/aarch64 to avoid unaligned issue

ptitSeb:
Added proper definition of ucontext for macos/aarch64 to avoid unaligned issue

For https://github.com/wasmerio/wasmer/issues/4072

### DIFF
--- a/lib/vm/src/trap/traphandlers.rs
+++ b/lib/vm/src/trap/traphandlers.rs
@@ -51,10 +51,7 @@ struct ucontext_t {
     uc_mcontext: libc::mcontext_t,
 }
 
-#[cfg(any(
-    target_os = "linux",
-    all(not(target_arch = "aarch64"), target_os = "macos")
-))]
+#[cfg(all(unix, not(all(target_arch = "aarch64", target_os = "macos"))))]
 use libc::ucontext_t;
 
 /// Default stack size is 1MB.
--- a/lib/vm/src/trap/traphandlers.rs
+++ b/lib/vm/src/trap/traphandlers.rs
@@ -38,7 +38,7 @@ static MAGIC: u8 = 0xc0;
 
 static DEFAULT_STACK_SIZE: AtomicUsize = AtomicUsize::new(1024 * 1024);
 
-// Current definition of `ucontext_t` in the `libc` crate is incorrect 
+// Current definition of `ucontext_t` in the `libc` crate is incorrect
 // on aarch64-apple-drawin so it's defined here with a more accurate definition.
 #[repr(C)]
 #[cfg(all(target_arch = "aarch64", target_os = "macos"))]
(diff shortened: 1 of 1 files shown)


syrusakbary (approved):

lib/vm/src/trap/traphandlers.rs
```diff
@@ -38,6 +38,25 @@ static MAGIC: u8 = 0xc0;
...
+    uc_link: *mut libc::ucontext_t,
+    uc_mcsize: usize,
+    uc_mcontext: libc::mcontext_t,
+}
+
+#[cfg(any(
+    target_os = "linux",
+    all(not(target_arch = "aarch64"), target_os = "macos")
+))]
+use libc::ucontext_t;
```
sug0, line 58:
```suggestion
#[cfg(not(all(target_arch = "aarch64", target_os = "macos")))]
use libc::ucontext_t;
```
I think this might be sufficient, and doesn't exclude other operating systems, assuming wasmer even runs on e.g. FreeBSD.

sug0 (commented):
Some of my team mates at @heliaxdev have suggested that a better fix might be to invoke `read_unaligned`, which does not check for the alignment of the `ucontext` pointer. If we implemented a fix based on `read_unaligned`, we would not need to redefine `ucontext_t` for different platforms/architectures. Once I have some disposable time to work on this issue again, I'll investigate this possibility. That said, otherwise, lgtm.

lib/vm/src/trap/traphandlers.rs
ptitSeb: That will fail on windows.
there might be an issue with freebsd and/or android yes. Maybe I should just exclude windows and macos/aarch64.

sug0: I see. maybe, then, you could use something like:
```
#[cfg(all(not(target_os = "windows"), not(all(target_arch = "aarch64", target_os = "macos"))))]
```
?

ptitSeb: Yeah, I can do that.

I was also looking at the "unix" general cfg, instead of excluding windows.

ptitSeb: done

ptitSeb:
> Some of my team mates at @heliaxdev have suggested that a better fix might be to invoke `read_unaligned`, which does not check for the alignment of the `ucontext` pointer. If we implemented a fix based on `read_unaligned`, we would not need to redefine `ucontext_t` for different platforms/architectures. Once I have some disposable time to work on this issue again, I'll investigate this possibility. That said, otherwise, lgtm.

Well, I think I'll merge this fix for now then. If you get a better fix with read unaligned, feel free to do a PR with it.

PR #4127, opened 2023-08-04
Add regression test for path_rename

### LINKED ISSUE #4125: Add Regression Test for path_rename Bug
Add a regression test for the `path_rename` bug fixed in https://github.com/wasmerio/wasmer/pull/4121 .

Arshia001:
Add regression test for changes in #4121. Closes #4125.


Arshia001:
@theduke I'm not sure where I cat put the source of the .wasm file in this PR?

tests/integration/cli/tests/snapshot.rs
```diff
@@ -1268,3 +1268,12 @@ fn test_snapshot_quickjs() {
         .run_wasm(include_bytes!("./wasm/qjs.wasm"));
     assert_json_snapshot!(snapshot);
 }
+
+#[cfg_attr(any(target_env = "musl", target_os = "windows"), ignore)]
```
theduke, line 1272: Why are we ignoring these?

tests/integration/cli/tests/snapshot.rs
```diff
@@ -1268,3 +1268,12 @@ fn test_snapshot_quickjs() {
...
+#[test]
+fn test_snapshot_fs_rename() {
+    let snapshot = TestBuilder::new()
+        .with_name(function!())
+        .run_wasm(include_bytes!("./wasm/fs-rename.wasm"));
```
theduke, line 1277: I assume the test is written in Rust?

I'd say we take this opportunity to add a new `tests/rust-wasi-tests` crate, which as a bunch of `src/bin/xxx.rs` files for each `.wasm`. 

(We can then also add a `rust-wasix-tests` crate with wasix tests, etc. )

tests/integration/cli/tests/snapshot.rs
Arshia001: No idea, I just noticed they were on every one of these tests, so I assumed there was a reason. Isn't there?

tests/integration/cli/tests/snapshot.rs
Arshia001: This makes sense, I'll do it in this PR.

theduke: We can later also add a make command that compiles all the tests and copies the wasm files to a well-known location, where they can be included from in tests.

tests/integration/cli/tests/snapshot.rs
Arshia001: I just removed the filters, let's see if it'll work.

tests/integration/cli/tests/snapshot.rs
Arshia001: I just realized commenting "Done" on a conversation resolves it instead of adding a comment!

tests/integration/cli/tests/snapshot.rs
Arshia001: I assume this is why the filters are created: https://github.com/wasmerio/wasmer/actions/runs/5760653049/job/15617657063#step:26:208

tests/integration/cli/tests/snapshot.rs
Arshia001: @theduke it appears that the code in that new crate is being compiled by one of the tests, and since the APIs don't exist on Windows, it's failing. Any idea how we might fix that?

theduke: I think we can just remove the crate from the workspace.

tests/integration/cli/tests/snapshot.rs
theduke: Ah yeah.... seems like we messed up the path detection here. 

Let's create a followup issue to fix this, but out of scope for this PR.

tests/integration/cli/tests/snapshot.rs
Arshia001: let me try that.

tests/integration/cli/tests/snapshot.rs
Arshia001: Done.

theduke (approved):

PR #4130, opened 2023-08-06
Update to tracing-subscriber 0.3

tottoto:
# Description

This pull request updates to use `tracing-subscriber` 0.3, which has been already used in the other part of `wasmer` project.


theduke (approved):
Good catch, thanks.

PR #4134, opened 2023-08-07
Enable the api sys feature which is required to build

tottoto:
# Description

The motivation is to fix build `wasmer-emscripten`. `api`'s `sys` feature seems to be required to build `wasmer-emscripten`.


tottoto:
I overlooked JSC build.

PR #4139, opened 2023-08-09
0-copy module deserialization

### LINKED ISSUE #3882: mmap Support for Artifacts / deserialized Modules
Serialized module artifacts can currently only be used through deserialization, which means the embedded data has to be copied. 

We want have the ability to use module artifacts from an `mmap` backed buffer to make memory management easier.

`rkyv`, the serialization library we use, also provides a zero-copy mode via the `archive` feature.
See: https://rkyv.org/architecture/archive.html 

To make this possible, we need to change `Module` to optionally contain an Artifact backed by `ArchivedSerializableModule` instead of `SerializableModule`. 

Note: the `ArchviedXXX` types are generated by the rkyv derive macros.

This may requiring hiding all the access to different data elements behind accessor methods.

cc @john-sharratt

Arshia001:
Closes #3882. This change lets modules be loaded directly from archived structs by adding traits that are implemented by both plain and archived versions of structs used in modules.


lib/compiler/src/artifact_builders/artifact_builder.rs
```diff
@@ -165,8 +174,8 @@ impl ArtifactBuild {
     }
 
     /// Get Function Relocations
-    pub fn get_function_relocations(&self) -> PrimaryMap<LocalFunctionIndex, Vec<Relocation>> {
-        self.serializable.compilation.function_relocations.clone()
+    pub fn get_function_relocations(&self) -> &PrimaryMap<LocalFunctionIndex, Vec<Relocation>> {
+        &self.serializable.compilation.function_relocations
```
syrusakbary, line 178: Thank god for this! (the clones were a big hack)

lib/compiler/src/artifact_builders/artifact_builder.rs
```diff
@@ -242,3 +254,227 @@ impl ArtifactCreate for ArtifactBuild {
...
+
+    fn memory_styles(&self) -> &PrimaryMap<MemoryIndex, MemoryStyle> {
+        &self.compile_info.memory_styles
+    }
+
+    fn table_styles(&self) -> &PrimaryMap<TableIndex, TableStyle> {
+        &self.compile_info.table_styles
+    }
+
+    fn serialize(&self) -> Result<Vec<u8>, SerializeError> {
```
syrusakbary, line 456: I'm concerned on having some of this logic duplicated on both Plain and Archive... is it currently duplicated or did I get it wrong?

lib/compiler/src/engine/artifact.rs
```diff
@@ -319,7 +388,13 @@ impl Artifact {
...
         };
 
-        artifact.internal_register_frame_info();
+        // The error scenario could be improved here. We're deserializing frame info
+        // as part of registration, so that's returning a DeserializeError, but
+        // this function returns a CompileError that's then wrapped in a
+        // DeserializeError anyway.
+        artifact
+            .internal_register_frame_info()
+            .map_err(|e| CompileError::Validate(format!("{:?}", e)))?;
```
syrusakbary, line 397: Is this a Validation Error really? Maybe we can use other CompileError variant a bit more fit for a frame info registration error

syrusakbary (approved):
PR looks great other than for the few comments

lib/compiler/src/artifact_builders/artifact_builder.rs
Arshia001: The logic isn't exactly the same aside from the header, but I could easily refactor that bit out into a new function.

lib/compiler/src/engine/artifact.rs
Arshia001: Well, not really. The problem is that we're now doing a bit of on-demand deserialization as part of the compile process, so compilation _can_ also return deserialization errors. I could change the return type to `DeserializeError`, but since `from_parts` is public, that would be a breaking change.

lib/api/src/engine.rs
```diff
@@ -90,11 +99,43 @@ impl Engine {
...
+    }
+
+    #[cfg(all(feature = "sys", not(target_arch = "wasm32")))]
+    /// Load a serialized WebAssembly module from a memory mapped file and deserialize it.
+    ///
+    /// NOTE: you should almost always prefer [`Self::deserialize_from_mmapped_file`].
+    ///
+    /// # Safety
+    /// See [`Artifact::deserialize_unchecked`].
+    pub unsafe fn deserialize_from_mmapped_file_unchecked(
```
theduke, line 113: We started to move architecture specific implementations to `NativeEngineExt` instead of inherent impls.

Since mmap isn't available for WASM targets, let's move the `mmaped_*` methods to that trait.

lib/api/src/engine.rs
```diff
@@ -90,11 +99,43 @@ impl Engine {
...
+            bytes.into(),
+        )?))
+    }
+
+    #[cfg(all(feature = "sys", not(target_arch = "wasm32")))]
+    /// Load a serialized WebAssembly module from a memory mapped file and deserialize it.
+    ///
+    /// # Safety
+    /// See [`Artifact::deserialize`].
+    pub unsafe fn deserialize_from_mmapped_file(
```
theduke, line 129: As above, move to `NativeEngineExt`.

lib/compiler/src/artifact_builders/artifact_builder.rs
```diff
@@ -232,13 +244,234 @@ impl ArtifactCreate for ArtifactBuild {
...
+                    .map_err(|e| DeserializeError::CorruptedBinary(format!("{:?}", e)))?,
+            );
+            ModuleFromArchive::from_serializable_module(module)
+        })?;
+
+        // Safety: we know the lambda will execute before getting here and assign both values
+        Ok(Self {
+            cell,
+            compile_info: unsafe { compile_info.assume_init() },
+        })
```
theduke, line 319: Nit: for better readability, can we do:

```suggestion
        // Safety: we know the lambda will execute before getting here and assign both values
        let compile_info = unsafe { compile_info.assume_init() };
        Ok(Self {
            cell,
            compile_info,
        })
```

lib/compiler/src/engine/link.rs
```diff
@@ -130,18 +131,28 @@ fn apply_relocation(
 
 /// Links a module, patching the allocated functions with the
 /// required relocations and jump tables.
-pub fn link_module(
+pub fn link_module<'a>(
     _module: &ModuleInfo,
     allocated_functions: &PrimaryMap<LocalFunctionIndex, FunctionExtent>,
-    function_relocations: Relocations,
+    function_relocations: impl Iterator<
```
theduke, line 137: nit: this is very noisy in the argument signature, can we move that to a generic bound?

lib/types/src/compilation/function.rs
```diff
@@ -45,6 +48,42 @@ pub struct FunctionBody {
     pub unwind_info: Option<CompiledFunctionUnwindInfo>,
 }
 
+/// Any struct that acts like a `FunctionBody`.
+#[allow(missing_docs)]
+pub trait FunctionBodyLike<'a> {
```
theduke, line 53: I'm not a fan of the `*Like` naming scheme.

But then again, I don't have a significantly better idea...

The best I can come up with is `*Accessor` or `*Reader`, but I don't think those are great either.

We can leave it as is if we don't have other ideas.

lib/compiler/src/engine/artifact.rs
```diff
@@ -85,12 +91,17 @@ impl Default for ArtifactId {
...
 pub struct Artifact {
     id: ArtifactId,
-    artifact: ArtifactBuild,
+    artifact: ArtifactBuildVariant,
     // The artifact will only be allocated in memory in case we can execute it
     // (that means, if the target != host then this will be None).
     allocated: Option<AllocatedArtifact>,
 }
 
+pub enum ArtifactBuildVariant {
```
theduke, line 100: Doc comment explaining what this is and why it exists.

lib/compiler/src/engine/artifact.rs
```diff
@@ -931,7 +1159,7 @@ impl Artifact {
 
         Ok(Self {
             id: Default::default(),
-            artifact,
+            artifact: ArtifactBuildVariant::Plain(artifact),
             allocated: Some(AllocatedArtifact {
                 frame_info_registered: false,
                 frame_info_registration: None,
```
theduke, line 1165: Let's take this opportunity to add some sorely needed focued Artifact tests.

I don't know if the all the related types implement (Partial)Eq, but we could have a test that round-trips a module ,and then compares the two variants for equality.

We should also have some fixtures of serialized modules in files, and test that they can still be loaded so we don't accidentally break deserialization.

Note that there is an artifact version number that we currently manually increment if the layout changes, which is also used for constructing the artifact cache filenames.

https://github.com/wasmerio/wasmer/blob/7602de387cd6a4cc940098c170320f9f5229a847/lib/types/src/serialize.rs#L214

theduke (changes requested):
Great work!

Mostly some nitpicks, and a request for adding some tests.

lib/compiler/src/engine/artifact.rs
theduke: Tests should use some reasonably complex modules, like Python/PHP. 

I think there are already `.wasm` files in the repo, or if not, they can be extracted from the webcs.

lib/api/src/engine.rs
Arshia001: All of the deserialize methods are `#[cfg(all(feature = "sys", not(target_arch = "wasm32")))]`. Do we want to keep the other ones here and just move the mmapped functions?

lib/types/src/compilation/function.rs
Arshia001: I'm not a fan of *Like, *Accessor or *Reader either. Open to suggestions though!

lib/compiler/src/engine/link.rs
Arshia001: The generic version of this function definition is much less readable IMO:

```
pub fn link_module<
    'a,
    FunctionRelocationIterator,
    FunctionRelocationIteratorInner,
    FunctionRelocation,
    SectionRelocationIterator,
    SectionRelocationIteratorInner,
    SectionRelocation,
>(
    _module: &ModuleInfo,
    allocated_functions: &PrimaryMap<LocalFunctionIndex, FunctionExtent>,
    function_relocations: FunctionRelocationIterator,
    allocated_sections: &PrimaryMap<SectionIndex, SectionBodyPtr>,
    section_relocations: SectionRelocationIterator,
    libcall_trampolines: SectionIndex,
    trampoline_len: usize,
) where
    FunctionRelocationIterator:
        Iterator<Item = (LocalFunctionIndex, FunctionRelocationIteratorInner)>,
    FunctionRelocationIteratorInner: Iterator<Item = &'a FunctionRelocation>,
    FunctionRelocation: RelocationLike + 'a,
    SectionRelocationIterator: Iterator<Item = (SectionIndex, SectionRelocationIteratorInner)>,
    SectionRelocationIteratorInner: Iterator<Item = &'a SectionRelocation>,
    SectionRelocation: RelocationLike + 'a,
```

theduke: I just meant the one large type.

But that of course also is messy because of one generic vs 5 `impls`...

No objection to leaving it as is.

lib/api/src/engine.rs
theduke: We should probably have a followup task though to move them whenever `5.0` comes around.

Arshia001: Done. I also created #4143.

Arshia001:
According to my measurements, on average this change this much reduction in module load times:
* 40% for cowsay
* 50% for rustpython
* 47% for php

lib/compiler/src/engine/artifact.rs
Arshia001: I added two new tests, one for serializing freshly-compiled modules and one for deserializing existing compiled modules.

theduke: Awesome, thanks.

theduke (approved):

theduke:
This is approved, but we should wait with the merge until `4.1.2` is out, since this is a pretty big change and we want to give it some time to bake on master.

ptitSeb:
Why is this merged!!!!

Arshia001:
@ptitSeb sorry, I missed @theduke's laat comment. My bad.

PR #4141, opened 2023-08-10
Web http client

Michael-F-Bryan:
This introduces a web HTTP client that uses JavaScript's `fetch()` API.


lib/wasix/src/http/mod.rs
```diff
@@ -4,6 +4,12 @@ pub mod client_impl;
 #[cfg(feature = "host-reqwest")]
 pub mod reqwest;
 
+#[cfg(feature = "js")]
+mod web_http_client;
```
theduke, line 8: Do you think it makes sense to isolate all the web/js specific code in an `arch/web/`?

I think that would be better for code structure instead of sprinkling web related code around the crate.

lib/wasix/src/lib.rs
```diff
@@ -30,6 +30,9 @@ compile_error!(
     "The `js` feature must be enabled only for the `wasm32` target (either `wasm32-unknown-unknown` or `wasm32-wasi`)."
 );
 
+#[cfg(all(test, target_arch = "wasm32"))]
+wasm_bindgen_test::wasm_bindgen_test_configure!(run_in_browser);
```
theduke, line 34: Do you think it makes sense to isolate all the web/js specific code in an `arch/web/`?

I think that would be better for code structure instead of sprinkling web related code around the crate.

lib/wasix/src/utils/mod.rs
```diff
@@ -1,20 +1,22 @@
...
 mod owned_mutex_guard;
 pub mod store;
 mod thread_parker;
 
-mod dummy_waker;
-pub use self::dummy_waker::WasiDummyWaker;
-
-use std::collections::BTreeSet;
+#[cfg(feature = "js")]
+pub(crate) mod web;
```
theduke, line 7: Is this `js` or `web`?

Should we align the naming?

This code should potentially also work in non-browser environments , like Deno/Node, so `js` seems like a more sensible name.

On the other hand, `web` might then require some different setup over a generic `js`?

I primarily want to prevent us boxing ourselves into a suboptimal naming that will require a breaking change.
Thoughts?

lib/wasix/src/utils/web.rs
```diff
@@ -0,0 +1,15 @@
+use wasm_bindgen::{JsCast, JsValue};
+
+/// Try to extract the most appropriate error message from a [`JsValue`],
```
theduke, line 3: As mentioned above, should we have a dedicated module (tree) for web?

lib/wasix/src/http/web_http_client.rs
```diff
@@ -0,0 +1,272 @@
...
+    pub fn extend_default_headers(&mut self, map: HeaderMap) -> &mut Self {
+        self.default_headers.extend(map);
+        self
+    }
+}
+
+impl HttpClient for WebHttpClient {
+    fn request(&self, mut request: HttpRequest) -> BoxFuture<'_, Result<HttpResponse, Error>> {
+        let (sender, receiver) = futures::channel::oneshot::channel();
+
```
theduke, line 44: Note that `reqwest` actually supports the browser. 
Any drawbacks to just using reqwest for js builds as well?

(I think one potential complication is that the futures used by reqwest in browser builds aren't send/sync...)

Michael-F-Bryan: The biggest benefit I see is being able to drop `reqwest` and all the dependencies it pulls in. 

That'll have massive benefits in terms of both compile times and the size of the `*.wasm` file every user's browser needs to download.

lib/wasix/src/http/mod.rs
Michael-F-Bryan: I was actually thinking of doing a big re-structure of the `wasmer-wasix` crate at some point. At the moment, we've got a bunch of random stuff in random places and have made a lot of things public accidentally. For example, the `ModuleCache` and all that are stored under the `wasmer_wasix::runtime` namespace, which doesn't really make sense. 

It's a bit tricky because we have two orthogonal ways to organise things and neither one is obviously better than the other. One way is to split things up trait/component, while the other way is to split things up by architecture/target. 

I personally prefer to put trait implementations in their own module next to the trait they implement. It's just a case of making sure we work hard to keep the amount of conditional compilation under control.

lib/wasix/src/utils/mod.rs
Michael-F-Bryan: It's funny that you mention this... We were originally planning to rename the `js` feature to `web` in the main `wasmer` crate, but backed it out because `web` is just as bad a name as `js` (https://github.com/wasmerio/wasmer/issues/3844).

I'm inclined to leave things as-is for now. The `wasmer-wasix` crate is still very much unstable and the public API needs an overhaul, so we can probably leave it until we are ready to do `1.0`.

lib/wasix/src/http/web_http_client.rs
theduke: That's a very valid argument.

ptitSeb (approved):

PR #4157, opened 2023-08-17
Remove `Send` and `Sync` from `wasmer` types backed by JavaScript objects

### LINKED ISSUE #4158: Soundness bug with Send+Sync implementations in the browser
### Describe the bug

It looks like we accidentally introduced a soundness bug in #3556 when adding `Send+Sync` bounds to things like `wasmer::js::Module` and `wasmer::js::Memory`.

Here's one example:

https://github.com/wasmerio/wasmer/blob/d53b9e26d3275483e6af9dfbddb69d57c825fc2b/lib/api/src/js/module.rs#L48-L57

The underlying assumption is that because a type implements `structuredClone` it's fine to implement `Send` and `Sync`.

However, `structuredClone` **only** applies when sending values to a worker using `postMessage`.

If you use any other method to transfer a value to another thread (e.g. channels) then the underlying `JsValue` will point to the wrong object on that thread's wasm-bindgen "heap" and you'll have a bad time.

### Additional context

See the conversation [on Slack](https://wasmerio.slack.com/archives/C04193Z3RHS/p1692254473023459) for more

Michael-F-Bryan:
This fixes a soundness bug introduced in #3556 which added `Send+Sync` bounds to things like `wasmer::js::Module` and `wasmer::js::Memory`.

Here's one example:

https://github.com/wasmerio/wasmer/blob/d53b9e26d3275483e6af9dfbddb69d57c825fc2b/lib/api/src/js/module.rs#L48-L57

The underlying assumption is that because a type implements `structuredClone` it's fine to implement `Send` and `Sync`.

However, `structuredClone` **only** applies when sending values to a worker using `postMessage`.

If you use any other method to transfer a value to another thread (e.g. channels) then the underlying `JsValue` will point to the wrong object on that thread's wasm-bindgen "heap" and you'll have a bad time.

Fixes #4158.

### DIFF
--- a/lib/api/src/js/externals/memory.rs
+++ b/lib/api/src/js/externals/memory.rs
@@ -53,20 +53,7 @@ pub struct Memory {
     pub(crate) handle: VMMemory,
 }
 
-// Only SharedMemories can be Send in js, becuase they support `structuredClone`.
-// Normal memories will fail while doing structuredClone.
-// In this case, we implement Send just in case as it can be a shared memory.
-// https://developer.mozilla.org/en-US/docs/Web/API/structuredClone
-// ```js
-// const memory = new WebAssembly.Memory({
-//   initial: 10,
-//   maximum: 100,
-//   shared: true // <--- It must be shared, otherwise structuredClone will fail
-// });
-// structuredClone(memory)
-// ```
-unsafe impl Send for Memory {}
-unsafe impl Sync for Memory {}
+assert_not_implemented!(Memory: !Send + !Sync);
 
 impl Memory {
     pub fn new(store: &mut impl AsStoreMut, ty: MemoryType) -> Result<Self, MemoryError> {
--- a/lib/api/src/js/externals/table.rs
+++ b/lib/api/src/js/externals/table.rs
@@ -11,9 +11,7 @@ pub struct Table {
     pub(crate) handle: VMTable,
 }
 
-// Table can't be Send in js because it dosen't support `structuredClone`
-// https://developer.mozilla.org/en-US/docs/Web/API/structuredClone
-// unsafe impl Send for Table {}
+assert_not_implemented!(Table: !Send + !Sync);
 
 fn set_table_item(table: &VMTable, item_index: u32, item: &Function) -> Result<(), RuntimeError> {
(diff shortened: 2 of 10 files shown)


lib/api/src/js/macros.rs
```diff
@@ -0,0 +1,32 @@
+/// Assert that a type does **not** implement a set of traits.
+macro_rules! assert_not_implemented {
+    ($t:ty : !$first:ident $(+ !$rest:ident)*) => {
```
Arshia001, line 3: Just a small nit: `assert_not_implemented!(Instance: !Send + !Sync);` sounds like a double negative. so one might initially expect `Instance` to *not* implement `!Send`, i.e. that `Instance` should be `Send`. I don't have suggestions for alternate syntax though.

lib/api/src/js/externals/memory.rs
```diff
@@ -53,20 +53,7 @@ pub struct Memory {
...
-// ```js
-// const memory = new WebAssembly.Memory({
-//   initial: 10,
-//   maximum: 100,
-//   shared: true // <--- It must be shared, otherwise structuredClone will fail
-// });
-// structuredClone(memory)
-// ```
-unsafe impl Send for Memory {}
-unsafe impl Sync for Memory {}
```
theduke, line 69: Side note: should't clippy have required a Safety: ... comment here?

lib/api/src/js/externals/table.rs
```diff
@@ -11,9 +11,7 @@ pub struct Table {
     pub(crate) handle: VMTable,
 }
 
-// Table can't be Send in js because it dosen't support `structuredClone`
-// https://developer.mozilla.org/en-US/docs/Web/API/structuredClone
-// unsafe impl Send for Table {}
```
theduke, line 16: Can we keep a comment on all `assert_not_implemented` usage points so future readers know what's going on here?

theduke (approved):
LGTM

Although: this is technically a breaking change, and would require a `5.0`...

Although we can probably get by with just ignoring this here, since it is basically a bug fix.

syrusakbary:
This PR looks great, but it will require further rework on WASIX to make it work with this changes

PR #4165, opened 2023-08-18
Fix writing to middle of files and improve performance

### LINKED ISSUE #3903: Fix Incredibly Slow write Implementation of TmpFs In-Memory Files

Arshia001:
This PR fixes writing to the middle of temp FS files. The current behaviour slices the buffer and inserts the new buffer in the middle. This is in contrast to the behaviour of normal file systems where writing to the middle of a file overwrites existing data. As a consequence of this fix, there is no longer a performance issue either.

Closes #3903.

### DIFF
--- a/lib/virtual-fs/src/mem_fs/file.rs
+++ b/lib/virtual-fs/src/mem_fs/file.rs
@@ -1343,37 +1343,14 @@ impl File {
 
 impl File {
     pub fn write(&mut self, buf: &[u8], cursor: &mut u64) -> io::Result<usize> {
-        match *cursor {
-            // The cursor is at the end of the buffer: happy path!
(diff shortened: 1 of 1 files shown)


theduke:
Supersedes https://github.com/wasmerio/wasmer/pull/3905

theduke (approved):
LGTM

PR #4179, opened 2023-08-25
Faster compiles for debug by using release version of cranelift

john-sharratt:
Uses the release version of cranelift for debug as it takes about 5 mins for basic WASM apps on my machine without it.


syrusakbary (approved):

theduke:
I already added this ages ago, but @ptitSeb removed it again , I forgot why.

@ptitSeb do you remember the reason?

ptitSeb (approved):

PR #4180, opened 2023-08-25
NativeEngineExt::deserialize now returns Module

theduke:
This is done to make the result actually usable, because there is no
public constructor that allows going from Artifact to Module.

Followup to #4139


Arshia001 (approved):
Lgtm

PR #4184, opened 2023-08-28
Allow `VirtualTaskManager` to explicitly transfer a module to a blocking task on the threadpool

Michael-F-Bryan:
We need a way to explicitly transfer a `wasmer::Module` to tasks running on the threadpool. Using `VirtualTaskManager::task_dedicated()` will capture the value which means we'll hit #4158 when running in the browser.

### DIFF
--- a/lib/wasix/src/runtime/task_manager/mod.rs
+++ b/lib/wasix/src/runtime/task_manager/mod.rs
@@ -159,6 +159,23 @@ pub trait VirtualTaskManager: std::fmt::Debug + Send + Sync + 'static {
 
     /// Returns the amount of parallelism that is possible on this platform
     fn thread_parallelism(&self) -> Result<usize, WasiThreadError>;
+
+    /// Schedule a blocking task to run on the threadpool, explicitly
+    /// transferring a [`Module`] to the task.
+    ///
+    /// This should be preferred over [`VirtualTaskManager::task_dedicated()`]
+    /// where possible because [`wasmer::Module`] is actually `!Send` in the
+    /// browser and can only be transferred to background threads via
+    /// an explicit `postMessage()`. See [#4158] for more.
+    ///
+    /// [#4158]: https://github.com/wasmerio/wasmer/issues/4158
+    fn spawn_with_module(
+        &self,
+        module: Module,
+        task: Box<dyn FnOnce(Module) + Send + 'static>,
+    ) -> Result<(), WasiThreadError> {
+        self.task_dedicated(Box::new(move || task(module)))
+    }
 }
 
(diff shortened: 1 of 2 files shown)


lib/wasix/src/runtime/task_manager/mod.rs
```diff
@@ -159,6 +159,23 @@ pub trait VirtualTaskManager: std::fmt::Debug + Send + Sync + 'static {
...
+    /// Schedule a blocking task to run on the threadpool, explicitly
+    /// transferring a [`Module`] to the task.
+    ///
+    /// This should be preferred over [`VirtualTaskManager::task_dedicated()`]
+    /// where possible because [`wasmer::Module`] is actually `!Send` in the
+    /// browser and can only be transferred to background threads via
+    /// an explicit `postMessage()`. See [#4158] for more.
+    ///
+    /// [#4158]: https://github.com/wasmerio/wasmer/issues/4158
+    fn spawn_with_module(
```
theduke, line 172: Being able to transfer explicitly was the initial reason for having `task_wasm`. 

is there a reason why that isn't sufficient?

Michael-F-Bryan: The existing `task_wasm()` method seems to be specialised for use inside syscalls and pulls in a bunch of state or semantics that don't really make sense outside of a syscall context (or is just plain unavailable) like the `WasiEnv`, instance snapshots, creating/sharing linear memory, and so on.

For example, [here](https://github.com/wasmerio/wasmer-js/blob/79e556a361b3523f1eeadef56ba71e21fc076c4f/src/run.rs#L58-L70
) is how I'm using `spawn_with_module()`:

```rust
    // Note: The WasiEnvBuilder::run() method blocks, so we need to run it on
    // the thread pool.
    let tasks = runtime.task_manager().clone();
    tasks.spawn_with_module(
        module,
        Box::new(move |module| {
            let _span = tracing::debug_span!("run").entered();
            let result = builder.run(module).map_err(anyhow::Error::new);
            let _ = sender.send(ExitCondition(result));
        }),
    )?;
```

You can see that literally all I'm passing into the task is a WebAssembly module.

Yet this is what `task_wasm()` expects as inputs:

https://github.com/wasmerio/wasmer/blob/98b1c6a99746105e4fef66e93f78ba2a4d96a70f/lib/wasix/src/runtime/task_manager/mod.rs#L56-L65

I *could* shoe-horn my use case into `task_wasm()` by making the `env` field optional and giving dummy values for everything else, but then we'll have issues with mixing abstraction layers and the task manager doing a bunch of extra stuff which may or may not even work in that context.

theduke (approved):

PR #4186, opened 2023-08-28
Add stdin/stdout/stderr streams to `WasiRunner` and only use async threading when requested

Michael-F-Bryan:
While working on https://github.com/wasmerio/wasmer-js/pull/328, I noticed that we don't have a way to intercept the stdin, stdout, and stderr streams of a `WasiRunner`. 

I've also made it so we only use async threading when it is explicitly requested. Using async threads in the browser triggers #4158 because it uses things like `task_dedicated()` and `resume_wasm_after_poller()` which may smuggle `!Send` types to background threads.


theduke (approved):
lgtm

Issue #4188, opened 2023-08-28
Artifact Cache: Consider CPU Feature Flags When Loading Artifacts 


theduke:
Module artifacts are compiled with specific CPU feature flags.

This means artifacts aren't universally compatible, but can only run on CPUs that support all the features the artifact was compiled with.

It is possible to end up with a module cache directory which contains artifacts that are not compatible with the current CPU.
This results in a very confusing errors like ` missing required CPU features: "EnumSet(AVX512DQ | AVX512VL | AVX512F)"', lib/wasix/tests/runners.rs:75:21` .

The module cache implementation should either check the artifact features for support, or somehow include the cpu features in the cache path. Note: the latter is quite tricky to do in a sensible way that doesn't require lots of cache lookups.

Related to #4187

theduke:
Nothing needed anymore.

PR #4190, opened 2023-08-30
Early check that a cached artifact is compatible with current CPU Features

ptitSeb:
Early check that a cached artifact is compatible with current CPU Features

This will fix https://github.com/wasmerio/wasmer/issues/4188

This has been tested localy by:
1. clearing cache
2. Forcing an CPUFeatures with more feature than the CPU currently support with
3. Generatic a cache (running will fail because of the super set of feature with this PR)
4. Relaunching the wasm program with the PR, it will silently discard the cache and re-create it

Code to force more Feature than the cpu have (in my case at least, as my CPU is pretty old and only have `EnumSet(SSE2 | SSE3 | SSSE3 | SSE41 | SSE42 | POPCNT | AVX)`):
```rust
impl StoreOptions {
    /// Gets the store for the host target, with the compiler name selected
    pub fn get_store(&self) -> Result<(Store, CompilerType)> {
        let target = Target::default();
        let cpu_features = target.cpu_features().clone();
        cpu_features.insert(CpuFeature::AVX2);
        cpu_features.insert(CpuFeature::AVX512DQ);
        cpu_features.insert(CpuFeature::AVX512VL);
        cpu_features.insert(CpuFeature::AVX512F);
        let target = Target::new(target.triple().clone(), cpu_features);
        self.get_store_for_target(target)
    }
```

in `lib/cli/src/store.rs`

### DIFF
--- a/lib/compiler/src/engine/artifact.rs
+++ b/lib/compiler/src/engine/artifact.rs
@@ -282,6 +282,16 @@ impl Artifact {
                 artifact,
                 allocated: None,
             });
+        } else {
+            // check if cpu features are compatible before anything else
+            let cpu_features = artifact.cpu_features();
+            for feature in cpu_features {
+                if !target.cpu_features().contains(feature) {
+                    return Err(DeserializeError::Incompatible(
+                        "Some CPU Features needed for the artifact are missing".to_owned(),
+                    ));
+                }
+            }
         }
         let module_info = artifact.module_info();
         let (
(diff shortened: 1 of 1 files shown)


lib/compiler/src/engine/artifact.rs
```diff
@@ -282,6 +282,16 @@ impl Artifact {
                 artifact,
                 allocated: None,
             });
+        } else {
+            // check if cpu features are compatible before anything else
+            let cpu_features = artifact.cpu_features();
+            for feature in cpu_features {
+                if !target.cpu_features().contains(feature) {
+                    return Err(DeserializeError::Incompatible(
+                        "Some CPU Features needed for the artifact are missing".to_owned(),
```
theduke, line 291: Can we first collect the missing features and then add them to the error so users know what features are missing?

theduke (approved):

lib/compiler/src/engine/artifact.rs
ptitSeb: done

PR #4191, opened 2023-08-30
Fix invalid access to wasi instance handles in wasix proc_spawn

g-bartoszek:
# Description
Fixes `wasix` `proc_spawn` implementation by removing redundant attempt to retrieve memory from the uninitialized `inner` handle.


theduke (approved):
Thanks, @g-bartoszek !

LGTM

PR #4193, opened 2023-08-31
Fix sockets

Arshia001:
A couple of small fixes to sockets:

* Fixed endianness issue when deserializing IPv6 addresses that caused swapped bytes in each `u16`.
* Fixed receiving data from UDP sockets that have not been connected to a remote socket (i.e. server sockets).

### DIFF
--- a/lib/wasix/src/net/mod.rs
+++ b/lib/wasix/src/net/mod.rs
@@ -174,15 +174,9 @@ pub(crate) fn read_ip_port<M: MemorySize>(
             (IpAddr::V4(Ipv4Addr::new(o[2], o[3], o[4], o[5])), port)
         }
         Addressfamily::Inet6 => {
-            let [a, b, c, d, e, f, g, h] = {
-                let o = [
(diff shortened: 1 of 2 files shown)


ptitSeb:
This seems very similar to https://github.com/wasmerio/wasmer/pull/4166

Arshia001:
@theduke the problem is that UDP is still not really working. There's something wrong with either binding or receiving, but I haven't figured out exactly what it is yet.

theduke:
> @theduke the problem is that UDP is still not really working. There's something wrong with either binding or receiving, but I haven't figured out exactly what it is yet.

Even more important to have a test then.

Arshia001:
Absolutely, as soon as someone actually fixes it. I'm out of time for now, have to move on to other priorities, but I'll keep this on my task list for later.

Arshia001:
@theduke CI should be green soon, just need an approval to get this merged in

theduke (approved):

PR #4196, opened 2023-09-01
Replace stack pool mutex with lock-free queue

Arshia001:
Alternate implementation of #3922 using a lock-free queue instead of thread locals. I benchmarked this by creating 100 threads and running a simple WASM function in a loop 30,000 times in each thread. This PR reduces the total time by ~75% on my 10-core CPU, and does not create meaningful impact on single-threaded performance. Also, since the queue is shared among threads, there is no risk of high memory consumption in highly concurrent scenarios.

### DIFF
--- a/lib/vm/src/trap/traphandlers.rs
+++ b/lib/vm/src/trap/traphandlers.rs
@@ -21,7 +21,7 @@ use std::mem;
 use std::mem::MaybeUninit;
 use std::ptr::{self, NonNull};
 use std::sync::atomic::{compiler_fence, AtomicPtr, AtomicUsize, Ordering};
-use std::sync::{Mutex, Once};
+use std::sync::Once;
 use wasmer_types::TrapCode;
 
 /// Configuration for the the runtime VM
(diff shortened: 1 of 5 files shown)


lib/vm/src/trap/traphandlers.rs
```diff
@@ -894,14 +895,12 @@ fn on_wasm_stack<F: FnOnce() -> T, T>(
...
-        static ref STACK_POOL: Mutex<Vec<DefaultStack>> = Mutex::new(vec![]);
+        static ref STACK_POOL: crossbeam_queue::SegQueue<DefaultStack> = crossbeam_queue::SegQueue::new();
     }
     let stack = STACK_POOL
-        .lock()
-        .unwrap()
         .pop()
         .unwrap_or_else(|| DefaultStack::new(stack_size).unwrap());
-    let mut stack = scopeguard::guard(stack, |stack| STACK_POOL.lock().unwrap().push(stack));
+    let mut stack = scopeguard::guard(stack, |stack| STACK_POOL.push(stack));
```
theduke, line 903: This is not directly related, but should we use an upper bound for the number of stored stack here?
For example, if you had very bursty workloads and only ran a lot of instances for a short time, while mostly remaining idle, we'd have lots of redundant memory usage.

Happy to also just create a followup issue to discuss / track this.

theduke (approved):
I think this approach is preferable, let's go with it.

lib/vm/src/trap/traphandlers.rs
Arshia001: The `ArrayQueue` actually makes this possible, so it'd be simple to implement here. However, we'd want to make something like that configurable, so it'll be bigger in scope and we can do that separately.

PR #4198, opened 2023-09-02
chore: fix unavailable document url

Kikkon:
# Description

The links currently used in the Chinese documentation are unavailable. They need to be updated to the newest valid links.


ptitSeb:
Thank you for the fix.

ptitSeb (approved):

PR #4202, opened 2023-09-05
Add support for `JoinFlags::NON_BLOCKING` to `proc_join`

g-bartoszek:
# Description
Adds support for `JoinFlags::NON_BLOCKING` to `proc_join`

### DIFF
--- a/lib/wasix/src/syscalls/wasix/proc_join.rs
+++ b/lib/wasix/src/syscalls/wasix/proc_join.rs
@@ -33,7 +33,7 @@ pub fn proc_join<M: MemorySize + 'static>(
 pub(super) fn proc_join_internal<M: MemorySize + 'static>(
     mut ctx: FunctionEnvMut<'_, WasiEnv>,
     pid_ptr: WasmPtr<OptionPid, M>,
-    _flags: JoinFlags,
+    flags: JoinFlags,
     status_ptr: WasmPtr<JoinStatus, M>,
 ) -> Result<Errno, WasiError> {
     wasi_try_ok!(WasiEnv::process_signals_and_exit(&mut ctx)?);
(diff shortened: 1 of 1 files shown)


lib/wasix/src/syscalls/wasix/proc_join.rs
```diff
@@ -182,18 +182,30 @@ pub(super) fn proc_join_internal<M: MemorySize + 'static>(
...
-                let exit_code = process.join().await.unwrap_or_else(|_| Errno::Child.into());
-                tracing::trace!(%exit_code, "triggered child join");
-                JoinStatusResult::ExitNormal(pid, exit_code)
-            })?;
-        return match res {
-            AsyncifyAction::Finish(ctx, result) => ret_result(ctx, result),
-            AsyncifyAction::Unwind => Ok(Errno::Success),
-        };
+        if flags & JoinFlags::NON_BLOCKING != JoinFlags::from_bits_preserve(0) {
+            return if let Some(status) = process.try_join() {
```
theduke, line 186: I know this was already in the code before, but return statements before large expressions are very easy to miss. 

Also, unless I'm missing something, it seems like this code will always hit one of the two branches, right?

So we can just let the value fall through and remove the `Nothing` at the end?

lib/wasix/src/syscalls/wasix/proc_join.rs
```diff
@@ -182,18 +182,30 @@ pub(super) fn proc_join_internal<M: MemorySize + 'static>(
...
-            __asyncify_with_deep_sleep::<M, _, _>(ctx, Duration::from_millis(50), async move {
-                let exit_code = process.join().await.unwrap_or_else(|_| Errno::Child.into());
-                tracing::trace!(%exit_code, "triggered child join");
-                JoinStatusResult::ExitNormal(pid, exit_code)
-            })?;
-        return match res {
-            AsyncifyAction::Finish(ctx, result) => ret_result(ctx, result),
-            AsyncifyAction::Unwind => Ok(Errno::Success),
-        };
+        if flags & JoinFlags::NON_BLOCKING != JoinFlags::from_bits_preserve(0) {
```
theduke, line 185: I think this can just be `flags.contains(JoinFlags::NON_BLOCKING)`?

theduke (changes requested):

lib/wasix/src/syscalls/wasix/proc_join.rs
g-bartoszek: Unfortunately it was nested in one more outer if statement: https://github.com/wasmerio/wasmer/pull/4202/files#diff-26f88335b0c3e9515dca019801949f26b612efb713b4182b1b23a80d6a2d6dcaR175 . I changed it to if/else to remove the "returns".

theduke (approved):
LGTM

PR #4204, opened 2023-09-05
Stability improvements

john-sharratt:
- Reduced some excessive logging
- Fixed an issue with the polling loop
- Added more socket option flags for WASM sockets


ptitSeb:
it seems to help with my networking issue with libc-wasix apps, but not completly fix the issue. I need to debug more.

theduke (approved):
lgtm

PR #4206, opened 2023-09-08
Remove StdioMode::Reserved variant

g-bartoszek:
# Description
Removes redundant `StdioMode` variant to make it match the `wasix-abi-rust` definition.


theduke (approved):
This alligns the defifnitions with the current state in the Rust wasix crate and the libc, so this is what we want.

PR #4211, opened 2023-09-11
Fixes for sockets

john-sharratt:
- Disabled DONTROUTE which was breaking networking
- Removed some potential race conditions on the handlers being registered at the wrong times
- Removed the unsafe code and replaced it with a token system
- Implemented a `poll` call for host based networking which better represents reality
- Implemented a unit test so that this does not happen again

### DIFF
--- a/lib/virtual-io/src/selector.rs
+++ b/lib/virtual-io/src/selector.rs
@@ -84,7 +84,14 @@ impl Selector {
         let token = Token(token);
         guard.lookup.insert(token, handler);
 
-        source.register(&guard.registry, token.clone(), interests)?;
+        match source.register(&guard.registry, token.clone(), interests) {
+            Ok(()) => {}
+            Err(err) if err.kind() == io::ErrorKind::AlreadyExists => {
+                source.deregister(&guard.registry).ok();
+                source.register(&guard.registry, token.clone(), interests)?;
+            }
+            Err(err) => return Err(err),
+        };
 
         Ok(token)
     }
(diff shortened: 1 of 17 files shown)


lib/virtual-io/src/selector.rs
```diff
@@ -1,51 +1,62 @@
...
+    token_close: Token,
     inner: Mutex<EngineInner>,
-    #[derivative(Debug = "ignore")]
-    pub(crate) registry: mio::Registry,
-    pub(crate) tx_drop: Mutex<Sender<Token>>,
     closer: mio::Waker,
 }
 
+use lazy_static::lazy_static;
+lazy_static! {
```
theduke, line 30: Can we use `std::cell::OnceCell` here to avoid an extra dependency?
I know it's a bit awkward that `LazyCell` isn't stable yet, but alas...

lib/virtual-io/src/selector.rs
```diff
@@ -1,51 +1,62 @@
...
+    static ref GLOBAL_SELECTOR: Arc<Selector> = Selector::new_internal();
```
theduke, line 31: Can we document why there is a global selector, and what it is used for?

Can this be a contention point in highly concurrent scenarios? (eg Edge)

lib/virtual-io/src/selector.rs
john-sharratt: deleted it

lib/virtual-io/src/selector.rs
john-sharratt: deleted it

ptitSeb (approved):

PR #4220, opened 2023-09-18
Ability to detect a tainted instance

john-sharratt:
This needs to go in first
https://github.com/wasmerio/wasmer/pull/4211


theduke (approved):

Michael-F-Bryan (changes requested):
You'll generally want to share a `Runtime` to reuse the same thread pool, any connections that may still be held open by the HTTP client, in-memory caches, and so on.

With that in mind, I think we need to include some sort of ID (or a reference to the `wasmer::Instance`) in `on_taint()` that allows the runtime to know *which* instance became tainted. Otherwise we've just got a *"something went wrong"* callback and the only way to know which instance became tainted is through context, which is pretty brittle.

syrusakbary:
We'll keep this open for now, but ideally we should use proper signal propagation to implement this properly (based on @theduke feedback)

theduke (approved):

PR #4223, opened 2023-09-20
Allow packages to have commands that use a dependency's atom

Michael-F-Bryan:
This is the `wasmer` counterpart to https://github.com/wasmerio/pirita/issues/172 and https://github.com/wasmerio/wasmer-toml/pull/22, adding support for commands that can use atoms from a dependency.

### DIFF
--- a/tests/integration/cli/tests/packages/js-script/wasmer.toml
+++ b/tests/integration/cli/tests/packages/js-script/wasmer.toml
@@ -1,0 +1,18 @@
+[package]
+name = "wasmer/js-script"
+version = "0.0.0"
+description = "A package which uses quickjs to run a JavaScript program"
+
+[dependencies]
+"saghul/quickjs" = "0.0.3"
+
+[[command]]
+name = "main"
+module = "saghul/quickjs:quickjs"
+runner = "wasi"
+
+[command.annotations.wasi]
+main_args = ["/src/index.js"]
+
+[fs]
+"/src" = "./src"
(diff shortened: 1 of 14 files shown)


tests/integration/cli/tests/packages/js-script/wasmer.toml
```diff
@@ -12,7 +12,7 @@ module = "saghul/quickjs:quickjs"
 runner = "wasi"
 
 [command.annotations.wasi]
-main_args = ["/src/index.js"]
+main-args = ["/src/index.js"]
```
ptitSeb, line 15: can we have an alias or an error when using the wrong name?

ptitSeb (approved):

tests/integration/cli/tests/packages/js-script/wasmer.toml
Michael-F-Bryan: I was thinking of adding [`#[serde(deny_unknown_fields)]`](https://serde.rs/container-attrs.html#deny_unknown_fields) to trigger an error when we try to deserialize an unknown field, but then we lose forward-compatibility.

For example, a newer version of wasmer introduces a `enable-threads` field and now older versions of wasmer will error out with `unknown field, "enable-threads", while deserializing WASI metadata` instead of silently ignoring it.

There is also the `#[serde(alias = "...")]` attribute, but it means I'd need to add snake_case aliases to every field in every metadata type, which isn't very practical 😞

PR #4224, opened 2023-09-21
Use write instead of send for pipe, to accomodate socketpair

ptitSeb:
Use write instead of send for pipe, to accomodate socketpair.

Because C socketpair use fd_pipe to create the pair, waist pipe will need to handle `sock_send`

### DIFF
--- a/lib/wasix/src/syscalls/wasix/sock_send.rs
+++ b/lib/wasix/src/syscalls/wasix/sock_send.rs
@@ -27,12 +27,8 @@ pub fn sock_send<M: MemorySize>(
 ) -> Result<Errno, WasiError> {
     let env = ctx.data();
     let fd_entry = env.state.fs.get_fd(sock).unwrap();
-    let inode = fd_entry.inode.clone();
-    let guard = inode.read();
-    let use_write = match guard.deref() {
-        Kind::Pipe { .. } => true,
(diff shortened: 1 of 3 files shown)


lib/wasix/src/syscalls/wasix/sock_send.rs
```diff
@@ -25,7 +25,20 @@ pub fn sock_send<M: MemorySize>(
...
 ) -> Result<Errno, WasiError> {
-    sock_send_internal(ctx, sock, si_data, si_data_len, si_flags, ret_data_len)
+    let env = ctx.data();
+    let fd_entry = env.state.fs.get_fd(sock).unwrap();
+    let inode = fd_entry.inode.clone();
+    let guard = inode.read();
+    let use_write = match guard.deref() {
+        Kind::Pipe { .. } => true,
+        _ => false,
+    };
```
syrusakbary, line 35: You can probably use `matches!(guard.deref(), Kind::Pipe { .. })`. It's simpler :)

Michael-F-Bryan (approved):

lib/wasix/src/syscalls/wasix/sock_send.rs
ptitSeb: done

PR #4242, opened 2023-10-05
fix(wasi): Console: use package default args if no custom args specified

theduke:
Use the CLI args provided by the command if no custom args are specified.

This is already done by the CLI, but was not adopted in wasix.

### DIFF
--- a/lib/wasix/src/os/console/mod.rs
+++ b/lib/wasix/src/os/console/mod.rs
@@ -205,7 +205,34 @@ impl Console {
             }
         };
 
-        let wasi_opts = webc::metadata::annotations::Wasi::new(prog);
+        let command = if let Some(entrypoint) = pkg.entrypoint_cmd.as_deref() {
+            pkg.get_command(entrypoint).unwrap()
+        } else {
+            match pkg.commands.as_slice() {
+                [] => {
+                    return Err(SpawnError::Other(
+                        anyhow::anyhow!("The WEBC file doesn't contain any executable commands")
+                            .into(),
+                    ))
+                }
+                [one] => one,
+                [first, ..] => first,
+            }
+        };
+
+        let mut wasi_opts = command
+            .metadata()
+            .wasi()
+            .ok()
+            .and_then(|x| x)
+            .unwrap_or_else(|| webc::metadata::annotations::Wasi::new(prog));
+
+        // Prevent adding default args if custom args are specified.
+        if !args.is_empty() {
+            if let Some(wargs) = &mut wasi_opts.main_args {
+                wargs.clear();
+            }
+        }
 
         let root_fs = RootFileSystemBuilder::new()
             .with_tty(Box::new(CombineFile::new(


lib/wasix/src/os/console/mod.rs
```diff
@@ -205,7 +205,34 @@ impl Console {
             }
         };
 
-        let wasi_opts = webc::metadata::annotations::Wasi::new(prog);
+        let command = if let Some(entrypoint) = pkg.entrypoint_cmd.as_deref() {
+            pkg.get_command(entrypoint).unwrap()
```
Michael-F-Bryan, line 209: Can we return an error here?

It'd be annoying if we start panicking on Edge because an oversight in `wasmer publish` meant we publish packages with a bad `package.entrypoint` field.

lib/wasix/src/os/console/mod.rs
```diff
@@ -205,7 +205,34 @@ impl Console {
...
+        } else {
+            match pkg.commands.as_slice() {
+                [] => {
+                    return Err(SpawnError::Other(
+                        anyhow::anyhow!("The WEBC file doesn't contain any executable commands")
+                            .into(),
+                    ))
+                }
+                [one] => one,
+                [first, ..] => first,
```
Michael-F-Bryan, line 219: We should error out here because the situation is ambiguous. 

That's what we do [in the `webc` crate](https://github.com/wasmerio/pirita/blob/f99de1f75309f8f43e3b4585d3a6601bc5225ac3/crates/webc/src/wasmer_package/manifest.rs#L655-L668) when loading a package directly from disk (i.e. `wasmer run .` and tarball conversion).

```rs
/// Infer the package's entrypoint.
fn entrypoint(manifest: &wasmer_toml::Manifest) -> Option<String> {
    if let Some(entrypoint) = &manifest.package.entrypoint {
        return Some(entrypoint.clone());
    }

    if let [only_command] = manifest.commands.as_slice() {
        // For convenience (and to stay compatible with old docs), if there is
        // only one command we'll use that as the entrypoint
        return Some(only_command.get_name().to_string());
    }

    None
}
```

lib/wasix/src/os/console/mod.rs
```diff
@@ -205,7 +205,34 @@ impl Console {
...
+            .ok()
+            .and_then(|x| x)
+            .unwrap_or_else(|| webc::metadata::annotations::Wasi::new(prog));
+
+        // Prevent adding default args if custom args are specified.
+        if !args.is_empty() {
+            if let Some(wargs) = &mut wasi_opts.main_args {
+                wargs.clear();
+            }
+        }
```
Michael-F-Bryan, line 235: FYI - This doesn't line up with what the `wasmer` CLI does.

The `wasmer` CLI will start with the arguments provided in `[command.annotations.wasi]` and **append** the arguments passed in via the command-line.

@theduke I'll let you decide whether you are okay with having different behaviour - just thought I'd point out the possible inconsistency.

Michael-F-Bryan (changes requested):

lib/wasix/src/os/console/mod.rs
theduke: I assumed that it was impossible for the entrypoint to not exist.

lib/wasix/src/os/console/mod.rs
theduke: We can't error out here because that's what Edge is currently doing and expects.

lib/wasix/src/os/console/mod.rs
theduke: Yeah we briefly talked about this already.
We need some solution here IMO, because often (usually) you'll want to override the args, not amend them.

lib/wasix/src/os/console/mod.rs
theduke: I agree that ideally we would error out here with a misconfiguration error, but for now we need to validate that this won't break anything.

theduke:
The console is now a legacy feature and doens't have to be updated any longer.

PR #4247, opened 2023-10-10
Add support for publishing private packages

Michael-F-Bryan:
This updates the `wasmer publish` command to allow users to publish private packages. When a package is private, you will only be able to access it if you have contributor access (e.g. because you are the owner, have been added as a contributor, or are part of a namespace that owns the package).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -153,9 +153,9 @@ dependencies = [
 
 [[package]]
 name = "arbitrary"
-version = "1.3.1"
+version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a2e1373abdaa212b704512ec2bd8b26bd0b7d5c3f70117411a5d9a451383c859"
+checksum = "7d5a26814d8dcb93b0e5a0ff3c6d80a8843bafb21b39e8e18a6f05471870e110"
 dependencies = [
  "derive_arbitrary",
 ]
@@ -275,9 +275,9 @@ checksum = "35636a1494ede3b646cc98f74f8e62c773a38a659ebc777a2cf26b9b74171df9"
 
 [[package]]
 name = "basic-toml"
-version = "0.1.6"
+version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "24c12265665aebaa236af9bbe266681bcc9c5666192119e3d8335cf083aca26f"
+checksum = "2f2139706359229bfa8f19142ac1155b4b80beafb7a60471ac5dd109d4a19778"
 dependencies = [
  "serde",
 ]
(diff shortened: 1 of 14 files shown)


theduke (commented):
One concern I have here is that we are duplicating the responsibility for marking a package as private.

Is it the private setting in wasmer.toml? 
Is it the CLI flag and the GraphQL argument?

What happens when the GraphQL specifies private, but not the package manifest?
Will the backend alter the manifest before persisting it?

What happens on a re-publish when you forget the private flag?
Does the API return an error then because the package is private?
Does it just accept it?

It would be cleaner, I think, if the manifest was the source of truth.
An additional `--private` flag or the CLI arg could make sure the package is private and error out. Similar for the GraphQL API.

Thoughts, @Michael-F-Bryan , @ayys ?

ayys:
I agree that wasmer.toml should be the source of truth,

> What happens when the GraphQL specifies private, but not the package manifest?

the graphql `private` field and CLI argument `--private` should still exist, but they will be validated against wasmer.toml. If the config file does not have `private = true` set, then the backend should throw an error.



> What happens on a re-publish when you forget the private flag?

For visibility, right now piblic packages cannot be made private, but private packages can be made public. So if you remove the `private = true` field from the config, the package should become public (the CLI should confirm this change with the user?)


>  An additional --private flag or the CLI arg could make sure the package is set to private in the manifest and error out. Similar for the GraphQL API.

I agree that the `private` field for the CLI and for GraphQL should be there to make sure wasmer.toml has `private = true` set and then use that..

Michael-F-Bryan:
@theduke this is my understanding. @ayys feel free to correct me if I'm wrong.

> Is it the private setting in wasmer.toml?

Yep. We added a `private` flag to the `[package]` table in https://github.com/wasmerio/wasmer-toml/pull/31 (released as `wasmer-toml v0.9.1`) and the `webc` crate also added support for it in https://github.com/wasmerio/pirita/pull/186 (released as `webc v5.6.0`).

> It would be cleaner, I think, if the manifest was the source of truth.

100% agreed. It's traceable and immediately obvious to anyone consuming the packages.

> What happens when the GraphQL specifies private, but not the package manifest?

If there is disagreement between the manifest and the GraphQL mutation, the backend should return an error and publishing will fail.

> Will the backend alter the manifest before persisting it?

To the best of my knowledge, no. Altering the manifest that someone has published is something we should try to avoid as much as possible. 

When I publish something, I want to be confident that my users get exactly what I published.

> What happens on a re-publish when you forget the private flag?
> Does the API return an error then because the package is private?
> Does it just accept it?

@ayys probably knows better than me, but I think this would silently make the package public.

> An additional --private flag or the CLI arg could make sure the package is set to private in the manifest and error out. Similar for the GraphQL API.

I'd like to avoid a `--private` flag if we can. It feels like we could confuse users by having a `--private` flag which doesn't actually mark the package as private, but instead acts as a sort of validation. It also leads to ambiguities over whether the `wasmer.toml` or `--private` is the way you should mark a package as private, a lot of people won't read docs and will instead check `wasmer publish --help`, so they'll probably see the `--private` flag first.

PR #4250, opened 2023-10-11
replace mach with mach2

### LINKED ISSUE #4222: Is it possible to change the mach dependency
### Summary

When using `wasmer` version 4.2 we have a warning when running a `cargo audit` on a dependency that you have which is `mach`. The warning : https://rustsec.org/advisories/RUSTSEC-2020-0168 . They recommend to use an other crate called `mach2` which is more maintained. What do you think about this change ?

Arshia001:
Closes #4222.

Note, since `mach` was also being used by `region`, I had to fork and patch `region` too. The patch can be removed once the changes are merged upstream and released.

### DIFF
--- a/lib/cli/src/commands/app/deploy.rs
+++ b/lib/cli/src/commands/app/deploy.rs
@@ -678,7 +678,7 @@ pub async fn wait_app(
                                 eprintln!();
                             }
                             if !(res.status().is_success() || res.status().is_redirection()) {
-                                eprintln!("The app version was deployed correctly, but fails with a non-success status code of {}", res.status())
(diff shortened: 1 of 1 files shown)


syrusakbary (changes requested):
This is great, but we can't merge this upstream until the dependencies are in crates.io (as otherwise we can't publish wasmer to crates.io)

Arshia001:
Yes, I realised a few minutes too late that that this is not a wasix app 😄

Gonna have to create a PR in region and hope they accept and publish a new version.

theduke:
@Arshia001 what's the status here?

Arshia001:
@theduke we still need to open a PR in `region`.

syrusakbary:
Solution: create region2 create with the fix

PR #4263, opened 2023-10-20
WASI journal and stateful persistence

john-sharratt:
# WASM Journal Functionality

Wasmer now supports journals for the state of a WASM process. This gives the ability to persist changes made to the temporary file system and to save and store snapshots of the running process.

The journal file is a linear history of events that occurred when the process was running that if replayed will bring the process made to a discrete and deterministic state.

Journal files can be concatenated, compacted and filtered to change the discrete state.

These journals are maintained in a consistent and durable way thus ensuring that failures of the system while the process is running does not corrupt the journal.

# Snapshot Triggers

The journal will record state changes to the sandbox built around the WASM process as
it runs however it may be important to certain use-cases to take explicit snapshot
restoration points in the journal at key points that make sense.

When a snapshot is triggered all the running threads of the process are paused and
the state of the WASM memory and thread stacks are recorded into the journal so that
they can be restored.

In order to use the snapshot functionality the WASM process must be compiled with the `asyncify` modifications, this can be done using the `wasm-opt` tool.

Note: If a process does not have the `asyncify` modifications you can still use the journal functionality for recording the file system and WASM memory state however the stacks of the threads will be omitted meaning a restoration will
restart the main thread.

Various triggers are possible that will cause a snapshot to be taken at a specific point in time, these are as follows:

## On Idle

Triggered when all the threads in the process go into an idle state. This trigger is useful to take snapshots at convenient moments without causing unnecessary overhead.

For processes that have TTY/STDIN input this is particularly useful.

## On FirstListen

Triggered when a listen syscall is invoked on a socket. This can be an important milestone to take a snapshot when one wants to speed up the boot time of a WASM process up to the moment where it is ready to accept requests.
    
## On FirstStdin

Triggered when the process reads stdin for the first time. This can be useful to speed up the boot time of a WASM process.

## On FirstEnviron

Triggered when the process reads an environment variable for the first time. This can be useful to speed up the boot time of a CGI WASM process which reads the environment variables to parse the request that it must execute.

## On Timer Interval

Triggered periodically based on a timer (default 10 seconds) which can be specified using the `journal-interval` option. This can be useful for asynchronous replication of a WASM process from one machine to another with a particular lag latency.

## On Sigint (Ctrl+C)

Issued if the user sends an interrupt signal (Ctrl + C).

## On Sigalrm

Alarm clock signal (used for timers)
(see `man alarm`)

## On Sigtstp

The SIGTSTP signal is sent to a process by its controlling terminal to request it to stop temporarily. It is commonly initiated by the user pressing Ctrl-Z.

# On Sigstop

The SIGSTOP signal instructs the operating system to stop a process for later resumption

# On Non Deterministic Call

When a non-deterministic call is made from WASM process to the outside world (i.e. it reaches out of the sandbox)

# Limitations

- The WASM process that wish to record the state of the threads must have had the `asyncify`  post processing step applied to the binary (see `wasm-opt`).
- Taking a snapshot can consume large amounts of memory while its processing.
- Snapshots are not instant and have overhead when generating.
- The layout of the memory must be known by the runtime in order to take snapshots.

# Design

On startup if the restore journal file is specified then the runtime will restore the state of the WASM process by reading and processing the log entries in the snapshot journal. This restoration will bring the memory and the thread stacks back to a previous point in time and then resume all the threads.

When a trigger occurs a new journal will be taken of the WASM process which will take the following steps:

1. Pause all threads
2. Capture the stack of each thread
3. Write the thread state to the journal
4. Write the memory (excluding stacks) to the journal
5. Resume execution.

The implementation is currently able to save and restore the following:

- WASM Memory
- Stack memory
- Call stack
- Open sockets
- Open files
- Terminal text

# Journal Capturer Implementations

## Log File Journal

Writes the log events to a linear log file on the local file system as they are received by the trait. Log files can be concatenated together to make larger log files.

## Unsupported Journal

The default implementation does not support snapshots and will error out if an attempt is made to send it events. Using the unsupported capturer as a restoration point will restore nothing but will not error out.

## Compacting Journal

Deduplicates memory and stacks to reduce the number of volume of log events sent to its inner capturer. Compacting the events occurs in line as the events are generated

## Filtered Journal

Filters out a specific set of log events and drops the rest, this capturer can be useful for restoring to a previous call point but
retaining the memory changes (e.g. WCGI runner).


lib/cli/src/commands/run/wasi.rs
```diff
@@ -105,6 +107,34 @@ pub struct Wasi {
...
+    /// as well.
+    #[cfg(feature = "snapshooter")]
+    #[clap(long = "snapshot-on")]
+    pub snapshot_on: Vec<SnapshotTrigger>,
+
+    /// Time in seconds between taking snapshots of the process and dumping
+    /// them to the snapshot file.
+    #[cfg(feature = "snapshooter")]
+    #[clap(long = "snapshot-period")]
+    pub snapshot_period: Option<u64>,
```
syrusakbary, line 130: Why not making the dumping instant? (why would be wanted to be delayed?)

lib/cli/src/commands/run/wasi.rs
```diff
@@ -116,6 +146,47 @@ pub struct Wasi {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SnapshotTrigger {
+    /// Triggered when all the threads in the process goes idle
+    Idle,
+    /// Triggered when a listen syscall is invoked on a socket
+    Listen,
+    /// Triggered when the process reads stdin for the first time
+    Stdin,
+    /// Triggered periodically (default 10 seconds) which can be specified using the `snapshot-period` option
+    Periodic,
```
syrusakbary, line 160: I'm a bit confused by this. On the snapshot-period CLI argument it puts is as a delay between process -> dump data. But here it says to be a time period repeated over time (which I don't understand)

lib/cli/src/commands/run/wasi.rs
```diff
@@ -116,6 +146,47 @@ pub struct Wasi {
     pub deny_multiple_wasi_versions: bool,
 }
 
+/// Various triggers that will cause the runtime to take snapshot
+/// of the WASM state and store it in the snapshot file.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum SnapshotTrigger {
```
syrusakbary, line 152: I'd add another option: `NonDeterministicCall`. Thoughts?

lib/wasix/src/snapshot/log_file.rs
```diff
@@ -0,0 +1,261 @@
...
+/// means that forwards and backwards compatibility must be dealt with
+/// carefully.
+///
+/// When opening an existing journal file that was previously saved
+/// then new entries will be added to the end regardless of if
+/// its been read.
+///
+/// The logfile snapshooter uses a 64bit number as a entry encoding
+/// delimiter.
+pub struct LogFileSnapShooter {
```
syrusakbary, line 177: I'm not very fan of calling something "SnapShooter", I think just `SnapshotCapturer` or `Snapshot` would be better

lib/cli/Cargo.toml
```diff
@@ -160,9 +160,11 @@ default = [
   "wat",
   "wast",
   "compiler",
+  "snapshooter",
```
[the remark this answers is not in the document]
john-sharratt: Renamed here:
https://github.com/wasmerio/wasmer/pull/4263/commits/8f11ac78b00d3f07b774e776d20e54221dc9e33f

lib/cli/src/commands/run/wasi.rs
john-sharratt: It is actually instant already however this parameter allows for a timer trigger to be added and for the interval to be set. The idea of a timer trigger is so that those that want a real-time backup of a process that can be restored somewhere else can use this functionality.

Renamed it from `period` to `timer` to make it less ambiguous
https://github.com/wasmerio/wasmer/pull/4263/commits/8f11ac78b00d3f07b774e776d20e54221dc9e33f

lib/cli/src/commands/run/wasi.rs
john-sharratt: Added this as a placeholder and understand what you are doing with this, it makes sense.
https://github.com/wasmerio/wasmer/pull/4263/commits/8f11ac78b00d3f07b774e776d20e54221dc9e33f

lib/wasix/src/snapshot/log_file.rs
john-sharratt: Renamed here:
https://github.com/wasmerio/wasmer/pull/4263/commits/8f11ac78b00d3f07b774e776d20e54221dc9e33f

lib/cli/src/commands/run/wasi.rs
john-sharratt: Renamed this from periodic to timer to make it more clear what its is
https://github.com/wasmerio/wasmer/pull/4263/commits/8f11ac78b00d3f07b774e776d20e54221dc9e33f

john-sharratt:
Next up...
Add fake sockets
Compactor Refactor 
Adds heaps of integration tests

Tests need to be created for all the scenarios
Using an incorrect journal on a different WASM process
Basic file operations are preserved
Listening socket is preserved
Epoll is preserved
Static memory is preserved
Process exiting cleans up file descriptors
Process exiting sets memory to zero

MaratBR:
> Taking a snapshot can consume large amounts of memory while its processing.

Quick question if you don't mind.
What amount is "large"? Assuming I have a WASM runtime that uses 10MB of memory how much memory taking a snapshot will take? I assume it will be something like this right?
```
total memory required  = (runtime virtual RAM + file system size) * coefficient
```

john-sharratt:
> > Taking a snapshot can consume large amounts of memory while its processing.
> 
> Quick question if you don't mind.
> What amount is "large"? Assuming I have a WASM runtime that uses 10MB of memory how much memory taking a snapshot will take? I assume it will be something like this right?
> ```
> total memory required  = (runtime virtual RAM + file system size) * coefficient

For taking snapshots its the amount of activd memory which is equal to the size of the active memory buffer (the memory that grows not the program itself) behind your wasm app plus the active thread stack size. For normal apps this will be reasonable (few MB's) however for large databases this could be quite big with indexes.

When the logs are compacting it will store a lookup index while its removing duplicates and it will write a new file to replace the old file. So you need enough memory proportional to the number of events plus enough disk space for a copy of the log file before it deletes the old one.

Fairly reasonable constraints I think but ones you need to keep in mind

theduke (approved):
This is way to much to reasonably review in detail, but as iterated before the basics look good.

syrusakbary (commented):
The API changes (to the `wasmer` crate look reasonable to me).

I only have one thought: it would be great if we can separate the abstraction from the implementation, that is: keeping the Journal base abstraction outside of WASIX. I think that will help us immensely in the future refactor

lib/wasix/src/journal/concrete/archived.rs
```diff
@@ -0,0 +1,3558 @@
+use lz4_flex::block::{compress_prepend_size, decompress_size_prepended};
+use num_enum::{IntoPrimitive, TryFromPrimitive};
+use rkyv::ser::{ScratchSpace, Serializer};
+use rkyv::{Archive, CheckBytes, Deserialize as RkyvDeserialize, Serialize as RkyvSerialize};
+use std::borrow::Cow;
+use std::time::SystemTime;
+use virtual_net::{Duration, IpAddr, IpCidr, Ipv4Addr, Ipv6Addr, SocketAddr, StreamSecurity};
```
theduke, line 7: Can we use custom types for everthing in the journal?

Depending on types from separate contexts (virtual_net, SystemTime, wasmer_wasix_types) can make it easy to break the format accidentally, while we need this to be perma-stable.

As @syrusakbary mentioned, moving the journaling out to a separate crate (like `wasix_journal`) would help to ensure this.

theduke (changes requested):

lib/wasix/src/journal/concrete/archived.rs
syrusakbary: What I was mentioning was to have a `wasmer_journal` crate, independent from WASIX, and then WASIX can implement the types and such

theduke: That doesn't make much sense to me.

This is a journal for WASIX syscalls, it's inherently tied to what syscalls WASIX has.

That doesn't mean it needs to be in the WASIX crate though.

theduke: I guess you could argue that it's a syscalll journal, which happens to overlap with WASIX, but it doesn't really relate to Wasmer (the runtime).

lib/cli/src/commands/run/wasi.rs
```diff
@@ -105,6 +107,35 @@ pub struct Wasi {
...
+    /// that `--snapshot-on timer` has also been specified.
+    #[cfg(feature = "snapshot")]
+    #[clap(long = "snapshot-timer")]
+    pub snapshot_timer: Option<u64>,
+
+    /// When specified, the runtime will restore a previous snapshot
+    /// using the supplied file.
+    #[cfg(feature = "snapshot")]
+    #[clap(long = "resume-from")]
+    pub resume_from: Option<PathBuf>,
```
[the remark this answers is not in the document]
john-sharratt: This doesn't exist anymore anyway, instead what we did was modify the CLI to use the `--journal` parameter which where former arguments are read-only applied to the state and the later (or last) journal is first applied then appended to. If one wants to do the use case of a delta then only need specify an empty (or non-existent file) as the last journal and it will all stay immutable which is what this argument above you have referenced was designed to cover.

lib/wasix/src/snapshot/effector.rs
```diff
@@ -0,0 +1,143 @@
...
+
+use crate::{
+    mem_error_to_wasi, os::task::process::WasiProcessInner, syscalls::__asyncify_light,
+    utils::map_snapshot_err, WasiEnv, WasiError, WasiThreadId,
+};
+
+use super::*;
+
+#[derive(Debug, Clone)]
+pub struct SnapshotEffector {}
```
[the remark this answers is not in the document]
john-sharratt: It's now called a JournalEffector but regardless the point you make remains valid - added comments to clarify what it is and that the intent is to remove this in the upcoming syscall refactor
```
[dcgi eb96a9b95f] Added comments on the JournalEffector
 1 file changed, 10 insertions(+)
 ```

File: lib/wasix/src/snapshot/filter.rs [hunk 9]
```diff
@@ -0,0 +1,152 @@
+use futures::future::BoxFuture;
+
+use super::*;
+
+/// Filters out a specific set of log events and drops the rest, this
+/// capturer can be useful for restoring to a previous call point but
+/// retaining the memory changes (e.g. WCGI runner).
+pub struct FilteredSnapshotCapturer {
+    inner: Box<DynSnapshotCapturer>,
+    filter_memory: bool,
```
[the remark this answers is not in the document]
john-sharratt: This evolved from its earlier implementation as the design was built out. It's no longer called `FilteredSnapshotCapturer` and now called `FilteredJournal`. The responsibility was scoped down to purely filtering an input and output events.

Still a builder and config type on this filter is a good idea, so I've added one. 

```
john@AlienWorld:/prog/wasmer$ git commit -a -m "Split out the config responsibility from the FilteredJournal"
[dcgi fb6d4d639d] Split out the config responsibility from the FilteredJournal
 2 files changed, 117 insertions(+), 79 deletions(-)
john@AlienWorld:/prog/wasmer$
```

lib/wasix/src/snapshot/log_file.rs
```diff
@@ -0,0 +1,261 @@
...
+            file: tokio::fs::File::from_std(file),
+            at_end: false,
+        };
+        Ok(Self {
+            state: tokio::sync::Mutex::new(state),
+        })
+    }
+}
+
+#[async_trait::async_trait]
```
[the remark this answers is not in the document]
john-sharratt: This has already been removed

Makefile
```diff
@@ -439,10 +439,10 @@ test-build-docs-rs:
 		fi; \
 		printf "*** Building doc for package with manifest $$manifest_path ***\n\n"; \
 		if [ -z "$$features" ]; then \
-			RUSTDOCFLAGS="--cfg=docsrs" $(CARGO_BINARY) +nightly doc $(CARGO_TARGET_FLAG) --manifest-path "$$manifest_path" --locked || exit 1; \
+			RUSTDOCFLAGS="--cfg=docsrs" $(CARGO_BINARY) +nightly doc $(CARGO_TARGET_FLAG) --manifest-path "$$manifest_path" --exclude tokio --locked || exit 1; \
```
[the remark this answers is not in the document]
john-sharratt: Added a commment for this...
```
[dcgi 58a89419db] Added a comment on the excluding of tokio from the docs build
 1 file changed, 11 insertions(+)
```

lib/wasix/src/journal/concrete/archived.rs
john-sharratt: I would have to replicate all the WASIX enums to do this (and there are lots of them!).... including the infamous Errno enum - that's some serious boiler plate guys, we sure we really want to do this?

john-sharratt: Actually there might be a way to avoid that, let me give it a shot

john-sharratt: https://github.com/wasmerio/wasmer/pull/4263/commits/7c1b68df6b11c47c825ab2b01202a336c3e3370c****

john-sharratt: That commit moves out the crate however completely decoupling that crate from all other types will create a large amount of refactoring, needless boilerplate and duplication of code. Especially around enums and flags, etc...

@syrusakbary @theduke Are you really sure you want to do this? All that extra boilerplate looks like premature abstraction to me and really increases the code complexity while the actual on disk format (`ArchivedJournalEntry`) is already abstracted from those types and hence protected from changes that could impact it.

syrusakbary: I think the commit that @john-sharratt did is addressing what I commented perfectly. The PR looks good to me now

syrusakbary (approved):

lib/wasix/src/journal/concrete/archived.rs
theduke: I know it's a lot of boilerplate, *but*: we really need the journal to be perma-stable and to not have any breaking changes.

The types from `virtual-net` should definitely be duplicated, but those are not that many. 

I guess the WASIX enums and partially the types also more or less need to be perma-stable, but things will get more complicated if we add WASI preview 2 support, and if we make any kinds of changes to the representation of WASI types. 

My intuition is to go with all separate types.

PR #4287, opened 2023-10-31
feat(cli): Add package commands

theduke:
Adds new CLI commands for working with packages:

* package download: download package from registry
* package build: build a package from a wasmer.toml manifest
* package unpack: extract package contents to a directory


lib/cli/src/commands/package/build.rs
```diff
@@ -0,0 +1,131 @@
...
+                );
+            }
+            manifest_path
+        };
+
+        let pkg = webc::wasmer_package::Package::from_manifest(manifest_path)?;
+
+        let manifest = pkg
+            .manifest()
+            .package_annotation::<webc::metadata::annotations::Wapm>("wapm")
```
Michael-F-Bryan, line 56: Prefer to use the helper methods where possible. That way we don't use magic strings (although you also have access to `webc::metadata:annotations::Wapm::KEY`) and there's no possibility of deserializing to the wrong type.

```suggestion
            .wapm()
```

lib/cli/src/commands/package/download.rs
```diff
@@ -0,0 +1,177 @@
...
+
+use anyhow::Context;
+use futures::StreamExt;
+use tokio::io::AsyncWriteExt;
+use wasmer_registry::wasmer_env::WasmerEnv;
+use wasmer_wasix::runtime::resolver::PackageSpecifier;
+
+/// Download a package from the registry.
+#[derive(clap::Parser, Debug)]
+pub struct CmdPackageDownload {
```
Michael-F-Bryan, line 11: This should be called `PackageDownload` or `DownloadPackage`.

lib/cli/src/commands/package/download.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        let client = reqwest::Client::new();
+        let mut b = client
+            .get(url.clone())
+            .header(http::header::ACCEPT, "application/webc");
+        if let Some(token) = token {
+            b = b.header(http::header::AUTHORIZATION, format!("Bearer {token}"));
+        };
+
+        eprintln!("Downloading package from '{url}'...");
```
Michael-F-Bryan, line 79: Instead of printing progress messages directly to stderr, can you make a spinner using `indicatif::ProgressBar`?

lib/cli/src/commands/package/download.rs
```diff
@@ -0,0 +1,177 @@
...
+
+        let tmp_name = url
+            .as_str()
+            .chars()
+            .map(|x| match x {
+                'a'..='z' | 'A'..='Z' | '0'..='9' | '-' => x,
+                _ => '_',
+            })
+            .collect::<String>();
+        let tmp_path = std::env::temp_dir().join(format!("wasmer-download-{}-{}", tmp_name, time));
```
Michael-F-Bryan, line 100: Let's use the `tempfile` crate instead of rolling our own stuff. The `tempfile::NamedTempFile` type has a `persist()` method for saving the temporary file to the final destination and you can always do `tokio::fs::File::from(temp.as_file().try_clone()?)` if you want to write to it asynchronously.

lib/cli/src/commands/package/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+/// Package related commands.
+#[derive(clap::Subcommand, Debug)]
+// Allowing missing_docs because the comment would override the doc comment on
+// the command struct.
+#[allow(missing_docs)]
+pub enum CmdPackage {
+    Download(CmdPackageDownload),
+    Unpack(unpack::CmdPackageUnpack),
+    Build(build::CmdPackageBuild),
+}
```
Michael-F-Bryan, line 16: We don't need the `Cmd` prefix.

```suggestion
pub enum Package {
    Download(PackageDownload),
    Unpack(unpack::PackageUnpack),
    Build(build::PackageBuild),
}
```

lib/cli/src/commands/package/unpack.rs
```diff
@@ -0,0 +1,90 @@
+use std::path::PathBuf;
+
+use anyhow::Context;
+
+/// Extract contents of a container to a directory.
+#[derive(clap::Parser, Debug)]
+pub struct CmdPackageUnpack {
```
Michael-F-Bryan, line 7: Rename to `PackageUnpack` or `UnpackPackage` (which sounds more natural).

lib/cli/src/commands/package/unpack.rs
```diff
@@ -0,0 +1,90 @@
...
+    #[clap(long)]
+    overwrite: bool,
+
+    /// Path to the package.
+    package_path: PathBuf,
+}
+
+impl CmdPackageUnpack {
+    pub(crate) fn execute(&self) -> Result<(), anyhow::Error> {
+        eprintln!("Unpacking...");
```
Michael-F-Bryan, line 22: We should be able to use `indicatif` here. It'll give us nice things like a spinner and automatically hide output if stderr isn't a TTY.

lib/cli/src/commands/package/unpack.rs
```diff
@@ -0,0 +1,90 @@
...
+        })?;
+
+        let outdir = &self.out_dir;
+        std::fs::create_dir_all(outdir)
+            .with_context(|| format!("could not create output directory '{}'", outdir.display()))?;
+
+        pkg.unpack(outdir, self.overwrite)
+            .with_context(|| "could not extract package".to_string())?;
+
+        eprintln!("Extracted package contents to '{}'", self.out_dir.display());
```
Michael-F-Bryan, line 38: You can use `indicatif::ProgressBar`'s `finish_with_message()` for this bit.

lib/registry/src/wasmer_env.rs
```diff
@@ -42,6 +42,22 @@ impl WasmerEnv {
...
+            .host_str()
+            .context("url has no host")?
+            .strip_prefix("registry.")
+            .context("could not derive registry public url")?
+            .to_string();
+        url.set_host(Some(&domain))
+            .context("could not derive registry public url")?;
+
+        Ok(url)
+    }
```
Michael-F-Bryan, line 59: This feels a bit hacky. We should really be [asking the registry](https://registry.wasmer.io/graphql?query=query%20%7B%0A%20%20info%20%7B%0A%20%20%20%20defaultFrontend%0A%20%20%7D%0A%7D) for the frontend URL.

```graphql
query {
  info {
    defaultFrontend
  }
}
```

Michael-F-Bryan (changes requested):
Thanks for adding this to the `wasmer` CLI. I've been wanting to do it for ages but never got around to it.

I've added some suggestions around coding style so it matches up with the rest of the CLI code.

lib/cli/src/commands/package/download.rs
theduke: Not doing this now, can be a followup.

lib/cli/src/commands/package/unpack.rs
theduke: Let's do this in a followup.

lib/cli/src/commands/package/unpack.rs
theduke: Let's do this in a followup.

Michael-F-Bryan (approved):
LGTM :+1:

PR #4293, opened 2023-11-05
When the main thread was exiting it was not flushing and properly closing file descriptors

john-sharratt:


syrusakbary (approved):

syrusakbary:
I'll revert this PR because it caused an issue where python can't run `import sys; help(sys);` See more: #4303 

cc @john-sharratt @theduke

PR #4294, opened 2023-11-06
Terminate after flushing file descriptors

john-sharratt:
This patch will modify the behavior of a closing WASM process so that it only sends the termination signal after `stdout` and `stderr` are flushed.


syrusakbary:
We need this: Because WCGI will stream the response waiting for either of these conditions:
1. the stdout to finish streaming
2. the process to exit

So if we exit the process before we finish the streaming there is a race condition where the data could be truncated

syrusakbary (approved):

PR #4296, opened 2023-11-06
fix: prevent potential UB by deriving repr C for union

0xalpharush:
# Description
Use repr C for `VMFunctionContext` to prevent undefined behavior. I believe this union may exhibit UB since it [crosses an FFI boundary](https://github.com/wasmerio/wasmer/blob/d18aa794978e6b951e36632e29e58f4ef7e0da03/lib/vm/src/trap/traphandlers.rs#L642-L670). Also, I think `eq` and `is_null` may be called on both instances of the union despite the layout not being guaranteed. This is explained further [this](https://rust-lang.github.io/rust-clippy/master/index.html#/default_union_representation) by this clippy lint so perhaps it's worth configuring it to be `deny`.


syrusakbary:
Looks great, thanks!

PR #4297, opened 2023-11-07
Moved tmp as tmpdirectory outside of RootFileSystemBuilder

syrusakbary:
This fixes an issue with the tmpfilesystem being reused across spawning processes (aka clang)


theduke (changes requested):
Good catch!

We should add a regression test for this, eg in the wasix snapshot tests.

The test can just: 

* write a file in /tmp
* spawn a new process
* check content in file of /tmp

theduke (commented):
Also: I think we'll want to add a tmp fs option to the RootFileSystemBuilder, and only create a new  tmp directory if the args don't provide one, instead of just not creating one.

Otherwise things break that don't use the wasi runner.

PR #4298, opened 2023-11-07
Virtual fs refactor

### LINKED ISSUE #4182: Improve how mapdir works (proposal)
Currently, mapping into `.` in the host is problematic, because... what is `.`: is it `/` or is it other dir?

My thoughts on how to resolve this, to minimize collisions with root:
* set `/home` as the current working directory by default
* Anything mounted on `.` will be mounted on `/home`. For example: `wasmer run myprogram --dir=.` would be equivalent to `wasmer run myprogram --mapdir=.:.` which will be equivalent to `wasmer run --mapdir=.:/home`

That will resolve multiple issues that we have in the filesystem, like mounting `.` in `/` and many more

### LINKED ISSUE #4262: operating with tmp files fails in subprocess
When you run a program, and that program creates a subprocess that tries to operate with a temporal directory, it fails.

## Steps to reproduce

```
$ wasmer run python/python
```

Then:
```
import subprocess; subprocess.run(["/bin/python"])
```

Then:

```
import tempfile; tempfile.mkdtemp(prefix=f"pip-a-")
```

Should succeed. Now it fails:

```
➜  wasmer run python
Could not find platform dependent libraries <exec_prefix>
Python 3.12.0 (heads/wasix-compatibility-dirty:186ebc7b8b, Oct  3 2023, 10:47:09) [Clang 15.0.6 ] on wasix
Type "help", "copyright", "credits" or "license" for more information.
>>> import tempfile
>>> tempfile.mkdtemp(prefix=f"pip-a-")
'/tmp/pip-a-kzbg1opp'
>>> import subprocess; result = subprocess.run(["/bin/python"])
Could not find platform dependent libraries <exec_prefix>
Python 3.12.0 (heads/wasix-compatibility-dirty:186ebc7b8b, Oct  3 2023, 10:47:09) [Clang 15.0.6 ] on wasix
Type "help", "copyright", "credits" or "license" for more information.
>>> import tempfile
>>> tempfile.mkdtemp(prefix=f"pip-a-")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/cpython/lib/python3.12/tempfile.py", line 368, in mkdtemp
    _os.mkdir(file, 0o700)
PermissionError: [Errno 2] Permission denied: '/tmp/pip-a-xycr51b9'
>>>
```

As we see, the initial `tempfile.mkdtemp(prefix=f"pip-a-")` works in the first python process, but not in the child python one

syrusakbary:
This PR aims to clean up the virtual filesystem by using better abstraction strategies.

* VirtualFS
  * [x] Refactor Native Filesystem: NativeFS will replace HostFS: it's paths are relative and the absolute host path is not leaked outside. Basically, instead of doing `HostFS::new()` (and HostFS having full access to the host filesystem) you will do `HostFS::new("/path/in/the/host")` to then be able to do all the fs operations coming from that relative dir
  * [x] ~~Reimplement UnionFS with a simpler approach~~ **Removed Union**
    * [x] Make sure all tests in UnionFS are working (needed to get wast properly working) - **USED mem_fs instead**
  * [x] Add the Directory abstraction in VirtualFS
  * [x] Make `<dyn VirtualFile>` and `<dyn Directory>` Hasheable
  * ~~Reimplement OverlayFS~~
  * ~~Replace the filesystem in WASIX to use UnionFS for the mount directories and OverlayFS on top of the container~~
* [x] Fix `wasmer run xyz --dir=.`
* [x] Set the cwd to `/home` when running WASI (and allow customizing it via config). Plug it also from `wasmer run ...`. This will fix #4182 and also the weird mounting hack we do in packages to make relative paths absolute
* [x] Unified even more the logic between running containers and single wasm files
* [x] Improved the ergonomics and simplified the logic in wasmer run code
* Extras:
  * [ ] Get clang fully working with the /tmp folder
  * [ ] Fix #4262 where `/tmp` folder is not writeable in spawned sub processes


lib/wasix/src/runners/wasi_common.rs
```diff
@@ -101,87 +85,37 @@ impl CommonWasiOptions {
...
-    if !mapped_dirs.is_empty() {
-        let host_fs: Arc<dyn FileSystem + Send + Sync> = Arc::new(crate::default_fs_backing());
-        build_directory_mappings(builder, &mut root_fs, &host_fs, mapped_dirs)?;
-    }
+type ContainerFs = virtual_fs::TraceFileSystem<
+    OverlayFileSystem<
+        TmpFileSystem,
+        [RelativeOrAbsolutePathHack<Arc<dyn FileSystem + Send + Sync>>; 1],
+    >,
+>;
```
syrusakbary, line 113: TODO: Don't use the Tracing File System

lib/virtual-fs/src/ops.rs
```diff
@@ -329,3 +327,209 @@ mod tests {
         assert_eq!(super::read(&fs, "/file.txt").await.unwrap(), b"");
     }
 }
+
```
syrusakbary, line 330: Note: we don't need this

PR #4300, opened 2023-11-08
Use authentication when running a package

Michael-F-Bryan:
As part of private packages and #4247, we want to make sure API tokens are sent to the backend when doing `wasmer run some/private-package`.

We'd like the CLI to automatically figure out which token to use while not sending API tokens to random endpoints, so updated `WapmSource` (the component that queries the Wasmer registry when looking up packages) to accept an API token for its registry and modified `BuiltinPackageLoader` (the component for downloading those packages) so it has a hashmap of hostnames to API tokens that it will consult when downloading a `*.webc` file.


theduke (commented):
A `dyn Trait` should be a last resort, IMO.

It makes code more obtuse an harder to debug.

In this case, all it does it provide a token... why not just pass in the auth token as a setting?

theduke (approved):
LGTM now.

PR #4301, opened 2023-11-08
refactor(cli): run: Unify wasi env setup

theduke:
Unify the setup of the WasiEnv.
Previously there were different setup paths for running plain .wasm
files and for running packages.

This required some modifications to the WasiRunner, but only a
reasonable amount.

This also brings some additional improvements:

* Allow customizing the current_dir , both in the WasiEnvBuilder and in
  the WasiRunner
* Properly validate and merge --dir and --mapdir flags


lib/cli/src/commands/run/wasi.rs
```diff
@@ -238,6 +317,119 @@ impl Wasi {
...
+                    guest: guest.clone(),
+                }
+            };
+            mapped_dirs.push(mapping);
+        }
+
+        Ok(mapped_dirs)
+    }
+
+    pub fn build_mapped_commands(&self) -> Result<Vec<MappedCommand>, anyhow::Error> {
```
syrusakbary, line 404: What is this? What are map-commands?

theduke: It's a way to access host commands form inside the guest.

This is already supported in the CLI.

syrusakbary (approved):
Approving and merging now as I need this updates

PR #4302, opened 2023-11-09
Allow `WasiRunner` to mount `FileSystem` instances

Michael-F-Bryan:
As part of https://github.com/wasmerio/wasmer-js/pull/334 we need the ability to mount entire `FileSystem` instances at a particular point in the WASIX instance's filesystem. 

I've added an adaptor so the existing `map_directory()` methods for mapping a directory on the host filesystem goes through this new `mount()` mechanism.

### DIFF
--- a/lib/wasix/src/runners/wasi_common.rs
+++ b/lib/wasix/src/runners/wasi_common.rs
@@ -9,7 +9,7 @@ use derivative::Derivative;
 use futures::future::BoxFuture;
 use virtual_fs::{
     DirEntry, FileOpener, FileSystem, FsError, OverlayFileSystem, RootFileSystemBuilder,
-    TmpFileSystem, ScopedDirectoryFileSystem,
+    ScopedDirectoryFileSystem, TmpFileSystem,
 };
 use webc::metadata::annotations::Wasi as WasiAnnotation;
 
@@ -265,7 +265,6 @@ impl From<MappedDirectory> for MountedDirectory {
         MountedDirectory { guest, fs }
     }
 }
-
 
 #[derive(Debug)]
 struct RelativeOrAbsolutePathHack<F>(F);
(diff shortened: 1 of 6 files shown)


Michael-F-Bryan:
@syrusakbary or @theduke can you review this? It's the last thing Wasmer JS needs before we can make it use `master` directly.

lib/wasix/src/runners/wasi_common.rs
```diff
@@ -257,6 +239,120 @@ fn create_dir_all(fs: &dyn FileSystem, path: &Path) -> Result<(), Error> {
...
+    inner: F,
+}
+
+impl<F> HostFolderFileSystem<F> {
+    fn new(inner: F, root: PathBuf) -> Self {
+        HostFolderFileSystem { root, inner }
+    }
+
+    fn path(&self, path: &Path) -> Result<PathBuf, virtual_fs::FsError> {
+        let path = path.strip_prefix("/").unwrap_or(path);
```
theduke, line 283: This allows operations outside of the directory with relative paths.

Note: I added something similar in Edge, which uses this code to check the path:

https://github.com/wasmerio/edge/blob/e6eb175163372dfd5261d7d3da281a1ad6bfb5bf/crates/edge/src/instance/volume/local.rs#L61

https://github.com/wasmerio/edge/blob/e6eb175163372dfd5261d7d3da281a1ad6bfb5bf/crates/edge/src/instance/volume/local.rs#L136

lib/wasix/src/runners/wasi_common.rs
```diff
@@ -257,6 +239,120 @@ fn create_dir_all(fs: &dyn FileSystem, path: &Path) -> Result<(), Error> {
...
+            Arc::new(HostFolderFileSystem::new(crate::default_fs_backing(), host));
+
+        MountedDirectory { guest, fs }
+    }
+}
+
+/// A [`FileSystem`] implementation that is scoped to a specific directory on
+/// the host.
+#[derive(Debug, Clone)]
+struct HostFolderFileSystem<F> {
```
theduke, line 272: Why is this in here?

Shouldn't this be in `virtual-fs`?

As something like `ScopedDirectoryFilesystem` or similar?

Michael-F-Bryan: Moved and renamed :+1:

theduke (approved):

PR #4314, opened 2023-11-16
fix(cli): Prevent temporary file issues in "package download"

theduke:
tempfile::NamedTempFile s can not be moved between different file
systems.

Many OSes create temporary files on a separate file system, which leads
to errors when downloading.

This commit changes things to just use a regular file instead.


Michael-F-Bryan (approved):
LGTM.

The only thing we might want to think about is automatically cleaning up the temporary file if there is an error, but that's not a blocker.

PR #4317, opened 2023-11-19
Bump min enumset version to 1.1.0

webmaster128:
# Description

The symbol `enumset::__internal::EnumSetTypeRepr` used in wasmer-types was only made public in 1.1.0:
https://github.com/Lymia/enumset/compare/v1.0.13...v1.1.0

In order to ensure a project embedding Wasmer does not use incompatible enumset versions ^1.0.0, < 1.1.0 we should bump the version here to the actual min required version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,6 +84,7 @@ rust-version = "1.70"
 version = "4.2.3"
 
 [workspace.dependencies]
+enumset = "1.1.0"
 memoffset = "0.9.0"
 wasmer-toml = "0.9.2"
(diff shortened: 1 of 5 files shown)


syrusakbary (changes requested):
Hey @webmaster128 , I think it may be a better idea to set enumset as a workspace dependency, and then just do:

```yaml
enumset.workspace = true
```

webmaster128:
Sure.

Do you know what happens to `enumset.workspace = true` in the package when the crate is published? Is that replaces with the entry from the workspace?

PR #4322, opened 2023-11-23
fix: Prevent panicking in VirtualTaskManagerExt::spawn_and_block_on

theduke:
This helper method previously would panic if the inner async task failed
to send a result, instead of returning an error that can be handled.

It also:
* used an mpsc channel. a oneshot channel is better here
* used the InlineWaker to wait for the result, which is also redundant
  because tokio channels have sync receive methods


Michael-F-Bryan (approved):
The `??` looks a bit funny, but this totally makes sense.

syrusakbary (approved):

PR #4326, opened 2023-11-23
Always re-execute a registry query when cache lookups fail

### LINKED ISSUE #4288: wasmer run doesn't pick the latest package from the registry
### Describe the bug

looking at https://wasmer.io/yamt/toywasm
as of writing this, the latest version of this package is 34.0.0.

however, for some reasons,
`wasmer run yamt/toywasm` picks and runs 29.0.0.

```
% wasmer run yamt/toywasm -- --version                           [~]
toywasm v29.0.0
:
:
:
```

specifying later versions doesn't work either.
```
wasmer run yamt/toywasm@30 -- --version
error: Unable to find "yamt/toywasm@^30" in the registry
╰─▶ 1: Not found
```

i tried two versions, one from https://github.com/wasmerio/wasmer/releases/tag/v4.2.3
another from homebrew. the symptom was the same.
```sh
wasmer 4.2.3 (7cb550d 2023-10-27)
binary: wasmer-cli
commit-hash: 7cb550db60656352cc527eaf869981d39448e38d
commit-date: 2023-10-27
host: x86_64-apple-darwin
```

```
wasmer 4.1.2 (??????? 2023-08-21)
binary: wasmer-cli
commit-hash: 
commit-date: 2023-08-21
host: x86_64-apple-darwin
compiler: cranelift
```

### Steps to reproduce

see above.

### Expected behavior

pick the latest version.

### Actual behavior

for some reasons it can only see up to 29.0.0.

### Additional context

webassembly.sh picks the latest version as expected.
https://webassembly.sh/?run-command=toywasm%20--version

Michael-F-Bryan:
Fixes #4288.

The underlying problem is that we would always assume that a cached query to the backend is correct, so we wouldn't notice a newly published version. This rearranges the logic so any cached query that doesn't match a version constraint will force us to re-query the backend.

### DIFF
--- a/lib/wasix/src/runtime/resolver/wapm_source.rs
+++ b/lib/wasix/src/runtime/resolver/wapm_source.rs
@@ -133,7 +133,7 @@ impl WapmSource {
 impl Source for WapmSource {
     #[tracing::instrument(level = "debug", skip_all, fields(%package))]
     async fn query(&self, package: &PackageSpecifier) -> Result<Vec<PackageSummary>, QueryError> {
-        let (full_name, version_constraint) = match package {
+        let (package_name, version_constraint) = match package {
             PackageSpecifier::Registry { full_name, version } => (full_name, version),
             _ => return Err(QueryError::Unsupported),
         };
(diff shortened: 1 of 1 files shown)


theduke (commented):
Can we have a test for this?

I guess it's a bit awkward because it would require publishing a package.

theduke (approved):

PR #4328, opened 2023-11-27
Add `--wait` and `--timeout` flags to `wamer publish`

ayys:
## The `--wait` flag
Rasonale: When a large package is published, the user has to wait for a while to
be able to use it. This is because the webc of the package is not available on
the registry until the package is fully processed.

The registry now supports passing a wait flag to the publish command. This will
cause the command to wait until the webc of the package is available on the
registry before exiting. By default, this is set to false (for backwards
compatibility), and the `--wait` flag on the CLI enables it.

## The `--timeout` flag

Rasonale: When publishing a large package with the `--wait` flag - it may take a long
time.

The `--timeout` flag can be used to set the timeout for http requests made while
publishing the package. This is important because processing large webc files
may take a while. The default timeout is set to 30 seconds, but can be increased
if necessary (e.g. when publishing large packages).

### DIFF
--- a/lib/cli/src/commands/publish.rs
+++ b/lib/cli/src/commands/publish.rs
@@ -27,11 +27,10 @@ pub struct Publish {
     /// Defaults to current working directory.
     #[clap(name = "PACKAGE_PATH")]
     pub package_path: Option<String>,
-    /// Wait for package to be available on the registry before exiting
+    /// Wait for package to be available on the registry before exiting.
     #[clap(long)]
     pub wait: bool,
-
-    /// Timeout (in seconds) for the publish query to the registry
+    /// Timeout (in seconds) for the publish query to the registry.
     #[clap(long, default_value = "30")]
     pub timeout: u64,
 }
--- a/lib/cli/src/commands/publish.rs
+++ b/lib/cli/src/commands/publish.rs
@@ -31,6 +31,10 @@ pub struct Publish {
     #[clap(long)]
     pub wait: bool,
     /// Timeout (in seconds) for the publish query to the registry.
+    ///
+    /// Note that this is not the timeout for the entire publish process, but
+    ///
+    /// for each individual query to the registry during the publish flow.
     #[clap(long, default_value = "30")]
     pub timeout: u64,
 }
--- a/lib/cli/src/commands/publish.rs
+++ b/lib/cli/src/commands/publish.rs
@@ -27,6 +27,13 @@ pub struct Publish {
     /// Defaults to current working directory.
     #[clap(name = "PACKAGE_PATH")]
     pub package_path: Option<String>,
+    /// Wait for package to be available on the registry before exiting
+    #[clap(long)]
+    pub wait: bool,
+
+    /// Timeout (in seconds) for the publish query to the registry
+    #[clap(long, default_value = "30")]
+    pub timeout: u64,
 }
 
 impl Publish {
@@ -46,6 +53,8 @@ impl Publish {
             token,
             no_validate: self.no_validate,
             package_path: self.package_path.clone(),
+            wait: self.wait,
+            timeout: std::time::Duration::from_secs(self.timeout),
         };
         publish.execute().map_err(on_error)?;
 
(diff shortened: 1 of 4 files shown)


lib/cli/src/commands/publish.rs
```diff
@@ -27,6 +27,16 @@ pub struct Publish {
     /// Defaults to current working directory.
     #[clap(name = "PACKAGE_PATH")]
     pub package_path: Option<String>,
+    /// Wait for package to be available on the registry before exiting.
+    #[clap(long)]
+    pub wait: bool,
+    /// Timeout (in seconds) for the publish query to the registry.
+    ///
+    /// Note that this is not the timeout for the entire publish process, but
+    ///
```
Michael-F-Bryan, line 36:
```suggestion
```

lib/registry/src/publish.rs
```diff
@@ -112,11 +179,19 @@ pub fn try_chunked_uploading(
...
+fn upload_package(
+    signed_url: &str,
+    archive_path: &PathBuf,
+    archived_data_size: u64,
+    timeout: Duration,
+) -> Result<(), anyhow::Error> {
+    let url = url::Url::parse(signed_url).unwrap();
     let client = reqwest::blocking::Client::builder()
         .default_headers(reqwest::header::HeaderMap::default())
+        .timeout(timeout)
```
Michael-F-Bryan, line 194: Note that doing it this way means we'll *always* apply the timeout, not just when the user does `--wait`. This should be fine because the non-waiting publish returns pretty quickly, but I just thought I'd point it out.

lib/registry/src/publish.rs
```diff
@@ -112,11 +179,19 @@ pub fn try_chunked_uploading(
...
 
-    let signed_url = url.url;
-    let url = url::Url::parse(&signed_url).unwrap();
+fn upload_package(
+    signed_url: &str,
+    archive_path: &PathBuf,
+    archived_data_size: u64,
+    timeout: Duration,
+) -> Result<(), anyhow::Error> {
+    let url = url::Url::parse(signed_url).unwrap();
```
Michael-F-Bryan, line 191: I know this shouldn't ever fail, but we're returning `anyhow::Error` anyway so may as well do proper error handling.

```suggestion
    let url = url::Url::parse(signed_url).context("Invalid signed URL")?;
```

lib/cli/src/commands/publish.rs
```diff
@@ -27,6 +27,16 @@ pub struct Publish {
...
+    /// for each individual query to the registry during the publish flow.
+    #[clap(long, default_value = "30")]
+    pub timeout: u64,
```
Michael-F-Bryan, line 39: We should probably use [`humantime::Duration`](https://docs.rs/humantime/latest/humantime/struct.Duration.html) instead of `u64` here. That'll let users do `--timeout="1 min 42s"` or whatever.

There is also a convenient `.into()` implementation for converting to a normal `std::time::Duration` that we can use when initializing the `Publish` struct.

```suggestion
    #[clap(long, default_value = "30s")]
    pub timeout: humantime::Duration,
```

(probably requires adding `humantime` to `wasmer-cli` as a dependency.

Michael-F-Bryan (commented):
Looks like a mostly straightforward addition. 

I've added some comments for how we could make the `--timeout` argument more user-friendly (i.e. let them do `--timeout=60s`), but don't see any issues.

Thanks for cleaning up the `wasmer-registry` code, by the way!

lib/cli/src/commands/publish.rs
ayys: @Michael-F-Bryan this is pretty cool, but I noticed that humantime is not a dependency for wasmer-cli. Are we ok with adding it as a new dependency?

ayys: I added the dependency, closing this.

lib/registry/src/publish.rs
ayys: Added the context, closing this thread.

lib/cli/src/commands/publish.rs
ayys: Done!

Michael-F-Bryan (approved):

PR #4331, opened 2023-11-28
Fix visibility validation to work when publishing a new package

### LINKED ISSUE #4329: CLI: wasmer publish fails for private packages that don't exist yet
### Describe the bug

`wasmer publish` with a `private = true` package fails if the package does not exist yet, because it tries to check for package privacy status.

```
❯ wasmer publish
error: Unable to look up package information
╰─▶ 1: error sending query: no package version for "christoph/winterjs-test2"
```

Michael-F-Bryan:
Fixes #4329.


Arshia001 (approved):
LGTM

PR #4334, opened 2023-11-29
Add `application/wasm` to list of accepted content-types for webcs

ayys:
-------

When downloading a webc file, accept files with `application/wasm` header as
well.
The backend sets `Content-Type: application/wasm` header on webc files in the
object storage, so they can have dynamic compression enabled on them.
This makes downloading them faster by enabling `Accept-Encoding: gzip` when
downloading the file.


Michael-F-Bryan (approved):
It's a bit of a hack to pretend a `*.webc` file is `application/wasm` just so our CDN will automatically add compression, but I guess that's the easiest solution right now.

Issue #3851, opened 2023-05-08
WASIX's file-based module cache will block the runtime thread


Michael-F-Bryan:
### Describe the bug

The fs-based cache WASIX uses for compiled modules is called from `async` code, but uses synchronous IO for reading the file and deserializing. This happens in both the current version ([here](https://github.com/wasmerio/wasmer/blob/5cab85f85fb8b1debdec2dbeeb56e4f354a747fc/lib/wasi/src/bin_factory/module_cache.rs#L111) and [here](https://github.com/wasmerio/wasmer/blob/5cab85f85fb8b1debdec2dbeeb56e4f354a747fc/lib/wasi/src/bin_factory/module_cache.rs#L167-L171)) and the version introduced in #3841 ([here](https://github.com/wasmerio/wasmer/blob/e633eb3e52f0b5a8f27c726ee1ffd52f0070bf7f/lib/wasi/src/runtime/module_cache/filesystem.rs#L117-L152)).

We need to tweak the new `FileSystemCache` implementation to either use the `tokio::fs` module (typing the `FileSystemCache` to a tokio runtime) or make the `FileSystemCache` accept an `Arc<dyn VirtualTaskManager>` parameter in its constructor and do the blocking IO via `spawn_blocking()`.

maminrayej:
Resolved by #4345.

PR #4345, opened 2023-12-07
Asyncify filesystem cache

maminrayej:
In the middle of my investigations for improving startup time, I encountered issue #3851 in one of the comments.
This fix isn't ideal since it uses the `tokio` thread pool instead of the pool in `TokioTaskManager`. I couldn't figure out how to design a trait that is both object-safe, and is able to model the `spawn_await` function that I had in mind properly. This is the best that I came up with:
```rust
pub trait AsyncTaskManager {
    type Future<T>: Future<Output = T>;
    
    fn spawn_await<T>(&self, f: impl Future<Output = T>) -> Self::Future<T>;
}
```
which isn't object-safe.

I could clone the `TokioTaskManager` directly into `FileSystemCache`, but I was not sure about it.

### DIFF
--- a/lib/wasix/src/runtime/module_cache/filesystem.rs
+++ b/lib/wasix/src/runtime/module_cache/filesystem.rs
@@ -37,7 +37,7 @@ impl FileSystemCache {
 
 #[async_trait::async_trait]
 impl ModuleCache for FileSystemCache {
-    #[tracing::instrument(level = "debug", skip_all, fields(%key))]
+    #[tracing::instrument(level = "debug", skip_all, fields(% key))]
     async fn load(&self, key: ModuleHash, engine: &Engine) -> Result<Module, CacheError> {
         let path = self.path(key, engine.deterministic_id());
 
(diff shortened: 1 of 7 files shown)


lib/wasix/src/runtime/module_cache/filesystem.rs
```diff
@@ -37,45 +40,57 @@ impl FileSystemCache {
...
+                            }
+                            Err(e) => {
+                                tracing::debug!(
+                            %key,
+                            path=%path.display(),
+                            error=&e as &dyn std::error::Error,
+                            "Deleting the cache file because the artifact couldn't be deserialized",
+                        );
+
+                                if let Err(e) = std::fs::remove_file(&path) {
```
Michael-F-Bryan, line 72: Should this be done using a `tokio` function?

maminrayej: Since the code is run on one of `rusty_pool` threads, it's okay.

maminrayej:
@Michael-F-Bryan are there any blockers for this PR?

Michael-F-Bryan (approved):
Let's merge this! 🚀

PR #4350, opened 2023-12-09
Optimize sleep_now for TokioTaskManager 

xuanyuan300:
# Description

Currently, the `sleep_now` method in `VirtualTaskManager` is used in many places, such as socket. I found that according to the current `sleep_now`, after the task execution is completed, since the timeout task is just `recv` future, the sleep task that actually executes the timeout cannot be really canceling, this may cause tasks not being recycled immediately, so it will bring some potential resource issues.
This pr optimizes `sleep_now` and uses `AbortHandle` to abort the timeout task when dropping.


theduke (approved):
Makes sense.

We would want some additional improvements to reduce the overhead here, but this is a good fix, thanks!

PR #4352, opened 2023-12-11
Wait for webc, bindings, exe generation during package-publish with `--wait` flag

ayys:
When `wasmer publish` is called with the `--wait` flag, it now triggers a graphql subscription that waits for webc generation, language bindings generation and native executables generation.

~Note: This feature is only available on dev right now (https://registry.wasmer.wtf/graphql)~
This feature is on deployed on prod backend.

https://github.com/wasmerio/wasmer/assets/48376475/1b8b1134-fdc7-4f41-8a05-d62cea92d3ef

### DIFF
--- a/lib/cli/src/commands/publish.rs
+++ b/lib/cli/src/commands/publish.rs
@@ -40,7 +40,8 @@ pub struct Publish {
 
 impl Publish {
     /// Executes `wasmer publish`
-    pub fn execute(&self) -> Result<(), anyhow::Error> {
+    #[tokio::main]
+    pub async fn execute(&self) -> Result<(), anyhow::Error> {
         let token = self
             .env
             .token()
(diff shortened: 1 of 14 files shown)


syrusakbary (approved):
It looks good on my side, although I think it will be better for @Michael-F-Bryan to review it

lib/cli/src/commands/publish.rs
```diff
@@ -34,13 +34,14 @@ pub struct Publish {
     ///
     /// Note that this is not the timeout for the entire publish process, but
     /// for each individual query to the registry during the publish flow.
-    #[clap(long, default_value = "30s")]
+    #[clap(long, default_value = "2m")]
```
ayys, line 37: I am increasing this for all the publish queries because 20 seconds is too small anyway.

PR #4355, opened 2023-12-14
Fix the `virtual_fs::StaticFile` implementation

Michael-F-Bryan:
The `AsyncRead` implementation for `virtual_fs::StaticFile` never updates its internal cursor, so reading from a `StaticFile` would always yield the same bytes without ever encountering an EOF.

I've fixed this by throwing away a lot of the custom code and just making `StaticFile` a thin wrapper around `std::io::Cursor`. I've also updated the other filesystem trait implementations so trying to modify a `virtual_fs::StaticFile` now returns a permission error instead of silently doing nothing.

### DIFF
--- a/lib/virtual-fs/src/static_file.rs
+++ b/lib/virtual-fs/src/static_file.rs
@@ -1,27 +1,23 @@
-use futures::future::BoxFuture;
-use tokio::io::{AsyncRead, AsyncSeek, AsyncWrite};
+use std::{
+    convert::TryInto,
+    io::{self, Cursor},
+    pin::Pin,
+    task::{Context, Poll},
+};
 
-use std::borrow::Cow;
-use std::convert::TryInto;
-use std::io::{self, SeekFrom};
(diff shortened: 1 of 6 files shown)


lib/virtual-fs/src/static_file.rs
```diff
@@ -30,49 +31,44 @@ impl VirtualFile for StaticFile {
...
-        let bytes = &bytes[cursor..];
-
-        if bytes.len() > buf.remaining() {
-            let remaining = buf.remaining();
-            buf.put_slice(&bytes[..remaining]);
-        } else {
-            buf.put_slice(bytes);
-        }
-        Poll::Ready(Ok(()))
+        Pin::new(&mut self.0).poll_read(cx, buf)
```
theduke, line 71: Nit: this is usually done with `x.as_mut()` (method on Pin)

Also applies to the other uses.

theduke (approved):

Issue #4358, opened 2023-12-17
`filestat` should contain timestamps in nanoseconds, not seconds


yagehu:
### Describe the bug

The [`filestat`](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-filestat-record) type returned by `fd_filestat_get` should contain timestamps in nanoseconds.  However, currently Wasmer returns timestamps in seconds for directories, e.g. preopened directories.

```sh
wasmer 4.2.4 (1abcd2c 2023-12-16)
binary: wasmer-cli
commit-hash: 1abcd2cc63deeca13af6506add6029edaba4d657
commit-date: 2023-12-16
host: x86_64-unknown-linux-gnu
compiler: singlepass,cranelift
rustc 1.73.0 (cc66ad468 2023-10-03)
binary: rustc
commit-hash: cc66ad468955717ab92600c770da8c1601a4ff33
commit-date: 2023-10-03
host: x86_64-unknown-linux-gnu
release: 1.73.0
LLVM version: 17.0.2
```

### Steps to reproduce
Compile this C snippet with wasi-sdk and run the binary with Wasmer.

```bash
clang wasmer-dir-time.c
wasmer run --dir . a.out
```

```c
#include <unistd.h>
#include <sys/stat.h>
#include <stdio.h>

int main(void) {
  struct stat buf;

  int ret = fstat(4, &buf);
  if (ret != 0) {
    perror("fstat");
    return 1;
  }

  printf("atime %lld seconds", buf.st_atime);

  return 0;
}
```

### Expected behavior
It should print a reasonable timestamp like:

> atime 1702794075 seconds

### Actual behavior
It prints a very small number:

> atime 1 seconds

PR #4359, opened 2023-12-17
Use nanoseconds in `filestat` for directories

yagehu:
This commit fixes the `filestat` type which contains three timestamps. Per [WASI spec], they should be in nanoseconds.  This is only a problem for directories.

[WASI spec]: https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-timestamp-u64

fixes https://github.com/wasmerio/wasmer/issues/4358

### DIFF
--- a/lib/virtual-fs/src/mem_fs/filesystem.rs
+++ b/lib/virtual-fs/src/mem_fs/filesystem.rs
@@ -1520,11 +1520,13 @@ mod test_filesystem {
                     modified,
                     len: 0
                 }) if
-                    accessed == foo_metadata.accessed &&
-                    created == foo_metadata.created &&
+                    accessed <= foo_metadata.accessed &&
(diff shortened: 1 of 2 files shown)


lib/virtual-fs/src/mem_fs/filesystem.rs
```diff
@@ -1520,11 +1520,13 @@ mod test_filesystem {
                     modified,
                     len: 0
                 }) if
-                    accessed == foo_metadata.accessed &&
-                    created == foo_metadata.created &&
+                    accessed <= foo_metadata.accessed &&
+                    created <= foo_metadata.created &&
```
yagehu, line 1524: This test seemed to be spec'd wrong. Since the root directory was created before `foo`, `accessed` and `created` should be less than or equal to foo's.

Issue #4361, opened 2023-12-18
`path_open` not working with wasi-sdk


yagehu:
### Describe the bug

This following simple C program compiled with wasi-sdk does not work with Wasmer.

```sh
$ wasmer -vV; rustc -vV
wasmer 4.2.4 (1abcd2c 2023-12-16)
binary: wasmer-cli
commit-hash: 1abcd2cc63deeca13af6506add6029edaba4d657
commit-date: 2023-12-16
host: x86_64-unknown-linux-gnu
compiler: singlepass,cranelift
rustc 1.73.0 (cc66ad468 2023-10-03)
binary: rustc
commit-hash: cc66ad468955717ab92600c770da8c1601a4ff33
commit-date: 2023-10-03
host: x86_64-unknown-linux-gnu
release: 1.73.0
LLVM version: 17.0.2
```

### Steps to reproduce

Compile with wasi-sdk and run with Wasmer.

```bash
$ clang wasmer-creat.c
$ mkdir -p tmptest
$ wasmer run --dir . a.out
```

```c
#include <fcntl.h>
#include <stdio.h>

int main(void) {
  int fd = open("tmptest/a", O_CREAT | O_RDWR);
  if (fd == -1) {
    perror("open");
    return 1;
  }

  return 0;
}
```

### Expected behavior
This snippet should create a file `a` in the `tmptest` directory.

### Actual behavior
> open: No such file or directory

### Additional context
It seems the parent directory is resolved to the virtual root in the `path_open` handler.  I'll be able to submit a patch.

theduke:
The file system is sandboxed by default.

You need to mount host directories into the WASM context by using `--mapdir <GUEST_PATH>:<HOST_PATH>`.

In your example: `--mapdir tmptest:tmptest`.

theduke:
Feel free to comment/reopen if you run into additional issues.

yagehu:
@theduke I'm aware. I am mounting a directory with `--dir .`. I've raised a PR that fixes this. https://github.com/wasmerio/wasmer/pull/4363

yagehu:
I believe this is just an error with mapping `.` whichthe PR fixes.

PR #4363, opened 2023-12-18
Fix preopened current directory not working 

yagehu:
This commit fixes a bug where creating a file with an implicit relative
path returns `0` indicating success, but no file is created on the host.
Here, implicit relative means a relative path that does not begin with
`./`. For example, calling `open("somefile", O_CREAT | O_WRONLY)` will
try to create a file called `somefile` in the current working directory.

The fix is to change the current directory if the user has explicitly
mounted any host directory to `.` in guest.  In this case, the user's
intention is clear: the current directory *is* the mounted host
directory.

fixes https://github.com/wasmerio/wasmer/issues/4361

### DIFF
--- a/lib/virtual-fs/src/mem_fs/file_opener.rs
+++ b/lib/virtual-fs/src/mem_fs/file_opener.rs
@@ -187,7 +187,7 @@ impl FileSystem {
                             let time = time();
                             Metadata {
                                 ft: FileType {
-                                    file: true,
+                                    dir: true,
                                     ..Default::default()
                                 },
                                 accessed: time,
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -582,6 +582,16 @@ impl WasiFs {
                 .metadata(path)
                 .map_err(|e| format!("Could not get metadata for file {:?}: {}", path, e))?;
 
+            if let Some(alias) = alias {
+                // User has mounted some host directory to `.` in guest. By default, `current_dir` was
+                // set to `/`. But the user's intention is clear here: the host directory *is* the current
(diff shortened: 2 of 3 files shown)


yagehu:
@theduke Gentle ping. I think this is a reasonable fix for a common use case not working on Wasmer. Specifically, something like:

```c
#include <fcntl.h>
#include <stdio.h>

int main(void) {
  int fd = open("somefile", O_CREAT | O_WRONLY);\
  if (fd == -1) {
    perror("open");
    return 1;
  }

  return 0;
}
```

syrusakbary:
This needs a bit more discussion between @theduke and @syrusakbary (me). Stay tuned, and thanks for opening the PR @yagehu (and the patience!)

yagehu:
> This needs a bit more discussion between @theduke and @syrusakbary (me). Stay tuned, and thanks for opening the PR @yagehu (and the patience!)

Hi @syrusakbary, any update? Just to jolt your memory, the motivation for this PR is that a simple `path_open()` with an implicit relative path (a relative path without a leading `./`) does not work with Wasmer while all other runtimes can handle this behavior.  In a contrived C example:

```
open("somefile", O_CREAT | O_RDWR)
```

should create a file `somefile` at the current directory.

For context, I'm differential fuzzing various wasm runtimes at the raw WASI level (bypassing wasi-libc).  Wasmer is the only runtime where I need to apply this patch so the file system behavior is consistent with other runtimes.

I can update this PR

yagehu:
> Hey @yagehu , most of the tests are failing on this branch, which is why I'm hesitant to merge atm

To be clear, I was not asking to merge this PR as is since a lot has changed since my initial PR.  I'll update this PR if there's interest in aligning Wasmer path_open behavior with other runtimes.

PR #4368, opened 2023-12-21
fix(wasix): http client: use reasonable default connect timeout

theduke:
Without this, requests can stall for an infinite amount of time.

This is particularly problematic because the default http client is also
used for interacting with the registry.


Michael-F-Bryan (approved):
Make sense :+1:

Issue #4383, opened 2024-01-02
WASIX: Custom atomics handling


theduke:
Currently WASIX does not have any provisions for atomics handling, which can be customized through `Tunables`.

This can lead to WASIX instances not handling signals properly, because they are waiting on atomics. 

We need to wake up any atomics waiters in an instance when signals come in.

PR #4385, opened 2024-01-03
feat(api): Add SharedMemoryHandle

theduke:
Allows to trigger atomics related operations on a memory that supports
interacting with the atomics layer though a shared handle.

Note that this needs to be a separate handle that is not a store reference, because it needs to be accessible while an instance is executing.

Right now only final operations are exposed that wake up all atomic
waiters, and that allow disabling atomics.

This is required for forceful shutdown of instances that keep
hot-looping on atomics, and will be used from WASIX.

Part of #4383

### DIFF
--- a/lib/api/src/js/externals/memory.rs
+++ b/lib/api/src/js/externals/memory.rs
@@ -184,7 +184,7 @@ impl Memory {
         true
     }
 
-    pub fn as_shared(&self, _store: &impl AsStoreRef) -> Option<crate::SharedMemoryHandle> {
+    pub fn as_shared(&self, _store: &impl AsStoreRef) -> Option<crate::SharedMemory> {
         // Not supported.
         None
     }
--- a/lib/api/src/js/externals/memory.rs
+++ b/lib/api/src/js/externals/memory.rs
@@ -183,6 +183,11 @@ impl Memory {
     pub fn is_from_store(&self, _store: &impl AsStoreRef) -> bool {
         true
     }
+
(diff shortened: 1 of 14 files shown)


theduke:
Pushed a small change that makes the handle only hold a weak reference to the underlying atomics handler to avoid prolonging the memory lifetime.

lib/api/src/sys/externals/memory.rs
```diff
@@ -99,12 +101,73 @@ impl Memory {
         mem.copy()
     }
 
+    pub fn to_shared(&self, store: &impl AsStoreRef) -> Option<crate::SharedMemory> {
+        let mem = self.handle.get(store.as_store_ref().objects());
```
syrusakbary, line 105: We need to check that the memory is actually a shared memory.

lib/api/src/js/externals/memory.rs
```diff
@@ -183,6 +183,11 @@ impl Memory {
     pub fn is_from_store(&self, _store: &impl AsStoreRef) -> bool {
         true
     }
+
+    pub fn to_shared(&self, _store: &impl AsStoreRef) -> Option<crate::SharedMemoryHandle> {
```
syrusakbary, line 187: probably want to name it `as_shared` since this is not consuming self?

theduke: Good point, done.

lib/api/src/sys/externals/memory.rs
theduke: That already happened implicitly, because only shared memories can hand out a shared memory handle. 

But I added an additional sanity check regardless, can't hurt.

PR #4389, opened 2024-01-04
remove memory footprint computation

maminrayej:
We calculate the `module_memory_footprint` in `load_package_tree` by collecting `atoms` in a set and adding their lengths together which is a costly operation. But, we don't use the calculated value anywhere in the code base. `load_package_tree` is in the critical path when running a package so removing these computation is beneficial.

before:
```
wasmer_wasix::runtime::package_loader::load_package_tree: close time.busy=60.9ms time.idle=2.12µs
```
after
```
wasmer_wasix::runtime::package_loader::load_package_tree: close time.busy=64.0µs time.idle=1.20µs
```


Michael-F-Bryan (approved):
I checked with @theduke and`module_memory_footprint` isn't used anywhere, so this should be an easy win :+1:

PR #4394, opened 2024-01-08
ignore append when truncate is set

### LINKED ISSUE #4376: Cannot open a file with O_TRUNC and O_APPEND
### Describe the bug

This might be a wontfix but I figure I should report this anyway since out of all the runtimes I've tested only Wasmer and wasmi have this behavior.  Native Linux also does not have this behavior.

Opening a file with O_TRUNC and O_APPEND together results in errno 28 `inval`. This is a result of the [host-fs `FileSystem`](https://github.com/wasmerio/wasmer/blob/master/lib/virtual-fs/src/host_fs.rs#L207) uses Rust `std::fs::OpenOptions`, which performs [this check](https://github.com/rust-lang/rust/blob/15755f38cd94f160713a6117829b0d1da2535b05/library/std/src/sys/unix/fs.rs#L1103).

I think Wasmtime doesn't have this behavior because it uses rustix which directly invokes the `open` syscall without this extra check.

```sh
$ zwasmer -vV; rustc -vV
wasmer 4.2.4 (0d782f1 2023-12-22)
binary: wasmer-cli
commit-hash: 0d782f18f99ba503f9aff89669c6ffd8e1631f53
commit-date: 2023-12-22
rustc 1.74.1 (a28077b28 2023-12-04)
binary: rustc
commit-hash: a28077b28a02b92985b3a3faecf92813155f1ea1
commit-date: 2023-12-04
release: 1.74.1
LLVM version: 17.0.4
```

### Steps to reproduce

Compile the snippet with wasi-sdk and run with wasmer, mounting a directory.

```c
#include <fcntl.h>
#include <stdio.h>

int main(void) {
    int fd = open("tmp/j", O_CREAT | O_RDWR | O_TRUNC | O_APPEND);
    if (fd == -1) {
        perror("open");
        return 1;
    }

    return 0;
}
```

### Expected behavior

It should create the file `tmp/j` and exit with no error.

### Actual behavior

> open: Invalid argument

maminrayej:
Resolves #4376.

### DIFF
--- a/lib/virtual-fs/src/host_fs.rs
+++ b/lib/virtual-fs/src/host_fs.rs
@@ -222,13 +222,20 @@ impl crate::FileOpener for FileSystem {
         // TODO: handle create implying write, etc.
         let read = conf.read();
         let write = conf.write();
-        let append = conf.append();
+
+        // according to Rust's stdlib, specifying both truncate and append is nonsensical,
+        // and it will return an error if we try to open a file with both flags set.
+        // in order to prevent this, and stay compatible with native binaries, we just ignore
+        // the append flag if truncate is set. the rationale behind this decision is that
+        // truncate is going to be applied first and append is going to be ignored anyway.
+        let append = if conf.truncate { false } else { conf.append() };
+
         let mut oo = fs::OpenOptions::new();
         oo.read(conf.read())
             .write(conf.write())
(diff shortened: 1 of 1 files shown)


Arshia001 (approved):
LGTM

syrusakbary (approved):

PR #4397, opened 2024-01-09
Fix WasiProcess::terminate()

theduke:
fix(wasix): Fix WasiProcess::terminate()

Previously the code would mix up the output status of a thread/process with
an instruction to exit with a specific code.

The Process::terminate() method would just set the output status,
leaving no way to check if a thread had actually terminated.

This commit adds an additonal atomic flag that records if the process
should exit, and changes the relevant code (signal processing etc) to
check that flag.

The terminate method now just sets this code and then sends a kill
signal.

**NOTE**: The PR also contains some cleanup of the WasiProcess internal state.
Best to review commit by commit.

- chore(wasix): Clean up WasiProcess state
- refactor(wasix): Rename WasiProcessCheckpoint to ProcessCheckpoint
- chore(vm): Fix clippy lints
- fix(wasix): Fix WasiProcess::terminate()


syrusakbary (approved):
LGTM 👍

theduke:
Got some changes in the integration tests. 

Need to investigate.

PR #4398, opened 2024-01-09
make public api consistent using each backend

### LINKED ISSUE #4033: The `wasmer` crate's `sys`, `js`, and `jsc` feature flags expose an inconsistent public API

maminrayej:
This PR tries to combine all the suggestions in #4033:

- By hiding `sys`, `js` and `jsc` behind the `backend` module, it makes sure all three backends expose the same API.
- In order to not break the public API, stub structs, traits, ... are provided via the `stub` module. `js` and `jsc` use this module to expose the same API as `sys` by re-exporting items that they don't have an implementation for.
- Unfortunately it seems using `#[deprecated]` with re-exports doesn't work in Rust yet and is a known [issue](https://github.com/rust-lang/rust/issues/30827). So I didn't have a way to incorporate this suggestion from the original issue.

Resolves #4033.


syrusakbary:
> By hiding sys, js and jsc behind the backend module, it makes sure all three backends expose the same API.

I don't see how. This PR just repositioning the files, creating a `lib/api/src/backend/unsupported.rs` (should those  types really exist in other platforms?).

maminrayej:
> I don't see how. This PR just repositioning the files, creating a lib/api/src/backend/unsupported.rs (should those types really exist in other platforms?).

The problem at hand are the items that are exposed as part of `lib/api` that depend on a specific backend like `sys`. One example is what Michael encountered. For example `Artifact` is exposed when `sys` is enabled so one can use it and then expect it to work, just to find out their project won't build with `js` or `jsc`. This is the inconsistency that this PR is trying to solve. From here on out we have two options:

1. Remove things like `Artifact` from our api and put it behind a feature gated module like `sys`. This will make the api exposed by `sys` consistent with `js` and `jsc`, but requires a breaking change. I don't think we can afford it right now.
2. Add a temporary stub implementation of `Artifact` for `js` and `jsc`. This also makes the api exposed by these backends consistent with each other. It doesn't require a breaking change and is the way Rust std manages OS specific things for example.

This PR tries to implement the second solution. Also, in order to prevent this mistake from happening again in the future, it explicitly re-exports everything from the backend. For example, assume we add `Foo` and expose it via `sys`. We have to expose it via `backend` too. But, since `backend` expects all of its `imps` to expose `Foo`, `lib/api` won't get compiled until we provide an impl of `Foo` for all backends, preventing any inconsistency to happen again.

So in conclusion, this PR makes the api exposed by different backends consistent without a breaking change, and prevents inconsistencies in the future.

Note that this PR is marked as draft and is still completely ready to be reviewed.

lib/api/src/backend/unsupported.rs
```diff
@@ -0,0 +1,83 @@
+#![allow(dead_code)]
+
+use std::marker::PhantomData;
```
syrusakbary, line 3: We chatted about not doing this strategy (as it can infinitely open up the API surface), but only exposing the sys API through the main lib as deprecated exported fields to prompt the user to use `::sys` instead

lib/api/src/lib.rs
```diff
@@ -446,11 +446,36 @@ mod value;
 pub mod vm;
 
 #[cfg(feature = "sys")]
-mod sys;
+/// sys
+pub mod sys;
 
 #[cfg(feature = "sys")]
 pub use sys::*;
```
syrusakbary, line 453: Do we want this?

lib/api/src/lib.rs
```diff
@@ -446,11 +446,36 @@ mod value;
...
 
+#[cfg(feature = "sys")]
+#[deprecated(note = "Use sys::Artifact instead")]
```
syrusakbary, line 456:
```suggestion
#[deprecated(note = "wasmer::Artifact is deprecated, use wasmer::sys::Artifact instead")]
```

lib/api/src/lib.rs
maminrayej: Well we're saying `use wasmer::sys::Artifact instead`, so this module should be available for users to be able to migrate.

syrusakbary (approved):

PR #4404, opened 2024-01-16
Fix for duplicate __wasm_signal exports - using a unique name for the second occurance

john-sharratt:


Arshia001 (approved):

lib/wasix/src/state/env.rs
```diff
@@ -155,7 +155,13 @@ impl WasiInstanceHandles {
             signal: instance
                 .exports
                 .get_typed_function(&store, "__wasm_signal")
-                .ok(),
+                .ok()
+                .or_else(|| {
+                    instance
```
theduke, line 160: Can we add a comment adding context and describing why this is happening?

syrusakbary: ping @john-sharratt

john-sharratt:
This is no longer needed

PR #4407, opened 2024-01-25
CLI: Big CLI cleanup + merge Edge CLI

### LINKED ISSUE #4405: Move wasmer-api and edge-client-cli to Wasmer Repo
We should the base client-cli crates from Edge to this repo to help iteration and the release process.

We might as well refactor the registry-api crate while we do this

theduke:
This PR cleans up the CLI and merges in both the `wasmer-api` (backend API bindings) crate and the the `wasmer-edge-cli` create (Edge related commands).

The diff is very large, reviewers should look at the individual commits.

There is a lot of cleanup, like re-structuring the modules, removing a bunch of
dead code, normalizing all the `use` imports to the standard format, ...

My initial intention was to keep them separate, and it would potentially make
sense to factor out all commands that use the Wasmer Backend in the future, but
this was not feasible in this PR without prevent future required cleanups and
normalization, particular regarding the wasmer_registry crate.

**NOTE:**: This PR makes all the CLI internals private.
This is a breaking change, but the CLI internals are heavily idiomatic and should
not be exposed to external users.
We were already not really paying attention to potential breakage.
Still needs to be discussed if we can actually do this.

Closes #4405 

Required followup work:

* Rewrite `wasmer-registry` GraphQL API usage to the `wasmer-api` crate : #4408 
* Clean up/split up/remove the `wasmer-registry` crate: #4409


theduke:
@syrusakbary I added two followup issues.

syrusakbary (approved):

PR #4410, opened 2024-01-25
fix(cli): publish: Fix panic + Make waiting for build results more modular

theduke:
* Fixes a panic due to nested tokio runtimes

* Exposes a --wait flag, which only waits for the container, and a
--wait-all flag, which also waits for bindings and native executables.

Also adds a timeout for waiting.

### DIFF
--- a/lib/cli/src/commands/publish.rs
+++ b/lib/cli/src/commands/publish.rs
@@ -1,6 +1,6 @@
 use anyhow::Context as _;
 use clap::Parser;
-use wasmer_registry::wasmer_env::WasmerEnv;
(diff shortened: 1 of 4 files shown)


syrusakbary (approved):

ayys:
Have you considered making it, so users can do `--wait=[webc|bindings|native-exes|all]`?
`wasmer publish --wait` would wait for all. Or I can wait for just the webc and bindings if that's what I care about with `wasmer publish --wait=webc,bindings`.

theduke:
@ay\ I actually did, sadly that's a bit of a pain with `clap`, there is no way to do it with the derive, we'd have to right the parser impl by hand.

PR #4428, opened 2024-01-31
Try to build the package before publishing it

maminrayej:
This PR adds a preliminary step to `wasmer run` that tries to build the package before publishing it. As a bonus, `wasmer package build` now has the `--check` flag which only checks whether the package could be built or not.

Related to #4288 and this [comment](https://github.com/wasmerio/wasmer/issues/4288#issuecomment-1919086857).

### DIFF
--- a/lib/cli/src/commands/package/build.rs
+++ b/lib/cli/src/commands/package/build.rs
@@ -20,6 +20,10 @@ pub struct PackageBuild {
     ///
     /// Defaults to current directory.
     package: Option<PathBuf>,
+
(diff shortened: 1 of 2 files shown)


syrusakbary (approved):
LGTM

ayys:
I updated the description thinking the diff will show up as an activity on this page, but it didn't.
```diff
```diff
- This PR adds a preliminary step to `wasmer run` that tries to build the package before publishing it
+ This PR adds a preliminary step to `wasmer publish` that tries to build the package before publishing it
```

PR #4429, opened 2024-02-01
Return Notsup for absolute symlinks instead of panicking

boomshroom:
Generally when userspace tries to do something the environment can't do (even if it might be able to do it in the future), the usual response is the return a "Not Supported" error, rather than crashing the entire environment.

This addresses, but doesn't quite solve #2243.

### DIFF
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -930,7 +930,8 @@ impl WasiFs {
                                 let (pre_open_dir_fd, relative_path) = if link_value.is_relative() {
                                     self.path_into_pre_open_and_relative_path(&file)?
                                 } else {
-                                    unimplemented!("Absolute symlinks are not yet supported");
+                                    tracing::error!("Absolute symlinks are not yet supported");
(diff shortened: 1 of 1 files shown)


syrusakbary (commented):
This looks great, will wait from feedback from @theduke or @john-sharratt prior to merge

lib/wasix/src/fs/mod.rs
```diff
@@ -930,7 +930,8 @@ impl WasiFs {
                                 let (pre_open_dir_fd, relative_path) = if link_value.is_relative() {
                                     self.path_into_pre_open_and_relative_path(&file)?
                                 } else {
-                                    unimplemented!("Absolute symlinks are not yet supported");
+                                    tracing::error!("Absolute symlinks are not yet supported");
```
theduke, line 933: We don't wan't logging in the `virtual-fs` crate, that should be kept to usage sites like `wasix`.

theduke (approved):
Agreed, this is a leftover, returning an error is the proper way to go here.

Thanks!

Approved, but we should remove the trace statement. ( @boomshroom  )

PR #4434, opened 2024-02-02
Update Rust toolchain to 1.73

syrusakbary:
We should be normally 1-2 versions behind than the official version.

I got an error when trying to compile Wasmer as another crate was failing compilation. This fixes it


lib/c-api/src/wasm_c_api/store.rs
```diff
@@ -1,11 +1,11 @@
 use super::engine::wasm_engine_t;
 use std::cell::UnsafeCell;
-use std::sync::Arc;
+use std::rc::Rc;
 use wasmer_api::{AsStoreMut, AsStoreRef, Store, StoreMut, StoreRef as BaseStoreRef};
 
 #[derive(Clone)]
 pub struct StoreRef {
-    inner: Arc<UnsafeCell<Store>>,
```
theduke, line 8: Why this change?
This would make the StoreRef non-transferable between threads.

syrusakbary: We are having an Arc over an element that is not `Send` and `Sync`, which basically triggers a compilation error in latest Rust @theduke

theduke (approved):

syrusakbary:
The two failing tests are flaky, nothing to do with this PR

ayys:
This broke the schema validation CI on registry with error
```
cargo fmt --all -- --check
error: 'cargo-fmt' is not installed for the toolchain '1.73-x86_64-unknown-linux-gnu'
To install, run `rustup component add rustfmt --toolchain 1.73-x86_64-unknown-linux-gnu`
```

fix was simple though, just updated the toolchain being installed as part of the CI

```diff
      - name: Install Rust
        uses: dtolnay/rust-toolchain@stable
        with:
-          toolchain: "1.70"
+          toolchain: "1.73"
          components: rustfmt, clippy
```

PR #4440, opened 2024-02-06
cli: allow filtering by log stream

### LINKED ISSUE #4412: Edge-CLI: instance logs: show log stream + allow filtering by log stream
Extend the app logs command with:

- Showing the log stream (stdout,stderr,runtime)
- Filtering by log stream

This needs to wait for GraphQL API extensions in the backend. (done by @ayys )

maminrayej:
Extends the `app logs` command to allow filtering different log streams.

Closes #4412  / RUN-69


ayys:
@maminrayej Do you think it makes sense to do it like `wasmer app logs --streams [stdout|stderr|all]`?

maminrayej:
I think your suggestion of allowing the user to list the desired streams is better. It's going to be a little bit more awkward to parse, but I think it's worth it since if any more streams get added in the future, having `--stream1 --stream2 --stream3` is ugly.

But I'm not the biggest fan of `|` between options. I prefer:
```console
wasmer app logs --streams stdout, stderr
```
It's similar to how you can specify a set of features in Cargo:
```console
cargo build --features "feature1, feature2"
```

What do you think?

lib/backend-api/src/types.rs
```diff
@@ -591,6 +591,12 @@ mod queries {
         pub token: String,
     }
 
+    #[derive(cynic::Enum, Clone, Copy, Debug)]
+    pub enum LogStream {
+        Stdout,
+        Stderr,
```
theduke, line 597: We are missing `Runtime` here, right? (or what did we call it exactly?)

theduke: Maybe that was not added in the backend?

lib/cli/src/commands/app/logs.rs
```diff
@@ -59,6 +59,14 @@ pub struct CmdAppLogs {
...
     /// - namespace/name@version
     ident: Identifier,
+
+    /// Enables STDOUT log stream
+    #[clap(long)]
+    stdout: bool,
+
+    /// Enables STDERR log stream
+    #[clap(long)]
+    stderr: bool,
```
theduke, line 69: I agree with @ayys  

* By default, load all streams
* Have a `--streams stdout[,...]` flag

theduke (commented):
We can potentially do this in a followup PR, but we should also allow client-side sorting of logs into streams. 

And also allow filtering by instance ID.
I'm not sure if that was a separate issue or not.

lib/backend-api/src/types.rs
maminrayej: According to the schema returned by backend, `Runtime` is not included:
```
enum LogStream {
  STDOUT
  STDERR
}
```

lib/cli/src/commands/app/logs.rs
maminrayej: I agree with using commas.

theduke (approved):

Issue #4417, opened 2024-01-29
Edge-CLI: The `wasmer app logs`: panic when no logs present


theduke:
It appears the `wasmer app logs` command can currently run into a situation where retrieving the logs panics.

## Reproduction

```
`wasmer app create -t static-website ...
wasmer app logs XXX
```

```
thread 'main' panicked at 'range end index 1000 out of range for slice of length 10', /home/theduke/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wasmer-edge-cli-0.1.4/src/cmd/app/logs.rs:113:25
```

theduke:
cc @maminrayej , might this be related to your recent refactoring?

maminrayej:
Do you mean the metrics refactor? I don't think so.
Doesn't this command just send a request to backend? If so, I just added the `--watch` flag to the code path.
I'll investigate.

PR #4452, opened 2024-02-13
fix logic error in indexing

### LINKED ISSUE #4442: Fetching logs for an app fails
When fetching logs for an app that have many log entries, the CLI fails with the following error messsage,

```sh
$ wasmer app logs <app-name>
thread 'main' panicked at 'range end index 1000 out of range for slice of length 10', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-edge-cli-0.1.4/src/cmd/app/logs.rs:113:25
```

I've made sure there are logs for the app, and those logs can be fetched via the registry.

### Steps to reproduce
```sh
$ wasmer app logs <app-name>
thread 'main' panicked at 'range end index 1000 out of range for slice of length 10', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-edge-cli-0.1.4/src/cmd/app/logs.rs:113:25
```

### Expected behavior
Logs for the app should appear.

### Actual behavior
The CLI errors out.

Could be related to #4441

maminrayej:
Resolves #4442 and #4417.


lib/cli/src/commands/app/logs.rs
```diff
@@ -113,7 +113,7 @@ impl crate::commands::AsyncCliCommand for CmdAppLogs {
         while let Some(logs) = logs_stream.next().await {
             let logs = logs?;
 
-            let limit = std::cmp::max(logs.len(), rem);
+            let limit = std::cmp::min(logs.len(), rem);
 
             let logs = &logs[..limit];
```
theduke, line 118: Let's prevent a similar issue in the future by not using slice indexing.

Actually, why are we indexing here at all?
The limit is just supposed to be for querying, why would we discard logs returned by the API?

theduke (changes requested):

lib/cli/src/commands/app/logs.rs
maminrayej: It's not just for querying. Since there is no direct relation between `max` and the maximum amount of logs returned by the backend, we have to discard logs somewhere in order to respect `max`. For example, if user wants to see at most 30 logs, and backend returns 20 at a time, we have to discard 10 logs in the second batch.

IMO Indexing is fine as long as its guard is correct. Plus, using `get` here would swallow this logical bug. This is a type of bug that should've been caught in a test, or during review. But I'm fine with using `get` as well.

theduke: Indexing can panic, and it has panicked here and can again if we change the code somehow. 

There's a reason why iterators exist, let's just use those.
It can also just be `slice.get(0..limit)`, which doesn't panic.

theduke (approved):

PR #4454, opened 2024-02-15
add pagination to app list queries

### LINKED ISSUE #4413: Edge-CLI: Fix "app list" command pagination
The `app list` command currently uses some GraphQL queries without pagination.
The backend only returns 10 apps per query.
We need to alter the code flow to allow for pagination, and add a CLI argument for it.

There also should be a `--all` flag to get all outputs.

NOTE: this is a bit complicated by the fact that the current `--all` flag tries to fetch all user-accessible apps. This should be restricted to single namespaces only.

maminrayej:
Adds pagination to `wasmer app list` in order to enable users to list all apps. `--max` is also added to limit the number of displayed apps.

Resolves #4413.

### DIFF
--- a/lib/cli/src/commands/app/list.rs
+++ b/lib/cli/src/commands/app/list.rs
@@ -32,6 +32,10 @@ pub struct CmdAppList {
     /// Maximum number of apps to display
     #[clap(long, default_value = "1000")]
     max: usize,
+
+    /// Asks whether to display the next page or not
+    #[clap(long, default_value = "false")]
+    paging_mode: bool,
 }
 
 #[async_trait::async_trait]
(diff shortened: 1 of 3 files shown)


lib/cli/src/commands/app/list.rs
```diff
@@ -32,15 +41,37 @@ impl AsyncCliCommand for CmdAppList {
...
+            let limit = std::cmp::min(apps.len(), rem);
+
+            if limit == 0 {
+                break;
+            }
+
+            display_apps.extend(apps.drain(..limit));
+
+            rem -= limit;
+        }
```
syrusakbary, line 72: I'd do this conditionally, kind of asking the user if they want to go to the next page.

maminrayej: I think it's going to be annoying if the user wants to list 30 apps and has to type `y` every 10 apps. Maybe a dedicated flag like `--paging-mode` which asks at the end of each page?

syrusakbary (approved):

PR #4455, opened 2024-02-15
Implemented an exponential CPU backoff that kicks in when a run token is not held

john-sharratt:


lib/wasix/src/os/task/process.rs
```diff
@@ -133,6 +141,22 @@ pub struct WasiProcessInner {
     /// Represents a checkpoint which blocks all the threads
     /// and then executes some maintenance action
     pub checkpoint: WasiProcessCheckpoint,
+    /// Referenced list of wakers that will be triggered
```
theduke, line 144: Can be put all related fields on a nested struct to encupsulate things a bit better?
And move all related types and this struct to a submodule?
(including the `acquire_` methods)

lib/wasix/src/syscalls/wasi/fd_read.rs
```diff
@@ -44,6 +44,7 @@ pub fn fd_read<M: MemorySize>(
         fd_entry.offset.load(Ordering::Acquire) as usize
     };
```
theduke, line 46: We probably also would want to include `sock_send` and `sock_recv`?

lib/wasix/src/os/task/process.rs
```diff
@@ -534,6 +570,147 @@ impl WasiProcess {
             thread.set_status_finished(Ok(exit_code))
         }
     }
```
theduke, line 572: Should we have an exposed API to modify the backoff behaviour at runtime?

lib/wasix/src/os/task/process.rs
john-sharratt: done

lib/wasix/src/syscalls/wasi/fd_read.rs
john-sharratt: Those syscalls are only active when the process is active (i.e. something is being actively done and we do not want to slow down) - thus adding the overhead of the check is redundant (although likely unmeasurable)

lib/wasix/src/os/task/process.rs
john-sharratt: I don't think we'll need it - so best we avoid the complexity and cost of doing this - its quite a bit harder to wake up threads should the backoff be changed and it would be unexpected behavior if it didn't wake them up when invoking the API

theduke (approved):

PR #4462, opened 2024-02-22
DProxy

john-sharratt:
Journal support for pause and resume of a WASIX compiled HTTP based server
Fixes for various aspects that were not working properly in the Jounral


lib/virtual-net/src/loopback.rs
```diff
@@ -0,0 +1,240 @@
...
+    tcp_listeners: HashMap<SocketAddr, LoopbackTcpListener>,
+    ip_addresses: Vec<IpCidr>,
+}
+
+#[derive(Debug, Clone)]
+pub struct LoopbackNetworking {
+    state: Arc<Mutex<LoopbackNetworkingState>>,
+}
+
+impl LoopbackNetworking {
```
theduke, line 28: Yay!

lib/virtual-net/src/composite.rs
```diff
@@ -0,0 +1,121 @@
+use std::net::SocketAddr;
+use std::task::{Context, Poll};
+
+use crate::{Ipv4Addr, Ipv6Addr, NetworkError, VirtualIoSource, VirtualTcpListener};
+use derivative::Derivative;
+use virtual_mio::ArcInterestHandler;
+
+#[derive(Derivative)]
+#[derivative(Debug)]
+pub struct CompositeTcpListener {
```
theduke, line 10: In what context do we need this?
It's not used anywhere in the added code I think, just exposed?

Is it used in implementations?

lib/wasix/src/journal/effector/syscalls/path_create_directory.rs
```diff
@@ -20,14 +26,18 @@ impl JournalEffector {
...
     ) -> anyhow::Result<()> {
-        crate::syscalls::path_create_directory_internal(ctx, fd, path).map_err(|err| {
-            anyhow::format_err!(
-                "journal restore error: failed to create directory path (fd={}, path={}) - {}",
-                fd,
-                path,
-                err
-            )
-        })?;
+        if fd == VIRTUAL_ROOT_FD {
```
theduke, line 29: Why is this necessary?
I assume for restoration?

Should have a comment explaining the need.

theduke: How does the restoration interact with mounting?

lib/wasix/src/journal/effector/syscalls/path_rename.rs
```diff
@@ -26,16 +28,26 @@ impl JournalEffector {
...
-        if ret != Errno::Success {
-            bail!(
-                "journal restore error: failed to rename path (old_fd={}, old_path={}, new_fd={}, new_path={}) - {}",
-                old_fd,
-                old_path,
-                new_fd,
-                new_path,
-                ret
-            );
+        if old_fd == VIRTUAL_ROOT_FD && new_fd == VIRTUAL_ROOT_FD {
```
theduke, line 31: As above, why is this needed, needs an explanatory comment. 
Can we avoid the custom logic?

lib/wasix/src/journal/effector/syscalls/path_set_times.rs
```diff
@@ -32,19 +34,23 @@ impl JournalEffector {
...
-                    flags,
-                    path,
-                    st_atim,
-                    st_mtim,
-                    fst_flags,
-                    err
-                )
-            })?;
+        if fd == VIRTUAL_ROOT_FD {
+            // we ignore this record as its not implemented yet
```
theduke, line 38: As above.

lib/wasix/src/journal/effector/syscalls/path_unlink.rs
```diff
@@ -20,14 +26,18 @@ impl JournalEffector {
...
     ) -> anyhow::Result<()> {
-        let ret = crate::syscalls::path_unlink_file_internal(ctx, fd, path)?;
-        if ret != Errno::Success {
-            bail!(
-                "journal restore error: failed to remove file (fd={}, path={}) - {}",
-                fd,
-                path,
-                ret
-            );
+        if fd == VIRTUAL_ROOT_FD {
```
theduke, line 29: As above.

lib/wasix/src/journal/effector/thread_state.rs
```diff
@@ -15,8 +27,60 @@ impl JournalEffector {
...
                 memory_stack: Cow::Owned(memory_stack.into()),
                 store_data: Cow::Owned(store_data.into()),
+                start,
+                layout,
                 is_64bit: M::is_64bit(),
             },
         )
     }
+
+    pub fn apply_thread_state<M: MemorySize>(
```
theduke, line 37: Comment explaining what this does.

lib/wasix/src/net/socket.rs
```diff
@@ -72,8 +72,28 @@ pub enum InodeSocketKind {
         socket: Box<dyn VirtualUdpSocket + Sync>,
         peer: Option<SocketAddr>,
     },
-    RemoteTcpStream {
+    RemoteSocket {
+        family: Addressfamily,
+        ty: Socktype,
+        pt: SockProto,
```
theduke, line 78: This is getting really complex, should probably live in a separate struct.

lib/wasix/src/runners/dproxy/hyper_proxy/stream.rs
```diff
@@ -0,0 +1,93 @@
...
+use futures::Stream;
+use hyper::client::connect::Connected;
+use tokio_stream::wrappers::BroadcastStream;
+use virtual_net::tcp_pair::{TcpSocketHalfRx, TcpSocketHalfTx};
+
+use super::*;
+
+#[derive(Debug)]
+pub struct HyperProxyStream {
+    pub(super) tx: TcpSocketHalfTx,
```
theduke, line 12: I'm not entirely sure why we need a special DProxy runner rather than just a journaled runner.

Is it primarily for the sharding?
And/or snapshotting after each request?

lib/wasix/src/syscalls/journal/actions/close_thread.rs
```diff
@@ -0,0 +1,33 @@
+use super::*;
+
+impl<'a, 'c> JournalReplayRunner<'a, 'c> {
```
theduke, line 3: This and the below files aren't syscalls, should they live in the syscalls directory?

lib/wasix/src/syscalls/wasix/thread_spawn.rs
```diff
@@ -89,14 +89,33 @@ pub(crate) fn thread_spawn_internal<M: MemorySize>(
...
     Span::current().record("tid", thread_id);
 
+    // Spawn the thread
+    thread_spawn_internal_phase2::<M>(ctx, thread_handle, layout, start_ptr_offset, None)?;
+
+    // Success
+    Ok(thread_id)
+}
+
+pub fn thread_spawn_internal_phase2<M: MemorySize>(
```
theduke, line 99: Comment: what are the phases, why are they needed?

lib/wasix/Cargo.toml
```diff
@@ -49,6 +49,7 @@ tokio = { version = "1", features = [
 ], default_features = false }
 tokio-stream = { version = "0.1", features = [ "sync" ] }
 futures = { version = "0.3" }
+async-ctrlc = { version = "1.2.0", features = [ "stream" ], optional = true }
```
theduke, line 52: Tokio has built-in support for this: https://docs.rs/tokio/latest/tokio/signal/fn.ctrl_c.html

lib/wasix/src/syscalls/wasix/thread_spawn.rs
john-sharratt: renamed to make it easier to understand

lib/wasix/src/syscalls/journal/actions/close_thread.rs
john-sharratt: they replay the journal by doing syscalls

john-sharratt: ill rename them to make that easier to see

lib/wasix/src/runners/dproxy/hyper_proxy/stream.rs
john-sharratt: The dproxy runner spawns and reuses instances while the dcgi runner creates them and throws them away. they are very different operating models. Dproxy actually proxies the requests as well while dcgi does not

lib/virtual-net/src/loopback.rs
john-sharratt: indeed this is good news for the browser as well

lib/wasix/src/net/socket.rs
john-sharratt: done

lib/wasix/src/journal/effector/thread_state.rs
john-sharratt: done

lib/wasix/src/journal/effector/syscalls/path_create_directory.rs
john-sharratt: Added this comment

``` rust
/// Used for interacting with the file system when it has no
/// pre-opened file descriptors at the root level. Normally
/// a WASM process will do this in the libc initialization stage
/// however that does not happen when the WASM process has never
/// been run. Further that logic could change at any time in libc
/// which would then break functionality. Instead we use this fixed
/// file descriptor
///
/// This is especially important for fuse mounting journals which
/// use the same syscalls as a normal WASI application but do not
/// run the libc initialization logic
```

lib/wasix/src/journal/effector/syscalls/path_set_times.rs
john-sharratt: See above

lib/wasix/src/journal/effector/syscalls/path_rename.rs
john-sharratt: See above

lib/virtual-net/src/composite.rs
john-sharratt: its used in `Edge` for the `DistributedNetworking`. It allows combining multiple listeners into a single listener which is useful when there are multiple switches (e.g. loopback networking and remote networking)

lib/wasix/src/fs/mod.rs
```diff
@@ -411,14 +411,68 @@ fn create_dir_all(fs: &dyn FileSystem, path: &Path) -> Result<(), virtual_fs::Fs
...
+        Self {
+            next_fd: Arc::new(AtomicU32::new(self.next_fd.load(Ordering::SeqCst))),
+        }
+    }
+
+    pub fn next_val(&self) -> WasiFd {
+        self.next_fd.fetch_add(1, Ordering::SeqCst)
+    }
+
+    pub fn set_val(&self, val: WasiFd) {
```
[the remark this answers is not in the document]
john-sharratt: Unfortunately yes... its indeed needed for the journal restoration to prevent duplicates as we can not otherwise guarantee overlaps

lib/virtual-fs/src/mem_fs/filesystem.rs
```diff
@@ -29,6 +31,20 @@ impl FileSystem {
...
     }
 
+    /// Uses a mmap'ed file as a cache for file data thus removing the
+    /// need to copy the data into memory.
+    ///
+    /// This is especially important for journals as it means that the
+    /// data stored within the journals does not need to be copied
+    /// into memory, for very large journals this would otherwise be
+    /// a problem.
+    pub fn with_backing_offload(self, buffer: OffloadBackingStore) -> Result<Self> {
```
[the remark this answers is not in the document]
john-sharratt: That does significantly increase the complexity though - more than I would like.

`mem_fs` already significantly violates the separation of concerns principle - its better to tackle this during the file system refactoring that is upcoming.

lib/virtual-fs/src/lib.rs
```diff
@@ -359,6 +359,12 @@ pub trait VirtualFile:
         None
     }
 
+    /// Writes to this file using an mmap offset and reference
+    /// (this method only works for mmap optimized file systems)
+    fn write_from_mmap(&mut self, _offset: u64, _len: u64) -> std::io::Result<()> {
```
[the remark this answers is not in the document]
john-sharratt: not without significant rework, its the same reason this exists
```
/// This method will copy a file from a source to this destination where
    /// the default is to do a straight byte copy however file system implementors
    /// may optimize this to do a zero copy
    fn copy_reference(
        &mut self,
        mut src: Box<dyn VirtualFile + Send + Sync + 'static>,
    ) -> BoxFuture<'_, std::io::Result<()>> {
        Box::pin(async move {
            let bytes_written = tokio::io::copy(&mut src, self).await?;
            tracing::trace!(bytes_written, "Copying file into host filesystem",);
            Ok(())
        })
    }
    ```
    
    Better we tackle this during the file system refactor

File: lib/virtual-fs/Cargo.toml [hunk 16]
```diff
@@ -40,6 +40,7 @@ getrandom = { version = "0.2", features = [ "js" ] }
 [dev-dependencies]
 pretty_assertions = "1.3.0"
 tempfile = "3.6.0"
+tracing-test = "0.2.4"
```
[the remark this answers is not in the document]
john-sharratt: its used in the unit tests of `offloaded_file.rs`

theduke:
@john-sharratt we should tackle the clean abstractions now rather than punting it to someone else in the upcoming refactor.

As is this adds fresh complications to the refactor that will be hard to untangle. 
We should make the refactor easier, not harder.

john-sharratt:
Tackling the abstractions now... *is* the refactor. That's the whole problem in the first place, this PR is not appropriate for that work as its significant and makes this PR too big

theduke (approved):

PR #4467, opened 2024-02-25
fix(cli): Fix auto-package version bumping in 'wasmer deploy'

### LINKED ISSUE #4464: CLI: "wasmer deploy" auto package version bump does not work if package version is out of date
Running "wasmer deploy" is supposed to automatically bump and publish the package version if a local package is used.

This does not work correctly at the moment if the package version in wasmer.toml is out of date.

We need to query the current package version and bump appropriately instead.

This currently breaks CI setups that rely on auto-bumping.

theduke:
THe backend does not return reliable version information, so we have to
introduce a loop that checks for the next available version.

Closes #4464
Closes RUN-100

### DIFF
--- a/lib/cli/src/commands/create_exe.rs
+++ b/lib/cli/src/commands/create_exe.rs
@@ -1958,7 +1958,7 @@ pub(super) mod utils {
     #[test]
     fn test_filter_tarball() {
         use std::str::FromStr;
-        let test_paths = vec![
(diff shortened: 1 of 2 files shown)


theduke:
@ayys the cause for this problem is https://linear.app/wasmer/issue/BE-206

maminrayej (approved):

PR #4470, opened 2024-02-28
chore: Completely remove wasix_http_client

theduke:
Followup to a previous removal commit.
This now completely removes all usage of wasix_http_client.
(exported definitions, implementations, etc)

Also removes the related wai-bindgen-gen-wasmer crate, which is not
needed anymore.

Followup to #4466

NOTE: removing the `wai-bindgen-gen-wasmer` crate is essentially an admission that WAI is unmaintained , and will not receive future development.

I think this is the right move.


syrusakbary (changes requested):
I'm happy with this except with removing `wai-bindgen-wasmer`. If we undo that, then we should be good to go

syrusakbary (approved):

PR #4476, opened 2024-03-06
feat: add `wasmer-argus`

xdoardo:
# Description

`wasmer-argus` is a test runner. It fetches packages from a user-defined WebContainer registry using its graphql endpoints: by default, this endpoint is `http://registry.wasmer.io/graphql`. Internally, it uses the `wasmer-api` crate, therefore the target registry must be compatible with the queries defined in that crate. 

Since registries, to fetch *every package* in them, might need an authorization token, `wasmer-argus` has support for user-defined tokens, which can be gathered automatically from the environment using the `WASMER_TOKEN` variable or from the CLI arguments with a dedicated flag. 

I tried to make the testing process as parallelised as possible. `wasmer-argus` gathers informations about the available parallelism in the host machine using `std::thread::available_parallelism`; the user can, however, specify "the amount of parallelism" desired -- which controls the number of concurrent tests -- using another dedicated CLI flag.

Further documentation can be found in the `README.md`. 

## Further steps
* [ ] better documentation (`--help` is already quite explainatory)
* [ ] a bit more testing
* [ ]  gather feedback

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,30 @@ Looking for changes that affect our C API? See the [C API Changelog](lib/c-api/C
 
 ## **Unreleased**
 
+## 4.2.7 - 19/03/2024
+
+This release adds the `wasmer domain` command for DNS records management, and also includes an important fix to the `stack_restore` WASIX syscall (used by the `longjmp` function).
+
+## Added
+
+  - [#4478](https://github.com/wasmerio/wasmer/pull/4478) chore(backend-api): Add size to PackageDistribution
+
+## Changed
+
+  - [#4492](https://github.com/wasmerio/wasmer/pull/4492) No longer restoring the thread local memory when we longjmp
+  - [#4487](https://github.com/wasmerio/wasmer/pull/4487) Manage DNS records
+  - [#4220](https://github.com/wasmerio/wasmer/pull/4220) Ability to detect a tainted instance
+  - [#4455](https://github.com/wasmerio/wasmer/pull/4455) Implemented an exponential CPU backoff that kicks in when a run token is not held
+  - [#4470](https://github.com/wasmerio/wasmer/pull/4470) chore: Completely remove wasix_http_client
+
+## Fixed
+
+  - [#4490](https://github.com/wasmerio/wasmer/pull/4490) Fix for a panic in the sock_recv when a file handle is missing
+  - [#4335](https://github.com/wasmerio/wasmer/pull/4335) Fixed an issue where the package loader was blocking the tokio runtime
+  - [#4473](https://github.com/wasmerio/wasmer/pull/4473) fix: fix feature = "cargo-clippy" deprecation
+
+
+
 ## 4.2.6 - 03/03/2024
 
 This release includes a number of DX improvements for the Wasmer CLI, as well as fixes to WASI and its filesystem implementation.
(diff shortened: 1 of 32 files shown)


tests/wasmer-argus/src/argus/packages.rs
```diff
@@ -0,0 +1,217 @@
...
+                .template(&format!(
+                    "[{test_id}] [{{bar:40.cyan/blue}}] {{bytes}}/{{total_bytes}} - {{msg}}"
+                ))
+                .unwrap()
+                .progress_chars("#>-"),
+        );
+
+        p.set_message(format!("downloading from {url}"));
+
+        let mut outfile = match File::create(&path.join("package.webc")) {
```
theduke, line 122: Using a sync file in an async context, this should be a tokio File.

tests/wasmer-argus/src/argus/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+        );
+
+        p.enable_steady_tick(Duration::from_millis(100));
+
+        p.set_message("package downloaded");
+        let filepath = path.join("package.webc");
+
+        let start = time::Instant::now();
+
+        let res = match std::panic::catch_unwind(|| {
```
theduke, line 152: nit: a huge expression in a match is hard to read, I'd move this out to a variable first

theduke (commented):
Mostly just nitpicks, and one thing to fix.

tests/wasmer-argus/src/argus/mod.rs
xdoardo: This and the other small changes: [6195830](https://github.com/wasmerio/wasmer/pull/4476/commits/619583065ca574c1ce0e958ffcb9c990af90eac6)

tests/wasmer-argus/src/argus/packages.rs
xdoardo: Edited in [9f066cc](https://github.com/wasmerio/wasmer/pull/4476/commits/9f066cccd38ac80756f25d4c6d2bab1be2bf6110)

theduke:
@xdoardo one of the tests is failing: https://github.com/wasmerio/wasmer/actions/runs/8185273449/job/22381307580?pr=4476#step:12:3263

tests/wasmer-argus/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+cynic = "3.4.3"
+url = "2.5.0"
+futures = "0.3.30"
+tracing = "0.1.40"
+tokio = { version = "1.36.0", features = ["rt-multi-thread", "sync", "time", "fs"] }
+clap = {version = "4.4.11", features = ["derive", "string"]}
+tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
+serde = { version = "1.0.197", features = ["derive"] }
+serde_json = "1.0.114"
+wasmer = { version = "4.2.6", path = "../../lib/api", features = ["engine", "core", "singlepass", "cranelift", "llvm"] }
```
syrusakbary, line 34: Rather than depending on wasmer (crate), should we just depend on the CLI instead (not the crate, but just calling a subprocess with `wasmer compile xyz.wasm --singlepass` or `wasmer compile xyz.wasm --llvm`.

That way we don't need to constantly recompile wasmer

xdoardo: [edd40c2](https://github.com/wasmerio/wasmer/pull/4476/commits/edd40c2e6545859582a25e4702c5da516092f7aa) adds a feature that enables user to use both the CLI and the library. By default,  only the CLI will be used.

PR #4483, opened 2024-03-11
feat(backend-api): Add DNS Record queries

theduke:
- **Update backend-api schema**
- **feat(backend-api): Add types and queries for DNS records**
- **docs(backend-api): Fix cynic web ui link in README**


theduke:
Subsumed by #4487

PR #4487, opened 2024-03-13
Manage DNS records

ayys:
This adds subcommands to manage domains and DNS records.
This PR gives you the ability to list, get, and sync zone files for domains.

Merges PR #4483 

``` shell
$ wasmer domains list # list all domains

$ wasmer domains list <namespace> # list all domains in the namespace

$ wasmer domain get <domain-name>

$ wasmer domain get-zone-file <domain-name>

$ wasmer domain sync-zone-file <zone-file> [--no-delete-missing-records]
```

### DIFF
--- a/lib/backend-api/src/query.rs
+++ b/lib/backend-api/src/query.rs
@@ -865,6 +865,29 @@ pub async fn get_domain_with_records(
     Ok(opt)
 }
 
+/// Register a new domain
+pub async fn register_domain(
+    client: &WasmerClient,
+    name: String,
+    namespace: Option<String>,
+    import_records: Option<bool>,
+) -> Result<types::DnsDomain, anyhow::Error> {
+    let vars = types::RegisterDomainVars {
+        name,
+        namespace,
+        import_records,
+    };
+    let opt = client
+        .run_graphql_strict(types::RegisterDomain::build(vars))
+        .await
+        .map_err(anyhow::Error::from)?
+        .register_domain
+        .context("Domain registration failed")?
+        .domain
+        .context("Domain registration failed, no associatede domain found.")?;
+    Ok(opt)
+}
+
 /// Retrieve all DNS records.
 ///
 /// NOTE: this is a privileged operation that requires extra permissions.
--- a/lib/backend-api/src/query.rs
+++ b/lib/backend-api/src/query.rs
@@ -883,12 +883,19 @@ pub async fn get_all_dns_records(
 pub async fn get_all_domains(
     client: &WasmerClient,
     vars: types::GetAllDomainsVariables,
-) -> Result<types::DnsDomainConnection, anyhow::Error> {
-    client
+) -> Result<Vec<DnsDomain>, anyhow::Error> {
+    let connection = client
         .run_graphql_strict(types::GetAllDomains::build(vars))
         .await
         .map_err(anyhow::Error::from)
         .map(|x| x.get_all_domains)
+        .context("no domains returned")?;
+    Ok(connection
+        .edges
+        .into_iter()
+        .flatten()
+        .filter_map(|x| x.node)
+        .collect())
 }
 
 /// Retrieve a domain by its name.
@@ -948,19 +955,13 @@ pub async fn upsert_domain_from_zone_file(
         delete_missing_records: Some(delete_missing_records),
     };
     let res = client
-        .run_graphql_raw(types::UpsertDomainFromZoneFile::build(vars))
+        .run_graphql_strict(types::UpsertDomainFromZoneFile::build(vars))
         .await?;
 
-    if let Some(domain) = res
-        .data
-        .and_then(|d| d.upsert_domain_from_zone_file)
-        .map(|d| d.domain)
-    {
-        Ok(domain)
-    } else {
-        Err(GraphQLApiFailure::from_errors(
-            "could not sync zone file",
-            res.errors,
-        ))
-    }
+    let domain = res
+        .upsert_domain_from_zone_file
+        .context("Upserting domain from zonefile failed")?
+        .domain;
+
+    Ok(domain)
 }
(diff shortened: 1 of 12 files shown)


lib/cli/src/commands/domain/get.rs
```diff
@@ -0,0 +1,32 @@
...
+impl AsyncCliCommand for CmdDomainGet {
+    type Output = ();
+
+    async fn run_async(self) -> Result<(), anyhow::Error> {
+        let client = self.api.client()?;
+        if let Some(domain) = wasmer_api::query::get_domain_with_records(&client, self.name).await?
+        {
+            println!("{}", self.fmt.format.render(&domain));
+        } else {
+            println!("Domain not found");
```
theduke, line 28: Return error so the command fails with a non-zero exit code.

Eg: `anyhow::bail!("Domain not found");`

lib/cli/src/commands/domain/get.rs
```diff
@@ -0,0 +1,32 @@
...
+        }
+        Ok(())
```
theduke, line 30: We probably want to fail with a not found error here.

lib/cli/src/commands/domain/list.rs
```diff
@@ -0,0 +1,47 @@
...
+        let domains_connection = wasmer_api::query::get_all_domains(
+            &client,
+            GetAllDomainsVariables {
+                first: None,
+                after: None,
+                namespace: self.namespace,
+            },
+        )
+        .await?;
+        let domains = domains_connection
```
theduke, line 35: We can leave this for later, but this will need pagination / a --all flag to fetch all domains.

lib/cli/src/commands/domain/zonefile.rs
```diff
@@ -0,0 +1,80 @@
...
+    api: ApiOpts,
+
+    /// filename of  zone-file to sync
+    zone_file_path: String,
+
+    /// Do not delete records that are not present in the zone file
+    #[clap(
+        short = 'n',
+        long = "no-delete-missing-records",
+        required = false,
```
theduke, line 36: Neither required nor default_valueis required here, these are automatic.

lib/cli/src/commands/domain/zonefile.rs
```diff
@@ -0,0 +1,80 @@
...
+    type Output = ();
+
+    async fn run_async(self) -> Result<(), anyhow::Error> {
+        let client = self.api.client()?;
+        if let Some(domain) =
+            wasmer_api::query::get_domain_zone_file(&client, self.domain_name).await?
+        {
+            let zone_file_contents = domain.zone_file;
+            if let Some(zone_file_path) = self.zone_file_path {
+                std::fs::write(zone_file_path, zone_file_contents).expect("Unable to write file");
```
theduke, line 53: Don't use expect, return the error instead.

Hint: you can use `.context("unable ...")?`, (you need to import `use anyhow::Context` first)

lib/cli/src/commands/domain/zonefile.rs
```diff
@@ -0,0 +1,80 @@
...
+            } else {
+                println!("{}", zone_file_contents);
+            }
+        } else {
+            println!("Domain not found");
```
theduke, line 58: Return error ( eg `anyhow::bail!("Domain not found)`)

lib/cli/src/commands/domain/list.rs
```diff
@@ -0,0 +1,47 @@
...
+            .edges
+            .into_iter()
+            .map(|edge| {
+                edge.expect("domain not found")
```
theduke, line 39: No `.expect()`, need to return an error instead.

lib/cli/src/commands/domain/zonefile.rs
```diff
@@ -0,0 +1,80 @@
...
+        Ok(())
+    }
+}
+
+#[async_trait::async_trait]
+impl AsyncCliCommand for CmdZoneFileSync {
+    type Output = ();
+
+    async fn run_async(self) -> Result<(), anyhow::Error> {
+        let data = std::fs::read(&self.zone_file_path).expect("Unable to read file");
```
theduke, line 69: No `.expect()`.

lib/cli/src/types.rs
```diff
@@ -1,8 +1,73 @@
...
+                "Name".to_string(),
+                "TTL".to_string(),
+                "Value".to_string(),
+            ]);
+        let mut rows: Vec<Vec<String>> = vec![];
+        if let Some(ref records) = self.records {
+            records.iter().flatten().for_each(|record| {
+                rows.push(vec![
+                    record.record_type().to_string(),
+                    record.name().expect("Expected record name").to_string(),
```
theduke, line 45: No `.expect()`, use a stub value instead , eg `.unwrap_or_else(|| "<no name>".to_string()`

lib/backend-api/src/query.rs
```diff
@@ -826,3 +936,31 @@ fn unix_timestamp(ts: OffsetDateTime) -> f64 {
...
+    client: &WasmerClient,
+    zone_file_contents: String,
+    delete_missing_records: bool,
+) -> Result<DnsDomain, anyhow::Error> {
+    let vars = UpsertDomainFromZoneFileVars {
+        zone_file: zone_file_contents,
+        delete_missing_records: Some(delete_missing_records),
+    };
+    let res = client
+        .run_graphql_raw(types::UpsertDomainFromZoneFile::build(vars))
```
theduke, line 951: Should use `.run_graphql_strict()` here to fail on any returned GraphQL errors.

theduke (changes requested):

lib/cli/src/opts.rs
```diff
@@ -95,15 +95,23 @@ impl ApiOpts {
...
 pub struct ItemFormatOpts {
-    /// Output format. (json, text)
+    /// Output format. (yaml, json, table)
     #[clap(short = 'f', long, default_value = "yaml")]
     pub format: crate::utils::render::ItemFormat,
 }
 
+/// Formatting options for a single item.
+#[derive(clap::Parser, Debug, Default)]
+pub struct ItemTableFormatOpts {
```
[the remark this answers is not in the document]
ayys: Resolved on slack: https://wasmerio.slack.com/archives/C03MX4KL6KH/p1710759045973139?thread_ts=1710758497.967439&cid=C03MX4KL6KH

theduke (approved):

syrusakbary (approved):

PR #4514, opened 2024-03-22
Fix for snapshots in certain use cases

john-sharratt:


lib/wasix/src/state/builder.rs
```diff
@@ -1032,7 +1042,23 @@ impl WasiEnvBuilder {
         module_hash: ModuleHash,
         mut store: Store,
     ) -> Result<(), WasiRuntimeError> {
+        let attach_ctrl_c = self.attach_ctrl_c;
+
         let (_, env) = self.instantiate_ext(module, module_hash, &mut store)?;
+
+        // Install the ctrl-c handler
+        #[cfg(feature = "ctrlc")]
+        if attach_ctrl_c {
```
theduke, line 1051: Putting this logic here in the runner seems a bit hacky to me, can't we do this on the outside without having to smush it into the runner?

Isn't it enough to be in the CLI?

john-sharratt: It's also in the runner (which is in the CLI) - however there is another code path left in the CLI which does not use the runner and hence comes down this path. The fix is to unify the code of both into one, that's quite a bit more work than this patch.

theduke (approved):

Issue #4493, opened 2024-03-18
Support Function Externals


syrusakbary:

xdoardo:
A first step towards this goal is #4501.

PR #4515, opened 2024-03-25
Experimental: Add interpreter support for Wasmer

syrusakbary:
This PR adds support for running Wasm modules in interpreted mode (in this case, via WAMR).

Progress:
* [x] Make the baseline functionality work
* [x] Most of the examples work
* [x] Most of the test work
* [x] Integrate it into CI
* [x] Document the differences when using WAMR (for example: memory grow can't be called outside of the host)
* [x] Test this on iOS
* [x] Fix https://github.com/wasmerio/wasmer/issues/4971
* [ ] Improve testing to test way more example and test cases


examples/memory.rs
```diff
@@ -108,43 +108,50 @@ fn main() -> anyhow::Result<()> {
...
 
     Ok(())
 }
 
 #[test]
-#[cfg_attr(feature = "wasm-c-api", ignore = "wasm-c-api has a different memory testuite")]
+#[cfg_attr(
+    feature = "wasm-c-api",
+    ignore = "wasm-c-api has a different memory testuite"
+)]
```
syrusakbary, line 154: We can probably remove this, since we are skipping the failing tests

Cargo.toml
```diff
@@ -128,6 +124,7 @@ default = ["wat", "wast", "cache", "wasi", "engine", "emscripten", "middlewares"
 # That means: that is able to execute modules
 backend = []
 jsc = ["wasmer/jsc", "backend", "wat", "wasmer/std"]
+wasm-c-api = ["wasmer/wasm-c-api", "backend", "wat", "wasmer/std"]
```
syrusakbary, line 127: We should add a new one: `wamr = ["wasm-c-api"]`

xdoardo: f2148f462fc398b0676a8587e97fa010b8e1e90a

lib/cli/src/error.rs
```diff
@@ -27,11 +26,12 @@ macro_rules! warning {
     })
 }
 
-#[cfg(not(feature = "jsc"))]
+#[cfg(not(any(feature = "jsc", feature = "wasm-c-api")))]
 impl PrettyError {
     /// Process a `Result` printing any errors and exiting
     /// the process after
     pub fn report<T>(result: Result<T, Error>) -> ! {
+        use wasmer::RuntimeError;
```
syrusakbary, line 34: Can we move the import up, and do it with a cfg decorator instead?

tests/integration/cli/tests/run.rs
```diff
@@ -61,7 +65,11 @@ fn run_python_create_temp_dir_in_subprocess() {
         .output()
         .unwrap();
 
-    assert_eq!(output.stdout, "0".as_bytes().to_vec());
+    if cfg!(not(feature = "wamr")) {
+        assert_eq!(output.stdout, "0".as_bytes().to_vec());
+    } else {
+        assert!(output.status.success())
+    }
```
syrusakbary, line 72: This is extremely weird. the stdout should be the same

tests/integration/cli/tests/run.rs
```diff
@@ -500,7 +536,11 @@ fn wasi_runner_on_disk() {
         .env("RUST_LOG", &*RUST_LOG)
         .assert();
 
-    assert.success().stdout(contains("Hello, World!"));
+    if cfg!(not(feature = "wamr")) {
+        assert.success().stdout(contains("Hello, World!"));
+    } else {
+        assert.success().stderr(contains("Hello, World!"));
+    }
```
syrusakbary, line 543: Why stderr on wamr but stdout normally? This seems like wamr is doing things that it shouldn't.

I found `wasm_runtime_init_wasi` to be the clulprit. But it seems we are not using it.. right?

tests/integration/cli/tests/run.rs
xdoardo: It's not that weird, the reason is that wamr prints some warnings to stdout: 
```
[tests/integration/cli/tests/run.rs:68] String::from_utf8(output.stdout.clone()).unwrap() = "[01:54:07:569 - 170257000]: doesn't support import memories and tables for now, ignore them\n"
```

tests/integration/cli/tests/run.rs
```diff
@@ -9,7 +9,11 @@ use std::{
 
 use assert_cmd::{assert::Assert, prelude::OutputAssertExt};
 use once_cell::sync::Lazy;
-use predicates::str::contains;
+use predicates::{
+    boolean::AndPredicate,
+    str::{contains, is_match},
+    Predicate,
+};
```
syrusakbary, line 16: Note: undo all the sdout/sderr changes on this file

tests/integration/cli/tests/run.rs
```diff
@@ -164,12 +179,15 @@ fn test_wasmer_run_pirita_works() {
         .assert()
         .success();
 
-    assert.stdout("hello\n");
+    if cfg!(not(feature = "wamr")) {
+        assert.stdout("hello\n");
+    } else {
+        assert.stdout(contains("hello\n"));
+    }
```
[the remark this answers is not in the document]
xdoardo: As mentioned above, WAMR can print spurious messages to stdout. Added a comment in the code to explain this.

tests/integration/cli/tests/run.rs
```diff
@@ -228,7 +250,11 @@ fn test_wasmer_run_works() {
         .assert()
         .success();
 
-    assert.stdout("hello\n");
+    if cfg!(not(feature = "wamr")) {
+        assert.stdout("hello\n");
+    } else {
+        assert.stderr(contains("hello\n"));
+    }
```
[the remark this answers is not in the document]
xdoardo: I removed the `ignore`s.

xdoardo:
Note: merging this PR closes issues https://github.com/wasmerio/wasmer/issues/4496, https://github.com/wasmerio/wasmer/issues/4495, https://github.com/wasmerio/wasmer/issues/4494, https://github.com/wasmerio/wasmer/issues/4493 and https://github.com/wasmerio/wasmer/issues/4486 as well.

lib/api/src/c_api/mod.rs
```diff
@@ -33,6 +33,9 @@ mod lib {
     }
 }
 
+/// A constant that indicates to users that the WAMR (interpreter) backend is in use. 
+pub const WASMER_WAMR: bool = true;
```
syrusakbary, line 37: We should not use this. The engine should return this @xdoardo

syrusakbary: Here: https://github.com/wasmerio/wasmer/pull/4515/files#diff-5d10c39c36980e1ba420d516d4bd3d7202ddf63c08a3366940d73dcd3f522868R30

syrusakbary:
Merging this into 5.0 so then we can review and merge wasmi and v8

PR #4525, opened 2024-03-27
Fix bug with `app deploy`: app URL is stale

ayys:
The app was being fetched before the new version was published. Now that aliases are properly changed on the backend, this resulted in stale URL being pinged.

This fixes by querying the backend after the app has been deployed.

### DIFF
--- a/lib/cli/src/commands/app/mod.rs
+++ b/lib/cli/src/commands/app/mod.rs
@@ -115,7 +115,7 @@ pub enum WaitMode {
 /// Same as [Self::deploy], but also prints verbose information.
 pub async fn deploy_app_verbose(
     client: &WasmerClient,
-    mut opts: DeployAppOpts<'_>,
+    opts: DeployAppOpts<'_>,
 ) -> Result<(DeployApp, DeployAppVersion), anyhow::Error> {
     let owner = &opts.owner;
     let app = &opts.app;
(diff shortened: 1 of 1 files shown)


lib/cli/src/commands/app/mod.rs
```diff
@@ -130,20 +130,6 @@ pub async fn deploy_app_verbose(
 
     eprintln!("Deploying app {pretty_name}...\n");
 
-    let (owner, app_opt) = if let Some(owner) = owner {
```
theduke, line 133: I think there was a reason for this pre-fetch, because otherwise there could be some odd behaviour when the owner didn't match the app id.
I don't remember the exact details.
Need to make sure the backend does the right thing without this.

What happens if the specified app_id does not match a manually specified `--owner`?

ayys: the backend has validation for that, so we should be good.

ayys:
```
╰─▶ 2: GraphQL API failure: Expected the app specified by app_id (`app_id`) to be `test-tiinyhost`, but found 
`ayush-test-tiinyhost` instead. Check your `name` field in the config to make sure it's correct.
```

theduke (approved):

Issue #4422, opened 2024-01-30
Improve Singlepass performance on Aarch64 (popcnt, select)


syrusakbary:
In ARM, Wasmer (with singlepass compiler) is not generating optimal code for the following opcodes:
* i32PopCnt: https://github.com/wasmerio/wasmer/blob/master/lib/compiler-singlepass/src/machine_arm64.rs#L3070-L3115
* i64PopCnt: https://github.com/wasmerio/wasmer/blob/master/lib/compiler-singlepass/src/machine_arm64.rs#L5146-L5189
* Select

We should aim for the Aarch64 code to be as optimal to the x64

PR #4526, opened 2024-03-27
feat(singlepass): use SIMD insts for popcount

xdoardo:
Tackles #4422.

### DIFF
--- a/lib/compiler-singlepass/src/machine_arm64.rs
+++ b/lib/compiler-singlepass/src/machine_arm64.rs
@@ -113,8 +113,6 @@ pub struct MachineARM64 {
     pushed: bool,
     /// Vector of unwind operations with offset
     unwind_ops: Vec<(usize, UnwindOps)>,
-    /// The actual compilation target.  
-    target: Option<Target>,
     /// A boolean flag signaling if this machine supports NEON.
     has_neon: bool,
 }
@@ -142,6 +140,9 @@ enum ImmType {
 #[allow(dead_code)]
 impl MachineARM64 {
     pub fn new(target: Option<Target>) -> Self {
+        // If and when needed, checks for other supported features should be
+        // added as boolean fields in the struct to make checking if such
+        // features are available as cheap as possible.
         let has_neon = match target {
             Some(ref target) => target.cpu_features().contains(CpuFeature::NEON),
             None => false,
@@ -156,7 +157,6 @@ impl MachineARM64 {
             src_loc: 0,
             pushed: false,
             unwind_ops: vec![],
-            target,
             has_neon,
         }
     }
--- a/lib/compiler-singlepass/src/machine_arm64.rs
+++ b/lib/compiler-singlepass/src/machine_arm64.rs
@@ -143,7 +143,7 @@ enum ImmType {
 impl MachineARM64 {
     pub fn new(target: Option<Target>) -> Self {
         let has_neon = match target {
-            Some(target) => target.cpu_features().contains(CpuFeature::NEON),
+            Some(ref target) => target.cpu_features().contains(CpuFeature::NEON),
             None => false,
         };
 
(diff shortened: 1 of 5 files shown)


lib/compiler-singlepass/src/machine_arm64.rs
```diff
@@ -5185,46 +5174,37 @@ impl Machine for MachineARM64 {
...
-            })?;
-            temps.push(tmp);
-            Location::GPR(tmp)
-        };
-        let label_loop = self.assembler.get_label();
-        let label_exit = self.assembler.get_label();
-        self.assembler
-            .emit_mov(Size::S32, Location::GPR(GPR::XzrSp), dest)?; // dest <= 0
-        self.assembler.emit_cbz_label(Size::S64, src, label_exit)?; // src == 0, then goto label_exit
-        self.assembler.emit_label(label_loop)?;
```
syrusakbary, line 5213: We need to use this codepath if the chipset doesn't support NEON (fallback).
So please wrap the old / new code in a if/else

lib/compiler-singlepass/src/machine_arm64.rs
```diff
@@ -3072,49 +3072,38 @@ impl Machine for MachineARM64 {
...
-            let tmp = self.acquire_temp_gpr().ok_or_else(|| {
-                CompileError::Codegen("singlepass cannot acquire temp gpr".to_owned())
-            })?;
-            temps.push(tmp);
-            Location::GPR(tmp)
-        };
-        let label_loop = self.assembler.get_label();
-        let label_exit = self.assembler.get_label();
-        self.assembler
-            .emit_mov(Size::S32, Location::GPR(GPR::XzrSp), dest)?; // 0 => dest
```
syrusakbary, line 3101: We need to use this codepath if the chipset doesn't support NEON (fallback).
So please wrap the old / new code in a if/else

syrusakbary (changes requested):
Please use Wasmer::CPUFeatures to detect if chipset is NEON (using https://doc.rust-lang.org/std/arch/macro.is_aarch64_feature_detected.html )

(note: we should update CPUFeatures to have a new enum choice for NEON)

https://naqvis.github.io/wasmer-crystal/Wasmer/CpuFeatures.html

And update the code accordingly.

lib/compiler-singlepass/src/machine_arm64.rs
xdoardo: see 4a43746

lib/compiler-singlepass/src/machine_arm64.rs
```diff
@@ -3072,38 +3075,90 @@ impl Machine for MachineARM64 {
...
-            CompileError::Codegen("singlepass cannot acquire temp gpr".to_owned())
-        })?;
-        neon_temps.push(neon_temp);
+        if self.target.is_some()
+            && self
+                .target
+                .as_ref()
+                .unwrap()
+                .cpu_features()
+                .contains(CpuFeature::NEON)
```
syrusakbary, line 3084: This might be a bit time-consuming if the `i32_popcnt` is called a lot (which can happen), I think we can rather have this precomputed on machine creation, and simply do:
`self.supports_neon`

Thoughts @xdoardo ?

lib/compiler-singlepass/src/machine_arm64.rs
```diff
@@ -5173,36 +5228,89 @@ impl Machine for MachineARM64 {
...
```
syrusakbary, line 5237: This might be a bit time-consuming if the `i64_popcnt` is called a lot (which can happen), I think we can rather have this precomputed on machine creation, and simply do:
self.supports_neon

Thoughts @xdoardo ?

syrusakbary (approved):
Looks good once the last comments are addressed

lib/compiler-singlepass/src/machine_arm64.rs
xdoardo: Addressed in  [f553d9f](https://github.com/wasmerio/wasmer/pull/4526/commits/f553d9fe871e4ad091c57f423e5edc9461dc9fac)

lib/compiler-singlepass/src/machine_arm64.rs
xdoardo: Addressed in  [f553d9f](https://github.com/wasmerio/wasmer/pull/4526/commits/f553d9fe871e4ad091c57f423e5edc9461dc9fac)

PR #4527, opened 2024-03-28
allow owner field in app.yaml

ayys:
- [x] support `owner` field in app.yaml
- [x] add `owner` field to app.yaml generated by `wasmer app create`

makes it so app.yaml can process owner field in two ways,

In descending order of priority:
- via the `--owner` flag passed to the CLI
- via the `owner` field in the app.yaml

sample config:

```yaml
kind: wasmer.io/App.v0
name: my-fun-little-app
owner: <user or namespace> # new owner field
package: wasmer/hello
```

### DIFF
--- a/lib/cli/src/commands/app/mod.rs
+++ b/lib/cli/src/commands/app/mod.rs
@@ -117,7 +117,7 @@ pub async fn deploy_app_verbose(
     client: &WasmerClient,
     opts: DeployAppOpts<'_>,
 ) -> Result<(DeployApp, DeployAppVersion), anyhow::Error> {
-    let owner = &opts.owner.or_else(|| opts.app.owner.clone());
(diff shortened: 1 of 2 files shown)


lib/cli/src/commands/app/mod.rs
```diff
@@ -117,7 +117,7 @@ pub async fn deploy_app_verbose(
     client: &WasmerClient,
     opts: DeployAppOpts<'_>,
 ) -> Result<(DeployApp, DeployAppVersion), anyhow::Error> {
-    let owner = &opts.owner;
+    let owner = &opts.owner.or_else(|| opts.app.owner.clone());
```
theduke, line 120: While we are at it, let's also update the `app create` command to add the `owner` field.

ayys: Done

theduke (approved):

PR #4532, opened 2024-04-01
unnamed packages

ayys:
this adds support for unnamed packages, which are packages that don't have a
name.

This is useful for packages that are not meant to be used by other
people/packages but are used to deploy apps.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1370,6 +1370,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fea41bba32d969b513997752735605054bc0dfa92b4c56bf1189f2e174be7a10"
 
 [[package]]
+name = "document-features"
+version = "0.2.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ef5282ad69563b5fc40319526ba27e0e7363d552a896f0297d54f767717f9b95"
+dependencies = [
+ "litrs",
+]
+
+[[package]]
 name = "dyn-clone"
 version = "1.0.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1890,6 +1899,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d2fabcfbdc87f4758337ca535fb41a6d701b65693ce38287d856d1674551ec9b"
 
 [[package]]
+name = "globset"
+version = "0.4.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "57da3b9b5b85bd66f31093f8c408b90a74431672542466497dcbdfdc02034be1"
+dependencies = [
+ "aho-corasick",
+ "bstr 1.9.1",
+ "log 0.4.21",
+ "regex-automata 0.4.6",
+ "regex-syntax 0.8.2",
+]
+
+[[package]]
 name = "graphql-introspection-query"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2311,6 +2333,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "ignore"
+version = "0.4.22"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b46810df39e66e925525d6e38ce1e7f6e1d208f72dc39757880fcb66e2c58af1"
+dependencies = [
+ "crossbeam-deque 0.8.5",
+ "globset",
+ "log 0.4.21",
+ "memchr",
+ "regex-automata 0.4.6",
+ "same-file",
+ "walkdir",
+ "winapi-util",
+]
+
+[[package]]
 name = "indexmap"
 version = "1.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2627,6 +2665,12 @@ name = "linux-raw-sys"
 version = "0.4.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "01cda141df6706de531b6c46c3a33ecca755538219bd484262fa09410c13539c"
+
+[[package]]
+name = "litrs"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b4ce301924b7887e9d637144fdade93f9dfff9b60981d4ac161db09720d39aa5"
 
 [[package]]
 name = "llvm-sys"
@@ -5755,7 +5799,7 @@ dependencies = [
  "tracing",
  "tracing-test",
  "typetag",
- "webc",
+ "webc 5.9.0",
 ]
 
 [[package]]
@@ -6204,7 +6248,7 @@ dependencies = [
  "tracing",
  "url",
  "uuid",
- "webc",
+ "webc 5.8.1",
 ]
 
 [[package]]
@@ -6228,7 +6272,7 @@ dependencies = [
  "url",
  "wasmer",
  "wasmer-api",
- "webc",
+ "webc 5.9.0",
 ]
 
 [[package]]
@@ -6273,7 +6317,7 @@ dependencies = [
  "wasmer-middlewares",
  "wasmer-types",
  "wasmer-wasix",
- "webc",
+ "webc 5.9.0",
 ]
 
 [[package]]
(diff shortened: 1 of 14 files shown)


syrusakbary (changes requested):
This is failing on the branch (using the prod registry):


```
$ wasmer run python/python
error: Unable to find "python/python" in the registry
│   1: "https://registry.wasmer.io/graphql" replied with 400 Bad Request
╰─▶ 2: "https://registry.wasmer.io/graphql" replied with 400 Bad Request
```

lib/wasix/src/bin_factory/binary_package.rs
```diff
@@ -69,7 +70,6 @@ pub struct BinaryPackage {
     pub webc_fs: Arc<dyn FileSystem + Send + Sync>,
     pub commands: Vec<BinaryPackageCommand>,
     pub uses: Vec<String>,
```
syrusakbary, line 72: Should this be using `Vec<PackageId>` eventually? We may need to create a ticket about this so we don't forget

lib/wasix/Cargo.toml
```diff
@@ -102,13 +104,13 @@ web-sys = { version = "0.3.64", features = [
...
 default-features = false
-features = ["rustls-tls", "json", "stream"]
+features = ["rustls-tls", "json", "stream", "blocking"]
 optional = true
 
 [target.'cfg(target_arch = "riscv64")'.dependencies.reqwest]
 version = "0.11"
 default-features = false
-features = ["native-tls", "json", "stream"]
+features = ["native-tls", "json", "stream", "blocking"]
```
syrusakbary, line 113: We don't need blocking any longer

lib/backend-api/src/query.rs
```diff
@@ -24,10 +25,21 @@ use crate::{
...
 pub async fn fetch_webc_package(
     client: &WasmerClient,
-    ident: &WebcIdent,
+    ident: &PackageIdent,
     default_registry: &Url,
 ) -> Result<webc::compat::Container, anyhow::Error> {
-    let url = ident.build_download_url_with_default_registry(default_registry);
+    let url = match ident {
+        PackageIdent::Named(n) => Url::parse(&format!(
+            "{default_registry}/{}:{}",
```
theduke, line 33: `Url::to_string()` ends with a trailing slash, so this would have double slashes.
We could just clone the URL and use `Url::set_path()`.

But also, note that `NamedPackageIdent` has a `registry` field.

Also, this method needs an optional auth token, because unnamed private packages won't be downloadable otherwise, but we can do that as a followup.

But since this isn't actually actively used anywhere, we can do all of that in a followup to not block the merge.
=> create followup issue, @xdoardo

lib/cli/src/commands/package/download.rs
```diff
@@ -93,43 +90,53 @@ impl PackageDownload {
...
+                    client.with_auth_token(token.clone())
+                } else {
+                    client
+                };
+
+                let rt = tokio::runtime::Runtime::new()?;
+                let pkg = rt.block_on(wasmer_api::query::get_package_release(&client, &hash.to_string()))?
+                    .with_context(|| format!("Package with {hash} does not exist in the registry, or is not accessible"))?;
+
+                (pkg.webc_url, token)
```
syrusakbary, line 135: I believe we fixed this... but we didn't @maminrayej can you address this to just get the hash directly from the backend?

syrusakbary: We should not use a new query to the backend for this, we can just retrieve the hash directly

lib/cli/src/commands/app/create.rs
```diff
@@ -1,34 +1,54 @@
...
-    },
-    opts::{ApiOpts, ItemFormatOpts},
-    utils::package_wizard::{CreateMode, PackageType, PackageWizard},
-};
+use super::deploy::CmdAppDeploy;
+
+async fn write_app_config(app_config: &AppConfigV1, dir: Option<PathBuf>) -> anyhow::Result<()> {
+    let raw_app_config = app_config.clone().to_yaml()?;
+
+    let app_dir = match dir {
```
theduke, line 27: very tiny nitpick: seeems to me that the caller should always pass in a valid path here, and should be resposible for falling back to current dir

lib/cli/src/commands/app/create.rs
```diff
@@ -1,34 +1,54 @@
...
-    template: Option<AppType>,
-
+    pub template: Option<AppType>,
+    /// Whether or not to deploy the application once it is created.
+    ///
+    /// If selected, this might entail the step of publishing the package related to the
+    /// application. By default, the application is not deployed and the package is not published.
     #[clap(long)]
-    publish_package: bool,
+    pub deploy_app: bool,
```
theduke, line 51: Should this just be `--deploy`?
The `-app` seems redundant.

Being annoying here because flags are public API and should be changed very rarely.

lib/cli/src/commands/package/download.rs
theduke: There is no download URL in the frontend/backend to download by hash correctly.

The backend will need to implement this, so it's a followup task.

syrusakbary: Got it, resolving the question

lib/wasix/src/bin_factory/binary_package.rs
syrusakbary: @theduke

lib/cli/src/commands/app/create.rs
```diff
@@ -74,6 +90,320 @@ pub struct CmdAppCreate {
...
     #[clap(long, short = 'p')]
     pub package: Option<String>,
+
+    /// Whether or not to search (and use) a local manifest.
+    #[clap(long)]
+    pub use_local_manifest: bool,
+
+    /// Name to use when creating a new package from a template.
+    #[clap(long)]
+    pub new_package_name: Option<String>,
```
syrusakbary, line 100: This should be `--template-package-name` (is longer, but more exhaustive than `new`

lib/cli/src/commands/publish.rs
```diff
@@ -44,17 +52,134 @@ pub struct Publish {
...
     #[clap(long, default_value = "2m")]
     pub timeout: humantime::Duration,
+
+    /// Whether or not the patch field of the version of the package - if any - should be bumped.
+    #[clap(long)]
+    pub autobump: bool,
+
+    /// Do not prompt for user input.
+    #[clap(long)]
+    pub non_interactive: bool,
```
syrusakbary, line 62: This should fall back to a default value

lib/backend-api/src/query.rs
theduke: Ignore the `auth token` part, that's already on the client of course, I thought this was a standalone method.

lib/wasix/Cargo.toml
syrusakbary: @maminrayej

lib/wasix/Cargo.toml
```diff
@@ -102,13 +104,13 @@ web-sys = { version = "0.3.64", features = [
 [target.'cfg(not(target_arch = "riscv64"))'.dependencies.reqwest]
 version = "0.11"
 default-features = false
-features = ["rustls-tls", "json", "stream"]
+features = ["rustls-tls", "json", "stream", "blocking"]
```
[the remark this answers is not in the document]
syrusakbary: @maminrayej

lib/cli/src/commands/app/create.rs
syrusakbary: @xdoardo

lib/cli/src/commands/app/create.rs
syrusakbary: @xdoardo

lib/cli/src/commands/app/create.rs
syrusakbary: Let's delete this `new_package_name` and use `package` instead (this flow will only be triggered if `--template` is provided or the user specifies the template

syrusakbary (commented):
@maminrayej will fix the linting and request flags, and package tests
@xdoardo will fix the cli tests

lib/wasix/src/bin_factory/binary_package.rs
theduke: followup issue: https://github.com/wasmerio/wasmer/issues/4597

lib/backend-api/src/query.rs
theduke: Tracked in https://github.com/wasmerio/wasmer/issues/4598

lib/cli/src/commands/app/create.rs
xdoardo: https://github.com/wasmerio/wasmer/issues/4598

lib/cli/src/commands/publish.rs
xdoardo: [62f0c2c](https://github.com/wasmerio/wasmer/pull/4532/commits/62f0c2c7bfa19edcd7980324e52643ed78c6513a)

syrusakbary (approved):

PR #4537, opened 2024-04-03
Fix owner issues with app create

ayys:
the owner field was being set to the same namespace as the package. This means
if the package is `wasmer/hello` then the owner becomes `wasmer`.

This PR also adds the owner field to app.yaml generated by `wasmer app create`


syrusakbary (approved):

theduke (approved):
@ayys : need to update the tests with the new `owner` field.

PR #4546, opened 2024-04-08
Expose `store::StoreObjects`

Twey:
# Description

Currently `store::AsStoreMut` is exposed, including its function `objects_mut`.  This function returns a `StoreObjects` type that is not exported.  If we're native, it can be imported from `wasmer_vm`, but under `js` `StoreObjects` is not visible and therefore the trait can't be implemented.

This PR exports `StoreObjects` so that JS and native code alike can implement `AsStoreMut`.


syrusakbary (approved):

syrusakbary:
Good catch. Merging

PR #4549, opened 2024-04-09
feat: (Partial) unnamed package support

theduke:
- **feat(backend-api): Add get_package_release query and types**
- **feat: Partial support for unnamed packages**


lib/cli/src/commands/package/download.rs
```diff
@@ -93,18 +93,48 @@ impl PackageDownload {
...
+                .with_context(|| {
+                    format!(
+                "could not retrieve package information for package '{}' from registry '{}'",
+                full_name, endpoint,
+            )
+                })?;
+
+                let download_url = package
+                    .pirita_url
+                    .context("registry does provide a container download container download URL")?;
```
xdoardo, line 118: typo: `"registry '{}' does not provide a container download URL", endpoint`, maybe?

lib/wasix/src/runtime/resolver/outputs.rs
```diff
@@ -27,20 +27,55 @@ pub struct ItemLocation {
...
-        let PackageId {
-            package_name,
+        write!(f, "{}@{}", self.name, self.version)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum PackageId {
+    Named(PackageIdent),
+    HashSha256(String),
```
xdoardo, line 45: nit: I think that it'd be good if we stuck to one name for the same thing; if we agree, we should either change this to `Sha256Hash` or change the name in `edge-schema` to match this one.

Cargo.toml
```diff
@@ -86,13 +86,14 @@ version = "4.2.8"
 [workspace.dependencies]
 enumset = "1.1.0"
 memoffset = "0.9.0"
-wasmer-toml = "0.9.2"
+wasmer-toml = "0.10.0"
```
syrusakbary, line 89: We need to delete this dependency

theduke:
SUbsumed by #4532

PR #4552, opened 2024-04-11
Module cache optimization round2

john-sharratt:
Improvements as follows:
- Fixed the mmap accelerated module load which was not using mmap
- Removed a needless copy of the module code which the module is archived (i.e. it was not fully zero copy)
- Various other minor tweaks

### DIFF
--- a/lib/compiler/src/artifact_builders/artifact_builder.rs
+++ b/lib/compiler/src/artifact_builders/artifact_builder.rs
@@ -126,7 +126,7 @@ impl ArtifactBuild {
         };
         let serializable = SerializableModule {
             compilation: serializable_compilation,
-            compile_info,
+            compile_info: compile_info,
             data_initializers,
             cpu_features: cpu_features.as_u64(),
         };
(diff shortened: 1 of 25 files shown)


syrusakbary (approved):

lib/compiler/src/artifact_builders/artifact_builder.rs
```diff
@@ -134,19 +135,18 @@ impl ArtifactBuild {
...
-        _data: &[u8],
-        _target: &Target,
-        _memory_styles: PrimaryMap<MemoryIndex, MemoryStyle>,
-        _table_styles: PrimaryMap<TableIndex, TableStyle>,
-    ) -> Result<Self, CompileError> {
-        Err(CompileError::Codegen(
-            "Compilation is not enabled in the engine".to_string(),
-        ))
-    }
+    // #[cfg(not(feature = "compiler"))]
```
theduke, line 138: If we don't need this anymore, we should just remove it, not comment it out?

syrusakbary: Good call, fixing

syrusakbary:
Merging manually as tests are blocked by one issue on the backend, and not really related to the Wasmer runtime

Issue #4554, opened 2024-04-11
`fd_advise` on invalid file descriptors report ok


yagehu:
### Describe the bug

Calling `fd_advise` on a clearly invalid file descriptor returns 0 in all conditions.

Minimal Rust repro:

```rust
use wasi::ADVICE_WILLNEED;

fn main() {
    unsafe {
        let result = wasi::fd_advise(6, 0, 0, ADVICE_WILLNEED);

        match result {
            Ok(()) => eprintln!("fd_advise ok"),
            Err(errno) => eprintln!("fd_advise failed with {errno}"),
        }
    }
}

```

```sh
$ cargo build --target wasm32-wasi
```

```sh
$ wasmer -vV; rustc -vV
wasmer 4.2.5 (0460050 2024-03-05)
binary: wasmer-cli
commit-hash: 04600507844bae978239531c8e6265d410aefe6d
commit-date: 2024-03-05
host: x86_64-unknown-linux-gnu
compiler: singlepass,cranelift
rustc 1.77.1 (7cf61ebde 2024-03-27)
binary: rustc
commit-hash: 7cf61ebde7b22796c69757901dd346d0fe70bd97
commit-date: 2024-03-27
host: x86_64-unknown-linux-gnu
release: 1.77.1
LLVM version: 17.0.6
```

### Steps to reproduce
Run the above Rust snippet compiled to `wasm32-wasi`. It should output `fd_advise failed with <errno>`, but it reports `fd_advise ok`.

### Expected behavior
When the fd is invalid, like a random or very large number, `fd_advise` should report something like `ebadf`.

### Actual behavior

`fd_advise` returns ok (0) in all cases.

### Additional context
This was uncovered in differential fuzzing between Wasmer and other Wasm/WASI runtimes. I have a simple PR to fix this behavior.

PR #4555, opened 2024-04-11
Fd validity and basic bound checks on `fd_advise`

yagehu:
This commit fixes `fd_advise` unconditionally returning ok (0) on all inputs.  The details are discussed in issue #4554.

# Description
#4554 discusses the motivation. This is a behavior divergence between Wasmer and all other Wasm runtimes.

fixes [#4554][1]

[1]: https://github.com/wasmerio/wasmer/issues/4554

### DIFF
--- a/lib/wasix/src/syscalls/wasi/fd_advise.rs
+++ b/lib/wasix/src/syscalls/wasi/fd_advise.rs
@@ -41,7 +41,19 @@ pub(crate) fn fd_advise_internal(
     len: Filesize,
     advice: Advice,
 ) -> Result<(), Errno> {
-    // this is used for our own benefit, so just returning success is a valid
-    // implementation for now
+    // Instead of unconditionally returning OK.  This barebones implementation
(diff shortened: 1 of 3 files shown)


theduke:
@yagehu once again, many thanks!

I pushed a commit that adds a test validating the behaviour.

The invalid file descriptor error is as expected, but the overflow still returns a `0`, could you check why?
(if not we will follow up)

To run the tests:

* compile the release binary
* `cd tests/wasi-fyi; ./test.sh`

yagehu:
@theduke Thanks for demonstrating the fyi test! I was having a hard time finding the right test suite to add to.

Since [`filesize`](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-filesize-u64) is spec'd as `u64`, I adjusted the `extern "C"` signature accordingly and now it indeed errors as expected. Previously, `i64::MAX` added to itself would not overflow an `u64`.

theduke (approved):

Issue #4556, opened 2024-04-12
`fd_seek` with `whence::set` and a negative offset should error


yagehu:
### Describe the bug

`fd_seek` with [`whence`](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-whence-variant) set to `set` and a negative should error. It is the case on Linux host and other Wasm runtimes. However, Wasmer happily cast the negative offset to an unsigned offset and performs the seek.

```sh
$ wasmer -vV; rustc -vV
wasmer 4.2.8 (fe3c6d0 2024-04-11)
binary: wasmer-cli
commit-hash: fe3c6d0461f2ee646fd38d1d6480d7c3a30fec74
commit-date: 2024-04-11
host: x86_64-unknown-linux-gnu
compiler: singlepass,cranelift
rustc 1.77.1 (7cf61ebde 2024-03-27)
binary: rustc
commit-hash: 7cf61ebde7b22796c69757901dd346d0fe70bd97
commit-date: 2024-03-27
host: x86_64-unknown-linux-gnu
release: 1.77.1
LLVM version: 17.0.6
```

### Steps to reproduce

```rust
fn main() {
    unsafe {
        let result = wasi::fd_seek(4, -1, wasi::WHENCE_SET);

        match result {
            Ok(new_offset) => eprintln!("fd_seek ok new offset {new_offset}"),
            Err(errno) => eprintln!("fd_seek errno {errno}"),
        }
    }
}
```

Compile this snippet to `wasm32-wasi` and run it with wasmer.

### Expected behavior
The above snippet should print `fd_seek errno INVAL (error 28)`.

### Actual behavior
```
fd_seek ok new offset 18446744073709551615
```

### Additional context
I have a fix. Will submit a PR.

PR #4557, opened 2024-04-12
Safely handle offset in fd_seek

yagehu:
# Description

This PR fixes [#4556][1].  Passing in a negative offset along with `whence::set` causes wasmer to erroneously seek to a very large offset.

[1]: https://github.com/wasmerio/wasmer/issues/4556


theduke (approved):
Good catch, thanks!

PR #4617, opened 2024-04-25
4.3.0 alpha.1 leftover fixes

Arshia001:


lib/config/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "wasmer-config"
-version = "0.1.0"
+version = "0.1.1"
```
theduke, line 3: `wasmer-config` will need manual attention.
We need to observe semver.
We might have made some breaking changes mandating a `0.2.0`, can you check?

syrusakbary (changes requested):
`wasmer-config` should be a workspace dependency, not sure why we rolledback on that, but I don't think should be necessary

lib/config/Cargo.toml
theduke: Actually, I think we might have made 0 changes

Arshia001:
@syrusakbary we need to merge this at some point. Do you want to do it as is, or find a workaround for the crate version thing and include that in this PR?

PR #4618, opened 2024-04-25
propagate timestamps to virtual file handles

maminrayej:
This PR propagates the timestamps to the virtual file whenever `fd_filestat_set_times` is called.

### DIFF
--- a/lib/virtual-fs/src/lib.rs
+++ b/lib/virtual-fs/src/lib.rs
@@ -336,6 +336,12 @@ pub trait VirtualFile:
     /// the time at which the file was created in nanoseconds as a UNIX timestamp
     fn created_time(&self) -> u64;
 
+    /// sets accessed time
+    fn set_accessed(&mut self, _atime: u64) {}
+
+    /// sets modification time
+    fn set_modified(&mut self, _mtime: u64) {}
+
     /// the size of the file in bytes
     fn size(&self) -> u64;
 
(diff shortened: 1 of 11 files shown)


lib/virtual-fs/src/mem_fs/file.rs
```diff
@@ -144,6 +144,30 @@ impl VirtualFile for FileHandle {
...
+            Ok(fs) => fs,
+            _ => return,
+        };
+
+        let inode = fs.storage.get_mut(self.inode);
+        if let Some(node) = inode {
+            node.metadata_mut().modified = mtime;
+        }
+    }
+
```
syrusakbary, line 170: We should unite this two in one function, since the systemcall has two functions at the beginning.
Please see: https://github.com/bytecodealliance/cap-std/blob/8f2f60f48f39e2b1606ae2961a5fe935dce75ab9/cap-primitives/src/rustix/freebsd/fs/set_times_impl.rs#L9

maminrayej: I agree.

lib/virtual-fs/src/lib.rs
```diff
@@ -336,6 +336,9 @@ pub trait VirtualFile:
     /// the time at which the file was created in nanoseconds as a UNIX timestamp
     fn created_time(&self) -> u64;
 
+    /// sets accessed and modified time
+    fn set_times(&mut self, _atime: Option<u64>, _mtime: Option<u64>) {}
```
syrusakbary, line 340: Note: try to implement this for all the other implementations of VirtualFS.

We also need it for the `Dir`, and `Filesystem`

syrusakbary (approved):

Issue #4479, opened 2024-03-08
Login: error when HTTPS_PROXY is set


re-xmyl:
### Summary
i couldn't to login 

### Additional details
wasmer login
error: failed to setup proxy for reqwest Client
╰─▶ 1: Could not connect to proxy: builder error: unknown proxy scheme

$ echo $https_proxy
https://127.0.0.1:7890/
$ echo $http_proxy
http://127.0.0.1:7890/

the proxy actually works good.
how to  fix my question?

theduke:
HTTPS proxies should work just fine in theory, we will need to look into the cause of the issue.

PR #4624, opened 2024-04-28
respect proxy env vars

maminrayej:
This PR adds support for downloading packages through a proxy server provided by the `http_proxy`  or `HTTP_PROXY` env var.


syrusakbary (approved):

theduke:
I think there was a related issue.

theduke:
Here: https://github.com/wasmerio/wasmer/issues/4479

But I don't think that issue is fixed by this PR.

PR #4627, opened 2024-04-29
reuse atom signatures

maminrayej:
This PR tries to reuse the atom signature (sha256 hash of the atom) in the manifest to prevent recomputing the hash for that atom.
This will be done in two steps:
1. Add support for two hashing algorithms in `ModuleHash`, namely `xxhash` and `sha256`.
2. Use `sha256` for hashes of atoms in order to be able to reuse their signature.

### DIFF
--- a/lib/cli/src/commands/run/mod.rs
+++ b/lib/cli/src/commands/run/mod.rs
@@ -705,9 +705,10 @@ impl ExecutableTarget {
                     .load_module_sync(&wasm)
                     .with_context(|| format!("Unable to compile \"{}\"", path.display()))?;
 
+                // FIXME: Amin
                 Ok(ExecutableTarget::WebAssembly {
                     module,
-                    module_hash: ModuleHash::hash(&wasm),
+                    module_hash: ModuleHash::xxhash(&wasm),
                     path: path.to_path_buf(),
                 })
             }
@@ -717,7 +718,8 @@ impl ExecutableTarget {
                 let module = unsafe { Module::deserialize_from_file(&engine, path)? };
                 let module_hash = {
                     let wasm = std::fs::read(path)?;
-                    ModuleHash::hash(wasm)
+                    // FIXME: Amin
+                    ModuleHash::xxhash(wasm)
                 };
 
                 Ok(ExecutableTarget::WebAssembly {
--- a/lib/cli/src/commands/run/mod.rs
+++ b/lib/cli/src/commands/run/mod.rs
@@ -705,7 +705,6 @@ impl ExecutableTarget {
                     .load_module_sync(&wasm)
                     .with_context(|| format!("Unable to compile \"{}\"", path.display()))?;
 
-                // FIXME: Amin
                 Ok(ExecutableTarget::WebAssembly {
                     module,
                     module_hash: ModuleHash::xxhash(&wasm),
@@ -718,7 +717,6 @@ impl ExecutableTarget {
                 let module = unsafe { Module::deserialize_from_file(&engine, path)? };
                 let module_hash = {
                     let wasm = std::fs::read(path)?;
-                    // FIXME: Amin
                     ModuleHash::xxhash(wasm)
                 };
 
--- a/lib/wasix/src/runtime/module_cache/types.rs
+++ b/lib/wasix/src/runtime/module_cache/types.rs
@@ -138,58 +138,42 @@ pub enum ModuleHash {
 impl ModuleHash {
     /// Create a new [`ModuleHash`] from the raw xxhash hash.
     pub fn xxhash_from_bytes(key: [u8; 8]) -> Self {
-        Self::from_bytes(key)
+        Self::XXHash(key)
     }
 
     /// Create a new [`ModuleHash`] from the raw sha256 hash.
     pub fn sha256_from_bytes(key: [u8; 32]) -> Self {
-        Self::from_bytes(key)
-    }
-
-    fn from_bytes<const N: usize>(key: [u8; N]) -> Self {
-        if N == 8 {
-            let key = key.as_slice().try_into().unwrap();
-            ModuleHash::XXHash(key)
-        } else if N == 32 {
-            let key = key.as_slice().try_into().unwrap();
-            ModuleHash::Sha256(key)
-        } else {
-            panic!("Only keys with size 8 or 32 are accepted")
-        }
+        Self::Sha256(key)
     }
 
     /// Creates a random xxhash for the module
     pub fn xxhash_random() -> Self {
-        Self::random::<8>()
+        let mut rand = rand::thread_rng();
+        let mut key = [0u8; 8];
+        rand.fill_bytes(&mut key);
+        Self::xxhash_from_bytes(key)
     }
 
     /// Creates a random sha256 hash for the module
     pub fn sha256_random() -> Self {
-        Self::random::<32>()
-    }
-
-    // Creates a random hash for the module
-    fn random<const N: usize>() -> Self {
         let mut rand = rand::thread_rng();
-        let mut key = [0u8; N];
+        let mut key = [0u8; 32];
         rand.fill_bytes(&mut key);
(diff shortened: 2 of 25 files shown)


lib/wasix/src/runtime/module_cache/types.rs
```diff
@@ -127,50 +128,122 @@ impl CacheError {
...
+            let key = key.as_slice().try_into().unwrap();
+            ModuleHash::Sha256(key)
+        } else {
+            panic!("Only keys with size 8 or 32 are accepted")
+        }
+    }
+
+    /// Creates a random xxhash for the module
+    pub fn xxhash_random() -> Self {
+        Self::random::<8>()
```
syrusakbary, line 163: This is not a proper implementation. This should return `Self::XXHash(...)`.

lib/wasix/src/runtime/module_cache/types.rs
```diff
@@ -127,50 +128,122 @@ impl CacheError {
...
+    }
+
+    /// Creates a random sha256 hash for the module
+    pub fn sha256_random() -> Self {
+        Self::random::<32>()
```
syrusakbary, line 168: This is not a proper implementation. This should return `Self::SHA256(...)`.

lib/wasix/src/runtime/module_cache/types.rs
```diff
@@ -127,50 +128,122 @@ impl CacheError {
...
+        if N == 8 {
+            let key = key.as_slice().try_into().unwrap();
+            ModuleHash::XXHash(key)
+        } else if N == 32 {
+            let key = key.as_slice().try_into().unwrap();
+            ModuleHash::Sha256(key)
+        } else {
+            panic!("Only keys with size 8 or 32 are accepted")
+        }
+    }
```
syrusakbary, line 159: This function is not proper. We send a different type depending on the length, which is highly confusion and can lead to many errors

lib/wasix/src/runtime/module_cache/types.rs
```diff
@@ -127,50 +128,122 @@ impl CacheError {
...
+                }
+            }
+            ModuleHash::Sha256(bytes) => {
+                if N == 32 {
+                    bytes.as_slice().try_into().unwrap()
+                } else {
+                    panic!("Requested for {N} bytes, but Sha256 is 32 bytes")
+                }
+            }
+        }
```
syrusakbary, line 230: Why we need to expose this?

lib/wasix/src/runtime/module_cache/types.rs
```diff
@@ -127,50 +128,122 @@ impl CacheError {
...
+    }
+
+    /// Get the raw hash.
+    pub fn as_bytes<const N: usize>(self) -> [u8; N] {
+        match self {
+            ModuleHash::XXHash(bytes) => {
+                if N == 8 {
+                    bytes.as_slice().try_into().unwrap()
+                } else {
+                    panic!("Requested for {N} bytes, byt XXHash is 8 bytes");
```
syrusakbary, line 220: I'm not a fan of panics here...

lib/wasix/src/syscalls/journal/mod.rs
```diff
@@ -31,9 +31,9 @@ pub struct JournalSyscallPlayer<'a, 'c> {
     pub ctx: FunctionEnvMut<'c, WasiEnv>,
     pub bootstrapping: bool,
 
-    pub journal_module_hash: Option<[u8; 8]>,
+    pub journal_module_hash: Option<Vec<u8>>,
```
syrusakbary, line 34: We can probably use a Boxed slice instead, as we'll need no conversion? (`Box<[u8]>`)

lib/wasix/src/syscalls/journal/mod.rs
```diff
@@ -31,9 +31,9 @@ pub struct JournalSyscallPlayer<'a, 'c> {
...
     pub rewind: Option<RewindState>,
-    pub cur_module_hash: [u8; 8],
+    pub cur_module_hash: Vec<u8>,
```
syrusakbary, line 36: We can probably use a Boxed slice instead, as we'll need no conversion? (`Box<[u8]>`)

lib/wasix/src/runtime/package_loader/load_package_tree.rs
```diff
@@ -213,13 +217,32 @@ fn legacy_atom_hack(
...
+        .expect("malformed atom signature: does not have sha256 prefix");
+    let sha256 = base64::prelude::BASE64_STANDARD
+        .decode(base64_encoded)
+        .expect("malformed base64 encoded hash");
+    let hash: [u8; 32] = sha256
+        .as_slice()
+        .try_into()
+        .expect("sha256 hash is not 32 bytes");
+
+    ModuleHash::sha256_from_bytes(hash)
```
syrusakbary, line 243: I think this is quite hacky. We should just have a `FromStr` / `IntoStr` impl for the `ModuleHash`, that simply can do the `sha256:abc` -> `ModuleHash::SHA256(..)` conversion and back.

maminrayej: There are no hacks here. I'm just doing the steps that were taken in `pirita`, but in reverse. But I agree that your way may be more _convenient_.

lib/wasix/src/syscalls/journal/mod.rs
maminrayej: My main concern was introducing an allocation here which both `Vec` and `Box` do unfortunately. `Vec` already derefs to a slice, so I don't think here is any advantage in using `Box<[u8]>`.

lib/wasix/src/runtime/package_loader/load_package_tree.rs
syrusakbary: Let's do what's more convenient, is better and we can reuse it in more places. Maybe we can even add it to wasmer-config. In any case, let's do it proper

lib/wasix/src/syscalls/journal/mod.rs
syrusakbary: Size will not be changed, so `Box<[u8]>` would probably be more optimal. I'm not religious about this, but as `Box<[u8]>` has a fixed slice, less memory will be allocated than for the vector.

This is a nit pick in any case

lib/wasix/src/runtime/package_loader/load_package_tree.rs
theduke: This should not panick, but return an error instead...

theduke:
Note: we'll want a validation function that validates all these invariants to be correct, and run that validation on webc upload in the future.

lib/wasix/src/syscalls/journal/mod.rs
maminrayej: Ah you're right. I didn't consider the size.

syrusakbary (approved):

lib/cli/src/commands/run/mod.rs
```diff
@@ -717,7 +717,7 @@ impl ExecutableTarget {
                 let module = unsafe { Module::deserialize_from_file(&engine, path)? };
                 let module_hash = {
                     let wasm = std::fs::read(path)?;
-                    ModuleHash::hash(wasm)
+                    ModuleHash::xxhash(wasm)
```
syrusakbary, line 720: I just realized that this is completely wrong, `let wasm = std::fs::read(path)?` is not a wasm file, but a precompiled artifact.
Thus, we shall not use that as the hash in any way.

syrusakbary: Issue: https://github.com/wasmerio/wasmer/issues/4632

maminrayej: Since we decided to track this bug in a separate issue and solve it later on, I'll resolve this comment.

syrusakbary (approved):

PR #4634, opened 2024-05-01
Remove `Send` bound from `FunctionEnv`

Twey:
# Description

Requiring that function environments be `Send` is quite onerous on the Web, in which many objects (including the Wasmer `Instance` itself, should we wish to pass around a reference to it) are not `Send`.  The function environment is not passed between threads in Wasmer, and in typed usage the user never sees the trait object, only concrete instances of it after casting — which means that all user-visible types will remain `Send` if the user should wish to pass them around.

### DIFF
--- a/lib/api/src/externals/function.rs
+++ b/lib/api/src/externals/function.rs
@@ -140,7 +140,7 @@ impl Function {
     ///     Ok(vec![Value::I32(sum)])
     /// });
     /// ```
-    pub fn new_with_env<FT, F, T: Send + 'static>(
+    pub fn new_with_env<FT, F, T: 'static>(
         store: &mut impl AsStoreMut,
         env: &FunctionEnv<T>,
         ty: FT,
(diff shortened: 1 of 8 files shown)


Twey:
Hi @syrusakbary, the system automatically assigned you to review this — but if somebody else should do it please let me know whom to request :smiley:

syrusakbary:
Hey @Twey , we want to set a sound type system for the Store, in a way that if the `Store` is `Send+Sync` then the functions have to be as well. But if is not... then they don't need to.

But we need to define a bit better this, that's why the review on this PR is a bit halted

Twey:
Hey @syrusakbary, thanks for getting back to me!  The easiest way I can think of to do that would be to remove the `Box<dyn Any>` from the `VMFunctionEnvironment` struct and make it a parameter instead (which could be instantiated to `Box<dyn Any>` or `Box<dyn Any + Send>` as the user likes).  That way both the store and the function environment will inherit their `Send`ness from the parameter.  Do you think that's worth me making a PR for?  Or is there a reason we can't have type parameters on `VMFunctionEnvironment`?

Twey:
Closed in favour of #5660.

PR #4638, opened 2024-05-06
Handle local package resolution

maminrayej:
Right now, the way for handling a local package is the same as a local webc file. This will cause the code to rely on webc hash as a unique identifier in the resolution graph. But the cost of computing the hash for a local package is much higher than a local webc file since for the package, we must serialize the package to get the hash.

This PR adds a new code path for running a local package and uses the hash of the package path as its unique identifier in the dependency graph.

### DIFF
--- a/lib/cli/src/commands/package/build.rs
+++ b/lib/cli/src/commands/package/build.rs
@@ -3,6 +3,7 @@ use std::path::PathBuf;
 use anyhow::Context;
 use dialoguer::console::{style, Emoji};
 use indicatif::ProgressBar;
+use sha2::Digest;
 use wasmer_config::package::PackageHash;
 
 use crate::utils::load_package_manifest;
@@ -53,7 +54,10 @@ impl PackageBuild {
             )
         };
         let pkg = webc::wasmer_package::Package::from_manifest(manifest_path)?;
-        let pkg_hash = PackageHash::from_sha256_bytes(pkg.webc_hash());
+        let data = pkg.serialize()?;
+        let hash = sha2::Sha256::digest(&data).into();
+
+        let pkg_hash = PackageHash::from_sha256_bytes(hash);
         let name = if let Some(manifest_pkg) = manifest.package {
             if let Some(name) = manifest_pkg.name {
                 if let Some(version) = manifest_pkg.version {
@@ -115,8 +119,6 @@ impl PackageBuild {
                 out_path.display()
             );
         }
-
-        let data = pkg.serialize()?;
 
         pb.println(format!(
             "{} {}Writing package...",
--- a/lib/wasix/src/bin_factory/binary_package.rs
+++ b/lib/wasix/src/bin_factory/binary_package.rs
@@ -1,4 +1,4 @@
-use std::{os::unix::ffi::OsStrExt, path::Path, sync::Arc};
+use std::{path::Path, sync::Arc};
 
 use anyhow::Context;
 use derivative::*;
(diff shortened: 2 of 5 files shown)


lib/wasix/src/bin_factory/binary_package.rs
```diff
@@ -90,7 +121,9 @@ impl BinaryPackage {
 
         let manifest = container.manifest();
         let id = PackageInfo::package_id_from_manifest(manifest)?.unwrap_or_else(|| {
-            PackageId::Hash(PackageHash::from_sha256_bytes(container.webc_hash()))
+            PackageId::Hash(PackageHash::from_sha256_bytes(
+                container.webc_hash().unwrap(),
```
syrusakbary, line 125: I'd use `container.webc_hash().unwrap_or(|| container.path_hash())` ?

maminrayej: calling `from_webc` with a `Package`, is a bug (or not intended usage) from this PR forward since the hash won't be present. Although your suggestion works, I prefer the dedicated `from_dir` approach since it clearly signals the difference between a local package and a local webc.

To not panic, I changed the code to return an error upon the absence of the webc hash.

syrusakbary (approved):
Looks good for me, but I think since is touching Binary package will be better for @theduke to do the final approval

lib/cli/src/commands/package/build.rs
```diff
@@ -53,7 +54,10 @@ impl PackageBuild {
             )
         };
         let pkg = webc::wasmer_package::Package::from_manifest(manifest_path)?;
-        let pkg_hash = PackageHash::from_sha256_bytes(pkg.webc_hash());
+        let data = pkg.serialize()?;
+        let hash = sha2::Sha256::digest(&data).into();
+
+        let pkg_hash = PackageHash::from_sha256_bytes(hash);
```
theduke, line 60: Won't this differ from the embedded webc hash in the file (because it doesn't ignore the hash in the file)?

I see the hash is only returned, not used, but could this cause issues for other callers?
Like the new deployment flow?

lib/wasix/src/bin_factory/binary_package.rs
```diff
@@ -79,6 +80,36 @@ pub struct BinaryPackage {
...
+    #[tracing::instrument(level = "debug", skip_all)]
+    pub async fn from_dir(
+        dir: &Path,
+        rt: &(dyn Runtime + Send + Sync),
+    ) -> Result<Self, anyhow::Error> {
+        let source = rt.source();
+
+        // since each package must be in its own directory, hash of the `dir` should provide a good enough
+        // unique identifier for the package
+        let hash = sha2::Sha256::digest(dir.display().to_string().as_bytes()).into();
```
theduke, line 92: Minor, but to prevent lossiness from `.display()` we could also use:  `dir.as_os_str().as_encoded_bytes()`.

theduke (approved):

lib/wasix/src/bin_factory/binary_package.rs
maminrayej: I originally did this, but aren't we on `1.73`? `as_encoded_bytes` was stabilized on `1.74`.

lib/cli/src/commands/package/build.rs
maminrayej: The hash returned by `v2::OwnedReader` and `v3::OwnedReader`, implementors of `AbstractWebc`, is just a `digest` call on the underlying buffer equivalent to `data` here. So they should be the same.

PR #4640, opened 2024-05-06
add proxy support to the `package download` subcommand

maminrayej:
In continuation with #4624, this PR adds proxy support to the `wasmer package download` subcommand.

### DIFF
--- a/lib/cli/src/commands/package/download.rs
+++ b/lib/cli/src/commands/package/download.rs
@@ -138,7 +138,17 @@ impl PackageDownload {
             PackageSource::Url(url) => bail!("cannot download a package from a URL: '{}'", url),
         };
 
-        let client = reqwest::blocking::Client::new();
+        let client = {
+            let mut builder = reqwest::blocking::ClientBuilder::new();
+
+            let proxy = std::env::var("http_proxy").or_else(|_| std::env::var("HTTP_PROXY"));
+            if let Ok(scheme) = proxy {
+                builder = builder.proxy(reqwest::Proxy::all(scheme)?);
+            }
+
+            builder.build().context("Could not create reqwest client")?
+        };
+
         let mut b = client
             .get(download_url)
             .header(http::header::ACCEPT, "application/webc");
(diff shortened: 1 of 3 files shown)


lib/cli/src/commands/package/download.rs
```diff
@@ -138,7 +138,17 @@ impl PackageDownload {
...
+        let client = {
+            let mut builder = reqwest::blocking::ClientBuilder::new();
+
+            let proxy = std::env::var("http_proxy").or_else(|_| std::env::var("HTTP_PROXY"));
+            if let Ok(scheme) = proxy {
+                builder = builder.proxy(reqwest::Proxy::all(scheme)?);
+            }
+
+            builder.build().context("Could not create reqwest client")?
+        };
```
syrusakbary, line 150: We need to abstract this into a common function that is reused

syrusakbary (changes requested):
Address comment

ayys:
Potential future changes: we will probably need to do the same for wasmer run (I think that uses a different code path to download the package).
Also, the code to upload file to google storage would also break I suppose? so we will need to add proxy there as well. (cc: @xdoardo since he is working on the publish stuff)

maminrayej:
> Potential future changes: we will probably need to do the same for wasmer run (I think that uses a different code path to download the package).

The code path for `wasmer run` was covered in #4624 .

> Also, the code to upload file to google storage would also break I suppose? so we will need to add proxy there as well

That's a good point. I can check that code path as well.

lib/wasix/src/http/reqwest.rs
```diff
@@ -98,3 +92,25 @@ impl super::HttpClient for ReqwestHttpClient {
         Box::pin(f)
     }
 }
+
```
syrusakbary, line 95: We can have a function.

```rust
fn get_proxy(): Result<reqwest::Proxy> {
    let proxy = env::var("http_proxy").or_else(|_| env::var("HTTP_PROXY")).map(reqwest::Proxy::all(scheme))
}
```

syrusakbary (approved):
Approved once comment is addressed

PR #4645, opened 2024-05-08
sync modified timestamp upon creating the inode

maminrayej:


syrusakbary (approved):

Arshia001 (changes requested):
Did we also handle the zero timestamps for creation and access time? I believe we should set those to be equal to the modification time if they're not available, since returning zero is still a bad idea.

Arshia001 (approved):
~Did we also handle the zero timestamps for creation and access time? I believe we should set those to be equal to the modification time if they're not available, since returning zero is still a bad idea.~

Let's do this in another PR.

PR #4653, opened 2024-05-08
Revert `wasmer-wasix` dependency to per project instead of workspace

maminrayej:
Using `wasmer-wasix` as a workspace dependency causes trouble for the release script. So for now, this PR reverts the dependency back to a per project style.


syrusakbary:
We can merge this, if we create another issue to improve the release script and start using workspace dependencies for this crate (and other one we had before as well)

syrusakbary (approved):
Approved only with the condition of linking here the new issue (for refactoring build script and use workspace dependencies) prior to merging

PR #4659, opened 2024-05-09
Second solution to the rewind buffer issue: allocate memory for it using memory.grow

Arshia001:


Arshia001:
Abandoned in favor of #4658

PR #4667, opened 2024-05-10
Hacky solution to renaming files on a non-root FS, as long as both pa…

Arshia001:
…ths are within the same FS


syrusakbary (commented):
Missing virtual-fs unit tests for this

Arshia001:
@syrusakbary I've modified the wasi-fyi rename test to verify this change completely. Do we also need a unit test? I'm not very familiar with the virtual-fs code, so I don't know how to set up such a test right now.

Arshia001:
Closing this in favor of #4702

PR #4679, opened 2024-05-14
[bugfix] correct error message when app version is not found

### LINKED ISSUE #4647: wasmer app logs: Error "unknown package version"
When invoking `wasmer app logs wasmer/wasix-docs` , we get a strange error :

```
❮ wasmer app logs wasmer/wasix-docs
error: unknown package version
```

ayys:
This PR corrects the error message reported when the app version is not found on registry.

### DIFF
--- a/lib/backend-api/src/query.rs
+++ b/lib/backend-api/src/query.rs
@@ -884,7 +884,7 @@ fn get_app_logs(
                     .run_graphql(types::GetDeployAppLogs::build(variables.clone()))
                     .await?
                     .get_deploy_app_version
-                    .context("unknown package version")?;
+                    .context("app version not found")?;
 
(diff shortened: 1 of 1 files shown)


theduke:
This fixes the bad error message, but does not close https://github.com/wasmerio/wasmer/issues/4647 .

Getting logs with an app id should still work.

The issue is that the command uses `Identifier`, while most of the other commands use `AppIdentOpts`, which supports finding the app from an `app.yaml` in the local directory, ns/name, and app id.
Problem is that it doesn't support a specific app version, which `Identifier` supports.

But all of these variations should work.

theduke:
Since the query takes an owner + namespace, the CLI can detect that it's an app id, and do a query to resolve the owner/name itself.

theduke (approved):

PR #4689, opened 2024-05-14
redirect to child node in a different fs

### LINKED ISSUE #3575: remove_dir_all is broken with concurrent modification
### Describe the bug

`fd_readdir` [skips](https://github.com/wasmerio/wasmer/blob/0ebc872f062d6094261ff2c569dfb844e3824ab2/lib/wasi/src/syscalls/mod.rs#L1360) `cookie` directory entries, but cookie is just an index into the sorted list of files in the directory. If that list changes between two `fd_readdir` calls, `cookie` won't refer to the intended continuation point,

```sh
echo (wasmer -V) \| (rustc -V) \| (uname -m)
wasmer 3.1.0 | rustc 1.67.0 (fc594f156 2023-01-24) (Arch Linux rust 1:1.67.0-2) | x86_64
```

### Steps to reproduce
Compile the following rust program:
```rust
fn main() {
    std::fs::remove_dir_all("foo/").expect("Foo gone");
}
```
Execute:
```sh
cargo build --target wasm32-wasi
mkdir foo/; touch foo/{bar,blubb,gah}
wasmer run target/wasm32-wasi/debug/remove\_dir\_all.wasm --mapdir .:.

### Actual behavior
```
thread 'main' panicked at 'Foo gone: Os { code: 55, kind: Uncategorized, message: "Directory not empty" }', src/main.rs:8:37
```
and the file `foo/gah` will still remain (which file remains may depend on your fs - btrfs here).

### Expected behavior
`foo/` removed successfully.


### Additional context
* The behavior differs with `--dir` and `--mapdir`, both are broken.
* wasmtime seems to have the exact same bug. Made me think it's a rust bug at first.

### PR DESCRIPTION
**Pull request by maminrayej**
This PR adds the missing logic in `mem_fs` for redirecting when resolving a child node that resides in a different fs.

Resolves #3575.

### DIFF
--- a/tests/wasi-fyi/fs_remove_dir_all.rs
+++ b/tests/wasi-fyi/fs_remove_dir_all.rs
@@ -2,5 +2,14 @@ use std::fs;
 
 fn main() {
     assert!(fs::create_dir_all("/fyi/foo/bar").is_ok());
+    assert!(fs::create_dir_all("/fyi/foo/baz").is_ok());
+    assert_eq!(
+        fs::read_dir("/fyi/foo")
+            .unwrap()
+            .map(|e| e.unwrap().file_name())
+            .collect::<Vec<_>>(),
+        vec!["bar", "baz"]
+    );
     assert!(fs::remove_dir_all("/fyi/foo").is_ok());
+    assert!(fs::read_dir("/fyi/foo").is_err());
 }
--- a/tests/wasi-fyi/fs_remove_dir_all.rs
+++ b/tests/wasi-fyi/fs_remove_dir_all.rs
@@ -1,0 +1,6 @@
+use std::fs;
+
+fn main() {
+    assert!(fs::create_dir_all("/fyi/foo/bar").is_ok());
+    assert!(fs::remove_dir_all("/fyi/foo").is_ok());
+}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/wasi-fyi/fs_remove_dir_all.rs [hunk 1]
```diff
@@ -0,0 +1,6 @@
+use std::fs;
+
+fn main() {
+    assert!(fs::create_dir_all("/fyi/foo/bar").is_ok());
```
syrusakbary, line 4: Can you assert the contents correspond to what we expect?

```rust
    assert!(fs::create_dir_all("/fyi/foo/bar").is_ok());
    assert!(fs::create_dir_all("/fyi/foo/xyz").is_ok());
    assert_eq!(fs::read_dir("/fyi/foo"), ["bar", "xyz"]); // DEMO CODE
```

syrusakbary (approved):

jcaesar:
@maminrayej This does not resolve #3575. (The test vector I provided originally does not throw an error message anymore and now needs to be invoked with `--mapdir foo:foo` instead of `--mapdir .:.`, but it still doesn't remove foo or even all files in it on the host.)

Whatever this PR changes isn't even related to the initial sentence of that bug (readdir cookies not being stable against directory modifications), so I have no idea why you thought this PR would fix that issue.

maminrayej:
@jcaesar it was most probably a mistake in typing the issue number. I'll look into it and re-open the issue if that is the case.

PR #4698, opened 2024-05-14
Fix calculation of stat.st_size in the presence of combined seeks and writes

Arshia001:


maminrayej (commented):
Looks reasonable to me. But for this PR to be more inline with other recent fixes to wasix and virtual-fs, I'd like to see a wasi-fyi test that fails before this PR and passes after, so we know this fixes the issue and prevents any future regressions of this sort.

Arshia001:
@maminrayej technically, `wasi-fyi` tests weren't created by us, so we should put the test somewhere else I believe.

syrusakbary:
@Arshia001 we can add them in wasi-fyi for now and open PRs later, that's ok!

Arshia001:
@syrusakbary @maminrayej I added the test, can you please re-review and approve?

maminrayej (approved):

PR #4700, opened 2024-05-15
`wasmer app logs`: support app IDs, improve output format

ayys:
Using the `AppIdentOpts` instead of `Identifier`, the `logs`
subcommand can now fetch get a `DeployApp` instance from an ID, or a
name or an alias.

Misc: this commit also cleans up the formatting a little bit.

### DIFF
--- a/lib/cli/src/commands/app/logs.rs
+++ b/lib/cli/src/commands/app/logs.rs
@@ -1,6 +1,7 @@
 //! Show logs for an Edge app.
 
-use comfy_table::Table;
+use colored::Colorize;
+use comfy_table::{Cell, Table};
 use edge_schema::pretty_duration::parse_timestamp_or_relative_time;
 use futures::StreamExt;
 use time::{format_description::well_known::Rfc3339, OffsetDateTime};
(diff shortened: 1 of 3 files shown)


lib/cli/src/commands/app/logs.rs
```diff
@@ -131,9 +118,9 @@ impl crate::commands::AsyncCliCommand for CmdAppLogs {
 
         let logs_stream = wasmer_api::query::get_app_logs_paginated(
             &client,
-            name.clone(),
-            owner.to_string(),
-            version.clone(),
+            app.name.clone(),
+            app.owner.global_name.to_string(),
+            None, // keep version None since we want logs from all versions atm
```
theduke, line 123: Can be done in a followup, but since yesterday there is a `AppIdent::AppVersion`, so we should respect that to  limit to a specific version in the future.

theduke (approved):

theduke:
Big improvement!

lib/cli/src/commands/app/logs.rs
ayys: I'll create a ticket for it, thanks!

ayys: https://wasmerio.slack.com/archives/CT5U4HW2C/p1715856712738749

ayys: https://github.com/wasmerio/wasmer/issues/4704

Issue #4701, opened 2024-05-15
cli(deploy): Add creation flags


xdoardo:
Embed part of the flags for `wasmer app create` in `wasmer deploy` as well. In particular, the flags needed are those regarding the creation of new apps when calling `wasmer deploy` without a pre-existing app: `--package`, `--template` and `--use-local-manifest`.

Issue #4674, opened 2024-05-13
Wasmer deploy flow, incorrect package


syrusakbary:
When being in a dir, with a local package `wasmer.toml`, and running `wasmer deploy` (no `app.yaml`).

```
$ wasmer deploy
It seems you are trying to create a new app!
✔ Who should own this app? · wasmer-examples
✔ What should be the name of the app? · angular-wasmer-starter
A package manifest was found in path /Users/syrusakbary/Development/wasmer-examples/angular-wasmer-starter/wasmer.toml.
✔ Use it for the app? · yes
? Do you want to deploy the app now? (y/n) ›
```

Then the `app.yaml` is created as:

```
kind: wasmer.io/App.v0
name: angular-wasmer-starter
owner: wasmer-examples
package: /Users/syrusakbary/Development/wasmer-examples/angular-wasmer-starter/wasmer.toml
```

Package should be `.`, not the long package route.

Issue #4605, opened 2024-04-25
wasmer app create: Fails with custom path


theduke:
Command:

```
wasmer app create -t static-website --non-interactive --no-wait --owner wasmer-tests --new-package-name test-app-cli-info --name test-app-cli-info --path /tmp/wasmer-tests/apps/test-app-cli-info --deploy-app
```

Error:

```
Loading local package (manifest path: /home/theduke/dev/github.com/wasmerio/wasmer-integration-tests/watest/wasmer.toml)
error: Could not read or find manifest in path '.'!
```

Issue #4578, opened 2024-04-16
cli: Use registry templates in `app create`


theduke:
This will need updates to both the initial generic flow, and to each of the various templates.

PR #4703, opened 2024-05-15
Use registry templates to create new app

xdoardo:
As per title, this PR removes the old templating mechanism and institutes a new one using the registry templates, solving #4578. 

Mandatory screenshot: 
image

Side quests: 
- Passes the owner information from `wasmer deploy` to `wasmer app create` (solves #4701)
- Correctly serializes the package information when creating from an existing manifest (solves #4674)
- Implicitly solves #4605, but introduces a potentially breaking change: `new-package-name` is not used anymore and, as of now, the `static-website` template is not available from the registry (with that name, at least).


lib/cli/src/utils/prompts.rs
```diff
@@ -23,38 +22,38 @@ pub fn prompt_for_ident(message: &str, default: Option<&str>) -> Result<String,
     }
 }
 
-/// Ask a user for a package name.
-///
-/// Will continue looping until the user provides a valid name.
-pub fn prompt_for_package_ident(
-    message: &str,
-    default: Option<&str>,
-) -> Result<NamedPackageIdent, anyhow::Error> {
```
syrusakbary, line 32: Why we are not prompting anymore for a package?

lib/cli/src/utils/mod.rs
```diff
@@ -111,79 +108,79 @@ pub fn load_package_manifest(
...
-                        break Ok((name, None));
-                    } else {
-                        eprintln!("Package '{name}' already exists");
-                    }
-                }
-            }
-        }
-    }
-}
+// /// Defines how to check for a package.
```
syrusakbary, line 135: Remove me if you are not going to use it

lib/cli/src/utils/mod.rs
```diff
@@ -111,79 +108,79 @@ pub fn load_package_manifest(
...
-
-/// Defines how to check for a package.
-pub enum PackageCheckMode {
-    /// The package must exist in the registry.
-    MustExist,
-    /// The package must NOT exist in the registry.
-    #[allow(dead_code)]
-    MustNotExist,
-}
+///// Ask a user for a package name.
```
syrusakbary, line 111: Remove me if not longer needed

theduke:
Do we have a static site template?
Should have that before releasing the the version.

syrusakbary (approved):

PR #4706, opened 2024-05-16
Multiple improvements for the journal load times

john-sharratt:
- Compressed memory is now passed around rather than decompressed.
- Removed a copy operation on the journal save and restore
- Added improved journal compaction which removes more events
- Added new memory style properties that allow a backing file to be supplied


theduke (approved):
LGTM

syrusakbary (approved):

PR #4708, opened 2024-05-16
Fix wasi-fyi tests

maminrayej:
This PR aims to fix the issues in the virtual-fs layer in order to pass all wasi-fyi tests.

### DIFF
--- a/lib/virtual-fs/src/arc_fs.rs
+++ b/lib/virtual-fs/src/arc_fs.rs
@@ -18,6 +18,10 @@ impl ArcFileSystem {
 }
 
 impl FileSystem for ArcFileSystem {
+    fn readlink(&self, path: &Path) -> Result<PathBuf> {
+        self.fs.readlink(path)
+    }
+
     fn read_dir(&self, path: &Path) -> Result<ReadDir> {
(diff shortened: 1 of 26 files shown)


lib/c-api/Cargo.toml
```diff
@@ -38,7 +38,7 @@ virtual-fs = { version = "0.11.4", path = "../virtual-fs", optional = true, defa
 enumset.workspace = true
 cfg-if = "1.0"
 lazy_static = "1.4"
-libc = { version = "^0.2", default-features = false }
+libc = { version = "=0.2.153", default-features = false }
```
theduke, line 41: We really can't pin to a specific libc version, that will cause a lot of hassle for downstream users.

At most we can use a minimum version with the required features.

syrusakbary: This reminds me that we should use a workspace version for libc (only set in the root Cargo.toml)

maminrayej: `libc 0.2.154` was yanked so cargo deny was failing.

syrusakbary (approved):

PR #4709, opened 2024-05-17
Remove `clock_time_get` instrumentation

maminrayej:
This PR reverts the instrumentation added to `clock_time_get` since it proved to add too much noise to the logs. Also, it adds an explanation as to why this syscall does not have an instrumentation.

Although it may help debugging some scenarios, especially in case of clock related bugs (see #4682), it makes debugging other scenarios more difficult since it clutters the logs.

### DIFF
--- a/lib/wasix/src/syscalls/wasi/clock_time_get.rs
+++ b/lib/wasix/src/syscalls/wasi/clock_time_get.rs
@@ -1,6 +1,9 @@
 use super::*;
 use crate::syscalls::*;
 
+// NOTE: This syscall is not instrumented since it will be logged too much,
+// hence introducing too much noise to the logs.
+
 /// ### `clock_time_get()`
 /// Get the time of the specified clock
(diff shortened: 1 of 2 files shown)


lib/wasix/src/syscalls/wasi/clock_time_get.rs
```diff
@@ -11,8 +14,6 @@ use crate::syscalls::*;
 /// Output:
 /// - `Timestamp *time`
 ///     The value of the clock in nanoseconds
-//#[instrument(level = "trace", skip_all, fields(?clock_id, %precision), ret)]
-#[instrument(level = "trace", skip_all, ret)]
```
theduke, line 15: If it's useful, we can keep this instrumentation behind an additional feature flag, like

```
#[cfg_attr(feature = "extra-logging", tracing::instrument(...)]
```

theduke (approved):

PR #4726, opened 2024-05-21
Update LLVM install instructions to use v15

XinyuZeng:
# Description

Update LLVM instructions to use v15.

Also wondering whether using the lastest stable llvm would be ok instead of 15?


syrusakbary (approved):

syrusakbary:
We can upgrade to latest LLVM, as soon as inkwell doesn't complain

PR #4730, opened 2024-05-22
Mount current directory to `/home` and Set CWD to `/home` when running WASI

### LINKED ISSUE #4724: Fix `wasmer run xyz --dir=.`
In order to reproduce you can run:

```
wasmer run python --dir=. -- file.py
```

### LINKED ISSUE #4725: Set CWD to /home when running WASI
The solution should set the cwd to `/home` when running WASI and mount `.` automatically into `/home` by default.

Related to #4724 .

This should work:

```
wasmer run python/python --dir=. -- main.py
```

### LINKED ISSUE #4753: Fail to get the file discriptor?
### Describe the bug
Fail to get the file descriptor, while other Wasm runtimes and native could get it successfully.

### Steps to reproduce
The c file is :
```
#include <stdio.h>
#include <fcntl.h>
#include <unistd.h>


int get_fd(const char *filename, int flags) {
    int fd = open(filename, flags);
    
    if (fd == -1) {
        printf("Get file descriptor of file %s failed!\n", filename);
        return -1;
    } else {
        printf("Get file descriptor of file %s succeed!\n", filename);
        return fd;
    }
}


int main() {
    int fd = get_fd("./hello.txt", O_RDWR);
    return 0;
}
```

(1)compile to Wasm:
`./wasi-sdk-21.0/bin/clang --target=wasm32-unkown-wasi --sysroot=./wasi-sdk-21.0/share/wasi-sysroot read.c -o read.wasm`
(2)compile to native code:
`gcc read.c -o read`
(3)Running wasm:
`wasmer run --dir=. read.wasm`
(4)Running native binaries
`./read`

### Expected behavior
Output:
`Get file descriptor of file ./hello.txt succeed!`
And this is the output of native and other Wasm runtimes, such as wasmtime.

### Actual behavior
wasmer prints:
`Get file descriptor of file ./hello.txt failed!`

### Additional context
Ubuntu 20.04
wasmer-4.3.1

maminrayej:
Resolves #4725
Resolves #4724
Resolves #4753


theduke:
Note: this would be a breaking behaviour change.

Are we okay with that?
Have we thought through the consequences?
Should this be an opt-in config setting instead?

lib/cli/src/commands/run/wasi.rs
```diff
@@ -483,7 +482,21 @@ impl Wasi {
             mapped_dirs.push(mapping);
         }
 
-        Ok(mapped_dirs)
+        if !have_current_dir {
+            let current_dir =
+                std::env::current_dir().context("could not determine current directory")?;
```
syrusakbary, line 487: This needs to be explicit, we shouldn't do this (basically only `wasmer run XYZ --dir=.` should mount the current dir

syrusakbary (commented):
Hey, I saw that we went into a different framework for testing that wasix-fyi, and I think this has some tradeoffs.

That means, that we should have two steps:
1. Build (https://github.com/wasmerio/wasmer/blob/main/tests/wasi-fyi/build.sh)
2. Test (in the similar fashion that wasi-fyi, https://github.com/wasmerio/wasmer/blob/main/tests/wasi-fyi/wasm-test.sh) -> Please see how the mapped dirs (via `$input_base.dir`), args and so on are provided in a framework-like fashion). We should use the same strategy here

syrusakbary (approved):

PR #4732, opened 2024-05-22
Poll for write-readiness after connecting a blocking socket

Arshia001:


theduke (approved):
LGTM

We should create a followup issue about returning the proper EINPROGRESS for non-blocking sockets as appropriate.

PR #4747, opened 2024-05-25
Fix for interrupts

john-sharratt:


syrusakbary (approved):

lib/wasix/src/syscalls/mod.rs
```diff
@@ -380,18 +379,37 @@ where
                 Errno::Child.into()
             }))));
         }
-        if self.process_signals && env.thread.has_signals_or_subscribe(cx.waker()) {
```
theduke, line 383: I think there was a reason for the process_signals setting, wasn't there?

theduke:
Discussed in person: fix seems valid, good to merge.

PR #4749, opened 2024-05-26
Only warn if needed

john-sharratt:
This patch will suppress the stack corruption warnings if asyncify is not used and hence its not a problem. This is to prevent all the large number of warnings that are being generated.

### DIFF
--- a/lib/wasix/src/state/env.rs
+++ b/lib/wasix/src/state/env.rs
@@ -95,6 +95,10 @@ pub struct WasiInstanceHandles {
     /// when a CTRL-C is pressed.
     pub(crate) signal_set: bool,
 
+    /// Flag that indicates if the stack capture exports are being used by
+    /// this WASM process which means that it will be using asyncify
+    pub(crate) has_stack_checkpoint: bool,
+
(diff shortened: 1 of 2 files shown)


Arshia001:
I was thinking of implementing the same thing. I was going to check for the asyncify exports instead. Indeed, it's possible to have a WASIX module that hasn't had the asyncify pass performed on it and can't ever unwind. I guess the best course of action would be to check for both the asyncify exports *and* the wasix imports?

theduke:
Somewhat agree with both sides here:

* if we warning is sensible, then we should only skip it if no asyncify is ever done
* but, we might want to downgrade from a warning to some other trace

lib/wasix/src/state/env.rs
```diff
@@ -439,6 +448,12 @@ impl WasiEnv {
         self.thread.tid()
     }
 
+    /// Returns true if this WASM process will need and try to use
+    /// asyncify while its running which normally means.
+    pub fn will_use_asyncify(&self) -> bool {
```
syrusakbary, line 453: Can we #[inline] this function?

syrusakbary (approved):

PR #4752, opened 2024-05-27
Fix stale wat version

XinyuZeng:
# Description

wat is fixed at `=1.0.71` previously, which causes version conflicts when I compare wasmer and wasmtime in the same repo. Note that update wat also needs to update the rust-toolchain to 1.76 to make the code compile.


maminrayej:
Thanks for the PR. It seems the change is not as trivial as we thought. Feel free to work on it if that's urgent. We will follow up on this in the future.

theduke:
@XinyuZeng in particular, the wasmparer version needs to be updated across the whole repository so we don't have duplicate versions.

XinyuZeng:
I don't have bandwidth on this right now. Others can take this if they are in a hurry.

xdoardo:
This effort is now part of #5019. Closing this PR. Thanks @XinyuZeng!

PR #4754, opened 2024-05-27
feat(config): Add app instaboot config

theduke:
Add bootstrapping configuration in the app.yaml config.

### DIFF
--- a/lib/config/src/app/healthcheck.rs
+++ b/lib/config/src/app/healthcheck.rs
@@ -11,13 +11,9 @@ pub enum HealthCheckV1 {
     schemars::JsonSchema, serde::Serialize, serde::Deserialize, PartialEq, Eq, Clone, Debug,
 )]
 pub struct HealthCheckHttpV1 {
-    /// Path to the health check endpoint.
-    pub path: String,
-    /// HTTP method to use for the health check.
-    ///
-    /// Defaults to GET.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub method: Option<String>,
+    #[serde(flatten)]
+    pub request: super::HttpRequest,
+
     /// Interval for the health check.
     ///
     /// Format: 1s, 5m, 11h, ...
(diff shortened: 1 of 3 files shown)


lib/config/src/app/http.rs
```diff
@@ -0,0 +1,31 @@
+/// Defines an HTTP request.
+#[derive(
+    schemars::JsonSchema, serde::Serialize, serde::Deserialize, PartialEq, Eq, Clone, Debug,
+)]
+pub struct HttpRequest {
+    /// Path to the health check endpoint.
```
syrusakbary, line 6: This is not a health check endpoint. This is a warmup endpoint

lib/config/src/app/http.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub path: String,
+    /// HTTP method to use for the health check.
+    ///
+    /// Defaults to GET.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub method: Option<String>,
```
syrusakbary, line 12: This should be an enum of the following: GET, POST, PUT, HEAD, OPTIONS

lib/config/src/app/http.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub timeout: Option<String>,
+    /// Expected status codes that are considered a pass for the health check.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub expected_status_codes: Vec<u16>,
+    /// Optional text that is in the body of the response that is considered a pass for the health check.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expected_body_includes: Option<String>,
+    /// Regular expression tested against the body that is considered a pass for the health check
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expected_body_regex: Option<String>,
```
syrusakbary, line 30: This is not a health check

syrusakbary, line 30: I'd wrap all this in an expect struct:
```
expect:
  status_codes: []
  body:
    includes:
    regex:
```

lib/config/src/app/healthcheck.rs
```diff
@@ -11,8 +11,9 @@ pub enum HealthCheckV1 {
     schemars::JsonSchema, serde::Serialize, serde::Deserialize, PartialEq, Eq, Clone, Debug,
 )]
 pub struct HealthCheckHttpV1 {
-    /// Path to the health check endpoint.
-    pub path: String,
+    #[serde(flatten)]
+    pub request: super::HttpRequest,
```
syrusakbary, line 15: all the other fields of `HttpRequest` are duplicated/redefined here

lib/config/src/app/http.rs
theduke: * It's easier to not have a typing layer here, because we'd need to do conversions into `http` anyway. Keeping this a String was intentional
* HTTP methods are in theory arbitrary, the well-known ones are just the standard

lib/config/src/app/http.rs
theduke: It doesn't hurt to have checks available here as well, since you might want to fail the bootstrapping if certain things don't work.

theduke: This layout was already used for health checks.
We can change things, since health checks aren't really exposed yet, but we first need to check , can't just break the schema.

lib/config/src/app/http.rs
theduke: ?

lib/config/src/app/healthcheck.rs
theduke: Yep, some duplications left in, those are fixed.

lib/config/src/app/http.rs
syrusakbary: The comments are incorrect: we say "Path to the health check endpoint.". When in fact is not a health check endpoint

lib/config/src/app/http.rs
ayys: > HTTP methods are in theory arbitrary, the well-known ones are just the standard

oooh TIL! This is pretty cool. This took me down a rabbit hole of all the nifty things custom http methods are used for.

lib/config/src/app/http.rs
syrusakbary: Clarifying: The comment "    /// Regular expression tested against the body that is considered a pass for the health check" (and the rest of comments) are wrong. This is not a health check. Correct the comments

syrusakbary: If we were not using them, then better reason for doing any breaking change now rather than later?

lib/config/src/app/http.rs
theduke: Closing as I think this is implemented as it should be.
Please re-open if you disagree.

syrusakbary (approved):

PR #4762, opened 2024-05-29
fix(wasix): Remove redundant module hash generation in BinaryPackage

### LINKED ISSUE #4785: Resolve redundant BinaryPackage hashing
We currently have redundant hashing of the webc in BinaryPackages, which is no longer needed for webc v3s , since the hash is contained in the webc itself.<br><br>The redundant hashing step should be removed.

theduke:
The hash is already available, no need to re-compute.

Closes #4785

### DIFF
--- a/lib/wasix/src/bin_factory/binary_package.rs
+++ b/lib/wasix/src/bin_factory/binary_package.rs
@@ -173,18 +173,25 @@ impl BinaryPackage {
         self.commands.iter().find(|cmd| cmd.name() == name)
     }
 
-    /// Get the bytes for the entrypoint command.
-    pub fn entrypoint_bytes(&self) -> Option<&[u8]> {
+    /// Resolve the entrypoint command name to a [`BinaryPackageCommand`].
+    fn get_entrypoint_cmd(&self) -> Option<&BinaryPackageCommand> {
         self.entrypoint_cmd
             .as_deref()
             .and_then(|name| self.get_command(name))
-            .map(|entry| entry.atom())
     }
 
+    /// Get the bytes for the entrypoint command.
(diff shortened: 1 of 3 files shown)


maminrayej (approved):

lib/wasix/src/bin_factory/binary_package.rs
```diff
@@ -173,18 +173,25 @@ impl BinaryPackage {
...
         self.entrypoint_cmd
             .as_deref()
             .and_then(|name| self.get_command(name))
-            .map(|entry| entry.atom())
     }
 
+    /// Get the bytes for the entrypoint command.
+    pub fn entrypoint_bytes(&self) -> Option<&[u8]> {
+        self.get_entrypoint_cmd().map(|entry| entry.atom())
+    }
```
syrusakbary, line 186: It seems `entrypoint_bytes` is only used in `bin_factor/exec.rs`. We should remove this function and refactor that code that is using it to be simpler:

```rust
if let Some(cmd) = binary.get_command(name).or(binary.get_entrypoint_cmd()) {
   ...
}
else {
}
```

theduke: It's a public function, we shouldn't just remove those randomly.

syrusakbary: We can mark it as deprecated if you prefer, and stop using it on our own code.

I don't think it adds any value (as we are the main users of the crate), but is your call if you prefer to deprecate it instead of deleting it

theduke: Deprecated

PR #4777, opened 2024-05-31
Disallow `path_open` absolute path

### LINKED ISSUE #4774: Should not be able to open absolute path
### Describe the bug

Absolute paths are only supported if the user explicitly preopens root. I've verified this in this [WASI issue](https://github.com/WebAssembly/WASI/issues/374).

```sh
$ wasmer -vV; rustc -vV
wasmer 4.3.1 (b986b4a 2024-05-28)
binary: wasmer-cli
commit-hash: b986b4a0faa56a30daa3554feff29e191849da65
commit-date: 2024-05-28
host: x86_64-apple-darwin
rustc 1.78.0 (9b00956e5 2024-04-29)
binary: rustc
commit-hash: 9b00956e56009bab2aa15d7bff10916599e3d6d6
commit-date: 2024-04-29
host: x86_64-apple-darwin
release: 1.78.0
LLVM version: 18.1.2
```

### Steps to reproduce

Compile and run this Rust example with `wasi` crate v0.11 (for preview1 API). Do not preopen root `/`.

```rust
fn main() {
    unsafe {
        let fd = wasi::path_open(
            5,
            0,
            "/",
            0,
            wasi::RIGHTS_FD_READ | wasi::RIGHTS_FD_WRITE,
            0,
            0,
        )
        .unwrap();

        eprintln!("fd: {fd}");

        let fd = wasi::path_open(
            fd,
            0,
            "abc",
            wasi::OFLAGS_CREAT,
            wasi::RIGHTS_FD_READ | wasi::RIGHTS_FD_WRITE,
            0,
            0,
        )
        .unwrap();

        eprintln!("fd: {fd}");
    }
}
```

### Expected behavior

This first `path_open` call should error. This behavior is consistent among other Wasm runtimes (Wasmtime, WasmEdge, Wazero, Node, WAMR).

### Actual behavior

Wasmer returns a valid fd which the second call uses to successfully `creat` a file in the Wasmer virtual root.

yagehu:
This commit changes `path_open` to return `inval` when it is called with an absolute path.  This behavior is consistent with other runtimes (Wasmtime, WasmEdge, Wazero, WAMR, Node).  This does not break wasi-libc behavior.  WASI folks [have stated][1] absolute paths should not work at the raw WASI level.

fixes #4774

[1]: https://github.com/WebAssembly/WASI/issues/374

### DIFF
--- a/lib/wasix/src/syscalls/wasi/path_open.rs
+++ b/lib/wasix/src/syscalls/wasi/path_open.rs
@@ -128,6 +128,15 @@ pub(crate) fn path_open_internal(
     let (memory, mut state, mut inodes) =
         unsafe { env.get_memory_and_wasi_state_and_inodes(&ctx, 0) };
 
+    if !wasi_try_ok_ok!(state.fs.get_fd_inode(dirfd))
+        .deref()
+        .name
+        .starts_with('/')
+        && path.starts_with('/')
+    {
+        return Ok(Err(Errno::Inval));
+    }
+
     let path_arg = std::path::PathBuf::from(&path);
     let maybe_inode = state.fs.get_inode_at_path(
         inodes,
--- a/tests/wasi-fyi/fs_open_root.rs
+++ b/tests/wasi-fyi/fs_open_root.rs
@@ -1,0 +1,44 @@
+use std::ffi::CString;
+use std::fs::File;
+
+#[link(wasm_import_module = "wasi_snapshot_preview1")]
+extern "C" {
+    pub fn path_open(
+        fd: i32,
+        dirflags: i32,
+        path: i32,
+        path_len: i32,
+        oflags: i32,
+        fs_rights_base: i64,
+        fs_rights_inheriting: i64,
+        fdflags: i32,
+        result_fd: i32,
+    ) -> i32;
+}
+
+const ERRNO_INVAL: i32 = 28;
+
+fn main() {
+    unsafe {
+        let fd = 5;
+        let path = CString::new("/fyi/should-fail").unwrap();
+
+        let errno = path_open(
+            fd,
+            0,
+            path.as_ptr() as i32,
+            path.as_bytes().len() as i32 + 1,
+            0,
+            0,
+            0,
+            0,
+            1024,
+        );
+        assert_eq!(
+            errno, ERRNO_INVAL,
+            "open absolute path at WASI level shall fail"
+        );
+
+        assert!(File::open("/hamlet/README.md").is_ok());
+    }
+}


maminrayej:
According to the [spec](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#path_open), the `path` passed to `path_open` must be relative. So, I believe the check for path must be done in `path_open` and not in `path_open_internal`. `path_open_internal` expects an absolute path since `path_open` converts the relative path to absolute using the current working directory.

But returning an error from this check in `path_open`:
```rust
// Convert relative paths into absolute paths
if path_string.starts_with("./") {
    path_string = ctx.data().state.fs.relative_path_to_absolute(path_string);
    trace!(
        %path_string
    );
} else {
    return Ok(Errno::Inval);
}
```
breaks bunch of other things, so I'm not sure of the implications.

cc @syrusakbary

yagehu:
Can you give an example of what this change breaks? I didn't see any failing tests. I can certainly revise my change and move it into `path_open`. I thought `path_open_internal` is only called by `path_open`.

This change also didn't break any higher level test examples involving wasi-libc.

maminrayej:
Your changes are fine. The point I was trying to make is that `path_open` just should not accept absolute paths in the first place according to the spec.  So, calling `path_open` with an absolute path should fail in `path_open` directly (before calling `path_open_internal`). But doing so will break other behaviors.

lib/wasix/src/syscalls/wasi/path_open.rs
```diff
@@ -128,6 +128,15 @@ pub(crate) fn path_open_internal(
     let (memory, mut state, mut inodes) =
         unsafe { env.get_memory_and_wasi_state_and_inodes(&ctx, 0) };
 
+    if !wasi_try_ok_ok!(state.fs.get_fd_inode(dirfd))
+        .deref()
+        .name
+        .starts_with('/')
+        && path.starts_with('/')
```
maminrayej, line 135:
```rust
&& path.starts_with('/') // This should be redundant since path_open converts relative paths to absolute
 ```

File: tests/wasi-fyi/fs_open_root.rs [hunk 2]
```diff
@@ -0,0 +1,44 @@
...
+        fdflags: i32,
+        result_fd: i32,
+    ) -> i32;
+}
+
+const ERRNO_INVAL: i32 = 28;
+
+fn main() {
+    unsafe {
+        let fd = 5;
```
maminrayej, line 23: Does this `fd` change if the runtime pre-opens things differently? For example, it can be `6`, or `7`, or ...

yagehu: Yes, potentially. AFAIK wasmer is the only runtime that supplies additional preopens. Other runtimes mostly begins at 3 (after stdin, stdout, and stderr). With the [proposed update to wasi-afk](https://github.com/wasmerio/wasmer/issues/4771), we can add a function that calls `fd_prestat_get` and dynamically determines the base fd and share this helper function. I'll leave it up to you if you want to create a issue for that.

yagehu:
@maminrayej Actually this PR is problematic. I just realized `inode.name` is always `/` or `.` regardless of `--mapdir`. The `dirfd` that `path_open` gets is not actually the mapped directory. I misunderstood the code.

maminrayej (approved):

syrusakbary (changes requested):
This will break other behaviors in many other places (including places where Wasmer+WASI is being used for plugins for example), we can't accept this PR for now

yagehu:
> This will break other behaviors in many other places (including places where Wasmer+WASI is being used for plugins for example), we can't accept this PR for now

OK! Understood. I will close this PR. However, this should only break application if (1) it is using raw WASI, and (2) it calls path_open with absolute path. Calls like `std::fs::File::open()` and `open(2)` are not affected.

PR #4783, opened 2024-06-03
Fix metadata conversion to carry over the modified timestamp

maminrayej:
This PR fixes the issue for returning zero modified timestamp when calling `metadata` or `read_dir`.

### DIFF
--- a/lib/virtual-fs/src/webc_volume_fs.rs
+++ b/lib/virtual-fs/src/webc_volume_fs.rs
@@ -296,20 +296,34 @@ impl AsyncWrite for File {
 }
 
 fn compat_meta(meta: webc::compat::Metadata) -> Metadata {
+    // HACK: timestamps are not present in webc v2, so we have to return
+    // a stub modified time. previously we used to just return 0, but that
+    // proved to cause problems with programs that interpret the value 0.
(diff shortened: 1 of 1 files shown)


lib/virtual-fs/src/webc_volume_fs.rs
```diff
@@ -295,21 +293,37 @@ impl AsyncWrite for File {
...
 
+// HACK: timestamps are not present in webc v2, so we have to return
+// a stub modified time. previously we used to just return 0, but that
+// proved to cause problems with programs that interpret the value 0.
+// to circumvent this problem, we decided to return a non-zero value.
+fn get_modified(timestamps: Option<webc::Timestamps>) -> u64 {
+    static DEFAULT: OnceLock<u64> = OnceLock::new();
+
+    timestamps
+        .map(|t| t.modified())
```
theduke, line 304: Doesn't this introduce somewhat weird random behaviour?

Why not just use `1` instead?

maminrayej: I'm also fine with returning `1`.

theduke (approved):

PR #4805, opened 2024-06-04
CLI: app create - Template handling improvements

### LINKED ISSUE #4804: CLI: wasmer app create only shows a small subset of templates
The template selector shown by `wasmer app create` only shows a subset of the available , making it cumbersome to use.

This is due to us only doing a single request to the backend, and the low pagination size.
Need to do multiple requests, and ideally caching the templates for a while so they don't need to be re-downloaded.

theduke:
Adds a local template cache (currently 15 minutes), and fetches more templates
from the backend, currently up to 100.

Also sorts the templates by popularity now.

NOTE: we should really use dialoger::FuzzySelect to give users a search prompt,
but that currently conflicts with the custom stylings of the select items and
messes up the formatting.
(See https://github.com/console-rs/dialoguer/issues/312)

Closes #4804

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6380,7 +6380,7 @@ dependencies = [
 
 [[package]]
 name = "wasmer-api"
-version = "0.0.28"
+version = "0.0.29"
 dependencies = [
  "anyhow",
  "base64 0.13.1",
--- a/lib/backend-api/Cargo.toml
+++ b/lib/backend-api/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasmer-api"
-version = "0.0.28"
+version = "0.0.29"
 description = "Client library for the Wasmer GraphQL API"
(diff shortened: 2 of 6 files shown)


xdoardo (approved):

theduke:
Note: we should add a different step in here that improves timings:

* always fetch the first page of templates
* compare the results with the first page in the cache
* if equivalent, we can use the cache, otherwise we refetch

That'll improve performance, because right now the cache time is just set to 15 minutes.

syrusakbary (approved):

theduke:
Note, I improved the caching logic by :

    * Always use the cache for a given period (1h)
    * If cache is older than that, just fetch the first page, and compare
      if no changes, then can re-use the cache
    * Otherwise, refetch all

syrusakbary (approved):

PR #4810, opened 2024-06-05
feat(cli): Add "app purge-cache" command

### LINKED ISSUE #4795: feat(cli): Add "app purge-cache" command
Add a new command `app purge-cache <app-ident>`, which allows purging the cache for an app.<br><br>Should use the GraphQL mutation added in [BE-306](https://linear.app/wasmer/issue/BE-306/graphql-api-add-mutation-for-app-cache-purging).<br><br>The command should accept a cache scope (`instaboot`, `cdn`, …)

theduke:
Closes #4795

- **chore: Fix some warnings**
- **feat(backend-api): Add app cache purging**
- **docs: Add small comment to AppIdentOpts**
- **feat(cli): Add "app purge-cache" command**

### DIFF
--- a/lib/cli/src/commands/app/purge_cache.rs
+++ b/lib/cli/src/commands/app/purge_cache.rs
@@ -36,7 +36,7 @@ impl AsyncCliCommand for CmdAppPurgeCache {
 
         let version_id = app.active_version.id;
 
-        let name = format!("{}/{}", app.owner.global_name, app.name);
+        let name = format!("{} ({})", app.name, app.owner.global_name);
 
         println!(
             "Purging caches for {}, app version {}...",
--- a/lib/cli/src/commands/app/purge_cache.rs
+++ b/lib/cli/src/commands/app/purge_cache.rs
@@ -47,7 +47,7 @@ impl AsyncCliCommand for CmdAppPurgeCache {
         let vars = wasmer_api::types::PurgeCacheForAppVersionVars { id: version_id };
         wasmer_api::query::purge_cache_for_app_version(&client, vars).await?;
 
-        println!("Caches purged!");
+        println!("🚽 swirl! All caches have been purged!");
 
         Ok(())
     }
--- a/lib/cli/src/commands/app/purge_cache.rs
+++ b/lib/cli/src/commands/app/purge_cache.rs
@@ -1,0 +1,54 @@
+//! Get information about an edge app.
+
+use super::util::AppIdentOpts;
+use crate::{
+    commands::AsyncCliCommand,
+    opts::{ApiOpts, ItemFormatOpts},
+};
+
+/// Purge caches for applications.
+///
+/// Cache scopes that can be cleared:
+/// * InstaBoot startup snapshots
+///   Will delete all existing snapshots.
+///   New snapshots will be created automatically.
+#[derive(clap::Parser, Debug)]
+pub struct CmdAppPurgeCache {
+    #[clap(flatten)]
+    #[allow(missing_docs)]
+    pub api: ApiOpts,
+    #[clap(flatten)]
+    #[allow(missing_docs)]
+    pub fmt: ItemFormatOpts,
+
+    #[clap(flatten)]
+    #[allow(missing_docs)]
+    pub ident: AppIdentOpts,
+}
+
+#[async_trait::async_trait]
+impl AsyncCliCommand for CmdAppPurgeCache {
+    type Output = ();
+
+    async fn run_async(self) -> Result<(), anyhow::Error> {
+        let client = self.api.client()?;
+        let (_ident, app) = self.ident.load_app(&client).await?;
+
+        let version_id = app.active_version.id;
+
+        let name = format!("{}/{}", app.owner.global_name, app.name);
+
+        println!(
+            "Purging caches for {}, app version {}...",
+            name,
+            version_id.inner()
+        );
+
+        let vars = wasmer_api::types::PurgeCacheForAppVersionVars { id: version_id };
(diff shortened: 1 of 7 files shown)


lib/cli/src/commands/app/purge_cache.rs
```diff
@@ -0,0 +1,54 @@
...
+        println!(
+            "Purging caches for {}, app version {}...",
+            name,
+            version_id.inner()
+        );
+
+        let vars = wasmer_api::types::PurgeCacheForAppVersionVars { id: version_id };
+        wasmer_api::query::purge_cache_for_app_version(&client, vars).await?;
+
+        println!("Caches purged!");
```
ayys, line 50: half joking:
```suggestion
        println!("🚽 Flush! All caches have been flushed!");
```

lib/cli/src/commands/app/purge_cache.rs
```diff
@@ -0,0 +1,54 @@
...
+    #[clap(flatten)]
+    #[allow(missing_docs)]
+    pub api: ApiOpts,
+    #[clap(flatten)]
+    #[allow(missing_docs)]
+    pub fmt: ItemFormatOpts,
+
+    #[clap(flatten)]
+    #[allow(missing_docs)]
+    pub ident: AppIdentOpts,
```
syrusakbary, line 26: This should not be required, by default we should check if there's an `app.yaml` in the current dir and get the app related to it in case it exists, so the user doesn't need to worry about the app id.

Also this should be an optional, and be called id:
```
    pub id: Optional<AppIdentOpts>,
```

lib/cli/src/commands/app/purge_cache.rs
```diff
@@ -0,0 +1,54 @@
...
+/// Purge caches for applications.
+///
+/// Cache scopes that can be cleared:
+/// * InstaBoot startup snapshots
+///   Will delete all existing snapshots.
+///   New snapshots will be created automatically.
+#[derive(clap::Parser, Debug)]
+pub struct CmdAppPurgeCache {
+    #[clap(flatten)]
+    #[allow(missing_docs)]
```
syrusakbary, line 18: We should have the docs for this

lib/cli/src/commands/app/purge_cache.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub api: ApiOpts,
+    #[clap(flatten)]
+    #[allow(missing_docs)]
```
syrusakbary, line 21: We should have the docs for this

lib/cli/src/commands/app/purge_cache.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub fmt: ItemFormatOpts,
+
+    #[clap(flatten)]
+    #[allow(missing_docs)]
```
syrusakbary, line 25: We should have the docs for this

lib/cli/src/commands/app/purge_cache.rs
```diff
@@ -0,0 +1,54 @@
...
+impl AsyncCliCommand for CmdAppPurgeCache {
+    type Output = ();
+
+    async fn run_async(self) -> Result<(), anyhow::Error> {
+        let client = self.api.client()?;
+        let (_ident, app) = self.ident.load_app(&client).await?;
+
+        let version_id = app.active_version.id;
+
+        let name = format!("{}/{}", app.owner.global_name, app.name);
```
ayys, line 39: nit: use `app.global_name` instead?

lib/cli/src/commands/app/purge_cache.rs
theduke: If you read the code, you will realize that this already happens.

lib/cli/src/commands/app/purge_cache.rs
theduke: No we should not, because the documentation is on the nested field.

I recommend reading the comments on the struct.

lib/cli/src/commands/app/purge_cache.rs
ayys: sidenote: all resources have a `globalName` field on the backend that should be used when refering to it, instead of building the name manually everywhere. This is to keep things consistent across clients/commands.

lib/cli/src/commands/app/purge_cache.rs
syrusakbary: Will the CLI show those comments?

lib/cli/src/commands/app/purge_cache.rs
ayys: it will be in 15 minutes ;)

lib/cli/src/commands/app/purge_cache.rs
syrusakbary: > Also this should be an optional, and be called **id**

Way clearer for the user

lib/cli/src/commands/app/purge_cache.rs
syrusakbary: If you don't want to fix this by using globalName, then please use another nomenclature for the app.

We use:
```suggestion
        let name = format!("{} ({})", app.name, app.owner.global_name);
```

lib/cli/src/commands/app/purge_cache.rs
theduke: Haha whatever, let's go for it.

lib/cli/src/commands/app/purge_cache.rs
theduke: Renamed to `app`

syrusakbary (approved):

PR #4819, opened 2024-06-07
Add support for spawning local wasm files

maminrayej:
This PR adds support for spawning local wasm files.


lib/wasix/src/bin_factory/exec.rs
```diff
@@ -29,15 +29,25 @@ pub async fn spawn_exec(
...
+pub async fn spawn_exec_wasm(
+    wasm: &[u8],
+    name: &str,
+    env: WasiEnv,
+    runtime: &Arc<dyn Runtime + Send + Sync + 'static>,
+) -> Result<TaskJoinHandle, SpawnError> {
     // Load the module
     let module = spawn_load_module(&env, name, wasm, runtime).await?;
 
-    // Spawn union the file system
```
john-sharratt, line 38: not sure what this comment means, does it mean Spawn the union file system? I don't think we spawn it though we set it up

lib/wasix/src/bin_factory/mod.rs
```diff
@@ -145,6 +144,90 @@ impl BinFactory {
...
+                    spawn_exec_wasm(&bytes, name.as_str(), env, &self.runtime).await
+                }
+                Executable::BinaryPackage(pkg) => {
+                    spawn_exec(pkg, name.as_str(), store, env, &self.runtime).await
+                }
+            }
+        })
+    }
+
+    // TODO: remove allow once BinFactory is refactored
```
john-sharratt, line 179: Can't we use the bin factory here now?

john-sharratt (approved):

lib/wasix/src/bin_factory/exec.rs
maminrayej: You're right. This is a leftover comment from `spawn_exec`.

lib/wasix/src/bin_factory/mod.rs
maminrayej: I'm not sure what you mean.

john-sharratt (approved):

PR #4826, opened 2024-06-11
chore(virtual-fs)!: Remove old stale WebcFileSystem

theduke:
The old WebcFileSystem in virt-fs is not used anymore, and was just for webc v1
support, which has long since been removed.

This commit removes the implementation.

The associated tests should be re-written to WebcVolumeFs, but
preferably in a followup.


lib/virtual-fs/src/overlay_fs.rs
```diff
@@ -1688,4 +1529,153 @@ mod tests {
             FsError::EntryNotFound
         )
     }
+
+    // OLD tests that used WebcFileSystem.
+    // Should be re-implemented with WebcVolumeFs
+    // #[tokio::test]
+    // async fn wasi_runner_use_case() {
```
maminrayej, line 1536: Is it possible to reuse this test for `webc_volume_fs`?

syrusakbary: Probably, but we can do it in a PR apart

syrusakbary (approved):

PR #4838, opened 2024-06-12
Don't pass auth header when downloading a package

xdoardo:
This small patch removes the auth header passed to Wasmer's cdn when downloading a package. Of course, this does not remove the mandatory auth when interacting with the graphql endpoint, which is still necessary. 

Furthermore, it makes the `--out-path` optional, and in case the user does not give a value for the flag it uses
`$(pwd)/namespace--name@version.webc` or `$(pwd)/hash.webc` as a filename.

### DIFF
--- a/lib/cli/src/commands/package/download.rs
+++ b/lib/cli/src/commands/package/download.rs
@@ -128,7 +128,7 @@ impl PackageDownload {
                 let download_url = package
                     .distribution_v3
                     .pirita_download_url
-                    .context("registry does provide a container download container download URL")?;
+                    .context("registry did not provide a container download URL")?;
 
                 let ident = format!("{}@{}", full_name, package.version);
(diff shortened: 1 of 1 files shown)


syrusakbary:
Where this PR originates from? What is this trying to fix? I thought this was needed in the past for downloading the files

theduke (approved):

syrusakbary:
(note: even if approved, please do not merge until resolving my question below)

syrusakbary:
Context of the issue:

> `wasmer package download sha256:...` currently doesn't work
>
> Ideally it would be the resolved version (for named), so <ns>--<name>@<version>.webc or similiar.
>
> and first 8 chars of hash/full hash for unnamed

syrusakbary (approved):

PR #4840, opened 2024-06-12
fix(journal): Correctly handle client socket closing during compaction

### LINKED ISSUE #4839: Journal compaction: Outbound sockets not handled
Outbound sockets are currently not properly recorded in the  compactor, and thus are not properly pruned from compacted journals, which can cause issues.

theduke:
Previously outbound sockets were not properly recorded as open in the
compactor, and thus closed sockets were not properly removed from the
journal.

This change implements tracking for sockets after connecting to them.

Closes #4839

### DIFF
--- a/lib/journal/src/concrete/compacting.rs
+++ b/lib/journal/src/concrete/compacting.rs
@@ -685,6 +685,17 @@ impl WritableJournal for CompactingJournalTx {
                 state.descriptor_seed += 1;
                 state.open_pipes.insert(*fd2, lookup);
             }
+            JournalEntry::SocketConnectedV1 { fd, .. } => {
+                let lookup = DescriptorLookup(state.descriptor_seed);
+                state.descriptor_seed += 1;
+                state.open_sockets.insert(*fd, lookup);
+                state
+                    .descriptors
+                    .entry(lookup)
+                    .or_default()
+                    .events
+                    .push(event_index);
+            }
             // Sockets that are accepted are suspect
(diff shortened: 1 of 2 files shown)


lib/journal/src/concrete/compacting.rs
```diff
@@ -467,7 +467,7 @@ impl WritableJournal for CompactingJournalTx {
                 //  thus the entire branch of events it represents is discarded)
                 let mut erase = false;
                 let lookup = if matches!(&entry, JournalEntry::CloseFileDescriptorV1 { .. }) {
-                    if state.open_sockets.remove(fd).is_some() {
+                    if dbg!(state.open_sockets.remove(fd).is_some()) {
```
syrusakbary, line 470: Undo `dbg!` ?

john-sharratt (commented):
Did you mean to comment out all unit tests?

lib/journal/src/concrete/compacting.rs
theduke: Done

theduke:
> Did you mean to comment out all unit tests?

They were already commented out, I un-commented the runner and added a new test.

Followup should fix up the other tests.
I had a brief look, but they require some work.

syrusakbary (approved):

theduke:
@john-sharratt for posterity, I phrased it a bit incorrectly: before, non of the tests were running because they were gated behind the `journal` feature flag, which is probably a copy paste leftover (from the wasix crate).

The journal crate doesn't have a `journal` feature flag, so the whole test module was inactive.

Issue #4837, opened 2024-06-12
CLI: Terminal cursor disappears when exiting prompts with CTRL-C


theduke:
In favrious CLI commands that have a prompt implemented with `dialoguer`, the terminal cursor disappears when exiting a command with CTRL-C.

The cause  is described in this issue: https://github.com/console-rs/dialoguer/issues/77

We'll need to manually reset the cursor in the respective commands.

PR #4841, opened 2024-06-12
Restore cursor after SIGINT during dialogue

xdoardo:
As per title. This PR adds a `setup` function to `AsyncCliCommand` with a default implementation that installs a signal handler that restores the cursor (see https://github.com/console-rs/dialoguer/issues/77).


syrusakbary:
Note: this PR is trying to fix #4837

syrusakbary (commented):
I find surprising that we were not using `ctrlc` before. Because I know for journaling we listen on Ctrl^C.

We need to make sure that handler for WASIX and this one are not colliding before merging the PR (please check with @john-sharratt )

xdoardo:
(I force-pushed to remove a commit that was not related to this PR). 

The signal handler is installed in those commands that implement the `AsyncCliComand` trait, and `run` is not one of them - this, of course, if you meant a conflict on the signal handling from the WASIX layer. I'll check with John.

syrusakbary:
Just more context: the current signal handler hops in when saving journaling upon doing ctrl-c

lib/cli/src/commands/mod.rs
```diff
@@ -70,12 +71,21 @@ pub(crate) trait AsyncCliCommand: Send + Sync {
     type Output: Send + Sync;
 
     async fn run_async(self) -> Result<Self::Output, anyhow::Error>;
+
+    fn setup(&self) -> Result<(), anyhow::Error> {
+        ctrlc::set_handler(move || {
```
theduke, line 76: nit: no need for the separate `crtlc` dependency, we can use `tokio::signal::ctrl_c` .

With the current structure it'll be easiest to just run it in a spawned task.

theduke, line 76: Note: we should probably restrict this to TTYs. 

`stdin.is_terminal()`.

See https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html#tymethod.is_terminal

theduke:
@syrusakbary this will not be invoked for the `run` command, so it won't interfere with the journaling.

syrusakbary (approved):

theduke (commented):
LGTM now

Issue #4788, opened 2024-06-03
the CLI doesn't take `--dry-run` into account when tagging


ayys:
I am getting this output when running `cargo wasix` on https://github.com/wasmerio/pirita repo.
I suspect the issue is that `wasmer package tag` does not take into account the flag `--dry-run`. 

```
2024-06-03T12:29:50.695914Z  INFO publish: cargo_wasmer::publish: Getting ready to publish dry_run=true pkg="wapm-to-w
ebc"
✔ Correctly built package locally                                                                                     
✔ Skipping push as dry-run is set                                                                                     
2024-06-03T12:30:03.259461Z  WARN ThreadId(01) wasmer_cli::commands::package::push: Unable to invalidate the cache use
d for package version queries error=No such file or directory (os error 2)


The package with the required hash does not exist in the selected registry.
HINT: you are running `wasmer publish` with `--dry-run` set.

error: Can't tag, no matching package found in the registry.
Error: Unable to publish "wapm-to-webc"

Caused by:
    The wasmer CLI exited unsuccessfully with exit code 1
```

## Minimal example

1. create an empty dir with this `wasmer.toml`
```
[dependencies]
"wasmer/static-web-server" = "^1"
```

2. run `wasmer publish` it

Issue #4672, opened 2024-05-13
The `wasmer package tag` command should not need a manifest


xdoardo:
Right now the `wasmer package tag` command requires the package's manifest to be specified in some way, either explicitly giving its path or having it in the current working directory. <br><br>We should use the manifest only as a fallback in case the user did not provide every information necessary: as an idea, we could have the command in the format

```
wasmer package tag sha256:abcd...... namespace/name --version 0.1.0
```

with the second positional argument (`namespace/name`) being optional and filled in from the manifest if present, and the same for the version.

Issue #4845, opened 2024-06-13
CLI: "wasmer deploy" does not allow specifying a particular app.yaml anymore


theduke:
In the recent refactors, "wasmer deploy" has lost the option to specify the path
to a speciic `app.yaml`.

Not having that option is annoying when you have multiple app.yamls for one package,
since it requires creating a separate sub-directory for each app...

xdoardo:
The `--dir` flag allows to pass a path to the `app.yaml` file. 
We either: 
1. Change `--dir` to `--path` (breaking)
2. Leave `--dir` as it is and add `--path`

theduke:
Option 2 sounds good to me.

PR #4848, opened 2024-06-14
feat(cli/package-tag): Manifest is not mandatory anymore while tagging

xdoardo:
This PR tries to solve #4672 and #4788. 
This involves necessary breaking changes to the `tag` command which we introduced in 4.3.0, as before we had
```
wasmer package tag sha256:abc....    ./path/to/manifest_dir
```
and, in this PR, it should now be 
```
wasmer package tag sha256:abc.... namespace/name[version] 
```
So, keeping both args (path and namespace) to maintain backwards compatibility would make this PR nonsensical - as it would still need a manifest. Users can still specify a manifest to draw the info regarding the package to tag from with the `--package-path` flag. 

In general, this PR still keeps the concept of manifest in the tagging logic. This, of course, is to keep a minimum of backwards compatibility, but also because some package values - such as description, readme and so forth -  are better read off of the manifest, rather than passed via flag from the user.

### DIFF
--- a/lib/cli/src/commands/package/tag.rs
+++ b/lib/cli/src/commands/package/tag.rs
@@ -75,7 +75,7 @@ pub struct PackageTag {
     /// Directory containing the `wasmer.toml`, or a custom *.toml manifest file.
     ///
     /// Defaults to current working directory.
-    #[clap(long = "manifest-path", default_value = ".")]
+    #[clap(long = "path", default_value = ".")]
     pub package_path: PathBuf,
 
     /// Wait for package to be available on the registry before exiting.
--- a/lib/cli/src/commands/package/tag.rs
+++ b/lib/cli/src/commands/package/tag.rs
@@ -14,7 +14,9 @@ use std::{
     str::FromStr,
 };
 use wasmer_api::WasmerClient;
-use wasmer_config::package::{Manifest, NamedPackageId, PackageBuilder, PackageHash, PackageIdent};
+use wasmer_config::package::{
+    Manifest, NamedPackageId, NamedPackageIdent, PackageBuilder, PackageHash, PackageIdent,
+};
 
 use super::PublishWait;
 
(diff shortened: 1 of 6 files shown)


syrusakbary (approved):

lib/cli/src/commands/package/tag.rs
```diff
@@ -66,11 +68,14 @@ pub struct PackageTag {
...
-    ///
+    /// The package to tag.
+    #[clap(name = "package_ident")]
+    pub package_id: Option<NamedPackageIdent>,
+
     /// Directory containing the `wasmer.toml`, or a custom *.toml manifest file.
     ///
     /// Defaults to current working directory.
-    #[clap(name = "path", default_value = ".")]
+    #[clap(long = "manifest-path", default_value = ".")]
```
ayys, line 78: I am not sure calling it manifest-path is the right approach. It makes things confusing. 

```
      --dir <DIR>
          The path to the directory containing the `app.yaml` file
```

wasmer publish has

```
  [path]
          Directory containing the `wasmer.toml`, or a custom *.toml manifest file.
```

lib/cli/src/opts.rs
```diff
@@ -33,8 +33,11 @@ fn parse_registry_url(registry: &str) -> Result<url::Url, String> {
 
 #[derive(clap::Parser, Debug, Clone, Default)]
 pub struct ApiOpts {
+    /// The (optional) authorization token to pass to the registry
     #[clap(long, env = "WASMER_TOKEN")]
     pub token: Option<String>,
+
+    /// The registry to interact with
```
ayys, line 40: Maybe say this?
`Change the current registry`

lib/cli/src/commands/package/tag.rs
xdoardo: Regarding the `--dir` flag in `deploy`: https://github.com/wasmerio/wasmer/issues/4845. I'll rename `--manifest-path` to `--path` in `tag - I think the former is more descriptive but longer (and breaking), so let's just keep the symmetry.

lib/cli/src/opts.rs
xdoardo: I think there's the possibility that there's no "current" registry, right? Maybe `Select the registry`? `Select the registry to interact with`?

ayys: no there is always a current registry. by default it's registry.wasmer.io (afaik i have never seen a situation where there is no registry specified, but I could be wrong).

lib/cli/src/commands/package/tag.rs
ayys: sgtm

ayys (approved):
lgtm other than the nit about registry.

Issue #4859, opened 2024-06-17
CLI: Newly intalled signal handler impedes normal termination of the program


xdoardo:
With the [new signal handler we recently installed](https://github.com/wasmerio/wasmer/pull/4841), most CLI commands just hang after the CLI function is finished (e.g. `wasmer deploy`) until CTRL+C is pressed. I think this is due to the fact that `tokio::signal::ctrl_c` [completes when the signal is received](https://docs.rs/tokio/latest/tokio/signal/fn.ctrl_c.html).

PR #4860, opened 2024-06-17
CLI: Use a channel to signal that the main fn is done

xdoardo:
Tries to solve #4859. Use a channel to notify the `setup` thread that the main CLI function (`deploy`, `tag`, `push`..) is done and, in this case, should not wait for SIGINT to be received. Another approach could be that of simply calling `.abort()` on the `JoinHandle`.

### DIFF
--- a/lib/cli/src/commands/mod.rs
+++ b/lib/cli/src/commands/mod.rs
@@ -103,14 +103,11 @@ impl<O: Send + Sync, C: AsyncCliCommand<Output = O>> CliCommand for C {
             let (snd, rcv) = tokio::sync::oneshot::channel();
             let handle = self.setup(rcv);
 
-            match AsyncCliCommand::run_async(self).await {
-                Ok(k) => return Ok(k),
-                Err(e) => {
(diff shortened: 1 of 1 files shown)


syrusakbary (approved):
I think it will be better for @theduke to review it

lib/cli/src/commands/mod.rs
```diff
@@ -91,10 +100,15 @@ impl<O: Send + Sync, C: AsyncCliCommand<Output = O>> CliCommand for C {
 
     fn run(self) -> Result<(), anyhow::Error> {
         tokio::runtime::Runtime::new()?.block_on(async {
-            let handle = self.setup();
+            let (snd, rcv) = tokio::sync::oneshot::channel();
+            let handle = self.setup(rcv);
 
             AsyncCliCommand::run_async(self).await?;
```
theduke, line 106: If an error is returned, this fails to trigger the shutdown channel.

theduke (changes requested):

theduke (approved):

Issue #4847, opened 2024-06-14
CLI: wasmer deploy - Improve App Reachability Checks


theduke:
Currently the `wasmer deploy` command waits for the deployed app version to become reachable.

At the moment it waits until the responses have the expected version header, AND until the responses have a 2xx response code.

This causes long wait times without proper log information if the app is deployed, but doesn't respond with 2xx codes.

We should rewrite the logic to only wait for the header values to match, and log an error if the response is not 2xx.

Also, 3xx headers are valid success responses, so those should also be counted as successful deploys.

PR #4862, opened 2024-06-17
feat(cli/deploy): Check status when waiting for the app to be available

xdoardo:
As per title. Solves (part of?) #4847

### DIFF
--- a/lib/cli/src/commands/app/deploy.rs
+++ b/lib/cli/src/commands/app/deploy.rs
@@ -669,7 +669,7 @@ pub async fn wait_app(
                                 eprintln!();
                             }
                             if !(res.status().is_success() || res.status().is_redirection()) {
-                                eprintln!("The app version was deployed correctly, but fails with a non-success status code of {}", res.status())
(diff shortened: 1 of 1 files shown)


lib/cli/src/commands/app/deploy.rs
```diff
@@ -668,7 +668,12 @@ pub async fn wait_app(
                             if !quiet {
                                 eprintln!();
                             }
-                            eprintln!("{} Deployment complete", "𖥔".yellow().bold());
+                            if !(res.status().is_success() || res.status().is_redirection()) {
+                                eprintln!("The app version was deployed correctly, but fails with a non-success status code of {}", res.status())
+                            } else {
+                                eprintln!("{} Deployment complete", "𖥔".yellow().bold());
```
theduke, line 674: Shouldn't this be green, and the above yellow?

theduke (changes requested):

lib/cli/src/commands/app/deploy.rs
xdoardo: I think we decided in 4.3.0 to have `"𖥔".yellow().bold()`.  I'll make line 672 yellow.

theduke (approved):

PR #4883, opened 2024-06-21
cli(app/create): Ask for directory to create the app in

xdoardo:
This PR introduces a mechanism that asks users the directory they want to save their app in in case the default app directory is not empty.

### DIFF
--- a/lib/cli/src/commands/app/create.rs
+++ b/lib/cli/src/commands/app/create.rs
@@ -32,8 +32,10 @@ async fn write_app_config(app_config: &AppConfigV1, dir: Option<PathBuf>) -> any
         None => std::env::current_dir()?,
     };
 
+    tokio::fs::create_dir_all(&app_dir).await?;
+
     let app_config_path = app_dir.join(AppConfigV1::CANONICAL_FILE_NAME);
-    std::fs::write(&app_config_path, raw_app_config).with_context(|| {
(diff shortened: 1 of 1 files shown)


syrusakbary:
@xdoardo I assume this doesn’t affect the behavior of Wasmer deploy in a full dir, right?

xdoardo:
If running `wasmer deploy` does not end up calling `wasmer app create` then no, it does not change. If `wasmer deploy` is called in a directory that does not contain a manifest (nor an `app.yaml`) then it falls back to `wasmer app create` which will eventually ask the user a directory to store their app in. 

Note, however, that creating apps from existing directories was previously explicitly unsupported. In the near future we'll want to adopt a `create` mechanism (and thus a `deploy` mechanism as well) that allows to create an app from an existing project with no manifest, if I understood correctly.

syrusakbary:
> then it falls back to wasmer app create which will eventually ask the user a directory to store their app in.

So, it will affect? Basically for `wasmer deploy` in a directory with files and without a `app.yaml` it should prompt to deploy the own directory. While `wasmer app create` can prompt to create in a new directory. Both behaviors should work well @xdoardo

syrusakbary:
I got confused here. Chatted with @xdoardo and we are in alignment with this solution. Approving

syrusakbary (approved):

PR #4896, opened 2024-06-25
Re-use previously closed FDs

Arshia001:
This aligns with the Unix requirement that any newly returned FD must be the lowest unopened FD, and improves apps that use `select`, which only supports 1K FDs.

This PR also makes the FD table use ahash, which should be faster.


maminrayej (approved):

maminrayej (commented):
Having a test to pin down this behavior would be nice.

syrusakbary (approved):

theduke:
Retroactively approved.

PR #4897, opened 2024-06-25
App create: Ask for framework before fetching templates

xdoardo:
(attaching images as this is pretty much a design patch)

In order to speed up the CLI reducing the size of the data requested to the backend and to provide a better UX, we wanted to introduce a new mechanism to present available templates to users.  Before,  we presented a list of all the available templates to the user: 

Screenshot 2024-06-25 at 16 40 11

With this patch, the user first selects the framework they want and, after that, the actual template from the matching ones. 

Screenshot 2024-06-25 at 16 40 51
Screenshot 2024-06-25 at 16 41 14

### DIFF
--- a/lib/backend-api/schema.graphql
+++ b/lib/backend-api/schema.graphql
@@ -1248,12 +1248,13 @@ type AppTemplate implements Node {
   updatedAt: DateTime!
   readme: String!
   useCases: JSONString!
-  language: String!
   repoLicense: String!
   usingPackage: Package
   defaultImage: String
   framework: String!
   templateFramework: TemplateFramework
+  language: String!
+  templateLanguage: TemplateLanguage
 }
 
 type AppTemplateCategory implements Node {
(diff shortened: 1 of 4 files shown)


syrusakbary:
I'm not that sure of doing it via framework instead of language as we indicated on the original issue (there will be many frameworks that we support, so language seems like a better initial filter). Any insights on why this was done this way?

theduke:
I dont quite get why we select by framework first, because the framework list could just grow huge again eventually.

Making the first step selecting the language seems much more sensible?

xdoardo:
Alright, agreed on filtering by language. Waiting for the backend API to implement it that way.

ayys:
nit: should there also be a <go-back> option for when I want to view templates from another category/language/framework?

xdoardo:
Latest commit changes it to display languages first.

syrusakbary (approved):

ayys (approved):
discussed with @xdoardo and lgtm :rocket:

PR #4920, opened 2024-07-08
Compile `wasmer-api` crate to `wasm32-unknown-unknown`

xdoardo:
This patch resolves compiler errors arising when compiling the `wasmer-api` (a.k.a. `backend-api`) crate to `wasm32-unknown-unknown` targets. 

The main complication was the explicit use of a type (`Pin<Box<dyn Stream<Item = Result<Vec<DeployApp>, anyhow::Error>> + Send + Sync>>`) that required a `Sync` that was not available in `wasm32`.

### DIFF
--- a/lib/backend-api/src/client.rs
+++ b/lib/backend-api/src/client.rs
@@ -53,12 +53,12 @@ impl WasmerClient {
     }
 
     pub fn new(graphql_endpoint: Url, user_agent: &str) -> Result<Self, anyhow::Error> {
-        #[cfg(all(target_family = "wasm", not(target = "wasm32-wasmer-wasi")))]
+        #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
         let client = reqwest::Client::builder()
             .build()
             .context("could not construct http client")?;
 
-        #[cfg(any(not(target_family = "wasm"), target = "wasm32-wasmer-wasi"))]
+        #[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
         let client = reqwest::Client::builder()
             .connect_timeout(Duration::from_secs(10))
             .timeout(Duration::from_secs(90))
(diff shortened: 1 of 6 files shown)


lib/backend-api/src/client.rs
```diff
@@ -53,11 +53,18 @@ impl WasmerClient {
     }
 
     pub fn new(graphql_endpoint: Url, user_agent: &str) -> Result<Self, anyhow::Error> {
+        #[cfg(target_family = "wasm")]
```
theduke, line 56: Timeout is available when compiling to `wasix`, which might be useful for certain things, so let's be more specific here and limit to the specific `-unknown`  target.

lib/backend-api/src/query.rs
```diff
@@ -1351,10 +1347,17 @@ fn get_app_logs(
...
                         /*
-                            TODO: the resolution of watch should be configurable
-                            TODO: should this be async?
-                        */
+                         * [TODO]: The resolution here should be configurable.
+                         */
+
+                        // No tokio::time::sleep on wasm32 as of now.
+                        // Probably not ideal?
+                        #[cfg(target_family = "wasm")]
```
theduke, line 1355: As above.

lib/backend-api/src/client.rs
xdoardo: [5dcc9f8](https://github.com/wasmerio/wasmer/pull/4920/commits/5dcc9f8229a53b065ff1a440ccb382fecbc90bb3)

lib/backend-api/src/query.rs
xdoardo: [5dcc9f8](https://github.com/wasmerio/wasmer/pull/4920/commits/5dcc9f8229a53b065ff1a440ccb382fecbc90bb3)

theduke (approved):

PR #4925, opened 2024-07-09
Fix stack_low detection when data_end is above stack_pointer and stac…

Arshia001:
…k_lower is missing

This PR removes a false warning, but the behavior should otherwise be identical to what it was before, although accidentally.

### DIFF
--- a/lib/wasix/src/state/func_env.rs
+++ b/lib/wasix/src/state/func_env.rs
@@ -191,10 +191,17 @@ impl WasiFunctionEnv {
                     _ => 0,
                 }
             } else if let Some(data_end) = data_end {
-                match data_end.get(store) {
+                let data_end = match data_end.get(store) {
                     wasmer::Value::I32(a) => a as u64,
                     wasmer::Value::I64(a) => a as u64,
                     _ => 0,
+                };
+                // It's possible for the data section to be above the stack, we check for that here and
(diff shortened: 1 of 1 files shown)


lib/wasix/src/state/func_env.rs
```diff
@@ -191,10 +191,17 @@ impl WasiFunctionEnv {
...
             } else if let Some(data_end) = data_end {
-                match data_end.get(store) {
+                let data_end = match data_end.get(store) {
                     wasmer::Value::I32(a) => a as u64,
                     wasmer::Value::I64(a) => a as u64,
                     _ => 0,
+                };
+                // It's possible for the data section to be above the stack, we check for that here and
+                // if it is, we'll assume the stack starts at address 0
+                if data_end >= stack_base {
```
theduke, line 201: Are we sure the stack actually starts at 0?
We check the LLVM configuration in the Rust targets.

Arshia001: I'm fairly confident that that should be the case, considering:
* The top of the stack falls at directly 2^20 (e.g. 1MB) by default.
* I don't know of any other memory regions that would live below the stack?

theduke (approved):
Approved, pending a manual test run of a Rust binary with journaling.

theduke (approved):

PR #4928, opened 2024-07-13
Prevent unnecessary panics when compiling on x86_64 CPUs that dont support AVX or SSE4.2

Firaenix:
# Description
My company, Bitping, operates a distributed network of user devices to perform network tests and other types of work using WASM binaries. We use wasmer as the interpreter for these binaries and recently after an update one of our node operators with an AMD Turion II processor reported the following error at start up, which caused our application to fail to start.

image

This PR attempts to solve the panic by returning a Result instead if compilation fails on older x86_64 CPUs that dont support AVX or SSE4.2.

I've successfully compiled the package locally as well as run the compiler_singlepass tests, which all pass.

### DIFF
--- a/lib/compiler-singlepass/src/compiler.rs
+++ b/lib/compiler-singlepass/src/compiler.rs
@@ -211,8 +211,8 @@ impl Compiler for SinglepassCompiler {
             .values()
             .collect::<Vec<_>>()
             .into_par_iter_if_rayon()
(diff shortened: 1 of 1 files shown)


Firaenix:
In regards to the cargo-deny CI step, do you want me to upgrade that crate? It's kind of outside of the scope of my fix

syrusakbary (approved):

PR #4929, opened 2024-07-15
Fix indirect call to dynamic imported function

### LINKED ISSUE #4565: Wasmer can't run Doom
I understand that this is a somewhat nebulous, perhaps frivolous issue. Here it is anyway.

Wasmer fails to run a Doom Wasm port. [This article](https://github.com/diekmann/wasm-fizzbuzz/blob/main/doom/README.md) by Cornelius Diekmann documents the process of porting Doom to WebAssembly and results in [this demo](https://diekmann.github.io/wasm-fizzbuzz/doom/). In [my own explorations of the subject](https://github.com/ashtonmeuser/godot-wasm-doom), I found that the Doom Wasm module runs as expected in the browser (tested on Chrome, Safari, Firefox) and Wasmtime, but fails using Wasmer.

The only clue as to what is going on is the following trap message on invocation of the `main` export function.

> out of bounds memory access

Wasmer version: 4.2.0

### Steps to reproduce

Descriptions for recreating the issue can be found in [my write up](https://github.com/ashtonmeuser/godot-wasm-doom). Alternatively, the module can be initialized with relatively few imports. I've attached the Wasm module in question.

[doom.wasm.zip](https://github.com/wasmerio/wasmer/files/14950274/doom.wasm.zip)

### Expected behaviour

Invocation of the `main` export function should result in the following expected logs.

> Hello, World, from JS Console! Answer=42 (101010 in binary)
Hello, world from rust! 🦀🦀🦀 (println! working)
Starting D_DoomMain
Triggering a printf
Doom's screen is 320x200
mallocing 12 bytes at 7078176

### Actual behavior

Invocation of the `main` export function results in the following logs before the function invocation fails.

> Hello, World, from JS Console! Answer=42 (101010 in binary)
Hello, world from rust! 🦀🦀🦀 (println! working)

Arshia001:
Fixes #4565.

The solution is to update the func_ptr of the anyfunc with the generated trampoline address during the resolve step. I'm not sure this is the best place for this code though; careful review is required.

### DIFF
--- a/lib/compiler/src/engine/resolver.rs
+++ b/lib/compiler/src/engine/resolver.rs
@@ -113,6 +113,11 @@ pub fn resolve_imports(
                         let index = FunctionIndex::new(function_imports.len());
                         let ptr = finished_dynamic_function_trampolines[index].0
                             as *mut VMFunctionBody as _;
+                        // The logic is currently handling the "resolution" of dynamic imported functions at instantiation time.
+                        // However, ideally it should be done even before then, as you may have dynamic imported functions that
+                        // are linked at runtime and not instantiation time. And those will not work properly with the current logic.
+                        // Ideally, this logic should be done directly in the `wasmer-vm` crate.
+                        // TODO (@syrusakbary): Get rid of `VMFunctionKind`
                         unsafe { f.anyfunc.as_ptr().as_mut() }.func_ptr = ptr;
                         ptr
                     }
--- a/lib/compiler/src/engine/resolver.rs
+++ b/lib/compiler/src/engine/resolver.rs
@@ -64,7 +64,7 @@ fn get_runtime_size(context: &StoreObjects, extern_: &VMExtern) -> Option<u32> {
 pub fn resolve_imports(
     module: &ModuleInfo,
     imports: &[VMExtern],
-    context: &StoreObjects,
+    context: &mut StoreObjects,
     finished_dynamic_function_trampolines: &BoxedSlice<FunctionIndex, FunctionBodyPtr>,
     memory_styles: &PrimaryMap<MemoryIndex, MemoryStyle>,
     _table_styles: &PrimaryMap<TableIndex, TableStyle>,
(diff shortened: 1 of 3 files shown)


xdoardo (approved):

syrusakbary (approved):

tests/compilers/imports.rs
```diff
@@ -14,15 +14,18 @@ use wasmer::*;
 
 fn get_module(store: &Store) -> Result<Module> {
     let wat = r#"
-        (import "host" "0" (func))
+        (type (func))
+        (import "host" "0" (func $host_func_0 (type 0)))
```
theduke, line 18: Will this ever actually be executed by a test?
Apparently not in the same file.

Would be good to have a dedicated test for this exact issue, which also executes the code.

syrusakbary: Oh wait... I assumed the module would instantiate

Arshia001: https://github.com/wasmerio/wasmer/blob/76b5870702f6535369cb09f56245eb41b726cc2b/tests/compilers/imports.rs#L54

This executes the module afaiu

lib/compiler/src/engine/resolver.rs
```diff
@@ -104,14 +104,17 @@ pub fn resolve_imports(
...
-                let f = handle.get(context);
+                let f = handle.get_mut(context);
                 let address = match f.kind {
                     VMFunctionKind::Dynamic => {
                         // If this is a dynamic imported function,
                         // the address of the function is the address of the
                         // reverse trampoline.
                         let index = FunctionIndex::new(function_imports.len());
-                        finished_dynamic_function_trampolines[index].0 as *mut VMFunctionBody as _
+                        let ptr = finished_dynamic_function_trampolines[index].0
```
theduke, line 114: This seems a somewhat hacky place to put this code, but if it is hard to put somewhere else, there shoud be a comment explaining why this is here.

syrusakbary: We can put this comment:

```
/// TODO (@syrusakbary): We need to do this properly, handling this well in the VM while getting rid of `VMFunctionKind`
```

theduke: And a bit more explanation as well. ;)

Arshia001: I'll let Syrus come up with the comment, as I don't quite understand the situation well enough.

tests/compilers/imports.rs
theduke: Ideally would also use a host function that returns an integer so the test makes sure it actually is executed correctly.

lib/compiler/src/engine/resolver.rs
Arshia001: @syrusakbary LMK what you want me to put in there, so I can add commits in both places (this and the test) and we can merge this in afterwards.

syrusakbary:
```
/// The logic is currently handling the "resolution" of dynamic imported functions at instantiation time.
/// However, ideally it should be done even before then, as you may have dynamic imported functions that are linked at runtime and not instantiation time. And those will not work properly with the current logic.
/// Ideally, this logic should be done directly in the `wasmer-vm` crate.
/// TODO (@syrusakbary): Get rid of `VMFunctionKind`
```

PR #4930, opened 2024-07-15
CLI: Add support for Secrets

### LINKED ISSUE #4864: Secrets: Implement a `secrets` subcommand in the `app` command
Needed sub-subcommands: `list`, `get`, `delete` and `set` <br>(full command: `wasmer app secrets [get|list|…]`)

xdoardo:
(Closes #4864). 
This PR adds subcommands to manage secrets in the `app` command: 
```
wasmer app secrets (create | update | list | reveal | delete)
```


syrusakbary (approved):

lib/backend-api/src/query.rs
```diff
@@ -11,19 +11,198 @@ use url::Url;
...
+                Some(edge) => edge,
+                None => continue,
+            };
+            let version = match edge.node {
+                Some(item) => item,
+                None => continue,
+            };
+
+            // Sanity check to avoid duplication.
+            if all_versions.iter().any(|v| v.id == version.id) == false {
```
theduke, line 171: As above.

lib/cli/src/commands/app/secrets/create.rs
```diff
@@ -0,0 +1,271 @@
...
+use wasmer_api::WasmerClient;
+
+/// Create a new secret related to an Edge app.
+#[derive(clap::Parser, Debug)]
+pub struct CmdAppSecretsCreate {
+    /// The path to the directory where the config file for the application will be written to.
+    #[clap(long = "app-dir", conflicts_with = "app")]
+    pub app_dir_path: Option<PathBuf>,
+
+    /// The name of the secret to create.
```
theduke, line 28: nit: I always like to put the positional args at the bottom so they can be identified easily.

lib/cli/src/commands/app/secrets/create.rs
```diff
@@ -0,0 +1,271 @@
...
+        } else {
+            let theme = ColorfulTheme::default();
+            Ok(dialoguer::Input::with_theme(&theme)
+                .with_prompt("Enter the value of the secret")
+                .interact_text()?)
+        }
+    }
+
+    async fn get_app_id(&self, client: &WasmerClient) -> anyhow::Result<String> {
+        if let Some(app_id) = &self.app_id.app {
```
theduke, line 92: Should re-use the logic from `AppIdent(Opts)`

lib/cli/src/commands/app/secrets/create.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// returns a list of secrets that must be upserted.
+    async fn filter_secrets(
+        &self,
+        client: &WasmerClient,
+        app_id: &str,
+        secrets: Vec<Secret>,
+    ) -> anyhow::Result<Vec<Secret>> {
+        let names = secrets.iter().map(|s| &s.name);
+        let app_secrets =
+            wasmer_api::query::get_all_app_secrets_filtered(client, app_id, names).await?;
```
theduke, line 169: Should this kind of logic really be in the CLI?
That makes it hard to update.

This also opens the door to race conditions, so it's not actually a guarantee that create actually creates.


I feel like the backend should handle this.
Eg with a "createNew" argument to the mutation.

lib/cli/src/commands/app/secrets/create.rs
```diff
@@ -0,0 +1,271 @@
...
+            "Backend did not return any payload to confirm the successful creation of the secret!",
+        )?;
+
+        if !res.success {
+            anyhow::bail!("Secret creation failed!")
+        } else {
+            if !self.quiet {
+                eprintln!("Succesfully created secret(s):");
+                for secret in secrets {
+                    eprintln!("{}", secret.name.bold());
```
theduke, line 232: Should show a hint that secrets are not deployed yet, and to re-deploy the app to apply the changes.

lib/cli/src/commands/app/secrets/delete.rs
```diff
@@ -0,0 +1,232 @@
...
+        } else {
+            let theme = ColorfulTheme::default();
+            Ok(dialoguer::Input::with_theme(&theme)
+                .with_prompt("Enter the name of the secret")
+                .interact_text()?)
+        }
+    }
+
+    async fn get_app_id(&self, client: &WasmerClient) -> anyhow::Result<String> {
+        if let Some(app_id) = &self.app_id.app {
```
theduke, line 85: Should re-use common logic.

lib/cli/src/commands/app/secrets/list.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// Do not prompt for user input.
+    #[clap(long, default_value_t = !std::io::stdin().is_terminal())]
+    pub non_interactive: bool,
+
+    #[clap(flatten)]
+    pub fmt: ListFormatOpts,
+}
+
+impl CmdAppSecretsList {
+    async fn get_app_id(&self, client: &WasmerClient) -> anyhow::Result<String> {
```
theduke, line 46: Reuse common logic.

lib/cli/src/commands/app/secrets/update.rs
```diff
@@ -0,0 +1,266 @@
...
+        } else {
+            let theme = ColorfulTheme::default();
+            Ok(dialoguer::Input::with_theme(&theme)
+                .with_prompt("Enter the value of the secret:")
+                .interact_text()?)
+        }
+    }
+
+    async fn get_app_id(&self, client: &WasmerClient) -> anyhow::Result<String> {
+        if let Some(app_id) = &self.app_id.app {
```
theduke, line 97: Reuse common logic.

lib/cli/src/commands/app/secrets/update.rs
```diff
@@ -0,0 +1,266 @@
...
+        )?;
+
+        if !res.success {
+            anyhow::bail!("Secret creation failed!")
+        } else {
+            if !self.quiet {
+                eprintln!("Succesfully updated secret(s):");
+                for secret in secrets {
+                    eprintln!("{}", secret.name.bold());
+                }
```
theduke, line 228: As in `create`, should show a hint that it's not deployed yet and user must re-deploy to apply.

lib/cli/src/commands/app/secrets/create.rs
xdoardo: This is something we thought about. The backend exposes an `upsert` mutation, abd adding a `createNew` would make it semantically inconsistent, in my opinion. (cc: @ayys)

lib/backend-api/src/query.rs
```diff
@@ -11,19 +11,198 @@ use url::Url;
...
```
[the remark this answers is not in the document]
xdoardo: Removed.

lib/cli/src/commands/app/secrets/create.rs
xdoardo: Reordered.

lib/cli/src/commands/app/secrets/create.rs
xdoardo: Added a message after printing that secrets were created/updated: 
```
Info: In order for secrets to appear in your app, re-deploy it.
```

lib/cli/src/commands/app/secrets/update.rs
xdoardo: As above.

lib/cli/src/commands/app/secrets/create.rs
xdoardo: I chose not to use a function from `AppIdent*Flag*` as that part of the code was pretty tied into the logic of secrets and searching to fallback to user input. What I mean is that a function `AppIdentFlag::resolve() -> Result<AppIdent>` does not suit this piece of code, and `AppIdentFlag::search_ident(maybe_search_path) -> Result<Option<AppIdent>>` does not suit the other either. So, a utility function - that could, perhaps, use more methods from `app::utils` encloses the behaviour shared between the `secrets` commands to get an app identifier.

theduke (approved):

lib/cli/src/commands/app/secrets/create.rs
theduke: Created a followup issue to figure this out, we can merge without it.

PR #4933, opened 2024-07-16
CLI: Manually exit upon SIGINT reception

xdoardo:
As per title. The CLI returns with a `130` (128 + SIGINT) exit code upon receiving a SIGINT (CTRL+C) signal.


lib/cli/src/commands/mod.rs
```diff
@@ -84,6 +84,7 @@ pub(crate) trait AsyncCliCommand: Send + Sync {
                     _ = tokio::signal::ctrl_c() => {
                         let term = console::Term::stdout();
                         let _ = term.show_cursor();
+                        std::process::exit(130);
```
syrusakbary, line 87: This only works in Unix. This code will not work in Windows. Also, can we do `128 + SIGINT` instead of `130` ?

syrusakbary (approved):

PR #4938, opened 2024-07-17
fix(cli/package/tag): Better messages when tagging an existing package

### LINKED ISSUE #4707: wasmer deploy fails with named local package and rejected package bump
### Describe the bug

Can't deploy with local named package without a version bump.

### Steps to reproduce

wasmer.toml:
```
[package]
name = "wasmer-tests/static-site"
version = "0.1.0"
description = "wasmer-tests/static-site website"

[dependencies]
"wasmer/static-web-server" = "^1"

[fs]
public = "public"
```

app.yaml:
```
name: xxx
owner: xxx
package: .
kind: wasmer.io/App.v0
```

* `wasmer deploy`
* Do you want to bump the package? => SELECT NO

=> ERROR
```
 endpoint=https://registry.wasmer.io/graphql
error: GraphQL query failed
╰─▶ 1: GraphQL API failure: This version of the package `wasmer-tests/static-site@0.1.0` already exists.
```

### Expected behavior

Command does not try to re-publish, but just uses the existing version.

### Actual behavior

Tries to republish, fails.

### LINKED ISSUE #4801: CLI: "deploy/publish" commands should not ask to bump version if tag already exists
Currently, when running `wasmer deploy` with an unchanged *named* package, the CLI will:

* build the package locally
* Upload it if it does not exist in the backend
* Ask to bump the version ( => tag new version)
* Deploy

If the package already has a version matching the package hash, the "ask to bump" step should be skipped, and the matching version should be used. 

There should potentially be a warning/confirm step if the version number in wasmer.toml does not match the auto-determined version.

xdoardo:
(Closes #4801, closes #4707, fixes RUN-276)

This PR adds a better message shown to users in the case they try to tag two different releases of the same package with the same version: 

Screenshot 2024-07-17 at 13 48 22

### DIFF
--- a/lib/cli/src/commands/package/tag.rs
+++ b/lib/cli/src/commands/package/tag.rs
@@ -12,7 +12,6 @@ use is_terminal::IsTerminal;
 use std::{
     path::{Path, PathBuf},
     str::FromStr,
-    sync::OnceLock,
 };
 use wasmer_api::{types::PackageVersionWithPackage, WasmerClient};
 use wasmer_config::package::{
@@ -20,9 +19,6 @@ use wasmer_config::package::{
 };
 
 use super::PublishWait;
-
-/// The package related to the NamedPackageId the user chose to tag.
-static MAYBE_USER_PACKAGE: OnceLock<Option<PackageVersionWithPackage>> = OnceLock::new();
 
 /// Tag an existing package.
 #[derive(Debug, clap::Parser)]
@@ -92,6 +88,10 @@ pub struct PackageTag {
             value_enum
         )]
     pub wait: PublishWait,
+
+    #[clap(skip)]
+    /// The package related to the NamedPackageId the user chose to tag.
+    pub(crate) maybe_user_package: Option<PackageVersionWithPackage>,
 }
 
 impl PackageTag {
@@ -377,7 +377,7 @@ impl PackageTag {
     }
 
     async fn get_version(
-        &self,
+        &mut self,
         client: &WasmerClient,
         manifest: Option<&Manifest>,
         manifest_path: Option<&Path>,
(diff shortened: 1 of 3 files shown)


lib/cli/src/commands/package/tag.rs
```diff
@@ -12,14 +12,18 @@ use is_terminal::IsTerminal;
...
-use wasmer_api::WasmerClient;
+use wasmer_api::{types::PackageVersionWithPackage, WasmerClient};
 use wasmer_config::package::{
     Manifest, NamedPackageId, NamedPackageIdent, PackageBuilder, PackageHash, PackageIdent,
 };
 
 use super::PublishWait;
 
+/// The package related to the NamedPackageId the user chose to tag.
+static MAYBE_USER_PACKAGE: OnceLock<Option<PackageVersionWithPackage>> = OnceLock::new();
```
theduke, line 25: A shared global for this is quite nasty... 

Why is this neccessary?

xdoardo: In order to avoid repeating the same query to the backend twice while keeping function types a bit cleaner. Otherwise, the `get_version` function should return a `Result<(Option<PackageVersionWithPackage>, semver::Version)>` lifting up to `synthesize_id`. To me - this is of course subjective - is less clean than a `OnceLock`.

lib/cli/src/commands/package/tag.rs
```diff
@@ -433,12 +459,34 @@ impl PackageTag {
...
                     new_version.patch += 1;
-                    if Confirm::with_theme(&theme)
+                    if must_bump {
+                        eprintln!("{}: Registry already has version {user_version} of {full_pkg_name}, but with different contents.", "Warn".bold().yellow());
+                        eprintln!(
+                            "{}: Not bumping the version will make this action fail.",
+                            "Warn".bold().yellow()
+                        );
+                        let res = Confirm::with_theme(&theme)
+                            .with_prompt(format!("Continue ({user_version} -> {new_version})?"))
```
syrusakbary, line 469: `Would you like to update the version from {user_version} to {new_version}` ?

syrusakbary (approved):

theduke (approved):

PR #4946, opened 2024-07-21
Fixes a panic caused on the recv_from path

john-sharratt:
This is caused because the slices do not line up - its not been detected till now because UDP sockets are vary used


theduke (approved):
Can we get a small regression test for this?

Issue #4918, opened 2024-07-05
If you login, then attempt to login again it panics


john-sharratt:
```bash
john@AlienWorld:~/prog/wasmer/lib$ wasmer login
Opening auth link in your default browser: https://wasmer.io/auth/cli?nonce_id=...-c
Waiting for session... Done!
✔ Login for Wasmer user "tokera" saved

john@AlienWorld:~/prog/wasmer/lib$ wasmer login
thread 'main' panicked at /home/john/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.37.0/src/runtime/blocking/shutdown.rs:51:21:
Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
john@AlienWorld:~/prog/wasmer/lib
```

It would also be handy if one could `logout`

PR #4955, opened 2024-07-24
Move `WasmerConfig` out of the `registry_api` crate and introduce a `logout` subcommand

### LINKED ISSUE #4951: CLI: Create `logout` command

xdoardo:
(Closes #4951 and - hopefully - others login-related issues). 

This PR does two things: 
1. Moves the CLI crate a bit further away from the double registry dependency (`backend-api` vs `registry-api`). In order to achieve this, it moves the `WasmerConfig` type out of the `registry-api` crate and into the `cli` crate. In turn, this allows to merge the `ApiOpts` and `WasmerEnv` types.  Rationale for doing this now: #4918  panics due to the runtime-inside-a-runtime quirk in turn due to the use of registry-api.
2.  Adds a `logout` command. 

This PR is preventive, and should initially be used just to test the behaviour of the CLI after the induced changes.
In order to be sure that the issues #4918 and #4918 are solved by this PR, a windows-native test run must be executed.


theduke (approved):
LGTM

PR #4965, opened 2024-07-26
CLI: Try to fetch owner from `app_id` if necessary while deploying

### LINKED ISSUE #4957: CLI: wasmer deploy: existing app not detected properly when no namespace specified
When running `wasmer deploy` with the below app config, the command prompts for chosing the owner, even though the app already exists and has an owner. An app id is present in the config.

Correct behaviour: do not prompt for an owner, just detect that the app already exists and deploy.

```yaml
kind: wasmer.io/App.v0
name: wasmer-tests-js-worker-basic
app_id: da_zRLmILtvUyj2
package: wasmer-tests/js-worker-basic
cli_args:
  - /src/index.js
  ```

xdoardo:
(Closes #4957 , fixes RUN-372)

As per title: if no owner is supplied via CLI and there's no `owner` in the `app.yaml`, in the case that an `app_id` is present, the `deploy` command tries to fetch the app owner from the backend using the `app_id`.

### DIFF
--- a/lib/cli/src/commands/app/deploy.rs
+++ b/lib/cli/src/commands/app/deploy.rs
@@ -180,6 +180,17 @@ impl CmdAppDeploy {
             return Ok((owner.clone(), r_ret));
         }
 
+        if let Some(app_id) = app.get("app_id") {
+            if let Some(app_id) = app_id.as_str() {
+                // Try to get the owner from the app_id. 
+                //
+                // In this case, don't edit the app config.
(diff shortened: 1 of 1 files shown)


syrusakbary:
Note: there can be no owner, or no owner and no name, or no name.

In general, if we have an app id, and there's a mismatch between the corresponding owner and name, then we need to update them both automatically and show that app.yaml has been updated in the stdout as information for the user.

theduke (approved):

PR #4974, opened 2024-07-30
Allow WebCVolumeFS to be backed by a host directory

Arshia001:


Arshia001:
Closing as this is a bad idea, will open another PR later with a better solution.

PR #4975, opened 2024-07-30
CLI: Ask users to allow networking capabilities if not set

### LINKED ISSUE #4953: CLI (wasmer run): Cache permission flags given by the user
bonus points: Detect needed permissions upon running a package and show a warning if a permission flag was missing (e.g. no `--net` for an http server)

xdoardo:
(Closes #4953, fixes RUN-371)

As per title. This patch introduces a mechanism that allows users to recover from not permitted networking accesses with a prompt: 

Screenshot 2024-07-30 at 14 18 47

In addition, as implied by the prompt, this patch adds a mechanism to cache the users' choice in response to the prompt above.

### DIFF
--- a/lib/cli/src/commands/run/capabilities/mod.rs
+++ b/lib/cli/src/commands/run/capabilities/mod.rs
@@ -1,0 +1,3 @@
+/// A custom implementation of the [`virtual_net::VirtualNetwork`] that asks users if they want to
+/// use networking features at runtime.
+pub(crate) mod net;
--- a/lib/cli/src/commands/run/capabilities/mod.rs
+++ b/lib/cli/src/commands/run/capabilities/mod.rs
@@ -1,3 +1,12 @@
 /// A custom implementation of the [`virtual_net::VirtualNetwork`] that asks users if they want to
 /// use networking features at runtime.
 pub(crate) mod net;
+
+/// The default name of the directory to store cached capabilities for packages.
+pub(crate) const DEFAULT_WASMER_PKG_CAPABILITY_CACHE_DIR: &str = "pkg_capabilities";
+
+/// A struct representing cached capabilities for a specific package.
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub(crate) struct PkgCapabilityCache {
+    pub enable_networking: bool,
+}
--- a/lib/cli/src/commands/run/capabilities/net.rs
+++ b/lib/cli/src/commands/run/capabilities/net.rs
@@ -1,9 +1,12 @@
 use std::{
     net::{IpAddr, SocketAddr},
+    path::PathBuf,
     sync::OnceLock,
     time::Duration,
 };
 
+use super::{super::PackageSource, PkgCapabilityCache};
+use anyhow::Context;
 use colored::Colorize;
 use dialoguer::theme::ColorfulTheme;
 use virtual_net::{
@@ -16,34 +19,20 @@ use virtual_net::{
 /// use networking features at runtime.
 #[derive(Debug, Clone)]
 pub(crate) struct AskingNetworking {
+    pkg_cache_path: PathBuf,
     enable: OnceLock<Result<bool>>,
     capable: DynVirtualNetworking,
     unsupported: DynVirtualNetworking,
 }
 
-fn ask_user(fn_name: &str) -> Result<bool> {
-    let theme = ColorfulTheme::default();
-
-    println!("Networking needs to be enabled to call function '{fn_name}'.");
-    if dialoguer::Confirm::with_theme(&theme)
-        .with_prompt("Enable networking?")
-        .interact()
-        .map_err(|_| NetworkError::UnknownError)?
-    {
-        eprintln!(
-            "{}: to enable networking by default, use the `--net` flag",
-            "Info".bold()
-        );
-        Ok(true)
-    } else {
-        Ok(false)
-    }
-}
-
 macro_rules! call {
     ($self: expr, $fn_name: ident, $( $arg: expr ),* ) => {
 
-        let enable_networking = $self.enable.get_or_init(|| ask_user(stringify!($fn_name))).map_err(|_| NetworkError::UnknownError)?;
+        let enable_networking = $self.enable.get_or_init(|| $self.ask_user(stringify!($fn_name)))
+            .map_err(|e| {
+                tracing::error!("{e}");
+                NetworkError::UnknownError
+            })?;
 
         if enable_networking {
             return $self.capable.$fn_name( $( $arg ),* ).await;
@@ -54,7 +43,10 @@ macro_rules! call {
 
     ($self: expr, $fn_name: ident) => {
 
-        let enable_networking = $self.enable.get_or_init(|| ask_user(stringify!($fn_name))).map_err(|_| NetworkError::UnknownError)?;
+        let enable_networking = $self.enable.get_or_init(|| $self.ask_user(stringify!($fn_name))).map_err(|e| {
+                tracing::error!("{e}");
+                NetworkError::UnknownError
+            })?;
 
         if enable_networking {
             return $self.capable.$fn_name().await;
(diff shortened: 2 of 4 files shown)


syrusakbary:
I believe it will be better for the user to input a single char (`y/n/a`, as yes/no/always) ). Rather than up/down selectable.

It's way more common for this kind of input

lib/cli/src/commands/run/capabilities/net.rs
```diff
@@ -73,27 +112,21 @@ impl AskingNetworking {
...
-            .items(&[
-                "no",
-                "yes",
-                "always (will remember this option for future runs of the same package)",
-            ])
-            .default(2)
+        match dialoguer::Input::with_theme(&theme)
+            .with_prompt(format!(
+                "Would you like to allow networking for this package? {}{}",
+                "".bold(),
```
syrusakbary, line 118: Empty string bold?

xdoardo: Yes. This is due to the fact that the string in `with_prompt` will be displayed in bold from `dialoguer`. Since I think it looks better to have `(yes/no/always)` as dimmed but not bold, the empty `.bold()` string is used to trigger the ANSI reset escape.

syrusakbary:
This is great, I just tried locally. Can we have the following? Have the letter `a` prefilled in the input (that way is one less keystroke for the user)

lib/cli/src/commands/run/capabilities/net.rs
syrusakbary: Got it

lib/cli/src/commands/run/mod.rs
```diff
@@ -140,9 +144,75 @@ impl Run {
...
+            .host_str()
+            .unwrap_or("unknown_registry")
+            .replace('.', "_");
+
+        // We don't have the bytes of the module yet, but we still want to have the
+        // package-capabilities cache be as close to an actual identifier as possible.
+        let package_cache_path = match &self.input {
+            PackageSource::File(f) => {
+                let full_path = f.canonicalize()?.to_path_buf();
+                let mut h = DefaultHasher::new();
```
syrusakbary, line 159: This hasher might not be consistent over rust versions. We need to use a consistent one

syrusakbary: I'd probably recommend using something like xxhash or other fast hasher

syrusakbary: Or a cryptographically secure hasher (like blake3 or sha)

syrusakbary (changes requested):
I tried this and it doesn't work. You can't delete the "a" and can't just press enter either (need to add another "a")

lib/cli/src/commands/run/capabilities/net.rs
```diff
@@ -0,0 +1,306 @@
...
+        println!("The current package is requesting networking access.");
+        println!("Run the package with `--net` flag to bypass the prompt.");
+        match dialoguer::Input::with_theme(&theme)
+            .with_prompt(format!(
+                "Would you like to allow networking for this package? {}{}",
+                "".bold(),
+                "(yes/no/always)".dimmed()
+            ))
+            .with_initial_text("a")
+            .interact()
```
syrusakbary, line 122: I bet this should be:

```suggestion
            .interact_text()
```

Or perhaps we need to do

```rust
            .default("a") // Instead of with_initial_text
            .interact()
```

lib/cli/src/commands/run/mod.rs
xdoardo: [9f9c47a](https://github.com/wasmerio/wasmer/pull/4975/commits/9f9c47a2f75643b1eca6ce8a22ba3ead2d63553f)

lib/cli/src/commands/run/capabilities/net.rs
xdoardo: [e0859cc](https://github.com/wasmerio/wasmer/pull/4975/commits/e0859cc71ee018687e2565c7c4b5b5e4b9b0c6d7)

lib/cli/src/commands/run/capabilities/net.rs
```diff
@@ -0,0 +1,306 @@
...
+impl FromStr for UserSelection {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        let s = s.trim();
+        if s.is_empty() {
+            anyhow::bail!("No input!")
+        }
+
+        let c: char = s.chars().next().unwrap().to_ascii_lowercase();
```
theduke, line 87: Super tiny nit: this stuff is nicer to write without unwraps:

```
let Some(c) = s..trim().chars().next() else {
  bail!(...)
}
```

Although in this case you could just match on the full string as well?

lib/cli/src/commands/run/mod.rs
```diff
@@ -140,9 +143,81 @@ impl Run {
...
+
+                let mut hash = twox_hash::XxHash64::default();
+                full_path.hash(&mut hash);
+                metadata.hash(&mut hash);
+
+                format!("path_{}.json", hash.finish())
+            }
+            PackageSource::Package(p) => match p {
+                PackageSpecifier::Ident(id) => {
+                    format!("id_{id}.json")
```
theduke, line 182: Should we only use the namespace + name here, and re-use across versions?

Would be sensible and more convenient IMO.

lib/cli/src/commands/run/mod.rs
```diff
@@ -140,9 +143,81 @@ impl Run {
...
+                }
+            },
+        };
+
+        let runtime = self.wasi.prepare_runtime(
+            engine,
+            &self.env,
+            &self
+                .env
+                .cache_dir()
```
theduke, line 214: nit: Can we move this path construction to a dedicated helper method somewhere?

lib/cli/src/commands/run/wasi.rs
```diff
@@ -548,10 +551,33 @@ impl Wasi {
         let tokio_task_manager = Arc::new(TokioTaskManager::new(rt_or_handle.into()));
         let mut rt = PluggableRuntime::new(tokio_task_manager.clone());
 
-        if self.networking {
+        let has_networking = self.networking
+            || {
+                if pkg_cache_path.is_file() {
```
theduke, line 556: As above, can we have this in a helper method?

lib/cli/src/commands/run/wasi.rs
```diff
@@ -548,10 +551,33 @@ impl Wasi {
...
+                    let raw = std::fs::read_to_string(pkg_cache_path)?;
+                    if let Ok(pkg_capability_cache) =
+                        serde_json::from_str::<PkgCapabilityCache>(&raw)
+                    {
+                        tracing::info!("Cache hit for user-set capabilities at path {pkg_cache_path:?}: {pkg_capability_cache:?}");
```
theduke, line 561: nit: maybe `debug` is enough here?

lib/cli/src/commands/run/capabilities/net.rs
xdoardo: > Although in this case you could just match on the full string as well?

I think we want to consider correct many combinations (`y`/`Y`/`yes`/`YES`...). Of course, an input such as `yna` resolves to `yes`

lib/cli/src/commands/run/capabilities/mod.rs
```diff
@@ -0,0 +1,12 @@
+/// A custom implementation of the [`virtual_net::VirtualNetwork`] that asks users if they want to
+/// use networking features at runtime.
+pub(crate) mod net;
+
+/// The default name of the directory to store cached capabilities for packages.
+pub(crate) const DEFAULT_WASMER_PKG_CAPABILITY_CACHE_DIR: &str = "pkg_capabilities";
+
+/// A struct representing cached capabilities for a specific package.
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub(crate) struct PkgCapabilityCache {
```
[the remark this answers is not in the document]
xdoardo: I think these two are kind of orthogonal. Maybe a name different from `PkgCapabilityCache`, something like `RunFlags` would make it more evident?

lib/cli/src/commands/run/wasi.rs
xdoardo: [5b8321d](https://github.com/wasmerio/wasmer/pull/4975/commits/5b8321d098b02353e0a960f805851a1456a29338)

lib/cli/src/commands/run/wasi.rs
xdoardo: [5b8321d](https://github.com/wasmerio/wasmer/pull/4975/commits/5b8321d098b02353e0a960f805851a1456a29338)

lib/cli/src/commands/run/mod.rs
xdoardo: [5b8321d](https://github.com/wasmerio/wasmer/pull/4975/commits/5b8321d098b02353e0a960f805851a1456a29338)

lib/cli/src/commands/run/mod.rs
xdoardo: [5b8321d](https://github.com/wasmerio/wasmer/pull/4975/commits/5b8321d098b02353e0a960f805851a1456a29338)

Cargo.toml
```diff
@@ -35,6 +35,7 @@ tokio = { version = "1", features = [
     "macros",
 ], optional = true }
 crossbeam-queue = "0.3.8"
+twox-hash = "1.6.3"
```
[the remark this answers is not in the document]
xdoardo: [5b8321d](https://github.com/wasmerio/wasmer/pull/4975/commits/5b8321d098b02353e0a960f805851a1456a29338)

syrusakbary (approved):

PR #4981, opened 2024-07-31
Allow WebCVolumeFS to be backed by a host directory - Take 2

Arshia001:


Arshia001:
Also a bad idea, closing in favor of #4982

PR #4982, opened 2024-07-31
Mount [fs] entries as additional mapped directories when running a lo…

Arshia001:
…cal package


theduke:
@Arshia001 there should probably be a test for this in the CLI integration tests?

Eg: start a static webserver , add a file ,fetch the file with an http client, make sure the file is as expected?

maminrayej (commented):
LGTM. Just needs a test that reads from, writes to, and creates file on the host filesystem.

maminrayej (approved):

PR #4986, opened 2024-08-05
Multiple fixes for the journal to fix bootstrapping

john-sharratt:


lib/journal/src/concrete/compacting.rs
```diff
@@ -31,7 +32,7 @@ impl From<Range<u64>> for MemoryRange {
 }
 
 #[derive(Debug, Clone, Copy, PartialOrd, Ord, PartialEq, Eq, Hash)]
-struct DescriptorLookup(u64);
+struct SubEventsLookup(u64);
```
theduke, line 35: Can we call this `EventId`? Would be clearer I think.

theduke, line 35: The name here was changed, but the type itself is still called `StateDscriptor`.
That's somewhat confusing.

Would this be better: `EventBatch` + `EventBatchId` ?

theduke (commented):
Can we get some high level comments on the major parts of the code in here so readers have an easier time following what's going on?

What are the types, what's the `events` field on StateDescriptor, , etc.

theduke (approved):

PR #4990, opened 2024-08-07
Added a patch for corrupt freed FD lists after replaying journals

john-sharratt:


lib/wasix/src/fs/mod.rs
```diff
@@ -566,6 +566,11 @@ impl WasiFs {
         }
     }
 
+    pub fn clear_free_fd_list(&self) {
```
theduke, line 569: Comment why this is needed.

Also, maybe call it  `unchecked_clear...` and add a `DANGEROUS` comment.

Arshia001 (approved):

theduke (approved):

PR #4992, opened 2024-08-07
fix(cli/WasmerEnv): strip `registry.` only if hostname starts with it

xdoardo:
Currently, if a user has a non-canonical registry URL set as default url - for example, `http://localhost:11/graphql` - the CLI will fail as it tries to strip the `registry.` part of the URL, which might not be there. This small patch applies that change only in the case that the hostname actually starts with `registry.`.


theduke (approved):

theduke:
Is it worth adding a test that checks a few different variants to prevent regressions?

PR #4994, opened 2024-08-08
Check if a user can deploy an app before deploying it

### LINKED ISSUE #4961: Improve deploy process
As a user, if I go to a repo with a set `app.yaml` for which I don't have permissions to push (eg: https://github.com/wasmer-examples/wordpress-wasmer-starter )

Basically, an app.yaml with either [owner & name] or [app_id] (or both). If I don't have permissions to deploy, it should prompt me if I want to deploy to a new app, and if I confirm.. then delete `owner, name and app_id` and trigger again the deploy process.

Example:

```bash
$ git clone https://github.com/wasmer-examples/wordpress-wasmer-starter && cd wordpress-wasmer-starter
$ wasmer deploy
It seems the app {name} ({owner}) doesn't exist, or you don't have permissions for it.
Do you want to deploy this as a new app? (y/n)
# Delete owner, name and app_id from app.yaml
# Trigger the deploy flow again
```

xdoardo:
(Closes #4961)

Currently, if a user `a` tries to deploy an app that in the config shows as owner the name `b` but `a` has no permissions to deploy on `b` the deploy fails. This PR adds a preventive check to make sure that the user has the correct permissions to deploy.  In case the user can't the CLI asks for a new owner and app name.

### DIFF
--- a/lib/cli/src/commands/app/deploy.rs
+++ b/lib/cli/src/commands/app/deploy.rs
@@ -284,12 +284,9 @@ impl AsyncCliCommand for CmdAppDeploy {
             .await?;
 
         if !wasmer_api::query::viewer_can_deploy_to_namespace(&client, &owner).await? {
-            eprintln!(
-                "Cannot deploy app to namespace {}, as the current user is not authorized.",
-                owner.bold()
-            );
+            eprintln!("It seems you don't have access to {}", owner.bold());
             if self.non_interactive {
-                anyhow::bail!("Please, check the app configuration or the current user with the `whoami` command!");
+                anyhow::bail!("Please, change the owner before deploying or check your current user with `{} whoami`.", std::env::args().next().unwrap_or("wasmer".into()));
             } else {
                 let user = wasmer_api::query::current_user_with_namespaces(&client, None).await?;
                 owner = crate::utils::prompts::prompt_for_namespace(
--- a/lib/cli/src/commands/app/deploy.rs
+++ b/lib/cli/src/commands/app/deploy.rs
@@ -301,6 +301,14 @@ impl AsyncCliCommand for CmdAppDeploy {
                     .as_mapping_mut()
                     .unwrap()
                     .insert("owner".into(), owner.clone().into());
+
+                if app_yaml.get("app_id").is_some() {
+                    app_yaml.as_mapping_mut().unwrap().remove("app_id");
+                }
+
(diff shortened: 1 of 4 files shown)


lib/cli/src/commands/app/deploy.rs
```diff
@@ -279,10 +279,40 @@ impl AsyncCliCommand for CmdAppDeploy {
             None
         };
 
-        let owner = self
+        let mut owner = self
             .get_owner(&client, &mut app_yaml, maybe_edge_app.as_ref())
             .await?;
 
+        if !wasmer_api::query::viewer_can_deploy_to_namespace(&client, &owner).await? {
```
syrusakbary, line 286: This is not the way. We can simply fetch that the app exists (given the app_id if existing, or using the namespace/name), and if not... try to deploy it as normal and let it fail if the backend fails

lib/cli/src/commands/app/deploy.rs
```diff
@@ -279,10 +279,40 @@ impl AsyncCliCommand for CmdAppDeploy {
...
+            eprintln!(
+                "Cannot deploy app to namespace {}, as the current user is not authorized.",
```
syrusakbary, line 288:
```suggestion
                "It seems you don't have access to the app {name} ({owner})",
```

lib/cli/src/commands/app/deploy.rs
```diff
@@ -279,10 +279,40 @@ impl AsyncCliCommand for CmdAppDeploy {
...
+                owner.bold()
+            );
+            if self.non_interactive {
+                anyhow::bail!("Please, check the app configuration or the current user with the `whoami` command!");
```
syrusakbary, line 292:
```suggestion
                anyhow::bail!("Please, change the owner before deploying or check your current user with `wasmer whoami`.");
```

syrusakbary (commented):
Comments

syrusakbary (approved):

lib/cli/src/commands/app/deploy.rs
xdoardo: FTR: We agreed that just checking if the user can deploy to a namespace is simpler than fetching the app, as receiving just `None` from the backend does not tell us which situation the user is in: do they have permissions to deploy an app and it does not exist (yet) or the user can't deploy to the selected namespace and should thus change the owner?

PR #5005, opened 2024-08-11
Enable feature `net` for mio in virtual-net, fix #5004

### LINKED ISSUE #5004: Build faild after update some deps.
Build faild after update some deps.

The `virtual-net` with feature `host-net`  need `mio 0.8` with feature `net`, but we don't enable the `net` feature.

We could compile it now because we use `tokio 1.38`, which use `mio 0.8` with feature `net`. And after I run `cargo update` to update tokio to `1.39`, which doesn't use `mio 0.8` anymore, `virtual-net` will no longer compile successfully.

### Steps to reproduce
```shell
cargo update
cargo build -p virtual-net
```
```shell
error[E0433]: failed to resolve: could not find `net` in `mio`
  --> lib/virtual-net/src/host.rs:71:34
   |
71 |                     stream: mio::net::TcpListener::from_std(sock),
   |                                  ^^^ could not find `net` in `mio`
```

Dirreke:
Close #5004

### DIFF
--- a/lib/virtual-net/Cargo.toml
+++ b/lib/virtual-net/Cargo.toml
@@ -16,7 +16,7 @@ async-trait = { version = "^0.1" }
 tracing = "0.1"
 tokio = { version = "1", default-features = false, features = ["io-util"] }
 libc = { workspace = true, optional = true }
-mio = { version = "0.8", optional = true }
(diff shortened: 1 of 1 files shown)


theduke (approved):
Thanks!

PR #5007, opened 2024-08-12
add access to php integration test secrets

barankaraaslan:
these env vars are required for running soon to be enabled php tests

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -873,6 +873,20 @@ jobs:
     - uses: actions/download-artifact@v3
       with:
         name: wasmer-cli-linux-x64 
+    - name: Cargo Registry Cache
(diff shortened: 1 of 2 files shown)


xdoardo:
While it does not strictly concern this PR, we should note that the CI test `test-sys / Build and test C-API on macos-x64 (pull_request)`  failed - I think re-running it might pass - but also printed this warning: 

```
Warning: The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
Error: The operation was canceled.
```

xdoardo (approved):

theduke (approved):

PR #5008, opened 2024-08-12
Allow nested mounted paths in `UnionFilesystem`

maminrayej:
This PR adds support for nested mounted paths such as:
```toml
[fs]
"/app/a" = "a"
"/app/b" = "b"
```
when using the `UnionFilesystem`.


lib/backend-api/src/client.rs
```diff
@@ -6,6 +6,16 @@ use anyhow::{bail, Context as _};
...
 
+pub fn get_proxy() -> Result<Option<reqwest::Proxy>, anyhow::Error> {
+    if let Ok(scheme) = std::env::var("http_proxy").or_else(|_| std::env::var("HTTP_PROXY")) {
+        let proxy = reqwest::Proxy::all(scheme)?;
+
+        Ok(Some(proxy))
+    } else {
+        Ok(None)
+    }
+}
```
syrusakbary, line 17: I believe we already have this somewhere (CLI?). We should reuse it, not copy&paste

theduke: We can move it to the api crate, and re-use it in the CLI.

But I believe it might be in the registry crate ATM...

lib/virtual-fs/src/union_fs.rs
```diff
@@ -158,17 +158,19 @@ impl UnionFileSystem {
...
+            match mount.fs.read_dir(Path::new(p.as_str())) {
                 Ok(dir) => {
                     if ret.is_none() {
                         ret = Some(Vec::new());
                     }
                     let ret = ret.as_mut().unwrap();
-                    for sub in dir.flatten() {
+                    for mut sub in dir.flatten() {
+                        let sub_path: PathBuf = sub.path.components().skip(1).collect();
+                        sub.path = PathBuf::from(&mount.path).join(&sub_path);
```
theduke, line 173: Does this code need to handle empty sub-paths?

theduke (commented):
Looks reasonable to me, but not very familiar with the union fs.

lib/backend-api/src/client.rs
maminrayej: but `backend-api` does not depend on `api`. It'll be a dependency just for a single function.

lib/virtual-fs/src/union_fs.rs
maminrayej: `sub` contains entries with this format: `/path/to/entry`. The smallest number of components `sub` can have is when you read the root directory and get something like this: `/foo`. so `sub_path` will contain `foo`. The `read_dir` must return `/` for `sub_path` to become empty which isn't a valid value.

lib/virtual-fs/src/union_fs.rs
```diff
@@ -178,6 +180,43 @@ impl UnionFileSystem {
             }
         }
 
+        // also take into account partial matchings
```
syrusakbary, line 183: What is partial matchings? This seems to be doing things that the previous for loop should handle

lib/virtual-fs/src/union_fs.rs
```diff
@@ -4,88 +4,100 @@
...
-    sync::{Arc, Mutex, Weak},
-};
+use std::path::Path;
 
-use tracing::{debug, trace};
+#[derive(Debug)]
+pub enum UnionOrFs {
+    Union(Box<UnionFileSystem>),
+    FS(Box<dyn FileSystem>),
+}
```
syrusakbary, line 13: I don't think we need this

lib/virtual-fs/src/union_fs.rs
```diff
@@ -4,88 +4,100 @@
...
-    pub should_sanitize: bool,
-    pub new_path: Option<String>,
-}
-
 /// Allows different filesystems of different types
 /// to be mounted at various mount points
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Default)]
 pub struct UnionFileSystem {
     pub mounts: Vec<MountPoint>,
```
syrusakbary, line 102: I think this should be a HashMap of `<directory, MountPoint>`

lib/virtual-fs/src/union_fs.rs
```diff
@@ -5,89 +5,37 @@
...
-    pub new_path: Option<String>,
-}
-
 /// Allows different filesystems of different types
 /// to be mounted at various mount points
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Default)]
 pub struct UnionFileSystem {
-    pub mounts: Vec<MountPoint>,
+    pub mounts: RwLock<HashMap<PathBuf, MountPoint>>,
```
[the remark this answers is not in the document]
Arshia001: I remember we also had `AHashMap` as a dependency, right?

PR #5013, opened 2024-08-13
CLI: Volumes commands

theduke:
Add volumes commands to the CLI:

* app volume s3-credentials
* app volume list

- **feat(backend-api): Update schema + add volumes queries**
- **feat(cli): Add "app volumes" commands**

### DIFF
--- a/lib/cli/src/commands/app/volumes/list.rs
+++ b/lib/cli/src/commands/app/volumes/list.rs
@@ -1,0 +1,38 @@
+//! List Edge apps.
+
+use super::super::util::AppIdentOpts;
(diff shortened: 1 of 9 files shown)


lib/cli/src/commands/app/volumes/list.rs
```diff
@@ -0,0 +1,38 @@
+//! List Edge apps.
```
xdoardo, line 1:
```suggestion
//! List volumes tied to an edge app.
```

xdoardo: (I know, I know, nitpicky!)

xdoardo (approved):

PR #5019, opened 2024-08-19
Cranelift update

xdoardo:
An in-progress effort to upgrade cranelift.

Side quests: 
+ [x] update the MSRV to 1.78. 
+ [x] update `wat` to latest
+ [x] update `wasmparser`


lib/compiler-cranelift/src/config.rs
```diff
@@ -137,11 +136,11 @@ impl Cranelift {
...
-        // There are two possible traps for division, and this way
-        // we get the proper one if code traps.
-        flags
-            .enable("avoid_div_traps")
-            .expect("should be valid flag");
+        //// There are two possible traps for division, and this way
+        //// we get the proper one if code traps.
+        //flags
+        //    .enable("avoid_div_traps")
+        //    .expect("should be valid flag");
```
syrusakbary, line 143: This setting no longer exists in cranelift? If so, can we just remove this?

lib/compiler-cranelift/src/config.rs
```diff
@@ -177,15 +176,15 @@ impl Cranelift {
...
-        }
+        //if is_riscv {
+        //    flags
+        //        .set("enable_simd", "false")
+        //        .expect("should be valid flag");
+        //} else {
+        //    flags
+        //        .set("enable_simd", "true")
+        //        .expect("should be valid flag");
+        //}
```
syrusakbary, line 187: Remove me

lib/api/src/sys/tunables.rs
```diff
@@ -255,6 +255,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature = "cranelift")]
```
syrusakbary, line 258: This should be for any engine. Undo

lib/compiler-cranelift/Cargo.toml
```diff
@@ -14,21 +14,32 @@ rust-version.workspace = true
...
+	"isle-in-source-tree"
+] }
+cranelift-frontend = { version = "0.110.2", default-features = false }
+itertools = "0.12.0"
 tracing = "0.1"
 hashbrown = { version = "0.11", optional = true }
 rayon = { version = "1.5", optional = true }
 more-asserts = "0.2"
-gimli = { version = "0.26", optional = true }
+gimli = { version = "0.28.0", optional = true }
```
syrusakbary, line 37: Nice. can we update gimli as well on the singlepass crate and make the version global for the Cargo workspace?

lib/compiler-cranelift/src/func_environ.rs
```diff
@@ -119,7 +125,9 @@ pub struct FuncEnvironment<'module_environment> {
     /// The memory styles
     memory_styles: &'module_environment PrimaryMap<MemoryIndex, MemoryStyle>,
 
-    /// The table styles
+    /// Cranelift tables we have created to implement Wasm tables.
+    tables: SecondaryMap<TableIndex, Option<TableData>>,
```
syrusakbary, line 129: Why this is an option?

It seems to me it should be:

```suggestion
    tables: SecondaryMap<TableIndex, TableData>,
```

lib/compiler-cranelift/Cargo.toml
syrusakbary: Note: use the same gimli version as `cranelift-codegen` (that is: `0.28` (.1) )

lib/compiler-cranelift/src/func_environ.rs
xdoardo: The reason, I think, is that there's no sensible default value for `TableData`, and creating a new `SecondaryMap` (both from `new` or `default`) [requires the value type to implement `Default`](https://docs.rs/cranelift-entity/latest/cranelift_entity/struct.SecondaryMap.html#method.new). Perhaps, not using `SecondaryMap`s might be viable as well.

syrusakbary (approved):

syrusakbary (approved):

PR #5022, opened 2024-08-19
feat(cli): Add --print-rclone-config to "app volumes s3-credentials"

theduke:


ayys (approved):
we could have a `--outfile` or `--out` flag to store the config in a file `edge-<name>.rclone.config` and then add a comment to the output letting the user know

something like,

```
[stderr] # You can append the app config to rclone by running this command:
[stderr] # wasmer app volumes s3-credentials --rclone-print >> edge-{app_name}.config
[stderr] # rclone --config  edge-{app_name}.config ls edge-{app_name}:/
[edge-{app_name}]
type = s3
provider = Other
acl = private
access_key_id = {access_key}
secret_access_key = {secret_key}
endpoint = {endpoint}
```

PR #5025, opened 2024-08-20
CLI: Allow sorting in "app list" command

theduke:
- **feat(backend-api): Allow sorting in app queries**
- **feat(cli): Allow sorting in "app list" command**

### DIFF
--- a/lib/cli/src/commands/app/list.rs
+++ b/lib/cli/src/commands/app/list.rs
@@ -3,7 +3,7 @@
 use std::pin::Pin;
 
 use futures::{Stream, StreamExt};
-use wasmer_api::types::DeployApp;
+use wasmer_api::types::{DeployApp, DeployAppsSortBy};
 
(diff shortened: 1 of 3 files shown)


lib/cli/src/commands/app/list.rs
```diff
@@ -34,6 +34,19 @@ pub struct CmdAppList {
     /// Asks whether to display the next page or not
     #[clap(long, default_value = "false")]
     paging_mode: bool,
+
+    /// Sort order for apps.
+    ///
+    /// Use: --newest, --oldest or --last-updated
```
xdoardo, line 40: Doesn't clap print possible values already? Maybe it's different from `ValueEnum`'d enums.

xdoardo (approved):

lib/cli/src/commands/app/list.rs
theduke: It didn't use to, they might have added it.

PR #5030, opened 2024-08-21
Improve loop metering tests

webmaster128:
# Description

I ran a few tests around metering the `loop` instruction, wondering if the accounting for `Operator::Loop { .. }` can be avoided. However, this is not the case as that would subtract everything before the loop for every iteration. So I ended up with a few more tests that might be helpful to show and harden loop metering behaviour for terminating and infinite loops.


xdoardo:
LGTM!

xdoardo (approved):

PR #5053, opened 2024-08-29
Reduce overhead of chunk timeouts in wasix package downloads

### LINKED ISSUE #5052: WASIX: Reduce overhead of chunk timeouts in package downloads
Currently a fresh timeout is created for each chunk, which is is a lot of overhead.

theduke:
Closes #5052

Closes RUN-414

### DIFF
--- a/lib/wasix/src/http/reqwest.rs
+++ b/lib/wasix/src/http/reqwest.rs
@@ -69,12 +69,59 @@ impl ReqwestHttpClient {
         let headers = std::mem::take(response.headers_mut());
 
         let status = response.status();
+
+        // Download the body.
         let data = if let Some(timeout) = self.response_body_chunk_timeout {
+            // Download the body with a chunk timeout.
+            // The timeout prevents long stalls.
+
             let mut stream = response.bytes_stream();
             let mut buf = Vec::new();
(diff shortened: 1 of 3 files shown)


lib/wasix/src/http/reqwest.rs
```diff
@@ -69,12 +69,59 @@ impl ReqwestHttpClient {
...
+                        }
+
+                        _ = &mut timeout => {
+                            if chunk_count == 0 {
+                                return Err(anyhow::anyhow!("Timeout while downloading response body"));
+                            } else {
+                                // Timeout, but chunks were still downloaded, so
+                                // just continue with a fresh timeout.
+                                continue 'OUTER;
+                            }
```
maminrayej, line 119: should we reset `chunk_count` to `0` when we `continue` since `if chunk_count == 0` will never be `true` again if one chunk has been downloaded regardless how many timeouts happen.

theduke: Oh whoops, that got lost when I refactored a bit.

theduke: Oh no: so the chunk_count is defined within the loop, so on the next loop iteration it'll be re-initialized to 0.

maminrayej (approved):

PR #5057, opened 2024-09-01
Add `cwd` to the manifest as a command annotation

maminrayej:
Adds support for specifying the working directory for each command as a wasi annotation. For example:

```toml
[command.annotations.wasi]
cwd = "/data"
```


lib/wasix/src/runners/wasi.rs
```diff
@@ -353,6 +353,10 @@ impl crate::runners::Runner for WasiRunner {
             }
         }
 
+        if let Some(cwd) = cmd.metadata().annotation::<String>("cwd")? {
```
theduke, line 356: Shouldn't this also be a `wasi` annotation?
Not every kind of runner would definitely have a CWD, and it fits together with the other wasi settings.

theduke, line 356: Also, shouldn't we add this to the type definitions in webc?

maminrayej: It was added as a custom annotation which does not have a type definition and is just a plain `Value`, but adding it as a wasi annotation will require a change to webc.

tests/integration/cli/tests/packages/list-cwd/wasmer.toml
```diff
@@ -0,0 +1,14 @@
...
+[fs]
+"/data" = "."
+
+[[command]]
+name = "run"
+module = "main"
+runner = "wasi"
+
+[command.annotations]
+cwd = "/data"
```
syrusakbary, line 14: This needs to be part of the WASI annotations, not just the command

syrusakbary (approved):

PR #5065, opened 2024-09-03
Remove merging the filesystems

### LINKED ISSUE #5064: VFS: Regression in Union FS: Excessive copying of files
The new union FS implementation leads to excessive copying of files into the host file system, delaying startup significantly and causing lots of extra work.

To reproduce:

```
RUST_LOG=virtual_fs=trace,wasmer_wasix=trace wasmer run --registry wasmer.wtf  --net sha256:68c1f418474d98f0c4ca44d6754f144fe355525d1ba347dc52ab934cb98c5dad &| tee log.txt &

curl localhost:8080

❯ cat log.txt | grep 'Copying file into host filesystem' | wc -l
5846
```

This might also cause some kind of deadlocking or freezes, as seen in the app above.
The PHP code stalls and the request never returns.

maminrayej:
Resolves #5064.

### DIFF
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -5,7 +5,7 @@ mod notification;
 use std::{
     borrow::{Borrow, Cow},
     cmp::Reverse,
-    collections::{BinaryHeap, HashMap, HashSet, VecDeque},
+    collections::{BinaryHeap, HashMap, HashSet},
     ops::{Deref, DerefMut},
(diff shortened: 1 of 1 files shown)


syrusakbary:
As chatted, this PR would make `wasmer run xyz` inside of a `wasmer run sharrattj/bash` fail, as we need merge functionality at runtime. Tests pass on this PR because we are not testing this functionality atm.

Closing this PR in favor of #5066

PR #5066, opened 2024-09-03
Prevent redundant merging of the filesystems

### LINKED ISSUE #5064: VFS: Regression in Union FS: Excessive copying of files

maminrayej:
Another approach related to #5065.

Resolves #5064.

### DIFF
--- a/lib/wasix/src/runners/wasi.rs
+++ b/lib/wasix/src/runners/wasi.rs
@@ -242,6 +242,7 @@ impl WasiRunner {
         let container_fs = if let Some(pkg) = pkg {
             builder.add_webc(pkg.clone());
             builder.set_module_hash(pkg.hash());
+            builder.include_package(pkg.id.clone());
             Some(Arc::clone(&pkg.webc_fs))
         } else {
             None
--- a/lib/wasix/src/runners/wasi.rs
+++ b/lib/wasix/src/runners/wasi.rs
@@ -242,7 +242,7 @@ impl WasiRunner {
         let container_fs = if let Some(pkg) = pkg {
             builder.add_webc(pkg.clone());
             builder.set_module_hash(pkg.hash());
-            builder.include_package(pkg.id.clone());
+            builder.include_packages(pkg.package_ids.clone());
             Some(Arc::clone(&pkg.webc_fs))
         } else {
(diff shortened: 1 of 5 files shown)


lib/wasix/src/runners/wasi.rs
```diff
@@ -242,6 +242,7 @@ impl WasiRunner {
         let container_fs = if let Some(pkg) = pkg {
             builder.add_webc(pkg.clone());
             builder.set_module_hash(pkg.hash());
+            builder.include_package(pkg.id.clone());
```
theduke, line 245: Isn't this too limited?

A BinaryPackage can consist of multiple dependencies providing file systems, do all of those need to be included here?

(I think this would require an extension to BinaryPackage, because the file system entries don't record their source package IDs, IIRC).

theduke: Do we have a test that gets FS volumes from multiple dependencies?

theduke:
@maminrayej please no force pushes here, I'll integrate into the edge branch.

lib/wasix/src/runners/wasi.rs
maminrayej: Yes, makes sense.

lib/wasix/src/state/env.rs
```diff
@@ -1096,7 +1096,7 @@ impl WasiEnv {
 
         // We first need to copy any files in the package over to the
         // main file system
```
syrusakbary, line 1098: This comment is no longer relevant

syrusakbary (approved):

PR #5070, opened 2024-09-05
Add `rotate-secrets` subcommand for volumes and minor changes

xdoardo:
This PR adds a `rotate-secrets` subcommand for the `app volumes` command in the CLI. Furthermore, rearranges a bit the code and renames `s3-credentials` to `credentials`, adding support for different format options in this command. 

This also adds a small patch to prevent the `deploy` command to enter `create` if run in non-interactive mode without an available app, which is not worth another PR.


lib/cli/src/commands/app/volumes/credentials/rotate_secrets.rs
```diff
@@ -0,0 +1,36 @@
...
+    type Output = ();
+
+    async fn run_async(self) -> Result<Self::Output, anyhow::Error> {
+        let client = self.env.client()?;
+        let (_ident, app) = self.ident.load_app(&client).await?;
+
+        wasmer_api::query::rotate_s3_secrets(&client, app.id).await?;
+
+        println!(
+            "Correctly rotated s3 secrets for app {} ({})",
```
theduke, line 29: When you rotate you'll pretty much always want to see the new secrets, no?

Would make sense for the command to also show the new values.

lib/cli/src/commands/app/volumes/credentials/s3.rs
```diff
@@ -0,0 +1,120 @@
+//! Retrieve credentials to manage app volumes.
+
+use crate::{
+    commands::{app::util::AppIdentOpts, AsyncCliCommand},
+    config::WasmerEnv,
+};
+
+/// Retrieve access credentials for the volumes of an app.
```
theduke, line 8: File is named `s3.rs` while it should be `credentials.rs`.

theduke (commented):
Docs on docs.wasmer.io also need to be updated for the changes here.

lib/cli/src/commands/app/volumes/credentials/s3.rs
theduke: Ohh it's an extra nested submodule...

Doesn't really make sense to me when the command is called credentials?

xdoardo: I thought as follows: `rotate-secrets` is still something related to credentials, but we don't want to have a command hierarchy such as: 
```
wasmer app volumes credentials show
wasmer app volumes credentials rotate-secrets
```
as, imo, it's just too long. Therefore, we wanted to have something like: 
```
wasmer app volumes credentials <...>
wasmer app volumes rotate-secrets
```
Nonetheless, rotating secrets is related to credentials, and that's why there's a nested submodule. The contents of `s3.rs` can go in `mod.rs`, yes; the rationale behind the decision was to expose users to the fact that the infrastructure backing volumes is an s3-compatible storage only when it's needed - but the credentials still are s3 credentials indeed.

lib/cli/src/commands/app/volumes/credentials/rotate_secrets.rs
xdoardo: ([5f80b94](https://github.com/wasmerio/wasmer/pull/5070/commits/5f80b948eba4ed1f9989676560388ee280135fc2))

theduke (approved):

PR #5071, opened 2024-09-06
Annotation Builder

### LINKED ISSUE #5069: Make creating annotations for a command easier

maminrayej:
Resolves #5069.


syrusakbary (changes requested):
This PR is still storing things as a toml in memory, which is not ideal and doesn't cause the issue with PartialEq being solved. It also attaches things as always stored in toml, but we may want in the future to target other formats.

Ideally, we should have something like:

```rust
#[derive(Debug, Clone, Default)]
pub struct CommandAnnotationBuilder {
    annotations: HashMap<String, Box<dyn Serializable>>,
}
```

PR #5089, opened 2024-09-15
feat(cli): Restore packages from webcs

### LINKED ISSUE #5085: CLI: allow synthesizing a wasmer.toml in "wasmer package unpack"
Extracted packages only contain a webc manifest, not a wasmer.toml file.

Converting a manifest to a wasmer.toml by hand is very tedious.

The unpack command should have an extra flag `--generate-wasmer-toml` or similar to make this easier.

theduke:
Allow "package unpack" to restore packages

Add a new --format option to package unpack.

* package: Try to restore a usable package dir with a wasmer.toml
* webc: just unpack the webc

Also adds a new wasmer-package crate that can hold package related functionality.

As a drive-by chore, also lifts pretty_assertions to be a workspace dependency.

Closes #5085


lib/package/src/convert.rs
```diff
@@ -0,0 +1,367 @@
...
+
+    for (atom_name, data) in webc.atoms() {
+        let atom_path = module_dir.join(&atom_name);
+
+        std::fs::create_dir_all(&module_dir).map_err(|err| {
+            ManifestConversionError::with_cause(
+                format!("Could not create directory '{}'", module_dir.display(),),
+                err,
+            )
+        })?;
```
maminrayej: should this be called outside of the `for` loop?

lib/package/src/convert/webc_to_package.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    for (atom_name, data) in webc.atoms() {
+        let atom_path = module_dir.join(&atom_name);
+
+        std::fs::create_dir_all(&module_dir).map_err(|err| {
+            ManifestConversionError::with_cause(
+                format!("Could not create directory '{}'", module_dir.display(),),
+                err,
+            )
+        })?;
```
maminrayej, line 142: should this be called outside of the `for` loop?

maminrayej:
Otherwise LGTM.

lib/package/src/convert/webc_to_package.rs
theduke: Putting it in the loop was deliberate because the directory should only be created if there are modules, and I was to lazy to put an extra if check outside of the loop. ;)

maminrayej: I think it's fine merging it as-is. But I think it's also fine to create an empty `modules` directory.
A `TODO` would be nice to mark that section for a future clean up.

theduke: Done

maminrayej (approved):

PR #5090, opened 2024-09-15
Add the right to fetch stats on pipes

### LINKED ISSUE #5080: Make aio-http work with latest python and wasmer

maminrayej:
This PR adds the right to fetch the stats of file descriptors returned by the `fd_pipe` syscall.

Resolves #5080.


syrusakbary (approved):
Can we add a simple test case for this?

PR #5091, opened 2024-09-16
Apply the proxy setting in wasmer config

maminrayej:
There is support for setting a proxy in wasmer config, but it was not applied when creating a client. This PR fixes this issue. 

Note: Unfortunately, there is still need for `HTTP_PROXY` env var since there are clients being created deep in the source code that do not have access to `WasmerEnv`. Sometimes the combination of both is needed for the entire traffic to be rerouted through the proxy. Fixing this requires a non-trivial refactor that is beyond the scope of this PR.

### DIFF
--- a/lib/backend-api/src/client.rs
+++ b/lib/backend-api/src/client.rs
@@ -75,10 +75,10 @@ impl WasmerClient {
         })
     }
 
-    pub fn new_with_proxy(
+    pub fn new(
         graphql_endpoint: Url,
         user_agent: &str,
-        proxy: reqwest::Proxy,
+        proxy: Option<reqwest::Proxy>,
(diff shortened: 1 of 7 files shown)


lib/backend-api/src/client.rs
```diff
@@ -75,10 +75,10 @@ impl WasmerClient {
         })
     }
 
-    pub fn new_with_proxy(
```
syrusakbary, line 78: To not do breaking api, can we call this `new_with_proxy` and then have a `new` that just calls `new_with_proxy`?

maminrayej: Yeah makes sense.

syrusakbary (approved):

Issue #5047, opened 2024-08-27
Cgi runner fails on second http requests


barankaraaslan:
Create an rust cgi app with `wasmer app create`

build and run the app<br>send 2 consecutive get requests through curl

the second request get stuck<br><br><br>But. If you send the first request through web browser, it works

PR #5095, opened 2024-09-17
Remove always false condition in test CI

maminrayej:


maminrayej:
I've disabled the WCGI test for now, because of #5047.

PR #5098, opened 2024-09-18
`js::module`: leave warning comment on the `Send` impl

Twey:
# Description

Until https://github.com/wasmerio/wasmer/issues/4158 is fixed, let's at least leave a signpost so future readers don't trip on it like I did :)

### DIFF
--- a/lib/api/src/js/module.rs
+++ b/lib/api/src/js/module.rs
@@ -44,14 +44,10 @@ pub struct Module {
     raw_bytes: Option<Bytes>,
 }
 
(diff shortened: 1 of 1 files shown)


syrusakbary:
Modules should be `Send`, but we need to spend time to investigate why they were not working.
We can merge this until gets solved

Twey:
> Modules should be `Send`, but we need to spend time to investigate why they were not working.
> We can merge this until gets solved

I don't think that JavaScript modules can ever be `Send`, since we must always `postMessage` the JS object (or have some other explicit transfer step).

PR #5107, opened 2024-09-23
Add new --invoke arg for choosing entry function for single WASM modules + fix --invoke not working for WASI(X) modules

### LINKED ISSUE #4691: `wasmer run . --net --entrypoint=abc` doesn't work
Wasmer entrypoint doesn't work.

Steps:
1. Clone https://github.com/wasmer-examples/laravel-wasmer-starter
2. Run `wasmer run . --net --entrypoint=artisan`
3. It doesn't do anything -> It should show

It should show something like:

```
$ wasmer run . --net --entrypoint=artisan
wasm execute failed [/bin/sh] - could not find binary '/bin/sh'
wasm execute failed [/usr/local/bin/stty] - could not find binary '/usr/local/bin/stty'
wasm execute failed [/usr/local/bin/stty] - could not find binary '/usr/local/bin/stty'
Laravel Framework 11.7.0

Usage:
  command [options] [arguments]

Options:
 ...
```

You can check the right output if you change in `wasmer.toml` the entrypoint from `php` to `artisan`

Arshia001:
This closes #4691.

### DIFF
--- a/lib/cli/src/commands/run/mod.rs
+++ b/lib/cli/src/commands/run/mod.rs
@@ -76,9 +76,12 @@ pub struct Run {
     /// Set the default stack size (default is 1048576)
     #[clap(long = "stack-size")]
     stack_size: Option<usize>,
-    /// The function or command to invoke.
-    #[clap(short, long, aliases = &["command", "invoke", "command-name"])]
+    /// The entrypoint module for webc packages.
+    #[clap(short, long, aliases = &["command", "command-name"])]
     entrypoint: Option<String>,
+    /// The function to invoke.
+    #[clap(short, long)]
+    invoke: Option<String>,
     /// Generate a coredump at this path if a WebAssembly trap occurs
     #[clap(name = "COREDUMP_PATH", long)]
     coredump_on_trap: Option<PathBuf>,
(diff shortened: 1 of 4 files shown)


xdoardo (approved):

syrusakbary (changes requested):
I think there's misunderstanding on what entrypoint is for.

Entrypoint is used for calling a specific command inside of webc. Basically, it needs to be a command inside of a package. Not a function inside of wasm.

For a function inside of wasm we should have (if we don't have already) an `--invoke` flag in the cli

Arshia001:
@syrusakbary my understanding is that, for pure WASM modules, the entrypoint flag specifies a function inside the module which is to be invoked: https://github.com/wasmerio/wasmer/blob/d11fccdd74cfba48ac3cad78a58a343a611e3f6b/lib/cli/src/commands/run/mod.rs#L380-L390

Considering that this code deals with running a single WASIX module (as opposed to an entire webc) I'd say the code is consistent with what was already there.

Arshia001:
@syrusakbary Please review again

Note: this is a breaking change, and should be merged only for 5.0.

syrusakbary (approved):

PR #5120, opened 2024-09-26
Improve QueryError

### LINKED ISSUE #5119: wasix: Improve the QueryError
The package resolution QueryError does not carry the queried identifier, which leads to very confusing errors propagating through the system.

The error should be refactored into a struct with the queried specified and an error kind.

theduke:
Extend the QueryError to hold the queried package for improved error reporting.

Closes #5119

### DIFF
--- a/lib/wasix/src/runtime/resolver/backend_source.rs
+++ b/lib/wasix/src/runtime/resolver/backend_source.rs
@@ -250,10 +250,7 @@ impl Source for BackendSource {
                         });
                     }
                     Err(error) => {
-                        return Err(QueryError::Other {
-                            query: package.clone(),
-                            error,
-                        });
+                        return Err(QueryError::new_other(error, package));
                     }
                 }
             }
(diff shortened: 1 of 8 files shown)


lib/wasix/src/runtime/resolver/backend_source.rs
```diff
@@ -217,7 +217,7 @@ impl BackendSource {
     async fn query_by_hash(
         &self,
         hash: &PackageHash,
-    ) -> Result<Option<PackageSummary>, QueryError> {
+    ) -> Result<Option<PackageSummary>, anyhow::Error> {
```
maminrayej, line 220: Are we returning any other error than `QueryError`, for this to be needed?

theduke: This is a private helper function that never returns a specific QueryError, so I changed the return type to avoid having to  write `.map_err()` 5 times.

maminrayej (approved):

PR #5126, opened 2024-10-01
BROKEN: Fix running multi-threaded WASIX modules with data initialize…

Arshia001:
…rs by messing everything up and breaking WASIX on JS and JSC


Arshia001:
This was a fun experiment. Time to get real though.

PR #5127, opened 2024-10-02
Add support for `exec-name` and transfer args and envs after execve

maminrayej:
This PR adds support for `exec-name` annotation for overriding the name that is passed to an executable as arg0. Also, previously when spawning a command using `execve`, the args and env vars specified in the manifest for that command, were not passed along properly when spawning the package. This would break some command that expect a specific env vars or args to function properly.

This PR is marked as draft since using `exec-name` in the code path used by `wasmer run` is still not done.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7377,9 +7377,9 @@ dependencies = [
 
 [[package]]
 name = "webc"
-version = "6.0.1"
+version = "6.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c48441419be082f8d2537c84d8b1f502624d77bc08fbbd09ab17cadfe7f0ac53"
+checksum = "cdea84cf234555864ca9b7a5084c1a99dbdf2d148035f62a09b19ce5606532c1"
 dependencies = [
  "anyhow",
  "base64",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,7 +96,7 @@ wasmer-config = { path = "./lib/config" }
 wasmer-wasix = { path = "./lib/wasix" }
 
 # Wasmer-owned crates
-webc = { version = "6.0.1", default-features = false, features = ["package"] }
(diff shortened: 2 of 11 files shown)


lib/wasix/src/bin_factory/binary_package.rs
```diff
@@ -224,6 +224,25 @@ impl BinaryPackage {
...
+
+        match self.commands.as_slice() {
+            [] => anyhow::bail!("The WEBC file doesn't contain any executable commands"),
+            [one] => Ok(one.name()),
+            [..] => {
+                let mut commands: Vec<_> = self.commands.iter().map(|cmd| cmd.name()).collect();
+                commands.sort();
+                anyhow::bail!(
+                    "Unable to determine the WEBC file's entrypoint. Please choose one of {:?}",
+                    commands,
```
syrusakbary, line 241: This is not a good message. Users should not care a bit about webc.

```suggestion
                anyhow::bail!(
                    "Unable to determine the package's entrypoint. Please choose one of {:?}",
                    commands,
```

lib/wasix/src/bin_factory/binary_package.rs
```diff
@@ -224,6 +224,25 @@ impl BinaryPackage {
...
         })
     }
+
+    pub fn infer_entrypoint(&self) -> Result<&str, anyhow::Error> {
+        if let Some(entrypoint) = self.entrypoint_cmd.as_deref() {
+            return Ok(entrypoint);
+        }
+
+        match self.commands.as_slice() {
+            [] => anyhow::bail!("The WEBC file doesn't contain any executable commands"),
```
syrusakbary, line 234: This is not a good message. Users should not care a bit about webc.

```suggestion
        match self.commands.as_slice() {
            [] => anyhow::bail!("The package doesn't contain any executable commands"),
```

syrusakbary (approved):

PR #5139, opened 2024-10-07
deps: Update corosensei

ognevny:
# Description
Fixes some recent rustc lints and compilation for `*-pc-windows-gnullvm` targets


syrusakbary (approved):
That's a great upgrade, thanks for the patch!

PR #5151, opened 2024-10-17
Remove negative variant of exit code

### LINKED ISSUE #5131: Can Wasi ExitCode be negative?
### Summary

A clear and concise summary of your question.

I see that https://docs.rs/wasmer-wasix-types/0.27.0/wasmer_wasix_types/wasi/enum.ExitCode.html has an `i32` and as such when you ask for the raw exit code it gives an `i32`.

But after searching around a bit, I don't see a situation in which it would be negative. I see that `Errno` is `repr(u16)`. Can this actually be negative anywhere?

I could be missing it somewhere since a from impl exists:
```rust

impl From<ExitCode> for i32 {
    fn from(val: ExitCode) -> Self {
        match val {
            ExitCode::Errno(err) => err.to_native(),
            ExitCode::Other(code) => code,
        }
    }
}
```

### Additional details

N/A

maminrayej:
Resolves #5131.


theduke (commented):
Since we are making a breaking change anyway, I would be in favour of just changing things to:

```rust
struct ExitCode(u16)
```

We can still do a `Into` implementation for the Errno enum to keep code changes smaller.

theduke (approved):

PR #5153, opened 2024-10-17
Terminate entire wasix process when a worker thread fails or calls exit

### LINKED ISSUE #5130: Panic in spawned thread causes dead lock (?)
### Describe the bug

See reproduction steps

```
wasmer 4.3.7 (c6c348d 2024-09-09)
binary: wasmer-cli
commit-hash: c6c348d00fa3582eb3b3fda07e69fdc695156193
commit-date: 2024-09-09
host: x86_64-pc-windows-msvc
binary: rustc
commit-date: 2024-09-04
host: x86_64-pc-windows-msvc
release: 1.81.0
LLVM version: 18.1.7
```

### Steps to reproduce

```rs
fn main() {
    std::thread::spawn(|| {
        panic!("dead lock?"); // dead locks
    })
    .join()
    .unwrap();
}
```
run it with `cargo wasix run`

### Expected behavior

Process should exit with non-zero code, just like `cargo run` does
```
thread '<unnamed>' panicked at src/main.rs:3:9:
dead lock?
thread 'main' panicked at src/main.rs:6:6:
called `Result::unwrap()` on an `Err` value: Any { .. }
error: process didn't exit successfully: `target\debug\wasix-threading.exe` (exit code: 101)
```

Arshia001:
Fixes #5130.

On native platforms, when a thread causes a segfault, panics, calls `exit()`, or otherwise terminates abnormally, the entire process will exit. This was not the case on WASIX before, but should now be the case with this PR.


lib/wasix/src/state/env.rs
```diff
@@ -1277,7 +1277,7 @@ impl WasiEnv {
         }
 
         // If this is the main thread then also close all the files
-        if self.thread.is_main() {
+        if self.thread.is_main() || matches!(exit_code, Some(e) if !e.is_success()) {
```
theduke, line 1280: We have to make sure this code path isn't hit for subprocesses.
First with manual testing, but I think we also need both negative and positive tests here.

Arshia001: I've already tested everything I can think of with bash. This code path _is_ hit for subprocesses, but only terminates the subprocess, not the parent process.

lib/wasix/src/syscalls/wasix/thread_spawn.rs
```diff
@@ -226,20 +235,23 @@ fn call_module<M: MemorySize>(
                         .runtime
                         .on_taint(TaintReason::UnknownWasiVersion);
                     ret = Errno::Noexec;
+                    exit_code = Some(ExitCode::Other(128 + ret as i32));
```
theduke, line 238: what's going on here?
128 + 19, I think? Aka SIGSTOP?

That's super confusing to read.
Shouldn't we just define that as a constant and use it directly?

lib/wasix/src/syscalls/wasix/thread_spawn.rs
```diff
@@ -226,20 +235,23 @@ fn call_module<M: MemorySize>(
...
                 }
                 Err(err) => {
                     debug!("failed with runtime error: {}", err);
                     env.data(&store)
                         .runtime
                         .on_taint(TaintReason::RuntimeError(err));
                     ret = Errno::Noexec;
+                    exit_code = Some(ExitCode::Other(128 + ret as i32));
```
theduke, line 246: As above.

tests/integration/cli/tests/snapshot.rs
```diff
@@ -1281,3 +1281,57 @@ fn test_snapshot_fs_rename() {
         .run_wasm(include_bytes!("./wasm/fs-rename.wasm"));
     assert_json_snapshot!(snapshot);
 }
+
+#[cfg_attr(any(target_env = "musl", target_os = "windows"), ignore)]
+#[test]
+fn test_snapshot_exit_0_from_main() {
```
theduke, line 1287: Yay, tests, amazing.

I think these would have been a better fit for the new Rust-based test library, since that allows for definining the expected exit code I think ... but fine with leaving them here also.

Although will need to be very careful with the outputs here.
If someone else sees changes here it would be easy to accept a new snapshot with changed values.

Can we add an additional assert_eq() for the expected exit code?

lib/wasix/src/syscalls/wasix/thread_spawn.rs
Arshia001: Apparently, 128 + signal value is the semi-standard exit code when you receive a signal, or so the web seemed to think. No idea if this is actually a thing or not. We may as well have used a constant 1 here, but trying to follow best practices...

tests/integration/cli/tests/snapshot.rs
Arshia001: Snapshots already include the exit code as well, so there's no need for anything more here: https://github.com/wasmerio/wasmer/pull/5153/files#diff-499a0af35079fb7373fe5bc28ca85c9b84fae898427d4bcba948dbf56ebaebdfR19

lib/wasix/src/syscalls/wasix/thread_spawn.rs
theduke: But it's 128+ret, and with the offset of ret being Noexec that would amount to what's associated with SIGSTOP, so that's fine!

This is really only a readability nit.

tests/integration/cli/tests/snapshot.rs
theduke: Yeah they do, my point is that it would be easy to accept a snapshot with a different exit code if you don't know the expected exit code. 

If someone else touches the tests in the future.

lib/wasix/src/syscalls/wasix/thread_spawn.rs
Arshia001: Let's leave it in, so people will later find and think we knew what we were doing and be amazed that we knew it.

tests/integration/cli/tests/snapshot.rs
Arshia001: Well, truth be told, snapshot tests are the only thing I know how to make. I do think this is fine (the exit code is mentioned in the test names as well as the source code and the snapshot) but if you think we absolutely need to change it, feel free to point me at the new test library you mentioned.

theduke (approved):

PR #5160, opened 2024-10-19
Fix fd with append flag can't seek correctly

### LINKED ISSUE #5159: Cannot `fd_seek` with `fdflags::append`
### Describe the bug

If an fd was opened with `fdflags::append`, `fd_seek` with it does not work properly.

```sh
wasmer -vV; rustc -vV
wasmer 4.4.0 (8b97cfe 2024-10-18)
binary: wasmer-cli
commit-hash: 8b97cfe3992ae3c2f0002f9955fcb23057f666a9
commit-date: 2024-10-18
c_api backend: 
binary: rustc
commit-date: 2024-09-04
release: 1.81.0
LLVM version: 18.1.7
```

### Steps to reproduce

Compile the snippet with crate `wasi` v0.11, and run:

```
cargo build --target wasm32-wasip1
mkdir dir
wasmer run --mapdir /dir:dir target/wasm32-wasip1/debug/repro.wasm
```

```rust
fn main() {
    unsafe {
        let base_fd = 5;
        let rights = wasi::RIGHTS_FD_WRITE | wasi::RIGHTS_FD_TELL | wasi::RIGHTS_FD_SEEK;
        let fd = wasi::path_open(
            base_fd,
            0,
            "dir/file",
            wasi::OFLAGS_CREAT | wasi::OFLAGS_EXCL,
            rights,
            rights,
            0,
        )
        .unwrap();
        let fd_append = wasi::path_open(
            base_fd,
            0,
            "dir/file",
            wasi::OFLAGS_CREAT,
            rights,
            rights,
            wasi::FDFLAGS_APPEND,
        )
        .unwrap();

        assert_eq!(wasi::fd_seek(fd, 100, wasi::WHENCE_SET).unwrap(), 100);
        assert_eq!(
            wasi::fd_seek(fd_append, 100, wasi::WHENCE_SET).unwrap(),
            100
        );
    }
}
```

### Expected behavior

The snippet should run successfully. The two seeks should return the same offset.

### Actual behavior

Wasmer fails when running the snippet because the second `fd_seek` returns 0 as the offset.

yagehu:
This commit fixes an incompatible behavior (with other runtimes, and POSIX) when an fd that was opened with `fdflags::append` cannot seek to the right offset with `fd_seek`.  The append flag should not "lock" the offset at the end, as it's possible to seek to another offset and read from it.

This passes all the `wasi-fyi` tests.

fixes #5159

### DIFF
--- a/tests/wasi-fyi/fd_seek_append.rs
+++ b/tests/wasi-fyi/fd_seek_append.rs
@@ -1,32 +1,15 @@
-use std::os::fd::AsRawFd;
-
-#[link(wasm_import_module = "wasi_snapshot_preview1")]
-extern "C" {
-    pub fn fd_seek(fd: i32, offset: i64, whence: i32, filesize: i32) -> i32;
-}
-
-const ERRNO_SUCCESS: i32 = 0;
-const WHENCE_SET: i32 = 0;
+use std::io::{Seek as _, SeekFrom};
 
 fn main() {
-    unsafe {
-        let offset = 100u64;
-        let mut new_offset = 0u64;
-        let f = std::fs::OpenOptions::new()
-            .read(true)
-            .write(true)
-            .create(true)
-            .append(true)
-            .open("/fyi/fs_seek_append.dir/file")
-            .unwrap();
-        let errno = fd_seek(
-            f.as_raw_fd(),
-            offset as i64,
-            WHENCE_SET,
-            &mut new_offset as *mut u64 as usize as i32,
-        );
+    let offset = 100u64;
+    let mut f = std::fs::OpenOptions::new()
+        .read(true)
+        .write(true)
+        .create(true)
+        .append(true)
+        .open("/fyi/fs_seek_append.dir/file")
+        .unwrap();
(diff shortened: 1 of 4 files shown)


tests/wasi-fyi/fd_seek_append.rs
```diff
@@ -0,0 +1,32 @@
...
+            .create(true)
+            .append(true)
+            .open("/fyi/fs_seek_append.dir/file")
+            .unwrap();
+        let errno = fd_seek(
+            f.as_raw_fd(),
+            offset as i64,
+            WHENCE_SET,
+            &mut new_offset as *mut u64 as usize as i32,
+        );
```
syrusakbary, line 27: Could we have an example that triggers using only rust std (without importing any module from `wasi_snapshot_preview1`)?

That way we can have future-proof tests

yagehu: Absolutely. Fixed!

syrusakbary (commented):
Great, thanks. Will wait for @maminrayej review

PR #5162, opened 2024-10-20
Fix append-ness of new fd affecting old fds

### LINKED ISSUE #5161: Opening a file with append should not change existing fd offset
### Describe the bug

If file has been opened with a descriptor `fd0`, opening the same file with the append flag&mdash;producing `fd1`&mdash;should not change the offset of `fd0`.  `fd0` and `fd1` should have independent offsets.

This behavior is consistent across Linux and other Wasm runtimes (Wasmtime, WAMR, WasmEdge, Wazero), but Wasmer changes the offset of the original descriptor.

```sh
wasmer -vV; rustc -vV
wasmer 4.4.0 (8b97cfe 2024-10-18)
binary: wasmer-cli
commit-hash: 8b97cfe3992ae3c2f0002f9955fcb23057f666a9
commit-date: 2024-10-18
c_api backend: 
binary: rustc
commit-date: 2024-09-04
release: 1.81.0
LLVM version: 18.1.7
```

### Steps to reproduce

Compile and run this snippet, mounting some diretory:

```
cargo build --target wasm32-wasip1
wasmer run --mapdir /dir:dir target/wasm32-wasip1/debug/repro.wasm
```

```rust
use std::{
    fs,
    io::{Seek, SeekFrom, Write},
};

fn main() {
    let mut f0 = fs::OpenOptions::new()
        .create(true)
        .write(true)
        .open("file")
        .unwrap();

    f0.write_all(b"abc").unwrap();
    f0.seek(SeekFrom::Start(1)).unwrap();

    assert_eq!(fs::read_to_string("file").unwrap(), "abc");

    // This open with append should not affect the offset of f0.
    let _f1 = fs::OpenOptions::new()
        .create(true)
        .write(true)
        .append(true)
        .open("file")
        .unwrap();

    f0.write_all(b"d").unwrap();

    assert_eq!(fs::read_to_string("file").unwrap(), "adc");
}
```

### Expected behavior

Running this snippet should not panic.

### Actual behavior

Wasmer panics, producing a file with the content `abcd` instead of `adc`.

yagehu:
This PR fixes a case where opening a new file descriptor with the append flag makes previously opened non-append fds append as well.

The fix is relatively simple, since the append-ness of an fd is already tracked in the virtual fs implementation, there's no need to open the actual file with the append flag.

This behavior is consistent across Linux and other Wasm runtimes.

fixes #5161


maminrayej:
This PR seems to have some overlap with #5160. Let's first merge #5160 and then have only new changes in this PR.

yagehu:
> This PR seems to have some overlap with #5160. Let's first merge #5160 and then have only new changes in this PR.

@maminrayej Updated!

maminrayej (approved):
LGTM, just one question though. `handle.size()` (size of the file from the file's perspective) should be the same as `fd_entry.inode.stat.read().unwrap().st_size` (size of the file from the perspective of the inode). is there a reason you changed it? did the test fail with `handle.size()`?

yagehu:
> LGTM, just one question though. `handle.size()` (size of the file from the file's perspective) should be the same as `fd_entry.inode.stat.read().unwrap().st_size` (size of the file from the perspective of the inode). is there a reason you changed it? did the test fail with `handle.size()`?

Right. The test failed with `handle.size()` because it may be the old valid sometimes. I'm guessing because of async disk operation.

maminrayej:
Thanks.

PR #5184, opened 2024-10-28
fix(api/wamr): Build `wamr` on `iOS`

xdoardo:
A small patch that fixes a build error when targeting iOS (`aarch64-apple-ios`) while building the `wasmer` crate with the `wamr` feature enabled.

### DIFF
--- a/lib/api/build.rs
+++ b/lib/api/build.rs
@@ -94,7 +94,16 @@ fn main() {
         }
 
         if target_os == "ios" {
-            // XXX: Hacky..
+            // XXX: Hacky
+            //
+            // Compiling wamr targeting `aarch64-apple-ios` results in 
+            //
+            // ```
+            //  clang: error: unsupported option '-mfloat-abi=' for target 'aarch64-apple-ios'
+            // ```
+            // So, here, we simply remove that setting.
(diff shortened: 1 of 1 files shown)


lib/api/build.rs
```diff
@@ -85,14 +85,26 @@ fn main() {
...
+        if target_os == "windows" {
             dst.define("CMAKE_CXX_COMPILER", "cl.exe");
             dst.define("CMAKE_C_COMPILER", "cl.exe");
             dst.define("CMAKE_LINKER_TYPE", "MSVC");
             dst.define("WAMR_BUILD_PLATFORM", "windows");
             dst.define("WAMR_BUILD_LIBC_UVWASI", "0");
         }
 
+        if target_os == "ios" {
+            // XXX: Hacky..
```
syrusakbary, line 97: Can you open an issue on the wamr repo and link it here.
Also, add a better comment on why we need to do this

xdoardo: Sure, here it is: https://github.com/bytecodealliance/wasm-micro-runtime/pull/3889

syrusakbary (approved):

Issue #5217, opened 2024-11-05
Fix `library 'zstd' not found` aarch64 build issue with LLVM


xdoardo:
CI fails with this error:

```
= note: ld: warning: ignoring duplicate libraries: '-lc', '-lm'
          ld: library 'zstd' not found
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

PR #5220, opened 2024-11-05
fix(ci): Move addition of `brew` libs to other step

xdoardo:
A small patch that moves the addition of `brew` libs to another step. 
(Reference issue: https://github.com/wasmerio/wasmer/issues/5217)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -271,11 +271,6 @@ jobs:
           echo "${LLVM_DIR}/bin" >> $GITHUB_PATH
         env:
           LLVM_DIR: .llvm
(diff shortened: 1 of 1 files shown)


maminrayej (commented):
This adds homebrew libs for all macs (x86 and arm). Is this intended?

xdoardo:
In short, on `x86` system libraries installed by brew live in a different place than `arm64`, and for the former this addition should be superfluous and not damaging. Nonetheless, just to be sure, I'll move this addition to another different step. 

(Note: the previous PR checked against `matrix.metadata.os` which does not exists!)

maminrayej (approved):

PR #5235, opened 2024-11-08
c-api/README.md: Fix double include in usage section

winspool:
While reading the affected README, i saw, that the file "wasmer.h" was mentioned twice
in the list of installed include files. That's not correct.

Instead of creating an issue, i created this PR to change the README,
but I have never used or installed "wasmer", so i was not able to verify the result.

-- 
Regards ... Detlef


maminrayej (approved):

maminrayej:
Thanks.

PR #5243, opened 2024-11-12
Support multiple heterogeneous backends at the same time

xdoardo:
What's missing: 

- [x] jsc
- [x] wasmi (https://github.com/wasmi-labs/wasmi/pull/1312, https://github.com/wasmi-labs/wasmi/pull/1315)


lib/api/Cargo.toml
```diff
@@ -166,6 +168,7 @@ tar = "0.4.42"
 ureq = "2.10.1"
 xz = "0.1.0"
 zip = "2.2.0"
+which = "7.0.0"
```
theduke, line 171: Should this be an optional dependency, only enabled for the backends that require it?

lib/api/build.rs
```diff
@@ -1,178 +1,215 @@
...
+
+    if target_os == "windows" {
+        dst.define("CMAKE_CXX_COMPILER", "cl.exe");
+        dst.define("CMAKE_C_COMPILER", "cl.exe");
+        dst.define("CMAKE_LINKER_TYPE", "MSVC");
+        dst.define("WAMR_BUILD_PLATFORM", "windows");
+        dst.define("WAMR_BUILD_LIBC_UVWASI", "0");
+    }
+
+    //if target_os == "ios" {
```
theduke, line 98: So no iOS support for WAMR?

I don't think merging with WAMR only supported on some architectures is good... that create a feature matrix mess.
At the very least this should mention why it's commented out.

lib/api/build.rs
```diff
@@ -1,178 +1,215 @@
...
-            .write_to_file(out_path.join("bindings.rs"))
-            .expect("Couldn't write bindings");
+    let objcopy = objcopy.unwrap();
+
+    unsafe {
+        let syms: Vec<String> = WAMR_RENAMED
+            .iter()
+            .map(|(old, new)| format!("--redefine-sym=_{old}={new}"))
+            .collect();
+        dbg!(std::process::Command::new(objcopy)
```
theduke, line 190: Leftover dbg

lib/api/build.rs
```diff
@@ -1,178 +1,215 @@
...
+    println!("cargo:rustc-link-lib=wamr");
+}
 
-        let url = match (
+#[cfg(feature = "v8")]
+fn build_v8() {
+    use bindgen::callbacks::ParseCallbacks;
+    use std::{env, path::PathBuf};
+
+    let header_url = "https://raw.githubusercontent.com/laper32/v8-cmake/refs/heads/master/v8/third_party/wasm-api/wasm.h";
```
theduke, line 210: Build steps REALLY should not download random stuff at build time... can this be embedded in the repo?

I guess v8 is pretty huge?

lib/api/build.rs
```diff
@@ -185,63 +222,134 @@ fn main() {
...
 
-        bindings
-            .write_to_file(out_path.join("bindings.rs"))
-            .expect("Couldn't write bindings");
+    unsafe {
+        let syms: Vec<String> = WEE8_RENAMED
+            .iter()
+            .map(|(old, new)| format!("--redefine-sym=_{old}={new}"))
+            .collect();
+        dbg!(std::process::Command::new(objcopy)
```
theduke, line 336: Leftover dbg

lib/api/src/entities/engine/inner.rs
```diff
@@ -0,0 +1,150 @@
...
+                "The selected runtime does not support `deserialize_from_file`".into(),
+            )),
+        }
+    }
+}
+
+impl Default for RuntimeEngine {
+    fn default() -> Self {
+        #[cfg(feature = "sys")]
+        {
```
theduke, line 124: This code will probably give build warnings if multiple backends are enabled.

Probably either need to write that in a more if/else style, or ignore the warning.

lib/api/src/entities/engine/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+/// Create temporary handles to engines.
+mod engine_ref;
+
+/// The actual (private) definition of the engines.
+mod inner;
+pub(crate) use inner::RuntimeEngine;
+
+pub use engine_ref::*;
+
+static ENGINE_ID_COUNTER: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
```
theduke, line 22: nit, but I'd probably move the static and the next() function to be statics in the Engine struct impl.

lib/api/src/entities/engine/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+
+pub(crate) fn atomic_next_engine_id() -> u64 {
```
theduke, line 24: The ID should not be a plain u64.

We want to have the ability to change it.
It should be an opaque type like

```
#[derive(Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)]
struct EngineId(u64);
```

lib/api/src/entities/engine/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+impl Engine {
+    /// Returns the deterministic id of this engine.
+    pub fn deterministic_id(&self) -> &str {
+        self.rt.deterministic_id()
+    }
+
+    /// Returns the unique id of this engine.
+    pub fn id(&self) -> u64 {
```
theduke, line 53: opaque type

lib/api/src/entities/function/host/imp.rs
```diff
@@ -0,0 +1,730 @@
...
+                 Err(JSValue::string(&ctx, format!("panic: {:?}", panic)))
+             },
+         }
+
+     }
+
+     Some(fn_callback::<T, $( $x, )* Rets, RetsAsResult, Self > as _)
+   }
+
+  #[cfg(feature = "js")]
```
theduke, line 122: These impulse are quite a lot of code. 

Could they life in the backend-specific modules instead?

lib/api/src/entities/function/host/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+/// `VMFunctionCallback`.
+pub trait HostFunction<T, Args, Rets, Kind>
+where
+    Args: WasmTypeList,
+    Rets: WasmTypeList,
+    Kind: HostFunctionKind,
+{
+    #[cfg(feature = "jsc")]
+    /// Get the pointer to the function body for a given runtime.
+    fn jsc_function_callback(&self) -> crate::rt::jsc::vm::VMFunctionCallback;
```
theduke, line 20: Having backend-specific functions in a trait, that are enabled conditionally, seems like a huge hack.

Why is this needed?

theduke (commented):
This is only a first review round. 

This is a huge PR... so there will be other review rounds later.

Note: if the dbg!() s are intentional, we can leave them there, but I'd probably use an `eprintln!()` instead to differentiate from temporary test dbg!().

lib/api/build.rs
xdoardo: (I'll reply to the other comments soon)

WAMR does support iOS. This bit was a hack to remove the `-mfloat-abi=` parameter from the make file in WAMR itself. Should not be needed anymore after https://github.com/bytecodealliance/wasm-micro-runtime/pull/3889 got merged.

lib/api/build.rs
xdoardo: Yep, you're totally right. I forgot to embed that file in the repo.

lib/api/Cargo.toml
xdoardo: [0a0324e](https://github.com/wasmerio/wasmer/pull/5243/commits/0a0324e7cfc31dbbb2a1d3817409b630fc45bf9a)

lib/api/build.rs
xdoardo: [0a0324e](https://github.com/wasmerio/wasmer/pull/5243/commits/0a0324e7cfc31dbbb2a1d3817409b630fc45bf9a)

lib/api/src/entities/function/host/mod.rs
xdoardo: I agree, it's hacky and ugly. The problem is [this](https://github.com/wasmerio/wasmer/blob/5198-cli-support-wasm_c_api-backends-together-with-native-ones/lib/api/src/entities/function/mod.rs#L112) function - and the sister `_with_env` function - which has a constraint on the parameter `F` being a `HostFunction<...>`.

We could address this with per-runtime traits (e.g. `SysHostFunction`, `WamrHostFunction` and so on), but new problems would arise, in particular: what would be the signature of the top-level `Function::new_typed(_with_env)` function? In some way, we'll need to be able to make the `F` parameter into a `SysHostFunction` when actually calling the internal API to create the `Function` (`Sys::new_typed`, so to speak). If we keep it `HostFunction`: 
```
pub trait HostFunction<T, Args, Rets, Kind>
where
    Args: WasmTypeList,
    Rets: WasmTypeList,
    Kind: HostFunctionKind,
{
   /* empty */
}

pub trait SysHostFunction<T, Args, Rets, Kind>
where
    Args: WasmTypeList,
    Rets: WasmTypeList,
    Kind: HostFunctionKind,
{
   fn fn_callback(&self) -> ...;
}
....

/// Creates a new host `Function` from a native function.
where
        Args: WasmTypeList,
        Rets: WasmTypeList,
{ 
      ...
      cfg sys  {
         ...sys::Function::new_typed(store, f /* <cast to SysHostFunction if not automatic>/*)
      }
} 

// in sys
/// Creates a new host `Function` from a native function.
pub(crate) fn new_typed<F, Args, Rets>(store: &mut impl AsStoreMut, func: F) -> Self
where
        F: SysHostFunction<(), Args, Rets, WithoutEnv> + 'static + Send + Sync,
        Args: WasmTypeList,
        Rets: WasmTypeList,
{
    let cback = func.fn_callback();
    ....
}
```
We then have to implement something like 
```
impl< $( $x, )* Rets, RetsAsResult, Func> SysHostFunction<(), ( $( $x ),* ), Rets, WithoutEnv> for Func where
    $( $x: FromToNativeWasmType, )*
    Rets: WasmTypeList,
    RetsAsResult: IntoResult<Rets>,
    T: Send + 'static,
    Func: HostFunction<(), $($x),*), Rets, WithoutEnv> + 'static,
{
   fn fn_callback(&self) -> ...  {
       /* How do we keep track of Func being a callable object here? */ 
   }  
}
```
A possibility would be to make `HostFunction: Fn(Args) -> R` where `R: IntoResult<Rets>`, but I'm not sure it would work. What do you think? Probably, a fresh pair of eyes can see a completely different way to do it :)

lib/api/src/entities/engine/mod.rs
xdoardo: [3c54101](https://github.com/wasmerio/wasmer/pull/5243/commits/3c541018a7d44fda01ebd56c7defed032b29ef4e)

lib/api/src/entities/engine/mod.rs
xdoardo: [8e5ce03](https://github.com/wasmerio/wasmer/pull/5243/commits/8e5ce03a184cd0851ec4709e9cfdb6312d6726db)

lib/api/build.rs
xdoardo: [f381470](https://github.com/wasmerio/wasmer/pull/5243/commits/f3814701f8abb16d6481d2b91816583982b6220c)

lib/api/src/entities/engine/inner.rs
xdoardo: 1005cfc6f721b42dad2e6b519d5dc0f440618780

lib/api/src/entities/engine/inner.rs
```diff
@@ -0,0 +1,156 @@
...
+                "The selected runtime does not support `deserialize_from_file`".into(),
+            )),
+        }
+    }
+}
+
+impl Default for RuntimeEngine {
+    #[allow(unreachable_code)]
+    fn default() -> Self {
+        #[cfg(feature = "sys")]
```
syrusakbary, line 124: One thing is enabling a `sys` runtime, and other a default runtime being `sys`. I think we need a new feature that let you choose the default (eg: `sys-default`), that can't be activated when other default runtime (eg: `js-default`) or similar are activated (compiler error)

xdoardo: We already have a feature called `sys-default`. What would happen if `sys` is enabled but not `sys-default`?

syrusakbary: Basically, we should not have default for `RuntimeEngine` if there's no default defined

syrusakbary: If we don't want to have that because is a breaking change, then I'd have what you have now as a fallback, not as the main path

syrusakbary: Yup

xdoardo: [c111880](https://github.com/wasmerio/wasmer/pull/5243/commits/c11188021375a0e74cfb1fe9754c7524f40f0fbb)

lib/api/src/entities/function/host/imp.rs
xdoardo: [81f4195](https://github.com/wasmerio/wasmer/pull/5243/commits/81f41957f94a32924733020ad2373d0158fe60d1)

lib/api/src/entities/function/host/mod.rs
xdoardo: [81f4195](https://github.com/wasmerio/wasmer/pull/5243/commits/81f41957f94a32924733020ad2373d0158fe60d1)

Issue #45, opened 2018-12-10
usage intructions (run ngnix) fail


dominictarr:
I did this
```
curl https://get.wasmer.io -sSfL | sh
git clone https://github.com/wasmerio/wasmer.git
# sourcing the file as the output instructed me didn't work so I just ran it like this:
~/.wasmer/bin/wasmer run examples/nginx/nginx.wasm -- -p examples/nginx/ -c examples/nginx/nginx.conf 
```

and the output:
```
â  Compiling [======>                                 ] 16%  
â  Compiling [============>                           ] 32%  
â  Compiling [==============================>         ] 75% ^[[?64;1;2;6;9;15;18;21;22c
â  Compiling [====================================>   ] 91%  
Runtime error: trap at 0x0 - segmentation violation   ] 92% 
```
seems serious

this was on 64 bit linux. (archlinux)

syrusakbary:
We are still investigating into the segmentation violation issue, to get a fix asap.

handicraftsman:
Looks like this is still an issue.

syrusakbary:
This should be fixed in master.
This week we will publish a new version: wasmer 0.2.0 with all the improvements

handicraftsman:
Thanks, i will check it as soon as 0.2.0 is released

syrusakbary:
Wasmer 0.2.0 is now released.

You can update it locally via:

```bash
wasmer self-update
```

Or

```bash
curl https://get.wasmer.io -sSfL | sh
```

PR #5244, opened 2024-11-12
Alias `.` to cwd in wasi env

### LINKED ISSUE #5229: WASIX: Working Directory mismatch
There seems to be a problem with working directory handling in the runtime and in the WASIX libc.

To illustrate:

`wasmer --dir .` is supposed to mount the current dir at `/home` , and to set the working dir to `/home`.

```
wasmer run wasmer/python --dir . -- -c "import os; print(os.getcwd())"
> `/home`
```

This works correctly!

But:
`wasmer run wasmer/python --dir . -- -c "import os; os.makedirs('./test123)"`

This succeeds, but the directory is not created on the host file system.

Investigation reveals that `os.getcwd()` here leads to the `getcwd()` syscall being invoked, so it returns the expected value.
But the path_create syscall invoked actually uses the path `/test123`, so apth resolution logic in libc assumes the CWD to be `/`.

maminrayej:
Partially resolves #5229. The full fix requires the sync added in wasix-libc in [(#45)](https://github.com/wasix-org/wasix-libc/pull/45).

### DIFF
--- a/lib/wasix/src/runners/wasi_common.rs
+++ b/lib/wasix/src/runners/wasi_common.rs
@@ -73,7 +73,8 @@ impl CommonWasiOptions {
 
         if self.mounts.iter().all(|m| m.guest != ".") {
             // The user hasn't mounted "." to anything, so let's map it to "/"
-            builder.add_map_dir(".", "/")?;
+            let path = builder.get_current_dir().unwrap_or(PathBuf::from("/"));
+            builder.add_map_dir(".", path)?;
         }
 
         builder.set_fs(Box::new(fs));
--- a/lib/wasix/src/runners/wasi.rs
+++ b/lib/wasix/src/runners/wasi.rs
@@ -262,6 +262,18 @@ impl WasiRunner {
             None
         };
 
+        if self.wasi.is_home_mapped {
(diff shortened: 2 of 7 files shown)


lib/wasix/src/runners/wasi_common.rs
```diff
@@ -73,7 +73,8 @@ impl CommonWasiOptions {
 
         if self.mounts.iter().all(|m| m.guest != ".") {
             // The user hasn't mounted "." to anything, so let's map it to "/"
-            builder.add_map_dir(".", "/")?;
+            let path = builder.get_current_dir().unwrap_or(PathBuf::from("/"));
```
Arshia001, line 76: ah, this'll conflict with my PR. Also, I do believe my change (where I put `/` after `.`) will change how this behaves. @maminrayej let's make sure to account for that in whichever PR gets merged second.

maminrayej: I believe wasix-libc does maximal matching when trying to find the preopen dir. So even if `/` is before for example `/home`, `/home/foo` should get matched with the `/home` regardless, but yeah agreed will double check.

Arshia001 (approved):

Issue #3602, opened 2023-02-22
WASIX: Architecture-Independent StackSnapshot Alignment


theduke:
The `StackSnapshot` type used by wasix is currently defined as a `u128`, which creates alignment issues on different CPU architectures, for example because ARM has different alignment requirements.

The type should be replaced to be free of alignment issues, like a multiple `u32`.

This may also require updates to:

* WASIX Rust bindings
   https://github.com/john-sharratt/wasix
  (This currently has the type defined with two `u64` fields...
* WASIX libc
    https://github.com/wasmerio/wasix-libc

ptitSeb:
assign to syrus for later assignment.

ptitSeb:
Main type aligment done. Need some 2nd look on that later.

syrusakbary:
Not stale

syrusakbary:
This was fixed by #5259 . Closing the issue

PR #5259, opened 2024-11-18
Fix alignment and padding of the WASIX snapshot type to align with wasix-libc

Arshia001:
This fixes the PHP crash when going through `zend_catch`.

### DIFF
--- a/lib/wasi-types/src/wasi/wasix_manual.rs
+++ b/lib/wasi-types/src/wasi/wasix_manual.rs
@@ -108,7 +108,24 @@ pub union EventUnion {
 #[repr(C)]
 pub struct StackSnapshot {
     pub user: u64,
-    pub hash: u128,
+    // The hash is defined as two u64s in wasix-libc, so we must do the same
+    // here to make sure the alignment and padding are the same.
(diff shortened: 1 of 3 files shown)


Arshia001:
Incidentally, this fixes #3602.

maminrayej (approved):
I would use `u64::MAX` instead of the handwritten hex value to make it more readable, otherwise LGTM.

lib/wasi-types/src/wasi/wasix_manual.rs
```diff
@@ -108,7 +108,24 @@ pub union EventUnion {
...
+    // The hash is defined as two u64s in wasix-libc, so we must do the same
+    // here to make sure the alignment and padding are the same.
+    pub hash_lower: u64,
+    pub hash_upper: u64,
+}
+
+impl StackSnapshot {
+    pub fn new(user: u64, hash: u128) -> Self {
+        Self {
+            user,
```
theduke, line 120: Let's add for the conversion.

theduke (changes requested):

theduke (approved):

PR #5283, opened 2024-12-05
Better network sandbox

### LINKED ISSUE #5280: Allow passing explicit flags for --net
We should allow having things like:

```
wasmer run xyz --net DOMAIN,IP,IP-RANGE,IP-PORT
```

Example:

```
wasmer run xyz --net=xyz.com,1.2.3.4,1.2.3.4/20,1.2.3.4:80,1.2.3.4/20:80
```

Note:

* Domains:
  * xyz.com is a valid domain
  * abc.xyz.com is a valid domain
  * \*.xyz.com is a valid domain
* IP:
  * 1.2.3.4 is a valid one (single ip)
  * 1.2.3.4/20 is a valid one (ip range)
  * 1.2.3.4/20:80 is a valid one (ip range, one port)
  * 1.2.3.4:80 is a valid one (single ip, single port)
  * 1.2.3.4:80-100 is a valid one (single ip, port range)
  * 1.2.3.4:\* is a valid one (single ip, any port)
  * \*:80 is a valid one (any ip, port 80)
  * *:* is a valid one (any ip, any port)
  * *:80-1000 (port range)

maminrayej:
## Overview
This PR adds a more fine-grained sandbox for the network layer. Previously, the wasm module did not have access to any networking functionality by default and the user could give this access by providing the `--net` flag to the wasmer client. But this would expose all of host's network stack to the wasm module. Now users can provide a list of  of patterns that will act as a whitelist or blacklist. This is done by adding the new `--net-filter` flag:
```
wasmer run --net-filter=<comma_separated_list_of_rules>
```

## Rule Specification
Each member of the `<comma_separated_list_of_rules>` can be expressed like:
```
<rule_kind>:<rule_action>=<rule_expr>

<rule_kind>: dns, ipv4, ipv6

<rule_action>: allow | deny

dns:
<rule_expr>:
{<domain_spec>}:{<port_spec>} (this will be expanded to an outbound IP rule)
<domain_spec>: domain | domain glob | *

ipv4:
<rule_expr>:
<ipv4_specs>:<port_specs>/<in|out>
<ipv4_specs>: <ipv4_spec> | {<ipv4_spec>,}
<ipv4_spec>: ipv4 | ipv4_range | *

ipv6:
<rule_expr>:
<ipv6_specs>:<port_specs>/<in|out>
<ipv6_specs>: <ipv6_spec> | {<ipv6_spec>,}
<ipv6_spec>: ipv6 | ipv6_range | *

<port_specs>: <port_spec> | {<port_specs>,}
<port_spec>: port | start_port-end_port | *
```
Some examples:
- Allow a specific domain and port: `dns:allow=example.com:80`
- Deny a domain and all its subdomains on all ports: `dns:deny=*danger.xyz:*`
- Allow opening ipv4 sockets only on a specific IP and port: `ipv4:allow=127.0.0.1:80/in`.

### Whitelisting and Blacklisting
Each rule can be expressed as an `allow` (whitelist) or `deny` (blacklist). A socket or domain is only accessible if at least one rule whitelists it and no rule blacklists it.

### Directional Filtering
IP based rules can be either directional by specifying `/in` or `/out` postfixes to the rule, or bidirectional which is the default setting for these rules.

### Rule Combination
In order to prevent repetition, the parts before and after the `:` could hold multiple values. For example:
```
ipv4:deny={127.0.0.1/24, 192.168.1.1/24}:{80, 443}
```
This is equivalent to:
```
ipv4:deny=127.0.0.1/24:80,
ipv4:deny=127.0.0.1/24:443,
ipv4:deny=192.168.1.1/24:80,
ipv4:deny=192.168.1.1/24:443
```

Resolves #5280.

### DIFF
--- a/lib/cli/src/commands/run/wasi.rs
+++ b/lib/cli/src/commands/run/wasi.rs
@@ -12,7 +12,7 @@ use clap::Parser;
 use tokio::runtime::Handle;
 use url::Url;
 use virtual_fs::{DeviceFile, FileSystem, PassthruFileSystem, RootFileSystemBuilder};
-use virtual_net::ruleset::RuleSet;
+use virtual_net::ruleset::Ruleset;
 use wasmer::{Engine, Function, Instance, Memory32, Memory64, Module, RuntimeError, Store, Value};
 use wasmer_config::package::PackageSource as PackageSpecifier;
 use wasmer_types::ModuleHash;
@@ -566,7 +566,7 @@ impl Wasi {
             .networking
             .clone()
             .flatten()
-            .map(|ruleset| RuleSet::from_str(&ruleset))
+            .map(|ruleset| Ruleset::from_str(&ruleset))
             .transpose()?;
 
         let network = if let Some(ruleset) = ruleset {
--- a/lib/cli/src/commands/run/wasi.rs
+++ b/lib/cli/src/commands/run/wasi.rs
@@ -9,7 +9,6 @@ use std::{
 use anyhow::{bail, Context, Result};
 use bytes::Bytes;
 use clap::Parser;
-use futures::future::FlattenSink;
 use tokio::runtime::Handle;
 use url::Url;
 use virtual_fs::{DeviceFile, FileSystem, PassthruFileSystem, RootFileSystemBuilder};
--- a/lib/cli/src/commands/run/wasi.rs
+++ b/lib/cli/src/commands/run/wasi.rs
@@ -103,7 +103,21 @@ pub struct Wasi {
     #[clap(long = "net")]
     pub networking: bool,
 
-    /// Define a set of filters to control the network sandbox
+    /// Define a set of network filters
+    ///
+    /// Allows fine-grained control over the network sandbox.
+    ///
+    /// Syntax:
+    ///
+    /// <rule-type>:<allow|deny>=<rule-expression>
+    ///
+    /// Examples:
+    ///
+    ///  - Allow a specific domain and port: dns:allow=example.com:80
+    ///
+    ///  - Deny a domain and all its subdomains on all ports: dns:deny=*danger.xyz:*
+    ///
+    ///  - Allow opening ipv4 sockets only on a specific IP and port: ipv4:allow=127.0.0.1:80/in.
     #[clap(long, requires = "networking")]
     pub net_filter: Option<String>,
 
--- a/lib/cli/src/commands/run/wasi.rs
+++ b/lib/cli/src/commands/run/wasi.rs
@@ -9,6 +9,7 @@ use std::{
 use anyhow::{bail, Context, Result};
 use bytes::Bytes;
 use clap::Parser;
+use futures::future::FlattenSink;
 use tokio::runtime::Handle;
 use url::Url;
 use virtual_fs::{DeviceFile, FileSystem, PassthruFileSystem, RootFileSystemBuilder};
@@ -100,14 +101,11 @@ pub struct Wasi {
     /// Enable networking with the host network.
     ///
     /// Allows WASI modules to open TCP and UDP connections, create sockets, ...
-    #[clap(long = "net")]
-    pub networking: bool,
-
-    /// Define a set of network filters
     ///
-    /// Allows fine-grained control over the network sandbox.
+    /// Optionally, a set of network filters could be defined which allows fine-grained
+    /// control over the network sandbox.
     ///
-    /// Syntax:
+    /// Rule Syntax:
     ///
     /// <rule-type>:<allow|deny>=<rule-expression>
     ///
@@ -118,8 +116,10 @@ pub struct Wasi {
     ///  - Deny a domain and all its subdomains on all ports: dns:deny=*danger.xyz:*
     ///
     ///  - Allow opening ipv4 sockets only on a specific IP and port: ipv4:allow=127.0.0.1:80/in.
-    #[clap(long, requires = "networking")]
-    pub net_filter: Option<String>,
+    #[clap(long = "net", require_equals = true)]
+    // Note that when --net is passed to the cli, the first Option will be initialized: Some(None)
+    // and when --net=<ruleset> is specified, the inner Option will be initialized: Some(Some(ruleset))
+    pub networking: Option<Option<String>>,
 
     /// Disables the TTY bridge
     #[clap(long = "no-tty")]
@@ -575,14 +575,15 @@ impl Wasi {
         let tokio_task_manager = Arc::new(TokioTaskManager::new(rt_or_handle.into()));
         let mut rt = PluggableRuntime::new(tokio_task_manager.clone());
 
-        let has_networking = self.networking
+        let has_networking = self.networking.is_some()
             || capabilities::get_cached_capability(pkg_cache_path)
                 .ok()
                 .is_some_and(|v| v.enable_networking);
 
         let ruleset = self
-            .net_filter
+            .networking
             .clone()
+            .flatten()
             .map(|ruleset| Ruleset::from_str(&ruleset))
             .transpose()?;
 
--- a/lib/cli/src/commands/run/wasi.rs
+++ b/lib/cli/src/commands/run/wasi.rs
@@ -101,7 +101,11 @@ pub struct Wasi {
     ///
     /// Allows WASI modules to open TCP and UDP connections, create sockets, ...
     #[clap(long = "net")]
-    pub networking: Option<Option<String>>,
+    pub networking: bool,
+
+    /// Define a set of filters to control the network sandbox
+    #[clap(long, requires = "networking")]
+    pub net_filter: Option<String>,
 
     /// Disables the TTY bridge
     #[clap(long = "no-tty")]
@@ -557,15 +561,14 @@ impl Wasi {
         let tokio_task_manager = Arc::new(TokioTaskManager::new(rt_or_handle.into()));
         let mut rt = PluggableRuntime::new(tokio_task_manager.clone());
 
-        let has_networking = self.networking.is_some()
+        let has_networking = self.networking
             || capabilities::get_cached_capability(pkg_cache_path)
                 .ok()
                 .is_some_and(|v| v.enable_networking);
 
         let ruleset = self
-            .networking
+            .net_filter
             .clone()
-            .flatten()
             .map(|ruleset| Ruleset::from_str(&ruleset))
             .transpose()?;
 
--- a/lib/cli/src/commands/run/wasi.rs
+++ b/lib/cli/src/commands/run/wasi.rs
@@ -12,6 +12,7 @@ use clap::Parser;
 use tokio::runtime::Handle;
 use url::Url;
 use virtual_fs::{DeviceFile, FileSystem, PassthruFileSystem, RootFileSystemBuilder};
+use virtual_net::ruleset::RuleSet;
 use wasmer::{Engine, Function, Instance, Memory32, Memory64, Module, RuntimeError, Store, Value};
 use wasmer_config::package::PackageSource as PackageSpecifier;
 use wasmer_types::ModuleHash;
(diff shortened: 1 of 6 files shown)


ibuildthecloud:
Is there a way to allow all domains? Basically in the syntax what does * mean, or it is even valid?  Does * mean any domain and : is any IP?

ibuildthecloud:
We couldn't figure out a way to allow only specific IPs and all domains. For example we wanted to allow *:80, *:443. This ends up not being possible because we need to add additionally some pattern for domains like *.com.

lib/cli/src/commands/run/wasi.rs
```diff
@@ -100,7 +101,7 @@ pub struct Wasi {
     ///
     /// Allows WASI modules to open TCP and UDP connections, create sockets, ...
     #[clap(long = "net")]
-    pub networking: bool,
+    pub networking: Option<Option<String>>,
```
theduke, line 104: What is `Option<Option<String>>` supposed to mean?

This at the very least needs a comment and a doc comment explanation with examples.
I don't see how `Option<Option<>` has any meaning with clap also?

lib/virtual-net/Cargo.toml
```diff
@@ -14,6 +14,8 @@ base64.workspace = true
 hyper = { workspace = true, optional = true }
 rkyv = { workspace = true, optional = true }
 
+ipnet = "2.10.1"
```
theduke, line 17: Do we really need two new dependencies?
We already have enough of them...

lib/virtual-net/src/host.rs
```diff
@@ -132,6 +166,13 @@ impl VirtualNetworking for LocalNetworking {
         port: Option<u16>,
         dns_server: Option<IpAddr>,
     ) -> Result<Vec<IpAddr>> {
+        if let Some(ruleset) = self.ruleset.as_ref() {
+            if !ruleset.matches_domain(host) {
+                tracing::warn!(%host, "ruleset matching failed");
+                return Err(NetworkError::PermissionDenied);
+            }
```
theduke, line 173: The resolved IPs have to be added to the whitelist.

lib/virtual-net/src/lib.rs
```diff
@@ -7,6 +7,7 @@ pub mod composite;
 pub mod host;
 pub mod loopback;
 pub mod meta;
+pub mod ruleset;
 #[cfg(feature = "remote")]
```
theduke, line 11: Consider adding a `FirewallNetworking<T>` which applies rules. 
This is optional.

Note: one annoyance is that you would also have to add wrapping socket implementations.

theduke (changes requested):

ibuildthecloud:
I don't think the current design will satisfy these two common use cases.

Use case 1
- I want to only allow access to example.com

Use case 2
- I want to only allow access to *:80, *:443

The first fails because there is no way to allow all IPs but just specific domains.
The second falls because there is no way to allow only IPs but all domains.

lib/cli/src/commands/run/wasi.rs
maminrayej: The first `Option` will be initialized then only `--net` is specified: `Some(None)` and the inner `Option` will be initialized when `--net=<ruleset>` is specified: `Some(Some(ruleset))`.

I tried a bunch of other patterns but neither worked the same way. I can add the same explanation above to the code to make it clear.

lib/virtual-net/Cargo.toml
maminrayej: I prefer not to implement IP-range logic (such as `contains`) myself if I can use a crate for it.

lib/virtual-net/src/host.rs
maminrayej: Done

lib/virtual-net/src/lib.rs
maminrayej: As discussed, we decided to not do this for now since methods like `try_accept` cannot be protected from outside and we need the inner network to actually run and accept the connection which makes enforcing the ruleset a bit weird and dependent on the implementation details of the inner network.

lib/virtual-net/src/ruleset.rs
```diff
@@ -0,0 +1,1388 @@
...
+            ("3001:db8::".parse::<Ipv6Addr>().unwrap(), 8080),
+            Direction::Inbound
+        ));
+        assert!(ruleset.is_socket_allowed(
+            ("3001:db8::".parse::<Ipv6Addr>().unwrap(), 8080),
+            Direction::Inbound
+        ));
+    }
+
+    #[test]
```
theduke, line 1375: Leftover debugging test?

lib/virtual-net/src/ruleset.rs
```diff
@@ -0,0 +1,1388 @@
...
+
+        let (expr, rem) = input.split_at(next_entry);
+
+        let rem = rem.strip_prefix(',').unwrap_or(rem);
+
+        Ok((RuleExpr(expr.to_string()), rem))
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
```
theduke, line 542: What's a "segment"?
Doc comments!

lib/virtual-net/src/ruleset.rs
```diff
@@ -0,0 +1,1388 @@
...
+struct RulesetSegment {
+    ty: RuleType,
+    action: RuleAction,
+    expr: RuleExpr,
+}
+
+fn ruleset_segments(s: impl AsRef<str>) -> Result<Vec<RulesetSegment>, RulesetError> {
```
theduke, line 549:
```suggestion
fn parse_ruleset_segments(s: impl AsRef<str>) -> Result<Vec<RulesetSegment>, RulesetError> {
```

lib/virtual-net/src/ruleset.rs
```diff
@@ -0,0 +1,1388 @@
...
+//
+// ipv6:
+// <rule_expr>:
+// {<ipv6_spec>}:{<port_spec>}/<in|out>
+// <ipv4_spec>: ipv6 | ipv6_range | *
+//
+// <port_spec>: port | start_port-end_port | *
+
+#[derive(Debug, thiserror::Error)]
+pub enum RulesetError {
```
theduke, line 33:
```suggestion
pub enum RuleParseError {
```

lib/virtual-net/src/ruleset.rs
```diff
@@ -0,0 +1,1388 @@
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
```
theduke, line 1: Add doc comment explaining what this is for, and giving an overview of the syntax.

lib/virtual-net/src/ruleset.rs
```diff
@@ -0,0 +1,1388 @@
...
+// <rule_expr>:
+// {<ipv6_spec>}:{<port_spec>}/<in|out>
+// <ipv4_spec>: ipv6 | ipv6_range | *
+//
+// <port_spec>: port | start_port-end_port | *
+
+#[derive(Debug, thiserror::Error)]
+pub enum RulesetError {
+    #[error("invalid connection direction: {0}")]
+    DirectionParsingError(String),
```
theduke, line 35:
```suggestion
    Direction(String),
```

No need to repeat Error in the variants, also, all of them are parse errors.

lib/virtual-net/src/ruleset.rs
```diff
@@ -0,0 +1,1388 @@
...
+        Ok(Self {
+            rules: Arc::new(RwLock::new(rules)),
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
```
theduke, line 695: Good to add some negative tests that assert failure cases.

lib/virtual-net/src/host.rs
```diff
@@ -65,6 +76,13 @@ impl VirtualNetworking for LocalNetworking {
         reuse_port: bool,
         reuse_addr: bool,
     ) -> Result<Box<dyn VirtualTcpListener + Sync>> {
+        if let Some(ruleset) = self.ruleset.as_ref() {
+            if !ruleset.is_socket_allowed(addr, Direction::Inbound) {
+                tracing::warn!(%addr, "ruleset matching failed");
```
theduke, line 81: More descriptive error, like `listen_tcp blocked by firewall rule`

theduke: Same for the other logs below.

lib/virtual-net/src/ruleset.rs
```diff
@@ -0,0 +1,1388 @@
...
+pub struct Ruleset {
+    rules: Arc<RwLock<Vec<Rule>>>,
+}
+
+impl Ruleset {
+    pub fn is_socket_allowed(&self, addr: impl Into<SocketAddr>, dir: Direction) -> bool {
+        let addr = addr.into();
+
+        let ruleset = self.rules.read().unwrap();
+        let is_whitelisted = ruleset.iter().any(|r| r.is_socket_allowed(addr, dir));
```
theduke, line 576: Small perf improvement: no need to process whitelist rules if a deny rule matches.

lib/virtual-net/src/ruleset.rs
```diff
@@ -0,0 +1,1388 @@
...
+        is_whitelisted && !is_blacklisted
+    }
+
+    pub fn is_domain_allowed(&self, domain: impl AsRef<str>) -> bool {
+        let domain = domain.as_ref();
+
+        let ruleset = self.rules.read().unwrap();
+        let is_whitelisted = ruleset.iter().any(|r| r.is_domain_allowed(domain));
+        let is_blacklisted = ruleset.iter().any(|r| r.is_domain_blocked(domain));
+        drop(ruleset);
```
theduke, line 589: As above

lib/virtual-net/src/ruleset.rs
```diff
@@ -0,0 +1,1388 @@
...
+}
+
+impl Ruleset {
+    pub fn is_socket_allowed(&self, addr: impl Into<SocketAddr>, dir: Direction) -> bool {
+        let addr = addr.into();
+
+        let ruleset = self.rules.read().unwrap();
+        let is_whitelisted = ruleset.iter().any(|r| r.is_socket_allowed(addr, dir));
+        let is_blacklisted = ruleset.iter().any(|r| r.is_socket_blocked(addr, dir));
+        drop(ruleset);
```
theduke, line 578: nit: this isn't really needed, since all that follows is a single bool comparison.

But, for future reference, it's super easy to lose the `drop` during refactors, or do changes that extend the lock lifetime.

What I like to do is a nested scope:

```rust
let x = {
  // Nested scope to limit lock lifetime.
  let lock = self.y.lock().unwrap();
  ...
};
```

lib/virtual-net/src/ruleset.rs
```diff
@@ -0,0 +1,1388 @@
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
+use std::ops::RangeInclusive;
+use std::str::FromStr;
```
theduke, line 3: All public types and functions should have doc comments

lib/cli/src/commands/run/wasi.rs
maminrayej: We decided to opt for a dedicated flag: `--net-filter` for the ruleset specification.

lib/cli/src/commands/run/wasi.rs
```diff
@@ -102,6 +103,10 @@ pub struct Wasi {
     #[clap(long = "net")]
     pub networking: bool,
 
+    /// Define a set of filters to control the network sandbox
```
theduke, line 106: This needs an explanation and examples for the rules syntax.

lib/cli/src/commands/run/wasi.rs
```diff
@@ -102,6 +103,10 @@ pub struct Wasi {
...
+    #[clap(long, requires = "networking")]
```
theduke, line 107: Does `--net-filter` implicitly also enable networking (`--net`)?
Makes sense to do so, but should be documented.

maminrayej: No the `requires` attribute will make it throw an error if the user did not also pass `--net` since a ruleset without a network permission does not make sense.

theduke (approved):

PR #5292, opened 2024-12-09
Pass readme content instead of readme file name

ayys:
the readme field in wasmer.toml is supposed to be a pathbuf, but it is
being passed to the backend as a string. This should be the content
of the readme file, and not the file path itself.

the commit fixes this by reading the file, and passing the content
correctly.


xdoardo (approved):

syrusakbary (approved):

lib/cli/src/commands/package/tag.rs
```diff
@@ -200,6 +200,18 @@ impl PackageTag {
             None
         };
 
+        let maybe_readme_content = match maybe_readme {
+            Some(readme) => {
+                let readme_path = Path::new(&readme);
```
syrusakbary, line 205: This path has to be relative to where the manifest lives. I don't think we are accounting for that on this code

syrusakbary (changes requested):
Please address last comment

syrusakbary (approved):

PR #5293, opened 2024-12-10
LoongArch: Switch to the medium code model

heiher:
This patch switches LoongArch to the `medium` code model, achieving a balance between performance and addressing capability. It also resolves the issue where the `large` code model generates GOT-based function accesses, making it challenging to achieve proper linking.


xdoardo:
LGTM. Thanks!

xdoardo (approved):

Issue #5247, opened 2024-11-14
The runtime does not allow to reuse socket address for UDP connections


rodjjo:
### Describe the bug

The runtime exposes functions to allow us to set SO_REUSEADDR and SO_REUSEPORT, however it's not working:
```sh
wasmer run --net examples/rust_udp/target/wasm32-wasmer-wasi/debug/udp_example.rustc.wasm
Starting UDP server
Binding to 0.0.0.0:8080
Error: Os { code: 3, kind: AddrInUse, message: "Address in use" }
```

The example code:
```rust
use tokio::net::UdpSocket;
use std::io;
use std::net::{SocketAddr};
use socket2::{Socket, Domain, Type};


#[tokio::main]
async fn main() -> io::Result<()> {
    println!("Starting UDP server");
    let std_sock = Socket::new(Domain::IPV4, Type::DGRAM, None)?;
    std_sock.set_reuse_address(true)?;

    println!("Binding to 0.0.0.0:8080");

    let address: SocketAddr = "0.0.0.0:8080".parse().unwrap();
    let address = address.into();
    std_sock.bind(&address)?;
    println!("Binding to 0.0.0.0:8080");
   
    let sock = UdpSocket::from_std(std_sock.into())?;

    let mut buf = [0; 1024];
    loop {
        let (len, addr) = sock.recv_from(&mut buf).await?;
        println!("{:?} bytes received from {:?}", len, addr);

        let len = sock.send_to(&buf[..len], addr).await?;
        println!("{:?} bytes sent", len);
    }
}
```

Libraries in use (cargo.toml):
```text
[package]
name = "udp_example"
version = "0.1.0"
edition = "2021"

[dependencies]
 
socket2 = { git = "https://github.com/wasix-org/socket2.git", branch = "master" }
tokio = { git = "https://github.com/wasix-org/tokio.git", branch = "wasix-1.35.1", version = "=1.35.1", features = ["fs", "io-util", "io-std","net", "rt",  "signal", "macros", "rt", "rt-multi-thread", "signal", "sync", "time", "socket2" ] }
```

### Steps to reproduce

Compile the application and try to open two instances of it using wasmer.

### Actual behavior

It shows an error:
```text
Starting UDP server
Binding to 0.0.0.0:8080
Error: Os { code: 3, kind: AddrInUse, message: "Address in use" }
```

### Additional context

Journal report:
```text
init-module (hash=[cc, 76, 0, fc, cb, e3, 44, c8])
[...]

PR #5312, opened 2024-12-19
 Forward reuse_port and reuse_addr when it's possible to

rodjjo:
The runtime exposes functions to allow us to set SO_REUSEADDR and SO_REUSEPORT, however it's not working:
```sh
wasmer run --net examples/rust_udp/target/wasm32-wasmer-wasi/debug/udp_example.rustc.wasm
Starting UDP server
Binding to 0.0.0.0:8080
Error: Os { code: 3, kind: AddrInUse, message: "Address in use" }
```

The example code:
```rust
use tokio::net::UdpSocket;
use std::io;
use std::net::{SocketAddr};
use socket2::{Socket, Domain, Type};


#[tokio::main]
async fn main() -> io::Result<()> {
    println!("Starting UDP server");
    let std_sock = Socket::new(Domain::IPV4, Type::DGRAM, None)?;
    std_sock.set_reuse_address(true)?;

    println!("Binding to 0.0.0.0:8080");

    let address: SocketAddr = "0.0.0.0:8080".parse().unwrap();
    let address = address.into();
    std_sock.bind(&address)?;
   
    let sock = UdpSocket::from_std(std_sock.into())?;

    let mut buf = [0; 1024];
    loop {
        let (len, addr) = sock.recv_from(&mut buf).await?;
        println!("{:?} bytes received from {:?}", len, addr);

        let len = sock.send_to(&buf[..len], addr).await?;
        println!("{:?} bytes sent", len);
    }
}
```

This pull request intend to solve [issue-5247](https://github.com/wasmerio/wasmer/issues/5247)

**I need help to understand the impact of the changes on other platforms, I tested it in my use case and I was able to reuse the address when I ran Wasmer on Ubuntu.**

### DIFF
--- a/lib/virtual-net/src/host.rs
+++ b/lib/virtual-net/src/host.rs
@@ -103,7 +103,7 @@ impl VirtualNetworking for LocalNetworking {
     async fn bind_udp(
         &self,
         addr: SocketAddr,
-        _reuse_port: bool,
+        reuse_port: bool,
         reuse_addr: bool,
     ) -> Result<Box<dyn VirtualUdpSocket + Sync>> {
         use socket2::{Socket, Domain, Type};        
@@ -116,12 +116,15 @@ impl VirtualNetworking for LocalNetworking {
         }
 
         let std_sock = Socket::new(Domain::IPV4, Type::DGRAM, None).map_err(io_err_into_net_error)?;
-        std_sock.set_reuse_address(reuse_addr).map_err(io_err_into_net_error)?;
-        //std_sock.set_reuse_port(reuse_port).map_err(io_err_into_net_error)?;
-
-        std_sock.bind(&addr.into()).map_err(io_err_into_net_error)?;
-
-        let socket = mio::net::UdpSocket::from_std(std_sock.into());
+        #[cfg(not(windows))]
+        let socket = {
+            std_sock.set_reuse_address(reuse_addr).map_err(io_err_into_net_error)?;
+            std_sock.set_reuse_port(reuse_port).map_err(io_err_into_net_error)?;
+            std_sock.bind(&addr.into()).map_err(io_err_into_net_error)?;
+            mio::net::UdpSocket::from_std(std_sock.into())
+        };
+        #[cfg(windows)]
+        let socket = mio::net::UdpSocket::bind(addr).map_err(io_err_into_net_error)?;
 
         #[allow(unused_mut)]
         let mut ret = LocalUdpSocket {
--- a/lib/virtual-net/src/host.rs
+++ b/lib/virtual-net/src/host.rs
@@ -106,7 +106,7 @@ impl VirtualNetworking for LocalNetworking {
         reuse_port: bool,
         reuse_addr: bool,
     ) -> Result<Box<dyn VirtualUdpSocket + Sync>> {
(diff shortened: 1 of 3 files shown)


rodjjo:
@maminrayej what do you think, would be that the way ?

maminrayej (commented):
LGTM, although it needs a test to pin this behavior. You could try to add a [WASIX](https://github.com/wasmerio/wasmer/tree/main/tests/wasix) test by looking at other tests and how they are set up.

As a bonus (not required for the PR to be merged), you could fix the same issue in other places such as [listen_tcp](https://github.com/wasmerio/wasmer/blob/3c32c74685063a02ad33fb09e557b1ac43bf5c29/lib/virtual-net/src/host.rs#L69-L101).

lib/virtual-net/src/host.rs
```diff
@@ -103,16 +103,32 @@ impl VirtualNetworking for LocalNetworking {
...
+        use socket2::{Domain, Socket, Type};
+
         if let Some(ruleset) = self.ruleset.as_ref() {
             if !ruleset.allows_socket(addr, Direction::Inbound) {
                 tracing::warn!(%addr, "bind_udp blocked by firewall rule");
                 return Err(NetworkError::PermissionDenied);
             }
         }
 
+        #[cfg(not(windows))]
```
syrusakbary, line 118: Any reason why we are not using the same logic for Windows? It seems that the API for socket reusing is also available there, so the more commonalities between systems, the better

syrusakbary:
Hey @rodjjo, we are happy to merge the PR once we make the logic with Windows similar as with Unix and hopefully a test that verifies the fix to make sure we are shielded of this issue in the future.

Let me know if you need any help with that

xdoardo:
@rodjjo Are you still interested in moving this forward?

rodjjo:
> @rodjjo Are you still interested in moving this forward?

I'm going to update it this week.

lib/virtual-net/src/host.rs
rodjjo: https://github.com/wasmerio/wasmer/actions/runs/13051954340/job/36414084837?pr=5312

I think the dependence does not support it on windows.

rodjjo:
> LGTM, although it needs a test to pin this behavior. You could try to add a [WASIX](https://github.com/wasmerio/wasmer/tree/main/tests/wasix) test by looking at other tests and how they are set up.
> 
> As a bonus (not required for the PR to be merged), you could fix the same issue in other places such as [listen_tcp](https://github.com/wasmerio/wasmer/blob/3c32c74685063a02ad33fb09e557b1ac43bf5c29/lib/virtual-net/src/host.rs#L69-L101).

I'm going to create another pull request for the listen_tcp issue.

rodjjo:
> Hey @rodjjo, we are happy to merge the PR once we make the logic with Windows similar as with Unix and hopefully a test that verifies the fix to make sure we are shielded of this issue in the future.
> 
> Let us know if you need any help with that

image
I created a simple tests that tries to reuse the same address and the same port by creating two sockets and binding the address and port to check if it has success or not.

syrusakbary:
Looks good. Test failing has nothing to do with this PR. Merging

PR #5319, opened 2025-01-02
Fix partial writes to host files when app shuts down

Arshia001:
* Flush files before closing
* Make sure files are actually closed when fd_close is called


Arshia001:
The approach implemented here breaks cross-process shared FDs, and is not suitable for merging.

PR #5323, opened 2025-01-04
Close host files when the last FD referencing them is closed

Arshia001:


lib/wasix/src/fs/mod.rs
```diff
@@ -121,20 +121,62 @@ impl Inode {
...
+    pub fn handle_count(&self) -> u32 {
+        self.open_handles.load(Ordering::SeqCst) as u32
+    }
+
+    pub fn acquire_handle(&self) {
+        self.open_handles.fetch_add(1, Ordering::SeqCst);
+    }
+
+    pub fn drop_one_handle(&self) {
+        if self.open_handles.fetch_sub(1, Ordering::SeqCst) != 1 {
```
theduke, line 157: Should this be `<= 1` to be safe?

Arshia001: Since it's an atomic, I believe it's impossible for the `1` value to go unobserved. Indeed, I believe when I was looking at the `Arc` sources from std, they also use a `== 1` test.

lib/wasix/src/fs/mod.rs
```diff
@@ -146,6 +188,10 @@ impl std::ops::Deref for InodeGuard {
 #[derive(Debug, Clone)]
 pub struct InodeWeakGuard {
     ino: Inode,
+    // Needed for when we want to upgrade back. We don't exactly
+    // care too much when the AtomicI32 is dropped, so this is
+    // a strong reference to keep things simple.
+    open_handles: Arc<AtomicI32>,
```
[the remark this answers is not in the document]
Arshia001: For just the weak struct, or the main one as well? I'm honestly not too obsessed with keeping the code readable at this point, but I suppose we can do that.

theduke (changes requested):
I'm quite against abusing the notion of unsafe here to emulate some notion of scoping. 

Why don't we do things properly by providing a way to get a borrow to the inner content, while making the public fields private to prevent misuse?

Yes, the file system is not in a great state, but adding a huge hack on top of other hacks is bound to lead to more issues in the future.

Arshia001:
> Why don't we do things properly by providing a way to get a borrow to the inner content, while making the public fields private to prevent misuse?

Mostly because it'll have rippling effects across the code. The FS code is not very well encapsulated, and adding proper encapsulation requires non-trivial amounts of refactoring. I'm not a huge fan of this either, but I'm trying to minimise the amount of time we spend refactoring code we know has to go as soon as possible.

I'll try to see if we can't make a small refactor.

theduke (approved):

PR #5334, opened 2025-01-10
Implement app.yaml job schema

Arshia001:


lib/config/src/app/job.rs
```diff
@@ -0,0 +1,239 @@
...
+    serde::Serialize, serde::Deserialize, schemars::JsonSchema, Clone, Debug, PartialEq, Eq,
+)]
+pub struct ExecutableJob {
+    /// The package that contains the command to run. Defaults to the app config's package.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    package: Option<PackageSource>,
+
+    /// The command to run. Defaults to the package's entrypoint.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    command: Option<String>,
```
ayys, line 49: command should be a required field

lib/config/src/app/job.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// Only applicable for runners that accept CLI arguments.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    cli_args: Option<Vec<String>>,
+
+    /// Environment variables.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub env: Option<HashMap<String, String>>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub capabilities: Option<ExecutableJobCompatibilityMapV1>,
```
ayys, line 61: not sure if this is the right way to set capabilities. @theduke can you comment?

lib/config/src/app/job.rs
```diff
@@ -0,0 +1,239 @@
...
+#[derive(
+    serde::Serialize, serde::Deserialize, schemars::JsonSchema, Clone, Debug, PartialEq, Eq,
+)]
+pub struct Job {
+    name: String,
+    trigger: JobTrigger,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    fetch: Option<HttpRequest>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    execute: Option<ExecutableJob>,
```
ayys, line 19: use the OneOf schema validation here?
https://docs.rs/schemars/latest/schemars/schema/struct.SubschemaValidation.html#structfield.one_of

lib/config/src/app/job.rs
Arshia001: > Defaults to the package's entrypoint.

Command being optional would make sense for e.g. bash. One could do:

```yaml
jobs:
  - execute:
      package: bash/bash
      cli_args: "cp xxx yyy"
```

without needing to know/guess what the main command is called.

lib/config/src/app/job.rs
Arshia001: Honestly, I'm not so sure about this one either. There was a similar field in `AppConfig`, and having a (configurable) memory limit for jobs made sense to me... Not sure though, will wait for @theduke to comment on this.

lib/config/src/app/job.rs
Arshia001: Good point! Done in d1c6dffe31cfe62f1baf1cc7cd9a4bc5f43b64fa.

lib/config/src/app/job.rs
ayys: but package is also optional. maybe we can use `anyOf` here? https://docs.rs/schemars/latest/schemars/schema/struct.SubschemaValidation.html#structfield.any_of

theduke: It makes sense to require either cli_args or a package, but the fields should be optional.

The BE can do validation and provide an error.

lib/config/src/app/job.rs
theduke: Yup, makese sense this way.

lib/config/src/app/job.rs
ayys: could be a warning the CLI shows letting the user know about this.
`You are not specifying any fields for the job. Unless you know what you're doing, you should specify atleast one of the following fields: 'package', 'command' or 'cli_args'.`

But yeah, thats not relevant for this PR.

theduke (approved):

lib/config/src/app/job.rs
```diff
@@ -0,0 +1,262 @@
...
+    pub env: Option<HashMap<String, String>>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub capabilities: Option<ExecutableJobCompatibilityMapV1>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub volumes: Option<Vec<AppVolume>>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub timeout: Option<PrettyDuration>,
```
theduke, line 74: These settings (timeout,drift,retries) should be on the cron, not on this action.
They are also relevant for fetch jobs.

lib/config/src/app/job.rs
```diff
@@ -0,0 +1,262 @@
...
+)]
+pub enum CronType {
+    Daily,
+    Hourly,
+    Weekly,
+    CronExpression(String),
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum JobTrigger {
```
theduke, line 40: Here we could just use `#[serde(rename_all = "kebab-case")]` so we don't need a custom De/Ser implementation.

We need that just for the CronType.

lib/config/src/app/job.rs
```diff
@@ -0,0 +1,262 @@
...
+#[serde(rename_all = "lowercase")]
+pub enum JobAction {
+    Fetch(HttpRequest),
+    Execute(ExecutableJob),
+}
+
+#[derive(
+    serde::Serialize, serde::Deserialize, schemars::JsonSchema, Clone, Debug, PartialEq, Eq,
+)]
+pub enum CronType {
```
theduke, line 32: Maybe `CronSchedule`?

lib/config/src/app/job.rs
Arshia001: Nice catch!

lib/config/src/app/job.rs
Arshia001: if it was

```yaml
trigger:
  cron: "* * * * *"
```

then yeah, but currently it's just a string and we need to decide based on the string's format, so I don't think it's possible.

theduke (approved):

PR #5355, opened 2025-01-20
chore: remove lazy_static dependency, use std instead

nyurik:
# Description
Since 1.80, Rust has [std::sync::LazyLock](https://doc.rust-lang.org/std/sync/struct.LazyLock.html), making lazy_static crate no longer needed.


xdoardo (approved):

xdoardo:
Looks good to me, thanks! 

(note to self: the failing CI test for unit-tests on std+macos-arm is due to `brew` libs not being read without the `RUSTDOCFLAGS` env variable; this is already fixed in another branch!)

PR #5360, opened 2025-01-21
chore: inline format args to improve readability

nyurik:
Using this command, fixed format argument inlining to improve readability, and fix a few minor related styling issues like trailing commas in a single line.

```
cargo clippy --all-targets --workspace --exclude wasmer-cli --exclude wasmer-swift -- -D clippy::uninlined_format_args
```

P.S.  Make sure to hide whitespace changes when reviewing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -967,7 +967,7 @@ install-wasmer-headless-minimal:
 update-testsuite:
 	git subtree pull --prefix tests/wast/spec https://github.com/WebAssembly/testsuite.git master --squash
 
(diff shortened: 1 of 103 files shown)


xdoardo:
Thanks! I think we can enforce this lint from now on: this would mean adding the `clippy::uninlined_format_args` to the `RUSTFLAGS` in the `lint-packages` command in the `Makefile`; would you like to add that?

nyurik:
thx, I updated the make file.  In theory, all Clippy configurations could be done much better using package.json workspace. Note that it should use `warn` rather than `deny` - because for the overall health of the project CI should simply run `cargo clippy ...  -- -D warnings` to treat warnings as errors.

#### main Cargo.toml
```toml
[workspace.lints.rust]
unused_qualifications = "warn"

[workspace.lints.clippy]
uninlined_format_args = "warn"
```

#### all crates
```
[lints]
workspace = true
```

xdoardo:
Thanks a lot, and thanks for the tip as well!

xdoardo (approved):

PR #5363, opened 2025-01-23
chore: fix all warnings

nyurik:
This fixes all compiler warnings. Note that `__cbindgen_hack__ = "yes"` is a bit weird - there is no need to make it support a value -- presence of a keyword is an indicator enough. Also, I couldn't figure out what actually sets that value.

### DIFF
--- a/lib/c-api/Cargo.toml
+++ b/lib/c-api/Cargo.toml
@@ -102,6 +102,7 @@ wasmer-artifact-create = ["wasmer-compiler/wasmer-artifact-create"]
 static-artifact-load = ["wasmer-compiler/static-artifact-load"]
 static-artifact-create = ["wasmer-compiler/static-artifact-create"]
 webc_runner = ["virtual-fs", "webc"]
+dylib = []
 # Deprecated features.
 jit = ["compiler"]
 
--- a/lib/cli-compiler/Cargo.toml
+++ b/lib/cli-compiler/Cargo.toml
@@ -75,6 +75,8 @@ cranelift = ["wasmer-compiler-cranelift", "compiler"]
 debug = ["fern", "log"]
 disable-all-logging = []
 jit = []
+llvm = []
+run = []
 
 [package.metadata.docs.rs]
 rustc-args = ["--cfg", "docsrs"]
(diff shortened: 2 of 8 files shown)


lib/c-api/Cargo.toml
```diff
@@ -102,6 +102,7 @@ wasmer-artifact-create = ["wasmer-compiler/wasmer-artifact-create"]
 static-artifact-load = ["wasmer-compiler/static-artifact-load"]
 static-artifact-create = ["wasmer-compiler/static-artifact-create"]
 webc_runner = ["virtual-fs", "webc"]
+dylib = []
```
syrusakbary, line 105: This feature was deprecated long ago. We shouldn't add it back

lib/sys-utils/Cargo.toml
```diff
@@ -29,3 +29,7 @@ tracing = "0.1.37"
 
 [package.metadata.docs.rs]
 rustc-args = ["--cfg", "docsrs"]
+
+[features]
+default = []
+tracing = []
```
syrusakbary, line 35: Why is tracing a feature if it's not used anywhere?

nyurik:
@syrusakbary thx for the quick review!  The reason I added those is because you have a lot of `cfg(feature = "...")` that still use those features - generating warnings.  IF they are deprecated, they should be removed at the same time from the code and cargo.toml, or else it is really confusing why the warnings are being generated

lib/c-api/Cargo.toml
nyurik: You use it in https://github.com/wasmerio/wasmer/blob/d431db9a3cbcf9a34ce3f5e83c5deddac29b52f7/lib/c-api/src/wasm_c_api/engine.rs#L258 - generating compiler warning because `dylib` is not declared. Should it be removed there?

lib/cli-compiler/Cargo.toml
```diff
@@ -75,6 +75,8 @@ cranelift = ["wasmer-compiler-cranelift", "compiler"]
 debug = ["fern", "log"]
 disable-all-logging = []
 jit = []
+llvm = []
+run = []
```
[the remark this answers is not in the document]
nyurik: If the features are used in code, they should be declared in cargo.toml - otherwise its like using a variable without declaring it first - ok for some languages, but usually a bad form for Rust :)  Should they be removed there, or should these be added here until then?

syrusakbary:
Sounds good. I'll do a quick fix removing some of the features that are no longer needed in a PR after

PR #5366, opened 2025-01-28
Reduce memory usage when loading webc packages by ensuring binaries m…

Arshia001:
…ounted in /bin aren't cloned

### DIFF
--- a/lib/wasix/src/state/env.rs
+++ b/lib/wasix/src/state/env.rs
@@ -1093,28 +1093,13 @@ impl WasiEnv {
                 let path = format!("/bin/{}", command.name());
                 let path = Path::new(path.as_str());
 
-                // FIXME(Michael-F-Bryan): This is pretty sketchy.
-                // We should be using some sort of reference-counted
-                // pointer to some bytes that are either on the heap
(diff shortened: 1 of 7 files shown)


lib/wasix/src/state/env.rs
```diff
@@ -1093,28 +1093,13 @@ impl WasiEnv {
...
 
                 match root_fs {
                     WasiFsRoot::Sandbox(root_fs) => {
-                        // As a short-cut, when we are using a TmpFileSystem
-                        // we can (unsafely) add the file to the filesystem
-                        // without any copying.
                         if let Err(err) = root_fs
                             .new_open_options_ext()
-                            .insert_ro_file(path, atom.into())
+                            .insert_ro_file(path, atom.as_slice().to_vec().into())
```
theduke, line 1102: Can we extend the fs to support `SharedBuffers` for readonly files?

Then we can skip this copy.

theduke (approved):

lib/wasix/src/state/env.rs
Arshia001: Done, though I don't know of any scenario where this code actually runs.

PR #5374, opened 2025-02-03
Add VirtualFile::copy_from_owned_buffer, fixes running mounted comman…

Arshia001:
…ds multiple times within the same process

# Description

### DIFF
--- a/lib/virtual-fs/src/mem_fs/mod.rs
+++ b/lib/virtual-fs/src/mem_fs/mod.rs
@@ -54,7 +54,7 @@ struct ArcFileNode {
     metadata: Metadata,
 }
 
-// TODO: this is broken!!! A `VirtualFile` stores its own offset,
+// FIXME: this is broken!!! A `VirtualFile` stores its own offset,
 // so a file stored this way can only be read once!
 #[derive(Debug)]
 struct CustomFileNode {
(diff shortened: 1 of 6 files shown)


theduke (approved):

lib/virtual-fs/src/mem_fs/mod.rs
```diff
@@ -54,6 +54,8 @@ struct ArcFileNode {
     metadata: Metadata,
 }
 
+// TODO: this is broken!!! A `VirtualFile` stores its own offset,
```
theduke, line 57: Should be a `FIXME: ` ;) .

And probably a followup issue.

lib/wasix/src/state/env.rs
```diff
@@ -1125,18 +1125,17 @@ impl WasiEnv {
                         }
                     }
                     WasiFsRoot::Backing(fs) => {
+                        // FIXME: we're counting on the fs being a mem_fs here. Otherwise, memory
+                        // usage will be very high.
                         let mut f = fs.new_open_options().create(true).write(true).open(path)?;
```
theduke, line 1130: Can we get a small regression test for this, just with a temp FS?

Arshia001: Would be quite difficult to set up, since we'd need to go through the code path that sets up the WASI FS and then make sure the fs is actually the correct type...

lib/virtual-fs/src/mem_fs/mod.rs
Arshia001: Sure :D

PR #5379, opened 2025-02-07
Allow memory growing to exactly 65536 pages, as per the specification

charmitro:
The Wasm spec allows memory to grow up to 2^16 pages[1]. We were a bit more restrictive, failing at `>= 65536` instead of `> 65536`. This fixes the check to allow valid 65536-page allocations.

Added test verifying correct memory growth behavior through `example/memory_grow.rs`.

[1]: https://webassembly.github.io/spec/core/bikeshed/index.html#growing-memories%E2%91%A0
1: https://webassembly.github.io/spec/core/bikeshed/index.html#growing-memories%E2%91%A0

```
If len is larger than 2^16, then fail
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -341,6 +341,11 @@ path = "examples/memory.rs"
 required-features = ["backend"]
 
 [[example]]
+name = "memory-grow"
+path = "examples/memory_grow.rs"
+required-features = ["backend"]
+
+[[example]]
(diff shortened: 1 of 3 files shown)


charmitro:
> Nice. LGTM

Failing the test-wasmer-integration-tests, but it doesn't seems relevant to my changes. All other tests succeded.

Issue #5224, opened 2024-11-06
Enable `c_api`-backends to be used in Wasmer's implementation of the `wasm_c_api`


xdoardo:
Due to duplicate symbols we can't, as of now, produce a [wasm-c-api](https://github.com/WebAssembly/wasm-c-api)-compatible header to interact with wasmer compiled with a backend that uses a `c_api` backend. 

We should investigate how to enable this use case. A possible solution might be to use `objcopy` with the `--prefix-symbols` flag.

--- 

Following the comments in https://github.com/wasmerio/wasmer/pull/5389, we note that solving this issue might be also a good chance to revamp a bit the structure of the `wasm_c_api`, especially the feature mechanisms.

PR #5389, opened 2025-02-14
fix(c_api): Feature gate compiler-related types and functions behind …

xdoardo:
This small patch feature-gates some of the `wasmer_c_api` types and functions behind the already-used `"compiler"` feature. Affected are the `wasmer_target_t` type and the `wasm_config_set_target` function. Also, the `target` field of the `wasm_config_type` is behind the same feature-gate -- note that the `compiler` field was already behind the same feature gate.

### DIFF
--- a/lib/c-api/src/wasm_c_api/engine.rs
+++ b/lib/c-api/src/wasm_c_api/engine.rs
@@ -1,18 +1,23 @@
-#[cfg(feature = "compiler")]
-pub use super::unstable::engine::wasmer_is_compiler_available;
-pub use super::unstable::engine::{wasm_config_set_features, wasm_config_set_target};
+pub use super::unstable::engine::wasm_config_set_features;
 use super::unstable::features::wasmer_features_t;
+use crate::error::update_last_error;
+use cfg_if::cfg_if;
+
+#[cfg(feature = "compiler")]
+pub use super::unstable::engine::{wasm_config_set_target, wasmer_is_compiler_available};
+#[cfg(feature = "compiler")]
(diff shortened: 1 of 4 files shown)


lib/c-api/src/wasm_c_api/engine.rs
```diff
@@ -84,6 +89,7 @@ pub struct wasm_config_t {
     pub(super) middlewares: Vec<wasmer_middleware_t>,
     pub(super) nan_canonicalization: bool,
     pub(super) features: Option<Box<wasmer_features_t>>,
+    #[cfg(feature = "compiler")]
```
syrusakbary, line 92: We should do a breaking change and unite all the sys features into one field (for sys).

Those are: middleware/nan_canonicalization/target

lib/c-api/src/wasm_c_api/unstable/engine.rs
```diff
@@ -48,6 +49,7 @@ use super::target_lexicon::wasmer_target_t;
 /// # }
 /// ```
 #[no_mangle]
+#[cfg(feature = "compiler")]
 pub extern "C" fn wasm_config_set_target(config: &mut wasm_config_t, target: Box<wasmer_target_t>) {
```
syrusakbary, line 53: wasm_config_sys_set_target?

xdoardo:
Note: we will probably address the comments above in the near future (https://github.com/wasmerio/wasmer/issues/5224)

syrusakbary (approved):

PR #5390, opened 2025-02-14
chore: Fix doc comments in the api crate

xdoardo:
As per title.

### DIFF
--- a/lib/api/src/entities/module/inner.rs
+++ b/lib/api/src/entities/module/inner.rs
@@ -18,17 +18,6 @@ use crate::{
     AsEngineRef,
 };
 
-/// IO errors that can happen while compiling a [`BackendModule`].
-#[derive(Error, Debug)]
-pub enum IoCompileError {
-    /// An IO error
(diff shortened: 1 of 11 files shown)


lib/api/src/entities/module/inner.rs
```diff
@@ -18,7 +18,7 @@ use crate::{
     AsEngineRef,
 };
 
-/// IO errors that can happen while compiling a [`Module`].
+/// IO errors that can happen while compiling a [`BackendModule`].
```
syrusakbary, line 21: This should be Module, not BackendModule. `BackendModule` should be not be usable/public by the user

syrusakbary: I believe IoCompileError is an error for all backends... isn't it?

xdoardo: Sure. This type shouldn't be there at all actually. Removing in next commit.

syrusakbary (approved):

PR #5394, opened 2025-02-15
Graceful Handling of OOM in `table.grow`

charmitro:
Introduce `DEFAULT_MAX_TABLE_SIZE` set to 1,048,576 elements (~1
million, 2^20) as a default cap for tables lacking a specified
maximum (`self.maximum = None`). This prevents impractical growth
attempts, such as a delta of 0xff_ff_ff_ff (4,294,967,295 elements),
which could succeed misleadingly on overcommitting systems like macOS.

Replace panic-prone `Vec::resize` with `try_reserve_exact` to attempt
memory reservation before resizing. Returns `None` (mapping to
WebAssembly’s -1) on allocation failure, ensuring graceful handling of
out-of-memory (OOM) conditions rather than crashing, as seen
previously on Linux with large deltas.

Update growth logic to compute the effective
maximum (`self.maximum.unwrap_or(DEFAULT_MAX_TABLE_SIZE)`), failing
early if `new_len` exceeds this limit, followed by a safe reservation
and resize sequence.

These changes align with the WebAssembly 2.0 (Draft 2025-01-28) spec’s
allowance for implementation-defined resource limits (Section 7.2) and
`table.grow` semantics (returning -1 on exceeding limits or resource
constraints).

### DIFF
--- a/examples/table_large_grow.rs
+++ b/examples/table_large_grow.rs
@@ -1,0 +1,48 @@
+//! Example demonstrating WebAssembly table growth limits
+//!
+//! Tests growing a table with a very large delta (0xff_ff_ff_ff) which should fail
+//! gracefully by returning -1 according to the WebAssembly specification.
+//!
+//! You can run the example directly by executing:
+//! ```shell
+//! cargo run --example table-large-grow --release --features backend
+//! ```
+
+use wasmer::{imports, Instance, Module, Store};
+
+fn main() -> anyhow::Result<()> {
+    let module_wat = r#"
+    (module
+      (table $t0 0 externref)
+      (table $t1 10 externref)
+      (func $init (export "init") (result i32)
+	(if (i32.ne (table.size $t1) (i32.const 10))
+	  (then (unreachable))
+	)
+	;; Store the result instead of dropping it
+	(table.grow $t0 (ref.null extern) (i32.const 0xff_ff_ff_ff))
+      )
+    )
+    "#;
+
+    let mut store = Store::default();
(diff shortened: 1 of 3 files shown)


examples/table_large_grow.rs
```diff
@@ -0,0 +1,55 @@
...
+        wasmer::Value::I32(0),
+        "On macOS, table.grow should return 0"
+    );
+
+    #[cfg(target_os = "linux")]
+    assert_eq!(
+        result[0],
+        wasmer::Value::I32(-1),
+        "On Linux, table.grow should return -1"
+    );
```
syrusakbary, line 47: This shouldn't happen (different output for same program, under different OSs). Good catch.
We need to make the API deterministic no matter what OS is running on

charmitro: Not sure/haven't put more thought into whether this is possible.

Pasting the scenarios here (before this PRs changes), since I have them handy, for some food for thought.

Linux:
* Linux with `vm.overcommit_memory = 0` results in `memory allocation of ... bytes failed`
* Linux with `vm.overcommit_memory = 1` results in `killed`
* Linux with `vm.overcommit_memory = 2` results in `memory allocation of ... bytes failed`

MacOS always works without exiting.

Maybe there's not much we can do here. But doing nothing will not adere the specification (i.e. _The [table.grow](https://webassembly.github.io/spec/core/bikeshed/index.html#syntax-instr-table) instruction grows table by a given delta and returns the previous size, or −1 if enough space cannot be allocated_).

syrusakbary (approved):

PR #5396, opened 2025-02-17
CI: Disable V8 for linux-aarch64, bump ubuntu version for linux-x64 and add search paths for linux-musl

xdoardo:
As per title.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,7 +54,7 @@ jobs:
             enable_wasmi: true
             enable_wamr: true
           - build: linux-x64
-            os: ubuntu-24.04
+            os: ubuntu-22.04
             artifact_name: 'wasmer-linux-amd64'
             llvm_url: 'https://github.com/wasmerio/llvm-custom-builds/releases/download/18.x/llvm-linux-amd64.tar.xz'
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -50,11 +50,11 @@ jobs:
             use_sccache: false
             build_wasm: true
             enable_llvm: true
-            enable_v8: true 
+            enable_v8: false
             enable_wasmi: true
             enable_wamr: true
           - build: linux-x64
-            os: ubuntu-22.04
```
theduke, line 57: NOTE: I don't know if this is the worfklow also used for building release artifacts, but if so we can not just do this bump because it makes the artifacts incompatible with OSes that are on an older libc.

We got plenty of complaints about that in the past.

xdoardo: I see. This entails rebuilding wee8. We can either merge it like this for now and downgrade later, skip wee8 for linux-x64 until it builds as expected on older ubuntu, or just block the release until wee8 is built for the correct target.

xdoardo: Note: we went for option number 3.

theduke (approved):

PR #5397, opened 2025-02-17
fix: have wasmer-wasix (lib/wasix) compile with `cargo build`

ayys:
When trying to compile wasmer-wasix with a simple `cargo build`, I noticed that it needed the sys flag from lib/api crate to be enabled. So I am adding the flag to `sys-default` which is enabled by default.

### DIFF
--- a/lib/wasix/Cargo.toml
+++ b/lib/wasix/Cargo.toml
@@ -213,8 +213,9 @@ sys-default = [
 	"sys-thread",
 	"host-vnet",
 	"host-threads",
-	"host-reqwest",
-	"ctrlc",
+        "host-reqwest",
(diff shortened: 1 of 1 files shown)


xdoardo:
To me the change looks good, since the `sys-default` feature on the `wasix` crate seems to imply, well, that it is supposed to enable the `sys` feature. However, I'm not sure if this entails something that I'm not aware of. @Arshia001 probably knows better. 

Also, a relevant note is that the [docs build](https://docs.rs/crate/wasmer-wasix/latest) of `wasmer-wasix` have been failing since version 0.32.0 (the last four versions).

Arshia001 (approved):

PR #5401, opened 2025-02-17
Fix returning from a failed exec + fix vfork

Arshia001:


lib/wasix/src/runners/wasi.rs
```diff
@@ -328,13 +329,78 @@ impl WasiRunner {
...
+                                    WasiRuntimeError::Anyhow(a) => {
+                                        WasiRuntimeError::Anyhow(a.clone())
+                                    }
+                                })
+                                .unwrap_or_else(|| {
+                                    WasiRuntimeError::Anyhow(Arc::new(anyhow::format_err!(
+                                        "{}", msg
+                                    )))
+                                })
+                        })
```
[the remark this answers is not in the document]
Arshia001: TBH, I probably thought the same thing at least 10 times... I'll fix it.

lib/wasix/src/runners/wasi.rs
```diff
@@ -367,22 +433,35 @@ impl crate::runners::Runner for WasiRunner {
...
         #[cfg(feature = "journal")]
         {
             for journal in self.wasi.journals.clone() {
-                env.add_journal(journal);
+                builder.add_journal(journal);
             }
 
-            for snapshot_trigger in self.wasi.snapshot_on.iter().cloned() {
-                env.add_snapshot_trigger(snapshot_trigger);
+            for trigger in self.wasi.snapshot_on.iter().cloned() {
```
theduke, line 446: small nit: this is a bit confusing, would be clearer with an `if wasi.snapshot_on.is_empty() { ... } else { ... }`

lib/wasix/src/runtime/task_manager/mod.rs
```diff
@@ -91,6 +97,7 @@ pub struct TaskWasm<'a> {
     pub spawn_type: SpawnMemoryType<'a>,
     pub trigger: Option<Box<WasmResumeTrigger>>,
     pub update_layout: bool,
+    pub pre_run: Option<Box<TaskWasmPreRun>>,
```
theduke, line 100: This is technically redundant , because one could just wrap the `run` and replace it, which would keep complexity down a bit.

BUT it also makes things more confusing , so this is probably better.

lib/wasix/src/syscalls/wasix/fd_pipe.rs
```diff
@@ -1,8 +1,12 @@
+use std::sync::atomic::AtomicUsize;
+
 use virtual_fs::Pipe;
 
 use super::*;
 use crate::syscalls::*;
 
+static PIPE_NUMBER: AtomicUsize = AtomicUsize::new(0);
```
theduke, line 8: docs for why this is needed and what's going on here.

theduke: Shouldn't this be a counter somewhere on the env data?

Arshia001: Good point! We use a hash of the file name to calculate inode numbers, so all pipes were getting the same. This serves to make pipe names unique. It doesn't really matter otherwise. Will add comments.

lib/wasix/src/runtime/task_manager/mod.rs
Arshia001: That's a fair point... Initially, I wanted to make this fallible, but through multiple revisions ended up with this, which doesn't make a lot of sense.

Arshia001: Although, I think the main callback isn't async?

lib/wasix/src/runners/wasi.rs
Arshia001: Funny how most of the comments came from code I copy-pasted 😄

lib/wasix/src/runners/wasi.rs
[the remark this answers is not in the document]
Arshia001: Except you can't clone every case, so you can't give it a `Clone` impl.

lib/wasix/src/runtime/task_manager/mod.rs
Arshia001: yep, the `run` field isn't async, and I don't want to make it async since it's used in many places. Let's just keep this.

theduke (approved):

PR #5414, opened 2025-02-19
LLVM: Use `Default` and `Static` code model and reloc mode when targeting Mach-O

xdoardo:
Also, disregard the `addend` hack that we usually compute while reading relocations when the relocation read has a specific kind.


syrusakbary (approved):
Relocations are fun

PR #5415, opened 2025-02-19
chore(wasix): Avoid redundant hashing when spawning commands

theduke:
When loading modules for package commands, re-use the hash from the
command to avoid redundant re-hashing, which can be quite expensive.

### DIFF
--- a/lib/wasix/src/bin_factory/exec.rs
+++ b/lib/wasix/src/bin_factory/exec.rs
@@ -19,7 +19,7 @@ use virtual_mio::InlineWaker;
 use wasmer::{Function, Memory32, Memory64, Module, RuntimeError, Store, Value};
 use wasmer_wasix_types::wasi::Errno;
 
-use super::BinaryPackage;
+use super::{BinaryPackage, BinaryPackageCommand};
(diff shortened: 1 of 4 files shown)


lib/wasix/src/bin_factory/exec.rs
```diff
@@ -54,25 +53,36 @@ pub async fn spawn_exec_wasm(
...
-            package_id: binary.id.clone(),
-        });
+        match pkg.commands.as_slice() {
+            [] => {
+                return Err(SpawnError::MissingEntrypoint {
+                    package_id: pkg.id.clone(),
+                });
+            }
+            [first] => first,
+            [_first, ..] => {
```
Arshia001, line 77: This time around, I'm not approving this 😄 this case can be merged with the empty slice case.

Arshia001 (changes requested):

lib/wasix/src/bin_factory/exec.rs
theduke: Buuut it's supposed to signal a difference between "no commands" and "multiple commands, don't know which to pick".

We can extend the error with a message to reflect this (that also mentions the command names).

Arshia001 (approved):

PR #5426, opened 2025-02-24
Add proper posix_spawn support without needing asyncify

Arshia001:


lib/wasix/src/lib.rs
```diff
@@ -549,13 +549,16 @@ fn wasix_exports_32(mut store: &mut impl AsStoreMut, env: &FunctionEnv<WasiEnv>)
         "proc_fork" => Function::new_typed_with_env(&mut store, env, proc_fork::<Memory32>),
         "proc_join" => Function::new_typed_with_env(&mut store, env, proc_join::<Memory32>),
         "proc_signal" => Function::new_typed_with_env(&mut store, env, proc_signal::<Memory32>),
+        "proc_signals_count_get" => Function::new_typed_with_env(&mut store, env, proc_signals_count_get::<Memory32>),
```
syrusakbary, line 552: The get seems redundant. IMO we should name this `proc_signals_count`

lib/wasix/src/lib.rs
```diff
@@ -678,13 +681,16 @@ fn wasix_exports_64(mut store: &mut impl AsStoreMut, env: &FunctionEnv<WasiEnv>)
         "proc_fork" => Function::new_typed_with_env(&mut store, env, proc_fork::<Memory64>),
         "proc_join" => Function::new_typed_with_env(&mut store, env, proc_join::<Memory64>),
         "proc_signal" => Function::new_typed_with_env(&mut store, env, proc_signal::<Memory64>),
+        "proc_signals_count_get" => Function::new_typed_with_env(&mut store, env, proc_signals_count_get::<Memory64>),
+        "proc_signals_get" => Function::new_typed_with_env(&mut store, env, proc_signals_get::<Memory64>),
```
syrusakbary, line 685: `_get`

lib/wasi-types/src/wasi/bindings.rs
```diff
@@ -2328,6 +2328,93 @@ impl core::fmt::Debug for Signal {
...
+            SigAction::Default => f.debug_tuple("SigAction::Default").finish(),
+            SigAction::Ignore => f.debug_tuple("SigAction::Ignore").finish(),
+        }
+    }
+}
+
+#[doc = " A signal and its corresponding action."]
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct SignalAndAction {
```
syrusakbary, line 2350: I think, based on the rest of the code, it may be more consistent to name `SignalAndAction` -> `Signal`, and then the current `Signal` name it `SignalKind`

lib/wasix/src/lib.rs
syrusakbary: Count is already a getter. In WASI they have `args_sizes_get`. We can have `proc_signals_sizes_get` which is consistent. But not `count_get` those are two verbs. We should either `count`, or `get`.

syrusakbary (approved):

Issue #5423, opened 2025-02-23
SSL/proxy flexibility for corporate environment


mihailik:
### Motivation

Currently in a corporate environment you often have (a) authenticated HTTP/HTTPS proxies (b) SSL certificate spoofing.

Other registries like Maven for Java or NPM for JavaScript work, often requiring a bit of tweaking with environment variables and/or extra arguments.

Wasmer doesn't seem to work in any way at the moment. SSL certificates could be hardcoded into the binary? Or additional SSL/proxy arguments not expected? Whatever the cause, there is no visible solution at this point.

### Proposed solution

Allow customising proxies, and customising the root SSL certificates. Conservatively that may come as command line options, forcing to be explicitly passed.

### Alternatives

Another option could be to rely on the same environment variables as node.js for example. However, that can come with security implications, where end user may be having those options thinking they only affect node environment, and they are also kicking in in Wasmer.

These unexpected environment shifts can be most uncooperative in CI/CD environments causing mysterious behaviour changes.

Explicit command line argument are probably least disruptive.

### Additional context

I am going to have a stab at it this Sunday, because that way I can at least have my own private clone of Wasmer tool to show POC of what we can do at my day job.

If everything goes smooth (a big if) we may adopt Wasmer for many interesting workloads internally. Which could benefit us: we have so many environments and jobs to run, and there's a mandate in 2025 to unify and make these data workloads more coherent and managed. But it could also benefit Wasmer, if the bosses like it and want to have some kind of Enterprise support deal etc. That's beyond my pay grade though, I'm focusing on the technical solution here.

mihailik:
This PR deals with some of these issues, but still not working in a corporate locked-out environment:

[#5091 Apply the proxy setting in wasmer config [merged]](https://github.com/wasmerio/wasmer/pull/5091)

I'll try to step through in debugger and see what's breaking.

mihailik:
I've done a bit of research in the codebase, there seem to be 4 call sites that create `reqwest::ClientBuilder::new()` that need to be made aware of the extra parameters related to proxy, SSL certificate store and possible self-signed certificate opt-in.

[lib/backend-api/src/client.rs](https://github.com/wasmerio/wasmer/blob/d21706ef723d823003a3cce22be3f8c899e5f305/lib/backend-api/src/client.rs#L96)
[...]

PR #5429, opened 2025-02-25
Support and propagate HTTP options for proxy, ssl.

mihailik:
# Command line parameters for proxy, ssl root, ssl opt-in for self-signed certificates

See issue for long rationale [**#5423** SSL/proxy flexibility for corporate environment](https://github.com/wasmerio/wasmer/issues/5423) 

This PR introduces ClientBuilderConfig in lib/wasix/src/http/client_builder.rs and corresponding CliClientBuilderConfig at lib/cli/src/config/env.rs.

These include:
* **proxy** to specify HTTP proxy
* **ca_file** to specify SSL root certificate override
* **unsafe_disable_ssl_verify** to allow self-signed certificates

These are parsed from the command line, and passed around during runtime initialisation flow. When any HTTP request is made, ClientBuilderConfig::create_client_builder is the one producing ClientBuilder with necessary settings.

Every creation of reqwest::ClientBuilder anywhere in Wasmer codebase now goes through ClientBuilderConfig::create_builder.

## Corporate environment case

SSL spoofing and seemingly insecure HTTP proxies are commonplace in large corporate environments. You just can't reach out to the big internet without a bit of customisation.

This PR allows that customisation from a command line, explicitly.

It also allows for future tweaking of that aspect, should you want to cater to particular commercial need of your business client. The model of capturing the options is provided, and the single central place where these options come to fruition is ClientBuilderConfig::create_client_builder.

## Testing and coding style

I've updated the existing tests to account for the changes, and tried to adhere to the coding styles.

Any feedback, guidance, refactoring/naming suggestions are welcome! I'm not fussy about how this feature implemented, just really want to enable Wasmer for a possible corporate use.


marxin:
Thanks a lot for putting this together — the motivation makes total sense, and we agree the proxy / custom CA / self-signed cert opt-in story is important for corporate environments (and #5423 captures that well).

That said, we’re going to close this PR for now, mainly because we’d like to implement this a bit differently on our side, and we don’t want to merge something that we’ll likely need to reshape soon after.

If you’re still interested in helping with the final version: is the original request in #5423 still valid for your current use case?

Appreciate the effort here 🙏

Issue #5432, opened 2025-02-25
Remove unnecessary panic when an import is missing


xdoardo:
When running a module that imports some functions that are not available, a user might experience this: 
```
λ wasmer run new.wasm --llvm
thread 'TokioTaskManager Thread Pool_thread_2' panicked at lib/wasix/src/runtime/task_manager/tokio.rs:266:18:
called `Result::unwrap()` on an `Err` value: InstanceCreateFailed(Link(Import("env", "__wasm_longjmp", UnknownImport(Function(FunctionType { params: [I32, I32], results: [] })))))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

We should remove the `unwrap` and properly dispose of (and report) the error. 
A reproduction of this behaviour can be seen with [this](https://github.com/user-attachments/files/18973833/mre.wasm.zip) on macOS-aarch64 using the LLVM backend.

PR #5440, opened 2025-02-27
Remove `unwrap` in code that instantiates a WASI module

xdoardo:
(fixes: https://github.com/wasmerio/wasmer/issues/5432)

This patch does two things: 
* fixes an `unwrap` introduced [here](https://github.com/wasmerio/wasmer/pull/5284/files#diff-869ef3b52ca7df003ad6395ca1448c280a51f1fe06345a6d4fe6fe7609c923c6R259)
* fixes an `#[error]` annotation to print the inner error

So that 
```
λ wasmer run new.wasm --llvm
thread 'TokioTaskManager Thread Pool_thread_2' panicked at lib/wasix/src/runtime/task_manager/tokio.rs:266:18:
called `Result::unwrap()` on an `Err` value: InstanceCreateFailed(Link(Import("env", "__wasm_longjmp", UnknownImport(Function(FunctionType { params: [I32, I32], results: [] })))))
```

is now 

```
λ wasmer run new.wasm  --llvm
error: Spawn failed
╰─▶ 1: Failed to create the instance - Error while importing "env"."__wasm_longjmp": unknown import. Expected Function(FunctionType { params: [I32, I32], results: [] })
```

### DIFF
--- a/lib/wasix/src/runtime/task_manager/tokio.rs
+++ b/lib/wasix/src/runtime/task_manager/tokio.rs
@@ -253,17 +253,27 @@ impl VirtualTaskManager for TokioTaskManager {
         } else {
             tracing::trace!("spawning task_wasm in blocking thread");
 
+            let (sx, rx) = std::sync::mpsc::channel();
+
             // Run the callback on a dedicated thread
             self.pool.execute(move || {
                 tracing::trace!("task_wasm started in blocking thread");
-                let (mut ctx, mut store) = WasiFunctionEnv::new_with_store(
+                let (mut ctx, mut store) = match WasiFunctionEnv::new_with_store(
                     task.module,
                     env,
                     task.globals,
                     make_memory,
                     task.update_layout,
-                )
-                .unwrap();
+                ) {
+                    Ok(x) => {
+                        sx.send(Ok(())).unwrap();
+                        x
+                    }
+                    Err(c) => {
+                        sx.send(Err(c)).unwrap();
+                        return;
+                    }
+                };
                 if let Some(pre_run) = pre_run {
                     InlineWaker::block_on(pre_run(&mut ctx, &mut store));
                 }
@@ -276,6 +286,8 @@ impl VirtualTaskManager for TokioTaskManager {
                     recycle,
                 });
             });
+
+            rx.recv().unwrap()?;
         }
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


xdoardo:
@theduke: I'm not too sure about the use of the channel here. Do you think it's fine? The alternative I had in mind was to create a `WasiFunctionEnv::check_imports(&module)` (or `can_instantiate`) function, but that would've entailed introducing a list of all the WASI(X) exports mapped to their names to be able to match them against the list of imports of the module without having a store just yet.

lib/wasix/src/runtime/task_manager/tokio.rs
```diff
@@ -253,17 +253,27 @@ impl VirtualTaskManager for TokioTaskManager {
         } else {
             tracing::trace!("spawning task_wasm in blocking thread");
 
+            let (sx, rx) = std::sync::mpsc::channel();
+
             // Run the callback on a dedicated thread
             self.pool.execute(move || {
                 tracing::trace!("task_wasm started in blocking thread");
-                let (mut ctx, mut store) = WasiFunctionEnv::new_with_store(
+                let (mut ctx, mut store) = match WasiFunctionEnv::new_with_store(
```
theduke, line 261: nit: I find code like this super unreadable - I always do 

```
let res = ...;
match res { ...
```

lib/wasix/src/runtime/task_manager/tokio.rs
```diff
@@ -276,6 +286,8 @@ impl VirtualTaskManager for TokioTaskManager {
                     recycle,
                 });
             });
+
+            rx.recv().unwrap()?;
```
theduke, line 290: There can be weird edge cases where the sender gets dropped, leading to this unwrap() panicking. 
(I guess if new_with_store were to panic)

Can you convert the error returned by `.recv()` and return it instead?

lib/wasix/src/runtime/task_manager/tokio.rs
xdoardo: [dc81648](https://github.com/wasmerio/wasmer/pull/5440/commits/dc816488050c232e3ef9179302d067d818b1afb2)

lib/wasix/src/runtime/task_manager/tokio.rs
xdoardo: [76826d2](https://github.com/wasmerio/wasmer/pull/5440/commits/76826d2b2217cbe0151f130487fc17e1c223a5db)

theduke (approved):

PR #5443, opened 2025-02-28
Enable multiple backends supported by the api crate to be used in the c-api as well

xdoardo:
(fixes https://github.com/wasmerio/wasmer/issues/5224)

This PR adds to the  `wasm_engine_t` enum the kinds `V8`, `WAMR`, `WASMI` and `JSC` to reflect the different possible kinds of engine that can actually be used in the underlying implementation. These enum variants can be by users used as a means to set the wanted engine in the underlying implementation. 
Together with this addition, this PR changes the `wasm_config_t` type to (internally) hold an engine-specific configuration: 
```
pub struct wasm_config_t {
    pub(super) engine: wasmer_engine_t,
    pub(super) engine_config: wasmer_engine_config_t, <------
    ...
}
```
where `wasmer_engine_config_t` is an enumeration of all the configurations for all the available engines: 
```
enum wasmer_engine_config_t {
    #[cfg(feature = "sys")]
    Sys(sys::wasmer_sys_engine_config_t),

    #[cfg(feature = "jsc")]
    Jsc(jsc::wasmer_jsc_engine_config_t),
    
    ...
}

```

The two separate fields are needed because one, `wasmer_engine_t`, is public and users can set it; the other is definitely private and users cannot interact directly with it.  This PR also introduces the relevant features to enable each of the backends cited above, with the matching `X-default` features to chose one as default.

Note: this PR produces a number of breaking changes in the API. Most notably, functions that were implicitly tied to the `sys` engine only, are now scoped with the `sys` name: `wasm_config_canonicalize_nans` is now `wasm_config_sys_canonicalize_nans`. Same goes for `wasm_config_set_sys_target` (previously `wasm_config_set_target`),  `wasm_config_sys_push_middleware` (previously `wasm_config_push_middleware`), and `wasm_config_set_sys_compiler` (previously `wasm_config_set_compiler`).

Furthermore, a chage in the semantics of these functions is (debatably) introduced: what does it mean to call `wasm_config_sys_push_middleware` on a config whose selected `wasmer_engine_t` is not `sys`? In this PR the introduced behaviour is that of regardlessly change the selected engine to `sys`: 

```
pub extern "C" fn wasm_config_set_sys_compiler(
    config: &mut wasm_config_t,
    compiler: wasmer_compiler_t,
) {
    ... 
    config.engine = wasmer_engine_t::UNIVERSAL;
    ... 
}
```
On one hand, this behaviour is transparent to the user - which should nonetheless expect the underlying engine to be the `sys` one, while on the other this function does not return a value, so we can't signal the user that the an error occurred in the case the current `config.engine != sys`. 

Notice, also, that to maintain a bit of backwards compatibility, the enum constructor in the `wasmer_engine_t` corresponding to `sys` is still `UNIVERSAL`. 

-- edit 

In the process of testing these additions, I discovered the following bugs which are now fixed: 
* An error in the code that performs deserialization of V8 modules
* Multiple errors in V8, WAMR, WASMI upon memory instantiation
* Multiple errors in V8, WAMR, WASMI in `Instance::exports()`

--

Missing bits: 
- [x] More tests with all the backends enabled
- [x] CI jobs to target those tests


lib/c-api/src/wasm_c_api/engine/config/jsc.rs
```diff
@@ -0,0 +1,38 @@
...
+pub fn wasm_jsc_engine_new_with_config(config: wasm_config_t) -> Option<Box<wasm_engine_t>> {
+    if !matches!(config.engine, wasmer_engine_t::JSC) || !config.engine_config.is_jsc() {
+        update_last_error("Cannot create a new `sys` engine with a non-sys-specific config!");
+        return None;
+    }
+
+    Some(Box::new(wasm_engine_t {
+        inner: wasmer_api::jsc::Engine::default().into(),
+    }))
+}
```
syrusakbary, line 28: This function is public, therefore why are we checkign the engine if what we believe it should be? I'd expect to receive a `wasm_jsc_config_t` instead of a generic `wasm_config_t`

lib/c-api/src/wasm_c_api/engine/config/jsc.rs
```diff
@@ -0,0 +1,38 @@
...
+/// This is a Wasmer-specific type with Wasmer-specific functions for
+/// manipulating it.
+///
+/// cbindgen:ignore
+#[repr(C)]
+#[derive(Debug, Default)]
+pub struct wasmer_jsc_engine_config_t;
+
+/// Create a new  [`wasm_engine_t`] backed by a `jsc` engine.
+pub fn wasm_jsc_engine_new_with_config(config: wasm_config_t) -> Option<Box<wasm_engine_t>> {
```
syrusakbary, line 19: This should receive a `wasmer_jsc_engine_config_t`

lib/c-api/src/wasm_c_api/engine/config/mod.rs
```diff
@@ -0,0 +1,54 @@
+#[cfg(feature = "sys")]
+pub mod sys;
+#[cfg(feature = "sys")]
+pub use sys::wasm_config_set_sys_compiler;
```
syrusakbary, line 4: I'd call this `wasm_config_sys_set_compiler`, so we can customize more settings in the future via `wasm_config_sys_*`

lib/c-api/src/wasm_c_api/engine/config/jsc.rs
xdoardo: Because this would mean introducing new `wasm(er)_config_js_new`. Currently we only have `wasm_config_new`, which returns a `wasm_config_t`. With the current API, the check is justified by the fact that code like 
```
let mut config = wasm_config_new();
wasm_config_set_engine(&mut config, V8);
wasm_config_sys_set_target(&mut config, my_target); 
``` 
does not produce type errors. 

Do you suggest to introduce the `wasm(er)_config_<backend>_new` functions? This would also mean duplicating the `features` field that we currently have in the `wasm_config_t` type.

lib/c-api/src/wasm_c_api/engine/config/mod.rs
xdoardo: [0f304c4](https://github.com/wasmerio/wasmer/pull/5443/commits/0f304c409bc12cabb57ae28af7a5bfc146120dd4)

lib/api/src/backend/mod.rs
```diff
@@ -47,3 +47,69 @@ pub enum BackendKind {
     /// The `jsc` runtime.
     Jsc,
 }
+
+impl Default for BackendKind {
+    fn default() -> Self {
+        #[cfg(feature = "sys-default")]
```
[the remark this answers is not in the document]
xdoardo: Or, wait, collide in what sense?

lib/c-api/src/wasm_c_api/engine/config/jsc.rs
syrusakbary: Then why is `wasm_jsc_engine_new_with_config` public and how is expected to be used?

I'd expect the users just call `wasm_engine_new_with_config` and never use the other function

xdoardo: Ah, that explains it. All these engine-specific config functions should be `pub(crate)` and used only in the top-level `wasm_engine_new_with_config` and not exposed to users.

lib/c-api/src/wasm_c_api/engine/config/wasmi.rs
```diff
@@ -0,0 +1,38 @@
...
+/// This is a Wasmer-specific type with Wasmer-specific functions for
+/// manipulating it.
+///
+/// cbindgen:ignore
+#[repr(C)]
+#[derive(Debug, Default)]
+pub struct wasmer_wasmi_engine_config_t;
+
+/// Create a new  [`wasm_engine_t`] backed by a `wasmi` engine.
+pub fn wasm_wasmi_engine_new_with_config(config: wasm_config_t) -> Option<Box<wasm_engine_t>> {
```
syrusakbary, line 19: This should not be public

lib/c-api/src/wasm_c_api/engine/config/sys.rs
```diff
@@ -0,0 +1,436 @@
...
+        let sys_config = wasmer_sys_engine_config_t {
+            nan_canonicalization: enable,
+            ..Default::default()
+        };
+        config.engine_config = wasmer_engine_config_t::Sys(sys_config);
+    }
+}
+
+/// Create a new  [`wasm_engine_t`] backed by a `sys` engine.
+pub fn wasm_sys_engine_new_with_config(config: wasm_config_t) -> Option<Box<wasm_engine_t>> {
```
syrusakbary, line 315: This should not be public

lib/c-api/src/wasm_c_api/engine/config/jsc.rs
syrusakbary, line 19: This should not be public

xdoardo: [5c08932](https://github.com/wasmerio/wasmer/pull/5443/commits/5c08932b8602ade591afaabe379c52da6f8d57d4)

lib/c-api/src/wasm_c_api/engine/config/sys.rs
xdoardo: [5c08932](https://github.com/wasmerio/wasmer/pull/5443/commits/5c08932b8602ade591afaabe379c52da6f8d57d4)

lib/c-api/src/wasm_c_api/engine/config/v8.rs
```diff
@@ -0,0 +1,38 @@
...
+/// This is a Wasmer-specific type with Wasmer-specific functions for
+/// manipulating it.
+///
+/// cbindgen:ignore
+#[repr(C)]
+#[derive(Debug, Default)]
+pub struct wasmer_v8_engine_config_t;
+
+/// Create a new  [`wasm_engine_t`] backed by a `v8` engine.
+pub fn wasm_v8_engine_new_with_config(config: wasm_config_t) -> Option<Box<wasm_engine_t>> {
```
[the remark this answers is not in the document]
xdoardo: [5c08932](https://github.com/wasmerio/wasmer/pull/5443/commits/5c08932b8602ade591afaabe379c52da6f8d57d4)

lib/c-api/src/wasm_c_api/engine/config/wamr.rs
```diff
@@ -0,0 +1,38 @@
...
+/// This is a Wasmer-specific type with Wasmer-specific functions for
+/// manipulating it.
+///
+/// cbindgen:ignore
+#[repr(C)]
+#[derive(Debug, Default)]
+pub struct wasmer_wamr_engine_config_t;
+
+/// Create a new  [`wasm_engine_t`] backed by a `wamr` engine.
+pub fn wasm_wamr_engine_new_with_config(config: wasm_config_t) -> Option<Box<wasm_engine_t>> {
```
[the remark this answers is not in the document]
xdoardo: [5c08932](https://github.com/wasmerio/wasmer/pull/5443/commits/5c08932b8602ade591afaabe379c52da6f8d57d4)

lib/c-api/src/wasm_c_api/engine/config/wasmi.rs
xdoardo: [5c08932](https://github.com/wasmerio/wasmer/pull/5443/commits/5c08932b8602ade591afaabe379c52da6f8d57d4)

syrusakbary (approved):

lib/c-api/src/wasm_c_api/unstable/engine.rs
```diff
@@ -163,7 +63,13 @@ pub extern "C" fn wasmer_is_headless() -> bool {
 /// compiled library.
 #[no_mangle]
 pub extern "C" fn wasmer_is_engine_available(engine: wasmer_engine_t) -> bool {
-    matches!(engine, wasmer_engine_t::UNIVERSAL if cfg!(feature = "compiler"))
+    match engine {
+        wasmer_engine_t::UNIVERSAL => cfg!(feature = "sys"),
+        wasmer_engine_t::V8 => cfg!(feature = "v8"),
+        wasmer_engine_t::WASMI => cfg!(feature = "wasmi"),
+        wasmer_engine_t::WAMR => cfg!(feature = "wamr"),
+        wasmer_engine_t::JSC => cfg!(feature = "wasmi"),
```
syrusakbary, line 71: This is wrong. it should be `jsc`:

```suggestion
        wasmer_engine_t::JSC => cfg!(feature = "jsc"),
```

lib/c-api/src/wasm_c_api/engine/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+#[no_mangle]
+pub extern "C" fn wasm_engine_new_with_config(
+    config: Option<Box<wasm_config_t>>,
+) -> Option<Box<wasm_engine_t>> {
+    #[allow(unused)]
+    let config = *(config?);
+
+    match config.engine {
+        #[cfg(feature = "sys")]
+        wasmer_engine_t::UNIVERSAL => config::sys::wasm_sys_engine_new_with_config(config),
```
syrusakbary, line 148: Should we have the Cranelift / LLVM / Singlepass as well here?

xdoardo: Perhaps, to keep symmetry with the (non-c-) api, we should also rename `wasmer_engine_t` to `wasmer_backend_t`.

syrusakbary (approved):

Issue #5444, opened 2025-02-28
Check `WasmSliceAccess` alignment before reading from it


xdoardo:
Currently, the code that create slice accesses [does not check](https://github.com/wasmerio/wasmer/blob/main/lib/api/src/utils/mem/access.rs#L249) for alignment of the pointer: 

```
            let buf_ptr: *mut u8 = slice.buffer.base().add(slice.offset as usize);
            let buf_ptr: *mut T = std::mem::transmute(buf_ptr);
            std::slice::from_raw_parts_mut(buf_ptr, slice.len as usize)
```

This can cause panics, since the `offset` is given as input (to the `WasmSlice`).

PR #5449, opened 2025-03-03
fix(api): Check if `WasmSliceAccess` ptr is aligned before reading

xdoardo:
As per title. Fixes https://github.com/wasmerio/wasmer/issues/5444

### DIFF
--- a/lib/api/src/utils/mem/mod.rs
+++ b/lib/api/src/utils/mem/mod.rs
@@ -28,6 +28,9 @@ pub enum MemoryAccessError {
     /// String is not valid UTF-8.
     #[error("string is not valid utf-8")]
     NonUtf8String,
+    /// Pointer to memory is unaligned.
+    #[error("unaligned pointer read")]
+    UnalignedPointerRead,
 }
 
 impl From<MemoryAccessError> for RuntimeError {
--- a/lib/api/src/utils/mem/access.rs
+++ b/lib/api/src/utils/mem/access.rs
@@ -246,6 +246,9 @@ where
         let buf = unsafe {
             let buf_ptr: *mut u8 = slice.buffer.base().add(slice.offset as usize);
             let buf_ptr: *mut T = std::mem::transmute(buf_ptr);
+            if !buf_ptr.is_aligned() {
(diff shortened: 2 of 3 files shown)


syrusakbary (approved):

lib/api/src/utils/mem/mod.rs
```diff
@@ -103,8 +106,11 @@ impl<'a, T: ValueType> WasmRef<'a, T> {
     #[inline]
     pub fn read(self) -> Result<T, MemoryAccessError> {
         let mut out = MaybeUninit::uninit();
-        let buf =
-            unsafe { slice::from_raw_parts_mut(out.as_mut_ptr() as *mut u8, mem::size_of::<T>()) };
+        let buf_ptr = out.as_mut_ptr() as *mut u8;
+        if !buf_ptr.is_aligned() {
+            return Err(MemoryAccessError::UnalignedPointer);
+        }
+        let buf = unsafe { slice::from_raw_parts_mut(buf_ptr, mem::size_of::<T>()) };
```
charmitro, line 113: Isn't `.is_aligned()` always going to return true due to [this](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#layout-1)?

> _MaybeUninit<T> is guaranteed to have the same size, alignment, and ABI as T:_

Arshia001: Depends... These guarantees only hold as long as we don't have unsafe code, and we do have a lot of unsafe code near the syscall boundaries. Still, @xdoardo would know better.

Arshia001: But also, since we're casting to `*mut u8`... Let's add a test?

Arshia001 (changes requested):

lib/api/src/utils/mem/mod.rs
xdoardo: [b7cd7a4](https://github.com/wasmerio/wasmer/pull/5449/commits/b7cd7a4a384df9e95c8137924ba58412caa5e381)

PR #5455, opened 2025-03-06
Engine selection improved

syrusakbary:
This PR aims to simplify the logic for backends

cc @xdoardo @charmitro


lib/cli/src/backend.rs
```diff
@@ -844,116 +665,63 @@ impl BackendType {
...
 
 impl From<&BackendType> for wasmer::BackendKind {
     fn from(backend_type: &BackendType) -> Self {
         match backend_type {
-            BackendType::Singlepass => wasmer::BackendKind::Sys,
-            BackendType::Cranelift => wasmer::BackendKind::Sys,
-            BackendType::LLVM => wasmer::BackendKind::Sys,
+            BackendType::Singlepass => wasmer::BackendKind::Singlepass,
+            BackendType::Cranelift => wasmer::BackendKind::Cranelift,
+            BackendType::LLVM => wasmer::BackendKind::LLVM,
```
charmitro, line 708: When building with LLVM disable,
```
error[E0599]: no variant or associated item named `LLVM` found for enum `BackendKind` in the current scope
   --> lib/cli/src/backend.rs:708:55
    |
708 |             BackendType::LLVM => wasmer::BackendKind::LLVM,
    |                                                       ^^^^ variant or associated item not found in `BackendKind`
```

Should we fallback to Cranelift in this case or headless? @syrusakbary  @xdoardo

syrusakbary: I checked the code, this is now solved

PR #5451, opened 2025-03-04
Select engine based on WASM module feature requirements

charmitro:
Implement automatic WebAssembly feature detection to choose compatible
backends:

  - Add `default_features_for_target` and
    `supported_features_for_target` on Engine
  - Implement feature detection from WebAssembly modules
  - Add special handling for exceptions support detection
  - Respect user-specified backends when provided
  - Provide backend filtering based on feature support
  - Enable automatic LLVM selection for modules with exceptions

PR #5458, opened 2025-03-06
Select engine based on WASM module feature requirements (v2)

xdoardo:
This PR builds on https://github.com/wasmerio/wasmer/pull/5451, fixing various build errors, feature flags. It also adds support for feature detection from incoming webcs. 

cc: @syrusakbary , @charmitro

### DIFF
--- a/lib/package/src/package/manifest.rs
+++ b/lib/package/src/package/manifest.rs
@@ -16,6 +16,8 @@ use webc::{
     sanitize_path,
 };
 
+use crate::utils::features_to_wasm_annotations;
+
 use webc::metadata::{
     annotations::{
         Atom as AtomAnnotation, FileSystemMapping, FileSystemMappings, VolumeSpecificPath, Wapm,
@@ -304,29 +306,7 @@ fn transform_atoms_shared(
 
         if let Ok(features) = features_result {
             // Convert wasmer_types::Features to webc::metadata::annotations::Wasm
-            let mut feature_strings = Vec::new();
-
-            if features.simd {
-                feature_strings.push("simd".to_string());
-            }
-            if features.bulk_memory {
-                feature_strings.push("bulk-memory".to_string());
-            }
-            if features.reference_types {
-                feature_strings.push("reference-types".to_string());
-            }
-            if features.multi_value {
-                feature_strings.push("multi-value".to_string());
-            }
-            if features.threads {
-                feature_strings.push("threads".to_string());
-            }
-            if features.exceptions {
-                feature_strings.push("exception-handling".to_string());
-            }
-            if features.memory64 {
-                feature_strings.push("memory64".to_string());
-            }
+            let feature_strings = features_to_wasm_annotations(&features);
 
             // Only create annotation if we detected features
             if !feature_strings.is_empty() {
--- a/lib/package/src/package/manifest.rs
+++ b/lib/package/src/package/manifest.rs
@@ -296,9 +296,59 @@ fn transform_atoms_shared(
     let mut metadata = IndexMap::new();
 
     for (name, (kind, content)) in atoms.iter() {
+        // Create atom with annotations including Wasm features if available
+        let mut annotations = IndexMap::new();
+
+        // Detect required WebAssembly features by analyzing the module binary
+        let features_result = wasmer_types::Features::detect_from_wasm(content);
+
+        if let Ok(features) = features_result {
+            // Convert wasmer_types::Features to webc::metadata::annotations::Wasm
+            let mut feature_strings = Vec::new();
+
+            if features.simd {
+                feature_strings.push("simd".to_string());
+            }
+            if features.bulk_memory {
+                feature_strings.push("bulk-memory".to_string());
+            }
+            if features.reference_types {
+                feature_strings.push("reference-types".to_string());
+            }
+            if features.multi_value {
+                feature_strings.push("multi-value".to_string());
+            }
+            if features.threads {
+                feature_strings.push("threads".to_string());
+            }
+            if features.exceptions {
+                feature_strings.push("exception-handling".to_string());
+            }
+            if features.memory64 {
+                feature_strings.push("memory64".to_string());
+            }
+
+            // Only create annotation if we detected features
+            if !feature_strings.is_empty() {
+                let wasm = webc::metadata::annotations::Wasm::new(feature_strings);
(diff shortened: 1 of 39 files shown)


lib/types/src/features.rs
```diff
@@ -264,6 +265,126 @@ impl Features {
...
+    /// 3. Parsing the module to detect certain common patterns
+    ///
+    /// # Arguments
+    ///
+    /// * `wasm_bytes` - The binary content of the WebAssembly module
+    ///
+    /// # Returns
+    ///
+    /// A new `Features` instance with the detected features enabled.
+    pub fn detect_from_wasm(wasm_bytes: &[u8]) -> Result<Self, wasmparser::BinaryReaderError> {
```
syrusakbary, line 284: This needs to be either in wasmer or wasmer-cli to not force the wasmparser dependency on wasmer_types.

Once we move it to wasmer, we should put it behind a feature flag (as wasmparser might not always be available), which should be named `detect-wasm-features`

syrusakbary: It's a utility similar to `wat`/`wat2wasm`

syrusakbary: TL;DR this is needed by package, so we can't move to wasmer. We will feature flag this function for now

charmitro: @xdoardo did the feature flag, resolving and opening a linear ticket for it.

lib/wasix/src/bin_factory/binary_package.rs
```diff
@@ -60,35 +63,18 @@ impl BinaryPackageCommand {
...
-            features.exceptions(true);
-
-            tracing::debug!("WebC module features from annotations: {:?}", features);
-            return Some(features);
+        // Return only the pre-computed features from the container manifest
+        if let Some(features) = &self.features {
+            tracing::debug!(
+                "Using pre-computed WebAssembly features from annotations: {:?}",
+                features
+            );
```
syrusakbary, line 71: Not needed, too verbose and only needed for the PR. If any logging, we should do it at the moment of reading the webc with a serialization of the whole manifest

lib/wasix/src/bin_factory/binary_package.rs
```diff
@@ -60,35 +63,18 @@ impl BinaryPackageCommand {
...
-                tracing::debug!("WebC module features detected from binary: {:?}", features);
-                Some(features)
-            }
-            Err(err) => {
-                tracing::warn!("Failed to detect WebAssembly features from atom: {}", err);
-                None
-            }
-        }
+        // If no annotations were found, return None
+        tracing::debug!("No WebAssembly feature annotations found for this module");
```
syrusakbary, line 76: Not useful tracing, just for this PR

lib/wasix/src/runtime/package_loader/load_package_tree.rs
```diff
@@ -180,7 +180,54 @@ fn load_binary_command(
...
+            features.simd(wasm_annotation.features.contains(&"simd".to_string()));
+            features.bulk_memory(
+                wasm_annotation
+                    .features
+                    .contains(&"bulk-memory".to_string()),
+            );
+            features.reference_types(
+                wasm_annotation
+                    .features
+                    .contains(&"reference-types".to_string()),
```
syrusakbary, line 206: This should be abstracted out, on the same layer that we are doing write/read.

Basically we should have two functions:

* `Features` -> Wasm annotations
And the reverse
*  Wasm annotations -> `Features`

Both living on the same space. One should be used when building a webc binary. Other when reading from a webc binary (aka here)

lib/wasix/src/runtime/package_loader/load_package_tree.rs
```diff
@@ -245,11 +292,58 @@ fn legacy_atom_hack(
...
+        if let Ok(Some(wasm_annotation)) = atom_metadata
+            .annotation::<webc::metadata::annotations::Wasm>(webc::metadata::annotations::Wasm::KEY)
+        {
+            tracing::debug!(
+                "Found wasm features in atom annotation: {:?}",
+                wasm_annotation
+            );
+
+            // Create features object from the annotation strings
+            let mut features = wasmer_types::Features::default();
```
syrusakbary, line 306: This logic is duplicated. It should be one function as previously commented

lib/wasix/src/runtime/package_loader/load_package_tree.rs
```diff
@@ -245,11 +292,58 @@ fn legacy_atom_hack(
...
+    // Get WebAssembly features from manifest atom annotations
+    let features = if let Some(atom_metadata) = webc.manifest().atoms.get(&name) {
+        if let Ok(Some(wasm_annotation)) = atom_metadata
+            .annotation::<webc::metadata::annotations::Wasm>(webc::metadata::annotations::Wasm::KEY)
+        {
+            tracing::debug!(
+                "Found wasm features in atom annotation: {:?}",
+                wasm_annotation
+            );
+
```
syrusakbary, line 304: Tracing not needed, for reasons exposed before

lib/wasix/src/runtime/package_loader/load_package_tree.rs
charmitro: IMO only Wasm annotations -> Features is needed. The otherway around is only used once.

lib/package/src/package/manifest.rs
```diff
@@ -296,9 +296,59 @@ fn transform_atoms_shared(
...
+                feature_strings.push("reference-types".to_string());
+            }
+            if features.multi_value {
+                feature_strings.push("multi-value".to_string());
+            }
+            if features.threads {
+                feature_strings.push("threads".to_string());
+            }
+            if features.exceptions {
+                feature_strings.push("exception-handling".to_string());
```
syrusakbary, line 325: Can we abstract this in another function, next to the one where we do the read?

lib/wasix/src/runtime/package_loader/load_package_tree.rs
syrusakbary: Even if is used once, having logic in two different places can become confusing and is good to tie them both together in one file

charmitro: Should be OK now.

syrusakbary (approved):
LGTM. But we are missing tests

PR #5469, opened 2025-03-07
Use engine-specific default features

charmitro:
Uses the engine's default features for backend instead of generic default features.


xdoardo:
Could you also add a log message such as [this](https://github.com/wasmerio/wasmer/blob/94726f9608cc05a3cfc54378970b9bbdfd743038/lib/cli/src/commands/run/mod.rs#L133) one using `engine.deterministic_id()`?

syrusakbary (approved):

PR #5472, opened 2025-03-10
feat(types): add supports_all_features utility function

charmitro:
Add a helper method to verify whether a compiler's supported features can handle a module's required features. This allows runtime validation of compiler selection based on WebAssembly feature requirements.

Suggested-by: Christoph Herzog <christoph@wasmer.io>

### DIFF
--- a/lib/types/src/features.rs
+++ b/lib/types/src/features.rs
@@ -387,6 +387,23 @@ impl Features {
 
         Ok(features)
     }
+
+    /// Checks if all needed features are supported by a specific compiler
+    pub fn supports_all_features(needed_features: &Features, compiler_features: &Features) -> bool {
+        // Check each needed feature that is true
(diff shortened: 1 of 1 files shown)


lib/types/src/features.rs
```diff
@@ -387,6 +387,23 @@ impl Features {
 
         Ok(features)
     }
+
+    /// Checks if all needed features are supported by a specific compiler
+    pub fn supports_all_features(needed_features: &Features, compiler_features: &Features) -> bool {
```
syrusakbary, line 392: How is this different from `contains_features` ? Basically supports `supports_all_features(needed_features, compiler_features)` is the same as `compiler_features.contains_features(needed_features)`, isn't it?

charmitro:
Closing, functionality already covered via `contains_features(&self, required: &Self)`.

PR #5475, opened 2025-03-12
Disallow mounting folders on the guest's root for WASIX modules

Arshia001:


lib/wasix/src/runtime/package_loader/load_package_tree.rs
```diff
@@ -488,6 +488,12 @@ fn filesystem_v2(
             continue;
         }
 
+        if mount_path.as_path() == Path::new("/") {
+            bail!(
+                "The \"{package}\" package wants to mount a volume at \"/\", but that's not allowed",
+            );
+        }
```
charmitro, line 495: Maybe a helper function for this? It is used in fs v2 and v3. Something like (not tested),
```
fn check_volume_path(
    mount_path: &PathBuf,
    package: &str,
) -> Result<(), Error> {
    if mount_path.as_path() == Path::new("/") {
        bail!(
            "The \"{package}\" package wants to mount a volume at \"/\", but that's not allowed",
        );
    }
    Ok(())
}
```

lib/cli/src/commands/run/wasi.rs
```diff
@@ -491,8 +492,14 @@ impl Wasi {
 
                 MappedDirectory {
                     host: current_dir,
-                    guest: MAPPED_CURRENT_DIR_DEFAULT_PATH.to_string(),
+                    guest: if is_wasix {
+                        MAPPED_CURRENT_DIR_DEFAULT_PATH.to_string()
+                    } else {
+                        "/".to_string()
```
charmitro, line 498: I'd love a const about `"/"`, `const ROOT_PATH: &str = "/";` or similar. What do you think?

charmitro:
Also,

Tested-by: Charalampos Mitrodimas <charalampos@wasmer.io>

lib/wasix/src/runtime/package_loader/load_package_tree.rs
Arshia001: if only the code was that clean to begin with... it has been my experience that attempting to make clean implementations in code that's not clean by design usually leads to more complexity down the line, so when I see duplicate code, I also duplicate mine, to at least keep the code *consistently* unclean.

lib/cli/src/commands/run/wasi.rs
Arshia001: I'd happily approve of someone else doing that XD `/` is well-known enough to warrant not giving it a const of its own though.

syrusakbary:
This PR might have some implications, that we don't want. I remember other programs using `mapdir=.:/` to make Python work for example. We may want to use that directory instead (and not mount any other dir) if / is provided. Lets follow up in a sync

Arshia001:
@syrusakbary in fact, that's exactly what this PR is trying to prevent. Mounting things on `/` messes up a lot of assumptions in WASIX, including:

* We put commands from packages in `/bin` and `/usr/bin`. If `/` is mounted, those will get overwritten and be inaccessible.
* Packages mount volumes, which contain necessary files without which they won't run. Python is in fact a great example of this. If I were to just mount something on `/`, I'd lost all the data files that came with the python package, ending up with a broken application.
  * Same with PHP, if I mount a website's root on `/`, I lose all the openssl data files and HTTPS will be broken for example.
  * I do believe python _was_ the motivation behind the custom behavior of mounting `.` on `/home` with `--dir .`. I think that's what you're thinking of as well.

marxin:
@Arshia001 The PR seems to be pretty close to be merged, can you rebase it?

syrusakbary (commented):
Need tests (and convert this into a warning)

Arshia001:
@syrusakbary since we turned the errors into warnings, there's not much we can actually test.

lib/wasix/src/runtime/package_loader/load_package_tree.rs
zebreus: And that is why our code is not getting any better :(

lib/cli/src/commands/run/wasi.rs
```diff
@@ -569,7 +589,11 @@ impl Wasi {
 
                 MappedDirectory {
                     host: resolved_host,
-                    guest: MAPPED_CURRENT_DIR_DEFAULT_PATH.to_string(),
+                    guest: if is_wasix {
+                        MAPPED_CURRENT_DIR_DEFAULT_PATH.to_string()
+                    } else {
+                        "/".to_string()
+                    },
```
zebreus, line 596: This is the only thing that really changed. I don't understand why this change is made and if it is an improvement.

zebreus: @Arshia001  If you are 100% certain this is a good idea, then go ahead and merge

Arshia001: @zebreus this is kind of the main issue! With WASI modules, nothing is mounted in the module's FS by default, so it's OK to mount wherever. With WASIX, we mount a lot of stuff in by default, so you don't want a mount on the root.

Now, with WASI, since there is no host-side CWD and you're at `/` by default, you just have to mount directly in `/` to make the files available to the module with relative paths. With WASIX, we mount into `/home` and cd into it.

zebreus: What I don't understand is that when `MAPPED_CURRENT_DIR_DEFAULT_PATH` is already defined to `/home` and that seems to work just fine.

If I understand you correctly the reason for the change is to make sure that the mountpoint is always the CWD, which it can't be for WASI. But we also can't always mount to `/` because that won't work for WASIX...

zebreus: So in the past, the dir was just mounted to `/home` for WASI and now it isn't anymore. While this could be breaking, it's probably an improvement. However I don't like that it's one more inconsistency.

PR #5481, opened 2025-03-16
epoll_wait timeout is in milliseconds

cheplv:
Expecting that timeout must be in milliseconds.
Currently it is in nanoseconds so we max can allocate 2.1 seconds timeout in int32 function argument.

# Description
Porting nodejs to wasm having libuv and other epoll_(p)wait function expect that duration of timeout is in milliseconds.
Currently nanoseconds is used so we can use maximum 2.1second timeout on int32 wasm parameter.

### DIFF
--- a/lib/wasix/src/syscalls/wasix/epoll_wait.rs
+++ b/lib/wasix/src/syscalls/wasix/epoll_wait.rs
@@ -126,7 +126,7 @@ pub fn epoll_wait<'a, M: MemorySize + 'static>(
         let timeout = if timeout == TIMEOUT_FOREVER {
             None
         } else {
-            Some(ctx.data().tasks().sleep_now(Duration::from_nanos(timeout)))
+            Some(ctx.data().tasks().sleep_now(Duration::from_millis(timeout)))
         };
         async move {
             if let Some(timeout) = timeout {


Arshia001:
Hey @cheplv, thanks for this contribution! Let me have a look at the existing WASIX code and see what it expects, as we may need to calibrate wasix-libc etc. as well.

Arshia001:
so, as it turns out, the syscall code is actually correct. The parameter to the syscall is of type `__wasi_timestamp_t`, which is a 64-bit nanosecond precision timestamp.

What's broken is the `epoll_pwait` implementation in wasix-libc, here:

https://github.com/wasix-org/wasix-libc/blob/e2329b4c10a21b5ee5d2b5d25cd42cd17fb37b60/libc-top-half/musl/src/linux/epoll.c#L55-L64

Since the parameter to `epoll_pwait` is a 32-bit millisecond precision timestamp, that code should cast to uint64 and multiply by 1000000 before passing the `to` parameter to the wasix syscall. I'll push a fix to wasix-libc in a minute, and you can use the updated wasix-libc in your build process. Closing this PR.

cheplv:
@Arshia001 Yes, this is correct.
Found another one bug in implementation of epoll - there is no filtering of events after epoll_ctl_add only EPOLLIN - receive EPOLLOUT event.
With -v -v -v for verbose - there is event happens from system and seems passed without filtering.
Anyway - for now I switched nodejs build to posix-poll (poll_oneoff) and it works correct.

Arshia001:
Can you please create a separate issue for the other epoll bug? thanks!

Are you building nodejs to wasix? I wasn't aware that it's possible to run v8 in WASM.

cheplv:
@Arshia001 I will return back to epoll issue later due this is not a main issue right now.
Yes, nodejs was build again wasix and working fine with networking.
Uploaded video to youtube: https://youtu.be/EBhoep7Ld78

cheplv:
@syrusakbary Nice to meet you. Will reach you by email shortly.

PR #5482, opened 2025-03-18
deps: Update to zip 2.4.x

Pr0methean:
Due to https://github.com/zip-rs/zip2/security/advisories/GHSA-94vh-gphv-8pm8, the `zip` crate should be updated to 2.4.x immediately, to prevent specially-crafted templates from writing files outside the destination directory.


xdoardo:
Thanks! Looks good to me, let's see if CI is green now.

xdoardo (approved):

PR #5498, opened 2025-04-02
Simply object initialization path with relocations

wakabat:
Previous code relies on C glue code to piece metadata and actual function addresses after linking together. This change refines the process, so proper relocations are generated for each function / trampoline when building the object. This way a linker should be able to fulfill the task of filling in actual function addresses in the final binary, there is no need to use glue code to do the memcpy work anymore.


syrusakbary:
Thanks for the PR! We need to test the branch locally, and try a few cases... but in general looks good!

wakabat:
Rebased against tip.

xdoardo (approved):

syrusakbary:
We tested locally (in aarch64 and x86_64) and it works well. We should be good to merge once the majority of tests pass (some jobs seem to be failing, but are unrelated to this PR)

Issue #5496, opened 2025-04-01
LLVM engine fails producing AOT binary from golang's outputs


wakabat:
### Describe the bug

Current tip has issue building runnable native binaries using LLVM engine from WASM programs generated by golang.

### Steps to reproduce

```
$ git clone https://github.com/wakabat/minimal-go-example
$ cd minimal-go-example
$ GOOS=wasip1 GOARCH=wasm go build -o main.wasm .
$ wasmer run main.wasm
$ wasmer create-exe main.wasm -o main_singlepass --singlepass -m avx
$ ./main_singlepass
WASI exited with code: ExitCode::0
$ wasmer create-exe main.wasm -o main_cranelift --cranelift
$ ./main_cranelift
WASI exited with code: ExitCode::0
```

The above steps all work as expected, but the following one does not:

```
$ wasmer create-exe main.wasm -o main_llvm --llvm
$ ./main_llvm
out of bounds memory access
```

The build process completes, but running the program fails.

I've done a little bisecting, and navigate to this PR: https://github.com/wasmerio/wasmer/pull/5142 However it is too huge so I cannot pinpoint the exact commit causing the error.

xdoardo:
Very interesting. Thanks, I'll tackle it as soon as possible.

wakabat:
One more finding:

This command actually works:

```
$ wasmer create-exe main.wasm -o main_llvm --llvm -m avx
$ ./main_llvm
WASI exited with code: ExitCode::0
```

But this does not:

```
$ wasmer create-exe main.wasm -o main_llvm --llvm
$ ./main_llvm
out of bounds memory access
```

I'm testing on a Ryzen 3900x chip, lldb also shows that the failing instruction is [punpckldq](https://www.felixcloutier.com/x86/punpcklbw:punpcklwd:punpckldq:punpcklqdq), so maybe the issue has to do with AVX2 support.

wakabat:
I think I've found out what is happening here, please refer to this PR: https://github.com/wasmerio/wasmer/pull/5506

xdoardo:
I think this was fixed with the recent PRs to fix create-exe. I tested this on my machine (aarch64-darwin). @wakabat, are you still seeing the issue?

xxuejie:
Nope, I'm not seeing this issue anymore after recent fixes.

wakabat:
Nope I'm not seeing this anymore, closing it, thanks!

PR #5506, opened 2025-04-12
fix(compiler): Preserve original alignment of custom sections

wakabat:
In LLVM engine(could also happen in other engines as well, but it is the LLVM engine that gets me to look into this bug), certain custom sections in generated objects have alignment settings. For example, `.rodata.cst16` might use an alignment of 16, while `.rodata.cst8` uses an alignment of 8.

However, current compiler would ignore those alignment settings when merging multiple functions into a single object. This change fixes the code so proper alignment setup is respected.

Note even though alignment value is inserted to CustomSection(which is a structure defined in wasmer-types and shared by all 3 engines), current change only fixes the alignment value in LLVM engine. My test does not reveal the necessity of alignment values in the other 2 engines, however a second confirmation might always be useful.

This might be a potential fix for #5496


syrusakbary:
That's an awesome fix. We are waiting for a full review of `create-exe`, but please keep the PRs coming... we'll probably merge them prior to the upcoming 6.0 release (note: the build failures have likely nothing to do with this PR)

xdoardo:
Thanks! Awesome. I have been using this branch to fix more errors happening on arm64 with create-exe - namely, the metadata header size being 0 when read off Mach-O objects. Thanks again, we'll merge this as soon as I fix the remaining problems with LLVM on arm64.

PR #5518, opened 2025-04-22
feat: Add `SuggestedCompilerOptimization` type(s) to automatically infer when to enable g0m0 optimization

xdoardo:
As per title. This PR introduces two (equally-shaped) types in `wasmer_config` (to refer to the annotation observed from a Wasmer manifest) and in `wasmer_types::target` (for the actual types to pass to engines) that represent annotations from users, which can be used to specify suggested compiler optimizations. For example: 
```toml
# ....
[[module]]
name = "php"
source = "sapi/cli/php.wasm"
abi = "wasi"
annotations.suggested_compiler_optimizations = { pass_params = true }
# ...
```

This PR also adds the code necessary to serialize and deserialize this metadata in the manifest (and keep it in the related webc).

### DIFF
--- a/lib/types/src/target.rs
+++ b/lib/types/src/target.rs
@@ -230,12 +230,12 @@ impl Default for Target {
     }
 }
 
-/// Suggested optimization that might be operated on the module when (and if) compiled.
+/// User-suggested optimization that might be operated on the module when (and if) compiled.
 ///
 // Note: This type is a copy of `wasmer_config::package::SuggestedCompilerOptimizations`, so to
 // avoid dependencies on `wasmer_config` for crates that already depend on `wasmer_types`.
 #[derive(Clone, Debug, PartialEq, Eq, Default)]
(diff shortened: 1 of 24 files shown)


lib/types/src/target.rs
```diff
@@ -229,3 +229,13 @@ impl Default for Target {
         }
     }
 }
+
+/// Suggested optimization that might be operated on the module when (and if) compiled.
+///
+// Note: This type is a copy of `wasmer_config::package::SuggestedCompilerOptimizations`, so to
+// avoid dependencies on `wasmer_config` for crates that already depend on `wasmer_types`.
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+pub struct SuggestedCompilerOptimizations {
```
syrusakbary, line 238:
```suggestion
pub struct UserCompilerOptimizations {
```

syrusakbary (approved):

PR #5521, opened 2025-04-23
feat: Enable rv32 target in LLVM engine

wakabat:
Simply turning on rv32 target can be trivial in LLVM setting, all we need is just tweaking a few configs. However, this change also comes with several bug fixes that are required in rv32 (possibly other 32 bit target targets as well), and in rv64:

* Certain pointer width are hardcoded to be 8, which will break on 32-bit platforms
* Metadata require an alignment of 8
* RISC-V relocations are not properly handled
* Certain floating-point & 64-bit routine functions are required in libcalls
* Previous transmute-based way of setting ABI in rv64 platform would
crash, this change also fixes it

I believe those additional fixes might be quite valuable as well, hence I put this PR together.


wakabat:
In addition to this PR, there is one additional bullet that I want to run with your suggestions first: as of right now, the features to be enabled for RISC-V targets are [hardcoded](https://github.com/wasmerio/wasmer/blob/3c648758f3f1dcebe688a9f0c0950b0bf7705bee/lib/compiler-llvm/src/config.rs#L236) and cannot be changed. In my use cases, I need to tweak those feature set, for example, I might want to turn off floating point features, while adding other new features.

Current [CpuFeature](https://github.com/wasmerio/wasmer/blob/3c648758f3f1dcebe688a9f0c0950b0bf7705bee/lib/types/src/target.rs#L33-L52) enum only defines certain x86 and ARM features, even if I can add new RISC-V features there, there is no way we can use command line arguments to disable one or more features. One possible path I can think of, is to add a new `Custom` feature:

```
pub enum CpuFeature {
    // ...
   Custom(String),
}
```

And you can use the `Custom` variant to inject features like `-f` or `-d`, the LLVM engine can be modified accordingly to pick up those disabling of features. However, `CpuFeature` implements `EnumSetType`, which makes it impossible to add `Custom(String)` variant.

In this sense, how do you suggest we allow disabling of certain features?

marxin:
@wakabat sorry the PR review is overdue. Generally speaking, we would be interested in the target support - if you're still interested, please rebase the PR and I'm going to review it then instantly.

wakabat:
@marxin Thanks for the reply! We definitely love to submit it upstream, let me clean this PR a bit(there might be some new fixes we found), and then it should be good for review.

wakabat:
Replaced by #6007

PR #5526, opened 2025-04-23
feat(wasix): Improve module loading performance and API correctness

theduke:
This commit contains two notable changes:

* The VirtualFile trait is extended with a new as_owned_buffer() method.
  This allows cheap copying of file data if the file is already fully in
  memory, and enables optimisations for code that is aware of this
  functionality by allowing zero-copy cloning.
  This is especially useful for data that is already mmaped.

  The new method has an empty default impl, but is implemented for
  appropriate files (ReadOnlyFile, webc volume files).

* Modifies the Runtime::load_module method to take a new
  `HashedModuleData` struct

  This has multiple benefits:
  - HashedModuleData is safe by construction, thanks to private fields,
    so the hash it contains will always be valid for the module data.
    This removes a source of bugs due to passing a wrong hash for module
    data to module loaders etc.
  - HashedModuleData can be cheaply creating from a
    BinaryPackageCommand, preventing redundant cloning of data
  - The load_module method now receives owned data that is cheaply
    clonable.
    Previously it received a slice, which introduces the neccessity of
    cloning the data if it needs to be passed on to a different thread.

In combination this PR can significantly reduce memory usage for
high-concurrency scenarios.

### DIFF
--- a/lib/wasix/src/runtime/mod.rs
+++ b/lib/wasix/src/runtime/mod.rs
@@ -5,7 +5,7 @@ pub mod task_manager;
 
 pub use self::task_manager::{SpawnMemoryType, VirtualTaskManager};
 use self::{module_cache::CacheError, task_manager::InlineWaker};
-use wasmer_types::ModuleHash;
+use module_cache::HashedModuleData;
 
 use std::{
     fmt,
@@ -106,14 +106,7 @@ where
         &self,
         cmd: &BinaryPackageCommand,
     ) -> BoxFuture<'_, Result<Module, SpawnError>> {
-        let engine = self.engine();
-        let hash = *cmd.hash();
-        let wasm = cmd.atom();
-        let module_cache = self.module_cache();
-
-        let task = async move { load_module(&engine, &module_cache, &wasm, hash).await };
-
-        Box::pin(task)
+        self.load_module(HashedModuleData::from_command(cmd))
     }
 
     /// Sync version of [`Self::load_command_module`].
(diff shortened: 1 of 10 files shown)


Arshia001 (approved):

examples/wasi.rs
```diff
@@ -43,7 +43,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     println!("Compiling module...");
     // Let's compile the Wasm module.
-    let module = runtime.load_module_sync(&wasm_bytes[..])?;
+    let data = wasmer_wasix::runtime::module_cache::HashedModuleData::new_sha256(wasm_bytes);
+    let module = runtime.load_module_sync(data)?;
```
syrusakbary: We should not break the API like this.

We can simply create one function `load_module_sync_from_hash` and let `load_module_sync` call `HashedModuleData::new_sha256` and then `load_module_sync_from_hash`

lib/wasix/src/runtime/mod.rs
```diff
@@ -147,21 +145,29 @@ where
...
+        let task = async move { load_module(&engine, &module_cache, &module).await };
 
         Box::pin(task)
     }
 
     /// Load a a Webassembly module, trying to use a pre-compiled version if possible.
     ///
     /// Non-async version of [`Self::load_module`].
-    fn load_module_sync(&self, wasm: &[u8]) -> Result<Module, SpawnError> {
+    fn load_module_sync(&self, wasm: HashedModuleData) -> Result<Module, SpawnError> {
```
syrusakbary: Create new method, don't break this one

theduke: There is no real point in keeping API stability in WASIX, we often make breaking changes when it makes sense.

syrusakbary: There is for the basic usage of external API.
SWC is using this API as well, we are leaking internals of how things are implemented rather than having the API designed by "how things should be used".

Fixing this is a quite easy thing, so I don't see the point of breaking API when we can simply choose to not do it

PR #5539, opened 2025-04-28
feat: add `enable_email` field to app.yaml configuration

ayys:
this allows users to specify if they want an email address associated
with their app. When enabled, the following happens:

- a randomly generated email address is associated with the app
- env vars are set to allow the app to access the email credentials


theduke (commented):
Woudln't this be better as a capability, since we may need a bunch of custom additional settings in the future?

```
capabilities:
  email:
    enabled: true
```

marxin:
@ayys Are you still interested in the PR, or can we close it?

PR #5116, opened 2024-09-25
Fix WASIX additional imports so it works with new threads as well

Arshia001:

PR #5544, opened 2025-04-29
Implement dynamic linking and loading of side modules

Arshia001:
PR is in draft status as of now, there's more left to do. Creating so people can start looking at the code.


lib/wasix/src/state/env.rs
```diff
@@ -327,6 +154,12 @@ pub struct WasiEnv {
     /// Implementation of the WASI runtime.
     pub runtime: Arc<dyn Runtime + Send + Sync + 'static>,
 
+    /// The handle of the current module in the Linker. If the module is
+    /// statically linked, this will always be `MAIN_MODULE_HANDLE` which is
+    /// zero. Non-zero handles will correspond to the module's entry in the
+    /// linker, which will live in the `WasiEnv::inner` field.
+    pub dl_handle: ModuleHandle,
```
theduke, line 161: Why a zero rather than Option<>?

lib/wasix/src/state/env.rs
```diff
@@ -580,61 +428,45 @@ impl WasiEnv {
...
+        let pid = self.process.pid();
 
         let mut store = store.as_store_mut();
 
-        let tasks = env.runtime.task_manager().clone();
-        let mut func_env = WasiFunctionEnv::new(&mut store, env);
-
-        // Determine if shared memory needs to be created and imported
-        let shared_memory = module.imports().memories().next().map(|a| *a.ty());
-
```
theduke, line 609: I'm positive that this was there for a reason, because some WASI (not WASIX) modules , depending on how they are built, sometimes either export or import the memory, so this was here to make sure things always work.

lib/wasix/src/state/env.rs
```diff
@@ -668,17 +561,30 @@ impl WasiEnv {
             func_env
                 .data(&store)
                 .blocking_on_exit(Some(Errno::Noexec.into()));
-            return Err(err.into());
+            return Err(WasiThreadError::ExportError(err));
         }
 
         // If this module exports an _initialize function, run that first.
         if call_initialize {
+            if let Ok(apply_data_relocs) = instance.exports.get_function("__wasm_apply_data_relocs")
```
theduke, line 569: Maybe a comment why this exists and what it's for?

Maybe just a link to https://github.com/WebAssembly/tool-conventions/blob/main/DynamicLinking.md ?

lib/wasix/src/state/func_env.rs
```diff
@@ -73,31 +72,11 @@ impl WasiFunctionEnv {
...
-        ctx.initialize_with_memory(&mut store, instance, memory, update_layout)
-            .map_err(|err| {
-                tracing::warn!("failed initialize environment - {}", err);
-                WasiThreadError::ExportError(err)
-            })?;
+        let (_, ctx) =
+            env.instantiate(module, &mut store, memory, update_layout, call_initialize)?;
 
+        // FIXME: shouldn't this happen _before_ instantiating, so the startup code in the instance
+        // has access to the globals?
```
theduke, line 79: I'm not sure this has been used for anything in a while, it's related to deep sleep, I believe.
When restoring, you would NOT want to run the initializer again.

lib/wasix/src/state/linker.rs
```diff
@@ -0,0 +1,751 @@
...
+
+impl From<u32> for ModuleHandle {
+    fn from(handle: u32) -> Self {
+        ModuleHandle(handle)
+    }
+}
+
+const DEFAULT_RUNTIME_PATH: [&str; 3] = ["/lib", "/usr/lib", "/usr/local/lib"];
+
+pub struct MemoryAllocator {}
```
theduke, line 45: Nit for forwards compatibility: add a `_private: ()` field so it can't be constructed manually.

theduke, line 45: Comment please what this is for.

lib/wasix/src/state/linker.rs
```diff
@@ -0,0 +1,751 @@
...
+    // TODO: It makes more sense to move the entire instantiation flow here, but that requires a bigger
+    // refactor in WasiEnv::instantiate.
+    pub fn new_for_main_module(
+        main_module: &Module,
+        store: &mut impl AsStoreMut,
+        memory: Option<Memory>,
+        imports: &mut Imports,
+        stack_size: u64,
+    ) -> Result<(Self, LinkedMainModule), LinkError> {
+        let dylink_section = super::linker::parse_dylink0_section(&main_module)?;
```
theduke, line 204: I know the linker is probably insecure about it's identity, but that function is still a part of it's own module!

lib/wasix/src/state/linker.rs
```diff
@@ -0,0 +1,751 @@
...
+
+        let loaded_module = DlModule {
+            instance: instance.clone(),
+            memory_base,
+            table_base,
+            instance_handles,
+        };
+
+        let handle = ModuleHandle(guard.next_module_handle);
+        guard.next_module_handle += 1;
```
theduke, line 414: Should we compare the stack offsets here and fail if the module was compiled differently than the main?

lib/wasix/src/state/linker.rs
```diff
@@ -0,0 +1,751 @@
...
+    stack_high: u64,
+    #[allow(dead_code)]
+    stack_low: u64,
+
+    // TODO: cache functions already placed in the table, so we don't add them again
+    indirect_function_table: Table,
+}
+
+#[derive(Clone)]
+pub struct Linker {
```
theduke, line 99: Docs on the Linker, what it does.
Same on all the methods.

lib/wasix/src/state/linker.rs
```diff
@@ -0,0 +1,751 @@
...
+    memory_base: u64,
+    #[allow(dead_code)]
+    table_base: u64,
+    #[allow(dead_code)]
+    instance_handles: WasiModuleInstanceHandles,
+}
+
+struct LinkerState {
+    main_instance: Option<Instance>,
+    side_modules: HashMap<ModuleHandle, DlModule>,
```
theduke, line 81: nit: FnvHashMap ?

lib/wasix/src/state/builder.rs
```diff
@@ -767,50 +765,50 @@ impl WasiEnvBuilder {
...
+    // /// Add an item to the list of importable items provided to the instance.
+    // pub fn import(
+    //     mut self,
+    //     namespace: impl Into<String>,
+    //     name: impl Into<String>,
+    //     value: impl Into<Extern>,
+    // ) -> Self {
+    //     self.add_imports([((namespace, name), value)]);
+    //     self
+    // }
```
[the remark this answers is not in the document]
Arshia001: Well, we kind of do, but custom imports don't work with threads and it was non-trivial getting them to work (as much as they used to) with the refactored code, so I just removed them. _that_ PR is still open though.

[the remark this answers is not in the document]
Arshia001: #5116 to be precise.

lib/wasix/src/state/linker.rs
```diff
@@ -0,0 +1,751 @@
...
+        store: &mut impl AsStoreMut,
+        size: u64,
+        _alignment: u32,
+    ) -> Result<u64, MemoryError> {
+        // TODO: no need to allocate entire pages of memory, but keeping it simple for now...
+        // also, pages are already aligned, so no need to take the alignment into account
+        let mut to_grow = size / WASM_PAGE_SIZE as u64;
+        if size % WASM_PAGE_SIZE as u64 != 0 {
+            to_grow += 1;
+        }
```
theduke, line 64: This requires a certain "cooperation" from the WASM level allocator, to only use new pages that it gets from growing the memory. 

I reckon that should be fine for most allocators though?

lib/wasix/src/state/linker.rs
Arshia001: Dl modules don't have stack offsets, not even the main. We can just configure the stack at will for the main, and sides will use __stack_pointer.

lib/wasix/src/state/linker.rs
Arshia001: Lol, cut-pasted that from an external module 😂

lib/wasix/src/state/func_env.rs
Arshia001: I believe it's also used for forking, setjmp, etc. Definitely something we want to keep. But the order really only matters for dl, and dl+asyncify is not a thing yet.

lib/wasix/src/state/env.rs
Arshia001: Docs are coming...

lib/wasix/src/state/env.rs
Arshia001: The functionality still exists somewhere around TaskWasm. There was 2 of it, now there's just one.

lib/wasix/src/state/env.rs
Arshia001: I need to remove this! Thanks for the reminder 😄

lib/wasix/src/state/linker.rs
Arshia001: We need to add fields to this, the current 'allocate a page per module' approach is terrible. But yeah, makes sense in the meantime.

lib/wasix/src/state/linker.rs
Arshia001: The only way for a guest-side allocator to get more memory is to call `memory.grow`, which is not in any way guarateed to return consecutive addresses. If an allocator does rely on that behavior, it's just faulty.

Copilot agrees:
> No, the memory.grow instruction in WebAssembly does not guarantee consecutive addresses. When memory is grown, the new memory pages are added to the existing memory, but the underlying memory layout is managed by the WebAssembly runtime. If an allocator relies on consecutive addresses being returned, it would be considered faulty.

lib/wasix/src/state/linker.rs
Arshia001: which it doesn't....

lib/wasix/src/runners/wasi.rs
```diff
@@ -238,32 +238,32 @@ impl WasiRunner {
...
+
+    // /// Add multiple import functions.
+    // ///
+    // /// This method will accept a [`&Imports`][wasmer::Imports] object.
+    // pub fn with_imports<I, S1, S2, E>(&mut self, imports: I) -> &mut Self
+    // where
+    //     I: IntoIterator<Item = ((S1, S2), E)>,
+    //     S1: Into<String>,
+    //     S2: Into<String>,
+    //     E: Into<Extern>,
```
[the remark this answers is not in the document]
Arshia001: will do XD

lib/wasix/src/state/builder.rs
```diff
@@ -767,50 +765,50 @@ impl WasiEnvBuilder {
...
-        I: IntoIterator<Item = ((S1, S2), E)>,
-        S1: Into<String>,
-        S2: Into<String>,
-        E: Into<Extern>,
-    {
-        self.add_imports(imports);
-        self
-    }
+    // /// Add an item to the list of importable items provided to the instance.
+    // pub fn import(
```
theduke, line 769: Remove dead code.

lib/wasix/src/state/builder.rs
```diff
@@ -1025,7 +1023,7 @@ impl WasiEnvBuilder {
             #[cfg(feature = "journal")]
             stop_running_after_snapshot: self.stop_running_after_snapshot,
             skip_stdio_during_bootstrap: self.skip_stdio_during_bootstrap,
-            additional_imports: self.additional_imports,
+            // additional_imports: self.additional_imports,
```
theduke, line 1026: Remove

lib/wasix/src/state/builder.rs
```diff
@@ -1077,120 +1075,132 @@ impl WasiEnvBuilder {
...
+                wasmer::ExternType::Memory(ty) => Some(*ty),
+                _ => None,
+            })
+            .map(|ty| wasmer::Memory::new(store, ty))
+            .transpose()
+            .map_err(WasiThreadError::MemoryCreateFailed)?;
+        Ok(env.instantiate(module, store, memory, true, call_init)?)
+    }
+
+    // FIXME: None of these should exist, running WASIX modules should happen through BinFactory
```
theduke, line 1092: Remove dead code

lib/wasix/src/state/linker.rs
```diff
@@ -0,0 +1,1444 @@
...
+    Uninitialized,
+    // Main module linked, now waiting for its instance
+    Pending(InProgressLinkState),
+    // Main module linked and initialized
+    Initialized(Instance),
+}
+
+/// The linker is responsible for loading and linking dynamic modules at runtime,
+/// and managing the shared memory and indirect function table.
+#[derive(Clone)]
```
theduke, line 276: Can we put an explanation for the flow and functionality somewhere, maybe here on the linker or just in the module doc comments?

Arshia001: That's probably a very good idea, if I can figure out the exact flow myself 😄

lib/wasix/src/runners/wasi_common.rs
```diff
@@ -47,7 +46,8 @@ pub(crate) struct CommonWasiOptions {
     pub(crate) stop_running_after_snapshot: bool,
     pub(crate) skip_stdio_during_bootstrap: bool,
     pub(crate) current_dir: Option<PathBuf>,
-    pub(crate) additional_imports: Imports,
+    // TODO: reintroduce in a way that works with threads
+    // pub(crate) additional_imports: Imports,
```
syrusakbary, line 50: Remove me

lib/wasi-types/src/wasi/bindings_manual.rs
```diff
@@ -391,3 +391,21 @@ unsafe impl wasmer::FromToNativeWasmType for JoinFlags {
...
+        self.bits() as i32
+    }
+
+    fn from_native(n: Self::Native) -> Self {
+        Self::from_bits_truncate(n as u32)
+    }
+
+    fn is_from_store(&self, _store: &impl wasmer::AsStoreRef) -> bool {
+        // TODO: find correct implementation
+        false
```
syrusakbary, line 409: We probably want to return true here

Arshia001: That's how it's done for every one of these. I just copy-paste the existing definitions. Not sure if that function is being called at all though, or what the consequences of returning true here would be. Let sleeping dogs lie?

Arshia001:
Superceded by #5581

PR #5564, opened 2025-05-13
build: Make `ureq` optional (build) dependency

kdy1:
# Description

Same as https://github.com/wasmerio/wasmer/pull/5333

 - `ureq` pulls `ring`, which is not cross compilable to aarch64. https://github.com/briansmith/ring/issues/1789

I looked at the code, and it seems like `ureq` is used only if `wamr` or `v8` is enabled. So I marked `ring` as optional build dependency.

CI failure due to `ring`: https://github.com/swc-project/swc/actions/runs/14975161792


syrusakbary (approved):

kdy1:
Can you release a new version?

PR #5569, opened 2025-05-15
Fix a couple of webc FS issues

Arshia001:
* Fix using CopyOnWriteFile with files larger than 8KiB
* Fix error when an OverlayFS::CopyOnWriteFile receives a flush/shutdown request while in read-only state

Note we have two things named CopyOnWriteFile, and this PR references both.

### DIFF
--- a/lib/virtual-fs/src/cow_file.rs
+++ b/lib/virtual-fs/src/cow_file.rs
@@ -1,12 +1,16 @@
 //! Used for /dev/zero - infinitely returns zero
 //! which is useful for commands like `dd if=/dev/zero of=bigfile.img size=1G`
 
+use derive_more::Debug;
 use replace_with::replace_with_or_abort;
-use std::io::{self, *};
 use std::pin::Pin;
 use std::task::{Context, Poll};
+use std::{
+    future::Future,
+    io::{self, *},
+};
 
-use tokio::io::{AsyncRead, AsyncSeek, AsyncWrite, ReadBuf};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncSeek, AsyncSeekExt, AsyncWrite};
 
 use crate::{BufferFile, VirtualFile};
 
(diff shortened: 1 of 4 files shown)


lib/virtual-fs/src/cow_file.rs
```diff
@@ -56,28 +56,37 @@ impl CopyOnWriteFile {
...
+            if *just_started {
+                *just_started = false;
+                if let Err(err) = Pin::new(inner.as_mut()).start_seek(SeekFrom::Start(0)) {
+                    return Poll::Ready(Err(err));
+                }
+            }
+
+            match Pin::new(inner.as_mut()).poll_complete(cx).map_ok(|_| ()) {
+                Poll::Ready(Ok(())) => {}
+                p => return p,
```
theduke, line 73: Regression test for the condition this fixes?

lib/virtual-fs/src/cow_file.rs
```diff
@@ -56,28 +56,37 @@ impl CopyOnWriteFile {
...
+        } = &mut self.state
+        {
+            if *just_started {
+                *just_started = false;
+                if let Err(err) = Pin::new(inner.as_mut()).start_seek(SeekFrom::Start(0)) {
+                    return Poll::Ready(Err(err));
+                }
+            }
+
+            match Pin::new(inner.as_mut()).poll_complete(cx).map_ok(|_| ()) {
```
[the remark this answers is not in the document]
Arshia001: I guess we do...  Not to mention the initial seek was (and still is) basically a hack, we shouldn't be calling `poll_complete` each time.

The cowfile implementation in virtual FS is much more robust, but it has virtual-fs-specific stuff built in, and I didn't want to make big changes, especially since there are probably very few tests.

lib/virtual-fs/src/cow_file.rs
Arshia001: Done

syrusakbary (approved):

PR #5581, opened 2025-05-24
Dynamic linking with support for threads

Arshia001:
Supercedes #5544.

Note, this PR contains a commit that bumps our rust version to 1.84. The 1.81 rust-analyzer was too slow, and we needed to update at some point anyway.


lib/wasix/src/state/env.rs
```diff
@@ -713,14 +570,36 @@ impl WasiEnv {
...
+        Self::do_pending_link_operations(ctx, true)?;
+        _ = Self::process_signals_and_exit(ctx)?;
+        Ok(())
+    }
+
+    pub fn do_pending_link_operations(
+        ctx: &mut FunctionEnvMut<'_, Self>,
+        fast: bool,
+    ) -> Result<(), WasiError> {
+        if let Some(linker) = ctx.data().inner().linker().cloned() {
```
theduke, line 585: Avoid this clone if no work is required.

Arshia001:
Switching to draft since the changes in 97daedd are just a hack.

lib/types/src/module.rs
```diff
@@ -504,6 +504,12 @@ impl ModuleInfo {
         index.index() < self.num_imported_tables
     }
 
+    /// Test whether the given table index is growable. Will return `None` if the
+    /// index is out of bounds.
+    pub fn is_table_growable(&self, index: TableIndex) -> Option<bool> {
+        Some(self.tables.get(index)?.maximum.is_none())
+    }
+
```
syrusakbary, line 512: Move this logic out of the types into an utility function inside of llvm intrinsics, we need to have an API on the module very similar to the Web API

syrusakbary: Also, the function will likely be not used as we will replace the logic for checking if a table is static (fixed size) or dynamic (growable)

lib/wasix/src/state/env.rs
Arshia001: borrowing issue, we pass all of ctx in, so the linker can't be borrowed.

PR #5582, opened 2025-05-26
Add support for weak imports and hidden exports

Arshia001:
Drafting PR for visibility. Should be merged into main after #5581 is in. Make sure to switch the base branch before merging.


Arshia001:
Closing this, since there are more changes that need to happen to the linker. Will push the commits to #5581 .

PR #5583, opened 2025-05-27
feat(wasix): Extend BuiltinPackageLoader with cache scanning and purging

theduke:
Expose the cache of BuiltinPackageLoader, and provide `scan()` and
`retain()` methods to observe the cache contents.

Also does a small change to keep the temporary files in a separate
directory to avoid temp files being included in scanning and retaining
operations.

### DIFF
--- a/lib/wasix/src/runtime/package_loader/builtin_loader.rs
+++ b/lib/wasix/src/runtime/package_loader/builtin_loader.rs
@@ -80,6 +80,10 @@ impl BuiltinPackageLoader {
         }
     }
 
+    pub fn cache(&self) -> Option<&FileSystemCache> {
+        self.cache.as_ref()
+    }
+
     pub fn validate_cache(
         &self,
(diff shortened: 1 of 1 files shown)


lib/wasix/src/runtime/package_loader/builtin_loader.rs
```diff
@@ -419,13 +423,17 @@ pub enum CacheValidationMode {
...
 #[derive(Debug)]
-struct FileSystemCache {
+pub struct FileSystemCache {
     cache_dir: PathBuf,
 }
 
 impl FileSystemCache {
     const FILE_SUFFIX: &'static str = ".bin";
 
+    fn temp_dir(&self) -> PathBuf {
```
zebreus, line 433: I am not familiar with that part of the codebase, but won't using a shared  `__temp__` lead to issues, when multiple wasmer processes are unpacking packages at the same time? (edited

PR #5591, opened 2025-06-05
feat: Include rkyv pointer width in module cache keys

theduke:
The pointer width used by rkyv is configurable through compile-time
features.

If the width is changed, old artifacts currently become invalidated.

This commit exposes related constants for the pointer width , and
includes it in the cache key if it differs from the default.

### DIFF
--- a/lib/types/src/serialize.rs
+++ b/lib/types/src/serialize.rs
@@ -15,6 +15,24 @@ impl MetadataHeader {
     /// to the format of the serialized data.
     pub const CURRENT_VERSION: u32 = 10;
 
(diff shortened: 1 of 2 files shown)


Arshia001 (changes requested):
As discussed on Slack, this does not spark joy. We want wasmer to be able to tell immediately if it can run a given .wasmu, so the information needs to exist within the artifact itself rather than in the name of a cache dir.

syrusakbary:
> the information needs to exist within the artifact itself

100%. Closing this PR, so we can open it again later

PR #5598, opened 2025-06-12
Add syscalls required to support libffi

zebreus:
This PR adds syscalls required to support libffi


lib/api/src/entities/value.rs
```diff
@@ -74,6 +75,20 @@ impl Value {
         Self::ExternRef(None)
     }
 
+    /// Creates a new `Value` with the default value for the given type.
```
Arshia001: Leaving review of changes in lib/api code to @syrusakbary.

lib/wasix/src/state/handles/mod.rs
```diff
@@ -276,4 +276,13 @@ impl WasiModuleTreeHandles {
             Self::Dynamic { linker, .. } => Some(linker),
         }
     }
+
+    pub fn main_module_indirect_function_table(&self) -> Option<Table> {
+        match self {
+            Self::Static(a) => {
+                a.instance.exports.get_table("__indirect_function_table").cloned().ok()
```
Arshia001: Other exports are cached when the handles struct is created, this one should be cached as well.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,80 @@
...
+    results: WasmPtr<u8, M>,
+) -> Result<Errno, WasiRuntimeError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
+    let (env, mut store) = ctx.data_and_store_mut();
+
+    let Some(indirect_function_table) = env.inner().main_module_indirect_function_table() else {
+        // No function table is available, so we cannot call any functions dynamically.
+        // TODO: This should cause a hard crash, but we return an error for now
+        return Ok(Errno::Notsup);
```
Arshia001: How is this error handled on the guest side? I don't suppose the libffi API allows for an error to be returned?

However, I believe returning an error from the wasix syscall and trapping on the guest side when the error can't be communicated to calling code is the correct approach.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,80 @@
...
+    };
+    let Some(function_value) = indirect_function_table.get(&mut store, function_id) else {
+        // TODO: This should cause a trap similar to calling a function with an invalid index in the function table.
+        return Ok(Errno::Inval);
```
Arshia001: Same as above.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,80 @@
...
+    };
+        
+    let Value::FuncRef(Some(function)) = function_value else {
+        // The table does not contain functions, but something else.
+        // TODO: This should cause a crash 
+        return Ok(Errno::Inval);
```
Arshia001: This should be caught at a much earlier stage, and be an internal error at this point. When caching the table's export in the handles struct (see my comment above) we can verify it has the correct type.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,80 @@
...
+        Ok(value)
+    }).collect::<Result<Vec<_>, WasiRuntimeError>>()?;
+
+    let result_values = function.call(&mut store, values_buffer.as_slice())?;
+
+    let memory = unsafe { env.memory_view(&store) };
+    let mut current_results_offset: u64 = results.offset().into();
+    result_values.iter().try_for_each(|result_value| {
+        let Some(bytes) = result_value.as_slice() else {
+            panic!("Function returned an unsupported type");
```
Arshia001: Is this possible at all? If not, unreachable! would be a better alternative.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,315 @@
+use std::path::PathBuf;
+
+use super::*;
+use crate::{state::WasmLoader, syscalls::*};
+use wasmer::{imports, wat2wasm, FunctionType, Table, Type};
+
+// TODO: Move to wasix-types
```
Arshia001: Indeed!

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,315 @@
...
+pub fn closure_free<M: MemorySize>(
+    mut ctx: FunctionEnvMut<'_, WasiEnv>,
+    function_id: u32,
+) -> Result<Errno, WasiRuntimeError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
+    let (env, mut store) = ctx.data_and_store_mut();
+
+    let Some(linker) = env.inner().linker() else {
+        panic!("Closures only work for dynamic modules.");
```
Arshia001: This is a notsup, not a hard crash.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,315 @@
...
+pub fn closure_prepare<M: MemorySize>(
+    mut ctx: FunctionEnvMut<'_, WasiEnv>,
+    // The backing function that will receive the serialized parameters
+    // It needs to have the following signature:
+    // * pointer to the values
+    // * pointer to the results
+    // * pointer to the userdata
+    backing_function_id: u32,
+    // The ID of the function that will be registered
+    function_id: u32,
```
Arshia001: Why the hardcoded function ID?

lib/wasix/src/state/linker.rs
```diff
@@ -877,6 +885,139 @@ impl std::fmt::Debug for Linker {
...
+// TODO: Rename me
+pub enum WasmLoader<'a> {
+    Filesystem {
+        module_name: &'a str,
+        ld_library_path: &'a [&'a Path],
+    },
+    Memory {
+        module_name: &'a str,
+        bytes: &'a [u8],
+        ld_library_path: &'a [&'a Path],
```
Arshia001: Is this actually used?

lib/wasix/src/state/linker.rs
```diff
@@ -1351,16 +1496,109 @@ impl Linker {
         }
     }
 
+    /// TODO: Arshia told me not to do this but I want to implement the other stuff first.
```
Arshia001: Love the honorable mention 😄 indeed, this will break when threads are involved.

lib/wasix/src/state/linker.rs
```diff
@@ -1351,16 +1496,109 @@ impl Linker {
...
+    pub fn allocate_dynamic_function(&self, store: &mut impl AsStoreMut) -> Result<u32, LinkError> {
+        lock_instance_group_state!(
+            group_state_guard,
+            group_state,
+            self,
+            LinkError::InstanceGroupIsDead
+        );
+        // TODO: Implement caching and allocating in chunks to avoid locking for each allocation
+
+        // TODO: Why does this return a u64? We dont implement table64 afaik
```
Arshia001: Because I gave it the wrong type, probably.

lib/wasix/src/state/linker.rs
```diff
@@ -1351,16 +1496,109 @@ impl Linker {
...
+            .clone()
+    }
+
+    /// Populate a previously registered function
+    /// 
+    /// You will get an error, if the index was not previously allocated by `allocate_function`
+    /// 
+    /// It is explicitly allowed to prepare a previously allocated dynamic function again using this.
+    /// However, if the old function is already running while calling this, the behavior of that running function is undefined. 
+    pub fn populate_dynamic_function(
```
Arshia001: The code around 'dynamic' functions (can we get a better name?) is mostly in prototype stage, with known issues, so I won't review it in-depth right now.

lib/wasix/src/state/linker.rs
```diff
@@ -877,6 +885,139 @@ impl std::fmt::Debug for Linker {
     }
 }
 
+/// Contains everything that is required to find a module and load it into memory
+// TODO: Rename me
+pub enum WasmLoader<'a> {
```
Arshia001: This enum can be just data for a module's spec; it does not need to encapsulate logic for loading the specified module as well. That is to say, let's put the code for locating modules back where it was.

Arshia001 (changes requested):
All other comments.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
zebreus: The `closure_*` syscalls work by first allocating a index in the function table and then preparing a closure at that index. The libffi interface requires us to do this in two steps.

lib/wasix/src/state/linker.rs
zebreus: I assume this refers to `ld_library_path`. It is used if the module has shared libs as dependencies in `dylink.0`.

However, we can probably remove it as it should probably get the LD_LIBRARY_PATH from the environment variables when loading shared libraries.

lib/wasix/src/state/linker.rs
zebreus: I renamed the closure-related functions in the linker to `closure_*`

lib/wasix/src/state/linker.rs
zebreus: The logic for actually resolving a path and loading the module bytes doesn't belong to dynamic linking. I think moving it here (and moving the module loader to a separate Rust module in a second step) simplifies the linker a bit.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
zebreus: It should not be possible, I decided to unwrap instead.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
zebreus: Moved

lib/wasix/src/state/linker.rs
zebreus: I am not really sure where to move the function table access, if not into the linker, because the function table isn't an export of a module but is stored in the instance group. I could move a copy outside early on. Do you have any suggestions on where to put it?

I at least tried to mitigate it by acquiring the instance group lock. Can you take another look?

lib/wasix/src/state/linker.rs
zebreus: I'll leave it as is; it doesn't hurt and makes implementing table64 a bit easier when we need it.

lib/wasi-types/src/wasi/bindings.rs
```diff
@@ -2437,6 +2437,35 @@ impl DlFlags {
...
+    I32,
+    #[doc = " A i64 parameter."]
+    I64,
+    #[doc = " A f32 parameter."]
+    F32,
+    #[doc = " A f64 parameter."]
+    F64,
+    #[doc = " A v128 parameter."]
+    V128,
+}
```
syrusakbary, line 2456: We already have the `Type` type in `wasmer_types`

https://github.com/wasmerio/wasmer/blob/main/lib/types/src/types.rs#L24-L41

We should use this rather than recreate another type.

lib/api/src/entities/value.rs
```diff
@@ -263,6 +278,78 @@ impl Value {
...
+    #[cfg(target_endian = "little")]
+    /// Get a slice to the content of this value if it is a scalar type.
+    ///
+    /// This is useful for writing a value to memory.
+    ///
+    /// Returns `None` if the value is not representable as a byte slice.
+    ///
+    /// Not available on big-endian architectures, because the result of this function
+    /// should be compatible with wasm memory, which is little-endian.
+    pub fn as_slice(&self) -> Option<&[u8]> {
```
syrusakbary: I don't think it make sense to provide `as_slice` in the API for a single `Value`, as normally is designed for types that inherently represent a sequence or collection of elements (with the exception of `Option` that can be sliced with a 0-or-1 element slice).

Please move this function out

lib/wasi-types/src/wasi/bindings.rs
zebreus: This one is different in that it is a serializable representation of a `Type` that can be used as a function parameter in wasi functions. The implementation is generated from [wasix-org/wasix-witx](https://github.com/wasix-org/wasix-witx/blob/e44cdd2a372f2d7f2365f72bb2ab76b68611a6d7/phases/snapshot/witx/typenames.witx#L1735-L1751)
to ensure that the serialized representation matches up with what the wasix-libc uses.

lib/api/src/entities/value.rs
zebreus: The API here provides a **byte** slice to the underlying value without copying the data. For a `Value::I32`, this would be a slice with 4 bytes, and for a `Value::I64`, it would be a slice with 8 bytes.

I agree that the name is somewhat misleading. Would it be acceptable to keep this API here, but rename it to `as_bytes`?

I can do this outside of `Value`, but the behaviour is not inherently related to closures, but closely coupled to `Value`. It belongs here.

lib/wasi-types/src/wasi/bindings.rs
zebreus: After considering this for a bit, I decided to leave it as it is. This enum only describes the types that can be serialized (are supported in `call_dynamic`/`closure*`) as opposed to the other one, which represents all types.

lib/api/src/entities/value.rs
zebreus: Moved this change to #5632

lib/wasix/src/state/linker.rs
zebreus: We reviewed this together last week, and the new implementation works with multiple threads by initializing the function table indices in a separate initialization function.

lib/api/src/entities/value.rs
Arshia001: I'm going to agree with @syrusakbary on this one. This function can exist within the syscall code, no need for it to be in the wasmer crate.

lib/wasi-types/src/wasi/bindings.rs
Arshia001: On this one, I agree with @zebreus. This is a recreation of the `Type` enum, but in a different domain, namely the domain of wasix syscalls, so it makes sense to have the duplication.

Cargo.lock
```diff
@@ -7278,6 +7288,7 @@ dependencies = [
  "wasm-bindgen",
  "wasm-bindgen-futures",
  "wasm-bindgen-test",
+ "wasm-encoder 0.235.0",
```
Arshia001, line 7291: Having two versions of a crate isn't good... It looks like we already had wasmparser/wasm-encoder deps, let's reuse those (moving them to the workspace if necessary) and upgrade them to the newest version if needed.

lib/wasix/src/state/linker.rs
Arshia001: On the contrary, this code is specifically 'load a DL module given the specific format the linker expects', which belongs only inside the linker.

Cargo.lock
Arshia001: Just noticed they're already in the workspace. We still need to find whatever's causing the duplicated dep.

lib/api/src/entities/value.rs
```diff
@@ -74,6 +75,20 @@ impl Value {
         Self::ExternRef(None)
     }
 
+    /// Creates a new `Value` with the default value for the given type.
+    pub fn default_typed(ty: &Type) -> Self {
```
Arshia001: I do wonder it it makes sense to move this to the syscall code as well, where we can use the WasmValueType enum directly.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,150 @@
+use super::*;
+use crate::syscalls::*;
+use wasmer::Type;
+
+// TODO: Decide on whether to move this back into Value or not.
```
Arshia001: I don't think this comment is needed. The code looks happy right here.

lib/wasix/src/state/handles/mod.rs
```diff
@@ -276,4 +289,36 @@ impl WasiModuleTreeHandles {
             Self::Dynamic { linker, .. } => Some(linker),
         }
     }
+
+    pub fn main_module_indirect_function_table_lookup(
```
Arshia001: nit: The "main module" vs "static module" naming only makes sense when accessing the overall handles themselves. Since we expect all modules within the same module tree to share the same lookup table all the time (at least for now), this should be called just `indirect_function_table_lookup`.

lib/wasix/src/state/handles/mod.rs
```diff
@@ -276,4 +289,36 @@ impl WasiModuleTreeHandles {
...
+        &self,
+        store: &mut impl AsStoreMut,
+        index: u32,
+    ) -> Result<Function, Errno> {
+        let value = match self {
```
Arshia001: nit: This can just go through `main_module_instance_handles`. There's no need to go through the linker here, and we'll save the extra lock operation that happens in the linker as well.

lib/wasix/src/state/handles/mod.rs
```diff
@@ -276,4 +289,36 @@ impl WasiModuleTreeHandles {
...
+            Self::Dynamic { linker, .. } => linker
+                .lookup_indirect_function_table(store, index)
+                .map_err(|_| Errno::Notsup)?,
+        };
+        let Some(value) = value else {
+            trace!(
+                function_id = index,
+                "Function not found in indirect function table"
+            );
+            return Err(Errno::Inval);
```
Arshia001: I think returning a `Result<Option<Function>>` makes more sense here, as this is an operation that could be useful outside the `call_dynamic` syscall. The `Inval` error code for missing function table entries can be implemented within the syscall itself.

lib/wasix/src/state/handles/mod.rs
```diff
@@ -276,4 +289,36 @@ impl WasiModuleTreeHandles {
...
+        };
+        let Value::FuncRef(funcref) = value else {
+            unreachable!("Function table contains something other than a funcref. At this point this should never happen");
+        };
+        let Some(funcref) = funcref else {
```
Arshia001, line 322: This should also be a `return None`.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,150 @@
...
+        Value::V128(value) => {
+            // Safety: This function is only enabled on little-endian architectures,
+            Some(unsafe { std::slice::from_raw_parts(value as *const u128 as *const u8, 16) })
+        }
+        // ExternRef, FuncRef, and ExceptionRef cannot be represented as byte slices
+        _ => None,
+    }
+}
+
+// TODO: Decide on whether to move this back into Value or not.
```
Arshia001: Same with the comment

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,150 @@
...
+#[cfg(target_endian = "little")]
+/// Get a slice to the content of this value if it is a scalar type.
+///
+/// Returns `None` for Value that can not be freely shared between contexts.
+/// Returns `None` if the value is not representable as a byte slice.
+///
+/// Not available on big-endian architectures, because the result of this function
+/// should be compatible with wasm memory, which is little-endian.
+fn value_as_bytes(value: &Value) -> Option<&[u8]> {
```
Arshia001: We don't need to do pointer casting at all here. These two functions can simply be `read_value` and `write_value`, and access the memory directly, going through the correct rust numeric type's `to_le_bytes` and `from_le_bytes` to do their thing. This eliminates the unsafe code and removes the `#[cfg]`.

Arshia001: ... which has the added benefit that we won't need a "default" for `Value`s either, and can eliminate the `default_typed` function as well, which also doesn't generally make a lot of sense.

lib/api/src/entities/value.rs
Arshia001: ... or eliminate it completely. See comment on `value_as_bytes`.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,150 @@
...
+// TODO: Decide on whether to move this back into Value or not.
+#[cfg(target_endian = "little")]
+/// Get a mutable slice to the content of this value if it is a scalar type.
+///
+/// Returns `None` for Value that can not be freely shared between contexts.
+/// Returns `None` if the value is not representable as a byte slice.
+///
+/// Not available on big-endian architectures, because the result of this function
+/// should be compatible with wasm memory, which is little-endian.
+fn value_as_bytes_mut(value: &mut Value) -> Option<&mut [u8]> {
```
Arshia001: Same thing as `value_as_bytes`, this doesn't need to be so difficult.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,150 @@
...
+            Some(unsafe { std::slice::from_raw_parts_mut(value as *mut u128 as *mut u8, 16) })
+        }
+        // ExternRef, FuncRef, and ExceptionRef cannot be represented as byte slices
+        _ => None,
+    }
+}
+
+/// Call a function from the `__indirect_function_table` with parameters and results from memory.
+///
+/// This function can be used to call functions whose types are not known at compile time of the caller. It is the callers responsibility to ensure that the passed parameters and results match the signature of the function beeing called.
```
Arshia001: Let's break these comments into multiple lines for better readability in IDEs. And may I point out how absolutely smart it was for markdown to render a single line break into a space. Whoever came up with that has my eternal respect.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// ### Format of the values and results buffer
+///
+/// The buffers contain all values sequentially. i32, and f32 are 4 bytes, i64 and f64 are 8 bytes, v128 is 16 bytes.
+///     
+/// For example if the function takes an i32 and an i64, the values buffer will be 12 bytes long, with the first 4 bytes being the i32 and the next 8 bytes being the i64.
```
Arshia001: I assume this doesn't create any headaches with alignments? I don't think the contents of the array are ever used as e.g. `uint64_t *`?

lib/api/src/entities/value.rs
zebreus: I have already removed this from this PR. Please let's continue discussing it in the other one.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,150 @@
...
+///   
+///   The buffer will be interpreted as described above
+///
+///   If the function does not have any parameters, this can be a nullptr (0).
+///
+/// * results: Pointer to a sequence of values
+///   
+///   If the function does not return a value, this can be a nullptr (0).
+///
+///   The buffer needs to be large enough to hold all return values.
```
Arshia001, line 91: This is inherently unsafe; If you happen to get your function ID wrong, and allocate too little space, the syscall code will start happily overwriting memory and causing corruption. Same with the inputs array, where it'll happily read past the end of the array. In the absence of reflection, it is impossible to use this API safely. It's basically a `strcpy` vs `strncpy` situation.

What I think we should do is to add parameters for the length of both arrays, thus making the syscall itself safe. This will probably be useless for `libffi`, but that's not the only use-case for this syscall. `libffi` can just pass an arbitrary large length to "re-enable" its unsafeness on top of our safe syscall.

This has the added benefit that we can feed functions zero values instead of randomly reading from memory, thus implicitly recreating the behaviour of JS trampolines  for wasm function where, if you provide too few arguments, the rest are just filled with zeroes.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+#[instrument(level = "trace", skip_all, fields(%function_id, values_ptr = values.offset().into(), results_ptr = results.offset().into()), ret)]
```
Arshia001: Readability matters XD

```suggestion
#[instrument(
    level = "trace", 
    skip_all, 
    fields(%function_id, values_ptr = values.offset().into(), results_ptr = results.offset().into()),
    ret
)]
```

lib/wasix/src/syscalls/wasix/call_dynamic.rs
```diff
@@ -0,0 +1,150 @@
...
+pub fn call_dynamic<M: MemorySize>(
+    mut ctx: FunctionEnvMut<'_, WasiEnv>,
+    function_id: u32,
+    values: WasmPtr<u8, M>,
+    results: WasmPtr<u8, M>,
+) -> Result<Errno, WasiRuntimeError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
```
Arshia001: This is already slow enough as it is, let's not put this call here. There are already enough checks for pending ops.

lib/wasix/src/state/linker.rs
Arshia001: I'm quite certain this is unused as it stands anyway. If we want this to be useful, any syscalls that load module data from memory where the module data can have more required modules (we don't have any right now) need to receive the value of the env var from guest code (env vars are stored guest-side only).

This field needs to go.

lib/wasix/src/state/linker.rs
Arshia001: In fact, after giving it more thought, I'm of the idea that this entire enum only makes sense as an input to `load_module` and `load_module_tree`. The actual module location logic is:
* Based around the idea of *finding* modules on a *file system*
* Is very highly linker-specific, in that it implements logic according to the expectations of `dlopen`
* Is therefore not suitable for any other kind of "locating modules" that we may want to do; this includes logic for locating executables, which has its own separate flow.

So the thing to do here is:
* Move the actual module locating code back where it was
* Separate the data this enum stores from that logic
* Make sure only `load_module` and `load_module_tree` use this enum, and that it's a data-only enum

The final enum should be this:

```rust
enum DlModuleSpec {
  FsLookup(path, ld_library_path),
  InMemory(name, bytes)
}
```

and, importantly, we should take care to prefix the module names of in-memory modules with something that keeps them separate from modules user code may want to look up; something like `::in-memory::{module_name}`.

lib/wasix/src/state/linker.rs
Arshia001: ... and `load_module_tree` needs to raise an error when encountering in-memory modules with further required modules, since that's not a supported scenario and, indeed, an impossible scenario ATM. Note that POSIX does not provide a way to load a lib from memory, only from disk.

lib/wasix/src/state/linker.rs
```diff
@@ -798,12 +812,18 @@ struct LinkerState {
     // iterate over it, we'll get the modules from lowest handle to highest, and
     // order is preserved.
     side_modules: BTreeMap<ModuleHandle, DlModule>,
-    side_modules_by_name: HashMap<PathBuf, ModuleHandle>,
+    side_modules_by_name: HashMap<String, ModuleHandle>,
```
Arshia001: This change breaks non-UTF8 FS paths. We can always go the other way and store a string as a pathbuf.

lib/wasix/src/state/linker.rs
```diff
@@ -798,12 +812,18 @@ struct LinkerState {
...
     next_module_handle: u32,
 
     memory_allocator: MemoryAllocator,
     heap_base: u64,
 
+    /// Tracks which slots in the function table are currently used for closures
+    used_closure_functions: BTreeSet<u32>,
+    /// Slots in the indirect function table that were allocated for closures but are currently not in use.
+    /// These can be given out without needing to lock all threads.
+    available_closure_functions: BTreeSet<u32>,
```
Arshia001: Since we don't look up or remove specific elements from this, it can be a min-heap or a simple ring buffer, which would do fewer allocations.

lib/wasix/src/state/linker.rs
```diff
@@ -877,6 +897,139 @@ impl std::fmt::Debug for Linker {
...
+        let (module_name, library_path) = match self {
+            WasmLoader::Filesystem {
+                module_name,
+                ld_library_path,
+            } => (module_name, ld_library_path),
+            WasmLoader::Memory {
+                module_name, bytes, ..
+            } => {
+                // TODO: Dont clone here
+                return Ok((PathBuf::from(module_name), bytes.to_vec()));
```
Arshia001: Notice how, in the case of in-memory modules, this function does nothing and just returns its input. That's a tell-tale sign of logic that doesn't belong in a function.

lib/wasix/src/state/linker.rs
```diff
@@ -1013,8 +1169,10 @@ impl Linker {
             main_module_dylink_info: dylink_section,
             side_modules: BTreeMap::new(),
             side_modules_by_name: HashMap::new(),
-            next_module_handle: 1,
+            next_module_handle: MAIN_MODULE_HANDLE.0 + 1,
```
Arshia001, line 1064: gusta mucho.

lib/wasix/src/state/linker.rs
```diff
@@ -1067,11 +1225,14 @@ impl Linker {
             trace!(name = needed, "Loading module needed by main");
             let wasi_env = func_env.data(store);
             linker_state.load_module_tree(
-                needed,
+                WasmLoader::Filesystem {
+                    module_name: needed.as_str(),
+                    // TODO: Use real ld_library_path
```
Arshia001: I'd implement this straight away, it's a rather important thing, and super-easy to do; just need to look for `LD_LIBRARY_PATH` in the env vars list and pass it into this function.

```suggestion
                    // TODO: Pass value of LD_LIBRARY_PATH into this function if it exists
                    // within the process' initial env vars (such as from `--env`
```

lib/wasix/src/state/linker.rs
```diff
@@ -1354,16 +1518,109 @@ impl Linker {
         }
     }
 
+    /// Perform a lookup in the indirect function table
```
Arshia001: As stated elsewhere, this function doesn't need to exist in the linker; the indirect function table is already available from the instance handles.

lib/wasix/src/state/linker.rs
```diff
@@ -1354,16 +1518,109 @@ impl Linker {
...
+                .unwrap();
+            linker_state.used_closure_functions.insert(function_index);
+            return Ok(function_index);
+        }
+
+        // Allocate more closures than we need to reduce the number of sync operations
+        const CLOSURE_ALLOCATION_SIZE: u32 = 100;
+
+        let function_index = group_state
+            .allocate_function_table(&mut store, CLOSURE_ALLOCATION_SIZE, 1)
```
Arshia001: The real alignment value is `2 ** table_alignment`, so the correct value here would be 0, not 1.

lib/wasix/src/state/linker.rs
```diff
@@ -1354,16 +1518,109 @@ impl Linker {
...
+            DlOperation::AllocateFunctionTable {
+                index: function_index,
+                size: CLOSURE_ALLOCATION_SIZE,
+            },
+            linker_state,
+            group_state,
+            &ctx.data().process,
+            ctx.data().tid(),
+        );
+        // TODO: Cleanup the previous function when we are overriding something
```
Arshia001: Left-over comment? I don't know what this means.

lib/wasix/src/state/linker.rs
```diff
@@ -1492,6 +1748,13 @@ impl Linker {
             call_initialization_function::<()>(instance, store, "__wasm_call_ctors")?;
         }
 
+        trace!("Calling wasix on load hook");
```
Arshia001: We really, *really* should be able to just piggyback off of `__wasm_call_ctors` for our closure modules here. I don't see why that name wouldn't work and this one would.

lib/wasix/src/state/linker.rs
```diff
@@ -2223,22 +2498,29 @@ impl InstanceGroupState {
         trace!(
             ?index,
             ?func,
-            "Placing function into table at pre-defined index"
+            "Appending function in table at pre-defined index"
```
Arshia001: "appending" means to add something to the end of a list. This is, indeed, a "placing" (i.e. "inserting") operation.

lib/wasix/src/state/linker.rs
```diff
@@ -2223,22 +2498,29 @@ impl InstanceGroupState {
...
+            );
+            table.grow(store, delta, Value::FuncRef(None))?;
         } else {
             let existing = table.get(store, index).unwrap();
             if let Value::FuncRef(Some(_)) = existing {
                 panic!("Internal error: function table index {index} already occupied");
             }
         }
 
+        let signature = func.ty(store).to_string();
```
Arshia001: nit: we call this a function "type", or "ty" when we want an identifier-friendly version.

lib/wasix/src/state/linker.rs
```diff
@@ -2479,7 +2758,10 @@ impl InstanceGroupState {
         }
 
         for tls in &pending_resolutions.tls {
-            let tls_base = self.tls_base(tls.resolved_from);
+            let Some(tls_base) = self.tls_base(tls.resolved_from) else {
+                panic!("Internal error: Tried to import TLS symbol from module {} that has no TLS base", tls.resolved_from.0);
```
Arshia001: This is not an internal error though, is it? A bad module (that's not necessarily generated by clang) could have TLS symbols but no TLS base.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,447 @@
...
+    let memory = unsafe { env.memory_view(&store) };
+    closure.write(&memory, function_id);
+    return Ok(Errno::Success);
+}
+
+/// Free a previously allocated slot for a closure in the `__indirect_function_table`
+///
+/// After calling this it is undefined behavior to call the function at the given index.
+#[instrument(level = "trace", fields(%closure), ret)]
+pub fn closure_free<M: MemorySize>(
```
Arshia001: We always place each syscall in its own source file.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,447 @@
...
+    mut ctx: FunctionEnvMut<'_, WasiEnv>,
+    closure: u32,
+) -> Result<Errno, WasiError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
+    let (env, mut store) = ctx.data_and_store_mut();
+
+    let Some(linker) = env.inner().linker().cloned() else {
+        trace!("Closures only work for dynamic modules.");
```
Arshia001: This can easily be a warning or an error.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,447 @@
...
+        return Ok(Errno::Notsup);
+    };
+
+    let free_result = linker.free_closure_index(&mut ctx, closure);
+    if let Err(e) = free_result {
+        // Should never happen
```
Arshia001: Yes it can, when a severe linker error happens... although that's a separate concern and this panic is, indeed, correct.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,447 @@
...
+    return Ok(Errno::Success);
+}
+
+/// Allocate a new slot in the __indirect_function_table for a closure
+///
+/// Until the slot is prepared with [`closure_prepare`], it is undefined behavior to call the function at the given index.
+///
+/// The slot should be freed with [`closure_free`] when it is no longer needed.
+#[instrument(level = "trace", skip_all, ret)]
+pub fn closure_allocate<M: MemorySize>(
```
Arshia001: We always place each syscall in its own source file.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,447 @@
...
+    mut ctx: FunctionEnvMut<'_, WasiEnv>,
+    closure: WasmPtr<u32, M>,
+) -> Result<Errno, WasiError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
+    let (env, mut store) = ctx.data_and_store_mut();
+    let Some(linker) = env.inner().linker().cloned() else {
+        trace!("Closures only work for dynamic modules.");
```
Arshia001: warning or error

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,447 @@
...
+        return Ok(Errno::Notsup);
+    };
+
+    let function_id = match linker.allocate_closure_index(&mut ctx) {
+        Ok(f) => f,
+        Err(e) => {
+            // Should never happen
+            panic!("Failed to allocate closure index: {e}");
```
Arshia001: This can _definitely_ happen when for example the table fails to grow due to an OOM or something.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,447 @@
...
+    result_types_length: u32,
+    environment: WasmPtr<u8, M>,
+) -> Result<Errno, WasiError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
+    let (env, mut store) = ctx.data_and_store_mut();
+    let memory = unsafe { env.memory_view(&store) };
+
+    let Some(linker) = env.inner().linker().cloned() else {
+        trace!("Closures only work for dynamic modules.");
```
Arshia001: warning/error

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,447 @@
...
+    let ld_library_path: [&Path; 0] = [];
+    let wasm_loader = WasmLoader::Memory {
+        module_name: &module_name,
+        bytes: &wasm_bytes,
+        ld_library_path: ld_library_path.as_slice(),
+    };
+    let module_handle = match linker.load_module(wasm_loader, &mut ctx) {
+        Ok(m) => m,
+        Err(e) => {
+            // Should never happen
```
Arshia001, line 382: this one can happen due to so many reasons! :D

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,447 @@
...
+    closure: u32,
+    backing_function: u32,
+    environment_offset: u64,
+    argument_types: &[ValType],
+    result_types: &[ValType],
+) -> Vec<u8> {
+    let mut wasm_module = wasm_encoder::Module::new();
+
+    // Add dylink section
+    let dylink = CustomSection {
```
Arshia001, line 128: I know wasmparser already supports the dylink section. Doesn't wasm-encoder also support it?

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,447 @@
...
+    wasm_module.section(&types);
+
+    // Add a imports section
+    let mut imports = ImportSection::new();
+    imports.import(
+        "env",
+        "memory",
+        MemoryType {
+            minimum: 1,
+            maximum: Some(65536),
```
Arshia001: A `None` would work just as well here I think.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
```diff
@@ -0,0 +1,447 @@
...
+                memory_index,
+            }),
+            // Not supported in closures
+            Self::Ref(_) => panic!("Cannot load reference type"),
+        };
+    }
+}
+
+/// Build a dynamically linkable WASM module for the given closure.
+fn build_closure_wasm_bytes(
```
Arshia001, line 117: Meticulously crafted! Well done.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
zebreus: It sadly does not

lib/wasix/src/syscalls/wasix/closure_prepare.rs
zebreus: Fair, but we can be sure our module is valid.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
zebreus: I am not 100% sure if it's happy here. I mean the functions are prefixed with `value_` which is usually a indication that they want to go to something called `Value`.

But let's keep them here for now.

Cargo.lock
zebreus: Can you open an issue for that?

lib/wasix/src/syscalls/wasix/call_dynamic.rs
zebreus: I removed the comment. If I need that functionality in another location I can still add it back later.

lib/wasix/src/state/handles/mod.rs
zebreus: The table in `main_module_instance_handles` is only defined, if the main module owns/exports the table. Which it does not for dynamic modules.

lib/wasix/src/state/handles/mod.rs
zebreus: Sounds good, changed it

lib/wasix/src/syscalls/wasix/call_dynamic.rs
zebreus: Theoretically `do_pending_operations` could be a quite cheap, if there are no pending operations.

zebreus: Our current implementation is not however, so I will comment this here and add a comment

lib/wasix/src/syscalls/wasix/call_dynamic.rs
zebreus: Agree, thought rustfmt would do this one for me

lib/wasix/src/syscalls/wasix/call_dynamic.rs
zebreus: I thought about this a lot while writing the code. It is definitely not like a strcpy situation because the reads always have a clear ending. The guest just does not always know about them ^^.

However I agree that reading outside of the memory is undefined behaviour. I think the proper interface for this function would be to have a properly aligned buffers and to pass the size that the guest allocated for the buffers together with the buffers, so the host can make sure that the sizes match up. We could then add another flag like `strict` (or `lenient`), that toggles between erroring on mismatched calls and filling missing values with zeroes. The only real hurdle here is that a missing return value can't really be filled, as we don't have memory for that.

As for proper alignment, the easiest solution would be to require 8 bytes per value even if its only a 4 byte value. I don't think this matters here, as the values are never used directly, but only ever copied out of memory and used from there.

I think we can add 2 interfaces for this functions. One that is more performant for the case when you are sure about the signature of the function (`call_dynamic_unchecked`) and one that does the checks on described above (`call_dynamic`).

For a first stage, I would basically rename the function to `call_dynamic_unchecked` and implement `call_dynamic` in a second PR.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
zebreus: Yeah, I agree. In the closures situation this could cause alignment issues on the client, but here it is fine.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
zebreus: I have my IDE configured to automatically wrap long lines, so that's no issue for me. However I see the benefit of wrapping long comments.

I changed it so they are wrapped now, but we should enable the rustfmt option for that and process all of our long comments in a separate PR.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
zebreus: I will look into that. However, I am pretty sure that the current solution is the cleanest one that does not require additional copies.

lib/wasix/src/state/linker.rs
zebreus: I think I phrased my previous a bit unclear. This clearly only belongs to the **dynamic linker**, however it does not belong to the **dynamic linking** part of the dynamic linker. Moving it to a separate module referred to moving it ot a module under linker, like `linker/file_loader.rs`. We can move all the library locating/filesystem stuff code in there and focus on the actual dynamic linking in the linker. The current link module is cluttered and way too big.

The in-memory modules also need the dynamic-lbirary path, because they can load shared libs.

I really don't like `DlModuleSpec` (2 abbreviations, `spec` is not descriptive), how about `DynamicLibraryFetcher` that makes it clear that the job of this unit of code is to fetch dynamic libraries. `Loader` could be interpreted that this does some linking logic, which it does not. I think `Fetcher` is more clear about that. I would also agree to `DynamicLibrarySpec` but don't think that's as descriptive.

lib/wasix/src/state/linker.rs
zebreus: I think so, don't remember what this relates to

lib/wasix/src/state/linker.rs
zebreus: Thats a unexpected API, I changed it to 0

lib/wasix/src/state/linker.rs
zebreus: No it's not. See my other comment elsewhere.

lib/wasix/src/state/linker.rs
zebreus: @Arshia001 can you implement this feature on top of the `dynamic_calls` branch? I'll have enough work doing the other things.

lib/wasix/src/state/linker.rs
zebreus: I think I disagree here. The value here lies in providing it's content in a unified form, so the user of `DynamicLibraryFetcher` does not need to know how a dynamic library is fetched, because that's an implementation detail.

lib/wasix/src/state/linker.rs
zebreus: Could be, I need to address this after taking a break.

lib/wasix/src/state/linker.rs
zebreus: I think actually a stack/`Vec` is the ideal data structure here. I changed it.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
zebreus: I think it does, but let's discuss this in the other one.

lib/wasix/src/syscalls/wasix/closure_prepare.rs
zebreus: We also don't implement table64, so this could maybe even happen without OOM. I changed this to an error, but not sure if it is the correct one

lib/wasix/src/state/handles/mod.rs
Arshia001: oh, right, dl modules import the table... then it should be fed into the instance handles by the linker.

lib/wasix/src/state/linker.rs
Arshia001: >way too big

yes, very much so, but it's also doing something quite complex. *Most* of it is just essential complexity.

>(2 abbreviations, spec is not descriptive

"DL" is a very clean abbreviation here, since the entire code is about loading dynamically linked stuff. `spec` is ubiquitous as a shorthand for specification.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
Arshia001: We're dealing with 4- or 8-byte numbers. A pointer on X64 is 8 bytes. Copying the numbers is not more expensive than passing around (copying) a reference.

lib/wasix/src/syscalls/wasix/call_dynamic.rs
Arshia001: I think alignment issues can only occur if the guest decides to just reinterpret the buffer's data as a pointer, which would be quite bad indeed. For the case of copying things into/out of the buffer, alignment shouldn't be a problem, but we can always go back to it if needed. I do believe the alignment is a completely guest-side problem, though.

lib/wasix/src/state/linker.rs
Arshia001: It's the standard dylink.0 uses, I assume to reduce space requirements, since non-power-of-two alignments aren't even a thing.

lib/wasix/src/state/linker.rs
Arshia001: nvm, it is

lib/wasix/src/syscalls/wasix/closure_prepare.rs
Arshia001: god damn.

PR #5603, opened 2025-06-12
Update most shared dependencies to be workspace defined

### LINKED ISSUE #5597: `wasmer-package` 0.600.1 depends on an exact version of `url`
### Describe the bug

According to `crates.io` (and cargo), `wasmer-package` depends `url` version `=2.5.2` instead of `^2.5.2`: https://crates.io/crates/wasmer-package/0.600.1/dependencies

However, this doesn't seem intentional, as that's not reflected in the `Cargo.toml` in the source code for `wasmer-package`: https://github.com/wasmerio/wasmer/blob/v6.0.1/lib/package/Cargo.toml#L29

Is there some publishing script that modifies these version ranges?

This creates issues because cargo attempts to prevent multiple patch versions of a package from being installed:

```
cargo check
    Updating crates.io index
error: failed to select a version for `url`.
    ... required by package `wasmer-package v0.600.1`
    ... which satisfies dependency `wasmer-package = "^0.600.1"` of package `virtual-fs v0.600.1`
    ... which satisfies dependency `virtual-fs = "^0.600.1"` of package `swc_plugin_runner v13.0.0 (/home/bgw.linux/swc/crates/swc_plugin_runner)`
    ... which satisfies path dependency `swc_plugin_runner` (locked to 13.0.0) of package `swc v25.0.0 (/home/bgw.linux/swc/crates/swc)`
    ... which satisfies path dependency `swc` (locked to 25.0.0) of package `swc_estree_compat v15.0.0 (/home/bgw.linux/swc/crates/swc_estree_compat)`
versions that meet the requirements `=2.5.2` are: 2.5.2

all possible versions conflict with previously selected packages.

  previously selected package `url v2.5.4`
    ... which satisfies dependency `url = "^2.5.4"` of package `dbg-swc v21.0.0 (/home/bgw.linux/swc/crates/dbg-swc)`

failed to select a version for `url` which could resolve this conflict
```

### Steps to reproduce

Create a project using both `wasmer-package` and `url` where `url` is newer than `2.5.2`.

### Expected behavior

The published package uses `url` version `^2.5.2` instead of `=2.5.2`.

### Actual behavior

The published package uses `url` version `=2.5.2`.

syrusakbary:
This PR fixes #5597 and future issues caused by version mismatch by making most deps be workspace-defined


lib/cli/Cargo.toml
```diff
@@ -160,49 +160,49 @@ time01 = { package = "time", version = "0.1.45", optional = true }
...
+bytesize.workspace = true
+cfg-if.workspace = true
+tempfile.workspace = true
+serde.workspace = true
+serde_json.workspace = true
 dirs = "4.0"
-serde_json = { version = "1.0" }
-target-lexicon = { version = "0.12", features = ["std"] }
+target-lexicon = { workspace = true, features = ["std"] }
 indexmap = { workspace = true }
```
zebreus:
```suggestion
indexmap.workspace = true
```

Cargo.toml
```diff
@@ -114,17 +141,26 @@ rkyv = { version = "0.8.8", features = ["indexmap-2", "bytes-1"] }
 memmap2 = { version = "0.6.2" }
 toml = { version = "0.8", features = ["preserve_order"] }
 indexmap = "2"
+serde = { version = "1", features = ["derive"], default-features = false }
```
zebreus: Disabling default features on serde, disables impls for Vec and HashMap. I dont think we are using them, but we should probably keep the `std` feature enabled

Enabling dervie by default is fine, because we were doing that everywhere anyway

lib/compiler-llvm/Cargo.toml
```diff
@@ -20,15 +20,15 @@ wasmer-compiler = { path = "../compiler", version = "=6.0.1", features = [
...
 libc.workspace = true
-byteorder = "1"
-itertools = "0.10"
-rayon = "1.5"
+byteorder.workspace = true
+itertools.workspace = true
+rayon.workspace = true
 phf = { version = "0.11.2", features = ["macros"] }
-tracing = { version = "0.1", default-features = false, features = ["log"] }
+tracing = { workspace = true, features = ["log"] }
```
zebreus, line 31: Enabling the default features here should be fine.

We should be able to remove the log feature, as we don't use the log crate.

zebreus, line 31:
```suggestion
tracing.workspace = true
```

zebreus (changes requested):
Builds on my machine. I left some minor nitpicks

lib/compiler-llvm/Cargo.toml
syrusakbary: Your suggestion is not compatible. Note the `log` feature

syrusakbary: We use the `log` crate in the CLI

Cargo.toml
```diff
@@ -114,17 +141,26 @@ rkyv = { version = "0.8.8", features = ["indexmap-2", "bytes-1"] }
...
 serde_yaml = { package = "serde_yml", version = "0.0.12" }
+serde_json = { version = "1" }
+bytecheck = { version = "0.6.8" }
 libc = { version = "^0.2", default-features = false }
 gimli = { version = "0.28.1" }
 futures-util = { version = "0.3.31" }
 mio = "1"
 # MIO 1.0 starts at tokio version 1.39, hence the minimum requirement.
 tokio = { version = "1.39.0", default-features = false }
+tokio-tungstenite = { version = "0.24.0" }
```
theduke, line 154: Initially thought we should use no-default here, but tungstenite deps are actually lean by default!

theduke (approved):
Only comment is that I usually like to split workspace deps and individual deps into separate groups in the Cargo.toml, aka into separate paragraphs.

But that's a style choice.

lib/compiler-llvm/Cargo.toml
zebreus: We register a tracing subscriber, so according to the docs, the `log` feature should have no effect anyway. But I am not sure what our subscribers do, maybe they are accessing log data.

We probably don't want the `log` feature of tracing enabled only when we add the llvm compiler as a dependency. So this is probably not the right place to set it, and it should be removed, or moved to the workspace, or the cli Cargo.toml

zebreus: See my comment below

syrusakbary: I don't disagree, but this PR is keeping the current behavior of the codebase.
If we want to change the behavior that should be done in a PR apart

PR #5607, opened 2025-06-15
Make error messages for missing shared libraries more helpful

zebreus:
The error messages for missing dynamic libraries do not provide any details about why a dynamic library could not be found. This PR makes them more helpful by including the searched paths and the reasons why the library was not found there.

The previous implementation would always show `entry not found` even when the actual filesystem error was something else.

Previous error message:
```
error: Spawn failed
╰─▶ 1: Linker error: File system error: entry not found
```

New error message:
```
error: Spawn failed
╰─▶ 1: Linker error: Failed to find shared library libexample.so: 
           /lib/libexample.so: entry not found
           /usr/lib/libexample.so: entry not found
           /usr/local/lib/libexample.so: entry not found
```

# Description

### DIFF
--- a/lib/wasix/src/state/linker.rs
+++ b/lib/wasix/src/state/linker.rs
@@ -472,7 +472,7 @@ pub enum LinkError {
     TableAllocationError(RuntimeError),
 
     #[error("Failed to find shared library {0}: {1}")]
-    SharedLibraryMissing(String, FsErrors),
+    SharedLibraryMissing(String, LocateModuleError),
 
     #[error("Module is not a dynamic library")]
     NotDynamicLibrary,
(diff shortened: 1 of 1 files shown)


lib/wasix/src/state/linker.rs
```diff
@@ -519,6 +519,25 @@ pub enum LinkError {
     MissingTlsInitializer,
 }
 
+#[derive(Debug)]
+pub enum FsErrors {
+    SingleError(FsError),
+    MultipleErrors(Vec<(PathBuf, FsError)>),
+}
+impl std::fmt::Display for FsErrors {
```
Arshia001: Blank line between enum and impl please

lib/wasix/src/state/linker.rs
```diff
@@ -519,6 +519,25 @@ pub enum LinkError {
     MissingTlsInitializer,
 }
 
+#[derive(Debug)]
+pub enum FsErrors {
+    SingleError(FsError),
```
Arshia001: The naming here can be improved:

* `FsErrors` makes no sense unless you know it comes from the linker. Let's rename this to `ModuleLocationError` (location here meaning "to locate") or something similar that mentions the act of locating modules.
* `Error` in the case names is redundant; let's rename them to `Single` and `Multiple`.

Arshia001 (changes requested):

lib/wasix/src/state/linker.rs
zebreus: I think `LocateModuleError` makes the locate part more clear

Arshia001 (approved):

PR #5608, opened 2025-06-15
Stop requiring __stack_pointer import for dynamic wasi main modules

zebreus:
This PR allows for dynamically linked wasix main modules that don't export a `__stack_pointer`. This is primarily useful for testing modules that don't do anything.

### DIFF
--- a/lib/wasix/src/state/linker.rs
+++ b/lib/wasix/src/state/linker.rs
@@ -961,12 +961,11 @@ impl Linker {
             "Memory layout"
         );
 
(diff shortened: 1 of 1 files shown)


Arshia001:
"Modules that don't do anything" isn't really a valid usecase, and I'd rather fail fast on the ones we actually do care about. We can revisit this once we have an actual usecase that requires it, but that'd require more changes to the code, since without the stack pointer imports, the WASM stack for DL modules can't work.

PR #5611, opened 2025-06-18
Allow loading the main module via dlopen

zebreus:
Calling dl_open with a `NULL` path now returns the handle of the main module. The PR also changes the dlopen handle of the main module to 1, because 0 means `search everywhere`.

# Description


Arshia001:
@zebreus we don't have the 'search all modules after the given handle' functionality; if you give dlopen a non-zero module handle, it'll look in that module only. It's likely whatever required this change will depend on that functionality as well. Otherwise, lgtm.

zebreus:
This is already merged with #5598

PR #5616, opened 2025-07-01
Fix threads breaking when dlsym is called before thread spawn

Arshia001:


lib/wasix/src/state/linker.rs
```diff
@@ -2602,14 +2629,21 @@ impl InstanceGroupState {
...
                                 import.module().to_string(),
                                 import.name().to_string(),
                                 import.ty().clone(),
                             ));
-                        }
+                        };
                         trace!(?module_handle, ?import, "Main memory");
+
+                        // Make sure the memory is big enough for the module being instantiated
+                        let current_size = self.memory.grow(store, 0)?;
```
syrusakbary, line 2642: I'm not sure calling grow to get the current size is the good way, which implies we are missing some important APIs.

Can we have a method to have the current size of a memory `.size(&store)` and use it here? (in the `api/` crate)

Arshia001: I would, indeed, very much like to have a function for getting memory sizes directly. I can implement that in a separate PR (to keep things clean) and we can update the linker afterwards, since this is not the only instance of this call.

lib/wasix/src/state/linker.rs
```diff
@@ -2486,9 +2489,9 @@ impl InstanceGroupState {
     }
 
     fn apply_resolved_function(
-        &mut self,
+        &self,
         store: &mut impl AsStoreMut,
-        name: String,
+        name: &String,
```
zebreus: Should probably be a `&str`

lib/wasix/src/state/linker.rs
```diff
@@ -1288,6 +1288,9 @@ impl Linker {
         trace!("Finalizing pending functions");
         instance_group.finalize_pending_resolutions_from_linker(&pending_resolutions, store)?;
 
+        trace!("Applying externally-requested function table entries");
+        instance_group.apply_requested_symbols_from_linker(store, &linker_state)?;
+
         let linker = Self {
```
zebreus, line 1294: At this point I am thinking it would be less error prone store the `DlOperations` and replay them once a new instance group is created. That way we can be 100% sure that the state is always the same.

Arshia001: there's no need and it doesn't really work without a sizable refactor, since most of the effects of the DL ops are stored directly into the linker's shared state, and get replayed automatically as part of instantiation.

lib/wasix/src/state/linker.rs
Arshia001: probably true, I'll make a mental note to fix it next time we make changes to the linker (or maybe you can fix it in the libffi branch?) since I don't want to wait another hour for CI just for this.

lib/wasix/src/state/linker.rs
zebreus: I think the refactor could avoid similar issues in the future. But let's look into that once that occurs.

lib/wasix/src/state/linker.rs
Arshia001: eh, nvm, had to push another change anyway. Fixed this too.

Issue #5614, opened 2025-06-25
WASI support for WASM reactors without WASIX


imDema:
### Motivation

When using WASM to load plugins or dynamic application components I want to be able to have full control of the execution of the functions in the WASM module.
Without WASI this is easy in wasmer as I can just import and export whatever I want and call them when I need to, however this limits the applicability to languages that can compile to a "WASI-less" target.
In order to support more targets we can use wasip1 to give the functions that (Go, Kotlin, ...) expect through the imported `wasi_snapshot_preview1` interface.

The problem is that, the way that WASI is implemented (and documented) in wasmer essentially forces you into the _command_ module execution flows, with examples showing you creating a WASI env and just running it as if it were basically a standalone process.

The only way I managed to get this to work is to generate the imports from the `generate_import_object_from_env` in `wasmer_wasix` (with the added side effect of **very** long compile time and binary size) then extracting them and registering to another `Imports` object with the interface I require.

Still this is not ideal as the wasmer_wasix implementation is much more than what is needed to support `wasm32-wasip1` targets. Additionally it initializes many resources that are not needed for a simple wasip1 target: I am currently also encountering a `too many files open` OS error (`virtual-mio-0.600.1/src/selector.rs:27:37`) when trying to benchmark my implementation based on wasmer as I am quickly instantiating many Instances and wasmer-wasix is initializing stuff that is not needed as I haven't even enabled FS or networking support for the instance.

This is a non-issue with other wasm engines, as in my implementation using Wasmtime and Wasmi I can just use `wasmtime_wasi::preview1::add_to_linker_sync(&mut linker, |s| &mut s.wasi_ctx)` (or wasmi equivalent) to link the WASI functionality with a lightweight context that doesn't suffer from the same problems as the wasmer implementation.

##### Summary
Consequently as far as I can tell, it is currently impossible to add WASI support without the compilation, size, resource utilization disadvantages of the WASIX implementation, additionally I see no easy way to do it without the workaround I described above. This leaves Wasmer as the most cumbersome runtime to use for dynamic application components that support WASI.

### Proposed solution and Alternatives

-  Split the `wasmer-wasix` implementation allowing to use only the WASI components without the overhead introduced by the rest of the features **OR** Provide a lightweight implementation of the WASI snapshot specifications with a lightweight context that is cheaper to instantiate if resources are not used
[...]

Issue #5612, opened 2025-06-21
[Feature]: support enable a subset of wasi features in wasmer-wasix


hardfist:
Thanks for proposing a new feature!

### Motivation
We're using wasmer in [swc](https://github.com/swc-project/swc) to support run wasm plugin to modify AST, which works great except the binary size is too large.
We only use subset of wasi feature(host-fs) and other features like network is not useful to us, but it seems there's no easy way to disable feature like `reqwest` cause it's all enabled by default in `sys` feature and we have to enable `sys` feature to use wasi.

I try to [disable all features](https://github.com/hardfist/wasmer/commit/26e11ec3207d8c6be26b649559994df540e358f8#diff-abef2cb64f6c893eaa5cfb2b1e77c70a650faf999b05b4a2f0c59dfcaba87eafR206-R210) in `sys` feature by default  and try integrated in [swc](https://github.com/swc-project/swc/pull/10638) which reduce 5M binary size for us which is significant.

### Proposed solution
support another feature like `sys-raw` or `raw` that don't enable all wasi features by default, so the users can enable their needed feature manually

### Alternatives
A clear and concise description of any alternative solutions or features you've considered, and why you're proposed solution is better.

Arshia001:
@hardfist thanks for creating this issue! `wasmer-wasix` is currently designed to operate in one of two modes: `sys` (running on a host machine) or `js` (running in a browser). Allowing a subset of WASIX features to be enabled requires, off the top of my head:

* Stub implementations for all syscalls; you need to provide each WASIX module with the full set of syscalls. Simply leaving some syscalls out will severely limit what modules can be loaded and is not a good option.
* Conditional compilation of parts of the state. There is currently a single `WasiState` struct that holds all the state a module has.
* Some kind of separation of the network and file system code; this is non-trivial because in WASIX (and unix) both sockets and files are described by FDs, hence the existence of `wasmer_wasix::fs::fd::Kind`, which holds files, sockets, and more. This also extends to the epoll logic and basically everything else that touches FDs in some way.
* Careful consideration of the impact of removed syscalls on guest code; for example, WASIX modules retrieve the CWD at startup, which would break if that syscall was configured out of a build.

All in all, this is a non-trivial feature to implement. We're happy to accept a contribution, and to help with the implementation details if needed.

syrusakbary:
We are going to review things a bit further and see what would be path forward @hardfist.

hardfist:
[...]

PR #5621, opened 2025-07-07
feat: support sys-minimal feature

hardfist:
This pull request introduces support for the `sys-minimal` feature in the `wasix` library, provide minimal `sys` implementation to reduce binary size for consumers. 
Related to 
* https://github.com/wasmerio/wasmer/issues/5612
* https://github.com/wasmerio/wasmer/issues/5614
* https://github.com/swc-project/swc/pull/10638#issuecomment-3042328241

@syrusakbary I'm not sure how I should add tests about this, may need your guidance

### DIFF
--- a/lib/wasix/Cargo.toml
+++ b/lib/wasix/Cargo.toml
@@ -203,7 +203,7 @@ webc_runner_rt_dproxy = [
 	"journal",
 ]
 # the minimal wasip1 implementation
-wasi-common = ["wasmer/sys", "sys-thread", "tokio/fs"]
+sys-minimal = ["wasmer/sys", "sys-thread", "tokio/fs"]
 sys = [
 	"webc/mmap",
 	"time",
--- a/lib/wasix/Cargo.toml
+++ b/lib/wasix/Cargo.toml
@@ -202,7 +202,8 @@ webc_runner_rt_dproxy = [
 	"tower-http",
 	"journal",
 ]
-
+# the minimal wasip1 implementation
+wasi-common = ["wasmer/sys", "sys-thread", "tokio/fs"]
 sys = [
 	"webc/mmap",
 	"time",
--- a/lib/wasix/Cargo.toml
+++ b/lib/wasix/Cargo.toml
@@ -202,7 +202,7 @@ webc_runner_rt_dproxy = [
 	"tower-http",
 	"journal",
 ]
-# the minimal wasip1 implementation
+# the minimal sys implementation
 sys-minimal = ["wasmer/sys", "sys-thread", "tokio/fs"]
 sys = [
 	"webc/mmap",
--- a/lib/wasix/src/lib.rs
+++ b/lib/wasix/src/lib.rs
@@ -17,15 +17,18 @@
 #[cfg(all(
     not(feature = "sys"),
     not(feature = "js"),
-    not(feature = "wasi-common")
+    not(feature = "sys-minimal")
 ))]
 compile_error!(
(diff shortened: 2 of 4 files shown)


hardfist:
@syrusakbary  https://github.com/wasmerio/wasmer/actions/runs/16106520659/job/45444368736?pr=5621#step:4:257 It seems these tests fail caused by CI permission?

syrusakbary:
Hey @hardfist,
I see you introduced the `wasi-common` feature, but it doesn't do any code restrictions at all that would decrease the code size.
Ideally, we would see some systemcalls being restricted/not added on the WASI `lib.rs` file, which would let the compiler optimize the not-used code away.

Also, You made the `wasi-common` feature being xor'ed with `js` or `sys` which doesn't make that much sense. `wasi-common` should/could be used with either `js` or `sys`m

Arshia001:
Hey @hardfist, thank you for this contribution!

From what I can tell, the change adds a new feature, which disables some of the dependencies. I have doubts about the exact set of dependencies enabled, though. For example, disabling the host networking feature isn't 100% compatible with wasip1, as with wasip1, you do get networking, just not the ability to open new sockets.

Since the set of features enabled in the `wasi-common` feature is not something that aligns 100% with wasip1, I wonder if it'd be better to just enable just these features in your own crate that depends on `wasmer-wasix`? I understand this is not possible today because we refuse to compile without exactly one of the `sys` or `js` features, so my suggestion is:

* we revert this change
* instead, we add in a new feature, named `sys-minimal`, which will be equivalent to "enable native stuff such as `state::handles::globals`, but nothing else"
* `sys` and `sys-minimal` can both be active at the same time, since they don't clash in any way, but if at least one is enabled, `js` must be disabled. Also, at least one of `js` or (`sys`/`sys-minimal`) must be enabled for the crate to compile.
  * This is, BTW, another issue with the current implementation; you can, for example, have `wasi-common` and `js` active at the same time and not trigger the compile errors
* client crates can then enable `sys-minimal` and cherry pick any other features they need, leaving out unused ones to avoid code bloat.

Would this solve your usecase, @hardfist?

hardfist:
The `wasi-common` semantic I want is actually the same as @Arshia001 's suggestion of `sys-minimal`(I use `wasi-common` just because wasmtime use this naming 🤣), so @Arshia001 your suggestion would be perfect for my case.

> I see you introduced the wasi-common feature, but it doesn't do any code restrictions at all that would decrease the code size.

I think it can reduce the binary size cause it doesn't enforce all features `sys`  and `js` do, so consumer can pick minimal feature

Arshia001:
Cool! Do you want to make the implementation?

hardfist:
> Cool! Do you want to make the implementation?

yeah, I can give it a second try

Arshia001 (approved):

Arshia001:
LGTM now!

hardfist:
@Arshia001 I also notice that some of wasmer's dependency is not necessary for our case like `wat` and wonder whether it could be feature gated too like wasix, if so I can continue improve in following PR

hardfist:
@Arshia001 it seems these failing tests is not related to the change?

Arshia001:
@hardfist I'm quite sure we already have a `wat` feature in wasmer. I remember it having a sizable impact on binary size.

The integration tests are flaky, and hence not marked as required. Merging this in. Thanks again!

hardfist:
@Arshia001 sorry to bother, when will wasmer release next version?

Arshia001:
I don't have an exact time, but 6.1 is coming soon, probably as soon as the current work on dynamic linking is done, which is very close to the finish line.

PR #5629, opened 2025-07-10
Make bundling webcs fail if there are major errors

zebreus:
The current behaviour silently omits fs entries from a webc if there is a problem reading that directory and strictness is set to loose. However, an I/O error when reading a directory is a critical error and should be reported to the user. This PR changes the behaviour to abort and report that error to the user. That behaviour mirrors the behaviour when including a single file.


syrusakbary (changes requested):
Only missing a test

PR #5634, opened 2025-07-14
Improve thread error handling

zebreus:
# Description
This commit improves error handling when spawning a thread fails. Critical error messages are now printed to stderr instead of debug logs. Additionally, the exit codes for these types of errors are now hardcoded to 129 to facilitate easier tracing of errors.

The first commit contains the logic changes. The second commit refactors the code to remove the unused `ret` value and splits a huge function into multiple smaller ones. The third one commits some fixes in other files applied by `clippy --fix`


syrusakbary (commented):
We need a test for the thread behavior

syrusakbary (approved):

PR #5636, opened 2025-07-15
Implement strict mode switch in call_dynamic

Arshia001:


zebreus (changes requested):
Can you please add some trace logs, when there is a mismatch between the passed values and the expected values?

PR #5646, opened 2025-07-18
fix: fix wat gating

hardfist:
# Description
wat2wasm is not used and not feature gating by "wat", so just remove it 
I test it in swc_plugin_runner and it works good now


theduke:
Thanks!

PR #5653, opened 2025-07-21
Implement a reflection API in WASIX

zebreus:
# WASIX Reflection API

Reflections should provide a way to introspect the signature of a function

#### Use cases:

* Avoid a lenient dynamic call for a function where you are not sure about the signature. Imagine there are two valid sigs and you want to assert that the function is one of the two, and then call it appropriately.
* Another use case is to save on dynamic calls by inspecting the signature of every function once and then calling it directly with the correct signature. The same as described above.
  * For these use cases, we need caching of the signatures in wasix-libc.
* A use case that we currently don't have, but would fit nicely here, is to provide a way to validate the signature of a function against a known signature.

#### Performance considerations:

* The full call will be expensive because the guest needs to allocate a buffer beforehand.
* We could optimize here by allowing the guest to pass null pointers
* In case the guest needs to call again.
* We could add a second function that returns the number of parameters and results, allowing one to be called first to determine the size of the buffer to allocate.
* Then, the main function can be called with the allocated buffer and filled with the signature.
* We could also integrate both functions into one, where it always returns the size and only fills if the size is big enough. The guest can then call it with a guessed buffer size, which will work in most cases, but can call again if that one was too small.
* The full call is actually not that expensive. If we allow passing null pointers as buffer,s it should be more or less equivalent to the metadata call.

#### What factors affect the cacheability of the signature?

* The function is allocated as a closure, so the signature can change
* The function is allocated by the dynamic linker or the module itself. These functions will not change at runtime. \* should not change. Guarantees for internal functions?
* The function exceeds the bounds of the table so that it will be updated and populated later.

#### Which edge cases are there, and how do we handle them

* The function is not found
  * `exists` => false
* The buffers are too small
  * `buffers_too_small` => true
* The buffers are too big
  * The remaining values remain unchanged
* The buffers are nullptr
  * ok, if their length is null, otherwise error
* The function is nullptr/0
  * I think slot zero is always empty, so it should be fine as is.
* Multiple tables
  * We can not support this because libc is always in the main module. Once we have the garbage collection proposal, we could implement
* The signature of the function changed
  * The cacheable boolean would be set to false in that case
* The function is out of bounds
  * `exists` => false
* The table is not available/wrong
  * A missing indirect function table is treated like a maximum-sized indirect function table, that is completely empty.
* Memory error
  * error
* The function is a closure; its signature can change.
  * `cachable` => false
* The function is not yet initialized, but it has been allocated for a closure.
  * `cachable` => false
* The function is not yet initialized but will never be (e.g., empty table slots due to alignment)
  * `cachable` => true
* The function is ready and was populated by the dynamic linker or by the module initialization.
  * `cachable` => true

#### Unstructured requirements:

* Provide information about the signature for functions
* Make the call cacheable
* Python does not need to make that many syscalls
* Information about the existence of the function.
* Provide information about the cachability of the signature
* Provide information about the number of arguments and results
* Perhaps provide information about the module from which the function originates?

## Questions

1. 1 syscall or 2 syscalls?

Strong preference for 1 function. But let's not rule out two functions completely. A draft for this is attached as proposal 2.

2. What is the format of the return types?

The same as in the closures/call_dynamic. Reuse the types from there

3. Should we provide optimizations when passing null pointers optimization for the function, so the guest doesn't always need to allocate buffers?

Yes

4. Split it up into type IDs and signatures?

No, we will not do that. More work can be done, and we can implement it in wasix-libc if needed.

5. Should we provide information about the module from which a function is derived?

No, too complicated, and no use case

6. Support for multiple tables

Not feasible, as wasi-libc only has access to the main modules' tables

7. We could also improve performance by making the API work with type IDs and providing an additional function to retrieve type IDs.

Too complicated. The only use case I see for this is when reflection is heavily used with non-cacheable functions.

8. We could provide a separate validation function, or use the existing one and check on the guest side.

Using the existing one comes with only minor drawbacks. However, a signature for what this function could look like is attached as a proposal extra

9. Do we want to provide information about the module the function is from?

No

10. Do we want to provide information about whether a function is a closure

No

11. Should the function communicate errors?

By returning errno

## Proposal

Implement a single function that encompasses all functionality. It will always return the function metadata if you pass buffers that are too small to hold all parameter types on the first call. You can call it again with a correctly sized buffer, because you have the metadata.

## Related PRs

* wasix-org/wasix-witx#7 defines the signature of the function in witx
* wasix-org/wasix-libc#75 exposes the syscall in our libc


lib/wasix/src/state/handles/mod.rs
```diff
@@ -297,11 +297,17 @@ impl WasiModuleTreeHandles {
...
+    /// * Returns an Errno if an error occurred.
+    /// * Returns `Ok(None)` if the index is out of bounds.
+    /// * Returns `Ok(Some(None))` if there is no function at the index.
+    /// * Returns `Ok(Some(Some(function)))` if there is a function at the index.
     pub fn indirect_function_table_lookup(
         &self,
         store: &mut impl AsStoreMut,
         index: u32,
-    ) -> Result<Option<Function>, Errno> {
+    ) -> Result<Option<Option<Function>>, Errno> {
```
Arshia001: `Result<Option<T>>` is already bad enough. Let's turn this into a separate enum.

lib/wasix/src/syscalls/wasix/reflect_signature.rs
```diff
@@ -0,0 +1,143 @@
+use super::*;
+use crate::syscalls::*;
+use wasmer::Type;
+use wasmer_wasix_types::wasi::WasmValueType;
+
+/// A structure representing the reflection information for a function signature
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct FunctionSignatureInfo {
```
Arshia001: This belongs with all the other bindings, in `wasix_types` I assume?

lib/wasix/src/syscalls/wasix/reflect_signature.rs
```diff
@@ -0,0 +1,143 @@
...
+            trace!(
+                "Failed to look up function in indirect function table: {}",
+                e
+            );
+            wasi_try_mem_ok!(signature_info.write(FunctionSignatureInfo {
+                cacheable: 0,
+                arguments: 0,
+                results: 0,
+            }));
+            return Ok(e);
```
Arshia001: Not a fan of passing the error code through. We should have an explicit error code here.

lib/wasix/src/syscalls/wasix/reflect_signature.rs
```diff
@@ -0,0 +1,143 @@
...
+        cacheable,
+        arguments,
+        results,
+    }));
+
+    if (arguments > argument_types_len || results > result_types_len) {
+        return Ok(Errno::Overflow);
+    }
+
+    for (index, param) in function_type.params().iter().enumerate() {
```
Arshia001: no need to loop, `MemoryView` supports writing slices.

lib/wasix/src/state/linker.rs
```diff
@@ -1063,7 +1065,7 @@ impl Linker {
             side_modules_by_name: HashMap::new(),
             next_module_handle: MAIN_MODULE_HANDLE.0 + 1,
             memory_allocator: MemoryAllocator::new(),
-            used_closure_functions: BTreeSet::new(),
+            allocated_closure_functions: BTreeMap::new(),
```
Arshia001, line 1068: Why do we need the information on whether a closure slot is in use?

Arshia001 (changes requested):

lib/wasix/src/state/handles/mod.rs
syrusakbary: Agreed, Enum will make things more readable (note: I also thought about using a custom error type instead)

lib/wasix/src/state/linker.rs
zebreus: A closure was either in `used_closure_functions` or in `available_closure_functions`. To check if a pointer is allocated for a closure, we need to search for it in either. But as `available_closure_functions` is a Vec, we can't efficiently search it.

Previously, `used_closure_functions` only included the pointers handed out to the guest; now it consists of all pointers we reserved for closures, allowing us to easily check if a pointer is a closure. `available_closure_functions` remains unchanged, providing a more efficient data structure for retrieving the next available pointer.

lib/wasix/src/syscalls/wasix/reflect_signature.rs
zebreus: I wanted to avoid writing it into a slice before, but didn't realize at that point how big the loop content would become. Will change it

lib/wasix/src/state/linker.rs
zebreus: Didn't read your comment properly, I'll look into if we actually need that information

lib/wasix/src/syscalls/wasix/reflect_signature.rs
zebreus: As `indirect_function_table_lookup` is returning an `Errno` I think it's proper to return it as is.

lib/wasix/src/syscalls/wasix/reflect_signature.rs
```diff
@@ -0,0 +1,143 @@
...
+    let cacheable = if is_closure { 0 } else { 1 };
+
+    let Some(function) = function else {
+        wasi_try_mem_ok!(signature_info.write(FunctionSignatureInfo {
+            cacheable,
+            arguments: 0,
+            results: 0,
+        }));
+        return Ok(Errno::Inval);
+    };
```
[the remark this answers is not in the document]
zebreus: That's wrong, `function` changed.

lib/wasix/src/syscalls/wasix/reflect_signature.rs
```diff
@@ -0,0 +1,143 @@
...
+    for (index, param) in function_type.params().iter().enumerate() {
+        let Ok(value_type) = WasmValueType::try_from(*param) else {
+            trace!(
+                "Failed to convert parameter type {} to WasmValueType",
+                param
+            );
+            return Ok(Errno::Inval);
+        };
+        wasi_try_mem_ok!(argument_types.write(&memory, value_type));
+        wasi_try_mem_ok!(argument_types.add_offset(M::ONE));
```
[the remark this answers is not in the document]
zebreus: I am pretty sure this worked in tests, but it will be rewritten to slices anyway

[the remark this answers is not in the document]
zebreus: I am pretty sure this worked in tests, but it will be rewritten to slices anyway

lib/wasix/src/syscalls/wasix/reflect_signature.rs
zebreus: Changed it to write my values into a vec before and then copy that as a slice.

lib/wasix/src/state/handles/mod.rs
zebreus: I moved all the error conditions into an enum; the function now returns a `Result<Function, LookupFunctionError>`

lib/wasix/src/syscalls/wasix/reflect_signature.rs
zebreus: `indirect_function_table_lookup` now returns a proper error type that has `Into<Errno>` implemented.

lib/wasix/src/state/linker.rs
zebreus: Yes, we need to track that information. If the guest calls `closure_free` we need to know if the closure is currently in use, or not. Only if it is used, it should get pushed to `available_closure_functions`.

lib/wasi-types/src/wasi/bindings.rs
```diff
@@ -2466,6 +2466,56 @@ impl core::fmt::Debug for WasmValueType {
...
+        }
+    }
+}
+
+#[doc = " A structure representing the reflection information for a function signature"]
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct ReflectionResult {
+    #[doc = " Whether the result is cacheable"]
+    pub cacheable: u8,
```
Arshia001: This field should use the wasix Bool type.

Arshia001 (changes requested):

lib/wasi-types/src/wasi/bindings.rs
zebreus: Yes it should

zebreus: Fixed

Arshia001:
@zebreus LGTM, let's merge once we have everything else set up and have run this code through python.

PR #5655, opened 2025-07-22
Shamelessly fix multiple things in one PR

Arshia001:


lib/wasix/src/state/linker.rs
```diff
@@ -3174,11 +3200,20 @@ impl InstanceGroupState {
...
+                        if !matches!(import.ty(), ExternType::Tag(ty) if *ty.params == [Type::I32])
+                        {
+                            return Err(LinkError::BadImport(
+                                import.module().to_string(),
+                                import.name().to_string(),
+                                import.ty().clone(),
+                            ));
+                        }
+                        trace!(?module_handle, ?import, "C++ exception tag");
+                        imports.define(import.module(), import.name(), self.cpp_exception.clone());
```
zebreus, line 3216: This is the exact same as above. Is there a reason for not having this in a shared function?

scripts/publish.py
```diff
@@ -285,15 +302,6 @@ def publish(self):
...
 
-            # sleep for 16 seconds between crates to ensure the crates.io
-            # index has time to update
-            if not self.dry_run:
-                print(
-                    "Sleeping for 16 seconds to allow the `crates.io` index to update..."
-                )
-                time.sleep(16)
-            else:
-                print("In dry-run: not sleeping for crates.io to update.")
```
zebreus, line 296: Why don't we need that sleep anymore?

zebreus (changes requested):

lib/wasix/src/state/linker.rs
Arshia001: Yes, because we were initially lazy XD Let's do some refactoring on the Linker at some point, I'm sure there's more duplicate code we can eliminate.

scripts/publish.py
Arshia001: `cargo` now does the waiting for us. I assume this was a pain point for other people as well.

lib/wasix/src/state/linker.rs
Arshia001: (note: we = me personally, no intention to pass blame around needlessly)

zebreus:
There is a new failing test in `tests/validation/log.test.ts`

PR #5660, opened 2025-07-29
Support `?Send` function environments

Twey:
This PR is currently a draft, as the work needs to be ported to the other (non-`sys`) backends and also needs documentation in some places, but contains all public API changes so I'm submitting it early for feedback.

# Description

Apropos of https://github.com/wasmerio/wasmer/pull/4634, parameterize the `Store` on a type of type-erased `Object`s.  This enables building stores where the internal `Object` is `Box<dyn Any>` (instead of `Box<dyn Any + Send>`) to permit the storage of non-`Send` objects in the store.

Type defaults have been used to maintain API compatibility for the most part.  However, this is still technically a breaking change, since it changes some public but less-used APIs in incompatible ways (e.g. adding `type Object` to `AsStoreRef`).

A trait `Upcast<T>` is introduced to `wasmer-types` for types that can type-erase `T`, and implemented for `Box<dyn Any>` and `Box<dyn Any + Send>`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1364,6 +1364,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "derive-where"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "510c292c8cf384b1a340b816a9a6cf2599eb8f566a44949024af88418000c50b"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.100",
+]
+
+[[package]]
 name = "derive_arbitrary"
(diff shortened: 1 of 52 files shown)


syrusakbary:
Thanks for creating the PR, we believe is a step in the right direction. We need a bit more time to review and test in other places (like browser)

Twey:
Appreciate it! Currently only the `sys` backend is done (hence the draft) but if you approve of the direction I'll get porting the other backends.

marxin:
Thanks for the PR! However, it’s still a work in progress, and we currently don’t have the bandwidth to polish and complete it.

PR #5669, opened 2025-07-31
Fix WASIX networking w.r.t. aiohttp

Arshia001:
This PR is a collection of fixes required to get aiohttp working properly.

The most important change is to the epoll system. Previously, epoll for anything besides sockets would rely on a custom `Waker` to get notified of changes to the FD's status. This was, however, broken, because:

* The old system would rely on polling the FD for read-readiness, with the custom `Waker` pushing events to the epoll instance once it was woken
  * This is already a misuse of the waker system, but let's leave that aside
* Once the waker was triggered, an event would be pushed to the epoll instance
* The instance would then try to "re-arm" the waker by polling for read-readiness again
* However, since the FD would already be ready to read more data (at this point, user code wouldn't have had a chance to read anything yet), the call would just go through and the waker wouldn't get registered

The new approach uses an `InterestHandler` instead. This is well-tested (sockets already use it), is explicit, and is not hacky.

PR in draft status for now until I figure out if more fixes are needed.

### DIFF
--- a/lib/virtual-fs/src/pipe.rs
+++ b/lib/virtual-fs/src/pipe.rs
@@ -1,16 +1,20 @@
 use bytes::{Buf, Bytes};
-use std::io::{self, Read, Seek, SeekFrom};
 use std::pin::Pin;
 use std::sync::Arc;
 use std::sync::Mutex;
 use std::task::Context;
 use std::task::Poll;
 use std::{io::IoSlice, sync::MutexGuard};
+use std::{
+    io::{self, Read, Seek, SeekFrom},
+    sync::Weak,
+};
 use tokio::sync::mpsc;
 use tokio::{
     io::{AsyncRead, AsyncSeek, AsyncWrite},
     sync::mpsc::error::TryRecvError,
 };
+use virtual_mio::{InterestHandler, InterestType};
 
 use crate::{ArcFile, FsError, VirtualFile};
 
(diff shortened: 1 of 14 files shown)


syrusakbary:
Everything looks good for me so far, but I think @theduke will be probably better to do the review

Arshia001:
Tests are coming! We do have *some* epoll tests, wonder why they weren't catching the issue before.

lib/virtual-fs/src/pipe.rs
```diff
@@ -213,6 +234,15 @@ impl PipeTx {
             Poll::Ready(Ok(8192))
         }
     }
+
+    fn push_readable_interest_to_rx_end(&self) {
```
theduke: Nit: maybe `mark_read_ready` / `mark_readable`?

I know it's only an internal function, but pretty crytpic.

lib/virtual-net/src/host.rs
```diff
@@ -117,8 +117,15 @@ impl VirtualNetworking for LocalNetworking {
 
         #[cfg(not(windows))]
         let socket = {
-            let std_sock =
-                Socket::new(Domain::IPV4, Type::DGRAM, None).map_err(io_err_into_net_error)?;
```
theduke, line 121: uh good fix

theduke (commented):
Good cleanup, LGTM.

Should we have a test that validates the UDP behaviour and the previously stalling pipe behaviour?

lib/virtual-fs/src/pipe.rs
Arshia001: I was trying to convey the fact that we're pushing an interest to the *other* half of the pipe, since the tx end has no interest handler and even if it did, it'd only support write interest... Let me think of something.

PR #5670, opened 2025-07-31
Fix segfault on unterminated eh_frame by validating the frame before passing it to `__register_function`.

zebreus:
Previously we did not add a terminator to the `.eh_frame` section in our llvm compiler. This is fine most of the time, because there's a good chance that the memory afterwards is filled with `0`'s, which constitutes a terminator. However, if the eh_frame section end aligns with a memory page, and our process is not allowed to access the next page, we get a segfault. This buffer over-read may have security implications as it will enable a WASM program to indirectly access the memory of the process, although it's probably impossible to exploit. This only affects libgcc-based systems, as the `__register_frame` API of libunwind interprets its argument as a single record.

This PR fixes that by asserting that the structure of `.eh_frame` is valid.

Things done:
- [x] Implement validation of the eh_frame struct
- [x] Test this PR with cranelift on CI
- [x] Test this PR with singlepass on CI
- [x] Test this PR with llvm on CI
- [x] Test this PR with llvm locally
- [x] Figure out how to handle old artifacts

### DIFF
--- a/lib/compiler-llvm/src/compiler.rs
+++ b/lib/compiler-llvm/src/compiler.rs
@@ -486,9 +486,10 @@ impl Compiler for LLVMCompiler {
 
         if !eh_frame_section_bytes.is_empty() {
             let eh_frame_idx = SectionIndex::from_u32(module_custom_sections.len() as u32);
-            // Do not terminate dwarf info with a zero-length CIE.
-            // Because more info will be added later
-            // in lib/object/src/module.rs emit_compilation
+            // Terminate dwarf info with a zero-length CIE.
+            // This is ok, even though more info will be added later
+            // in lib/object/src/module.rs emit_compilation, but that won't be called on all code paths.
+            eh_frame_section_bytes.extend_from_slice(&[0, 0, 0, 0]);
             module_custom_sections.push(CustomSection {
                 protection: CustomSectionProtection::Read,
                 alignment: None,
--- a/lib/compiler-llvm/src/compiler.rs
+++ b/lib/compiler-llvm/src/compiler.rs
@@ -486,9 +486,10 @@ impl Compiler for LLVMCompiler {
 
         if !eh_frame_section_bytes.is_empty() {
             let eh_frame_idx = SectionIndex::from_u32(module_custom_sections.len() as u32);
-            // Terminate dwarf info with a zero-length CIE.
-            // This is ok, even though more info will be added later
-            // in lib/object/src/module.rs emit_compilation, but that won't be called on all code paths.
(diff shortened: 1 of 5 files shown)


lib/compiler-llvm/src/compiler.rs
```diff
@@ -486,9 +486,10 @@ impl Compiler for LLVMCompiler {
 
         if !eh_frame_section_bytes.is_empty() {
             let eh_frame_idx = SectionIndex::from_u32(module_custom_sections.len() as u32);
-            // Do not terminate dwarf info with a zero-length CIE.
-            // Because more info will be added later
-            // in lib/object/src/module.rs emit_compilation
+            // Terminate dwarf info with a zero-length CIE.
+            // This is ok, even though more info will be added later
```
theduke: Explain why it is okay!

lib/compiler/src/engine/unwind/systemv.rs
```diff
@@ -138,51 +138,71 @@ impl UnwindRegistry {
...
+        // Pointers to the registrations that will be registered with `__register_frame`.
+        // For libgcc based systems, these are CIEs.
+        // For libunwind based systems, these are FDEs.
+        let mut records_to_register = Vec::new();
 
-                    // Move to the next table entry (+4 because the length itself is not inclusive)
-                    current = current.add(len + 4);
+        let mut current = 0;
+        let mut last_len = 0;
+        while current < (eh_frame.len() - 3) {
```
theduke: Comment please for where the `3` comes from, for the future reader

theduke (commented):
Makes sense for me to fix.

We'll have to bump the artifact version indeed, but if we coordinate the bump and are aware of it when it goes downstream we can manage.

lib/compiler/src/engine/unwind/systemv.rs
zebreus: Changed it to `<= len - size_of::<u32>` to make it clear that this is to make sure that it is always safe to cast the current index to an u32

theduke (approved):

syrusakbary:
I have some concern which makes this PR a bit hard to review: we don't have tests to confirm the new behavior. How can we improve this?

zebreus:
@syrusakbary All of our existing tests confirm that this PR does not break anything catastrophically. We can't write an automated test for the bug without a lot of effort, as it depends on the specifics of the hardware it's running on, how it's loaded, and a lot of randomness (memory layout etc). The example I used for debugging only caused crashes on my machine, but not on Arshia's.

I think this PR is logically sound, does not seem to break anything, and matches up with the spec for eh_frames.

PR #5677, opened 2025-08-04
Dont use the task manager for module loading

zebreus:
For now this is just to test if this breaks CI

# Description

### DIFF
--- a/lib/wasix/src/runtime/module_cache/filesystem.rs
+++ b/lib/wasix/src/runtime/module_cache/filesystem.rs
@@ -43,51 +43,35 @@ impl ModuleCache for FileSystemCache {
     #[tracing::instrument(level = "debug", skip_all, fields(% key))]
     async fn load(&self, key: ModuleHash, engine: &Engine) -> Result<Module, CacheError> {
         let path = self.path(key, &engine.deterministic_id());
+        let engine = engine.clone();
 
-        self.task_manager
-            .runtime_handle()
-            .spawn({
-                let task_manager = self.task_manager.clone();
-                let engine = engine.clone();
+        let bytes = read_file(&path).await?;
 
-                async move {
-                    let bytes = read_file(&path).await?;
+        match deserialize(&bytes, &engine) {
+            Ok(m) => {
+                tracing::debug!("Cache hit!");
+                Ok(m)
+            }
+            Err(e) => {
+                tracing::debug!(
+                    %key,
+                    path=%path.display(),
+                    error=&e as &dyn std::error::Error,
+                    "Deleting the cache file because the artifact couldn't be deserialized",
+                );
 
-                    task_manager
-                    .spawn_await({
-
-                        move || match deserialize(&bytes, &engine) {
-                            Ok(m) => {
-                                tracing::debug!("Cache hit!");
-                                Ok(m)
-                            }
-                            Err(e) => {
-                                tracing::debug!(
-                                    %key,
-                                    path=%path.display(),
-                                    error=&e as &dyn std::error::Error,
-                                    "Deleting the cache file because the artifact couldn't be deserialized",
-                                );
-
-                                if let Err(e) = std::fs::remove_file(&path) {
-                                    tracing::warn!(
-                                        %key,
-                                        path=%path.display(),
-                                        error=&e as &dyn std::error::Error,
-                                        "Unable to remove the corrupted cache file",
-                                    );
-                                }
-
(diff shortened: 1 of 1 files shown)


lib/wasix/src/runtime/module_cache/filesystem.rs
```diff
@@ -43,51 +43,35 @@ impl ModuleCache for FileSystemCache {
...
-            .runtime_handle()
-            .spawn({
-                let task_manager = self.task_manager.clone();
-                let engine = engine.clone();
-
-                async move {
-                    let bytes = read_file(&path).await?;
-
-                    task_manager
-                    .spawn_await({
```
theduke, line 57: This nested spawn is important!

Module::deserialize is potentially expensive, and is not allowed to run on the async worker, it must be offloaded to a blocking thread.

Only becomes visible as a problem with high concurrency, or on Edge, which previously used this module cache before getting a more full-featured implementation.

lib/wasix/src/runtime/module_cache/filesystem.rs
```diff
@@ -43,51 +43,35 @@ impl ModuleCache for FileSystemCache {
     #[tracing::instrument(level = "debug", skip_all, fields(% key))]
     async fn load(&self, key: ModuleHash, engine: &Engine) -> Result<Module, CacheError> {
         let path = self.path(key, &engine.deterministic_id());
-
-        self.task_manager
```
theduke, line 47: I believe this initial spawn is there because we wanted to make suire that the task runs on the runtime used by the instance, rather than another runtime, which can happen in more complex setups , eg edge where there are different runtimes for the server and for instances, leading to potential stalling issues with futures not advancing.

Not relevant anymore at the moment because Edge does things differently now, so it can probably be removed.

lib/wasix/src/runtime/module_cache/filesystem.rs
zebreus: Can we move the nested spawn to a `spawn_blocking` on the main runtime? Your second comment makes it seem like that would be possible as that feature is not used by edge anymore.

lib/wasix/src/runtime/module_cache/filesystem.rs
zebreus: I have removed it. I assume the cases where this is useful are pretty limited, because the main runtime is still waiting for the other runtime. (Although there may still be cases where you want the file loading or the long blocking task to happen in a specific thread pool or something.)

For now, I removed it. If someone needs it, they should file an issue, and we can investigate how to add it back then.

lib/wasix/src/runners/wasi.rs
```diff
@@ -503,6 +496,12 @@ impl WasiRunner {
         let tasks = runtime.task_manager().clone();
         let pkg = pkg.clone();
 
+        // TODO: This will NOT work when loading shared libraries, because of the FileSystemCache
+        //
```
zebreus:
```suggestion
```

Remove outdated comment

lib/wasix/src/runtime/module_cache/filesystem.rs
```diff
@@ -38,66 +37,120 @@ impl FileSystemCache {
...
+        tracing::warn!(
+            dir=%parent.display(),
+            error=&e as &dyn std::error::Error,
+            "Unable to create the cache directory",
+        );
+    }
+
+    // Note: We save to a temporary file and persist() it at the end so
+    // concurrent readers won't see a partially written module.
+    let (file, temp) = NamedTempFile::new_in(parent)
```
theduke, line 105: This is a blocking operation, should be in a blocking function.

You could just leave a TODO.

lib/wasix/src/runtime/module_cache/filesystem.rs
```diff
@@ -38,66 +37,120 @@ impl FileSystemCache {
...
+
+    if let Err(e) = tokio::fs::create_dir_all(parent).await {
+        tracing::warn!(
+            dir=%parent.display(),
+            error=&e as &dyn std::error::Error,
+            "Unable to create the cache directory",
+        );
+    }
+
+    // Note: We save to a temporary file and persist() it at the end so
```
zebreus, line 103:
```suggestion
    // TODO: NamedTempFile is blocking and we should use the appropriate tokio function instead.
    // Note: We save to a temporary file and persist() it at the end so
```

theduke (approved):

PR #5696, opened 2025-08-18
Fix initialization of main DL modules

Arshia001:
Fixes a regression introduced in #5679


zebreus:
I haven't yet reviewed this in depth, but it seems strange to me that we are calling two initialization functions here. The previous PR replaced one invocation with another one. Why is this one adding two more?

Let's, at some point, write some documentation about which functions are supposed to be called in which order and under which circumstances.

Arshia001:
>The previous PR replaced one invocation with another one.

The previous PR added a new invocation, keeping the old one around. The same two functions need to be called for the main instance as well, but I failed to notice that the main instance doesn't exist in the link state, so the functions weren't being called for it, hence the issues we were seeing with everything up to and including `printf`.

zebreus (approved):
lgtm

PR #5697, opened 2025-08-19
Add WASIX DL tests

Arshia001:


tests/wasix/test.sh
```diff
@@ -1,72 +1,5 @@
...
--Wl,--export=__data_end \
--Wl,--export=__wasm_init_tls \
--Wl,--export=__wasm_signal \
--Wl,--export=__tls_size \
--Wl,--export=__tls_align \
--Wl,--export=__tls_base \
--lwasi-emulated-mman \
--O3 \
--g \
--flto"
```
zebreus, line 68: nit: Maybe keep these, but point them at wasixcc

zebreus (approved):
Looks good. I think these tests should suffice to inform us when dl is broken

tests/wasix/test.sh
Arshia001: These are all flags that are automatically supplied by wasixcc, so no need to keep them around anymore.

PR #5705, opened 2025-08-25
Refactor module handles to use a safe new type

zebreus:
This PR refactors our `ModuleHandle` Type to split it into two types: one that is guaranteed to hold a non-special handles and another that can also hold special handles (For now just `Invalid` and `RTLD_DEFAULT`)

### DIFF
--- a/lib/wasix/src/state/linker.rs
+++ b/lib/wasix/src/state/linker.rs
@@ -283,35 +283,18 @@ use wasmer::{
 };
 use wasmer_wasix_types::wasix::WasiMemoryLayout;
 
+use super::{WasiModuleInstanceHandles, WasiState};
 use crate::{
     fs::WasiFsRoot, import_object_for_all_wasi_versions, Runtime, SpawnError, WasiEnv, WasiError,
     WasiFs, WasiFunctionEnv, WasiModuleTreeHandles, WasiProcess, WasiThreadId,
 };
+pub(crate) use module_handle::{ModuleHandle, ModuleHandleWithFlags};
 
-use super::{WasiModuleInstanceHandles, WasiState};
-
-// Module handle 1 is always the main module. Side modules get handles starting from the next one after the main module.
-pub static MAIN_MODULE_HANDLE: ModuleHandle = ModuleHandle(1);
-static INVALID_MODULE_HANDLE: ModuleHandle = ModuleHandle(u32::MAX);
+mod module_handle;
 
 static MAIN_MODULE_MEMORY_BASE: u64 = 0;
 // Need to keep the zeroth index null to catch null function pointers at runtime
 static MAIN_MODULE_TABLE_BASE: u64 = 1;
-
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct ModuleHandle(u32);
-
-impl From<ModuleHandle> for u32 {
-    fn from(handle: ModuleHandle) -> Self {
-        handle.0
-    }
-}
-
-impl From<u32> for ModuleHandle {
-    fn from(handle: u32) -> Self {
-        ModuleHandle(handle)
-    }
-}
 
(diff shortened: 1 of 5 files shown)


lib/wasix/src/state/linker/module_handle.rs
```diff
@@ -0,0 +1,133 @@
+use derive_more::Debug;
+
+// IDs for special module handles
+// Not public to ensure they are only used through the safe API.
+// TODO: These should probably move to the C bindings at some point
```
Arshia001, line 5: Note: these are internal to the linker and not visible to the C bindings at all.

lib/wasix/src/state/linker/module_handle.rs
```diff
@@ -0,0 +1,133 @@
...
+/// Constant representing the main module ID (1)
+const RAW_MODULE_HANDLE_MAIN: u32 = 1;
+/// Constant representing an invalid module handle (u32::MAX)
+const RAW_MODULE_HANDLE_INVALID: u32 = u32::MAX;
+
+/// A handle of a dynamic loaded shared object returned by dlopen
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ModuleHandleWithFlags {
+    /// An invalid module handle
+    Invalid,
```
Arshia001, line 18: This case doesn't really make sense except for a very special case (that of recursively loading module trees), and there's no point in polluting the code with it.

lib/wasix/src/state/linker/module_handle.rs
```diff
@@ -0,0 +1,133 @@
...
+    }
+}
+impl TryFrom<u32> for ModuleHandle {
+    type Error = ();
+
+    /// Try to convert from u32 to ModuleHandle
+    ///
+    /// Will fail if the u32 represents a special handle.
+    fn try_from(handle: u32) -> Result<Self, Self::Error> {
+        ModuleHandleWithFlags::from(handle).try_into()
```
Arshia001, line 98: Why go through the other type here?

lib/wasix/src/syscalls/wasix/dl_invalid_handle.rs
```diff
@@ -1,8 +1,9 @@
-use state::ModuleHandle;
+use state::ModuleHandleWithFlags;
 
 use super::*;
-use crate::syscalls::*;
+use crate::{state::ModuleHandle, syscalls::*};
 
+// TODO: dl invalid handle
```
Arshia001, line 6: weird TODO comment?

lib/wasix/src/state/linker.rs
```diff
@@ -540,6 +526,7 @@ impl std::fmt::Display for LocateModuleError {
     }
 }
 
+// TODO: What does partially resolved mean?
```
Arshia001, line 529: Another weird TODO comment... we answer questions in comments, not ask them of the reader XD

lib/wasix/src/state/linker.rs
```diff
@@ -1554,7 +1542,7 @@ impl Linker {
         &self,
         module_spec: DlModuleSpec,
         ctx: &mut FunctionEnvMut<'_, WasiEnv>,
-    ) -> Result<ModuleHandle, LinkError> {
+    ) -> Result<ModuleHandleWithFlags, LinkError> {
```
Arshia001, line 1545: This is the wrong type to return... what does returning `RTLD_DEFAULT` from `dlopen` mean?

Arshia001 (changes requested):
While I agree with modelling domains as closely as possible so as to [disallow bad state](https://fsharpforfunandprofit.com/posts/designing-with-types-making-illegal-states-unrepresentable/), I believe there are problems with how this PR does it. I'll comment the specific problems, and we can fix this once we start the effort to refactor the linker. I will close this PR for now so we don't spend more time on it.

PR #5711, opened 2025-08-29
SinglePass: add support for riscv64 target

marxin:
The PR adds the rudimentary support for RISC-V 64-bit target into the SinglePass compiler. Even though, the code still needs some polishing and various TODOs must be addressed, the compiler can pass all (104) the `spec` tests.

The biggest known limitations:
- unwinding support
- FP status register - save/restore
- CLZ, CTZ and POPCOUNT have sub-optimal performance
- a general `location_cmp` + `jmp_on_X` needs polishing as RISC-V target does not provide a FLAGS register
- `spec/address.wast` is flaky (sometimes trap is not properly triggered)
- still some unimplemented entry points for the `Machine` trait
- more intensive testing needed

### DIFF
--- a/lib/compiler-singlepass/src/codegen.rs
+++ b/lib/compiler-singlepass/src/codegen.rs
@@ -4090,12 +4090,15 @@ impl<'a, M: Machine> FuncGen<'a, M> {
                 let table_label = self.machine.get_label();
                 let mut table: Vec<Label> = vec![];
                 let default_br = self.machine.get_label();
-                self.machine.emit_relaxed_cmp(
-                    Size::S32,
-                    Location::Imm32(targets.len() as u32),
+                // TODO: refactor
+                let tmp = self.machine.acquire_temp_gpr().unwrap();
+                self.machine.i32_cmp_lt_u(
                     cond,
+                    Location::Imm32(targets.len() as u32),
+                    Location::GPR(tmp),
                 )?;
-                self.machine.jmp_on_aboveequal(default_br)?;
+                self.machine.jmp_on_false(Location::GPR(tmp), default_br)?;
+                self.machine.release_gpr(tmp);
 
                 self.machine.emit_jmp_to_jumptable(table_label, cond)?;
 
--- a/lib/compiler-singlepass/src/codegen.rs
+++ b/lib/compiler-singlepass/src/codegen.rs
@@ -6698,7 +6698,7 @@ impl<'a, M: Machine> FuncGen<'a, M> {
         let mut body = self.machine.assembler_finalize()?;
         body.shrink_to_fit();
 
-        save_assembly_to_file(std::path::Path::new("/tmp/module-dump.o"), &body);
+        save_assembly_to_file("-module-dump.o", &body);
 
         Ok((
             CompiledFunction {
--- a/lib/compiler-singlepass/src/codegen.rs
+++ b/lib/compiler-singlepass/src/codegen.rs
@@ -2863,7 +2863,7 @@ impl<'a, M: Machine> FuncGen<'a, M> {
 
                 // Trap if signature mismatches.
                 // TODO: refactor
-                self.machine.i64_cmp_ne(
+                self.machine.i32_cmp_eq(
                     Location::GPR(sigidx),
                     Location::Memory(
                         table_count,
@@ -2871,7 +2871,6 @@ impl<'a, M: Machine> FuncGen<'a, M> {
                     ),
                     Location::GPR(cond),
                 )?;
-                // TODO
                 self.machine
                     .jmp_on_false(Location::GPR(cond), self.special_labels.bad_signature)?;
 
--- a/lib/compiler-singlepass/src/codegen.rs
+++ b/lib/compiler-singlepass/src/codegen.rs
@@ -2818,10 +2818,14 @@ impl<'a, M: Machine> FuncGen<'a, M> {
                     )?;
                 }
 
+                let cond = self.machine.acquire_temp_gpr().unwrap();
+                self.machine.i32_cmp_lt_u(
+                    func_index,
+                    Location::GPR(table_count),
+                    Location::GPR(cond),
+                )?;
                 self.machine
-                    .location_cmp(Size::S32, func_index, Location::GPR(table_count))?;
-                self.machine
-                    .jmp_on_belowequal(self.special_labels.table_access_oob)?;
+                    .jmp_on_false(Location::GPR(cond), self.special_labels.table_access_oob)?;
                 self.machine
                     .move_location(Size::S32, func_index, Location::GPR(table_count))?;
                 self.machine.emit_imul_imm32(
@@ -2843,13 +2847,15 @@ impl<'a, M: Machine> FuncGen<'a, M> {
                     Location::GPR(table_count),
                 )?;
                 // Trap if the FuncRef is null
-                self.machine.location_cmp(
-                    Size::S64,
-                    Location::Imm32(0),
+                // TODO: refactor
+                self.machine.i64_cmp_ne(
                     Location::GPR(table_count),
+                    Location::Imm64(0),
+                    Location::GPR(cond),
                 )?;
                 self.machine
-                    .jmp_on_equal(self.special_labels.indirect_call_null)?;
+                    .jmp_on_false(Location::GPR(cond), self.special_labels.indirect_call_null)?;
+
                 self.machine.move_location(
                     Size::S64,
                     Location::Memory(
@@ -2860,17 +2866,20 @@ impl<'a, M: Machine> FuncGen<'a, M> {
                 )?;
 
                 // Trap if signature mismatches.
-                self.machine.location_cmp(
-                    Size::S32,
+                // TODO: refactor
+                self.machine.i64_cmp_ne(
                     Location::GPR(sigidx),
                     Location::Memory(
                         table_count,
                         (self.vmoffsets.vmcaller_checked_anyfunc_type_index() as usize) as i32,
                     ),
+                    Location::GPR(cond),
                 )?;
+                // TODO
                 self.machine
-                    .jmp_on_different(self.special_labels.bad_signature)?;
+                    .jmp_on_false(Location::GPR(cond), self.special_labels.bad_signature)?;
 
+                self.machine.release_gpr(cond);
                 self.machine.release_gpr(sigidx);
                 self.machine.release_gpr(table_count);
                 self.machine.release_gpr(table_base);
--- a/lib/compiler-singlepass/src/codegen.rs
+++ b/lib/compiler-singlepass/src/codegen.rs
@@ -2847,14 +2847,10 @@ impl<'a, M: Machine> FuncGen<'a, M> {
                     Location::GPR(table_count),
                 )?;
                 // Trap if the FuncRef is null
-                // TODO: refactor
-                self.machine.i64_cmp_ne(
+                self.machine.jmp_on_false(
                     Location::GPR(table_count),
-                    Location::Imm64(0),
-                    Location::GPR(cond),
+                    self.special_labels.indirect_call_null,
                 )?;
-                self.machine
-                    .jmp_on_false(Location::GPR(cond), self.special_labels.indirect_call_null)?;
 
                 self.machine.move_location(
                     Size::S64,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7059,6 +7059,7 @@ dependencies = [
  "rayon",
  "smallvec",
  "target-lexicon 0.12.16",
+ "tempfile",
  "wasmer-compiler",
  "wasmer-types",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -304,6 +304,11 @@ path = "examples/riscv.rs"
 required-features = ["singlepass"]
 
 [[example]]
+name = "riscv2"
+path = "examples/riscv2.rs"
+required-features = ["singlepass"]
+
+[[example]]
 name = "engine-headless"
 path = "examples/engine_headless.rs"
 required-features = ["cranelift"]
--- a/examples/riscv.rs
+++ b/examples/riscv.rs
@@ -17,7 +17,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     let wasm_bytes = wat2wasm(
         r#"
 (module
-  (func (export "nearest") (param $x f64) (result f64) (f64.nearest (local.get $x)))
+  (func (export "nearest") (param $x f64) (result f64) (f64.floor (local.get $x)))
 )
     "#
         .as_bytes(),
@@ -39,7 +39,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("Calling `fn` function...");
     //let result = sample.call(&mut store, &[Value::I32(123456)])?;
     //let result = sample.call(&mut store, &[])?;
-    let result = func.call(&mut store, &[Value::F64(0.49999)])?;
+    let result = func.call(&mut store, &[Value::F64(-0.0)])?;
     let ret = result[0].unwrap_f64();
     println!("Result 0x{} {}", ret, ret);
 
--- a/examples/riscv.rs
+++ b/examples/riscv.rs
@@ -16,19 +16,9 @@ use wasmer_compiler_singlepass::Singlepass;
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     let wasm_bytes = wat2wasm(
         r#"
-  (module
-    (type $sum_t (func (param i32 i32 i32) (result i32)))
-    (func $sum_f (type $sum_t)
-        (param $p1 i32)
-        (param $p2 i32)
-        (param $p3 i32)
-        (result i32)
-    local.get $p1
-    local.get $p2
-    i32.add
-    local.get $p3
-    i32.add)
-    (export "sum" (func $sum_f)))
+(module
+  (func (export "f32.convert_i32_s") (param $x i32) (result f32) (f32.convert_i32_s (local.get $x)))
+)
     "#
         .as_bytes(),
     )?;
@@ -44,14 +34,14 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     println!("Instantiating module...");
     let instance = Instance::new(&mut store, &module, &import_object)?;
-    let func = instance.exports.get_function("sum")?;
+    let func = instance.exports.get_function("f32.convert_i32_s")?;
 
     println!("Calling `fn` function...");
     //let result = sample.call(&mut store, &[Value::I32(123456)])?;
     //let result = sample.call(&mut store, &[])?;
-    let result = func.call(&mut store, &[Value::I32(1), Value::I32(2), Value::I32(3)])?;
-    let ret = result[0].unwrap_i32();
-    println!("Result 0x{:x} {}", ret, ret);
+    let result = func.call(&mut store, &[Value::I32(-1)])?;
+    let ret = result[0].unwrap_f32();
+    println!("Result 0x{} {}", ret, ret);
 
     // for i in 0..16 {
     //     let result = sample2.call(&mut store, &[Value::I32(i)])?; //, &[Value::I32(1), Value::I32(123)])?;
@@ -59,7 +49,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     // }
 
     // let result = sample2.call(&mut store, &[Value::I32(14)])?; //, &[Value::I32(1), Value::I32(123)])?;
-    // println!("Result {:?}", result);
+    println!("Result {:?}", result);
 
     Ok(())
 }
--- a/examples/riscv.rs
+++ b/examples/riscv.rs
@@ -10,81 +10,24 @@
 //!
 //! Ready?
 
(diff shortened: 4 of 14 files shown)


wakabat:
Nice work! Huge kudos on making this happen!

We've been testing this branch and here's some early discoveries:

* `target-lexicon` used by wasmer don't have a default calling convention, so when we are doing cross-compilation, the code will always fail at [those lines](https://github.com/marxin/wasmer/blob/4702cebd38142155202bc19c5d4adb6852fc25a4/lib/compiler-singlepass/src/compiler.rs#L96-L98), by any means you feel like setting `SystemV` as the default calling convention for RISC-V?
* We do notice the emitter does not cover all cases of the code structure, for example, we hit codegen errors in SLL / SRL generations(in some cases `src1` is actually an immediate). We have summarized our changes in this [commit](https://github.com/wakabat/official-wasmer/commit/1a6f3fd66c0a8d34b640817cbf975d222ed1dbec), which also includes the above default calling convention change. It works for us for the moment, but it might not be an exhausive list of codegen fixes. If you like those changes, feel free to incorporate them into this PR.
* We do notice some unaligned memory loads(i.e., `LD` tries to load from an address which is not 8-byte aligned), on our platform we have strict alignment rules, so just want to check: is it a goal for wasmer to generate aligned loads / stores?

Thanks for the awesome work!

marxin:
> Nice work! Huge kudos on making this happen!

Thank you!

> target-lexicon used by wasmer don't have a default calling convention, so when we are doing cross-compilation, the code will always fail at [those lines](https://github.com/marxin/wasmer/blob/4702cebd38142155202bc19c5d4adb6852fc25a4/lib/compiler-singlepass/src/compiler.rs#L96-L98), by any means you feel like setting SystemV as the default calling convention for RISC-V?

Sounds good to me right now.

> We do notice the emitter does not cover all cases of the code structure, for example, we hit codegen errors in SLL / SRL generations(in some cases src1 is actually an immediate). We have summarized our changes in this [commit](https://github.com/wakabat/official-wasmer/commit/1a6f3fd66c0a8d34b640817cbf975d222ed1dbec), which also includes the above default calling convention change. It works for us for the moment, but it might not be an exhausive list of codegen fixes. If you like those changes, feel free to incorporate them into this PR.

Again, thank you. Right now, I am going to cover more of the wasmer's test-cases on RISC-V target and I am going to cover all the missing pieces of the codegen.

> We do notice some unaligned memory loads(i.e., LD tries to load from an address which is not 8-byte aligned), on our platform we have strict alignment rules, so just want to check: is it a goal for wasmer to generate aligned loads / stores?
Thanks for the awesome work!

Interesting! To be honest, based on the search on the Internet, the architecture itself does not mandate if unaligned memory operations are allowed or not. However, as you described, there are definitely strict alignment boards in the wild. What do you use for testing? My VisionFive 2 board works fine when it comes to unaligned memory operations.

wakabat:
> Interesting! To be honest, based on the search on the Internet, the architecture itself does not mandate if unaligned memory operations are allowed or not. However, as you described, there are definitely strict alignment boards in the wild. What do you use for testing? My VisionFive 2 board works fine when it comes to unaligned memory operations.

Yes RISC-V spec do not require alignments of memory loads / stores, but I'm willing to bet there will be hardwares (maybe not VisionFive 2) that rely on alignments to simplify the implementation. We actually use a software RISC-V VM in which alignments help us significantly simplify the implementation. In addition, the LLVM compiler in wasmer fully respect alignments in the generated code.

Give those thought, can we at least make it an option / flag, which would generate aligned loads / stores when flipped on?

wakabat:
One more thing: there is a fix to dynasm which we will need: https://github.com/CensoredUsername/dynasm-rs/pull/118. The fix is released in dynasm 4.0.1. We'd appreciate it very much if you can help upgrade wasmer to use dynasm 4.0.1.

wakabat:
Hey so I can confirm that all previous issues have been fixed(well, expect for unaligned loads / stores), thank you!

We've been testing more on the code, and have put together some more fixes in [this commit](https://github.com/wakabat/official-wasmer/commit/828e024b8b8e2c315f9e07ee0c382515f88ddce7). As always, feel free to incorporate those changes into this PR.

Some of the fixes we applied include:

* I've implemented unaligned loads / stores(you can search for `emit_maybe_unaligned_load` and `emit_maybe_unaligned_store`) for some cases. It's likely they still require changes but I consider them to be a starting point for more discussions.
* There are some more codegen missing cases we found in our tests.
* I wasn't so sure about all the `(disp & 0x3) == 0` asserts in [this code piece](https://github.com/marxin/wasmer/blob/fb255b846262dc475a85e8e9a9bbf133f56d0185/lib/compiler-singlepass/src/emitter_riscv.rs#L385-L420). To me it asserts that the offset part of a load / store operation is aligned to 4 bytes. But from the above discussion, it does seem that wasmer right now does not enforce alignments for loads / stores, so any particular reason that the offsets here are 4-byte aligned? In addition, if we do want to add alignment here, maybe we can align based on the bytes read? For example, when reading 2 bytes, we can expect the offset to be aligned by 2, when reading 8 bytes, we can expect the offset to be aligned by 8. And when reading a single byte, we should not need any alignment for the offset. What do you think?

Last, one of our tests hit on the issue that unreachable operator is [not implemented](https://github.com/marxin/wasmer/blob/fb255b846262dc475a85e8e9a9bbf133f56d0185/lib/compiler-singlepass/src/emitter_riscv.rs#L369-L374) yet. In other compilers, this would properly trigger a wasmer trap, but for the moment this is missing in the singlepass compiler for rv64 target, any plan for implementing this?

Again, thanks for all the awesome work!

marxin:
> Hey so I can confirm that all previous issues have been fixed(well, expect for unaligned loads / stores), thank you!

Glad for that :rocket:

> We've been testing more on the code, and have put together some more fixes in [this commit](https://github.com/wakabat/official-wasmer/commit/828e024b8b8e2c315f9e07ee0c382515f88ddce7). As always, feel free to incorporate those changes into this PR.

Happily taking the changes with the following tweaks:

> * I've implemented unaligned loads / stores(you can search for `emit_maybe_unaligned_load` and `emit_maybe_unaligned_store`) for some cases. It's likely they still require changes but I consider them to be a starting point for more discussions.

I've taken the change, where I made a small refactoring connected to the fact we can emit the instructions in a loop.

> * There are some more codegen missing cases we found in our tests.

When it comes to shift operations, I recently added `ImmType::Shift32` and `ImmType::Shift64`, and these types guarantee we won't emit an unexpected immediate values. Feel free to test my branch now.

> * I wasn't so sure about all the `(disp & 0x3) == 0` asserts in [this code piece](https://github.com/marxin/wasmer/blob/fb255b846262dc475a85e8e9a9bbf133f56d0185/lib/compiler-singlepass/src/emitter_riscv.rs#L385-L420). To me it asserts that the offset part of a load / store operation is aligned to 4 bytes. But from the above discussion, it does seem that wasmer right now does not enforce alignments for loads / stores, so any particular reason that the offsets here are 4-byte aligned? In addition, if we do want to add alignment here, maybe we can align based on the bytes read? For example, when reading 2 bytes, we can expect the offset to be aligned by 2, when reading 8 bytes, we can expect the offset to be aligned by 8. And when reading a single byte, we should not need any alignment for the offset. What do you think?

Correct, we should not expect any particular displacement value and the only correct (as far as I known) check is the run-time one (`emit_maybe_unaligned_load`/`emit_maybe_unaligned_store`).

> 
> Last, one of our tests hit on the issue that unreachable operator is [not implemented](https://github.com/marxin/wasmer/blob/fb255b846262dc475a85e8e9a9bbf133f56d0185/lib/compiler-singlepass/src/emitter_riscv.rs#L369-L374) yet. In other compilers, this would properly trigger a wasmer trap, but for the moment this is missing in the singlepass compiler for rv64 target, any plan for implementing this?

Hm, this should be correctly supported (and tested by various `spec` tests). Note RISC-V can't encode directly a payload into the `unimp` instructions and thus I save the `payload` as `li.12 a0, payload as _`. In the run-time library, the trap is correctly decoded and proper `TrapCode` is reported. Please provide a test-case where you won't get the expected output.

Generally speaking, I reached a point, where most of the wasmer tests are passing.

Would be great if you could test the codegen more intensively. Looking forward for a feedback.

wakabat:
>> Last, one of our tests hit on the issue that unreachable operator is [not implemented](https://github.com/marxin/wasmer/blob/fb255b846262dc475a85e8e9a9bbf133f56d0185/lib/compiler-singlepass/src/emitter_riscv.rs#L369-L374) yet. In other compilers, this would properly trigger a wasmer trap, but for the moment this is missing in the singlepass compiler for rv64 target, any plan for implementing this?
>
> Hm, this should be correctly supported (and tested by various spec tests). Note RISC-V can't encode directly a payload into the unimp instructions and thus I save the payload as li.12 a0, payload as _. In the run-time library, the trap is correctly decoded and proper TrapCode is reported. Please provide a test-case where you won't get the expected output.

I think the current code actually relies on an OS trap on `unimp` to happen, wasmer then capture the trap and translate it to correct TrapCode. I'm sure this works for native environment such as a real VisionFive 2 RISC-V board, but for embedded RISC-V chips in bare-metal environment, or in our case in a software VM, traps are not supported, `unimp` simply translates to a VM failure.

If we look at wasmer's LLVM code: https://github.com/wasmerio/wasmer/blob/817bc7af38da0ef41cdb977e685bb3652f5c1bf3/lib/compiler-llvm/src/translator/code.rs#L2383-L2387, `Unreachable` opcode in wasm actually translates to a wasmer trap call directly, which would be correctly picked up in those non-OS environments. Any chance we can change the implementation in the singlepass compiler to mirror the LLVM compiler?

Many thanks for all the other changes! They do look great!

marxin:
> Any chance we can change the implementation in the singlepass compiler to mirror the LLVM compiler?

Sure, that's definitely something that could be implemented by a direct call to the trap function. However, have you considered the other `TrapCode`s that are currently reported via the signalling mechanism:
https://github.com/wasmerio/wasmer/blob/fac92435d61799a9b83fe8c3bb8c9edc1d91be94/lib/types/src/trapcode.rs#L23-L66
?

How will it play with your environment?

wakabat:
Hey thanks for the reply!

> However, have you considered the other TrapCodes that are currently reported via the signalling mechanism

I think it might be a big ask, but ideally, I would appreciate it if we can have all of them wrapped properly in a wasmer lib trap. In fact some trap code is already processed this way in wasmer: https://github.com/wasmerio/wasmer/blob/fac92435d61799a9b83fe8c3bb8c9edc1d91be94/lib/vm/src/libcalls/eh/gcc.rs#L275

In practice, we might have to differ one trap code from another, as they might be processed differently:

* `IntegerDivisionByZero` and likely `IntegerOverflow` are never thrown in RISC-V code, division by zero is handled at ISA level.
* `BadConversionToInteger` is a non-issue for us since we don't use floating point number.
* For trap codes like `StackOverflow`, `HeapAccessOutOfBounds`, `HeapMisaligned`, `TableAccessOutOfBounds`, `IndirectCallToNull`, `BadSignature`, we would simply treat the source program to be faulty, so even if current `unimp` instruction is used, we shall be fine. A VM(well, our RISC-V VM) level error will be thrown anyway.
* `UnreachableCodeReached` is the one we need to be wrapped properly in a wasmer lib trap, since we want to explicitly detect that the wasm `Unreachable` opcode is called.
* For `UncaughtException`, as long as unwinding is added, I think wasmer would throw this in a lib trap anyway.

marxin:
All right, I switched the implementation of the `Unreachable` for the Singlepass in the PR to use the direct call to `wasmer_vm_raise_trap`.

> * `IntegerDivisionByZero` and likely `IntegerOverflow` are never thrown in RISC-V code, division by zero is handled at ISA level.

About this one, note the emitted code does an explicit check and jumps to the 2 generated labels connected to the `TrapCodes`. Thus, the signalling mechanism and the `unimp` instruction is hit. Please try to run a test-case that stress-tests the traps.

Happy to get feedback!

wakabat:
Sorry for the delay, I can confirm that current tip in this PR works for all our existing tests. Thanks for the awesome work!

marxin:
> All right, I switched the implementation of the Unreachable for the Singlepass in the PR to use the direct call to wasmer_vm_raise_trap.

@wakabat Just wanted to let you know that I've decided to skip the change from this initial PR, but it's going to come shortly after we merge this PR as it's a change that influences other targets as well.

Addressed in #5821.

marxin:
@syrusakbary The PR is basically ready to be merged. However, it depends on a smaller tweaks to the CI when it comes to testing of `riscv64gc` target: #5817 #5815 #5809

marxin:
The CI job `Run WAST tests on linux-riscv64` is green when running in CI under QEMU. Yay!

lib/compiler-singlepass/src/codegen.rs
```diff
@@ -6611,6 +6611,8 @@ impl<'a, M: Machine> FuncGen<'a, M> {
         let mut body = self.machine.assembler_finalize()?;
         body.shrink_to_fit();
 
+        save_assembly_to_file("-module-dump.o", &body);
```
syrusakbary: Remove me before merging, use similar strategy as LLVM debug dir

marxin: Sure, I removed it and added a tracking issue for the feature: #5833.

tests/compilers/issues.rs
```diff
@@ -564,6 +572,12 @@ fn gen_wat_sum_function(arguments: usize) -> String {
 fn huge_number_of_arguments_fn(
```
marxin, line 572: Do not forget to remove this.

marxin:
The PR received a couple of fixed that very identified and reduced from running the RISC-V Singlepass compiler on tests for the following Rust crates:
- aho-corasick
- bstr
- bytesize
- glob
- indexmap
- itertools
- itoa
- memchr
- quickcheck
- rustc-demangle
- rust-shlex
- rust-url
- ryu
- strsim-rs
- strum
- time
- tinyvec
- unicode-ident
- uuid

wakabat:
Hey so I was testing the latest commit from this branch, and I found an issue.

The quirky code can be found here: https://github.com/marxin/wasmer/blob/69720a4374dbeb56fac98431be56c5823ed70f70/lib/compiler-singlepass/src/codegen.rs#L2343-L2350

Per [this definition](https://github.com/marxin/wasmer/blob/934b1f68d6b9f933e4ff44b41888fe6342c54dd5/lib/types/src/vmoffsets.rs#L569-L571), we know that signature index / id is actually 32 bit in size, not 64 bit.

We had a [fix](https://github.com/wakabat/official-wasmer/commit/55daaecdca8d81285218a083792079e5c5065b9f) that solves the problem, however this fix is quite simple, so feel free to use any commit to include the fix.

marxin:
> Per [this definition](https://github.com/marxin/wasmer/blob/934b1f68d6b9f933e4ff44b41888fe6342c54dd5/lib/types/src/vmoffsets.rs#L569-L571), we know that signature index / id is actually 32 bit in size, not 64 bit.

Thanks for it, picked the revision.

wakabat:
Hey so upon more testings, we have more to report here.

First, here are some test wasm files: 
[test-wasms.tar.gz](https://github.com/user-attachments/files/24429356/test-wasms.tar.gz)

The first thing to mention is probably a bug: one of the wasm files attached above would fail the compilation process using the singlepass compiler, but works just fine using the LLVM compiler:

```
$ ./target/release/wasmer compile --target riscv64 --singlepass -o test-wasms/1_singlepass.wasmu test-wasms/1.wasm
Compiler: singlepass
Target: riscv64-unknown-unknown-elf
error: failed to compile `test-wasms/1.wasm`
╰─▶ 1: Compilation error: Assembler failed finalization with: ImpossibleRelocation(Dynamic(DynamicLabel(2)))
$ ./target/release/wasmer compile --target riscv64 --llvm -o test-wasms/1_llvm.wasmu test-wasms/1.wasm
Compiler: llvm-opta
Target: riscv64-unknown-unknown-elf
✔ File compiled successfully to `test-wasms/1_llvm.wasmu`.
```

I spent some time digging this, I think the issue is that singlepass compiler uses `j` when jumping to labels, but `j` only allows near jump(can be expressed in signed 20 bit integer), but cannot work with far jump. The included `1.wasm` happens to include code that does a far jump. I have a temporary [fix](https://github.com/wakabat/official-wasmer/commit/36425cfbd720749ca86378543905961ca446d960) that would make compilation work, however I cannot be sure it is free from bug so feel free to fix the behavior in a more proper way.

Next, I want to discuss on generated code performance of the singlepass compiler.

It is worth noting that singlepass compiler generates much larger wasm files than LLVM compiler, in our tests, most singlepass binaries are 10x the size of LLVM binaries:

```
$ ls -lh test-wasms/
total 20M
-rw-rw-r-- 1 mint mint 258K Jan  5 16:11 0_llvm.wasmu
-rw-rw-r-- 1 mint mint 2.3M Jan  5 16:10 0_singlepass.wasmu
-rw-rw-r-- 1 mint mint  71K Jan  5 16:10 0.wasm
-rw-rw-r-- 1 mint mint 404K Jan  5 16:11 1_llvm.wasmu
-rw-rw-r-- 1 mint mint  82K Jan  5 16:10 1.wasm
-rw-rw-r-- 1 mint mint  56K Jan  5 16:11 2_llvm.wasmu
-rw-rw-r-- 1 mint mint 199K Jan  5 16:10 2_singlepass.wasmu
-rw-rw-r-- 1 mint mint  40K Jan  5 16:10 2.wasm
-rw-rw-r-- 1 mint mint 268K Jan  5 16:11 3_llvm.wasmu
-rw-rw-r-- 1 mint mint 2.6M Jan  5 16:10 3_singlepass.wasmu
-rw-rw-r-- 1 mint mint  59K Jan  5 16:10 3.wasm
-rw-rw-r-- 1 mint mint 189K Jan  5 16:11 4_llvm.wasmu
-rw-rw-r-- 1 mint mint 2.0M Jan  5 16:11 4_singlepass.wasmu
-rw-rw-r-- 1 mint mint  52K Jan  5 16:10 4.wasm
-rw-rw-r-- 1 mint mint 292K Jan  5 16:12 5_llvm.wasmu
-rw-rw-r-- 1 mint mint 3.1M Jan  5 16:11 5_singlepass.wasmu
-rw-rw-r-- 1 mint mint  58K Jan  5 16:10 5.wasm
-rw-rw-r-- 1 mint mint 268K Jan  5 16:12 6_llvm.wasmu
-rw-rw-r-- 1 mint mint 2.9M Jan  5 16:11 6_singlepass.wasmu
-rw-rw-r-- 1 mint mint  65K Jan  5 16:10 6.wasm
-rw-rw-r-- 1 mint mint 355K Jan  5 16:12 7_llvm.wasmu
-rw-rw-r-- 1 mint mint 3.6M Jan  5 16:12 7_singlepass.wasmu
-rw-rw-r-- 1 mint mint  67K Jan  5 16:10 7.wasm
```

It's worth mentioning that the runtime performance is much worse: we have a test that involves running all the wasm files above. For LLVM binaries, the test finished in about ~347 seconds, but for singlepass binaries, the test won't stop after running for 11 hours(and it is still running). All the other running environment in the test stays the same, it's just that in one test we use LLVM binaries from those wasm files, in another test we use singlepass binaries from those wasm files.

Even if we did the math now(note one test is still running after 5 hours), we are talking about ~100x runtime performance, and ~10x code size between LLVM compiler and singlepass compiler. I totally get that singlepass compiler will be slower, but is it expected that for larger programs, singlepass compiler can be this slow?

Is there any chance that some particular code sequences in those wasm files cause a regression somewhere? Can you help confirm if this matches your own testing between the 2 compilers? Or perhaps I made a mistake in my fix, and the code runs in an infinite loop?

As always, huge kudos for the work, and thanks a lot for the help!

wakabat:
Upon further checking, I think I misunderstood what `SCRATCH_REG` is used for. And my proposed [fix](https://github.com/wakabat/official-wasmer/commit/36425cfbd720749ca86378543905961ca446d960) is wrong. It is not a performance issue, but that my code is wrong, and the program somehow goes to an infinite loop.

So what really needs to do here, is to change [this line](https://github.com/marxin/wasmer/blob/245fbc9a26b351cf3cdaed1c5fa5b7f2b16b5a24/lib/compiler-singlepass/src/emitter_riscv.rs#L1593) to a `jump` [pseudo-Instructions](https://censoredusername.github.io/dynasm-rs/language/langref_riscv.html#pseudo-instructions) so we can do far-jump. But the problem is: `jump` requires a temporary register, and we don't really have a register to use in `emit_j_label`.

Let me know what you think might be a proper way to solve this. Thanks!

wakabat:
Okay so after some more tries, I think I've fixed the issues. A proposed fix can be found here: https://github.com/wakabat/official-wasmer/commit/e065880e79728b4d15ffc9d4eccd5153d249de2c. This fix passes all our current tests. Feel free to review and incorporate it into this PR.

marxin:
> Okay so after some more tries, I think I've fixed the issues. A proposed fix can be found here: https://github.com/wakabat/official-wasmer/commit/e065880e79728b4d15ffc9d4eccd5153d249de2c. This fix passes all our current tests. Feel free to review and incorporate it into this PR.

Great, thanks for the commit, I've just included it in the PR. Good job.

syrusakbary (approved):

PR #5724, opened 2025-09-10
Singlepass: simplify dynamic regs usage

marxin:
With the latest release of `dynasm-rs` release, dynamic register arguments are supported for all types supporting `Into<u8>` and thus, the current implementation can be rapidly simplified.


syrusakbary:
@marxin this is weird. Some tests were failing on main (like `cargo-deny`), but there are other set of tests that were passing and now are failing (such as `Unit-test packages on std on macos-arm`): https://github.com/wasmerio/wasmer/actions/runs/17621980652/job/50069573724?pr=5724

This last failure is very likely to be related to changes on this PR. It may be just because of the dynasm upgrade or may be by the code of this PR

syrusakbary:
The failing tests are unrelated to this PR. merging

PR #5726, opened 2025-09-11
Multiple fixes to EH codegen

Arshia001:
This PR is a collection of fixes to EH codegen, with the end result being that all spec tests are now uncommented and pass. Specifically, these are the changes:

* support imported tags
* support nested try_table blocks
* support try_table blocks with inputs
* support catch-less try_table blocks

The PR is best reviewed commit-by-commit. The implementation of imported tags is documented in WARP-32.

### DIFF
--- a/lib/compiler/src/engine/artifact.rs
+++ b/lib/compiler/src/engine/artifact.rs
@@ -442,7 +442,7 @@ impl Artifact {
             }
         };
 
-        // This needs to be called before publishind the `eh_frame`.
+        // This needs to be called before publishing the `eh_frame`.
         engine_inner.register_compact_unwind(
             compact_unwind,
             get_got_address(RelocationTarget::LibCall(wasmer_vm::LibCall::EHPersonality)),
--- a/lib/types/src/vmoffsets.rs
+++ b/lib/types/src/vmoffsets.rs
@@ -8,7 +8,7 @@
 
 use crate::{
     FunctionIndex, GlobalIndex, LocalGlobalIndex, LocalMemoryIndex, LocalTableIndex, MemoryIndex,
-    ModuleInfo, SignatureIndex, TableIndex, TagIndex,
+    ModuleInfo, SignatureIndex, TableIndex,
 };
 use more_asserts::assert_lt;
 use std::convert::TryFrom;
@@ -218,7 +218,7 @@ pub struct VMOffsets {
     /// The number of imported memories in the module.
     num_imported_memories: u32,
     /// The number of tags in the module.
-    num_tags: u32,
+    num_tag_ids: u32,
     /// The number of imported globals in the module.
     num_imported_globals: u32,
     /// The number of defined tables in the module.
@@ -232,7 +232,7 @@ pub struct VMOffsets {
     vmctx_imported_functions_begin: u32,
     vmctx_imported_tables_begin: u32,
     vmctx_imported_memories_begin: u32,
-    vmctx_tags_begin: u32,
+    vmctx_tag_ids_begin: u32,
     vmctx_imported_globals_begin: u32,
     vmctx_tables_begin: u32,
     vmctx_memories_begin: u32,
@@ -254,7 +254,7 @@ impl VMOffsets {
             num_imported_functions: cast_to_u32(module.num_imported_functions),
             num_imported_tables: cast_to_u32(module.num_imported_tables),
             num_imported_memories: cast_to_u32(module.num_imported_memories),
-            num_tags: cast_to_u32(module.tags.len()),
+            num_tag_ids: cast_to_u32(module.tags.len()),
             num_imported_globals: cast_to_u32(module.num_imported_globals),
             num_local_tables: cast_to_u32(module.tables.len()),
             num_local_memories: cast_to_u32(module.memories.len()),
@@ -263,7 +263,7 @@ impl VMOffsets {
             vmctx_imported_functions_begin: 0,
             vmctx_imported_tables_begin: 0,
             vmctx_imported_memories_begin: 0,
-            vmctx_tags_begin: 0,
+            vmctx_tag_ids_begin: 0,
             vmctx_imported_globals_begin: 0,
             vmctx_tables_begin: 0,
             vmctx_memories_begin: 0,
@@ -290,7 +290,7 @@ impl VMOffsets {
             num_imported_functions: 0,
             num_imported_tables: 0,
             num_imported_memories: 0,
-            num_tags: 0,
+            num_tag_ids: 0,
             num_imported_globals: 0,
             num_local_tables: 0,
             num_local_memories: 0,
@@ -299,7 +299,7 @@ impl VMOffsets {
             vmctx_imported_functions_begin: 0,
             vmctx_imported_tables_begin: 0,
             vmctx_imported_memories_begin: 0,
-            vmctx_tags_begin: 0,
+            vmctx_tag_ids_begin: 0,
             vmctx_imported_globals_begin: 0,
             vmctx_tables_begin: 0,
             vmctx_memories_begin: 0,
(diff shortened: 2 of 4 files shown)


lib/vm/src/instance/mod.rs
```diff
@@ -70,8 +70,8 @@ pub(crate) struct Instance {
     /// WebAssembly global data.
     globals: BoxedSlice<LocalGlobalIndex, InternalStoreHandle<VMGlobal>>,
 
-    /// WebAssembly global data.
-    tags: BoxedSlice<LocalTagIndex, InternalStoreHandle<VMTag>>,
+    /// WebAssembly tag data. Notably, this stores *all* tags, not just local ones.
+    tags: BoxedSlice<TagIndex, InternalStoreHandle<VMTag>>,
```
syrusakbary, line 74: Great!

lib/types/src/vmoffsets.rs
```diff
@@ -228,11 +228,11 @@ pub struct VMOffsets {
     /// The number of defined globals in the module.
     num_local_globals: u32,
 
+    // Note: missing vmctx_tags_begin because it HAS to start at the zero index.
```
syrusakbary: Why this change?

lib/types/src/vmoffsets.rs
```diff
@@ -209,6 +209,8 @@ const fn align(offset: u32, width: u32) -> u32 {
 pub struct VMOffsets {
     /// The size in bytes of a pointer on the target.
     pointer_size: u8,
+    /// The number of tags in the module.
+    num_tags: u32,
```
syrusakbary: We should make it on the same offset as it was before (we can simply change the name)

lib/compiler/src/engine/artifact.rs
```diff
@@ -845,6 +845,9 @@ impl Artifact {
         self.preinstantiate()?;
 
         let module = self.create_module_info();
+
+        let tags = resolve_tags(&module, imports, context).map_err(InstantiationError::Link)?;
```
syrusakbary, line 849: I think we should do it in the same place as we are resolving the signatures

tests/wast/spec/proposals/exception-handling/binary.wast
syrusakbary, line 1: Why deleting this files?

theduke: Yeah, why all the test deletions?

Do we have equivalent tests?

lib/types/src/vmoffsets.rs
Arshia001: Because we don't have access to the `VMOffsets` instance _or_ the `Module` in the VM. Generally, you need the entire package to calculate the offset to something as `size_of_everything_else_before_me + index * size_of_this_field`. Without the `VMOffsets` we can't know the size of the other stuff, so we put tags first to zero out that information.

Arshia001: Interestingly, though, I think this _breaks_ existing compiled modules; @zebreus actually noticed some cached modules failed to run, and I think this is why. This change breaks hard-coded `VMOffsets` values for code that was compiled before this change.

tests/wast/spec/proposals/exception-handling/binary.wast
Arshia001: They're repeated in other folders.

lib/types/src/vmoffsets.rs
syrusakbary: > Because we don't have access to the VMOffsets

There are two ways to fix this:

1. Hardcode VMOffsets creation: (we have tests for this, so we should be safe)
```rust
let module = ModuleInfo::new();
let offsets = VMOffsets::new(size_of::<*mut u8>() as u8, &module);
```

2. (preferred) Get access to the offsets via the vmctx instance:

```rust
ctx.instance().offsets
```

tests/wast/spec/proposals/exception-handling/try_table.wast
```diff
@@ -19,220 +19,219 @@
...
+   )
+   (i32.const 2)
+ )
+
+  (func (export "catchless-try") (param i32) (result i32)
     (block $h
-      (try_table (result i32) (catch $imported-e0 $h)
-        (throw $imported-e0-alias (i32.const 1))
+      (try_table (result i32) (catch $e0 $h)
+        (try_table (result i32) (call $throw-if (local.get 0)))
```
syrusakbary, line 228: Why we are doing this? We should preserve the syntax/tests as they were

tests/wast/spec/proposals/exception-handling/try_table.wast
```diff
@@ -306,15 +318,17 @@
...
-;; (assert_return (invoke "catch-imported-alias") (i32.const 2))
+(assert_return (invoke "catch-imported-alias") (i32.const 2))
 
-;;(assert_return (invoke "catchless-try" (i32.const 0)) (i32.const 0))
-;;(assert_return (invoke "catchless-try" (i32.const 1)) (i32.const 1))
+(assert_return (invoke "catchless-try" (i32.const 0)) (i32.const 0))
+(assert_return (invoke "catchless-try" (i32.const 1)) (i32.const 1))
 
 ;; (assert_exception (invoke "return-call-in-try-catch"))
 ;; (assert_exception (invoke "return-call-indirect-in-try-catch"))
```
syrusakbary, line 327: Should we uncomment this?

Arshia001: no, this is still not supported because, AFAIK, we don't implement the tailcall proposal today.

lib/compiler/src/engine/artifact.rs
Arshia001: wdym?

tests/wast/spec/proposals/exception-handling/try_table.wast
Arshia001: We didn't change that test, GH's diffing algorithm is comparing code from the function above to this one.

lib/compiler/src/engine/artifact.rs
syrusakbary: Because otherwise it'll be hard to find. We can either move the logic of signatures here, or the tag ones there. But both logic should happen at the same time to put things more structured, so it's easy to find in the future

tests/wast/spec/proposals/exception-handling/try_table.wast
syrusakbary: Got it!

lib/types/src/vmoffsets.rs
```diff
@@ -217,8 +217,8 @@ pub struct VMOffsets {
     num_imported_tables: u32,
     /// The number of imported memories in the module.
     num_imported_memories: u32,
-    /// The number of imported tags in the module.
-    num_imported_tags: u32,
+    /// The number of tags in the module.
+    num_tags: u32,
```
syrusakbary:
```suggestion
    /// The number of tags declarations in the module.
    num_tag_ids: u32,
```

syrusakbary: (to keep consistency with the signatures)

lib/types/src/vmoffsets.rs
```diff
@@ -736,10 +735,10 @@ impl VMOffsets {
...
     }
 
-    /// Return the offset to `VMTagImport` index `index`.
-    pub fn vmctx_vmtag_import(&self, index: TagIndex) -> u32 {
-        assert_lt!(index.as_u32(), self.num_imported_tags);
-        self.vmctx_imported_tags_begin + index.as_u32() * u32::from(self.size_of_vmtag_import())
+    /// Return the offset to `VMSharedTagIndex` index `index`.
+    pub fn vmctx_vmshared_tag_id(&self, index: TagIndex) -> u32 {
+        assert_lt!(index.as_u32(), self.num_tags);
+        self.vmctx_tags_begin + index.as_u32() * u32::from(self.size_of_vmshared_tag_index())
```
syrusakbary: Delete me, it seems is no longer used

lib/types/src/vmoffsets.rs
```diff
@@ -232,7 +232,7 @@ pub struct VMOffsets {
     vmctx_imported_functions_begin: u32,
     vmctx_imported_tables_begin: u32,
     vmctx_imported_memories_begin: u32,
-    vmctx_imported_tags_begin: u32,
+    vmctx_tags_begin: u32,
```
syrusakbary:
```suggestion
    vmctx_tag_ids_begin: u32,
```

syrusakbary: (to keep consistency with the signatures fields naming)

PR #5727, opened 2025-09-11
Replace serde_yml with serde_norway

jayvdb:
# Description

See https://github.com/rustsec/advisory-db/issues/2395

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2123,7 +2123,7 @@ dependencies = [
  "bstr 1.12.0",
  "log",
  "regex-automata 0.4.9",
- "regex-syntax 0.8.5",
+ "regex-syntax",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -6129,6 +6101,12 @@ version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b55eedc365f81a3c32aea49baf23fa965e3cd85bcc28fb8045708c7388d124ef"
 
+[[package]]
+name = "unsafe-libyaml-norway"
+version = "0.2.15"
```
jayvdb, line 6106: Note that unsafe-libyaml-norway is likely also problematic

https://github.com/rustsec/advisory-db/issues/2132#issuecomment-2620400835

But less so than serde-yml

Arshia001:
Hey @jayvdb, thank you for this contribution!

After discussing the issue internally, we decided to switch back to `serde_yaml` for now. Closing this PR in favor of #5741.

PR #5737, opened 2025-09-16
Workaround tokio issues

zebreus:
These are my workarounds for the tokio issues I encountered yesterday.

@theduke maybe you have some good ideas on how to mitigate this properly.

# Description

### DIFF
--- a/lib/wasix/src/bin_factory/exec.rs
+++ b/lib/wasix/src/bin_factory/exec.rs
@@ -38,7 +38,16 @@ pub async fn spawn_exec(
     // any longer
     drop(binary);
 
-    spawn_exec_module(module, env, runtime)
+    // Transfering the module and env to a new thread may be expensive...
+    // We may also be able to use the wasix runtime to spawn the task although that may run in tokio too...
+    if let Ok(_) = tokio::runtime::Handle::try_current() {
+        let cloned_runtime = runtime.clone();
+        tokio::task::spawn_blocking(move || spawn_exec_module(module, env, &cloned_runtime))
+            .await
+            .unwrap()
+    } else {
+        spawn_exec_module(module, env, runtime)
+    }
 }
 
 #[tracing::instrument(level = "trace", skip_all, fields(%name))]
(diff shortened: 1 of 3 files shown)


zebreus:
Can be reproduced with `wasmer run --registry wasmer.wtf zebreus/python-fails-to-load@0.1.0`

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -38,7 +38,16 @@ pub async fn spawn_exec(
     // any longer
     drop(binary);
 
-    spawn_exec_module(module, env, runtime)
+    // Transfering the module and env to a new thread may be expensive...
+    // We may also be able to use the wasix runtime to spawn the task although that may run in tokio too...
+    if let Ok(_) = tokio::runtime::Handle::try_current() {
+        let cloned_runtime = runtime.clone();
+        tokio::task::spawn_blocking(move || spawn_exec_module(module, env, &cloned_runtime))
```
theduke: This is a no-go, spawning of threads must be done through the task manager.

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -38,7 +40,46 @@ pub async fn spawn_exec(
...
+    struct State {
+        result: Option<Result<TaskJoinHandle, SpawnError>>,
+        waker: Option<Waker>,
+    }
+    let state = Arc::new(Mutex::new(State {
+        result: None,
+        waker: None,
+    }));
+
+    let cloned_state = state.clone();
```
theduke, line 52: I don't quite get what's going on here.

We already have VirtualTaskManagerExt::spawn_await => https://github.com/wasmerio/wasmer/blob/56b3de2dde3119a112c42052645e3329c7c79928/lib/wasix/src/runtime/task_manager/mod.rs#L455

lib/wasix/src/state/linker.rs
```diff
@@ -2289,6 +2289,7 @@ impl LinkerState {
                 module_spec,
                 ld_library_path,
             } => {
+                // This is one of the problematic inline wakers
```
Arshia001, line 2292:
```suggestion
                // FIXME: This is one of the problematic inline wakers
```

... if this is a pending issue.

PR #5741, opened 2025-09-16
Replace serde_yml with serde_yaml

Arshia001:


theduke:
Note, for posterity: since none of the forks of serde-yaml are trustworthy, we are going back to the old serde-yaml.

It is deprecated/unmaintained, but no known vulnerabilities exist, so it should still be good to use.

PR #5745, opened 2025-09-17
fix(wasix): Linker must re-use the same Engine

theduke:
The Linker must re-use the same Engine for all loaded modules.
If we try to combine dynamic libraries built by different engines, we get call type mismatch errors when modules call into each other.


syrusakbary (changes requested):
Since this PR is based in #5526 , adding the same comment here:

> The PR looks good, but we need to do the small change to preserve API compatibility as discussed in Slack

PR #5749, opened 2025-09-17
Singlepass: to_dwarf - use directly `gimli::Register` type

marxin:
The PR simplifies the logic where right now, `to_dwarf` function returns an index value (both for GPR and FPR types) and later on, the `u16` is mapped to the `gimli::Register` type. I think it would be better to have the direct mapping between our register values and the `gimli::Registers` at one place (e.g. `GPR::X26 => AArch64::X26`). Plus, having that, we can simplify the `create_unwind_info_from_insts` function as the register order for the GPR `X64` registers maps to the expected order by Windows EH.

### DIFF
--- a/lib/compiler-singlepass/src/arm64_decl.rs
+++ b/lib/compiler-singlepass/src/arm64_decl.rs
@@ -153,8 +153,44 @@ impl AbstractReg for GPR {
         ];
         GPRS.iter()
     }
-    fn to_dwarf(self) -> u16 {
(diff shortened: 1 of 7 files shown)


syrusakbary:
This one looks great as well, but will merge after the final 6.1.0 release

marxin:
Sure, once merging, please squash the commits in the PR.

syrusakbary (approved):

PR #5750, opened 2025-09-17
Singlepass: refactor condition jumps to single fn (`jmp_on_condition`)

marxin:
For the purpose of the upcoming RISC-V support (#5711), I would like to merge the `location_cmp/emit_relaxed_cmp` and `jmp_on_$condition` into a single function. Mainly motivated by the fact the RISC-V architecture does not profile the FLAGS register used for the jump instruction. On the other hand, it provides [direct compare and jump instructions](https://projectf.io/posts/riscv-branch-set/#branch) that, with register operations, directly map to the newly introduced `jmp_on_condition`. Right now, the RISC-V PR contains unnecessary changes where a new temporary register is first used for comparison, and then `jmp_on_true/false` function do the actual jump.

Note the currently used approach might be seen tricky as the codegen is responsible that no other FLAGS-modifying instructions sneaks in between the CMP and the consequent JUMP instruction.

### DIFF
--- a/lib/compiler-singlepass/src/codegen.rs
+++ b/lib/compiler-singlepass/src/codegen.rs
@@ -7,7 +7,7 @@ use crate::{
     common_decl::*,
     config::Singlepass,
     location::{Location, Reg},
-    machine::{Label, Machine, MachineStackOffset, NATIVE_PAGE_SIZE},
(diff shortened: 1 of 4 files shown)


syrusakbary:
This looks good, but I'd like to merge after the 6.1.0 final release

syrusakbary (approved):

PR #5751, opened 2025-09-19
Add "print-artifact-version" command

Arshia001:
This PR adds a (hidden) `print-artifact-version` command, which can be used to decide if the artifact version has changed since the last Wasmer version. Mainly useful for CI.


baalimago:
This isn't needed, see [WARP-38](https://linear.app/wasmer/issue/WARP-38/simple-remote-compilation-caching-for-wasmer-run)

PR #5752, opened 2025-09-19
LLVM: Fix compilation breaking for modules with function names that have sp…

Arshia001:
…ecial meaning in LLVM IR

### DIFF
--- a/lib/compiler-llvm/src/compiler.rs
+++ b/lib/compiler-llvm/src/compiler.rs
@@ -10,6 +10,7 @@ use inkwell::DLLStorageClass;
 use rayon::iter::ParallelBridge;
 use rayon::prelude::{IntoParallelIterator, IntoParallelRefIterator, ParallelIterator};
 use rayon::ThreadPoolBuilder;
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use wasmer_compiler::types::function::{Compilation, UnwindInfo};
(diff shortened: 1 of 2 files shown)


zebreus:
This seems sane and fixes the issues I had with python

lib/compiler-llvm/src/compiler.rs
```diff
@@ -91,6 +92,11 @@ struct ModuleBasedSymbolRegistry {
 }
 
 impl ModuleBasedSymbolRegistry {
+    const PROBLEMATIC_PREFIXES: &[&'static str] = &[
```
zebreus, line 95: @Arshia001 I was not able to find any documentation on where these prefixes come from. Can you add a comment for documentation in a followup PR?

Arshia001:
@zebreus tests are passing too, just lint failures from the bits AI auto-completed for me... Can you push a quick fix for the lint issues so it can be merged in, seeing as I'm on leave?

lib/compiler-llvm/src/compiler.rs
Arshia001: .L, I discovered from this issue. I knew llvm. myself. Didn't come from any specific place. I tried asking AI for others, but it didn't come up with anything.

zebreus: Then let's hope these are all problematic prefixes for now

Arshia001: Indeed. We can add more if we come across them later.

PR #5755, opened 2025-09-19
Rework EH some more to allow native access to exceptions

Arshia001:
Note: PR should not be merged until 6.1.0 final is released.

### DIFF
--- a/lib/compiler-llvm/src/translator/code.rs
+++ b/lib/compiler-llvm/src/translator/code.rs
@@ -13139,13 +13139,6 @@ impl<'ctx> LLVMFunctionCodeGenerator<'ctx, '_> {
                     err!(self.builder.build_unconditional_branch(catch_end_block));
 
                     self.builder.position_at_end(catch_specific_block);
-                    // TODO: do we need to do this?
-                    // let vmctx_alloca = err!(self.build_or_get_vmctx_alloca());
-                    // let vmctx_ptr = err!(self.builder.build_load(
-                    //     self.intrinsics.ptr_ty,
-                    //     vmctx_alloca,
-                    //     "vmctx"
-                    // ));
                     let selector_value = err!(self.builder.build_call(
                         self.intrinsics.personality2,
                         &[self.ctx.basic().into(), uw_exc.into()],
--- a/lib/vm/src/store.rs
+++ b/lib/vm/src/store.rs
@@ -3,143 +3,27 @@ use crate::{
     VMTable, VMTag,
 };
 use core::slice::Iter;
-use fnv::FnvHashMap;
 use std::{cell::UnsafeCell, fmt, marker::PhantomData, num::NonZeroUsize, ptr::NonNull};
 use wasmer_types::StoreId;
-
-/// Trait to represent a collection of objects that can live in a store.
-pub trait StoreObjectList<T> {
-    /// Gets the object at the given index.
-    fn get(&self, index: NonZeroUsize) -> &T;
-
-    /// Mutably gets the object at the given index.
-    fn get_mut(&mut self, index: NonZeroUsize) -> &mut T;
-
-    /// Adds an object to the list, returning its index.
-    fn append(&mut self, obj: T) -> NonZeroUsize;
-
-    /// Gets a pair of mutable references to two elements in the list.
-    /// Only supported for the Vec<T> implementation.
-    fn pair_mut(&mut self, a: NonZeroUsize, b: NonZeroUsize) -> Result<(&mut T, &mut T), ()>;
-
-    /// Deletes the given index from the list, if the operation
-    /// is supported by the specific implementation. Notably,
-    /// the implementation for Vec<T> does *not* support deletion
-    /// because deleting an element would mess up existing indices.
-    #[allow(dead_code)]
-    // FIXME: remove the #allow after we implement exception deallocation
-    fn try_delete(&mut self, index: NonZeroUsize) -> Result<Option<T>, ()>;
-}
-
-impl<T> StoreObjectList<T> for Vec<T> {
-    fn get(&self, index: NonZeroUsize) -> &T {
-        &self[index.get() - 1]
-    }
-
-    fn get_mut(&mut self, index: NonZeroUsize) -> &mut T {
-        &mut self[index.get() - 1]
-    }
-
-    fn append(&mut self, obj: T) -> NonZeroUsize {
-        let idx = self.len();
-        self.push(obj);
-        NonZeroUsize::new(idx + 1).unwrap()
-    }
-
-    fn pair_mut(&mut self, a: NonZeroUsize, b: NonZeroUsize) -> Result<(&mut T, &mut T), ()> {
-        if a == b {
-            panic!("attempted to get two mutable references to the same object");
-        }
-        if a.get() < b.get() {
-            let (low, high) = self.split_at_mut(b.get() - 1);
-            Ok((&mut low[a.get() - 1], &mut high[0]))
-        } else {
-            let (low, high) = self.split_at_mut(a.get() - 1);
-            Ok((&mut high[0], &mut low[b.get() - 1]))
-        }
-    }
-
-    fn try_delete(&mut self, _index: NonZeroUsize) -> Result<Option<T>, ()> {
-        Err(())
-    }
-}
-
-/// A collection of exceptions managed by a store. Exceptions need to be
-/// (de)allocated on the fly, so they can't use the same Vec<T> storage
-/// as other store objects.
-#[derive(Debug)]
-pub struct ExceptionCollection {
-    next_index: usize,
-    exceptions: FnvHashMap<NonZeroUsize, VMExceptionObj>,
-}
-
-impl Default for ExceptionCollection {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
-impl ExceptionCollection {
-    fn new() -> Self {
-        Self {
-            next_index: 1,
-            exceptions: FnvHashMap::default(),
-        }
-    }
-}
-
-impl StoreObjectList<VMExceptionObj> for ExceptionCollection {
-    fn get(&self, index: NonZeroUsize) -> &VMExceptionObj {
-        self.exceptions.get(&index).expect("invalid index")
-    }
-
-    fn get_mut(&mut self, index: NonZeroUsize) -> &mut VMExceptionObj {
-        self.exceptions.get_mut(&index).expect("invalid index")
-    }
-
-    fn append(&mut self, obj: VMExceptionObj) -> NonZeroUsize {
-        let idx = NonZeroUsize::new(self.next_index).unwrap();
-        self.next_index += 1;
-        self.exceptions.insert(idx, obj);
-        idx
-    }
-
-    fn pair_mut(
-        &mut self,
-        _a: NonZeroUsize,
-        _b: NonZeroUsize,
-    ) -> Result<(&mut VMExceptionObj, &mut VMExceptionObj), ()> {
-        Err(())
-    }
-
-    fn try_delete(&mut self, index: NonZeroUsize) -> Result<Option<VMExceptionObj>, ()> {
-        Ok(self.exceptions.remove(&index))
-    }
-}
 
 /// Trait to represent an object managed by a context. This is implemented on
 /// the VM types managed by the context.
 pub trait StoreObject: Sized {
-    /// The type of the list that holds instances of this object.
-    type List: StoreObjectList<Self>;
-
     /// List the objects in the store.
-    fn list(ctx: &StoreObjects) -> &Self::List;
+    fn list(ctx: &StoreObjects) -> &Vec<Self>;
 
     /// List the objects in the store, mutably.
-    fn list_mut(ctx: &mut StoreObjects) -> &mut Self::List;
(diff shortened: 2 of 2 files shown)


lib/types/src/libcalls.rs
```diff
@@ -133,23 +133,18 @@ pub enum LibCall {
...
-    /// alloc_exception
+    /// allocate exception object and get an exnref for it
     AllocException,
-
-    /// delete_exception
-    DeleteException,
-
-    /// read_exception
-    ReadException,
+    /// Get the tag and values out of an exnref
```
Arshia001: TODO: update comment

lib/api/src/backend/sys/entities/function/mod.rs
```diff
@@ -543,7 +543,19 @@ macro_rules! impl_host_function {
                 // See: https://github.com/wasmerio/wasmer/pull/5700
                 match result {
                     Ok(Ok(result)) => return result.into_c_struct(&mut store),
-                    Ok(Err(trap)) => raise_user_trap(trap),
+                    Ok(Err(trap)) => {
+                        // TODO: is it safe to do all this on the WASM stack?
```
Arshia001: @syrusakbary please review this, I don't think it's safe to do it this way according to your comment above.

Arshia001: Another way to do it would be to extract the exception information while we're still in the `on_host_stack` call above, and return an enum with 3 cases (success, trap, exception), which should work.

lib/api/src/backend/js/entities/exception.rs
```diff
@@ -20,42 +20,6 @@ unsafe impl Sync for Exception {}
...
     /// Create a new [`Exception`].
     pub fn new(store: &mut impl AsStoreMut, tag: Tag, payload: &[Value]) -> Self {
-        todo!()
-    }
-}
-
-#[derive(Debug, Clone)]
-#[repr(transparent)]
-/// A WebAssembly `exnref` in `js`.
-pub(crate) struct ExceptionRef;
```
syrusakbary, line 30: Undo deletions, this is needed for wasmer-js

Arshia001: Are we sure? I didn't find any actual implementations of EH in any of the other backends/platforms.

syrusakbary: ~~Yes, pretty much needed~~ I need to double check this

syrusakbary: Also, is needed in v8

Arshia001: I'm super-sure I only deleted empty placeholder code. We can put it back, but it won't do anything. That is to say, I believe natively interacting with exceptions in JS is something we didn't have before.

lib/api/src/backend/sys/entities/function/mod.rs
Arshia001: @syrusakbary ping on this

lib/api/src/backend/sys/entities/function/mod.rs
```diff
@@ -613,7 +625,19 @@ macro_rules! impl_host_function {
...
```
zebreus, line 629: Address the TODO comment

lib/api/src/backend/js/entities/exception.rs
```diff
@@ -20,42 +20,6 @@ unsafe impl Sync for Exception {}
...
-    pub(crate) unsafe fn from_vm_exceptionref(
-        _store: &mut impl AsStoreMut,
-        _vm_exceptionref: VMExceptionRef,
-    ) -> Self {
-        unimplemented!("ExceptionRef is not yet supported in js");
-    }
-
-    pub fn is_from_store(&self, _store: &impl AsStoreRef) -> bool {
-        true
+        unimplemented!("Exception handling is not yet supported in js");
```
syrusakbary, line 23: We need to undo this changes, the ExceptionRef I believe is used in wasmer-js

lib/api/src/backend/sys/entities/function/mod.rs
```diff
@@ -428,9 +431,39 @@ impl Function {
...
+        Err(trap) => {
+            let dyn_err_ref = &trap as &dyn Error;
+            if let Some(runtime_error) = dyn_err_ref.downcast_ref::<RuntimeError>() {
+                if let Some(exception) = runtime_error.to_exception() {
+                    return InvocationResult::Exception(exception);
+                }
+            }
+            InvocationResult::Trap(Box::new(trap))
+        }
+    }
```
syrusakbary, line 460: I'd like to review this in a video call, so I understand it fully first

lib/api/src/entities/exception/inner.rs
```diff
@@ -18,38 +19,65 @@ gen_rt_ty!(Exception
...
-            )),
-            #[cfg(feature = "js")]
-            crate::BackendStore::Js(_) => Self::Js(crate::backend::js::exception::Exception::new(
-                store, tag, payload,
-            )),
-            #[cfg(feature = "jsc")]
-            crate::BackendStore::Jsc(_) => Self::Jsc(
-                crate::backend::jsc::exception::Exception::new(store, tag, payload),
-            ),
+            _ => unimplemented!("new is only implemented for the sys backend"),
```
syrusakbary, line 35: Undo this, we rather let each backend decide if they have implemented it or not

lib/api/src/entities/value.rs
```diff
@@ -404,14 +397,14 @@ impl From<Option<ExternRef>> for Value {
     }
 }
 
-impl From<ExceptionRef> for Value {
-    fn from(val: ExceptionRef) -> Self {
+impl From<Exception> for Value {
```
syrusakbary, line 400: Why changing from ExceptionRef to Exception?

lib/api/src/backend/sys/entities/exception.rs
```diff
@@ -2,87 +2,101 @@
...
 
-#[derive(Debug, Clone)]
-#[repr(transparent)]
-/// A WebAssembly `extern ref` in the `sys` runtime.
-pub(crate) struct ExceptionRef {
-    handle: StoreHandle<wasmer_vm::VMExceptionObj>,
-}
+        let BackendTag::Sys(tag) = &tag.0 else {
+            panic!("cannot create Exception with Tag from another backend");
+        };
```
syrusakbary: Move this logic one level up. This 'SysException" type can receive the Sys type inside of the `BackendTag::Sys instead of the BackendTag itself.

lib/api/src/backend/sys/entities/exception.rs
```diff
@@ -2,87 +2,101 @@
...
 impl Exception {
-    /// Create a new [`Exception`].
-    pub fn new(store: &mut impl AsStoreMut, tag: Tag, payload: &[Value]) -> Self {
-        todo!()
-    }
-}
+    /// Creates a new exception with the given tag and payload, and also creates
+    /// a reference to it, returning the reference.
+    #[allow(irrefutable_let_patterns)]
+    pub fn new(store: &mut impl AsStoreMut, tag: &Tag, payload: &[Value]) -> Self {
```
syrusakbary: tag: `&backend::sys::Tag`

lib/api/src/backend/sys/entities/exception.rs
```diff
@@ -2,87 +2,101 @@
...
-    /// Create an instance of [`Self`] from a [`VMExceptionRef`].
-    pub(crate) unsafe fn from_vm_exceptionref(
-        store: &mut impl AsStoreMut,
-        vm_exceptionref: VMExceptionRef,
-    ) -> Self {
-        Self {
-            handle: unsafe {
-                StoreHandle::from_internal(store.objects_mut().id(), vm_exceptionref.0)
-            },
+    pub fn tag(&self, store: &impl AsStoreRef) -> Tag {
```
syrusakbary: This should return the backend tag

lib/api/src/error.rs
```diff
@@ -153,6 +153,26 @@ impl RuntimeError {
...
 
+    /// Creates a `RuntimeError` containing an exception.
+    ///
+    /// If this error is returned from an imported function, the exception
+    /// will be thrown in the WebAssembly code instead of the usual trapping.
+    pub fn exception(ctx: &impl AsStoreRef, exception: Exception) -> Self {
+        let exnref = exception.vm_exceptionref();
+        let store = ctx.as_store_ref();
+        match store.inner.objects {
+            #[cfg(feature = "sys")]
```
syrusakbary, line 164: Move this one/two lines up, so depending on the Store Type, we do one thing or another

lib/api/src/error.rs
```diff
@@ -198,25 +218,44 @@ impl RuntimeError {
...
-}
 
-impl std::fmt::Debug for RuntimeError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("RuntimeError")
-            .field("source", &self.inner.source)
-            .field("wasm_trace", &self.inner.wasm_trace)
-            .finish()
+    /// Returns true if the `RuntimeError` is an uncaught exception.
+    pub fn is_exception(&self) -> bool {
```
syrusakbary, line 223: We need to check based on the backend of the runtime error. Not all backends will have a functional `self.inner.[...]`

lib/cli/src/commands/run/mod.rs
```diff
@@ -625,7 +631,7 @@ fn invoke_function(
     store: &mut Store,
     func: &Function,
     args: &[String],
-) -> Result<Box<[Value]>, Error> {
+) -> Result<Result<Box<[Value]>, RuntimeError>, Error> {
```
syrusakbary: What is this?

syrusakbary: Research if we can flatten the errors

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -13183,15 +13139,16 @@ impl<'ctx> LLVMFunctionCodeGenerator<'ctx, '_> {
...
-                        vmctx_alloca,
-                        "vmctx"
-                    ));
+                    // TODO: do we need to do this?
+                    // let vmctx_alloca = err!(self.build_or_get_vmctx_alloca());
+                    // let vmctx_ptr = err!(self.builder.build_load(
+                    //     self.intrinsics.ptr_ty,
+                    //     vmctx_alloca,
+                    //     "vmctx"
+                    // ));
```
syrusakbary: Discuss this in a meeting aside of the PR with @marxin

lib/compiler/src/engine/link.rs
```diff
@@ -63,6 +67,7 @@ fn apply_relocation(
             RelocationTarget::CustomSection(custom_section) => {
                 *allocated_sections[custom_section] as usize
             }
+            RelocationTarget::Invalid => panic!("Invalid relocation target encountered"),
```
syrusakbary: This type seems redundant `RelocationTarget::Invalid`. We can simply panic when we are constructing it atm

lib/compiler/src/types/relocation.rs
```diff
@@ -449,10 +449,16 @@ impl RelocationLike for ArchivedRelocation {
...
+    /// A relocation to a dynamic trampoline.
+    DynamicTrampoline(FunctionIndex),
     /// A compiler-generated libcall.
     LibCall(LibCall),
     /// Custom sections generated by the compiler
     CustomSection(SectionIndex),
+    /// An invalid relocation target; this is useful for things that shouldn't be the
+    /// target of a relocation, such as function call trampolines so we'll know to
+    /// error out if we encounter them.
+    Invalid,
```
syrusakbary: Removeme

lib/types/src/libcalls.rs
```diff
@@ -133,23 +133,18 @@ pub enum LibCall {
...
-
-    /// delete_exception
-    DeleteException,
-
-    /// read_exception
-    ReadException,
+    /// Get the values buffer pointer out of an exnref
+    ReadExnRef,
+    /// Given a caught native exception pointer, get the exnref and delete the exception itself
+    NativeExceptionIntoExnRef,
```
syrusakbary: Should we name this: `LibunwindExceptionIntoExnRef` ? Whatever you think fits best

lib/vm/src/store.rs
```diff
@@ -3,42 +3,160 @@ use crate::{
     VMTable, VMTag,
 };
 use core::slice::Iter;
+use fnv::FnvHashMap;
 use std::{cell::UnsafeCell, fmt, marker::PhantomData, num::NonZeroUsize, ptr::NonNull};
 use wasmer_types::StoreId;
 
+/// Trait to represent a collection of objects that can live in a store.
+pub trait StoreObjectList<T> {
```
syrusakbary: We need to think about this deeper (after everything else is solved)

lib/api/src/entities/value.rs
Arshia001: Because we renamed `ExceptionRef` to just `Exception`.

lib/vm/src/store.rs
syrusakbary: Resolution: undo this, and we'll address proper GC for all objects in store aside

syrusakbary (approved):

PR #5756, opened 2025-09-22
Fix stalling issue when using shared libraries

zebreus:
Currently the WASIX tokio runtime does not run `WasiFunctionEnv::new_with_store` in the same execution thread as the actual execution but on the original thread. This can cause issues, because when loading shared libraries `new_with_store` can block with an inline waker which in turn waits for some tokio io. However, in some cases the Inline waker may starve the tokio runtime, which results in a deadlock.

This PR mitigates that issue by executing the `WasiFunctionEnv::new_with_store` function using `tokio::task::block_in_place`. In the longterm we should aim for that initialization to run in the same thread as the execution, but for now this is not possible.

### DIFF
--- a/lib/virtual-io/src/waker.rs
+++ b/lib/virtual-io/src/waker.rs
@@ -34,12 +34,6 @@ impl InlineWaker {
         unsafe { Waker::from_raw(raw_waker) }
     }
 
-    #[cfg(not(feature = "js"))]
-    pub fn block_on<'a, A>(task: impl Future<Output = A> + 'a) -> A {
-        futures::executor::block_on(task)
-    }
-
-    #[cfg(feature = "js")]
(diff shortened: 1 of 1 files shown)


zebreus:
closes #5737

zebreus:
No tests, because I am not able to reproduce this in a more minimal scenario than python with shared libraries.

zebreus:
The solution with `block_in_place` causes issues when using `posix_spawn` to spawn additional processes. It causes an error that we ``cannot execute `LocalPool` executor from within another executor: EnterError`` when we try to use `futures::executor::block_on(task)` in the `InlineWaker`.

I pushed a commit to use tokio instead of the futures LocalPool executor in `virtual-io`, which is one way to fix it. We could also revert to using our bespoke runtime in all scenarios, rather than just the browser.

PR #5757, opened 2025-09-23
Fix order of arguments when spawning processes from WASI

zebreus:
The current behaviour is that the `main-args` from `wasmer.toml` get appended to the passed args when spawning a process inside WASI. This is not the expected behaviour and does not align with the behaviour when 

For example running `something --help` in the example above would result in `something --help main-args-from-wasmer-toml`.  The correct behaviour should be to insert main-args before user-specified arguments like: `something main-args-from-wasmer-toml --help`.

### DIFF
--- a/lib/wasix/src/state/env.rs
+++ b/lib/wasix/src/state/env.rs
@@ -1287,12 +1287,11 @@ impl WasiEnv {
                     .extend_from_slice(env_vars.as_slice());
             }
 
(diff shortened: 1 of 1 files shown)


syrusakbary (approved):
This looks good to me. Assuming the `wp-cli` works properly, this should be good to go

zebreus:
@theduke  Make sure to test this on edge

PR #5761, opened 2025-09-26
Fix deadlocks in Selector

zebreus:
Previously `Selector` would deadlock if the handlers were trying to register/deregister handlers while handling an event.

### DIFF
--- a/lib/virtual-io/src/selector.rs
+++ b/lib/virtual-io/src/selector.rs
@@ -134,7 +134,7 @@ impl Selector {
     }
 
     pub fn shutdown(&self) {
-        self.close_requested.store(true, Ordering::Relaxed);
+        self.close_requested.store(true, Ordering::SeqCst);
         self.wakeup.wake().ok();
     }
 
@@ -246,7 +246,7 @@ impl Selector {
                 let token = event.token();
 
                 if token == engine.token_wakeup {
-                    if engine.close_requested.load(Ordering::Relaxed) {
+                    if engine.close_requested.load(Ordering::SeqCst) {
                         // If exiting was requested, exit the loop
                         return;
                     }


syrusakbary:
@theduke will need to review this PR in depth

Arshia001:
So do I! XD

lib/virtual-io/src/selector.rs
```diff
@@ -88,44 +161,66 @@ impl Selector {
...
+    fn queue_modification(&self, modification: SelectorModification) {
+        // CONCURRENCY: This is safe, because we only lock `queued_modifications` nested which is also safe.
+        if let Ok(mut inner_lookup) = self.lookup.try_lock() {
+            // We got the inner_lookup lock
+            // Process all queued modifications first, to assure they are processed in the correct order
+            self.process_queued_modifications(&mut inner_lookup);
+            modification.apply(&mut inner_lookup);
+        } else {
+            // CONCURRENCY: This will never deadlock as queued_modifications is always the innermost lock and we don't call any potentially blocking functions while holding the lock.
+            self.queued_modifications.lock().unwrap().push(modification);
```
theduke: There is a condition here where this insertion here gets stuck until the next modification is done or the next OS event arrives and triggers an iteration in the run loop.

I think this needs to either spin on the lock until the modification succeeds , or trigger the run loop.

lib/virtual-io/src/selector.rs
```diff
@@ -181,6 +281,18 @@ impl Selector {
...
             }
+
+            // Process any queued up modifications that were caused by the handlers above
+            //
+            // While we could process the queued modifications already in each loop iteration above, we will do it here because:
+            // * If new modifications were added or removed while processing events (in the loop above), they will have been already registered with the source.
+            //   We can not delay that as, we only have a reference to the source when adding/removing, so we can not store it for later processing here.
+            // * However, as all events have already been collected _before_ entering the loop above, it does not matter if we change the set of registered events in there.
+            // * However if we were to switch out the handlers above, we would change the handler for already collected events.
+            // * By delaying the processing of the queued handler modifications until here, we ensure that all events will be processed with the handler that was active when the event was collected.
```
theduke: Not so sure about epoll semantics.

With multiple threads in the mix this is inherently somewhat indeterministic, so the above reasoning should be fine, I believe, but it may also lead to events getting missed.

lib/virtual-io/src/selector.rs
```diff
@@ -8,18 +8,96 @@ use mio::{Registry, Token};
...
 
 #[derive(Debug)]
 pub struct Selector {
     token_close: Token,
-    inner: Mutex<EngineInner>,
+    /// The core assumption here is that this will always be the innermost lock, so we will never deadlock
+    registry: Mutex<Registry>,
+    /// See the comment in `run` for the concurrency model
+    lookup: Mutex<HashMap<Token, Box<dyn InterestHandler + Send + Sync>>>,
+    previous_seed: Mutex<usize>,
```
theduke: This can just be an AtomicUsize , avoiding the need for the lock.

lib/virtual-io/src/selector.rs
```diff
@@ -1,25 +1,109 @@
...
 pub struct Selector {
     token_close: Token,
-    inner: Mutex<EngineInner>,
+    token_wakeup: Token,
+    /// The core assumption here is that this will always be the innermost lock, so we will never deadlock
+    registry: Mutex<Registry>,
+    /// See the comment in `run` for the concurrency model
+    lookup: Mutex<HashMap<Token, Box<dyn InterestHandler + Send + Sync>>>,
+    next_seed: AtomicUsize,
     closer: mio::Waker,
```
theduke: It seems multiple wakers for the same Poll is not allowed?

https://docs.rs/mio/latest/mio/struct.Waker.html#notes

lib/virtual-io/src/selector.rs
zebreus: Ok, that no longer applies

lib/virtual-io/src/selector.rs
zebreus: This should no longer be an issue in the latest patch, as modifications will always go through the queue now.

lib/virtual-io/src/selector.rs
```diff
@@ -88,50 +168,61 @@ impl Selector {
         token: Token,
         source: Option<&mut dyn mio::event::Source>,
     ) -> io::Result<()> {
-        let mut guard = self.inner.lock().unwrap();
-        guard.lookup.remove(&token);
-
+        self.queue_modification(SelectorModification::Remove { token });
+        // CONCURRENCY: This should never result in a deadlock, as long as source.deregister does not call remove or add again.
+        let inner_registry = self.registry.try_lock().unwrap();
```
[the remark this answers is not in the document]
zebreus: Fixed

theduke (approved):
Approved after iterating on details in separate channels.

Issue #5746, opened 2025-09-17
Fuzzing with ASan


StackOverflowExcept1on:
### Motivation

Since we've found at least 2 UBs on Windows and issues with arm64 lately, it might make sense to do some fuzz testing with tools like wasm-gen and the recently added ASan support in corosensei.

### Proposed solution

https://github.com/Amanieu/corosensei/pull/54

Issue #5721, opened 2025-09-04
Unable to link against static Linux Wasmer lib


ashtonmeuser:
### Describe the bug

Since Wasmer v5.0.0, I am unable to link against the static Linux Wasmer library. It seems this version dropped PIC (`-fPIC`) support. 

```sh
g++ -o instance.os -c -fPIC -DGDEXTENSION -DLIBWASM_STATIC -I. -Iwasmer/include instance.cpp
g++ -o build/libwasmer-test.so -shared instance.os wasmer/lib/libwasmer.a
/bin/ld: wasmer/lib/libwasmer.a(corosensei-f1537472378e5be1.corosensei.c1cf870ccca3bb9c-cgu.0.rcgu.o): warning: relocation against `stack_init_trampoline_return' in read-only section `.text'
/bin/ld: wasmer/lib/libwasmer.a(corosensei-f1537472378e5be1.corosensei.c1cf870ccca3bb9c-cgu.0.rcgu.o): relocation R_X86_64_PC32 against symbol `stack_init_trampoline_return' can not be used when making a shared object; recompile with -fPIC
/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
scons: *** [build/libwasmer-test.so] Error 1
scons: building terminated because of errors.
```

### Steps to reproduce

1. Link against Wasmer static library (_wasmer/lib/libwasmer.a_).
2. Note failure to link on Linux present for Wasmer >=v5.0.0 which is not present on <=v4.4.0.

### Expected behavior

Library links correctly.

### Actual behavior

Library fails to link with the following output.

```sh
g++ -o instance.os -c -fPIC -DGDEXTENSION -DLIBWASM_STATIC -I. -Iwasmer/include instance.cpp
g++ -o build/libwasmer-test.so -shared instance.os wasmer/lib/libwasmer.a
/bin/ld: wasmer/lib/libwasmer.a(corosensei-f1537472378e5be1.corosensei.c1cf870ccca3bb9c-cgu.0.rcgu.o): warning: relocation against `stack_init_trampoline_return' in read-only section `.text'
/bin/ld: wasmer/lib/libwasmer.a(corosensei-f1537472378e5be1.corosensei.c1cf870ccca3bb9c-cgu.0.rcgu.o): relocation R_X86_64_PC32 against symbol `stack_init_trampoline_return' can not be used when making a shared object; recompile with -fPIC
/bin/ld: final link failed: bad value
[...]

PR #5763, opened 2025-09-28
chore(lib/vm): update `corosensei` to `0.3.0` to support AddressSanitizer

StackOverflowExcept1on:
# Description

See #5746


syrusakbary:
Thanks for the PR.
As @Amanieu commented on the original issue we would need to use the ScopedCoroutine when upgrading to 0.3.0
https://github.com/wasmerio/wasmer/issues/5721#issuecomment-3342127298

StackOverflowExcept1on:
@syrusakbary thx for quick reply!
```
    Checking wasmer-vm v6.1.0-rc.5 (/mnt/tmpfs/wasmer/lib/vm)
error[E0597]: `stack` does not live long enough
   --> lib/vm/src/trap/traphandlers.rs:953:48
    |
950 |       let mut stack = scopeguard::guard(stack, |stack| STACK_POOL.push(stack));
    |           --------- binding `stack` declared here
...
953 |       let mut coro = Coroutine::with_stack(&mut *stack, move |yielder, ()| {
    |                      -                           ^^^^^ borrowed value does not live long enough
    |  ____________________|
    | |
954 | |         // Save the yielder to TLS so that it can be used later.
955 | |         YIELDER.with(|cell| cell.set(Some(yielder.into())));
...   |
958 | |     });
    | |______- argument requires that `stack` is borrowed for `'static`
...
980 |   }
    |   - `stack` dropped here while still borrowed

For more information about this error, try `rustc --explain E0597`.
error: could not compile `wasmer-vm` (lib) due to 1 previous error
```
looks like we also need to fix stack caching somehow

StackOverflowExcept1on:
@syrusakbary done! I checked API changes in benchmarks here: https://github.com/Amanieu/corosensei/pull/58, https://github.com/Amanieu/corosensei/pull/60.

syrusakbary:
In general it looks good to me (I wonder if it might be better to do `coro.scope` inside the Trap handler instead of outside). @Amanieu I'd love to hear your thoughts!

Note: the tests failing have nothing to do with the code changes from this PR

Amanieu:
It makes no difference, the scope is just there to ensure the coroutine is properly dropped at the end of the scope. This code is fine as it is.

PR #5767, opened 2025-09-29
fix(build): Meaningful compilation errors when no backend feature enabled

theduke:
Currently, when no backend is enabled, we get a horrible mess of errors.

This commit:
* Makes sure the wasmer-compiler crate can be compiled withou the
  compiler feature enabled
* Makes sure the compiler crate fails to compile with a single
  meaningful message if neither core nor std features are enabled
* Makes sure the api crate fails to compile with a single meaninful
  error when no backend is enabled

Note: our current module structure doesn't make it trivial to do this
cleanly.

I went for an approach that has more duplication, but is minimally
invasive.


Arshia001:
LGTM, but can we turn the overly verbose `cfg`s into a macro please?

marxin:
> LGTM, but can we turn the overly verbose `cfg`s into a macro please?

Done that.

@theduke please check my change hasn't changed the expected build behavior?

PR #5770, opened 2025-09-30
Make table methods don’t fail in the browser

syrusakbary:
Ideally we would move as well the Table constructor and grow methods to JS for the value initialization. But we can do this once the PR https://github.com/wasm-bindgen/wasm-bindgen/pull/4698 is merged

### DIFF
--- a/lib/api/src/backend/js/entities/table.rs
+++ b/lib/api/src/backend/js/entities/table.rs
@@ -19,12 +19,15 @@ fn set_table_item(table: &VMTable, item_index: u32, item: &Function) -> Result<(
     table.table.set(item_index, item).map_err(|e| e.into())
 }
 
-fn get_function(store: &mut impl AsStoreMut, val: Value) -> Result<Function, RuntimeError> {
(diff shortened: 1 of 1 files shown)


lib/api/src/backend/js/entities/table.rs
```diff
@@ -85,7 +91,10 @@ impl Table {
         val: Value,
     ) -> Result<(), RuntimeError> {
         let item = get_function(store, val)?;
-        set_table_item(&self.handle, index, &item)
+        if let Some(item) = item {
+            set_table_item(&self.handle, index, &item)?;
```
theduke, line 95: I'm quite worried about this not being fallible anymore.

What's the reason this is needed?

At least needs a comment explaining the behaviour.

syrusakbary: This is needed because otherwise it fails as set table item requires the item not being null and it was receiving a Null FuncRef which raised the exception on the browser.

The proper way of doing it is relying on the browser to doit, which is pending on this PR and it's already added as a comment on this PR:
https://github.com/wasm-bindgen/wasm-bindgen/pull/4698

PR #5774, opened 2025-09-30
SDK cleanup

theduke:
Clean up the wasmer-sdk crate to make it ready for publishing.

* Add some docs
* Move package publishing to submodule
* Rename PublishError to PackagePublishError
* Run expensive operations on blocking threads
* Add custom error variants instead of anyhow
* ...


lib/sdk/src/app/deploy.rs
syrusakbary: I think this should return a `DeployError`, not an `anyhow::Error`. In general any errors from an SDK should be explicit (the CLI is the one where using anyhow makes sense)

Arshia001: Agreed; I'd even go as far as to say this crate shouldn't have an `anyhow` dep at all.

theduke: Whoops, I had updated that stuff on Wednesday already, but didn't push.

theduke: Oh right, so the problem is that the backend-api crate still uses anyhow, so we can't get rid of that.

PR #5780, opened 2025-10-01
Fix/eh nested catch ref

Arshia001:
Note: merge after #5779


zebreus (approved):
I don't completely understand the code, but you added it to the spec tests and it will be tested by the integration tests.

LGTM

PR #5785, opened 2025-10-06
feat(api): Stub backend

### LINKED ISSUE #5787: Stub Backend
Add a new stub backend to the `wasmer` crate, which satisfies all the APIs and makes Wasmer compile , even if no specific backend is enabled.<br><br>**The stub backend would only be compiled in if all the default features are disabled (**`--no-default-features`**, and if no real backend feature is enabled.**

## Motivation

Libraries that build on top of Wasmer should not choose any specific backend implementation, that decision should be up to the "end-user".<br>In practice that means libraries should use `wasmer = { …, default-features = false}`

At the moment it is incredibly cumbersome to write such libraries, because the crate will not compile unless a backend feature is enabled.

This means that you must either:

* Accept that the library does not compile, unless the crate where it is used also enables a specific backend, or the user manually passes `--features wasmer/cranelift` or similar<br><br>This leads to a very suboptimal development experience.
* Mirror the backend feature flags in the library.<br>This  is incredibly burdensome, and leads to an explosion of features that have to be kept track of across libraries.

This is not a theoretical concern, we face this very issue with libraries like `wasmer-wasix`. <br>Just doing `cd lib/wasix; cargo check` will fail to compile. `cd lib/wasix; cargo check -F wasmer/cranelift` is required.<br><br>While manageable for a big core library like wasix, this gets way harder if developers try to compose multiple libraries that add functionality.

## Implementation

* Add a stub backend that mirrors the real APIs, but fails for every action
* Feature-gate the stub backend so it is only compiled in when no real backend is selected<br>This notably means that `cd lib/api; cargo check` will NOT have the stub backend, since cranelift and is enabled by default

## Drawbacks

* Confusion for devs because they get a successfully compiling project, but are not able to use Wasemer because no backend is enabled.
  This issue is mitigated by the fact that you have to manually set `default-features=false`, so you opt out of default features. If you do so, one should already be aware of the different backends and the need to enable features.
  At the moment the same thing just spews compile errors.
  We can further mitigate with documentation updates.

* Extra backend plumbing to maintain for the stub
   Effort should be minimal

theduke:
Add a stub backend.

Closes #5787 
See issue for details.

Still a draft, needs some polishing.


marxin:
Based on our discussion, I don't think adding a stub backend is the right approach.

PR #5799, opened 2025-10-15
fix(wasix): Prevent recursive merge fallback when mounting packages

theduke:
fix: Prevent filesystem merge for packages

    Previously adding additional dependencies to a WASI environment would
    lead to a recursive file system merge, which would clone all files in
    memory, which is very expensive.

    This was caused by a very unstructed creation and extension logic in the
    WasiFs, which could take on a complicated form of <dyn Filesystem>.

    There previosly were attempts to downcast and extend without merging,
    but those would fail in common use cases.

    This is fixed by adding a dedicated WasiFs::Overlay variant, which is a
    dedicated representation for the common wasix package environment use
    case. This allows to deterministically extend an existing overlay with
    additional read-only mounts.

### DIFF
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -9,6 +9,7 @@ mod fd;
 mod fd_list;
 mod inode_guard;
 mod notification;
+pub(crate) mod relative_path_hack;
 
 use std::{
     borrow::{Borrow, Cow},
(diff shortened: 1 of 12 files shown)


lib/wasix/src/fs/mod.rs
```diff
@@ -638,15 +662,28 @@ impl WasiFs {
...
+        match &self.root_fs {
+            WasiFsRoot::Sandbox(fs) => {
+                // TODO: this can be changed to switch to Self::Overlay instead!
+                let fdyn: Arc<dyn FileSystem + Send + Sync> = webc_fs.clone();
+                fs.union(&fdyn);
+                Ok(())
+            }
+            WasiFsRoot::Overlay(overlay) => {
+                let union = &overlay.secondaries()[0];
+                union.0.merge(webc_fs, virtual_fs::UnionMergeMode::Skip)
```
Arshia001, line 683: I am *super* skeptical about skipping entries here. Let's discuss this change over a meeting.

theduke: I don't like it either, but skipping is the EXISTING behaviour! At least for the FS built up for packages.

Changes to that should be done in another PR.

Arshia001 (approved):

PR #5818, opened 2025-10-22
refactor: remove unused parts of FunctionStateMap

marxin:
While working on the Singlepass implementation of the `Operator::Unreachable` to directly make it happen via a function call, I noticed there are plenty of data structures filled by the compiler, but unused. Thus I am suggesting to remove them.
When it comes to `Unreachable` support in VM, the key mapping from offset to WASM offset is based on `InstructionAddressMap`, which I preserve.


syrusakbary:
Good call. This was set in the old times when jumping between different compilers (in a Proof of concept JIT). I agree that shouldn't be needed now

marxin:
Great, then let's dig even deeper, it seems the entire `FunctionStateMap` is actually completely unused in the Singlepass compiler.

syrusakbary (approved):

PR #5820, opened 2025-10-22
ci: enable 2 issues tests to be run on all targets

marxin:


marxin:
This PR already payed off as it exposed an issue in RISC-V target.

PR #5826, opened 2025-10-24
chore(build): fix compilation warnings

marxin:
First part of the warning clean-up, these changes make it possible to run `cargo build --all` without any warning.

### DIFF
--- a/lib/api/tests/memory.rs
+++ b/lib/api/tests/memory.rs
@@ -5,6 +5,7 @@ use std::sync::{
 use wasmer::{Instance, Memory, MemoryLocation, MemoryType, Module, Store, imports};
 
 #[test]
+#[allow(unused_attributes)]
 #[cfg_attr(feature = "wamr", ignore = "wamr ignores import memories")]
 #[cfg_attr(feature = "wasmi", ignore = "wasmi does not support threads")]
 #[cfg_attr(
--- a/lib/api/tests/module.rs
+++ b/lib/api/tests/module.rs
@@ -275,6 +275,7 @@ fn calling_host_functions_with_negative_values_works() -> Result<(), String> {
 }
 
 #[universal_test]
+#[allow(unused_attributes)]
 #[cfg_attr(feature = "wamr", ignore = "wamr does not support custom sections")]
 #[cfg_attr(feature = "wasmi", ignore = "wasmi does not support custom sections")]
 #[cfg_attr(feature = "v8", ignore = "v8 does not support custom sections")]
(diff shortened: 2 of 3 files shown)


lib/api/src/backend/v8/vm/mod.rs
```diff
@@ -133,7 +133,7 @@ impl VMFuncRef {
     }
 }
 
-pub(crate) struct VMExceptionRef(*mut wasm_ref_t);
```
syrusakbary: This should be private. Why it needs to be public?

examples/memory.rs
```diff
@@ -71,8 +71,10 @@ fn main() -> anyhow::Result<()> {
     let mem_size: TypedFunction<(), i32> = instance
         .exports
         .get_typed_function(&mut store, "mem_size")?;
+    #[cfg(not(feature = "wamr"))]
```
syrusakbary, line 74: Why not in wamr?

marxin: Because the code down below makes use of the functions only conditionally:
https://github.com/wasmerio/wasmer/blob/b4bff3c554876b4300a28b102fb508ff2eb62f02/examples/memory.rs#L112-L145

lib/api/src/backend/v8/vm/mod.rs
marxin: The place where the type is exposed to a public API is here:
```
warning: type `backend::v8::vm::VMExceptionRef` is more private than the item `vm::VMExceptionRef::V8::0`
   --> lib/api/src/vm/mod.rs:22:16
    |
22  |             V8(crate::backend::v8::vm::[<VM $name>]),
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ field `vm::VMExceptionRef::V8::0` is reachable at visibility `pub`
...
445 | define_vm_like!(ExceptionRef);
    | ----------------------------- in this macro invocation
    |
note: but type `backend::v8::vm::VMExceptionRef` is only usable at visibility `pub(crate)`
```

marxin: I think the `VMExceptionRef` type must really by public as it's something exposed by a back-end by the following function:
https://github.com/wasmerio/wasmer/blob/cdc05aaa1ee12e46444fc76fbada87668dd062a1/lib/vm/src/trap/trap.rs#L141-L150

syrusakbary: `to_exception_ref` should be private (or public just to the crate), not public. Is not intended for the user to use externally (but by the backend to use internally)

marxin: Sorry, but it's too invasive change - just created an issue for that for now: #5961.

Issue #5735, opened 2025-09-14
I encountered an abnormal termination issue when using the "wasix-1.47.0" branch of Tokio.


zhezzz:
In a Tokio environment, performing a simple print operation using the Tokio dependency from the "wasix-1.47.0" branch will likely get stuck and exit after 10 seconds. However, after reverting to the "wasix-1.35.1" branch and trying multiple times, it works normally.
PS: On one or two occasions, the execution got stuck for so long that I had to force terminate it.
@Arshia001 Hello, perhaps you're more experienced with this issue.
I've attached the log file for the problem.

[1.35.1.log](https://github.com/user-attachments/files/22320325/1.35.1.log)
[1.47.0_normal_exit.log](https://github.com/user-attachments/files/22320326/1.47.0_normal_exit.log)
[1.47.0_stuck_exit.log](https://github.com/user-attachments/files/22320327/1.47.0_stuck_exit.log)

#-----------command
cargo wasix build --release
RUST_LOG=TRACE wasmer run ./target/wasm32-wasmer-wasi/release/hello_rust.wasm
#----------

#------platform info
 rustc --version && wasmer --version && cargo wasix version
rustc 1.89.0 (29483883e 2025-08-04)
wasmer 6.0.1
cargo-wasix 0.1.25
#-------

#-------cargo.toml
[dependencies]
#tokio = { version = "=1.47.0", git = "https://github.com/wasix-org/tokio.git", branch = "wasix-1.47.0", features = ["full"] }
tokio = { version = "=1.35.1", git = "https://github.com/wasix-org/tokio.git", branch = "wasix-1.35.1", features = ["full"] }
#-----

#------main.rs
#[tokio::main]
async fn main() {
    println!("=========");
    println!("=========");
    println!("=========");
}
#------

Arshia001:
Hey @zhezzz, thanks for reporting this! In the logs, I see the print statements ran successfully, so I guess this is a problem with `tokio`'s shutdown procedure? Also, in the "normal exit" log, I don't see any delays, the run starts at `2025-09-14T13:20:08.819347Z` and ends at `2025-09-14T13:20:08.855125Z` which is around 35ms for the entire run. Is Wasmer failing to exit afterwards? Or does it sometimes work as it's supposed to and only fail occasionally?

Arshia001:
And can you please run with `--log-events=all` and post the logs again? That'll make it so that each syscall also gets a log when it's entered, so we can know for sure what's causing the wait.

[...]

PR #5838, opened 2025-10-28
Fix a number of WASIX stalls

Arshia001:
Explanation of `ThreadConditions` changes as requested by @theduke:

The previous implementation had two major flaws:

* The only lock was that of the `DashMap`; once an entry was added or retrieved for waiting, the map would be unlocked before the thread would go to sleep. In the meantime, a different thread could acquire the lock, send a notify, and continue on before the first thread had a chance to go to sleep. This was mitigated by using a Mutex + Condvar for each address, rather than relying solely on the `DashMap`.
  * The Mutex + Condvar solution also means unlocking and going to sleep are now one atomic operation, performed by the OS. This is the only safe way to implement the wait logic.
* The check against the existing value in memory would be performed before any locks were taken; this meant that in the time between checking the value and going to sleep, a second thread could make an atomic write (that would not be observed by the first thread) and send wake up signals. The value check has to be atomic with the sleeping to prevent this. I can't for the life of me explain the entire situation, but the linux docs on futexes explain the same situation. To correct this, we now do a second value check with the `do_wait` function. The original value check is kept as a fast-path failure.

### DIFF
--- a/lib/wasix/src/os/command/builtins/cmd_wasmer.rs
+++ b/lib/wasix/src/os/command/builtins/cmd_wasmer.rs
@@ -4,10 +4,11 @@ use crate::{
     SpawnError,
     bin_factory::spawn_exec_wasm,
     os::task::{OwnedTaskStatus, TaskJoinHandle},
-    runtime::{module_cache::HashedModuleData, task_manager::block_on},
+    runtime::module_cache::HashedModuleData,
 };
 use shared_buffer::OwnedBuffer;
 use virtual_fs::{AsyncReadExt, FileSystem};
+use virtual_mio::block_on;
 use wasmer::FunctionEnvMut;
 use wasmer_package::utils::from_bytes;
 use wasmer_wasix_types::wasi::Errno;
--- a/lib/wasix/src/os/command/mod.rs
+++ b/lib/wasix/src/os/command/mod.rs
@@ -2,12 +2,11 @@ pub mod builtins;
 
 use std::{collections::HashMap, sync::Arc};
 
+use virtual_mio::block_on;
 use wasmer::FunctionEnvMut;
 use wasmer_wasix_types::wasi::Errno;
 
-use crate::{
-    Runtime, SpawnError, WasiEnv, runtime::task_manager::block_on, syscalls::stderr_write,
-};
+use crate::{Runtime, SpawnError, WasiEnv, syscalls::stderr_write};
 
 use super::task::{OwnedTaskStatus, TaskJoinHandle, TaskStatus};
 
--- a/lib/wasix/src/os/console/mod.rs
+++ b/lib/wasix/src/os/console/mod.rs
@@ -21,6 +21,7 @@ use virtual_fs::{
     ArcBoxFile, ArcFile, AsyncWriteExt, CombineFile, DeviceFile, DuplexPipe, FileSystem, Pipe,
     PipeRx, PipeTx, RootFileSystemBuilder, StaticFile, VirtualFile,
 };
+use virtual_mio::block_on;
 #[cfg(feature = "sys")]
 use wasmer::Engine;
 use wasmer_config::package::PackageSource;
@@ -33,7 +34,6 @@ use crate::{
     capabilities::Capabilities,
     os::task::{control_plane::WasiControlPlane, process::WasiProcess},
     runners::wasi::{PackageOrHash, RuntimeOrEngine},
-    runtime::task_manager::block_on,
 };
 
 #[derive(Debug)]
--- a/lib/wasix/src/runtime/mod.rs
+++ b/lib/wasix/src/runtime/mod.rs
@@ -3,8 +3,8 @@ pub mod package_loader;
 pub mod resolver;
 pub mod task_manager;
 
+use self::module_cache::CacheError;
 pub use self::task_manager::{SpawnType, VirtualTaskManager};
-use self::{module_cache::CacheError, task_manager::block_on};
 use module_cache::HashedModuleData;
 use wasmer_config::package::SuggestedCompilerOptimizations;
 use wasmer_types::target::UserCompilerOptimizations as WasmerSuggestedCompilerOptimizations;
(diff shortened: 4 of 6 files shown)


Arshia001:
Note: PR is in draft while I investigate #5735, which may be related.

Note to self: there's also a stray `eprintln!` I should remove.

lib/vm/src/instance/mod.rs
```diff
@@ -829,19 +829,22 @@ impl Instance {
...
-        match memory.do_wait(location, timeout) {
+        let res = match memory.do_wait(location, timeout) {
             Ok(count) => Ok(count),
             Err(_err) => {
                 // ret is None if there is more than 2^32 waiter in queue or some other error
                 Err(Trap::lib(TrapCode::TableAccessOutOfBounds))
             }
-        }
+        };
+        eprintln!("memory_wait returning {:?}", res);
```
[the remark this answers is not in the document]
Arshia001: Thanks, copilot!

Arshia001:
@syrusakbary @theduke PR is now ready for review. Please note this PR is best reviewed commit-by-commit.

lib/wasix/src/os/command/builtins/cmd_wasmer.rs
```diff
@@ -4,7 +4,7 @@ use crate::{
     SpawnError,
     bin_factory::spawn_exec_wasm,
     os::task::{OwnedTaskStatus, TaskJoinHandle},
-    runtime::{module_cache::HashedModuleData, task_manager::InlineWaker},
+    runtime::{module_cache::HashedModuleData, task_manager::block_on},
```
syrusakbary: Use the import from `virtual_mio` instead (where is defined)

lib/wasix/src/os/command/mod.rs
```diff
@@ -6,7 +6,7 @@ use wasmer::FunctionEnvMut;
 use wasmer_wasix_types::wasi::Errno;
 
 use crate::{
-    Runtime, SpawnError, WasiEnv, runtime::task_manager::InlineWaker, syscalls::stderr_write,
+    Runtime, SpawnError, WasiEnv, runtime::task_manager::block_on, syscalls::stderr_write,
```
syrusakbary: Use the `block_on` import from `virtual_mio` instead (where is defined)

lib/wasix/src/os/console/mod.rs
```diff
@@ -33,7 +33,7 @@ use crate::{
     capabilities::Capabilities,
     os::task::{control_plane::WasiControlPlane, process::WasiProcess},
     runners::wasi::{PackageOrHash, RuntimeOrEngine},
-    runtime::task_manager::InlineWaker,
+    runtime::task_manager::block_on,
```
syrusakbary: Use the `block_on` import from `virtual_mio` instead (where is defined)

lib/wasix/src/runtime/mod.rs
```diff
@@ -4,7 +4,7 @@ pub mod resolver;
 pub mod task_manager;
 
 pub use self::task_manager::{SpawnType, VirtualTaskManager};
-use self::{module_cache::CacheError, task_manager::InlineWaker};
+use self::{module_cache::CacheError, task_manager::block_on};
```
syrusakbary: Use the `block_on` import from `virtual_mio` instead (where is defined)

lib/wasix/src/state/env.rs
```diff
@@ -37,7 +37,7 @@ use crate::{
         process::{WasiProcess, WasiProcessId},
         thread::{WasiMemoryLayout, WasiThread, WasiThreadHandle, WasiThreadId},
     },
-    runtime::task_manager::InlineWaker,
+    runtime::task_manager::block_on,
```
syrusakbary: Use the `block_on` import from `virtual_mio` instead (where is defined)

lib/wasix/src/syscalls/mod.rs
```diff
@@ -120,7 +120,7 @@ use crate::{
         process::{MaybeCheckpointResult, WasiProcessCheckpoint},
         thread::{RewindResult, RewindResultType},
     },
-    runtime::task_manager::InlineWaker,
+    runtime::task_manager::block_on,
```
syrusakbary: Use the `block_on` import from `virtual_mio` instead (where is defined)

tests/lib/wast/src/wasi_wast.rs
```diff
@@ -15,7 +15,7 @@ use virtual_fs::{
 };
 use wasmer::{FunctionEnv, Imports, Module, Store};
 use wasmer_types::ModuleHash;
-use wasmer_wasix::runtime::task_manager::{InlineWaker, tokio::TokioTaskManager};
+use wasmer_wasix::runtime::task_manager::{block_on, tokio::TokioTaskManager};
```
syrusakbary, line 18: Use the `block_on` import from `virtual_mio` instead (where is defined)

lib/c-api/src/wasm_c_api/wasi/mod.rs
```diff
@@ -21,7 +21,7 @@ use wasmer_api::{AsStoreMut, Imports, Module};
 use wasmer_wasix::{
     Pipe, PluggableRuntime, WasiEnv, WasiEnvBuilder, WasiFunctionEnv, WasiVersion,
     default_fs_backing, get_wasi_version,
-    runtime::task_manager::{InlineWaker, tokio::TokioTaskManager},
+    runtime::task_manager::{block_on, tokio::TokioTaskManager},
```
syrusakbary, line 24: Use the `block_on` import from `virtual_mio` instead (where is defined)

Arshia001: This crate doesn't have a virtual_mio dep

tests/lib/wast/src/wasi_wast.rs
Arshia001: This crate doesn't have a virtual_mio dep

PR #5844, opened 2025-10-31
Linker: fix initialization order of main module

Arshia001:
This PR fixes the initialization order for main modules in the linker.

Previously, the main function's relocations would run _before_ pending globals were resolved, which would result in the main module reading incorrect values while applying its relocations.


zebreus (approved):
You split up `finalize_link_operation` into `finalize_link_operation` and `initialize_new_modules`, because the relocation functions need to be called after finalizing the globals. Seems like the correct fix for the issue

PR #5849, opened 2025-11-05
fix(singlepass): do not make multi_value default

marxin:
Right now, the multi_value is not supported by Singlepass and thus on can't run any `.wasm` file using the compiler.

Fixes the following issue for a trivial WA module:
```
❯ wasmer run --singlepass debug5.wasm --invoke test --disable-cache
error: The singlepass backend does not support the required features for the Wasm module. You can use --cranelift instead
```


marxin:
Apparently, the change of the default value is pretty invasive and thus I'm going to abandon the PR for now.
We believe we can implement the multi-value support for Singlepass pretty soon.

PR #5850, opened 2025-11-05
chore(LLVM): use write_all

marxin:


syrusakbary (approved):
No idea why we had this before. But looks good

PR #5852, opened 2025-11-05
Linker: stop giving main modules a memory_base of zero + stop assumin…

Arshia001:
…g tls_base is the same as memory_base

### DIFF
--- a/lib/wasix/src/state/linker.rs
+++ b/lib/wasix/src/state/linker.rs
@@ -2636,8 +2636,9 @@ impl InstanceGroupState {
         let dl_instance = DlInstance {
             instance: instance.clone(),
             instance_handles,
-            // The TLS area of a side module's main instance is at the beginning
-            // of its memory
+            // The TLS base of a side module's main instance is read from the module's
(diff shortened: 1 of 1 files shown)


zebreus (approved):

lib/wasix/src/state/linker.rs
```diff
@@ -516,8 +515,8 @@ pub enum LinkError {
     #[error("Failed to initialize WASI(X) module handles: {0}")]
     MainModuleHandleInitFailed(ExportError),
 
-    #[error("Dynamic function index {0} was not allocated via `allocate_function`")]
-    DynamicFunctionIndexNotAllocated(u32),
+    #[error("Bad __tls_base export, expected a global of type I32 or I64")]
```
theduke, line 518: Is this public?
If so this is a breaking change.

ps: Public errors should also always have `#[non_exhaustive]`.

Arshia001: Wasix is pre-release and gets breaking changes all the time, so... 😄

PR #5860, opened 2025-11-10
Async function API support

syrusakbary:
This PR adds two functions mimicing the JS-promise-API for WebAssembly (jspi):
https://github.com/WebAssembly/js-promise-integration

Following the new [JS methods](https://github.com/WebAssembly/js-promise-integration/blob/main/proposals/js-promise-integration/Overview.md), this PR introduces:
* A new `Function::new_async(...)`, `Function::new_with_env_async(...)`, `Function::new_typed_async(...)` and `Funciton::new_typed_with_env_async(...)` as to mimic `new WebAssembly.Suspending(...)`
* A new way of calling suspending-function that is as simple as `function.call_async(...)` instead of just `function.call(...)` to mimic `WebAssembly.promising` API

Note: This PR doesn't support typed native functions just yet.

### DIFF
--- a/lib/api/tests/jspi_async.rs
+++ b/lib/api/tests/jspi_async.rs
@@ -1,0 +1,95 @@
+use std::sync::OnceLock;
+
+use anyhow::Result;
+use futures::{executor::block_on, future};
+use wasmer::{
+    Function, FunctionEnv, FunctionEnvMut, FunctionType, Instance, Module, Store, Type, Value,
+    imports,
+};
+
+#[derive(Default)]
+struct DeltaState {
+    deltas: Vec<f64>,
+    index: usize,
+}
+
+impl DeltaState {
+    fn next(&mut self) -> f64 {
+        let value = self.deltas.get(self.index).copied().unwrap_or(0.0);
+        self.index += 1;
+        value
+    }
+}
+
+fn jspi_module() -> &'static [u8] {
+    static BYTES: OnceLock<Vec<u8>> = OnceLock::new();
+    const JSPI_WAT: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/../../tests/examples/jspi.wat");
+    BYTES.get_or_init(|| wat::parse_file(JSPI_WAT).expect("valid example module"))
+}
+
+#[test]
+fn async_state_updates_follow_jspi_example() -> Result<()> {
+    let wasm = jspi_module();
+    let mut store = Store::default();
+    let module = Module::new(&store, wasm)?;
+
+    let init_state = Function::new_async(
+        &mut store,
+        FunctionType::new(vec![], vec![Type::F64]),
+        |_values| async move {
+            future::ready(()).await;
+            Ok(vec![Value::F64(1.0)])
+        },
+    );
+
+    let delta_env = FunctionEnv::new(
+        &mut store,
+        DeltaState {
+            deltas: vec![0.5, -1.0, 2.5],
+            index: 0,
+        },
+    );
+    let compute_delta = Function::new_with_env_async(
+        &mut store,
+        &delta_env,
+        FunctionType::new(vec![], vec![Type::F64]),
+        |mut env: FunctionEnvMut<DeltaState>, _values| {
+            let delta = env.data_mut().next();
+            async move {
+                future::ready(()).await;
+                Ok(vec![Value::F64(delta)])
+            }
+        },
+    );
+
+    let import_object = imports! {
+        "js" => {
+            "init_state" => init_state,
+            "compute_delta" => compute_delta,
+        }
+    };
+
+    let instance = Instance::new(&mut store, &module, &import_object)?;
+    let get_state = instance.exports.get_function("get_state")?;
+    let update_state = instance.exports.get_function("update_state")?;
+
+    fn as_f64(values: &[Value]) -> f64 {
+        match &values[0] {
+            Value::F64(v) => *v,
+            other => panic!("expected f64 value, got {other:?}"),
+        }
+    }
+
+    assert_eq!(as_f64(&get_state.call(&mut store, &[])?), 1.0);
+
+    let step = |store: &mut Store, func: &wasmer::Function| -> Result<f64> {
+        let result = block_on(func.call_async(store, &[]))?;
+        Ok(as_f64(&result))
+    };
+
+    assert_eq!(step(&mut store, update_state)?, 1.5);
+    assert_eq!(step(&mut store, update_state)?, 0.5);
+    assert_eq!(step(&mut store, update_state)?, 3.0);
+
+    Ok(())
+}


lib/api/src/entities/function/inner.rs
```diff
@@ -247,6 +250,126 @@ impl BackendFunction {
...
+            #[cfg(feature = "wasmi")]
+            crate::BackendStore::Wasmi(_) => unsupported_async_backend("wasmi"),
+            #[cfg(feature = "v8")]
+            crate::BackendStore::V8(_) => unsupported_async_backend("v8"),
+            #[cfg(feature = "js")]
+            crate::BackendStore::Js(_) => unsupported_async_backend("js"),
+            #[cfg(feature = "jsc")]
+            crate::BackendStore::Jsc(_) => unsupported_async_backend("jsc"),
+        }
+    }
```
zebreus, line 276: Do we need new_async (probably yes, this is just a reminder for me)

lib/api/src/entities/function/mod.rs
```diff
@@ -147,6 +153,71 @@ impl Function {
...
+    ) -> Self
+    where
+        F: Fn(FunctionEnvMut<T>, Args) -> Fut + 'static + Send + Sync,
+        Fut: Future<Output = RetsAsResult> + 'static + Send,
+        Args: WasmTypeList,
+        Rets: WasmTypeList,
+        RetsAsResult: IntoResult<Rets>,
+    {
+        Self(BackendFunction::new_typed_with_env_async(store, env, func))
+    }
```
zebreus: Again, do we really need duplicate constructors for all async functions? (again, probably yes)

lib/api/src/backend/sys/entities/function/mod.rs
```diff
@@ -127,6 +121,96 @@ impl Function {
...
+                }
+                .into();
+                let sig = func_ty.clone();
+                let future = func(env, &args);
+                HostCallOutcome::Future {
+                    func_ty: sig,
+                    future: Box::pin(future)
+                        as Pin<Box<dyn Future<Output = Result<Vec<Value>, RuntimeError>> + Send>>,
+                }
+            }
```
zebreus, line 175: The only unsafe part here is handling the raw args

lib/api/tests/jspi_async.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn next(&mut self) -> f64 {
+        let value = self.deltas.get(self.index).copied().unwrap_or(0.0);
+        self.index += 1;
+        value
+    }
+}
+
+fn jspi_module() -> &'static [u8] {
+    static BYTES: OnceLock<Vec<u8>> = OnceLock::new();
+    const JSPI_WAT: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/../../tests/examples/jspi.wat");
```
zebreus, line 26: `jspi.wat` is not checked into the repository

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    let mut future = Box::pin(future);
+    let waker = unsafe { Waker::from_raw(noop_raw_waker()) };
+    let mut cx = Context::from_waker(&waker);
+    match future.as_mut().poll(&mut cx) {
+        Poll::Ready(result) => result,
+        Poll::Pending => Err(RuntimeError::new(
+            "async host functions can only yield inside `call_async`",
+        )),
+    }
```
zebreus, line 185: I suggest adding a doc comment to explain what this function does.

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    fn block_on_future(&self, future: HostFuture) -> Result<Vec<Value>, RuntimeError> {
+        let yielder = unsafe { &*self.yielder };
+        match yielder.suspend(AsyncYield::HostFuture(future)) {
+            AsyncResume::HostFutureReady(result) => result,
+            AsyncResume::Start => unreachable!("coroutine resumed without start"),
+        }
+    }
+}
```
zebreus, line 151: We must ensure that the `AsyncContextGuard`s are dropped in the same order we entered, otherwise we could really mess the global state up.

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -0,0 +1,186 @@
...
+
+impl<'a, S> AsyncCallFuture<'a, S>
+where
+    S: AsStoreMut + 'static,
+{
+    pub(crate) fn new(function: SysFunction, store: &'a mut S, params: Vec<Value>) -> Self {
+        let store_ptr = store as *mut S;
+        let coroutine =
+            Coroutine::new(move |yielder: &Yielder<AsyncResume, AsyncYield>, _resume| {
```
zebreus, line 32: This is nice. Creating a new host stack for every promising call and only suspending on that solves a lot of problems.

However, I think this belongs into the VM crate, so we can expose it to the compilers at some point

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -0,0 +1,186 @@
...
+    S: AsStoreMut + 'static,
+{
+    type Output = Result<Box<[Value]>, RuntimeError>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        loop {
+            if let Some(future) = self.pending_future.as_mut() {
+                match future.as_mut().poll(cx) {
+                    Poll::Ready(result) => {
+                        self.pending_future = None;
```
zebreus, line 63: Is there a reason for pending_future not being a variant of AsyncResume. I might be missing something here but I think they are mutally exclusive Options

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -0,0 +1,186 @@
...
+                    Poll::Pending => return Poll::Pending,
+                }
+            }
+
+            let resume_arg = self.next_resume.take().unwrap_or(AsyncResume::Start);
+            let coroutine = match self.coroutine.as_mut() {
+                Some(coro) => coro,
+                None => return Poll::Ready(self.result.take().expect("polled after completion")),
+            };
+            match coroutine.resume(resume_arg) {
```
Arshia001, line 75: Here, we're resuming the WASM computation on the same thread as the thing that polled us, which is going to be a tokio thread 9 out of 10 times; so we're blocking the tokio threads for WASM computation which will starve the runtime.

Arshia001: Also, we're mixing futures and thread locals which makes this very difficult (if not impossible) to get right; we should probably go for a `tokio::task_local` instead?

syrusakbary: Greenthreads in general need to be resumed from the same thread that was running them. Lets sync on this

Arshia001:
Superceded by #5906.

PR #5861, opened 2025-11-10
feature: rework `--debug-dir` option

marxin:
The PR is a mix of refactoring and new features:
- LLVMCallbacks - trait is removed and the implementation is moved from `cli` to the LLVM compiler crate
- `--debug-llvm-dir` is renamed to `--debug-dir` (while supporting the old option as well), where the option will be used by all of our compilers
- for every compiler, let's create a subdir in the `--debug-dir` (e.g. `cranelift`), which makes it easier to navigate
- name the output files based on the function names
- trampolines names are shortened (e.g. `trampoline_call_300xI_I.s`) as LLVM complains if a path name is too long

### DIFF
--- a/lib/compiler-cranelift/src/config.rs
+++ b/lib/compiler-cranelift/src/config.rs
@@ -38,6 +38,15 @@ impl CraneliftCallbacks {
             File::create(path).expect("Error while creating debug file from Cranelift IR");
         file.write_all(mem_buffer).unwrap();
     }
+
+    /// Writes the object file memory buffer to a debug file.
+    pub fn obj_memory_buffer(&self, kind: &CompiledKind, mem_buffer: &[u8]) {
+        let mut path = self.debug_dir.clone();
+        path.push(format!("{}.o", function_kind_to_filename(kind)));
+        let mut file =
+            File::create(path).expect("Error while creating debug file from Cranelift object");
+        file.write_all(mem_buffer).unwrap();
+    }
 }
 
 // Runtime Environment
--- a/lib/compiler-cranelift/src/compiler.rs
+++ b/lib/compiler-cranelift/src/compiler.rs
@@ -302,6 +302,15 @@ impl CraneliftCompiler {
                     .map_err(|error| CompileError::Codegen(format!("{error:#?}")))?;
                 code_buf.extend_from_slice(result.code_buffer());
 
+                if let Some(callbacks) = self.config.callbacks.as_ref() {
+                    use wasmer_compiler::misc::CompiledKind;
+
+                    callbacks.obj_memory_buffer(
+                        &CompiledKind::Local(compile_info.module.get_function_name(func_index)),
+                        &code_buf,
+                    );
+                }
+
                 let func_relocs = result
                     .buffer
                     .relocs()
--- a/lib/compiler-cranelift/src/trampoline/dynamic_function.rs
+++ b/lib/compiler-cranelift/src/trampoline/dynamic_function.rs
@@ -120,6 +120,13 @@ pub fn make_trampoline_dynamic_function(
         .map_err(|error| CompileError::Codegen(error.inner.to_string()))?;
     code_buf.extend_from_slice(compiled.code_buffer());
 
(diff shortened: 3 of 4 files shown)


lib/compiler-llvm/src/config.rs
```diff
@@ -1,14 +1,20 @@
 use crate::compiler::LLVMCompiler;
+use anyhow::Result;
```
syrusakbary: In libraries we should aim to throw specific errors instead of generic ones via `anyhow`, as it makes easier to catch specific errors from the users of the libraries

lib/compiler-llvm/src/config.rs
```diff
@@ -34,11 +40,88 @@ pub enum CompiledKind {
...
+            "trampoline_dynamic_{}_{}",
+            types_to_signature(func_type.params()),
+            types_to_signature(func_type.results())
+        ),
+        CompiledKind::Module => "module".into(),
+    }
+}
+
+impl LLVMCallbacks {
+    pub fn new(debug_dir: PathBuf) -> Result<Self> {
```
syrusakbary: I believe it would be better to throw the specific Result IO error here :)

lib/compiler-llvm/src/config.rs
```diff
@@ -34,11 +40,88 @@ pub enum CompiledKind {
...
+            types_to_signature(func_type.results())
+        ),
+        CompiledKind::DynamicFunctionTrampoline(func_type) => format!(
+            "trampoline_dynamic_{}_{}",
+            types_to_signature(func_type.params()),
+            types_to_signature(func_type.results())
+        ),
+        CompiledKind::Module => "module".into(),
+    }
+}
```
syrusakbary: I wonder if this logic can be moved one level up (I assume it would be helpful as well for singlepass?)

marxin:
> PS: we should delete the `cli-compiler` crate

Heh, what a target. Let's do it in #5862.

lib/compiler-llvm/src/config.rs
marxin: Oh yeah, will be handy there for the Singlepass support.

lib/compiler-llvm/src/config.rs
```diff
@@ -20,25 +24,51 @@ pub type InkwellModule<'ctx> = inkwell::module::Module<'ctx>;
...
-pub trait LLVMCallbacks: Debug + Send + Sync {
-    fn preopt_ir(&self, function: &CompiledKind, module: &InkwellModule);
-    fn postopt_ir(&self, function: &CompiledKind, module: &InkwellModule);
-    fn obj_memory_buffer(&self, function: &CompiledKind, memory_buffer: &InkwellMemoryBuffer);
-    fn asm_memory_buffer(&self, function: &CompiledKind, memory_buffer: &InkwellMemoryBuffer);
+impl LLVMCallbacks {
+    pub fn new(mut debug_dir: PathBuf) -> Result<Self, io::Error> {
+        // Create the debug dir in case it doesn't exist
+        debug_dir.push("llvm");
+        std::fs::create_dir_all(&debug_dir)?;
```
syrusakbary: Move this two one level up

lib/compiler-cranelift/src/config.rs
```diff
@@ -4,11 +4,43 @@ use cranelift_codegen::{
...
+pub struct CraneliftCallbacks {
+    debug_dir: PathBuf,
+}
+
+impl CraneliftCallbacks {
+    /// Creates a new instance of `CraneliftCallbacks` with the specified debug directory.
+    pub fn new(mut debug_dir: PathBuf) -> Result<Self, io::Error> {
+        // Create the debug dir in case it doesn't exist
+        debug_dir.push("cranelift");
+        std::fs::create_dir_all(&debug_dir)?;
```
syrusakbary: Move this two one level up

marxin: Sorry, don't understand the `one level up`?

marxin: Ah, got it, lemme fix it.

PR #5864, opened 2025-11-10
Added typed function implementation

syrusakbary:
This PR has the draft for typed function implementations.

It currently errors when compiling (which shall be solved by making the Fut not Send + Sync?)
```
error[E0277]: the trait bound `{async closure@jspi_async.rs:132:62}: AsyncHostFunction<(), _, _, WithoutEnv>` is not satisfied
   --> lib/api/tests/jspi_async.rs:132:62
    |
132 |       let async_double = Function::new_typed_async(&mut store, async move |value: i32| {
    |  ________________________-------------------------_____________^
    | |                        |
    | |                        required by a bound introduced by this call
133 | |         future::ready(()).await;
134 | |         Ok(value * 2)
135 | |     });
    | |_____^ the trait `AsyncHostFunction<(), _, _, wasmer::WithoutEnv>` is not implemented for `{async closure@lib/api/tests/jspi_async.rs:132:62: 132:85}`
    |
note: required by a bound in `wasmer::Function::new_typed_async`
   --> /Users/syrusakbary/Development/wasmer/lib/api/src/entities/function/mod.rs:197:12
    |
193 |     pub fn new_typed_async<F, Args, Rets>(store: &mut impl AsStoreMut, func: F) -> Self
    |            --------------- required by a bound in this associated function
...
197 |         F: AsyncHostFunction<(), Args, Rets, WithoutEnv>
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Function::new_typed_async`
    = note: the full name for the type has been written to '/Users/syrusakbary/Development/wasmer/target/debug/deps/jspi_async-a88d2a475a558c5e.long-type-4905548332457747066.txt'
    = note: consider using `--verbose` to print the full type name to the console
```

### DIFF
--- a/lib/api/src/backend/sys/entities/function/mod.rs
+++ b/lib/api/src/backend/sys/entities/function/mod.rs
@@ -7,6 +7,7 @@ use crate::{
     BackendFunction, FunctionEnv, FunctionEnvMut, FunctionType, HostFunction, RuntimeError,
     StoreInner, Value, WithEnv, WithoutEnv,
     backend::sys::{engine::NativeEngineExt, vm::VMFunctionCallback},
+    entities::function::async_host::{AsyncFunctionEnv, AsyncHostFunction},
     entities::store::{AsStoreMut, AsStoreRef, StoreMut},
     sys::async_runtime::AsyncRuntimeError,
     utils::{FromToNativeWasmType, IntoResult, NativeWasmTypeInto, WasmTypeList},
@@ -14,7 +15,8 @@ use crate::{
 };
 use std::panic::{self, AssertUnwindSafe};
 use std::{
-    cell::UnsafeCell, cmp::max, error::Error, ffi::c_void, future::Future, pin::Pin, sync::Arc,
+    cell::UnsafeCell, cmp::max, error::Error, ffi::c_void, future::Future, marker::PhantomData,
+    pin::Pin, sync::Arc,
 };
 use wasmer_types::{NativeWasmType, RawValue};
 use wasmer_vm::{
@@ -257,37 +259,51 @@ impl Function {
         }
     }
 
-    pub(crate) fn new_typed_async<F, Fut, Args, Rets, RetsAsResult>(
-        store: &mut impl AsStoreMut,
-        func: F,
-    ) -> Self
+    pub(crate) fn new_typed_async<F, Args, Rets>(store: &mut impl AsStoreMut, func: F) -> Self
     where
-        F: Fn(Args) -> Fut + 'static + Send + Sync,
-        Fut: Future<Output = RetsAsResult> + 'static + Send,
-        Args: WasmTypeList,
-        Rets: WasmTypeList,
-        RetsAsResult: IntoResult<Rets>,
+        Args: WasmTypeList + 'static,
+        Rets: WasmTypeList + 'static,
+        F: AsyncHostFunction<(), Args, Rets, WithoutEnv> + Send + Sync + 'static,
     {
         let env = FunctionEnv::new(store, ());
+        let signature = FunctionType::new(Args::wasm_types(), Rets::wasm_types());
+        let args_sig = Arc::new(signature.clone());
+        let results_sig = Arc::new(signature.clone());
         let func = Arc::new(func);
-        Self::new_typed_with_env_async(store, &env, move |_env, args| {
+        Self::new_with_env_async(store, &env, signature, move |mut env_mut, values| -> Pin<
+            Box<dyn Future<Output = Result<Vec<Value>, RuntimeError>> + Send>,
+        > {
+            let raw_store = RawStorePtr {
+                ptr: env_mut.as_store_mut().as_raw(),
+            };
+            let args_sig = args_sig.clone();
+            let results_sig = results_sig.clone();
             let func = func.clone();
-            func(args)
+            let args =
+                match typed_args_from_values::<Args>(raw_store, args_sig.as_ref(), values) {
+                Ok(args) => args,
+                Err(err) => return Box::pin(async { Err(err) }),
+            };
+            let future = func
+                .as_ref()
+                .call_async(AsyncFunctionEnv::new(), args);
+            Box::pin(async move {
+                let typed_result = future.await?;
+                typed_results_to_values::<Rets>(raw_store, results_sig.as_ref(), typed_result)
+            })
         })
     }
 
-    pub(crate) fn new_typed_with_env_async<T, F, Fut, Args, Rets, RetsAsResult>(
+    pub(crate) fn new_typed_with_env_async<T, F, Args, Rets>(
         store: &mut impl AsStoreMut,
         env: &FunctionEnv<T>,
         func: F,
     ) -> Self
     where
         T: Send + 'static,
-        F: Fn(FunctionEnvMut<T>, Args) -> Fut + 'static + Send + Sync,
-        Fut: Future<Output = RetsAsResult> + 'static + Send,
-        Args: WasmTypeList,
-        Rets: WasmTypeList,
-        RetsAsResult: IntoResult<Rets>,
+        F: AsyncHostFunction<T, Args, Rets, WithEnv> + Send + Sync + 'static,
+        Args: WasmTypeList + 'static,
+        Rets: WasmTypeList + 'static,
     {
         let signature = FunctionType::new(Args::wasm_types(), Rets::wasm_types());
         let args_sig = Arc::new(signature.clone());
@@ -302,16 +318,16 @@ impl Function {
             let args_sig = args_sig.clone();
             let results_sig = results_sig.clone();
             let func = func.clone();
-            let args = match typed_args_from_values::<Args>(raw_store, args_sig.as_ref(), values) {
+            let args =
+                match typed_args_from_values::<Args>(raw_store, args_sig.as_ref(), values) {
                 Ok(args) => args,
                 Err(err) => return Box::pin(async { Err(err) }),
             };
-            let future = (*func)(env_mut, args);
+            let future = func
+                .as_ref()
+                .call_async(AsyncFunctionEnv::with_env(env_mut), args);
             Box::pin(async move {
-                let typed_result = future
-                    .await
-                    .into_result()
-                    .map_err(|err| RuntimeError::user(Box::new(err)))?;
+                let typed_result = future.await?;
                 typed_results_to_values::<Rets>(raw_store, results_sig.as_ref(), typed_result)
             })
         })
--- a/lib/api/src/entities/function/async_host.rs
+++ b/lib/api/src/entities/function/async_host.rs
@@ -1,0 +1,178 @@
+use std::{future::Future, marker::PhantomData, pin::Pin};
+
+use crate::{
+    FunctionEnvMut, HostFunctionKind, RuntimeError, WasmTypeList, WithEnv, WithoutEnv,
+    utils::{FromToNativeWasmType, IntoResult},
+};
+
+/// Wrapper conveying whether an async host function receives an environment.
+pub enum AsyncFunctionEnv<'a, T, Kind> {
+    /// Used by host functions without an environment.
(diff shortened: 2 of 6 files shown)


lib/api/src/entities/store/store_ref.rs
```diff
@@ -56,13 +56,16 @@ impl StoreMut<'_> {
...
 
-    #[allow(unused)]
-    pub(crate) unsafe fn from_raw(raw: *mut StoreInner) -> Self {
+    /// TODO: temporarily public, make private again
+    ///
+    /// # Safety
+    /// Don't use it XD
+    pub unsafe fn from_raw(raw: *mut StoreInner) -> Self {
         Self {
             inner: unsafe { &mut *raw },
```
zebreus, line 70: Needs to be undone

lib/api/src/backend/sys/entities/function/mod.rs
```diff
@@ -300,16 +318,16 @@ impl Function {
...
-            let args = match typed_args_from_values::<Args>(raw_store, args_sig.as_ref(), values) {
+            let args =
+                match typed_args_from_values::<Args>(raw_store, args_sig.as_ref(), values) {
                 Ok(args) => args,
                 Err(err) => return Box::pin(async { Err(err) }),
             };
-            let future = (*func)(env_mut, args);
+            let future = func
+                .as_ref()
+                .call_async(AsyncFunctionEnv::with_env(env_mut), args);
```
syrusakbary, line 328: I don't think this is good. It enables calling an async function with call instead of call_async. 
It also implicitly creates a Coroutine under the hood by calling `call_async`.

lib/api/Cargo.toml
```diff
@@ -81,6 +81,7 @@ tempfile.workspace = true
 anyhow.workspace = true
 macro-wasmer-universal-test = { version = "6.1.0", path = "./macro-wasmer-universal-test" }
 futures = "0.3"
+tokio = { workspace = true, features = ["full"] }
```
zebreus, line 85: I found the futures executor easier to use for spawn_local and futures is already a dev dep of api

lib/vm/src/store.rs
```diff
@@ -241,7 +241,7 @@ pub struct InternalStoreHandle<T> {
 #[cfg(feature = "artifact-size")]
 impl<T> loupe::MemoryUsage for InternalStoreHandle<T> {
     fn size_of_val(&self, _tracker: &mut dyn loupe::MemoryUsageTracker) -> usize {
-        std::mem::size_of_val(&self)
```
syrusakbary, line 244: ?

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -85,86 +110,107 @@ where
...
 
-enum AsyncResume {
-    Start,
-    HostFutureReady(Result<Vec<Value>, RuntimeError>),
+thread_local! {
+    static CURRENT_CONTEXT: RefCell<Vec<*const CoroutineContext>> = const { RefCell::new(Vec::new()) };
 }
 
-struct AsyncContextState {
+struct CoroutineContext {
```
syrusakbary, line 160: I actually like more the AsyncContextState than Coroutine context. We are not calling coroutine anywhere, so I believe Async{...} is a bit more cohesive

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -85,86 +110,107 @@ where
...
+                // we can still directly run a future that doesn't block
+                // inline.
+                // Note, there can be an async context without an active
+                // coroutine in the following scenario:
+                //   call_async -> wasm code -> imported function ->
+                //   call (non-async) -> wasm_code -> imported async function
+                run_immediate(future)
+            }
+            Some(context) => {
+                let ctx_ref = unsafe { context.as_ref().expect("valid context pointer") };
```
syrusakbary, line 135: Should this return an AsyncRuntimeError ?

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -85,86 +110,107 @@ where
...
+        });
+    }
+
+    fn get_current() -> Option<*const Self> {
+        CURRENT_CONTEXT.with(|cell| cell.borrow().last().copied())
+    }
+
     fn block_on_future(&self, future: HostFuture) -> Result<Vec<Value>, RuntimeError> {
-        let yielder = unsafe { &*self.yielder };
+        let yielder = unsafe { self.yielder.as_ref().expect("yielder pointer valid") };
```
syrusakbary, line 203: Should this return an `AsyncRuntimeError` yield error?

lib/api/src/entities/function/inner.rs
```diff
@@ -147,8 +147,8 @@ impl BackendFunction {
     pub fn new_typed<F, Args, Rets>(store: &mut impl AsStoreMut, func: F) -> Self
     where
         F: HostFunction<(), Args, Rets, WithoutEnv> + 'static + Send + Sync,
-        Args: WasmTypeList,
-        Rets: WasmTypeList,
+        Args: WasmTypeList + 'static,
```
zebreus, line 150: I don't understand why we need `'static` on the WasmTypeList. Could you give me some detail on that?

lib/api/src/backend/sys/async_runtime.rs
Arshia001: not really, if we have a null yielder, it's an unrecoverable internal error

lib/api/src/backend/sys/async_runtime.rs
Arshia001: same here, null = internal implementation error

lib/vm/src/store.rs
Arshia001: just a stray warning, presumably from an update to clippy or sth. self is already a ref, so `size_of_val(&self)` is wrong.

lib/api/src/backend/sys/entities/function/mod.rs
Arshia001: I believe that's from your own changes, @syrusakbary. Also, aren't we already in `new_typed_with_env_async` here?

lib/api/src/backend/sys/async_runtime.rs
Arshia001: We're spawning coroutines, one instance of this type specifically corresponds to one coroutine, and the thing contained inside is a coroutine yielder, so I believe `CoroutineContext` is an appropriate name (especially considering the type is private), but happy to change it back if you don't agree on this.

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -85,86 +110,107 @@ where
...
-        CURRENT_CONTEXT.with(|cell| {
-            cell.set(self.previous);
-        });
-    }
-}
-
 fn run_immediate(
     future: impl Future<Output = Result<Vec<Value>, RuntimeError>> + Send + 'static,
-) -> Result<Vec<Value>, RuntimeError> {
+) -> Result<Vec<Value>, AsyncRuntimeError> {
```
zebreus, line 213: Why do we need a separate type for async runtime error?

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -85,86 +110,107 @@ where
     }
 }
 
-thread_local! {
-    static CURRENT_CONTEXT: Cell<*const AsyncContextState> = const { Cell::new(ptr::null()) };
+pub enum AsyncRuntimeError {
+    YieldOutsideAsyncContext,
+    RuntimeError(RuntimeError),
 }
```
zebreus, line 116: Why do we need a separate AsyncRuntimeError? This could be a runtime error with the YieldOutsideAsyncContext trap code

lib/api/src/entities/function/async_host.rs
```diff
@@ -0,0 +1,178 @@
...
+
+/// Async counterpart to [`HostFunction`](super::host::HostFunction).
+pub trait AsyncHostFunction<T, Args, Rets, Kind>
+where
+    Args: WasmTypeList + 'static,
+    Rets: WasmTypeList,
+    Kind: HostFunctionKind,
+{
+    /// Invoke the host function asynchronously.
+    fn call_async(
```
zebreus, line 46: Why do we need to differentiate between calling a host function synchronously vs async? Can't we use the same interface for both?

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -14,12 +16,34 @@ use crate::{AsStoreMut, RuntimeError, Value};
...
+) -> AsyncCallFuture<'a, S>
+where
+    S: AsStoreMut + 'static,
+{
+    AsyncCallFuture::new(function, store, params)
+}
+
+enum AsyncYield {
+    HostFuture(HostFuture),
+}
```
zebreus, line 32: Enum with one element

lib/api/src/backend/sys/async_runtime.rs
```diff
@@ -14,12 +16,34 @@ use crate::{AsStoreMut, RuntimeError, Value};
...
+pub(crate) fn call_function_async<'a, S>(
+    function: SysFunction,
+    store: &'a mut S,
+    params: Vec<Value>,
+) -> AsyncCallFuture<'a, S>
+where
+    S: AsStoreMut + 'static,
+{
+    AsyncCallFuture::new(function, store, params)
+}
```
zebreus, line 28: What is the difference to  `AsyncCallFuture::new`

lib/api/src/backend/sys/entities/function/mod.rs
```diff
@@ -763,6 +789,9 @@ where
                 )
             },
             Ok(InvocationResult::Trap(trap)) => unsafe { raise_user_trap(trap) },
+            Ok(InvocationResult::YieldOutsideAsyncContext) => unsafe {
+                raise_lib_trap(Trap::lib(TrapCode::YieldOutsideAsyncContext))
+            },
```
zebreus, line 794: Wouldn't be required in YieldOutsideAsyncContext were a trap

lib/api/src/backend/sys/async_runtime.rs
zebreus: Agree with @Arshia001, as this is one of the only things that deal with coroutines directly. Like `async_runtime.rs` contains our async runtime implementation which is using coroutines, so we can use that word in there.

lib/api/Cargo.toml
Arshia001: `future::ready(()).await` doesn't actually suspend; I need tokio for `tokio::task::yield()` which does suspend once.

lib/api/src/entities/function/inner.rs
Arshia001: Anything we can pass from wasm into host functions is already `'static` anyway.

lib/api/src/backend/sys/async_runtime.rs
Arshia001: New error condition, which you can find in the type itself.

lib/api/src/backend/sys/async_runtime.rs
Arshia001: It actually can't. To get a proper stack trace with your error, it has to be generated on the WASM stack (I could be wrong here, correct me if I am @syrusakbary?), so this error has the job of propagating the condition back to the WASM stack so it can raise a lib trap there.

lib/api/src/entities/function/async_host.rs
Arshia001: because they return different types and do different things, I assume?

lib/api/src/backend/sys/entities/function/mod.rs
Arshia001: it _is_ a trap, but you can't raise traps from the host stack.

Arshia001:
Superceded by #5906.

PR #5865, opened 2025-11-10
build: move dependency declarations to top-level Cargo.toml

marxin:
This PR consolidates all dependency declarations into the top-level Cargo.toml, with member crates using workspace = true. The primary motivation is to maintain a single source of truth for dependency versions, making it easier to keep dependencies up-to-date. This approach is already widely adopted in the project—there are currently 550 instances of foo.workspace = true.

The only minor trade-off involves default-features: when a sub-crate requires `default-features = false`, the workspace-level declaration must also use `default-features = false`, which propagates to all consuming crates. In practice, this rarely causes issues—the worst case is a build failure when a downstream crate needs additional features, which is easy to detect and fix. I encountered only a handful of such cases during the migration.

**EDITTED**: Done! And reverted as the `workspace = true` dependencies cannot have overridden `version = ` property. That said, let's stick with the current definition of the repository dependencies.

(As a bonus, we can apply the same strategy to internal dependencies like `wasmer-compiler-llvm = { path = "../compiler-llvm", version = "=6.1.0", optional = true }`. By moving these to workspace dependencies, we can define the version constraint (e.g., `=6.1.0`) once in the top-level Cargo.toml, and member crates simply use `workspace = true`. This eliminates version duplication across internal crates while maintaining the strict version pinning.)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -28,6 +28,7 @@ env:
   NEXTEST_PROFILE: "ci"
   RUSTUP_WINDOWS_PATH_ADD_BIN: 1
   WASI_SDK_VERSION: "22"
+  S3_CACHE_VERSION: "1"
 
 jobs:
   setup:
@@ -77,7 +78,7 @@ jobs:
           path: |
             ~/.cargo/*
             ./target/*
-          key: r22-${{ github.repository }}-${{ runner.os }}-${{ hashFiles('Cargo.lock') }}-wasmer-make-lint-linux-x64
+          key: cache-v${{ env.S3_CACHE_VERSION }}-${{ github.repository }}-${{ runner.os }}-${{ hashFiles('Cargo.lock') }}-wasmer-make-lint-linux-x64
           aws-s3-bucket: wasmer-rust-artifacts-cache
           aws-access-key-id: ${{ secrets.CLOUDFLARE_ARTIFACTS_CACHE_ACCESS_TOKEN }}
           aws-secret-access-key: ${{ secrets.CLOUDFLARE_ARTIFACTS_CACHE_ACCESS_KEY }}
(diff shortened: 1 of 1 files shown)


marxin:
I'm not a workspace expert, but I don't see a downside to declaring single-use dependencies in the top-level Cargo.toml. In fact, this is precisely how projects avoid duplicate dependency declarations across multiple crates. Having dependencies in one place makes it easy to see which versions are outdated at a glance. When using the VSCode Dependi extension, you even get a color-coded visualization of version status.

screenshot-2025-11-11_09-25-25

Cargo.toml
```diff
@@ -153,56 +153,205 @@ mio = "1"
...
+corosensei = "0.3.0"
+crossbeam-queue = "0.3.8"
+fnv = "1.0.3"
+scopeguard = "1.1.0"
+rustversion = "1.0"
+libunwind = "1.3.3"
+mach2 = "0.4.2"
+bitflags = "1.2"
+wai-bindgen-wasmer-impl = "0.2.2"
+tracing-lib = { version = "0.1.26", package = "tracing" }
```
theduke: Why is this using a different package name?

I know this was already present, but might be a good opportunity to clean it up.

Cargo.toml
```diff
@@ -87,10 +85,30 @@ version = "6.1.0"
 
 [workspace.dependencies]
 # Repo-local crates
-wasmer-package = { version = "0.601.0", path = "lib/package" }
-wasmer-config = { path = "./lib/config" }
-wasmer-wasix = { path = "./lib/wasix" }
-wasmer-sdk = { path = "./lib/sdk" }
+macro-wasmer-universal-test = { version = "=6.1.0", path = "lib/api/macro-wasmer-universal-test" }
```
theduke: We don't want to force pin the specific version of dependencies in the common case.

We did pin it before for some crates to ensure consistency, mainly in/for WASIX and the compilers I believe, but it should not be the default.

Eg, if I just depend on wasmer-package, I should not be forced into a specific wasmer-config crate version, etc.

It makes sense to keep these here, but with a regular version constraint, eg `0.601.0`, and keep manual pins in the places where we had them before.

theduke (changes requested):

Cargo.toml
marxin: Apparently, one can't combine `workspace = true, version = 'x.y.z'` and so I'm going to revert the top-level declaration of the in repository crates. Good point.

Cargo.toml
marxin: Yes, cleaned up.

lib/cli/Cargo.toml
```diff
@@ -225,24 +227,24 @@ tokio-tungstenite = { workspace = true, features = [
...
+mac_address = { workspace = true, optional = true }
+#tun-tap = { workspace = true, features = ["tokio"], optional = true }
 
-clap_complete = "4.5.59"
-clap_mangen = "0.2.31"
-zip = { version = "2.4", default-features = false, features = ["deflate"] }
-console = "0.15.8"
-dotenvy = "0.15.7"
-lzma-rs = "0.3.0"
+clap_complete.workspace = true
```
theduke, line 233: in lib/cli/Cargo.toml the wasm-target dep used to be clap = { version =
  "4.5.50", default-features = false, … }, but after switching to workspace = true it now inherits the root
  workspace entry clap = "=4.5.50" which leaves default features on.

PR #5869, opened 2025-11-12
Singlepass: add multivalue support

marxin:
This PR adds **multi-value** support to the Singlepass compiler, enabling blocks, loops, if-expressions, and function calls to handle arbitrary numbers of inputs and outputs instead of being limited to single values.

WebAssembly uses a virtual stack machine where operations consume and produce stack values. The key challenge is managing lifetimes when an operation produces multiple results that outlive existing stack values. We solve this by pre-allocating memory locations (which accommodate both integer and floating-point types) and swapping them with shorter-lived values on the stack.

For function calls, the ABI returns the first values in registers and places additional return values in a caller-allocated stack buffer. At the call site, we reuse the memory space originally allocated for function parameters to store these extra return values, since they have longer lifetimes than the parameters.

New WebAssembly constructs requiring special handling:

1. if/else with parameters — Both branches can consume stack inputs, so we must duplicate these inputs for the else branch. This is safe because only one branch executes at runtime.
1. if without else — When an if block takes parameters but has no else branch, we must implement implicit pass-through semantics: the inputs are copied directly to the outputs when the condition is false.
1. loop with parameters — Loops now require PHI node semantics for their inputs. These values flow from two sources: the loop pre-header (initial entry) and the back-edge (loop continuation). We pre-allocate stack slots for these inputs and restore them at the start of each iteration. Only branches at the end of the loop body propagate values back to the pre-allocate return values.

### DIFF
--- a/lib/compiler-singlepass/src/config.rs
+++ b/lib/compiler-singlepass/src/config.rs
@@ -45,9 +45,7 @@ impl CompilerConfig for Singlepass {
 
     /// Gets the supported features for this compiler in the given target
     fn supported_features_for_target(&self, _target: &Target) -> Features {
-        let mut features = Features::default();
-        features.multi_value(false);
-        features
+        Features::default()
     }
 
     /// Pushes a middleware onto the back of the middleware chain.


lib/compiler-singlepass/src/codegen.rs
```diff
@@ -5340,22 +5522,25 @@ impl<'a, M: Machine> FuncGen<'a, M> {
                             .emit_call_register(this.machine.get_grp_for_call())
                     },
                     // [vmctx, table_index, start_idx, item, len]
-                    [Location::Imm32(table), dest.0, val.0, len.0]
-                        .iter()
-                        .cloned(),
+                    [
+                        (Location::Imm32(table), CanonicalizeType::None),
```
syrusakbary, line 5532: Why we are carrying the CanonicalizeType here? (in general, we canonicalize or not for a program, not just for a loc?)

lib/compiler-singlepass/src/codegen.rs
```diff
@@ -153,24 +154,49 @@ impl WpTypeExt for WpType {
...
+struct ControlFrame<M: Machine> {
+    pub state: ControlState<M>,
     pub label: Label,
-    pub returns: SmallVec<[WpType; 1]>,
-    pub value_stack_depth: usize,
+    pub param_types: SmallVec<[WpType; 8]>,
+    pub return_types: SmallVec<[WpType; 1]>,
+    /// Value stack depth at the beginning of the frame (including params and results).
+    value_stack_depth: usize,
 }
```
syrusakbary, line 179: Nice

lib/compiler-singlepass/src/codegen.rs
```diff
@@ -153,24 +154,49 @@ impl WpTypeExt for WpType {
...
-pub enum ControlState {
+#[derive(Clone)]
+pub enum ControlState<M: Machine> {
     Function,
     Block,
     Loop,
-    If(Label),
+    If {
+        label_else: Label,
+        inputs: SmallVec<[LocationWithCanonicalization<M>; 1]>,
```
syrusakbary, line 166: A bit more info on why inputs here is used would be useful

lib/compiler-singlepass/src/codegen.rs
```diff
@@ -292,7 +319,7 @@ impl<'a, M: Machine> FuncGen<'a, M> {
 
         for (loc, _) in locs.iter().rev() {
             if let Location::Memory(..) = *loc {
-                self.check_location_on_stack(loc, stack_offset)?;
+                self.check_location_on_stack(loc, stack_offset).unwrap();
```
syrusakbary: This should be unwrap or `?` ?

marxin: Technically, that's an internal error (bug), but yes, let's use `?`.

lib/compiler-singlepass/src/codegen.rs
marxin: Well, since my recent refactoring related to `CanonicalizeType`, we tie the info with each `value_stack` value.
That's why the ABI for `emit_call_native` expects the `(Location, CanonicalizeType)` as an input.

lib/compiler-singlepass/src/codegen.rs
marxin: Yep, done.

lib/compiler-singlepass/src/codegen.rs
```diff
@@ -326,6 +354,72 @@ impl<'a, M: Machine> FuncGen<'a, M> {
...
+    ///
+    /// This method reserves memory slots that can accommodate both integer and
+    /// floating-point types, then swaps these slots with the last `stack_slots`
+    /// values on the stack to position them correctly for the block's return values.
+    /// that are already present at the value stack.
+    fn allocate_return_slots_and_swap(
+        &mut self,
+        stack_slots: usize,
+        return_slots: usize,
+    ) -> Result<(), CompileError> {
```
syrusakbary, line 368: Maybe there's some opportunity to optimize this function

syrusakbary: This is from ChatGPT:
https://chatgpt.com/share/e/6915ae29-9f74-8011-878e-0cb19b1ca3f2

SmallVec impl:
```rust
fn allocate_return_slots_and_swap(
    &mut self,
    stack_slots: usize,
    return_slots: usize,
    if return_slots == 0 {
        return Ok(());
    }

    // Drain the last `stack_slots` items from the value stack
    let mut latest_slots = Vec::with_capacity(stack_slots);
    for _ in 0..stack_slots {
        if let Some(val) = self.value_stack.pop() {
            latest_slots.push(val);
        }
    }

    // Allocate new return slots
    let mut extra_slots = Vec::with_capacity(return_slots);
    for _ in 0..return_slots {
        extra_slots.push(self.acquire_location_on_stack()?);
    }

    // Collect all memory locations (existing + new)
    let mut all_mem_slots = Vec::with_capacity(stack_slots + return_slots);
    for (loc, _) in &latest_slots {
        if matches!(loc, Location::Memory(..)) {
            all_mem_slots.push(*loc);
        }
    }
    all_mem_slots.extend(extra_slots.iter().copied());

    // Push the return slots first
    for &loc in &all_mem_slots[..return_slots] {
        self.value_stack.push((loc, CanonicalizeType::None));
    }

    // Then push the previous parameters, relocating if needed
    for (loc, canonicalize) in latest_slots.into_iter().rev() {
        let new_loc = if matches!(loc, Location::Memory(..)) {
            let dest = all_mem_slots.pop().unwrap();
            self.machine.emit_relaxed_mov(Size::S64, loc, dest)?;
            dest
        } else {
            loc
        };
        self.value_stack.push((new_loc, canonicalize));
    }

    Ok(())
}
```

marxin: I tried using as much as possible from the code snippet, but the code is incorrect and misses the reverse order in which the `latest_slots` must be appended to the `self.value_stack`.

PR #5879, opened 2025-11-14
Print more meaningful error message when `__tls_base` export is missing and tls symbols are used.

zebreus:
Previously it was fine not to exports `__tls_base` from the main module, #5852 changed that to be an error. This PR prints a more meaningful error message when the `__tls_base` export is missing but tls symbols are used.

### DIFF
--- a/lib/wasix/src/state/linker.rs
+++ b/lib/wasix/src/state/linker.rs
@@ -3132,11 +3132,9 @@ impl InstanceGroupState {
                             ));
                         }
                         Err(e) => {
-                            panic!("Internal error: failed to resolve exported symbol: {}", e)
+                            panic!("Internal error: failed to resolve exported symbol: {e}")
                         }
                     };
-
-                    // .expect("Internal error: bad in-progress symbol resolution");
 
                     match export {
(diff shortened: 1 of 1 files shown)


zebreus:
@copilot fix the clippy error

lib/wasix/src/state/linker.rs
```diff
@@ -3113,18 +3114,27 @@ impl InstanceGroupState {
...
+                        Ok(e) => e,
+                        Err(ResolveError::NoTlsBaseGlobalExport) => {
+                            return Err(LinkError::MissingTlsBaseExport(
+                                import.module().to_string(),
+                                import.name().to_string(),
+                            ));
+                        }
+                        Err(e) => {
+                            panic!("Internal error: failed to resolve exported symbol: {e}")
+                        }
```
zebreus, line 3136: @Arshia001 I don't really like having to match here, because that will still cause a panic if we unwrap that error type on another path. Do you have a better idea on how to handle this?

marxin:
@Arshia001 can you review it, please?

Arshia001:
Duplicate of #6008, closing.

PR #5881, opened 2025-11-18
Implement WASIX context switching API

zebreus:
# WASIX context switching API

This PR implements an API for context switching (aka, greenthreads, stackswitching, fibers, ...) in WASIX.

## Goal

Provide minimal stack-switching primitives in WASIX that enable implementing greenthreads.

## Motivation

We need some sort of context switching to implement greenlet. We need greenlet because it is a dependency of many other Python packages. To this end, we propose a WASIX context switching API that our greenlet implementation can use to switch between stacks.

## Other changes

This PR also changes some of the undefined behaviour around vfork. Previously, we tried really hard to handle all kinds of undefined behaviour gracefully, but this led to some edge cases and possibly memory leaks.

For example it was possible to fork or vfork in an active vfork. This is something that no sane code should ever do, and would have lead to some really funky states. 

## Testing and verification

- [x] All existing tests pass 
- [x] All calls from WASIX into the main function run through the ContextSwitchingEnvironment
- [] All calls to unwind have an assertion that they only run on the main context
- [x] New tests for context switching functionality

## C API
This is the proposed C API (WIP)
```C
// Header <wasix/context.h>

// --------------------------------------------------------------------------
// WASIX Context Switching API
// --------------------------------------------------------------------------
//
// The WASIX context switching API provides lightweight execution contexts
// with independent call stacks.
//
// A context represents an execution state with its own call stack. It may
// be in one of the following states:
//
// - **active**
//   Currently executing. Only one context is active within a given
//   context-switching environment.
// - **suspended**
//   Execution is suspended and may be resumed with `wasix_context_switch()`.
// - **terminated**
//   The entrypoint has returned, thrown an uncaught exception, or triggered a
//   trap.
// - **deleted**
//   All resources associated with the context have been released; equivalent to
//   a context that was never created.
//
// Contexts exist within a context-switching environment. A context may only
// be resumed in the environment in which it was created. Future versions
// of this API may relax this restriction.
//
// Each context-switching environment has a single main context which is always
// the first active context in an context-switching environment. The main
// context cannot be deleted. If the main context terminates, the entire
// context-switching environment terminates.
//
// If no context-switching environment is available, functions in this API fail
// with `EAGAIN`. The conditions under which a context-switching environment is
// available are undefined.
//
// Behavior not specified here is implementation-defined and may be specified
// in future revisions of this API.
//
// #### Wasmer implementation note
// In wasmer each thread corresponds to a context-switching environment
// beginning at the program’s `main` function. During early initialization, no
// context-switching environment may be available. This behavior may change in
// future versions. In practice you can assume that a context-switching
// environment is available starting from the call to `main`. This means
// that there is no context-switching environment available when static
// initializers are called. If you need that feature, open an issue.
//
// #### Interaction with forking
// Forking will NOT fork the context switching environment. If a process calls
// `fork()`, the replica of the process will not inherit the context-switching
// environment. Instead it will enter a new context_switching environment with
// the currently active context which will be its main context. Calling `fork()`
// is not supported on any context other than the main context.
//
// A process created by `vfork()` will not be in a context-switching
// environment until it calls a function from the `exec` family of functions.

// Opaque identifier referring to a WASIX context.
//
// A context identifier is valid only within the context-switching environment
// from which it was obtained with `wasix_context_create()` or
// `wasix_context_main`.
typedef __wasi_context_id_t wasix_context_id_t;

// Identifier of the main context of the current context-switching
// environment.
//
// If no context-switching environment is active, accessing
// `wasix_context_main` yields an undefined identifier.
#define wasix_context_main (__wasix_context_main())
extern wasix_context_id_t __wasix_context_main(void);

// Create a new context.
//
// Creates a new context in the suspended state. On its first resumption,
// `entrypoint` is invoked within that context.
//
// The entrypoint function must not return, raise an uncaught exception, or
// terminate with a trap. If it does, the context enters the terminated state.
// This typically causes termination of the WASIX program unless host code
// handles the trap or error. Exact behavior may change in future versions.
//
// #### Wasmer implementation note
// In the current Wasmer implementation, termination of the entrypoint always
// resumes the main context and produces a trap:
//
// - Entrypoint terminated with a trap:
//   The same trap is produced in the main context.
// - Entrypoint returned normally:
//   A custom user trap is produced in the main context.
// - Entrypoint threw an uncaught exception:
//   The exception is re-thrown in the main context.
//
// This behavior may change in future versions.
//
// #### Parameters
// - `context_id` pointer to where the new context identifier will be stored
// - `entrypoint` function invoked when the context is first resumed
//
// #### Return Value
// - `0` on success
// - `-1` on failure, with `errno` set
//
// #### Errors
// - `EINVAL` entrypoint is not a valid function with the required signature
// - `ENOTSUP` not in a context-switching environment
// - `<memory-error>` Implementations may set `errno` to one of the
// WASIX-defined
//   memory error codes if `context_id` or memory reachable from the arguments
//   is invalid or inaccessible.
int wasix_context_create(wasix_context_id_t *context_id,
                         void (*entrypoint)(void));

// Destroy a suspended or terminated context.
//
// Destroys a context that is suspended or terminated. After a successful call,
// the context enters the deleted state; its identifier becomes invalid and its
// resources are released. Destroying an already deleted context is a no-op.
//
// The main context cannot be destroyed.
//
// Terminated contexts are effectively equivalent to deleted contexts, but
// `wasix_context_destroy()` must still be called to release resources.
//
// #### Wasmer implementation note
// Destroying a suspended context may produce an opaque user trap in the target
// context, requiring the entrypoint to terminate. External functions that call
// back into WASIX must propagate the trap unchanged; otherwise, behavior is
// undefined.
//
// #### Parameters
// - `context_id` identifier of the context to destroy
//
// #### Return Value
// - `0` on success
// - `-1` on failure, with `errno` set
//
// #### Errors
// - `EINVAL` can not destroy the active context
// - `EINVAL` can not destroy the main context
// - `ENOTSUP` not in a context-switching environment
int wasix_context_destroy(wasix_context_id_t context_id);

// Suspend the active context and resume another.
//
// Suspends the active context and resumes execution in `target_context_id`.
// The resumed context continues from where it was last suspended, or from its
// entrypoint if it has never been resumed.
//
// If `target_context_id` refers to the active context, the call is a no-op.
//
// #### Wasmer implementation note
// In the current Wasmer implementation, calling `wasix_context_switch()` in
// the main context may produce a trap if another context terminates while the
// main context is suspended. See the implementation note in
// `wasix_context_create()`.
//
// #### Parameters
// - `target_context_id` identifier of the suspended context to resume
//
// #### Return Value
// - `0` on success
// - `-1` on error, with `errno` set
//
// #### Errors
// - `EINVAL` target context is not suspended
// - `ENOTSUP` not in a context-switching environment
int wasix_context_switch(wasix_context_id_t target_context_id);
```

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -175,18 +175,7 @@ jobs:
           sudo apt-get install -y git make lld curl
 
       - name: Install wasixcc
-        if: matrix.metadata.build != 'windows-x64'
-        run: |
-          export RUST_LOG=wasixcc=trace
-          export GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
-          cargo install --git https://github.com/wasix-org/wasixcc.git --rev 2df8392a1cb9e7b050130f6ac01cd84d9e39406f -F bin
-          mkdir wasixcc-install
-          wasixcc --install-executables $HOME/.wasixcc/bin
-          echo $HOME/.wasixcc/bin >> $GITHUB_PATH
-          wasixcc --download-sysroot v2025-12-10.1
-          wasixcc --download-llvm
-          wasixcc --download-binaryen
-          echo $HOME/.wasixcc/binaryen/bin >> $GITHUB_PATH
+        uses: wasix-org/wasixcc@main
 
       - name: Tool versions
         run: |
@@ -884,17 +873,7 @@ jobs:
 
       - name: Install wasixcc
         if: matrix.metadata.build != 'windows-x64'
-        run: |
-          export RUST_LOG=wasixcc=trace
-          export GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}
-          cargo install --git https://github.com/wasix-org/wasixcc.git --rev 2df8392a1cb9e7b050130f6ac01cd84d9e39406f -F bin
-          mkdir wasixcc-install
-          wasixcc --install-executables $HOME/.wasixcc/bin
-          echo $HOME/.wasixcc/bin >> $GITHUB_PATH
-          wasixcc --download-sysroot v2025-12-10.1
-          wasixcc --download-llvm
-          wasixcc --download-binaryen
-          echo $HOME/.wasixcc/binaryen/bin >> $GITHUB_PATH
+        uses: wasix-org/wasixcc@main
 
       - name: Install LLVM
         shell: bash


lib/wasix/src/state/env.rs
```diff
@@ -208,6 +217,14 @@ impl Clone for WasiEnv {
             replaying_journal: self.replaying_journal,
             skip_stdio_during_bootstrap: self.skip_stdio_during_bootstrap,
             disable_fs_cleanup: self.disable_fs_cleanup,
+            contexts: self.contexts.clone(),
+            // TODO: This is wrong; The contexts can't really be cloned. What do we even use clone on WasiEnv for?
```
Arshia001: for spawning processes and maybe threads IIRC. We should probably name the method differently so it's more descriptive. Also, when spawning threads or processes, the contexts should not be carried over.

lib/wasix/src/state/env.rs
```diff
@@ -249,6 +266,11 @@ impl WasiEnv {
             replaying_journal: false,
             skip_stdio_during_bootstrap: self.skip_stdio_during_bootstrap,
             disable_fs_cleanup: self.disable_fs_cleanup,
+            // TODO: Not sure if we can even properly fork coroutines at all
```
Arshia001: We can't, and shouldn't need to. Forking is not supported with the new EH configuration anyway. Safe to just panic/error out if a guest tries to fork with active contexts.

lib/wasix/src/syscalls/wasix/context_destroy.rs
```diff
@@ -0,0 +1,60 @@
...
+use std::collections::BTreeMap;
+use std::sync::atomic::AtomicU32;
+use std::sync::{Arc, OnceLock, RwLock};
+use wasmer::{
+    AsStoreMut, Function, FunctionEnv, FunctionEnvMut, FunctionType, Instance, Memory, Module,
+    RuntimeError, Store, Value, imports,
+};
+use wasmer::{StoreMut, Tag, Type};
+
+/// ### `context_delete()`
```
Arshia001: Most descriptive doc comment of 2025 award goes to.... XD

lib/wasix/src/syscalls/wasix/context_delete.rs
```diff
@@ -0,0 +1,60 @@
...
+pub fn context_delete(
+    mut ctx: FunctionEnvMut<'_, WasiEnv>,
+    target_context_id: u64,
+) -> Result<Errno, WasiError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
+    let env = ctx.data();
+    let memory: MemoryView<'_> = unsafe { env.memory_view(&ctx) };
+
+    // TODO: Review which Ordering is appropriate here
```
Arshia001: Relaxed should be fine, since coroutines (contexts) can't be active in Parallel.

lib/wasix/src/syscalls/wasix/context_destroy.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    if target_context_id == MAIN_CONTEXT_ID {
+        tracing::trace!(
+            "Context {} tried to delete the main context, which is not allowed",
+            own_context_id
+        );
+        return Ok(Errno::Inval);
+    }
+
+    // TODO: actually delete the context
```
Arshia001: This should probably happen by resuming the context with a special value (`ContextResume::Continue` and `ContextResume::Shutdown`?) and letting it take care of trapping itself out.

lib/wasix/src/syscalls/wasix/context_delete.rs
```diff
@@ -0,0 +1,60 @@
...
+    if target_context_id == MAIN_CONTEXT_ID {
+        tracing::trace!(
+            "Context {} tried to delete the main context, which is not allowed",
+            own_context_id
+        );
+        return Ok(Errno::Inval);
+    }
+
+    // TODO: actually delete the context
+    let removed_future = env.contexts.remove(&target_context_id);
```
Arshia001: That reminds me, what happens if a context traps?

lib/wasix/src/syscalls/wasix/context_new.rs
```diff
@@ -0,0 +1,187 @@
...
+    current_context_id.store(new_context_id, Ordering::SeqCst);
+
+    match prelaunch_result {
+        Ok(_) => (),
+        Err(canceled) => {
+            tracing::trace!(
+                "Context {new_context_id} was canceled before it even started: {canceled}",
+            );
+            // TODO: Handle cancellation properly
+            panic!("Sender was dropped: {canceled}");
```
Arshia001: I'm sure you know this already, but just to reiterate, it's *super* important that WASIX does not ever panic, since a panic can shut down an entire Edge server process.

lib/wasix/src/syscalls/wasix/context_new.rs
```diff
@@ -0,0 +1,187 @@
...
+            // TODO: Handle returning functions with a real error type
+            RuntimeError::user(
+                format!(
+                    "Context {new_context_id} returned a value ({v:?}). This is not allowed for now"
+                )
+                .into(),
+            )
+        }
+    };
+    main_context
```
Arshia001: Should the error be propagated to the main context, or just the "parent context"? I know we don't have parent-child contexts at this time, but sending all errors to main sounds like it could cause lots of complications.

lib/wasix/src/syscalls/wasix/context_switch.rs
```diff
@@ -0,0 +1,105 @@
...
+    match WasiEnv::do_pending_operations(&mut ctx) {
+        Ok(()) => {}
+        Err(e) => {
+            return Err(Err(RuntimeError::user(Box::new(e))));
+        }
+    }
+
+    let (data) = ctx.data_mut();
+    // TODO: Review which Ordering is appropriate here
+    let own_context_id = data
```
Arshia001: Actually, since contexts can't be active in parallel, why are we even using an atomic here?

lib/wasix/src/syscalls/wasix/context_switch.rs
```diff
@@ -0,0 +1,105 @@
...
+        .swap(target_context_id, Ordering::SeqCst);
+
+    if own_context_id == target_context_id {
+        tracing::trace!("Switching context {own_context_id} to itself, which is a no-op");
+        return Err(Ok(Errno::Success));
+    }
+
+    let (sender, wait_for_unblock) = oneshot::channel::<Result<(), RuntimeError>>();
+
+    let maybe_old_value = data.contexts.insert(own_context_id, sender);
```
Arshia001: I'd call that field `data.waiting_contexts` or similar if it's the intention that it should hold pending contexts only.

Arshia001:
LGTM overall. Looks like it should be easy to update this PR with the new async API once that hits, too.

lib/wasix/src/syscalls/wasix/context_delete.rs
zebreus: if the function causes a trap (that is not handled by a host function), the main context will be resumed, and the trap will be raised there.

This allows us to do stuff like exiting in contexts.

lib/wasix/src/syscalls/wasix/context_destroy.rs
zebreus: Yes, I am still working on what exactly to trap with in that case. `context_delete` should be fine as it is right now, as the oneshot::channel futures should resolve to canceled if the sender is dropped.

lib/wasix/src/syscalls/wasix/context_destroy.rs
zebreus: the linter wanted doc comments ^^. Proper documentation for the functions is maintained in the WARP and will be copied to all relevant places when the PR is ready

lib/wasix/src/state/env.rs
zebreus: Yeah, I think this is not `Clone`, but more like a _use the state of this WasiEnv as a template for a new one_. But I dont have any ideas for a descriptive name right now.

lib/wasix/src/syscalls/wasix/context_new.rs
zebreus: Usually, if a context traps, it will either handle it itself or it should take the quickest path to an exit that leads through the main context. The only alternative is to exit directly, although having it well-defined as going through main might be a bit more flexible.

lib/wasix/src/syscalls/wasix/context_switch.rs
zebreus: fair, probably be `blocked` or `suspended` something those are the words I used most often. The name `contexts` is a leftover from an earlier revision, where they were way more stateful.

lib/wasix/src/syscalls/wasix/context_delete.rs
zebreus: Fair

zebreus: Made it relaxed in all places

lib/wasix/src/syscalls/wasix/context_switch.rs
zebreus: Because otherwise we would have to put it into a mutex/rwlock to be able to pass a mutable version into the future

zebreus: which might actually be more appropriate

lib/wasix/src/state/env.rs
zebreus: Fixed it, now cloned/newly created envs always start with empty contexts

lib/wasix/src/state/env.rs
zebreus: I changed it so contexts always gets initialized with None

lib/wasix/src/syscalls/wasix/context_destroy.rs
zebreus: Added a short description plus a reference to the C header, which contains authoritative docs. That header is not in wasix-libc main yet; it's also in the WARP and this PR's description.
If you have a suggestion for a better place for the docs, please feel free to open a new comment.

lib/wasix/src/syscalls/wasix/context_destroy.rs
zebreus: This should be fine. As soon as the Sender is dropped, the corresponding context can be resumed by the executor and delete/cancel itself gracefully.

lib/wasix/src/syscalls/wasix/context_new.rs
zebreus: I know. However, I'll leave this comment open for now as a reminder to do a final review of the panicking behavior before merging this PR.

lib/wasix/src/syscalls/wasix/context_new.rs
zebreus: We have this well defined as the main context. Doing parent context logic becomes really complex, when you consider scenarios where the parent has already been destroyed. In which case we would have to fall back to main anyway.

I left the exact behaviour explicitly undefined in the documentation, so we can change this at a later point if required.

lib/wasix/src/syscalls/wasix/context_switch.rs
zebreus: An Atomic is the easiest way to get interior mutability here. Thinking about it, I don't know if we even need interior mutability for `next_available_context_id`. For `current_context_id` and `unblockers`, we need them, as we want to pass an `Arc<ContextSwitchingEnvironment>` to the new context, and it needs to mutate those. However, `next_available_context_id` can probably be just a normal integer, as we always have a mutable ref to the env when editing it.

zebreus: We also need interior mutability for `next_available_context_id` as it's stored in the inner Arc, and we can't get a mutable ref to the contents of an arc. We could move that field out of inner into ContextSwitchingEnvironment directly, but I don't think thats an improvement.

lib/wasix/src/syscalls/wasix/context_switch.rs
zebreus: I renamed the field `unblockers` as it contains the things that can unblock suspended contexts

lib/wasix/src/syscalls/wasix/context_switch.rs
```diff
@@ -0,0 +1,68 @@
...
+) -> Result<Errno, RuntimeError> {
+    // TODO: Should we call do_pending_operations here?
+    // match WasiEnv::do_pending_operations(&mut ctx) {
+    //     Ok(()) => {}
+    //     Err(e) => {
+    //         return Now(Err(RuntimeError::user(Box::new(e))));
+    //     }
+    // }
+
+    let mut write_lock = ctx.write().await;
```
zebreus, line 21: @Arshia001 We don't `do_pending_operations` here because we can't do that with an `AsyncFunctionEnvMut`, and there is no way to get a normal `FunctionEnvMut` from it.

zebreus:
This needs a test that verifies that returning from a context entrypoint exits with 129

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -289,85 +288,84 @@ fn call_module(
...
+        ContextSwitchingEnvironment::run_main_context(&ctx, store, start.clone(), vec![]);
 
-                // An error was encountered when restoring from the vfork, report it
-                Err(e) => {
-                    call_ret = Err(RuntimeError::user(Box::new(WasiError::Exit(e.into()))));
-                    break;
-                }
+    loop {
+        // Technically, it's an error for a vfork to return from main, but anyway...
+        match resume_vfork(&ctx, &mut store, &start, &call_ret) {
```
Arshia001: Note: resume_vfork calls main synchronously, which it shouldn't anymore. Specifically, it's not an error to _vfork_ in the presence of coroutines either, it's just fork that shouldn't happen.

lib/wasix/src/state/env.rs
```diff
@@ -179,6 +177,13 @@ pub struct WasiEnv {
     ///  not be cloned when `WasiEnv` is cloned)
     /// TODO: We should move this outside of `WasiEnv` with some refactoring
     inner: WasiInstanceHandlesPointer,
+
+    /// Tracks the active contexts of the WASIX context switching API
+    ///
+    /// This is `None` when the main function was not launched with context switching
```
Arshia001, line 183: But how can this happen? Should it be allowed at all? I don't think we should allow it.

lib/wasix/src/syscalls/wasix/context_create.rs
```diff
@@ -0,0 +1,96 @@
+use crate::{WasiEnv, WasiError};
+use tracing::instrument;
+use wasmer::{
+    AsStoreRef, Function, FunctionEnvMut, MemorySize, RuntimeError, StoreMut, TypedFunction, Value,
+    WasmPtr,
+};
+use wasmer_wasix_types::wasi::Errno;
+
+/// Return the function corresponding to the given entrypoint index if it exists and has the signature `() -> ()`
+pub fn lookup_typechecked_entrypoint(
```
Arshia001, line 11: When we implement JSPI-based async support in the JS backend, this will be a problem. For now, we can keep it though.

Arshia001: (problem because type checking)

lib/wasix/src/syscalls/wasix/context_create.rs
```diff
@@ -0,0 +1,96 @@
...
+                "Failed to lookup entrypoint function {}: {:?}",
+                entrypoint_id,
+                e
+            );
+            return Err(Errno::Inval);
+        }
+    };
+
+    let entrypoint_type = entrypoint.ty(&store);
+    let Ok(entrypoint) = entrypoint.typed::<(), ()>(&store) else {
```
Arshia001: Why `() -> ()`? Don't we pass give the new context its ID or something?

lib/wasix/src/syscalls/wasix/context_create.rs
```diff
@@ -0,0 +1,96 @@
...
+    mut ctx: FunctionEnvMut<'_, WasiEnv>,
+    new_context_ptr: WasmPtr<u64, M>,
+    entrypoint: u32,
+) -> Result<Errno, WasiError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
+    // Verify that we are in an async context
+    // We need to do this first, before we borrow the store mutably
+    let Some(async_store) = ctx.as_store_async() else {
+        tracing::warn!(
```
Arshia001, line 68: This message mentions engine support, but we don't even expose this function to modules if the engine doesn't support async. This can happen if we launch the module synchronously which, as stated elsewhere, I don't think we should ever do after this change.

lib/wasix/src/syscalls/wasix/context_create.rs
```diff
@@ -0,0 +1,96 @@
...
+        );
+        return Ok(Errno::Again);
+    };
+
+    let (data, mut store) = ctx.data_and_store_mut();
+
+    // Get the context-switching environment
+    let Some(environment) = &data.context_switching_environment else {
+        tracing::warn!(
+            "The WASIX context-switching API is only available in engines supporting async execution"
```
Arshia001: again, mention of engine

lib/wasix/src/syscalls/wasix/context_create.rs
```diff
@@ -0,0 +1,96 @@
...
+        );
+        return Ok(Errno::Again);
```
Arshia001: Why `Again`? I'd go with maybe `Notsup` here. Calling this function again can't produce different results.

lib/wasix/src/syscalls/wasix/context_destroy.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    let env = ctx.data();
+    let memory: MemoryView<'_> = unsafe { env.memory_view(&ctx) };
+
+    // Verify that we are in an async context
+    let environment = match &env.context_switching_environment {
+        Some(c) => c,
+        None => {
+            tracing::warn!(
+                "The WASIX context-switching API is only available in engines supporting async execution"
```
Arshia001: engine

lib/wasix/src/syscalls/wasix/context_destroy.rs
```diff
@@ -0,0 +1,71 @@
...
+            );
+            return Ok(Errno::Again);
```
Arshia001: again -> notsup

lib/wasix/src/syscalls/wasix/context_destroy.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    if target_context_id == main_context_id {
+        tracing::trace!(
+            "Context {} tried to delete the main context, which is not allowed",
+            own_context_id
+        );
+        return Ok(Errno::Inval);
+    }
+
+    let removed_unblocker = environment.remove_unblocker(&target_context_id);
```
Arshia001: Should we make this function async and yield here so the context gets a chance to clean up immediately? Otherwise, the context could stay around in memory until we exit (think doing 5 `context_destroy` calls from the main context and just happily continuing to run for another hour without doing any more context switching)

Arshia001: Alternatively, should we wait for the context to report it's now gone? as in, create a new channel, put the sender in the context switching env, and the context can use the sender to report once it's actually gone

lib/wasix/src/syscalls/wasix/context_switch.rs
Arshia001: Now we can do that, and I believe we should

Arshia001: oh wait, we do, nvm

lib/wasix/src/syscalls/wasix/context_switch.rs
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    let data = write_lock.data_mut();
+
+    // Verify that we are in an async context
+    let environment = match &data.context_switching_environment {
+        Some(c) => c,
+        None => {
+            tracing::trace!("Context switching is not enabled");
```
Arshia001: we can never get here, this should be a panic

lib/wasix/src/syscalls/wasix/context_switch.rs
```diff
@@ -0,0 +1,92 @@
...
+    mut ctx: FunctionEnvMut<'_, WasiEnv>,
+    _target_context_id: u64,
+) -> Result<Errno, WasiError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
+    tracing::warn!(
+        "The WASIX context-switching API is only available in engines supporting async execution"
+    );
+    // Indicate that no context-switching environment is available
+    Ok(Errno::Again)
```
Arshia001: again -> notsup

lib/wasix/src/utils/thread_local_executor.rs
```diff
@@ -0,0 +1,96 @@
+use futures::{
```
Arshia001, line 1: Why do we need this module at all? It looks like a thin wrapper that doesn't do anything extra.

lib/wasix/tests/context_switching.rs
```diff
@@ -0,0 +1,264 @@
...
+    let runner = WasiRunner::new();
+    runner.run_wasm(
+        RuntimeOrEngine::Engine(engine),
+        "wasix-test",
+        module,
+        ModuleHash::random(),
+    )
+}
+
+#[cfg(target_os = "linux")]
```
Arshia001: Why just linux? These should work on all platforms

lib/wasix/src/state/context_switching.rs
```diff
@@ -0,0 +1,453 @@
...
+};
+use thiserror::Error;
+use tracing::trace;
+use wasmer::{RuntimeError, Store};
+use wasmer_wasix_types::wasi::ExitCode;
+
+/// The context-switching environment represents all state for WASIX context-switching
+/// on a single host thread.
+#[derive(Debug)]
+pub(crate) struct ContextSwitchingEnvironment {
```
Arshia001, line 27: This lives within the function environment, which is already a single-thread entity. There's no need to use an Arc or RwLocks here.

lib/wasix/src/state/context_switching.rs
```diff
@@ -0,0 +1,453 @@
...
+    inner: Arc<ContextSwitchingEnvironmentInner>,
+}
+
+#[derive(Debug)]
+struct ContextSwitchingEnvironmentInner {
+    /// List of the unblockers for all suspended contexts
+    unblockers: RwLock<BTreeMap<u64, Sender<Result<(), RuntimeError>>>>,
+    /// The ID of the currently active context
+    current_context_id: AtomicU64,
```
Arshia001, line 39: ... or atomics for that matter.

lib/wasix/src/state/context_switching.rs
```diff
@@ -0,0 +1,453 @@
...
+/// Errors that can occur during a context switch
+#[derive(Debug, Error)]
+pub enum ContextSwitchError {
+    #[error("Target context to switch to is missing")]
+    SwitchTargetMissing,
+}
+
+const MAIN_CONTEXT_ID: u64 = 0;
+
+/// Contexts will trap with this error as a RuntimeError::user when they are canceled
```
Arshia001, line 56: cancelled*

lib/wasix/src/state/context_switching.rs
```diff
@@ -0,0 +1,453 @@
...
+/// When it bubbles up to the start of the entrypoint function of a context, it will be
+/// handled by just letting the context exit silently.
+#[derive(Error, Debug)]
+#[error("Context was canceled")]
+pub struct ContextCanceled(());
+impl ContextCanceled {
+    /// Defuse the ContextCanceled so it does not panic when dropped
+    pub fn defuse(self) {
+        // Consume self without panicking
+        forget(self);
```
Arshia001, line 73: A boolean would have been simpler here :D

lib/wasix/src/state/context_switching.rs
```diff
@@ -0,0 +1,453 @@
...
+/// on a single host thread.
+#[derive(Debug)]
+pub(crate) struct ContextSwitchingEnvironment {
+    inner: Arc<ContextSwitchingEnvironmentInner>,
+}
+
+#[derive(Debug)]
+struct ContextSwitchingEnvironmentInner {
+    /// List of the unblockers for all suspended contexts
+    unblockers: RwLock<BTreeMap<u64, Sender<Result<(), RuntimeError>>>>,
```
Arshia001, line 37: Why not HashMap?

lib/wasix/src/state/context_switching.rs
```diff
@@ -0,0 +1,453 @@
...
+    pub(crate) fn active_context_id(&self) -> u64 {
+        self.inner.current_context_id.load(Ordering::Relaxed)
+    }
+
+    /// Get the id of the main context (0)
+    pub(crate) fn main_context_id(&self) -> u64 {
+        MAIN_CONTEXT_ID
+    }
+
+    pub(crate) fn remove_unblocker(&self, target_context_id: &u64) -> bool {
```
Arshia001: I'd call this function `abort` or `cancel` or similar, `remove_unblocker` doesn't really explain the _why_, only the _what_, which requires knowledge of this module to understand.

lib/wasix/src/state/context_switching.rs
```diff
@@ -0,0 +1,453 @@
...
+    > {
+        let (own_unblocker, wait_for_unblock) = oneshot::channel::<Result<(), RuntimeError>>();
+        let wait_for_unblock = wait_for_unblock.map_err(|_| ContextCanceled(()));
+
+        // Lock contexts for this block
+        let mut unblockers = self.inner.unblockers.write().unwrap();
+        let own_context_id = self.active_context_id();
+
+        // Assert preconditions (target is blocked && we are unblocked)
+        if unblockers.get(&target_context_id).is_none() {
```
Arshia001: double map lookup, we can just `.remove()` here and keep the value for later

lib/wasix/src/state/context_switching.rs
```diff
@@ -0,0 +1,453 @@
...
+                    // When our context was canceled return the `ContextCanceled` error.
+                    // It will be handled by the entrypoint wrapper and the context will exit silently.
+                    //
+                    // If we reach this point, we must try to restore our context ID as it will not be read again
+                    return Err(RuntimeError::user(canceled.into()));
+                }
+            };
+
+            // Restore our own context ID
+            let Some(inner) = Weak::upgrade(&weak_inner) else {
```
Arshia001, line 287: Once the Arc and atomics and RwLock are gone, we can just give this closure a copy of the `AsyncFunctionEnvMut` (which is cloneable) and it can lock that and gain access to the context switching environment.

lib/wasix/src/state/context_switching.rs
zebreus: BTreeMap is usually more compact for small data sets with simple keys, but that doesn't really matter here. I mostly choose it because BtreeMap is my go-to map.

lib/wasix/src/state/context_switching.rs
zebreus: Renamed to `destroy_context`, `switch_context`, and `create_context`

lib/wasix/src/state/context_switching.rs
zebreus: I originally did it that way to split the precondition checks from the actual action, before it was clear that the other check should result in a panic.

Changed it to only have one lookup

lib/wasix/src/state/context_switching.rs
zebreus: afaik, this is the normal way to implement stuff like this. And it doesn't incur runtime overhead.

lib/wasix/src/state/context_switching.rs
zebreus: `canceled` is the American spelling, `cancelled` is the British spelling. `futures` uses the american spelling, so I also used that

lib/wasix/tests/context_switching.rs
zebreus: @Arshia001  Yes, but these tests depend on wasixcc and we only offer a linux build atm

lib/wasix/src/utils/thread_local_executor.rs
zebreus: This module provides a lightweight thread-local executor with a `Send`+`Sync` spawner.

The `context_switching` module is already quite big and hard to understand. The logic here is independent of that and is not concerned with context switching. To understand the context switching, you do not need to know how we make the spawner `Send`+`Sync` safely.

lib/wasix/src/syscalls/wasix/context_switch.rs
zebreus: will change that

lib/wasix/src/syscalls/wasix/context_create.rs
zebreus: Will change that

lib/wasix/src/syscalls/wasix/context_create.rs
zebreus: No, because we need the new context ID immediately to return it from the function. As this API only provides a minimal interface for context switching, the guest code can pass that ID to the new context in some other way if it wants to.

lib/wasix/src/syscalls/wasix/context_create.rs
zebreus: agree

lib/wasix/src/state/env.rs
zebreus: It can happen if the engine does not support async

lib/wasix/src/syscalls/wasix/context_destroy.rs
zebreus: I considered that, but it would make the implementation more complex than it would need to be. This way, it is easy to understand and review. For now, switch is the only function that causes other contexts to run. The logic behind switching is really peculiar. I don't want to maintain a second, similar but slightly different version of that.

If we need immediate cleanup, we can implement it later, since it does not break the API.

lib/wasix/src/syscalls/wasix/context_destroy.rs
zebreus: yes

lib/wasix/src/state/context_switching.rs
zebreus: The function environment needs to be `Send`+`Sync` for some reason, unless your PR changed that.

lib/wasix/src/syscalls/wasix/context_switch.rs
zebreus: We call some functions synchronously even if the engine supports async. For example, the `_initialize` for the main module will not be launched in an async context. Since static initializers are executed in `_initialize`, users might choose to call `context` functions there, which is an error but not a panic.

This behaviour is also documented in the API documentation in the C header

zebreus: The EAGAIN errors also stem from this consideration

zebreus: Changed the trace to a warn and changed the message to `The WASIX context-switching API is only available after entering the main function`.
I choose not to mention static initializers in the message, because there might be other ways to trigger it.

lib/wasix/src/syscalls/wasix/context_create.rs
zebreus: We expose the context functions even if the engine does not support async. They fail in a well-defined manner (see API documentation in PR description/C header) in that situation. We do this, as they are exposed by wasix-libc, which means they need to be available to run any executable linked against wasix-libc.

In practice, this allows us to use the same python binary for all targets.

zebreus: Also we can get into this path even if our engine supports async, during calls to `_initialize`. See my more detailed response to your comment in context_switch

lib/wasix/src/state/env.rs
zebreus: Or before the main module has finished initialization

Arshia001: hmmmm.... I don't like the fact that we're replacing the concurrent execution context each time. This breaks vforks, since a vfork can technically happen while other coroutines are active and there's nothing that prevents it from doing that. However, as a temporary work-around, and since we're not mixing vforks (which are outdated anyway) and context switching in anything yet, we can just make WASIX exit with an error if vforks + active contexts happen.

lib/wasix/src/syscalls/wasix/context_create.rs
Arshia001: Ah right, we only start the context after the first switch. Good design!

lib/wasix/src/syscalls/wasix/context_create.rs
Arshia001: > We expose the context functions

you're replacing these syscalls with versions that just return an error if the engine doesn't support contexts at all, so we can never be here with the wrong engine. Let's change the message to something that mentions the actual circumstances where this can happen.

Also, `_initialize` doesn't exist in any WASIX binaries; that was a mistake on our part to include calls to it. `_initialize` is only exported from WASI reactors, and we don't support reactors at all.

lib/wasix/src/syscalls/wasix/context_create.rs
Arshia001: 👍🏻

lib/wasix/src/syscalls/wasix/context_destroy.rs
Arshia001: OK, but let's add a TODO and an issue so we can add this in later on, as it can have serious memory implications; basically, the last context you destroy (at least) will always stay in memory (and waste it) until the module exits.

lib/wasix/src/syscalls/wasix/context_switch.rs
Arshia001: > Since static initializers are executed in _initialize

they are not; static initializers run as part of `_start`, which is already in an async context. `_initialize` never exists. The `(start)` function only does memory relocations.

lib/wasix/src/utils/thread_local_executor.rs
Arshia001: Oh right, we need function envs to be `Send + Sync`... even though they can't be sent across threads anyway. Ugh,

lib/wasix/tests/context_switching.rs
Arshia001: riiiiiiiiiiight..... let's leave a TODO here so we can enable them for more platforms once wasixcc has support. Since we now have the LLVM CI builds, it should be rather trivial to support other platforms there.

lib/wasix/src/state/context_switching.rs
Arshia001: `Send + Sync`, yes, but that's different from needing an `Arc` or atomics, which are primitives for safe multi-threaded access without a mutable ref. We already have mutable refs to the function environment. The data in an `Arc` already has to be `Send + Sync`, so putting it outside of an Arc should work as well. Or maybe I'm missing something?

Also, atomics are entirely unnecessary here.

lib/wasix/src/state/context_switching.rs
Arshia001: Dammit XD

lib/wasix/src/state/context_switching.rs
Arshia001: From what I've seen, everybody generally uses HashMaps unless they need ordered access to data... @theduke would have you use an `FnvHashMap` or `AHashMap` here.

lib/wasix/src/syscalls/wasix/context_create.rs
zebreus: Even if it's not `_initialize`, people can try to call into these function during other linker functions (like relocation), where we can not provide an async runtime, unless we refactor the whole linker to support working with an owned Store. Also it is unset during a asyncify vfork.

I think the current implementation is fine, we don't need to make this more complicated.

lib/wasix/src/state/env.rs
zebreus: Sounds good, a panic will solve the issue.

zebreus: Also something else is that while the execution context is set we can also rely on having an async store.

zebreus: vforks with coroutines is as tricky as I am not yet sure how they should interact. I added assertions that prevent vforks in a context-switching environment and prevent entering a context-switching environment when a vfork is present.

I think the correct interaction would be to add logic to clone all active contexts as well, but that is a ton of work for a feature we don't need right now.

An alternative could be to have a vfork start with a new context-switching environment as the main context. That's not really consistent with how vfork should behave (exact copy, yadayada), but it should cover most cases.

We could also do the above, but allow vfork only if there is only one (main) context present. This will let us decide which way to go in the future without breaking existing behaviour.

zebreus: This will happen during vforks too. The logic is quite simple:
 context_switching_env exists -> context switching is available

lib/wasix/src/syscalls/wasix/context_new.rs
zebreus: I reviewed it again. Panics now occur only when something in WASIX goes wrong internally. Guest code should not be able to trigger these actively.

lib/wasix/src/bin_factory/exec.rs
zebreus: `resume_vfork` will now exit if a context-switching environment was present, as we can't reenter the previous environment.

lib/wasix/src/state/context_switching.rs
zebreus: The `Arc` allows us to pass it into the async closures as a weak reference. The simple data structure makes it easier to reason about that and to be sure we don't run into memory leaks cause by cyclic ownership of reference counted pointers.

lib/wasix/src/state/context_switching.rs
zebreus: See the other comment. The simpler data structure was chosen to make it easier to reason about cyclic ownership here.

lib/wasix/tests/context_switching.rs
zebreus: and the current wasixcc build does not include the matching libc. I will be disabling the tests for now and open an issue to reenable them at a later point

lib/wasix/tests/context_switching.rs
```diff
@@ -41,223 +41,223 @@ fn test_with_wasixcc(name: &str) -> Result<(), anyhow::Error> {
     )
 }
 
-#[cfg(target_os = "linux")]
+#[cfg(any())]
```
Arshia001: ?

tests/wasix/vfork/run.sh
```diff
@@ -10,4 +10,5 @@ $WASMER -q run main.wasm --dir . -- cloexec
 $WASMER -q run main.wasm --dir . -- exiting_child
 $WASMER -q run main.wasm --dir . -- trapping_child
 $WASMER -q run main.wasm --dir . -- exit_before_exec
-$WASMER -q run main.wasm --dir . -- trap_before_exec
+# This test is expected to fail with a non-zero exit code
+! $WASMER -q run main.wasm --dir . -- trap_before_exec
```
Arshia001: Why this change?

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -442,6 +442,14 @@ fn resume_vfork(
         // Terminate the child process
         child_env.process.terminate(code);
 
+        // If the vfork contained a context-switching environment, exit now
+        if ctx.data(store).context_switching_environment.is_some() {
+            tracing::error!(
+                "Terminated a vfork in another way than exit or exec which is undefined behaviour. In this case the parent parent process will be terminated."
```
Arshia001, line 454: This IS the exit() case of vfork.

lib/wasix/src/state/context_switching.rs
```diff
@@ -171,9 +157,24 @@ impl ContextSwitchingEnvironment {
...
+            if env
+                .vfork
+                .as_ref()
+                .and_then(|vfork| vfork.env.context_switching_environment.as_ref())
+                .is_some()
+            {
+                // Grace for vforks, so they don't bring everything down with them.
+                // This is still an error.
+                tracing::error!(
+                    "Failed to remove wasix context-switching environment from WASIX env after main context finished, this means you triggered undefined behaviour"
```
Arshia001: What undefined behavior?

lib/wasix/src/syscalls/wasix/proc_fork.rs
```diff
@@ -32,10 +32,16 @@ pub fn proc_fork<M: MemorySize>(
         Errno::Notsup
     }));
 
-    wasi_try_ok!(ctx.data().context_switching_environment.as_ref().ok_or_else(|| {
-        warn!("process forking is mutally exclusive with WASIX context-switching features. If you need both, please open an issue.");
-        Errno::Notsup
-    }));
+    if let Some(context_switching_environment) = ctx.data().context_switching_environment.as_ref() {
+        if context_switching_environment.active_context_id()
+            != context_switching_environment.main_context_id()
```
Arshia001, line 37: This still doesn't check for other coroutines.

lib/wasix/src/syscalls/wasix/proc_fork.rs
```diff
@@ -280,7 +281,7 @@ fn run<M: MemorySize>(
             .start
             .clone()
             .unwrap();
-        start.call(&mut store)
+        ContextSwitchingEnvironment::run_main_context(&ctx, store, start.into(), vec![])
```
Arshia001, line 295: It is SUPER-DUPER important we get processes right; for example, spawning python from bash will go through this code path. Do we have tests verifying this does the right thing? If not, we should add them.

lib/wasix/src/syscalls/wasix/proc_fork.rs
```diff
@@ -31,6 +31,11 @@ pub fn proc_fork<M: MemorySize>(
         Errno::Notsup
     }));
 
+    wasi_try_ok!(ctx.data().context_switching_environment.as_ref().ok_or_else(|| {
```
Arshia001: I'm fairly sure this logic is wrong (but I'm going commit-by-commit, maybe it was changed?)

Arshia001 (changes requested):

lib/wasix/tests/context_switching.rs
syrusakbary: I believe Lennart mentioned he commented out the tests until wasixcc had the changes merged (we will re-enable them after, I assume?)

lib/wasix/src/syscalls/wasix/proc_fork.rs
syrusakbary: Somehow you commented on an outdated commit @Arshia001 ? We have this logic now: https://github.com/wasmerio/wasmer/pull/5881/files#diff-7c7ce57c39d00b24cf8ad615c058ea452b4914fb93615cfd42070d279a8552eeR35

Arshia001: OK, I'll resolve this, but the other one is also wrong afaict.

lib/wasix/tests/context_switching.rs
Arshia001: Question is what change is needed in wasixcc? I'd expect the EH configuration should just work for context switching as well.

lib/wasix/src/syscalls/wasix/proc_fork.rs
zebreus: I added a test for that, and it passes. Also, `run_main_context` is defensive about its precondition that no previously existing context-switching env exists, so we would at least get a helpful error if we had a significant bug.

lib/wasix/src/state/context_switching.rs
zebreus: The parent process will exit.  This is documented in our documentation

zebreus: Also it is not that this code path triggers the undefined behaviour, but that you can only reach this path if you did stuff you were not allowed to do before (concretely returning from a vfork/trapping in a vfork before calling `exec()` or `exit()`).

lib/wasix/src/bin_factory/exec.rs
zebreus: No, it's not.

zebreus: This is the _you terminated the vfork in any other way than explicitly calling `exit()` or an `exec()` function_ case. Calling `_exit` (aka `proc_exit`) normally cleans up the vfork and returns execution to the parent.

Note that it is only allowed to call the `exit()` function in a fork, but not to do other things that cause an exit.

tests/wasix/vfork/run.sh
zebreus: Because trapping in a vfork before calling exec is totally undefined behaviour, we can't easily do that with an async store without major restructuring.

If that is breaking something, then we can add it back, but no one should rely on that working.

lib/wasix/tests/context_switching.rs
zebreus: Mainly we needed the new release of wasix-libc to make our CI pass

lib/wasix/src/syscalls/wasix/proc_fork.rs
zebreus: It is well-defined behaviour to have other contexts open at that point. I added a test for that.

The main restriction is that it is only allowed to fork on the main context. In the case of a fork, the child process enters a new context-switching environment that only has a main context.

In the case of a vfork, the vforked child process is not in a context-switching environment, so it can't do weird stuff. However, calling an `exec` function (not only in vfork) enters a new context-switching environment. The alternative to calling `exec` is to `_exit`. In both cases, the parent gets its original context-switching environment back.

All this behaviour is documented in the documentation (in `context.h`)

lib/types/src/module.rs
```diff
@@ -564,7 +564,14 @@ impl ModuleInfo {
     pub fn name(&self) -> String {
         match self.name {
             Some(ref name) => name.to_string(),
-            None => "<module>".to_string(),
+            None => {
+                format!(
+                    "<module ({} imported globals) ({} imported functions) ({} functions)>",
```
Arshia001: Undo this

lib/wasix/src/state/context_switching.rs
Arshia001: Let's leave a TODO here so we can eliminate the Arc later on.

lib/wasix/src/syscalls/wasix/context_create.rs
Arshia001: Mention context switching env not being present instead of the engine

Arshia001: NVM

lib/wasix/src/syscalls/wasix/context_create.rs
Arshia001: Here, mention context switching env not being present instead of engines

lib/wasix/src/syscalls/wasix/context_destroy.rs
Arshia001: here too

lib/wasix/src/state/context_switching.rs
```diff
@@ -0,0 +1,476 @@
...
+    pub(crate) fn active_context_id(&self) -> u64 {
+        self.inner.current_context_id.load(Ordering::Relaxed)
+    }
+
+    /// Get the id of the main context (0)
+    pub(crate) fn main_context_id(&self) -> u64 {
+        MAIN_CONTEXT_ID
+    }
+
+    pub(crate) fn destroy_context(&self, target_context_id: &u64) -> bool {
```
Arshia001, line 201: Add a TODO specifying that contexts aren't properly deallocated until the next switch operation and that we should fix this because it's a waste of memory

lib/wasix/src/syscalls/wasix/context_switch.rs
Arshia001: Actually, this is all true only of modules that were built by clang/LLVM, and modules can do weird things otherwise, so let's not make it a trap.

Arshia001:
Good to merge from my side as soon as a few more TODOs are added in the code. The existing functionality should work correctly, with a few minor performance caveats.

lib/wasix/src/state/context_switching.rs
zebreus: added a detailed TODO comment to `destroy_context`

lib/types/src/module.rs
zebreus: I cherry-picked the commit from #5946 when debugging and forgot to remove it again. Removed it now.

lib/wasix/src/syscalls/wasix/context_create.rs
zebreus: fixed

lib/wasix/src/syscalls/wasix/context_destroy.rs
zebreus: added a detailed explanation of the current situation + a TODO in destroy_context.

lib/wasix/src/state/context_switching.rs
zebreus: Added a TODO

PR #5889, opened 2025-11-19
build: switch to LLVM 21

marxin:
I successfully bumped the LLVM compiler in the `compiler-llvm` crate from version **18** to **21**. I made the compilation time comparison for `php` and `python`, and the build time stays the same in between the versions.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -101,10 +101,7 @@ allow = [
 	"MPL-2.0",
 	"0BSD",
 	"CC0-1.0",
-	"OpenSSL",
 	"BSD-3-Clause",
-	"Unicode-DFS-2016",
-	"Unicode-3.0",
 	# BOOST source license
 	# NOTE: not to be confused with business source license!
 	"BSL-1.0",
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -985,7 +984,7 @@ jobs:
         run: |
           ./scripts/alpine-linux-install-deps.sh
           echo "ENABLE_LLVM=0" >> $GITHUB_ENV
-          echo "LLVM_CONFIG_PATH=/usr/bin/llvm-config-18" >> $GITHUB_ENV
+          echo "LLVM_CONFIG_PATH=/usr/bin/llvm-config-20" >> $GITHUB_ENV
```
theduke, line 987: Is this not supposed to be `21`? Or is the binary still called `20`?

flake.nix
```diff
@@ -73,22 +73,22 @@
           ];
 
           shellHook = ''
-            export LLVM_SYS_181_PREFIX="${pkgs.llvmPackages_18.llvm.dev}"
```
theduke, line 76: Tested , works properly, thanks for updating!

.github/workflows/test.yaml
marxin: The latest alpine release does not provide version `21`, that's why we're sticking to `20` here.

theduke: Is it wise to split to two different versions?
Can that lead to potential issues or limitations?

I believe Alpine is  used for the `musl` build, so I guess we would still catch problems on CI .

marxin: I see it valuable to test the latest 2 LLVM releases in CI, should uncover potential issues.
Frankly speaking, with the latest Inkwell 0.7.0, I was able to build and test `wasmer` with all versions from 18-21.

theduke: Sound good to me!

Issue #5877, opened 2025-11-14
cargo r --features=llvm -p wasmer-cli run --disable-cache cowsay aa - crashes occasionally


marxin:
If run the command from the current `main` branch, sometimes I hit the following segfault:

```
$ cargo r --features=llvm -p wasmer-cli run --disable-cache cowsay aa
...
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007f45e62989d3 in __pthread_kill_internal (threadid=<optimized out>, signo=6) at pthread_kill.c:89
#2  0x00007f45e623e3a0 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f45e622557a in __GI_abort () at abort.c:77
#4  0x00007f45e65a211e in __deregister_frame_info_bases (begin=<optimized out>) at /usr/src/debug/gcc/gcc/libgcc/unwind-dw2-fde.c:296
#5  __deregister_frame_info_bases (begin=<optimized out>) at /usr/src/debug/gcc/gcc/libgcc/unwind-dw2-fde.c:230
#6  0x00007f45e65c308a in __deregister_frame (begin=<optimized out>) at /usr/src/debug/gcc/gcc/libgcc/unwind-dw2-fde.c:311
#7  0x0000558cf6957081 in wasmer_compiler::engine::unwind::systemv::{impl#1}::drop (self=0x7f456409d6e0) at lib/compiler/src/engine/unwind/systemv.rs:261
#8  0x0000558cf694ff17 in core::ptr::drop_in_place<wasmer_compiler::engine::unwind::systemv::UnwindRegistry> ()
    at /home/marxin/.rustup/toolchains/1.88-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:524
#9  0x0000558cf694f36f in core::ptr::drop_in_place<wasmer_compiler::engine::code_memory::CodeMemory> ()
    at /home/marxin/.rustup/toolchains/1.88-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:524
#10 0x0000558cf695006d in core::ptr::drop_in_place<[wasmer_compiler::engine::code_memory::CodeMemory]> ()
[...]

PR #5912, opened 2025-11-26
Prevent crashes during exit due to deregistering frames on exit (v2)

marxin:
Second attempt suggested by @zebreus. Tested that locally and works reliably it seems.

Fixes: #5877

### DIFF
--- a/lib/compiler/src/engine/artifact.rs
+++ b/lib/compiler/src/engine/artifact.rs
@@ -132,6 +132,7 @@ impl Artifact {
         tunables: &dyn Tunables,
         hash_algorithm: Option<HashAlgorithm>,
     ) -> Result<Self, CompileError> {
+        let data = Box::leak(Box::from(data));
         let mut inner_engine = engine.inner_mut();
         let environ = ModuleEnvironment::new();
         let translation = environ.translate(data).map_err(CompileError::Wasm)?;
@@ -1305,8 +1306,6 @@ impl Artifact {
 
 impl Drop for Artifact {
     fn drop(&mut self) {
-        eprintln!("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Pausing before dropping Artifact");
-        sleep(Duration::from_millis(10));
         eprintln!("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Dropping Artifact");
     }
 }


lib/compiler/src/engine/unwind/systemv.rs
```diff
@@ -237,8 +247,21 @@ impl UnwindRegistry {
...
+    EXIT_CALLED.store(true, Ordering::SeqCst);
+}
+
 impl Drop for UnwindRegistry {
     fn drop(&mut self) {
+        // We don't want to deregister frames in UnwindRegistry::Drop as that could be called during
+        // program shutdown and can collide with release_registered_frames and lead to
+        // crashes.
+        if EXIT_CALLED.load(Ordering::SeqCst) {
+            return;
```
zebreus: This should be checked before every frame deregistration (even that it is still prone to race conditions but the best we can do).
The toy example does not have that many functions, but something like Python has ~20000 functions iirc, so we need to process a ton of deregistrations.

lib/compiler/src/engine/unwind/systemv.rs
zebreus, line 279: I am pretty sure this is dead code, as we never make push anything into `self.registrations` when we are on macos aarch64

zebreus: @syrus I think the correct fix would be to move the call to `self.compact_unwind_mgr.deregister();` outside of the loop (potentially also add the registrations to `self.registrations` on macOS aarch64). However, I can't really test this as I don't have a Mac.

lib/compiler/src/engine/unwind/systemv.rs
```diff
@@ -139,6 +143,12 @@ impl UnwindRegistry {
 
     #[allow(clippy::cast_ptr_alignment)]
     unsafe fn register_frames(&mut self, eh_frame: &[u8]) {
+        // Register atexit handler that will tell us if exit has been called.
+        static INIT: Once = Once::new();
+        INIT.call_once(|| unsafe {
+            libc::atexit(atexit_handler);
+        });
```
zebreus, line 155: `atexit` should be available on all platforms so this is fine

lib/compiler/src/engine/unwind/systemv.rs
marxin: Fair point. I would like to handle this in a separate PR as it's pretty unrelated to it.
@zebreus Does it work for you?

lib/compiler/src/engine/unwind/systemv.rs
marxin: Good! Hopefully it's not going to grad the deregistration much.

marxin:
Hm, it's causing a new segfault:
```
./run.sh: line 3: 15775 Segmentation fault      (core dumped) $WASMER -q run main.wasm --dir . -- failing_exec
Testing fork...
Testing fork ❌
```

@zebreus I was unable to reproduce it locally (running it 1000x). Can you check it?

zebreus:
I reran the CI job on this PR and it seems consistently broken. Not sure why, it seems completly unrelated to the change

zebreus:
Rebased this PR on the latest main, maybe that fixes it

zebreus:
I pushed an extra commit that reverts all changes to check if that makes CI pass to be sure that the changes actually cause the crash

marxin:
Closing in favor of #5953.

PR #5913, opened 2025-11-26
CI: drop unused sccache and handle GH workflow issues

marxin:
Based on the discussion with @theduke, the `sccache` tends to be really slow and was disabled some time ago. The other changes are related to GH workflow issues reported in the VSCode.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -200,7 +200,8 @@ jobs:
           echo "$(pwd)/wasixcc-install" >> $GITHUB_PATH
           mkdir -p ~/.wasixcc/llvm
           mkdir -p ~/.wasixcc/sysroot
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yaml
```diff
@@ -1097,7 +1096,7 @@ jobs:
           #   if: false # matrix.build != 'macos-arm'
           #   shell: bash
           #   run: |
-          #     export WASMER_PATH=`pwd`/target/${{ matrix.target }}/release/wasmer${{ matrix.exe }}
```
syrusakbary, line 1101: Having matrix exe is important for windows?

marxin: Correct, let me add the missing `exe` to the matrix.

PR #5914, opened 2025-11-26
CI: leave MacOS 13 for x86_64 target.

marxin:
The GitHub runners are already being disabled:
https://github.com/actions/runner-images/issues/13046


baalimago (approved):
LGTM

Issue #5891, opened 2025-11-20
CI cache not working


marxin:
Apparently, the we're not using the CI build cache: https://github.com/wasmerio/wasmer/actions/runs/19534986168/job/55926831893#step:14:29

due to `Warning: Error from S3: NoSuchBucket: NoSuchBucket`

marxin:
Now we hit: `Warning: Error from S3: InvalidArgument: Credential access key has length 64, should be 32`.

marxin:
Should be in action now - monitor if it works.

marxin:
Fixed now.

PR #5916, opened 2025-11-26
CI: add posix-compatible TAR for alpine

marxin:
Related: #5891


marxin:
Sure, I should have explained it better. The GH action passes `tar --posix` option which is supported by the GNU tar, but not by the `tar` installed on Alpine by default. That one comes from Busybox:
```
/ # tar --posix
tar: unrecognized option: posix
BusyBox v1.37.0 (2025-08-05 16:40:33 UTC) multi-call binary.
```

PR #5917, opened 2025-11-26
chore(singlepass): simplify arch_supports_x

marxin:
Right now, all targets support canonicalization and the indirect call trampolines are unnecessary.

### DIFF
--- a/lib/compiler-singlepass/src/machine_x64.rs
+++ b/lib/compiler-singlepass/src/machine_x64.rs
@@ -12,7 +12,10 @@ use crate::{
 use dynasmrt::{DynasmError, VecAssembler, x64::X64Relocation};
 #[cfg(feature = "unwind")]
 use gimli::{X86_64, write::CallFrameInstruction};
-use std::ops::{Deref, DerefMut};
+use std::{
+    collections::HashMap,
+    ops::{Deref, DerefMut},
+};
 use wasmer_compiler::{
     types::{
         address_map::InstructionAddressMap,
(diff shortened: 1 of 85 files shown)


lib/compiler-singlepass/src/machine_x64.rs
```diff
@@ -7078,248 +7069,232 @@ impl Machine for MachineX86_64 {
         loc_b: Location,
         ret: Location,
     ) -> Result<(), CompileError> {
-        if !self.arch_supports_canonicalize_nan() {
```
syrusakbary, line 7075: This should be more like if we want to use nan canonicalization or not (regardless if the architecture supports it)

marxin: If I do that, I get to the following spec issues even w/o the NaN canonicalization:
```
    Error: Failed directives on tests/wast/spec/f64.wast:
      • expected -0.0 (Core(F64(Value(F64 { bits: 9223372036854775808 })))), got F64(0.0) (9223372036854775808) (1620:1)
      • expected Core(F64(ArithmeticNan)), got F64(NaN) (18443366373989023744) (1652:1)
      • expected Core(F64(ArithmeticNan)), got F64(NaN) (9219994337134247936) (1654:1)
      • expected Core(F64(ArithmeticNan)), got F64(NaN) (18443366373989023744) (1656:1)
      • expected Core(F64(ArithmeticNan)), got F64(NaN) (9219994337134247936) (1658:1)
      • expected Core(F64(ArithmeticNan)), got F64(NaN) (18443366373989023744) (1692:1)
      • expected Core(F64(ArithmeticNan)), got F64(NaN) (9219994337134247936) (1694:1)
      • expected Core(F64(ArithmeticNan)), got F64(NaN) (18443366373989023744) (1696:1)
      • expected Core(F64(ArithmeticNan)), got F64(NaN) (9219994337134247936) (1698:1)
      • expected Core(F64(ArithmeticNan)), got F64(NaN) (18443366373989023744) (1732:1)
      • expected Core(F64(ArithmeticNan)), got F64(NaN) (9219994337134247936) (1734:1)
      • expected Core(F64(ArithmeticNan)), got F64(NaN) (18443366373989023744) (1736:1)
```

The similar thing is also done in the ARM64 target.

syrusakbary: Thanks for the point, can you add this as a comment on the code? It will help on future readers on the code on why we have this unconditionally.

marxin: Sure, I like the idea. To be honest, I think it's very same reason what @theduke faced in  #5796 (where only calling the LLVM min/max intrinsic is not enough).

syrusakbary (approved):

PR #5906, opened 2025-11-25
Async execution API + concurrent execution support

Arshia001:
## Summary and explanation

This rather large PR add support for async execution (through `Function::call_async` and the `Function::new_*_async` family).

The design is explained in WARP-43.

A call to `Function::call_async` starts a new async context and a coroutine. The async context allows imported functions to suspend; such functions may be created with a call to `Function::new_*_async`, and passing in an async or otherwise future-returning host function.

New coroutines may be started simultaneously by making further calls to `Function::call_async`, either alongside the first `call_async` or within imported functions. The embedder shall be responsible for driving the resulting futures to completion.

Note that this PR changes how `Store` and `AsStoreMut` work fundamentally, including breaking changes to their API. In the new model, `Store`s use an `Rc` internally, which lets them stay alive as long as necessary. However, since many coroutines may be accessing the store simultaneously, the `Store` also uses an internal single-threaded read-write lock. To access the store's data, this lock must be taken, which is why `Store` itself no longer implements `AsStoreRef` and `AsStoreMut`. Embedders may gain access to an `AsStoreMut` by calling `Store::as_mut`, which acquires a write lock on the store and places it in the returned `AsStoreMut` implementation.

Since async functions may suspend, the lifetime of the write lock must be handled with more care; this is why `Function::call_async` receives a `&Store` directly. However, as long as the store remains locked, the future cannot progress; For example, this code results in a deadlock:

```rust
// Create a new store, and a clone of it
let store = Store::new();

// Hypothetically clone the store - this is possible because an Rc
// is used, but the API is not exposed to avoid this exact deadlock
// scenario
let store_clone = store.clone();

// Lock the store down
let mut store_mut = store.as_mut();

// Get a function reference
let func : Function = ...;

// Call the function via the clone of the store and await it
func.call_async(&store_clone).await;

// Deadlock happens - the internal lock is held by `store_mut`. As soon as the
// lock is dropped, the future will receive a notification and resume. However,
// since `store_mut` is owned by this function, and we're already waiting for
// the future to run, we have a deadlock.
```

Note that, as of this version, the futures spawned by `Function::call_async` are not `Send`. A single-thread async runtime may be used to drive multiple coroutines forward on the same thread.

------------------

## Original PR description

This PR adds two functions mimicking the JS-promise-integration API for WebAssembly (jspi):
https://github.com/WebAssembly/js-promise-integration

[...]

PR #5920, opened 2025-11-26
Async execution API + concurrent execution support, take 2

Arshia001:
Supersedes #5906.

## Summary and explanation

This rather large PR adds support for async execution (through `Function::call_async` and the `Function::new_*_async` family).

A call to `Function::call_async` starts a new async context and a coroutine. The async context allows imported functions to suspend; such functions may be created with a call to `Function::new_*_async`, and passing in an async or otherwise future-returning host function.

New coroutines may be started simultaneously by making further calls to `Function::call_async`, either alongside the first `call_async` or within imported functions. The embedder shall be responsible for driving the resulting futures to completion.

This PR also adds a new `StoreAsync`, which can be constructed using `Store::into_async`. This takes ownership of the store, and moves the internal store data into a read-write lock. Each coroutine stores a handle to the lock, and has to lock it down for writing before it can start executing. This allows all coroutines executing against the same store to keep progressing without interleaving and possibly causing the store to get corrupted.

The same is also true of async imported functions with environments; those receive an `AsyncFunctionEnvMut`, which stores a handle to the lock, but not a write lock on the handle. When an import needs to modify the store's state, including to access its own function environment, it needs to lock the store down for writing using `AsyncFunctionEnvMut::write`. Note that, to let other coroutines be driven forward, an imported function should drop any write lock before `await`ing a future. If the write lock is held across an `await` point, other coroutines will not be able to progress, which can cause a deadlock if coroutines need to await one another or otherwise depend on one another's work. A simplified deadlock scenario can look like this:

```
// coroutine 1
let lock = env.write();
// do stuff
result_from_coroutine_2.await; // <- the write lock is still held here

// coroutine 2
let lock = env.write(); // <- this can't progress, since the coroutine 1 already holds the lock
// do stuff
report_result_to_coroutine_1();
```

This can be fixed by `drop`ing the lock before awaiting and reacquiring it later if needed.

Note that, as of this version, the futures spawned by `Function::call_async` are not `Send`. A single-thread async runtime may be used to drive multiple coroutines forward on the same thread.

## About the store context

Previously, the store would be owned by the embedder, and a mutable borrow would be passed into `Function::call`. The wasmer runtime would then store raw pointers to the store and use them as necessary. This can no longer be the case with async execution, since every coroutine must own its handle to the store.

This is where the new `StoreContext` comes into play. It understands sync vs async contexts, and has separate cases for each. In a sync context, `StoreContext` just saves a raw pointer, similar to the old behavior. In an async context, it saves a write lock on the shared store.

The `StoreContext` is also in charge of letting the rest of the runtime know what kind of context is currently active. Certain restrictions apply depending on the mix of async/sync context and async/sync imported functions:

* sync context, sync function: No change from previous version
* async context, async function: A lock handle is passed into the function. The function cannot access its environment or the store inline with the coroutine that called it; it must suspend (probably at an `env.write()` call) at least once. Once the function suspends, the async runtime will suspend the coroutine, making the store available for locking again, at which point the function can resume.
* async context, sync function: A reference to the write lock is passed into the function. The function itself cannot observe the difference.
* sync context, async function: As long as the async function does not suspend, it will work. If it suspends, a trap will be raised. A "fake" lock handle is passed into the function. This fake lock handle cannot be locked down, and `env.write()` will suspend indefinitely; `await`ing the resulting future will trigger the trap described above.

The intention is that, aside from an async function trying to suspend in a sync context (which is impossible for obvious reasons), every other scenario should *just work*.

------------------

## Original PR description

This PR adds two functions mimicking the JS-promise-integration API for WebAssembly (jspi):
https://github.com/WebAssembly/js-promise-integration

Following the new [JS methods](https://github.com/WebAssembly/js-promise-integration/blob/main/proposals/js-promise-integration/Overview.md), this PR introduces:
* A new `Function::new_async(...)`, `Function::new_with_env_async(...)`, `Function::new_typed_async(...)` and `Funciton::new_typed_with_env_async(...)` as to mimic `new WebAssembly.Suspending(...)`
* A new way of calling suspending-function that is as simple as `function.call_async(...)` instead of just `function.call(...)` to mimic `WebAssembly.promising` API

-------------------

## Tips for reviewers

This PR was designed in stages; as we discovered new potential for better usability and better APIs, we would go back to do more design work. As such, it is recommended to review the PR stage by stage. The stages are as follows:

* [Initial implementation of async function call API by @syrusakbary](https://github.com/wasmerio/wasmer/pull/5920/files/96ebd62): df08965..96ebd62
* [Initial feedback + support for out-of-order execution of coroutines](https://github.com/wasmerio/wasmer/pull/5920/files/96ebd62..b19c393): 87df386..b19c393
* [Proper implementation of `Function::new_typed_*_async` by @syrusakbary](https://github.com/wasmerio/wasmer/pull/5920/files/b19c393..a4f9896): a4f9896
* [Initial implementation of store locking mechanism and the new store APIs](https://github.com/wasmerio/wasmer/pull/5920/files/a4f9896..4ced3be): 8473345..4ced3be
* [Additional test by @zebreus](https://github.com/wasmerio/wasmer/pull/5920/files/4ced3be..36c1fad): 50d5b04..36c1fad
* [Make coroutines `!Send`](https://github.com/wasmerio/wasmer/pull/5920/files/36c1fad..fb77c93): fb77c93
* [The rest of the changes from the old PR](https://github.com/wasmerio/wasmer/pull/5920/files/b82c860..d909833): b82c860..a6dad39
* [Revised implementation of async API](https://github.com/wasmerio/wasmer/pull/5920/files/d909833..7825743): d909833 onwards

### DIFF
--- a/lib/c-api/src/wasm_c_api/externals/function.rs
+++ b/lib/c-api/src/wasm_c_api/externals/function.rs
@@ -8,7 +8,7 @@ use libc::c_void;
 use std::convert::TryInto;
 use std::mem::MaybeUninit;
 use std::sync::{Arc, Mutex};
-use wasmer_api::{Extern, Function, FunctionEnv, FunctionEnvMut, Value};
+use wasmer_api::{Extern, Function, FunctionEnv, FunctionEnvMut, RuntimeError, Value};
 
 #[derive(Clone)]
 #[allow(non_camel_case_types)]


lib/api/src/entities/function/env/mod.rs
```diff
@@ -83,6 +83,12 @@ impl<T: Send + 'static> FunctionEnvMut<'_, T> {
     pub fn data_and_store_mut(&mut self) -> (&mut T, StoreMut<'_>) {
         self.0.data_and_store_mut()
     }
+
+    /// Creates an [`AsStoreAsync`] from this [`AsyncFunctionEnvMut`] if the current
```
zebreus:
```suggestion
    /// Creates an [`AsStoreAsync`] from this [`FunctionEnvMut`] if the current
```

zebreus:
Being able to upgrade a `FunctionEnvMut` to an `AsyncStoreMut` is nice, but a bit unwieldy, for example usage like
```rust
let (data, mut store) = ctx.data_and_store_mut();
... some sync stuff
let mut async_store = ctx.as_store_async().unwrap();
... do something with the async store

... do more sync stuff with store
```
does not compile as I can't access `ctx`  while it's still borrowed from `data_and_store_mut`.

lib/api/src/entities/store/async_.rs
```diff
@@ -0,0 +1,199 @@
...
+    StoreContext, StoreInner, StoreMut, StorePtrWrapper, StoreRef,
+};
+
+use wasmer_types::StoreId;
+
+/// A store that can be used to invoke
+/// [`Function::call_async`](crate::Function::call_async).
+pub struct StoreAsync {
+    pub(crate) id: StoreId,
+    pub(crate) inner: LocalRwLock<StoreInner>,
```
zebreus: Having these fields acccesible outside of this module makes a safe implementation (and review) of StoreAsync much more complex, as we need to consider all uses to ensure they don't change. Also, it makes it possible to construct new `StoreAsyncs outside this module.

lib/api/src/entities/store/async_.rs
```diff
@@ -0,0 +1,199 @@
...
+    FromStoreContext(StorePtrWrapper),
+}
+
+/// A read lock on an async store.
+pub struct AsyncStoreReadLock<'a> {
+    pub(crate) inner: AsyncStoreReadLockInner,
+    pub(crate) _marker: PhantomData<&'a ()>,
+}
+
+impl<'a> AsyncStoreReadLock<'a> {
```
zebreus: I think this should be named `AsyncStoreReadGuard`, because that's usually the pattern for the name of guards in `RwLock`

lib/api/src/entities/store/async_.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+impl AsStoreAsync for StoreAsync {
+    fn store_ref(&self) -> &StoreAsync {
+        self
+    }
+}
+
+pub(crate) enum AsyncStoreReadLockInner {
```
zebreus: Use either `StoreAsync` or `AsyncStore` consistently

lib/api/src/entities/store/async_.rs
Arshia001: A lot of the runtime code does the same thing where, instead of creating a huge number of functions, we just expose a type's internals to the entire crate. I don't see a reason why `StoreAsync` should be different.

lib/api/src/entities/store/async_.rs
```diff
@@ -0,0 +1,199 @@
...
+/// [`Function::call_async`](crate::Function::call_async).
+pub struct StoreAsync {
+    pub(crate) id: StoreId,
+    pub(crate) inner: LocalRwLock<StoreInner>,
+}
+
+impl StoreAsync {
+    /// Transform this [`StoreAsync`] back into a [`Store`]
+    /// if this is the only clone of it and is unlocked.
+    pub fn into_store(self) -> Result<Store, Self> {
```
zebreus, line 35: The opposite is called `into_async` so I was somehow expecting this to be called `into_sync`

lib/api/src/entities/store/async_.rs
```diff
@@ -0,0 +1,199 @@
...
+use wasmer_types::StoreId;
+
+/// A store that can be used to invoke
+/// [`Function::call_async`](crate::Function::call_async).
+pub struct StoreAsync {
+    pub(crate) id: StoreId,
+    pub(crate) inner: LocalRwLock<StoreInner>,
+}
+
+impl StoreAsync {
```
zebreus, line 18: StoreAsync does not implement `Debug`, so I can not unwrap the result of `into_store`.

Arshia001:
@copilot review again

zebreus:
Not sure if it is a good or a bad sign that the copilot left no comments at all

examples/throw_exception.rs
```diff
@@ -83,10 +83,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     // Both Function::new_with_env and Function::new_typed_with_env can throw
     // exceptions if they return a Result with a RuntimeError.
-    fn throw1(
-        mut env: FunctionEnvMut<MyEnv>,
-        _params: &[Value],
-    ) -> Result<Vec<Value>, RuntimeError> {
+    fn throw1(mut env: FunctionEnvMut<MyEnv>, _params: &[Value]) -> DynamicFunctionResult {
```
syrusakbary: We adapted the API with this new type, but it hides mostly what is the type, basically... it makes things harder to understand, not easier.
I honestly think being explicit on this is better and easier to grasp.

We went from:
```
Result<Vec<Value>, RuntimeError>
```

To:
```
DynamicFunctionResult
```

(what is DynamicFunciton result? What is a dynamic function? it requires context to answer, where the other is much more straightforward to see the return type...).

We should undo this change.

lib/api/src/entities/function/env/inner.rs
```diff
@@ -85,11 +87,6 @@ impl<T> BackendFunctionEnv<T> {
         }
     }
 
-    //#[allow(dead_code)] // This function is only used in js
```
syrusakbary, line 88: Good call

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+
+/// A mutable handle to the [`FunctionEnv`] in an [`BackendAsyncFunctionEnvMut`].
+#[cfg(feature = "experimental-async")]
+pub enum BackendAsyncFunctionEnvHandleMut<T> {
+    #[cfg(feature = "sys")]
+    /// The function environment handle for the `sys` runtime.
+    Sys(crate::backend::sys::function::env::AsyncFunctionEnvHandleMut<T>),
+    #[cfg(not(feature = "sys"))]
+    /// Placeholder for unsupported backends.
+    Unsupported(PhantomData<T>),
```
syrusakbary: I believe the unsupported one should be active always.
Imagine you activate the sys feature AND the wasmi feature (so you can run things with the cranelift compiler and with wasmi), then the wasmi backend would not have this option.

```suggestion
    #[cfg(any(feature = "wasmi", feature = "wamr", [...] ))]
    /// Placeholder for unsupported backends.
    Unsupported(PhantomData<T>),
```

lib/api/src/entities/function/inner.rs
```diff
@@ -144,8 +145,8 @@ impl BackendFunction {
     pub fn new_typed<F, Args, Rets>(store: &mut impl AsStoreMut, func: F) -> Self
     where
         F: HostFunction<(), Args, Rets, WithoutEnv> + 'static + Send + Sync,
-        Args: WasmTypeList,
-        Rets: WasmTypeList,
+        Args: WasmTypeList + 'static,
```
syrusakbary: Is static still needed here?

lib/api/src/entities/function/inner.rs
```diff
@@ -203,8 +204,8 @@ impl BackendFunction {
     ) -> Self
     where
         F: HostFunction<T, Args, Rets, WithEnv> + 'static + Send + Sync,
-        Args: WasmTypeList,
-        Rets: WasmTypeList,
+        Args: WasmTypeList + 'static,
```
syrusakbary: Is static still needed here?

lib/api/src/backend/sys/entities/function/typed.rs
```diff
@@ -99,6 +117,54 @@ macro_rules! impl_native_traits {
                 // Ok(Rets::from_c_struct(results))
             }
 
+            #[allow(unused_mut)]
+            #[allow(clippy::too_many_arguments)]
+            #[cfg(feature = "experimental-async")]
+            pub(crate) fn call_async_sys(
```
syrusakbary: This function seems not not be used anywhere. Remove it

lib/api/src/backend/sys/entities/function/typed.rs
```diff
@@ -99,6 +117,54 @@ macro_rules! impl_native_traits {
...
+                let func = self.func.clone();
+                let store = store.store();
+                Self::call_async_sys_internal(func, store, $($x),*)
+            }
+
+            /// Call the typed func asynchronously.
+            #[allow(unused_mut)]
+            #[allow(clippy::too_many_arguments)]
+            #[cfg(feature = "experimental-async")]
+            pub(crate) fn call_async_sys_internal(
```
syrusakbary: Since `call_async_sys` is not used and we will remove, we can simply name this function `call_async_sys` instead of `call_async_sys_internal`

lib/api/src/entities/function/env/inner.rs
Arshia001: You're right, and it's weird we don't catch this in any of the CI pipelines.

examples/throw_exception.rs
Arshia001: Well, clippy was complaining rather severely about types being too complex in the async functions, which I kind of agree with, which gave birth to these types. I can remove them and silence the lints.

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+}
+
+#[cfg(feature = "experimental-async")]
+impl<T: 'static> BackendAsyncFunctionEnvHandleMut<T> {
+    /// Returns a mutable reference to the host state in this function environment.
+    pub fn data_mut(&mut self) -> &mut T {
+        match self {
+            #[cfg(feature = "sys")]
+            Self::Sys(f) => f.data_mut(),
+            #[cfg(not(feature = "sys"))]
```
syrusakbary: Change logic here (is not [not sys], is (any(wamr, wasmi, ...))

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+            _ => unsupported_async_backend(),
+        }
+    }
+
+    /// Returns both the host state and the attached StoreMut
+    pub fn data_and_store_mut(&mut self) -> (&mut T, &mut impl AsStoreMut) {
+        match self {
+            #[cfg(feature = "sys")]
+            Self::Sys(f) => f.data_and_store_mut(),
+            #[cfg(not(feature = "sys"))]
```
syrusakbary: Same

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+        }
+    }
+
+    /// Borrows a new [`BackendFunctionEnvMut`] from this
+    /// [`BackendAsyncFunctionEnvHandleMut`].
+    pub fn as_function_env_mut(&mut self) -> BackendFunctionEnvMut<'_, T> {
+        match self {
+            #[cfg(feature = "sys")]
+            Self::Sys(f) => BackendFunctionEnvMut::Sys(f.as_function_env_mut()),
+            #[cfg(not(feature = "sys"))]
```
syrusakbary: Same

lib/api/src/entities/function/env/inner.rs
```diff
@@ -204,12 +201,26 @@ impl<T: Send + 'static> BackendFunctionEnvMut<'_, T> {
...
+    /// Creates an [`AsStoreAsync`] from this [`BackendFunctionEnvMut`] if the current
+    /// context is async.
+    #[cfg(feature = "experimental-async")]
+    pub fn as_store_async(&self) -> Option<impl AsStoreAsync + 'static> {
+        match self {
+            #[cfg(feature = "sys")]
+            Self::Sys(f) => f.as_store_async(),
+            #[cfg(feature = "sys")]
+            _ => unsupported_async_backend(),
+            #[cfg(not(feature = "sys"))]
```
syrusakbary, line 216: Same

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+/// A shared handle to a [`FunctionEnv`], suitable for use
+/// in async imports.
+#[derive(derive_more::From)]
+#[non_exhaustive]
+#[cfg(feature = "experimental-async")]
+pub enum BackendAsyncFunctionEnvMut<T> {
+    #[cfg(feature = "sys")]
+    /// The function environment for the `sys` runtime.
+    Sys(crate::backend::sys::function::env::AsyncFunctionEnvMut<T>),
+    #[cfg(not(feature = "sys"))]
```
syrusakbary: Same

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+}
+
+/// A read-only handle to the [`FunctionEnv`] in an [`BackendAsyncFunctionEnvMut`].
+#[non_exhaustive]
+#[cfg(feature = "experimental-async")]
+pub enum BackendAsyncFunctionEnvHandle<T> {
+    #[cfg(feature = "sys")]
+    /// The function environment handle for the `sys` runtime.
+    Sys(crate::backend::sys::function::env::AsyncFunctionEnvHandle<T>),
+    #[cfg(not(feature = "sys"))]
```
syrusakbary: Same

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+            _ => unsupported_async_backend(),
+        }
+    }
+
+    /// Creates an [`AsStoreAsync`] from this [`BackendAsyncFunctionEnvMut`].
+    pub fn as_store_async(&self) -> impl AsStoreAsync + 'static {
+        match self {
+            #[cfg(feature = "sys")]
+            Self::Sys(f) => f.as_store_async(),
+            #[cfg(not(feature = "sys"))]
```
syrusakbary: Same

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+            _ => unsupported_async_backend(),
+        }
+    }
+
+    /// Returns both the host state and the attached StoreRef
+    pub fn data_and_store(&self) -> (&T, &impl AsStoreRef) {
+        match self {
+            #[cfg(feature = "sys")]
+            Self::Sys(f) => f.data_and_store(),
+            #[cfg(not(feature = "sys"))]
```
syrusakbary: Same

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+    }
+}
+
+#[cfg(feature = "experimental-async")]
+impl<T: 'static> AsStoreRef for BackendAsyncFunctionEnvHandleMut<T> {
+    fn as_store_ref(&self) -> StoreRef<'_> {
+        match self {
+            #[cfg(feature = "sys")]
+            Self::Sys(f) => AsStoreRef::as_store_ref(f),
+            #[cfg(not(feature = "sys"))]
```
syrusakbary: Same

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+    }
+}
+
+#[cfg(feature = "experimental-async")]
+impl<T: 'static> AsStoreMut for BackendAsyncFunctionEnvHandleMut<T> {
+    fn as_store_mut(&mut self) -> StoreMut<'_> {
+        match self {
+            #[cfg(feature = "sys")]
+            Self::Sys(f) => AsStoreMut::as_store_mut(f),
+            #[cfg(not(feature = "sys"))]
```
syrusakbary: Same

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+            _ => unsupported_async_backend(),
+        }
+    }
+
+    fn objects_mut(&mut self) -> &mut crate::StoreObjects {
+        match self {
+            #[cfg(feature = "sys")]
+            Self::Sys(f) => AsStoreMut::objects_mut(f),
+            #[cfg(not(feature = "sys"))]
```
syrusakbary: Same

lib/api/src/entities/function/env/inner.rs
```diff
@@ -238,3 +249,197 @@ where
...
+    }
+}
+
+#[cfg(feature = "experimental-async")]
+impl<T: 'static> AsStoreRef for BackendAsyncFunctionEnvHandle<T> {
+    fn as_store_ref(&self) -> StoreRef<'_> {
+        match self {
+            #[cfg(feature = "sys")]
+            Self::Sys(f) => AsStoreRef::as_store_ref(f),
+            #[cfg(not(feature = "sys"))]
```
syrusakbary: Same

syrusakbary (approved):

PR #5921, opened 2025-11-27
CI: move back to working actions-cache-s3@v2

marxin:
For some reason, after the version bump the Cache restore is not working any more. Let's move back to the working version.

Related: #5891


marxin:
Cool! It's already this PR where I see the Cache restore properly happening.

PR #5924, opened 2025-11-27
fix(backend-api): Update GraphQL schema

theduke:
The GraphQL API was updated, making some previously mandatory fields optional.

This caused parsing issues and deployment failures for remote builds.

Mitigations have been implemented upstream, but we need to update the
schema to align the definitions.


theduke:
You can never break public GraphQL APIs, or public APIs in general, anyway, because you don't know who is using it.

PR #5945, opened 2025-12-04
fix: Set LLVM target machine's ABI in a portable way

wakabat:
Previously we were using a hacking way to set the ABI for LLVM target machine in rv64 architecture. We can use LLVMTargetMachineOptions to properly set the ABI. This avoids having to change the code for (possibly) each LLVM version.


marxin:
Thanks for the PR, I really value the simplification.

PR #5952, opened 2025-12-09
chore: do not call process::exit from wasmer-cli + careful EH frame deregistration

marxin:


marxin:
Not calling the `process::exit` in the wasmer CLI is not enough it seems. I am still facing calls to `get_trace_and_trapcode` after my changes -> which eventually leads to crashes.

PR #5953, opened 2025-12-09
Prevent crashes during exit due to deregistering frames on exit (v3) 

marxin:
Fixes: #5877


lib/compiler/src/engine/trap/stack.rs
```diff
@@ -1,10 +1,21 @@
...
+
 use super::frame_info::{FRAME_INFO, GlobalFrameInfo};
 use backtrace::Backtrace;
 use wasmer_types::{FrameInfo, TrapCode};
 use wasmer_vm::Trap;
 
 /// Given a `Trap`, this function returns the Wasm trace and the trap code.
 pub fn get_trace_and_trapcode(trap: &Trap) -> (Vec<FrameInfo>, Option<TrapCode>) {
+    #[cfg(unix)]
+    // If the exit is called, we can access the back-trace information any longer (#5877)
```
syrusakbary:
```suggestion
    // If the exit is called, we can't access the back-trace information any longer (#5877)
```

PR #5957, opened 2025-12-10
feature: disable crate-exe and create-obj subcommands

marxin:


docs/dev/create-exe.md
```diff
@@ -1,5 +1,8 @@
 # wasmer create-exe
 
+> [!Caution] 
+This feature has been removed from wasmer-cli due to insufficient support capabilities.
```
syrusakbary:
```suggestion
This feature has been disabled from wasmer-cli due to unresolved issues and insufficient support capabilities to fix them. We would like to re-enable this as soon as all the existing issues are resolved (feel free to create a PR if you would like to help us with this quest: [LINK TO EXISTING ISSUES])
```

marxin:
@syrusakbary can be merged now

wakabat:
Hey I understand there might be unresolved bugs. However, the linked issues all have to do with `create-exe`. In this case, does it make sense to keep `create-obj`, and just disable `create-exe` for now?

If all we need to tackle is linker issue, I think we can keep `create-obj`. The generated object can still be linked using a full featured linker such as lld, right?

marxin:
I'm gonna take a look much deeper on the serialization format in the next week and will reach out to you after that.

PR #5962, opened 2025-12-11
Support EH for Cranelift compiler

marxin:
This change adds exception handling (EH) support to the Cranelift backend. We leverage Cranelift's built-in EH codegen while reusing Wasmer's VM API (consistent with the LLVM backend). Like LLVM, we depend on libunwind/libgcc for unwinding. However, we diverge from Cranelift's native unwinder, so we must emit the .gcc_except_table section (LSDA - Language-Specific Data Area) to enable proper EH integration. The implementation has been validated with `php-32` and `python` packages executing basic code snippets successfully.

### DIFF
--- a/tests/ignores.txt
+++ b/tests/ignores.txt
@@ -2,6 +2,7 @@
 singlepass spec::simd # Singlepass doesn't support yet SIMD (no one asked for this feature)
 singlepass spec::exception_handling # Singlepass doesn't support EH yet (no one asked for this feature)
 windows spec::exception_handling # No EH support on Windows yet
+macos+cranelift spec::exception_handling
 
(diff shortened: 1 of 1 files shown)


lib/compiler-cranelift/src/translator/code_translator.rs
```diff
@@ -76,6 +76,10 @@
 
 mod bounds_checks;
 
+const CATCH_ALL_TAG_VALUE: i32 = i32::MAX;
```
syrusakbary: Ideally, this would be a value in `wasmer-types` and used in the compilers and VM (that way, we assure always consistency on this across compilers and VM)

Issue #5958, opened 2025-12-10
`std::fs::hard_link` stopped after with rustc +nightly


marxin:
After the recent change in the main branch of the Rust compiler (https://github.com/rust-lang/rust/pull/147572), the following code snippet stopped working on Wasmer:

```
fn main() {
    std::fs::hard_link("/tmp/1.txt", "/tmp/2.txt").unwrap();
}
```

```
✦ ❯ cargo +nightly b --target=wasm32-wasip1
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.03s
✦ ❯ file /tmp/1.txt
/tmp/1.txt: ASCII text
✦ ❯ wasmer run target/wasm32-wasip1/debug/hardlink.wasm --mapdir /tmp:/tmp
⠁ Compiling to WebAssembly                                                                                                              
thread 'main' (1) panicked at src/main.rs:2:52:
called `Result::unwrap()` on an `Err` value: Os { code: 44, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
RuntimeError: unreachable
    at abort (hardlink-a3f77a71fb778855.wasm[161]:0x7aa1)
    at std[b3a36c8b3c2dd55]::sys::pal::wasi::helpers::abort_internal (hardlink-a3f77a71fb778855.wasm[87]:0x2ebf)
    at std[b3a36c8b3c2dd55]::process::abort (hardlink-a3f77a71fb778855.wasm[63]:0x1ffd)
    at __rustc[47f3c1cf3382b6b6]::__rust_abort (hardlink-a3f77a71fb778855.wasm[72]:0x2658)
    at __rustc[47f3c1cf3382b6b6]::__rust_start_panic (hardlink-a3f77a71fb778855.wasm[37]:0xbb0)
    at __rustc[47f3c1cf3382b6b6]::rust_panic (hardlink-a3f77a71fb778855.wasm[69]:0x249f)
    at std[b3a36c8b3c2dd55]::panicking::panic_with_hook (hardlink-a3f77a71fb778855.wasm[56]:0x186c)
    at std[b3a36c8b3c2dd55]::panicking::panic_handler::{closure#0} (hardlink-a3f77a71fb778855.wasm[53]:0x141c)
[...]

PR #5971, opened 2025-12-14
CI: re-enable test due to fixed rustc +nightly

marxin:
Fixes: #5958.


marxin:
The upstream (`rust`) change has been just merged, so it's going to be included in the tomorrow's release I guess.

PR #5978, opened 2025-12-16
Make all WASIX interop enums non_exhaustive to prevent UB in case of …

Arshia001:
…bad input from guest modules

# Description

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -176,6 +176,8 @@ jobs:
 
       - name: Install wasixcc
         uses: wasix-org/wasixcc@main
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Tool versions
         run: |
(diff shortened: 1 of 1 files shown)


lib/wasix/src/net/socket.rs
```diff
@@ -114,37 +114,40 @@ pub enum WasiSocketOption {
...
 
-impl From<Sockoption> for WasiSocketOption {
-    fn from(opt: Sockoption) -> Self {
+impl TryFrom<Sockoption> for WasiSocketOption {
+    type Error = Errno;
+
+    fn try_from(opt: Sockoption) -> Result<Self, Self::Error> {
         use WasiSocketOption::*;
         match opt {
-            Sockoption::Noop => Noop,
```
marxin, line 121: If you want, you don't have to use `Ok` on the RHS of `=>` statements, you can do:
```rust
Ok(match opt {...,  _ => return Err(Errno::Inval) })
```

Arshia001: Good point! The F# programmer in me likes expressions more than return statements I guess.

marxin: Sure!

PR #5979, opened 2025-12-16
Prevent failure due to rate limits when installing `wasixcc`

zebreus:
This PR attempts to fix random failures during `wasixcc` installation in the CI pipelines. The wasixcc install action makes about 10 calls to GitHub to determine the latest versions of `wasixcc` and its dependencies, and to download them from the GitHub release. The installation sometimes failed due to rate limits. By passing the `GITHUB_TOKEN` to that action, we can increase the rate limit and get fewer failures.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -176,6 +176,8 @@ jobs:
 
       - name: Install wasixcc
         uses: wasix-org/wasixcc@main
+        with:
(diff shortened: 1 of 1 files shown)


marxin (approved):

zebreus:
`Remote integ. tests / Run ssh tests` is still failing, but that also seems to fail in other pipelines.

PR #5980, opened 2025-12-16
setjmp/longjmp based vforking

zebreus:
Currently, the `vfork()` syscall requires asyncifying the entire module to enable unwinding/rewinding. This PR explores using setjmp/longjmp to implement the vfork syscall.

## Explanation

We want to move the entire unwinding into guestspace. 

For that, we provide a new vfork support syscall that does not perform any unwinding/rewinding itself, but instead swaps the currently active env with a shallow clone of itself with a new process ID. The `vfork()` function would then set up a setjmp, and `_exit()` and `exec*()` will then jump back to that point after restoring the env.

`vfork()` would need to be defined as a macro for this approach to work, so the `setjmp()` can happen in the same function as the vfork call.

## Pseudocode

```cpp
// ######### pseudo-c implementation

#include <setjmp.h>
#include <sys/types.h>

// __vfork_jump is the real jump buffer
jmp_buf __vfork_jump;
// This is passed to setjmp and moved to __vfork_jump once we are in a new
// environment
jmp_buf __vfork_jump_temporary;
// The pid of the vforked process
pid_t __child_pid;

pid_t __vfork_internal(int setjmp_result) {
  if (setjmp_result == 0) {
    // Swaps our env with a shallow clone
    int ret = __wasi_proc_fork_env(&__child_pid);
    if (ret != 0) {
      // Fork failed
      errno = ret;
      return (pid_t)-1;
    }

    // If the vfork was successful move the jump buffer to the real jumpbuffer
    __vfork_jump = __vfork_jump_temporary;

    // If the vfork succeeded we are now in the child

    // In the child vfork returns 0
    return (pid_t)0;
  } else {
    // In the parent vfork returns 0
    return __child_pid;
  }
}

#define vfork() __vfork_internal(setjmp(__vfork_jump_temporary))

// ######### pseudo c describing the modifications required to other library functions

void _exit(int status) {
  __wasi_proc_exit2(status);

  // proc_exit now returns if a active asyncify-free vfork is present
  longjmp(__vfork_jump, 1);
  __builtin_unreachable();
}

static int exec(const char *file, const char *arg0, ...) {
  int ret = exec(file, argv);

  if (ret == -1 && errno == 0) {
    // return from vfork
    longjmp(__vfork_jump, 1);
  }
  return ret;
}
```

## Drawbacks & Alternatives

Drawbacks:

* Moves a lot of complexity to wasix-libc, which is bad for debugging
* How to handle other types of exits
* You can not take the address of vfork when it's a macro
* We need a `proc_exit2` to have a exit function that can return.

Alternatives:

* ~~Don't use setjmp/longjmp, but use a special exception tag that can be thrown by the host~~
  * It cannot be done without the compiler support that setjmp gets.
* Use setjmp/longjmp, but register the jumpbuffer with the host and let it call into `__wasm_longjmp` directly
* Instead of having vforking logic in proc_exit and proc_exec, move that logic to a separate env_restore and never call proc_exit or proc_exec in those cases.
* Modify `proc_exit()` and allow it to return instead of introducing `proc_exit2()`

Set in stone:

* We must use `setjmp` and `longjmp`, because we need the compiler support we get by doing that (effectively splitting the function at the setjmp)
* We don't want to do any sort of preprocessing
* We must use a macro for vfork

## Appendix

<details><summary>Example for situations we can only do with the compiler support we get for setjmp/longjmp.</summary>

```cpp
int main(void) {
    int value =0 ;
    int ret;

    while (value < 10) {
      if (value == 5) {
        ret = setjmp(env);
      }
      value ++;
    }

    if (ret == 0) {
        // First time through
        printf("setjmp returned %d (initial call)\n", ret);
        risky_function();
        printf("This line is never reached\n");
    } else {
        // Returned via longjmp
        printf("setjmp returned %d (after longjmp)\n", ret);
    }

    printf("Program continues normally\n");
    return 0;
}
```

If setjmp were just a try_table, then the longjmp after the while and the if would not be able to jump to the setjmp

However, the clang/llvm splits up the function at the setjmp call, so we can jump back to it from all following parts of the function.

</details>

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -178,6 +178,7 @@ jobs:
         uses: wasix-org/wasixcc@main
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
+          sysroot-tag: v2025-12-22.1
 
       - name: Tool versions
         run: |
@@ -878,6 +879,7 @@ jobs:
         uses: wasix-org/wasixcc@main
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
+          sysroot-tag: v2025-12-22.1
 
       - name: Install LLVM
         shell: bash


lib/wasix/src/syscalls/wasix/proc_fork_env.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    // Write the child's PID to the provided pointer
+    let memory = unsafe { env.memory_view(&ctx) };
+    wasi_try_mem_ok!(pid_ptr.write(&memory, child_env.pid().raw()));
+
+    // Serialize the globals
+    let serialized_globals: Bytes = capture_store_snapshot(&mut ctx.as_store_mut())
+        .serialize()
+        .unwrap()
+        .into();
```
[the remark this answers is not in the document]
zebreus: This code was taken from proc_fork where it is done the same way.

lib/wasix/src/syscalls/wasi/proc_exit.rs
```diff
@@ -37,14 +37,24 @@ pub fn proc_exit<M: MemorySize>(
...
         child_env.process.terminate(code);
 
+        if vfork.rewind_stack.is_none() {
+            // If we are not using asyncify, we are actually done here :)
+            // See proc_vfork for information about this path
+
+            // TODO: Restore store data (globals)
+            // Or figure out if we really need to do that
+
+            return Ok(());
```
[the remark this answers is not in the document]
zebreus: obsolete

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -458,7 +459,8 @@ fn resume_vfork(
 
         // Jump back to the vfork point and current on execution
         let child_pid = child_env.process.pid();
-        let rewind_stack = vfork.rewind_stack.freeze();
+        // TODO: Write this in a more elegant fashion to prevent the unwrap here
+        let rewind_stack = vfork.rewind_stack.unwrap().freeze();
```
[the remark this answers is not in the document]
zebreus: obsolete

lib/wasix/src/syscalls/wasix/proc_exec3.rs
```diff
@@ -209,9 +210,20 @@ pub fn proc_exec3<M: MemorySize>(
...
                 assert!(ctx.data().context_switching_environment.is_some());
+
+                let Some(rewind_stack) = vfork.rewind_stack else {
+                    // If we are not using asyncify, we are actually done here :)
+                    // See proc_vfork for information about this path
+
+                    // TODO: Restore store data (globals)
+                    // Or figure out if we really need to do that
+
+                    return Ok(Errno::Success);
```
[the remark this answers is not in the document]
zebreus: obsolete

lib/wasix/src/syscalls/wasix/proc_exit2.rs
```diff
@@ -0,0 +1,82 @@
...
+/// place where the vfork happened is the responsibility of the
+/// caller.
+#[instrument(level = "trace", skip_all)]
+pub fn proc_exit2<M: MemorySize>(
+    mut ctx: FunctionEnvMut<'_, WasiEnv>,
+    code: ExitCode,
+) -> Result<(), WasiError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
+    debug!(%code);
```
Arshia001: This shouldn't be here, remove it please. I probably committed it by mistake at some point while debugging something else.

lib/wasix/src/syscalls/wasix/proc_fork_env.rs
```diff
@@ -0,0 +1,79 @@
...
+pub fn proc_fork_env<M: MemorySize>(
+    mut ctx: FunctionEnvMut<'_, WasiEnv>,
+    child_pid_ptr: WasmPtr<Pid, M>,
+) -> Result<Errno, WasiError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
+    let env = ctx.data();
+
+    if let Some(vfork) = env.vfork.as_ref() {
+        tracing::warn!("nesting vforks is not supported");
```
Arshia001:
```suggestion
        tracing::warn!("Nested vforks are not supported");
```

tests/wasix/vfork/run.sh
```diff
@@ -1,15 +1,35 @@
+#!/usr/bin/env bash
 set -e
 
+# Compile the main program with setjmp/longjmp based vfork
+WASIXCC_WASM_EXCEPTIONS=yes WASIXCC_PIC=yes wasixcc main.c -o main-eh.wasm -Wl,-pie
```
Arshia001: You technically don't need PIC for this, though it doesn't have any adverse effects either.

Arshia001: Looks like this test is half-relying on the top-level test script to compile the non-EH variant. This is rather brittle. You can disable that behavior with _something_ (a `.no-compile` file? You can find it in the other tests) and just build both variants here.

zebreus: Sounds good, I'll do that

zebreus: I wanted to be sure to get the ehpic sysroot, but thinking about it, it shouldn't be required

Arshia001:
@zebreus I see the WASIX pipeline is failing. I previously removed the `fork` and `vfork` functions from EH configurations because they wouldn't work. I assume you need to put `vfork` back in.

The correct way to do this would be to get the wasix-libc change reviewed and ready for merging, make sure tests run locally, merge wasix-libc and create a release, and update the wasix-libc release here so the pipeline works.

lib/wasix/src/syscalls/wasix/proc_exit2.rs
zebreus: It might be possible that I committed that. I included the exit code in the instrumentation macro for the function so we can see it when printing logs.

PR #5985, opened 2025-12-18
chore: don't call Features::detect_from_wasm if enable all option is on

marxin:
Plus, I added API for enabling `relaxed_simd`.


syrusakbary (approved):
Simpler. LGTM

PR #5987, opened 2025-12-18
chore: cleanup

marxin:
It's a collection of smaller changes I collected in my backlog.


tests/wasmer-argus/Cargo.toml
```diff
@@ -27,7 +27,7 @@ serde.workspace = true
 serde_json.workspace = true
 derive_more.workspace = true
 webc.workspace = true
-async-trait = { workspace = true }
+async-trait.workspace = true
```
syrusakbary, line 30: Side question: is this dependency really needed, since Rust added support for async in traits?

marxin: Yes, we still need it for trait objects (like `Arc<dyn VirtualNetworking>`). That's something still missing in the stable Rust.

Issue #5929, opened 2025-11-28
Update `wasmparser` crate


marxin:
The most recent version has adopted a new Visitor pattern we must migrate to:
https://docs.rs/wasmparser/0.242.0/wasmparser/macro.for_each_visit_operator.html

theduke:
Cranelift may be tied to a version of the parser, so not sure if we can freely update without waiting for Cranelift.

marxin:
Good point, that's something we should easily notice in the CI I guess.

PR #5989, opened 2025-12-19
deps: bump `wasmparser` crate and port to it

marxin:
Fixes: #5929

### DIFF
--- a/lib/api/src/utils/polyfill.rs
+++ b/lib/api/src/utils/polyfill.rs
@@ -622,7 +622,9 @@ pub fn parse_export_section(
                 module_info.declare_global_export(GlobalIndex::new(index), name)?
             }
             ExternalKind::Tag => module_info.declare_tag_export(TagIndex::new(index), name)?,
-            ExternalKind::FuncExact => unimplemented!("custom-descriptors not implemented yet"),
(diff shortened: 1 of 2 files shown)


zebreus (approved):
This wasix part looks good, I approve that. The rest also seems acceptable

lib/api/src/utils/polyfill.rs
```diff
@@ -621,6 +622,7 @@ pub fn parse_export_section(
                 module_info.declare_global_export(GlobalIndex::new(index), name)?
             }
             ExternalKind::Tag => module_info.declare_tag_export(TagIndex::new(index), name)?,
+            ExternalKind::FuncExact => unimplemented!("custom-descriptors not implemented yet"),
```
Arshia001: I'd avoid panics and do errors; we don't want Edge to crash with a bad wasm for example.

lib/compiler/src/translator/sections.rs
```diff
@@ -128,6 +128,7 @@ pub fn parse_import_section<'data>(
                     field_name,
                 )?;
             }
+            TypeRef::FuncExact(_) => unimplemented!("custom-descriptors not implemented yet"),
```
Arshia001: Same here

PR #5990, opened 2025-12-19
fix: provide unique function names

marxin:
In rare cases (cowsay), we have multiple function indices share a common function name (`dummy` in my case). Similarly to wasm2wat, provide a unique name to the function (useful for `--compiler-debug-dir` option).

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -20,9 +20,6 @@ jobs:
         include:
           - build: linux
             os: ubuntu-latest
-    env:
-      SCCACHE_AZURE_BLOB_CONTAINER: wasmerstoragesccacheblob
-      SCCACHE_AZURE_CONNECTION_STRING: ${{ secrets.SCCACHE_AZURE_CONNECTION_STRING }}
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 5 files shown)


lib/compiler/src/translator/sections.rs
```diff
@@ -524,10 +522,19 @@ pub fn parse_name_section<'data>(
...
+                        // In very rare cases a function name can have duplicates.
+                        if names_set.contains(&name) {
+                            for i in 0.. {
+                                let alternative = format!("{name}.{i}");
+                                if !names_set.contains(&alternative) {
+                                    name = alternative;
+                                    break;
+                                }
+                            }
+                        }
```
syrusakbary, line 541: I don't love this algorithm. It will have O(~n^2/2) in the worst case.
I think it would be better to keep track of the "taken" indices, and then just bump one on it.

Example:
```python
names_dict = defaultdict(int)
name = f"{name}.{names_dict[name]}"
names_dict[name] += 1
```

marxin: Yeah, I know the algorithm is super-lame, but we're speaking here about super rare case.
Anyways, fixed properly.

Issue #5998, opened 2025-12-22
build: Update webkit dependency flake.nix


theduke:
Currently some tests do not run due to mismatch in the webkit dependency.

PR #5999, opened 2025-12-22
fix(build): Update webkit in flake.nix

theduke:
Needed for some of the tests to run.

Fixes: #5998


Twey:
This breaks the build: `rusty-jsc-sys` [explicitly requires](https://github.com/wasmerio/rusty_jsc/blob/updated-api/sys/build.rs#L8) version 4.0.

PR #6007, opened 2025-12-24
Enable riscv32 target in LLVM compiler & relocations fixes

wakabat:
This commit contains several RISC-V fixes we've accumulated:

* Handle more RISC-V allocations
* Metadata must be aligned by 8
* Pass pointer width instead of hardcoding it to 8
* Implement trampoline for rv32

It fixes problems using wasmer in rv32, but for now the code does not aim to enable rv32. I feel like it is a bigger decision.

### DIFF
--- a/lib/compiler-llvm/src/compiler.rs
+++ b/lib/compiler-llvm/src/compiler.rs
@@ -183,8 +183,14 @@ impl LLVMCompiler {
             || {
                 let target_machine = self.config().target_machine(target);
                 let target_machine_no_opt = self.config().target_machine_with_opt(target, false);
-                FuncTranslator::new(target_machine, Some(target_machine_no_opt), binary_format)
-                    .unwrap()
+                let pointer_width = target.triple().pointer_width().unwrap().bytes();
+                FuncTranslator::new(
+                    target_machine,
+                    Some(target_machine_no_opt),
+                    binary_format,
+                    pointer_width,
+                )
+                .unwrap()
             },
             |func_translator, (i, input)| {
                 let module = func_translator.translate_to_module(
@@ -399,10 +405,12 @@ impl Compiler for LLVMCompiler {
                         let target_machine = self.config().target_machine_with_opt(target, true);
                         let target_machine_no_opt =
                             self.config().target_machine_with_opt(target, false);
+                        let pointer_width = target.triple().pointer_width().unwrap().bytes();
                         FuncTranslator::new(
                             target_machine,
                             Some(target_machine_no_opt),
                             binary_format,
+                            pointer_width,
                         )
                         .unwrap()
                     },
--- a/lib/compiler-llvm/src/config.rs
+++ b/lib/compiler-llvm/src/config.rs
@@ -244,14 +244,16 @@ impl LLVM {
                 info: true,
                 machine_code: true,
             }),
-            Architecture::Riscv64(_) => InkwellTarget::initialize_riscv(&InitializationConfig {
-                asm_parser: true,
-                asm_printer: true,
-                base: true,
-                disassembler: true,
-                info: true,
-                machine_code: true,
-            }),
+            Architecture::Riscv64(_) | Architecture::Riscv32(_) => {
+                InkwellTarget::initialize_riscv(&InitializationConfig {
+                    asm_parser: true,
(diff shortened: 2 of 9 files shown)


wakabat:
@marxin So this is what I have to replace #5521. I have to mention that many changes included in this PR has to do with rv32, while wasmer does not officially support rv32.

We can of course talk about it, and discuss how much in this PR shall be merged into wasmer.

marxin:
> @marxin So this is what I have to replace https://github.com/wasmerio/wasmer/pull/5521. I have to mention that many changes included in this PR has to do with rv32, while wasmer does not officially support rv32.

Thank you for the PR, generally speaking, LGTM.

> We can of course talk about it, and discuss how much in this PR shall be merged into wasmer.

I would like to see a full support of the rv32 target with LLVM - do you know what's missing on top of this pull request? Btw. can one test `rv32` on a Linux system running `rv64gc` system?

wakabat:
For our case, the only remaining work to support rv32, is merely enabling it in [this file](https://github.com/wasmerio/wasmer/blob/0de3304d2887a33963e50be5dde762553fc73160/lib/compiler-llvm/src/config.rs#L220-L314).

And I don't believe you can test `rv32` code in `rv64gc` platform, you need a real rv32 chip or VM setup.

marxin:
> For our case, the only remaining work to support rv32, is merely enabling it in [this file](https://github.com/wasmerio/wasmer/blob/0de3304d2887a33963e50be5dde762553fc73160/lib/compiler-llvm/src/config.rs#L220-L314).

All right, then include it in this PR so that it's a self-contained set of changes. And note I added a couple of relocations in the #6020 PR. This PR needs to be rebased.

> 
> And I don't believe you can test `rv32` code in `rv64gc` platform, you need a real rv32 chip or VM setup.

Frankly speaking, we can definitely add `rv32` support, but as of now, we don't have any development board we can test it on.

wakabat:
This is rebased, I've also tweaked LLVM compiler's config to introduce rv32 targets.

> as of now, we don't have any development board we can test it on.

I totally agree that a real board can never be replaced in many tests, e.g., performance analysis. But if we are limiting the goal to correctness testing, I personally believe qemu-user package can be quite handy.

In addition, I would recommend having an environment where the host is 64-bit and the target is 32-bit, it helps uncover [certain types of bugs](https://github.com/wasmerio/wasmer/pull/6007/files#diff-2b0b3efb6a510d2b18e37376c4ac1315d52057865ed8c11b878deb6a0923b596), where pointer size of the host platform is used for target code.

marxin:
> This is rebased, I've also tweaked LLVM compiler's config to introduce rv32 targets.

Great, please rename the PR as it's now purely about the `rv32` target addition. And one more rebase will be needed.

> But if we are limiting the goal to correctness testing, I personally believe qemu-user package can be quite handy.

Speaking about the qemu cross compilation - we inherently need to have a cross LLVM binaries (something like https://github.com/mean00/LLVM-embedded-toolchain-for-rv32). Do you have any experience with that? I am all ears!

wakabat:
This is rebased, the PR title is changed. I do agree that this PR is now mostly about riscv32, but I do believe some of the added relocations can be handy for riscv64 as well(especially [this one](https://github.com/wasmerio/wasmer/blob/8a86e976364c6a6fa917dcba206adc85781a0f44/lib/compiler/src/object/module.rs#L443)).

As for cross compilation, I have to say our workflow is slightly different: we only employ wasmer to generate `.o` or `.wasmu` file, then build the full binary with rustc. So unfortunately we don't use an embedded toolchain, sorry.

wakabat:
Thanks for merging this! Just to provide a more complete picture: [this commit](https://github.com/wakabat/official-wasmer/commit/72bc1374b9677f6a2e7b96423797c79598db533e) contains the remaining changes we have to use RISC-V target in wasmer. It basically includes:

* Tweaking of RISC-V target since our RISC-V platform does not support `c` extension or floating point number
* Due to the above change, we have to add a few libcall stubs so they can be linked to the correct softfloat implementations.

Wasmer might not need those since wasmer assumes floating point number support, but just want to make the information complete here, in case it helps. We do have patches for other purposes, but for generating RISC-V code, those are all the changes we need.

PR #6008, opened 2025-12-24
Fix/linker crash tls symbol without tls base

Arshia001:
This PR also contains a fix to a potential deadlock in the Linker. Mostly the same implementation as #5879 otherwise.

### DIFF
--- a/lib/wasix/src/state/linker.rs
+++ b/lib/wasix/src/state/linker.rs
@@ -1580,6 +1580,15 @@ impl Linker {
         function_id: u32,
         ctx: &mut FunctionEnvMut<'_, WasiEnv>,
     ) -> Result<bool, LinkError> {
+        // If we can get a read lock on the linker state, do it
+        if let Ok(linker_state) = self.linker_state.try_read() {
+            return Ok(linker_state
+                .allocated_closure_functions
+                .contains_key(&function_id));
+        }
+
+        // Otherwise, fall back to the path where we apply DL ops and acquire
+        // a write lock afterwards
         lock_instance_group_state!(
             group_state_guard,
             group_state,


lib/wasix/src/state/handles/mod.rs
```diff
@@ -361,10 +363,15 @@ impl WasiModuleTreeHandles {
...
     ///
     /// Returns false if the entry is not reserved for closures.
-    pub fn is_closure(&self, function_id: u32) -> bool {
-        match self {
-            WasiModuleTreeHandles::Static(_) => false,
-            WasiModuleTreeHandles::Dynamic { linker, .. } => linker.is_closure(function_id),
+    pub fn is_closure(
+        ctx: &mut FunctionEnvMut<'_, WasiEnv>,
+        function_id: u32,
+    ) -> Result<bool, LinkError> {
```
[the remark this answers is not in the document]
zebreus: i dont think we need doc changes here. Maybe move it to a freestanding function

lib/wasix/src/state/handles/mod.rs
```diff
@@ -361,10 +363,15 @@ impl WasiModuleTreeHandles {
...
+        match ctx.data().inner() {
+            WasiModuleTreeHandles::Static(_) => Ok(false),
+            WasiModuleTreeHandles::Dynamic { linker, .. } => {
+                linker.clone().is_closure(function_id, ctx)
```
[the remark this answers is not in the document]
zebreus: @Arshia001 investigate this

lib/wasix/src/state/linker.rs
```diff
@@ -583,10 +595,8 @@ pub enum ResolveError {
     #[error("Failed to perform pending DL operation: {0}")]
     PendingDlOperationFailed(#[from] LinkError),
 
-    #[error(
-        "Tried to resolve a tls symbol from a module that did not initialize thread local storage"
-    )]
-    TlsSymbolWithoutTls,
+    #[error("Module must export its __tls_base for exported TLS symbols to be resolved correctly")]
+    NoTlsBaseGlobalExport,
```
zebreus, line 599: Thats a good message

lib/wasix/src/state/linker.rs
```diff
@@ -3113,18 +3129,27 @@ impl InstanceGroupState {
...
+                    ) {
+                        Ok(export) => export,
+                        Err(ResolveError::NoTlsBaseGlobalExport) => {
+                            return Err(LinkError::MissingTlsBaseExport(
+                                import.name().to_string(),
+                                *module_handle,
+                            ));
+                        }
+                        Err(e) => {
+                            Err(e).expect("Internal error: bad in-progress symbol resolution")
```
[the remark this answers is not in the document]
zebreus: I think copilot has a point here

lib/wasix/src/state/linker.rs
```diff
@@ -1565,12 +1572,23 @@ impl Linker {
...
+        function_id: u32,
+        ctx: &mut FunctionEnvMut<'_, WasiEnv>,
+    ) -> Result<bool, LinkError> {
+        lock_instance_group_state!(
+            group_state_guard,
+            group_state,
+            self,
+            LinkError::InstanceGroupIsDead
+        );
+        write_linker_state!(linker_state, self, group_state, ctx);
```
zebreus, line 1598: Do we need a write lock here? If so add a comment why

zebreus (changes requested):

lib/wasix/src/state/linker.rs
Arshia001: Everything that touches the linker state must get a write lock first. This is because once a thread locks it down for writing and starts a DL op, every other thread will have to apply the operation before the lock can be released by the original thread. If, at that point, you request a read lock, you'll block indefinitely.

zebreus: fair

zebreus: why do we even have a read lock then

lib/wasix/src/syscalls/wasix/reflect_signature.rs
```diff
@@ -49,6 +49,12 @@ pub fn reflect_signature<M: MemorySize>(
     result_types_len: u16,
     result: WasmPtr<ReflectionResult, M>,
 ) -> Result<Errno, WasiError> {
+    let is_closure = WasiModuleTreeHandles::is_closure(&mut ctx, function_id).map_err(|e| {
+        trace!("Failed to check if function is a closure: {}", e);
+        WasiError::Exit(Errno::Noexec.into())
```
zebreus, line 54: I think this should be either `Ok(Errno::Noexec)` if it is expected to happen regularly, or a panic if it is not likely to occur.

I am not an expert on linker error states; I'm not sure how we can get here in the first place.

zebreus:
Also, please add a test that doesn't crash when a module is missing the TLS base export.

lib/wasix/src/syscalls/wasix/reflect_signature.rs
Arshia001: this happens when a DL operation fails to apply correctly, so it's neither a process-wide panic (one bad module/misimplementation in the linker shouldn't crash an Edge process) nor is it an error the guest can recover from; so we trap.

lib/wasix/src/state/linker.rs
Arshia001: for use within the DL operation application logic. Once the operation is processed on the original thread, every other thread just needs read access, and they can do it in parallel. I also added a happy path `try_read` to this method.

lib/wasix/src/state/linker.rs
zebreus, line 1601: In general I like the try_lock approach. However combining it with the direct locking increases the complexity even more. 

We could return an error in case the try_lock fails. The reflection function can then be adjusted to be like

```rust
fn reflection() {
  do_pending_linker_ops()?;
  let Ok(is_closure) = is_closure() else {
    trace!("Unlikely")
    return Ok(Errno::Again);
  };
  ...
```

I think that would simplify the control flow, however it would also mean that reflect_signature can now also return an error if the caller didn't do anything wrong, which is stupid.

zebreus: This is similar to what we are doing in `allocate_closure` where we return `Ok(Errno::Fault)` in that case.

zebreus: But I think both are bad ideas.

PR #6009, opened 2026-01-02
fix(wasix): Fix wasip1 preopen handling for Go compatibility

iamrajiv:
# Description

Go programs compiled with `GOOS=wasip1 GOARCH=wasm` failed to open files/directories on Wasmer with "Bad file number" errors. The same programs work correctly on wasmtime.

#### Root Causes

**1. Incorrect `pr_name_len` in `fd_prestat_get`** (`lib/wasix/src/fs/mod.rs`)

Wasmer reported `pr_name_len = name.len() + 1`, incorrectly including space for a null terminator. The WASI spec states `pr_name_len` should be the exact length of the path string without null terminator.

This caused Go to read preopen names like `"/\x00"` instead of `"/"`, breaking path prefix matching and making Go think no preopens existed.

**2. Wrong error code for root directory** (`lib/wasix/src/syscalls/wasix/path_open2.rs`)

When opening the root directory without `O_DIRECTORY` flag, Wasmer returned `ENOTCAPABLE` (76) instead of `EISDIR` (31). Go has retry logic that specifically handles `EISDIR` to retry with proper directory flags, but this never triggered with `ENOTCAPABLE`.

**3. Off-by-one in buffer length check** (`lib/wasix/src/syscalls/wasi/fd_prestat_dir_name.rs`)

The condition `name.len() < path_len` was too strict - changed to `<=` to allow exact-size buffers.

#### Changes

- `lib/wasix/src/fs/mod.rs`: Remove `+ 1` from `pr_name_len` calculation
- `lib/wasix/src/syscalls/wasix/path_open2.rs`: Return `Errno::Isdir` instead of `Errno::Notcapable` for `Kind::Root`
- `lib/wasix/src/syscalls/wasi/fd_prestat_dir_name.rs`: Change `<` to `<=` and remove unnecessary null terminator write

#### Testing

Tested with Go 1.25 compiled to wasip1:

```bash
➜  wasmer-fix cd /Users/rajiv.singh/Desktop/projects/wasmer-fix/go
GOOS=wasip1 GOARCH=wasm go build -o repro.wasm repro.go
➜  go git:(master) ✗ /Users/rajiv.singh/Desktop/projects/wasmer-fix/wasmer/target/release/wasmer run --dir / repro.wasm
--- Go wasip1 Repro Start ---
SUCCESS: Successfully read root directory. Found 2 entries.
 - /.
 - //
➜  go git:(master) ✗ wasmer repro.wasm
⠁ Compiling to WebAssembly
--- Go wasip1 Repro Start ---
FAIL: Error reading root directory: open /: Bad file number
➜  go git:(master) ✗
```

### DIFF
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -1579,10 +1579,8 @@ impl WasiFs {
         Prestat {
             pr_type: Preopentype::Dir,
             u: PrestatEnum::Dir {
-                // REVIEW:
-                // no need for +1, because there is no 0 end-of-string marker
-                // john: removing the +1 seems cause regression issues
-                pr_name_len: inode_val.name.read().unwrap().len() as u32 + 1,
+                // WASI spec: pr_name_len is the length of the path string, NOT including null terminator
+                pr_name_len: inode_val.name.read().unwrap().len() as u32,
             }
             .untagged(),
         }


paralin:
+1 for Go fixes and also it's funny the old comments say "TODO verify this..." etc.

paralin:
Test failures are flakes:

```
      "stderr": "error: GraphQL query failed\n╰─▶ 1: GraphQL API failure: You need to be logged in and active to perform this action.\n"
```

iamrajiv:
Could a maintainer re-run the tests or confirm this is expected?

cc: @syrusakbary

Arshia001:
Scary "seems to cause regression" comment in the old code... @zebreus you should also be aware of this change in case we do start seeing regressions anywhere.

Issue #6004, opened 2025-12-23
Cranelift backend uncaught exception in Python greenlet test


Arshia001:
Running the Python greenlet test with the LLVM backend works, but fails with an uncaught exception when using the Cranelift backend.

Commands to reproduce:

```bash
wasmer run --registry wasmer.wtf zebreus/python-with-packages-test@0.1.6 --llvm -- /tests/greenlet-test.py
```

```bash
wasmer run --registry wasmer.wtf zebreus/python-with-packages-test@0.1.6 --cranelift -- /tests/greenlet-test.py
```

The Cranelift run fails with an uncaught exception while the LLVM run succeeds.

marxin:
Seems even a simpler C++ reproducer fails with Cranelift:
```c++
struct ScopeLogger {
  int name;
  explicit ScopeLogger(int n) : name(n) {
//    __builtin_printf("[ENTER] %d\n", name);
  }
  ~ScopeLogger() noexcept(false) {
//    __builtin_printf("[EXIT]\n");
    if (name == 0) {
//      __builtin_printf("Destructor throwing!\n");
      throw 42;
    }
  }
};

int main() {
  {
    ScopeLogger s(1);
    try {
      ScopeLogger inner(0);
    } catch (const int &e) {
//      __builtin_printf("caught: %d\n", e);
    }
//    __builtin_printf("after caught\n");
  }

  return 0;
}
```

```
Uncaught exception with payload: [I32(8454320)]
    at _Unwind_RaiseException (a.out[232]:0xd004)
    at __cxa_throw (a.out[247]:0xd312)
    at __original_main (a.out[262]:0xd9cd)
    at _start (a.out[267]:0xdb46)
```

PR #6010, opened 2026-01-05
fix(cranelift): EH: ensure CatchAll is always present at the end of an action linked list

marxin:
~The PR simplifies the emission of Action table, where the key fix is that a landing pad catching both All and a Tag must have the `CatchAll` as the very last item in the Action table (in order to align with `find_eh_action`).~

We know the `CatchAll` is always the last "tag" for a particular landing pad - and thus by reverting the order of tags, we ensure it's always encoded as the very last value in the Action table linked list.

With the change, all tests in `zebreus/python-with-packages-test@0.1.6` test package are green!

Fixes: #6004

### DIFF
--- a/tests/compilers/issues.rs
+++ b/tests/compilers/issues.rs
@@ -642,6 +642,7 @@ fn issue_5795_memory_reset_size(mut config: crate::Config) {
     assert_eq!(memory.view(&store).size().0, 1);
 }
 
+#[cfg(not(target_os = "windows"))]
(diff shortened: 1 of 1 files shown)


zebreus:
Also some CI tests are failing

Arshia001 (changes requested):
I think this is the wrong change, but I'm not sure. Tentatively rejecting the change while we discuss it.

PR #6012, opened 2026-01-05
Fix relative_path_to_absolute

### LINKED ISSUE #6006: relative_path_to_absolute() off-by-one
This in wasmer/lib/wasix/src/fs/mod.rs doesn't seem right:

    pub(crate) fn relative_path_to_absolute(&self, mut path: String) -> String {
        if !path.starts_with("/") {
            let current_dir = self.current_dir.lock().unwrap();
            path = format!("{}{}", current_dir.as_str(), &path[1..]);
            if path.contains("//") {
                path = path.replace("//", "/");
            }
        }
        path
    }

If the WASM program passes in an empty string, then &path[1..] will
raise an error.

Version 6.1.0.

Robert Morris
rtm@mit.edu

zebreus:
Fixes #6006

### DIFF
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -701,15 +701,13 @@ impl WasiFs {
     }
 
     /// Converts a relative path into an absolute path
-    pub(crate) fn relative_path_to_absolute(&self, mut path: String) -> String {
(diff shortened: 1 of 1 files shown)


marxin:
@zebreus a rebase on top of the current `main` will handle the `cargo-deny` failure.

PR #6018, opened 2026-01-07
CI(aarch64-linux): include into testing matrix

marxin:


syrusakbary (approved):

syrusakbary:
Perhaps we should also modify the build workflow accordingly?

marxin:
> Perhaps we should also modify the build workflow accordingly?

Yep, good idea!

Issue #4664, opened 2024-05-10
Registry response cache does not differentiate by registry URL


theduke:
The cache used for registry API responses in `WapmSource` does not differentiate cache items by registry endpoint, leading to sharing of cache data between different registries, and confusing behaviour.

The cache should be updated to include the normalized registry URL in the cache path.

Relevant setup code here: [https://github.com/wasmerio/wasmer/blob/4fa2687a6bda9a9135637c0affeed8ff63e47b32/lib/cli/src/commands/run/wasi.rs#L646](https://github.com/wasmerio/wasmer/blob/4fa2687a6bda9a9135637c0affeed8ff63e47b32/lib/cli/src/commands/run/wasi.rs#L646)

marxin:
It's something pretty close to what I've recently adjusted in #5974. Let me fix this issue.

PR #6022, opened 2026-01-08
fix: include endpoint URL in response cache location

marxin:
Fixes: #4664

### DIFF
--- a/lib/cli/src/commands/run/wasi.rs
+++ b/lib/cli/src/commands/run/wasi.rs
@@ -502,7 +502,10 @@ impl Wasi {
         Ok(Vec::new())
     }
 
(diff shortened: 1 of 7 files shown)


syrusakbary (approved):

lib/cli/src/commands/run/wasi.rs
```diff
@@ -239,6 +239,10 @@ pub struct RunProperties {
     pub args: Vec<String>,
 }
 
+fn endpoint_to_folder(url: &Url) -> String {
+    url.to_string().replace(|c| "/:?&=#%\\".contains(c), "_")
```
syrusakbary: Can we also do this replacements?

* registry.wasmer.io -> wasmer.io
* registry.wasmer.wtf -> wasmer.wtf

marxin: Sure!

PR #6028, opened 2026-01-09
chore: drop arm,x86_64 - unsupported platforms

marxin:


marxin:
Sure, let's make a proper discussion about the platform support before!

PR #6033, opened 2026-01-09
feat: replace `--mapdir` and `--map` with `--volume` argument

marxin:
Currently, we support two arguments that largely overlap in functionality (`--map` and `--mapdir`), and the `--mapdir` syntax is especially counter-intuitive since it uses GUEST_DIR:HOST_DIR. Following the earlier discussion with Syrus, we should consolidate these flags and adopt the same convention as docker run by introducing a single `--volume` argument.

~Because this is a breaking change, we need to communicate it clearly to users. For example:~

We're going to support the legacy options, while a warning is reported to a user:
```
     Running `target/debug/wasmer run wasmer/bash -c --mapdir '/fuzz:fuzz' --dir docs -- --version`
warning: The `--mapdir` option is deprecated and will be removed in the next major release. Please use `--volume` instead.
warning: The `--dir` option is deprecated and will be removed in the next major release. Please use `--volume` instead.
```

### DIFF
--- a/lib/cli/src/commands/run/wasi.rs
+++ b/lib/cli/src/commands/run/wasi.rs
@@ -69,7 +69,7 @@ pub struct Wasi {
     )]
     volumes: Vec<MappedDirectory>,
 
-    /// Legacy option
+    // Legacy option
     #[clap(long = "dir", group = "wasi", hide = true)]
     pub(crate) pre_opened_directories: Vec<PathBuf>,
 


marxin:
@Arshia001 the PR tries to use exclusively `build_mapped_directories` function where all the validation is suppose to happen. Can you review the PR? Do you see any missing validation that needs to happen in `build_mapped_directories`?

lib/cli/src/commands/run/wasi.rs
```diff
@@ -60,16 +61,24 @@ const WAPM_SOURCE_CACHE_TIMEOUT: Duration = Duration::from_secs(10 * 60);
...
+        value_parser = parse_volume,
+    )]
+    volumes: Vec<MappedDirectory>,
+
+    /// Legacy option
+    #[clap(long = "dir", group = "wasi", hide = true)]
     pub(crate) pre_opened_directories: Vec<PathBuf>,
 
-    /// Map a host directory to a different location for the Wasm module
+    // Legacy option
```
Arshia001, line 76: One of these has `///`, while the other has `//`. Should use one or the other for consistency.

Issue #5930, opened 2025-11-28
Update `tokio-tungstenite` crate


marxin:
Apparently, we're waiting for a new release of the `graphql-ws-client` crate that will provide support for the latest `tungstenite` crate.

marxin:
The following commits needs to provided by a release: https://github.com/obmarg/graphql-ws-client/commit/fe58978ba6b15f26c5a4f8d13d977065fef2f456.

PR #6034, opened 2026-01-12
deps: bump tungstenite-related crates

marxin:
Resolves: #5930.

### DIFF
--- a/lib/virtual-net/src/rx_tx.rs
+++ b/lib/virtual-net/src/rx_tx.rs
@@ -6,6 +6,7 @@ use std::{
 
 use crate::Result;
 use bincode::config;
+use bytes::Bytes;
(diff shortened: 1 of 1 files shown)


lib/virtual-net/src/rx_tx.rs
```diff
@@ -232,8 +240,12 @@ where
                 };
 
                 let mut job = Box::pin(async move {
+                    use bytes::Bytes;
```
zebreus: We are using `bytes::Bytes` multiple times in this module. Can we move it to the top of the module?

zebreus (approved):

lib/virtual-net/src/rx_tx.rs
marxin: Nice observation, fixed that.

PR #6036, opened 2026-01-12
chore: use hex crate instead of custom format

marxin:


zebreus:
@copilot Verify that the things formatted with the hex crate are exactly identical to what was there before.

PR #6043, opened 2026-01-12
feat: `--compiler-threads` supported by all compilers

marxin:


lib/cli/src/backend.rs
```diff
@@ -206,11 +206,9 @@ pub struct RuntimeOptions {
...
     enable_pass_params_opt: bool,
 
-    /// Only available for the LLVM compiler. Sets the number of threads used to compile the
-    /// input module(s).
-    #[cfg(feature = "llvm")]
-    #[clap(long)]
-    llvm_num_threads: Option<NonZero<usize>>,
+    /// Sets the number of threads used to compile the input module(s).
+    #[clap(long, alias = "llvm-num-threads")]
+    num_threads: Option<NonZero<usize>>,
```
syrusakbary: This should be `num_compilation_threads` (or similar). Otherwise the name could be confusing as a reader might imply it means num of WASIX threads

marxin: Correct, good point. We should call it `--compiler-threads`, similarly to `--compiler-debug-dir`.

PR #6045, opened 2026-01-13
fix(make-release): use full pull

marxin:
When doing `git pull --depth=1` we might miss a commit if the release PR receives more commits (noticed the behavior).

Apart from that, I applies `ruff format` so that VSCode is not yelling at me. Review by commits is strongly recommended.

### DIFF
--- a/scripts/make-release.py
+++ b/scripts/make-release.py
@@ -417,7 +417,7 @@ def make_release(version):
     while not (already_released):
         print("git pull origin main...")
         proc = subprocess.Popen(
-            ["git", "pull", "origin", "main", "--depth", "1"],
+            ["git", "pull", "origin", "main"],
             stdout=subprocess.PIPE,
             cwd=temp_dir.name,
         )


scripts/make-release.py
```diff
@@ -39,36 +38,61 @@
...
-    import tempfile
 
     temp_dir = tempfile.TemporaryDirectory()
     print(temp_dir.name)
-    if os.system("git clone https://github.com/wasmerio/wasmer --branch " + TAG + " --depth 1 " + temp_dir.name) != 0:
+    if (
+        os.system(
+            "git clone https://github.com/wasmerio/wasmer --branch "
+            + TAG
+            + " --depth 1 "
```
zebreus, line 72: Here we still clone with depth set to one but I think that should be fine here.

zebreus (approved):

scripts/make-release.py
marxin: Yeah, that one is pretty handy as the full repo clone is quite big. Hopefully the pull (w/o the depth) will make it so we can access all the Release PR commits.

PR #6048, opened 2026-01-13
deps: bump MSRV to 1.91

marxin:
Even though, the next Rust release will be chopped in one week from now, I would prefer doing the bump a bit earlier as it's going to be the version used for `7.0` final release.


marxin:
> The previous policy was current - 3, I believe, so that would currently be 1.89.

No, we agreed on for latest 3 releases: https://github.com/wasmerio/wasmer/blob/main/docs/BUILD.md#installing-rustup
That means, currently, the oldest supported is `1.90`.

> What's the motivation for bumping now?

Expecting the `7.0` release in the upcoming week (or two), the MSRV would be `1.91` (will be released in 1 week from now) at that time. And so I would like to bump it a bit earlier.

> Also bumping to a minor version seems confusing to me, I haven't seen anyone do that.
> Users will upgrade to patch versions anyway.

Fair point, let me change it.

marxin:
@theduke seems we replaced `rust-toolchain` with `rust-toolchain.toml` as part of #6051.
@Twey How intentional was that?

marxin:
@theduke today is the release day - can you approve it?

Issue #6057, opened 2026-01-15
Package Uploads: Add support for new upload methods


theduke:
Extend the upload logic to support the new S3 upload flow.

More details to be provided by @artemyarulin .

PR #6063, opened 2026-01-15
feature(cli): Use a new R2 upload mechanism for publishing packages

artemyarulin:
Switches wasmer-cli upload to use new R2 method for uploading files. Added also integration tests which ensures uploading works, but I've left it disabled as it requires valid token/dev registry access.

Resolves: #6057


lib/cli/src/commands/package/common/mod.rs
```diff
@@ -83,39 +81,6 @@ pub(super) async fn upload(
         builder.build().unwrap()
     };
 
-    let res = client
```
artemyarulin, line 86: All the following logic is specific to GCP and as now we are using R2 via S3 upload API then there is no need to have any of those

Issue #5922, opened 2025-11-27
copilot fails to build wasmer with LLVM


zebreus:
Because the copilot environment does not have LLVM 21 installed

PR #6066, opened 2026-01-15
Add a better setup for copilot

zebreus:
This PR adds copilot setup to give it access to `cargo fmt`, `wasixcc`, our clang builds, clippy, and some other tools. This PR also adds copilot instructions that tell it to be diligent and to actually run `cargo fmt` and `cargo clippy`.

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -11,7 +11,6 @@ on:
 
 env:
   RUST_BACKTRACE: 1
-  MSRV: "1.90"
 
 jobs:
   # The job MUST be called `copilot-setup-steps`
@@ -28,12 +27,19 @@ jobs:
           sudo apt-get update -y
           sudo apt-get install -y --allow-downgrades libstdc++6
           sudo apt-get install -y --reinstall g++
-          sudo apt-get install -y creduce 
+          sudo apt-get install -y creduce
+
+      - name: Read rust version
+        id: read_rust_version
+        run: |
+          rust_toolchain=$(head -n1 rust-toolchain | tr -d '\n')
+          echo "MSRV=$rust_toolchain" >> $GITHUB_ENV
+          echo "MSRV=$rust_toolchain" >> $GITHUB_OUTPUT
 
       - name: Install Rust
         uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: ${{ env.MSRV }}
+          toolchain: ${{ steps.read_rust_version.outputs.MSRV }}
           target: x86_64-unknown-linux-gnu
           components: rustc,cargo,rust-std,rust-src,clippy,rustfmt,rust-analyzer,rust-docs
 


.github/workflows/copilot-setup-steps.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Install wasixcc
+        uses: wasix-org/wasixcc@main
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          sysroot-tag: v2025-12-22.1
+
+      - name: Ensure the setup works
+        run: |
+          cargo fmt --check || true
+          RUSTFLAGS="-D dead-code -D nonstandard-style -D unused-imports -D unused-mut -D unused-variables -D unused-unsafe -D unreachable-patterns -D bad-style -D improper-ctypes -D unused-allocation -D unused-comparisons -D while-true -D unconditional-recursion -D bare-trait-objects -D function_item_references -D clippy::uninlined_format_args " cargo clippy --all --exclude wasmer-swift --locked --allow-dirty -- -D clippy::all || true
```
marxin: Is the `|| true` intentional here?

marxin:
Btw. is this PR somehow related to #5922?

.github/workflows/copilot-setup-steps.yml
zebreus: Well, kind of yes, so it does not fail to setup if there are formatting errors. However, considering that, running those commands here is quite useless. I'll remove them tomorrow.

zebreus:
Yes, this fixes #5922

.github/copilot-instructions.md
```diff
@@ -0,0 +1,38 @@
...
+- Use version control (e.g., Git) for collaboration.
+
+## Naming Conventions
+- Use meaningful, descriptive names for variables, functions, and classes.
+- Names should reflect purpose and behavior; avoid names that require comments to explain intent.
+
+## Making your changes pass CI
+- Before submitting any changes, make sure to run the following commands to check and fix formatting and linting:
+  - `cargo fmt` to format all code according to the style guidelines.
+  - `RUSTFLAGS="-D dead-code -D nonstandard-style -D unused-imports -D unused-mut -D unused-variables -D unused-unsafe -D unreachable-patterns -D bad-style -D improper-ctypes -D unused-allocation -D unused-comparisons -D while-true -D unconditional-recursion -D bare-trait-objects -D function_item_references -D clippy::uninlined_format_args " cargo clippy --all --exclude wasmer-swift --locked --fix --allow-dirty -- -D clippy::all` to check for common mistakes and improve code quality.
```
theduke: Should we have this in the makefile?

zebreus: I think having it explicit is better for two reasons:
1. Copilot is not tempted to open the Makefile and get confused by its contents.
2. We directly communicate that linter errors are errors, ideally, which increases the likelihood of Copilot not making those errors in the first place.
3. At least locally, Copilot often wants to change the command invocations, which makes it more transparent.

zebreus:
> Note that this style of instructions, especially the clippy on everything, can increase the LLM job times A LOT.
> 
> Saw that on Edge...

I was hoping that by telling it to only run Clippy before submitting changes, it would only do it once in the end. I'll keep an eye on the copilot logs, to see if that works as expected

.github/workflows/copilot-setup-steps.yml
zebreus: Removed it

PR #6068, opened 2026-01-15
CI: restore cache + free some disk space

marxin:
This will enable Cache again on platforms and we'll get rid of occasional error:
```
  = note: /usr/lib/gcc/x86_64-alpine-linux-musl/15.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: final link failed: No space left on device
```

I got inspired `free-disk-space` GH Action, which works, but not within a container:
by https://github.com/jlumbroso/free-disk-space/issues/21#issuecomment-2679770755

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -831,7 +831,7 @@ jobs:
     container:
       image: ${{ matrix.metadata.container }}
       volumes:
-         - /usr/local:/host/usr/local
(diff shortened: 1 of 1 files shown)


marxin:
> I reckon it could slow down the jobs that populate that cache , but hopefully less than the other slowdowns?

Yes, I think the benefits overweight the costs.

Issue #6060, opened 2026-01-15
Support compressed packages in the download flow


theduke:
Add handling of compressed packages in the package download flow.

The code should:

* Send an `accept-encoding` header in the request
* Check the response `content-encoding` header 
* If present, decompress the downloaded bytes

Gzip and zstd compression algorithms should be supported.

PR #6069, opened 2026-01-15
Support compressed package downloads

theduke:
Add support for compressed package downloads, both in the CLI and
in WASIX.

Uses the builtin reqwest support in the CLI, and custom decoding logic in 
the WASIX BuiltinPackageLoader.

Resolves: #6060

### DIFF
--- a/lib/cli/Cargo.toml
+++ b/lib/cli/Cargo.toml
@@ -253,6 +253,7 @@ reqwest = { workspace = true, default-features = false, features = [
 	"json",
 	"multipart",
 	"gzip",
+	"zstd",
 ] }
 
 [target.'cfg(any(target_arch = "riscv64", target_arch = "loongarch64"))'.dependencies]
@@ -260,6 +261,7 @@ reqwest = { workspace = true, default-features = false, features = [
 	"native-tls",
 	"json",
 	"multipart",
+	"zstd",
 ] }
 
 
--- a/lib/wasix/src/runtime/package_loader/builtin_loader.rs
+++ b/lib/wasix/src/runtime/package_loader/builtin_loader.rs
@@ -1,11 +1,11 @@
 use std::{
     collections::HashMap,
-    io::{ErrorKind, Write as _},
+    io::{ErrorKind, Read, Write as _},
     path::PathBuf,
     sync::{Arc, RwLock},
 };
 
-use anyhow::{Context, Error};
+use anyhow::{Context, Error, bail};
 use bytes::Bytes;
 use http::{HeaderMap, Method};
 use tempfile::NamedTempFile;
@@ -283,6 +283,8 @@ impl BuiltinPackageLoader {
         }
 
         let body = response.body.context("package download failed")?;
+        let body = Self::decode_response_body(&response.headers, body)
+            .context("package download failed: could not decode response body")?;
         tracing::debug!(%url, "package_download_succeeded");
 
         let body = bytes::Bytes::from(body);
@@ -296,6 +298,15 @@ impl BuiltinPackageLoader {
         let mut headers = HeaderMap::new();
         headers.insert("Accept", "application/webc".parse().unwrap());
         headers.insert("User-Agent", USER_AGENT.parse().unwrap());
+
+        // Accept compressed responses.
+        // NOTE: gzip and zstd decoding is available on native platforms.
+        // In browser platforms, the fetch implementation should automatically
+        // handle decoding of gzip/zstd responses transparently.
+        headers.insert(
+            http::header::ACCEPT_ENCODING,
+            "zstd;q=1.0, gzip;q=0.8".parse().unwrap(),
+        );
 
         if url.has_authority()
             && let Some(token) = self.tokens.get(url.authority())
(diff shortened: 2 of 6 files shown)


lib/wasix/src/runtime/package_loader/builtin_loader.rs
```diff
@@ -298,6 +300,12 @@ impl BuiltinPackageLoader {
         headers.insert("Accept", "application/webc".parse().unwrap());
         headers.insert("User-Agent", USER_AGENT.parse().unwrap());
 
+        // Accept compressed responses.
+        // NOTE: gzip and zstd decoding is available on native platforms.
+        // In browser platforms, the fetch implementation should automatically
+        // handle decoding of gzip/zstd responses transparently.
+        headers.insert(http::header::ACCEPT_ENCODING, "gzip, zstd".parse().unwrap());
```
marxin: Should we use the qvalue weights to prefer ZSTD if possible?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Encoding#q

lib/wasix/src/runtime/package_loader/builtin_loader.rs
```diff
@@ -317,6 +325,53 @@ impl BuiltinPackageLoader {
...
         headers
     }
+
+    fn decode_response_body(headers: &HeaderMap, body: Vec<u8>) -> Result<Vec<u8>, anyhow::Error> {
+        let encodings = match headers.get(http::header::CONTENT_ENCODING) {
+            Some(header) => header
+                .to_str()
+                .context("invalid content-encoding header")?
+                .split(',')
+                .map(|encoding| encoding.trim().to_ascii_lowercase())
```
marxin, line 342: `filter_map(...).collect_vec()` ?

lib/wasix/src/runtime/package_loader/builtin_loader.rs
```diff
@@ -317,6 +325,53 @@ impl BuiltinPackageLoader {
...
+                .filter(|encoding| !encoding.is_empty())
+                .collect::<Vec<_>>(),
+            None => Vec::new(),
+        };
+
+        if encodings.is_empty() || (encodings.len() == 1 && encodings[0] == "identity") {
```
marxin, line 350: `exactly_one` from Itertools?

marxin (commented):
Can we write a test that will test the behavior?

lib/wasix/src/runtime/package_loader/builtin_loader.rs
```diff
@@ -317,6 +325,53 @@ impl BuiltinPackageLoader {
...
+                    #[cfg(not(target_arch = "wasm32"))]
+                    {
+                        reader = Box::new(
+                            zstd::stream::read::Decoder::new(reader)
+                                .context("failed to initialize zstd decoder")?,
+                        );
+                    }
+                    #[cfg(target_arch = "wasm32")]
+                    {
+                        bail!("zstd content-encoding is not supported on wasm32");
```
marxin, line 372: Theoretically we can rely on the Rust native implementation: https://docs.rs/ruzstd/latest/ruzstd/
However, it's abotu 30% slower (see: https://github.com/davidlattimore/wild/pull/126/changes). Though, don't know if it's worth the more complicated code.

theduke: I considered that initially.

Issues:

* reqwest uses zstd , so we'd either have duplicate deps or, a custom wasm32 path with ruzstd
* The zstd rust repo mentions way larger perfomance delta for highly compressed contents (up to 3.5x)
   In the browser we don't care, because modern browsers support zstd (except Safari... it'll get there), and on native we don't want to take the perf hit.

marxin: Fair enough! Let's not complicate the logic then ;)

lib/wasix/src/runtime/package_loader/builtin_loader.rs
theduke: Since the check is a bool condition, I don't think that would make the code more readable.

lib/wasix/src/runtime/package_loader/builtin_loader.rs
theduke: at_most_one might work well.

lib/wasix/src/runtime/package_loader/builtin_loader.rs
theduke: (and other WASM targets except the browser aren't relevant at the moment)

lib/wasix/src/runtime/package_loader/builtin_loader.rs
marxin: OK!

lib/wasix/src/runtime/package_loader/builtin_loader.rs
theduke: Good point. 

Server logic usually takes the best suited one anyway, and we will probably only support zstd on the server, but it's good to add anyway.

theduke:
Added a bunch of tests.

lib/wasix/src/runtime/package_loader/builtin_loader.rs
theduke: Done!

lib/cli/Cargo.toml
```diff
@@ -252,13 +252,15 @@ reqwest = { workspace = true, default-features = false, features = [
 	"json",
 	"multipart",
 	"gzip",
+	"zstd",
```
zebreus, line 256: This does not add zstd as a system dependency, so it's fine.

lib/wasix/src/runtime/package_loader/builtin_loader.rs
```diff
@@ -317,6 +328,56 @@ impl BuiltinPackageLoader {
 
         headers
     }
+
+    fn decode_response_body(headers: &HeaderMap, body: Vec<u8>) -> Result<Vec<u8>, anyhow::Error> {
```
zebreus, line 336: Is there a specific reason for this not being an associated function of Response.

lib/wasix/src/runtime/package_loader/builtin_loader.rs
zebreus: `matches!(encodings.as_slice(), [] | ["identity"])`

theduke: That one doesn't work , unfortunately, because it's String, not &str.

One of the annoying gotchas, I stumble over that all the time.

theduke: Let's just leave it as is, no need to over-complicate.

theduke:
Added some small doc comments, squashed commits. 

Good to go.

Issue #5977, opened 2025-12-16
Wasmer's LLVM compiler generates code to pass 64-bit value to `memory_copy`


wakabat:
Hey so I was testing the latest wasmer code on Rust 1.91.1 in RISC-V 64 bit architecture. And I noticed that wasmer would throw `HeapAccessOutOfBounds` error randomly. Upon further investigation, I noticed that the [checked_add](https://github.com/wasmerio/wasmer/blob/78ecea6d2cb08ef3343dcee87a00419ff39e1599/lib/vm/src/vmcontext.rs#L327) in `wasmer_vm::vmcontext::memory_copy` would overflow, but the passed values are small and should not really overflow.

Unfortunately my code is deeply coupled in a complex project so I don't have a reproducible test case, but I did managed to dump some logs with notes for you to check out: https://gist.github.com/wakabat/372009f1adf5f16eb5e121013016e485 And I'm happy to provide more information if needed.

I have to mention that I don't really have a need for Rust 1.91.1, it just happened that we have Rust 1.91.1 handy for our architecture when this quest started. I tested a series of Rust compilers, Rust 1.89.0 and Rust 1.90.0 are fine, but Rust 1.91.1 would trigger the bug. And as the note hints, I think this problem has existed in wasmer for some time, it's just Rust 1.91.1 finally exposes the issue to us.

I have to say that I'm not 100% sure that this is an issue with wasmer, it could be that I'm missing something in our code. Either case, we could use a little help here. Let us know what you think, many thanks!

FYI: I actually tried modifying code [here](https://github.com/wasmerio/wasmer/blob/78ecea6d2cb08ef3343dcee87a00419ff39e1599/lib/compiler-llvm/src/translator/code.rs#L11997-L12022), adding LLVM code pieces to truncate / mask `len` to a 32-bit value, but this does not really work. Apart from handling `Operator::MemoryCopy`, is there other part of the code that would invoke `wasmer_vm_memory32_copy`?

marxin:
Thanks for the report! Are you talking about running the native rustc compliler on the `riscv64gc` platform? If so, can you try running the Wasmer's LLVM compiler tests with the mentioned revision?

It would be great if you could come up with a smaller reproducer.

wakabat:
Okay so I still don't have a reproducible case(maybe I can have one when we finished cleaning up our WIP code and open source it properly), but I have a theory of what is going on here.

[...]

Issue #5943, opened 2025-12-03
missing host-fn support for multi-value


marxin:
Right now, the multi-value support is not properly covered when it comes to host function that return multiple arguments, e.g. `(i32, i32, i32)` in Rust language.

More precisely, we're missing the full support for **Singlepass**, **Cranelift** and the **LLVM+Windows** targets. The #5936 PR included intensive set of tests that stress the functionality (LLVM+linux and LLVM+macos-arm work correctly).

For the Cranelift, we have to implement the ABI ourselves (using `StructReturn` type[^1]), similarly to what we do for LLVM: https://github.com/wasmerio/wasmer/blob/main/lib/compiler-llvm/src/abi/x86_64_systemv.rs.

[^1]: https://docs.rs/alexcrichton-cranelift-codegen/latest/cranelift_codegen/ir/enum.ArgumentPurpose.html#variant.StructReturn

syrusakbary:
Once we fix multi-value, we can fix the "typed-function" implementation to use the native function trampolines, instead of the dynamic function ones.

https://github.com/wasmerio/wasmer/blob/main/lib/api/src/backend/sys/entities/function/typed.rs#L107-L117

https://github.com/wasmerio/wasmer/blob/main/lib/api/src/backend/sys/entities/function/typed.rs#L234-L244

marxin:
Moreover, `loongarch64-linux` and `riscv64gc-linux` targets are affected as well.

marxin:
The issue has been addressed in Cranelift and LLVM for all supported targets. Since full Singlepass support would effectively require a complete overhaul of the compiler, we’ll mark the feature as experimental and close this issue.

PR #6072, opened 2026-01-16
fix(LLVM,riscv64): fix calling conventions for `i32` type

marxin:
There's a special requirement for the RISC-V SysV ABI:
https://five-embeddev.com/riscv-user-isa-manual/Priv-v1.12/rv64.html
> The compiler and calling convention maintain an invariant that all 32-bit values are held in a sign-extended format in 64-bit registers.
> Even 32-bit unsigned integers extend bit 31 into bits 63 through 32. Consequently, conversion between unsigned and signed 32-bit integers
> is a no-op, as is conversion from a signed 32-bit integer to a signed 64-bit integer.

So far, we've been using `X86_64SystemV` for the RISC-V ABI, and so I copied it into a separate one, where extra param attributes are added. That's something we must address as part of #5943. On top of that I decided to put some common functionality into the `Abi` trait itself.

@wakabat can you give it a try?

Fixes: #5977


lib/compiler-llvm/src/translator/intrinsics.rs
```diff
@@ -480,16 +487,46 @@ impl<'ctx> Intrinsics<'ctx> {
...
+                            );
+                            function.add_attribute(
+                                AttributeLoc::Param(i as u32),
+                                context.create_enum_attribute(
+                                    Attribute::get_named_enum_kind_id("noundef"),
+                                    0,
+                                ),
+                            );
+                        }
+                    }
```
syrusakbary, line 511: I find a bit strange that this logic lives here, it feels a bit of a hack.
Do you think this could be improved?

marxin: Good point! We can do better by using the https://thedan64.github.io/inkwell/inkwell/intrinsics/struct.Intrinsic.html#method.find API, that should give us a properly annotated declaration that hopefully takes target ABI into account.

wakabat:
Hey thanks for this change! I have run all the tests on our side using Rust 1.92.0, and they all pass. So as far as our tests, this change did resolve the bug in #5977

marxin:
Cool, thanks for the confirmation.

syrusakbary (approved):

Issue #6073, opened 2026-01-18
poll_oneoff hangs forever when nsubscriptions == 0


artemyarulin:
poll_oneoff hangs forever when nsubscriptions == 0. With no subscriptions and no timeout, the poll loop can never resolve, so callers block indefinitely instead of getting an error. Expected behavior is to return EINVAL for empty subscription lists.

```c
#include <assert.h>
#include <pthread.h>
#include <sched.h>
#include <stdatomic.h>
#include <stdbool.h>
#include <wasi/api_wasi.h>

int main(void) {
    __wasi_subscription_t sub = {0};
    __wasi_event_t out = {0};
    __wasi_size_t events = 0;
    const __wasi_size_t nsubscriptions = 0; // Passing a 0 causes an infinite loop
    __wasi_errno_t ret = __wasi_poll_oneoff(&sub, &out, nsubscriptions, &events);
    assert(ret == __WASI_ERRNO_INVAL && "poll_oneoff(0) should return EINVAL");
}
```

Found via fuzzing

PR #6074, opened 2026-01-18
fix(wasix): handle empty subscription list in poll_oneoff

artemyarulin:
Fixes https://github.com/wasmerio/wasmer/issues/6073

### DIFF
--- a/lib/wasix/src/syscalls/wasi/poll_oneoff.rs
+++ b/lib/wasix/src/syscalls/wasi/poll_oneoff.rs
@@ -64,12 +64,12 @@ pub fn poll_oneoff<M: MemorySize + 'static>(
     nsubscriptions: M::Offset,
     nevents: WasmPtr<M::Offset, M>,
 ) -> Result<Errno, WasiError> {
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
     // An empty subscription list would otherwise block forever in the poll loop.
     if nsubscriptions == M::ZERO {
         return Ok(Errno::Inval);
     }
-
-    wasi_try_ok!(WasiEnv::process_signals_and_exit(&mut ctx)?);
 
     ctx = wasi_try_ok!(maybe_backoff::<M>(ctx)?);
     ctx = wasi_try_ok!(maybe_snapshot::<M>(ctx)?);


lib/wasix/src/syscalls/wasi/poll_oneoff.rs
```diff
@@ -64,6 +64,11 @@ pub fn poll_oneoff<M: MemorySize + 'static>(
     nsubscriptions: M::Offset,
     nevents: WasmPtr<M::Offset, M>,
 ) -> Result<Errno, WasiError> {
+    // An empty subscription list would otherwise block forever in the poll loop.
```
zebreus, line 69: There should probably be a `WasiEnv::do_pending_operations(&mut ctx)? even if we do this early return, so signals and linker synchronization are still advanced even if the guest loop-polls on an empty set.

lib/wasix/src/syscalls/wasi/poll_oneoff.rs
```diff
@@ -64,6 +64,11 @@ pub fn poll_oneoff<M: MemorySize + 'static>(
...
+    if nsubscriptions == M::ZERO {
+        return Ok(Errno::Inval);
```
zebreus, line 71: I think there's a guarantee that a successful poll_oneoff always returns at least one event. If that is correct, then `Inval` is correct here.

zebreus (changes requested):
Looks like a good change, however this changes the existing behaviour of the poll_oneoff API, so we need to be a bit careful.

It seems to me that previously, the API could be used with 0 subscriptions as a way to block indefinitely until a signal happens. I don't think that was intended or defined behaviour, so this change should be fine.

We need to make an accompanying change in wasix-libc to catch that case early and return EINVAL there without a syscall.

We should also run linker synchronization in the early return case here, just in case.

lib/wasix/src/syscalls/wasi/poll_oneoff.rs
zebreus: The behaviour and `EINVAL` is correct here, see https://github.com/WebAssembly/WASI/pull/193

zebreus:
`wasix-libc` is already ready for the behaviour added in this PR.

Looking at the code, I realized we are handling signals around poll incorrectly. I will open a separate issue for that.

lib/wasix/src/syscalls/wasi/poll_oneoff.rs
```diff
@@ -64,6 +64,11 @@ pub fn poll_oneoff<M: MemorySize + 'static>(
...
+    }
+
     wasi_try_ok!(WasiEnv::process_signals_and_exit(&mut ctx)?);
```
zebreus: >There should probably be a `WasiEnv::do_pending_operations(&mut ctx)? even if we do this early return, so signals and linker synchronization are still advanced even if the guest loop-polls on an empty set.

@Arshia001  Is there a reason not to use `do_pending_operations` instead of `process_signals_and_exit` here, I am missing?

```suggestion
    WasiEnv::do_pending_operations(&mut ctx)?;

    // An empty subscription list would otherwise block forever in the poll loop.
    if nsubscriptions == M::ZERO {
        return Ok(Errno::Inval);
    }
```

lib/wasix/src/syscalls/wasi/poll_oneoff.rs
zebreus: Added a suggested fix for this in another comment that highlights the whole section

zebreus (changes requested):

lib/wasix/src/syscalls/wasi/poll_oneoff.rs
```diff
@@ -64,7 +64,12 @@ pub fn poll_oneoff<M: MemorySize + 'static>(
     nsubscriptions: M::Offset,
     nevents: WasmPtr<M::Offset, M>,
 ) -> Result<Errno, WasiError> {
-    wasi_try_ok!(WasiEnv::process_signals_and_exit(&mut ctx)?);
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
```
zebreus, line 68: @Arshia001 We now also do linker synchronization here

zebreus: arshia also does not know more about this

PR #6075, opened 2026-01-19
Experimental support for interruptable wasm computation

Arshia001:
This PR adds a `Store::interrupter()` method, which returns an `Interrupter` instance that can then be used to kill running WASM instances.

This works by raising a `SIGUSR` signal against the thread that's running the store's instance, which is then caught and transformed into a `HostInterrupt` trap.

If the instance happens to be calling an imported function when the signal is raised, the trap will happen as soon as the imported function returns.

The implementation correctly interacts with the async runtime; when interrupting an async store, all futures spawned from `call_async` on that store's instances will be terminated with the `HostInterrupt` trap.

The implementation is gated behind the `experimental-host-async` feature flag.

### DIFF
--- a/lib/vm/src/interrupt_registry/unix.rs
+++ b/lib/vm/src/interrupt_registry/unix.rs
@@ -20,7 +20,12 @@ struct StoreInterruptState {
     /// The pthread of the thread the store is running on, used to
     /// send the interrupt signal. Note that multiple stores may
     /// be executing WASM code within the same OS thread.
-    pthread: libc::pthread_t,
+    ///
+    /// We store this as a plain integer because `libc::pthread_t` is a raw
+    /// pointer on some Unix targets, which would make the global `DashMap`
+    /// fail its `Send` bounds even though we only treat the value as an opaque
+    /// thread identifier.
+    pthread: usize,
     /// Whether this store was interrupted.
     interrupted: bool,
     /// See comments in [`ThreadInterruptState`].
@@ -89,7 +94,7 @@ pub fn install(store_id: StoreId) -> Result<InterruptInstallGuard, InstallError>
         // TODO: isn't there a way to get this without reaching for libc APIs?
         // Since stores can't be sent across threads once they start executing code,
         // we don't need to update this value for recursive calls.
-        let pthread = unsafe { libc::pthread_self() };
+        let pthread = unsafe { libc::pthread_self() as usize };
 
         StoreInterruptState {
             pthread,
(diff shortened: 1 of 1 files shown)


zebreus:
That's a nice way to implement interruptions on `sys`.

marxin:
A general question: do you want to have it included in 7.0 release (under the experimental feature gate)? Or is it something we don't have to push now?

lib/api/src/backend/sys/entities/function/mod.rs
```diff
@@ -513,6 +515,15 @@ impl Function {
         // Call the trampoline.
         let result = {
             let store_id = store.objects_mut().id();
+
+            #[cfg(feature = "experimental-host-interrupt")]
+            let interrupt_guard = match interrupt_registry::install(store_id) {
```
theduke, line 517: Putting this behind a #[cfg] is a bit awkard, since it requires re-compilation.

Could we instead use a setting?
Eg configured through the existing capabilities?

lib/api/src/backend/sys/entities/function/mod.rs
```diff
@@ -850,13 +863,29 @@ where
         // AS WE ARE IN THE WASM STACK, NOT ON THE HOST ONE.
         // See: https://github.com/wasmerio/wasmer/pull/5700
         match result {
-            Ok(InvocationResult::Success(())) => {}
+            Ok(InvocationResult::Success(())) => unsafe {
+                // Note: can't acquire a proper ref-counted context ref here, since we can switch
```
theduke, line 864: Seems like the whole block should be behind the interrupt check?

lib/api/src/backend/sys/entities/function/mod.rs
```diff
@@ -964,17 +993,27 @@ macro_rules! impl_host_function {
                     Ok(InvocationResult::Success(result)) => unsafe {
                         // Note: can't acquire a proper ref-counted context ref here, since we can switch
                         // away from the WASM stack at any time.
-                        // Safety: The pointer is only used for the duration of the call to `into_c_struct`.
+                        // Safety: The pointer is only used for the duration of the call to
+                        // `into_c_struct` and `get_current_transient`.
                         let mut store_wrapper = StoreContext::get_current_transient(env.store_id);
                         let mut store = store_wrapper.as_mut().unwrap();
+                        #[cfg(feature = "experimental-host-interrupt")]
```
theduke, line 997: Repeating the same code 4 times isn't so great, could be an extra check up front?

marxin:
a follow-up fix: #6378

Arshia001:
Thanks for taking the time to update the code, @marxin!

Issue #6031, opened 2026-01-09
tests/wast/spec/f32.wast fails with LLVM on rv64gc target


marxin:
Likely started failing after #5796 got merged into main.

```
---- wast::spec::f32::llvm::llvm stdout ----
Running wast `tests/wast/spec/f32.wast`
Error: Failed directives on tests/wast/spec/f32.wast:
  • expected Core(F32(ArithmeticNan)), got F32(NaN) (4288675840) (2456:1)
  • expected Core(F32(ArithmeticNan)), got F32(NaN) (2141192192) (2458:1)
  • expected Core(F32(ArithmeticNan)), got F32(NaN) (4288675840) (2476:1)
  • expected Core(F32(ArithmeticNan)), got F32(NaN) (2141192192) (2478:1)
  • expected Core(F32(ArithmeticNan)), got F32(NaN) (4288675840) (2496:1)
  • expected Core(F32(ArithmeticNan)), got F32(NaN) (2141192192) (2498:1)
  • expected Core(F32(ArithmeticNan)), got F32(NaN) (4288675840) (2516:1)
  • expected Core(F32(ArithmeticNan)), got F32(NaN) (2141192192) (2518:1)


---- wast::spec::f64::llvm::llvm stdout ----
Running wast `tests/wast/spec/f64.wast`
Error: Failed directives on tests/wast/spec/f64.wast:
  • expected Core(F64(ArithmeticNan)), got F64(NaN) (18443366373989023744) (2456:1)
  • expected Core(F64(ArithmeticNan)), got F64(NaN) (9219994337134247936) (2458:1)
  • expected Core(F64(ArithmeticNan)), got F64(NaN) (18443366373989023744) (2476:1)
  • expected Core(F64(ArithmeticNan)), got F64(NaN) (9219994337134247936) (2478:1)
  • expected Core(F64(ArithmeticNan)), got F64(NaN) (18443366373989023744) (2496:1)
  • expected Core(F64(ArithmeticNan)), got F64(NaN) (9219994337134247936) (2498:1)
  • expected Core(F64(ArithmeticNan)), got F64(NaN) (18443366373989023744) (2516:1)
  • expected Core(F64(ArithmeticNan)), got F64(NaN) (9219994337134247936) (2518:1)
```

PR #6079, opened 2026-01-19
fix(LLVM): RISC-V rounding operations need an extra check

marxin:
Similarly to RISC-V Singlepass implementation, trunc/ceil/nearest rounding operations need a special check for NaN values:
https://github.com/wasmerio/wasmer/blob/840cfdfdb261628572f66ea98e8339862550b450/lib/compiler-singlepass/src/machine_riscv.rs#L803-L812

Fixes: #6031

### DIFF
--- a/lib/compiler-llvm/src/translator/code.rs
+++ b/lib/compiler-llvm/src/translator/code.rs
@@ -2080,12 +2080,11 @@ impl<'ctx> LLVMFunctionCodeGenerator<'ctx, '_> {
                     ty,
                     "canonical_nan",
                 ));
-                let res = err!(self.builder.build_select(
-                    is_nan,
-                    canonical_nan,
-                    value,
-                    "canonical_nan",
-                ));
+                let res =
+                    err!(
+                        self.builder
+                            .build_select(is_nan, canonical_nan, value, "canonical_nan",)
+                    );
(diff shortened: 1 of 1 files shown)


lib/compiler-llvm/src/translator/code.rs
```diff
@@ -2035,6 +2044,123 @@ impl<'ctx> LLVMFunctionCodeGenerator<'ctx, '_> {
         }
     }
 
+    fn finalize_rounding_result(
+        &self,
+        value: BasicValueEnum<'ctx>,
+        info: ExtraInfo,
+    ) -> Result<(BasicValueEnum<'ctx>, ExtraInfo), CompileError> {
+        let ty = value.get_type();
```
syrusakbary, line 2052: What if the canonicalization is not active? This seems to me that no matter if that setting it's true or not, it will always do the extra compute?

marxin: Yes, unfortunately on RISC-V, the `@llvm.floor.f32` intrinsic cannot handle the following SPEC test:
`(assert_return (invoke "floor" (f32.const -nan:0x200000)) (f32.const nan:arithmetic))`
and so, if we want to fully support specification, we need to do the extra check.

PR #6080, opened 2026-01-19
chore: drop `xxhash` hashing algorithm

marxin:
Based on the discussion with @theduke, another hashing algorithm:

> It was introduced to speed up some flows in WASIX - not exactly with my support.
> Having multiple hash algos makes things very complex and error-prone.

The only leftover is `ModuleHash` in the Artifact format. We need it for a compatibility reasons.

@Arshia001 @zebreus The last remaining usage is:
```
pub fn from_path(str: &str) -> Self {
        Inode(xxhash_rust::xxh64::xxh64(str.as_bytes(), 0))
    }
```
Do you want to keep it, or are you fine with sticking to the existing `sha256`?


theduke:
Some context:

Shared compute environments , or in the future, our own CLI with a public shared artifact cache, can not use a weak, non-cryptographic hash like xxhash for security reasons.

In addition, having multiple hashes in the code makes it very easy to break shared caching in some places if the environment uses a different hash for storing artifacts - we've had this problem multiple times on Edge.

From that perspective, taking the performance hit and removing xxhash makes a lot of sense.

We have also done quite a bit of work to take hashes from metadata if the origin is a package (which it almost always is), reducing the performance impact of more expensive hashes a lot.

zebreus:
I am not particularly attached to xxhash. In the snippet you posted, it might even be dangerous. SHA256 seems like a good replacement.

lib/wasix/src/fs/mod.rs
```diff
@@ -121,7 +122,13 @@ impl Inode {
     }
 
     pub fn from_path(str: &str) -> Self {
-        Inode(xxhash_rust::xxh64::xxh64(str.as_bytes(), 0))
+        let digest = sha2::Sha256::digest(str.as_bytes());
+        // SAFETY: we take only 1/4 of the output digest
```
marxin: @zebreus are you fine with taking just 1/4 of the hash. Or should I widen the type of `Inode`?

zebreus: That's a fair consideration. I don't think we can easily widen the size of an inode.

But with only 64bits we don't benefit from the security of sha256, so it is probably best to leave this as an xxhash.

marxin: All right, works for me.

marxin:
I think @theduke is aligned with the change - waiting for @syrusakbary for the approval connected to our API.

syrusakbary:
Note: do we know that all the webcs don't have any xxhash inside? I assume not, but good to triple chck

marxin:
> Note: do we know that all the webcs don't have any xxhash inside? I assume not, but good to triple chck

@theduke Can you check it, please?

Issue #6076, opened 2026-01-19
macos-arm64: random exception handling failures under `cargo test`


marxin:
I noticed a random failures if I replace the default `cargo nextest` harness with the classical `cargo test` on MacMini M4:

```
$ cargo t --features=llvm exception
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.13s
     Running tests/compilers/main.rs (target/debug/deps/compilers-f11f4309f34d0a2c)

running 6 tests
test wast::spec::exception_handling::tag::llvm::llvm ... ok
test issues::issue_6004_exception::llvm::llvm ... ok
test wast::spec::exception_handling::throw::llvm::llvm ... ok
error: test failed, to rerun pass `--test compilers`

Caused by:
  process didn't exit successfully: `/Users/apple/Programming/wasmer/target/debug/deps/compilers-f11f4309f34d0a2c exception` (signal: 5, SIGTRAP: trace/breakpoint trap)
...
$ cargo t --features=llvm issues
failures:

---- issues::issue_6004_exception::llvm::llvm stdout ----

thread 'issues::issue_6004_exception::llvm::llvm' panicked at tests/compilers/issues.rs:697:10:
called `Result::unwrap()` on an `Err` value: RuntimeError { source: Sys(UncaughtException { exnref: VMExceptionRef(StoreHandle { id: StoreId(9), internal: 1 }), backtrace:    0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
 }), wasm_trace: [FrameInfo { module_name: "<module>", func_index: 0, function_name: None, func_start: SourceLoc(4294967295), instr: SourceLoc(4294967295) }] }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

@syrusakbary can you try reproducing it on your laptop?

syrusakbary:
Just got this:
```
running 6 tests
test wast::spec::exception_handling::tag::llvm::llvm ... ok
test issues::issue_6004_exception::llvm::llvm ... ok
test wast::spec::exception_handling::throw_ref::llvm::llvm ... ok
test wast::spec::exception_handling::throw::llvm::llvm ... ok
test wast::spec::exception_handling::easy_throw::llvm::llvm ... ok
fatal runtime error: Rust cannot catch foreign exceptions, aborting
error: test failed, to rerun pass `--test compilers`

Caused by:
[...]

PR #6087, opened 2026-01-21
fix(LLVM,macOS): fix compact unwinding bookkeeping

marxin:
There were 3 fundamental issues with the macOS compact unwind info:
1. we register a global callback (for all threads) that is responsible for frame lookup - `find_dynamic_unwind_sections`; we forgot to remove entries if a module is unregistered (basically use-after-free type of error)
2. there was quite some dead code in `systemv.rs` for compact unwinding -> fixed and made it `#[cfg]` based; doing that one can't basically call a bad function
3. `unsafe { UNWIND_INFO.try_lock() };` - having `static mut` let to races even with the previous 2 bullet points addressed

Note there's a still one missing TODO about a better data structure, will address that in a separate PR. We should not do a linear scan of an address in `find_dynamic_unwind_sections`: O(N).

Fixes: #6076

### DIFF
--- a/lib/compiler/src/engine/artifact.rs
+++ b/lib/compiler/src/engine/artifact.rs
@@ -467,9 +467,7 @@ impl Artifact {
             )?;
         }
         #[cfg(any(target_os = "linux", all(windows, target_arch = "x86_64")))]
-        if let Some(eh_frame) = eh_frame {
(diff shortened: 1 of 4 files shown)


syrusakbary (approved):
I have run `cargo t --features=llvm exception` in my mac, and everything passes now.

Lets move this forward!

PR #6091, opened 2026-01-21
fix: compact unwind info - use rangemap

marxin:
For a faster address look-up in `find_dynamic_unwind_sections`, we should use something better than O(N) linear scan.


syrusakbary (approved):
Way better

theduke:
Note: was still reviewing this, had some minor comments...

PR #6102, opened 2026-01-22
fix(macos-x64): restore unwind info

marxin:


.github/workflows/test.yaml
```diff
@@ -947,7 +947,8 @@ jobs:
         shell: bash
         run: make ${{ matrix.stage.make }}
         # For MacOS x64 - run tests only on main or release branches
-        if: github.event_name != 'pull_request' || startsWith(github.head_ref, 'release-') || matrix.metadata.os != 'macos-15-intel'
+        #  || matrix.metadata.os != 'macos-15-intel'
+        # if: github.event_name != 'pull_request' || startsWith(github.head_ref, 'release-')
```
syrusakbary: We can just delete this instead of commenting it

marxin: I am going to move the piece back (tested only on `main`, not as part of  a PR).

lib/compiler/src/engine/artifact.rs
```diff
@@ -462,7 +462,10 @@ impl Artifact {
                 get_got_address(RelocationTarget::LibCall(wasmer_vm::LibCall::EHPersonality)),
             )?;
         }
-        #[cfg(any(target_os = "linux", all(windows, target_arch = "x86_64")))]
+        #[cfg(all(
+            not(target_arch = "wasm32"),
+            not(all(target_os = "macos", target_arch = "aarch64"))
+        ))]
```
syrusakbary: I wonder if changing a bit the logic would make it a bit more readable. Thoughts?

```suggestion
        #[cfg(not(any(
            target_arch = "wasm32",
            all(target_os = "macos", target_arch = "aarch64")
        )))]
```

Issue #5831, opened 2025-10-24
Re-enable `oss-fuzz` again


marxin:
The fuzzing was disable in de8811cb2fb5f86ca16c11e6abb13da9cf7428e2, consider enabling it again in the future.

marxin:
I updated the dependencies in the OSS-fuzz repository: https://github.com/google/oss-fuzz/pull/14771.

But it seems I face a linker errors (LLVM vs. GCC C++ library) when running inside Docker container.

PR #6103, opened 2026-01-22
CI: enable fuzzing again

marxin:
Resolves: #5831


marxin:
Will return to it at some point..

PR #6096, opened 2026-01-22
build: Switch back to plain rust-toolchain file

theduke:
To prevent downloading redundant components or targets in CI or other
environments, and to keep things simple, switch back to a plain rust-toolchain
file over the recently introduced rust-toolchain.toml.

Closes #6095

PR #6051, opened 2026-01-13
Add Rust tools to Nix development environment

Twey:
# Description

Currently the Nix expression doesn't include the Rust tools that are needed to build Wasmer.

Use [oxalica/rust-overlay](https://github.com/oxalica/rust-overlay) to interpret the rustup `rust-toolchain.toml` file and provide the correct toolchain, with support for native and `wasm32-unknown-unknown` compilation, in the development shell.

There's also a Nix package in `scripts/nix` that looks like it needs some love, but I didn't tackle it here.

PR #6110, opened 2026-01-23
Switch from legacy `rust-toolchain` format

Twey:
# Description

https://github.com/wasmerio/wasmer/pull/6051 switched the Rust toolchain format to the newer TOML syntax and added development tools so that they'd be included in the development environment described in `flake.nix`.  However, this resulted in them being installed in CI contexts in which they aren't needed.

https://github.com/wasmerio/wasmer/pull/6096 fixed this by moving the tools to `flake.nix`, but also reverted the `rust-toolchain.toml` file to the legacy `rust-toolchain` syntax.

This PR reverts that change, keeping the newer TOML format, and just moves the tools from the `rust-toolchain.toml` to the `flake.nix` using the [override functionality](https://github.com/oxalica/rust-overlay?tab=readme-ov-file#cheat-sheet-common-usage-of-rust-bin) from [`oxalica/rust-overlay`](https://github.com/oxalica/rust-overlay).

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,3 @@
 [toolchain]
 channel = "1.90"
+components = ["rustfmt", "clippy"]


Twey:
It looks like some of those components are actually required by CI anyway.  I've added them back.

rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.90"
```
marxin, line 2: Bump to `1.91` ;)

marxin:
Note the PR conflicts with #6107.

PR #6114, opened 2026-01-26
fix(LLVM): fixed ABI issues noticed while fuzzing

marxin:


tests/ignores.txt
```diff
@@ -59,13 +59,13 @@ llvm+riscv64      wasmer::nan_canonicalization::llvm::universal
...
 #6032
 llvm+riscv64      wasmer::stack_overflow_sret::llvm
-llvm+loongarch64      wasmer::stack_overflow_sret::llvm
 
 # riscv support on Cranelift is also very young
 cranelift+riscv64 spec::align::cranelift
 cranelift+riscv64 spec::memory_copy::cranelift
 cranelift+riscv64 spec::memory_trap::cranelift
 cranelift+riscv64 spec::r#if::cranelift
+cranelift+riscv64 wasmer::exception_handling
```
syrusakbary, line 68: I assume this will be fixed somewhere else?

marxin: Yeah, it's more about a missing `simd` (`v128`) support, nothing much connected to exceptions to be honest. Will split the test-case in a new PR.

PR #6119, opened 2026-01-26
feat(LLVM): use verbose asm output for debug output

marxin:
Makes the assembly more readable.

### DIFF
--- a/lib/compiler-llvm/src/config.rs
+++ b/lib/compiler-llvm/src/config.rs
@@ -113,6 +113,7 @@ pub struct LLVM {
     pub(crate) middlewares: Vec<Arc<dyn ModuleMiddleware>>,
     /// Number of threads to use when compiling a module.
     pub(crate) num_threads: NonZero<usize>,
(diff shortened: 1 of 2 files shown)


lib/compiler-llvm/src/config.rs
```diff
@@ -335,9 +335,11 @@ impl LLVM {
         if let Architecture::Riscv64(_) = triple.architecture {
             llvm_target_machine_options = llvm_target_machine_options.set_abi("lp64d");
         }
-        llvm_target
+        let target_machine = llvm_target
             .create_target_machine_from_options(&target_triple, llvm_target_machine_options)
-            .unwrap()
+            .unwrap();
+        target_machine.set_asm_verbosity(true);
```
theduke: Could this have negative consequences for perf?
Like generating lots more output?

Or is this only applicable when debug dumping?

marxin: Yeah, good point, let me enable it only if `--compiler-debug-dir` is used!

PR #6120, opened 2026-01-26
chore(fuzzing): tweak options and save file early

marxin:


marxin:
> Why disable relaxed simd?

We don't support it yet: https://webassembly.org/features/

Issue #6126, opened 2026-01-27
Mounted FS read_dir drops mount prefix


artemyarulin:
When a filesystem is mounted at a guest path, `read_dir()` returns entries without the mount prefix (e.g. `/file.txt` instead of `/mnt/file.txt`). This breaks overlay dedupe and results in duplicate directory entries.

```rust
use virtual_fs::{mem_fs::FileSystem as MemFS, FileSystem};

let fs = MemFS::default();
let mounted = MemFS::default();
virtual_fs::ops::touch(&mounted, "/file.txt").unwrap();

fs.mount("/mnt".into(), &mounted, "/".into()).unwrap();

let entries: Vec<_> = fs
    .read_dir("/mnt".as_ref())
    .unwrap()
    .map(|e| e.unwrap().path)
    .collect();

assert_eq!(entries, vec!["/mnt/file.txt".into()]); // actual: "/file.txt"

PR #6127, opened 2026-01-27
fix(virtual-fs): Fix read_dir missing mount prefix

artemyarulin:
When a filesystem is mounted at a guest path, read_dir() returns entries without the mount prefix (e.g. /file.txt instead of /mnt/file.txt). This breaks overlay dedupe and results in duplicate directory entries.

Fixes https://github.com/wasmerio/wasmer/issues/6126

### DIFF
--- a/lib/virtual-fs/src/overlay_fs.rs
+++ b/lib/virtual-fs/src/overlay_fs.rs
@@ -1131,8 +1131,43 @@ mod tests {
     use std::path::PathBuf;
 
     use super::*;
-    use crate::mem_fs::FileSystem as MemFS;
+    use crate::{TmpFileSystem, mem_fs::FileSystem as MemFS};
     use tokio::io::{AsyncReadExt, AsyncSeekExt, AsyncWriteExt};
+
+    #[test]
+    fn overlay_read_dir_rebases_mounted_entries() {
+        let primary = TmpFileSystem::new();
+        ops::create_dir_all(&primary, "/app").unwrap();
+
+        let volume = MemFS::default();
+        ops::create_dir_all(&volume, "/themes/twentytwentyfour").unwrap();
+
+        let container = MemFS::default();
+        ops::create_dir_all(&container, "/app/wp-content/themes/twentytwentyfour").unwrap();
+
+        let volume: Arc<dyn FileSystem + Send + Sync> = Arc::new(volume);
+        primary
+            .mount(
+                PathBuf::from("/app/wp-content"),
+                &volume,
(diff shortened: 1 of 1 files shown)


artemyarulin:
The bug is in `mem_fs::read_dir()` when it hits a mounted filesystem: it returns entries from the mounted FS namespace (`/themes/...`) instead of rebasing them to the guest mount path (`/app/wp-content/themes/...`). The bug has existed for years (bisect points somewhere around ~2024), but it was mostly invisible because the container filesystem was merged into the root, so there wasn’t a second layer under like `/app/wp-content` to collide with.

PR #5799 changed the layering to keep the container as an overlay secondary, which made both layers visible and exposed the duplicates. This mirrors what we saw in WordPress:

```rust
let root = TmpFileSystem::new();              // mem_fs root
let volume = MemFS::default();                // “shared volume” FS
ops::create_dir_all(&volume, "/themes/twentytwentyfour").unwrap();
root.mount("/app/wp-content".into(), &Arc::new(volume), "/".into()).unwrap();

let container = MemFS::default();             // “container” FS
ops::create_dir_all(&container, "/app/wp-content/themes/twentytwentyfour").unwrap();

let overlay = OverlayFileSystem::new(root, [container]);

// BUG: read_dir returns both paths because mount entries are unrebased
let names: Vec<_> = overlay
    .read_dir("/app/wp-content/themes".as_ref())
    .unwrap()
    .map(|e| e.unwrap().file_name().to_string_lossy().to_string())
    .collect();

assert_eq!(names, vec!["twentytwentyfour", "twentytwentyfour"]); // Without fix we got duplicates
```

We now have two tests that cover different layers, so they behave differently in relation to #5799:

`read_dir_rebases_mount_paths`: directly checks the `mem_fs` mount behavior — `when I mount /app/wp-content, do read_dir() entries come back with the /app/wp-content prefix?` This would fail even before #5799 because the underlying bug already existed.

`overlay_read_dir_rebases_mounted_entries`: exercises the WordPress‑like layering (root FS + mounted shared volume + container FS).

Before #5799, the container FS was effectively merged into the root, so there wasn’t a second visible layer at `/app/wp-content`. The unrebased `/themes/...` path didn’t collide → no duplicates → test would pass → no WP issue.

After #5799, the container FS is kept as an overlay secondary, so it overlaps `/app/wp-content`. The unrebased `/themes/...` entry collides → duplicates appear → test fails → WP duplicate issue.

So #5799 didn’t introduce the bug, it simply uncovered a long‑standing issue that was there for a long time.

PR #6134, opened 2026-01-28
Switch from legacy `rust-toolchain` format

marxin:
V2 of #6110.

### DIFF
--- a/tests/integration/cli/Cargo.toml
+++ b/tests/integration/cli/Cargo.toml
@@ -10,7 +10,7 @@ publish = false
 [dev-dependencies]
 rand.workspace = true
 target-lexicon.workspace = true
-serde.workspace = true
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -57,7 +53,10 @@
 
 
             # Rust tooling
-            rust-toolchain
+            (rust-toolchain.override {
+              targets = [ "wasm32-unknown-unknown" ];
+              extensions = [ "clippy" "rustfmt" "rust-analyzer" "rust-src" ];
+            })
 
             # Snapshot testing
```
zebreus, line 61: I think this should work, but haven't tested it.

Twey: I have :)

Twey:
Thanks for polishing this up!  I didn't have much free time last week.

PR #6138, opened 2026-01-29
CI: include smoketests for final release binaries

marxin:
Based on the latest experience of final binaries crashing with LLVM, I would like to introduce smoke tests.
CC: @artemyarulin


theduke:
Note: we already have CLI tests that run a bunch of things.

I forgot where they are located , but this would probably fit in better there.

marxin:
Closing as the right place for the tests is the `wasmer-integration-tests-cli` project under the `wasmer` repository.

Issue #4169, opened 2023-08-21
Unsupported ref types in ArgumentRegisterAllocator


yarikbratashchuk:
Hello,

Here is a code that crashes on `todo!` [macro](https://github.com/wasmerio/wasmer/blob/master/lib/compiler-singlepass/src/x64_decl.rs#L323):
```rust
    let wasm_bytes = wat2wasm(r#"
        (module
            (type $t0 (func (param funcref externref)))
            (import "" "" (func $hello (type $t0)))
        )
        "#.as_bytes(),
    )?;

    let compiler = Singlepass::default();
    let mut store = Store::new(compiler);

    let module = Module::new(&store, wasm_bytes)?;
```

chipshort:
I think this has been fixed in #5309

PR #6139, opened 2026-01-29
fix(Singlepass): support ExternRef and FuncRef

marxin:
Add support for the types as the function arguments.

Fixes: #4169


marxin:
> Don't we need to enable current spec tests to see if things fully work?

The spec connected to these types are already enabled - this particular PR is about trampoline support of the types.

PR #6150, opened 2026-02-01
fix(sdk): Inaccurate handling of command-line arguments in `proc_exec`.

corwin-of-amber:
# Description

In POSIX, invoking a program from the shell with empty arguments (e.g., `funky fir "" tern`) should pass all arguments, including the empty ones, to the subprocess.

The implementation was filtering by `.is_empty`, presumably to get rid of the last element returned from `.split`, as the arguments are passed as a string, as per the WASI specifications, and the last argument has a trailing EOL.

### DIFF
--- a/lib/wasix/src/syscalls/wasix/proc_exec3.rs
+++ b/lib/wasix/src/syscalls/wasix/proc_exec3.rs
@@ -1,3 +1,4 @@
+use itertools::Itertools;
 use wasmer::FromToNativeWasmType;
 
 use super::*;
@@ -55,8 +56,8 @@ pub fn proc_exec3<M: MemorySize>(
     })?;
     let args: Vec<_> = args
         .split(&['\n', '\r'])
+        .dropping_back(1) // `split` returns one last (empty) element
         .map(|a| a.to_string())
-        .filter(|a| !a.is_empty())
         .collect();
 
     let envs = if !envs.is_null() {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7386,6 +7386,7 @@ dependencies = [
  "http-body-util",
  "hyper",
  "hyper-util",
+ "itertools 0.14.0",
  "js-sys",
  "libc",
  "linked_hash_set",
--- a/lib/wasix/Cargo.toml
+++ b/lib/wasix/Cargo.toml
@@ -41,6 +41,7 @@ shared-buffer.workspace = true
 hyper = { workspace = true, features = ["server"], optional = true }
 derive_more.workspace = true
 wasm-encoder.workspace = true
+itertools.workspace = true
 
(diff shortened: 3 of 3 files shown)


lib/wasix/src/syscalls/wasix/proc_exec3.rs
```diff
@@ -55,8 +56,8 @@ pub fn proc_exec3<M: MemorySize>(
     })?;
     let args: Vec<_> = args
         .split(&['\n', '\r'])
+        .dropping_back(1) // `split` returns one last (empty) element
         .map(|a| a.to_string())
-        .filter(|a| !a.is_empty())
         .collect();
```
syrusakbary, line 61: It could happen that the args don't end with `\n` or `\r\n`. If that's the case, it would remove the last char (which is not ideal).

I think it would be best to have a function:
```rust
fn strip_trailing_newline(input: &str) -> &str {
    input
        .strip_suffix("\r\n")
        .or(input.strip_suffix("\n"))
        .unwrap_or(input)
}
```

And then do:

```suggestion
    let args: Vec<_> = strip_trailing_newline(args)
        .split(&['\n', '\r'])
        .map(|a| a.to_string())
        .collect();
```

marxin: We can do even better:
```
args.trim_end_matches(&['\r', '\n'])
        .lines()
        .map(str::to_owned)
        .collect_vec();
```

Arshia001:
The _proper_ way to fix this would be to drop the string-join-based implementation (I don't know where or why we got that), and go for a proper array-of-strings impl. This implementation is also quite inefficient on the guest side because space must be allocated for the overall string argument to this syscall. Short of that, I agree with @marxin's code.

corwin-of-amber:
Seems like there is no way to avoid the string parsing here, since this is part of the WASIX API. I don't know how changeable this is.

```c
void __wasi_proc_exec(
    /**
     * Name of the process to be spawned
     */
    const char *name,
    /**
     * List of the arguments to pass the process
     * (entries are separated by line feeds)
     */
    const char *args
);
```
https://github.com/wasix-org/wasix-libc/blob/e7af091259489c514784d9125f29ac3d70ede1d8/libc-bottom-half/headers/public/wasi/api_wasix.h#L4232

As the operation requested is launching a new process, the copying of the arguments in memory, in term of efficiency, is acceptable IMHO.

marxin:
> As the operation requested is launching a new process, the copying of the arguments in memory, in term of efficiency, is acceptable IMHO.

Yep! Please update the PR based on the suggestion and we're ready to merge it.

```rust
args.trim_end_matches(&['\r', '\n'])
        .lines()
        .map(str::to_owned)
        .collect_vec();
```

marxin:
Applied the change in #6168.

PR #6153, opened 2026-02-02
feat(LLVM+Cranelift): parallel compilation revamp

marxin:
Currently, when building a module with LLVM, we utilize Rayon's `par_iter` on the function bodies. That works fine, except we have no control about the distribution of the function sizes. And so, we can easily end up with a range of functions that are really huge:

`static-web-server`:
screenshot-2026-02-02_06-30-46

The PR builds the buckets with uniform size and then the compilation starts with the largest buckets first:
screenshot-2026-02-02_06-26-06

### DIFF
--- a/tests/compilers/progress.rs
+++ b/tests/compilers/progress.rs
@@ -4,8 +4,6 @@ use std::sync::{Arc, Mutex};
 use wasmer::{sys::NativeEngineExt, wat2wasm};
 use wasmer_types::{CompilationProgressCallback, CompileError, UserAbort};
 
-use crate::Compiler;
-
 const SIMPLE_WAT: &str = r#"(module
   (import "env" "div" (func $div (param i32 i32) (result i32)))
   (func (export "add") (param i32 i32) (result i32)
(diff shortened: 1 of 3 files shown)


marxin:
@syrusakbary reworked the PR so that we have a common infrastructure that is currently used both by LLVM and Cranelift.

tests/compilers/progress.rs
```diff
@@ -44,8 +46,14 @@ fn reports_progress_steps(mut config: crate::Config) -> Result<()> {
         "expected at least one progress notification"
     );
     let last = events.last().unwrap();
-    assert_eq!(last.phase_step_count(), Some(4));
-    assert_eq!(last.phase_step(), Some(4));
+    // LLVM/Cranelift compiler uses bitcode size for the total.
+    if matches!(config.compiler, Compiler::LLVM | Compiler::Cranelift) {
+        assert_eq!(last.phase_step_count(), Some(2014));
+        assert_eq!(last.phase_step(), Some(2014));
```
syrusakbary: Steps should be invariant regardless of the backend, I don't think is a good idea to make it dependent of the backend

syrusakbary (changes requested):
Only one comment to address

tests/compilers/progress.rs
marxin: Yep, works for me and I've just fixed that.

syrusakbary (approved):

PR #6156, opened 2026-02-02
feat(LLVM): enable `pass-params` by default

marxin:
Based on the recent benchmark effort, I measured a significant speed-up with `pass-params` (aka. `g0m0` aka. stack-pointer and heap arguments) across the board:
rustc_perf_runtime2

The PR preserves the `suggested_compiler_opts` logic for the package files - currently only:
```
[module.annotations.suggested_compiler_optimizations]
pass_params = false
```

is taken into account (pass_params = true is on by default). Similarly for the Wasmer CLI, I ignore `--enable-pass-params-opt` option and a user can disable the optimization with `--disable-pass-params-opt`.

Apart from that, the PR addresses one bug where we wrongly included the 2 extra arguments for host-fn call.

### DIFF
--- a/lib/cli/src/backend.rs
+++ b/lib/cli/src/backend.rs
@@ -200,11 +200,19 @@ pub struct RuntimeOptions {
     #[clap(long, value_enum)]
     profiler: Option<Profiler>,
 
-    /// Only available for the LLVM compiler. Enable the "pass-params" optimization, where the first (#0)
+    /// Deprecated, "pass-params" optimization enabled by default, where the first (#0)
     /// global and the first (#0) memory passed between guest functions as explicit parameters.
+    /// Only available for the LLVM compiler.
     #[cfg(feature = "llvm")]
     #[clap(long)]
-    enable_pass_params_opt: bool,
+    _enable_pass_params_opt: bool,
+
+    /// Disable the "pass-params" optimization, where the first (#0)
+    /// global and the first (#0) memory passed between guest functions as explicit parameters.
+    /// Only available for the LLVM compiler.
+    #[cfg(feature = "llvm")]
+    #[clap(long)]
+    disable_pass_params_opt: bool,
 
     /// Sets the number of threads used to compile the input module(s).
     #[clap(long, alias = "llvm-num-threads")]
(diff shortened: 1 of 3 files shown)


syrusakbary:
I'm not sure about this. RIght now pass params embeds the stack pointer so it's not actually a variable from the outside. This has wide implications, for example when using dynamic linked modules.

However, if we move the strategy of the stack pointer being in one register (rather than in a local function variable that is "passed" as a copy), then we should be good to merge

marxin:
All right, I am getting into the nitty-gritty.

First, now I understand the `__stack_pointer` is a Clang (name your other favorite compiler) ABI concept, where the compiler overcomes the limitation of the WebAssembly as one can't take addresses of a stack variables. That's where `__stack_pointer` comes into play to serve languages like C and C++. Apart from that, our LLVM codegen depends on a classical stack-pointer register (and frame-pointer) that is used to access the `alloca`-based local stack slots.

That being said, I tried an experimental branch where I replaced the `g0m0` just with `m0`. So only the memory(0) (aka start of the heap) pointer is being added to the calling conventions. By doing that, I get vast **majority of the speed-up** provided by the original g0m0!

> I'm not sure about this. RIght now pass params embeds the stack pointer so it's not actually a variable from the outside. This has wide implications, for example when using dynamic linked modules.

Can you depict the exact scenario where that would break? Still, I am getting into all the WA aspects.
By still, I think we can get the biggest value from the heap pointer optimization (`m0`). Plus, I noticed we're currently using `rbp` register for non-leaf functions - we might adjust it as our unwinding mechanism depends on the DWARF info.

> However, if we move the strategy of the stack pointer being in one register (rather than in a local function variable that is "passed" as a copy), then we should be good to merge

There appears to be recent activity on this front: https://github.com/llvm/llvm-project/issues/179036. At the moment, support for x86_64 is still missing for a critical LLVM option. It’s worth emphasizing that in this case we must rely entirely on the compiler—otherwise subtle but serious issues can arise, for example with exception handling and correct stack unwinding.

syrusakbary:
> I get vast majority of the speed-up provided by the original g0m0

I guess depends on what benchmark you run. If you run the interpreter I attached in slack a few days ago, the major speedup came from g0 (if I recall correctly).

> Can you depict the exact scenario where that would break? Still, I am getting into all the WA aspects.

In the scenario that you run multiple threads (or you fork). Right now, things are cloned as you call it (namely, the globals, in which the stack pointer is included). When you move the stack pointer out of the global to be a function variable (that gets passed as value, not reference), then cloning will do nothing, and might break the Wasm behavior.
One way to alleviate that is to put the global value up to date by the end of Wasm function (when it goes outside of the module), but feels a bit hacky.

marxin:
Once #6187 gets in, we can basically enable the `g0` optimization by default (effectively removing the `--enable-pass-params-opt` option).

PR #6161, opened 2026-02-03
virtual-fs: Hide package contents under bind‑mounts by filtering overlay secondaries

artemyarulin:
Host bind‑mounts were incorrectly merged with package filesystem contents, causing package files (and even directories) to appear inside the mounted path and leak back to the host. 

```bash
$ cd /tmp
$ touch foo.txt
$ # Folders from the package merged with the mounted folder
$ wasmer run --registry wasmer.wtf wasmer/bash --use sha256:3a4d0b7932f16e70bdd38cd84ad478129c8e9dedd4ed41040938799ad21e6993 --mapdir /app/wp-content:$(pwd) -- -c "ls -a /app/wp-content; ls  /app/wp-content"
themes
plugins
foo.txt
$ ls /tmp # Folders from package itself leak to the host after
themes 
plugins
```

This PR makes bind‑mounted paths opaque to the overlay by filtering secondary filesystems under those prefixes, so the host mount replaces the package path (basically like in Docker) while keeping package mounts working for other paths.

### DIFF
--- a/lib/virtual-fs/src/overlay_fs.rs
+++ b/lib/virtual-fs/src/overlay_fs.rs
@@ -65,6 +65,7 @@ use crate::{
 pub struct OverlayFileSystem<P, S> {
     primary: Arc<P>,
     secondaries: S,
+    opaque_prefixes: Vec<PathBuf>,
 }
 
 impl<P, S> OverlayFileSystem<P, S>
@@ -78,6 +79,21 @@ where
         OverlayFileSystem {
             primary: Arc::new(primary),
             secondaries,
+            opaque_prefixes: Vec::new(),
+        }
+    }
+
+    /// Create a new [`FileSystem`] with opaque prefixes that hide any secondary
+    /// entries under those paths.
+    pub fn new_with_opaque_prefixes(
+        primary: P,
+        secondaries: S,
+        opaque_prefixes: Vec<PathBuf>,
+    ) -> Self {
+        OverlayFileSystem {
+            primary: Arc::new(primary),
+            secondaries,
+            opaque_prefixes,
         }
     }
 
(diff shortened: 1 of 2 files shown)


lib/virtual-fs/src/overlay_fs.rs
```diff
@@ -96,8 +112,31 @@ where
...
+        if self.is_opaque(path) {
+            Box::new(std::iter::empty())
+        } else {
+            Box::new(self.secondaries.filesystems().into_iter())
+        }
+    }
+
     fn permission_error_or_not_found(&self, path: &Path) -> Result<(), FsError> {
-        for fs in self.secondaries.filesystems() {
+        for fs in self.secondaries_iter(path) {
```
artemyarulin, line 139: I've tried to find a better way to avoid replacing every call to `secondaries`, but couldn't find it. Suggestion are welcome

lib/virtual-fs/src/overlay_fs.rs
artemyarulin, line 106: `secondaries` (and following `secondaries_mut`) are public and contains old behaviour without filtering. I didn't find the usage of it in wasmer though, only in tests, so not sure if that behaviour should be fixed here or not? 

And I guess as it's public then changing function signature is not possible, so changing that to the iterator is not acceptable, so not sure how to proceed with those

lib/virtual-fs/src/overlay_fs.rs
```diff
@@ -1169,6 +1210,72 @@ mod tests {
...
+            .create(true)
+            .write(true)
+            .open("/app/wp-content/themes/foo.txt")
+            .unwrap_err();
+        assert_eq!(err, FsError::EntryNotFound);
+
+        assert_eq!(
+            overlay
+                .primary()
+                .metadata(Path::new("/app/wp-content/themes"))
```
artemyarulin, line 1280: Test reproduces exact manual test that we've discovered

lib/wasix/src/runners/wasi_common.rs
```diff
@@ -430,6 +448,43 @@ mod tests {
...
+            use virtual_fs::FileSystem;
+            assert!(
+                overlay_fs
+                    .metadata("/app/wp-content".as_ref())
+                    .unwrap()
+                    .is_dir()
+            );
+            assert_eq!(
+                overlay_fs
+                    .metadata("/app/wp-content/themes".as_ref())
```
artemyarulin: Same test, but on the overlay layer, like an integration

lib/virtual-fs/src/overlay_fs.rs
```diff
@@ -297,7 +336,8 @@ where
             // the secondaries, in which case we need to copy it to the
             // primary rather than rename it
             if !had_at_least_one_success {
-                for fs in self.secondaries.filesystems() {
+                let secondaries: Vec<_> = self.secondaries_iter(&from).collect();
```
marxin, line 355: What about directly iterating the `self.secondaries_iter(&from)` Iteration output? Any reason for the collect here?

lib/virtual-fs/src/overlay_fs.rs
```diff
@@ -309,7 +349,8 @@ where
             // If the rename operation was a success then we need to update any
             // whiteout files on the primary before we return success.
             if had_at_least_one_success {
-                for fs in self.secondaries.filesystems() {
+                let secondaries: Vec<_> = self.secondaries_iter(&from).collect();
+                for fs in secondaries {
```
marxin, line 369: Likewise here.

theduke:
This is a tricky one, will need to dive in a bit more

artemyarulin:
@theduke ping!

theduke (commented):
I believe this PR may now be redundant, after the recent changes made to package mounting and `MountFs`. 

@Arshia001 , can you validate?

theduke:
Closing, since this is not needed anymore with our new mounting structure.

PR #6165, opened 2026-02-04
Deploy improvements

syrusakbary:
This PR:
* Improves the output of the deploy log
* Adds banner/alerts when the apps perish

### Before
*(note: the alert did not existed before)*
image

### After

image

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1002,9 +1002,6 @@ jobs:
     container: ${{ matrix.container }}
     steps:
       - uses: actions/checkout@v6
-      - uses: goto-bus-stop/setup-zig@v2
-        with:
-          version: 0.10.0
       - name: Set up base deps on musl
         if: matrix.build == 'linux-musl'
         run: |
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2189,6 +2189,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "getrandom"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "139ef39800118c7683f2fd3c98c1b23c09ae076556b435f8e9064ae108aaeeec"
+dependencies = [
+ "cfg-if",
+ "js-sys",
+ "libc",
+ "r-efi",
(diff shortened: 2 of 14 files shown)


lib/cli/src/commands/app/deploy.rs
```diff
@@ -1265,6 +1266,85 @@ pub async fn wait_app(
...
+    OffsetDateTime::try_from(datetime.clone())
+        .ok()
+        .and_then(|value| value.format(&format).ok())
+        .unwrap_or_else(|| datetime.0.clone())
+}
+
+fn format_duration_words(duration: TimeDuration) -> String {
+    let total_seconds = duration.whole_seconds().max(0);
+
+    if total_seconds >= 86_400 {
```
marxin: I would rather use `Duration::from_hours(24)` (seems `Duration::from_days(1)` is unstable yet).

lib/cli/src/commands/app/deploy.rs
```diff
@@ -1265,6 +1266,85 @@ pub async fn wait_app(
...
+        let days = total_seconds / 86_400;
+        format!("{days} day{}", if days == 1 { "" } else { "s" })
+    } else if total_seconds >= (3_600 - 60) {
```
marxin: Same here, where I would directly make a floating-point division with rounding, instead of the manually implemented rounding mechanism.

lib/backend-api/schema.graphql
```diff
@@ -3900,6 +3929,19 @@ type Mutation {
   generateUploadUrl(input: GenerateUploadUrlInput!): GenerateUploadUrlPayload
 }
 
+"""Sync all blog posts from Notion to the database."""
+type SyncNotionBlogPostsPayload {
```
marxin, line 3933: Am I right the graphql is a bit unrelated to the `deploy` adjustement? Or how does it work actually?

syrusakbary: This PR requires updating the schema, since it uses new fields that were not available before (the perishReason being the main offender)

marxin: Understood.

PR #6174, opened 2026-02-10
Resurrect Wasmer CLI integration tests

marxin:
The tests partially duplicate coverage elsewhere, but they still provide value and are worth keeping. Some legacy tests were removed.

Fixes: #6180

PR #6199, opened 2026-02-13
CI: do not create ignored jobs on PRs

marxin:
Having that we'll simplify the CI by not creating extra configurations that actually skip the actual `make XYZ` step.


marxin:
@syrusakbary I think we have to adjust the list of mandatory checks by removing:
```
Cranelift - no-std - macos-x64
Examples - macos-x64
Packages - std - macos-x64
Singlepass - no-std - macos-x64
Wasmer-cli - macos-x64
WAST (all comp.) - macos-x64
```

marxin:
@syrusakbary and these need to be removed from mandatory tests:
```
CLI tests - linux-musl
CLI tests - linux-x64
CLI tests - macos-arm
CLI tests - windows-x64
```

they got dropped as part of #6174

PR #6203, opened 2026-02-16
CI(macos-arm/x86): skip on PR (main branch only)

marxin:
Based on the Depot bill from the previous month, we must reduce usage of the Depot runners for MacOS.


marxin:
@theduke @syrusakbary again, we'll need to adjust the mandatory runs..

PR #6207, opened 2026-02-17
fix(wasix-fs): resolve absolute symlink targets from virtual root

theduke:
The WASIX symlink resolver incorrectly treated absolute symlink targets as
if they were relative to the directory containing the symlink.

This caused lookups like:
- create /test/hello -> /test2/hello
- open/cat /test/hello

to fail with ENOENT when the symlink lived below a non-root directory,
because resolution restarted from base_po_dir/path_to_symlink context
instead of the virtual root.

What changed:
- fs resolver (WasiFs::get_inode_at_path_inner):
  - for Kind::Symlink, detect relative_path.is_absolute()
  - resolve absolute targets from VIRTUAL_ROOT_FD
  - keep existing relative-target behavior (resolve from symlink parent)
- open syscall resolver (path_open_internal in path_open2):
  - mirror the same absolute-vs-relative split before recursive reopen
  - use VIRTUAL_ROOT_FD for absolute targets

Why both places:
- path traversal and open have separate symlink follow paths; fixing only one
  leaves inconsistent behavior between inode walking and open-by-path.

Tests:
- extended tests/wasix/symlink-open-read-write/main.c with a regression case
  where a nested symlink (nested/hello) points to an absolute target
  (/host/target.txt) and must open/read successfully.
- added required <sys/stat.h> include for mkdir in that test.

Validation performed:
- cargo test -p wasmer-wasix test_relative_path_to_absolute --lib passes.
- WASIX C integration test compile/run was not executed locally because
  wasixcc is not available in this environment.

### DIFF
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -1269,20 +1269,29 @@ impl WasiFs {
                         path_to_symlink,
                         relative_path,
                     } => {
-                        let new_base_dir = *base_po_dir;
-                        let new_base_inode = self.get_fd_inode(new_base_dir)?;
-
-                        // allocate to reborrow mutabily to recur
-                        let new_path = {
-                            /*if let Kind::Root { .. } = self.inodes[base_po_dir].kind {
-                                assert!(false, "symlinks should never be relative to the root");
-                            }*/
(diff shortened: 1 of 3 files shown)


theduke:
Not ready yet!

PR #6208, opened 2026-02-17
fix(wasix-fs): Fix symlink behaviour

theduke:
fix(wasix-fs): align symlink semantics across resolution, persistence, and mutation flows

    This consolidates and finalizes the symlink fixes so runtime-created and backing-FS symlinks follow the same representation and resolve consistently across WASIX filesystem code paths.

    Implemented changes:
    - Fixed absolute symlink target resolution in both core path traversal and `path_open2` symlink-reopen flow by resolving absolute targets from `VIRTUAL_ROOT_FD`.
    - Standardized runtime-created symlink representation to match backing-FS-discovered symlinks:
      `(base_po_dir, path_to_symlink_within_preopen, relative_path_raw_target)`.
    - Removed depth-based `..` rewriting in `path_symlink`; now stores raw `old_path` symlink targets.
    - Added `WasiFs::path_into_pre_open_and_relative_path_owned` helper for owned-path conversion.

    Backing FS + mem-fs integration:
    - Added symlink-node support in mem-fs:
      - new `Node::Symlink` variant,
      - symlink target persistence,
      - `readlink()` support,
      - `create_symlink(source, target)` support.
    - Exposed mem-fs symlink creation via `TmpFileSystem::create_symlink()`.
    - Updated mem-fs redirect behavior for symlink creation:
      - delegate to redirected mem-fs when possible,
      - return `FsError::Unsupported` otherwise.
    - Updated `path_symlink` to persist symlinks in backing FS when supported (sandbox / overlay primary).
    - Fallback to ephemeral symlink registration is now only used on `FsError::Unsupported`; other backing-FS errors are propagated.

    Ephemeral symlink behavior:
    - Kept ephemeral symlink registry for filesystems without native symlink creation.
    - Hardened key normalization:
      - prevent `..` from escaping above `/`,
      - ignore `Component::Prefix(_)` for virtual key normalization.
    - Ephemeral symlinks are resolved transiently and not inserted into directory-entry caches.
    - Added `WasiFs` helpers for ephemeral lifecycle:
      - `unregister_ephemeral_symlink`,
      - `move_ephemeral_symlink`.

    Mutation-flow consistency:
    - `path_rename` now handles symlink inodes coherently:
      - attempts backing-FS rename for real symlinks,
      - allows `Errno::Noent` only for known ephemeral symlinks,
      - updates stored symlink location (`base_po_dir`, `path_to_symlink`) after rename,
      - moves ephemeral registry entries for renamed ephemeral symlinks.
    - `path_unlink_file` now handles symlink deletion coherently:
      - removes real symlinks from backing FS,
      - unregisters ephemeral symlink entries when applicable,
      - preserves removal error behavior for real failures.

    Tests and validation:
    - Extended `tests/wasix/symlink-open-read-write/main.c` coverage with nested symlink open/read checks.
    - Fixed test setup robustness by allowing `EEXIST` from `mkdir`.
    - Verified with:
      - `cargo test -p wasmer-wasix test_relative_path_to_absolute --lib`
      - `cargo test --release --features cranelift --locked --test compilers path_symlink::cranelift::cranelift`
      - `cargo test --release --features cranelift --locked --test compilers path_rename::cranelift::cranelift` (selection builds; cases are ignored in this suite)

### DIFF
--- a/lib/virtual-fs/src/mem_fs/filesystem.rs
+++ b/lib/virtual-fs/src/mem_fs/filesystem.rs
@@ -254,6 +254,64 @@ impl FileSystem {
 
         Ok(())
     }
+
+    pub fn create_symlink(&self, source: &Path, target: &Path) -> Result<()> {
+        let (inode_of_parent, name_of_symlink) = {
+            let guard = self.inner.read().map_err(|_| FsError::Lock)?;
+            let path = guard.canonicalize_without_inode(target)?;
+            let parent_of_path = path.parent().ok_or(FsError::BaseNotDirectory)?;
+            let name_of_symlink = path
+                .file_name()
+                .ok_or(FsError::InvalidInput)?
+                .to_os_string();
+            let inode_of_parent = match guard.inode_of_parent(parent_of_path)? {
+                InodeResolution::Found(a) => a,
+                InodeResolution::Redirect(fs, mut redirected_path) => {
+                    redirected_path.push(&name_of_symlink);
+                    drop(guard);
+                    let fs_ref: &dyn crate::FileSystem = fs.as_ref();
+                    if let Some(mem_fs) = fs_ref.downcast_ref::<Self>() {
+                        return mem_fs.create_symlink(source, redirected_path.as_path());
+                    }
+                    return Err(FsError::Unsupported);
+                }
+            };
+            (inode_of_parent, name_of_symlink)
+        };
+
+        let mut fs = self.inner.write().map_err(|_| FsError::Lock)?;
+        if fs.canonicalize(target).is_ok() {
+            return Err(FsError::AlreadyExists);
+        }
+
+        let inode_of_symlink = fs.storage.vacant_entry().key();
+        let real_inode_of_symlink = fs.storage.insert(Node::Symlink(SymlinkNode {
+            inode: inode_of_symlink,
+            name: name_of_symlink,
+            target: source.to_path_buf(),
+            metadata: {
+                let time = time();
+                Metadata {
+                    ft: FileType {
+                        symlink: true,
+                        ..Default::default()
+                    },
+                    accessed: time,
+                    created: time,
+                    modified: time,
+                    len: source.as_os_str().len() as u64,
+                }
+            },
+        }));
+
+        assert_eq!(
+            inode_of_symlink, real_inode_of_symlink,
+            "new symlink inode should have been correctly calculated",
+        );
+
+        fs.add_child_to_node(inode_of_parent, inode_of_symlink)?;
+        Ok(())
+    }
 }
 
 impl crate::FileSystem for FileSystem {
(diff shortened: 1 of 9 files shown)


marxin:
Can we review the PR together with #6161 - seems to me there might be some interaction (read: the FS is can of worms).

lib/virtual-fs/src/mem_fs/filesystem.rs
```diff
@@ -254,6 +254,64 @@ impl FileSystem {
 
         Ok(())
     }
+
+    pub fn create_symlink(&self, source: &Path, target: &Path) -> Result<()> {
```
zebreus, line 258: This function should probably exist in the Filesystem trait and not just for the memfs. All other fs types can probably return that it's not supported.

zebreus, line 258: To align this more with a linux style fs implementation, I would name it just `symlink` and name the parameters `target` (The path the symlink should point to (the contents of the symlink)) and `linkpath` (The path where the new symlink should be created) like for fuse filesystems. 

The parameter currently named target is usually named source

lib/virtual-fs/src/tmp_fs.rs
```diff
@@ -58,6 +58,10 @@ impl TmpFileSystem {
     pub fn canonicalize_unchecked(&self, path: &Path) -> Result<PathBuf> {
         self.fs.canonicalize_unchecked(path)
     }
+
+    pub fn create_symlink(&self, source: &Path, target: &Path) -> Result<()> {
```
zebreus, line 62: Move to FileSystem trait

lib/wasix/src/fs/mod.rs
```diff
@@ -238,6 +238,14 @@ impl InodeWeakGuard {
     }
 }
 
+#[derive(Debug)]
+#[cfg_attr(feature = "enable-serde", derive(Serialize, Deserialize))]
+struct EphemeralSymlinkEntry {
+    base_po_dir: WasiFd,
```
zebreus, line 244: It is not obvious what the abbreviation `po` refers to. Please expand.

lib/wasix/src/fs/mod.rs
```diff
@@ -581,6 +612,65 @@ impl WasiFs {
...
+        guard.insert(
+            new_key,
+            EphemeralSymlinkEntry {
+                base_po_dir,
+                path_to_symlink: normalize_virtual_symlink_key(&path_to_symlink),
+                relative_path,
+            },
+        );
+    }
+
```
zebreus, line 673: It seems to me like most of this behaviour should be implemented in the virtual fs. Having another layer of emulated "ephemeral" symlinks above that makes the already complex and error-prone system even worse.

zebreus (dismissed):
It seems to me that this PR puts a lot of effort into working around the missing `symlink(target, linkname)` function in the FileSystem trait. 

It seems to me like most of this behaviour should be implemented in the virtual fs. Having another layer of emulated "ephemeral" symlinks above that makes the already complex and error-prone system even worse. Can you move that to a FileSystem implementation?

lib/wasix/src/fs/mod.rs
zebreus: This is still a problem, but not my problem ^^

zebreus (commented):
.

PR #6212, opened 2026-02-18
feat(LLVM): mark memory loads/stores as non-volatile for WASIX modules

marxin:
The current approach of marking all memory loads and stores as volatile has a significant negative impact on performance (with this change, the PHP benchmark runs about **10%** faster). I’m not an expert on the WebAssembly memory model, but we should allow LLVM to perform its usual optimizations—such as merging memory operations and eliminating unused loads—instead of preventing them across the board.

However, the change is not compatible with the current SPEC tests (for obvious reasons).

The change can be nicely explained by example:
```
  (func (export "_start") (result i32)
    (i32.load8_u (i32.const 65536))
    (drop)
    (i32.const 42)
  )
```
Will not hit out-of-bounds as the load is unused/dead.

```
  (func (export "_start") (result i32)
    (i32.load8_u (i32.const 65536))
  )
```

Will trigger out-of-bounds error!

### DIFF
--- a/lib/compiler-llvm/src/translator/code.rs
+++ b/lib/compiler-llvm/src/translator/code.rs
@@ -1211,8 +1211,7 @@ impl<'ctx> LLVMFunctionCodeGenerator<'ctx, '_> {
 
     // If this memory access must trap when out of bounds (i.e. it is a memory
     // access written in the user program as opposed to one used by our VM)
-    // then mark that it can't be delete. For WASIX modules, we are more aggressive
-    // as the volatile memory access is an optimization barrier.
+    // then mark that it can't be delete.
     fn mark_memaccess_nodelete(
         &mut self,
         memory_index: MemoryIndex,


marxin:
> but it should be enabled automatically when running WASIX modules

@syrusakbary Done in the updated version of the PR, please review.

marxin:
> The flag should be named differently, and should be the wasix library triggering it when compiling/creating the engine

Addressed that.

lib/compiler/src/artifact_builders/artifact_builder.rs
```diff
@@ -79,6 +79,10 @@ impl ArtifactBuild {
             .apply_on_module_info(&mut module)
             .map_err(|err| CompileError::MiddlewareError(err.to_string()))?;
         module.hash = Some(ModuleHash::new(data));
+        let volatile_memory_ops = !module
+            .imports
+            .keys()
+            .any(|import| import.module.starts_with("wasix_"));
```
syrusakbary: That's a bit hacky way of checking.
We can check instead if we are executing things with the WASI/WASIX runner from the webc (not sure if this file is where the logic should live, I forgot about where this lives)

lib/compiler-llvm/src/compiler.rs
```diff
@@ -201,6 +201,11 @@ impl LLVMCompiler {
                     binary_format,
                     pointer_width,
                     *target.cpu_features(),
+                    !compile_info
+                        .module
+                        .imports
+                        .keys()
+                        .any(|import| import.module.starts_with("wasix_")),
```
syrusakbary: Very hacky, should be provided from the upper level

marxin:
Based on the discussion with @theduke and @Arshia001, we concluded with a solution where a new Config option will be added (and enabled by default for the Wasmer CLI).

lib/cli/src/backend.rs
```diff
@@ -503,6 +503,7 @@ impl RuntimeOptions {
                 use wasmer_compiler_llvm::LLVMCallbacks;
                 use wasmer_types::entity::EntityRef;
                 let mut config = LLVM::new();
+                config.enable_non_volatime_memops();
```
Arshia001: volatile*

typo? XD

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -1207,7 +1211,8 @@ impl<'ctx> LLVMFunctionCodeGenerator<'ctx, '_> {
 
     // If this memory access must trap when out of bounds (i.e. it is a memory
     // access written in the user program as opposed to one used by our VM)
-    // then mark that it can't be delete.
+    // then mark that it can't be delete. For WASIX modules, we are more aggressive
```
Arshia001: I'd remove the mention of WASIX here.

lib/compiler-llvm/src/compiler.rs
```diff
@@ -445,6 +446,7 @@ impl Compiler for LLVMCompiler {
                     binary_format,
                     pointer_width,
                     *target.cpu_features(),
+                    self.config.enable_non_volatime_memops,
```
Arshia001: more typo

lib/compiler-llvm/src/config.rs
```diff
@@ -103,6 +103,7 @@ impl LLVMCallbacks {
 #[derive(Debug, Clone)]
 pub struct LLVM {
     pub(crate) enable_nan_canonicalization: bool,
+    pub(crate) enable_non_volatime_memops: bool,
```
Arshia001: typo

lib/compiler-llvm/src/config.rs
```diff
@@ -121,6 +122,7 @@ impl LLVM {
     pub fn new() -> Self {
         Self {
             enable_nan_canonicalization: false,
+            enable_non_volatime_memops: false,
```
Arshia001: typo

lib/compiler-llvm/src/config.rs
```diff
@@ -155,6 +157,13 @@ impl LLVM {
         self
     }
 
+    /// For the LLVM compiler, we can use non-volatile memory operations which lead to a better performance
+    /// (but are not 100% SPEC compliant).
+    pub fn non_volatime_memops(&mut self, enable_non_volatime_memops: bool) -> &mut Self {
```
Arshia001: typo

lib/compiler-llvm/src/config.rs
```diff
@@ -357,6 +366,12 @@ impl CompilerConfig for LLVM {
         self.enable_verifier = true;
     }
 
+    /// For the LLVM compiler, we can use non-volatile memory operations which lead to a better performance
+    /// (but are not 100% SPEC compliant).
+    fn enable_non_volatime_memops(&mut self) {
```
Arshia001: typo! 😆

lib/compiler/src/compiler.rs
```diff
@@ -50,6 +50,10 @@ pub trait CompilerConfig {
         // in case they create an IR that they can verify.
     }
 
+    /// For the LLVM compiler, we can use non-volatile memory operations which lead to a better performance
+    /// (but are not 100% SPEC compliant).
+    fn enable_non_volatime_memops(&mut self) {}
```
Arshia001: aaaaand typo.

marxin: Fixed.

PR #6216, opened 2026-02-19
Add WASM feature detection for WASIX subprocess spawning

Arshia001:
This PR implements additional checks for WASM features when spawning a WASIX subprocess. Fixes scenarios such as bash (non-EH) calling python (EH).

### DIFF
--- a/lib/wasix/src/runtime/mod.rs
+++ b/lib/wasix/src/runtime/mod.rs
@@ -673,11 +673,9 @@ impl Runtime for PluggableRuntime {
             }
         };
 
-        let engine;
-
         cfg_if! {
             if #[cfg(feature = "sys-default")] {
-                engine = match features {
+                let engine = match features {
                     Some(f) => match self.engine_with_extended_features(&base_engine, &f) {
                         Ok(engine) => engine,
                         Err(e) => {
@@ -692,7 +690,7 @@ impl Runtime for PluggableRuntime {
                     None => base_engine,
                 };
             } else {
-                engine = base_engine;
+                let engine = base_engine;
             }
         };
 


lib/wasix/Cargo.toml
```diff
@@ -30,6 +30,7 @@ virtual-net = { path = "../virtual-net", version = "0.700.1", default-features =
 ] }
 wasmer-journal = { path = "../journal", version = "0.700.1", default-features = false }
 wasmer-config = { version = "0.700.1", path = "../config" }
+wasmer-compiler = { version = "7.0.1", path = "../compiler" }
```
zebreus: What are the implications of adding a compiler dependency for wasmer builds without sys? Can we make this optional and gate it on the `sys` feature?

lib/wasix/src/syscalls/wasix/proc_exec3.rs
```diff
@@ -124,8 +124,6 @@ pub fn proc_exec3<M: MemorySize>(
         }
     };
 
-    let new_store = ctx.data().runtime.new_store();
```
zebreus, line 127: I assume this was dead code?

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -33,14 +35,33 @@ pub async fn spawn_exec(
     spawn_union_fs(&env, &binary).await?;
 
     let cmd = package_command_by_name(&binary, name)?;
+
+    // Build a store whose engine is configured for this command's required
+    // WebAssembly features (e.g. SIMD, relaxed-SIMD, threads, etc.).
+    let store = cmd.wasm_features().and_then(|features| {
```
theduke: This breaks external users overwriting engine selection.

Can we instead rely on `Runtime::resolve_module()` , put the new logic in the default implementation, and then rely on `let engine = returned_module.as_engine_ref().engine().clone()` (or whatever the exact API is) for the engine to use for the new store?

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -50,9 +71,28 @@ pub async fn spawn_exec_wasm(
     env: WasiEnv,
     runtime: &Arc<dyn Runtime + Send + Sync + 'static>,
 ) -> Result<TaskJoinHandle, SpawnError> {
-    let module = spawn_load_module(name, wasm, runtime).await?;
-
-    spawn_exec_module(module, env, runtime)
+    // Detect WebAssembly features required by this module so that we can
+    // configure the engine appropriately for the subprocess.
+    let store = wasmer_compiler::Features::detect_from_wasm(wasm.wasm().as_slice())
```
theduke: As above regarding engine selection.

lib/wasix/src/runtime/task_manager/mod.rs
```diff
@@ -100,6 +100,7 @@ pub struct TaskWasm<'a> {
     pub module: Module,
     pub globals: Option<StoreSnapshot>,
     pub spawn_type: SpawnType<'a>,
+    pub store_override: Option<wasmer::Store>,
```
theduke: Do we need this? If we use the engine from the module that should be redundant?

lib/wasix/src/runtime/task_manager/mod.rs
```diff
@@ -132,6 +134,11 @@ impl<'a> TaskWasm<'a> {
         }
     }
 
+    pub fn with_store_override(mut self, store: wasmer::Store) -> Self {
```
zebreus: Maybe add some doc comment for store override

lib/wasix/src/runtime/mod.rs
```diff
@@ -178,6 +178,27 @@ where
         }
     }
 
+    /// Create a new [`wasmer::Store`].
+    fn new_store_with_extended_features(
+        &self,
+        features: &Features,
+    ) -> Result<wasmer::Store, wasmer_types::CompileError> {
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "sys")] {
```
zebreus: If other backends have similar restrictions, it might be nice to have that functionality for them as well

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -50,9 +71,28 @@ pub async fn spawn_exec_wasm(
...
+                .new_store_with_extended_features(&features)
+                .map_err(|e| {
+                    tracing::debug!(
+                        error = %e,
+                        "Could not create store with extended features; falling back to default"
+                    );
+                    e
+                })
+                .ok()
+        });
```
zebreus: using feature detection from wasmer_compiler should probably be gated to sys.

lib/wasix/src/bin_factory/exec.rs
zebreus: wasm_features returns `None` if the manifest doesn't explicitly specify the wasm features. We should probably do feature detection in that case to align commands with wasm modules

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -50,9 +71,28 @@ pub async fn spawn_exec_wasm(
...
+
+    let engine_ref = store.as_ref().map(|s| s.as_engine_ref());
+    let module =
+        spawn_load_module(name, wasm, runtime, engine_ref.as_ref().map(|e| e.engine())).await?;
```
zebreus: spawn_load_module is just a light wrapper around  `resolve_module`. Maybe use that to make this code path more similar to the one in spawn_exec.

lib/wasix/src/bin_factory/exec.rs
```diff
@@ -136,20 +178,24 @@ pub fn spawn_exec_module(
...
+        let task = if let Some(store) = store_override {
+            task.with_store_override(store)
+        } else {
+            task
+        };
+
+        tasks_outer.task_wasm(task).map_err(|err| {
+            error!("wasi[{}]::failed to launch module - {}", pid, err);
+            SpawnError::Other(Box::new(err))
+        })?
```
zebreus: I like the simplification here

lib/wasix/Cargo.toml
Arshia001: good point!

lib/wasix/src/syscalls/wasix/proc_exec3.rs
Arshia001: yep, smh

lib/wasix/src/bin_factory/exec.rs
Arshia001: safe to assume such packages are corrupt, @theduke ?

theduke:
That's actually what Edge does, it will pick the most compatible engine across all binaries in the package.

marxin:
Closing as went with the other approach.

Issue #6170, opened 2026-02-06
Windows library not static


ashtonmeuser:
### Describe the bug

With the release of v7.0.0-alpha.1, the Windows static library (_lib/wasmer.lib_) is not correctly compiled as a static lib and is instead a DLL import lib (same as _lib/wasmer.dll.lib_).

See [this workflow run](https://github.com/ashtonmeuser/godot-wasm/actions/runs/21764041280) where I've built against Wasmer v6.1.0, v7.0.1, v7.0.0-alpha.1, v7.0.0-alpha.2, v7.0.0-rc.1, v7.0.0-rc.2, v7.0.0. All fail linking except v6.1.0. See [this run](https://github.com/ashtonmeuser/wasmer-test/actions/runs/21767104563) for a more minimal reproduction (and possibly jumping off point for a static lib integration test).

Downloading the asset and poking around (via `ar -t wasmer-windows-amd64/lib/wasmer.lib | head` on MacOS) with various versions confirms the lib is static in v6.1.0 and a DLL import lib in v7.0.0-alpha.1.

C library configuration seems to be a constant pain point for Wasmer. I'll look into contributing even a simple test that links against the lib to avoid these issues in the future. Libs were broken from Wasmer 4 → 5.0.0, fixed in 6.1.0, and broken once again (in a different way) from 6 → 7.0.0.

Possibly related: https://github.com/wasmerio/wasmer/issues/5721.

### Steps to reproduce

Statically link the Wasmer C lib. Note failure.

### Expected behavior

Library links successfully.

### Actual behavior

Library fails to link.

<details>
<summary>Error log</summary>

```
Linking Shared Library addons\godot-wasm\bin\windows\godot-wasm.dll ...
wasm-memory.windows.template_release.x86_64.obj : error LNK2019: unresolved external symbol wasm_engine_delete referenced in function "private: __cdecl godot_wasm::Store::~Store(void)" (??1Store@godot_wasm@@AEAA@XZ)
wasm.windows.template_release.x86_64.obj : error LNK2001: unresolved external symbol wasm_engine_delete
wasm-memory.windows.template_release.x86_64.obj : error LNK2019: unresolved external symbol wasm_engine_new referenced in function "private: __cdecl godot_wasm::Store::Store(void)" (??0Store@godot_wasm@@AEAA@XZ)
wasm.windows.template_release.x86_64.obj : error LNK2001: unresolved external symbol wasm_engine_new
[...]

PR #6221, opened 2026-02-20
fix(packaging): take correct wasmer.lib file for packaging

marxin:
Fixes: #6170


marxin:
The change does not fix the underlying issue..

marxin:
Superseded by #6275.

PR #6222, opened 2026-02-23
fix(Backend): Sys backend should use all supported features

marxin:
Let's enable every feature the engine supports, since the same engine may later be used with a module that requires more features than the one used during engine detection (e.g. a non-EH bash calling EH Python).

An alternative approach for #6216.


Arshia001 (commented):
The code looks good, will wait for @syrusakbary's input on whether we want this change from a product perspective

PR #6223, opened 2026-02-23
feat: drop SuggestedCompilerOptimizations as it's unused

marxin:


lib/config/src/package/mod.rs
```diff
@@ -503,26 +503,6 @@ pub struct Module {
     /// Interface definitions that can be used to generate bindings to this
     /// module.
     pub bindings: Option<Bindings>,
-    /// Miscellaneous annotations from the user.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub annotations: Option<UserAnnotations>,
```
theduke, line 508: This is a breaking change, we can't do that!
The annotation plumbing is sensible to keep in any case, there can be other use cases for it.

marxin: Sure, good point, fixed.

PR #6228, opened 2026-02-23
feat: embed local `Globals` and funcref `Table` in `VMContext`

marxin:
By placing the local Globals and the funcref Table directly in VMContext, we eliminate one level of indirection (though it’s typically cached by the CPU).

This results in a small performance improvement, but it’s debatable whether the added complexity is worth it. For context, I also experimented with the -O2 LLVM optimization pipeline, and it appears that those optimizations alone bring us close to the same gains as the proposed change.

interpreters_runtime_benchmark
rustc_perf_runtime2

@syrusakbary Your thoughts on that?

### DIFF
--- a/lib/types/src/serialize.rs
+++ b/lib/types/src/serialize.rs
@@ -13,7 +13,7 @@ pub struct MetadataHeader {
 impl MetadataHeader {
     /// Current ABI version. Increment this any time breaking changes are made
     /// to the format of the serialized data.
-    pub const CURRENT_VERSION: u32 = 16;
+    pub const CURRENT_VERSION: u32 = 17;
 
     /// Magic number to identify wasmer metadata.
     const MAGIC: [u8; 8] = *b"WASMER\0\0";


syrusakbary:
I find very weird that doing this increased the times on `hashmap_remove` (and got even worse with O2 optimization)

syrusakbary:
I think we are tricking things including non-volatile in the benchmarks when comparing with pass-params. It would be good to test without non-volatile so we can compare apples with apples

marxin:
> I find very weird that doing this increased the times on `hashmap_remove` (and got even worse with O2 optimization)

Just verified that manually - the test is really unstable I can get easily +- 5% results when running multiple times (with 100 iterations) with the same compiler (so let's don't take the one much seriously).

marxin:
> I think we are tricking things including non-volatile in the benchmarks when comparing with pass-params. It would be good to test without non-volatile so we can compare apples with apples

Yes, let me include a new graph that also lists the `main` branch after we did `m0` optimization (but before the non-volatile memops were merged) (green color):
interpreters_runtime_benchmark

That being said, if we exclude all the experiments with `-O2` options, the PR **does not help PHP**, but improves **Python by 10%**.

marxin:
Decided to rework it and make v2 of it in #6290.

PR #6236, opened 2026-02-24
CI: simplify artifact tests

marxin:
By doing that, we don't have to persist the huge `.wasmu` binary blobs.


tests/compilers/artifact.rs
```diff
@@ -1,15 +1,31 @@
...
 use anyhow::Result;
+use sha2::{Digest, Sha256};
 
 use wasmer::Module;
 use wasmer_types::Features;
 
-const ARTIFACT_FILES: [&str; 3] = ["bash.wasm", "cowsay.wasm", "python-3.11.3.wasm"];
+// Instead of preserving the .wasmu files in the repository,
+// we can just compare the expected SHA 256 of the files.
+const ARTIFACT_FILES: [(&str, &str); 3] = [
```
theduke, line 11: Not sure if I have context on this, but wasn't the point of the test to ensure that old artifacts keep working so that we don't accidentally break the artifact format?

theduke: If so, this would kind of negate the point, because every small compiler change etc would require updating the hash, so we wouldn't find accidental artifact breakage.

marxin: Yes — by knowing the sha256 hash of the artifacts, we can verify that they were serialized into exactly the same content (i.e., they produce the same hash). That guarantees we can deserialize them correctly.

marxin: > If so, this would kind of negate the point, because every small compiler change etc would require updating the hash, so we wouldn't find accidental artifact breakage.

Correct!

PR #6241, opened 2026-02-25
chore: simplify WasmFeatures init

marxin:
Let's start with an empty list and intentionally enable what we support.


theduke (approved):
Looks good.

Did a quick double-check against https://docs.rs/wasmparser/latest/wasmparser/struct.WasmFeatures.html to make sure we cover everything, and looks fine.

marxin:
Thanks for the review @theduke.

Issue #6235, opened 2026-02-24
support Extended Constant Expressions


marxin:
Spec link: https://github.com/WebAssembly/extended-const/blob/master/proposals/extended-const/Overview.md

PR #6242, opened 2026-02-26
feat: support Extended constant expressions WA feature

marxin:
With the feature, the compile does not need to emit a series of dynamic relocations (`__wasm_apply_data_relocs`) that can be rather large:

```
  (func $__wasm_apply_data_relocs (type 1)
    i32.const 3001748
    global.get $__memory_base
    i32.add
    global.get $__memory_base
    i32.const 279405
    i32.add
    i32.store
    i32.const 3001752
    global.get $__memory_base
    i32.add
    global.get $__memory_base
    i32.const 359472
    i32.add
    i32.store
    i32.const 3001756
    global.get $__memory_base
    i32.add
    global.get $__memory_base
    i32.const 360496
    i32.add
    i32.store
    i32.const 3001760
    global.get $__memory_base
    i32.add
    global.get $__memory_base
    i32.const 362032
    i32.add
    i32.store
...
```

Resolves: #6235.

### DIFF
--- a/lib/compiler/src/engine/artifact.rs
+++ b/lib/compiler/src/engine/artifact.rs
@@ -688,7 +688,7 @@ impl DataInitializerLocationVariant<'_> {
             Self::Plain(p) => (*p).clone(),
             Self::Archived(a) => DataInitializerLocation {
                 memory_index: a.memory_index(),
-                offset_expr: rkyv::deserialize::<_, rkyv::rancor::Error>(&a.offset_expr).unwrap(),
+                offset_expr: a.offset_expr(),
             },
         }
     }


lib/compiler/src/engine/artifact.rs
```diff
@@ -688,8 +688,7 @@ impl DataInitializerLocationVariant<'_> {
             Self::Plain(p) => (*p).clone(),
             Self::Archived(a) => DataInitializerLocation {
                 memory_index: a.memory_index(),
-                base: a.base(),
-                offset: a.offset(),
+                offset_expr: rkyv::deserialize::<_, rkyv::rancor::Error>(&a.offset_expr).unwrap(),
```
theduke: Mhm, why do we need to do nested deserialize here, and and `unwrap()`?

Don't want to panic on malformed artifacts.

marxin: Used API function, however it seems there are multiple places where we actually call `rkyv::deserialize.*.unwrap()`.
Do you want me to create an issue for that and address it?

PR #6243, opened 2026-02-26
feat: support Wide arithmetic WA proposal

marxin:
While working on the [`rustc-perf benchmarks`](https://github.com/rust-lang/rustc-perf/blob/master/collector/runtime-benchmarks/hashmap/src/main.rs), I noticed the default hashing algorithm (`foldhash`) of `hashbrown` provides a feature gate for the [`wide-arithmetic`](https://github.com/orlp/foldhash/blob/master/src/lib.rs#L159). With the feature implemented, we got optimal code for `x86_64` and some of the hashing benchmarks are ~10% faster.

Feature docs: https://github.com/WebAssembly/wide-arithmetic/blob/main/proposals/wide-arithmetic/Overview.md

### DIFF
--- a/fuzz/fuzz_targets/universal_cranelift.rs
+++ b/fuzz/fuzz_targets/universal_cranelift.rs
@@ -27,6 +27,7 @@ impl Arbitrary<'_> for CraneliftPassFuzzModule {
         config.tail_call_enabled = false;
         config.simd_enabled = false;
         config.relaxed_simd_enabled = false;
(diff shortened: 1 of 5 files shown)


syrusakbary (approved):
Nice! It would be interesting to add it to cranelift / singlepass

marxin:
> Nice! It would be interesting to add it to cranelift / singlepass

Cranelift support is added! For Singlepass, let me list it on my TODO list ;)

PR #6247, opened 2026-02-26
fix(wasix): make fd_renumber close-and-insert atomic

sumleo:
## Summary
- `fd_renumber_internal` has a TOCTOU race between closing the target FD and inserting the source FD at the target slot
- Fixed by holding a single write lock across both operations, making the rename truly atomic as documented

## Details
The function's documentation says "Atomically copy file descriptor", but the implementation was not atomic. It performed two steps that each acquired and released the `fd_map` write lock independently:

1. `state.fs.close_fd(to)` -- acquires fd_map write lock, removes FD, releases lock
2. `fd_map.insert(true, to, new_fd_entry)` -- acquires fd_map write lock again

Between these two lock acquisitions, another thread could allocate a new FD into the now-free `to` slot. When the original thread then tries the exclusive insert, it would find the slot occupied and panic:
```
panic!("Internal error: expected FD {to} to be free after closing in fd_renumber");
```

The fix inlines the close operation (`fd_map.remove(to)`) under the same write lock scope as the insert. The flush is still performed outside the lock since it may involve async I/O and should not block other fd_map operations.

### Before (two separate lock acquisitions):
```rust
wasi_try_ok!(state.fs.close_fd(to));  // acquires + releases fd_map lock
// RACE WINDOW: another thread can allocate into slot `to`
let mut fd_map = state.fs.fd_map.write().unwrap();  // re-acquires lock
fd_map.insert(true, to, new_fd_entry);  // may panic if slot taken
```

### After (single lock acquisition):
```rust
let mut fd_map = state.fs.fd_map.write().unwrap();
let _ = fd_map.remove(to);  // close under same lock
fd_map.insert(true, to, new_fd_entry);  // slot guaranteed free
```

## Testing
- The race condition requires concurrent FD allocation during `fd_renumber`, which is difficult to trigger deterministically but can occur under multi-threaded WASIX workloads
- Existing fd_renumber tests should continue to pass since the fix preserves identical behavior for the single-threaded case
- Can be verified by code inspection: the close and insert now happen under a single `fd_map.write()` guard with no intermediate lock release


theduke:
Thanks!

Your commits have been included in #6261 as part of additional fixes.

PR #6252, opened 2026-03-02
fix(singlepass): read float return values from SIMD registers per C ABI

deuszx:
## Summary

The singlepass compiler's `emit_call_native` reads all return values from GPR registers (RAX/RDX on x86_64, X0-X7 on ARM64, X10+ on RISC-V). However, the platform C ABIs return floating-point values in SIMD registers:
- **x86_64 System V**: XMM0-XMM1
- **ARM64 AAPCS64**: V0-V7
- **RISC-V**: FA0 (F10), FA1 (F11)

This causes **f32/f64 return values from imported host functions to contain garbage data**.

## Root Cause

This regression was introduced when return value handling was moved from the type-aware `Operator::Call` handler into `emit_call_native`, which uses `get_return_value_location()` — a function that unconditionally returns GPR locations regardless of the return value type.

In older wasmer versions (pre-4.x refactor), the `Operator::Call` handler checked `is_float()` on each return type and read from the appropriate SIMD register (e.g., `get_simd_for_ret()` returning V0) for float types, and GPR for integer types. After the refactoring, this type-awareness was lost.

## Fix

- Adds `get_simd_return_register(float_idx)` to the `Machine` trait, returning the correct SIMD register for the nth float return value per the platform's C ABI
- In `emit_call_native`, after building the return value location list, float entries are replaced with the correct SIMD register
- Implementations provided for all three backends: ARM64, x86_64, and RISC-V

## Minimal Reproduction

```rust
use wasmer::{Store, Module, Instance, Function, imports};

fn main() {
    let wat = r#"(module
        (import "host" "get" (func $get (result f32)))
        (func (export "test") (result f32) call $get)
    )"#;

    let engine = wasmer::sys::EngineBuilder::new(
        wasmer::sys::Singlepass::new()
    ).engine().into();
    let mut store = Store::new(engine);
    let module = Module::new(&store.engine().clone(), wat).unwrap();
    let get = Function::new_typed(&mut store, || -> f32 { -0.125 });
    let imports = imports! { "host" => { "get" => get } };
    let instance = Instance::new(&mut store, &module, &imports).unwrap();
    let test = instance.exports.get_function("test").unwrap();
    let result = test.call(&mut store, &[]).unwrap();
    // Before fix: garbage value (e.g., 0x54304080)
    // After fix: correct value -0.125 (bits 0xBE000000)
    println!("{:?}", result);
}
```

## Test Plan

- [x] Added `float_return_values_from_imports` test in `tests/compilers/imports.rs` covering f32, f64, i32, and i64 return values from imports
- [x] All existing import tests pass on both singlepass and cranelift
- [x] Verified fix on ARM64 (Apple Silicon) with singlepass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/compiler-singlepass/src/codegen.rs
+++ b/lib/compiler-singlepass/src/codegen.rs
@@ -663,7 +663,8 @@ impl<'a, M: Machine> FuncGen<'a, M> {
             _ => Size::S64,
         };
         let param_sizes = params_type.map(get_size).collect_vec();
-        let return_value_sizes = return_types.map(get_size).collect_vec();
+        let return_wptypes = return_types.collect_vec();
+        let return_value_sizes = return_wptypes.iter().map(|t| get_size(*t)).collect_vec();
 
         /* We're going to reuse the memory param locations for the return values. Any extra needed slots will be allocated on stack. */
         let used_stack_params = stack_params
@@ -711,6 +712,18 @@ impl<'a, M: Machine> FuncGen<'a, M> {
                 &mut stack_offset,
                 self.calling_convention,
             ));
+        }
+        // The C ABI returns floating-point values in SIMD registers (XMM0 on x86_64,
+        // V0 on ARM64, FA0 on RISC-V), not in GPRs. Fix up the return value locations
+        // so that we read float results from the correct registers.
+        let mut float_idx = 0usize;
+        for (i, wp_type) in return_wptypes.iter().enumerate() {
+            if matches!(wp_type, WpType::F32 | WpType::F64) {
+                if let Some(simd_loc) = self.machine.get_simd_return_register(float_idx) {
+                    return_args[i] = simd_loc;
+                }
+                float_idx += 1;
+            }
         }
 
         // Allocate space for arguments relative to SP.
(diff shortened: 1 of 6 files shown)


marxin:
The PR is about to fix #5943, but it's far from being complete. For a complete Singlepass fix, we need to factor out the logic in `x86_64_systemv.rs` (and other files in the folder), and use it for both LLVM and Singlepass:
https://github.com/wasmerio/wasmer/blob/492ef3b658d6a4ddfcfe5f689b9388b20673918b/lib/compiler-llvm/src/abi/x86_64_systemv.rs

deuszx:
Thanks @marxin for responding. This is a hard blocker (at least for our project) and a regression – this used to work in older versions for us and now stopped.

marxin:
> Thanks @marxin for responding.

We are really sorry about the limitation and we're planning to work on it in the upcoming future.

> This is a hard blocker (at least for our project) and a regression – this used to work in older versions for us and now stopped.

Can you be more concrete about the used scenario and the project? What type of host functions do you import into the WA modules? I am all ears.

deuszx:
I really appreciate all the effort, I didn't want to come across any different than that.

If I remember correctly, the minimal example that reproduces it is returning floating points. The [test is](https://github.com/linera-io/wasmer/blob/af787430b80e65552973e2c7a6f7e409934630ee/tests/compilers/imports.rs#L523) `float_return_values_from_imports` .

PR #6253, opened 2026-03-02
feat(debugging): add find-fn-in-perfmap script

marxin:
The PR includes more verbose debugging messages related to the EH, and on top of that, I include a script that prints a function name for a given address (from the perf file).

### DIFF
--- a/scripts/find-fn-in-perfmap.py
+++ b/scripts/find-fn-in-perfmap.py
@@ -1,4 +1,9 @@
 #!/usr/bin/env python3
+#
+# Wasmer maps JIT-compiled code directly into memory, so resolving function names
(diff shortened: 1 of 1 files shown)


scripts/find-fn-in-perfmap.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
+
```
theduke, line 7: Can we add an explanation for what the script is for / how to use it?

(it's cryptic for someone with no context)

marxin: Sure, good idea - resolved.

theduke (approved):

PR #6263, opened 2026-03-04
wasix: Epoll refactor + various file system fixes

theduke:
* Refactor EPOLL implementation
  - fix race conditions and invalid behaviours
  - much better performance
* Fix various file system / FD related race conditions and bugs


marxin:
@theduke Can we merge it so that it gets into the `7.1` planned to be released later this week?

PR #6265, opened 2026-03-05
deps: bump MSRV to 1.92

marxin:
Today is a new Rust release day:
https://releases.rs/docs/1.94.0/


theduke:
@marxin let's please stop constantly bumping versions just because. 

This causes a lot of uneccessary churn. 

Is there any reason why we need Rust 1.92?

marxin:
All right, I'll make any MSRV bump always connected to a real need.

PR #6267, opened 2026-03-05
feat(wasix): make builtin commands pluggable

Cookee24:
## Summary

This PR makes builtin commands in wasmer-wasix pluggable via a simple public API, while keeping the internal cmd_wasmer
implementation internal.

## What changed

- Added public BuiltinCommand::new(name, handler) for easy closure-based builtin commands.
- Added builder APIs to register custom builtins:
    - builtin_command(...)
    - builtin_command_with_path(...)
- Added builder toggle:
    - disable_default_builtins(bool) (replaces use_default_builtins style)
- Added BinFactory public APIs:
    - register_builtin_command(...)
    - register_builtin_command_with_path(...)
    - clear_builtin_commands()
- Kept default behavior backward-compatible (default builtins still enabled unless explicitly disabled).

Also, I notice that there is a TODO on `Commands::new_with_builtins`, mabey there is a better implementation?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -208,14 +208,16 @@ jobs:
           target = "${{ matrix.target }}"
           EOF
         if: matrix.target
-      - name: Build C API headless
+      - name: Build & Package C API headless
         shell: bash
         run: |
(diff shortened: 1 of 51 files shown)


syrusakbary:
This PR is very interesting, I like the abstraction. @theduke @Arshia001 can you take a look?

PR #6268, opened 2026-03-05
wasix(tests): Make test runnable on MacOS

artemyarulin:
- New wasix tests were disabled on MacOS completely, but it can actually run fine on Mac
- Main issue that it was using default engine (=Cranelift) which has very limited support on Mac, so for MacOS we simply switch to LLVM
- Also this PR drops LLVM optimisation to None for MacOS which makes tests dramatically faster to run and for tests we don't care about runtime performance that much

This is preparation PR for the upcoming POSIX tests syscalls merging as it will make my life a bit easier :)

### DIFF
--- a/lib/wasix/Cargo.toml
+++ b/lib/wasix/Cargo.toml
@@ -184,6 +184,13 @@ wasmer = { path = "../api", version = "=7.1.0-alpha.1", default-features = false
 	"cranelift",
 ] }
 
+[target.'cfg(all(not(target_arch = "wasm32"), target_os = "macos"))'.dev-dependencies]
+wasmer = { path = "../api", version = "=7.1.0-alpha.1", default-features = false, features = [
+	"wat",
+	"js-serializable-module",
+	"llvm",
+] }
+
 [target.'cfg(target_arch = "wasm32")'.dependencies.getrandom03]
 # Enable wasm_js for the getrandom 0.3.x instance pulled in by rand 0.9.
 package = "getrandom"
--- a/lib/wasix/tests/wasm_tests/mod.rs
+++ b/lib/wasix/tests/wasm_tests/mod.rs
@@ -20,6 +20,7 @@ use std::pin::Pin;
 use std::process::Command;
 use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll};
+use wasmer::{sys::EngineBuilder, sys::LLVMOptLevel, sys::Target};
 use wasmer_wasix::VirtualFile as VirtualFileTrait;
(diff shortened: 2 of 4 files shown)


lib/wasix/Cargo.toml
```diff
@@ -184,6 +184,13 @@ wasmer = { path = "../api", version = "=7.1.0-alpha.1", default-features = false
 	"cranelift",
 ] }
 
+[target.'cfg(all(not(target_arch = "wasm32"), target_os = "macos"))'.dev-dependencies]
+wasmer = { path = "../api", version = "=7.1.0-alpha.1", default-features = false, features = [
+	"wat",
+	"js-serializable-module",
+	"llvm",
```
marxin, line 191: I would rather include `llvm` unconditionally in the previous `[target.'cfg(all(not(target_arch = "wasm32"), target_os = "macos"))'.dev-dependencies]` section (even for e.g. Linux).

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -273,6 +274,35 @@ fn get_cache_dir() -> PathBuf {
...
+        // required by these tests, especially exception handling. Use the slower LLVM
+        // backend instead so the WASIX test suite can run reliably on macOS.
+        let target = Target::default();
+        let features = wasmer_types::Features::detect_from_wasm(wasm_bytes).unwrap_or_else(|_| {
+            wasmer::Engine::default_features_for_backend(&wasmer::BackendKind::LLVM, &target)
+        });
+
+        // Reduce opt level as it speeds up compilation dramatically and for tests no opt is fine
+        let mut compiler = wasmer::sys::LLVM::default();
+        compiler.opt_level(LLVMOptLevel::None);
```
marxin, line 290: Can you measure how much slower is that? Note I would rather stick to defaults for `LLVM` as it's going to make the testing more closer to a real use of the compiler.

Ideally, similarly to the existing old-school BASH wasix tests, we should run the tests for all compilers (Cranelift, LLVM). As shown in the previous PR, the Cranelift testing discovered yet another EH issue in the compiler.

marxin (changes requested):

marxin:
Note, we've currently disable macOS runs in CI for PRs. But if you want to measure it and test, you may want to temporarily disable:

```
        exclude:
          - is_pull_request: true
            metadata:
              build: macos-x64
```
in the following file: https://github.com/wasmerio/wasmer/blob/main/.github/workflows/test.yaml.

artemyarulin:
Replaced by https://github.com/wasmerio/wasmer/pull/6269

PR #6269, opened 2026-03-06
Migrate c-wasi-tests/WASI tests to the new test runner

artemyarulin:
~~Depends on #6268~~ 

- Migrate the rest of the c-wasi-tests to the new runner and delete associated build files with that
- Migrate some of the old wasi tests as well to the new runner. 
- Rest of the wasi tests requires some changes in the runner, so it will be migrated in the follow up PR

Part of the preparation work for POSIX test migration


artemyarulin:
Blocked by https://github.com/wasmerio/wasmer/issues/6324

Issue #5907, opened 2025-11-25
Alarm signals seem to be broken in WASIX


zebreus:
### Describe the bug

Alarm signals currently do nothing in WASIX.

### Steps to reproduce
1. Create the following C file:
```
#include <errno.h>
#include <signal.h>
#include <unistd.h>

static void handle_alarm(int sig) { _exit(1); }

int main(void) {
  signal(SIGALRM, handle_alarm);
  alarm(1);
  printf("Calling sleep\n");
  sleep(2);
  printf("Calling sleep\n");
  sleep(2);

  // We waited long enough and should have called handle_alarm by now
  // If we reach this point, the SIGALRM got lost.
  return 2;
}
```
2. Compile it for WASIX using wasixcc
3. Run with wasmer
4. Observe the exit code

### Expected behavior

It should exit with code 1 as specified in `handle_alarm`

### Actual behavior

It will exit with code 2, which indicates that the alarm signal got lost.

### Additional context

Required for configuring GNU sed

aeryz:
Hey @marxin. I had multiple issues open last night and mistakenly commented under #5904 while my investigation is all about this issue. Moving my comment here :)

I investigated this and realized that the bug is a cause of several different bugs:

## 1. Inconsistency between `alarm` and `setitimer` impls
The [alarm function sets `it_value`](https://github.com/wasix-org/wasix-libc/blob/a584e627718bf3e1eecf687ed592d636cf356b12/libc-top-half/musl/src/unistd/alarm.c#L7) when calling `setitimer` while `setitimer` only takes `tv_internal` into account. This results in the interval to be always `0` which eventually lead to the signal being lost [due to this](https://github.com/wasmerio/wasmer/blob/main/lib/wasix/src/os/task/process.rs#L740).

## 2. Inconsistency between `setitimer` and `proc_raise_interval`
See #6284.

## 3. Deadlock when handling intervals
See #6283 

If we can confirm #6284 is actually a bug, we can:
1. Merge #6283 
2. Decide on which unit to use and address #6284 
3. And eventually fix the `it_value` vs `it_interval` issue.

I would be happy to do the fixes.

aeryz:
Hey @zebreus and @marxin, in order for me to fully fix this issue, I need some clarification on [this issue](https://github.com/wasix-org/wasix-libc/issues/104). Could you check it out please when you have a moment? I'd like to confirm the expected behaviour before implementing the fix.

Issue #5904, opened 2025-11-25
Abort in a WASIX thread causes an exit with exitcode 0


zebreus:
### Describe the bug

Calling the WASIX abort() function (which is just a wrapper around __builtin_trap) from a thread causes the  wasmer to exit with code 0. The exitcode should actually be non-zero. Calling __builtin_trap from the main thread works and has a non-zero exit code.

### Steps to reproduce

Compile the following code to wasix WASM

```
#include <stdlib.h>
#include <stdio.h>
#include <assert.h>
#include <errno.h>
#include <pthread.h>

#ifndef NULL
#define NULL ((void*)0)
#endif

void* abort_in_thread(void* arg) {
    // Abnormal termination should never result in exit code 0
    abort();
}

int main() {
    pthread_t thread;
    pthread_create(&thread, NULL, abort_in_thread, NULL);
    pthread_join(thread, NULL);

    return 5;
}
```

Then run it with wasmer

### Expected behavior

Exit with a non-zero exit code

### Actual behavior

Exits with exitcode zero 

### Additional context

Calling abort in the main thread works as intended

zebreus:
@copilot try again

marxin:
Thank you for the investigations (and the initial PR). Feel free to continue working on it, we'll appreciate that.

PR #6283, opened 2026-03-10
fix: deadlock while signal handling with intervals

aeryz:
Hey, I'm new here!

While investigating #5907, I realized that there are several issues that eventually lead to `SIGALARM` not being handled. I also raised #6284 and explained another issue under #5904. But in this PR, I only fixed the part where the handling of the intervals deadlocks since the same Mutex is tried to be locked when the interval is hit.


theduke:
Fix is sensible. Thanks.

PR #6293, opened 2026-03-12
feat: use hash-based signature checks for indirect calls

marxin:
Right now we’re using a separate signature table, which requires an additional memory load. I’m proposing a SHA-256–based approach where each type signature is mapped to a hash, and we use 64 bits from the computed hash. I also added a module-level validation step to ensure we don’t end up with collisions.

Sample LLVM IR:
```llvmir
%sig_hash = load i64, ptr %sig_hash_ptr, align 8
%sig_hashes_equal = icmp eq i64 5823885051365845289, %sig_hash
```


syrusakbary:
Do we have benchmarks on this one @marxin ?

lib/vm/src/sig_registry.rs
```diff
@@ -37,39 +35,32 @@ impl SignatureRegistry {
...
-            }
+
+        if let Some(sig_hash) = inner.signature_to_hash.get(sig) {
+            return *sig_hash;
+        }
+
+        let sig_hash = VMSignatureHash::new(sig.signature_hash());
+        if inner.hash_to_signature.contains_key(&sig_hash) {
+            // We are checking type signature checking at the ModuleInfo level, thus the can't reach this point.
+            unreachable!("type signature collision");
```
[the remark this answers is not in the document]
marxin: I am aware of the limitation - covered with a comment.

marxin:
> Do we have benchmarks on this one @marxin ?

For PHP, the impact is rather small as all the hot functions share a common signature and so the current index lookup touches the same memory location. But still, I believe it's a valuable change we want to have in.

lib/types/src/types.rs
```diff
@@ -286,6 +288,24 @@ impl FunctionType {
     pub fn results(&self) -> &[Type] {
         &self.results
     }
+
+    /// Returns a stable 64-bit signature hash derived from the Wasm value types.
+    pub fn signature_hash(&self) -> u64 {
+        fn update_types(hasher: &mut Sha256, types: &[Type]) {
```
theduke: Doing a (somewhat expensive) sha256 and only taking the top bits seems a bit hacky to me.

Why not just use eg a CRC32?

That should not really run into collisions and should be better for perf, both startup (think: doing thousands of sha256 hashes for large modules) and for runtime.

marxin: Yeah, good idea. to be honest, even with the current SHA algorithm, I couldn't spot in the perf report. For packages like `php-32`, or `python`, we're speaking about ~600 different signature types that need to be hashed.

marxin:
Moved the `SignatureHash` into `ModuleInfo` which will make start up of an existing artifact much faster. No hash will be computed (except the host-fn imports).

lib/types/src/serialize.rs
```diff
@@ -13,7 +13,7 @@ pub struct MetadataHeader {
 impl MetadataHeader {
     /// Current ABI version. Increment this any time breaking changes are made
     /// to the format of the serialized data.
-    pub const CURRENT_VERSION: u32 = 16;
+    pub const CURRENT_VERSION: u32 = 18;
```
syrusakbary, line 16: Two jumps? Have we forgot to rebase from main?

lib/compiler/src/engine/inner.rs
```diff
@@ -157,16 +158,20 @@ impl Engine {
 
     /// Register a signature
     #[cfg(not(target_arch = "wasm32"))]
-    pub fn register_signature(&self, func_type: &FunctionType) -> VMSharedSignatureIndex {
+    pub fn register_signature(
+        &self,
+        func_type: &FunctionType,
+        hash: SignatureHash,
+    ) -> VMSignatureHash {
```
syrusakbary: The hash is computed directly from the func_type, making this an argument to pass to the function can make things drift.

Move back to:
```suggestion
    pub fn register_signature(&self, func_type: &FunctionType) -> VMSignatureHash {
```

And then inside the function, just do: 
```rust
        let hash = SignatureHash::new(func_type.signature_hash());
```

lib/types/src/serialize.rs
marxin: Another just happened in #6290 - got merged into `main`.

lib/compiler/src/engine/inner.rs
marxin: Like the suggested encapsulation!

PR #6295, opened 2026-03-12
Simplify and generalize runtime macros

Twey:
# Description

As part of https://github.com/wasmerio/wasmer/pull/5660 I want to be able to treat some VM entities a little differently to others (e.g. different type parameters).  Currently this involves a lot of code duplication: the boilerplate is mostly handled by the `define_vm_like` macro, but if I need to opt out of the macro temporarily I have to write the boilerplate myself.

To avoid doing this I:
- replaced the macro-generated `into_*` functions with `derive(Unwrap)` from `derive-more`, a crate already used in this codebase
- generalized the `gen_rt_ty` macro to handle custom lifetimes, type parameters, names, and visibilities, covering all the use cases I needed
- rewrote `define_vm_like` in terms of `gen_rt_ty`, which gives it a much smaller definition

### DIFF
--- a/lib/api/src/vm/mod.rs
+++ b/lib/api/src/vm/mod.rs
@@ -2,31 +2,16 @@
 
 mod impls;
 
-use crate::VMExternToExtern;
+use crate::{VMExternToExtern, macros::backend::gen_rt_ty};
 use wasmer_types::RawValue;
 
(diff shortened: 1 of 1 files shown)


marxin:
Thanks for a nice simplification PR.

PR #6298, opened 2026-03-13
fix(wasix): Fix connect_tcp readiness, and epoll del cleanup

artemyarulin:
- `EPOLL_CTL_DEL` removed subscriptions from the epoll map but did not detach their join guards, which allowed stale deleted subscriptions to interfere with later watchers on the same fd.
- `LocalNetworking::connect_tcp()`now returns an already connected TCP stream. Previously `connect_tcp()` returned `mio::net::TcpStream::connect(peer)` directly, which exposed a socket that could still be in the middle of a nonblocking connect. Some async guest networking paths expect `connect_tcp()` to complete connection establishment before the socket is handed back.

Basically that PR plus a previous one #6297 fixes that Puthon repro:
```python
import asyncio, httpx

def req_sync():
  print(httpx.get("https://example.com/", timeout=15.0).status_code) # works fine

async def main():
    async with httpx.AsyncClient(timeout=15.0) as c:
        print((await c.get("https://example.com/")).status_code) # ERROR TIMEOUT

asyncio.run(main())
```

### DIFF
--- a/lib/virtual-net/src/host.rs
+++ b/lib/virtual-net/src/host.rs
@@ -175,7 +175,16 @@ impl VirtualNetworking for LocalNetworking {
             return Err(NetworkError::PermissionDenied);
         }
 
-        let stream = mio::net::TcpStream::connect(peer).map_err(io_err_into_net_error)?;
+        // This future may be polled outside Tokio's executor context, so run
+        // the connect itself on the stored runtime handle to guarantee a reactor.
+        let stream = self
+            .handle
+            .spawn(tokio::net::TcpStream::connect(peer))
+            .await
+            .map_err(|_| NetworkError::IOError)?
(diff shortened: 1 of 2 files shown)


lib/virtual-net/src/host.rs
```diff
@@ -175,7 +175,17 @@ impl VirtualNetworking for LocalNetworking {
             return Err(NetworkError::PermissionDenied);
         }
 
-        let stream = mio::net::TcpStream::connect(peer).map_err(io_err_into_net_error)?;
+        let stream = self
+            .handle
+            .spawn(async move { tokio::net::TcpStream::connect(peer).await })
```
theduke: Why are we spaning a new task for the connect?

Issue #6279, opened 2026-03-10
`wasmer inspect php.wat | less` crashes with broken pipe


marxin:
I noticed that whenever we inspect a non-trivial file piped to `less`, then if I close the less command with `q`, I got:
```
thread 'main' (46578) panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
stack backtrace:
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Issue #6148, opened 2026-02-01
Non‑blocking pipe writes never return EAGAIN (unbounded pipe buffer)


artemyarulin:
When using a non‑blocking pipe `O_NONBLOCK`, `write()` should eventually return `-1` with `errno=EAGAIN/EWOULDBLOCK` once the pipe buffer is full. In WASIX it never does, because the pipe implementation is effectively unbounded. This breaks back pressure semantics and can cause OOM

```c
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <unistd.h>

int main(void) {
    int fds[2]; 
    pipe(fds);
    fcntl(fds[1], F_SETFL, fcntl(fds[1], F_GETFL, 0) | O_NONBLOCK);
    while (write(fds[1], "x", 1) == 1) {

    }
    // BUG: write never returns -1/EAGAIN on a non-blocking pipe
}
```

PR #6299, opened 2026-03-15
Fix/pipe posix compliance

valydumitru01:
# Description

Rewrite `pipe.rs` with a POSIX-compliant, byte-bounded buffer backed by
`Mutex<VecDeque<u8>>`, replacing the previous mpsc channel.

**Pipe fixes:**
- Enforce PIPE_CAPACITY (64 KB) at byte level, not message count
- Enforce PIPE_BUF (4096) atomicity guarantee for writes
- Correct WouldBlock vs BrokenPipe error mapping
- Fix AsyncWrite which was silently using blocking `tx.send()`
- Three-layer wake system: Condvar (blocking threads), Waker (async
  tasks), InterestHandler (external event loop)

**fd_write.rs refactor** (no behavioral changes outside of pipe paths):
- Extract write_to_file, write_to_socket, write_to_pipe,
  write_to_event_notifications, write_to_buffer
- Unify PipeTx and DuplexPipe arms via `write_to_pipe(&mut dyn Write)`
- Introduce WriteOutcome struct and PipeWriteResult enum for type-safe
  broken-pipe signaling

**Tests:**
- 27 unit tests covering the full pipe contract
- E2E C tests: fork, vfork, SIGPIPE, EAGAIN, atomicity, dup2,
  drain-resume cycle, capacity bounds

# Motivation

The pipe implementation was not POSIX-conformant and had multiple reported
bugs: #6279, #6148.

The fd_write refactor was done alongside for readability and to cleanly
support the new pipe error semantics (WouldBlock now surfaces as EAGAIN
through the existing `map_io_err` path).


marxin:
@Arshia001 can you make a review?

Issue #6017, opened 2026-01-06
LLVM(LoongArch64): some simd spec tests crash in LLVM


marxin:
Apparently, the following tests crash when LLVM emits code:
```
     SIGSEGV [   0.271s] wasmer-workspace::compilers wast::spec::simd::simd_f32x4::llvm::llvm
     SIGSEGV [   0.181s] wasmer-workspace::compilers wast::spec::simd::simd_f64x2::llvm::llvm
     SIGSEGV [   0.082s] wasmer-workspace::compilers wast::spec::simd::simd_splat::llvm::llvm
     SIGSEGV [   0.421s] wasmer-workspace::compilers wast::spec::simd::simd_load::llvm::llvm
     SIGSEGV [   0.338s] wasmer-workspace::compilers wast::spec::simd_f32x4::llvm::llvm
     SIGSEGV [   0.186s] wasmer-workspace::compilers wast::spec::simd_f64x2::llvm::llvm
     SIGSEGV [   0.364s] wasmer-workspace::compilers wast::spec::simd_load::llvm::llvm
     SIGSEGV [   0.261s] wasmer-workspace::compilers wast::spec::simd_splat::llvm::llvm
```

Back-trace:
```
(gdb) bt
#0  0x00007fffed7eed40 in llvm::SelectionDAG::ReplaceAllUsesWith(llvm::SDNode*, llvm::SDValue const*) () from /lib/libLLVM.so.21.1
#1  0x00007fffed5bec5c in ?? () from /lib/libLLVM.so.21.1
#2  0x00007fffed5d8ae0 in ?? () from /lib/libLLVM.so.21.1
#3  0x00007fffed5e5114 in ?? () from /lib/libLLVM.so.21.1
#4  0x00007fffed5e8384 in llvm::SelectionDAG::Legalize() () from /lib/libLLVM.so.21.1
#5  0x00007fffed817a10 in llvm::SelectionDAGISel::CodeGenAndEmitDAG() () from /lib/libLLVM.so.21.1
#6  0x00007fffed819d90 in llvm::SelectionDAGISel::SelectAllBasicBlocks(llvm::Function const&) () from /lib/libLLVM.so.21.1
#7  0x00007fffed81b674 in llvm::SelectionDAGISel::runOnMachineFunction(llvm::MachineFunction&) () from /lib/libLLVM.so.21.1
[...]

PR #6300, opened 2026-03-16
deps(LLVM): bump to 22.1

marxin:
The PR is waiting for a proper Inkwell release, but would be nice to have it fully tested upfront before it goes out of the door.


theduke:
What's the benefit of requiring 22.1 ?

Are there any significant changes we'd like to pull in?

theduke:
Ah, I thought we were on 22.0, not 21.

Makes sense.

marxin:
> Ah, I thought we were on 22.0, not 21.

Yeah! A big change should be fix for LoongArch64 SIMD: #6017.

marxin:
Will be reopened once a `inkwell` release is available.

PR #6309, opened 2026-03-16
Add `proc_raise_interval2` syscall

### LINKED ISSUE #6284: Inconsistent time unit for intervals in signal handling
### Describe the bug

While investigating #5904, when the exact same code is compiled, I realized that the `wasix-libc` [passes the timestamp in nanoseconds](https://github.com/wasix-org/wasix-libc/blob/a584e627718bf3e1eecf687ed592d636cf356b12/libc-top-half/musl/src/signal/setitimer.c#L34) while the [proc_raise_interval function expects it in milliseconds](https://github.com/wasmerio/wasmer/blob/main/lib/wasix/src/syscalls/wasi/proc_raise.rs#L35).

Dropping the corresponding function in the WAT file as a reference as well:
```
     (call $__wasi_proc_raise_interval
      (i32.const 14)
      (i64.add
       (i64.mul
        (i64.load offset=8
         (local.get $1)
        )
        (i64.const 1000)
       )
       (i64.mul
        (i64.load
         (local.get $1)
        )
        (i64.const 1000000000)
       )
      )
      (i32.const 1)
     )
    )
```

aeryz:
Upon investigating #5907, I realized that [proc_raise_interval](https://github.com/wasmerio/wasmer/blob/4e3af188e22c41185cf1117d240245d62d40d3ef/lib/wasix/src/syscalls/wasi/proc_raise.rs#L26) is not sufficient for a complete [setitimer](https://linux.die.net/man/2/setitimer) implementation. The details are explained in depth [here](https://github.com/wasix-org/wasix-libc/issues/104). 

Under the issue, @zebreus [suggested that](https://github.com/wasix-org/wasix-libc/issues/104#issuecomment-4042997144) we introduce a new syscall called `proc_raise_interval2` which closely follows `setitimer`. 

This PR is an attempt to add it. Since I'm new here and don't know the process of introducing a new API, I wanted to create the draft PR and show how would it look like. In addition to this PR, here are the changes done to other repos:
1. **wasix-libc**: [Generated the C bindings](https://github.com/wasix-org/wasix-libc/compare/main...aeryz:wasix-libc:add/proc-raise-interval2)
2. **WASI**: [Added the witx definitions](https://github.com/wasix-org/wasix-witx/commit/a92d636b6fa0fe5072429105f76875f0c8b50a47)

Additionally, `proc_raise_interval` syscall was previously missing the support for repetition which is now supported for both `proc_raise_interval` and `proc_raise_interval2`.

Fixes: #6284 

Also note that, due to the sync nature of how the signal handling works, the runtime only checks for the intervals during syscalls. This means on CPU-heavy applications, the signal timer will not work properly (it will never execute before the interval but might take longer). We can note this as a known limitation for now and in the future work on async signal handling. 

If we agree on adding this syscall, I will also add some e2e tests.


zebreus (commented):
I can do a proper review once this is no longer a draft. @Arshia001 @theduke what do you think of this change?

Please make sure this does not break existing use of `proc_wasix_interval`. When you add the tests for `proc_raise_interval` and `proc_raise_interval2` add them in  `lib/wasix/tests/wasm_tests` and not in `tests/wasix`.

Also the `proc_raise_interval2` implementation should live in its own file in `syscalls/wasix`.

Usually we make the old version just forward to the new implementation of the syscall but this might be hard here given the new signature reads the values from memory. Maybe create a shared helper or adjust the signature?

I agree that we should add documentation on the limitations of our signal handling implementation,. For now a comment to the syscall implementation is probably good enough.

lib/wasix/src/syscalls/wasi/proc_raise.rs
```diff
@@ -29,15 +29,84 @@ pub fn proc_raise_interval(
     interval: Timestamp,
     repeat: Bool,
 ) -> Result<Errno, WasiError> {
+    println!("called regular raise");
```
theduke: Leftover print.

lib/wasix/src/syscalls/wasi/proc_raise.rs
```diff
@@ -29,15 +29,84 @@ pub fn proc_raise_interval(
...
+    let env = ctx.data();
+    let memory = unsafe { env.memory_view(&ctx) };
+
+    let new_ptr = new_ptr.deref(&memory);
+    let new = wasi_try_ok!(new_ptr.read().map_err(crate::mem_error_to_wasi));
+
+    let value = timeval_to_duration(new.value);
+    let interval = timeval_to_duration(new.interval);
+
+    println!("duration: {value:?} {interval:?}");
```
theduke: Leftover print.

lib/wasix/src/syscalls/wasi/proc_raise.rs
```diff
@@ -29,15 +29,84 @@ pub fn proc_raise_interval(
...
+            Some(old_value) => __wasi_itimerval_t {
+                interval: if let Some(interval) = old_value.interval {
+                    __wasi_timeval_t {
+                        sec: interval.as_secs(),
+                        usec: interval.subsec_micros().into(),
+                    }
+                } else {
+                    zero
+                },
+                value: __wasi_timeval_t {
```
theduke: If we are sticking to Linux semantics of `setitimer` here, which I reckon we should, then the old value is supposed to be the time until the next trigger time, not the originally configured time.

lib/wasix/src/syscalls/wasi/proc_raise.rs
```diff
@@ -29,15 +29,84 @@ pub fn proc_raise_interval(
     interval: Timestamp,
     repeat: Bool,
 ) -> Result<Errno, WasiError> {
+    println!("called regular raise");
     let env = ctx.data();
     let interval = match interval {
```
theduke, line 34: As @zebreus said, an we implement a helper function that uses Rust types, and make both old and new implementation use the helper, to avoid duplication?

lib/wasix/src/syscalls/wasi/proc_raise.rs
```diff
@@ -29,15 +29,84 @@ pub fn proc_raise_interval(
...
+    Ok(Errno::Success)
+}
+
+fn timeval_to_duration(tv: __wasi_timeval_t) -> Option<Duration> {
+    if tv.sec == 0 && tv.usec == 0 {
+        None
+    } else {
+        Some(Duration::from_secs(tv.sec) + Duration::from_micros(tv.usec))
+    }
+}
```
theduke: Should add a test/tests in the tests/wasix test suite.

The pattern of existing tests should be easy to emulate. 
Note: you'll need wasixcc ( https://github.com/wasix-org/wasixcc )

lib/wasix/src/syscalls/wasi/proc_raise.rs
```diff
@@ -29,15 +29,84 @@ pub fn proc_raise_interval(
...
+        wasi_try_ok!(ret_ptr.write(ret_timer).map_err(crate::mem_error_to_wasi));
+    }
+
+    WasiEnv::do_pending_operations(&mut ctx)?;
+
+    Ok(Errno::Success)
+}
+
+fn timeval_to_duration(tv: __wasi_timeval_t) -> Option<Duration> {
+    if tv.sec == 0 && tv.usec == 0 {
```
theduke: I believe this can overflow the range of Duration.
Should probably return an error here, so the caller gets back an `Errno::INVAL`.

theduke (changes requested):
Thanks!

The direction makes sense.

Needs some modifications.

lib/wasix/src/syscalls/wasi/proc_raise.rs
aeryz: Oh good catch! My understanding was when a timer is hit, the `value` becomes `interval` and `interval` stays as is. It's actually true but I missed the part where `value` always gives the time until the next trigger time. 
Relevant info: [posix settimer](https://man7.org/linux/man-pages/man2/timer_settime.2.html), [setitimer](https://man7.org/linux/man-pages/man2/setitimer.2.html).

I'll return the `value - now - last_signal` instead.

lib/wasix/src/syscalls/wasi/proc_raise.rs
aeryz: Yeah better to use `checked_add` here

lib/wasix/src/syscalls/wasi/proc_raise.rs
aeryz: Yeah yeah, I wanted to open a PR to have a minimal PoC on the new syscall. Since we are good with the API now, I will do the refactors and add the tests to complete the PR.

lib/wasix/src/syscalls/wasi/proc_raise.rs
aeryz: Fixed [here](https://github.com/wasmerio/wasmer/pull/6309/changes#diff-9cb3aba0b96987c35a4d965dca02fd42061438e1efb1b3132b464ba3bd740884R42)

lib/wasix/src/syscalls/wasi/proc_raise.rs
aeryz: Giving it some more thought,  I think the functions are currently already implemented in a way that the main logic is shared. Note that the shared logic is:
1. timer is set with the given signal
2. previous timer is removed when the timer.value is `0`
Both of these are already implemented by `WasiProcess::signal_interval`. The functions differ in:
1. `proc_raise_interval` works with millisecond timestamps, but `proc_raise_interval2` works with `itimerval`
2. `proc_raise_interval2` makes use of the old timer.
These are already implemented seperately and the only duplicate part is `WasiEnv::do_pending_operations` which I would say it's good to keep it explicit per syscall unless the legacy implementation is directly forwarded to the new implementation.  

And in terms of the use of proper Rust types. I kept the fields of `__wasi_timeval_t` private and make it convertible to `Duration`. So if anyone wants to do any meaningful thing with `__wasi_timeval_t`, they would have to convert it to `Duration` first. And our internal Rust types and the signal interval related functions all use `Duration`s. Please let me know if you see a possible footgun.

lib/wasix/src/syscalls/wasi/proc_raise.rs
aeryz: Tests are added under `libc_tests` for `alarm` and `setitimer`.

lib/wasix/src/syscalls/wasi/proc_raise.rs
aeryz: Fixed [here](https://github.com/wasmerio/wasmer/pull/6309/changes#diff-93a5ab428871750a388e0e51eed5570f1ebcf2b6b6fcac51392f342982e91df6R329)

PR #6316, opened 2026-03-17
chore: update option help output + hide ignored --enable-pass-params-opt

marxin:


marxin:
> I kind of like that the naming corresponds to flags, like `--llvm`, and the corresponding cargo features.

Oh, got it, hasn't thought about it this way.
But then we should name it like: `Available for --cranelift, --llvm and --singlepass.`, right?

PR #6320, opened 2026-03-18
CI: ignore some heavy integration tests on Windows and macOS

marxin:
Since #6315, the compile time for Python has increased again and thus we face long build times, especially on Windows.


marxin:
> Windows is sort of a B tier platform, so I could see ignoring Windows, but Mac OS seems problematic.

Well, right now, we completely disabled macOS as it's pretty slow - let me test it as part of this PR.

marxin:
Ok, so for macOS, we have it disabled right now. And it seems the tests are also terribly slow (timeouts after 360s due to parallel LLVM compilation). That said, I would like to disable a couple of tests, while keeping and enabling macOS runs on `main`.

PR #6333, opened 2026-03-20
NAPI: split into 2 versioned parts

marxin:
Right now, we’re using the `napi` namespace for all functionality, but for version compatibility it would be better to introduce a versioned namespace, separating the official APIs (`napi`) from the unofficial one (`napi_extension_wasmer`).


marxin:
Superseded by #6353.

PR #6323, opened 2026-03-18
Add app secrets import and export commands

0x-pankaj:
## Summary
Added two new CLI commands for Wasmer Edge app secrets management to solve issue #6258.
### Problem
Previously, users had no straightforward way to export/import app secrets with values. The `wasmer app secrets list -f json` only returns metadata (created/updated timestamps), so exporting required scripting to list names and call `wasmer app secrets reveal` per secret.
### Solution
Added two new commands:
1. **`wasmer app secrets export`** - Export secrets from Wasmer Edge to local file
   
   ```bash
   # Export to stdout (default env format)
   wasmer app secrets export --app namespace/app-name
   
   # Export to file
   wasmer app secrets export --app namespace/app-name -o secrets.env
   
   # Export as JSON/YAML
   wasmer app secrets export --app namespace/app-name -f json
   ```
2. **`wasmer app secrets import`** - Import secrets from local file to Wasmer Edge
   
   ```bash
   # Import from env file
   wasmer app secrets import --app namespace/app-name --from-file secrets.env
   
   # Import and update existing secrets
   wasmer app secrets import --app namespace/app-name --from-file secrets.env --update-existing
   
   # Import with auto-redeploy
   wasmer app secrets import --app namespace/app-name --from-file secrets.env --redeploy
   ```
### Features
- **Export formats**: env (default), JSON, YAML
- **Import formats**: env (default), JSON, YAML
- **Output**: stdout or file (`-o, --output`)
- **Import options**: 
  - `--update-existing`: Update secrets that already exist
  - `--redeploy`: Auto-redeploy after import
### Testing
All commands tested on real Wasmer Edge deployment:
| Test Case | Status |
|-----------|--------|
| Export to stdout (env format) | ✅ Pass |
| Export to file | ✅ Pass |
| Export as JSON | ✅ Pass |
| Export as YAML | ✅ Pass |
| Import from env file | ✅ Pass |
| Import from JSON file | ✅ Pass |
| Import from YAML file | ✅ Pass |
| Import with --update-existing | ✅ Pass |
| Import with --redeploy | ✅ Pass |
### Files Changed
- `lib/cli/src/commands/app/secrets/export.rs` (new)
- `lib/cli/src/commands/app/secrets/import.rs` (new)
- `lib/cli/src/commands/app/secrets/mod.rs` (modified)
### Visual Flow
```
EXPORT:  Wasmer Edge (Cloud) ──> Local File (secrets.env)
[...]

PR #6346, opened 2026-03-24
fix(cli): --quiet blocks some operations

marxin:
CC: @0x-pankaj

Related: #6323


0x-pankaj:
Thanks @marxin for the follow-up fix!. This make sense the --quite flag shouldn't block functional operations.

Issue #6024, opened 2026-01-09
Testing cranelift on RISC-V


wakabat:
Hey, so we began testing wasmer's cranelift compiler on riscv64 platform as well. And just want to share something we find.

## Unaligned load / store

We disallow unaligned loads / stores in our platform. So like the singlepass compiler, we hit on unaligned loads / stores immediately. I put together a [fix](https://github.com/wakabat/official-wasmer/commit/da498b34492c52515c7faa91ffcb208a7460dc90) that works for our tests. However I'm not sure if wasmer can simply include this fix. There are performance implications. Maybe it makes sense to add a feature that enables this optionally?

## Difference in trapping

I notice trapping, or at least trapping triggered by some opcodes, are implemented by calling `wasmer_vm_raise_trap` in both LLVM compiler and singlepass compiler:

* https://github.com/wasmerio/wasmer/blob/bd91673c1c3976d3b01d2dffd80c7ec33effc67f/lib/compiler-llvm/src/translator/code.rs#L2471-L2475
* https://github.com/wasmerio/wasmer/blob/bd91673c1c3976d3b01d2dffd80c7ec33effc67f/lib/compiler-singlepass/src/codegen.rs#L3407-L3433

But the cranelift compiler [traps](https://github.com/wasmerio/wasmer/blob/bd91673c1c3976d3b01d2dffd80c7ec33effc67f/lib/compiler-cranelift/src/translator/code_translator.rs#L260) simply, and relies on OS-level signal handler to process the trap.

We are running wasmer in a special environment, which can be thought as a bare-metal embedded system. So OS signal handler is not available. So the question is: is there any reason the cranelift compiler simply traps? Can we change the cranelift compiler to call `wasmer_vm_raise_trap` as well?

As always, thanks for the help!

marxin:
The issue can be closed now - both sub-issues were covered by the linked PRs.

PR #6348, opened 2026-03-24
feat(Cranelift): use call to wasmer_vm_raise_trap for Unreachable

marxin:
Similarly to other compilers, we should directly call `wasmer_vm_raise_trap` and not rely on the signal handling.

Implements: #6024

### DIFF
--- a/lib/compiler-cranelift/src/func_environ.rs
+++ b/lib/compiler-cranelift/src/func_environ.rs
@@ -1305,7 +1305,9 @@ impl BaseFuncEnvironment for FuncEnvironment<'_> {
         let trap_code = pos
             .ins()
             .iconst(I32, wasmer_types::TrapCode::UnreachableCodeReached as i64);
-        pos.ins().call_indirect(func_sig, func_addr, &[trap_code]);
+        builder
+            .ins()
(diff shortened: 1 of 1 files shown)


lib/compiler-cranelift/src/func_environ.rs
```diff
@@ -1285,6 +1298,22 @@ impl BaseFuncEnvironment for FuncEnvironment<'_> {
...
+    fn translate_unreachable(&mut self, builder: &mut FunctionBuilder) -> WasmResult<()> {
+        let (func_sig, func_idx) = self.get_raise_trap_func(builder.func);
+        let mut pos = builder.cursor();
+        let (_, func_addr) = self.translate_load_builtin_function_address(&mut pos, func_idx);
+        let trap_code = pos
+            .ins()
+            .iconst(I32, wasmer_types::TrapCode::UnreachableCodeReached as i64);
+        builder
+            .ins()
+            .call_indirect(func_sig, func_addr, &[trap_code]);
```
syrusakbary, line 1310: Why call indirect?

marxin: It's the host VM function that are called through `VMBuiltinFunctionsArray` offset.

PR #6351, opened 2026-03-25
feat(Singlepass): add config for strict memory boundary checking

marxin:
So far, we've been using the strict boundary checking for the RISC-V target of the Singlepass compiler. Works fine, however, it makes the memory operations rather slow and so, similarly to other targets, let's rely on the access violation memory signals.

@tsahee @wakabat 
For preserving the old behavior, I added a new config option `strict_memory_boundary_checks`.


lib/compiler-singlepass/src/machine_riscv.rs
```diff
@@ -938,23 +939,25 @@ impl MachineRiscv {
         self.emit_relaxed_load(Size::S64, false, Location::GPR(tmp_base), base_loc)?;
 
         // Load bound into temporary register.
-        self.emit_relaxed_load(Size::S64, false, Location::GPR(tmp_bound), bound_loc)?;
+        if need_check {
+            self.emit_relaxed_load(Size::S64, false, Location::GPR(tmp_bound), bound_loc)?;
```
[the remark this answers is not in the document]
marxin: It's OK, not leading to any run-time cost.

lib/compiler-singlepass/src/codegen.rs
```diff
@@ -837,10 +837,11 @@ impl<'a, M: Machine> FuncGen<'a, M> {
...
-            MemoryStyle::Static { .. } => false,
-            MemoryStyle::Dynamic { .. } => true,
-        };
+        let need_check = self.config.strict_memory_boundary_checks
+            || match self.memory_styles[MemoryIndex::new(0)] {
+                MemoryStyle::Static { .. } => false,
+                MemoryStyle::Dynamic { .. } => true,
+            };
 
         let offset = if self.module.num_imported_memories != 0 {
```
theduke, line 848: Does singlepass not support multiple memories?

lib/compiler-singlepass/src/config.rs
```diff
@@ -92,12 +95,26 @@ impl Singlepass {
...
         Self {
             enable_nan_canonicalization: true,
+            strict_memory_boundary_checks: false,
             middlewares: vec![],
             callbacks: None,
             num_threads: std::thread::available_parallelism().unwrap_or(NonZero::new(1).unwrap()),
         }
     }
 
+    /// Configure whether Singlepass should always emit strict memory boundary
```
theduke, line 105: Shouldn't a major behavior change like this use the old behaviour by default?

We could switch the default with the next major version?

lib/compiler-singlepass/src/codegen.rs
marxin: No. Right now, it's unsupported features by all Wasmer compilers: https://webassembly.org/features/

lib/compiler-singlepass/src/config.rs
marxin: Given the change is tightly communicated with our customer, we can do it. Moreover, it's going to align with other Singlepass platforms.

theduke: Our customers are a limited subset of our users though.

In general I'm against significant behaviour changes in minor versions.

Guess should have @syrusakbary chime in.

marxin: Sure, that’s a good point! Keep in mind that the current RISC-V port of Singlepass is still considered experimental, so we may have some flexibility to make this change.

As for minor versions, it’s a bit tricky. On one hand, we aim to do a major release once a year, but in reality, many minor releases introduce breaking changes and arguably should be treated as major versions. This might be worth discussing further.

syrusakbary (changes requested):
The strict memory boundary checking is what MemoryStyle::Dynamic implies.

All MemoryStyle::Dynamic memories must do strict memory boundary checking, so we don't need an extra setting?

marxin:
> The strict memory boundary checking is what `MemoryStyle::Dynamic` implies.

Yes, that behavior has been preserved.

> All `MemoryStyle::Dynamic` memories must perform strict boundary checking, so we don’t need an extra setting?

The situation comes from a divergence in the original Singlepass RISC-V implementation compared to x86_64 and AArch64. On those architectures, `MemoryStyle::Static` relies on signal handling to catch out-of-bounds accesses.

When I added `rv64gc` support, I mistakenly enabled strict boundary checking by default to comply with the spec. The newly introduced `strict_memory_boundary_checks` flag is still useful, though - particularly for scenarios like running modules on bare metal (e.g., Offchain use cases).

marxin:
@theduke review please

theduke (approved):
After some discussions, we have decided that aligning the behaviour across architectures is the right thing to do.

The behaviour change should be called out loudly in the next change log though.

lib/compiler-singlepass/src/machine_riscv.rs
```diff
@@ -900,6 +900,7 @@ impl MachineRiscv {
         memarg: &MemArg,
         check_alignment: bool,
         value_size: usize,
+        need_check: bool,
```
syrusakbary, line 903: Why do we have `need_check` and `check_alignment` ?

lib/compiler-singlepass/src/machine_riscv.rs
```diff
@@ -939,23 +940,25 @@ impl MachineRiscv {
         self.emit_relaxed_load(Size::S64, false, Location::GPR(tmp_base), base_loc)?;
 
         // Load bound into temporary register.
-        self.emit_relaxed_load(Size::S64, false, Location::GPR(tmp_bound), bound_loc)?;
+        if need_check {
```
syrusakbary, line 943: This should be `if check_alignment`

lib/compiler-singlepass/src/machine_riscv.rs
```diff
@@ -1017,17 +1020,22 @@ impl MachineRiscv {
...
         let cond = tmp_base;
 
-        // Trap if the end address of the requested area is above that of the linear memory.
-        self.assembler.emit_cmp(
-            Condition::Le,
-            Location::GPR(tmp_addr),
-            Location::GPR(tmp_bound),
-            Location::GPR(cond),
-        )?;
+        if need_check {
```
syrusakbary, line 1023: `if check_alignment`

lib/compiler-singlepass/src/config.rs
```diff
@@ -92,12 +95,26 @@ impl Singlepass {
     pub fn new() -> Self {
         Self {
             enable_nan_canonicalization: true,
+            strict_memory_boundary_checks: false,
```
syrusakbary, line 98: This should be implied from the memory used. We don't need an extra var

lib/compiler-singlepass/src/codegen.rs
```diff
@@ -839,10 +839,11 @@ impl<'a, M: Machine> FuncGen<'a, M> {
...
     ) -> Result<(), CompileError> {
-        let need_check = match self.memory_styles[MemoryIndex::new(0)] {
-            MemoryStyle::Static { .. } => false,
-            MemoryStyle::Dynamic { .. } => true,
-        };
+        let need_check = self.config.strict_memory_boundary_checks
+            || match self.memory_styles[MemoryIndex::new(0)] {
+                MemoryStyle::Static { .. } => false,
+                MemoryStyle::Dynamic { .. } => true,
+            };
```
syrusakbary, line 846: This is a red-herring. People should be using `MemoryStyle::Dynamic` if they need strict boundary checks

syrusakbary (changes requested):
Not aligned with the PR.

We are adding a new variable: `strict_memory_boundary_checks`, rather than telling the user to use `MemoryStyle::Dynamic` for that memory. -> `MemoryStyle::Dynamic` literally means that the memory needs strict boundary checks.

The, we have the diverging logic to check if needs check.

It needs check just only if the memory style is dynamic.

I don't see any use case for memory not being dynamic and needing a check. If needs a check -> Memory should be dynamic

marxin:
> We are adding a new variable: `strict_memory_boundary_checks`, rather than telling the user to use `MemoryStyle::Dynamic` for that memory. -> `MemoryStyle::Dynamic` literally means that the memory needs strict boundary checks.

No, you're right, we're making the situation more complex! I'm aligned with your.

@tamirhemo can you double-check `MemoryStyle::Dynamic` memory style will serve your needs for the RISC-V simulator you have?

tamirhemo:
> > We are adding a new variable: `strict_memory_boundary_checks`, rather than telling the user to use `MemoryStyle::Dynamic` for that memory. -> `MemoryStyle::Dynamic` literally means that the memory needs strict boundary checks.
> 
> No, you're right, we're making the situation more complex! I'm aligned with your.
> 
> @tamirhemo can you double-check `MemoryStyle::Dynamic` memory style will serve your needs for the RISC-V simulator you have?

I am actually hoping we could use your new strict boundary checking, i.e. relying on traps. In our platform, we have the ability to protect memory (something like an `mprotect` syscall) that would then trap on an unauthorized write/access. Would this be enough to satisfy the needs of static? it wasn't clear to me from the code changes

marxin:
> In our platform, we have the ability to protect memory (something like an mprotect syscall) that would then trap on an unauthorized write/access. Would this be enough to satisfy the needs of static? it wasn't clear to me from the code changes

Yes, that would work as expected with the `MemoryStyle::Static`. So please give it a try and if you spot any missing gaps, we can take a look at it. Generally speaking, the Static mode provide a better performance characteristics.

PR #6354, opened 2026-03-25
feat(Singlepass, Cranelift): add --enable-nonaligned-memory-accesses

marxin:
@tsahee @wakabat

Right now, the RISC-V single-pass compiler emits 1-byte loads and stores (as a runtime check) to avoid strict alignment violations. This PR changes the default to use full-size native load/store operations, while still providing an option to retain the previous behavior. Similarly, for Cranelift, we now support this functionality by incorporating @wakabat’s changes.

Relates: #6024


marxin:
Superseded in #6388.

PR #6365, opened 2026-03-27
ci: make napi-v8 opt-in + fix header file installation

marxin:
For Linux distribution packaging, disable `napi-v8` by default and make it available only through our official builds.


Makefile
```diff
@@ -954,7 +954,9 @@ install-wasmer:
 	install -Dm755 target/release/wasmer $(DESTDIR)/bin/wasmer
 
 install-capi-headers:
-	for header in lib/c-api/*.h; do install -Dm644 "$$header" $(DESTDIR)/include/$$(basename $$header); done
```
theduke, line 945: This could make us miss headers when added.

(unlikely, but always possible...)

Would we see CI failures in that case?

marxin: Yeah, but it's probably unlikely to happen.

PR #6367, opened 2026-03-27
Simplify FunctionMiddleware trait: remove lifetime parameter

bragaigor:
# Description

- Add lifetime parameter to FunctionMiddleware trait for more expressive borrowing
- Add locals_info() callback so middlewares can inspect function local types before operator processing
- Add From<MiddlewareError> for CompileError for ergonomic error handling
- Deduplicate store handle lookups in Global get()/set()

 ## Motivation                                                                                                                                                                                  
                                                                                                                                                                                              
OffchainLabs maintains a fork of wasmer with changes needed for Arbitrum Stylus (on-chain WASM execution). These changes have been carried across multiple wasmer version upgrades (v4.2.3 → v4.3.7 → v7.1.0). Proposing them upstream would reduce the maintenance burden and benefit other wasmer users who need richer middleware capabilities. 

The current middleware API prevents implementing middlewares that need to buffer operators or know about function locals. Specifically:

1. Operator buffering: With the lifetime on `feed` (`fn feed<'a>(..., op: Operator<'a>)`), a middleware cannot store `Operator<'a>` in its own fields — `'a` is a per-call parameter unrelated to `Self`. Moving it to the trait (`FunctionMiddleware<'a>`) lets implementations accumulate operators. This is needed for whole-function analysis passes (e.g., computing worst-case stack depth before emitting instrumented output).                                                                                                                             
                                                                                                                                                                                              
2. Locals info: Middleware that computes stack frame sizes needs the number of locals, but locals are parsed before operators and never passed to the middleware. The `locals_info` callback fills this gap. It has a default no-op implementation, so existing code is unaffected.

### DIFF
--- a/lib/api/src/backend/sys/entities/global.rs
+++ b/lib/api/src/backend/sys/entities/global.rs
@@ -48,17 +48,9 @@ impl Global {
 
     pub(crate) fn get(&self, store: &mut impl AsStoreMut) -> Value {
         unsafe {
-            let raw = self
-                .handle
-                .get(store.as_store_ref().objects().as_sys())
-                .vmglobal()
-                .as_ref()
-                .val;
-            let ty = self
-                .handle
-                .get(store.as_store_ref().objects().as_sys())
-                .ty()
-                .ty;
+            let global = self.handle.get(store.as_store_ref().objects().as_sys());
+            let raw = global.vmglobal().as_ref().val;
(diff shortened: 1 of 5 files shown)


bragaigor:
will open a new PR against https://github.com/OffchainLabs/wasmer-upstream

PR #6368, opened 2026-03-27
Expose stack pool drain and stack size getters

bragaigor:
# Description

- Expose `MAX_STACK_SIZE` constant (100 MB) and `get_stack_size()` getter for the process-wide default                                  
- Move `STACK_POOL` from function-local static to module-level static so it can be accessed by the new `drain_stack_pool()` function
- Add `drain_stack_pool()` to clear cached coroutine stacks before retrying with a larger size — without this, the pool serves cached undersized stacks indefinitely (see Background)                                                                                         
- Add comments documenting that the pool is not size-aware and callers must drain before resizing   

The `STACK_POOL` caches coroutine stacks for reuse but does not track their allocation size. When the default stack size is increased at runtime (e.g., doubling on native stack overflow), the pool continues to serve old, smaller stacks:

1. Normal calls fill pool with 1MB stacks                                                   
2. Call overflows → set default to 2MB → retry
3. Retry pops 1MB stack from pool → overflows again
4. scopeguard returns 1MB stack to pool → infinite loop

`drain_stack_pool()` breaks this cycle by clearing all cached stacks so the next allocation uses the new size.

### DIFF
--- a/lib/vm/src/trap/traphandlers.rs
+++ b/lib/vm/src/trap/traphandlers.rs
@@ -40,6 +40,9 @@ static MAGIC: u8 = 0xc0;
 
 static DEFAULT_STACK_SIZE: AtomicUsize = AtomicUsize::new(1024 * 1024);
 
+/// Maximum allowed stack size (100 MB).
+pub const MAX_STACK_SIZE: usize = 100 * 1024 * 1024;
+
 // Current definition of `ucontext_t` in the `libc` crate is incorrect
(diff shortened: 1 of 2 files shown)


lib/vm/src/trap/traphandlers.rs
bragaigor, line 1: add tests

bragaigor:
Closing this PR in pro of https://github.com/wasmerio/wasmer/pull/6374 (base branch from https://github.com/OffchainLabs/wasmer-upstream)

Issue #6343, opened 2026-03-24
Implement "readonly" `funcref` `Table`


marxin:
It's something than can fasten the `indirect_call` for PHPIX (#6181).

PR #6370, opened 2026-03-30
feat: readonly fixed funcref Table

marxin:
The change allows a more relaxed run-time checks when it comes to locally known and read-only `funcref` Tables as emitted by wasixcc. Leveraging the optimization we get 5% speed-up for the PHPIX benchmark.

Resolves: #6343

### DIFF
--- a/lib/vm/src/instance/mod.rs
+++ b/lib/vm/src/instance/mod.rs
@@ -1593,30 +1593,28 @@ fn initialize_tables(instance: &mut Instance) -> Result<(), Trap> {
             return Err(Trap::lib(TrapCode::TableAccessOutOfBounds));
         }
 
-        // Disable read-only check as we're initializing the table.
-        let is_readonly = table.ty().readonly;
-        table.ty_as_mut().readonly = false;
         if let wasmer_types::Type::FuncRef = table.ty().ty {
             for (i, func_idx) in init.elements.iter().enumerate() {
                 let anyfunc = instance.func_ref(*func_idx);
                 table
-                    .set(
+                    .set_with_construction(
                         u32::try_from(start + i).unwrap(),
                         TableElement::FuncRef(anyfunc),
+                        true,
                     )
                     .unwrap();
             }
         } else {
             for i in 0..init.elements.len() {
                 table
-                    .set(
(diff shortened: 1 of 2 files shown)


marxin:
> Basically, rather than going through the table index we can skip it altogether?

No, it's not possible. The table **index** is a variable and thus we must go through the `Table`. Note the Clang ABI uses a single table for all potential indirect calls in a module, that's why we can't make any extra assumption about the function signature (and the run-time check is mandatory).

lib/vm/src/instance/mod.rs
```diff
@@ -1613,6 +1616,7 @@ fn initialize_tables(instance: &mut Instance) -> Result<(), Trap> {
                     .unwrap();
             }
         }
+        table.ty_as_mut().readonly = is_readonly;
```
syrusakbary: I really don't like this, this should not live here. The ty_as_mut is a hack that hints it

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -3293,8 +3293,8 @@ impl<'ctx> LLVMFunctionCodeGenerator<'ctx, '_> {
                     ))
                     .into_pointer_value();
 
-                    // trap if we're trying to call a null funcref
-                    {
+                    if !table.readonly {
+                        // trap if we're trying to call a null funcref
```
syrusakbary, line 3297: I don't think we should do it here.

On the lines before, we can check if is read only, and skip loading the table at all. We already know the funcrefs so we just need to call the function directly

lib/vm/src/instance/mod.rs
marxin: Yeah, found a more feasible approach, good point.

lib/compiler-llvm/src/translator/code.rs
marxin: No, we cannot call the function directly as the called function is variable and depends on the **func_index**. Know we're not in run-time, but at the time, we emit the LLVM IR for a function.

marxin:
@syrusakbary @theduke PING

PR #6373, opened 2026-03-30
feat(runtime): Add cache eviction method to BuiltinPackageLoader

artemyarulin:
Add `BuiltinPackageLoader::evict_cached()` to allow callers to remove a cached in-memory WEBC Container by hash. This is needed by Edge, where pruning a WEBC from disk may not reclaim space because the runtime still retains the mmap-backed container and keeps the deleted file alive in-process.

### DIFF
--- a/lib/wasix/src/runtime/package_loader/builtin_loader.rs
+++ b/lib/wasix/src/runtime/package_loader/builtin_loader.rs
@@ -876,6 +876,18 @@ mod tests {
         cache_misses_will_trigger_a_download_internal().await
     }
 
+    #[tokio::test]
+    async fn evict_cached_removes_in_memory_container() {
+        let loader = BuiltinPackageLoader::new();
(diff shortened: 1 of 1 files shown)


lib/wasix/src/runtime/package_loader/builtin_loader.rs
```diff
@@ -158,6 +158,11 @@ impl BuiltinPackageLoader {
         self.in_memory.save(container, hash);
     }
 
+    /// Remove a container from the in-memory cache.
+    pub fn evict_cached(&self, hash: &WebcHash) -> bool {
```
marxin: What about using the same signature as the underlying `HashMap` function (`Option<Value>`)?

artemyarulin: Agree

artemyarulin:
Test added

PR #6374, opened 2026-03-30
Expose stack pool drain and stack size getters

bragaigor:
# Description

- Expose `MAX_STACK_SIZE` constant (100 MB) and `get_stack_size()` getter for the process-wide default                                  
- Move `STACK_POOL` from function-local static to module-level static so it can be accessed by the new `drain_stack_pool()` function
- Add `drain_stack_pool()` to clear cached coroutine stacks before retrying with a larger size — without this, the pool serves cached undersized stacks indefinitely (see Background)                                                                                         
- Add comments documenting that the pool is not size-aware and callers must drain before resizing   

The `STACK_POOL` caches coroutine stacks for reuse but does not track their allocation size. When the default stack size is increased at runtime (e.g., doubling on native stack overflow), the pool continues to serve old, smaller stacks:

1. Normal calls fill pool with 1MB stacks                                                   
2. Call overflows → set default to 2MB → retry
3. Retry pops 1MB stack from pool → overflows again
4. `scopeguard` returns 1MB stack to pool → infinite loop

`drain_stack_pool()` breaks this cycle by clearing all cached stacks so the next allocation uses the new size.

### DIFF
--- a/lib/vm/src/trap/traphandlers.rs
+++ b/lib/vm/src/trap/traphandlers.rs
@@ -63,7 +63,7 @@ struct ucontext_t {
 use libc::ucontext_t;
 
 /// Sets the process-wide default stack size for new Wasmer coroutines.
-/// The value is clamped to [8 KB, MAX_STACK_SIZE].
+/// The value is clamped to [8 KiB, MAX_STACK_SIZE].
 pub fn set_stack_size(size: usize) {
     DEFAULT_STACK_SIZE.store(
         size.clamp(ByteSize::kib(8).as_u64() as usize, MAX_STACK_SIZE),
@@ -1245,7 +1245,7 @@ mod tests {
     fn set_stack_size_clamps_to_min() {
         let _lock = GLOBAL_STATE.lock().unwrap();
         let _restore = RestoreStackSize(get_stack_size());
-        set_stack_size(1); // way below 8 KB minimum
+        set_stack_size(1); // way below 8 KiB minimum
         assert_eq!(get_stack_size(), ByteSize::kib(8).as_u64() as usize);
     }
 
@@ -1278,9 +1278,9 @@ mod tests {
     /// The stack pool is not size-aware, so after a stack size increase it keeps
     /// serving cached undersized stacks. `drain_stack_pool()` breaks the cycle.
     ///
-    /// 1. A call fills the pool with 500 KB stacks (simulating normal execution).
-    /// 2. The caller doubles the default to 1 MB (simulating overflow retry).
-    /// 3. WITHOUT draining, the pool still hands back a 500 KB stack — the
+    /// 1. A call fills the pool with 500 KiB stacks (simulating normal execution).
+    /// 2. The caller doubles the default to 1 MiB (simulating overflow retry).
+    /// 3. WITHOUT draining, the pool still hands back a 500 KiB stack — the
     ///    retry would overflow again, creating an infinite loop.
     /// 4. After `drain_stack_pool()`, the pool is empty and the next allocation
     ///    must use the new, larger size.
@@ -1290,7 +1290,7 @@ mod tests {
         let _restore = RestoreStackSize(get_stack_size());
         drain_stack_pool();
 
-        // --- Phase 1: simulate normal execution that returns a 500 KB stack ---
+        // --- Phase 1: simulate normal execution that returns a 500 KiB stack ---
         let small_size = ByteSize::kib(500).as_u64() as usize;
         let small_stack = DefaultStack::new(small_size).unwrap();
         STACK_POOL.push(small_stack);


lib/vm/src/trap/traphandlers.rs
```diff
@@ -57,9 +61,35 @@ struct ucontext_t {
...
+///
+/// Note that `STACK_POOL` is a global, concurrently used queue. Other threads
+/// may push stacks back into the pool (for example, when their Wasm execution
+/// finishes) while or after this function is running. As a result, this
+/// function provides only a best-effort drain of the pool: there is no
+/// guarantee that no undersized stacks exist immediately after it returns
+/// unless the caller ensures, via external synchronization, that no other
+/// Wasm executions can return stacks to the pool while this function runs.
+pub fn drain_stack_pool() {
+    while STACK_POOL.pop().is_some() {}
```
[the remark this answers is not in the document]
bragaigor: 1. The loop is already best-effort and documented as such — the doc comment on `drain_stack_pool()` explicitly says other threads may push stacks back during or after the drain, and that callers must use external synchronization if they need a hard guarantee.                                                                                     
2. A bounded drain could be worse, not better — if you drain only N items (snapshot length), you miss any stacks that were pushed between your snapshot and the actual pops. You'd add complexity for the same lack of guarantee.
3. In practice this can't spin — stacks only enter the pool when a Wasm execution finishes (`scopeguard` in `on_wasm_stack`). The number of in-flight Wasm executions is finite and bounded by the thread count. The pool can't grow faster than threads return stacks, and each thread returns at most one at a time.                                                                 
4. The real use case is single-threaded

lib/vm/src/trap/traphandlers.rs
```diff
@@ -1174,3 +1204,110 @@ pub fn lazy_per_thread_init() -> Result<(), Trap> {
...
+
+    #[test]
+    fn get_set_stack_size_roundtrip() {
+        let _lock = GLOBAL_STATE.lock().unwrap();
+        let original = get_stack_size();
+        let new_size = 4 * 1024 * 1024; // 4 MB
+        set_stack_size(new_size);
+        assert_eq!(get_stack_size(), new_size);
+        set_stack_size(original);
+    }
```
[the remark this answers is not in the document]
bragaigor: Currently the mutex kind of takes care of that by panicking threads; however, adding a drop guard would make it cleaner.

lib/vm/src/trap/traphandlers.rs
```diff
@@ -40,6 +40,10 @@ static MAGIC: u8 = 0xc0;
 
 static DEFAULT_STACK_SIZE: AtomicUsize = AtomicUsize::new(1024 * 1024);
```
marxin: Given all the constants in the file, what about us using e.g. `ByteSize::mib(1).0` rather than the manual `* 1024 ...` multiplications?

lib/vm/src/trap/traphandlers.rs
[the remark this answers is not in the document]
marxin: Your comment and the implementation seems fine to me.

lib/vm/src/trap/traphandlers.rs
```diff
@@ -1174,3 +1204,110 @@ pub fn lazy_per_thread_init() -> Result<(), Trap> {
...
+    use std::sync::Mutex;
+
+    // Guards tests that mutate global state (DEFAULT_STACK_SIZE, STACK_POOL).
+    // Rust runs tests in parallel by default; this mutex serializes them so
+    // they don't step on each other.
+    static GLOBAL_STATE: Mutex<()> = Mutex::new(());
+
+    #[test]
+    fn max_stack_size_is_100mb() {
+        assert_eq!(MAX_STACK_SIZE, 100 * 1024 * 1024);
```
marxin: Again, can we use the suggested `ByteSize` crate? Will it make the code more readable?

lib/vm/src/trap/traphandlers.rs
bragaigor: Oh yeah, I like that suggestion, updating to use `ByteSize`

lib/vm/src/trap/traphandlers.rs
```diff
@@ -1174,3 +1208,113 @@ pub fn lazy_per_thread_init() -> Result<(), Trap> {
...
+        let new_size = ByteSize::mib(4).as_u64() as usize;
+        set_stack_size(new_size);
+        assert_eq!(get_stack_size(), new_size);
+    }
+
+    #[test]
+    fn set_stack_size_clamps_to_min() {
+        let _lock = GLOBAL_STATE.lock().unwrap();
+        let _restore = RestoreStackSize(get_stack_size());
+        set_stack_size(1); // way below 8 KB minimum
```
marxin: A nit: `8 KiB`

lib/vm/src/trap/traphandlers.rs
```diff
@@ -1174,3 +1208,113 @@ pub fn lazy_per_thread_init() -> Result<(), Trap> {
...
+        let _lock = GLOBAL_STATE.lock().unwrap();
+        drain_stack_pool();
+        drain_stack_pool(); // second call on empty pool should not panic
+        assert!(STACK_POOL.is_empty());
+    }
+
+    /// The stack pool is not size-aware, so after a stack size increase it keeps
+    /// serving cached undersized stacks. `drain_stack_pool()` breaks the cycle.
+    ///
+    /// 1. A call fills the pool with 500 KB stacks (simulating normal execution).
```
marxin: `KiB` and `MiB` here.

PR #6375, opened 2026-03-30
build: move binary blobs to a separate git repository

marxin:


marxin:
> Except one thing, I don't think the `.wasmu` files should be moved to a repo outside.

Actually, these are one of the biggest drivers - they occupy ~100MB and each artifact bump needs to adjust them. That said, they're responsible for a big size bloat.

It's not actually a pain - one can basically commit into `main` of `wasmer-test-files`, and then use the `hash` as a sub-module for a PR. From that point, no need to the submodule will be needed.

syrusakbary:
> these are one of the biggest drivers - they occupy ~100MB and each artifact bump needs to adjust them. That said, they're responsible for a big size bloat.

Maybe using git LFS might be a better fit for this? I'm not opposed to the changes, but I think it may complicate some things down the line, so we need to be very certain this is the best way to go

marxin:
> Maybe using git LFS might be a better fit for this?

Given my previous poor LFS user experience, I would rather go with the standard submodule.

> I'm not opposed to the changes, but I think it may complicate some things down the line, so we need to be very certain this is the best way to go.

Right now, we are in a situation where the current git repo is terribly large (`.git` is **2.2GiB**) and a history rewrite (where we'll ignore large files) will be necessary. We discussed that briefly with @theduke. If the submodule scheme will not play well, we can always adjust while not paying the repository as (will be moved to a different repository).

PR #6376, opened 2026-03-30
Fix Global borrow conflicts, middleware lifetimes, and error routing

bragaigor:
# Description

- Fix borrow conflict in `Global::set` and `Global::get` (`lib/api/src/backend/sys/entities/global.rs`): The previous code held a `&mut` borrow of the store (via `get_mut`) while simultaneously trying to borrow it again for `val.as_raw(store)`. Restructured to perform type/mutability checks and raw value computation before taking the mutable reference. Also deduplicated redundant `handle.get()` calls in `get()`.
- Add lifetime parameter to `FunctionMiddleware` trait and expose locals info to middlewares (`lib/compiler/src/translator/middleware.rs`, `lib/middlewares/src/metering.rs`, `tests/compilers/middlewares.rs`): The `FunctionMiddleware` trait now carries a lifetime `'a`, allowing feed to borrow operators with the reader's lifetime instead of an anonymous one. Added `locals_info` callback so middlewares can inspect function locals before processing operators. `MiddlewareReaderState` now tracks local declarations and emits locals info once all declarations are read.
- Add From<MiddlewareError> for CompileError for ergonomic error handling

 ## Motivation                                                                                                                                                                                  
                                                                                                                                                                                              
OffchainLabs maintains a fork of wasmer with changes needed for Arbitrum Stylus (on-chain WASM execution). These changes have been carried across multiple wasmer version upgrades (v4.2.3 → v4.3.7 → v7.1.0). Proposing them upstream would reduce the maintenance burden and benefit other wasmer users who need richer middleware capabilities. 

The current middleware API prevents implementing middlewares that need to buffer operators or know about function locals. Specifically:

1. Operator buffering: With the lifetime on `feed` (`fn feed<'a>(..., op: Operator<'a>)`), a middleware cannot store `Operator<'a>` in its own fields — `'a` is a per-call parameter unrelated to `Self`. Moving it to the trait (`FunctionMiddleware<'a>`) lets implementations accumulate operators. This is needed for whole-function analysis passes (e.g., computing worst-case stack depth before emitting instrumented output).                                                                                                                             
                                                                                                                                                                                              
2. Locals info: Middleware that computes stack frame sizes needs the number of locals, but locals are parsed before operators and never passed to the middleware. The `locals_info` callback fills this gap. It has a default no-op implementation, so existing code is unaffected.

### DIFF
--- a/lib/compiler/src/translator/middleware.rs
+++ b/lib/compiler/src/translator/middleware.rs
@@ -73,8 +73,8 @@ pub struct MiddlewareReaderState<'a> {
     /// Number of local declaration groups (each group is a count + type pair).
     local_decls_group: u32,
 
-    /// Number of local declarations read so far.
-    local_decls_read: u32,
+    /// Number of local declaration groups read so far.
+    local_decls_group_read: u32,
 
     /// Locals read so far.
     locals: Vec<ValType>,
@@ -143,7 +143,7 @@ impl<'a> MiddlewareBinaryReader<'a> {
                 }),
                 pending_operations: VecDeque::new(),
                 local_decls_group: 0,
-                local_decls_read: 0,
+                local_decls_group_read: 0,
                 locals: vec![],
             },
             chain: vec![],
@@ -202,8 +202,8 @@ impl<'a> FunctionBinaryReader<'a> for MiddlewareBinaryReader<'a> {
             self.state.locals.push(ty);
         }
 
-        self.state.local_decls_read += 1;
-        if self.state.local_decls_read == self.state.local_decls_group {
+        self.state.local_decls_group_read += 1;
+        if self.state.local_decls_group_read == self.state.local_decls_group {
             self.emit_locals_info();
         }
         Ok((count, ty))


lib/compiler/src/translator/middleware.rs
```diff
@@ -66,26 +69,35 @@ pub struct MiddlewareReaderState<'a> {
...
     /// Generates a function middleware chain.
-    fn generate_function_middleware_chain(
+    fn generate_function_middleware_chain<'a>(
         &self,
         local_function_index: LocalFunctionIndex,
-    ) -> Vec<Box<dyn FunctionMiddleware>> {
+    ) -> Vec<Box<dyn FunctionMiddleware<'a> + 'a>> {
         self.iter()
             .map(|x| x.generate_function_middleware(local_function_index))
             .collect()
```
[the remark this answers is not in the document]
bragaigor: Humm `generate_function_middleware<'a>` only introduces a lifetime, not a type parameter. `dyn ModuleMiddleware` remains valid.

lib/compiler/src/translator/middleware.rs
```diff
@@ -18,10 +18,10 @@ pub trait ModuleMiddleware: Debug + Send + Sync {
     /// Here we generate a separate object for each function instead of executing directly on per-function operators,
     /// in order to enable concurrent middleware application. Takes immutable `&self` because this function can be called
     /// concurrently from multiple compilation threads.
-    fn generate_function_middleware(
+    fn generate_function_middleware<'a>(
         &self,
         local_function_index: LocalFunctionIndex,
-    ) -> Box<dyn FunctionMiddleware>;
+    ) -> Box<dyn FunctionMiddleware<'a> + 'a>;
```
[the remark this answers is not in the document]
bragaigor: similar reason to https://github.com/wasmerio/wasmer/pull/6376#discussion_r3012500372. `'a` is not "unconstrained" — it's a late-bound lifetime parameter that the caller chooses when invoking the method. This is standard  Rust for "works for any lifetime."

lib/compiler/src/translator/middleware.rs
```diff
@@ -169,7 +196,16 @@ impl<'a> FunctionBinaryReader<'a> for MiddlewareBinaryReader<'a> {
                 message: "locals must be read before the function body".to_string(),
                 offset: self.current_position(),
             }),
+        }?;
+        for _ in 0..count {
+            self.state.locals.push(ty);
+        }
```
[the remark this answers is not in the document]
bragaigor: I don't think it's worth the complexity for the following reasons; so this isn't a practical concern because of where middleware sits in the pipeline:                                                        
                                                                                                                                          
1. Validation runs first. `Module::from_binary()` calls `Self::validate(engine, binary)?` before compilation ever starts. This calls `wasmparser`'s `Validator::validate_all()`, which enforces the `WebAssembly` spec limits on the module - including local counts - and rejects malformed modules before they reach compilation.
2. `wasmparser` caps locals at `50,000`. The hard limit is `MAX_WASM_FUNCTION_LOCALS: u32 = 50_000` in `wasmparser/src/limits.rs:32`. Since `ValType` is 1 byte, the worst-case Vec is ~50 KB per function; negligible. This limit is agreed upon across Wasm engines for consistency (see the JS API limits spec).
3. Middleware runs after validation. The `MiddlewareBinaryReader` that executes `read_local_decl` only processes function bodies during compilation, which happens after validation succeeds. By the time this code runs, `wasmparser` has already verified the module is well-formed.                                                                                                                            
                                                                                                                                          
In short, untrusted Wasm is validated by wasmparser before any middleware code touches it. Adding redundant bounds-checking here would be defending against inputs that have already been rejected.

marxin:
@bragaigor mark it ready when you feel it's worth reviewing

bragaigor:
Thanks @marxin ! I kept being pulled away and wanted to make sure the changes made sense to bring it. Now I think it's ready

lib/compiler/src/translator/middleware.rs
```diff
@@ -66,26 +69,35 @@ pub struct MiddlewareReaderState<'a> {
 
     /// The pending operations added by the middleware.
     pending_operations: VecDeque<Operator<'a>>,
+
+    /// Number of local declarations that will ever be read.
+    local_decls: u32,
```
marxin: Can't we just drop this and rely on `locals.len()` instead?

tests/compilers/middlewares.rs
```diff
@@ -205,3 +211,120 @@ fn middleware_chain_order_2(mut config: crate::Config) -> Result<()> {
...
+    let wat = r#"(module
+        (func (export "test") (param i32 i32) (result i32)
+           (local i64)
+           (local f32 f32)
+           (i32.add (local.get 0) (local.get 1)))
+    )"#;
+    let _module = Module::new(&store, wat)?;
+
+    let locals = captured.lock().unwrap();
+    assert_eq!(
```
marxin, line 261: What about directly doing `assert_eq!(&locals, &[&[ValType::I64, ValType::F32, ValType::F32]])` ?

tests/compilers/middlewares.rs
```diff
@@ -205,3 +211,120 @@ fn middleware_chain_order_2(mut config: crate::Config) -> Result<()> {
...
+    let store = config.store();
+    // Function with no locals at all — locals_info should still be called with an empty slice.
+    let wat = r#"(module
+        (func (export "test") (param i32 i32) (result i32)
+           (i32.add (local.get 0) (local.get 1)))
+    )"#;
+    let _module = Module::new(&store, wat)?;
+
+    let locals = captured.lock().unwrap();
+    assert_eq!(
```
marxin: Similarly here.

tests/compilers/middlewares.rs
```diff
@@ -205,3 +211,120 @@ fn middleware_chain_order_2(mut config: crate::Config) -> Result<()> {
...
+        locals.len(),
+        2,
+        "expected locals_info called once per function"
+    );
+    // Order may vary due to concurrent compilation, so check both are present.
+    let mut sorted: Vec<Vec<ValType>> = locals.clone();
+    sorted.sort_by_key(|v| v.len());
+    assert_eq!(sorted[0], vec![ValType::F64]);
```
marxin: and also here?

lib/compiler/src/translator/middleware.rs
bragaigor: `local_decls` counts the number of local declaration groups, while `locals.len()` counts individual locals. In WASM binary format, locals are declared in groups like (3, i32), (2, f64) — that's 2 declaration groups but 5 individual locals. `read_local_count()` reads the number of declaration groups from the binary and stores it in `local_decls`. We could probably use a better comment to highlight `local_decls` refers to the count of local declaration groups? And probably rename to something like `local_decls_group`?

marxin: Oh, gotcha. Yes, the name confused me, so I would go with `local_decls_group`.

tests/compilers/middlewares.rs
bragaigor: ah yes that's cleaner, applied in https://github.com/wasmerio/wasmer/pull/6376/commits/43564c09677b131bd5f6284fc490b08f481f345e

lib/compiler/src/translator/middleware.rs
bragaigor: cool done in https://github.com/wasmerio/wasmer/pull/6376/commits/43564c09677b131bd5f6284fc490b08f481f345e

tests/compilers/middlewares.rs
bragaigor: simplified in https://github.com/wasmerio/wasmer/pull/6376/commits/43564c09677b131bd5f6284fc490b08f481f345e

tests/compilers/middlewares.rs
bragaigor: simplified in https://github.com/wasmerio/wasmer/pull/6376/commits/43564c09677b131bd5f6284fc490b08f481f345e

lib/compiler/src/translator/middleware.rs
```diff
@@ -66,26 +69,35 @@ pub struct MiddlewareReaderState<'a> {
 
     /// The pending operations added by the middleware.
     pending_operations: VecDeque<Operator<'a>>,
+
+    /// Number of local declaration groups (each group is a count + type pair).
+    local_decls_group: u32,
+
+    /// Number of local declarations read so far.
+    local_decls_read: u32,
```
marxin: Should be probably named `local_decls_group_read`, right? Comment needs adjustment as well.

bragaigor: yes, I should have done in previous commit. Done in https://github.com/wasmerio/wasmer/pull/6376/commits/1aae135b7c04b214b32126d174f938a6b9876c44

PR #6377, opened 2026-03-31
size-check pooled stacks on pop to prevent undersized reuse

bragaigor:
# Description                                                                                                              

- Add a size check when popping coroutine stacks from the pool in `on_wasm_stack`, discarding undersized stacks instead of blindly reusing them
                                                                                 
When a native stack overflow is detected, a caller may retry with a larger stack size (e.g. doubling) and call `drain_stack_pool()` to clear cached undersized stacks. However, `drain_stack_pool()` is a point-in-time operation; threads that are still mid-execution at the old size will push their (now undersized) stacks back into the pool after the drain completes.
                                                
Example timeline:
```
Thread A: starts wasm execution with 50K stack
Thread B: hits stack overflow, doubles to 1MB, drains pool
Thread A: finishes, pushes 50K stack back into pool 
Thread B: pops from pool → gets 50K stack → overflows again 
```
                                                                                                                              
This causes an unnecessary extra retry and another doubling of the stack size.                                              

The proposed solution checks the size of each popped stack against the requested `stack_size` before using it. If the stack is too small, discard it (`munmap`) and allocate a fresh one. This is done with a single pop attempt — if the stack doesn't fit, we go straight to a fresh allocation rather than continuing to pop, which keeps the logic simple and avoids any risk of spinning through a large pool.                                        
```rust
  let stack = STACK_POOL                                                                                                      
      .pop()                                                                                                                  
      .filter(|s| s.base().get() - s.limit().get() >= stack_size)                                                             
      .unwrap_or_else(|| DefaultStack::new(stack_size).unwrap());                                                             
```
`drain_stack_pool()` remains valuable as a complementary mechanism: it bulk-clears the pool so that the number of undersized stacks any future caller can encounter is bounded by the number of concurrently executing threads (rather than the total pool size, which can be much larger). The size check on pop handles the stragglers that slip through after the drain.


lib/vm/src/trap/traphandlers.rs
```diff
@@ -1317,4 +1317,33 @@ mod tests {
             "after drain, pool must be empty so a fresh stack is allocated at the new size"
         );
     }
+
+    /// `on_wasm_stack` discards undersized stacks from the pool and allocates
+    /// a fresh one instead of blindly reusing whatever the pool returns.
+    #[test]
+    fn on_wasm_stack_discards_undersized_stack() {
+        let _lock = GLOBAL_STATE.lock().unwrap();
+        let _restore = RestoreStackSize(get_stack_size());
```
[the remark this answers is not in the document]
bragaigor: That is correct but I'd rather keep it for consistency given all other tests need it and in case this test gets updated in the future

lib/vm/src/trap/traphandlers.rs
```diff
@@ -1005,15 +1005,15 @@ fn on_wasm_stack<F: FnOnce() -> T + 'static, T: 'static>(
...
-    // and before relying on the new value returned by `get_stack_size()`.
+    // Reuse a cached stack from the pool if it is large enough, otherwise
+    // allocate a fresh one. The size check prevents using undersized stacks
+    // that were returned by threads still running at the old size after
+    // `drain_stack_pool()` was called. `base() - limit()` is the full mmap
+    // region (including guard page), which is always >= the requested size
+    // for stacks allocated with that size.
     let stack = STACK_POOL
         .pop()
+        .filter(|s| s.base().get() - s.limit().get() >= stack_size)
```
marxin: Can we add `size()` function into the `Stack` trait and use it here?

bragaigor: Good idea, I used a blanket extension trait since Stack is coming from external crate https://github.com/wasmerio/wasmer/pull/6377/changes/406dd726e3b653423e0ca8f920d7cf54a7c45a16

marxin: Fine, thanks.

marxin:
Can you merge `main` where we just disabled one flaky test-case?

bragaigor:
`main` merged

PR #6380, opened 2026-03-31
feat(runtime): Allow disabling automatic in-memory cache in BuiltinPackageLoader

artemyarulin:
To follow up on #6373 it's actually better to make `InMemoryCache` totally optional as for example in Edge we have our own caching layer and it doesn't make sense to have two layers at the same time. 

This PR makes it optional, but default behavior stays the same, so a safe change.

### DIFF
--- a/lib/wasix/src/runtime/package_loader/builtin_loader.rs
+++ b/lib/wasix/src/runtime/package_loader/builtin_loader.rs
@@ -31,7 +31,7 @@ use crate::{
 #[derive(Debug)]
 pub struct BuiltinPackageLoader {
     client: Arc<dyn HttpClient + Send + Sync>,
-    in_memory: InMemoryCache,
+    in_memory: Option<InMemoryCache>,
     cache: Option<FileSystemCache>,
     /// A mapping from hostnames to tokens
     tokens: HashMap<String, String>,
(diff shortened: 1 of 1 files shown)


lib/wasix/src/runtime/package_loader/builtin_loader.rs
```diff
@@ -447,8 +466,9 @@ impl PackageLoader for BuiltinPackageLoader {
         // The sad path - looks like we don't have a filesystem cache so we'll
         // need to keep the whole thing in memory.
         let container = crate::spawn_blocking(move || from_bytes(bytes)).await??;
-        // We still want to cache it in memory, of course
```
marxin, line 450: Can we preserve the comment line?

lib/wasix/src/runtime/package_loader/builtin_loader.rs
```diff
@@ -79,6 +79,14 @@ impl BuiltinPackageLoader {
         }
     }
 
+    /// Disable automatic promotion of loaded containers into the in-memory cache.
+    pub fn without_automatic_in_memory_cache(self) -> Self {
```
theduke: I'd drop the `automatic_`, that's kind of redundant?

marxin: Yes, I like the suggestion.

lib/wasix/src/runtime/package_loader/builtin_loader.rs
artemyarulin: resolved

marxin:
@artemyarulin take a look at CI

PR #6388, opened 2026-04-02
feat(Singlepass, Cranelift): add `--enable-experimental-unaligned-memory-accesses`

marxin:
Right now, the RISC-V single-pass compiler emits 1-byte loads and stores (as a runtime check) to avoid strict alignment violations. This PR changes the default to use full-size native load/store operations, while still providing an option to retain the previous behavior. Similarly, for Cranelift, we now support this functionality by incorporating @wakabat’s changes.

@tsahee @wakabat - can you test the change if it fits your needs?

Relates: #6024


lib/compiler/src/compiler.rs
```diff
@@ -54,6 +54,10 @@ pub trait CompilerConfig {
     /// (but are not 100% SPEC compliant).
     fn enable_non_volatile_memops(&mut self) {}
 
+    /// Enable run-time handling of potentially unaligned memory accesses.
+    /// Available for Cranelift and Singlepass (RISC-V).
+    fn enable_unaligned_memory_accesses(&mut self) {}
```
theduke, line 59: I know this is the existing pattern, but should these methods have default impls?
Would be saner to force implementing them in each concrete implementation.

theduke (commented):
Some AI findings: 

- High: Singlepass on RISC-V still leaves f32/f64 memory ops on the raw path, so the new flag does not actually make unaligned float accesses work. lib/compiler-singlepass/src/
    machine_riscv.rs:5884 and lib/compiler-singlepass/src/machine_riscv.rs:5939 still call emit_relaxed_load, and lib/compiler-singlepass/src/machine_riscv.rs:5908 / lib/compiler-singlepass/src/
    machine_riscv.rs:5963 still call emit_relaxed_store / canonicalize_nan directly instead of emit_maybe_unaligned_*. Valid Wasm like f32.load at address 1 will still fault/trap on the target this
    feature is meant for. The new WAST won’t catch this unless CI actually exercises Singlepass on RISC-V.
 
==> Is this intentionally only for ints?

- Medium: Cranelift’s new slow path explicitly excludes 16-byte accesses, so v128.load / v128.store are still not handled even when allow_unaligned_memory_accesses is enabled. lib/compiler-
    cranelift/src/translator/code_translator.rs:962 and lib/compiler-cranelift/src/translator/code_translator.rs:1072 route SIMD loads/stores through translate_load / translate_store, but the
    fallback only runs for mem_op_size > 1 && mem_op_size < 16 at lib/compiler-cranelift/src/translator/code_translator.rs:3032 and lib/compiler-cranelift/src/translator/code_translator.rs:3119. So
    the flag currently means “some unaligned accesses”, not all Wasm memory accesses.

==> As above, is this fine?
  
- Medium: The new generic compiler API is only wired for Cranelift, not Singlepass. The trait adds enable_unaligned_memory_accesses as a default no-op in lib/compiler/src/compiler.rs:57; Cranelift
    overrides it at lib/compiler-cranelift/src/config.rs:298; Singlepass does not override it in lib/compiler-singlepass/src/config.rs:130. Any caller holding Box<dyn CompilerConfig> and enabling
    the feature generically will get the expected behavior for Cranelift and a silent no-op for Singlepass.

I reckon this is an actual oversight?

lib/cli/src/backend.rs
```diff
@@ -218,6 +218,12 @@ pub struct RuntimeOptions {
     #[clap(long = "enable-nan-canonicalization")]
     enable_nan_canonicalization: bool,
 
+    /// Allow unaligned memory accesses.
```
theduke, line 221: If we keep both, should we test both variants on CI?

theduke: We do have the new test for unaligned memory access, which gets the feature enabled. 

Seems like there are several instructions that are not covered:

  - i32.load8_{u,s} and i64.load8_{u,s}
  - i32.store8 and i64.store8
  - v128.load / v128.store
  - SIMD: v128.load8x8_*, load16x4_*, load32x2_*, load*_splat, load*_zero, load*_lane, store*_lane

tamirhemo:
Hi @marxin apologies for the late reply (I am taking over for @wakabat on this issue). So if I understand the code correctly, the new setting will emit the word load/store in any case and so potentially unaligned ones will be emitted?

marxin:
> So if I understand the code correctly, the new setting will emit the word load/store in any case and so potentially unaligned ones will be emitted?

The new settings will allow run-time detection of unaligned memory accesses. When such cases occur, the generated assembly will handle them by performing byte-wise loads/stores and reconstructing the intended data type. Hopefully this works well for your RISC-V simulator.

tamirhemo:
> > So if I understand the code correctly, the new setting will emit the word load/store in any case and so potentially unaligned ones will be emitted?
> 
> The new settings will allow run-time detection of unaligned memory accesses. When such cases occur, the generated assembly will handle them by performing byte-wise loads/stores and reconstructing the intended data type. Hopefully this works well for your RISC-V simulator.

Yes, that’s exactly what we need!

syrusakbary (approved):

lib/compiler/src/compiler.rs
marxin: I am pretty aligned with the default implementation here ;)

lib/cli/src/backend.rs
marxin: Good point - extended the set of tested types (not the SIMD types are unsupported as of now).

marxin:
Decided to name both the option and the configuration parameter with the `experimental` prefix as the feature is supported only by some compilers and platforms.

PR #6391, opened 2026-04-06
Go from UnionFS to MountFS, which properly supports nested mounts and eliminate mounting from all other filesystems

Arshia001:
Draft PR, for opinion-gathering, though the code should be in relatively good shape. See early design ideas in `docs/dev/union-fs-vfs-design.md`.

### DIFF
--- a/tests/lib/wast/src/wasi_wast.rs
+++ b/tests/lib/wast/src/wasi_wast.rs
@@ -221,21 +221,23 @@ impl<'a> WasiTest<'a> {
                 let fs = host_fs::FileSystem::new(Handle::current(), base_dir.clone()).unwrap();
 
                 for (alias, real_dir) in &self.mapped_dirs {
-                    let mut dir = base_dir.clone();
-                    dir.push(real_dir);
-                    builder.add_map_dir(alias, dir)?;
+                    let mut guest_dir = PathBuf::from("/");
+                    guest_dir.push(real_dir);
+                    builder.add_map_dir(alias, guest_dir)?;
                 }
 
                 // due to the structure of our code, all preopen dirs must be mapped now
                 for dir in &self.dirs {
-                    let mut new_dir = base_dir.clone();
+                    let mut new_dir = PathBuf::from("/");
                     new_dir.push(dir);
                     builder.add_map_dir(dir, new_dir)?;
                 }
 
                 for alias in &self.temp_dirs {
                     let temp_dir = tempfile::tempdir_in(&base_dir)?;
-                    builder.add_map_dir(alias, temp_dir.path())?;
+                    let guest_temp_dir =
+                        Path::new("/").join(temp_dir.path().strip_prefix(&base_dir).unwrap());
+                    builder.add_map_dir(alias, guest_temp_dir)?;
                     host_temp_dirs_to_not_drop.push(temp_dir);
                 }
 
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -2572,7 +2572,7 @@ mod tests {
         let init = WasiEnvBuilder::new("test_prog")
             .engine(Engine::default())
             .fs(Arc::new(host_fs) as Arc<dyn FileSystem + Send + Sync>)
-            .map_dir("hamlet", &hamlet_dir)
+            .map_dir("hamlet", "/hamlet")
             .unwrap()
             .build_init()
             .unwrap();
@@ -2595,14 +2595,10 @@ mod tests {
 
     #[tokio::test]
     async fn dot_mapped_preopen_uses_guest_current_dir() {
-        let root_dir = tempdir().unwrap();
-        let hamlet_dir = root_dir.path().join("hamlet");
-        std::fs::create_dir_all(&hamlet_dir).unwrap();
-
         let init = WasiEnvBuilder::new("test_prog")
             .engine(Engine::default())
             .current_dir("/work")
-            .map_dir(".", &hamlet_dir)
+            .map_dir(".", "/work")
             .unwrap()
             .build_init()
             .unwrap();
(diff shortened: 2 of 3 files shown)


marxin (commented):
Can you make CI green?

Arshia001:
will do after all changes are in @marxin

lib/wasix/src/state/env.rs
```diff
@@ -1107,75 +1127,23 @@ impl WasiEnv {
...
-                        }
-                        let mut f = fs.new_open_options().create(true).write(true).open(path2)?;
-                        if let Err(e) = f.copy_from_owned_buffer(&atom).await {
-                            tracing::warn!(
-                                error = &e as &dyn std::error::Error,
-                                "Unable to copy file reference",
-                            );
-                        }
-                    }
+                if let Err(err) = write_buffer_to_fs(root_fs, path, &atom).await {
```
theduke: This loses the COW semantics for package binaries, which would introduce very notable overhead on Edge.

This may be a bit more tricky now with the MountFS being the new root, but we have to find a layering that still allows preserving this.

lib/wasix/src/state/env.rs
```diff
@@ -1248,31 +1216,17 @@ impl WasiEnv {
             })?;
             let file = OwnedBuffer::from(file);
 
-            if let WasiFsRoot::Sandbox(root_fs) = &self.state.fs.root_fs {
-                let _ = root_fs.create_dir(Path::new("/bin"));
-                let _ = root_fs.create_dir(Path::new("/usr"));
-                let _ = root_fs.create_dir(Path::new("/usr/bin"));
+            let path = format!("/bin/{command}");
+            let path = Path::new(path.as_str());
+            if let Err(err) = block_on(write_buffer_to_fs(&self.state.fs.root_fs, path, &file)) {
```
theduke: As above.

lib/wasix/src/state/env.rs
```diff
@@ -1248,31 +1216,17 @@ impl WasiEnv {
...
-                    tracing::debug!("failed to add atom command [{}] - {}", command, err);
-                    continue;
-                }
-            } else {
-                tracing::debug!(
-                    "failed to add atom command [{}] to the root file system as it is not sandboxed",
-                    command
-                );
+            let path = format!("/usr/bin/{command}");
+            let path = Path::new(path.as_str());
```
theduke, line 1237: As above.

lib/virtual-fs/src/builder.rs
```diff
@@ -56,18 +62,44 @@ impl RootFileSystemBuilder {
         self
     }
 
+    pub fn with_memory_limiter(mut self, limiter: DynFsMemoryLimiter) -> Self {
```
theduke, line 65: I believe we do not preserve the memory limiter for the additional nested tmp file systems that are created for nested mounts.

We will need to somehow store and forward the limiter.

lib/virtual-fs/src/mount_fs.rs
```diff
@@ -0,0 +1,2133 @@
...
+
+#[derive(Debug, Clone)]
+struct MountedFileSystem {
+    fs: DynFileSystem,
+    source_path: PathBuf,
+}
+
+#[derive(Debug, Default)]
+struct MountNode {
+    mount: DashMap<(), MountedFileSystem>,
```
theduke: Using a dashmap here can be racy ,since we don't guard against races.

Example flow below.

Also, updates to the mount topology are rare, while lookups are very common, and the dashmap increases baseline overhead for all lookups.
I think I'd stick to eiter an `RwLock<FnvHashMap<..>>` here, or do something with `arc_swap`, which would be cheaper thana lock (though this may be tricky to do safely because file systems are not all `Clone`).

Racy:


  - A reader can observe a partially constructed subtree while mount() is creating intermediate nodes.
  - A reader can observe a subtree temporarily missing during ReplaceExisting because descendants are cleared before the new mount is visible.
  - read_dir() / exact_node() can mix child names from one moment with mount presence from another moment; see lib/virtual-fs/src/mount_fs.rs:230 and lib/virtual-fs/src/mount_fs.rs:319.

tests/lib/wast/src/wasi_wast.rs
```diff
@@ -261,40 +261,24 @@ impl<'a> WasiTest<'a> {
...
                         let f = mem_fs::FileSystem::default();
 
-                        let union = union_fs::UnionFileSystem::new();
+                        let mount_fs = mount_fs::MountFileSystem::new();
 
-                        union.mount(
-                            "mem_fs".to_string(),
-                            PathBuf::from("/test_fs").as_ref(),
-                            Box::new(a),
+                        mount_fs.mount_with_source("mem_fs", "/test_fs", Arc::new(a))?;
```
theduke: AI finding, not sure if sensible:

```
- Medium: the UnionHostMemory WAST harness migration swapped mount_with_source() arguments. The new API is (guest_path, source_path, fs), but the code is passing the old union “name” as the guest path and the old guest
  path as the source path, so mounts land at mem_fs, mem_fs_2, etc. instead of /test_fs, /snapshot1, /tests, and so on. That means this test mode no longer models the old topology and can silently stop testing the intended
  mounted-path behavior. See tests/lib/wast/src/wasi_wast.rs:266.
```

theduke (changes requested):

lib/virtual-fs/src/builder.rs
Arshia001: oh right, the overlay primaries are not covered by the limiter... hmmmm.

lib/virtual-fs/src/mount_fs.rs
Arshia001: TBH, I think the current design is _too concurrent_. We don't change topology during execution, only during instance setup, so having an RW lock on the whole structure sounds like the better option to me; instance setup can just take write locks for free since it's the only owner of the FS, and during execution read operations always get a read lock as they need it. WDYT, @theduke?

marxin: Probably not critical in this context, but even a read-only requests for a RW lock can be really slow:
https://www.youtube.com/watch?v=tND-wBBZ8RY
https://jon.thesquareplanet.com/slides/are-mutexes-slow/#6

marxin:
@Arshia001 Could you clarify whether this PR affects #6161, or even better, whether it fully supersedes it?

Arshia001:
@marxin entirely separate, that one deal with overlay, this one with union (now mount) fs.

PR #6416, opened 2026-04-08
fix(LLVM): add --disable-non-volatile-memops and use it consistently

### LINKED ISSUE #6401: LLVM backend does not trap on out-of-bounds SIMD load when result is dropped
### Describe the bug

When a `v128.load16x4_s` instruction accesses an out-of-bounds address and its result is immediately dropped, the LLVM backend returns normally instead of trapping. The Cranelift backend and other runtimes (wasmtime, wasmedge) trap as expected.

```sh
$> wasmer -vV; rustc -vV
wasmer 7.1.0
binary: wasmer-cli
commit-hash: 7a01a2680beba20eb2e719732973ed49a3763636
commit-date: 2026-03-27
CPU flags: sse2 sse3 ssse3 sse4.1 sse4.2 popcnt avx bmi bmi2 avx2 avx512dq avx512vl avx512f lzcnt fma
runtimes: Singlepass, Cranelift, LLVM, WAMR, Wasmi
features: wasix, napi_v10, napi_extension_wasmer_v0
rustc 1.95.0-nightly (873d4682c 2026-01-25)
binary: rustc
commit-hash: 873d4682c7d285540b8f28bfe637006cef8918a6
commit-date: 2026-01-25
release: 1.95.0-nightly
LLVM version: 21.1.8
```

#### buggy (save as tmp_drop.wat):
```
(module
  (memory 1)
  (func (export "main")
    i32.const 0xFFFFFFFE
    v128.load16x4_s align=1
    drop
  )
)
```
Then run: `wasm-tools parse -o tmp_drop.wasm tmp_drop.wat`

#### non-buggy (save as tmp_return.wat):
```
(module
  (memory 1)
  (func (export "main") (result v128)
    i32.const 0xFFFFFFFE
    v128.load16x4_s align=1
    ;; drop
  )
)
```
Then run: `wasm-tools parse -o tmp_return.wasm tmp_return.wat`

### Expected behavior

The buggy version with the `drop` at the end causes the execution to terminate normally in llvm mode. The non-buggy version with a `return` instead of a drop correctly traps with an out of bounds exception in llvm mode.

### Actual behavior
llvm mode should also trap with an out of bounds exception for the version with the `drop` instruction.

### Additional context

I am sharing the pre/post optimization IR dumps for both the `drop` and `return` versions below:

#### `drop` version:
```
$> diff function_0.preopt.ll function_0.postopt.ll
355,366c355
< entry:
<   br label %start_of_code
< 
< start_of_code:                                    ; preds = %entry
<   %mem_value_ptr = getelementptr i8, ptr %m0_base_ptr, i64 4294967294
<   %0 = load i64, ptr %mem_value_ptr, align 8
<   %1 = bitcast i64 %0 to <4 x i16>
<   %2 = sext <4 x i16> %1 to <4 x i32>
<   %3 = bitcast <4 x i32> %2 to i128
<   br label %return
< 
< return:                                           ; preds = %start_of_code
---
> return:
```
Due to unused value, the entire `start_of_code` block is being optimized away. As a result, no out of bounds exception is being produced.

#### `return` version:
```
$> cat function_0.postopt.ll
... ... ...
return:
  %mem_value_ptr = getelementptr i8, ptr %m0_base_ptr, i64 4294967294
  %0 = load i64, ptr %mem_value_ptr, align 8
  %1 = bitcast i64 %0 to <4 x i16>
  %2 = sext <4 x i16> %1 to <4 x i32>
  %3 = bitcast <4 x i32> %2 to i128
  ret i128 %3
}
... ... ...
``` 
On the other hand, in the `return` version, the loading and returning code is preserved since %3 value is used.

Similar to wasmer's cranelift mode and other runtimes, shouldn't wasmer's llvm mode also trap with out of bounds exception regardless of the liveness of the values?  I think doing `load volatile` or including an explicit bounds checking step might resolve the issue.

marxin:
This PR introduces the `--disable-non-volatile-memops` CLI option, which preserves the use of the `volatile` keyword for memory operations (as it was prior to #6212). Additionally, it refactors how load and store operations are emitted, significantly simplifying and unifying the `Operator::X` expansion.

Fixes: #6401


lib/compiler-llvm/src/translator/code.rs
```diff
@@ -1265,7 +1243,22 @@ impl<'ctx> LLVMFunctionCodeGenerator<'ctx, '_> {
...
+                memory_index,
+                self.intrinsics,
+                self.module,
+                self.memory_styles,
+            )? {
+                // The best we've got is `volatile`.
+                memaccess.set_volatile(true).map_err(|err| {
+                    CompileError::Codegen(format!(
+                        "could not set volatile on memory operation: {err}"
+                    ))
```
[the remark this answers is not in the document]
marxin: No, the `annotate_user_memaccess` was called also for atomic operations.

tests/compilers/issues.rs
```diff
@@ -1149,3 +1149,49 @@ fn issue_6334_foldable_comparison_expressions(mut config: crate::Config) -> Resu
 
     Ok(())
 }
+
+#[cfg(feature = "llvm")]
```
theduke, line 1156: nit: can we add a short description of what this test is for?

I like to have those in general, avoiding the need to look up the issue to get a baseline idea.

lib/compiler-llvm/src/translator/code.rs
```diff
@@ -11032,2114 +10636,466 @@ impl<'ctx> LLVMFunctionCodeGenerator<'ctx, '_> {
...
-            Operator::I64AtomicRmw32AddU { ref memarg } => {
+            Operator::I64AtomicStore32 { ref memarg } => {
                 let value = self.state.pop1()?.into_int_value();
                 let offset = self.state.pop1()?.into_int_value();
-                let memory_index = MemoryIndex::from_u32(0);
-                let effective_address = self.resolve_memory_ptr(
-                    memory_index,
-                    memarg,
-                    self.intrinsics.ptr_ty,
+                self.build_annotated_atomic_store(
```
theduke, line 10759: Nice to get rid of all this duplication...

theduke (approved):

tests/compilers/issues.rs
marxin: Sure, done.

PR #6457, opened 2026-04-12
test(wasix): migrate process smoke POSIX tests to wasm_tests

artemyarulin:
This migrates the next tests-only POSIX batch from `tests/wasix-tests` into `lib/wasix/tests/wasm_tests`.

This PR adds:
- `proc_id`
- `proc_parent`
- `proc_snapshot`

These are test migrations only. There are no runtime or syscall implementation changes in this PR.


marxin (approved):

Arshia001 (changes requested):
I'm _highly_ skeptical of the proc_snapshot test. snapshotting only works with asyncify, which means we're either:
* not running these tests properly (unlikely), or
* we're using an old enough version of wasixcc that would default to the asyncify configuration; hence, updating wasixcc will break this test.

marxin:
@artemyarulin ping here

PR #6458, opened 2026-04-12
Fix: Windows uninstaller re-adds package bin path to PATH

jedisct1:
SUMMARY
The Windows uninstall hook removes `{app}\bin` from PATH but mistakenly calls `EnvAddPath` for `{app}\globals\wapm_packages\.bin`, leaving that stale path behind after uninstall.

PROVENANCE

PRECONDITIONS
- The Windows installer is used to install Wasmer.
- The product is later uninstalled, triggering `CurUninstallStepChanged`.

PROOF
1. Input/source/state origin: installation adds both `{app}\bin` and `{app}\globals\wapm_packages\.bin` to PATH in `scripts/windows-installer/wasmer.iss:79-85`.
2. Control-flow and data-flow path: uninstall runs `CurUninstallStepChanged` in `scripts/windows-installer/wasmer.iss:88-95`.
3. Failing condition or violated invariant: the uninstall hook removes `{app}\bin` with `EnvRemovePath(...)` at line 92, but line 93 calls `EnvAddPath(...)` for `{app}\globals\wapm_packages\.bin` instead of removing it.
4. Resulting impact: uninstall leaves a dead Wasmer path in the user PATH variable, so command resolution and environment state remain incorrect after removal.
5. Why this is reachable in the current code: `CurUninstallStepChanged` is the tracked uninstall handler, and the wrong function call is unconditional inside its `usPostUninstall` branch.

WHY THIS IS A REAL BUG
This is a real lifecycle bug in shipped installer logic: uninstall mutates PATH in the wrong direction, leaving stale machine state behind.

PATCH RATIONALE
The patch changes the single incorrect `EnvAddPath` call to `EnvRemovePath`. It is the smallest fix and only affects uninstall cleanup.

RESIDUAL RISK
None

# Description


marxin (approved):
LGTM, merge current main for addressing the CI failure.

jedisct1:
> LGTM, merge current main for addressing the CI failure.

Done.

PR #6459, opened 2026-04-12
Fix: WASIX decodes IPv6 addresses in host endianness

jedisct1:
SUMMARY
Several WASIX helpers reinterpret network-order IPv6 octets as host-endian `u16`s, so every IPv6 address, CIDR, and route is byte-swapped on little-endian hosts.

PROVENANCE

PRECONDITIONS
- The host is little-endian, which is the current mainstream target for this repo’s supported Linux/macOS/Windows architectures.
- A WASIX syscall path reads an IPv6 address, CIDR, or route through these helpers.

PROOF
1. Input/source/state origin: raw IPv6 bytes come from guest memory via `__wasi_addr_t`, `__wasi_addr_ip6_t`, `__wasi_cidr_t`, and `Route` reads in `lib/wasix/src/net/mod.rs:22-29`, `:51-57`, `:94-101`, and `:231-268`.
2. Control-flow and data-flow path: each IPv6 read path uses `transmute::<[u8; 16], [u16; 8]>` and then passes the resulting words to `Ipv6Addr::new(...)` in `lib/wasix/src/net/mod.rs:33-34`, `:58-59`, `:113-116`, `:253-256`, and `:268-269`.
3. Failing condition or violated invariant: network byte order is big-endian, but `transmute` produces host-endian `u16`s. On little-endian machines, octets `20 01` become segment `0x0120` instead of `0x2001`.
4. Resulting impact: IPv6 bind/connect/send/route operations use the wrong addresses and routes, corrupting network behavior for every call site that depends on these helpers.
5. Why this is reachable in the current code: the helper functions are the shared decode path for IPv6 values consumed by socket and route syscalls, and the file already shows the correct octet-preserving form elsewhere (`Ipv6Addr::from(octets)` at `lib/wasix/src/net/mod.rs:179`).

WHY THIS IS A REAL BUG
This is a concrete cross-platform data corruption bug: the same encoded IPv6 bytes deterministically decode to the wrong address on little-endian hosts.

PATCH RATIONALE
The patch replaces the host-endian `transmute` paths with `Ipv6Addr::from(octets)`, which preserves the encoded network byte order and is the smallest correct fix.

RESIDUAL RISK
None

# Description

### DIFF
--- a/lib/wasix/src/net/mod.rs
+++ b/lib/wasix/src/net/mod.rs
@@ -29,10 +29,7 @@ pub(crate) fn read_ip<M: MemorySize>(
     let o = addr.u.octs;
     Ok(match addr.tag {
         Addressfamily::Inet4 => IpAddr::V4(Ipv4Addr::new(o[0], o[1], o[2], o[3])),
-        Addressfamily::Inet6 => {
-            let [a, b, c, d, e, f, g, h] = unsafe { transmute::<[u8; 16], [u16; 8]>(o) };
-            IpAddr::V6(Ipv6Addr::new(a, b, c, d, e, f, g, h))
-        }
+        Addressfamily::Inet6 => IpAddr::V6(Ipv6Addr::from(o))
         _ => return Err(Errno::Inval),
     })
 }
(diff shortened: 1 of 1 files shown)


jedisct1:
@copilot apply changes based on the comments in [this thread](https://github.com/wasmerio/wasmer/pull/6459#pullrequestreview-4095409744)

PR #6460, opened 2026-04-12
Fix: Module loading panics on non-UTF-8 file paths

jedisct1:
SUMMARY
Loading a module from an existing file whose canonical path is not valid UTF-8 panics instead of returning an `IoCompileError`.

PROVENANCE

PRECONDITIONS
- `BackendModule::from_file()` is called with a path that exists and can be read.
- The canonicalized filesystem path contains non-UTF-8 bytes, which is allowed on Unix filesystems.

PROOF
1. Input/source/state origin: `BackendModule::from_file()` accepts an arbitrary `impl AsRef<Path>` in `lib/api/src/entities/module/inner.rs:73-76`.
2. Control-flow and data-flow path: it canonicalizes the path, reads the file, creates the module, then converts the canonical path with `canonical.as_path().to_str().unwrap()` in `lib/api/src/entities/module/inner.rs:78-84`.
3. Failing condition or violated invariant: `Path::to_str()` returns `None` for valid non-UTF-8 Unix paths, and `unwrap()` panics instead of propagating an error.
4. Resulting impact: the embedding process aborts while loading a valid file path, violating the function’s `Result<_, IoCompileError>` contract.
5. Why this is reachable in the current code: no earlier step rejects non-UTF-8 paths; canonicalization and file reading both operate on `Path`, not UTF-8 strings.

WHY THIS IS A REAL BUG
This is a concrete filesystem-boundary bug, not a style issue: valid inputs cause a panic in a function that advertises recoverable error handling.

PATCH RATIONALE
The patch uses `to_string_lossy()` for the debug name instead of `unwrap()`. That preserves existing naming behavior for UTF-8 paths and avoids crashing on valid non-UTF-8 paths without changing module loading semantics.

RESIDUAL RISK
None


marxin (changes requested):
Good observation, can you please include a test-case that will test it?

jedisct1:
> Good observation, can you please include a test-case that will test it?

Done!

lib/api/tests/module.rs
```diff
@@ -292,3 +297,27 @@ fn module_custom_sections() -> Result<(), String> {
     );
     Ok(())
 }
+
+#[test]
+#[cfg(unix)]
+fn module_from_file_non_utf8_path() -> Result<(), String> {
+    let store = Store::default();
+    // Minimal valid wasm module: magic number + version
+    let wasm_bytes: &[u8] = b"\x00asm\x01\x00\x00\x00";
```
marxin: I would rather use `wat2wasm` of something like `"(module)"`?

lib/api/tests/module.rs
```diff
@@ -292,3 +297,27 @@ fn module_custom_sections() -> Result<(), String> {
...
+
+    let dir = tempfile::tempdir().map_err(|e| format!("{e:?}"))?;
+    let non_utf8_name = OsStr::from_bytes(b"module_\xff\xfe.wasm");
+    let path = dir.path().join(non_utf8_name);
+
+    // Some filesystems (e.g. macOS APFS/HFS+) reject non-UTF-8 filenames.
```
marxin: I would simplify it to just `Err(..) => return Ok(())`, would not hard-code any specific exit code.

lib/api/tests/module.rs
```diff
@@ -292,3 +297,27 @@ fn module_custom_sections() -> Result<(), String> {
...
+    // Skip the test on those platforms instead of failing.
+    match std::fs::write(&path, wasm_bytes) {
+        Err(e) if e.raw_os_error() == Some(92) => return Ok(()),
+        Err(e) => return Err(format!("{e:?}")),
+        Ok(()) => {}
+    }
+
+    let module = Module::from_file(&store, &path).map_err(|e| format!("{e:?}"))?;
+    assert!(module.name().is_some());
```
marxin: Can we ensure it matches the expected `non_utf8_name`?

marxin (changes requested):

lib/api/tests/module.rs
jedisct1: Done!

lib/api/tests/module.rs
jedisct1: Done :)

lib/api/tests/module.rs
jedisct1: Ok!

marxin (approved):
LGTM, fix the CI and we're ready to go.

jedisct1:
Done.

marxin (approved):

PR #6462, opened 2026-04-12
Fix: Argus test runners record elapsed time backwards

jedisct1:
SUMMARY
Argus computes durations as `start - Instant::now()`, which saturates to zero instead of measuring elapsed time, so stored test timing data is wrong.

PROVENANCE

PRECONDITIONS
- An Argus CLI or library test run completes and creates a `TestReport`.

PROOF
1. Input/source/state origin: both test runners capture a start instant before running tests in `tests/wasmer-argus/src/argus/tester/cli_tester.rs:148` and `tests/wasmer-argus/src/argus/tester/lib_tester.rs:67`.
2. Control-flow and data-flow path: the CLI runner passes `start_time - Instant::now()` to `TestReport::new()` in both success and error paths at `tests/wasmer-argus/src/argus/tester/cli_tester.rs:87-110`; the library runner does the same at `tests/wasmer-argus/src/argus/tester/lib_tester.rs:152-158`.
3. Failing condition or violated invariant: the later instant is subtracted from the earlier one. `Instant` subtraction in this direction does not yield elapsed time, so the recorded duration is not the time spent running the test.
4. Resulting impact: Argus persists incorrect runtime telemetry for every recorded test result.
5. Why this is reachable in the current code: every `TestReport` created by these runners uses the reversed subtraction expression.

WHY THIS IS A REAL BUG
This is a concrete data-integrity bug in persisted test results, not a stylistic issue. The recorded values are wrong for all executions of these code paths.

PATCH RATIONALE
The patch replaces the reversed subtraction with `start_time.elapsed()` / `start.elapsed()`. This is the minimal, standard way to measure elapsed time and does not alter test outcomes beyond fixing the recorded duration.

RESIDUAL RISK
None


marxin:
LGTM, can you please merge the main branch to address the CI failure?

marxin (approved):

jedisct1:
> LGTM, can you please merge the main branch to address the CI failure?

Done.

PR #6463, opened 2026-04-12
wasix: migrate proc_exec POSIX tests to wasm_tests

artemyarulin:
This migrates the next tests-only POSIX batch from `tests/wasix-tests` into `lib/wasix/tests/wasm_tests`.

This PR adds:
- `proc_exec`
- `proc_exec2`

These are test migrations only. There are no runtime or syscall implementation changes in this PR.


marxin (approved):

Arshia001 (approved):
This _might_ also depend on asyncify if a collection of things happens at the same time, but I don't think they will:
* deep sleep must be enabled
* the new process should take over 50ms to execute
I don't think deep sleep is enabled for these tests, so we should be fine.

PR #6468, opened 2026-04-14
fix(wasix): Make connect nonblocking when requested and implement real socket::status

artemyarulin:
Around a month ago we had an issue where Python was getting a timeout when trying to connect to external hosts in nonblocking mode. In [#6298](https://github.com/wasmerio/wasmer/pull/6298) I tried to fix it, but did it in the wrong way: I made WASIX only return the socket after the connection was fully established. It fixed the Python issue, but it also effectively made `connect()` blocking.

This PR reverts that and fixes the real issue instead: `socket::status()` was always returning hardcoded `Ready` for TCP sockets.

So what was really happening was:
- nonblocking connect() returned a socket
- client code checked its status
- status said Ready
- client started sending data,
- but the TCP handshake could still still be in progress.
So it was racy and wrong.

This change keeps `connect()` nonblocking again and fixes `socket::status()` so it returns the real state of the socket instead of always pretending it is already ready.

### DIFF
--- a/lib/virtual-net/src/host.rs
+++ b/lib/virtual-net/src/host.rs
@@ -19,7 +19,7 @@ use std::os::fd::RawFd;
 #[cfg(windows)]
 use std::os::windows::io::AsRawSocket;
 
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::task::Poll;
 use std::time::Duration;
 use tokio::runtime::Handle;
@@ -175,16 +175,7 @@ impl VirtualNetworking for LocalNetworking {
             return Err(NetworkError::PermissionDenied);
         }
 
-        // This future may be polled outside Tokio's executor context, so run
-        // the connect itself on the stored runtime handle to guarantee a reactor.
-        let stream = self
-            .handle
-            .spawn(tokio::net::TcpStream::connect(peer))
-            .await
-            .map_err(|_| NetworkError::IOError)?
-            .map_err(io_err_into_net_error)?;
-        let stream = stream.into_std().map_err(io_err_into_net_error)?;
-        let stream = mio::net::TcpStream::from_std(stream);
+        let stream = mio::net::TcpStream::connect(peer).map_err(io_err_into_net_error)?;
 
         if let Ok(p) = stream.peer_addr() {
             peer = p;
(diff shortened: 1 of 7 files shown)


lib/virtual-net/src/host.rs
```diff
@@ -394,6 +392,7 @@ pub struct LocalTcpStream {
     selector: Arc<Selector>,
     handler_guard: HandlerGuardState,
     buffer: BytesMut,
+    connect_state: Mutex<ConnectState>,
```
artemyarulin, line 395: Mutex because status could in theory be checked from multiple threads. RWLock I think is not worth in here

lib/virtual-net/src/host.rs
```diff
@@ -614,7 +614,40 @@ impl VirtualSocket for LocalTcpStream {
     }
 
     fn status(&self) -> Result<SocketStatus> {
-        Ok(SocketStatus::Opened)
+        // `take_error()` consumes the latched socket error, so once the
+        // connect resolves we cache the terminal state to keep status() stable.
+        let mut connect_state = self.connect_state.lock().unwrap();
+        match *connect_state {
+            ConnectState::Opened => return Ok(SocketStatus::Opened),
+            ConnectState::Failed => return Ok(SocketStatus::Failed),
```
theduke, line 622: Should retain the error, but can be done as a followup.
Add a TODO

theduke (approved):

artemyarulin:
Python upstream result: IMPROVEMENT

Baseline: main @ 03769ab
Candidate: fix/socket-non-blocking @ 445554e

Summary:
- Regressions: 0
- Improvements: 2
- FLAKY: 0 -> 57
- PASS: 0 -> 28313
- FAIL: 424 -> 2639
- TIMEOUT: 0 -> 4915
- SKIP: 0 -> 2040

See full details at [commit](https://github.com/wasmerio/compat-tests/commit/73eeaa53f6fbf829e5209ca48999411fd8f8e494).

Top improvements:
- `test.test_extcall` (FAIL -> PASS)
- `test.test_syntax` (FAIL -> PASS)

More info:
- compat-tests workflow: https://github.com/wasmerio/compat-tests/actions/runs/24392839151
- compat-tests results commit: https://github.com/wasmerio/compat-tests/commit/73eeaa53f6fbf829e5209ca48999411fd8f8e494
- compat-tests branch: https://github.com/wasmerio/compat-tests/tree/fix/socket-non-blocking
- full test log: https://github.com/wasmerio/compat-tests/actions/runs/24392839151/artifacts/6425940351

PR #6469, opened 2026-04-14
Remove build-scripts dependency

Arshia001:
Having a dependency on build-scripts is the wrong choice, especially since we're going to archive it soon anyway. This PR reworks the dynamic call/closure tests so they don't depend on libffi directly (reducing the possibility of false negatives), which was the only thing that relied on external build artifacts, and eliminates that dependency.


marxin (approved):
LGTM, good simplification.

PR #6480, opened 2026-04-16
compiler/engine: fix 'occured' -> 'occurred' in error doc comment

SAY-5:
Doc comment in `lib/compiler/src/engine/error.rs` line 48 reads `A runtime error occured`. Fixed to `occurred`. Comment-only change.


marxin:
Does not make sense fixing a random spell checking issue - my bet is we've got hundreds of them.
A more systematic approach would be leveraging any of the existing automation tools: #6481.

PR #6486, opened 2026-04-17
Stop handing out zero timestamps for mount points from mountfs

Arshia001:
# Description

### DIFF
--- a/lib/virtual-fs/src/mount_fs.rs
+++ b/lib/virtual-fs/src/mount_fs.rs
@@ -10,6 +10,7 @@ use std::{
     ffi::OsString,
     path::{Path, PathBuf},
     sync::{Arc, RwLock},
+    time::{SystemTime, UNIX_EPOCH},
 };
 
 const DEFAULT_METADATE_TIME: u64 = 1_000_000_000; // 1 second in nano seconds
@@ -27,6 +28,7 @@ pub enum ExactMountConflictMode {
 struct MountedFileSystem {
     fs: DynFileSystem,
     source_path: PathBuf,
+    created_at: u64,
 }
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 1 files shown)


syrusakbary (approved):

theduke:
Superseeded by https://github.com/wasmerio/wasmer/pull/6487

lib/virtual-fs/src/mount_fs.rs
```diff
@@ -87,14 +93,26 @@ impl MountPoint {
...
+impl Default for MountFileSystem {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl MountFileSystem {
     pub fn new() -> Self {
-        Self::default()
+        let ts = Self::now_nanos();
```
theduke, line 111: Is there a motivation for using `now()`.
This makes the whole thing highly indeterministic between different runs.

Arshia001: The idea is that a mount point's (and a synthetic node's) timestamps should correspond to the time it was created or updated. The root is created alongside the mount FS instance, and other nodes get the current timestamp as they're created. I strongly disagree with the implementation in #6487, which simply responds with 1 second past epoch for everything.

lib/virtual-fs/src/mount_fs.rs
```diff
@@ -174,12 +196,18 @@ impl MountFileSystem {
         normalized
     }
 
-    fn directory_metadata() -> Metadata {
+    fn now_nanos() -> u64 {
+        SystemTime::now()
```
theduke, line 202: If there was a system setting system time to 0 for reproducibility, should we do a `.max(DEFAULT_METADATE_TIME)` here to avoid the `0` problems?

Arshia001: hmmmmm....................... very interesting point.

PR #6492, opened 2026-04-20
chore: drop `Universal` from naming

marxin:
It’s been a while since Wasmer became a `universal` runtime (v3.0), and I think it’s time to remove the remaining occurrences of that terminology from the codebase. It was confusing to me and now feels counter-intuitive.


theduke (approved):
Makes sense to me, though @syrusakbary should approve.

syrusakbary (approved):

PR #6495, opened 2026-04-20
ci: use wasmer packages instead of sharrattj + enable more tests

marxin:
I noticed we're testing a couple of legacy `sharrattj/*` packages, would be better using the `wasmer` namespace. Plus, a couple of tests re-enabled.


theduke (approved):
Good catch, those packages are ancient.

But may lead to failing tests.

PR #6499, opened 2026-04-20
fix(virtual-fs): resolve symlinks in host_fs root check

### LINKED ISSUE #6477: test_rejects_host_absolute_paths_inside_root test fails on macOS
Fails since #6391 PR: https://github.com/wasmerio/wasmer/actions/runs/24361027884/job/71140534166

artemyarulin:
On macOS `TempDir` returns paths under `/var/folders/...`, but `/var` is a symlink to `/private/var`. This PR canonicalize absolute paths in `prepare_path` before the root check so it sees through symlink aliases uniformly across platforms. 

It's not only MacOS specific, so fix is applied to all platforms.

Should fix https://github.com/wasmerio/wasmer/issues/6477

### DIFF
--- a/lib/virtual-fs/src/host_fs.rs
+++ b/lib/virtual-fs/src/host_fs.rs
@@ -83,6 +83,14 @@ impl FileSystem {
             return Err(FsError::InvalidInput);
         }
 
+        // Resolve symlinks so the root check below sees through aliases, example `/var` -> `/private/var`
+        // on macOS. Skip relative paths to avoid resolving against the process CWD.
+        let path = if path.is_absolute() {
+            dunce::canonicalize(&path).unwrap_or(path)
+        } else {
+            path
+        };
+
         if self.root != Path::new("/") && path.starts_with(&self.root) {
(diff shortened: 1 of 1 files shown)


Arshia001 (changes requested):
Tentatively requesting changes until I do a deep dive, I think we may be missing more important things here.

lib/virtual-fs/src/host_fs.rs
```diff
@@ -83,6 +83,14 @@ impl FileSystem {
             return Err(FsError::InvalidInput);
         }
 
+        // Resolve symlinks so the root check below sees through aliases, example `/var` -> `/private/var`
```
marxin, line 86: Re-enable the `test_rejects_host_absolute_paths_inside_root` test - later in the file.

Arshia001:
OK, this is definitely the wrong fix because we're host-canonicalizing a guest path. this works for `--volume /path/to/dir` because the host and guest path would be the same, but not for any other case. Closing this PR.

PR #6517, opened 2026-04-25
fix(wasix): honor socket connect_timeout in TCP connect

### LINKED ISSUE #6450: TCP connect() ignores short socket timeouts under Wasmer
Under Wasmer, a TCP `connect()` to a blackholed address does not respect a very short socket timeout. Instead of timing out in about 0.001s , it can block for about 30s. This breaks `test.test_timeout.TCPTimeoutTestCase.testConnectTimeout` CPython test

```python
import socket
import time

sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.settimeout(0.001)

t1 = time.monotonic()
try:
    sock.connect(("pythontest.net", 56666))
except Exception as e:
    print(type(e).__name__, getattr(e, "errno", None), repr(e))
finally:
    print("elapsed", time.monotonic() - t1)
    sock.close()

```

This looks like a Wasix socket timeout bug. The `TCP` connect path in `socket.rs` appears to await the underlying connection attempt too directly, so the caller ends up waiting on the network stack’s own timing instead of the Python socket timeout. This is related to the nonblocking-connect issues, but distinct: here the problem is that short connect timeouts are not honored.

dhir1007:
## Problem
When a WASI program calls `settimeout()` on a socket and then `connect()`, 
Wasmer ignores the timeout and blocks for up to 30 seconds instead.

This is because `sock_connect_internal` hardcoded `None` as the timeout 
argument to `socket.connect()`, which caused it to always fall back to the 
30s default inside `socket.rs`.

## Fix
Read `connect_timeout` from the socket via `opt_time(TimeType::ConnectTimeout)` 
and pass it through to `socket.connect()`, so the user-set timeout is respected.

This follows the exact same pattern already used in `sock_accept` for 
`accept_timeout`.

## Testing
The original reproducer from the issue:
```python
import socket, time
sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.settimeout(0.001)
t1 = time.monotonic()
try:
    sock.connect(("pythontest.net", 56666))
except Exception as e:
    print(type(e).__name__, getattr(e, "errno", None))
finally:
    print("elapsed", time.monotonic() - t1)
    sock.close()
```
Should now timeout in ~0.001s instead of ~30s.

Fixes #6450

### DIFF
--- a/lib/wasix/src/net/socket.rs
+++ b/lib/wasix/src/net/socket.rs
@@ -582,7 +582,9 @@ impl InodeSocket {
         let new_write_timeout;
         let new_read_timeout;
 
-        let timeout = timeout.unwrap_or(Duration::from_secs(30));
+        let timeout = timeout
+            .or_else(|| self.opt_time(TimeType::ConnectTimeout).ok().flatten())
+            .unwrap_or(Duration::from_secs(30));
 
         let handler;
         let connect = {
--- a/lib/wasix/src/syscalls/wasix/sock_connect.rs
+++ b/lib/wasix/src/syscalls/wasix/sock_connect.rs
@@ -80,10 +80,7 @@ pub(crate) fn sock_connect_internal(
                 .auto_bind_udp(tasks.deref(), net.deref())
                 .await?
                 .unwrap_or(socket);
-            let timeout = socket
-                .opt_time(TimeType::ConnectTimeout)
-                .ok()
-                .flatten();
+            let timeout = socket.opt_time(TimeType::ConnectTimeout).ok().flatten();
             socket
                 .connect(
                     tasks.deref(),
--- a/lib/wasix/src/syscalls/wasix/sock_connect.rs
+++ b/lib/wasix/src/syscalls/wasix/sock_connect.rs
@@ -1,6 +1,6 @@
 use super::*;
 use crate::syscalls::*;
-
+use crate::net::socket::TimeType;
 /// ### `sock_connect()`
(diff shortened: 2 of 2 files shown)


dhir1007:
The failing test `merged_filesystem_contains_all_files` in the Windows 
integration job appears unrelated to this change.

I verified this by checking that the test file `tests/integration/cli/tests/run.rs` 
contains no references to `sock_connect` or `connect_timeout` - only the 
`merged_filesystem` test function itself appears. This change only touches 
`sock_connect_internal` in `lib/wasix/src/syscalls/wasix/sock_connect.rs`.

This looks like a pre-existing failure on the Windows CI runner. Happy to 
investigate further if needed.

Arshia001 (changes requested):
Thanks for submitting this fix! I think the change is correct in principle, but implemented at the wrong level. Rather than passing the timeout into the socket's method, we should let the `connect` method read it out of the socket's properties itself, so the logic in `connect` becomes: `timeout.or_else(self.opt_time()).unwrap_or(30secs)`.

dhir1007:
Hii @Arshia001 Thanks for the feedback! Moved the timeout resolution into `socket.connect()` 
itself - it now falls back to the socket's `connect_timeout` property if no 
explicit timeout is passed, following your suggested pattern. 
`sock_connect.rs` is back to passing `None`.

Arshia001 (approved):

Arshia001:
@dhir1007 thanks! Merging.

PR #6520, opened 2026-04-27
Fix files opened in read-only mode being writable

Arshia001:


syrusakbary (approved):

artemyarulin:
@Arshia001 what about comments just before in the code? Should we remove those or is that about something else?

```
// TODO: traverse rights of dirs properly
    // COMMENTED OUT: WASI isn't giving appropriate rights here when opening
    //              TODO: look into this; file a bug report if this is a bug
    //
```

syrusakbary (approved):

PR #6532, opened 2026-05-01
Fix/wasix subprocess feature detection

### LINKED ISSUE #6515: macOS: `bash` invoking a `coreutils` command leads to: `compile error: "exceptions proposal not enabled"`
I noticed the CI results on `main` are red again and it's something unrelated to the changes in the Wasmer repo.

What happens on macOS:
```
❯ echo "ls /" | wasmer run wasmer/bash
bash-dist#
/usr/bin/ls: /usr/bin/ls: Operation not permitted
bash-dist# exit
```

while on Linux I get a correct result:
```
❯ echo "ls /" | wasmer run wasmer/bash
bash-dist#
dev
etc
tmp
usr
bash-dist# exit
```

I suspect it's related to a recent release of the `coreutils` package:
https://wasmer.io/wasmer/coreutils/releases

Arshia001:
This PR allows WASIX subprocesses to select a different engine or different feature set relative to the parent process.

Fixes #6515.


lib/compiler/src/engine/inner.rs
```diff
@@ -327,6 +327,29 @@ impl Engine {
 
         Ok(())
     }
+
+    /// Build a new Engine with the same compiler and configuration, but with this Engine's
+    /// features extended by the given features.
+    #[cfg(feature = "compiler")]
+    pub fn new_with_extended_features(&self, features: &Features) -> Result<Self, CompileError> {
```
marxin, line 334: I thought it's already something we've been doing, see #6222.

lib/c-api/Cargo.toml
```diff
@@ -80,16 +80,18 @@ sys = ["wasmer-api/sys", "wasmer-compiler"]
...
 	"dep:wasmer-compiler-singlepass",
 	"wasmer-api/singlepass",
+	"wasmer-wasix/singlepass",
 	"compiler",
 ]
 # Enable the `cranelift` compiler. 
 cranelift = [
 	"dep:wasmer-compiler-cranelift",
 	"wasmer-api/cranelift",
+	"wasmer-wasix/cranelift",
```
theduke, line 90: Why is this needed? Should cranelift already be pulled in through wasmer-api/cranelift ?

lib/c-api/Cargo.toml
```diff
@@ -80,16 +80,18 @@ sys = ["wasmer-api/sys", "wasmer-compiler"]
...
 	"compiler",
 ]
 # Enable the `llvm` compiler.
-llvm = ["dep:wasmer-compiler-llvm", "wasmer-api/llvm", "compiler"]
+llvm = ["dep:wasmer-compiler-llvm", "wasmer-api/llvm", "wasmer-wasix/llvm", "compiler"]
```
theduke, line 94: Same question as above

lib/compiler/src/engine/inner.rs
theduke: Yeah, we already enable all features, so this shouldn't be required.

This also breaks in-memory caching of modules, because Modules are tied to a specific engine ID, and ::new will lead to a new engine. 

Although this is not problematic for this PR in particular because of a separate engine caching mechanism.

lib/wasix/src/runtime/mod.rs
```diff
@@ -568,6 +632,132 @@ impl PluggableRuntime {
...
+        if let Some(current) = current_backend {
+            candidate_backends.push(current);
+        }
+        for backend in SysBackend::all() {
+            if !candidate_backends.contains(&backend) {
+                candidate_backends.push(backend);
+            }
+        }
+
+        for backend in candidate_backends {
```
theduke, line 671: Redundant extra loop , and redundant vec allocation?

theduke (commented):
I do not like the tight coupling between specific sys engines and WASIX.

WASIX should ideally be mostly agnostic/unaware of different backends.

I'd much rather have something like we have on Edge: an `EngineSet` abstraction that holds the available engines, and that can be used to determine the best available engine. 

Then users like the CLI can create an EngineSet and configure  it on the PluggableRuntime, with a prioritization.

That keeps things cleanly composable and flexible.

Thoughts?

marxin:
The EH support got added for Cranelift+macOS, so the need for this PR has decreased significantly. That said, I’m closing it for now.

PR #6536, opened 2026-05-03
fix(wasix): Interrupt atomic waiters on signals

### LINKED ISSUE #6585: wasix: Threads get stuck on atomic waiting, even after SIGKILL
Currently the atomics waiting system is not properly integrated with signal handling in WASIX.

Threads can get stuck indefinitely when waiting on atomic notify.<br><br>This happens even for KILL signals.

theduke:
Previously threads could get stuck idefinitely , even on KILL signals,
because they would be stuck in atomic waiting.

The downside of this implementation is that it can lead to spurious
wakeups, but most data structures building on top of atomics can usually
handle these!

Closes #6585

### DIFF
--- a/lib/vm/src/instance/mod.rs
+++ b/lib/vm/src/instance/mod.rs
@@ -931,11 +931,7 @@ impl Instance {
         };
         match unsafe { memory.do_wait(dst, expected, timeout) } {
             Ok(count) => Ok(count),
-            Err(_err) => {
-                // ret is None if there is more than 2^32 waiter in queue or some other error
-                // TODO: why THIS specific trap code tho? -.-
-                Err(Trap::lib(TrapCode::TableAccessOutOfBounds))
-            }
+            Err(_err) => Err(Trap::lib(TrapCode::HostInterrupt)),
         }
     }
 
--- a/lib/wasix/src/os/task/process.rs
+++ b/lib/wasix/src/os/task/process.rs
@@ -16,7 +16,7 @@ use std::{
     time::Duration,
 };
 use tracing::trace;
-use wasmer::FunctionEnvMut;
+use wasmer::{FunctionEnvMut, SharedMemory};
 use wasmer_types::ModuleHash;
 use wasmer_wasix_types::{
     types::Signal,
@@ -179,6 +179,10 @@ pub struct WasiProcessInner {
     pub snapshot_on: HashSet<SnapshotTrigger>,
     /// Any wakers waiting on this process (for example for a checkpoint)
     pub wakers: Vec<Waker>,
+    /// If true then the process has started cleaning up
+    pub cleanup_started: bool,
+    /// Shared process memory.
+    pub memory: Option<SharedMemory>,
     /// The snapshot memory significantly reduce the amount of
     /// duplicate entries in the journal for memory that has not changed
     #[cfg(feature = "journal")]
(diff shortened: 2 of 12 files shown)


lib/wasix/src/os/task/process.rs
```diff
@@ -885,11 +897,42 @@ fn signal_process_internal(process: &LockableWasiProcessInner, signal: Signal) {
...
+    }
     for thread in guard.threads.values() {
         thread.signal(signal);
     }
 }
 
+fn should_wake_atomic_waiters_for_signal(signal: Signal) -> bool {
+    // Atomic wait wakeups are memory-wide, so only use them for signals
+    // that should interrupt or terminate execution anyway.
+    matches!(
```
Arshia001: We definitely also need SIGWAKEUP here.

lib/wasix/src/os/task/process.rs
```diff
@@ -885,11 +897,42 @@ fn signal_process_internal(process: &LockableWasiProcessInner, signal: Signal) {
...
+            | Signal::Sigint
+            | Signal::Sigstop
+            | Signal::Sigpipe
+    )
+}
+
+fn wake_atomic_waiters(process: &WasiProcessInner) {
+    let pid = process.pid;
+    if let Some(memory) = &process.memory {
+        if let Err(err) = memory.wake_all_atomic_waiters() {
```
Arshia001: Does this API both exist _and_ work properly for all backends? Or are we doing a sys-only thing here?

Arshia001 (changes requested):
LGTM overall, but this needs intergration with the linker as well (waking atomic waiters up for SIGWAKEUP and adding a test for the two things working together.) Happy to take over from here @theduke?

lib/wasix/src/os/task/process.rs
Arshia001: Eh, definitely sys-only.

lib/vm/src/instance/mod.rs
```diff
@@ -934,7 +934,7 @@ impl Instance {
             Err(_err) => {
                 // ret is None if there is more than 2^32 waiter in queue or some other error
                 // TODO: why THIS specific trap code tho? -.-
```
Arshia001: @theduke feel free to remove this piece of art from yours truly 😄

lib/wasix/src/os/task/process.rs
theduke: yeah, sys only unfortunately.

lib/wasix/src/os/task/process.rs
```diff
@@ -597,6 +603,12 @@ impl WasiProcess {
         signal_process_internal(&self.inner, signal);
     }
 
+    /// Registers the shared memory used by this process.
+    pub fn register_memory(&self, memory: SharedMemory) {
+        let mut inner = self.inner.0.lock().unwrap();
+        inner.memory = Some(memory);
+    }
```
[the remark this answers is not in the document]
theduke: Handled properly now.

lib/wasix/src/os/task/process.rs
```diff
@@ -885,11 +913,49 @@ fn signal_process_internal(process: &LockableWasiProcessInner, signal: Signal) {
     }
 
     // Otherwise just send the signal to all the threads
+    wake_atomic_waiters(&guard, signal);
     for thread in guard.threads.values() {
         thread.signal(signal);
     }
 }
 
+fn wake_atomic_waiters(process: &WasiProcessInner, signal: Signal) {
```
Arshia001, line 922: This function only ever does useful things for SIGKILL, so we should eliminate all the other cases.

Arshia001 (approved):
Added one comment which needs to be addressed, LGTM otherwise. Approving already.

theduke:
We decided to scope the changes here and only handle SIGKILL for now, since we have low trust in this being useful for other signals.

PR #6547, opened 2026-05-04
Fix append + Seek and support of writing beyond EOF

### LINKED ISSUE #6543: ftruncate followed by write past new EOF writes at the wrong offset
`ftruncate()` does not change the file offset, so if the offset was 17 and the file is truncated to 0, the next write should create a sparse file with zeroes before the written byte. Wasmer instead writes the byte at the beginning and the file reads back as length 1.

```c
#include <fcntl.h>
#include <stdio.h>
#include <unistd.h>

int main() {
    int fd = open("x", O_CREAT | O_RDWR, 0666);
    write(fd, "12345678901234567", 17);
    ftruncate(fd, 0);
    write(fd, "Z", 1);
    lseek(fd, 0, SEEK_SET);
    unsigned char b[32] = {0};
    int n = read(fd, b, sizeof(b));
    printf("n=%d hex=", n);
    for (int i = 0; i < n; i++) printf("%02x", b[i]);
    printf("\n");
}
```

Native
```
n=18 hex=00000000000000000000000000000000005a
```

wasmer
```
n=1 hex=5a
```

Related to the https://github.com/wasmerio/wasmer/issues/6542, but different as it requires sparse file support fix

artemyarulin:
DEPENDS ON https://github.com/wasmerio/wasmer/pull/6545 and could be merged (rebased) only after that.

## Append + Seek

`mem_fs` treated `O_APPEND` as if seeking was disabled for the descriptor. I assume that came from reading the POSIX/Linux `O_APPEND` wording too broadly: the spec says the file offset is moved to `EOF` before each `write()`, but it does not make `lseek()` a no-op. The expected behavior is that `lseek()` still updates the descriptor offset for reads/tells, while writes on an append-mode descriptor are still performed at EOF.

This matters for log readers/rotators or database-style append logs that open a file with `O_APPEND` for safe appends, but also seek around the same descriptor to read headers, inspect existing records, or recover the last valid entry. Those reads must use the position set by `lseek()`, even though writes still append at `EOF`.

## Sparse

This PR also fixes sparse writes by allowing regular-file seeks past `EOF`, returning `EOF` for reads at or beyond `EOF`, and zero-filling the gap when a later write materializes the sparse region. 

This matters for tools that preallocate or patch files by seeking to an offset before writing, such as archive writers, database files, binary patchers, and language runtimes that use `ftruncate()`/`lseek()` to resize or rewrite structured files. Without sparse write support, Wasmer writes at the wrong offset after truncation or seek-past-EOF, corrupting the file layout instead of preserving the expected zero-filled hole.

Fixes https://github.com/wasmerio/wasmer/issues/6543

### DIFF
--- a/lib/virtual-fs/src/mem_fs/file.rs
+++ b/lib/virtual-fs/src/mem_fs/file.rs
@@ -809,10 +809,6 @@ impl AsyncRead for FileHandle {
 
 impl AsyncSeek for FileHandle {
     fn start_seek(mut self: Pin<&mut Self>, position: io::SeekFrom) -> io::Result<()> {
-        if self.append_mode {
-            return Ok(());
-        }
-
         let mut cursor = self.cursor;
         let ret = {
             let mut fs = self
@@ -868,24 +864,6 @@ impl AsyncSeek for FileHandle {
     }
 
     fn poll_complete(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<u64>> {
-        // In `append` mode, it's not possible to seek in the file. In
-        // [`open(2)`](https://man7.org/linux/man-pages/man2/open.2.html),
-        // the `O_APPEND` option describes this behavior well:
-        //
-        // > Before each write(2), the file offset is positioned at
-        // > the end of the file, as if with lseek(2).  The
-        // > modification of the file offset and the write operation
-        // > are performed as a single atomic step.
-        // >
-        // > O_APPEND may lead to corrupted files on NFS filesystems
-        // > if more than one process appends data to a file at once.
-        // > This is because NFS does not support appending to a file,
-        // > so the client kernel has to simulate it, which can't be
-        // > done without a race condition.
-        if self.append_mode {
-            return Poll::Ready(Ok(0));
-        }
-
         let mut fs = self
             .filesystem
             .inner
@@ -1505,6 +1483,12 @@ impl File {
 impl File {
     pub fn read(&self, buf: &mut [u8], cursor: &mut u64) -> io::Result<usize> {
         let cur_pos = *cursor as usize;
+
+        // POSIX regular files return EOF, not an error, when reading at or beyond EOF.
+        if cur_pos >= self.buffer.len() {
+            return Ok(0);
+        }
+
         let max_to_read = cmp::min(self.buffer.len() - cur_pos, buf.len());
         let data_to_copy = &self.buffer[cur_pos..][..max_to_read];
 
@@ -1546,10 +1530,8 @@ impl File {
             ));
         }
 
-        // In this implementation, it's an error to seek beyond the
-        // end of the buffer.
         let next_cursor = next_cursor.try_into().map_err(to_err)?;
-        *cursor = cmp::min(self.buffer.len() as u64, next_cursor);
+        *cursor = next_cursor;
 
         let cursor = *cursor;
         Ok(cursor)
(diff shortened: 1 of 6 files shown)


lib/virtual-fs/src/mem_fs/file.rs
```diff
@@ -1557,15 +1541,23 @@ impl File {
 impl File {
     pub fn write(&mut self, buf: &[u8], cursor: &mut u64) -> io::Result<usize> {
         let position = *cursor as usize;
+        let end = position
+            .checked_add(buf.len())
+            .ok_or(io::ErrorKind::InvalidInput)?;
+
+        if position > self.buffer.len() {
+            // Materialize the sparse gap with zeroes in this contiguous in-memory buffer.
```
Arshia001, line 1564: eh, this will hurt memory usage for things that depend on sparse gaps to be SPARSE...

Arshia001 (approved):
LGTM, but CI is red

artemyarulin:
@Arshia001 can you re-review follow up commit https://github.com/wasmerio/wasmer/pull/6547/commits/bf1e6d4801ec4663e5a8fa91992413f16a639ed4 in this PR? 


Basically it fixes cursor semantics to match POSIX/Linux `O_APPEND`, as in docs:

> Before each write(2), the file offset is positioned at the end of the file, as if with lseek(2).

Previously, `mem_fs` implemented append mode by moving the handle cursor to `EOF` when opening an existing file with append enabled. That made append writes land at the end, but it put the `EOF` positioning in the wrong place: the visible file offset after `open(..., O_APPEND)` became `EOF` instead of the normal initial offset.

After I've removed seek being no-op for seek it uncovered that existing issue: opening an existing file in append mode reported the cursor at EOF which is wrong as per doc only write handle should move to EOF upon writing. 

But now that commit does:

- append-mode opens keep the normal initial cursor position
- seeks and reads observe the handle cursor normally
- But all the writes uses a write cursor from the current file size immediately before writing

artemyarulin:
# Shield - Improvement 🎉🎉🎉

| Language | Tests  | Pass rate now | PASS | FAIL | TIMEOUT | CRASH |
| -------- | ------ | ------------- | ---- | ---- | ------- | ----- |
| Python | 37,920 | 69.0% | $${\color{green}+13}$$ | $${\color{red}+36}$$ | $${\color{green}-49}$$ | 0 |
| Node.js | 4,861 | 32.7% | $${\color{red}-4}$$ | $${\color{green}-30}$$ | $${\color{red}+35}$$ | $${\color{green}-4}$$ |
| PHP | 19,912 | 72.8% | $${\color{green}+1}$$ | $${\color{green}-1}$$ | 0 | 0 |
| Rust | 22,337 | 94.3% | 0 | 0 | 0 | 0 |

- Examples from [tests_python_results.json](https://github.com/wasmerio/compat-tests/blob/dc3871627103a2ce5544443a00c9ddba47cb38fa/tests_python_results.json):
  - `test.test_largefile.CLargeFileTest.test_lseek` (`FAIL -> PASS`)
  - `test.test_largefile.CLargeFileTest.test_osstat` (`FAIL -> PASS`)
  - `test.test_largefile.CLargeFileTest.test_seek_read` (`FAIL -> PASS`)
  - `test.test_largefile.CLargeFileTest.test_truncate` (`FAIL -> PASS`)
  - `test.test_largefile.PyLargeFileTest.test_lseek` (`FAIL -> PASS`)

- Examples from [tests_php_results.json](https://github.com/wasmerio/compat-tests/blob/dc3871627103a2ce5544443a00c9ddba47cb38fa/tests_php_results.json):
  - `ext/standard/tests/strings/vfprintf_basic.phpt` (`FAIL -> PASS`)


## Artifacts

- GitHub Action: [https://github.com/wasmerio/anti-regression-shield/actions/runs/25362341834](https://github.com/wasmerio/anti-regression-shield/actions/runs/25362341834)
- Results commit: [https://github.com/wasmerio/compat-tests/commit/dc3871627103a2ce5544443a00c9ddba47cb38fa](https://github.com/wasmerio/compat-tests/commit/dc3871627103a2ce5544443a00c9ddba47cb38fa)

PR #6552, opened 2026-05-04
Migrate old shell-based wasix tests to the new format in wasmer-wasix

Arshia001:


lib/wasix/tests/wasm_tests/libc_tests.rs
```diff
@@ -18,3 +61,50 @@ wasm_test!(
     "variadic-args",
     stdout = "Printing 5, 6, 0, 42"
 );
+
+#[test]
+#[ignore = "file-backed mmap writeback does not currently persist under WasiRunner"]
+fn test_msync_end_of_file() {
```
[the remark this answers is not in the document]
Arshia001: old tests were wrong and weren't catching the impl being wrong, so non-issue

lib/wasix/tests/wasm_tests/context_switching.rs
```diff
@@ -93,3 +110,18 @@ wasm_test!(
     test_contexts_with_setjmp,
     "contexts_with_setjmp"
 );
+
+#[test]
+#[cfg(unix)]
+fn test_legacy_process_switching() {
```
marxin: Can't we manually create 6 invocations of `wasm_test!` macro instead of this loop?

lib/wasix/tests/wasm_tests/libc_tests/msync-end-of-file/main.c
```diff
@@ -22,7 +22,7 @@ int main() {
   fstat(fd, &statbuf);
   size_t filesize = statbuf.st_size;
 
-  data = mmap(NULL, 3, PROT_READ | PROT_WRITE, MAP_PRIVATE, fd, 3);
```
marxin, line 25: Why are these changes needed?

lib/wasix/tests/wasm_tests/fd_tests.rs
```diff
@@ -1,3 +1,54 @@
...
+    }
+
+    if path.is_dir() {
+        let _ = fs::remove_dir_all(path);
+    } else {
+        let _ = fs::remove_file(path);
+    }
+}
+
+fn run_fd_test(test_name: &str, cleanup_paths: &[&str]) {
```
marxin: There's a only one invocation of `run_fd_test`, can we inline it directly into caller? And note the `cleanup_paths` is an empty slice.

lib/wasix/tests/wasm_tests/libc_tests.rs
```diff
@@ -1,3 +1,46 @@
+use std::fs;
+use std::path::Path;
+
+use super::{run_build_script, run_wasm_with_result};
+
+fn remove_path_if_exists(path: &Path) {
```
marxin: A duplicated code `remove_path_if_exists` - can we move it to a shared place?

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -112,18 +115,95 @@ mod socket_tests;
...
+use tracing_subscriber::EnvFilter;
+use tracing_subscriber::fmt::MakeWriter;
+use tracing_subscriber::layer::SubscriberExt;
 use wasmer_wasix::VirtualFile as VirtualFileTrait;
 use wasmer_wasix::runners::MappedDirectory;
 use wasmer_wasix::runners::wasi::{RuntimeOrEngine, WasiRunner};
 use wasmer_wasix::runtime::module_cache::{HashedModuleData, ModuleCache};
 use wasmer_wasix::virtual_fs::{AsyncRead, AsyncSeek, AsyncWrite};
 
+static TRACE_SUBSCRIBER_INIT: OnceLock<()> = OnceLock::new();
```
marxin, line 216: Why do we need it? Can't we just stick to `RUST_LOG=trace cargo t --nocapture` if anybody needs to debug a failing test-case?

lib/wasix/tests/wasm_tests/path_tests.rs
```diff
@@ -15,6 +17,22 @@ fn remove_path_if_exists(path: &Path) {
     }
 }
 
+fn run_path_test_stdout_0(test_name: &str, cleanup_paths: &[&str]) {
+    let wasm = run_build_script(file!(), test_name).unwrap();
+    let test_dir = wasm.parent().unwrap();
+    for path in cleanup_paths {
```
marxin: Generally speaking, can't we rather build and run the tests in a temporary folder so that we don't have to clean up files afterwards?

lib/wasix/tests/wasm_tests/path_tests.rs
```diff
@@ -24,24 +42,98 @@ fn test_chdir_getcwd() {
 }
 
 #[test]
-fn test_create_move_open() {
-    let wasm = run_build_script(file!(), "create-move-open").unwrap();
+fn test_closing_pre_opened_dirs() {
+    run_path_test_stdout_0("closing-pre-opened-dirs", &[]);
```
marxin: What prevents us migrating that to the newly introduced `wasm_test` macro?

marxin (changes requested):

lib/wasix/tests/wasm_tests/path_tests.rs
Arshia001: the fact that the old tests have C code that prints a "0" to stdout and we check for it... which is weird, granted, but also changing the sources is more work, which I decided against since my goal of "move stuff to proper rust-based tests" is already achieved. Can do the extra work if you feel it's necessary, @marxin

lib/wasix/tests/wasm_tests/libc_tests/msync-end-of-file/main.c
Arshia001: They're not, the mmap/msync tests don't really work, so it doesn't really matter what we put in them... once we fix msync, we can go back to these and make sure they're in the right shape. I'm keeping them around in the interest of not losing the C sources completely.

lib/wasix/tests/wasm_tests/context_switching.rs
Arshia001: we _can_, but that would require argv support in wasm_test!, which is not a bad idea anyway. adding it in.

lib/wasix/tests/wasm_tests/mod.rs
Arshia001: no, because:

1. the tests do not automatically install a subscriber, so running with `RUST_LOG` will have no effect; the usual output comes from the logging setup in the wasmer CLI, which the tests do not use
2. some of the tests actually need to inspect tracing output to ensure things happened in the correct order.

lib/wasix/tests/wasm_tests/path_tests.rs
Arshia001: We can, this is just a direct port of the existing code. will do this.

lib/wasix/tests/wasm_tests/libc_tests/msync-end-of-file/main.c
theduke: Can we add a comment here than that this test is basically broken?

theduke (approved):
Can you write a wasm_tests/README.md that explains how tests are structured, how to add tests and how to run them?

lib/wasix/tests/wasm_tests/path_tests.rs
marxin: Well, given the support for the optional macro arguments, can't we just use: `wasm_test!(test_x, "x", stdout = "0");`

.github/workflows/test.yaml
```diff
@@ -793,6 +743,9 @@ jobs:
           target = "${{ matrix.metadata.target }}"
           EOF
         if: matrix.metadata.target
+      - name: Install wasixcc
+        uses: wasix-org/wasixcc@v0.4.3
+        if: matrix.stage.install_wasixcc == true && (matrix.metadata.build == 'linux-x64' || matrix.metadata.build == 'linux-arm64' || matrix.metadata.build == 'linux-musl')
```
marxin: Don't like how we combine a matrix property with a filter, what about directly doing:
`matrix.description == "CLI integ. tests" && startsWith(matrix.metadata.build, 'linux')`

Arshia001: Done. I try not to rely on string comparisons too much, but happy to do this since you're the runtime person!

lib/wasix/tests/wasm_tests/libc_tests/msync-end-of-file/main.c
Arshia001: the #[ignore] comment already clearly mentions this:

```
test wasm_tests::libc_tests::test_msync_end_of_file ... ignored, file-backed mmap writeback does not currently persist under WasiRunner
```

lib/wasix/tests/wasm_tests/path_tests.rs
Arshia001: true. fixed.

marxin (approved):
Thanks for addressing my comments, it's much better now ;)

PR #6554, opened 2026-05-04
wasm bindgen update to the latest

maksimryndin:
Hey guys!

thanks for the fantastic work!

The pr updates the trap conversion from the raw pointer for the latest `wasm-bindgen` as there is now a transparent wrapper types over `u32` (https://github.com/wasm-bindgen/wasm-bindgen/blob/0.2.120/src/rt/mod.rs#L366.

(the initial approach was a transmute to avoid importing the hidden public type of `WasmPtr` but clippy requires an explicit type, so probably it is actually future proof to import to ensure if something changes in future upgrades, it explicitly stops compiling)

Would it be possible to roll this change to the version 7.1.* as well?


marxin:
Thank you for the your PR.

> Would it be possible to roll this change to the version 7.1.* as well?

Sorry, but we're not planning any 7.1 minor release as of now.

marxin (approved):

Issue #6412, opened 2026-04-08
lock.acquire(timeout=...) never times out under Wasmer


artemyarulin:
Timed lock acquisition does not time out under Wasmer.

```python
import _thread
import threading
import time

lock = _thread.allocate_lock()
lock.acquire()

def worker():
    start = time.monotonic()
    result = lock.acquire(timeout=0.5)
    elapsed = time.monotonic() - start
    print(result, elapsed)

threading.Thread(target=worker).start()
time.sleep(2)
print("done")
```

Native:

```bash
$ python3.13 repro_lock_timeout.py
False 0.50...
done
```

Wasmer:

```bash
./target/release/wasmer run --net python/python --volume $PWD:/src -- /src/repro_lock_timeout.py
done
```

The worker thread never returns from `lock.acquire(timeout=0.5)`. Related: `threading.Condition.wait(timeout=0.1)` also hangs under Wasmer, so this likely affects more than one synchronization primitive.

theduke:
The issue is in the wasix libc.

pthread_mutex_timedlock internally defers to some helpers that drop the timeout.
Fix needs to be done there.

PR #6558, opened 2026-05-05
test: added regression test for lock timeout - futex_wait

Fliqqr:
# Description
Added regression test for lock timeout hanging (issue: https://github.com/wasmerio/wasmer/issues/6412), see this PR for fix: https://github.com/wasix-org/wasix-libc/pull/109


marxin (changes requested):
Note the currently used test harness under `tests/wasix` will be converted soon, see #6552.
Can you migrate it based on the schema present in the PR?

Arshia001 (approved):
LGTM. Should remain a draft until a wasix-libc release with the fix mentioned is published.

.github/workflows/copilot-setup-steps.yml
```diff
@@ -71,7 +71,6 @@ jobs:
         uses: wasix-org/wasixcc@v0.4.3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
-          sysroot-tag: v2026-02-16.1
```
marxin, line 74: Why this removal, shouldn't we stick to a new release instead?

lib/wasix/tests/wasm_tests/pthread_tests/mutex-lock-timeout/main.c
```diff
@@ -0,0 +1,61 @@
...
+  (void)arg;
+
+  struct timespec ts;
+  if (clock_gettime(CLOCK_REALTIME, &ts) != 0) {
+    perror("clock_gettime");
+    return (void*)(intptr_t)errno;
+  }
+
+  ts.tv_nsec += 500 * 1000 * 1000;  // 0.5 seconds
+  if (ts.tv_nsec >= 1000000000) {
```
marxin: What about using `'` as the thousand separator: `1'000'000'000`. And maybe we can define a constant given the fact we're using the const twice.

.github/workflows/copilot-setup-steps.yml
Fliqqr: Wasixcc pulls the latest sysroot when installing, so I think the tests should match that

marxin: Great, I am fine with that.

marxin (approved):

lib/wasix/tests/wasm_tests/pthread_tests/mutex-lock-timeout/main.c
marxin: It's only a test, leaving to you if you want to adjust it or now.

Fliqqr: Apparently using `'` as a separator is only possible in C++, made it into a constant however

lib/wasix/tests/wasm_tests/pthread_tests/mutex-lock-timeout/main.c
```diff
@@ -6,6 +6,8 @@
 #include <time.h>
 
 pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
+const int LOCK_TIMEOUT_NSEC = 500 * 1000 * 1000; // 0.5 seconds
```
marxin: What about:
```
const int LOCK_TIMEOUT_NSEC = 500'000'000; // 0.5 seconds
const int NSEC_OVERFLOW_GUARD = 1'000'000'000; // 1 second
```

lib/wasix/tests/wasm_tests/pthread_tests/mutex-lock-timeout/main.c
marxin: Good!

lib/wasix/tests/wasm_tests/pthread_tests/mutex-lock-timeout/main.c
Fliqqr: ' as a separator is not a thing in C

```
failures:

---- wasm_tests::pthread_tests::test_lock_timeout stdout ----
Build stdout: 
Build stderr: main.c:9:34: error: expected ';' after top level declarator
    9 | const int LOCK_TIMEOUT_NSEC = 500'000'000;  // 0.5 seconds
      |                                  ^
      |                                  ;
1 error generated.
```

marxin: Understood - we can theoretically change the file to `.cpp` and then use it, but not worth doing that.

PR #6565, opened 2026-05-06
CI: simplify and unify job names

marxin:
The PR simplifies the CI jobs in the following manner:
- extra `lib/config` checks dropped (part of `All tests`)
- `Packages` renamed to `All tests` and `WAST` got dropped as they are effectively part of `All tests`
- the `cranelift/singlepass` `std` tests used to have a confusing name - rename to `only-std` and unified under one job run
- the `stage-X` Makefile targets dropped - don't see a reason for that

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -359,7 +359,7 @@ jobs:
       matrix:
         build-what:
           - key: wast-tests
-            build-cmd: "make test-stage-0-wast"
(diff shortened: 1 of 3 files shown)


theduke (approved):
Did a double-check to make sure no tests are dropped.

Looks good.

marxin:
Great, @theduke can you adjust the mandatory CI jobs so that the PR can be merged?

PR #6575, opened 2026-05-07
Revert "ci: disable flaky tests"

### LINKED ISSUE #6538: flaky `wasm_tests` WASIX tests
I noticed that some tests in CI fail intermittently, so I tried running them locally over an extended period:

```
$ while true ; do cargo nextest run --nff 2>&1 | rg '\.\.\. FAILED' ; done
    test wasm_tests::exit_tests::test_exit_nonzero_in_thread ... FAILED
    test wasm_tests::exit_tests::test_exit_nonzero_in_closure_call_thread ... FAILED
    test wasm_tests::exit_tests::test_abort_in_thread ... FAILED
    test wasm_tests::exit_tests::test_exit_nonzero_in_call_dynamic_thread ... FAILED
```

@Arshia001 can you take a look?

Arshia001:
I believe the issue was already resolved when we merged #6552, as that introduced a lock that prevents these tests running in parallel, which was the cause of the flakiness. This also resolves #6538.

### DIFF
--- a/lib/wasix/tests/wasm_tests/exit_tests.rs
+++ b/lib/wasix/tests/wasm_tests/exit_tests.rs
@@ -2,7 +2,6 @@ wasm_test!(test_exit_zero, "exit-zero");
 wasm_test!(test_exit_nonzero, "exit-nonzero", should_fail);
 wasm_test!(test_exit_zero_in_thread, "exit-zero-in-thread");
 wasm_test!(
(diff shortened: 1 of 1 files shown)


marxin (changes requested):
No no, it does not help - still seeing the same failure (and some more -> will include them in the issue).

PR #6577, opened 2026-05-08
fix(wasix): add missing return

kilyanni:
# Description

`BinFactory::get_executable` was missing a `return` keyword, making an early return silently do nothing instead.

- added `return`
- added `#![deny(unused_must_use)]` to hopefully make the linter catch this in the future
- added `let _` discards for intended violations of `unused_must_use`

### DIFF
--- a/lib/wasix/src/bin_factory/mod.rs
+++ b/lib/wasix/src/bin_factory/mod.rs
@@ -170,7 +170,7 @@ impl BinFactory {
         {
             let cache = self.local.read().unwrap();
             if let Some(data) = cache.get(&name) {
-                data.clone().map(Executable::BinaryPackage);
+                return data.clone().map(Executable::BinaryPackage);
             }
         }
 


lib/wasix/src/bin_factory/mod.rs
```diff
@@ -1,4 +1,5 @@
 #![allow(clippy::result_large_err)]
+#![deny(unused_must_use)]
```
marxin: I don’t think this is a systematic approach. If we really want a type whose return value must be used, I’d wrap it in something like `WrapperType(...)` and then declare its methods with `#[must_use]`.

marxin (changes requested):

lib/wasix/src/bin_factory/mod.rs
kilyanni: I'm not entirely sure what you mean; this turns warnings about not using a `must_use` marked value into errors, it doesn't mark any types `must_use`.

marxin: Ok. So generally speaking, we don't promote many compiler warnings to errors, but one regular basis, I run `cargo b --all-target --all-features` and address the warnings. In case of the file, the `![deny(unused_must_use)]` won't help here as `HashMap::insert` is not marked with `#[must_use]`.

kilyanni: Ah, that makes sense. I misremembered `Option::map` as being `#[must_use]`, which is what prompted me to reach for the lint in the first place. Reverted the linter setting

Should the `let _ =`s stay, or would you prefer those be dropped as well?

edit: I swear there was some lint that was supposed to catch stuff like this, but even with `clippy::pedantic` it's green.
edit 2: https://github.com/rust-lang/rust/issues/71368

marxin: > Should the let _ =s stay, or would you prefer those be dropped as well?

I would just drop them.

> edit 2: https://github.com/rust-lang/rust/issues/71368

Cool you've found the upstream issue.

marxin:
@Arshia001 a review reminder

Arshia001 (approved):
LGTM

marxin (approved):

PR #6578, opened 2026-05-08
fix(wasix): Fix behaviour of eventfd polling to match POSIX

kilyanni:
# Description

`NotificationInner::poll()` gated wakeups with an edge-triggered check (`last_poll != counter`), which had two bugs:

- **Spurious `Ready(0)` on first poll.** A new eventfd starts with `counter=0` and `last_poll=u64::MAX`, so the check fires immediately and the caller sees `POLLIN|POLLHUP`. A subsequent `read()` then blocks on a zero counter and hangs.
- **Missed wakeups.** After a write, the first `poll()` fires and sets `last_poll = counter`. If a second `poll()` runs before any `read()`, it returns `Pending`. An event loop that polls twice per iteration without reading silently drops the event.

Per [eventfd(2)](https://man7.org/linux/man-pages/man2/eventfd.2.html#:~:text=%E2%80%A2%20The%20file%20descriptor%20is%20readable%20%28the%20select%282%29%20readfds%20argument%3B%20the%20poll%282%29%20POLLIN%20flag%29%20if%20the%20counter%20has%20a%20value%20greater%20than%200%2E), the fd is readable iff the counter is greater than 0.

## Changes
- `notification.rs`: drop `last_poll` and check `counter > 0` directly.
- `fd-event-readable`: C test covering both failure modes above. Run with:
  ```
  cargo test -p wasmer-wasix --features sys wasm_tests::fd_tests::test_fd_event_readable
  ```

### DIFF
--- a/lib/wasix/src/fs/notification.rs
+++ b/lib/wasix/src/fs/notification.rs
@@ -12,8 +12,6 @@ struct NotificationState {
     /// Used for event notifications by the user application or operating system
     /// (positive number means there are events waiting to be processed)
     counter: u64,
-    /// Counter used to prevent duplicate notification events
-    last_poll: u64,
     /// Flag that indicates if this is operating
     is_semaphore: bool,
     /// All the registered wakers
@@ -32,7 +30,6 @@ impl NotificationState {
     }
 
     fn wake_all(&mut self) {
-        self.last_poll = u64::MAX;
         while let Some(waker) = self.wakers.pop_front() {
             waker.wake();
         }
@@ -75,7 +72,6 @@ impl NotificationInner {
         Self {
             state: Mutex::new(NotificationState {
                 counter: initial_val,
-                last_poll: u64::MAX,
                 is_semaphore,
                 wakers: Default::default(),
                 interest_handler: None,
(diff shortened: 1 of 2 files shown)


kilyanni:
Copilots low confidence comment sounds plausible, though I'm pretty sure it's not caused or affected by this change / pre-existing. Probably worth investigating seperately

Arshia001 (changes requested):
This change fixes non-ET mode, but breaks ET mode. AFAIK, almost all the polling code in WASIX right now assumes ET mode by default (which is a mistake, but that is the current state...) if we want to start being proper about this, rather than change the existing behavior, which can have real effects on other software enabling ET mode (ignored today by WASIX) and relying on it to only receive change notifications, I'd implement proper switching based on ET mode. That's a bit more work, but ensures we don't break existing software unexpectedly.

kilyanni:
Ah, that's good to know. Though I'm unsure how exactly it pipes into `NotificationInner::poll`.
Can you clarify how ET mode would be set/unset in the case of this function? Do you have a specific scenario in mind where this would regress?

My understanding is that `epoll` with `EPOLLET` is unaffected here, since that path is driven by `wake_all` / `push_interest` (?), which are called the same as before. If there's another polling path in WASIX that implicitly relies on ET semantics today and would be broken by this change, could you give me a pointer so I can add a regression test and correct or verify this diff?

kilyanni:
(rebased to fix merge conflicts)

Arshia001 (approved):

PR #6579, opened 2026-05-09
fix(wasix): handle pipe-backed stdio in flush, fd_sync, and poll

kilyanni:
# Description

After `dup2(pipe_write_end, STDOUT_FILENO)`, any `fsync` or `fdatasync` on stdout would fail.

The implementations all special-cased stdio, which (as far as I can tell) originated as a workaround for a bug that no longer exists, thus making the special case redundant. This special handling only handled files, so non-file stdio would always fail with `FsError::NotAFile`.

  Changes:

  - `WasiFs::flush()`: remove stdout/stderr special case; unify all fds through get_fd + kind-match; add pipe/socket as a no-op
  - `fd_sync`: pipes/sockets return `Errno::Success` instead of `Errno::Inval` since syncing them is a no-op
  - `poll_fd_guard`: remove stdout/stderr special case; `InodeValFilePollGuard::new` already handles `Kind::PipeTx`/`PipeRx`/`DuplexPipe`, it was just never reached for stdio
  - `FlushPoller`: deduplicate from three identical local definitions into one `pub(crate) struct` in `fs/mod.rs`.
  - Add regression test: dup2 write-end of pipe onto stdout, then fsync and assert success.

### DIFF
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -60,6 +60,21 @@ use crate::{ALL_RIGHTS, bin_factory::BinaryPackage, state::PreopenedDir};
 // use a Linux-like 64k ceiling until WASIX models per-process fd limits.
 pub(crate) const MAX_FD: WasiFd = (64 * 1024) - 1;
 
+pub(crate) struct FlushPoller {
+    pub(crate) file: Arc<RwLock<Box<dyn VirtualFile + Send + Sync>>>,
+}
+
+impl Future for FlushPoller {
+    type Output = Result<(), Errno>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut file = self.file.write().unwrap();
+        Pin::new(file.as_mut())
+            .poll_flush(cx)
+            .map_err(|_| Errno::Io)
+    }
+}
+
 /// the fd value of the virtual root
 ///
 /// Used for interacting with the file system when it has no
@@ -1698,19 +1713,6 @@ impl WasiFs {
                     }
                 };
                 drop(fd);
-
-                struct FlushPoller {
-                    file: Arc<RwLock<Box<dyn VirtualFile + Send + Sync>>>,
-                }
-                impl Future for FlushPoller {
-                    type Output = Result<(), Errno>;
-                    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-                        let mut file = self.file.write().unwrap();
-                        Pin::new(file.as_mut())
-                            .poll_flush(cx)
-                            .map_err(|_| Errno::Io)
-                    }
-                }
                 FlushPoller { file }.await?;
             }
         }
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -331,19 +331,11 @@ impl WasiInodes {
         }
     }
 
-    /// Get the `VirtualFile` object at stdout
-    pub(crate) fn stdout(fd_map: &RwLock<FdList>) -> Result<InodeValFileReadGuard, FsError> {
-        Self::std_dev_get(fd_map, __WASI_STDOUT_FILENO)
-    }
     /// Get the `VirtualFile` object at stdout mutably
     pub(crate) fn stdout_mut(fd_map: &RwLock<FdList>) -> Result<InodeValFileWriteGuard, FsError> {
         Self::std_dev_get_mut(fd_map, __WASI_STDOUT_FILENO)
     }
 
-    /// Get the `VirtualFile` object at stderr
-    pub(crate) fn stderr(fd_map: &RwLock<FdList>) -> Result<InodeValFileReadGuard, FsError> {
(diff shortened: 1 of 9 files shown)


lib/wasix/src/syscalls/wasi/fd_sync.rs
```diff
@@ -54,14 +54,15 @@ pub fn fd_sync(mut ctx: FunctionEnvMut<'_, WasiEnv>, fd: WasiFd) -> Result<Errno
...
+            Kind::Buffer { .. } | Kind::Symlink { .. } => return Ok(Errno::Inval),
+            // Writes to pipes and sockets go directly into the underlying channel;
+            // there is no intermediate write buffer to drain, so flush is a no-op.
+            Kind::Socket { .. }
             | Kind::PipeTx { .. }
             | Kind::PipeRx { .. }
             | Kind::DuplexPipe { .. }
             | Kind::EventNotifications { .. }
-            | Kind::Epoll { .. } => return Ok(Errno::Inval),
+            | Kind::Epoll { .. } => return Ok(Errno::Success),
```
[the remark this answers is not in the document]
kilyanni: @copilot apply changes based on [this feedback](https://github.com/wasmerio/wasmer/pull/6579#discussion_r3212127979)

lib/wasix/src/syscalls/wasi/poll_oneoff.rs
```diff
@@ -175,40 +175,22 @@ pub(crate) fn poll_fd_guard(
...
-            {
-                return Err(Errno::Access);
-            }
-            let inode = fd_entry.inode;
+    let fd_entry = state.fs.get_fd(fd)?;
+    let requires_access = match s.type_ {
+        Eventtype::FdRead => Rights::FD_READ,
+        Eventtype::FdWrite => Rights::FD_WRITE,
+        _ => Rights::empty(),
+    };
```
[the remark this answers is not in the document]
kilyanni: @copilot apply changes based on [this feedback](https://github.com/wasmerio/wasmer/pull/6579#discussion_r3212127941)

lib/wasix/src/fs/mod.rs
```diff
@@ -1655,56 +1660,37 @@ impl WasiFs {
...
-                let mut file =
-                    WasiInodes::stderr_mut(&self.fd_map).map_err(fs_error_into_wasi_err)?;
-                file.flush().await.map_err(map_io_err)?
-            }
-            _ => {
-                let fd = self.get_fd(fd)?;
-                if !fd.inner.rights.contains(Rights::FD_DATASYNC) {
-                    return Err(Errno::Access);
-                }
+        if fd == __WASI_STDIN_FILENO {
```
Arshia001: why are we still checking against the FD number here? any program may renumber its FDs so that the stdin FD does not reflect an actual input-only stream of characters.

Arshia001 (changes requested):
Since this PR touches a lot of functionality, we need more tests. Please add separate tests addressing each of the scenarios that changed.

lib/wasix/src/fs/mod.rs
kilyanni: Ah good catch
I did that because it matches previous behavior and flushing/syncing generally a no-op if it is stdin (I think Linux might actually give `EINVAL` instead of success, though I'm not sure)

But you're right that that isn't actually correct in all cases. Gonna add some tests for stdin and drop the special handling

lib/wasix/src/fs/mod.rs
```diff
@@ -318,19 +331,11 @@ impl WasiInodes {
...
-    }
     /// Get the `VirtualFile` object at stdout mutably
     pub(crate) fn stdout_mut(fd_map: &RwLock<FdList>) -> Result<InodeValFileWriteGuard, FsError> {
         Self::std_dev_get_mut(fd_map, __WASI_STDOUT_FILENO)
     }
 
-    /// Get the `VirtualFile` object at stderr
-    pub(crate) fn stderr(fd_map: &RwLock<FdList>) -> Result<InodeValFileReadGuard, FsError> {
-        Self::std_dev_get(fd_map, __WASI_STDERR_FILENO)
-    }
```
kilyanni, line 333: I removed these two because they're unused after this change, but they might be worth keeping around.
LMK if you'd prefer them to be `#[allow(dead_code)]`

lib/wasix/src/fs/mod.rs
kilyanni: Dropped the stdin special handling

kilyanni:
Upon closer investigation, this PR fixed a real bug but incorrectly. 
The core issue was/is that `close(STDOUT_FILENO)` would return EINVAL after replacing stdout with a pipe.
However in POSIX, close never returns EINVAL.

Wasix was propagating the EINVAL from flushing, and the previous version of this PR changed the return values of flushing.
I now updated it to match POSIX better; flushing now returns EINVAL for all relevant cases (including some where it previously returned EIO) and `close` swallows EINVAL now.

All stdio special handling in `flush` is now removed.
There are now three tests for the relevant changes in behavior.

kilyanni:
(rebased to fix merge conflicts)

Arshia001 (approved):

kilyanni:
Thanks ^^

PR #6580, opened 2026-05-09
docs(wasix): add warning about CC env var

kilyanni:
# Description

The wasix wasm test suite uses the environments `$CC` variable if set, which can produce unexpected results.

I had `CC=gcc` set in my environment, and it took me quite a while to figure out that the test suite isn't in fact 50% failing and that I was in fact compiling it with `gcc` instead of `wasixcc` 😅

To save others from the same confusion, this PR adds a small note to the README calling out this behavior.


syrusakbary (approved):

kilyanni:
Thanks!

PR #6583, opened 2026-05-11
fix(wasix): Fix race conditions in {fd,path}_filestat_get

### LINKED ISSUE #6582: WASIX: filestat_get race conditions
the fd/path_filestat_get syscalls have race conditions that can lead to invalid FD acccess and wrong BADF errors.

theduke:
Previously the fd was re-acccessed at a later point in the syscall,
leading to potential race conditions.

This resulted in real bugs in applications with parallelism.

Closes #6582

### DIFF
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -1497,6 +1497,16 @@ impl WasiFs {
         self.get_inode_at_path_inner(inodes, base_inode, path, 0, follow_symlinks)
     }
 
+    pub(crate) fn get_inode_at_path_from_inode(
+        &self,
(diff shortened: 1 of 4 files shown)


Arshia001 (changes requested):
Good find! Can we add regression tests?

theduke:
Requesting a review from @Fliqqr so he can see the kind of problems we have to deal with in WASIX...

theduke:
> Good find! Can we add regression tests?

Not really easily, since this is a race condition.
We could have a stress test with like 5 threads trying to prestat the same file over and over, but I don't think we want those kinds of tests in the main suite.

theduke:
@Arshia001 is this good to go?

Issue #6574, opened 2026-05-07
wasix tests: support multiple run-time engines, consider libtest-mimic


marxin:
The issue is a follow-up to #6552, where we consolidated the WASIX tests that were previously implemented as shell scripts.

With the upcoming integration of V8, we want to run the full WASIX test suite against LLVM, Cranelift (where supported), and V8. We could also consider using `libtest-mimic`, which can generate tests dynamically from a filesystem layout. For an example of this approach, see the Wild linker integration tests: https://github.com/wild-linker/wild/blob/main/wild/tests/integration_tests.rs. With the custom directives (present directly in the tested source files), we can check for exit code, expected stdout, etc.

Arshia001:
My 2 cents on why `libtest-mimic` would be an improvement over adapting the `wasm_test!` macro to run each test multiple times:
* we can clearly separate the wasm code + engine into a separate test case, such as `wasix_wasm::filesystem::fd_read::cranelift`. With the macro, all engines will be tested together in the same test.
* We get to put the test's C code and expected results together, which will result in better readability and maintainability (for humans at least!)

sonukapoor:
I opened #6584 as a small first slice for this: a separate `libtest-mimic` target that reports `basic_tests::helloworld` separately for LLVM and Cranelift while leaving the existing `wasix-wasm` suite behavior unchanged.

That PR is green in CI now. If this direction looks acceptable, the next useful slice seems to be either:
- adding metadata-driven trials for a small stable fixture group, or
- adding explicit per-engine skip/unsupported handling before expanding coverage.

I can follow whichever direction is more useful for the V8/multi-engine rollout.

marxin:
One can annotate the source files with directives like:
```
//#Args: foo bar
//#ExpectedStdout: 0
//#ExpectedExitCode: 42
//#UseTemp: true
//#MapDirectory:/tmp:$tmp
```

marxin:
I'm going to work on the `libmimic-test` integration.

PR #6584, opened 2026-05-11
test(wasix): generate helloworld cases per engine

sonukapoor:
This is a small first slice for #6574. It keeps the existing `wasix-wasm` test target and adds an opt-in `engines = [...]` form to the existing `wasm_test!` macro so a fixture can generate one Rust test per runtime engine.

The initial opt-in fixture is `basic_tests::test_helloworld`, which now reports separate LLVM and Cranelift cases under the existing harness. Other `wasm_test!` cases continue through the current default engine path until they are explicitly migrated.

Based on review, this no longer introduces `libtest-mimic`; that can be evaluated separately if maintainers want a full harness replacement later.

Verified locally with:

```sh
. ~/.wasixcc/env
unset CC CXX
export LLVM_SYS_221_PREFIX="$PWD/.llvm"

cargo fmt --check
cargo test -p wasmer-wasix --features sys --test wasix-wasm test_helloworld -- --nocapture
cargo test -p wasmer-wasix --features sys --test wasix-wasm process_tests -- --nocapture
git diff --check
```


lib/wasix/tests/wasix-wasm-engines.rs
```diff
@@ -0,0 +1,50 @@
+#![allow(dead_code, unused_imports)]
+
```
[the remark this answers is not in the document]
sonukapoor: Updated in 1caab3a. The lint allow is now scoped to the imported `wasm_tests` module, which keeps this harness file lint-visible while avoiding unused warnings from the existing test modules included by `mod.rs`.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -575,26 +575,77 @@ fn get_cache_dir() -> PathBuf {
...
+    });
+
+    let compiler = wasmer::sys::LLVM::default();
+
+    EngineBuilder::new(compiler)
+        .set_features(Some(features))
+        .set_target(Some(target))
+        .engine()
+        .into()
+}
```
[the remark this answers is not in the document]
sonukapoor: Updated in 1caab3a. The shared feature-detection and `EngineBuilder` setup now lives in `create_compiler_engine_for_wasm`, with LLVM and Cranelift wrappers supplying the backend kind and compiler.

lib/wasix/tests/wasix-wasm-engines.rs
```diff
@@ -0,0 +1,49 @@
+#[cfg(all(unix, not(feature = "js")))]
```
marxin: I would like to see `libmimic-test` introduction as a separate change with the ability to replace the existing `wasm_test` macro.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -575,26 +575,80 @@ fn get_cache_dir() -> PathBuf {
...
+
+fn create_cranelift_engine_for_wasm(wasm_bytes: &[u8]) -> wasmer::Engine {
+    create_compiler_engine_for_wasm(
+        wasm_bytes,
+        wasmer::BackendKind::Cranelift,
+        wasmer::sys::Cranelift::default(),
+    )
+}
+
+fn create_engine_for_wasm_with_engine(
```
marxin, line 765: Am I right we'll still run the tests only with a single engine?

marxin (changes requested):

lib/wasix/tests/wasm_tests/mod.rs
marxin: Note the inspiration can be taken from https://github.com/wasmerio/wasmer/blob/main/tests/lib/compiler-test-derive/src/lib.rs, where we generate multiple test runs for different run-times.

sonukapoor: Updated in 4a6ab3f. `basic_tests::test_helloworld` now runs as two separate tests under the existing `wasix-wasm` target: `wasm_tests::basic_tests::test_helloworld::llvm` and `wasm_tests::basic_tests::test_helloworld::cranelift`. Other fixtures still use the current default engine path until explicitly opted in.

sonukapoor: I followed that direction in 4a6ab3f by generating per-engine Rust tests from the macro opt-in, similar to the compiler-test runtime split, rather than adding a separate harness target in this PR.

lib/wasix/tests/wasix-wasm-engines.rs
sonukapoor: Updated in 4a6ab3f. This PR no longer introduces `libtest-mimic`; it keeps that as a separate possible follow-up and instead adds an opt-in `engines = [...]` form to the existing `wasm_test!` macro, which gives a direct replacement path for current macro users.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -43,6 +46,94 @@ macro_rules! wasm_test {
             runner.with_args([$($arg),+]);
         })
     };
+    (@run_engine $engine:ident, $wasm:ident, $run_dir:expr, []) => {
```
marxin, line 49: Looking at the explosive number of combinations, to be honest, the declarative macro does not scale well.
`libtest-mimic` would be much more systematic approach for the future.

marxin (changes requested):
Rejecting. @Arshia001 what's your take on it?

PR #6587, opened 2026-05-12
Fix/wasix spawn store on final thread

### LINKED ISSUE #6546: v8: make sure `Store` is not moved between threads
Based on the limitations of the C API for wee8, we cannot move a created Store in between threads, otherwise we'll get a weird run-time error at the place where a host function is invoked.

`"Fatal error (v8): current thread is different from the thread the store was created in!"` - based on my testing experience, the constraint is real and we must switch the logic in https://github.com/wasmerio/wasmer/blob/f7241ae5241230788f225bae2e6fa223e71cd91f/lib/wasix/src/runtime/task_manager/tokio.rs#L142. Ideally, the `Store` of `v8` should be marked as `!Send` and we won't need the `check_isolate` logic any longer, Rust rich type system will guide us.

Arshia001:
(Eventually) resolves #6546. Implementation plan:

* [x] Make `Store` `!Send`, as it was meant to be
* [x] Split `Memory::share_in_store` and `Memory::copy_in_store` into two separate functions that can be called from the thread with access to the proper store; e.g., thread 1 calls `Memory::share_and_detach`, sends the resulting `DetachedMemory` to thread 2, thread 2 calls `DetachedMemory::attach` with its own store. This avoids the need to have both stores available on the same thread.
* [x] Refactor `TokioTaskManager` and surrounding code to spawn new tasks' stores on the final thread
* [x] Same as the point above, but for tasks with a trigger
* [x] Refactor the linker to use a token-based locking mechanism with `Send`able tokens for updates to its topology
* [x] Refactor `Linker::create_instance_group` so it can be split into two functions, one called from the parent thread and one called from the new thread
* [x] Bonus: refactor the linker into smaller modules and encapsulate the locking/syncing mechanisms because it was near-impossible to add any more code to the already 4.5k-LoC file.

As a second bonus, this PR also removed dcgi, wcgi, and dproxy since the assumptions made while implementing the changes to `TokioTaskManager` are not compatible with those runners, and they were unused anyway.

Note to reviewers: the linker changes are mostly mechanical. The thing that needs to be looked at in detail is the sync module, which has received all the new development work. Everything exists as it did before, just in a new file.


lib/api/src/backend/sys/entities/memory/mod.rs
```diff
@@ -119,7 +119,9 @@ impl Memory {
         &self,
         store: &impl AsStoreRef,
     ) -> Result<Box<dyn LinearMemory + 'static>, MemoryError> {
-        let mut mem = self.try_clone(store)?;
+        let store = store.as_store_ref();
+        let objects = store.objects();
+        let mem = self.handle.get(objects.as_sys());
```
syrusakbary: Why this? I'm not sure is needed

lib/api/src/entities/memory/inner.rs
```diff
@@ -170,39 +170,20 @@ impl BackendMemory {
...
-    /// Attempts to duplicate this memory (if its clonable) in a new store
-    /// (copied memory)
+    /// Attempts to duplicate this memory in a new store with a byte-for-byte copy
     #[inline]
+    #[deprecated(
+        since = "8.0.0",
+        note = "Since `Store` is no longer `Send + Sync`, this method cannot be used meaningfully. \
+                Use `copy_and_detach`, then `attach` on the thread owning the other `Store` instead."
+    )]
     pub fn copy_to_store(
```
syrusakbary, line 180: Does this work in the browser? (aka. wasmer-js)

lib/cli/src/commands/run/mod.rs
```diff
@@ -429,13 +424,7 @@ impl Run {
 
         let uses = self.load_injected_packages(&runtime)?;
 
-        if DcgiRunner::can_run_command(cmd.metadata())? {
-            self.run_dcgi(id, pkg, uses, runtime)
-        } else if DProxyRunner::can_run_command(cmd.metadata())? {
-            self.run_dproxy(id, pkg, runtime)
-        } else if WcgiRunner::can_run_command(cmd.metadata())? {
-            self.run_wcgi(id, pkg, uses, runtime)
-        } else if WasiRunner::can_run_command(cmd.metadata())? {
```
syrusakbary, line 438: Nice

lib/api/src/entities/memory/inner.rs
```diff
@@ -249,37 +230,53 @@ impl BackendMemory {
...
+    /// Attempts to create a detached shared memory handle that can
+    /// later be attached to a different store. This operation is only supported
+    /// for shared memory instances.
     #[inline]
-    pub fn share_in_store(
-        &self,
-        store: &impl AsStoreRef,
-        new_store: &mut impl AsStoreMut,
-    ) -> Result<Self, MemoryError> {
+    pub fn share_and_detach(&self, store: &impl AsStoreRef) -> Result<VMMemory, MemoryError> {
```
syrusakbary: If the memory is shared... what do you mean with share and detach?

A memory is a memory, and if is shared, we can simply clone it into another store.

I think the name `share_into` is clearer (or `clone_into_store`)

lib/api/src/entities/memory/inner.rs
```diff
@@ -249,37 +230,53 @@ impl BackendMemory {
...
             });
         }
 
+        self.try_clone(store)
+    }
+
+    /// Attempts to create a detached copied memory handle that can later be
+    /// attached to a different store.
+    #[inline]
+    pub fn copy_and_detach(&self, store: &impl AsStoreRef) -> Result<VMMemory, MemoryError> {
```
syrusakbary: I think the name `copy_into` is clearer

lib/api/src/entities/memory/mod.rs
```diff
@@ -15,6 +15,58 @@ pub(crate) mod view;
 pub(crate) use inner::*;
 pub use view::*;
 
+/// A memory handle that has been cloned from an existing shared memory but
+/// has not yet been attached to a store.
+#[derive(Debug)]
+pub struct DetachedMemory {
+    kind: DetachedMemoryKind,
+    memory: VMMemory,
+}
```
syrusakbary: We should not have this type. We should only clone into a store directly. No intermediary types

lib/wasix/src/runtime/task_manager/mod.rs
```diff
@@ -209,26 +206,7 @@ pub trait VirtualTaskManager: std::fmt::Debug + Send + Sync + 'static {
                 })?;
                 Ok(Some(mem))
             }
-            SpawnType::ShareMemory(mem, old_store) => {
```
syrusakbary, line 212: I believe v8 was fine on accessing memory from one store and copying into another, right? (at least, browsers do this all time)

lib/api/src/entities/memory/inner.rs
marxin: I think at some point we should also make a serious investigation if `wasmer-js` works. My impression is it's gonna be similarly broken to V8.

lib/api/src/entities/memory/mod.rs
marxin: Why? Note it's probably a key component, where we need to mindfully move a memory to the other thread (without having the newly created store yet). I think @Arshia001 will explain better here.

lib/api/src/entities/memory/mod.rs
```diff
@@ -15,6 +15,58 @@ pub(crate) mod view;
...
+    }
+
+    /// Returns true if this detached memory will be attached as a copied
+    /// memory.
+    pub fn is_copied(&self) -> bool {
+        self.kind == DetachedMemoryKind::Copied
+    }
+
+    /// Attaches this detached memory to the provided store.
+    pub fn attach(self, new_store: &mut impl AsStoreMut) -> Memory {
```
marxin: Neat API, like it.

lib/cli/Cargo.toml
```diff
@@ -110,9 +110,11 @@ tokio-subscriber = ["dep:console-subscriber", "tokio/tracing"]
 [dependencies]
 # Repo-local dependencies.
 
-wasmer = { version = "=7.2.0-alpha.2", path = "../api", default-features = false, features = ["headless"] }
```
marxin: What linter are you using for formatting of the TOML files? We should align on that.

lib/wasix/src/runtime/task_manager/mod.rs
marxin: Yes, I have the same feeling!

marxin:
I think this is a new build warning:
```
warning: unused variable: `linker_state`
    --> lib/wasix/src/state/linker.rs:903:13
     |
 903 |         let mut $guard = loop {
     |             ^^^^^^^^^^
...
1303 |         write_linker_state!(linker_state, self, parent_group_state, parent_ctx);
     |         ----------------------------------------------------------------------- in this macro invocation
     |
help: `linker_state` is captured in macro and introduced a unused variable
    --> lib/wasix/src/state/linker.rs:903:13
     |
 903 |         let mut $guard = loop {
     |             ^^^^^^^^^^
...
1303 |         write_linker_state!(linker_state, self, parent_group_state, parent_ctx);
     |         ----------------------------------------------------------------------- in this macro invocation
     = note: `#[warn(unused_variables)]` (part of `#[warn(unused)]`) on by default
     = note: this warning originates in the macro `write_linker_state` (in Nightly builds, run with -Z macro-backtrace for more info)
```

lib/api/src/entities/memory/inner.rs
Arshia001: @marxin actually, `wasmer-js` does work, at least partially. The newer features aren't implemented properly (async execution comes to mind immediately) but you can see it in action on https://wasmer.sh.

Arshia001: @syrusakbary yes, both operations (`try_clone` and `new_from_existing`) are implemented for the JS backend as well.

lib/api/src/backend/sys/entities/memory/mod.rs
Arshia001: `try_clone` first clones a handle, then copies that. Cloning a handle is only supported for shared memories. However, copying a memory is not a shared-only operation, hence this change.

I believe the code that was already in place did it by cloning because `VMMemory::copy` required a mutable reference, which cannot be obtained from a read-only `StoreRef`. Hence, the code would create an entirely new mutable memory handle and then copy that. The mutability requirement for `VMMemory::copy` turned out to be superficial, so I changed that to take `&self` and now we can copy the memory without an additional clone. This makes the code work with non-shared memory instances as well.

lib/api/src/entities/memory/inner.rs
Arshia001: While I agree the naming is bad (suggestions welcome!) this operation only builds a detached memory handle from `self`, it still doesn't go into any stores yet. See my reply to your other comment for more details.

lib/api/src/entities/memory/mod.rs
Arshia001: The problem we're solving here is:

* We're making `Store` `!Send`, as it was always meant to be. This is especially important for the V8 backend, where the engine flat out refuses to work if a store jumps threads
* Our own sys stores aren't exactly supposed to be Send either, we were just doing `unsafe impl Send for Store {}` probably because at some point we needed it; the comment mentioned setting the trap handler, but that works with `!Send` stores today
* Once a `Store` is `!Send`, it has to remain on the thread where it was created
* This, in turn, means you can't have a function that receives two stores from different threads as input; hence, why `share_in_store(old_store, new_store)` is no longer a viable option
* To let the memory traverse the thread boundary, we need a type that's not bound to any store, and hence can be Send. This is where DetachedMemory comes into the picture. You first either "share" or copy the memory in the parent thread, then pass the DetachedMemory instance into the other thread where it can be installed into its intended destination store.

lib/api/src/entities/memory/inner.rs
Arshia001: same as above, this is not copying into the store, just creating a detached memory.

lib/api/src/entities/memory/inner.rs
marxin: > @marxin actually, wasmer-js does work, at least partially. The newer features aren't implemented properly (async execution comes to mind immediately) but you can see it in action on https://wasmer.sh/.

Ideally once we're done with V8, we should include the `wasmer-js` into the testing matrix.

lib/wasix/src/runtime/task_manager/mod.rs
Arshia001: >browsers do this all time

With browsers, it's not really as simple... a single JS worker tends to stay single-threaded, and sending things to other workers is notoriously difficult. While memory instances can be sent across workers, stores themselves are implicit in JS IIRC, so "accessing memory from one store and copying into another" in a browser doesn't map cleanly to what we're doing here.

However, while it is true that `share_in_store` (as we implemented it) did work with V8, either by luck or by design, our new design around `!Send` stores doesn't allow that function to be used meaningfully any longer. And even if we did go back on that design (which I don't believe we should), this new design in WASIX is much cleaner and I want to keep it.

Arshia001:
@marxin the build warning exists because the linker is still WIP.

lib/cli/Cargo.toml
Arshia001: The `Even Better TOML` VS Code extension comes with a default formatter. I'm all for standardising our TOML formatting, merge conflicts due to different formatters can be a big headache from time to time.

Arshia001: AI tells me Even Better TOML uses `taplo`.

marxin: Great, let's start using it: #6588. I was using the extension some time ago, but it was annoying if the upstream project didn't use it by default.

lib/wasix/src/runtime/task_manager/mod.rs
syrusakbary: In V8 you can't RUN code generated from one store in another. But I believe you can access a shared memory from any other thread.

So we may need to keep Send stores for that to happen.

> a single JS worker tends to stay single-threaded, and sending things to other workers is notoriously difficult

I disagree with this. Sharing a memory is quite trivial between workers

lib/wasix/src/state/linker/instance_group/imports.rs
```diff
@@ -0,0 +1,766 @@
...
+            }
+        }
+
+        trace!(?module_handle, "Imports object populated successfully");
+
+        Ok(())
+    }
+
+    // For when we receive a module loaded DL operation
+    pub(in crate::state::linker) fn populate_imports_from_linker(
```
marxin, line 220: Hm, the function is pretty similar to the previous one (`populate_imports_from_link_state`) - can we share the logic?

marxin (changes requested):
I noticed the number of changes in this PR has grown significantly, and I have two questions:

* Can we separate the changes related to V8’s `Store` not being `Send` from the missing dylink support?
* The PR split `lib/wasix/src/state/linker.rs` into multiple source files, but in its current form that makes the PR very difficult to review.

lib/wasix/src/state/linker/instance_group/imports.rs
Arshia001: not without considerable parameterization that would defeat the gains, but I'll take another look just in case I missed something

Arshia001:
@marxin this is now a rather huge PR... but the changes are all connected, from the `DetachedMemory` to the `!Send` stores and the fixes in wasix and the linker. For CI to pass, we need all of these things together, so splitting it up into separate PRs is not the best of ideas IMO.

lib/wasix/src/runtime/task_manager/mod.rs
Arshia001: a memory is easy, yes, but not a whole lot of other things.

lib/wasix/src/state/linker/instance_group/imports.rs
Arshia001: well, there's one refactor we can make, I'll push the change shortly.

lib/vm/src/memory.rs
```diff
@@ -452,16 +454,13 @@ impl LinearMemory for VMOwnedMemory {
...
-    mmap: Rc<RwLock<WasmMmap>>,
+    mmap: Arc<RwLock<WasmMmap>>,
     // Configuration of this memory
     config: VMMemoryConfig,
     // waiters list for this memory
     conditions: ThreadConditions,
 }
 
-unsafe impl Send for VMSharedMemory {}
-unsafe impl Sync for VMSharedMemory {}
```
syrusakbary: Why removing this? A SharedMemory should be Send and Sync

lib/wasix/src/state/linker/memory_allocator.rs
```diff
@@ -0,0 +1,197 @@
+use std::cmp::{Ord, PartialEq, PartialOrd};
+
+use tracing::trace;
+use wasmer::{AsStoreMut, Memory, MemoryError};
+
+const WASM_PAGE_SIZE: u32 = wasmer::WASM_PAGE_SIZE as u32;
+
+struct AllocatedPage {
```
syrusakbary, line 8: What this is used for?

syrusakbary: Oh, I see this was here before as well, so it must be a splitting into chunks thing

Arshia001: the constant or the whole module? the constant is WASM_PAGE_SIZE cast to u32. the whole module is the memory allocator for shared modules, which require some amount of memory for their static/TLS data and such.

lib/vm/src/memory.rs
Arshia001: it is. I moved the impls to `WasmMmap` so it's more local to the actual place where it's needed; if the impl had been there (next to where we have a raw pointer) all along, we wouldn't have used an Rc by mistake here.

Arshia001:
As agreed with @syrusakbary in chat, I'll make another PR that will offer an improved memory API, replacing the `DetachedMemory` design in this PR, which should be merged into this one. Marking as draft again in the meantime.

syrusakbary (changes requested):
Small update.

We realized that we need a better structure for the SharedMemory, as currently we need to do many hacks to make things work (DetachedMemory)

We are going to create a new PR improving the SharedMemory API, review and merge into main, and then we will rebase this PR with that, to simplify this PR implementation

marxin (approved):
Made review of all the non-linker changes and it looks good to me.

lib/wasix/src/state/linker/instance_group/imports.rs
```diff
@@ -0,0 +1,694 @@
...
+                                trace!(?requesting_module, name, "Failed to lock linker state");
+                                None
+                            }
+                            Err(TryLockError::Poisoned(_)) => {
+                                *resolved_guard = Some(None);
+                                return Err(mk_error());
+                            }
+                        };
+
+                        let group_guard = linker.instance_group_state.lock().unwrap();
```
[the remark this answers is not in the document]
Arshia001: non-issue, we only try to lock the linker state here on best-effort basis, so no deadlock can happen.

syrusakbary (approved):
Looks great

PR #6589, opened 2026-05-13
CI: disable 2 more flaky tests

marxin:
Related: #6538.


theduke (approved):
Not terribly happy about just disabling tests.

Have they been flaky forever, or just turned flaky recently?

Should get someone to triage if these indicate problems. cc @Arshia001

marxin:
> Not terribly happy about just disabling tests.

Not happy about it either, but it's a low priority right now.

> Have they been flaky forever, or just turned flaky recently?

Note for one of the tests - I disabled it 2 months ago, so probably the frequency has increased recently (note we're using 3 retries in nextest).

PR #6594, opened 2026-05-15
tests: port WASIX tests to libtest-mimic

marxin:
The PR ports the currently used `wasm_test` macro to the libtest-mimic with the custom directives strongly inspired by: https://github.com/wild-linker/wild/blob/main/wild/tests/integration_tests.rs.

If I am not mistaken all the tests were ported (except `mutex-lock-timeout`), which I will cover in the follow-up PR - AI has been used to verify all the existing tests are covered.

The goal was to keep most of the `mod.rs` untouched - the unification (and the introduction of support for other targets will come as a follow-up PR).

Issue: #6574


lib/wasix/tests/wasm_tests/context_switching/legacy_process_switching/main.c
```diff
@@ -1,3 +1,27 @@
+//#Config: test_legacy_process_switching_basic_switching
```
Arshia001, line 1: What is a config, exactly? Is it just a way to name one way to run the test, or does it refer to something that exists outside the test, maybe in the harness code?

marxin: Take a look at the beginning of `wasm_tests.rs` - I tried documenting there what is it all about (including all the supported directives).

Arshia001: I did read the documentation, but wasn't sure exactly what it is, which is why I asked. I think we may want to clarify the documentation a bit.

marxin: Addressed.

Arshia001 (approved):

Arshia001 (approved):

PR #6596, opened 2026-05-15
Add new `Memory` APIs in preparation for `!Send` stores

Arshia001:
This is mainly a prelude to #6587. See design details and discussion in WARP-60.


lib/api/src/backend/sys/entities/memory/mod.rs
```diff
@@ -177,16 +184,16 @@ impl SharedMemoryOps for ThreadConditionsHandle {
...
+            .ok_or(crate::AtomicsError::Unimplemented)?
             .disable_atomics();
         Ok(())
     }
 
-    fn wake_all_atomic_waiters(&self) -> Result<(), MemoryError> {
+    fn wake_all_atomic_waiters(&self) -> Result<(), crate::AtomicsError> {
         self.upgrade()
-            .ok_or_else(|| MemoryError::Generic("memory was dropped".to_string()))?
+            .ok_or(crate::AtomicsError::Unimplemented)?
```
syrusakbary: The unimplemented error seems weird. I’d use other name

lib/api/src/entities/memory/mod.rs
```diff
@@ -151,14 +182,18 @@ impl Memory {
         self.0.reset(store)
     }
 
-    /// Attempts to duplicate this memory (if its cloneable) in a new store
-    /// (copied memory)
+    /// Attempts to duplicate this memory in a new store with a byte-for-byte copy
+    #[deprecated(
+        since = "8.0.0",
+        note = "Since `Store` is no longer `Send + Sync`, this method cannot be used meaningfully. \
```
syrusakbary: Why copy to store can't be used meaningfully?

lib/api/src/entities/memory/mod.rs
```diff
@@ -151,14 +182,18 @@ impl Memory {
...
+                Use `copy`, then `attach` on the thread owning the other `Store` instead."
+    )]
     pub fn copy_to_store(
         &self,
         store: &impl AsStoreRef,
         new_store: &mut impl AsStoreMut,
     ) -> Result<Self, MemoryError> {
-        self.0.copy_to_store(store, new_store).map(Self)
+        self.copy(store).map(|memory| memory.attach(new_store))
     }
```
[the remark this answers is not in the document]
syrusakbary: I would expect this to perform a byte-for-byte copy even on shared memories.

lib/sys-utils/src/memory/fd_memory/memories.rs
```diff
@@ -35,6 +40,9 @@ struct WasmMmap {
     vm_memory_definition: MaybeInstanceOwned<VMMemoryDefinition>,
 }
 
+unsafe impl Send for WasmMmap {}
+unsafe impl Sync for WasmMmap {}
```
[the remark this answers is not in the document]
syrusakbary: Why we need this? It's likely we will need Send/Sync on SharedMemory. But here?

lib/api/src/entities/memory/mod.rs
```diff
@@ -9,12 +10,42 @@ use crate::{
...
+    )
+}
+
+/// A detached memory handle that is either an owned non-shared
+/// memory or a shared memory reference.
+pub enum OwnedOrSharedMemory {
+    /// A shared memory reference.
+    Shared(SharedMemory),
+    /// A detached owned memory.
+    Owned(OwnedMemory),
```
syrusakbary: The concept of `OwnedMemory` is incredibly weird for me. It should not exist.
OwnedMemory is literally a LinearMemory.

Yes, I know that OwnedMemory means *is not yet attached*. But I don't think that makes sense.

This happens because we have separated the clone operation from the attachment into other store. But we can do it in one step, so there's no Unattached Memory. Just a LinearMemory that was copied into another store (or a SharedMemory that can be on any store).

Arshia001: > separated the clone operation from the attachment into other store

Please keep in mind, this is the direct result of how store construction should work after #6587 and for V8. Essentially:

* Thread A owns its store and memory
* Thread B needs to be started, and should get its memory from thread A
* Thread A has to construct and pass a memory handle into thread B
* Between the construction of the memory handle on thread A and the construction of thread B's store, the memory has nowhere to be attached to
* The old way of calling `copy_to_store` with both stores doesn't work even if we pretend stores are sync, because thread A only has a temporary reference to its store (`StoreMut<'a>`). That cannot be passed into thread B because of the lifetime parameter.

Now, another option is to just not support copying non-shared memories. `OwnedMemory` exists purely for non-shared memories. If we think copying non-shared memories is not important and are happy restricting both sharing _and_ copying to only shared memories, this can be removed.

lib/api/src/entities/memory/mod.rs
Arshia001: see here: https://github.com/wasmerio/wasmer/pull/6596#discussion_r3257563796

lib/api/src/backend/sys/entities/memory/mod.rs
Arshia001: agreed

lib/sys-utils/src/memory/fd_memory/memories.rs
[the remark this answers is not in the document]
Arshia001: Basically, we had a long-standing bug: `VMMemory` was storing an `Rc<RwLock>`. This should not work, but it did, because we also had `unsafe impl Send` for `VMMemory`, which would circumvent the `!Send`-ness of `Rc`. The thing that's actually "unsafely Send" is the raw pointer in `WasmMmap`, not `VMMemory` itself; this is why I moved the `unsafe impl Send` here, so it's closer to the actual raw pointer, while also fixing `VMMemory` to use an `Arc<RwLock>` properly. The end result is the same, this is just more future-proof.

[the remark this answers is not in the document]
Arshia001: Note that with this, `VMMemory` is now automatically `Send + Sync` as well.

lib/api/src/entities/memory/mod.rs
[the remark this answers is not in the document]
Arshia001: Yes, it has to do a copy, I'll fix the impl.

lib/api/build.rs
```diff
@@ -45,29 +29,60 @@ fn build_v8() {
...
+        });
+        let tar = xz::read::XzDecoder::new(tar_data.as_slice());
+        let mut archive = tar::Archive::new(tar);
+        archive.unpack(&cache_dir).unwrap_or_else(|err| {
+            panic!(
+                "failed to unpack v8 archive cache {} into {}: {err}",
+                archive_path.display(),
+                cache_dir.display()
+            )
+        });
```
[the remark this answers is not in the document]
Arshia001: Moving the V8 cache thing to its own PR: #6601

lib/api/src/entities/memory/mod.rs
Arshia001: We decided to drop `OwnedMemory` outright.

lib/api/src/entities/memory/mod.rs
Arshia001: Going to remove this since the API can still be used if multiple stores live on the same thread.

lib/api/build.rs
```diff
@@ -2,40 +2,24 @@
 fn build_v8() {
     use bindgen::callbacks::ParseCallbacks;
     use std::{
-        env,
-        path::PathBuf,
+        env, fs,
```
marxin, line 5: Needs to be merged after #6601.

marxin (approved):
With my limited knowledge I have about  the WebAssembly internals - the changes look reasonable to me.

lib/api/src/backend/v8/entities/memory/mod.rs
```diff
@@ -147,83 +146,42 @@ impl Memory {
...
-    pub fn try_copy(&self, store: &impl AsStoreRef) -> Result<VMMemory, MemoryError> {
-        check_isolate(store);
-
-        let res = unsafe { wasm_memory_copy(self.handle) };
-        if res.is_null() {
-            Err(MemoryError::Generic("memory copy failed".to_owned()))
-        } else {
-            Ok(res)
-        }
+        // FIXME: implement memory copying
```
syrusakbary, line 163: Implement this so it works in V8

lib/api/src/entities/memory/mod.rs
```diff
@@ -151,14 +161,17 @@ impl Memory {
...
+    /// Since Wasmer 8.0, this function can no longer be used for stores
+    /// in different threads; for that, use `copy`, and then `attach`
+    /// on the thread owning the other `Store`.
     pub fn copy_to_store(
         &self,
         store: &impl AsStoreRef,
         new_store: &mut impl AsStoreMut,
     ) -> Result<Self, MemoryError> {
-        self.0.copy_to_store(store, new_store).map(Self)
+        self.copy(store).map(|memory| memory.attach(new_store))
```
syrusakbary, line 174: We should do a copy & attach into the same operation.

lib/api/src/entities/memory/shared.rs
```diff
@@ -24,24 +20,51 @@ impl std::fmt::Debug for SharedMemory {
...
+    ) -> Self {
         Self {
             memory,
-            ops: std::sync::Arc::new(ops),
+            ops: Some(ops),
         }
     }
 
+    /// Attach this shared memory to the provided store.
+    pub fn attach(self, store: &mut impl AsStoreMut) -> Memory {
```
syrusakbary, line 53: I don't think this function is needed.

Right now is needed because first we copy, then we attach. But we can simply have one function that does this two things at the same time.

lib/api/src/entities/memory/inner.rs
```diff
@@ -170,39 +170,20 @@ impl BackendMemory {
...
-            Self::V8(s) => s
-                .try_copy(store)
-                .map(|new_memory| Self::new_from_existing(new_store, VMMemory::V8(new_memory))),
-            #[cfg(feature = "js")]
-            Self::Js(s) => s
-                .try_copy(store)
-                .map(|new_memory| Self::new_from_existing(new_store, VMMemory::Js(new_memory))),
-        }
+        self.copy(store)
+            .map(|new_memory| new_memory.attach(new_store).0)
```
syrusakbary, line 186: Join copy & attach on the same operation (so attach is not needed)

lib/api/src/entities/memory/mod.rs
```diff
@@ -170,24 +183,33 @@ impl Memory {
...
         new_store: &mut impl AsStoreMut,
     ) -> Result<Self, MemoryError> {
-        self.0.share_in_store(store, new_store).map(Self)
+        if !self.ty(store).shared {
+            return Err(MemoryError::MemoryNotShared);
+        }
+
+        self.as_shared(store)
+            .ok_or_else(shared_memory_detach_error)
+            .map(|memory| memory.attach(new_store))
```
syrusakbary, line 212: Join copy & attach on the same operation (so attach is not needed)

syrusakbary (approved):
The PR have some minor comments that should be trivial to fix, then we can merge

lib/api/src/backend/v8/entities/memory/mod.rs
Arshia001: That will be the topic of a separate PR by @marxin, which will happen as part of the V8 efforts. Also see #6607 for why this is less-than-trivial to implement at this stage.

PR #6597, opened 2026-05-16
test(wasix): serialize fixture builds across nextest processes

sonukapoor:
This is a narrow follow-up for #6538.

The WASIX wasm test harness had an in-memory `BUILD_LOCKS` map around fixture builds. That only serializes builds inside one test process. `cargo nextest` can run tests from the same binary in separate processes, so fixtures that share a build directory can still compile to the same `main` output concurrently.

This replaces the process-local build lock with a temp-directory lock keyed by fixture path, so shared fixture builds serialize across nextest processes as well. The lock has a timeout so a stale lock fails with a clear error instead of waiting forever.

I could not reproduce the linux-musl flake locally on macOS, but the affected `legacy_process_switching` group exercises the same shared fixture build path.

Verified locally with:

```sh
. ~/.wasixcc/env
unset CC CXX
export LLVM_SYS_221_PREFIX="$PWD/.llvm"

cargo fmt --check
cargo test -p wasmer-wasix --features sys --test wasix-wasm legacy_process_switching -- --nocapture
for i in {1..20}; do
  cargo nextest run -p wasmer-wasix --features sys --test wasix-wasm legacy_process_switching --no-fail-fast -j 5 >/tmp/wasmer-nextest-loop.log 2>&1 || { cat /tmp/wasmer-nextest-loop.log; exit 1; }
done
git diff --check
```

### DIFF
--- a/lib/wasix/tests/wasm_tests/mod.rs
+++ b/lib/wasix/tests/wasm_tests/mod.rs
@@ -198,13 +198,17 @@ mod socket_tests;
 mod threadlocal_tests;
 
 use std::borrow::Cow;
-use std::collections::HashMap;
+use std::collections::hash_map::DefaultHasher;
+use std::fs;
+use std::hash::{Hash, Hasher};
 use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
 use std::process::Command;
 use std::sync::{Arc, Mutex, OnceLock};
 use std::task::{Context, Poll};
+use std::thread;
+use std::time::{Duration, Instant};
 use tracing_subscriber::EnvFilter;
 use tracing_subscriber::fmt::MakeWriter;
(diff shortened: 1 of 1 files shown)


marxin:
I am aiming for a more straightforward fix where test builds and runs happen in a separate temporary folder. This is mandatory, especially with additional Engines, where the issue will become even worse.

@sonukapoor Please do not pick random issues to work on. First, ask to be assigned to a bug.

sonukapoor:
@marxin Please share your discord username so that I can reach out to you. Thanks.

marxin:
Please let's communicate your interest in the issues directly and then I'm going to assign you the selected issue and you can start working on it.

sonukapoor:
@marxin, sure - Could you assign an easy task for me to get my feet wet?

PR #6598, opened 2026-05-18
Fix `isize` overflow in `AArch64` PC-relative relocation range assertions

pmikolajczyk41:
# Description

The two range assertions:
```rust
((-1 << 32)..(1 << 32)).contains(&delta)  // Aarch64AdrPrelPgHi21
((-1 << 20)..(1 << 20)).contains(&delta)  // Aarch64AdrPrelLo21
```

cast the `u64` delta from `for_address()` to `isize` before the check. On a 32-bit host, `isize` is 32 bits, so `1 << 32` overflows and panics in debug builds. This is reproducible when cross-compiling `wasmer` for `AArch64` from a 32-bit host.

Fix: cast to `i64`, which is the correct type for a signed 64-bit offset regardless of the host pointer width.


marxin (approved):
LGTM, thank you for the contribution

PR #6599, opened 2026-05-18
tests: port WASIX tests to libtest-mimic (part 2)

### LINKED ISSUE #6574: wasix tests: support multiple run-time engines, consider libtest-mimic

marxin:
* Addresses https://github.com/wasmerio/wasmer/issues/6538#issuecomment-4476351547 by first copying each configuration into its own dedicated folder, where both the build and execution steps take place. As a result, the `#Tempdir` directive is no longer needed.
* Resolves conflicts with tests that relied on source files located in sibling or parent directories by adding support for multiple `.sh` files within a single folder, removing dependencies between folders.
* Runs all tests against multiple engines: Cranelift and LLVM.

Resolves: #6574


Arshia001 (approved):
LGTM!

PR #6600, opened 2026-05-18
build(v8): bump V8 backend to 15.0.1

synchwire:
Bumps the V8 backend to **V8 15.0.1**, using release assets from
[synchwire/v8-custom-builds](https://github.com/synchwire/v8-custom-builds)
tag `15.0.1-4` (tracking [v8-custom-builds#12](https://github.com/wasmerio/v8-custom-builds/pull/12)).
Replaces the previous 11.9.7 `wee8` prebuilt from `wasmerio/wee8-custom-builds`.

## Dependency

Requires v8-custom-builds release **15.0.1-4**, which forward-ports two
wasm-c-api patches the V8 backend links against:

- **shared-memory (`0007`)** — makes wasm-c-api memory a *sharable ref*,
  providing `wasm_memory_share` / `wasm_shared_memory_delete`. Without it the
  V8 test binaries fail to link (undefined references).
- **funcref-null (`0006`)** — initialises C-API funcref tables with the
  internal wasm-null sentinel.

`build.rs` points at `synchwire/v8-custom-builds`; once v8-custom-builds#12
merges and tags under `wasmerio/`, only the release URL/version constant needs
updating.

## Changes

**`build.rs`** — download the V8 15.0.1-4 assets and keep the `wee8_` symbol
prefixing. V8 15 object files use `SHT_CREL` sections (an LLVM ELF extension)
that GNU objcopy silently skips, leaving symbols unrenamed and breaking the
link. `build.rs` now prefers versioned `llvm-objcopy` and fails the build with
an actionable message (verified via `nm`) if nothing was renamed. The linux V8
CI job installs `llvm` so `llvm-objcopy` is on `PATH`.

**Trap messages** — V8 15 splits the old combined `call_indirect` trap
`"null function or function signature mismatch"` into `"null function"`
(null / uninitialized element) and `"function signature mismatch"` (signature
mismatch on a present entry). Spec-test and `traps.rs` expectations updated to
accept both.

**Windows** — V8 on Windows is not wired up: the v8-custom-builds Windows recipe
builds `v8_monolith`, which does not export the `wasm_*` C-API symbols the
backend links against. `build.rs` disables the Windows arm and `make test-all`
no longer adds V8 to the Windows test compilers.

**CI** — two informational (`continue-on-error: true`) cross-compile checks:
Android (green) and iOS (fails until a `v8-ios` asset is published — mobile is
tracked separately).

## Known V8 15 behavior differences

- Unbounded memory reports its type maximum as the 32-bit spec cap
  (65536 pages) instead of `None`. Only affects doctests run under a v8-only
  build, not the default multi-backend test config.

### DIFF
--- a/.github/workflows/android-v8-check.yml
+++ b/.github/workflows/android-v8-check.yml
@@ -1,0 +1,97 @@
+name: Android V8 cross-compile
+# Informational check that wasmer's v8 backend cross-compiles to
+# aarch64-linux-android. Same caveats as ios-v8-check.yml: the v8
+# feature isn't exercised by the main build.yml matrix, and this
+# workflow sits on top of a third-party prebuilt V8 archive plus
+# the NDK clang toolchain, both of which drift independently of
+# wasmer itself. `continue-on-error: true` keeps failures visible
+# without blocking merges.
+#
+# Only aarch64 is covered — Google Play requires 64-bit ARM since
+# 2021, the upstream wee8-custom-builds / synchwire/v8-custom-builds
+# only ship aarch64 android archives, and wasmer's build.rs URL
+# map only knows about aarch64-android. armv7 and x86_64 android
+# are out of scope.
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+jobs:
+  android-v8:
+    name: cargo build --target aarch64-linux-android --features v8
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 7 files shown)


theduke:
@marxin note the user description ... 

`I am an AI agent`.

marxin:
> `I am an AI agent`.

I know, but there's a person behind the changes that can clean up and polished the LLM-generated changes.

The PR is waiting for the change in the wee8 repository.

PR #6601, opened 2026-05-18
Cache downloaded v8 artifact during builds of wasmer with the V8 feature enabled

Arshia001:
Per title. Downloading the whole archive once per build on a slow internet connection takes a lot of time.

### DIFF
--- a/lib/api/build.rs
+++ b/lib/api/build.rs
@@ -29,7 +29,8 @@ fn build_v8() {
     let out_path = PathBuf::from(&out_dir);
     let crate_root = env::var("CARGO_MANIFEST_DIR").unwrap();
     let v8_header_path = PathBuf::from(&crate_root).join("third-party").join("wee8");
-    let cache_dir = stable_target_dir(&out_path)
+    let cache_dir = out_path
+        .join("../../../../")
         .join("wee8-artifacts")
         .join(WEE8_RELEASE_VERSION)
         .join(platform_name);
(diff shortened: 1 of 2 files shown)


lib/api/build.rs
```diff
@@ -45,29 +29,137 @@ fn build_v8() {
     let out_path = PathBuf::from(&out_dir);
     let crate_root = env::var("CARGO_MANIFEST_DIR").unwrap();
     let v8_header_path = PathBuf::from(&crate_root).join("third-party").join("wee8");
+    let cache_dir = stable_target_dir(&out_path)
```
marxin: To be honest, this change introduces a lot of temporary folder creation, removal, and renaming, which makes the code more convoluted. Couldn’t we just download directly to a temporary file and then rename it in the root folder to something like `wee8-$WEE8_RELEASE_VERSION-$platform.tar.xz`?

marxin (changes requested):

lib/api/build.rs
Arshia001: yeah, agreed. also, why can't I get copilot to agree to this (initially!) very simple change is beyond me :D

lib/api/build.rs
```diff
@@ -179,6 +252,16 @@ fn build_v8() {
     }
 
     println!("cargo:rustc-link-lib=static=v8prefixed");
+
+    fn stable_target_dir(out_path: &Path) -> PathBuf {
```
marxin: Can't we use rather `path.join("../../../../")` - took me while to realize what's happening in the function.

marxin (approved):

PR #6604, opened 2026-05-18
Add soft-float libcall support for LLVM backend on hardware-FP-less targets

pmikolajczyk41:
# Description

  ## Motivation

  When targeting platforms without hardware floating-point (e.g. soft-float RISC-V), LLVM emits references to compiler-rt / libgcc arithmetic helpers such as `__adddf3`, `__mulsf3`, `__eqdf2`, etc. Wasmer's
  LLVM backend resolves libcall symbols by name at JIT link time, but these routines were not registered, causing link failures for any wasm module compiled to such a target.

  ## Changes

Adds the full set of f32/f64 soft-float routines — arithmetic, conversion, and comparison — as `LibCall` variants. The variants and their ordering follow the GCC runtime library documentation (§3.2). Each variant maps to its standard compiler-rt symbol name (e.g. `__addsf3`) so Cranelift's relocation resolver can find it in the host process.

  On the VM side, all routines are declared as typed `extern "C"` symbols, letting the linker resolve them from the host's compiler-rt / libgcc at build time. The soft-float lookup in the LLVM object linker is restricted to ELF targets and gated on `#[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]`, because LLVM never emits these relocations when hardware floating-point is available, and no supported Apple (Mach-O) platform requires soft-float.
  
  ## Sources
  * https://gcc.gnu.org/onlinedocs/gccint/Soft-float-library-routines.html
  * https://github.com/llvm/llvm-project/tree/main/compiler-rt/lib/builtins

  ## Coverage
  
  The set of routines was cross-checked against the GCC internals manual and the LLVM compiler-rt builtins directory. All symbols present in the implementation are standard; the following categories were deliberately omitted:

  * **Extended precision (`tf`/`xf` variants, e.g. `__addtf3`)** — 128-bit and 80-bit `long double` helpers. Only emitted for targets with native TFmode/XFmode; not relevant for wasm targets.
  * **Complex arithmetic (`__muldc3`, `__divsc3`, etc.)** — only generated for C99 complex number operations, not present in wasm semantics.
  * **Power functions (`__powisf2`, `__powidf2`)** — GCC extension, not emitted by LLVM for standard wasm targets.
  * **BitInt conversions (`__floatbitintdf`, etc.)** — a recent GCC/Clang extension for arbitrary-width integers, not reachable from wasm.
  * **Half-precision / bfloat16 (`__extendhfsf2`, `__truncsfbf2`, etc.)** — only emitted when the target has native `_Float16` / `__bf16` types, which wasm does not expose.
  * **Three-way comparison** — `__cmpdf2` / `__cmpsf2` are the internal primitives that `__eqdf2` etc. are built on; LLVM emits the specific comparison functions directly, never the generic three-way variant.
  * **Integer arithmetic helpers (`__divdi3`, `__mulsi3`, etc.)** — 64-bit and 32-bit integer runtime helpers; out of scope for this PR, which focuses on floating-point emulation only.


lib/compiler-llvm/src/object_file.rs
```diff
@@ -150,6 +208,66 @@ static LIBCALLS_MACHO: phf::Map<&'static str, LibCall> = phf::phf_map! {
...
+    "___ledf2" => LibCall::Ledf2,
+    "___lesf2" => LibCall::Lesf2,
+    "___ltdf2" => LibCall::Ltdf2,
+    "___ltsf2" => LibCall::Ltsf2,
+    "___moddi3" => LibCall::Moddi3,
+    "___modsi3" => LibCall::Modsi3,
+    "___muldf3" => LibCall::Muldf3,
+    "___muldi3" => LibCall::Muldi3,
+    "___mulsf3" => LibCall::Mulsf3,
+    "_wasmer_vm__mulsi3" => LibCall::Mulsi3,
```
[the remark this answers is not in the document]
pmikolajczyk41: fixed in [3590973](https://github.com/wasmerio/wasmer/pull/6604/commits/35909731bd0bbbc338156eb4853daeb60a2cf0e1)

lib/compiler-llvm/src/object_file.rs
```diff
@@ -90,6 +90,62 @@ static LIBCALLS_ELF: phf::Map<&'static str, LibCall> = phf::phf_map! {
     "wasmer_eh_personality2" => LibCall::EHPersonality2,
     "wasmer_vm_dbg_usize" => LibCall::DebugUsize,
     "wasmer_vm_dbg_str" => LibCall::DebugStr,
+    // Soft-float and 64-bit integer arithmetic routines that LLVM emits for
```
marxin: I would like to see it guarded by the `cfg` so that it gets compiled in only for the desired platforms.

lib/compiler-llvm/src/object_file.rs
```diff
@@ -90,6 +90,62 @@ static LIBCALLS_ELF: phf::Map<&'static str, LibCall> = phf::phf_map! {
...
+    "__subdf3" => LibCall::Subdf3,
+    "__subsf3" => LibCall::Subsf3,
+    "__truncdfsf2" => LibCall::Truncdfsf2,
+    "__udivdi3" => LibCall::Udivdi3,
+    "__udivsi3" => LibCall::Udivsi3,
+    "__umoddi3" => LibCall::Umoddi3,
+    "__umodsi3" => LibCall::Umodsi3,
+    "__unorddf2" => LibCall::Unorddf2,
+    "__unordsf2" => LibCall::Unordsf2,
+    "memset" => LibCall::Memset,
```
marxin: Do you really need `memset`? It's a risky operation from the VM perspective - can we map it to `wasmer_vm_memory32_fill`?

lib/compiler-llvm/src/object_file.rs
```diff
@@ -90,6 +90,62 @@ static LIBCALLS_ELF: phf::Map<&'static str, LibCall> = phf::phf_map! {
...
+    "sqrt" => LibCall::Sqrt,
 };
 
 static LIBCALLS_MACHO: phf::Map<&'static str, LibCall> = phf::phf_map! {
```
marxin, line 148: Is there any supported Mach-O platform that needs these math routines?

lib/vm/src/libcalls.rs
```diff
@@ -1006,5 +1006,127 @@ pub fn function_pointer(libcall: LibCall) -> usize {
...
+        LibCall::Unordsf2 => __unordsf2 as *const () as usize,
+        LibCall::Memset => memset as *const () as usize,
+        LibCall::Sqrt => sqrt as *const () as usize,
     }
 }
+
+// Soft-float and 64-bit integer arithmetic routines. These are provided by
+// compiler-rt / libgcc on every standard Rust target. We declare them here so
+// wasmer can resolve JIT-compiled references to them at link time.
+unsafe extern "C" {
```
marxin, line 1108: Am I right we will hit a linker error if some of the defined routines is undefined?

marxin (changes requested):
Generally, I am fine with extending the libcalls, but I would like to see them guarded specifically for the affected platforms. This would ensure that currently supported platforms do not end up using any of the newly introduced libcall functions.

lib/compiler-llvm/src/object_file.rs
syrusakbary: memset needs to be out, otherwise it will break the sandbox requirements

lib/vm/src/libcalls.rs
pmikolajczyk41: we hit compilation error like:
```
wasm compilation: Codegen("relocation Relocation { kind: Unknown, encoding: Generic, size: 0, target: Symbol(SymbolIndex(172)), addend: 0, implicit_addend: false, flags: Elf { r_type: 19 } } targets unknown symbol 'Symbol { name: \"__adddf3\", address: 0, size: 0, kind: Unknown, section: Undefined, scope: Unknown, weak: false, flags: Elf { st_info: 16, st_other: 0 } }'")
```

lib/compiler-llvm/src/object_file.rs
pmikolajczyk41: Initially I added it for consistency, but you are right - I can't find any supported Mach-O platform that would use it. Dropped in [735bb64](https://github.com/wasmerio/wasmer/pull/6604/commits/735bb64305ba31b446d1731103295c09ae1ddadf)

lib/compiler-llvm/src/object_file.rs
pmikolajczyk41: Removed in [376dc93](https://github.com/wasmerio/wasmer/pull/6604/commits/376dc93d0505c87c3571cef2cbf48ffe3d01e53b)

lib/compiler-llvm/src/object_file.rs
pmikolajczyk41: done ✅

lib/compiler-llvm/src/object_file.rs
```diff
@@ -90,6 +90,67 @@ static LIBCALLS_ELF: phf::Map<&'static str, LibCall> = phf::phf_map! {
...
+    "__subdf3" => LibCall::Subdf3,
+    "__subsf3" => LibCall::Subsf3,
+    "__truncdfsf2" => LibCall::Truncdfsf2,
+    "__udivdi3" => LibCall::Udivdi3,
+    "__udivsi3" => LibCall::Udivsi3,
+    "__umoddi3" => LibCall::Umoddi3,
+    "__umodsi3" => LibCall::Umodsi3,
+    "__unorddf2" => LibCall::Unorddf2,
+    "__unordsf2" => LibCall::Unordsf2,
 };
```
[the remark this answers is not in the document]
pmikolajczyk41: outdated

lib/vm/src/libcalls.rs
```diff
@@ -1006,5 +1013,122 @@ pub fn function_pointer(libcall: LibCall) -> usize {
...
+    fn __nesf2();
+    fn __subdf3();
+    fn __subsf3();
+    fn __truncdfsf2();
+    fn __udivdi3();
+    fn __udivsi3();
+    fn __umoddi3();
+    fn __umodsi3();
+    fn __unorddf2();
+    fn __unordsf2();
```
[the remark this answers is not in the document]
pmikolajczyk41: outdated

lib/vm/src/libcalls.rs
```diff
@@ -1006,5 +1013,122 @@ pub fn function_pointer(libcall: LibCall) -> usize {
...
+        LibCall::Umoddi3 => __umoddi3 as *const () as usize,
+        LibCall::Umodsi3 => __umodsi3 as *const () as usize,
+        LibCall::Unorddf2 => __unorddf2 as *const () as usize,
+        LibCall::Unordsf2 => __unordsf2 as *const () as usize,
     }
 }
+
+// Soft-float and 64-bit integer arithmetic routines. These are provided by
+// compiler-rt / libgcc on every standard Rust target. We declare them here so
+// wasmer can resolve JIT-compiled references to them at link time.
```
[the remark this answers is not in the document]
pmikolajczyk41: outdated

lib/vm/src/libcalls.rs
```diff
@@ -1006,5 +1013,122 @@ pub fn function_pointer(libcall: LibCall) -> usize {
...
+        LibCall::Gtsf2 => __gtsf2 as *const () as usize,
+        LibCall::Ledf2 => __ledf2 as *const () as usize,
+        LibCall::Lesf2 => __lesf2 as *const () as usize,
+        LibCall::Ltdf2 => __ltdf2 as *const () as usize,
+        LibCall::Ltsf2 => __ltsf2 as *const () as usize,
+        LibCall::Moddi3 => __moddi3 as *const () as usize,
+        LibCall::Modsi3 => __modsi3 as *const () as usize,
+        LibCall::Muldf3 => __muldf3 as *const () as usize,
+        LibCall::Muldi3 => __muldi3 as *const () as usize,
+        LibCall::Mulsf3 => __mulsf3 as *const () as usize,
```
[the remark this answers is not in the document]
pmikolajczyk41: outdated

lib/vm/src/libcalls.rs
```diff
@@ -145,6 +145,12 @@ pub extern "C" fn wasmer_vm_f64_nearest(x: f64) -> f64 {
     }
 }
 
+/// Implementation of f64.sqrt
+#[unsafe(no_mangle)]
+pub extern "C" fn wasmer_vm_sqrt(x: f64) -> f64 {
```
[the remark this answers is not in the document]
pmikolajczyk41: outdated

lib/compiler-llvm/src/object_file.rs
```diff
@@ -150,8 +211,25 @@ static LIBCALLS_MACHO: phf::Map<&'static str, LibCall> = phf::phf_map! {
...
+        BinaryFormat::Elf => &LIBCALLS_ELF,
+        BinaryFormat::Macho => &LIBCALLS_MACHO,
+        _ => return None,
+    };
+    if let Some(&lc) = base.get(name) {
+        return Some(lc);
+    }
+    #[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
+    if let Some(&lc) = SOFTFLOAT_LIBCALLS_ELF.get(name) {
+        return Some(lc);
```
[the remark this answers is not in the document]
pmikolajczyk41: outdated

lib/vm/src/libcalls.rs
```diff
@@ -1006,5 +1006,99 @@ pub fn function_pointer(libcall: LibCall) -> usize {
...
+        LibCall::Ledf2 => __ledf2 as *const () as usize,
+        LibCall::Gtsf2 => __gtsf2 as *const () as usize,
+        LibCall::Gtdf2 => __gtdf2 as *const () as usize,
     }
 }
+
+// Soft-float arithmetic routines. These are provided by compiler-rt / libgcc
+// on every standard Rust target. We declare them here so wasmer can resolve
+// JIT-compiled references to them at link time.
+unsafe extern "C" {
```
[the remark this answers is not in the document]
pmikolajczyk41: MSRV is 1.92 and unsafe extern "C" { } was stabilized in 1.82, so the current form is fine.

lib/compiler-llvm/src/object_file.rs
```diff
@@ -92,6 +92,58 @@ static LIBCALLS_ELF: phf::Map<&'static str, LibCall> = phf::phf_map! {
     "wasmer_vm_dbg_str" => LibCall::DebugStr,
 };
 
+// Soft-float routines that LLVM emits for targets without hardware floating-point. Not needed on
+// x86-64 or AArch64 because LLVM lowers float and integer arithmetic to native instructions on
+// those targets, never to these helpers.
+#[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
```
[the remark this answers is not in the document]
pmikolajczyk41: marxin specifically requested this

lib/compiler-llvm/src/object_file.rs
```diff
@@ -298,10 +364,10 @@ where
...
                         }
                         // Maybe a libcall then?
-                    } else if let Some(libcall) = libcalls.get(symbol_name) {
-                        RelocationTarget::LibCall(*libcall)
-                    } else if let Ok(Some(reloc_target)) =
-                        symbol_name_to_relocation_target(symbol_name)
+                    } else if let Some(libcall) = lookup_libcall(symbol_name, binary_fmt) {
+                        RelocationTarget::LibCall(libcall)
+                    } else if let Some(reloc_target) =
+                        symbol_name_to_relocation_target(symbol_name)?
```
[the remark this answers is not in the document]
pmikolajczyk41: [e2cc632](https://github.com/wasmerio/wasmer/pull/6604/commits/e2cc63242c1e52e7ad9a873ff22349a60e9370d5)

lib/vm/src/libcalls.rs
```diff
@@ -1006,5 +1006,99 @@ pub fn function_pointer(libcall: LibCall) -> usize {
...
+        LibCall::Nesf2 => __nesf2 as *const () as usize,
+        LibCall::Nedf2 => __nedf2 as *const () as usize,
+        LibCall::Gesf2 => __gesf2 as *const () as usize,
+        LibCall::Gedf2 => __gedf2 as *const () as usize,
+        LibCall::Ltsf2 => __ltsf2 as *const () as usize,
+        LibCall::Ltdf2 => __ltdf2 as *const () as usize,
+        LibCall::Lesf2 => __lesf2 as *const () as usize,
+        LibCall::Ledf2 => __ledf2 as *const () as usize,
+        LibCall::Gtsf2 => __gtsf2 as *const () as usize,
+        LibCall::Gtdf2 => __gtdf2 as *const () as usize,
```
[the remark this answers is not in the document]
pmikolajczyk41: [665a758](https://github.com/wasmerio/wasmer/pull/6604/commits/665a758e5d6174ce60466b7f72e64b326355bec0)

pmikolajczyk41:
@marxin @syrusakbary I have reduced the scope: the original PR was broader than necessary. Here's what changed:
  - Removed `sqrt`
  - Removed integer helpers (`__divdi3`, `__moddi3`, `__mulsi3`, etc.) these are integer arithmetic routines, not soft-float; wrong PR
  - The PR now covers exactly the GCC §3.2 soft-float routines: arithmetic, conversion, and comparison for f32/f64, ordered to match the official documentation (https://gcc.gnu.org/onlinedocs/gccint/Soft-float-library-routines.html)

lib/compiler-llvm/src/object_file.rs
marxin: Can we have a rather explicit listing of architectures where we want the soft FP operations? `!x86_64 || !aarch64` does not seem correct to me.

lib/compiler-llvm/src/object_file.rs
```diff
@@ -152,6 +204,24 @@ static LIBCALLS_MACHO: phf::Map<&'static str, LibCall> = phf::phf_map! {
...
+fn lookup_libcall(name: &str, fmt: BinaryFormat) -> Option<LibCall> {
+    let base = match fmt {
+        BinaryFormat::Elf => &LIBCALLS_ELF,
+        BinaryFormat::Macho => &LIBCALLS_MACHO,
+        _ => return None,
+    };
+    if let Some(&lc) = base.get(name) {
+        return Some(lc);
+    }
+    #[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
```
marxin: Similarly here.

lib/vm/src/libcalls.rs
```diff
@@ -1006,5 +1006,189 @@ pub fn function_pointer(libcall: LibCall) -> usize {
         }
         LibCall::DebugUsize => wasmer_vm_dbg_usize as *const () as usize,
         LibCall::DebugStr => wasmer_vm_dbg_str as *const () as usize,
+        // --- Soft-float libcalls ---
+        // compiler-rt / libgcc provides these on every std Rust target.
+        // On wasm32 the JIT engine is never active, so these variants are unreachable.
+        #[cfg(not(target_arch = "wasm32"))]
```
marxin: Can we just one `#[cfg]` gate where we'll do a sub-match for all the soft FP calls?

lib/vm/src/libcalls.rs
marxin: That's a reasonable error message.

marxin (changes requested):
We're getting quite close, please take a look at my inline comments.

lib/vm/src/libcalls.rs
pmikolajczyk41: done in [66a9591](https://github.com/wasmerio/wasmer/pull/6604/commits/66a9591da7d0904cc7308151f18136b8b884941b)

lib/compiler-llvm/src/object_file.rs
pmikolajczyk41: good point, updated in [9005fbd](https://github.com/wasmerio/wasmer/pull/6604/commits/9005fbdfd8534e09ab219c79f6f0046c8195aa4a)

lib/compiler-llvm/src/object_file.rs
pmikolajczyk41: done in [9005fbd](https://github.com/wasmerio/wasmer/pull/6604/commits/9005fbdfd8534e09ab219c79f6f0046c8195aa4a)

lib/compiler-llvm/src/object_file.rs
```diff
@@ -92,10 +92,8 @@ static LIBCALLS_ELF: phf::Map<&'static str, LibCall> = phf::phf_map! {
     "wasmer_vm_dbg_str" => LibCall::DebugStr,
 };
 
-// Soft-float routines that LLVM emits for targets without hardware floating-point. Not needed on
-// x86-64 or AArch64 because LLVM lowers float and integer arithmetic to native instructions on
-// those targets, never to these helpers.
-#[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
+// Soft-float routines emitted by LLVM for targets without hardware floating-point.
+#[cfg(any(target_arch = "riscv32", target_arch = "riscv64"))]
```
marxin: I think it's still imprecise - we should combine the target architecture with `"target_feature = "f"`?

lib/vm/src/libcalls.rs
```diff
@@ -1006,5 +1006,149 @@ pub fn function_pointer(libcall: LibCall) -> usize {
...
+        | LibCall::Nedf2
+        | LibCall::Gesf2
+        | LibCall::Gedf2
+        | LibCall::Ltsf2
+        | LibCall::Ltdf2
+        | LibCall::Lesf2
+        | LibCall::Ledf2
+        | LibCall::Gtsf2
+        | LibCall::Gtdf2) => {
+            #[cfg(target_arch = "wasm32")]
```
marxin, line 1054: Should be aligned with the previously used `cfg` selection.

lib/vm/src/libcalls.rs
```diff
@@ -1006,5 +1006,149 @@ pub fn function_pointer(libcall: LibCall) -> usize {
...
+                LibCall::Gtsf2 => __gtsf2 as *const () as usize,
+                LibCall::Gtdf2 => __gtdf2 as *const () as usize,
+                _ => unreachable!(),
+            }
+        }
     }
 }
+
+// Soft-float arithmetic routines. Provided by compiler-rt / libgcc on all non-wasm targets.
+#[cfg(not(target_arch = "wasm32"))]
```
marxin, line 1107: Same here.

marxin (changes requested):
Once we make the `cfg` filtering correctly, we can merge the PR.

pmikolajczyk41:
Hey @marxin, following up on the cfg alignment issue (I thought it would be easier to discuss it here):

I tried using the same `all(any(riscv32, riscv64), not(target_feature = "f"))` gate in both `object_file.rs` and `libcalls.rs`, and CI failed on aarch64 (again, I tried it also yesterday, before your comment): https://github.com/wasmerio/wasmer/actions/runs/26502620838/job/78046740120?pr=6604#step:20:969. The same happens on [arm64](https://github.com/wasmerio/wasmer/actions/runs/26502620838/job/78046740301?pr=6604#step:20:972) or [windows-x64](https://github.com/wasmerio/wasmer/actions/runs/26502620838/job/78046740091?pr=6604#step:20:1014).

The reason is `make_libcall_trampolines` in `trampoline.rs`, which eagerly pre-builds trampolines for every `LibCall` variant via `enum_iterator::all::<LibCall>()`. On x86_64 and aarch64 each trampoline carries an `Abs8` relocation, so `link.rs` ends up calling `function_pointer` for every variant — including all soft-float ones — at module-link time, even when the compiled WASM module never touches floats (the failing CI job is the `compiler_cranelift` example, which only compiles `i32.add`).

So these two gates serve fundamentally different roles and it seems that they can't be the same:
  - `object_file.rs` (riscv+no-f): controls when we recognize a soft-float symbol name in the object file as a `LibCall` relocation target — correct, since LLVM only emits those symbols on targets without hardware float.
  - `libcalls.rs`: controls when we can resolve a LibCall variant to a runtime address — must cover all non-wasm targets because of the eager trampoline pre-building described above.

My proposed fix is to use `#[cfg(not(target_arch = "wasm32"))]` in `libcalls.rs` (on wasm32 the JIT is never active, so soft-float variants are genuinely unreachable there; on every other target compiler-rt/libgcc provides the symbols).

An alternative would be to gate every soft-float variant in the `LibCall` enum itself with `#[cfg(all(any(riscv32, riscv64), not(target_feature = "f")))]`, which would make `enum_iterator::all::<LibCall>()` skip them on other targets and avoid the trampoline issue entirely. But besides being visually awful, it would break serialization: the variant indices baked into serialized artifacts depend on the full enum layout, so stripping variants on non-riscv targets would make cross-compiled artifacts unloadable on those hosts.

marxin:
Thank you for your explanation and I think we can merge the PR. Good job.

marxin (approved):

PR #6605, opened 2026-05-19
fix(wasix): allow unlinking symlinks discovered on the backing FS

kilyanni:
# Description

`path_unlink_file` returned `EINVAL` when called on a symlink that exists on the backing filesystem but was never created through the `path_symlink` syscall. Path resolution deliberately doesn't cache such symlinks in the parent directory's entry map, but the unlink path assumed anything reachable would be cached and bailed out otherwise.

The fix detects this case and removes the symlink directly from the backing FS (without following it), so backing-FS symlinks behave the same as ones created in-process. The added test pre-creates a host symlink, unlinks it from wasm, and asserts the link is gone while its target survives.


lib/wasix/src/syscalls/wasi/path_unlink_file.rs
```diff
@@ -75,18 +75,46 @@ pub(crate) fn path_unlink_file_internal(
...
+                    Ok(()) => {
+                        state.fs.unregister_ephemeral_symlink(host_path);
+                        Errno::Success
+                    }
+                    Err(Errno::Noent)
+                        if state.fs.ephemeral_symlink_at(host_path).is_some() =>
+                    {
+                        state.fs.unregister_ephemeral_symlink(host_path);
+                        Errno::Success
+                    }
```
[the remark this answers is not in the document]
kilyanni: @copilot apply changes based on [this feedback](https://github.com/wasmerio/wasmer/pull/6605#discussion_r3263402665)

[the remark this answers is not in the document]
kilyanni: I think this is a very niche nearly irreproducible scenario, but we're forced to cover it lest rust complains about uncovered match arms

Unless we mark it as `unreachable!()`, but I'm not convinced it's truly unreachable

Arshia001:
@kilyanni we recently merged two rather big PRs, one around unlinking and one around symlinks, and this one exists at the intersection of those two things. Can you merge main in, and make sure everything works as expected and CI is green afterwards please?

lib/wasix/src/syscalls/wasi/path_unlink_file.rs
```diff
@@ -75,18 +75,49 @@ pub(crate) fn path_unlink_file_internal(
...
-                "Internal logic error in wasi::path_unlink_file, parent is not a directory"
-            ),
+            None => {
+                // Symlinks discovered during path resolution aren't cached in `entries`
+                // (path lookup re-resolves them each time), so a backing-FS symlink is
+                // absent here even though the host file exists. Symlinks created via
+                // `path_symlink` are cached and take the `Some` branch above.
+                let inode_is_symlink = matches!(inode.read().deref(), Kind::Symlink { .. });
+                if !inode_is_symlink {
+                    unreachable!(
```
Arshia001: Are we sure this is actually unreachable? IIUC, this is unreachable because only symlinks get the special treatment of "don't cache intermediate ones", which can be the case for other types of inodes, either today or in the future.

Arshia001 (changes requested):

lib/wasix/src/syscalls/wasi/path_unlink_file.rs
kilyanni: Thinking about it, even if it is only symlinks, a race between two deletes on the same target could cause this to be hit.

It is still a weird edge-case. Dropped the `unreachable!` but instead marked it with a `warn!` for now so we can see if it does actually ever happen

lib/wasix/src/syscalls/wasi/path_unlink_file.rs
```diff
@@ -75,18 +75,49 @@ pub(crate) fn path_unlink_file_internal(
...
+                if !inode_is_symlink {
+                    tracing::warn!(
+                        "wasi::path_unlink_file: path resolution returned inode {:?} for {:?}, but parent directory had no matching entry",
+                        inode.ino(),
+                        child_name
+                    );
+                    return Ok(Errno::Noent);
+                }
+                let host_path = host_adjusted_path.as_path();
+                let errno = match state.fs_remove_file(host_path) {
```
Arshia001: This runs `fs_remove_file` while `parent_inode.write()` is still held (`drop(guard)` is below on line 118). In the `Some(removed_inode)` path, the write lock is released when the block ends, *before* host I/O later in the function.

Holding the parent lock across host filesystem I/O increases contention and is inconsistent with the rest of this function. Could you `drop(guard)` (or restructure) before calling `fs_remove_file` here?

lib/wasix/tests/wasm_tests/path/unlink-backing-symlink/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
Arshia001, line 1: Nit: similar path/symlink tests live under `lib/wasix/tests/wasm_tests/path/` (e.g. `path/symlink-open-read-write`), not `path_tests/`. The harness will pick up either layout, but `path/unlink-backing-symlink` would match existing convention.

lib/wasix/tests/wasm_tests/path/unlink-backing-symlink/main.c
```diff
@@ -0,0 +1,56 @@
+#include <errno.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+// Regression test for path_unlink_file on a symlink that exists in the backing
```
Arshia001, line 7: This test covers backing-FS symlinks pre-created on the host, which is the reported bug — great.

The new `None` branch in `path_unlink_file` also applies to ephemeral symlinks discovered during path lookup (they are resolved but not inserted into `entries`, same as backing symlinks). Could you add coverage for that path as well? For example, a companion test where the symlink is created via `path_symlink` on a filesystem that falls back to the ephemeral registry (`Unsupported` from `create_symlink`), then unlinked from wasm.

Arshia001 (changes requested):
Thanks for addressing the earlier `unreachable!` feedback — the `warn!` + `ENOENT` fallback looks reasonable.

A few remaining items before merge:

1. **Lock ordering:** the new `None` branch calls `fs_remove_file` while still holding `parent_inode.write()`. The normal `Some` path releases that lock before any host I/O. Please drop the guard before `fs_remove_file` (see inline comment).
2. **Test layout:** please move the regression test under `path/` to match existing wasix wasm test conventions (`path/symlink-open-read-write`, etc.).
3. **Coverage:** the backing-FS symlink case is well covered; please also add (or extend with) a case for ephemeral symlinks that are resolved but not inserted into `entries` — the same `None` branch handles both.

lib/wasix/tests/wasm_tests/path/unlink-backing-symlink/main.c
kilyanni: Added some unit tests for that case

Arshia001 (approved):

PR #6608, opened 2026-05-19
fix(wasix): Fix EPOLL to deliver multiple events per FD at the same time to avoid race conditions

artemyarulin:
Fix WASIX epoll readiness handling so multiple readiness bits for the same FD are returned as a single coalesced event, matching Linux epoll behavior.

Previously, WASIX could split readiness like `EPOLLIN | EPOLLOUT` into two separate events for the same FD. If the app handled the first event by closing the socket, the numeric FD could be reused for another resource before the stale second event was processed. In production this manifested as sporadic `Bad file descriptor` errors in PHP/WordPress workloads, and may also explain rare file corruption where bytes intended for a socket were written to a newly opened file that reused the same FD.

The fix coalesces pending readiness bits into one `epoll_event` per FD, so callers no longer receive stale follow-up events for an FD they already handled and closed. A WASIX regression test covers the FD reuse/file corruption scenario.

Based on  https://man7.org/linux/man-pages/man7/epoll.7.html 
```
- If more than one event occurs between [epoll_wait(2)](https://man7.org/linux/man-pages/man2/epoll_wait.2.html) calls, are they combined or reported separately?
- They will be combined.
```

### DIFF
--- a/lib/wasix/tests/wasm_tests/poll_tests/epoll-fd-reuse-file-corruption/main.c
+++ b/lib/wasix/tests/wasm_tests/poll_tests/epoll-fd-reuse-file-corruption/main.c
@@ -1,0 +1,65 @@
+//#ExpectedStdout: 0
+#include <arpa/inet.h>
+#include <fcntl.h>
+#include <netinet/in.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/epoll.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+int main(void) {
+  unlink("victim.php");
+
+  // Build a real TCP loopback connection. The accepted socket is the fd watched
+  // by epoll, matching the socket lifecycle used by HTTP servers.
+  int listener = socket(AF_INET, SOCK_STREAM, 0);
+  struct sockaddr_in addr = {.sin_family = AF_INET,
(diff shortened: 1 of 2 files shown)


lib/wasix/tests/wasm_tests/poll_tests/epoll-fd-reuse-file-corruption/main.c
```diff
@@ -0,0 +1,65 @@
...
+  // file open should reuse the same numeric fd, creating the stale-fd hazard.
+  close(fd);
+  int victim_fd = open("victim.php", O_CREAT | O_TRUNC | O_WRONLY, 0600);
+  if (victim_fd != fd) return 2;
+
+  // A stale second epoll event still carries the old fd number. With the bug,
+  // writing through that stale event writes into victim.php instead of a
+  // socket.
+  write(victim_fd, "<?php echo 'ok'; ?>\n", 19);
+  write(events[1].data.fd, "TLS_HANDSHAKE_BYTES", 19);
```
artemyarulin, line 55: Unfortunately I couldn't repro file corruption using WordPress/PHP load as whole thing is a race condition, so hard to replicate. But this test is how I imagine it could have happen

Arshia001 (approved):
Great find!

PR #6610, opened 2026-05-19
feat(V8): enable V8 feature on linux-x86_64 and run V8 WASIX tests

### LINKED ISSUE #6564: wasix: run tests with all supported compilers
Right now, we're speaking about `LLVM`, `Cranelift`, but in the future, we might want to include V8 as well.

marxin:
The PR is currently blocked by #6587, but the change itself is pretty trivial and can be here in a form of a draft.

Resolves: #6564

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -296,7 +296,9 @@ build_wasmer_extra_features_csv = $(subst $(space),$(comma),$(build_wasmer_extra
 test_compilers := $(compilers)
 ifeq ($(IS_LINUX), 1)
 	ifeq ($(IS_AMD64), 1)
-		test_compilers += v8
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -293,8 +293,17 @@ space := $() $()
 comma := ,
 build_wasmer_extra_features_csv = $(subst $(space),$(comma),$(build_wasmer_extra_features))
 
+test_compilers := $(compilers)
```
marxin, line 296: Currently a temporary solution - we should enable V8 for all targets we currently support.

marxin:
Superseded by #6635.

PR #6613, opened 2026-05-21
re-enable auto discovery of WASIX integration tests, clean up the tes…

Arshia001:
…t folder

I also moved the `wasm_tests` code around a bit. It's a simple move of the two files, but shows up in git as big changes because the new file names clash with the old ones.


marxin (changes requested):
To be honest, I don't like the moving of the existing code as it seems unnecessary for the test discovery. What about doing just:

```patch
diff --git i/lib/wasix/Cargo.toml w/lib/wasix/Cargo.toml
index 9b7f492dca0..50a200562f4 100644
--- i/lib/wasix/Cargo.toml
+++ w/lib/wasix/Cargo.toml
@@ -11,7 +11,6 @@ homepage.workspace = true
 license.workspace = true
 repository.workspace = true
 rust-version.workspace = true
-autotests = false

 [[test]]
 name = "wasm_tests"
```

and deleting: `lib/wasix/tests/wasix-wasm.rs` and `lib/wasix/tests/socket_wasm.rs`?

Arshia001:
@marxin tbh, I never quite liked how we have the code for wasm_tests split over multiple places, makes it hard to discover... I can move the code around a bit more to make the change cleaner, but I insist we do this change.

marxin:
You are right - that said, can we then combine both of the existing files (`wasm_tests.rs` and `mod.rs`) into the `mod.rs` file by not having an extra helper file?

Arshia001:
@marxin I did a bit of git magic to make it so the changes are properly recognized. Also, to me, it makes sense to keep the code in two separate modules, since there's 1k+ lines of it. wdyt?

marxin (approved):

PR #6617, opened 2026-05-21
Improve dependency unification + add 'package tree' command

theduke:
This PR does two things.

A) improve dependency unification

Improve the dependency resolver to unify dependencies, so overlapping
dependencies with slightly different version ranges are still accepted.

Previously, when separate dependencies pulled in the same package, but with 
different minimum versions, resolution would be rejected.

The new code instead properly fetches all relevant deps first, then unifies to
the most compatible versions.

B) Add 'package tree' CLI command

It shows the dependency tree of a package, including specified and resolved
versions.

### DIFF
--- a/lib/cli/src/commands/package/tree.rs
+++ b/lib/cli/src/commands/package/tree.rs
@@ -149,10 +149,17 @@ fn format_dependency(alias: &str, dependency: &Dependency, resolved_id: &Package
             .map_or("*".to_string(), |tag| tag.to_string());
 
         if is_fixed_to_resolved(&dependency.pkg, resolved_id) {
-            return format!("{alias}: @{specified_version}");
+            return format!(
+                "{alias}: {}@{specified_version}",
+                aliased_package_name(alias, &specified.full_name())
+            );
         }
 
-        return format!("{alias}: @{specified_version}=>{}", resolved.version);
+        return format!(
+            "{alias}: {}@{specified_version}=>{}",
+            aliased_package_name(alias, &specified.full_name()),
+            resolved.version
+        );
     }
 
     if is_fixed_to_resolved(&dependency.pkg, resolved_id) {
@@ -160,6 +167,10 @@ fn format_dependency(alias: &str, dependency: &Dependency, resolved_id: &Package
     } else {
         format!("{alias}: {} -> {resolved_id}", dependency.pkg)
     }
+}
+
+fn aliased_package_name<'a>(alias: &str, full_name: &'a str) -> &'a str {
+    if alias == full_name { "" } else { full_name }
 }
 
 fn is_fixed_to_resolved(specified: &PackageSource, resolved_id: &PackageId) -> bool {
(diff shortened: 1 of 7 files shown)


lib/wasix/src/runtime/resolver/resolve.rs
```diff
@@ -81,30 +92,94 @@ fn print_cycle(packages: &[PackageId]) -> String {
...
+    // First try a fast fixpoint pass that merges obviously compatible versions.
+    let heuristic = discover_merged_dependencies(root_id, root, source).await?;
+    if discovered_packages_are_unified(&heuristic) {
+        return Ok(heuristic);
+    }
+
+    // If the greedy pass still leaves duplicate named packages behind, search
+    // alternate version choices so lower direct versions can win when they are
+    // the only way to keep the full graph consistent.
+    if let Some(discovered) = discover_dependencies_with_backtracking(root_id, root, source).await?
```
marxin, line 141: In general, the suggested approach seems reasonable. Am I correct that the backtracking approach essentially tries all valid combinations, with early exits when a selected dependency version can no longer satisfy another dependency in the graph?

lib/wasix/src/runtime/resolver/resolve.rs
```diff
@@ -8,16 +8,20 @@ use petgraph::{
...
+    Dependency, DependencyGraph, ItemLocation, PackageInfo, PackageSummary, QueryError, Resolution,
     ResolvedPackage, Source,
     outputs::{Edge, Node},
 };
 
 use super::ResolvedFileSystemMapping;
 
+const MAX_DISCOVERED_PACKAGES: usize = 1_000;
+const MAX_PENDING_DEPENDENCIES: usize = 4_000;
+const MAX_BACKTRACK_STATES: usize = 10_000;
```
marxin, line 23: Might be interesting to construct an artificial test case with a deep dependency tree where each node has many incompatible versions. I’m curious what the elapsed time would look like in that scenario.

marxin (approved):
Good job, looks good to me. Maybe, I would consider using a different marker for the version assignment:

```
wasmer/cli -> wasmer/cli@0.1.3
|-- curl/curl: @*=>8.4.0+build.05
|-- wasmer/bash: @*=>1.0.25
|   `-- wasmer/coreutils: @^1.0.19=>1.0.25
|-- wasmer/find: @*=>4.10.0
```

note the `>=` has its own meaning in the context of version specifiers (greater or equal than).

lib/wasix/src/runtime/resolver/resolve.rs
theduke: Yes, essentially, it's a relatively simple approach.

Note: dependency resolution is a pretty big topic.... can of worms.
This meaningfully extends support without going too deep.

lib/wasix/src/runtime/resolver/resolve.rs
theduke: Added a bit of a stress test case.

lib/wasix/src/runtime/resolver/resolve.rs
marxin: I know it's very hard problem on its own :)

theduke:
@marxin suggestion for a different separator/format? For the "specified => resolved" version?

marxin:
> @marxin suggestion for a different separator/format? For the "specified => resolved" version?

Tried looking at what `cargo tree` and `uv tree` do and it seems they just present the resolved version in the output:
```
❯ uv tree
Resolved 2 packages in 0.47ms
foo v0.1.0
└── termcolor v3.3.0
...
❯ cargo tree
benchmark-runner v0.1.0 (/home/marxin/Programming/wild/benchmarks/runner)
├── anyhow v1.0.102
├── clap v4.5.60
│   ├── clap_builder v4.5.60
│   │   ├── anstream v0.6.21
│   │   │   ├── anstyle v1.0.13
```

so what about presenting the version specifier in parenthesis by using a different color. Anyway, it's just a nit, leaving that up to you if you want to invest more time into it.

marxin:
@theduke the PR is basically ready to be merged - can you polish the nits and merge it?

lib/cli/src/commands/package/tree.rs
```diff
@@ -0,0 +1,195 @@
...
+            .tag
+            .as_ref()
+            .map_or("*".to_string(), |tag| tag.to_string());
+
+        if is_fixed_to_resolved(&dependency.pkg, resolved_id) {
+            return format!("{alias}: @{specified_version}");
+        }
+
+        return format!("{alias}: @{specified_version}=>{}", resolved.version);
+    }
```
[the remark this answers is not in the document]
theduke: Fixed

Issue #6121, opened 2026-01-27
`wasmer run --v8 cowsay Meh` crashes (Fatal error (v8): current thread is different from the thread the store was created in!)


marxin:
```
Thread 45 "tokio-runtime-w" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff807f86c0 (LWP 2489780)]
0x00007ffff7d8600d in ?? () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff7d8600d in ?? () from /usr/lib/libc.so.6
#1  0x000055555a630b9d in v8::internal::wasm::WasmSerializer::SerializeNativeModule(v8::base::Vector<unsigned char>) const ()
#2  0x00005555599e3393 in wasm::Module::serialize() const ()
#3  0x00005555599e862f in wee8_wasm_module_serialize ()
#4  0x000055555996fd09 in wasmer::backend::v8::entities::module::ModuleHandle::serialize (self=0x7fff84094040) at lib/api/src/backend/v8/entities/module.rs:116
#5  0x00005555599723c3 in wasmer::backend::v8::entities::module::Module::serialize (self=0x7fff807f68e8) at lib/api/src/backend/v8/entities/module.rs:207
#6  0x000055555997be47 in wasmer::entities::module::inner::BackendModule::serialize (self=0x7fff807f68e0) at lib/api/src/entities/module/inner.rs:261
#7  0x00005555599b21f6 in wasmer::entities::module::Module::serialize (self=0x7fff807f68e0) at lib/api/src/entities/module/mod.rs:179
#8  0x00005555584c7ffa in wasmer_wasix::runtime::module_cache::filesystem::tokio_save::{async_fn#0}::{async_block#0}::{closure#0} ()
```

marxin:
Seems the original issue can be _mitigated_ with `--disable-cache`, however, anything non-trivial does not work either:
```
thread 'TokioTaskManager Thread Pool_thread_1' (3127254) panicked at lib/api/src/backend/v8/entities/mod.rs:18:9:
Fatal error (v8): current thread is different from the thread the store was created in!
stack backtrace:
```

Arshia001:
@marxin one more thing we probably don't have yet is support for the async runtime and `Function::call_async*`. This is necessary for WASIX to run properly, and will impact running Python codebases.

[...]

PR #6621, opened 2026-05-22
fix(V8): address various limitations

### LINKED ISSUE #6559: v8: support shared libraries
As of now, the Python package cannot be run with `--v8`:
```
     Running `target/debug/wasmer run --disable-cache --v8 python/python -- --version`

thread 'TokioTaskManager Thread Pool_thread_1' (33194) panicked at lib/api/src/backend/v8/entities/instance.rs:110:40:
Export ImportType { module: "env", name: "_ZTH5errno", ty: Function(FunctionType { params: [], results: [] }) } not found
```

```
$ wasm-tools print python
...
(module
  (@dylink.0
    (mem-info (memory 3753556 6) (table 7089 0))
    (needed "libsqlite3.so" "libssl.so" "libcrypto.so" "libcrypto.so")
...
    (import-info "env" "_ZTH5errno" binding-weak undefined)
  )
```

@Arshia001 do you know what would it take to support the feature with V8?

marxin:
The PR addresses several V8 limitations. First, it includes a small fix for `grow_at_least`, which previously mixed byte sizes and page sizes; this is now resolved through a new `Pages` method.

It also adds support for parsing and preserving the `dylink.0` section. In addition, I introduced a new `V8ModuleInfo` structure (used for serialization/deserialization), where we now explicitly copy imports and exports. The motivation for this change is that I encountered crashes when `wasm_module_exports` / `wasm_module_imports` were accessed from another thread during process and thread creation.

Also, I noticed that the `place_in_function_table_at` change is required; otherwise, I hit an assertion failure in the debug V8 build.

With the provided changes, I can run `python/python` package.

Issue: #6121
Fixes: #6559

### DIFF
--- a/lib/api/src/backend/v8/entities/module.rs
+++ b/lib/api/src/backend/v8/entities/module.rs
@@ -12,6 +12,7 @@ use wasmer_types::{
     GlobalType, ImportType, ImportsIterator, MemoryType, ModuleInfo, Mutability, Pages,
     SerializeError, TableType, Type,
 };
+use wasmparser::{Parser, Payload};
 
 #[derive(Debug)]
 pub(crate) struct ModuleHandle {
@@ -132,11 +133,16 @@ impl Drop for ModuleHandle {
     }
 }
 
+const DYLINK_SECTION_NAME: &str = "dylink.0";
+
 #[derive(Clone, PartialEq, Eq)]
 /// A WebAssembly `module` in the `v8` runtime.
 pub struct Module {
     pub(crate) handle: Arc<ModuleHandle>,
     name: Option<String>,
+    // Copy of the section data needed by the dynamic linker, since the current
+    // API cannot retrieve it later from the handle.
+    dylink_section_data: Option<Vec<u8>>,
 }
 
 unsafe impl Send for Module {}
@@ -165,10 +171,12 @@ impl Module {
         let info = crate::utils::polyfill::translate_module(&binary[..])
             .unwrap()
             .info;
+        let dylink_section_data = get_dylink_section_data(&binary)?;
 
         Ok(Self {
             handle: Arc::new(module),
             name: info.name,
+            dylink_section_data,
         })
     }
 
@@ -227,10 +235,12 @@ impl Module {
         let name = String::from_utf8_lossy(name_bytes).to_string();
         let mod_bytes = &binary[(8 + off)..];
         let module = ModuleHandle::deserialize(engine, mod_bytes)?;
+        let dylink_section_data = get_dylink_section_data(mod_bytes)?;
 
         Ok(Self {
             handle: Arc::new(module),
             name: Some(name),
+            dylink_section_data,
         })
     }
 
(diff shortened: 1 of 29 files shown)


lib/api/src/backend/v8/entities/module.rs
```diff
@@ -132,11 +135,27 @@ impl Drop for ModuleHandle {
     }
 }
 
+const DYLINK_SECTION_NAME: &str = "dylink.0";
```
Arshia001: Why are we only parsing out the dylink.0 section? I know this is the only thing we use internally in WASIX, but client code may as well need to look at other sections as well. If memory usage is the concern, we can probably skip well-known debug sections and store everything else; I don't assume there'll be a lot of custom sections to keep around anyway.

lib/api/tests/module.rs
```diff
@@ -280,8 +280,11 @@ fn calling_host_functions_with_negative_values_works() -> Result<(), String> {
 }
 
 #[engine_test]
+#[cfg_attr(
```
Arshia001: I think this also skips the test for other backends if V8 is enabled, no?

lib/wasix/src/state/linker/instance_group/table.rs
```diff
@@ -82,7 +82,9 @@ impl InstanceGroupState {
                 "Growing indirect function table"
             );
             table.grow(store, delta, Value::FuncRef(None))?;
-        } else {
+        } else if !table.is_v8() {
```
Arshia001: If we're going to claim WASIX works with v8 (sans the async bits) after we're done, we need to investigate this some more. This condition can end up breaking DL modules on V8 is highly spectacular fashion.

Arshia001 (dismissed):

lib/api/src/backend/v8/entities/module.rs
```diff
@@ -380,14 +298,35 @@ impl Module {
...
+        if name == DYLINK_SECTION_NAME {
+            Box::new(
+                self.info
+                    .dylink_section_data
+                    .iter()
+                    .map(|data| data.clone().into_boxed_slice()),
+            )
+        } else {
+            Box::new(std::iter::empty())
+        }
```
syrusakbary: This seems a hack. I'd serialize all custom sections inside of the V8ModuleInfo (not just DYLINK_SECTION_NAME), so then we can retrieve any of them easily after.

lib/api/src/backend/v8/entities/module.rs
```diff
@@ -132,11 +135,27 @@ impl Drop for ModuleHandle {
...
+    name: Option<String>,
+    // Cache imports and exports instead of calling `wasm_module_imports` or
+    // `wasm_module_exports`: a `Module` can be queried from a different thread
+    // (store) than the one V8 associated with the module, which can crash.
+    imports: Vec<ImportType>,
+    exports: Vec<ExportType>,
+    v8_module_bytes: Vec<u8>,
+    // Copy of the section data needed by the dynamic linker, since the current
+    // API cannot retrieve it later from the handle.
+    dylink_section_data: Option<Vec<u8>>,
```
syrusakbary: Rather than having only dylink_section_data I'd just have a section_data being a HashMap of name -> &[u8].

Note: why &[u8]? Because we have already the whole module stored in `v8_module_bytes`, so we can have a reference (or a offset to those bytes)

lib/api/src/backend/v8/entities/module.rs
syrusakbary: Yeah, exactly. I added a comment about this also.

lib/wasix/src/state/linker/instance_group/table.rs
marxin: Yep, you are right. The investigation payed off - have a proper fix: https://github.com/wasmerio/v8-custom-builds/pull/14.

lib/api/src/backend/v8/entities/module.rs
marxin: Yes - I decided to ignore other sections because of the memory constraints, but I must admit it's weird given `custom_sections` function should return all of them.

lib/api/src/backend/v8/entities/module.rs
marxin: Yes - let's do that.

lib/api/src/backend/v8/entities/module.rs
marxin: Note we can't make any assumption about the `v8_module_bytes` as it's the internal representation of the module built by V8. We don't even know if the custom section are kept, probably not.

lib/api/tests/module.rs
marxin: Yes are right. Anyway, it's gonna be unnecessary as we'll preserve all the custom sections.

syrusakbary (approved):

PR #6614, opened 2026-05-21
Add regression test for ERRORCHECK mutexes becoming immediately NOTRE…

Arshia001:
…COVERABLE (see wasix-org/wasix-libc#110)

Note this PR will only turn green after the PR mentioned above is merged in and a new wasix-libc release is created.

PR #6622, opened 2026-05-22
Enable WASIX tests on macOS LLVM

sadhbh-c0d3:
Allow WASIX tests to run on macOS.

See also https://github.com/wasix-org/wasix-libc/pull/112

~This is enabled with explicit command line option: `--enable-macos-wasm-tests`, e.g.~

~cargo test --test wasm_tests -- --enable-macos-wasm-tests wasm/context_switching/contexts_with_signals~

This is enabled with explicit environment variable: `WASMER_ENABLE_MACOS_WASM_TESTS=1`.

Rationale: Currently WASIX tests are disabled on macOS, and this is to allow them to run.

### DIFF
--- a/lib/wasix/tests/wasm_tests/README.md
+++ b/lib/wasix/tests/wasm_tests/README.md
@@ -48,9 +48,9 @@ Cargo and nextest filtering both work. Before running the suite, make sure
 `wasixcc` is installed and available in your shell environment.
 
 On macOS, this suite is opt-in because Cranelift exception-handling support is
-still incomplete there. Pass `--enable-macos-wasm-tests` after Cargo's `--` to
-collect and run the macOS-supported LLVM variants:
+still incomplete there. Set `WASMER_ENABLE_MACOS_WASM_TESTS=1` to collect and
+run the macOS-supported LLVM variants:
 
 ```sh
-cargo test --test wasm_tests -- --enable-macos-wasm-tests wasm/context_switching/contexts_with_signals
+WASMER_ENABLE_MACOS_WASM_TESTS=1 cargo test --test wasm_tests wasm/context_switching/contexts_with_signals
 ```
--- a/lib/wasix/tests/wasm_tests/README.md
+++ b/lib/wasix/tests/wasm_tests/README.md
@@ -46,3 +46,11 @@ the primary source, for example `//#BuildEnv: WASIXCC_PIC=1` in C/C++ sources or
 These tests run through the normal `wasix` integration test target, so standard
 Cargo and nextest filtering both work. Before running the suite, make sure
 `wasixcc` is installed and available in your shell environment.
+
+On macOS, this suite is opt-in because Cranelift exception-handling support is
+still incomplete there. Pass `--enable-macos-wasm-tests` after Cargo's `--` to
+collect and run the macOS-supported LLVM variants:
+
+```sh
+cargo test --test wasm_tests -- --enable-macos-wasm-tests wasm/context_switching/contexts_with_signals
+```
--- a/lib/wasix/tests/wasm_tests/README.md
+++ b/lib/wasix/tests/wasm_tests/README.md
@@ -49,7 +49,3 @@ Cargo and nextest filtering both work. Before running the suite, make sure
 
 On macOS, this suite collects and runs the LLVM variants only because Cranelift
 exception-handling support is still incomplete there:
-
-```sh
-cargo test --test wasm_tests wasm/context_switching/contexts_with_signals
-```
--- a/lib/wasix/tests/wasm_tests/README.md
+++ b/lib/wasix/tests/wasm_tests/README.md
@@ -47,10 +47,9 @@ These tests run through the normal `wasix` integration test target, so standard
 Cargo and nextest filtering both work. Before running the suite, make sure
 `wasixcc` is installed and available in your shell environment.
 
-On macOS, this suite is opt-in because Cranelift exception-handling support is
-still incomplete there. Set `WASMER_ENABLE_MACOS_WASM_TESTS=1` to collect and
-run the macOS-supported LLVM variants:
+On macOS, this suite collects and runs the LLVM variants only because Cranelift
+exception-handling support is still incomplete there:
 
 ```sh
-WASMER_ENABLE_MACOS_WASM_TESTS=1 cargo test --test wasm_tests wasm/context_switching/contexts_with_signals
+cargo test --test wasm_tests wasm/context_switching/contexts_with_signals
 ```
--- a/lib/wasix/tests/wasm_tests/mod.rs
+++ b/lib/wasix/tests/wasm_tests/mod.rs
@@ -55,8 +55,6 @@ use walkdir::WalkDir;
 
 mod runner;
 
-const ENABLE_MACOS_WASM_TESTS_ENV: &str = "WASMER_ENABLE_MACOS_WASM_TESTS";
-
 fn should_emit_colour() -> bool {
     std::io::stdout().is_terminal()
         || std::env::var("CARGO_TERM_COLOR").as_deref() == Ok("always")
@@ -64,7 +62,7 @@ fn should_emit_colour() -> bool {
 }
 
 fn main() -> Result<std::process::ExitCode> {
-    let mut args = libtest_mimic::Arguments::from_iter(std::env::args_os());
+    let mut args = libtest_mimic::Arguments::from_args();
     if should_emit_colour() {
         args.color = Some(libtest_mimic::ColorSetting::Always);
     }
@@ -585,9 +583,6 @@ fn identify_primary_sources(test_src_dir: &Path) -> Result<Vec<PrimarySource>> {
 fn collect_tests(tests: &mut Vec<Trial>) -> Result<()> {
     // Windows runtime support is still limited, so skip these tests on that platform.
     if cfg!(target_os = "windows") {
-        return Ok(());
-    }
-    if cfg!(target_os = "macos") && std::env::var_os(ENABLE_MACOS_WASM_TESTS_ENV).is_none() {
         return Ok(());
     }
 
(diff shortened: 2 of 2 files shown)


lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -621,9 +654,11 @@ fn collect_tests(tests: &mut Vec<Trial>) -> Result<()> {
 
             for config in configs {
                 for engine in [Engine::Cranelift, Engine::LLVM] {
-                    // The EH support for macOS is still missing: #6419
-                    if cfg!(target_os = "macos") {
-                        return Ok(());
+                    // Cranelift EH support for macOS is still missing: #6419.
+                    // Keep collecting LLVM trials so focused macOS filters do
+                    // not silently report zero tests.
+                    if cfg!(target_os = "macos") && engine == Engine::Cranelift {
```
marxin: Can we rather use an environment variant instead of the extra argument? Btw. does the extra argument work with `cargo nextest`?

marxin (changes requested):

lib/wasix/tests/wasm_tests/mod.rs
marxin: Similarly to: `tests/compilers/config.rs:        let debug_dir = std::env::var("WASMER_COMPILER_DEBUG_DIR")`

sadhbh-c0d3: Changed to env

marxin: Oh, now I got it. We unintentionally skipped `Engine::LLVM` on Mach-O. That's something we want to do unconditionally. So please drop the introduced env. variable and just keep this hunk.

Arshia001 (commented):
LGTM; will leave final approval to @marxin.

lib/wasix/tests/wasm_tests/mod.rs
sadhbh-c0d3: It's unconditional now - removed env.

sadhbh-c0d3:
This was branched off https://github.com/wasmerio/wasmer/pull/6614 so it contains the context switch test added.

lib/wasix/tests/wasm_tests/pthread_tests/mutex-errorcheck-notrecoverable/main.c
```diff
@@ -0,0 +1,53 @@
+// Regression test for issue where wasix-libc would set 0x3fffffff as the TID
```
marxin: This test-case should belong to a separate PR, unrelated to the test enablement.

lib/wasix/tests/wasm_tests/README.md
```diff
@@ -46,3 +46,10 @@ the primary source, for example `//#BuildEnv: WASIXCC_PIC=1` in C/C++ sources or
 These tests run through the normal `wasix` integration test target, so standard
 Cargo and nextest filtering both work. Before running the suite, make sure
 `wasixcc` is installed and available in your shell environment.
+
+On macOS, this suite collects and runs the LLVM variants only because Cranelift
+exception-handling support is still incomplete there:
+
+```sh
```
marxin: The note about `macOS` is fine, but please omit the `cargo test` example.

marxin (changes requested):

marxin:
And please merge main, that will unblock the failing CI.

lib/wasix/tests/wasm_tests/pthread_tests/mutex-errorcheck-notrecoverable/main.c
sadhbh-c0d3: Yes, but the thing is that if I enable macOS tests when I was working on this issue, and that's how I included changes from that other branch. When this https://github.com/wasmerio/wasmer/pull/6614 PR is merged, diff won't show that. Alternatively I would need to create new branch for enabling macOS tests, cherry-pick my changes from this branch, and create new PR, because I cannot change source branch in this PR. I cannot revert this change on this branch, because then https://github.com/wasmerio/wasmer/pull/6614 will be reverted. I think most efficient would be to get https://github.com/wasix-org/wasix-libc/pull/113 merged first into wasix-libc, then https://github.com/wasmerio/wasmer/pull/6614 to confirm changes work well, and then this PR.

marxin: Sorry, am pretty confused. If you remove the test from this PR, then we'll merge it pretty soon (it's almost the only blocker). Once done, we'll have #6614 (only adding a test), that will be green as soon as wasix-org/wasix-libc#113 gets merged. What's wrong with the merge order?

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -657,21 +657,60 @@ fn collect_tests(tests: &mut Vec<Trial>) -> Result<()> {
...
-                for engine in [
+            let supported_engines = [
                     Engine::Cranelift,
                     Engine::LLVM,
                     // TODO: enable once the WASIX tests are green with V8
-                    //#[cfg(feature = "v8")]
-                    //Engine::V8,
-                ] {
-                    // The EH support for macOS is still missing: #6419
-                    if cfg!(target_os = "macos") {
```
marxin, line 669: Why are we introducing so big change? I would like to see the original version, which was simple an one-liner: `if cfg!(target_os = "macos") && engine == Engine::Cranelift {`.

marxin (changes requested):

lib/wasix/tests/wasm_tests/README.md
sadhbh-c0d3: done

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -649,6 +649,18 @@ fn collect_tests(tests: &mut Vec<Trial>) -> Result<()> {
...
         let primary_sources = identify_primary_sources(entry.path())?;
 
+        let mut supported_engines = vec![Engine::LLVM];
+
+        // TODO: enable once the WASIX tests are green with V8
+        // #[cfg(feature = "v8")]
+        // supported_engines.push(Engine::V8);
+
+        // Cranelift EH support for macOS is still missing: #6419.
+        // Keep collecting LLVM trials so focused macOS filters do not silently report zero tests.
```
marxin: Can we omit the line? Seems to me natural we're running LLVM tests on macOS as they are properly supported by the target.

marxin (approved):
Fine with expect one misleading comment line. Thanks for working on that.

lib/wasix/tests/wasm_tests/pthread_tests/mutex-errorcheck-notrecoverable/main.c
sadhbh-c0d3: Problem is that the test in question comes from a different branch, and this branch was based on that other branch. There is no safe way of removing a commit from a branch, if you want that commit to be included in another pull request. If I reverted that commit on my branch, then merging this PR would result in reverting what other PR was proposing, should the other PR be merged, because if on my branch I'm reverting a commit from other branch, then I'm effectively cancelling that commint altogether also on other branch, i.e. if you try to merge that other branch after I reverted that commit on my branch, then merging of that other branch will be successful, but comit will be "successfully" removed, because my branch has removed it. That's why revert is bad idea. Rebase is possible, and this is what I actually did, but requires rewriting history, and effects of rebasing are undegfined behaviour if there is more than one contributor. I have put this to test, and because my change is on my fork, and there is no other contributors to my fork, I was able to safely replace history of my branch, and as a result commit in question is gone from this branch, and at the same it it is not reverted, so that the other PR, when it gets merged, won't be automatically reverted by this PR.

Nevertheless, both reverting and rebasing with rewriting history is unsafe practice leading to problematic git history and undefined behaviour of pull requests that dependent on it. To avoid it, best would be to create new branch, and cherry-pick changes that I want to include, which is what I actually did, but then I have forced this new branch do be history of this branch, which could have corrupted state if someone had cloned this branch.

Best approach in this case would be to recognise that effort required and risk involved does not justify doing it in first place. Once the other two PR's I have mentioned are merged to main, the commit in question would already be part of main, and this PR wouldn't show it in changed files. That approach would be safest and cheapest.

sadhbh-c0d3: Marking as resolved, since commit is gone now from this branch.

lib/wasix/tests/wasm_tests/mod.rs
sadhbh-c0d3: Not sure what are you referring to. After merging your change, I have refactored the loop to filter the supported engines based on current configuration and platform.

sadhbh-c0d3:
> Fine with expect one misleading comment line. Thanks for working on that.

I am only questioning whether effort justifies the value there. We're talking about test runner, which as long as it runs tests should be good enough.

lib/wasix/tests/wasm_tests/mod.rs
sadhbh-c0d3: Comment is here, so that someone who doesn't know that, has it explained for them.

sadhbh-c0d3: Reduced to first line

marxin: Thank you.

marxin:
> We're talking about test runner, which as long as it runs tests should be good enough.

Well, yes, but I don't want to introduce a new comment that is misleading. Thanks for addressing it.

marxin:
Note we've got a way (`macos` PR label) how to run optional macOS CI jobs (there are enabled only on `main` for the resource reasons). Can you please make a git commit change (can be empty commit) so that the Tests are re-created and run also for macOS.

PR #6624, opened 2026-05-22
Err if WASIX task ID allocation exceeds the supported TID range

sadhbh-c0d3:
WASIX reserves special high values in guest-visible thread/mutex state, including the main-thread TID `0x3fffffff`. 

Generated task IDs should stay in the documented host-allocated range `1..=0x1fffffff`, so this adds an explicit upper bound to allocation and returns `TaskLimitReached` once that range is exhausted.

Also adds a focused unit test for the boundary case.

### DIFF
--- a/lib/wasix/src/os/task/control_plane.rs
+++ b/lib/wasix/src/os/task/control_plane.rs
@@ -174,16 +174,12 @@ impl WasiControlPlane {
 impl MutableState {
     fn next_process_id(&mut self) -> Result<WasiProcessId, ControlPlaneError> {
         // TODO: reuse terminated ids, handle wrap-around, ...
-        let id = self.process_seed.checked_add(1).ok_or({
-            ControlPlaneError::TaskLimitReached {
-                max: MAX_WASI_TASK_ID as usize,
-            }
-        })?;
-        if id > MAX_WASI_TASK_ID {
+        if self.process_seed == MAX_WASI_TASK_ID {
             return Err(ControlPlaneError::TaskLimitReached {
                 max: MAX_WASI_TASK_ID as usize,
             });
         }
+        let id = self.process_seed.checked_add(1).expect("Invalid arithmetic operation");
         self.process_seed = id;
         Ok(WasiProcessId::from(id))
     }
--- a/lib/wasix/src/os/task/control_plane.rs
+++ b/lib/wasix/src/os/task/control_plane.rs
@@ -10,6 +10,8 @@ use std::{
 use crate::{WasiProcess, WasiProcessId};
 use wasmer_types::ModuleHash;
 
+const MAX_WASI_TASK_ID: u32 = 0x1fffffff;
+
 #[derive(Debug, Clone)]
 pub struct WasiControlPlane {
     state: Arc<State>,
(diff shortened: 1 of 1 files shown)


lib/wasix/src/os/task/control_plane.rs
```diff
@@ -174,9 +176,14 @@ impl MutableState {
         // TODO: reuse terminated ids, handle wrap-around, ...
         let id = self.process_seed.checked_add(1).ok_or({
```
Arshia001: This is the wrong error to return; if `checked_add` fails, it means we've reached `u32::MAX`, not `MAX_WASI_TASK_ID`.

Arshia001 (changes requested):
This is the wrong change; if the end goal is to enforce the thread ID maximum of 0x1fffffff from wasix-libc, we should be limiting thread IDs, not process IDs.

lib/wasix/src/os/task/control_plane.rs
sadhbh-c0d3: I agree. I have changed the implementation. The important fact here is to understand that increment by 1 that would overflow an u32 is an impossible situation, and it is guaranteed by new form.

I first check if current seed is already =MAX, and if so I return error of known type to the caller (user of the function). Caller does not need to know implementation detail of the function itself, and instead caller has expectation to obtain new ID or TaskLimitReached error.

There is no situation in which increment by 1 would fail, and if it does then panic is fully justified within this implementation, and that is because implementation does check before that potential panic, and that check should prevent panic in 100% of cases. It would be a programming error if it didn't, and in such case panic is correct answer, i.e. code that is incorrect should panic.

sadhbh-c0d3:
> This is the wrong change; if the end goal is to enforce the thread ID maximum of 0x1fffffff from wasix-libc, we should be limiting thread IDs, not process IDs.

https://github.com/sadhbh-c0d3/wasmer/blob/680110b3ecf1dc4c7e41e8926e017e6150e705e9/lib/wasix/src/os/task/process.rs#L531

Arshia001:
Interesting... so all of this is to just limit the process IDs so the main thread doesn't get a weird ID? But there's a pending open question. We've looked at the main thread in wasix-libc, and it always gets a thread ID of `0x3fffffff`, correct? This is both outside the `[0x00000000, 0x1fffffff]` range (which is fine) and not tied to the PID.

The corresponding comment in wasix-libc also mentions that WASIX generated worker thread IDs in the `[0x00000000, 0x1fffffff]` range. However, I haven't seen this check anywhere. I think both comments may be out-of-date.

As a bonus, realistically, it's _very_ difficult to run into a situation where we have half a billion processes, so this change is done in the spirit of being super-pedantic. With that said, I feel we should get the semantics 100% right, or just drop the change.

sadhbh-c0d3:
@Arshia001 This PR is a bonus one. We assume that thread ID shouldn't be outside range 1..1fffffff.

I agree that running out of ID's would be unusual situation, but to be pragmatic since we have such unwritten assumption stated in the comment on TLS initialization, it's better to have such assertion.

The main thread ID = 3fffffff, which is deliberately chosen outside of the 1..1fffffff range, so it won't collide with any dynamically allocated ID.

There is no distinction between Process ID and Thread ID - same function is used - and as the consequence PID would share the assumption. Could be moved though closer to thread creation.

Having assertion in place can save us bugs, e.g. if the seed got corrupted from unsafe memory access.

This PR can be closed if we belive such assertion is unnecessary.

Arshia001:
Yes, I do believe that without some more research into the exact code, this PR alone doesn't achieve much. I'll close it.

PR #6625, opened 2026-05-22
test(wasix): port wasi-fyi suite to wasm harness

### LINKED ISSUE #6566: port wasi-fyi to wasmer-wasix tests
I think the tests are pretty similar to the WASIX tests:
- an input file (`rustc +nightly --target=wasm32-wasip1 -o "$output" "$input"`) is compiled to `.wasm` file
- a test is run with a provided `stdin`
- and expected `stdout`/`stderr` are sometimes expected

@Arshia001 is it something you can take a look once #6552 got merged?

sonukapoor:
This ports the complete wasi-fyi fixture suite into the shared wasmer-wasix wasm_tests harness, so the fixtures are discovered as normal directive-driven wasm tests and run per engine through libtest-mimic.

The shared harness now supports Rust primary sources, runtime env/stdin directives, stderr and file-backed output checks, program-name overrides, and disabling default mounts when a fixture needs explicit guest mappings. The wasi-fyi Rust sources carry those directives directly, including /hamlet and /fyi mappings, and the old standalone tests/wasi-fyi shell target plus the separate wasi_fyi.rs collector are removed.

Closes #6566.

Verified locally with:

```sh
cargo test -p wasmer-wasix --features sys --test wasm_tests wasi_fyi -- --nocapture
git diff --check origin/main..HEAD
```


lib/wasix/tests/wasm_tests/wasi_fyi.rs
```diff
@@ -0,0 +1,281 @@
...
+        return Ok(None);
+    };
+    let contents = String::from_utf8(contents)?;
+    let envs = contents
+        .lines()
+        .filter(|line| !line.trim().is_empty())
+        .map(|line| {
+            let (key, value) = line
+                .split_once('=')
+                .ok_or_else(|| anyhow::anyhow!("invalid env sidecar line: {line}"))?;
```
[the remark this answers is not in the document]
sonukapoor: Env sidecar UTF-8 and invalid-line errors now include the sidecar path.

lib/wasix/tests/wasm_tests/wasi_fyi.rs
```diff
@@ -0,0 +1,281 @@
+use std::fs;
+use std::os::unix::fs::symlink;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+use wasmer_wasix::runners::MappedDirectory;
+use wasmer_wasix::virtual_fs::StaticFile;
+
```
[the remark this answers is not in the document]
sonukapoor: Symlink copying is now cfg-gated so this module compiles on non-Unix hosts; the wasi-fyi suite remains skipped on Windows.

lib/wasix/tests/wasm_tests/wasi_fyi.rs
```diff
@@ -0,0 +1,281 @@
...
+                runner.with_args(args);
+            }
+
+            if let Some(envs) = read_envs(&fixture_dir, fixture).unwrap() {
+                runner.with_envs(envs);
+            }
+
+            if let Some(stdin) = read_sidecar(&fixture_dir, fixture, "stdin").unwrap() {
+                runner.with_stdin(Box::new(StaticFile::new(stdin)));
+            }
```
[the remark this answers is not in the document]
sonukapoor: Sidecars are now read before runner setup and propagated through the test result with fixture-specific context.

lib/wasix/tests/wasm_tests/wasi_fyi.rs
```diff
@@ -0,0 +1,281 @@
...
+    };
+    let expected = String::from_utf8(expected)?.trim().parse::<i32>()?;
+    let actual = result.exit_code.unwrap_or_default();
+    anyhow::ensure!(
+        actual == expected,
+        "{} should exit with code {}\n{}",
+        fixture,
+        expected,
+        super::format_captured_output(result),
+    );
```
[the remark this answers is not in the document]
sonukapoor: The current branch uses the runner concrete exit code here; unknown exits use the runner sentinel and fail the equality check instead of defaulting to success.

lib/wasix/tests/wasm_tests/wasi_fyi.rs
```diff
@@ -0,0 +1,281 @@
...
+wasi_fyi_test!(process_exit_120, "process_exit-120");
+
+fn run_wasi_fyi_test(fixture: &str) -> Result<(), anyhow::Error> {
+    let fixture_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("tests/wasm_tests/wasi_fyi");
+    let wasm_dir = tempfile::tempdir()?;
+    let wasm_path = wasm_dir.path().join("main");
+    let test_fs_dir = wasm_dir.path().join("test_fs");
+
+    copy_dir_all(fixture_dir.join("test_fs"), &test_fs_dir)?;
+    compile_rust_fixture(&fixture_dir, fixture, &wasm_path)?;
```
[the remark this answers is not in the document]
sonukapoor: Fixture compilation is now serialized and cached per fixture within the test process. The mutable test_fs copy remains per invocation to preserve isolation for filesystem tests.

marxin (changes requested):
Thank you for working on that, but I don't like the introduction of a new test harness (`wasi_fyi.rs`), I would rather like seeing extending the existing `wasm_tests/mod.rs`.

> The fixtures keep their existing sidecar expectations for args, env, stdin, stdout, stderr, and exit status.

Please migrate the sidecar expectation to the directive annotation directly to the source files.

> The runner maps the host directories that wasi-fyi expects at /hamlet and /fyi, and the old standalone tests/wasi-fyi shell target is removed now that the suite is covered by wasm_tests.

This might be tricky a bit, but I believe the `MappedDirectory` can be easily extended to our needs for `/hamlet` and `/fyi` folders.

sonukapoor:
Reworked this to use the shared `wasm_tests` harness instead of a separate `wasi_fyi.rs` collector.

The wasi-fyi Rust sources are now discovered as Rust primary sources, and the former arg/env/status sidecars are represented as source directives. `/hamlet` and `/fyi` are mapped through `MappedDirectory` directives, with default mounts disabled for these fixtures.

Local verification:

```sh
cargo test -p wasmer-wasix --features sys --test wasm_tests wasi_fyi -- --nocapture
git diff --check origin/main..HEAD
```

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -384,13 +461,11 @@ fn run_build_script(config: &Config) -> anyhow::Result<PathBuf> {
...
     create_dir_all(&build_test_path)?;
 
-    let mut options = fs_extra::dir::CopyOptions::new();
-    options.content_only = true;
-    fs_extra::dir::copy(&config.test_src_dir, &build_test_path, &options).with_context(|| {
-        anyhow!(
-            "cannot copy {:?} to the temporary directory: {:?}",
-            config.test_src_dir,
-            &build_test_path,
+    copy_dir_all(&config.test_src_dir, &build_test_path).with_context(|| {
```
marxin: Do we know why is the previous variant not sufficient?

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -298,6 +336,19 @@ fn process_directive(
         "ExpectedStdout" => {
             config.expected_stdout.push(arg.to_owned());
         }
+        "ExpectedStderr" => {
+            config.expected_stderr.push(arg.to_owned());
+        }
+        "ExpectedStdoutFile" => {
+            config
```
marxin: What about mapping this and `ExpectedStderrFile` by filling up `config.expected_stdout.extend(... read lines from file)`?

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -516,17 +683,31 @@ fn run_integration_test(config: Config) -> Result<libtest_mimic::Completion> {
...
+        assert_expected_lines(
+            &config,
+            "stdout",
+            &config.expected_stdout,
+            &result.stdout,
+            &result,
+        )?;
+    }
+
+    if !config.expected_stderr.is_empty() {
```
marxin, line 746: Here it's going to simplify as well (no extra case for stdout_files and stderr_files.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -546,6 +727,48 @@ fn run_integration_test(config: Config) -> Result<libtest_mimic::Completion> {
     Ok(libtest_mimic::Completion::Completed)
 }
 
+fn assert_expected_lines(
```
marxin: I would not introduce 2 new functions given the previous suggestion.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -580,10 +810,14 @@ impl PrimarySource {
 
     fn is_default(&self) -> bool {
         match self {
-            Self::SourceFile(_) => true,
+            Self::SourceFile(filename) => matches!(filename.as_str(), "main.c" | "main.cpp"),
             Self::BashScript(filename) => filename == "build.sh",
         }
     }
+
+    fn is_rust(&self) -> bool {
```
marxin: I am thinking if `Self::RustSourceFile`, `Self::CSourceFile` and `Self::CppSourceFile` would not be a better option?

lib/wasix/tests/wasm_tests/README.md
```diff
@@ -13,16 +13,21 @@ as a test fixture:
...
-arguments, expected output, expected exit status, mapped directories, prefilled files, and file checks.
+arguments, runtime environment, stdin, expected output, expected exit status, mapped directories, prefilled files, and file checks.
 The supported directives are documented in [`../wasm_tests.rs`](../wasm_tests.rs).
 
 If a fixture has more than one `.sh` file, each shell file is treated as a
-primary source, where`build.sh` is the default shell source name.
+primary source, where `build.sh` is the default shell source name. If a fixture
+has more than one `.rs` file, each Rust file is treated as a primary source.
+
+The former `tests/wasi-fyi` shell suite now lives in [`wasi_fyi/`](./wasi_fyi/)
```
marxin, line 27: No need for documenting it.

lib/wasix/tests/wasm_tests/README.md
```diff
@@ -46,3 +52,4 @@ the primary source, for example `//#BuildEnv: WASIXCC_PIC=1` in C/C++ sources or
 These tests run through the normal `wasix` integration test target, so standard
 Cargo and nextest filtering both work. Before running the suite, make sure
 `wasixcc` is installed and available in your shell environment.
+The `wasi_fyi` tests also require the Rust `wasm32-wasip1` target because they compile Rust fixtures to WebAssembly.
```
marxin: I would rather mention that a `.rs` test requires the cross `rustc` compiler installed.

lib/wasix/tests/wasm_tests/runner.rs
```diff
@@ -302,17 +302,38 @@ pub(crate) struct WasmRunResult {
     pub(crate) error: Option<String>,
 }
 
-pub(crate) fn run_wasm_with_runner_config(
+pub(crate) fn run_wasm_with_runner_config_options(
```
marxin: Let's extend the current function - no need for a wrapper.

marxin (changes requested):
Thank you for the effort - generally speaking looking very promising.

Have a couple more ideas:
- can we get rid of the `.gitignore` files?
- I would use `#ProgramName` only if required by a test
- looking at the repetitive directives for the individual files - what about introducing a concept of `#AbstractConfig: {path}` that will be defined in the folder once and then inherited by all test fixtures?

sonukapoor:
@marxin, Really appreciate your guidance here. I will take a look at everything today.

lib/wasix/tests/wasm_tests/wasi_fyi.rs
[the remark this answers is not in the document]
sonukapoor: Resolved by the harness rewrite: args/env/stdin/status metadata now lives in source directives, so this sidecar-reading path no longer exists.

lib/wasix/tests/wasm_tests/wasi_fyi.rs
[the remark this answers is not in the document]
sonukapoor: Resolved by the harness rewrite: the separate wasi_fyi.rs module and its Unix-only import were removed.

lib/wasix/tests/wasm_tests/wasi_fyi.rs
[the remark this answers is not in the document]
sonukapoor: Resolved by the harness rewrite: wasi-fyi fixtures now use the shared wasm_tests build/run path instead of a separate per-fixture runner.

lib/wasix/tests/wasm_tests/wasi_fyi.rs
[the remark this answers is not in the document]
sonukapoor: Resolved by the harness rewrite: env expectations now use source directives, so this .env sidecar parsing path no longer exists.

lib/wasix/tests/wasm_tests/wasi_fyi.rs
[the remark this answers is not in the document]
sonukapoor: Resolved by the harness rewrite: this separate file was removed, and exit status checks now run through the shared wasm_tests directive path.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: This is simplified now; output files are parsed into the same expected stdout/stderr path, so the runtime assertion has no separate file-output case.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: I retested the previous `fs_extra::dir::copy` path after the Rust-source discovery cleanup. It still fails before build for the `wasi_fyi` fixtures because the fixture tree contains symlinks. The local copy helper is only preserving files, directories, and symlinks so those fixtures can run from the shared build directory.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Changed this to feed `ExpectedStdoutFile` and `ExpectedStderrFile` into the existing expected-line vectors. I kept the expectation explicit even when the referenced file is empty, so empty expected output still asserts that the stream stays empty.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Changed the source enum to `CSourceFile`, `CppSourceFile`, and `RustSourceFile`.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Removed the separate output-file assertion helpers. The file directives now reuse the same expected-line comparison path.

lib/wasix/tests/wasm_tests/README.md
sonukapoor: Trimmed this down and removed the extra discovery detail from the README.

lib/wasix/tests/wasm_tests/runner.rs
sonukapoor: Removed the wrapper and extended the existing `run_wasm_with_runner_config` function instead.

lib/wasix/tests/wasm_tests/README.md
sonukapoor: Updated the README to mention that Rust `.rs` tests require `rustc` with the `wasm32-wasip1` target installed.

sonukapoor:
Addressed the follow-up cleanup and pushed the rebased branch. The `.gitignore` placeholders under `wasi_fyi` are gone, `ProgramName` remains only on the `env_args-*` fixtures that assert argv[0], and the common `/hamlet` / `/fyi` mapping setup now lives in `wasi_fyi/wasm_test_config` instead of being repeated in every source file.

I also retested the previous directory-copy path. It still fails on the symlink fixtures before build, so the local copy helper remains limited to preserving files, directories, and symlinks for the shared build directory.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -55,9 +72,12 @@ use std::str::FromStr;
 use anyhow::bail;
 use libtest_mimic::Trial;
 use walkdir::WalkDir;
+use wasmer_wasix::virtual_fs::StaticFile;
 
 mod runner;
 
+const DIRECTORY_CONFIG_FILE: &str = "wasm_test_config";
```
marxin: Can we please make it a new "AbstractConfigFile: wasi-fyi.config" directive instead (and explicitly listing it in all the FYI tests files)?

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -207,13 +243,41 @@ fn parse_configs(default_config: &Config) -> Result<Vec<Config>> {
     let mut config = default_config.clone();
     let mut build_env = Vec::new();
 
+    let directory_config = default_config.test_src_dir.join(DIRECTORY_CONFIG_FILE);
```
marxin: As mentioned, let's make it an explicit directive.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -296,7 +360,36 @@ fn process_directive(
                 .collect();
         }
         "ExpectedStdout" => {
-            config.expected_stdout.push(arg.to_owned());
+            config
+                .expected_stdout
+                .get_or_insert_with(Vec::new)
+                .push(arg.to_owned());
+        }
+        "ExpectedStderr" => {
```
marxin, line 344: Let's directly read the file content at this place and put the lines to the `config.expected_stderr` which should remain `Vec<String>`.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -404,26 +528,39 @@ fn run_build_script(config: &Config) -> anyhow::Result<PathBuf> {
...
+                PrimarySource::RustSourceFile(_) => "rustc".to_string(),
+                PrimarySource::BashScript(_) => unreachable!("handled above"),
             };
             let mut cmd = Command::new(&compiler);
-            cmd.arg(&primary_source)
-                .arg("-o")
-                .arg("main")
-                .current_dir(&build_test_path)
-                .env("WASIXCC_DISCARD_UNSUPPORTED_FLAGS", "yes");
+            if config.source.is_rust() {
```
marxin: I would remove `is_rust` and make it based on the previous match statement where we handle `RustSourceFile`.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -442,6 +579,53 @@ fn run_build_script(config: &Config) -> anyhow::Result<PathBuf> {
     Ok(build_test_path.join("main"))
 }
 
+fn copy_test_tree(from: &Path, to: &Path) -> Result<()> {
```
marxin, line 589: I’m curious whether there’s an existing Rust crate that already provides this functionality. To be honest, I’m not a big fan of the custom homegrown approach.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -463,40 +647,71 @@ fn run_integration_test(config: Config) -> Result<libtest_mimic::Completion> {
         File::create(run_dir.join(path))?.write_all(file_content.as_bytes())?;
     }
 
+    let program_name = config
+        .program_name
```
marxin: If the directive is missing, I would not overwrite the program name.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -463,40 +647,71 @@ fn run_integration_test(config: Config) -> Result<libtest_mimic::Completion> {
...
+        .clone()
+        .unwrap_or_else(|| wasm.to_string_lossy().to_string());
+    let stdin = match &config.stdin {
```
marxin: As mentioned, please read it at the time of the Directive parsing.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -607,17 +860,53 @@ fn identify_primary_sources(test_src_dir: &Path) -> Result<Vec<PrimarySource>> {
...
-            return Ok(vec![PrimarySource::SourceFile(file.to_string())]);
+            return Ok(vec![match file {
+                "main.c" => PrimarySource::CSourceFile(file.to_string()),
+                "main.cpp" => PrimarySource::CppSourceFile(file.to_string()),
+                _ => unreachable!("primary source file list is fixed"),
+            }]);
         }
     }
 
+    let rust_sources = read_dir(test_src_dir)?
```
marxin, line 870: Please add a comment that explains we allow multiple Rust source files in a same directory.

marxin (changes requested):
We're getting very close to the finish line - please address the remaining smaller issues.
Thank you for working on the task.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: I checked the existing dependency path again. `fs_extra::dir::copy` is already available in this crate, but it does not seem to preserve the wasi-fyi fixture tree correctly because these fixtures include symlinks, including broken symlinks used by the filesystem tests.

The local helper is intentionally narrow: it stages files, directories, and symlinks without following them, so the shared `wasm_tests` build directory matches the fixture tree closely enough for these tests.

Would you prefer keeping this small helper with that explicit scope, or should I look for a crate that preserves symlinks and use that instead?

marxin: All right, then I think the current function is fine. Thanks for investigating.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Kept the scoped copy helper. `fs_extra::dir::copy` was retested and does not preserve the symlink fixture tree correctly; the later thread converged on keeping the current function.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Addressed in the pushed update. Removed the implicit directory config lookup; shared FYI setup is loaded only through explicit `AbstractConfigFile` directives.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Addressed in the pushed update. Added `AbstractConfigFile: wasi-fyi.config` and listed it explicitly in all 67 wasi-fyi Rust fixtures.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Addressed in the pushed update. Removed `is_rust`; Rust compilation is handled by the `PrimarySource::RustSourceFile` match arm.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Addressed in the pushed update. `StdinFile` is read during directive parsing and stored as stdin bytes, so runtime setup no longer opens the fixture file.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Addressed in the pushed update. Expected output file directives now read their files during parsing and append lines into `config.expected_stdout` and `config.expected_stderr`, which remain `Vec<String>`.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Addressed in the pushed update. The config layer no longer synthesizes a program name; missing `ProgramName` is passed through as `None` to the runner.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Addressed in the pushed update. Added a comment before Rust source discovery explaining that multiple Rust files in one fixture directory are independent tests.

marxin (changes requested):
Can you please check the `.ignore` files that were moved - do we actually need them?

With that last missing bit, the PR can be merged. Thank you for your work.

sonukapoor:
Checked and pushed a small cleanup for the moved `.ignore` files.

Two placeholders were not needed and are removed: `fs_create_dir-non_existent.rs/.ignore` and `fs_rename-file.dir/.ignore`. I also emptied the old ignore-pattern contents from the remaining placeholders, so the remaining `.ignore` files only preserve fixture directories that need to exist before the test runs.

Verified locally with:

```sh
LLVM_SYS_221_PREFIX=<local LLVM prefix> cargo test -p wasmer-wasix --features sys --test wasm_tests wasi_fyi -- --nocapture
git diff --check
```

marxin (approved):
LGTM

PR #6626, opened 2026-05-22
Fix WASIX symlink resolution for path components

sadhbh-c0d3:
This fixes WASIX path resolution so symlinks are handled consistently during component traversal.

Changes:

- Resolve intermediate symlink components even when the final component should not be followed.
- Follow the final symlink only when `follow_symlinks` is enabled.
- Cache backing filesystem symlinks as `Kind::Symlink` entries.
- Keep ephemeral symlinks virtual and avoid caching them in directory entries.
- Split component discovery from inode creation/cache insertion so parent directory locks are not held while resolving preopen-relative symlink paths.
- Add explicit loop labels in the resolver to make traversal control flow clearer.


artemyarulin:
@sadhbh-c0d3 could you please rebase on top of the latest wasmer?

lib/wasix/src/fs/mod.rs
```diff
@@ -1135,101 +1135,147 @@ impl WasiFs {
...
+                continue 'path_iter;
             }
 
             // used to terminate symlink resolution properly
             let last_component = i + 1 == n_components;
             // for each component traverse file structure
             // loading inodes as necessary
-            'symlink_resolution: while symlink_count < MAX_SYMLINKS {
+            'component_lookup: loop {
+                enum ComponentResolution {
```
theduke: For readability, let's declare that outside the function.

lib/wasix/src/syscalls/wasix/chdir.rs
```diff
@@ -32,12 +32,29 @@ pub fn chdir<M: MemorySize>(
...
+        .fs
+        .root_fs
+        .read_dir(Path::new(&resolved_path))
+        .is_err()
+    {
         return Err(Errno::Noent);
     }
 
-    state.fs.set_current_dir(path);
+    state.fs.set_current_dir(&resolved_path);
```
theduke, line 86: We do not want to use the resolved path here, we want to use the originally specified path!
That aligns with how it works in Linux.

Also, as written, this introduces bugs/leaks, because the path stored in dirs can be the actual resolved host path for mounted directories!

lib/wasix/src/syscalls/wasix/chdir.rs
```diff
@@ -32,12 +32,29 @@ pub fn chdir<M: MemorySize>(
 
 pub fn chdir_internal(env: &WasiEnv, path: &str) -> Result<(), Errno> {
     let state = &env.state;
-
-    // Check if the directory exists
-    if state.fs.root_fs.read_dir(Path::new(path)).is_err() {
+    let path = state.fs.relative_path_to_absolute(path.to_string());
+    let inode = state
```
theduke, line 62: This line and the below read_dir() validation can be limited to the Kind::Dir branch below.

lib/wasix/src/fs/mod.rs
```diff
@@ -1272,135 +1315,159 @@ impl WasiFs {
...
+                        }
+                        // Special files cannot be traversed further, so return the inode directly.
+                        return Ok(new_inode);
                     }
                 }
-                break 'symlink_resolution;
+            }
+
+            // We continue with Symlink resolution...
+            if follow_symlinks || !last_component {
```
theduke: This can leave chained intermediate symlinks unresolved when follow_symlinks=false. 

Example: /a -> c, /c -> d, and /d/b exists. Resolving /a/b with
    follow_symlinks=false should still follow both a and c because they are non-final path components. The new recursive call resolves a to c with follow_symlinks=false, returns the final symlink c
    unresolved, then the outer loop advances to component b and returns/uses the c symlink instead of applying b under d. For longer paths it can consume one component while only resolving the
    pending symlink. This affects callers that intentionally pass false for final-link behavior, such as path_readlink and path_unlink_file.

theduke (changes requested):

lib/wasix/src/syscalls/wasix/chdir.rs
syrusakbary: I don't think the resolved path is the host resolved path @theduke . The resolved path seem to be the symlink resolved path (within the virtual fs)

artemyarulin:
# Shield - Regression 💩💩💩

| Language | Tests  | Pass rate now | PASS | FAIL/TIMEOUT | SKIP | CRASH |
| -------- | ------ | ------------- | ---- | ------------ | ---- | ----- |
| Python | 37,920 | 69.6% | $${\color{green}+39}$$ | $${\color{red}+480}$$ | -519 | 0 |
| Node.js | 4,861 | 33.8% | $${\color{green}+3}$$ | $${\color{green}-6}$$ | +3 | $${\color{red}+5}$$ |
| Rust | 22,337 | 94.3% | 0 | $${\color{green}-40}$$ | +40 | $${\color{green}-1}$$ |

### More changed tests

- Python: [tests_python_results.json](https://github.com/wasmerio/compat-tests/blob/aa1e11ea93f277902b494bfb817d085733f7b342/tests_python_results.json)
- Node.js: [tests_node_results.json](https://github.com/wasmerio/compat-tests/blob/aa1e11ea93f277902b494bfb817d085733f7b342/tests_node_results.json)
- Rust: [tests_rust_results.json](https://github.com/wasmerio/compat-tests/blob/aa1e11ea93f277902b494bfb817d085733f7b342/tests_rust_results.json)

## Install shield

- `git clone https://github.com/wasmerio/compat-tests.git`
- `cd compat-tests`
- `cargo build`
- `./target/debug/shield run --lang <LANG> --wasmer [WASMER BINARY] <TEST OR BATCH>`

## Artifacts

- GitHub Action: [https://github.com/wasmerio/anti-regression-shield/actions/runs/26491941723](https://github.com/wasmerio/anti-regression-shield/actions/runs/26491941723)
- Results commit: [https://github.com/wasmerio/compat-tests/commit/aa1e11ea93f277902b494bfb817d085733f7b342](https://github.com/wasmerio/compat-tests/commit/aa1e11ea93f277902b494bfb817d085733f7b342)

lib/wasix/src/syscalls/wasix/chdir.rs
sadhbh-c0d3: At the first glance it appears that we're setting current directory to resolved destination path, so if we were doing `cd A/B/C` and B -> /Z, then resolved path would be `/Z/C` if I understand correctly. Need to double-check.

Here is macOS behaviour:
```
% cd .next-bundle/functions/_not-found.func 
% pwd
/Users/sadhbh/src/ImperfectProtocol/private-poker/.next-bundle/functions/_not-found.func
% realpath .
/Users/sadhbh/src/ImperfectProtocol/private-poker/.next-bundle/functions/about.rsc.func 
```

@theduke  I think you are right.

lib/wasix/src/fs/mod.rs
sadhbh-c0d3: @theduke I think what you mean is that /A/B/C where B->/Z and C->Y should resolve to:
- follow=true => /Z/Y
- follow=false => /Z/C

There is a bug, but it's not `if follow_symlinks || !last_component` condition. See my comment below.

lib/wasix/src/fs/mod.rs
```diff
@@ -1272,135 +1315,159 @@ impl WasiFs {
...
+                    };
+                    (new_base_inode, new_path)
+                };
+                debug!("Following symlink recursively");
+                let symlink_inode = self.get_inode_at_path_inner(
+                    inodes,
+                    new_base_inode,
+                    &new_path,
+                    symlink_count + 1,
+                    follow_symlinks,
```
sadhbh-c0d3: This should be changed to: `follow_symlinks || !last_component`, because for /A/B/C where B->Z and C->Y:

- follow=true => /Z/Y
- follow=false => /Z/C

So for B we pass `!last_component` (always true), and for C `follow_symlinks`.

sadhbh-c0d3:
I have done refactor by hand and then I asked AI to check for bugs, and cross-check against compat-tests. Still need to run those to validate.

sadhbh-c0d3:
Shield results after fixes:

| Suite | Compared | Main | Yesterday | Current | Current vs Main | Current vs Yesterday |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `pathlib` | 1510 | 539/979 (55.06%) | 542/1497 (36.21%) | 958/976 (98.16%) | +419 passes | +416 passes |
| `os` | 367 | 132/217 (60.83%) | 139/217 (64.06%) | 175/215 (81.40%) | +43 passes | +36 passes |
| `tarfile` | 619 | 541/606 (89.27%) | 544/606 (89.77%) | 588/606 (97.03%) | +47 passes | +44 passes |
| `zipfile` | 336 | 248/331 (74.92%) | 250/331 (75.53%) | 251/331 (75.83%) | +3 passes | +1 pass |
| **TOTAL** | **2832** | **1460/2133 (68.45%)** | **1475/2651 (55.64%)** | **1972/2128 (92.67%)** | **+512 passes** | **+497 passes** |

```
pathlib  = test.test_pathlib
os       = test.test_os
tarfile  = test.test_tarfile
zipfile  = test.test_zipfile
```

Regressions vs main/yesterday: none verified.

artemyarulin:
# Shield - Improvement 🎉🎉🎉

| Language | Tests  | Pass rate now | PASS | FAIL/TIMEOUT | SKIP | CRASH |
| -------- | ------ | ------------- | ---- | ------------ | ---- | ----- |
| Python | 37,920 | 71.5% | $${\color{green}+774}$$ | $${\color{green}-782}$$ | +8 | 0 |
| Node.js | 4,861 | 33.9% | $${\color{green}+12}$$ | $${\color{green}-13}$$ | +1 | $${\color{red}+3}$$ |
| Rust | 22,337 | 94.3% | 0 | $${\color{green}-4}$$ | +4 | 0 |

- Examples from [tests_python_results.json](https://github.com/wasmerio/compat-tests/blob/687c687af2f6726def34a6b83c7681cc32cdd84c/tests_python_results.json):
  - `test.test_asyncio.test_base_events.BaseLoopSockSendfileTests.test__sock_sendfile_native_failure` (`FAIL -> PASS`)
  - `test.test_cprofile.CProfileTest.test_output_file_when_changing_directory` (`FAIL -> PASS`)
  - `test.test_genericpath.TestGenericTest.test_samefile_on_symlink` (`FAIL -> PASS`)
  - `test.test_genericpath.TestGenericTest.test_samestat_on_symlink` (`FAIL -> PASS`)
  - `test.test_glob.GlobTests.test_glob_symlinks` (`FAIL -> PASS`)


## Artifacts

- GitHub Action: [https://github.com/wasmerio/anti-regression-shield/actions/runs/26618653352](https://github.com/wasmerio/anti-regression-shield/actions/runs/26618653352)
- Results commit: [https://github.com/wasmerio/compat-tests/commit/687c687af2f6726def34a6b83c7681cc32cdd84c](https://github.com/wasmerio/compat-tests/commit/687c687af2f6726def34a6b83c7681cc32cdd84c)

marxin:
@sadhbh-c0d3 I'm going to address the Alpine issues in a separate PR (#6643).

sadhbh-c0d3:
> @sadhbh-c0d3 I'm going to address the Alpine issues in a separate PR (#6643).

But I already addressed them here

marxin:
> But I already addressed them here

The change in this PR is very convoluted, probably generated by LLM? We just need to install one more package, no need for a fancy detection.

lib/wasix/src/fs/mod.rs
```diff
@@ -1272,136 +1514,143 @@ impl WasiFs {
...
+                        } => {
+                            let new_inode = self.create_inode(inodes, kind, false, name)?;
+                            (entry_name, new_inode, !is_ephemeral, false)
+                        }
+                        ComponentResolution::BackingSymlink {
+                            file,
+                            link_value,
+                            entry_name,
+                        } => {
+                            let (pre_open_dir_fd, _) =
```
Arshia001: Discarding the relative path from `path_into_pre_open_and_relative_path` breaks the invariant expected by `Kind::symlink`, which specifically requires the `base_po_dir + path_to_symlink` pair to point to the correct path. I understand this PR is trying to represent symlinks with absolute paths (questionable choice in and of itself, but let's ignore that.) This still breaks the invariant and creates brittle coupling between this code and everything else that touches `Kind::Symlink`, and makes it very easy to introduce bugs into the code at a later time.

lib/wasix/src/fs/mod.rs
```diff
@@ -1485,6 +1734,122 @@ impl WasiFs {
...
+                .mount_entries()
+                .into_iter()
+                .filter(|entry| path_to_symlink.strip_prefix(&entry.path).is_ok())
+                .max_by_key(|entry| entry.path.as_os_str().len())
+                .map(|entry| entry.path)
+                .unwrap_or_else(|| PathBuf::from("/"));
+
+            let mut symlink_relative = path_to_symlink
+                .strip_prefix(&mount_path)
+                .unwrap_or_else(|_| path_to_symlink.strip_prefix("/").unwrap_or(path_to_symlink))
```
Arshia001: Why do we need to retry the prefix strip 3 separate times? Sounds to me like the first strip_prefix must always succeed, and if it doesn't, we're looking at an internal error. Covering up internal errors with retries is never a good idea and leads to hard-to-discover bugs later on.

lib/wasix/src/syscalls/wasi/path_symlink.rs
```diff
@@ -90,11 +90,15 @@ pub fn path_symlink_internal(
...
-    // Resolve symlink location to (preopen fd, relative path within that preopen)
-    // so runtime-created symlinks behave the same as symlinks discovered via readlink().
-    let (base_po_dir, path_to_symlink) = state
-        .fs
-        .path_into_pre_open_and_relative_path_owned(&symlink_path)?;
+    // Guest-created symlinks live in the virtual filesystem namespace. Keep
+    // their location relative to the virtual root so targets like
+    // `/temp/link -> ../hamlet/file` can cross sibling preopens without
+    // escaping the guest sandbox.
+    let base_po_dir = crate::VIRTUAL_ROOT_FD;
```
Arshia001: AFAICT, this PR uses two separate representations for symlinks:
* `VIRTUAL_ROOT_FD` + non-absolute, but otherwise full path, such as `hamlet/act3`
* absolute path, such as `/hamlet/act3`, in which case the base PO is ignored

we need to unify the code to use only one of these. My suggestion would be to remove the base_po field from `Kind::Symlink`, and always store `VIRTUAL_ROOT_FD`-relative paths, i.e. full path without leading slash, such as `hamlet/act3`.

lib/wasix/src/syscalls/wasi/path_rename.rs
```diff
@@ -254,9 +256,25 @@ pub fn path_rename_internal(
...
+                    if *base_po_dir == crate::VIRTUAL_ROOT_FD {
+                        (
+                            crate::VIRTUAL_ROOT_FD,
+                            host_adjusted_target_path
+                                .strip_prefix("/")
+                                .unwrap_or(host_adjusted_target_path.as_path())
+                                .to_owned(),
+                        )
+                    } else if path_to_symlink.is_absolute() {
+                        let (new_base_po_dir, _) =
```
Arshia001: Here we're doing the same "discard the relative path" logic, which results in the wrong `base_po+path` pair (see comment on `get_inode_at_path_inner`)

lib/wasix/src/fs/mod.rs
```diff
@@ -1272,136 +1514,143 @@ impl WasiFs {
...
-                                    );
-                                }
-                            }
-                            cur_inode = new_inode;
-
-                            if loop_for_symlink && follow_symlinks {
-                                debug!("Following symlink to {:?}", cur_inode);
-                                continue 'symlink_resolution;
-                            }
+                            (entry_name, new_inode, true, false)
```
Arshia001: Previously, we didn't cache symlinks here. This entire caching mechanism is the wrong choice because it makes underlying FS changes (such as a host-mapped directory changing) invisible to the WASIX process. However, I'd at least be inclined to keep the semantics we used to have, lest we introduce unforeseen regressions.

lib/wasix/src/fs/mod.rs
```diff
@@ -1485,6 +1734,122 @@ impl WasiFs {
...
+            match component {
+                Component::CurDir => {}
+                Component::ParentDir => {
+                    if !resolved.pop() {
+                        return Err(Errno::Perm);
+                    }
+                }
+                Component::Normal(component) => {
+                    remaining.push(component);
+                    preserve_remaining = true;
```
Arshia001: We're ignoring further `..` components embedded in the symlink, which can end up pointing to a path above our virtual root. If the intent is to validate that the path falls within our expected tree here, we should process the entire link, not just up to the first non-`..` component.

lib/wasix/src/syscalls/wasix/path_open2.rs
```diff
@@ -409,6 +454,60 @@ pub(crate) fn path_open_internal(
...
+                    }
+                    Err(_) => None,
+                };
+
+                if let Some((resolved_base_fd, resolved_path)) = final_symlink_target {
+                    if o_flags.contains(Oflags::EXCL) {
+                        return Ok(Err(Errno::Exist));
+                    }
+
+                    return path_open_internal(
```
Arshia001: While this change is correct from the perspective of POSIX compliance, a recursive call here can introduce a possible stack overflow given enough links on the backing FS. We should _at least_ track and limit the number of recursions. A better fix would be to budget this with the number of links traversed while resolving the path.

Arshia001 (changes requested):

lib/virtual-fs/src/host_fs.rs
```diff
@@ -138,6 +138,17 @@ impl crate::FileSystem for FileSystem {
         fs::create_dir(path).map_err(Into::into)
     }
 
+    fn hard_link(&self, source: &Path, target: &Path) -> Result<()> {
```
Arshia001: The host FS should not support creating links to avoid sandbox escape scenarios.

lib/wasix/src/syscalls/wasi/fd_readdir.rs
```diff
@@ -66,16 +66,17 @@ pub fn fd_readdir<M: MemorySize>(
...
-                            stat.st_ino,
-                        )
-                    },
-                ));
+                let entry_names: std::collections::HashSet<_> =
+                    entry_vec.iter().map(|(name, _, _)| name.clone()).collect();
+                entry_vec.extend(
+                    entries
+                        .iter()
+                        .filter(|(name, _)| !entry_names.contains(*name))
```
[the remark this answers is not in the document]
sadhbh-c0d3: The type of `name` is `&&String`, so `*name` is just `&String`

lib/wasix/src/fs/mod.rs
sadhbh-c0d3: I agree. I have now made symlink locations stored as paths relative to virtual root, and resolution is in one place.

lib/wasix/src/fs/mod.rs
sadhbh-c0d3: You're right. I have changed that, so symlink resolution now finds one longest matching mount prefix and treats strip failure as an internal mismatch instead of retrying through fallbacks.

lib/wasix/src/syscalls/wasi/path_symlink.rs
sadhbh-c0d3: Agree with you. Changed that as well. Symlinks location is now unified as relative to virtual root, and then SymlinkKind selects the policy for link resolution: guest-created or backing filesystem.

lib/wasix/src/syscalls/wasi/path_rename.rs
sadhbh-c0d3: This is fixed now. Rename now rebases the symlink location relative to virtual root instead of making it from base and path pair.

lib/wasix/src/fs/mod.rs
sadhbh-c0d3: Ok. So now symlink discoveries in backing filesystem are no longer cached inside parent directory entries. This makes sense, as it will allow changes made on the host side being reflected.

lib/wasix/src/fs/mod.rs
sadhbh-c0d3: Yeah, true. I changed that. We validate full relative symlink target path now.

lib/virtual-fs/src/host_fs.rs
sadhbh-c0d3: I checked. It doesn't. hard_link() was added for WASI path_link. HostFileSystem does not implement it directly, and default implementation returns Unsupported, so it is not possible for guest to create hard link inside host filesystem.

lib/wasix/src/syscalls/wasix/path_open2.rs
sadhbh-c0d3: Yes, it's improved now. Symlink path is resolved (path_open2.rs, line 353) before aquiring the write lock on fd_map (path_open2.rs, line 418), and also each reentry into path_open_internal_with_symlink_depth() increments symlink_depth, and then if it's greater thatn MAX_SYMLINKS, you'll get Errno::Loop returned.

sadhbh-c0d3:
@Arshia001 Code has been refactored to reduce mental overload, all guest path operations are now in one place, and rest of the code is using them. The results of `os`, `pathlib`, `zipfile`, and `tarfile` compat-tests are very good.

Here’s the branch-vs-main table using the documented comparable baseline:

| Suite | Main | Current branch | Delta |
| --- | ---: | ---: | ---: |
| `test.test_pathlib` | 539/979 55.06% | 958/976 98.16% | +419 passes, +43.10 pts |
| `test.test_os` | 132/217 60.83% | 176/215 81.86% | +44 passes, +21.03 pts |
| `test.test_tarfile` | 541/606 89.27% | 589/606 97.19% | +48 passes, +7.92 pts |
| `test.test_zipfile` | 248/331 74.92% | 251/331 75.83% | +3 passes, +0.91 pts |
| TOTAL | 1460/2133 68.45% | 1974/2128 92.76% | +514 passes, +24.31 pts |

lib/wasix/src/syscalls/wasix/chdir.rs
sadhbh-c0d3: man chdir(2)

  The chdir() function changes the current directory to the directory
  indicated by the path parameter.

  The fchdir() function changes the current directory to the directory
  indicated by the filedes parameter.  **If the path parameter refers to a
  symbolic link, the chdir() function sets the current directory to the
  directory pointed to by the symbolic link.**

  The current directory, also called the current working directory, is the
  starting point of searches for pathnames that do not begin with a /
  (slash).   In order for a directory to become the current directory, the
  calling process must have search access to the directory.

@theduke  This means that current `fs.set_current_dir(&resolved_path)` is correct.

sadhbh-c0d3:
> > But I already addressed them here
> 
> The change in this PR is very convoluted, probably generated by LLM? We just need to install one more package, no need for a fancy detection.

I replaced that with your changes.

lib/wasix/src/fs/mod.rs
sadhbh-c0d3: @theduke this is resolved, we always resolve symlinks, unless it's a last component of the top level path and follow is off. So in any other case, whether it is inner component of the path, or recursive symlink resolution, we always follow all symlinks.

sadhbh-c0d3:
> probably generated by LLM

FYI, the core function to this whole business, `get_inode_at_path_inner()`, was rewritten by myself without any AI, but then I asked AI to improve it. The process was long, because improving meant getting all compat-tests regressions fixed, and because now symlinks work well, lots of tests started failing, and that is why smaller changes were required to make all those tests pass.

lib/wasix/src/fs/mod.rs
```diff
@@ -302,26 +331,181 @@ pub struct WasiInodes {
...
+    path.to_string_lossy().into_owned()
+}
+
+pub(crate) fn strip_guest_root_prefix(path: &Path) -> PathBuf {
+    let path_str = path.to_string_lossy();
+    PathBuf::from(path_str.strip_prefix('/').unwrap_or(&path_str))
+}
+
+pub(crate) fn guest_path_is_absolute(path: &Path) -> bool {
+    path.to_string_lossy().starts_with('/')
```
syrusakbary: Why we need this function? 

Is not this one better? (it doesn't require string lossy conversion)

```suggestion
    path.is_absolute()
```

lib/wasix/src/fs/mod.rs
```diff
@@ -302,26 +331,181 @@ pub struct WasiInodes {
...
+    components
+}
+
+pub(crate) fn guest_path_to_string(path: &Path) -> String {
+    path.to_string_lossy().into_owned()
+}
+
+pub(crate) fn strip_guest_root_prefix(path: &Path) -> PathBuf {
+    let path_str = path.to_string_lossy();
+    PathBuf::from(path_str.strip_prefix('/').unwrap_or(&path_str))
```
syrusakbary: Why not this?

```suggestion
    PathBuf::from(path.strip_prefix("/").unwrap_or(path))
```

lib/wasix/src/fs/mod.rs
sadhbh-c0d3: The idea is that guest path is not platform dependent path, so it's absolute only when it starts with `/`. Otherwise we'd use Path as ususal, which was the case before. To be POSIX compliant path needs to follow strict rules.

lib/wasix/src/fs/mod.rs
sadhbh-c0d3: While at first glance looks like you're right, unfortunately the reason for not using Path methods is to be strictly POSIX compliant, so operations are made on strings and not platform independent Path. 

PS. We're probably passing and storing Path in many places. Ideally we should have PosixPath that replaces it.

sadhbh-c0d3: For compat-tests to pass path must behave identitally as POSIX.

Even if different variant of the same path is valid POSIX path, it is not compliant for system layer to replace one with the other. That is the sole reason to have all those guest path operations defined on strings and not use platform independent Path.

WASIX is implementing system layer and not user toolkit, and compat-tests rely on that, and expect same behavior.

sadhbh-c0d3: I will move all POSIX path stuff into separate module.

Arshia001 (approved):
LGTM!

lib/wasix/src/fs/mod.rs
sadhbh-c0d3: ^ done

lib/wasix/src/fs/mod.rs
Arshia001: to be extra pedantic, we actually need a parallel type to stdlib's `Path`. Just using the built-in type is wrong on Windows, and converting to string is non-POSIX because we're limiting ourselves to unicode paths. Still, strings are the better short-term hack here.

lib/virtual-fs/src/host_fs.rs
Arshia001: Actually, the hard_link impl didn't have sandbox escape risks, but I figured this out after a second review. I'd still want to avoid creating hard links in host backed folders because hard links can get really weird.

syrusakbary:
Checked bunch of things, they all work:
1. WordPress
2. Bash (with symlinks)
3. Next.js

We can be confident enough on this PR to merge (based on @Arshia001's review as well)

syrusakbary (approved):

PR #6627, opened 2026-05-25
fix: advance static WebC file cursors on read

immanuwell:
## What
Static WebC reads in `StaticFileSystem` were not moving the file cursor.
If a caller reads the file in chunks, it gets the same bytes again, which is pretty annoying.

This path is live in the C API WASI `prepare_webc_env` flow too, so this is not just test-only stuff.

## Repro
Use any static WebC file that is read in 2 chunks.

1. Read 4 bytes from an 8-byte wasm file and get `[0, 97, 115, 109]`
2. Read 4 more bytes
3. Before this patch it returns `[0, 97, 115, 109]` again
4. After this patch it returns `[1, 0, 0, 0]`

## Fix
Bump the cursor by the copied byte count and add a small regression test.

tiny fix, but pretty legit.

### DIFF
--- a/lib/virtual-fs/src/static_fs.rs
+++ b/lib/virtual-fs/src/static_fs.rs
@@ -135,27 +135,27 @@ impl AsyncRead for WebCFile {
         _cx: &mut Context<'_>,
         buf: &mut tokio::io::ReadBuf<'_>,
     ) -> Poll<io::Result<()>> {
-        let bytes = self
+        let this = self.get_mut();
+        let bytes = this
             .volumes
-            .get(&self.volume)
+            .get(&this.volume)
             .ok_or_else(|| {
                 IoError::new(
                     IoErrorKind::NotFound,
-                    anyhow!("Unknown volume {:?}", self.volume),
+                    anyhow!("Unknown volume {:?}", this.volume),
                 )
             })?
-            .get_file_bytes(&self.entry)
+            .get_file_bytes(&this.entry)
             .map_err(|e| IoError::new(IoErrorKind::NotFound, e))?;
 
-        let cursor: usize = self.cursor.try_into().unwrap_or(u32::MAX as usize);
-        let _start = cursor.min(bytes.len());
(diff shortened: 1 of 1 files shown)


lib/virtual-fs/src/static_fs.rs
```diff
@@ -135,27 +135,27 @@ impl AsyncRead for WebCFile {
         _cx: &mut Context<'_>,
         buf: &mut tokio::io::ReadBuf<'_>,
     ) -> Poll<io::Result<()>> {
-        let bytes = self
+        let this = self.get_mut();
```
marxin, line 138: Let's stick to the usage of `self`, introduction of `this` seem not idiomatic to Rust.

lib/virtual-fs/src/static_fs.rs
```diff
@@ -402,3 +402,59 @@ fn translate_file_type(f: FsEntryType) -> crate::FileType {
...
+    use webc::{
+        metadata::Manifest,
+        v1::{DirOrFile, Volume, WebC},
+    };
+
+    fn fileblock_with_wasm_file() -> &'static [u8] {
+        let volume_bytes = Volume::serialize_files(
+            [(
+                DirOrFile::File(Path::new("lib/python.wasm").to_path_buf()),
+                vec![0, 97, 115, 109, 1, 0, 0, 0],
```
marxin, line 421: Can we create rather a text input with `b"foo bar ..."`?

marxin (changes requested):
Good catch, ready with a small nits.

lib/virtual-fs/src/static_fs.rs
theduke: This is relatively common, I also do it quite often in a similar context. 

Can't re-declare self, obviously, and no reason to write `self.get_mut()` 5 times.

theduke:
Let's have @Arshia001 review this one as well, please.

Arshia001 (approved):
LGTM! Thanks for the contribution.

PR #6632, opened 2026-05-25
chore: unify registry setting for the CLI int. tests (and enable them)

Fliqqr:
# Description
- reenabled integration tests
- added explicit registry URL var for tests
- gitignore `.env`

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -609,9 +609,8 @@ jobs:
             make: "test-wasmer-cli"
           - description: "Examples"
             make: "test-examples"
-        # TODO: temporarily disabled
-        #          - description: "CLI integ. tests"
-        #            make: test-integration-cli-ci
+          - description: "CLI integ. tests"
+            make: test-integration-cli-ci
         metadata:
           - build: linux-x64
             os: ubuntu-22.04
--- a/.gitignore
+++ b/.gitignore
@@ -35,4 +35,3 @@ link.tar.gz
 result
 result-*
 /lib/wasix/tests/wasm_tests/build/
-.env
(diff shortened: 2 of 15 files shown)


marxin (changes requested):
Do we know what exactly changed that we're facing a different expected results?
Note the PR should ideally include re-enabling of the integration tests (disabled in #6623).

Fliqqr:
Reenabling integration tests is still blocked by [BE-1462](https://linear.app/wasmer/issue/BE-1462/wasmer-runtime-integration-tests-failing-on-dev), this one is just a small fix for other failing tests

`wasmer/hello` has changed from returning <title>wasmer/hello</title> to <title>Hello World</title>, and some of the assert snapshots were outdated

We can keep this as a draft until the main integration test fix is ready

marxin:
> `wasmer/hello` has changed from returning <title>wasmer/hello</title> to <title>Hello World</title>, and some of the assert snapshots were outdated

All right.

> We can keep this as a draft until the main integration test fix is ready

I would rather included revert of the commit where the int. tests were disabled and we can keep it opened until the BE issue is resolved.

maminrayej (dismissed):
It needs to be investigated why snapshot values have changed.

Fliqqr:
There is seems to be discrepancy between what the tests expect when running locally and when ran in the CI. Could revert the snapshots to how they were, which should work for CI, but will not work for local testing.

Gonna come back to this tomorrow, not as easy as it initially seemed.

Fliqqr:
It was a local config issue, even though I rebuilt wasmer-cli it was still pulling the registry URL from the global installation, which was pointed to wasmer.wtf for some reason, but integration tests expect .io as the packages seem to differ between the two slightly.

Fliqqr:
Issue was on my side with local config pulling in the wrong registry url, made the tests more explicit with what registry URL is used so it hopefully does not happen again.

marxin (approved):
LGTM

PR #6633, opened 2026-05-25
fix(wasix): Don't log error on HostInterrupt as it's expected behaviour

artemyarulin:
We've introduced new `HostInterrupt` behaviour a few weeks ago and currently we are almost hitting rate limits in Sentry in Edge. I think we should simply silence that as it's totally normal behavour

Fixes https://linear.app/wasmer/issue/EDGE-1775/sentry-runtimeerror-interrupted-by-host

### DIFF
--- a/lib/wasix/src/bin_factory/exec.rs
+++ b/lib/wasix/src/bin_factory/exec.rs
@@ -381,8 +381,16 @@ fn call_module(
             Some(s) => s,
             None => {
                 let err_display = err.display(&mut store);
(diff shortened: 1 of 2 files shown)


Arshia001 (changes requested):
I'd still log the message at debug level. Removing logs outright is never a good idea in WASIX.

artemyarulin:
Yeah, good idea - fixed

Arshia001 (approved):

PR #6635, opened 2026-05-26
fix(V8): use proper new memory sharing mechanism

marxin:
The PR migrates to the new memory sharing API and enables WASIX tests on V8 for platforms where we build wee8 library.

Issue: #6121

### DIFF
--- a/lib/api/src/backend/v8/entities/memory/mod.rs
+++ b/lib/api/src/backend/v8/entities/memory/mod.rs
@@ -182,6 +182,9 @@ impl Memory {
 
         let memorytype = unsafe { wasm_memorytype_new(&limits) };
         let copied_memory = unsafe { wasm_memory_new(v8_store.inner, memorytype) };
+        unsafe {
+            wasm_memorytype_delete(memorytype);
+        }
         if copied_memory.is_null() {
             return Err(MemoryError::Generic(
                 "failed to allocate copied V8 memory".to_owned(),
--- a/lib/api/src/backend/v8/entities/memory/mod.rs
+++ b/lib/api/src/backend/v8/entities/memory/mod.rs
@@ -1,7 +1,7 @@
 //! Data types, functions and traits for `v8` runtime's `Memory` implementation.
 use std::{marker::PhantomData, mem::MaybeUninit};
 
-use tracing::{debug, warn};
+use tracing::warn;
 pub use wasmer_types::MemoryError;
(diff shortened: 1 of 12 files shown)


lib/api/src/backend/v8/entities/memory/mod.rs
```diff
@@ -58,25 +58,29 @@ impl Memory {
...
     }
 
     pub fn ty(&self, store: &impl AsStoreRef) -> MemoryType {
         check_isolate(store);
         let store = store.as_store_ref();
 
-        let memory_type: *mut wasm_memorytype_t = unsafe { wasm_memory_type(self.handle.0) };
+        let memory_type: *mut wasm_memorytype_t =
+            unsafe { wasm_memory_type(self.handle.as_memory()) };
         let limits: *const wasm_limits_t = unsafe { wasm_memorytype_limits(memory_type) };
```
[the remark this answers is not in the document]
Arshia001: @marxin sounds to me like Copilot is right about this?

syrusakbary (approved):

syrusakbary (changes requested):
We have mutated the VMMemory to be either attached (like now), or Shared.

However, we have a type that we use for Shared Memory: `VMSharedMemory`.

We should use this type instead of using always VMMemory

lib/api/third-party/wee8/wasm.h
```diff
@@ -498,7 +498,7 @@ WASM_API_EXTERN bool wasm_table_grow(wasm_table_t*, wasm_table_size_t delta, was
 
 // Memory Instances
 
-WASM_DECLARE_REF(memory)
+WASM_DECLARE_SHARABLE_REF(memory)
```
syrusakbary, line 501: This is weird.

We should have shared memory as a shareable_ref and memory as a normal ref, right?

marxin:
> We have mutated the VMMemory to be either attached (like now), or Shared.
> 
> However, we have a type that we use for Shared Memory: `VMSharedMemory`.
> 
> We should use this type instead of using always VMMemory

Should be properly reflected in the new PR: #6644

lib/api/third-party/wee8/wasm.h
marxin: Note this change is only copying of what we merged in https://github.com/wasmerio/v8-custom-builds/pull/15.
Don't know the exact details of the WASM C API macro declaration magic - but the underlying types seem reasonable to me.

marxin:
Superceded in #6644

PR #6639, opened 2026-05-27
fix: Prevent memory leak for subprocesses

### LINKED ISSUE #6637: Regression: repeated spawned PHP child processes use much more RAM after bc34d15d
Repeated short-lived PHP child processes now retain much more memory in Wasmer/WASIX.

- last good: 4263a29798a884562904cb8a9e8643358957bfae
- first bad: bc34d15d2489764438c4b2b5cd0db2203e8d762c (fix(wasix): Interrupt atomic waiters on signals)
- latest main still reproduces

```php
<?php
if ($argv[1] ?? 0) return;
for ($i = 0; $i < 12; $i++) shell_exec(PHP_BINARY . " --no-php-ini " . __FILE__ . " 1");
```

Running as:
```
$ wasmer run --wasmer-dir /tmp/wasmer-test --volume "$PWD:/work" phpix/phpix-83-32bit@=0.2.0-rc.4 /work/repro_php_spawn_min.php
```

- commit: 4263a29798a884562904cb8a9e8643358957bfae, time: 0.78s, max RSS: 655 MiB
- commit: bc34d15d2489764438c4b2b5cd0db2203e8d762c, time: 0.82s, max RSS: 2.32 GiB

bc34d15d started attaching SharedMemory to WasiProcess, and Wasix already has a FIXME that terminated processes are not deregistered from the control plane. So leaked process records may now pin memory per spawned child.

theduke:
A previous modification registered memory on process handles.

Since process lifetime management is not properly implemented, this lead
to leaking the memory of subprocesses until the main process exits.

Fixed by just keeping a handle to the memory, which exposes operations
through a weak reference.

Fixes #6637

### DIFF
--- a/lib/api/src/entities/memory/shared.rs
+++ b/lib/api/src/entities/memory/shared.rs
@@ -14,9 +14,24 @@ pub struct SharedMemory {
     ops: Option<Arc<dyn SharedMemoryOps + Send + Sync>>,
 }
 
+/// Shared memory operations that do not hold the underlying memory alive.
+///
+/// This handle is intended for operations, such as waking atomic waiters, that
+/// may be attempted after the original memory owner has started shutting down.
+#[derive(Clone)]
+pub struct MemoryOps {
+    ops: Option<Arc<dyn SharedMemoryOps + Send + Sync>>,
+}
+
 impl std::fmt::Debug for SharedMemory {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("SharedMemory").finish()
+    }
+}
+
+impl std::fmt::Debug for MemoryOps {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("MemoryOps").finish()
     }
 }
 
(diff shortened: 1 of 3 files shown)


marxin:
Does it address #6637? If so, please use `Fixes: ` directive for the PR description?

theduke:
Requesting review from @marxin since it extends the core API.

lib/api/src/entities/memory/shared.rs
```diff
@@ -14,12 +14,27 @@ pub struct SharedMemory {
     ops: Option<Arc<dyn SharedMemoryOps + Send + Sync>>,
```
marxin, line 14: Shouldn't this be converted to the `MemoryOps` as well given it's equivalent? Please investigate also the other places where the type is used.

marxin (changes requested):
LGTM,

lib/api/src/entities/memory/shared.rs
```diff
@@ -14,12 +14,27 @@ pub struct SharedMemory {
...
 }
 
+/// Shared memory operations that do not hold the underlying memory alive.
+///
+/// This handle is intended for operations, such as waking atomic waiters, that
+/// may be attempted after the original memory owner has started shutting down.
+#[derive(Clone)]
+pub struct MemoryOps {
+    ops: Option<Arc<dyn SharedMemoryOps + Send + Sync>>,
```
Arshia001, line 23: Nit: This type promises to not keep the memory alive, but really depends on an internal detail of the only existing implementation, which keeps memories around via a `Weak<T>`, so it's easy to break the contract in the future.

lib/api/src/entities/memory/shared.rs
theduke: I didn't want to couple the internal representation to the public api, to make it clear that MemoryOps is purely for , hence not re-using MemoryOps is for public consumers.

May be debatable?

marxin: Gotcha - sure, leaving that up to you.

syrusakbary (dismissed):
I don't think we should have a new type for memory.

If we have a leak on memory on subprocess, we should clean that memory upon the subprocess ending, right?

It seems to me we are fixing the issue on the wrong place. (we can do a meeting to clarify what could be a better solution)

marxin (approved):
I covered the described limitation of the `WasiControlPlane` in an issue so that it does not fall through crack. But, for now, this change (already deployed to the Edge) is a solid workaround to me.

PR #6642, opened 2026-05-29
Fix FS race resulting in None handle, take 2

Arshia001:
Supersedes #6636.


lib/wasix/src/fs/fd.rs
```diff
@@ -15,6 +15,9 @@ use crate::os::epoll::EpollState;
 
 use super::{InodeGuard, InodeWeakGuard, NotificationInner};
 
+/// Shared handle to an open [`VirtualFile`].
+pub(crate) type VirtualFileLock = Arc<RwLock<Box<dyn VirtualFile + Send + Sync + 'static>>>;
```
theduke, line 19: Java vibes, double boxing , not so great for perf.

*if* we never need to move out or swap the file, we can use: `Arc<RwLock<dyn VirtualFile>>` without the extra Box, I believe.

lib/wasix/src/fs/fd.rs
```diff
@@ -89,7 +92,7 @@ pub enum Kind {
     File {
         /// The open file, if it's open
         #[cfg_attr(feature = "enable-serde", serde(skip))]
-        handle: Option<Arc<RwLock<Box<dyn VirtualFile + Send + Sync + 'static>>>>,
+        handle: Option<VirtualFileLock>,
```
theduke, line 95: Side question, but why Option<>.
So we can retain info for a closed fd?

lib/wasix/src/syscalls/wasi/fd_renumber.rs
```diff
@@ -38,71 +38,15 @@ pub(crate) fn fd_renumber_internal(
...
-    let flush_target = old_fd.as_ref().and_then(|fd_entry| {
-        let guard = fd_entry.inode.read();
-        match guard.deref() {
-            Kind::File {
-                handle: Some(file), ..
-            } => Some(file.clone()),
-            _ => None,
-        }
-    });
-    drop(old_fd);
```
theduke, line 104: I don't think this is new, it was already broken before, but fd_renumber is supposed to just RENUMBER, not to keep the old fd open!

But since this is not new behaviour, can come as a followup.

lib/wasix/src/fs/fd.rs
Arshia001: Mainly because the WASIX-level FS tree that caches inodes never drops them.

lib/wasix/src/fs/fd.rs
Arshia001: that's complicated. also, I'm not sure you can put trait objects in locks directly?

lib/wasix/src/syscalls/wasi/fd_renumber.rs
Arshia001: actually, when you `dup2` an FD, the old one is supposed to stay open too. I guess `fd_renumber` is a bit of misnomer, but the syscall is meant to back `dup2`, meaning the current logic is correct.

lib/wasix/src/fs/fd.rs
theduke: Yeah you can.

theduke: Is it that complicated? I guess we already return Box<dyn VirtualFile> from some APIs? If so, then yeah.

theduke (approved):

Issue #6555, opened 2026-05-05
Repeated `data.drop` / `elem.drop` in a hot loop is much slower in Wasmer (LLVM and Cranelift) than in other runtimes


gaaraw:
### Describe the bug

A very small bulk-memory/reference-types microbenchmark that repeatedly executes `data.drop` or `elem.drop` in a hot loop is much slower in Wasmer than in other runtimes tested.

The slowdown appears in **both Wasmer LLVM and Wasmer Cranelift**, and the same pattern shows up for both `data.drop` and `elem.drop`.

The benchmark is not a loop-template artifact: replacing the target instruction with `nop` collapses runtime from multi-second to sub-second.

### Environment

- OS: Linux wuhao-Precision-3660 6.8.0-85-generic (Ubuntu 22.04.1)
- Architecture: x86_64
- Wasmer version: `6.1.0`
- Wasmtime version used for comparison: `41.0.0 (4898322a4 2025-12-18)`
- rustc:  rustc 1.90.0 (1159e78c4 2025-09-14)
	binary: rustc
	commit-hash: 1159e78c4747b02ef996e55082b704c09b970588
	commit-date: 2025-09-14
	host: x86_64-unknown-linux-gnu
	release: 1.90.0
	LLVM version: 20.1.8

### Steps to reproduce

[test_cases.zip](https://github.com/user-attachments/files/27391051/test_cases.zip)

Primary reproducer: `data_drop_2.wat`

Compile:

```bash
wat2wasm data_drop_2.wat -o data_drop_2.wasm
```

Run with Wasmer LLVM:

```bash
/path/to/wasmer run --disable-cache --llvm data_drop_2.wasm
```

Run with Wasmer Cranelift:

```bash
/path/to/wasmer run --disable-cache --cranelift data_drop_2.wasm
```

Supporting controls:

- `nop_control.wat`: same wrapper shape, but loop body is `nop`
- `elem_drop_1.wat`: sibling reproducer for `elem.drop`
- `elem_drop_nop_control.wat`: same wrapper shape, but loop body is `nop`

### Expected behavior

I would expect `data.drop` / `elem.drop` to be measurably slower than `nop`, but not to be such a large outlier relative to other runtimes, especially when the same slowdown pattern appears in both Wasmer backends.

#### Primary reproducer: `data_drop_2.wat`

The hot loop body is only:

```wat
(loop $body
  data.drop 1
  (local.set $i (i64.sub (local.get $i) (i64.const 1)))
[...]

PR #6648, opened 2026-05-29
test(wasix): port wasi-wast tests to wasm harness

### LINKED ISSUE #6646: `wasitests` can be ported to the new `wasm_tests` harness
The tests under `tests/wasi-wast/src/wasitests.rs` can be ported to the new test harness, where can removed the existing `.wasm`/`.wat` files and generate them on fly.

Should be pretty close to wasmerio/wasmer#6566

sonukapoor:
This ports the old `tests/wasi-wast` coverage into the current WASIX wasm test harness under `lib/wasix/tests/wasm_tests/wasi_wast`.

The old generator and checked-in generated `.wasm` / `.wast` fixtures are removed, and `make test-wasi` now runs the ported `wasi_wast` cases through the WASIX harness. The WASI examples now use small inline WAT modules instead of depending on the deleted generated fixtures.

The port keeps the old source coverage in parity with the new test list, including `fd_rights`, and moves the `test_fs` fixture tree next to the new tests.

Verified locally:

```sh
cargo fmt
cargo metadata --locked --no-deps
LLVM_SYS_221_PREFIX=.llvm cargo test -p wasmer-wasix --features sys --test wasm_tests wasi_wast -- --nocapture
cargo test -p wasmer-wast --features wasmer/cranelift --locked
git diff --check origin/main..HEAD
```

The root example compile path is currently blocked locally by missing `tests/wast/spec/proposals/wide-arithmetic`, before it reaches these example changes.

Resolves: #6646


Arshia001:
@marxin despite touching WASIX, I think you should be the one to review this since it mostly touches how tests are run.

tests/lib/wast/src/wasi_wast.rs
```diff
@@ -1,765 +0,0 @@
...
-use wasmer_wasix::types::wasi::{Filesize, Timestamp};
-use wasmer_wasix::{
-    FsError, PluggableRuntime, VirtualFile, WasiEnv, WasiEnvBuilder, WasiError, WasiVersion,
-    generate_import_object_from_env, get_wasi_version,
-};
-use wast::parser::{self, Parse, ParseBuffer, Parser};
-
-/// The kind of filesystem `WasiTest` is going to use.
-#[derive(Debug)]
-pub enum WasiFileSystemKind {
```
marxin, line 31: That's one aspect of the port that I overlooked - can we please add similar functionality to the `wasm_tests` and with a new directive, list use file systems?

marxin (changes requested):
Very good job - in general looks fine, with one comment related to the support of testing multiple mounting options.

tests/lib/wast/src/wasi_wast.rs
sonukapoor: Added this in `e45beae` with a new `FileSystems:` directive in `wasm_tests`.

The ported `wasi_wast` cases now run with filesystem variants in the test name, for example `wasm/wasi_wast/fd_read/default/mem/llvm`. Most cases use `host,mem,tmp,passthru,union,root`; `unix_open_special_files` remains host-only because it maps `/dev`.

Verified locally:

```sh
LLVM_SYS_221_PREFIX=.llvm cargo test -p wasmer-wasix --features sys --test wasm_tests wasi_wast -- --nocapture
cargo metadata --locked --no-deps
git diff --check origin/main..HEAD
```

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -135,6 +144,45 @@ impl Engine {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum FileSystemKind {
```
marxin, line 143: All right - I think it's probably right time to start using `strum` crate, please port it and drop `fn name` and `fn from_str`. Similarly, please port also the `Engine` enum.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -135,6 +144,45 @@ impl Engine {
...
+        }
+    }
+}
+
+impl FromStr for FileSystemKind {
+    type Err = anyhow::Error;
+
+    fn from_str(value: &str) -> Result<Self> {
+        match value.trim().to_lowercase().as_str() {
+            "host" | "host_fs" => Ok(Self::Host),
```
marxin: Do not allow multiple variants - just a single one. And please add also `all` that will enable all of them.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -206,13 +260,30 @@ impl Config {
     }
 
     fn full_test_name(&self) -> String {
-        if self.source.is_default() {
+        if self.source.is_default() && self.include_file_system_in_name {
```
marxin: We should likely build a vectors and later join them with '/' instead.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -450,6 +521,16 @@ fn process_directive(
         "DefaultMappedDirectories" => {
             config.default_mapped_directories = arg.parse::<bool>()?;
         }
+        "FileSystems" => {
+            let file_systems = arg
+                .split(|c: char| c == ',' || c.is_whitespace())
```
marxin: Just comma.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -634,6 +715,191 @@ fn copy_symlink(from: &Path, _to: &Path) -> Result<()> {
...
+    let path = Path::new(guest);
+    create_virtual_dir_all(fs, path).with_context(|| {
+        format!(
+            "failed to create empty mapped directory {} in virtual filesystem",
+            path.display()
+        )
+    })
+}
+
+fn copy_host_tree_to_virtual_fs(
```
marxin, line 748: This seems like another copy of `copy_test_tree`.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -145,6 +193,9 @@ struct Config {
     test_name: String,
     config_name: String,
     engine: Engine,
+    file_system: FileSystemKind,
```
marxin: Why do we need 2 fields here? Note `include_file_system_in_name` should probably somehow depend on either `file_system`, or `Vec<FileSystemKind>`.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -634,6 +715,191 @@ fn copy_symlink(from: &Path, _to: &Path) -> Result<()> {
...
+
+    Ok(())
+}
+
+fn configure_mapped_directories(
+    runner: &mut wasmer_wasix::runners::wasi::WasiRunner,
+    config: &Config,
+    extra_temporary_folders: &mut Vec<tempfile::TempDir>,
+) -> Result<()> {
+    if config.file_system == FileSystemKind::Host {
```
marxin: If possible, I would like to see it unified with `for directory in &config.mapped_directories`.

marxin (changes requested):

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 8ceb04e. `Engine` and `FileSystemKind` now derive `strum::Display` and `strum::EnumString`, and the manual name/from_str helpers were removed.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 8ceb04e. `FileSystems` now accepts one backend or `all`; comma-separated lists are rejected. The ported `wasi_wast` tests use `all`, while `unix_open_special_files` stays host-only.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 8ceb04e. `full_test_name` now builds the path components in a vector and joins them with `/`.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: This is now single-value only per the other comment: commas are rejected, and `all` expands to every backend.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 8ceb04e. The shared `WalkDir` traversal is now factored into `copy_host_tree`, with host and virtual-FS callbacks for directories, files, and symlinks.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 8ceb04e. `Config` now stores a single `FileSystemSelection`; name inclusion is derived from whether the selection came from an explicit `FileSystems` directive.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 8ceb04e. `configure_mapped_directories` now uses one loop over `config.mapped_directories` and branches inside it for host vs. virtual filesystem setup.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -116,27 +123,74 @@ struct MappedDirectory {
     guest: String,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, strum::Display, strum::EnumString)]
+#[strum(ascii_case_insensitive)]
 pub enum Engine {
+    #[strum(serialize = "cranelift")]
```
marxin: No need for extra annotation if I'm not mistaken: https://docs.rs/strum/latest/strum/derive.Display.html

lib/wasix/tests/wasm_tests/mod.rs
marxin: Same here - I probably like the longer names (enum variants).

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -116,27 +123,74 @@ struct MappedDirectory {
...
+    Tmp,
+    #[strum(serialize = "passthru")]
+    PassthruMemory,
+    #[strum(serialize = "union")]
+    Union,
+    #[strum(serialize = "root")]
+    Root,
+}
+
+const ALL_FILE_SYSTEMS: &[FileSystemKind] = &[
```
marxin: https://docs.rs/strum/latest/strum/derive.EnumIter.html

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -116,27 +123,74 @@ struct MappedDirectory {
...
+    FileSystemKind::Host,
+    FileSystemKind::InMemory,
+    FileSystemKind::Tmp,
+    FileSystemKind::PassthruMemory,
+    FileSystemKind::Union,
+    FileSystemKind::Root,
+];
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum FileSystemSelection {
```
marxin: Don't like extra enum, in parsing, for "All" - just fill up all the variants.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -116,27 +123,74 @@ struct MappedDirectory {
...
-            Self::Singlepass => "singlepass",
-            #[cfg(feature = "v8")]
-            Self::V8 => "v8",
+            Self::Default => &[FileSystemKind::Host],
+            Self::Selected(kind) => std::slice::from_ref(kind),
+            Self::All => ALL_FILE_SYSTEMS,
         }
     }
+
+    fn selected(&self) -> FileSystemKind {
```
marxin: Should be removed.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -389,30 +439,21 @@ fn process_directive(
             let (engine, reason) = arg
                 .split_once(':')
                 .ok_or_else(|| anyhow!("missing colon separator for SkipEngine"))?;
-            if let Some(engine) = match engine.to_lowercase().as_str() {
```
marxin, line 394: I would stick with the original version, even though it's not ideal either.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -600,43 +656,57 @@ fn run_build_script(config: &Config) -> anyhow::Result<PathBuf> {
     Ok(build_test_path.join("main"))
 }
 
-fn copy_test_tree(from: &Path, to: &Path) -> Result<()> {
-    create_dir_all(to).with_context(|| format!("failed to create {}", to.display()))?;
+fn copy_host_tree(
+    from: &Path,
```
marxin, line 669: I would replace the arguments with a defined `struct`.

lib/wasix/tests/wasm_tests/wasi_wast/path_symlink.rs
```diff
@@ -1,6 +1,9 @@
-// WASI:
-// tempdir: temp
-// mapdir: hamlet:test_fs/hamlet
+//#DefaultMappedDirectories: false
```
marxin: Can we abstract some of the commonly used directives to an AbstractConfig?

marxin (changes requested):

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 1c103e2. Removed the explicit filesystem serialize names as well, so the filesystem segment now uses enum variant names such as `InMemory` and `PassthruMemory`.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 1c103e2. Removed the explicit `strum(serialize = ...)` annotations from `Engine`; the derived display now uses the enum variant names.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 1c103e2. `FileSystemKind` now derives `EnumIter`, and `FileSystems: all` expands through `FileSystemKind::iter()`.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 1c103e2. Replaced the separate copy-helper callback arguments with a `CopyHostTreeActions` struct.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 1c103e2. Removed the extra selection enum; `FileSystems: all` now stores the expanded filesystem variants directly in the config.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 1c103e2. The `FileSystemSelection` helper methods are gone with the enum; the config now uses an optional filesystem list.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 1c103e2. Restored the explicit `SkipEngine` match style while keeping the Singlepass case from current main.

lib/wasix/tests/wasm_tests/wasi_wast/path_symlink.rs
sonukapoor: Done in 1c103e2. Added `wasi_wast/wasi-wast.config` for the common directives and updated the ported fixtures to use `AbstractConfigFile`; tests with `/tmp` or host-only `/dev` keep local overrides.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -209,23 +218,30 @@ impl Config {
         self.tests_build_root.join(self.full_test_name())
     }
 
+    fn file_system_kinds(&self) -> &[FileSystemKind] {
+        self.file_systems
+            .as_deref()
+            .unwrap_or(&[FileSystemKind::Host])
```
marxin: Don't like this `unwrap_or`, I would rather start with `file_systems: vec[FileSystemKind::Host]`, and then replace it if `FileSystems` directive takes action.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -464,6 +480,21 @@ fn process_directive(
         "DefaultMappedDirectories" => {
             config.default_mapped_directories = arg.parse::<bool>()?;
         }
+        "FileSystems" => {
+            ensure!(!arg.is_empty(), "FileSystems must not be empty");
+            ensure!(
```
marxin: Well, it's a bit strange the name contains plural, but we're expecting a single value. I would support parsing a list of values separated by comma.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -126,17 +135,15 @@ pub enum Engine {
...
-            Self::LLVM => "llvm",
-            #[cfg(feature = "singlepass")]
-            Self::Singlepass => "singlepass",
-            #[cfg(feature = "v8")]
-            Self::V8 => "v8",
-        }
-    }
+#[derive(Debug, Clone, Copy, PartialEq, Eq, strum::Display, strum::EnumIter, strum::EnumString)]
+#[strum(ascii_case_insensitive)]
+enum FileSystemKind {
```
marxin, line 143: `#[strum(serialize_all = "lowercase")]`

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -116,7 +124,8 @@ struct MappedDirectory {
     guest: String,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, strum::Display, strum::EnumString)]
+#[strum(ascii_case_insensitive)]
 pub enum Engine {
```
marxin, line 131: `#[strum(serialize_all = "lowercase")]`

marxin (changes requested):
We're pretty close - please address the remaining issue and merge `main` branch.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 3b31db8. `Config` now initializes `file_systems` with `Host`; the separate `explicit_file_systems` flag is only used to preserve the existing default test names.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 3b31db8. `FileSystemKind` now uses `#[strum(serialize_all = "lowercase")]`.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 3b31db8. `Engine` now uses `#[strum(serialize_all = "lowercase")]`.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Done in 3b31db8. `FileSystems` now accepts comma-separated filesystem values, while `all` still expands through `FileSystemKind::iter()`.

marxin: So I would rather make a quick check of `if arg == "all" ...` and then directly split by comma and call `kind.parse::<FileSystemKind>(`, no extra checking is necessary.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -989,37 +1209,41 @@ fn collect_tests(tests: &mut Vec<Trial>) -> Result<()> {
...
+                    for engine in &supported_engines {
+                        // In general, the WASIX tests expect support for more advanced WebAssembly extensions (like exception handling),
+                        // but we can still run selectively some tests with Singlepass.
+                        #[cfg(feature = "singlepass")]
+                        if entry
+                            .path()
+                            .file_name()
+                            .expect("must be valid filename")
+                            .to_string_lossy()
+                            != "wasi_fyi"
```
marxin: We should be able to run the new tests with `Singlepass` as well. Please enable it.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -211,23 +237,28 @@ impl Config {
         self.tests_build_root.join(self.full_test_name())
     }
 
+    fn file_system_kinds(&self) -> &[FileSystemKind] {
```
marxin: No need for the function.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -151,6 +173,8 @@ struct Config {
     test_name: String,
     config_name: String,
     engine: Engine,
+    file_systems: Vec<FileSystemKind>,
+    explicit_file_systems: bool,
```
marxin: Don't like this bool, we should rather use `Option<Vec<FileSystemKind>>` for `file_systems`.

Makefile
```diff
@@ -316,6 +316,10 @@ endif
 compiler_features := --features $(subst $(space),$(comma),$(compilers)),wasmer-artifact-create,static-artifact-create,wasmer-artifact-load,static-artifact-load
 test_compiler_features := --features $(subst $(space),$(comma),$(test_compilers)),wasmer-artifact-create,static-artifact-create,wasmer-artifact-load,static-artifact-load
 build_compiler_features = --features $(subst $(space),$(comma),$(build_compilers))$(if $(build_wasmer_extra_features_csv),$(comma)$(build_wasmer_extra_features_csv)),wasmer-artifact-create,static-artifact-create,wasmer-artifact-load,static-artifact-load
+wasix_wasm_test_features := sys
```
marxin: Drop please.

Makefile
```diff
@@ -667,7 +671,7 @@ test-wasi-unit:
 	$(CARGO_BINARY) test $(CARGO_TARGET_FLAG) --manifest-path lib/wasi/Cargo.toml --release --locked
 
 test-wasi:
```
marxin: Drop as well, we'll be run automatically as part of WASIX tests.

marxin (changes requested):
Thank you, I think we're pretty close.

lib/wasix/tests/wasm_tests/mod.rs
marxin: Done.

Makefile
sonukapoor: Dropped in `270bd2b`; `wasi_wast` remains covered through the WASIX wasm tests. Re-ran `cargo fmt --check`, `cargo metadata --locked --no-deps`, and the focused `wasi_wast` Singlepass subset.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Removed in `270bd2b`; the collection loop now expands `config.file_systems` directly, defaulting to host when the directive is absent.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Changed in `270bd2b`: `file_systems` is now `Option<Vec<FileSystemKind>>`, with `None` representing the implicit host filesystem.

Makefile
sonukapoor: Dropped in `270bd2b`.

lib/wasix/tests/wasm_tests/mod.rs
sonukapoor: Updated in `270bd2b`: `FileSystems` now handles `all` directly and otherwise splits only on commas before parsing each `FileSystemKind`.

marxin (approved):
Thanks, I put a bit of clean up on top of your changes and we're ready to go.

sonukapoor:
> Thanks, I put a bit of clean up on top of your changes and we're ready to go.

Thank you @marxin

sonukapoor:
@marxin, please assign the next issue to me. I want to work on a feature, if possible.

sonukapoor:
@marxin Is there any new issue I can contribute to?

marxin:
There are a bunch of performance issues we've got, you might be interested in these:
https://github.com/wasmerio/wasmer/issues?q=is%3Aissue%20state%3Aopen%20author%3Agaaraw

sonukapoor:
I looked through the list. Would #6555 be a useful one for me to take a first pass on?

It looks unassigned, and my initial scope would be limited: reproduce it on current `main`, compare LLVM/Cranelift behavior, check whether the repeated `data.drop` / `elem.drop` cost is still present, and summarize where the time appears to go before proposing any change.

I avoided the issues already assigned to you or Arshia, but if another one would be more useful, I’m happy to follow your preference.

marxin:
Sure, let's start with this one, thanks.

Issue #6504, opened 2026-04-21
Revamp Artifact format (Linux LLVM)


marxin:
Currently, we serialize compiler-produced object files using the `rkyv` crate and store them on disk. At startup, we load these objects, arrange them in memory, and apply relocations. This approach is suboptimal and results in complex logic, particularly when constructing the `.text` section and unwind information.

Instead, we should delegate this responsibility to a linker and produce a final shared library, which can be mapped directly at runtime in a much simpler and more efficient way.

dhir1007:
hi @marxin, this enhancement look really good, can you share a bit more about the point where u mention that the current approach is complex and that we should create a final shared library.

marxin:
Sure - right now we bundle all the `.o` object files into an archive (similar to a `.a`). At startup, we map those objects into memory and perform symbol relocation ourselves - essentially doing the work a linker like `lld` would handle automatically if we were linking a shared library directly.

dhir1007:
thanks for the explanation! that makes sense - using lld to produce 
a proper shared library and letting the os handle mapping is much 
cleaner than managing relocations manually. is this something you're 
planning to tackle yourself for v8.0, or is there a specific piece 
of it that would be suitable for a contributor to help with?

marxin:
> is there a specific piece of it that would be suitable for a contributor to help with?

None that come to mind.

marxin:
Implemented on the current `main` branch.

PR #6651, opened 2026-05-30
perf(vm): per-thread Mmap pool with madvise(DONTNEED) reset

BlobMaster41:
Adds a per-thread pool of idle anonymous-private `Mmap` regions in `wasmer-vm`. Linear-memory mappings that drop are reset via `madvise(MADV_DONTNEED)` and stored thread-locally instead of being `munmap`'d; subsequent `Mmap::accessible_reserved` calls of the same shape pick them up before going to the kernel.

For a per-request wasm-host workload that builds a fresh `Store` and `Instance` per call and the wasm has a linear memory, this fixes a hard kernel scaling wall: at 28 threads, throughput on a memory-touching wasm module goes from 60 k to 1.15 M ops/sec (19x). The pool is Linux-only via `cfg`; non-Linux targets are stubs that fall through to the existing `mmap`/`munmap` path.

No public API change, no serialized format change, no behaviour change for any wasm program.

> [!CAUTION]
> This PR touches the `Mmap` allocation and `Drop` paths in `wasmer-vm`, which back the linear memory of every wasm Instance. Please confirm that I thought of every security-critical path. The Safety section walks every invariant the change depends on, and the `mmap_pool::tests::*` suite includes an adversarial test for each (including a `fork`/`SIGSEGV` check that the guard is actually restored).

## Why

The standard wasmer usage pattern for HTTP-style workloads is "compile once, instantiate per request". Each `Instance::new` for a memory-using wasm module ends up at `Mmap::accessible_reserved`, which calls `mmap(2)` to reserve and commit the linear-memory range. Each `Instance` drop unmaps it.

Both `mmap` and `munmap` take `mm_struct.mmap_lock` write-locked. The lock is process-wide. At 28 threads doing fresh per-request instantiate-and-call against a wasm with a 1-page linear memory, the lock spends most of its time held write-locked by some thread. Per-thread throughput collapses 40x compared to single-thread, and aggregate throughput at 28 threads is lower than at 1 thread.

A `perf record` at 28 threads on a stock build:

```
14.21%  swapper          [k] intel_idle              ← CPUs blocked
12.79%  bench            [k] osq_lock                ← rwsem optimistic spin
 8.40%  swapper          [k] intel_idle_irq
 2.98%  bench            [k] rwsem_optimistic_spin
 2.04%  bench            [k] zap_pmd_range           ← munmap page-table teardown
 1.82%  bench            [k] rwsem_spin_on_owner
 ...
```

Userland code is nowhere in the top 10 symbols. The whole box is serialized on the kernel's mm semaphore.

A standalone microbench of `mmap+munmap` vs `madvise(MADV_DONTNEED)` on the same region at 28 threads gives a 6x throughput improvement for the second pattern, which is the design point this PR builds on:

```
mmap+munmap      28T:   157 k ops/sec  (  5.6 k/thread)
madvise_dontneed 28T:   995 k ops/sec  ( 35.5 k/thread)
```

## The fix

`lib/vm/src/mmap_pool.rs` (new, ~280 lines): per-thread `HashMap<PoolKey, Vec<Mmap>>` indexed by `(accessible_size, mapping_size)`. The pool is held in a thread-local `RefCell`.

On `Mmap` drop, an anonymous-private mapping (no backing file, no `MmapType::Shared`) is reset and pushed into its bucket instead of being `munmap`'d. On `Mmap::accessible_reserved` the pool is checked before the kernel.

Reset has two cases:

* **No `memory.grow` happened during the previous tenant's lifetime.** The accessible range was the only part with backing pages; the guard range is still PROT_NONE from the original `accessible_reserved`. `madvise(MADV_DONTNEED)` on `[0..accessible_size)` releases backing physical pages so the next access faults in fresh zero pages from the kernel allocator. No `mprotect` is needed, which avoids the cross-CPU TLB-shootdown IPI that `mprotect` on a multi-core-resident VMA triggers.
* **`memory.grow` extended RW access past `accessible_size`.** The previous tenant could have written secret data to pages above `accessible_size`. `madvise(MADV_DONTNEED)` covers the entire mapping (the kernel skips ranges without present PTEs efficiently, so this is cheap even for sparse 4 GiB/64 KiB mappings). Then `mprotect(PROT_NONE)` on `[accessible_size..total_size)` restores the guard.

Whether growth happened is tracked by a new `extended_beyond_accessible: bool` field on `Mmap`, set in `make_accessible` only when `start + len > accessible_size`. The initial-setup call into `make_accessible(0, accessible_size)` from `accessible_reserved` leaves the flag false because its end is exactly `accessible_size`, not greater.

## Safety

The pool is shared across wasm Instances within a single process. Most wasm hosts run untrusted code, so an incomplete reset is a tenant-isolation bug. Every assumption is documented inline in `mmap_pool.rs`; the high points:

### Reset must produce a zero-filled mapping

`MADV_DONTNEED` on a `MAP_PRIVATE | MAP_ANON` mapping on Linux releases backing pages so subsequent accesses fault in fresh zero pages from the kernel page allocator. The kernel guarantees that any page handed to userland is freshly zeroed; old contents cannot leak. See `man 2 madvise` (`MADV_DONTNEED`) and `mm/madvise.c` (`madvise_dontneed_single_vma` → `zap_page_range`).

Test: `pool_reuse_is_zero_filled` writes a sentinel pattern across the accessible range, drops, takes again, asserts every byte reads as zero.

### Guard pages must be restored even when growth happened

`make_accessible` calls `mprotect(PROT_READ | PROT_WRITE)` to extend the accessible range but does not update `Mmap::accessible_size`. If a mapping that had been extended via `make_accessible` were pooled without restoring the guard, the next caller would see the same `accessible_size` they originally requested but pages above that boundary would still be RW. A wasm program could read past its declared memory.

Mitigation: when `extended_beyond_accessible` is true, the pool `mprotect(PROT_NONE)`s the entire `[accessible_size..total_size)` range back to guard.

Test: `pool_reuse_restores_guard` extends RW access to the full mapping, writes a sentinel into the extended range, drops; then takes a fresh mapping of the same shape and forks a child that writes into what should be the guard region. The parent asserts the child died from `SIGSEGV`.

### Reset must zero pages the previous tenant wrote above `accessible_size`

This was a real vulnerability caught during the audit. `mprotect` does not zero pages. A previous tenant calling `memory.grow`, writing secret data to the extended range, then dropping, would leave the physical pages allocated and full of secret data. After guard restoration, a future tenant could call `memory.grow` to mprotect the same pages back to RW and read the secrets directly.

Fix: in the `extended_beyond_accessible == true` path, `madvise(MADV_DONTNEED)` is run on the entire mapping before `mprotect`. This releases backing pages for the previously-extended region too. The future tenant's growth fault now goes to a fresh zero page.

Test: `pool_reuse_no_leak_across_grown_range` exercises exactly this attack path. Tenant A extends, writes a sentinel into the extended range, drops. Tenant B takes the same shape, grows again, reads. Every byte must be zero.

### Shared and file-backed mappings are not pooled

Another thread or process could hold a live alias to a `MAP_SHARED` or file-backed mapping. The exclusion is enforced by `if backing_file.is_none() && memory_type == MmapType::Private` in `accessible_reserved` and by setting `Mmap::poolable = false` on every other path.

### Thread-exit destruction

When a thread exits, its TLS pool drops. The pool's `Drop` sets `draining = true` before letting its `buckets` field drop. Each `Mmap` that drops off the inner `Vec`s sees `draining = true` via a `try_borrow_mut` on the same `RefCell` and falls through to direct `munmap` instead of trying to re-insert into the pool that is currently being torn down.

Test: `thread_exit_drains_pool_safely` spawns a thread that builds and drops 32 pooled mappings, then joins; the parent thread's pool keeps working afterwards.

### Reset partial failure

If `madvise` succeeds and `mprotect` later fails, `reset` returns `Err`, `try_pool` returns `Some(mmap)`, and `Mmap::drop` falls through to direct `munmap`. The mapping is unmapped instead of pooled, so a partially-reset region can never be observed by a future tenant.

### Send / Sync invariants

`Mmap` was already `Send + Sync` (asserted by `_assert_send_sync::<Mmap>()` at the end of `mmap.rs`). The pool is per-thread (`thread_local!`), so pool data is never crossed-thread; cross-thread Mmap ownership transfer is unaffected. Cross-thread leak through Send is tested by `pool_cross_thread_send_no_leak`.

### Non-Linux platforms

The entire pool module is `#[cfg(target_os = "linux")]`. On macOS, FreeBSD, NetBSD, and Windows the pool functions are stubs (`try_take` always returns `None`, `try_pool` always returns `Some(mmap)`), so callers fall back to the existing `mmap`/`munmap` path. `MADV_DONTNEED` semantics differ across BSD lineages and the zero-fill guarantee has not been verified there.

## Tests

Nine new unit tests in `lib/vm/src/mmap_pool.rs`. Three of them are the security invariants quoted above; the other six exercise lifecycle and edge cases:

```
test mmap_pool::tests::pool_reuse_is_zero_filled              ... ok
test mmap_pool::tests::pool_reuse_restores_guard              ... ok
test mmap_pool::tests::pool_reuse_no_leak_across_grown_range  ... ok
test mmap_pool::tests::pool_round_trip_preserves_layout       ... ok
test mmap_pool::tests::pool_does_not_reuse_across_shapes      ... ok
test mmap_pool::tests::pool_bucket_overflow_unmaps_excess     ... ok
test mmap_pool::tests::pool_cross_thread_send_no_leak         ... ok
test mmap_pool::tests::pool_stress_no_leak_under_concurrency  ... ok
test mmap_pool::tests::thread_exit_drains_pool_safely         ... ok
```

The stress test is the heaviest of the bunch: 28 threads, three distinct `(accessible, mapping)` shapes, 2000 iterations per thread, with `memory.grow`-style extension on 1 of every 3 iterations. Each iteration's accessible (and grown) range is byte-checked for the all-zero invariant before the iteration writes its own thread-unique sentinel pattern. 168 000 cycles total, completes in under half a second, zero leaks observed.

```
$ cargo test -p wasmer-vm
test result: ok. 37 passed; 0 failed
```

(28 pre-existing wasmer-vm unit tests + 9 new pool tests.)

## Numbers

Intel i9-10940X (14 cores / 28 SMT threads), CPU governor `performance`, glibc allocator, otherwise idle box. Each measurement is a 5-second window after a per-thread warmup; the harness is the standalone reproducer at the bottom of this PR.

### Workload A: memory-store-heavy wasm without `memory.grow`

```
(module
    (memory 1)
    (func (export "store_100_i32") (param $val i32)
        ;; 100 i32 stores into linear memory at offsets 0, 4, 8, ...
        ...))
```

Per-request lifecycle: build fresh `Store`, instantiate, call, drop.

| threads | upstream main | this PR | improvement |
|---|---|---|---|
| 1 | 86 k | 378 k | 4.4x |
| 4 | 74 k | 566 k | 7.6x |
| 8 | 67 k | 865 k | 12.9x |
| 16 | 61 k | 1.15 M | 18.7x |
| 28 | 60 k | **1.15 M** | **19.1x** |

Per-thread at 28T: 2.1 k → 41 k (19.5x). The workload now scales the way an in-register workload does.

### Workload B: realistic-shape wasm with `memory.grow`, globals, locals, function calls

Initial memory of 2 pages, two `memory.grow` calls per top-level invocation (one into the original guard, one past the first grow), three mutable globals (`i32` / `i64` / `f64`) written and read, many locals, two internal function calls (`$mix` is an fnv1a-style step and `$sum` is a 256-i32 reducer), a 32-iter stir loop deriving an accumulator from the seed, a 256-i32 write loop into linear memory, mixed-type stores (`i8` / `i16` / `i32` / `i64` / `f32` / `f64`) at distinct offsets, a 1024-byte write into the grown range, and a return value composed from every side effect so dead-code elimination cannot fold the body away.

```wat
(module
    (memory $mem 2)
    (global $g_i32 (mut i32) (i32.const 0))
    (global $g_i64 (mut i64) (i64.const 0))
    (global $g_f64 (mut f64) (f64.const 0))

    (func $mix (param $h i32) (param $b i32) (result i32)
        local.get $h
        local.get $b
        i32.xor
        i32.const 16777619
        i32.mul)

    (func $sum (param $start i32) (result i32)
        (local $i i32)
        (local $acc i32)
        (loop $L
            local.get $acc
            local.get $start
            local.get $i
            i32.const 4
            i32.mul
            i32.add
            i32.load
            i32.add
            local.set $acc
            local.get $i
            i32.const 1
            i32.add
            local.tee $i
            i32.const 256
            i32.lt_s
            br_if $L)
        local.get $acc)

    (func (export "stress") (param $seed i32) (result i32)
        (local $i i32)
        (local $acc i32)
        (local $acc64 i64)
        (local $accf f64)
        (local $offset i32)
        (local $tmp i32)

        local.get $seed
        local.set $acc
        (loop $L_stir
            local.get $acc
            local.get $i
            call $mix
            local.set $acc
            local.get $i
            i32.const 1
            i32.add
            local.tee $i
            i32.const 32
            i32.lt_s
            br_if $L_stir)

        i32.const 1
        memory.grow
        drop

        i32.const 0
        local.set $i
        (loop $L_w
            local.get $i
            i32.const 4
            i32.mul
            local.get $acc
            local.get $i
            i32.xor
            i32.store
            local.get $i
            i32.const 1
            i32.add
            local.tee $i
            i32.const 256
            i32.lt_s
            br_if $L_w)

        i32.const 4096
        local.get $acc
        i32.store8
        i32.const 4098
        local.get $acc
        i32.store16
        i32.const 4100
        local.get $acc
        i32.store
        i32.const 4108
        local.get $acc
        i64.extend_i32_s
        i64.const 7919
        i64.mul
        local.tee $acc64
        i64.store
        i32.const 4116
        local.get $acc
        f64.convert_i32_s
        f64.const 2.718281828459045
        f64.mul
        local.tee $accf
        f64.store
        i32.const 4124
        local.get $accf
        f64.const 2.0
        f64.div
        f32.demote_f64
        f32.store

        i32.const 1
        memory.grow
        drop
        i32.const 200000
        local.set $offset
        i32.const 0
        local.set $i
        (loop $L_grown
            local.get $offset
            local.get $i
            i32.add
            local.get $i
            i32.const 0xFF
            i32.and
            i32.store8
            local.get $i
            i32.const 1
            i32.add
            local.tee $i
            i32.const 1024
            i32.lt_s
            br_if $L_grown)

        local.get $acc
        global.set $g_i32
        local.get $acc64
        global.set $g_i64
        local.get $accf
        global.set $g_f64

        i32.const 0
        call $sum
        local.set $tmp

        i32.const 4096
        i32.load8_u
        i32.const 4098
        i32.load16_u
        i32.add
        i32.const 4100
        i32.load
        i32.add
        i32.const 4108
        i64.load
        i32.wrap_i64
        i32.add
        i32.const 4116
        i64.load
        i32.wrap_i64
        i32.add
        i32.const 4124
        i32.load
        i32.add
        local.get $tmp
        i32.add
        i32.const 200000
        i32.load8_u
        i32.add
        i32.const 200512
        i32.load8_u
        i32.add
        global.get $g_i32
        i32.add
        global.get $g_i64
        i32.wrap_i64
        i32.add
        global.get $g_f64
        i64.reinterpret_f64
        i32.wrap_i64
        i32.add))
```

Driver: the same reproducer template from the bottom of this PR, with `WAT` set to the module above and the function signature changed to `<i32, i32>` (it takes the seed and returns the combined accumulator).

| threads | upstream main | this PR | improvement |
|---|---|---|---|
| 1 | 56.7 k | 69.0 k | 1.22x |
| 4 | 56.3 k | 122.3 k | 2.17x |
| 8 | 53.7 k | 99.1 k | 1.85x |
| 16 | 53.8 k | 104.2 k | 1.94x |
| 28 | 54.0 k | **107.3 k** | **1.99x** |

The relative improvement is smaller because the in-wasm body dominates per-call cost. Per-call at 1T is 14.5 µs on this PR vs 17.6 µs on main; the pool saves the same ~3 µs of `Instance::new`+`Instance::drop` mmap overhead, but on this workload the remaining ~14 µs of wasm body is what bounds the cycle. Workloads with a bigger wasm body (a JSON-parsing handler, a CosmWasm contract, etc.) will see the pool's ~3 µs as a smaller fraction of total cost; the absolute throughput improvement at 28 threads is still around 2x for any per-request shape.

Post-patch hot symbols on workload B at 28T show the remaining bottleneck is no longer the Instance-boundary `mmap_lock`. It is now the `mprotect` that `memory.grow` does inside the wasm runtime to extend the accessible range:

```
14.5%  intel_idle
 3.6%  zap_pmd_range
 3.1%  smp_call_function_many_cond     ← cross-CPU IPI from mprotect
 2.6%  osq_lock
 2.3%  native_queued_spin_lock_slowpath
```

That cost is not addressable from inside this PR; it would require either a different memory-protection model or pre-mapping the maximum range RW at `Instance::new`.

## How to reproduce

Drop the following into `examples/repro_mmap_pool.rs` in a crate that depends on `wasmer`, `wasmer-compiler`, `wasmer-compiler-singlepass`, and `wat`. Override `THREADS=` and `SECS=` to vary.

```rust
use std::sync::{Arc, atomic::{AtomicBool, AtomicU64, Ordering}, Barrier};
use std::thread;
use std::time::{Duration, Instant};

use wasmer::{imports, EngineBuilder, Instance, Module, Store};
use wasmer_compiler_singlepass::Singlepass;

const WAT: &str = r#"
(module
    (memory 1)
    (func (export "store_100_i32") (param $val i32)
        (local $i i32)
        (loop $L
            local.get $i
            local.get $val
            i32.store
            local.get $i
            i32.const 4
            i32.add
            local.tee $i
            i32.const 400
            i32.lt_s
            br_if $L)))
"#;

fn main() {
    let threads: usize = std::env::var("THREADS")
        .ok().and_then(|s| s.parse().ok()).unwrap_or(28);
    let secs: u64 = std::env::var("SECS")
        .ok().and_then(|s| s.parse().ok()).unwrap_or(5);

    let engine = EngineBuilder::new(Singlepass::new()).engine();
    let wasm = wat::parse_str(WAT).unwrap();
    let store = Store::new(engine.clone());
    let module = Arc::new(Module::new(&store, &wasm).unwrap());
    drop(store);

    let barrier = Arc::new(Barrier::new(threads + 1));
    let stop = Arc::new(AtomicBool::new(false));
    let total = Arc::new(AtomicU64::new(0));

    let handles: Vec<_> = (0..threads).map(|_| {
        let engine = engine.clone();
        let module = module.clone();
        let barrier = barrier.clone();
        let stop = stop.clone();
        let total = total.clone();
        thread::spawn(move || {
            // warmup
            for _ in 0..10_000 {
                let mut store = Store::new(engine.clone());
                let inst = Instance::new(&mut store, &module, &imports!{}).unwrap();
                let f = inst.exports
                    .get_typed_function::<i32, ()>(&store, "store_100_i32")
                    .unwrap();
                let _ = f.call(&mut store, 0x5A5A_5A5A_u32 as i32).unwrap();
            }
            barrier.wait();

            let mut local = 0u64;
            while !stop.load(Ordering::Relaxed) {
                let mut store = Store::new(engine.clone());
                let inst = Instance::new(&mut store, &module, &imports!{}).unwrap();
                let f = inst.exports
                    .get_typed_function::<i32, ()>(&store, "store_100_i32")
                    .unwrap();
                f.call(&mut store, 0x5A5A_5A5A_u32 as i32).unwrap();
                local += 1;
            }
            total.fetch_add(local, Ordering::Relaxed);
        })
    }).collect();

    barrier.wait();
    let start = Instant::now();
    thread::sleep(Duration::from_secs(secs));
    stop.store(true, Ordering::Relaxed);
    for h in handles { h.join().unwrap(); }
    let elapsed = start.elapsed();
    let ops = total.load(Ordering::Relaxed);
    println!(
        "{} threads, {} ops in {:.3}s = {:.0} ops/sec ({:.0} ops/sec/thread)",
        threads, ops, elapsed.as_secs_f64(),
        ops as f64 / elapsed.as_secs_f64(),
        (ops as f64 / elapsed.as_secs_f64()) / threads as f64,
    );
}
```

```bash
git checkout main
cargo run --release --example repro_mmap_pool
# Expect ~60 k ops/sec at 28 threads

git checkout mmap-tls-pool
cargo run --release --example repro_mmap_pool
# Expect ~1.15 M ops/sec at 28 threads
```

For hot-symbol verification:

```bash
THREADS=28 perf record -F 2999 -a -g -- target/release/examples/repro_mmap_pool
perf report --stdio --no-children --percent-limit 1.0 | head -20
# On main: osq_lock at ~13%, rwsem_optimistic_spin at ~3%, zap_pmd_range at ~2%.
# On this PR: those symbols drop out of the hot list.
```

## Compatibility

* No public API change. `Mmap::accessible_reserved`, `Mmap::with_at_least`, and `Mmap::make_accessible` keep their signatures and external semantics.
* No serialization format change. The pool is runtime state; it does not appear in any persisted artifact.
* No behaviour change visible to wasm programs. The pool returns mappings that are observationally identical to fresh `mmap` allocations: zero-filled accessible range, PROT_NONE guard region, same `(accessible_size, mapping_size)` shape.
* Linux only. macOS, FreeBSD, NetBSD, and Windows fall through to the existing path.

## Composition with other in-flight perf PRs

This PR is independent of #6640 (`tls-stack-cache`), #6649 (`vmoffsets-cache`), and #6650 (`storeid-thread-local`). It does not require any of them, and it stacks additively on top of all of them.

## Out of scope

Items that show up in the post-patch profile and are addressable independently in follow-up work:

* `memory.grow` inside the wasm runtime still does `mprotect`, which triggers cross-CPU TLB shootdowns. Would require either pre-mapping the maximum range RW at `Instance::new` (loses the guard model) or a different protection scheme.
* `Store::new` internals (signature registry allocation, per-store bookkeeping). Roughly 6% at 28 threads on the realistic workload.
* glibc `malloc/free` contention on the `VMContext` allocation inside `InstanceAllocator::new`. A per-thread `VMContext` buffer pool with the same pattern as this PR could remove it.

Happy to follow up on any of these.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7546,6 +7546,7 @@ dependencies = [
  "cc",
  "cfg-if",
  "corosensei",
+ "criterion",
  "crossbeam-queue",
  "dashmap",
  "enum-iterator",
--- a/lib/vm/Cargo.toml
+++ b/lib/vm/Cargo.toml
@@ -55,6 +55,15 @@ windows-sys = { workspace = true, features = [
 cc.workspace = true
 rustversion.workspace = true
 
+# Workspace default has default-features = false (no plotters HTML
+# report dep). Bench still runs and prints stats to stdout.
+[dev-dependencies]
+criterion = { workspace = true }
+
+[[bench]]
+name = "mmap_pool_alloc"
+harness = false
+
 [badges]
 maintenance = { status = "actively-developed" }
 
--- a/lib/vm/Cargo.toml
+++ b/lib/vm/Cargo.toml
@@ -55,10 +55,8 @@ windows-sys = { workspace = true, features = [
 cc.workspace = true
 rustversion.workspace = true
 
-# Workspace default has default-features = false (no plotters HTML
-# report dep). Bench still runs and prints stats to stdout.
 [dev-dependencies]
-criterion = { workspace = true }
+criterion.workspace = true
 
 [[bench]]
 name = "mmap_pool_alloc"
--- a/lib/vm/benches/mmap_pool_alloc.rs
+++ b/lib/vm/benches/mmap_pool_alloc.rs
@@ -1,0 +1,69 @@
+//! Micro-benchmarks for the per-thread `Mmap` pool.
+//!
+//! The pool's value comes from avoiding `mm_struct.mmap_lock` traffic
+//! under concurrent `Mmap::accessible_reserved` / drop pairs. The
+//! relevant signals are:
+//!
+//!   1. single-thread steady state - the pool always hits after the
+//!      first iteration. Per-iteration cost is dominated by the
+//!      `MADV_DONTNEED` reset rather than `mmap`/`munmap`.
+//!   2. multi-thread contention - N threads independently allocating
+//!      + dropping. The previous design serialised on the process-
+//!      wide mmap_lock; the per-thread pool should give a near-flat
+//!      scaling curve up to the point where the underlying `madvise`
+//!      syscall itself becomes a bottleneck.
+//!
+//! Run with:
+//!
+//! ```text
+//! cargo bench -p wasmer-vm --bench mmap_pool_alloc
+//! ```
+//!
+//! Linux-only payoff. The pool is gated on `target_os = "linux"`; on
+//! other targets the benches still build but exercise the fallback
+//! direct-mmap path.
+
+use std::thread;
+
+use criterion::{Criterion, criterion_group, criterion_main};
+use wasmer_vm::{Mmap, MmapType};
+
+/// One wasm linear memory's worth of mapping. Picks a size in the
+/// same order of magnitude as a typical AssemblyScript tenant's
+/// initial allocation.
+const MAPPING_SIZE: usize = 64 * 1024;
+
+fn one_alloc_free_cycle() {
+    let m = Mmap::accessible_reserved(MAPPING_SIZE, MAPPING_SIZE, None, MmapType::Private)
+        .expect("mmap");
+    drop(m);
+}
+
+fn single_thread_steady_state(c: &mut Criterion) {
+    c.bench_function("mmap_pool::alloc_drop/single_thread", |b| {
+        b.iter(one_alloc_free_cycle);
+    });
+}
+
+fn multi_thread_contention(c: &mut Criterion) {
+    let mut group = c.benchmark_group("mmap_pool::alloc_drop/multi_thread");
+    for &threads in &[1usize, 2, 4, 8] {
+        group.bench_function(format!("{threads}_threads"), |b| {
+            b.iter(|| {
+                thread::scope(|s| {
+                    for _ in 0..threads {
+                        s.spawn(|| {
+                            for _ in 0..64 {
+                                one_alloc_free_cycle();
+                            }
+                        });
+                    }
+                });
+            });
+        });
+    }
+    group.finish();
+}
+
+criterion_group!(benches, single_thread_steady_state, multi_thread_contention);
+criterion_main!(benches);
--- a/lib/vm/src/lib.rs
+++ b/lib/vm/src/lib.rs
@@ -23,6 +23,7 @@ mod imports;
 mod instance;
 mod memory;
 mod mmap;
+mod mmap_pool;
 mod probestack;
 mod sig_registry;
 mod store;
--- a/lib/vm/src/mmap.rs
+++ b/lib/vm/src/mmap.rs
@@ -103,9 +103,7 @@ impl Mmap {
         #[cfg(target_os = "linux")]
         {
             if backing_file.is_none() && memory_type == MmapType::Private {
-                if let Some(pooled) =
-                    crate::mmap_pool::try_take(accessible_size, mapping_size)
-                {
+                if let Some(pooled) = crate::mmap_pool::try_take(accessible_size, mapping_size) {
                     return Ok(pooled);
                 }
             }
@@ -493,12 +491,7 @@ impl Drop for Mmap {
                                 unpooled.total_size_for_pool(),
                             )
                         };
-                        assert_eq!(
-                            r,
-                            0,
-                            "munmap failed: {}",
-                            io::Error::last_os_error()
-                        );
+                        assert_eq!(r, 0, "munmap failed: {}", io::Error::last_os_error());
                         std::mem::forget(unpooled);
                         return;
                     }
--- a/lib/vm/src/mmap.rs
+++ b/lib/vm/src/mmap.rs
@@ -67,10 +67,24 @@ impl Mmap {
     }
 
     /// Create a new `Mmap` pointing to at least `size` bytes of page-aligned accessible memory.
+    ///
+    /// Mappings returned from this constructor are excluded from the
+    /// per-thread Mmap pool. The only caller is `CodeMemory`, which
+    /// calls `region::protect` directly to make pages read-execute
+    /// after writing function bodies in. That protection change happens
+    /// outside the `Mmap` API surface, so the pool cannot track it; a
+    /// recycled mapping would still be PROT_READ_EXECUTE when the next
+    /// caller tried to write to it, faulting on the first store. See
+    /// `lib/compiler/src/engine/code_memory.rs::publish`.
     pub fn with_at_least(size: usize) -> Result<Self, String> {
         let page_size = region::page::size();
         let rounded_size = size.next_multiple_of(page_size);
-        Self::accessible_reserved(rounded_size, rounded_size, None, MmapType::Private)
+        let mut m = Self::accessible_reserved(rounded_size, rounded_size, None, MmapType::Private)?;
+        #[cfg(target_os = "linux")]
+        {
+            m.poolable = false;
+        }
+        Ok(m)
     }
 
     /// Create a new `Mmap` pointing to `accessible_size` bytes of page-aligned accessible memory,
--- a/lib/vm/src/mmap.rs
+++ b/lib/vm/src/mmap.rs
@@ -101,12 +101,11 @@ impl Mmap {
         // bypasses the pool and falls through to a fresh mmap. See
         // `mmap_pool` for the security argument.
         #[cfg(target_os = "linux")]
+        if backing_file.is_none()
+            && memory_type == MmapType::Private
+            && let Some(pooled) = crate::mmap_pool::try_take(accessible_size, mapping_size)
         {
-            if backing_file.is_none() && memory_type == MmapType::Private {
-                if let Some(pooled) = crate::mmap_pool::try_take(accessible_size, mapping_size) {
-                    return Ok(pooled);
-                }
-            }
+            return Ok(pooled);
         }
 
         // If there is a backing file, resize the file so that its at least
--- a/lib/vm/src/mmap.rs
+++ b/lib/vm/src/mmap.rs
@@ -22,6 +22,21 @@ pub struct Mmap {
     accessible_size: usize,
     #[cfg_attr(target_os = "windows", allow(dead_code))]
     sync_on_drop: bool,
+    /// True iff this Mmap was created in a way that makes it safe to
+    /// return to the per-thread pool on Drop: anonymous PRIVATE mapping,
+    /// no backing file, no MAP_SHARED. Set to false for any mapping the
+    /// pool cannot guarantee tenant isolation for.
+    #[cfg_attr(target_os = "windows", allow(dead_code))]
+    poolable: bool,
+    /// True iff `make_accessible` was called with a range that extends
+    /// beyond the original `accessible_size`. Tracks whether the
+    /// guard region has been promoted to PROT_RW so the pool knows
+    /// whether it needs to mprotect it back to PROT_NONE on recycle.
+    /// Initial setup calls into `make_accessible(0, accessible_size)`
+    /// during construction; that call does NOT extend beyond the
+    /// declared accessible size, so it leaves this flag false.
+    #[cfg_attr(target_os = "windows", allow(dead_code))]
+    extended_beyond_accessible: bool,
 }
 
 /// The type of mmap to create
@@ -46,6 +61,8 @@ impl Mmap {
             total_size: 0,
             accessible_size: 0,
             sync_on_drop: false,
+            poolable: false,
+            extended_beyond_accessible: false,
         }
     }
 
@@ -77,6 +94,21 @@ impl Mmap {
         // special-case that.
         if mapping_size == 0 {
             return Ok(Self::new());
+        }
+
+        // Per-thread Mmap pool fast path. Only the simple anonymous
+        // private case is poolable; any backing file or shared mapping
+        // bypasses the pool and falls through to a fresh mmap. See
+        // `mmap_pool` for the security argument.
+        #[cfg(target_os = "linux")]
+        {
+            if backing_file.is_none() && memory_type == MmapType::Private {
+                if let Some(pooled) =
+                    crate::mmap_pool::try_take(accessible_size, mapping_size)
+                {
+                    return Ok(pooled);
+                }
+            }
         }
 
         // If there is a backing file, resize the file so that its at least
@@ -145,6 +177,8 @@ impl Mmap {
                 total_size: mapping_size,
                 accessible_size,
                 sync_on_drop: memory_fd != -1 && memory_type == MmapType::Shared,
+                poolable: memory_fd == -1 && memory_type == MmapType::Private,
+                extended_beyond_accessible: false,
             }
         } else {
             // Reserve the mapping size.
@@ -167,6 +201,8 @@ impl Mmap {
                 total_size: mapping_size,
                 accessible_size,
                 sync_on_drop: memory_fd != -1 && memory_type == MmapType::Shared,
+                poolable: memory_fd == -1 && memory_type == MmapType::Private,
+                extended_beyond_accessible: false,
             };
 
             if accessible_size != 0 {
@@ -222,6 +258,8 @@ impl Mmap {
                 total_size: mapping_size,
                 accessible_size,
                 sync_on_drop: false,
+                poolable: false,
+                extended_beyond_accessible: false,
             }
         } else {
             // Reserve the mapping size.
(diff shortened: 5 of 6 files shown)


BlobMaster41:
Might be possible to also do the same thing on windows and other OS / distro. More complex to implement.

marxin:
Here's a polished rewrite that improves grammar, tone, and clarity while keeping the message constructive:

Thank you for the contribution. After reviewing the proposal, I've decided not to merge this PR for now.

1. We're planning a significant refactoring of the Artifact format (#6504), with the goal of enabling a direct mapping of executables into memory. This approach will naturally distribute `mmap` activity over time through on-demand page faults, which may address some of the concerns this PR is targeting.
2. The proposed change could potentially introduce a new attack vector, where stale memory mappings might become accessible.
3. I believe we may be able to mitigate the lock contention through THP (Transparent Huge Pages), which is already part of our roadmap.
4. I'd also like to see benchmark results from a more realistic workload where a non-trivial amount of time is spent executing the workload rather than exiting immediately.

Overall, I appreciate the idea and the effort that went into this contribution. It's an interesting direction, and we may revisit it in the future.

BlobMaster41:
All right, ill rebench once #6504 is done, for context, im putting wasmer behind an HTTP server and handling everything as the backend with it. Without this PR, performance are highly impacted. We are talking 60k possible req/s vs 1 milion (on a static page, dynamic, recomputed every request).

From my benchmarks, over a 8kbit payload im able to push 670176.14 req/s with wasmer backing it (this is via dpdk, bypassing the kernel, otherwise it's impossible to hit such result, I have TSO off so expect ~1.5,2x improvement with TSO on, (offload compute to the NIC).

```
  39 threads and 30000 connections
    Latency     1.95ms   21.92ms   1.79s    99.61%
    Req/Sec    20.02k    19.34k   67.60k    76.93%
  Latency Distribution
     50%  490.00us
     75%    2.05ms
     90%    4.04ms
     99%   12.22ms
  20172913 requests in 30.10s, 159.84GB read
Requests/sec: 670176.14
Transfer/sec:      5.31GB
```

And over a 100mbit payload:

```
  39 threads and 500 connections
    Latency     1.82s    90.58ms   2.00s    60.84%
    Req/Sec     3.69      5.07    40.00     82.48%
  Latency Distribution
     50%    1.82s
     75%    1.91s
     90%    1.93s
     99%    1.98s
  1732 requests in 30.10s, 180.07GB read
Requests/sec:     57.54
Transfer/sec:      5.98GB
```

I am working off my own fork at the moment, but you may tag me later to check the results I have on the rework in progress.

These results are with a 100gbit NIC on my private network.

PR #6652, opened 2026-05-31
fix(virtual-fs): correct mem_fs poll_write_ready behavior

immanuwell:
# Description
`poll_write_ready()` in `mem_fs` was checking read perms, so write-only handles came back `PermissionDenied`.
`CustomFile` and `ArcFile` also got forwarded to `poll_read_ready()`, so write readiness was kinda wrong there too.

Repro before this patch:
`cargo test -p virtual-fs poll_write_ready -- --nocapture`

That fails on `main` with both regression tests.
This patch checks `writable`, calls `poll_write_ready()`, and adds the tests.


marxin (commented):
LGTM, but I would like to get a feedback from @Arshia001 or @theduke.

Arshia001 (approved):

Arshia001:
Thanks for the contribution!

PR #6667, opened 2026-06-04
Make proper atomic loads from the guest's memory in the atomics libcalls

Arshia001:
Previously, we were constructing an `AtomicPtr` and then loading from it, which does the wrong thing; it loads the _pointer_ atomically, then make a non-atomic read of the target address. This change fixes that mistake by using `AtomicU32/U64::from_ptr` which will load the target value atomically.

I tried implementing a test that would catch the regression, but couldn't build one that failed with the existing impl.

Thanks to @sadhbh-c0d3 for discovering the issue!

### DIFF
--- a/lib/vm/src/threadconditions.rs
+++ b/lib/vm/src/threadconditions.rs
@@ -1,8 +1,7 @@
 use std::{
-    sync::atomic::AtomicPtr,
     sync::{
         Arc,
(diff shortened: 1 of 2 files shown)


marxin (approved):

sadhbh-c0d3:
lgtm

PR #6669, opened 2026-06-04
Fix UDP datagram receive and add local multicast fanout

sadhbh-c0d3:
## Features

- Add high-performance local virtual UDP multicast for guest-to-guest delivery inside the same Wasmer instance.
- Use a mostly lock-free fanout path with shared packet segments and per-subscriber read queues, so multiple local sockets can receive the same multicast packet without copying payload buffers.

## Fixes

- Preserve empty UDP packets in the socket backlog so zero-length datagrams can be received by the guest.
- Make `try_recv_from()` drain the backlog before polling the host socket again, so packets already observed during readiness polling are returned correctly.


lib/virtual-net/src/client.rs
```diff
@@ -961,15 +961,19 @@ impl VirtualIoSource for RemoteSocket {
             Poll::Pending => {}
         }
         if !self.buffer_recv_with_addr.is_empty() {
-            let total = self
+            let len = self
                 .buffer_recv_with_addr
-                .iter()
+                .front()
                 .map(|a| a.data.len())
-                .sum();
```
sadhbh-c0d3, line 968: We only take one, because in UDP, user receives exactly one dgram (it's not a stream)

Arshia001 (changes requested):
The implementation and logic looks correct to me. As a rule of thumb, when making changes to WASIX, we add a test for it and make sure it failed before and passes after the change. Can you add one? AI will be super-helpful here.

lib/wasix/src/syscalls/wasi/fd_write.rs
```diff
@@ -254,14 +254,19 @@ pub(crate) fn fd_write_internal<M: MemorySize>(
...
                                             Some(timeout),
                                             nonblocking,
                                         )
-                                        .await?;
+                                        .await
+                                    {
+                                        Ok(sent) => sent,
+                                        Err(_) if sent > 0 => break,
+                                        Err(err) => return Err(err),
+                                    };
```
sadhbh-c0d3, line 269: We need to not return Err in case when there was some bytes written, as otherwise same bytes will be written again. We just need to return number of bytes that were factually written so that user will continue from where they finished.

Following principle: Some success wins over error in order to guarantee progress and determinism.

Arshia001: but also store the error that we received so we can report it on the next write I assume?

sadhbh-c0d3: @Arshia001 Good idea!

sadhbh-c0d3: I asked Codex, and I think it's right though - what you think?

Yes, the principle is right: **if a write makes positive progress, success wins for that syscall**.

But I would **not generally queue the error in `fd_write` and force-return it on the next write**.

The POSIX-ish rule is:

- If `write()` / `writev()` writes `N > 0` bytes, return `N`.
- If it writes `0` bytes and hits an error, return the error.
- If it writes some bytes and then hits `EAGAIN`, `EPIPE`, etc., the caller only gets the byte count.
- The next write is a fresh operation against the current socket state.

Why not queue the error?

- `EAGAIN` / `WouldBlock` is readiness state, not a durable error. Returning it later after progress would be stale and could confuse libuv.
- For real fatal errors like `EPIPE` / `ECONNRESET`, the socket itself should carry that failed state. The next write will naturally fail because the socket is broken.
- A manual queued error at the `fd_write` layer could poison an unrelated later write, especially if libuv already consumed the partial byte count and adjusted its buffers.
- It would duplicate responsibility: stream/socket state should own persistent errors; `fd_write` should report what happened during this call.

So the model I’d use:

```text
sent == 0, error happens      => return error
sent > 0, error happens       => return sent
later write sees same socket  => underlying socket decides success/error
```

For UDP/datagrams, different rule: datagrams are atomic. Either the whole datagram is accepted, or return an error like `EMSGSIZE`. No partial datagram success.

sadhbh-c0d3: man write() only says:

       On success, the number of bytes written is returned.  On error, -1
       is returned, and errno is set to indicate the error.

       Note that a successful write() may transfer fewer than count
       bytes.  Such partial writes can occur for various reasons; for
       example, because there was insufficient space on the disk device
       to write all of the requested bytes, or because a blocked write()
       to a socket, pipe, or similar was interrupted by a signal handler
       after it had transferred some, but before it had transferred all
       of the requested bytes.  In the event of a partial write, the
       caller can make another write() call to transfer the remaining
       bytes.  The subsequent call will either transfer further bytes or
       may result in an error (e.g., if the disk is now full).
       
but I think it aligns with that subsequent write doesn't need to fail necessarily, and returning old error would be inaccurate.

lib/virtual-net/src/host.rs
```diff
@@ -66,6 +74,253 @@ impl Default for LocalNetworking {
     }
 }
 
+const MULTICAST_RING_CAPACITY: usize = 1024;
```
sadhbh-c0d3: Using ringbuffer was my idea, but I think we need to configure this, and not use constant.

lib/virtual-net/src/multicast.rs
```diff
@@ -0,0 +1,430 @@
...
+    }
+}
+
+#[derive(Debug)]
+struct MulticastReadState {
+    read_seg: Option<Arc<PacketSeg>>,
+    read_index: usize,
+    queued_segs: usize,
+    ready_segs: SegQueue<Arc<PacketSeg>>,
+}
```
sadhbh-c0d3, line 113: To achieve almost lock-free high-performance multicast I designed a solution where:
- packet segments are pushed to SegQeueue `ready_segs` by publishers.
- subscriber reads current segment `read_seq`, and when it reaches end, pops next segment from SegQueue `ready_segs`.
- SegQueue is high-performance lock-free queue, so publisher can push without blocking subscriber.

sadhbh-c0d3:
The Unix domain socket listen bucket was reworked in
`wasix-libc` and `wasmer`. `wasix-libc` now round-trips `sockaddr_un` paths in
the WASIX address ABI, and `wasmer` now binds, listens, connects, accepts, and
reports addresses for local AF_UNIX stream sockets through an in-process virtual
listener registry. A focused C probe reached `socket`, `bind`, and `listen`
successfully.

lib/wasix/src/syscalls/wasi/fd_write.rs
Arshia001: after reading up on this, I agree. a partially successful read/write should be reported as successful, and any following operation will report the error back. Also, the error should be recovered from the underlying network impl with the next operation, storing it makes very little sense.

Arshia001 (changes requested):
This PR needs to be split up into smaller ones which implement one feature cleanly.

The unix domain socket implementation is incorrect, and unless implemented properly end-to-end, should be removed.

The multicast implementation is also incomplete and should be removed.

The change to reading from UDP sockets looks sane, but should go into its own PR.

If there are other changes, please let me know what they are, @sadhbh-c0d3.

sadhbh-c0d3:
@Arshia001 Cherry picked changes into new PR here https://github.com/wasmerio/wasmer/pull/6685

PR #6675, opened 2026-06-06
fix(cli): return an error when binfmt_misc is missing

immanuwell:
# Description
`wasmer binfmt` panics if `--binfmt-misc` points to a path that does not exist.

This is easy to hit in practice because the flag is user supplied, and `binfmt_misc` is not always mounted either. Small papercut, but the panic is kinda rough rn.

This swaps the panic for a normal CLI error and adds a regression test.

Repro:
`cargo run -p wasmer-cli --no-default-features --features 'backend,sys' -- binfmt --binfmt-misc /tmp/wasmer-binfmt-missing register`

Before:
panics in `lib/cli/src/commands/binfmt.rs`

After:
`error: /tmp/wasmer-binfmt-missing does not exist`

### DIFF
--- a/lib/cli/src/commands/binfmt.rs
+++ b/lib/cli/src/commands/binfmt.rs
@@ -64,7 +64,7 @@ impl Binfmt {
     /// execute [Binfmt]
     pub fn execute(&self) -> Result<()> {
         if !self.binfmt_misc.exists() {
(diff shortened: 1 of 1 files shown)


marxin (approved):
LGTM, thank you for the contribution.

PR #6676, opened 2026-06-08
fix(Artifact): move new enum values to the end

marxin:
@pmikolajczyk41 The libcalls enum variant was extended in #6672, but unfortunately the new values very not put at the end, but in the middle of the variant space. That broke already existing artifacts.

### DIFF
--- a/lib/types/src/libcalls.rs
+++ b/lib/types/src/libcalls.rs
@@ -32,13 +32,6 @@ pub enum LibCall {
 
     /// nearest.f64
     NearestF64,
-
-    /// sqrt.f32
-    SqrtF32,
-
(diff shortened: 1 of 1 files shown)


pmikolajczyk41:
@marxin oh, apologies for that regression. Do you maybe want me to add a test for this enum that would catch similar problems in the future? The simplest would be to assert a hardcoded list of encoded variants

marxin:
> Do you maybe want me to add a test for this enum that would catch similar problems in the future?

Probably not - we're planning some bigger Artifact format revamp, so maybe after that.

artemyarulin (approved):

PR #6681, opened 2026-06-08
fix poll_read_ready underflow when the file cursor is past EOF

netliomax25-code:
1. poll_read_ready reports the readable byte count as `len - cursor`, but the cursor can sit past EOF after a seek (POSIX allows seeking beyond the end), so the subtraction underflows.
2. This reaches the mem_fs File/OffloadedFile/ReadOnlyFile handles, the host_fs file, and the webc volume file. In release builds it wraps to a near-`usize::MAX` count that gets reported back through poll_oneoff/epoll, and in debug builds it panics with "attempt to subtract with overflow". A guest reaches it by seeking a fd past the end and then polling it for readability.
3. Switched each site to `saturating_sub`, so the count is 0 once the cursor is at or past the end, which matches how the read paths already treat EOF.

Added a mem_fs regression test that seeks past the end and checks that poll_read_ready returns 0.

### DIFF
--- a/lib/virtual-fs/src/host_fs.rs
+++ b/lib/virtual-fs/src/host_fs.rs
@@ -584,7 +584,7 @@ impl VirtualFile for File {
         };
         let _ = self.inner_std.seek(io::SeekFrom::Start(cursor));
 
-        let remaining = end - cursor;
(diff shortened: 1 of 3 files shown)


netliomax25-code:
any update?

PR #6685, opened 2026-06-08
Udp datagram receive and last err

sadhbh-c0d3:
- Preserve empty UDP packets in the socket backlog so zero-length datagrams can be received by the guest.
- Make `try_recv_from()` drain the backlog before polling the host socket again, so packets already observed during readiness polling are returned correctly.
- Support getting last socket error.


sadhbh-c0d3:
Split this into smaller PRs.

PR #6687, opened 2026-06-09
fix err_buf_len underflow in write_dl_error

netliomax25-code:
1. err_buf_len is the guest-supplied error buffer length passed to dlopen/dlsym; the truncation path computed `err_buf_len - 1`, which underflows usize when a guest passes 0 (subtract-overflow panic in debug, wraps to a bad str slice in release).
2. when err_buf_len equals the message length the writer emitted the message plus its NUL, one byte past the guest buffer.
Reserve a byte for the terminator and return early on an empty buffer. Added two regression tests covering the zero-length and exact-length cases.

### DIFF
--- a/lib/wasix/src/syscalls/wasix/dlopen.rs
+++ b/lib/wasix/src/syscalls/wasix/dlopen.rs
@@ -94,14 +94,13 @@ pub(crate) fn write_dl_error<M: MemorySize>(
         err = &err[..err_len];
     }
 
-    let mut buf = vec![0; err_len + 1];
-    buf[0..err_len].copy_from_slice(err.as_bytes());
-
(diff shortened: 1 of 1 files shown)


Arshia001:
Hey @netliomax25-code, thank you for this contribution!

Since you're already improving the function, can you please also remove the extra buffer allocation on line 97, writing the error message directly into the guest memory instead?

netliomax25-code:
Done. Dropped the temporary `Vec` and now write the message plus the trailing NUL straight into the guest slice via `as_mut()`. Both regression tests still pass.

Arshia001 (approved):

Arshia001:
Thanks! Will merge this as soon as CI turns green.

PR #6689, opened 2026-06-09
fix: `enable-serde` feature (deprecate usage in wasix, virtual-fs)

marxin:
The PR removes usage of `enable-serde` in the `wasix` and `virtual-fs` crates, where it does not serve any real use case. The feature is still preserved, but now emits a warning.

For the other crates, I fixed the relevant places in the Singlepass and Cranelift compilers.
In the future, we may want to reduce usage of this feature even further.


marxin:
Seems we have a build configuration in our CI:
```
# Check all the features compatible with the `sys` backend.
check-api-features:
	cargo check --package wasmer --features=$(subst $(space),$(comma),$(compilers)),default,artifact-size,core,enable-serde,wasmer-artifact-load,wasmer-artifact-create,static-artifact-load,static-artifact-create
```
where we enable both `enable-serde` and the compiler:  Cargo check command enables `enable-serde` only on the `wasmer` package. That propagates down to `wasmer-compiler/wasmer-vm/wasmer-`types, but stops there. The
 `wasmer-compiler-cranelift` crate's enable-serde feature is never activated.

syrusakbary (dismissed):
I believe this is needed for instaboot? @Arshia001 can confirm before moving forward

theduke (changes requested):
`serde` serialization is a very old feature.

I don't believe it works correctly at the moment, and is not actively used, at least on the `wasix` side.

I believe we can probably just remove the functionality, but potentially keep the feature as a stub , with a deprecation warning, and remove it with the next breaking change?

Maybe @syrusakbary can chime in here.

syrusakbary:
@Arshia001 will need to chime in here for instaboot requirement

Arshia001:
@syrusakbary @theduke journal code uses `rkyv` afaik. `serde` does not play into the journal implementation.

theduke:
Yeah, journal code doesn't use serde IIRC.

marxin:
Reworked the PR based on the discussion - another round of review, please.

theduke:
@marxin ~~I thought the result of our discussion was to just yoink the feature?~~

theduke (approved):
LGTM

PR #6690, opened 2026-06-09
Added sock_recv_msg, sock_send_msg

sadhbh-c0d3:
## Motivation

WASIX currently exposes POSIX `sendmsg()` / `recvmsg()` in libc, but the backing implementation only forwards byte iovecs through `sock_send`, `sock_send_to`, `sock_recv`, and `sock_recv_from`.

That means `msg_control` / `msg_controllen` is ignored. In practice this breaks POSIX ancillary data, especially `SOL_SOCKET` / `SCM_RIGHTS`, which libuv uses for IPC handle passing. EdgeJS/Node cluster and shared-handle flows need this path to behave like normal POSIX rather than silently dropping control messages.

## What Changed

This adds the first WASIX ABI/plumbing layer for message sockets:

- Adds WITX definitions for socket control-message metadata.
- Adds new WASIX syscalls:
  - `sock_send_msg`
  - `sock_recv_msg`
- Adds wasix-libc declarations and import shims.
- Adds public `cmsghdr`, `CMSG_*`, and `SCM_RIGHTS` socket header surface.
- Updates `sendmsg()` to translate POSIX control data into WASIX control data.
- Updates `recvmsg()` to translate WASIX control data back into POSIX `cmsghdr`.
- Wires Wasmer exports for the new syscalls.

The initial Wasmer implementation is intentionally strict: outgoing ancillary data currently returns `Notsup` instead of being silently dropped. Payload-only `sendmsg()` / `recvmsg()` still delegates to the existing socket send/receive paths.

## Context

This is groundwork for correct POSIX-compatible `SCM_RIGHTS` support over WASIX socket pairs. The next runtime step is replacing the current pipe-backed socketpair behavior with a real socketpair/control-message queue that can associate fd rights with stream data and install received fds into the receiving process fd table.

Related context:

- `sendmsg()` / `recvmsg()` are required by libuv IPC paths.
- libuv uses `SCM_RIGHTS` to pass handles between parent and child processes.
- Current WASIX byte-only socket syscalls cannot represent ancillary data.
- Silently ignoring `msg_control` is incorrect; unsupported control data must fail visibly until fully implemented.


Arshia001:
There are multiple issues that need to be resolved for this to work, and seeing as the implementation is also only half-done, I'll close this PR; we can get back to implementing this at some other point in time.

PR #6691, opened 2026-06-09
Duplex pipe readiness

sadhbh-c0d3:
## Summary

Fixes two WASIX runtime issues that blocked Node/EdgeJS-style child IPC over `socketpair()`:

- allows child-process fd actions to place the IPC channel on fd `3` without colliding with the virtual root/preopen fd;
- fixes `DuplexPipe` readiness delivery so epoll/poll can observe already-queued socketpair data.

## Motivation

Node/libuv cluster and child-process IPC expect the child IPC channel to be available on fd `3`. In WASIX, fd `3` can also be occupied by the virtual root/preopen entry, so spawn fd actions that duplicate the IPC socketpair endpoint onto fd `3` could fail or leave path resolution in a bad state.

After the fd setup was fixed, the child could write the IPC message, but the parent could still miss it if the data was already queued before epoll registered its interest handler. That left the parent waiting even though the `DuplexPipe` had readable bytes.

Together these issues prevented cluster/diagnostics-channel IPC from making progress.

## Changes

- Relocate a displaced preopen fd when `dup2_at(src, dst)` targets that fd, instead of refusing the operation.
- Make the virtual root fd available internally so absolute path opens can resolve from the root inode even when the virtual root is not present at its usual fd slot.
- Prime EPOLLIN readiness for `DuplexPipe` and `PipeRx` when registering epoll interest if data is already readable.
- Give socketpair-backed `DuplexPipe` inodes unique names to make debugging and tracing endpoint ownership clearer.
- Add unit coverage for prequeued `DuplexPipe` data being visible through epoll readiness.

## Context

This was found while debugging EdgeJS/QuickJS WASIX Node tests. The failing path was:

```text
sock_pair_internal()
  -> Kind::DuplexPipe

proc_spawn2 fd actions
  -> dup2 IPC child endpoint onto fd 3

child writes IPC payload on fd 3
parent waits for readable event on the other endpoint

### DIFF
--- a/lib/wasix/src/fs/mod.rs
+++ b/lib/wasix/src/fs/mod.rs
@@ -2110,7 +2110,7 @@ impl WasiFs {
         }
     }
 
-    fn virtual_root_fd(root_inode: InodeGuard) -> Fd {
+    pub(crate) fn virtual_root_fd(root_inode: InodeGuard) -> Fd {
         Fd {
             inner: FdInner {
                 rights: ALL_RIGHTS,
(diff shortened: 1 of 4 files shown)


Arshia001:
This change is wrong in principle. Moving preopen FDs to a different number will break WASIX modules' guest-side state.

PR #6693, opened 2026-06-10
fix(virtual-fs): remove anyhow from offloaded file test

immanuwell:
# Description

Small test-only fix in `virtual-fs`.

`mem_fs::offloaded_file::tests::test_offload_file` returned `anyhow::Result<()>`, but `anyhow` is optional and not enabled for `--no-default-features --features host-fs`. So a real host-fs test build just falls over.

Swapped it to `std::io::Result<()>`. Same test behavior, no runtime change, just removes that accidental feature tie-in

Repro:
`cargo test -p virtual-fs --no-default-features --features host-fs test_remove_file -- --nocapture`

Before this PR it fails with:
`error[E0433]: failed to resolve: use of unresolved module or unlinked crate anyhow`

After this PR that command passes, and `cargo test -p virtual-fs --no-default-features --features host-fs test_offload_file -- --nocapture` still passes too.

### DIFF
--- a/lib/virtual-fs/src/mem_fs/offloaded_file.rs
+++ b/lib/virtual-fs/src/mem_fs/offloaded_file.rs
@@ -394,10 +394,11 @@ impl OffloadedFile {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::io;
 
     #[test]
     #[tracing_test::traced_test]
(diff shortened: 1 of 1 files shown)


marxin (approved):
LGTM

PR #6698, opened 2026-06-10
feat(wasix): build wasm_tests Rust fixtures with cargo-wasix

Arshia001:
Replace direct rustc invocation with cargo wasix build, using ephemeral Cargo.toml files for single-file fixtures and supporting explicit Cargo projects with ## directives and BuildBin selection.

Since the WASIX Rust toolchain is not published for every platform CI tests on, the Rust fixtures are now prebuilt once on linux-x64 (`make build-wasm-tests-fixtures`, exporting via `WASM_TESTS_BUILD_ONLY_DIR`) and shared with all test jobs as an artifact consumed through `WASM_TESTS_PREBUILT_DIR`; test hosts no longer need `cargo-wasix`. Singlepass was also dropped from the wasm_tests harness entirely, as the WASIX toolchain emits EH opcodes it doesn't support.


lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -635,16 +757,12 @@ fn run_build_script(config: &Config) -> anyhow::Result<PathBuf> {
...
-            let mut cmd = rustc_command(source.contains("#![feature(").then_some("nightly"));
-            cmd.arg("--target=wasm32-wasip1")
-                .arg("-o")
-                .arg("main")
-                .arg(&primary_source)
-                .current_dir(&build_test_path);
-            cmd
+            write_ephemeral_cargo_toml(&build_test_path, filename)?;
+            cargo_wasix_build_command(&build_test_path)
+        }
```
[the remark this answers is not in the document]
Arshia001: cargo-wasix toolchain is always nightly anyway.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -21,6 +21,9 @@
 //!
 //! `BuildEnv:{key}={value}` sets an environment variable before building.
 //!
+//! `BuildBin:{name}` selects which Cargo binary to copy to `main` after a
```
marxin: What about simplifying the workflow and just expecting a single binary in a `Cargo.toml` file?

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -575,18 +574,146 @@ fn read_fixture_bytes(test_src_dir: &Path, arg: &str, directive: &str) -> Result
...
 
-fn rustc_command(toolchain: Option<&str>) -> Command {
-    if let Some(toolchain) = toolchain {
-        // rustc +version multiplexing is unsupported on Windows, use the documented approach:
-        // https://rust-lang.github.io/rustup/concepts/toolchains.html#custom-toolchains
-        let mut cmd = Command::new("rustup");
-        cmd.arg("run").arg(toolchain).arg("rustc");
-        cmd
+fn parse_cargo_toml_directive_line(line: &str) -> Option<&str> {
+    let line = line.trim();
```
marxin: `line.trim().strip_prefix(..).trim().split_once(':').map(...)` ?

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -575,18 +574,146 @@ fn read_fixture_bytes(test_src_dir: &Path, arg: &str, directive: &str) -> Result
...
+        .env("CARGO_TARGET_DIR", build_dir.join("target"));
+    cmd
+}
+
+fn write_ephemeral_cargo_toml(build_dir: &Path, source_filename: &str) -> Result<()> {
+    let manifest = format!(
+        r#"[package]
+name = "main"
+version = "0.0.0"
+edition = "2021"
```
marxin, line 782: `2024`?

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -575,18 +574,146 @@ fn read_fixture_bytes(test_src_dir: &Path, arg: &str, directive: &str) -> Result
...
+path = "{source_filename}"
+
+[workspace]
+"#
+    );
+    fs::write(build_dir.join("Cargo.toml"), manifest)
+        .with_context(|| format!("failed to write {}", build_dir.join("Cargo.toml").display()))
+}
+
+fn ensure_standalone_cargo_workspace(build_dir: &Path) -> Result<()> {
```
marxin: I would not introduce the function but let `cargo wasix build` just fail - I think we'll get to a reasonable error message in that case.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -575,18 +574,146 @@ fn read_fixture_bytes(test_src_dir: &Path, arg: &str, directive: &str) -> Result
...
+    let contents = fs::read_to_string(&manifest_path)
+        .with_context(|| format!("failed to read {}", manifest_path.display()))?;
+    if contents.contains("[workspace]") {
+        return Ok(());
+    }
+    fs::write(&manifest_path, format!("{contents}\n[workspace]\n"))
+        .with_context(|| format!("failed to update {}", manifest_path.display()))
+}
+
+fn cargo_bin_name_from_manifest(manifest_path: &Path, build_dir: &Path) -> Result<String> {
```
marxin: As mentioned, I would prefer a simpler approach with exactly one binary being defined in a `Cargo.toml` file.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -575,18 +574,146 @@ fn read_fixture_bytes(test_src_dir: &Path, arg: &str, directive: &str) -> Result
...
+        "could not determine Cargo binary name for {}; add src/main.rs or a ##BuildBin: name directive",
+        manifest_path.display()
+    )
+}
+
+fn copy_cargo_wasix_artifact(build_dir: &Path, bin_name: &str) -> Result<PathBuf> {
+    let wasm = build_dir
+        .join(CARGO_WASIX_ARTIFACT_DIR)
+        .join(format!("{bin_name}.wasm"));
+    ensure!(
```
marxin: Probably unnecessary as the consequent `fs::copy` will fail.

marxin (changes requested):

marxin (approved):
Much better, thank you for the changes!

lib/wasix/tests/wasm_tests/runner.rs
```diff
@@ -266,8 +266,6 @@ fn create_engine_for_wasm(wasm_bytes: &[u8], engine: Engine) -> wasmer::Engine {
         Engine::Cranelift => wasmer::BackendKind::Cranelift,
         #[cfg(feature = "llvm")]
         Engine::LLVM => wasmer::BackendKind::LLVM,
-        #[cfg(feature = "singlepass")]
```
marxin, line 271: This change is undesirable, as it would effectively eliminate Singlepass coverage for all `WASI-FYI` tests. We should preserve the current approach, where some Rust tests are built using the `wasm32-wasip1` target rather than `cargo wasix`.

marxin (changes requested):

lib/wasix/tests/wasm_tests/runner.rs
Arshia001: Singlepass and WASIX were never meant to be mixed. Do we gain a lot of value from this?

marxin: Please, preserve it, it's the only non-SPEC test coverage for the Singlepass compiler for anything non-trivial.
I think having a directive that will allow `.rs` -> WASM compilation based on the direct current `rustc` invocation should not make it so much complicated.

Arshia001: Very well, I'm putting wasip1 back in, but only for certain tests. Some rust tests (only those in a `*.rs` file, if not excluded via annotations) will get a wasip1+singlepass run on top of the usual wasix runs against other engines.

marxin: Works for me! Thanks.

marxin (changes requested):
To be honest, I am not a fan of the pre-built binary approach for 2 primary reasons:

1. Given our `cargo-wasix` test suite has been effectively just created, I would rather test a subset of the platforms natively end-to-end, rather than testing a half of the story . Moreover, if we take cargo-wasix seriously, we should include the platform in a short term.
2. Complexity increase for the test-harness and the S3 bucket sharing among the CI jobs.

Arshia001:
@marxin I disagree with your first point. The end goal here is to test the Wasmer codebase E2E as much as possible, not testing the entire world. While I agree cargo-wasix and the rust toolchain could use the extra platform support, that is not something we should concern ourselves with on this specific repository, and cargo-wasix does have its own testing happening elsewhere.

On the second point, while I don't like the complexity either, it's the only _good_ way we have to run the tests on every platform without spending considerable time adding builds for the rust toolchain on multiple new platforms. All in all, I think the current approach is the best trade-off, at least in the short term.

marxin:
> The end goal here is to test the Wasmer codebase E2E as much as possible, not testing the entire world.

Yes, agree - that's something currently fully achievable by doing a C, C++ or a Rust `wasm32-wasip1` test-cases.

> While I agree cargo-wasix and the rust toolchain could use the extra platform support, that is not something we should concern ourselves with on this specific repository, and cargo-wasix does have its own testing happening elsewhere.

If the goal of this effort is to improve test coverage for `cargo-wasix` - which I think is a worthwhile initiative - then running the tests against a target that `cargo-wasix` does not support does not make much sense to me.
Also, how many tests are we considering, and which specific features are we interested in covering with the WASM tests?

Arshia001:
The goal of this effort is testing that _modules produced by cargo-wasix_ behave correctly, so it's really a test of the rust stdlib code + rust-specific syscalls, should there be any (I don't think there are today), at the same time making sure the expectations are upheld on the native side. It is not, however, a test of the _cargo-wasix toolchain compiling code for WASIX_. While both sets of code are produced from the rust toolchain build, we're only interested in code that builds to WASM.

Do note that `wasip1` tests will not give us any coverage for rust stdlib+WASIX.

It's also worth noting that putting stdlib tests (rust, but also wasix-libc) in the Wasmer repo is a very conscious, deliberate choice, with the alternative being a full build of Wasmer in the rust/wasix-libc CI pipelines like we had to do for EdgeJS. This is something we decided against.

As for how many tests we're cosidering, the answer is "as many as we can reasonably fit in", with immediate priority going to regression tests for any fixes we push to the rust stdlib code. We don't do nearly as many fixes there as we do in wasix-libc, mainly owing to the fact that the rust stdlib routes almost all of its logic through wasix-libc anyway so fixes are propagated automatically. However, I do wish to add more rust tests as soon as I get the chance to.

marxin:
I see. Still, I’m not convinced the added complexity is worth the benefit, given that the only truly interesting target we’ll test in CI is `linux-aarch64` - setting aside how similar the `linux-musl` and `linux-x86_64` targets are.

marxin:
There are conflicts related to us dropping Windows support for the Compilers - please rebase.

Arshia001:
@marxin at this point, I wonder if this will _ever_ get merged in XD

marxin:
It will, the surface got much smaller, can review it in the afternoon ;)

.github/workflows/test.yaml
```diff
@@ -696,8 +714,10 @@ jobs:
         with:
           toolchain: ${{ steps.load_toolchain.outputs.rust_toolchain }}
           target: ${{ matrix.metadata.target }}
+      # For the wasip1-toolchain (Singlepass) variants of the Rust wasm_tests
+      # fixtures.
       - name: Install Rust WASI targets
-        if: matrix.stage.make == 'test-all'
+        if: matrix.stage.make == 'test-all' && matrix.metadata.wasix_rust_toolchain
```
marxin: Does not seem to me correct, `wasm32-wasip1` as `wasm32-wasip1` is available for all platforms we support.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -45,6 +45,11 @@
 //! `SkipEngine:{engine}:{reason}` marks the configuration as ignored for
 //! a given engine (LLVM, Cranelift, V8, Singlepass).
 //!
+//! `Toolchains:{list}` selects which toolchains build a single-file Rust
+//! fixture: `wasix` (cargo-wasix; runs on every engine except Singlepass) and
+//! `wasip1` (rustc with the wasm32-wasip1 target; runs on Singlepass only).
```
marxin: I would like to keep the current logic where the `wasm32-wasip1` fixtures are run for all Engines.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -100,6 +105,22 @@ mod runner;
 
 const TESTED_LIBC_VERSIONS: &[Option<&str>] = &[None, Some("v2026-05-12.1")];
 
+/// Whether Rust fixtures are collected on this host.
+///
+/// Building them needs the WASIX Rust toolchain, which wasix-org/rust only
```
marxin: Can we shorted the comment by saying we don't provide the `wasix` toolchain for all platforms yet.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -100,6 +105,22 @@ mod runner;
...
+/// publishes for a few hosts; musl hosts in particular have no toolchain at
+/// all. Windows is excluded on purpose even though a toolchain exists for it:
+/// wasixcc does not cover Windows either, so the suite as a whole does not run
+/// there.
+const COLLECT_RUST_FIXTURES: bool = cfg!(any(
```
marxin: Shoudn't it be specific to WASIX fixtures (because the `wasip1` are always supported). If I see correctly the cfg! will succeed for all the platforms we support (with exception of `musl`) - or?

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -151,6 +172,37 @@ pub enum Engine {
     V8,
 }
 
+/// Which toolchain builds a Rust fixture. The WASIX toolchain emits
+/// exception-handling opcodes, so its output cannot run on Singlepass; the
+/// plain `wasm32-wasip1` rustup target can, and is available on every host
+/// platform.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, strum::Display, strum::EnumString)]
+#[strum(ascii_case_insensitive, serialize_all = "lowercase")]
+enum Toolchain {
```
marxin: `RustToolchain` ?

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -151,6 +172,37 @@ pub enum Engine {
...
+    Wasix,
+    Wasip1,
+}
+
+impl Toolchain {
+    /// Engines that can run this toolchain's output. Singlepass lacks
+    /// exception-handling support, so it only runs the wasip1-built variant;
+    /// conversely the other engines only run the WASIX-built variant since
+    /// running both toolchains' output there would duplicate coverage.
+    fn supports_engine(self, engine: Engine) -> bool {
```
marxin, line 201: As mentioned it will change the current behavior for `wasip1` which can run with any Runtime.

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -187,6 +239,8 @@ struct Config {
     test_name: String,
     config_name: String,
     engine: Engine,
+    toolchain: Toolchain,
```
marxin: `rust_toolchain` and `rust_toolchains`?

lib/wasix/tests/wasm_tests/mod.rs
```diff
@@ -1300,57 +1510,51 @@ fn collect_tests(tests: &mut Vec<Trial>) -> Result<()> {
...
-                            let test_name = entry
-                                .path()
-                                .file_name()
-                                .expect("must be valid filename")
-                                .to_string_lossy()
-                                .to_string();
-                            if *engine == Engine::Singlepass
-                                && !["wasi_fyi", "wasi_wast"].contains(&test_name.as_str())
-                            {
+                    for toolchain in config.toolchains.as_ref().unwrap_or(&default_toolchains) {
```
marxin: This iteration seems weird as it's only a `rust_toolchain` and makes sense only in context of `PrimarySource::RustSourceFile/CargoProject`. What about iterating that in the inner-most loop for configs that are of such type?

Makefile
```diff
@@ -950,6 +950,9 @@ lint-package-crate:
 lint-formatting:
 	cargo fmt --all -- --check
 	cargo fmt --manifest-path fuzz/Cargo.toml -- --check
+	# The wasm_tests Rust fixtures are not part of any crate, so `cargo fmt`
+	# does not cover them.
+	cd lib/wasix/tests/wasm_tests && find . -path ./build -prune -o -type f -name '*.rs' -exec rustfmt --edition 2024 --check {} +
```
marxin, line 955: Good catch!

marxin (changes requested):

marxin (approved):

PR #6701, opened 2026-06-11
ci: pin wasix-libc sysroot and verify it stays current

Arshia001:
Use a shared CI constant for the wasix-libc sysroot tag so tests and copilot setup install a known prerelease, and fail CI if the pin falls behind the latest wasix-libc release.


scripts/check-wasix-libc-sysroot-tag.sh
```diff
@@ -0,0 +1,38 @@
...
+  auth_args=(-H "Authorization: Bearer $GITHUB_TOKEN")
+fi
+
+latest="$(
+  curl -fsSL "${auth_args[@]}" \
+    -H "Accept: application/vnd.github+json" \
+    "https://api.github.com/repos/wasix-org/wasix-libc/releases/latest" \
+    | sed -n 's/.*"tag_name": "\([^"]*\)".*/\1/p' \
+    | head -1
+)"
```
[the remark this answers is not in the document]
Arshia001: stupid bot, checking final releases is the whole point

scripts/check-wasix-libc-sysroot-tag.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
marxin: Frankly speaking I don't like extra new script that is hard to read what's actually doing.
What about something easier like:
```
gh release list --exclude-pre-releases --order desc -R wasix-org/wasix-libc | head -n1 | grep v2026-06-03.1
v2026-06-03.1   Latest  v2026-06-03.1   2026-06-03T09:53:40Z
```

marxin (changes requested):
I’m fine with the concept, but I’m not a fan of the newly added bash script.

I also think we should somehow incorporate the version expectations into the `wasm_tests` harness, so we can deliberately test against both the latest pre-release and the latest stable release, if desired.

marxin:
And I think the following 2 tests will need `MinimalLibc` directive be set:
```
FAIL [   0.523s] wasmer-wasix::wasm_tests wasm/proc_exec4_newline_arg/default/v2026-05-12.1/cranelift
FAIL [   0.512s] wasmer-wasix::wasm_tests wasm/proc_exec4_newline_env/default/v2026-05-12.1/cranelift
```

scripts/check-wasix-libc-sysroot-tag.sh
Arshia001: that doesn't give us the information we seek. What we want is to know if the latest release is equal to or older than the tagged version we're using for CI.

Arshia001: I inlined it into the CI pipeline, should be self-explanatory now.

marxin (approved):
LGTM

scripts/check-wasix-libc-sysroot-tag.sh
marxin: I like the selected approach!

.github/workflows/test.yaml
```diff
@@ -609,6 +621,13 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: true
+      - name: Load CI constants (musl)
```
marxin, line 624: This seems to me unnecessary.

marxin (changes requested):
Noticed one duplicate.

.github/workflows/test.yaml
Arshia001: it's avoiding using `bash` in musl environments, where it's initially unavailable.

marxin: Gotcha.

marxin (approved):

PR #6705, opened 2026-06-11
fix(wasix): Child exit code race on cleanup

artemyarulin:
Wasix could sometimes report a spawned child as exiting with 0 even when the child actually exited nonzero. The root cause was a cleanup race: `Sigquit` was broadcast before the real process exit code was recorded, and because task status is effectively first-write-wins, waiters could observe the synthetic success status instead of the real child result.

The fix is to record the real exit code first and only then broadcast `Sigquit`. This keeps the actual child status authoritative and prevents `waitpid() / join` callers from seeing a false successful exit.

That resolves our ongoing issue with `wp core is-installed` returning 0 exit code sometimes, even when DB doesn't exist

### DIFF
--- a/lib/wasix/src/state/env.rs
+++ b/lib/wasix/src/state/env.rs
@@ -1305,11 +1305,11 @@ impl WasiEnv {
                         }
                     }
 
-                    // Now send a signal that the thread is terminated
-                    process.signal_process(Signal::Sigquit);
-
-                    // Terminate the process
+                    // Record the real exit code before broadcasting Sigquit.
+                    // Otherwise a pending Sigquit can win the status race and
+                    // make waiters observe a successful exit.
                     process.terminate(process_exit_code);
+                    process.signal_process(Signal::Sigquit);
                 }
             })
         } else {
--- a/lib/wasix/tests/wasm_tests/process_tests/spawn-exec-nonzero-exit-loop/build.sh
+++ b/lib/wasix/tests/wasm_tests/process_tests/spawn-exec-nonzero-exit-loop/build.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+##BuildEnv: WASIXCC_WASM_EXCEPTIONS=no
+##CurrentDirectory: /home
+##MappedDirectory: .:/home
+##ExpectedStdout: spawn-exec-nonzero-exit-loop passed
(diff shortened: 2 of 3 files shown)


artemyarulin:
Ah, CI is failing because test is running bunch of subprocesses and we still have a pending leak for that https://github.com/wasmerio/wasmer/issues/6637

marxin:
You'll need a similar trick to:
https://github.com/wasmerio/wasmer/blob/ca0ec1c88703dfb77600847cff42663e1b6b8228/lib/wasix/tests/wasm_tests/dynamic_library_tests/dl-spawn-tls-init-under-contention/build.sh#L13-L14

the current V8 memory consumption is a known limitation.

Arshia001 (approved):
Grrat find!!

artemyarulin:
Let me run a shield, just wondering if there was a test already in Python for that

artemyarulin:
# Shield - Improvement 🎉🎉🎉

| Language | Tests  | Pass rate now | PASS | FAIL/TIMEOUT | SKIP | CRASH |
| -------- | ------ | ------------- | ---- | ------------ | ---- | ----- |
| Python | 37,920 | 73.1% | $${\color{green}+1}$$ | $${\color{green}-48}$$ | +47 | 0 |
| Rust | 22,337 | 94.3% | 0 | 0 | 0 | 0 |

| Language | Status transitions |
| -------- | ------------------ |
| Python | FAIL->PASS:1, TIMEOUT->FAIL:10, TIMEOUT->SKIP:47 |

- Examples from [tests_python_results.json](https://github.com/wasmerio/compat-tests/blob/c5288ba721e7f755f9ab1c490a79cfb1ee2e66bd/tests_python_results.json):
  - `test.test_socket.TCPCloserTest.testClose` (`FAIL -> PASS`)


## Artifacts

- GitHub Action: [https://github.com/wasmerio/anti-regression-shield/actions/runs/27363680440](https://github.com/wasmerio/anti-regression-shield/actions/runs/27363680440)
- Results commit: [https://github.com/wasmerio/compat-tests/commit/c5288ba721e7f755f9ab1c490a79cfb1ee2e66bd](https://github.com/wasmerio/compat-tests/commit/c5288ba721e7f755f9ab1c490a79cfb1ee2e66bd)

PR #6709, opened 2026-06-12
docs: fix stale build instructions for Cranelift and LLVM

immanuwell:
Docs are a bit stale here.

`docs/BUILD.md` says to toggle Cranelift with `ENABLE_SINGLEPASS`, but that flag actually toggles Singlepass.
`docs/CONTRIBUTING.md` also still says LLVM 14, while the repo now expects LLVM 22.

Repro:
1. Run `make -n build-wasmer ENABLE_SINGLEPASS=0`
2. It still prints `Enabled Compilers: cranelift`
3. Run `make -n build-wasmer ENABLE_CRANELIFT=0`
4. Cranelift is gone

Fix:
- use `ENABLE_CRANELIFT` in the Cranelift section
- point CONTRIBUTING to the current LLVM 22 build docs

### DIFF
--- a/docs/BUILD.md
+++ b/docs/BUILD.md
@@ -87,8 +87,8 @@ make build-wasmer
 
 **Note**: you should see `cranelift` appear in the `Enabled Compilers: ...` message in the console.
 
-You may disable the Cranelift backend with the `ENABLE_SINGLEPASS=0` environment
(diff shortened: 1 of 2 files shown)


marxin (approved):
Thanks for the contribution.

PR #6711, opened 2026-06-13
sanitize volume and atom names in webc_to_package_dir

netliomax25-code:
# Description

1. `webc_to_package_dir` builds the on-disk destination for each volume as `target_dir.join(mapping.volume_name.trim_start_matches('/'))`, and for each atom as `module_dir.join(atom_name)`. Both names come straight from the (untrusted) webc, and `trim_start_matches('/')` only removes a leading slash, not `..`.
2. A package whose filesystem annotation names a volume `../escaped` then makes `volume.unpack` write under the parent of `target_dir`, so `wasmer package unpack` of a hostile package can drop files outside the chosen output directory.
3. Names are written verbatim (so unpack stays consistent with the generated wasmer.toml and remains round-trippable), but before joining we now check whether a volume or atom name would resolve outside the output directory. If it would and the new `--allow-escape` flag is not set, the conversion errors out with a hint to use the flag; with the flag set the original behavior is preserved.

Added a regression test that unpacks a webc whose volume is named `../escaped` and checks that the default rejects it (writing nothing to a sibling directory) while `allow_escape` lets it through.

### DIFF
--- a/lib/cli/src/commands/package/download.rs
+++ b/lib/cli/src/commands/package/download.rs
@@ -305,6 +305,7 @@ impl PackageDownload {
             let unpack_cmd = super::unpack::PackageUnpack {
                 out_dir,
                 overwrite: false,
+                allow_escape: false,
                 quiet: self.quiet,
                 package_path: out_path,
                 format: super::unpack::Format::Package,
--- a/lib/cli/src/commands/package/unpack.rs
+++ b/lib/cli/src/commands/package/unpack.rs
@@ -18,6 +18,14 @@ pub struct PackageUnpack {
     #[clap(long)]
     pub overwrite: bool,
 
+    /// Allow volume/atom names from the package to be written outside the
+    /// output directory (e.g. names containing "..").
+    ///
+    /// Only applies to the `package` format. Off by default, in which case an
+    /// escaping name is an error.
+    #[clap(long)]
+    pub allow_escape: bool,
+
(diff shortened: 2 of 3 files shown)


theduke (changes requested):
This isn't explicitly documented  , but we would like `unpack` to be "idempotent", meaning you can run `unpack` and `build` and get the same package back.

This is a valid concern though.

We could add a `--allow-escape` flag or similar.
If not provided , escaping the main package directory would lead to an error, with a hint to use the flag.

netliomax25-code:
Good point on idempotency. Reworked it along the lines you suggested:

1. Issue: the silent re-rooting I had broke round-tripping, since the on-disk path no longer matched the name kept in wasmer.toml.
2. Change: names are now written verbatim again, but before joining I check whether a volume or atom name would resolve outside the output directory (a leading `..` etc.). If it would and the new `--allow-escape` flag isn't set, the conversion errors out with a hint to pass the flag; with the flag the original escaping behavior is preserved.
3. Validation: updated the regression test to assert both paths (default rejects and writes nothing to a sibling dir; `allow_escape` lets it through). `webc_to_package_dir` now takes an `allow_escape: bool` and `wasmer package unpack` exposes `--allow-escape`. Existing roundtrip test still passes.

This way nothing is rewritten, so unpack/build stays consistent.

netliomax25-code:
any update?

PR #6712, opened 2026-06-14
feat(Cranelift): add support for EH on Mach-O

marxin:
The change is split into 2 parts - in the first one the code is ported to the upcoming `0.133` release and the second one actually implements the EH support.

Note the PR needs to wait for the next major version release of Cranelift, but it should be out pretty soon.

Resolves: #6419


Cargo.toml
```diff
@@ -134,9 +134,9 @@ console = "0.16.1"
 console-subscriber = "0.5.0"
 cooked-waker = "5"
 corosensei = "0.3.0"
-cranelift-codegen = { version = "=0.132.0", default-features = false }
-cranelift-entity = { version = "=0.132.0", default-features = false }
-cranelift-frontend = { version = "=0.132.0", default-features = false }
+cranelift-codegen = { git = "https://github.com/bytecodealliance/wasmtime.git", rev = "5d09d889b28224cd15b3cf501efc84c41af56e2e", default-features = false }
```
marxin: Redirect before merging the PR.

syrusakbary (approved):

lib/compiler-cranelift/src/config.rs
```diff
@@ -315,11 +321,9 @@ impl CompilerConfig for Cranelift {
         self.middlewares.push(middleware);
     }
 
-    fn supported_features_for_target(&self, target: &Target) -> wasmer_types::Features {
+    fn supported_features_for_target(&self, _target: &Target) -> wasmer_types::Features {
         let mut feats = Features::default();
-        if target.triple().operating_system == OperatingSystem::Linux {
-            feats.exceptions(true);
-        }
+        feats.exceptions(true);
```
syrusakbary, line 333: Note: we may want to set this to false for Windows?

marxin: Yes, correct.

Cargo.toml
marxin: Fixed, using latest official release.

Issue #6592, opened 2026-05-13
cli: Add 'package tree' command


theduke:
Add a CLI command to show the dependency tree of a package.

PR #6713, opened 2026-06-14
feat: Feature/add package tree command

varunguleriaCodes:
Adds #6592, added a package tree command which prints all the dependencies, sub-dependencies of a package
# Description

image


marxin:
Dup of already implemented feature: #6617.

PR #6715, opened 2026-06-16
feat(cli,sdk,backend-api): add package search

kilyanni:
# Description

Adds a `wasmer package search` command and wires up the respective backend/sdk.

It supports all filters offered by the backend; though at the moment the backend does not respect these (it doesn't error, just silently ignores them). Working on a separate PR to address that atm


lib/cli/src/commands/package/search.rs
```diff
@@ -0,0 +1,234 @@
...
+
+        let filter = PackagesFilter {
+            owner: self.owner,
+            published_by: self.published_by,
+            curated: self.curated,
+            deployable: self.deployable,
+            has_bindings: self.has_bindings.then_some(true),
+            has_commands: self.has_commands.then_some(true),
+            is_standalone: self.standalone.then_some(true),
+            with_interfaces,
```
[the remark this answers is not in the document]
kilyanni: This property is upheld by the backend, not the caller

Arshia001 (approved):
LGTM

lib/cli/src/types.rs
```diff
@@ -1,10 +1,48 @@
...
+/// Render the full name (`namespace/name`) of a package version's package.
+fn package_full_name(pv: &SearchPackageVersion) -> String {
+    match &pv.package.namespace {
+        Some(ns) => format!("{ns}/{}", pv.package.package_name),
+        None => pv.package.package_name.clone(),
+    }
+}
+
+impl CliRender for SearchPackageVersion {
+    fn render_item_table(&self) -> String {
```
marxin, line 18: We should not render a table if there are not results:
```
     Running `target/debug/wasmer package search foobar`
+---------+---------+---------+
| Package | Version | Created |
+=============================+
+---------+---------+---------+
```

lib/cli/src/commands/package/search.rs
```diff
@@ -0,0 +1,234 @@
...
+    #[clap(long, value_enum, default_value = "desc")]
+    sort: Sort,
+
+    /// Maximum number of results to display.
+    #[clap(long, default_value = "50")]
+    max: usize,
+
+    /// The search query. Leave empty to list all (matching) packages.
+    #[clap(default_value = "")]
+    query: String,
```
marxin, line 196: If the string is omitted and no other filters are used, the query can be probably quite heavy for the back-end right? Will correct server-side handling of `limit` cover it?

marxin (changes requested):
Could you please include some Wasmer CLI integration tests, similar to the ones in the other PR?

Regarding the server-side filters, do you have an estimate for when they’ll be covered? The tests should cover the filtering capability as well.

kilyanni:
> Regarding the server-side filters, do you have an estimate for when they’ll be covered?

I already have a draft open (https://github.com/wasmerio/backend/pull/2814/), but I'm blocked on actually testing/verifying it until I get access to edge

lib/cli/src/commands/package/search.rs
kilyanni: The backend limits requests to 1000 elements properly already. Running it with `--max 9999` still only yields 1000 results, so this should be fine I think

marxin: Sounds good.

lib/cli/src/types.rs
kilyanni: Now returns `No packages found.` (iff the format is set to a human-readable one)

marxin: Works for me, thanks.

lib/backend-api/src/query.rs
```diff
@@ -1742,6 +1742,70 @@ pub async fn get_package_versions(
...
+    query: impl Into<String>,
+    filter: Option<types::PackagesFilter>,
+    page_size: i32,
+) -> impl futures::Stream<Item = Result<Vec<types::SearchPackageVersion>, anyhow::Error>> + '_ {
+    let query = query.into();
+    futures::stream::try_unfold(Some(None), move |state| {
+        let query = query.clone();
+        let filter = filter.clone();
+        async move {
+            let after = match state {
```
marxin: Maybe `let Some(after) = state else { return None; };` ?

tests/integration/cli/tests/package/search.rs
```diff
@@ -0,0 +1,81 @@
+use insta::assert_json_snapshot;
+use tempfile::TempDir;
+use wasmer_integration_tests_cli::wasmer_command;
+
+fn package_search(args: &[&str]) -> std::process::Output {
+    let wasmer_dir = TempDir::new().unwrap();
+
+    wasmer_command()
+        .env("WASMER_DIR", wasmer_dir.path())
```
marxin: For testing the package search, I bet you won't need the 2 env variables, or?

tests/integration/cli/tests/package/search.rs
```diff
@@ -0,0 +1,81 @@
...
+    let packages: Vec<serde_json::Value> = serde_json::from_slice(&output.stdout).unwrap();
+    let package = packages
+        .iter()
+        .find(|value| {
+            value["package"]["namespace"] == "rustpython"
+                && value["package"]["package_name"] == "rustpython"
+        })
+        .expect("expected the backend to return rustpython/rustpython");
+
+    let normalized = serde_json::json!({
```
marxin: Do we really need the JSON repacking?

tests/integration/cli/tests/package/search.rs
```diff
@@ -0,0 +1,81 @@
...
+        "1",
+        "--curated",
+        "--order-by",
+        "alphabetically",
+        "--sort",
+        "asc",
+        "rustpython",
+    ]);
+
+    assert!(
```
marxin, line 30: Maybe we can also include the stdout/stderr to the JSON snapshot, similarly to what other snapshots look like?

marxin (approved):
Am generally fine with a couple of small nits. But, I would like to see it merged after the corresponding back-end change is in place. Having that, we can also test the filtering capabilities a bit in the unit tests.

tests/integration/cli/tests/package/search.rs
kilyanni: Copied them from another test, but yeah, seems they're not needed here

tests/integration/cli/tests/package/search.rs
kilyanni: It's not strictly required, but if a new version of `rustpython/rustpython` gets pushed, all the non-normalized values would not be stable I believe.

tests/integration/cli/tests/package/search.rs
kilyanni: The test now asserts stderr to be empty; wrt. stdout that depends on if we want normalization

tests/integration/cli/tests/package/search.rs
marxin: I see, then I would recommend not using the `snap` and just use a collection of `assert(...is_string())` calls.

marxin (approved):

tests/integration/cli/tests/package/search.rs
kilyanni: Makes sense, switched it over

kilyanni:
CI is failing due to network issues (504s) afaict

PR #6716, opened 2026-06-16
 feat(cli): add package show 

### LINKED ISSUE #6723: Command to show basic metadata about a package
### Motivation

When working with wasmer packages, you sometimes want to find out some basic information about a certain package (license, latest version, description). Right now this requires either `package download` + `package unpack` + `cat foobar/wasmer.toml` or using the website, which is a bit cumbersome.

### Proposed solution

A simple CLI command `package show` that displays/pretty-prints basic information about the package

### Alternatives

n/a

### Additional context

n/a

kilyanni:
# Description

Adds a small command to show basic metadata for packages. Works with local paths as well as registry names.

Example:

```
$ wasmer package show wasmer/bash@=1.0.24
Name:         wasmer/bash
Version:      1.0.24
License:      GNU
Entrypoint:   bash
```

Closes #6723

### DIFF
--- a/lib/cli/src/commands/package/get.rs
+++ b/lib/cli/src/commands/package/get.rs
@@ -1,0 +1,260 @@
+use std::path::Path;
+
+use anyhow::{Context, bail};
+use dialoguer::console::style;
+use wasmer_config::package::{Manifest, PackageIdent, PackageSource};
+
+use super::common::{get_manifest, manifest_from_webc_metadata, package_web_url};
+use crate::commands::AsyncCliCommand;
+use crate::config::WasmerEnv;
+
+/// Show basic metadata of a package without unpacking or downloading it.
+///
+/// The package can be a directory containing a `wasmer.toml` file, a `.webc`
+/// file, or a package name from the registry. If no argument is given, the
+/// current directory is used.
+#[derive(clap::Parser, Debug)]
+pub struct PackageGet {
+    #[clap(flatten)]
+    pub env: WasmerEnv,
+
+    /// The package to show.
+    ///
+    /// This can be a path to a package directory or a `.webc` file, or the name
+    /// of a package in the registry (e.g. `wasmer/hello@=0.1.0`).
+    #[clap(default_value = ".")]
+    pub package: String,
+}
+
+#[async_trait::async_trait]
+impl AsyncCliCommand for PackageGet {
+    type Output = ();
+
+    async fn run_async(self) -> Result<Self::Output, anyhow::Error> {
+        let (manifest, web_url) = self.load_manifest().await?;
+
+        let print_field = |label: &str, value: &str| {
+            println!("{:<13} {value}", style(format!("{label}:")).bold().dim());
+        };
+
+        if let Some(package) = manifest.package.as_ref() {
+            print_field("Name", package.name.as_deref().unwrap_or("<unnamed>"));
+            print_field(
+                "Version",
+                &package
+                    .version
+                    .as_ref()
+                    .map(|v| v.to_string())
+                    .unwrap_or_else(|| "<none>".to_string()),
+            );
+
+            if let Some(description) = &package.description {
+                print_field("Description", description);
+            }
+            if let Some(license) = &package.license {
+                print_field("License", license);
+            }
+            if let Some(homepage) = &package.homepage {
+                print_field("Homepage", homepage);
+            }
+            if let Some(repository) = &package.repository {
+                print_field("Repository", repository);
+            }
+            if let Some(entrypoint) = &package.entrypoint {
+                print_field("Entrypoint", entrypoint);
+            }
+            if package.private {
+                print_field("Private", "true");
+            }
+        } else {
+            println!("{}", style("Package has no metadata.").dim());
+        }
+
+        if !manifest.commands.is_empty() {
+            let commands = manifest
+                .commands
+                .iter()
+                .map(|c| c.get_name())
+                .collect::<Vec<_>>()
+                .join(", ");
+            print_field("Commands", &commands);
+        }
+
+        if !manifest.dependencies.is_empty() {
+            print_field("Dependencies", &manifest.dependencies.len().to_string());
+            for (name, version) in &manifest.dependencies {
+                println!("  {name} = {version}");
+            }
+        }
+
+        if let Some(web_url) = web_url {
+            print_field("URL", &web_url);
+        }
+
+        Ok(())
+    }
+}
+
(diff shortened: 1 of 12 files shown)


theduke:
@kilyanni please create an issue for this as well.

lib/cli/src/commands/package/get.rs
```diff
@@ -0,0 +1,228 @@
+use std::path::Path;
+
```
theduke, line 2: Our other commands for this are called `get`, eg `app get`, this should adhere to that naming.

theduke (dismissed):
Not a full review, just commenting on the naming.

lib/cli/src/commands/package/get.rs
kilyanni: From an outsiders perspective that sounds like it'd be an alias for `download` or something along those lines. There's also `wasmer app info`, would `info` also work as a name?

theduke: That's a bit of an unfortunate historical artifact. `get` shows more data, `info` is human-readable, but that split wasn't done well, that should just have been a different format toggle.

kilyanni: In this case the output is formatted to be human-readable fwiw, so `info` fits better here IMO

Though in that case; should I maybe add both? `get` -> raw wasmer.toml, `info` -> pretty-print?

theduke: At the moment we use `get` everywhere. `namespace get`, `app version get` `app deployments get`, so we have to adhere to that convention.

lib/cli/src/commands/package/get.rs
```diff
@@ -0,0 +1,228 @@
...
+        } else {
+            println!("{}", style("Package has no metadata.").dim());
+        }
+
+        if !manifest.commands.is_empty() {
+            let commands = manifest
+                .commands
+                .iter()
+                .map(|c| c.get_name())
+                .collect::<Vec<_>>()
```
marxin, line 81: The workspace uses `itertools` crate and so you can use `.collect_vec` instead here.

lib/cli/src/commands/package/get.rs
```diff
@@ -0,0 +1,228 @@
...
+            print_field(
+                "Version",
+                &package
+                    .version
+                    .as_ref()
+                    .map(|v| v.to_string())
+                    .unwrap_or_else(|| "<none>".to_string()),
+            );
+
+            if let Some(description) = &package.description {
```
marxin, line 54: Tried running the new command with:
```
$ target/debug/wasmer package get wasmer/cli
Name:         wasmer/cli
Version:      0.1.3
Entrypoint:   bash
```

The **description** is missing, but I can see it on the web:
https://wasmer.io/wasmer/cli

> CLI platform - a wrapper package with many common tools, useful for interactive environments.

lib/cli/src/commands/package/get.rs
```diff
@@ -0,0 +1,228 @@
...
+            if let Some(repository) = &package.repository {
+                print_field("Repository", repository);
+            }
+            if let Some(entrypoint) = &package.entrypoint {
+                print_field("Entrypoint", entrypoint);
+            }
+            if package.private {
+                print_field("Private", "true");
+            }
+        } else {
```
marxin, line 72: Maybe I would also include web link to the package, e.g. https://wasmer.io/wasmer/cli

lib/cli/src/commands/package/get.rs
```diff
@@ -0,0 +1,228 @@
...
+                    String::from("latest")
+                } else {
+                    version
+                };
+                let full_name = id.full_name();
+
+                let package = wasmer_backend_api::query::get_package_version(
+                    &client,
+                    full_name.clone(),
+                    version.clone(),
```
marxin, line 139: Maybe (leaving up to you) we could provide a clearer message when version is specified but the package isn’t found: say that the requested version is missing, and list the available versions.

lib/cli/src/commands/package/get.rs
```diff
@@ -0,0 +1,228 @@
...
+            }
+            PackageSource::Url(url) => {
+                bail!("showing a package directly from a URL is not supported: '{url}'")
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
marxin, line 291: Please include also snapshot-based test that will tests against the back-end, take a look at `test_snapshot_dash_python`.

marxin (changes requested):
Noticed a couple of smaller findings - very good first contribution to the project 🎊

lib/cli/src/commands/package/get.rs
kilyanni: Investigating this right now and noticed something odd: when you do `package download` + `package unpack`, the `wasmer.toml` does not contain the description (even though the toml shown by the webui under settings does).

Is this intentional/the desired behavior?

lib/cli/src/commands/package/get.rs
kilyanni: Sounds like a good idea, will do :+1:

lib/cli/src/commands/package/get.rs
marxin: @theduke maybe knows more details about the behavior?

kilyanni: For now, I made it pull from the registry directly. Would still be good to figure out the webc behavior either way though

lib/cli/src/commands/package/get.rs
kilyanni:
```
$ cargo run -q -p wasmer-cli -- package get wasmer/bash@4.4.4
error: package 'wasmer/bash' has no version matching '4.4.4' in registry 'wasmer.io'.
Available versions: 1.0.20, 1.0.21, 1.0.24, 1.0.25
```

marxin: Like it!

lib/cli/src/commands/package/get.rs
marxin: Can we rely on `snapshots`, rather than on a series of `contains`?

kilyanni: Absolutely; changed it to use snapshots

marxin (approved):

marxin:
Conflicts need to be resolved.

kilyanni:
Ah right, forgot this was gonna have merge conflicts with the search command

Rebased :+1:

PR #6717, opened 2026-06-16
fix(wasix-tests): stop parsing exit codes from error strings

immanuwell:
# Description

Tiny WASIX test harness fix.

Right now `lib/wasix/tests/wasm_tests/runner.rs` pulls exit codes out of free form error text by looking for `ExitCode::...`.
If some real runner error contains that text, the harness can treat it like a normal process exit. Kinda cursed, but real.

This patch reads typed `WasiError` and `WasiRuntimeError` values from the error chain instead.
Normal exit handling stays the same. Random strings stop getting treated like exits.

This is not some impossible edge tied to quotas or hard limits. Any error message with `ExitCode::N` can hit it, so the boundary is very reachable.

Repro before this patch:
1. Make `run_wasm_with_runner_config` return `anyhow!("Spawn failed: ExitCode::1")`.
2. The harness reports exit code `1` instead of returning the real error.

Checks:
- `cargo test -p wasmer-wasix --test runner_exit_code -- --nocapture`
- `cargo test -p wasmer-wasix --test runners -- --nocapture`
- `cargo build -p wasmer-wasix --tests`

Related recent harness work: #6599, #6531

### DIFF
--- a/lib/wasix/tests/runner_exit_code.rs
+++ b/lib/wasix/tests/runner_exit_code.rs
@@ -1,0 +1,38 @@
+#[path = "wasm_tests/error.rs"]
+mod error;
+
+use anyhow::{Result, anyhow};
+use wasmer_wasix::{WasiError, WasiRuntimeError, wasmer_wasix_types::wasi::ExitCode};
+
+#[test]
+fn extracts_exit_code_from_wasi_runtime_error_chain() -> Result<()> {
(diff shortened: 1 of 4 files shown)


immanuwell:
@marxin PTAL

marxin (approved):
LGTM, sorry for a bit longer review.

PR #6718, opened 2026-06-16
Last socket error

sadhbh-c0d3:
- Added `last_error()` to `VirtualSocket` and its implementations.
- Added `last_error()` to `InodeSocket` delegating to impl of virtual socket
- The `sock_get_opt_size()` calls `InodeSocket::last_error()`, which first calls `LocalTcpStream::status()` that sets `connect_state` including new socket error in it, and then `LocalTcpStream::last_error()` returns that as an `Option<NetworkError>`, which is then mapped into `Errno`, converted into `u16`, and eventually returned to libc caller.

In the baseline runtime, the socket operation that actually fails and the later getsockopt(SO_ERROR) call can be disconnected from each other. A nonblocking connect can fail with ECONNREFUSED, but if Wasmer does not record that error on the socket object, the next SO_ERROR read can return success or a generic error.

The visible result is wrong POSIX error reporting. Code that expects ECONNREFUSED may instead see EPIPE, ECONNRESET, or no pending error at all. The fix will store the last socket error in Wasmer socket state and expose that state through the WASIX Sockoption::LastError path used by getsockopt(SO_ERROR).

[Detailed description](https://github.com/wasmerio/edgejs/blob/test-wasix-quickjs-only/plans/wasix-plan/wasmer/03_last_socket_error_so_error.md#last-socket-error--so_error)

### DIFF
--- a/.github/ci-constants.env
+++ b/.github/ci-constants.env
@@ -1,3 +1,3 @@
 # Shared CI constants. Loaded into GITHUB_ENV by workflows that need them.
 # Pinned wasix-libc sysroot (wasix-org/wasix-libc release tag).
-WASIX_LIBC_SYSROOT_TAG=v2026-06-09.1
+WASIX_LIBC_SYSROOT_TAG=v2026-06-18.1
--- a/lib/virtual-net/src/host.rs
+++ b/lib/virtual-net/src/host.rs
@@ -527,6 +527,7 @@ enum ConnectState {
     Unknown,
     Opened,
     Failed(NetworkError),
+    Reset,
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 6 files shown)


Arshia001 (changes requested):
The change looks good in principle, with a couple of caveats:

* To stay POSIX-compliant, once the last error is returned, it must be cleared; POSIX reports socket errors only once.
* #6727 should be merged into this PR and then closed. Adding a test in a separate PR doesn't make a lot of sense.

sadhbh-c0d3:
Has dependency on get last error in wasix libc

Arshia001 (approved):

PR #6719, opened 2026-06-16
Partial Success for fd_write on Socket

sadhbh-c0d3:
- Fix: When number of bytes actually written is non-zero, then `writev()` must return that number of bytes written.

`man writev` says:
```
       Upon successful completion, writev() shall return the number of
       bytes actually written. Otherwise, it shall return a value of -1,
       the file-pointer shall remain unchanged, and [errno](https://man7.org/linux/man-pages/man3/errno.3.html) shall be set to
       indicate an error.
```

[Detailed description](https://github.com/wasmerio/edgejs/blob/test-wasix-quickjs-only/plans/wasix-plan/wasmer/05_partial_success_for_fd_write.md)

### DIFF
--- a/.github/ci-constants.env
+++ b/.github/ci-constants.env
@@ -1,3 +1,3 @@
 # Shared CI constants. Loaded into GITHUB_ENV by workflows that need them.
 # Pinned wasix-libc sysroot (wasix-org/wasix-libc release tag).
-WASIX_LIBC_SYSROOT_TAG=v2026-06-09.1
+WASIX_LIBC_SYSROOT_TAG=v2026-06-18.1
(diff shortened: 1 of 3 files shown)


Arshia001 (changes requested):
Change LGTM and can be merged once the test is merged in.

sadhbh-c0d3:
The test in this PR depends also on https://github.com/wasmerio/wasmer/pull/6721

The partial writev() on File was already present, this fixes the partial write on Socket.

Didn't push it here, but we would also need to consider what happens when one of the iovecs is invalid, say second one, and first was written. There is two options here:

a) validate all iovecs before writing, and return error if any of them is invalid before any writes
b) validate on the go, but then if there was at least one write, we must return success

I have solution to option b) available, but wasn't sure if I should add it to this PR.

Arshia001:
Superseded by a refreshed PR with additional stream TCP writev regression tests and documentation.

Arshia001:
Superseded by #6742.

PR #6720, opened 2026-06-16
UDP Datagram Receive, Readiness, and Backlog

sadhbh-c0d3:
- Fixed implementation of `poll_read_ready()` and `try_recv_from()` so that bytes received into `backlog` are consumed correctly.
- Additionally peek is now supported, and peeks bytes in backlog first.
- Flow is bytes are received from underlying stream, and placed into backlog, and then peeked or drained from that backlog. 
 
This solution guarantees consistency and without this fix packets would be dropped.

[Detailed description](https://github.com/wasmerio/edgejs/blob/test-wasix-quickjs-only/plans/wasix-plan/wasmer/01_udp_datagram_receive_readiness_and_backlog.md#udp-datagram-receive-readiness-and-backlog)

### DIFF
--- a/lib/virtual-net/src/client.rs
+++ b/lib/virtual-net/src/client.rs
@@ -1042,15 +1042,19 @@ impl VirtualIoSource for RemoteSocket {
             Poll::Pending => {}
         }
         if !self.buffer_recv_with_addr.is_empty() {
-            let total = self
+            let len = self
                 .buffer_recv_with_addr
-                .iter()
+                .front()
                 .map(|a| a.data.len())
-                .sum();
-            return Poll::Ready(Ok(total));
+                .unwrap_or_default();
+            return Poll::Ready(Ok(len));
         }
         match self.rx_recv_with_addr.poll_recv(cx) {
-            Poll::Ready(Some(data)) => self.buffer_recv_with_addr.push_back(data),
+            Poll::Ready(Some(data)) => {
+                let len = data.data.len();
+                self.buffer_recv_with_addr.push_back(data);
+                return Poll::Ready(Ok(len));
+            }
             Poll::Ready(None) => return Poll::Ready(Ok(0)),
             Poll::Pending => {}
         }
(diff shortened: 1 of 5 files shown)


Arshia001 (changes requested):
Approved in principle, but two outstanding issues (which I'm copy-pasting from AI, so no formatting! sorry!)

## Issue 1

  Critical: zero-length datagrams still break poll_oneoff

  This PR’s design doc and minimal example are built around a zero-length UDP 
  datagram. After this PR, when such a datagram is in the backlog:

  let len = self.backlog.front().map(|a| a.0.len()).unwrap_or_default();
  return Poll::Ready(Ok(len)); // len == 0

  That flows into inode_guard.rs:

   lib/wasix/src/fs/inode_guard.rs lines 245-254

                  Poll::Ready(bytes_available) => {
                      // ...
                      Some(EventResultType::Fd(EventFdReadwrite {
                          nbytes: bytes_available as u64,
                          flags: if bytes_available == 0 {
                              Eventrwflags::FD_READWRITE_HANGUP
                          } else {
                              Eventrwflags::empty()
                          },
                      }))

  So a readable zero-length datagram is reported as FD_READWRITE_HANGUP with nbytes = 
  0, not as a normal read-ready event.

  POSIX/Linux behavior:

  • poll(POLLIN) is set for a queued zero-length datagram.
  • recvfrom() returns 0 (valid datagram), not EOF.
  • UDP has no hangup semantics equivalent to TCP EOF.

  This is not a minor detail — it is the exact scenario the PR claims to fix. The
  virtual-net layer now preserves the datagram, but the wasix poll layer still
  misclassifies readiness.

## Issue 2:

  POSIX: backlog receive buffer size caps datagram size

  recv_into_backlog() always receives into a 10,240-byte buffer. Linux behavior when a
   datagram exceeds the buffer:

  • recvfrom returns up to buffer size.
  • Remaining bytes of that datagram are discarded and are unrecoverable.

  That was true before this PR too, but the backlog model makes the loss permanent at
  the guest boundary instead of only at one syscall. For strict POSIX fidelity on
  large UDP payloads, the backlog should store up to the platform’s max datagram size
  (65507 for IPv4 UDP), or use MSG_TRUNC/MSG_PEEK/recvmsg to detect truncation.

Arshia001:
One more issue:

  Filtering only exists in recv() / sock_recv, not everywhere:

  recv() (connected UDP) — filters by stored peer:

   lib/wasix/src/net/socket.rs lines 1517-1521

                          InodeSocketKind::UdpSocket { socket, peer } => {
                              if let Some(peer) = peer {
                                  match socket.try_recv_from(self.data, peek) {
                                      Ok((amt, addr)) if addr == *peer => Ok(amt),
                                      Ok(_) => Err(NetworkError::WouldBlock),

  On main that wrong-peer case returns WouldBlock (bad). PR #6720 improves this to
  discard non-matching datagrams and keep looping — but only on this path.

  recv_from() / sock_recv_from — no peer check at all:

   lib/wasix/src/net/socket.rs lines 1614-1616

                          InodeSocketKind::UdpSocket { socket, .. } => {
                              socket.try_recv_from(self.data, peek)
                          }

  So a guest calling recvfrom() on a connected UDP fd can still see datagrams from
  arbitrary senders, with the real source address filled in. That is not Linux
  connected-UDP behavior.

  PR #6720’s recv_into_backlog() pulls the next datagram off the real socket with no
  peer filter. Wrong-peer packets can land in the backlog via:

  • poll_read_ready() prefetch
  • any unfiltered try_recv_from() path

  Then:

  • recv() may eventually discard them (with #6720)
  • recv_from() will happily deliver them
  • poll may show readable because some datagram exists, even if recv() would skip
    it

  So you can get poll/recv inconsistency too.

Arshia001:
Superseded by #6741 — review feedback addressed and additional tests added on `fix/udp-datagram-receive-readiness-backlog`.

Arshia001:
Closing in favor of #6741.

PR #6721, opened 2026-06-16
Connected UDP Send, Peer State, and EMSGSIZE

sadhbh-c0d3:
- Connected UDP must send single batch of iovecs as a single datagram and not multiple.
- MessageSize error will occur when total size of the data gathered from all iovecs buffers exceeds maximum.
- Original connected peer address is preserved across sends.

The `InodeSocket` stores `inner: InodeSocketKind`, and initially it's `InodeSocketKind::PreSocket`. When `sock_connect_internal()` is called, it uses `__sock_upgrade()` to potentially upgrade `WasiFd` with new socket.

The reason `WasiFd` needs to be rebound to new `InodeSocket` is because `InodeSocket::bind_internal()` creates new instance of `InodeSocket` instead of modifying `self.inner.protected`.

```rust
Ok(Some(InodeSocket::new(InodeSocketKind::UdpSocket {
                                socket,
                                peer: None,
                            })))
```

Then `InodeSocket::connect()` sets peer address:

```rust
               InodeSocketKind::UdpSocket {
                    peer: target_peer, ..
                } => {
                    target_peer.replace(peer);
                    return Ok(None);
                }
```

[Detailed information](https://github.com/wasmerio/edgejs/blob/test-wasix-quickjs-only/plans/wasix-plan/wasmer/02_connected_udp_send_peer_state_and_emsgsize.md#connected-udp-send-peer-state-and-emsgsize)

### DIFF
--- a/lib/virtual-net/src/lib.rs
+++ b/lib/virtual-net/src/lib.rs
@@ -953,7 +953,6 @@ pub fn net_error_into_io_err(net_error: NetworkError) -> std::io::Error {
         NetworkError::Interrupted => ErrorKind::Interrupted.into(),
         NetworkError::InvalidData => ErrorKind::InvalidData.into(),
         NetworkError::InvalidInput => ErrorKind::InvalidInput.into(),
-        NetworkError::MessageSize => ErrorKind::InvalidInput.into(),
         NetworkError::NotConnected => ErrorKind::NotConnected.into(),
         NetworkError::NoDevice => ErrorKind::BrokenPipe.into(),
         NetworkError::PermissionDenied => ErrorKind::PermissionDenied.into(),
@@ -964,6 +963,16 @@ pub fn net_error_into_io_err(net_error: NetworkError) -> std::io::Error {
         NetworkError::Unsupported => ErrorKind::Unsupported.into(),
         NetworkError::UnknownError => ErrorKind::BrokenPipe.into(),
         NetworkError::InsufficientMemory => ErrorKind::OutOfMemory.into(),
+        NetworkError::MessageSize => {
+            #[cfg(all(target_family = "unix", feature = "libc"))]
+            {
+                std::io::Error::from_raw_os_error(libc::EMSGSIZE)
+            }
+            #[cfg(not(all(target_family = "unix", feature = "libc")))]
+            {
+                ErrorKind::Other.into()
+            }
+        }
         NetworkError::TooManyOpenFiles => {
             #[cfg(all(target_family = "unix", feature = "libc"))]
             {
--- a/lib/virtual-net/src/lib.rs
+++ b/lib/virtual-net/src/lib.rs
@@ -849,6 +849,9 @@ pub enum NetworkError {
     /// The provided data is invalid
     #[error("invalid input")]
     InvalidInput,
+    /// The message is too large to send as one packet
+    #[error("message too large")]
+    MessageSize,
     /// Could not perform the operation because there was not an open connection
     #[error("connection is not open")]
     NotConnected,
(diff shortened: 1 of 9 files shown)


lib/virtual-net/src/lib.rs
```diff
@@ -949,6 +953,7 @@ pub fn net_error_into_io_err(net_error: NetworkError) -> std::io::Error {
         NetworkError::Interrupted => ErrorKind::Interrupted.into(),
         NetworkError::InvalidData => ErrorKind::InvalidData.into(),
         NetworkError::InvalidInput => ErrorKind::InvalidInput.into(),
+        NetworkError::MessageSize => ErrorKind::InvalidInput.into(),
```
Arshia001: We should thread this error through properly instead of just pretending it's invalid input.

Arshia001 (changes requested):
There are multiple issues that need to be resolved:

## Issue 1

  sock_send_to() still sends multiple datagrams for multiple iovecs (not fixed)

  The PR only patches sock_send.rs. sock_send_to.rs still has the per-iovec loop:

   lib/wasix/src/syscalls/wasix/sock_send_to.rs lines 124-127

                          let mut sent = 0usize;
                          for iovs in iovs_arr.iter() {
                              // ...
                              socket.send_to(..., buf.as_ref(), addr, ...).await

  POSIX sendto() with iovecs is also a single gathered datagram (sendmsg semantics).
  Same bug, different syscall. If test-dgram-msgsize or other tests use sendto with
  multiple iovecs, they can still fail or behave non-POSIX.

  Recommendation: Apply the same coalescing logic to sock_send_to_internal, or factor
  out a shared helper.

## Issue 2

  EMSGSIZE mapping may be brittle across platforms

  It depends on hitting the raw_os_error() path. That works on current Linux
  (Uncategorized + errno 90). On some platforms or Rust versions, oversize datagram
  errors may appear as ErrorKind::InvalidInput first — which is mapped to generic
  NetworkError::InvalidInput → Errno::Inval, not Msgsize.

  Pedantic hardening:

  ErrorKind::InvalidInput if net_error.raw_os_error() == Some(libc::EMSGSIZE) => 
  NetworkError::MessageSize,

  (or equivalent), before the generic InvalidInput arm.

  Also, net_error_into_io_err(MessageSize) => ErrorKind::InvalidInput loses errno
  identity on round-trip. Minor, but inconsistent with the goal of preserving
  EMSGSIZE.

## Issue 3

  addr_peer() still has odd fallback for unconnected UDP

  After the new peer: Some arm, unconnected UDP still falls through to:

   lib/wasix/src/net/socket.rs lines 904-907

              InodeSocketKind::UdpSocket { socket, .. } => socket
                  .addr_peer()
                  // ...
                  .unwrap_or_else(|| { /* 0.0.0.0:0 */ })

  POSIX getpeername() on an unconnected UDP socket should be ENOTCONN, not 0.0.0.0:0.
  Pre-existing, not introduced here, but worth noting since this PR touches the same
  function.

## Issue 4

  Minor: dgram coalescing allocates unconditionally

  Building a Vec and extend_from_slice for every multi-iovec datagram send is correct
  but allocates. Acceptable for correctness; could use size-hint preallocation
  (data.reserve(total_len)) after a first pass over iovecs. Not a semantics issue.

Note from me on this one: it's important that we don't allow a malicious guest to get its host to over-allocate and run out of memory, so there should also be a hard cap on the total length of the data, and a guest trying to send more than that in one go gets an error before we allocate anything. while the OS's network layer will prevent us from sending too much data in one packet, we'll still end up allocating a lot of space before making the actual send call.

sadhbh-c0d3:
Issues 1 & 4 resolved by implementing `coalesce()` method for `FdWriteSource`, and usng `Vec::with_capacity()` there. Additionally added `is_dgram() -> bool` to `InodeSocket`.

sadhbh-c0d3:
Issue 3 also addressed (zero AI ☺️) 

I return ENOTCONN  only if address is not set. Wasn't sure if should always just return ENOTCONN if not connected.

sadhbh-c0d3:
Issue 2 addressed also no AI was harmed 🙃

Arshia001:
superseded by #6738.

PR #6727, opened 2026-06-17
Added test for getting last socket error

sadhbh-c0d3:
Test that `getsockopt()` can return value of the `SOL_ERROR` option.


Arshia001 (changes requested):
This should go in #6718.

PR #6728, opened 2026-06-17
Added test for UDP readiness fix

sadhbh-c0d3:
Test proves that UDP readiness doesn't currently work.

It is fixed in https://github.com/wasmerio/wasmer/pull/6720, the branch on which this test passes.


sadhbh-c0d3:
merged into implementation

PR #6729, opened 2026-06-17
Added test for connected udp peer address fix

sadhbh-c0d3:
Test that connected UDP socket remembers its peer address.


sadhbh-c0d3:
merged into implementation

PR #6730, opened 2026-06-17
Added test that writev on connected UDP socket writes single datagram packet

sadhbh-c0d3:
Test that `writev()` gathers single UDP packet from multiple buffers provided.


sadhbh-c0d3:
It's a test for https://github.com/wasmerio/wasmer/pull/6719

PR #6732, opened 2026-06-17
Test that polling UDP socket will timeout

sadhbh-c0d3:
Theoretically `poll()` tests themselves should suffice.


sadhbh-c0d3:
I think it's not needed

PR #6739, opened 2026-06-22
feat: cronjobs CLI support

Fliqqr:
# Changes
- added CLI support for `cronjobs`
- list cronjobs for an app `wasmer cron list --app <app_id>`
- list cronjob invocations `wasmer cron invocations <cronjob_id> --app <app_id>`
- enable/disable cronjobs `wasmer cron enable/disable <cronjob_id> --app <app_id>`

Cronjobs currently only exist in association with an app, but in future standalone cronjobs are a possibility. Hence why `cron` is not a subcommand under `app`.

### DIFF
--- a/lib/backend-api/src/query.rs
+++ b/lib/backend-api/src/query.rs
@@ -415,8 +415,20 @@ pub async fn get_cron_job_invocations_page(
     let cron_job = cron_job.as_ref();
     let owner = owner.into();
     let name = name.into();
-    let start = start.map(types::DateTime::try_from).transpose()?;
-    let end = end.map(types::DateTime::try_from).transpose()?;
+    let (start, end) = match (start, end) {
+        (Some(start), Some(end)) => (start, end),
+        (Some(start), None) => (start, OffsetDateTime::now_utc()),
+        (None, Some(end)) => (end - time::Duration::days(31), end),
+        (None, None) => {
+            let end = OffsetDateTime::now_utc();
+            (end - time::Duration::days(31), end)
+        }
+    };
+    if start > end {
+        bail!("invocation start must not be after end");
+    }
+    let start = types::DateTime::try_from(start)?;
+    let end = types::DateTime::try_from(end)?;
     let mut cron_after = None;
 
     loop {
@@ -425,8 +437,8 @@ pub async fn get_cron_job_invocations_page(
             name: name.clone(),
             cron_after: cron_after.clone(),
             cron_first: Some(100),
-            invocation_start: start.clone(),
-            invocation_end: end.clone(),
+            invocation_start: Some(start.clone()),
+            invocation_end: Some(end.clone()),
             invocation_after: invocation_after.clone(),
             invocation_first,
         };
--- a/lib/backend-api/src/query.rs
+++ b/lib/backend-api/src/query.rs
@@ -550,6 +550,7 @@ pub async fn get_cron_job_invocations_page_by_id(
     ))
 }
 
+#[allow(clippy::too_many_arguments)]
 pub async fn get_cron_job_invocation_logs(
     client: &WasmerClient,
     owner: impl Into<String>,
--- a/lib/backend-api/src/query.rs
+++ b/lib/backend-api/src/query.rs
@@ -324,6 +324,25 @@ pub async fn get_app_cron_jobs(
     Ok(cron_jobs)
 }
 
+/// Enable or disable a cron job.
+pub async fn toggle_cron_job(
+    client: &WasmerClient,
+    cron_job_id: impl Into<String>,
+    enabled: bool,
+) -> Result<types::CronJob, anyhow::Error> {
+    let res = client
+        .run_graphql_strict(types::ToggleCronJob::build(types::ToggleCronJobVars {
+            cron_job_id: types::Id::from(cron_job_id),
+            enabled,
+        }))
+        .await?;
+
+    Ok(res
+        .toggle_cron_job
+        .context("backend did not return toggled cron job")?
+        .cron_job)
+}
+
 /// Retrieve invocations for a cron job. The cron job can be referenced by id or name.
 pub async fn get_cron_job_invocations(
     client: &WasmerClient,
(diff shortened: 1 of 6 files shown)


lib/backend-api/src/query.rs
```diff
@@ -288,6 +288,204 @@ pub async fn get_app_databases(
...
+    let name = name.into();
+    let mut after = None;
+    let mut cron_jobs = Vec::new();
+
+    loop {
+        let vars = types::GetAppCronJobsVars {
+            owner: owner.clone(),
+            name: name.clone(),
+            after,
+            first: Some(100),
```
marxin: I would probably introduce a `CRON_JOB_PAGE_SIZE` constant we can re-use?

lib/cli/src/commands/cron.rs
```diff
@@ -0,0 +1,320 @@
...
+    ///
+    /// Defaults to now.
+    ///
+    /// Accepts RFC 3339, RFC 2822, date, Unix timestamp, or relative time.
+    #[clap(long = "end", alias = "until", value_parser = parse_timestamp_or_relative_time_negative_offset)]
+    end: Option<OffsetDateTime>,
+
+    /// Number of invocations to fetch per page.
+    #[clap(long, default_value = "100")]
+    page_size: i32,
```
marxin: Can we use directly `NonZeroU32` type here?

lib/cli/src/commands/cron.rs
```diff
@@ -0,0 +1,320 @@
...
+    ident: AppIdentArgOpts,
+    cron_job: String,
+    enabled: bool,
+) -> Result<(), anyhow::Error> {
+    let client = env.client()?;
+    let (_ident, app) = ident.to_opts().load_app(&client).await?;
+    let cron_jobs =
+        wasmer_backend_api::query::get_app_cron_jobs(&client, &app.owner.global_name, &app.name)
+            .await?;
+    let cron_job = cron_jobs
```
marxin: A shared logic with `CmdCronShow` where we also find `cron_job` the same way.

marxin (changes requested):
LGTM with small nits. Btw. we should include an Wasmer CLI integration tests as well, either in the this repository, or in a separate one, if it's a better fit.

And the graphql schema needs to be updated - it conflicts right now.

lib/cli/src/commands/cron.rs
Fliqqr: Refactored

lib/cli/src/commands/cron.rs
Fliqqr: Done

lib/backend-api/src/query.rs
Fliqqr: Done

Fliqqr:
Added integration tests for cronjob CLI as well as support for showing invocation logs

lib/backend-api/src/query.rs
```diff
@@ -288,6 +290,428 @@ pub async fn get_app_databases(
...
+        }))
+        .await?;
+
+    res.cron_job
+        .and_then(types::NodeCronJob::into_cron_job)
+        .with_context(|| format!("cron job '{cron_job_id}' not found"))
+}
+
+/// Retrieve invocations for a cron job. The cron job can be referenced by id or name.
+pub async fn get_cron_job_invocations(
```
marxin: If I'm not mistaken, the function is unused now.

lib/backend-api/src/query.rs
```diff
@@ -288,6 +290,428 @@ pub async fn get_app_databases(
...
+            .run_graphql_strict(types::GetCronJobInvocations::build(vars))
+            .await?;
+
+        let app = res.get_deploy_app.context("app not found")?;
+        let con = app.cron_jobs;
+        let page_info = con.page_info;
+        if let Some(cron) = con
+            .nodes
+            .into_iter()
+            .find(|node| node.id.inner() == cron_job || node.name == cron_job)
```
marxin, line 411: We discussed that briefly during our Triage call: isn't this wide query and filtering at the client site expensive? Shouldn't we rather extend the back-end API, so that one can directly specify the `cron_job`?

lib/backend-api/src/query.rs
```diff
@@ -288,6 +290,428 @@ pub async fn get_app_databases(
...
+        all_invocations,
+    ))
+}
+
+/// Retrieve one page of invocations for a cron job. The cron job can be referenced by id or name.
+#[allow(clippy::too_many_arguments)]
+pub async fn get_cron_job_invocations_page(
+    client: &WasmerClient,
+    owner: impl Into<String>,
+    name: impl Into<String>,
```
marxin, line 369: Looking at the function, what is actually `name` - is it a name of the cron job, or something else?

marxin (changes requested):

lib/backend-api/src/query.rs
Fliqqr: Its the name of the `app` the cronjob belongs to. The commands support two modes of getting cronjobs, either directly by cronjob ID or by specifying cornjob name + app name/id. I can make the models a bit more clear if the naming is an issue.

lib/backend-api/src/query.rs
Fliqqr: Backend currently does not support querying invocations directly by ID as Edge does not support it. I am working on extending the Edge API to optimize the queries.

lib/backend-api/src/query.rs
Fliqqr: Good catch, surprised AI reviews did not flag it

lib/backend-api/src/query.rs
marxin: Good to know you're gonna improve it in the future - it might easily happen we'll have a cronjob with thousands of the invocations.

lib/backend-api/src/query.rs
marxin: All right - that's for the clarification.

marxin (approved):

lib/backend-api/src/query.rs
marxin: Yep! So please fix this one and we can merge it.

Fliqqr: Done

PR #6742, opened 2026-06-22
fix(wasix): return partial success from stream socket writev

Arshia001:
## Summary

- Fix stream-socket `fd_write`/`writev` so a later `send()` failure does not fail the whole syscall when earlier iovecs already transferred bytes (POSIX partial-success semantics).
- Add wasm_tests coverage for connected UDP `writev`, live stream TCP `writev`, and stream TCP partial success after peer close.
- Document the partial-success test approach and its potential flakiness in the test source.

Supersedes #6719.

## Test plan

- [x] `cargo test -p wasmer-wasix --test wasm_tests -- wasm/socket/connected-udp-writev/default/cranelift`
- [x] `cargo test -p wasmer-wasix --test wasm_tests -- wasm/socket/stream-tcp-writev/default/cranelift`
- [x] `cargo test -p wasmer-wasix --test wasm_tests -- wasm/socket/stream-tcp-writev-partial/default/cranelift`
- [x] Verified `stream-tcp-writev-partial` fails on `main` without the `fd_write` fix (`writev` returns `-1` instead of `5`)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/ci-constants.env
+++ b/.github/ci-constants.env
@@ -1,3 +1,3 @@
 # Shared CI constants. Loaded into GITHUB_ENV by workflows that need them.
 # Pinned wasix-libc sysroot (wasix-org/wasix-libc release tag).
-WASIX_LIBC_SYSROOT_TAG=v2026-06-09.1
+WASIX_LIBC_SYSROOT_TAG=v2026-06-18.1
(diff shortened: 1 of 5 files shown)


Arshia001:
I'd like to keep the test around, no matter how much copilot complains about it. If we notice it being flaky, we can just ignore it later. cc @marxin

marxin:
> I'd like to keep the test around, no matter how much copilot complains about it. If we notice it being flaky, we can just ignore it later. cc @marxin

Sure, works for me.

marxin (approved):

